PR #419, opened 2018-11-19
Resource.allocated

SystemFw:


core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -137,6 +137,70 @@ sealed abstract class Resource[F[_], A] {
    */
   def flatMap[B](f: A => Resource[F, B]): Resource[F, B] =
     Bind(this, f)
+
+  /**
```
Daenyth: 💯 This is great doc

core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -137,6 +137,70 @@ sealed abstract class Resource[F[_], A] {
...
+            case (a, rel) =>
+              stack match {
+                case Nil => F.pure(a -> F.guarantee(rel(ExitCase.Completed))(release))
+                case f0 :: xs => continue(f0(a), xs, F.guarantee(rel(ExitCase.Completed))(release))
+              }
+          } {
+            case (_, ExitCase.Completed) =>
+              F.unit
+            case ((_, release), ec) =>
+              release(ec)
```
alexandru: At a cursory look, what jumps out at me is that you're calling `release` here, irregardless of the `ExitCase`. And I thought the purpose of `allocate` is to not call release, but to save it for later.

Am I missing something?

SystemFw: it's a `bracketCase`, not `bracket`. I'm calling release if the outer `F` fails or is interrupted during the process of acquiring resources, but if it completes properly, I then return the finaliser.
The rationale is here https://github.com/typelevel/cats-effect/pull/419/files#diff-ca16b2d95f362b8bc9b6c405f858be0cR146

alexandru: Ah, sorry, this is in case it isn't `Completed`.

alexandru:
PR looks good to me. Is it done?

SystemFw:
Code wise, I think it is. I would like some more tests but writing them appears tricky, maybe we can merge and iterate after?

alexandru:
OK 🙂

PR #421, opened 2018-11-19
Fix resource attempt on suspend

mitallast:
I found some problem with Resource attempt, while testing http4s.

Minimal reproduce code:
```sh
import cats.effect._
import cats.implicits._

val suspend = Resource.suspend[IO, Int](IO.raiseError(new Exception("")))
suspend.attempt.use(_ => IO.unit).unsafeRunSync()
```
expected: successful code execution
actual:  java.io.IOException:

I think problematic code at `cats.effect.ResourceMonadError#attempt` on case `Suspend`, and this change fixes it

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Resource.scala
+++ b/core/shared/src/main/scala/cats/effect/Resource.scala
@@ -350,7 +350,10 @@ private[effect] abstract class ResourceMonadError[F[_], E] extends ResourceMonad
           })
         })
       case Suspend(resource) =>
-        Suspend(resource.map(_.attempt))
+        Suspend(resource.attempt.map {
+          case Left(error) => Resource.pure(Left(error))
+          case Right(fa: Resource[F, A]) => fa.attempt
+        })
     }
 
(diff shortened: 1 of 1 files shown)


alexandru:
Looks good. Can you please add a test too?

rossabaker:
I agree that this is less surprising behavior, and it doesn't appear to violate any laws.  A note in the scaladoc clarifying the expected behavior would be great, too.

Thanks for finding and fixing this!

mitallast:
I added the test, but not sure how best to write a note in the scaladoc - to be honest, I do not know well cats & effect.

rossabaker:
Okay. Don't worry about the doc. I will add that today or tomorrow. Thanks for the PR.

rossabaker:
The scaladoc would go on a typeclass instance, where I doubt anybody looks.  The doc would have been more compelling if `.attempt` were on the data type, or if it applied generally to `MonadError`.  I think there's little extra value in doing it here.

PR #422, opened 2018-11-22
Convert expression to Single Abstract Method

shokohara:
I tried to simplify some expressions.


alexandru:
Hey @shokohara, the problem with these conversions is that they don't work on top of Java 7 and Scala 2.11.

PR #424, opened 2018-11-24
Fix #380: Deferred, Semaphore and MVar

### LINKED ISSUE #380: Improve Semaphore.release behavior
This 
```scala
for {
  semaphore <- Semaphore[IO](0)
  _ <- {
    for {
      _ <- semaphore.acquire
      _ <- IO().foreverM
    } yield ()
  }.start
  _ <- IO(Thread.sleep(1000))
  _ <- semaphore.release
} yield ()
```
is hanging forever (in case you run it from the main method for example) because semaphore.release will add permits and complete waiting readers in a synchronous way.

(we discovered this while debugging flaky tests. if we remove from given example  `IO(Thread.sleep(1000))`, there is a race between `semaphore.acquire` and `semaphore.release`. This is even harder to detect and quite error-prone.)

In our application, I am trying to build some consumer-like loop based on `Semaphore.aquire` and `Monad.iterateWhile`, `Concurrent` is required for `Semaphore` construction and the logic of synchronous completion is quite unexpected.

My proposal is :
1. make `cats.effect.concurrent.Semaphore.AbstractSemaphore#open` abstract
2. add default implementation to `AsyncSemaphore`
3. add implementation with wrapping in `Concurrent.start` to `ConcurrentSemaphore`.

As a result, releasing readers will be asynchronous and should handle similar situations.

WDYT?

P.S. Thank you for the amazing lib!

alexandru:
Fixes #380.

TL;DR about the issue:

1. when we have an `acquire` and `release` pair, like we have in `Deferred` (get / complete), `Semaphore` (acquire / release) and `MVar` (take / put), all of them being similar, the `release` part can block indefinitely in case the bind continuation of `acquire` is a never ending loop
2. the fix would be for `acquire` to always fork after an await, however we do not have `shift` available because we don't have a `ContextShift` restriction specified
3. `F.unit.start.flatMap(_.join)` is NOT a replacement for `shift` 

The solution is, when notifying consumers that are waiting for a value, to `start` and then forget about the result, something like this:

```scala
def notifyReaders(a: A, r: Iterable[A => Unit]): F[Unit] = {
  r.foldLeft(F.unit) { (acc, cb) => 
    acc.flatMap { _ =>
      F.delay(cb(a)).start.map(_ => ())
    }
  }
}
```

This is the only reasonable solution for the `Concurrent`-enabled implementations. Some notes:

- ordering callback calls is wrong, because a single callback that blocks will block all the rest, just like it blocks the thread calling `release` (this is related to a remark by @SystemFw that `Semaphore` relies on listeners waking up in order and I really hope you only rely on the priority of the listeners and not on how the actual callbacks are being called, because that would be very wrong)
- we really need `startAndForget` in the type class hierarchy (Cats-Effect 2.0) ... unfortunately we cannot add it for Scala 2.11 without breaking binary compatibility, but maybe we can find alternative ways until then
- I believe we need a `suspendWithContext[A](f: ExecutionContext => F[A]): F[A]` in `ContextShift`; this too breaks binary compatibility, plus it involves some work in IO's run-loop, but that part is I believe doable without binary compat changes

---

Unfortunately in fixing `MVar` I needed this function and it made sense to expose it:

```scala
object Concurrent {
  // ...

  def cancelableF[F[_], A](k: (Either[Throwable, A] => Unit) => F[CancelToken[F]])
    (implicit F: Concurrent[F]): F[A] = {

    F.asyncF { cb =>
      val state = new AtomicReference[Either[Throwable, Unit] => Unit](null)
      val cb1 = (a: Either[Throwable, A]) => {
        try {
          cb(a)
        } finally {
          // This CAS can only succeed in case the operation is already finished
          // and no cancellation token was installed yet
          if (!state.compareAndSet(null, Callback.dummy1)) {
            val cb2 = state.get()
            state.lazySet(null)
            cb2(Callback.rightUnit)
          }
        }
      }
      // Until we've got a cancellation token, the task needs to be evaluated
      // uninterruptedly, otherwise risking a leak, hence the bracket
      F.bracketCase(k(cb1)) { _ =>
        F.async[Unit] { cb =>
          if (!state.compareAndSet(null, cb)) {
            cb(Callback.rightUnit)
          }
        }
      } { (token, e) =>
        e match {
          case ExitCase.Canceled => token
          case _ => F.unit
        }
      }
    }
  }  
}
```


alexandru:
I broke `MVar` apparently. Have to fix it.

core/shared/src/main/scala/cats/effect/concurrent/Deferred.scala
```diff
@@ -179,25 +180,42 @@ object Deferred {
...
+          }
+      }
 
-      F.delay(loop())
+    private def notifyReadersLoop(a: A, r: Iterable[A => Unit]): F[Unit] = {
+      var acc = F.unit
+      val cursor = r.iterator
+      while (cursor.hasNext) {
+        val next = cursor.next()
+        val task = F.map(F.start(F.delay(next(a))))(mapUnit)
```
mpilquist: `F.void` instead of `F.map(..)(mapUnit)`?

core/shared/src/main/scala/cats/effect/Concurrent.scala
```diff
@@ -388,6 +388,80 @@ object Concurrent {
   def timeout[F[_], A](fa: F[A], duration: FiniteDuration)(implicit F: Concurrent[F], timer: Timer[F]): F[A] =
     timeoutTo(fa, duration, F.raiseError(new TimeoutException(duration.toString)))
 
+  /**
+   * Function that creates an async and cancelable `F[A]`, similar with
```
mpilquist: What guarantees are provided w.r.t. cancelation of the returned action?

core/shared/src/main/scala/cats/effect/concurrent/Deferred.scala
alexandru: `F.void` is more wasteful, because it translates to `F.map(task)(_ => ())` and the Scala compiler won't reuse the created `_ => ()`.

When doing performance-sensitive work, in general I care about these issues in about this order:

1. avoiding extraneous async or lazy boundaries
2. avoiding extraneous `compareAndSet` operations
3. avoiding extraneous memory allocations

Caching dummy functions like `_ => ()` falls into point 3, and sometimes I screw up due to the code becoming unclear, but it doesn't happen that often 🙂

core/shared/src/main/scala/cats/effect/Concurrent.scala
alexandru: This is a very good question.

I realized that the task generated by the provided generator function should be evaluated uninterruptedly, otherwise it can leak. 

I fixed it and added a ScalaDoc comment about it.

PR #437, opened 2018-12-03
Add a `map` method directly onto `Resource`

andimiller:
This enables it to be found by IDEs which have issues with partial
unification and feels like it should be there since the class already
provides `flatMap`.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Resource.scala
+++ b/core/shared/src/main/scala/cats/effect/Resource.scala
@@ -139,11 +139,10 @@ sealed abstract class Resource[F[_], A] {
     Bind(this, f)
 
   /**
-    * Direct implementation of the `map` operation which is otherwise
-    * available via the `cats.Monad` instance for `Resource[F, ?]`
+    *  Given a mapping function, transforms the resource provided by
+    *  this Resource.
     *
-    * This allows IDEs which have issues with Partial Unification to
-    * see `map`.
+    *  This is the standard `Functor.map`.
     */
   def map[B](f: A => B)(implicit F: Applicative[F]): Resource[F, B] =
     flatMap(a => Resource.pure[F, B](f(a)))
--- a/core/shared/src/main/scala/cats/effect/Resource.scala
+++ b/core/shared/src/main/scala/cats/effect/Resource.scala
@@ -146,7 +146,7 @@ sealed abstract class Resource[F[_], A] {
     * see `map`.
     */
   def map[B](f: A => B)(implicit F: Monad[F]): Resource[F, B] =
-    flatMap(a => Resource.applyCase(F.pure((f(a), _ => F.unit))))
+    flatMap(a => Resource.pure[F, B](f(a)))
 
   /**
     * Given a `Resource`, possibly built by composing multiple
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -138,6 +138,16 @@ sealed abstract class Resource[F[_], A] {
...
 
+  /**
+    * Direct implementation of the `map` operation which is otherwise
+    * available via the `cats.Monad` instance for `Resource[F, ?]`
+    *
+    * This allows IDEs which have issues with Partial Unification to
+    * see `map`.
+    */
+  def map[B](f: A => B)(implicit F: Monad[F]): Resource[F, B] =
+    flatMap(a => Resource.applyCase(F.pure((f(a), _ => F.unit))))
```
SystemFw: I think you can just use `Resource.pure` instead of the manual apply case (or conversely just build the tree using `Bind` and `Allocate`). In any case the constraint on `F`can probably be weakened to `Applicative`

andimiller: Good catch, I got slightly hung up on following how `map` was implemented via `Monad` and ended up using `applyCase`, I've simplified it to using `pure` and weakened the `F` to `Applicative`

alexandru:
Speaking of this, somebody should also feel a ticket with JetBrains for IntelliJ IDEA, maybe somebody ends up fixing their support.

It's been worsening lately, not sure what they are doing, probably rebuilding their engine or something.

laws/shared/src/test/scala/cats/effect/ResourceTests.scala
```diff
@@ -87,6 +87,13 @@ class ResourceTests extends BaseTestsSuite {
       Resource.liftF(fa).use(IO.pure) <-> fa
     }
   }
+  testAsync("map should be the same as going via Monad") { implicit ec =>
```
alexandru: Instead of adding this test, a better idea would be to override `map` in the provided `Monad` instance.

core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -138,6 +138,16 @@ sealed abstract class Resource[F[_], A] {
   def flatMap[B](f: A => Resource[F, B]): Resource[F, B] =
     Bind(this, f)
 
+  /**
+    * Direct implementation of the `map` operation which is otherwise
+    * available via the `cats.Monad` instance for `Resource[F, ?]`
+    *
+    * This allows IDEs which have issues with Partial Unification to
+    * see `map`.
```
alexandru: The reason for why this `map` is provided shouldn't be exposed in ScalaDoc.

Just do something generic like ...

```
Given a mapping function, transforms the source.

This is the standard `Functor.map`.
```

alexandru: But I think we need another ScalaDoc comment, on `Resource` class, mentioning:

1. `-Ypartial-unification`
2. `import cats.implicits._`

... to unlock all available operations.

PR #439, opened 2018-12-05
Add a semiflatMap method to Resource

### LINKED ISSUE #391: Enable effectful transformations inside Resource
Something like this, similar to `evalMap` from fs2:

```scala
//given
val a: Resource[F, A]

//and
val ab: A => F[B]

val foo: Resource[F, B] = a.modifyF(ab)
```

ksonj:
This adds `semiflatMap` (name taken from a similar method on [`EitherT`](https://github.com/typelevel/cats/blob/e1a7cfcddce0fd48403669a5744df50e4a08631c/core/src/main/scala/cats/data/EitherT.scala#L112)) to `Resource` to directly apply effectful transformations to the allocated resource.

Closes #391

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Resource.scala
+++ b/core/shared/src/main/scala/cats/effect/Resource.scala
@@ -201,6 +201,13 @@ sealed abstract class Resource[F[_], A] {
         (a.asInstanceOf[A], release)
     }
   }
+
+  /**
+    * Applies an effectful transformation to the allocated resource. Like a
+    * `flatMap` on `F[A]` while maintaining the resource context
+    */
+  def semiflatMap[B](f: A => F[B])(implicit F: Applicative[F]): Resource[F, B] =
+    this.flatMap(a => Resource.liftF(f(a)))
 }
 
 object Resource extends ResourceInstances {
--- a/laws/shared/src/test/scala/cats/effect/ResourceTests.scala
+++ b/laws/shared/src/test/scala/cats/effect/ResourceTests.scala
@@ -88,6 +88,12 @@ class ResourceTests extends BaseTestsSuite {
     }
   }
 
+  testAsync("semiflatMap") { implicit ec =>
+    check { fa: IO[String] =>
+      Resource.liftF(fa).semiflatMap(IO.pure).use(IO.pure) <-> fa
+    }
+  }
+
   testAsync("allocated produces the same value as the resource") { implicit ec =>
     check { resource: Resource[IO, Int] =>
       val a0 = Resource(resource.allocated).use(IO.pure).attempt
@@ -111,7 +117,7 @@ class ResourceTests extends BaseTestsSuite {
 
     prog.unsafeRunSync
   }
-  
+
   test("safe attempt suspended resource") {
     val exception = new Exception("boom!")
     val suspend = Resource.suspend[IO, Int](IO.raiseError(exception))


laws/shared/src/test/scala/cats/effect/ResourceTests.scala
```diff
@@ -88,6 +88,12 @@ class ResourceTests extends BaseTestsSuite {
     }
   }
 
+  testAsync("semiflatMap") { implicit ec =>
```
kubukoz: As mentioned on gitter, it'd be nice to have tests that check what happens in case of errors and actual effects in `f`

ksonj:
@kubukoz I've added a test case for cancellation in `f` and I've adapted the former test to include an actual effect. Now, I've tried adding another one where the effect throws an error, but that test fails and I suspect I'm just overlooking something. This is what I had:

```scala
  testAsync("(semiflatMap with error <-> IO.never") { implicit ec =>
    case object Foo extends Exception
    implicit val cs = ec.contextShift[IO]

    check { (g: Int => IO[Int]) =>
      val effect: Int => IO[Int] = a => (g(a) <* IO(throw Foo))
      Resource.liftF(IO(0)).semiflatMap(effect).use(IO.pure) <-> IO.never
    }
  }
```

kubukoz:
Wouldn't a failure in the effect result in IO.raiseError(Foo) instead of IO.never?

core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -201,6 +201,13 @@ sealed abstract class Resource[F[_], A] {
         (a.asInstanceOf[A], release)
     }
   }
+
+  /**
+    * Applies an effectful transformation to the allocated resource. Like a
+    * `flatMap` on `F[A]` while maintaining the resource context
+    */
+  def semiflatMap[B](f: A => F[B])(implicit F: Applicative[F]): Resource[F, B] =
+    this.flatMap(a => Resource.liftF(f(a)))
```
alexandru: Is an alias like this really useful?

I mean it is convenient, but it introduces yet another name that people have to learn.

For example this `semiflatMap` in Monix's `Iterant` is called `mapEval` and I'm pretty sure it has another name in FS2. And a divergence in naming is going to be more problematic than the absence of this helper.

So what precedent do we have for this name?

ksonj: The precedent is [this](https://github.com/typelevel/cats/blob/e1a7cfcddce0fd48403669a5744df50e4a08631c/core/src/main/scala/cats/data/EitherT.scala#L112) method on `EitherT` in cats. But to be honest I lilke `evalMap` better. Isn't that also what it's called on `fs2.Stream`?

As to whether it's useful, I do think so. This seems like a very common usecase and exposing it directly makes it a lot easier to discover, especially for beginners. Browsing the cats scaladoc sort of is my hoogle substitute and I missed `liftF` scanning through.

Then again, a section in the Resource docs mentioning `liftF` would probably work too. If that is preferred I'll PR that instead.

ksonj:
@kubukoz You're right, thanks! I wasn't aware of any difference there

core/shared/src/main/scala/cats/effect/Resource.scala
alexandru: We can add this too, we just have to agree about the naming.

ksonj: Oh, I didn't realize the difference `mapEval` vs `evalMap`. `Stream.evalMap` seems to not exactly be the same because the effect type is allowed to vary. `mapEval` still seems nicer than `semiflatMap`, though. But that's your call. I'll update this PR with an appropriate addition to the docs next time I find some time.

alexandru: I would be fine with either, pick one.

ksonj: That is a good argument for `semiflatMap`. I'll leave it then, because I think consistency is more important than an intuitive name. I've also updated the docs with examples on `liftF` and `semiflatMap` now

alexandru: I'm not very familiar with Cats' monad transformers. If @rossabaker says `semiflatMap` is OK, then I guess it's fine.

ksonj: I'm still struggling a little with that name (and others seem to as well). Other suggestions from the gitter channel were: `flatMapF` and `liftMap`. At least `flatMapF` seems to have some precedent on other monad transformers, e.g. `StateT`, `IdT` and `Kleisli`, and while maybe not perfect, it seems a bit nicer than `semiflatMap`. Especially since the name `semiflatMap` seems to have only been a compromise because `flatMapF` was already taken on `OptionT` and `EitherT` (https://github.com/typelevel/cats/pull/999) Thus the precedent for `flatMapF` is older. I'm updating accordingly

alexandru: 👍 for `flatMapF`, personally I like it.

rossabaker: :-1: `flatMapF` is a mess in cats, whose shape is determined by historical quirks

Preference:
1. `semiflatMap`: consistent upstream where it appears
2. `evalMap` or `mapEval`: used downstream, intuitive. If fs2 and monix had picked the same name, it would be my first choice.
4. `flatMapF`: I can't ever remember what this does on which type

This is a preference, not a veto.  If I'm outvoted, I won't pout. :+1: to the rest of the ticket.

ksonj: :slightly_smiling_face: I can live with either. What's needed to reach consensus though? Anything I could do?

One other thing that imo is a good argument for `flatMapF` is that it would be somewhat consistent with `liftF`, which already exists on `Resource`.

rossabaker: I'll mention the issue on the channel again, and if nobody else comes to argue against `flatMapF` in the next day or so, I'll admit defeat. :smile:

ChristopherDavenport: What is flatMapF? I think semiFlatMap makes more sense as well. Putting a letter F tells me nothing about the context. SemiFlatMap, tells my you are lifting into the transformers inner context. That increased clarity seems preferable to me as well.

SystemFw: I vote for `evalMap`

alexandru:
OMG the bike shedding 😄

I don't mind either of the alternatives TBH. Whatever you guys prefer. `semiflatMap`, `flatMapF`, at this point whatever is fine by me.

@rossabaker in Monix I went with `mapEval` instead of `evalMap` for 2 reasons:

1. if we are follow the `flatMap` naming convention (which isn't `mapFlat`), that one actually comes from function composition, so it's `flatten . map`, however the order in function composition notation is imo very confusing, plus we are not really composing two functions in `mapEval`
2. in Monix's `Observable` I've got multiple functions suffixed with `Eval`, e.g. `scanEval` vs `scan`, etc
  - I prefer suffixes in general because suffixes are friendlier with IDEs, even if nowadays IntelliJ IDEA is smart enough to suggest alternatives with arbitrary prefixes too

SystemFw:
So, this is just bike shedding so don't expect a strong argument, here's mine:

- Consistency is not an argument: we have precedent for like 6 different names
- Descriptive names is also not really an argument imho: this essentially an arbitrary transformation, best described by its type that by anything else. I don't think  you can really make a strong argument for either`semiFlatMap`, `mapEval` , `evalMap` or `flatMapF`.

The only argument I see is familiarity, and I think `evalMap` is the op with that shape which is used the most by users who are also likely to be using `Resource` (e.g. because they use fs2, or doobie or http4s that also use fs2).

eli-jordan:
+1 for `evalMap`

alexandru:
To add some fuel on fire ...

TBH `evalMap` / `mapEval` works more in the context of streaming, because it's very much like a `map`.

For `Resource` this more like a `flatMap` really.

So I'm going with the `semiflatMap` choice, seeing that `flatMapF` is less popular.

SystemFw:
> evalMap / mapEval works more in the context of streaming, because it's very much like a map.

I don't see how, `evalMap` in fs2 is literally `.flatMap( .. Stream.eval)` i.e exactly the same if you consider `Stream.eval` == `Resource.liftF`. If anything, we could rename `liftF` to `eval` in `Resource` :P

alexandru:
`map(f)` too is `flatMap(x => pure(f(x)))` and it isn't named `pureMap` 😛

SystemFw:
> map(f) too is flatMap(x => pure(f(x))) and it isn't named pureMap

True, but there we can deploy the consistency and formal name argument, which we can't with this one. But this is why `evalMap` to me is preferable because of familiarity, not any "deep" reasons.

alexandru:
Guys, @ksonj, just pick one.

oleg-py:
+1 for `evalMap` :)

ksonj:
I'll go with `evalMap` then, as everyone seems to be least opposed to that :smile:

PR #440, opened 2018-12-06
Add Resource#map

kubukoz:
This will make it easier for IntelliJ users to work with resources (both in for comprehensions and in direct `map` calls), as well as people who don't want to import syntax just to use `map` :)


kubukoz:
Duplicate of #437

PR #448, opened 2018-12-14
Change .html to .md to correctly point at the IO documentation

kschlorff16:
Was looking through the docs for the first time and found a broken link in `site/src/main/tut/tutorial/tutorial.md`. This fixes that! 

Also, 👋


gvolpe:
Hi @kschlorff16 , that's not a broken link. It is actually correct. Did you check here? https://typelevel.org/cats-effect/tutorial/tutorial.html

rossabaker:
Right, the links are for the generated site, not the Github view.  I wish they worked in both.

PR #449, opened 2018-12-15
Avoid Throwable#addSuppressed, use CompositeException

dwijnand:
I would say (without backing data) that it's fairly common in the
ecosystem to sometimes use 'case object' when defining a throwable.
This makes addSuppressed a memory leak risk.  There is the option to
disable suppressed exceptions, which makes addSuppressed a noop, but
that just makes it swallow errors then...

So for now I think it might be safest to just use CompositeException
(like we were already on JS).  Though I do share Ross' concern about
root causes...

Thoughts welcome.  There's also a bunch of thoughts by a handful of
people over in https://twitter.com/dwijnand/status/1073600688058458112.

/cc @rossabaker @alexandru
This undoes #255

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -190,22 +190,8 @@ val mimaSettings = Seq(
     import com.typesafe.tools.mima.core._
     import com.typesafe.tools.mima.core.ProblemFilters._
     Seq(
-      // All internals — https://github.com/typelevel/cats-effect/pull/376
-      exclude[DirectMissingMethodProblem]("cats.effect.internals.IOBracket#BracketStart.this"),
-      exclude[MissingClassProblem]("cats.effect.internals.ForwardCancelable$State$IsEmptyCanceled"),
-      exclude[MissingClassProblem]("cats.effect.internals.ForwardCancelable$State$IsEmptyCanceled$"),
-      exclude[MissingClassProblem]("cats.effect.internals.ForwardCancelable$State$Reference"),
-      exclude[MissingClassProblem]("cats.effect.internals.ForwardCancelable$State$IsEmpty$"),
-      exclude[DirectMissingMethodProblem]("cats.effect.internals.ForwardCancelable.:="),
-      exclude[DirectMissingMethodProblem]("cats.effect.internals.ForwardCancelable.this"),
-      exclude[DirectMissingMethodProblem]("cats.effect.internals.ForwardCancelable.plusOne"),
-      exclude[MissingClassProblem]("cats.effect.internals.ForwardCancelable$State$"),
-      exclude[MissingClassProblem]("cats.effect.internals.ForwardCancelable$State$Reference$"),
-      exclude[MissingClassProblem]("cats.effect.internals.ForwardCancelable$State$IsCanceled$"),
-      // https://github.com/typelevel/cats-effect/pull/377
-      exclude[DirectMissingMethodProblem]("cats.effect.internals.IOBracket#EnsureReleaseFrame.this"),
-      exclude[DirectMissingMethodProblem]("cats.effect.internals.IOBracket#BracketReleaseFrame.this"),
-      exclude[DirectMissingMethodProblem]("cats.effect.internals.IOBracket#BaseReleaseFrame.this"),
+      // Ignore any binary compatibility issues/problems that match the internals package
+      exclude[Problem]("cats.effect.internals.*"),
       // All internals - https://github.com/typelevel/cats-effect/pull/403
       exclude[DirectMissingMethodProblem]("cats.effect.concurrent.Semaphore#AbstractSemaphore.awaitGate"),
       exclude[DirectMissingMethodProblem]("cats.effect.concurrent.Semaphore#AsyncSemaphore.awaitGate"),
--- a/core/js/src/main/scala/cats/effect/internals/IOPlatform.scala
+++ b/core/js/src/main/scala/cats/effect/internals/IOPlatform.scala
@@ -16,9 +16,7 @@
 
 package cats.effect.internals
 
-import cats.data.NonEmptyList
 import cats.effect.IO
-import cats.effect.util.CompositeException
 
 import scala.concurrent.duration.Duration
 
@@ -75,23 +73,6 @@ private[effect] object IOPlatform {
         case Left(e) => Logger.reportFailure(e)
         case _ => ()
       }
-    }
-  }
-
-  /**
-   * Composes multiple errors together, meant for those cases in which
-   * error suppression, due to a second error being triggered, is not
-   * acceptable.
-   *
-   * On top of the JVM this function uses `Throwable#addSuppressed`,
-   * available since Java 7. On top of JavaScript the function would return
-   * a `CompositeException`.
-   */
-  def composeErrors(first: Throwable, rest: Throwable*): Throwable = {
-    rest.filter(_ != first).toList match {
-      case Nil => first
-      case nonEmpty =>
-        new CompositeException(first, NonEmptyList.fromListUnsafe(nonEmpty))
     }
   }
 }
--- a/core/jvm/src/main/scala/cats/effect/internals/IOPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/internals/IOPlatform.scala
@@ -107,18 +107,4 @@ private[effect] object IOPlatform {
    * `false` if it's JavaScript.
    */
   final val isJVM = true
-
-  /**
-   * Composes multiple errors together, meant for those cases in which
-   * error suppression, due to a second error being triggered, is not
-   * acceptable.
-   *
-   * On top of the JVM this function uses `Throwable#addSuppressed`,
-   * available since Java 7. On top of JavaScript the function would return
-   * a `CompositeException`.
-   */
-  def composeErrors(first: Throwable, rest: Throwable*): Throwable = {
-    for (e <- rest; if e != first) first.addSuppressed(e)
-    first
-  }
 }
--- a/core/shared/src/main/scala/cats/effect/internals/CancelUtils.scala
+++ b/core/shared/src/main/scala/cats/effect/internals/CancelUtils.scala
@@ -57,7 +57,7 @@ private[effect] object CancelUtils {
           case Nil =>
             IO.unit
           case first :: rest =>
-            IO.raiseError(IOPlatform.composeErrors(first, rest: _*))
+            IO.raiseError(IOFrame.composeErrors(first, rest: _*))
         }
       }
     }
--- a/core/shared/src/main/scala/cats/effect/internals/IOBracket.scala
+++ b/core/shared/src/main/scala/cats/effect/internals/IOBracket.scala
@@ -170,7 +170,7 @@ private[effect] object IOBracket {
     extends IOFrame[Unit, IO[Nothing]] {
 
     def recover(e2: Throwable): IO[Nothing] =
-      IO.raiseError(IOPlatform.composeErrors(e, e2))
+      IO.raiseError(IOFrame.composeErrors(e, e2))
 
     def apply(a: Unit): IO[Nothing] =
       IO.raiseError(e)
--- a/core/shared/src/main/scala/cats/effect/internals/IOFrame.scala
+++ b/core/shared/src/main/scala/cats/effect/internals/IOFrame.scala
@@ -16,7 +16,9 @@
 
 package cats.effect.internals
 
+import cats.data.NonEmptyList
 import cats.effect.IO
+import cats.effect.util.CompositeException
 
 /** A mapping function that is also able to handle errors,
   * being the equivalent of:
@@ -66,5 +68,21 @@ private[effect] object IOFrame {
 
     def apply(a: A): IO[B] = fs(a)
     def recover(e: Throwable): IO[B] = fe(e)
+  }
+
+  /**
+   * Composes multiple errors together, meant for those cases in which
+   * error suppression, due to a second error being triggered, is not
+   * acceptable.
+   *
+   * To avoid leaking memory (such as into a case object throwable)
+   * this function creates and returns a `CompositeException`.
+   */
+  def composeErrors(first: Throwable, rest: Throwable*): Throwable = {
+    rest.filter(_ != first).toList match {
+      case Nil => first
+      case nonEmpty =>
+        new CompositeException(first, NonEmptyList.fromListUnsafe(nonEmpty))
+    }
   }
 }
--- a/core/shared/src/main/scala/cats/effect/internals/IOParMap.scala
+++ b/core/shared/src/main/scala/cats/effect/internals/IOParMap.scala
@@ -110,7 +110,7 @@ private[effect] object IOParMap {
           other.cancel.unsafeRunAsync { r =>
             conn.pop()
             cb.async(Left(r match {
-              case Left(e2) => IOPlatform.composeErrors(e, e2)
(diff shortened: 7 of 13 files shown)


dwijnand:
Oh I forgot to say, I moved `composeErrors` to `IOFrame` as it seemed related (and shared across platforms), but let me know where it should live instead.

dwijnand:
Also, I tried to write a test the guard against it becoming a memory risk again, but I struggled to get something stable (and it's probably related to how much memory I'm feeding my sbt).

jdegoes:
@dwijnand Thank you for this!!! Partially addresses a big pet peeve of mine.

Now there's another problem that should also be addressed.

If a cascade of errors is going to be considered recoverable, then a user should have a guarantee that they can catch the type of one of the errors that they threw.

Without this guarantee, you would write code like this:

```scala
IO.raiseError(e1).guarantee(io2).attempt {
  case Left(`e1`) => handleError
  case Right(_) => handleSuccess
}
```

and the code will fail, with another exception, because the exception thrown by `IO.raiseError` is not the same as the exception caught by `attempt`.

So I think you have your choice:

 * You can choose to make composite errors non-recoverable, so the user cannot catch anything with appropriate logging / reporting of the error (ZIO);
 * OR, you can choose to discard / log / report the secondary error, and allow the user to catch the primary error, without knowledge of the secondary error.

But in no case should a user throw `E1` and `E2` and be able to catch some error which is neither `E1` nor `E2`. Indeed, this violates monad error laws (like `addSuppressed`), which is why the instance for `Throwable` equality in Cats-Effect merely returns `true`—the monad error laws are not being passed, rather the errors are not being compared for structural equality.

dwijnand:
> But in no case should a user throw `E1` and `E2` and be able to catch some error which is neither `E1` nor `E2`. Indeed, this violates monad error laws (like `addSuppressed`), which is why the instance for `Throwable` equality in Cats-Effect merely returns `true`—the monad error laws are not being passed, rather the errors are not being compared for structural equality.

I don't know the monad error laws (or monad error itself) but I found the instance for Throwable equality and an interesting comment:

```
      def eqv(x: Throwable, y: Throwable): Boolean = {
        // All exceptions are non-terminating and given exceptions
        // aren't values (being mutable, they implement reference
        // equality), then we can't really test them reliably,
        // especially due to race conditions or outside logic
        // that wraps them (e.g. ExecutionException)
        (x ne null) == (y ne null)
      }
```

I get the mutable part and the wrapping part, but not the race condition part.  But I'm mostly stuck on what's meant by the "non-terminating part".  What's meant there??

Going back to what you said earlier:

> If a cascade of errors is going to be considered recoverable, then a user should have a guarantee that they can catch the type of one of the errors that they threw.

Where does this guarantee come from?

Assuming it's a use case we care to provide, perhaps plain case matching isn't the API for it.  Maybe there should be some kind of API that accounts for: throwable root causes, suppressed exceptions, and `CompositeException`.

Maybe it should do something for special cases like `ExecutionException`, like the compiler's internal [`Exceptional`](https://github.com/scala/scala/blob/v2.12.8/src/compiler/scala/tools/nsc/util/Exceptional.scala) does.

jdegoes:
The comment is a justification for the current design. In my opinion, the instance should be replaced by one that does reference and structural equality. The library should _never tamper in any way with a value provided by the user_ —including calling mutable methods on that value or wrapping it in other types.

You can find Cats [MonadError Laws here](https://github.com/typelevel/cats/blob/master/laws/src/main/scala/cats/laws/MonadErrorLaws.scala).

One of the most surprising things to users about Future's API is that sometimes you throw an error and you can't catch it—you can only catch that error wrapped in `ExecutionException`. It's even worse if that (incorrect) behavior is hard to predict, e.g. it happens only on some platforms (Scala.js) or only in some runtime edge cases (finalizer throws an error), as it does for Cats IO.

> Assuming it's a use case we care to provide, perhaps plain case matching isn't the API for it. Maybe there should be some kind of API that accounts for: throwable root causes, suppressed exceptions, and CompositeException.

I think fixing the actual problem (the error model) will be too much work for this PR and too controversial, because different people will have different ideas about what needs to be done (ZIO, for example, preserves all errors losslessly, never swallows errors, and passes all monad error laws).

I think the best we could hope for here is:

 - Consistent behavior across Scala.js and JVM
 - Not (dangerously) mutating values passed in from the user
 - Not changing the type of exceptions thrown by user-land code

This will be a small change but have large positive effects on consistency and predictability of the API.

dwijnand:
I see, so fully reverting #255 so recovering does

```scala
Logger.reportFailure(e2)
IO.raiseError(e)
```

I'm happy to make that change, obviously, but I'd probably be best served waiting on what the cats effect maintainers think first.

alexandru:
`Throwable` objects aren't values, first of all because they contain a mutable stack-trace (see [setStackTrace](https://docs.oracle.com/javase/7/docs/api/java/lang/Throwable.html#setStackTrace(java.lang.StackTraceElement[]))).

Doing `object MyStuff extends Throwable` for that reason alone is problematic ... people like to pretend that these are values, but they are not values. I am doing that myself, but I'm doing so while being fully aware of the perils. 

Basically `object MyStuff extends Throwable` is only useful as an optimization (for code in which the error is actually used for flow control) and only for well encapsulated code, otherwise _it is an anti-pattern_.

In the `cats.effect.IO` the change would be easy, you'd just have to modify `Platform.composeErrors` to return a `CompositeException` instead of using `addSuppressed`, however:

1. `addSupressed` is now the de-facto standard on top of the JVM for composing errors, since Java 7, due to Java's [try-with-resources statement](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) which is very similar with our `bracket`
2. the reason for why it was introduced is because people are usually interested in seeing the first error that happened, with its full stack trace, but at the same time having the ability to also fetch the suppressed errors

Personally I don't care for the `object Err extends Throwable` use case. People should simply not do that, because this use case is incompatible with how Throwables are meant to be used on top of the JVM.

But if you think this is an important use case (although in my opinion the libraries / projects doing that should simply fix their code), then we can revert to `CompositeException`.

alexandru:
@dwijnand 

> I get the mutable part and the wrapping part, but not the race condition part. But I'm mostly stuck on what's meant by the "non-terminating part". What's meant there??

Code throwing exceptions is non-terminating. This is reflected by the type system via the `Nothing` type, which is Scala's bottom. And `IO.raiseError` is the same, since for any `A`, you have `IO[Nothing] <: IO[A]`, naturally.

Also `Throwable` can't have a good `Eq` instance.

A good way to understand why that `Eq[Throwable]` instance is the way it is would be to clone the project, change that instance into something that tries to do normal equality check and then notice and learn from the tests that fail.

And actually much better would be to do that in Monix, since we have a lot more tests there related to error handling, in the context of streaming too.

build.sbt
```diff
@@ -190,22 +190,8 @@ val mimaSettings = Seq(
...
-      exclude[DirectMissingMethodProblem]("cats.effect.internals.ForwardCancelable.plusOne"),
-      exclude[MissingClassProblem]("cats.effect.internals.ForwardCancelable$State$"),
-      exclude[MissingClassProblem]("cats.effect.internals.ForwardCancelable$State$Reference$"),
-      exclude[MissingClassProblem]("cats.effect.internals.ForwardCancelable$State$IsCanceled$"),
-      // https://github.com/typelevel/cats-effect/pull/377
-      exclude[DirectMissingMethodProblem]("cats.effect.internals.IOBracket#EnsureReleaseFrame.this"),
-      exclude[DirectMissingMethodProblem]("cats.effect.internals.IOBracket#BracketReleaseFrame.this"),
-      exclude[DirectMissingMethodProblem]("cats.effect.internals.IOBracket#BaseReleaseFrame.this"),
+      // Ignore any binary compatibility issues/problems that match the internals package
+      exclude[Problem]("cats.effect.internals.*"),
```
alexandru: 😀 OMG, I've been living in the dark

alexandru: Speaking of this, it would have been awesome to have an SBT setup that checks nothing is exposed from the `internals` package.

My workflow of choice for that is to generate the ScalaDoc and the search for "internal" in it 🙂

jdegoes:
@alexandru `Throwable` are values, just like `Array` and many other values. They are _mutable_ values, but that does not give an excuse for a _purely functional library_ to mutate those values.

Just because _Java_ does something, doesn't mean that a purely functional library should do that. After all, if we wanted to follow "Java conventions", we would not be using values to model failures; nor effect types like `IO`, for that matter.

In my view, the problems that must be fixed are as follows:

1. Cats IO must stop _dangerously mutating_ values passed to it by user-land code.
2. Cats IO must stop _dangerously altering_ exception types by wrapping them in other exceptions.

This means no `addSuppressed` and definitely no `CompositeException`.

_A user should have the guarantee they can catch the same exception type that they throw._ Moreover, we should have the additional guarantee the same exception handling code will work on both Scala.js and JVM. With the above changes, we can have these guarantees and come closer to passing (instead of ignoring) monad error laws.

alexandru:
@jdegoes 

> @alexandru Throwable are values, just like Array and many other values. They are mutable values, but that does not give an excuse for a purely functional library to mutate those values.

`Throwable` and `Array` are mutable objects (objects with identity) and in my vocabulary mutable objects aren't values, just like how impure functions aren't functions.

Mutable objects are buckets. You can wrap values in them and `Throwable` works great as a wrapper.

> Just because Java does something

The point is that your typical Scala app will run a lot of Java code, as the stack gets thicker and thicker as the app grows, and will mutate `Throwable` objects, whether we like it or note, therefore `object Err extends Throwable` is always dangerous, irregardless of what the `IO` implementation does.

> In my view, the problems that must be fixed are as follows:
>
>   - Cats IO must stop dangerously mutating values passed to it by user-land code.
>   - Cats IO must stop dangerously altering exception types by wrapping them in other exceptions.

If you want to recover from errors, those are the only two options and it's a widely used practice in the industry. The third option, that ZIO implements, is to have a callback that can be called for thrown exceptions (that interrupt your fibers), however that's not `MonadError`.

Note that ZIO:

1. does not implement `MonadError[ZIO, Throwable]`, like `cats.effect.IO` does, so you're basically talking of apples versus oranges
2. cannot recover from exceptions thrown in `bracket`'s `release` via `MonadError`, whereas `cats.effect.IO` can

In other words you cannot do:

```scala
IO(throw exception).handleErrorWith { ex => recover(ex) }
```

Also the `addSuppressed` logic has nothing to do with `MonadError`, the `composeErrors` functionality being used only for `bracket`.

> With the above changes, we can have these guarantees and come closer to passing (instead of ignoring) monad error laws.

The [MonadError laws](https://github.com/typelevel/cats/blob/v1.5.0/laws/src/main/scala/cats/laws/MonadErrorLaws.scala) are well defined, but if you think there's any law that `cats.effect.IO` is violating, then it needs to be added to the `cats-laws` project.

In other words, when we are talking laws, the proof must be in the tests, otherwise there is no proof.

alexandru:
Btw, I updated my wording in my last message ... my point about ZIO being that the handling of unexpected exceptions is not done via the facilities of `MonadError`, which is totally fine as a philosophy when you're coming Scalazzi.

But coming from the viewpoint that unexpected exceptions should be handled via `MonadError[IO, Throwable]` there are only two choices and the de-facto Java standard is `addSuppressed`.

dwijnand:
Why isn't `addSuppressed` used in Scala.js?

alexandru:
@dwijnand Last time I checked it wasn’t implemented in Scala.js

jdegoes:
> The point is that your typical Scala app will run a lot of Java code, as the stack gets thicker and thicker as the app grows, and will mutate Throwable objects, whether we like it or note, therefore object Err extends Throwable is always dangerous, irregardless of what the IO implementation does.

This is not strictly true. The stack is created upon *construction* via `fillStackTrace()`, but is not subsequently changed after creation, unless you decide to change it yourself or pass it to mutating code.

In any case, there is a notable qualitative difference between:

 1. Imperative Java code in some third-party library mutating a `Throwable` that you pass it
 2. A supposedly purely functional library mutating your exception non-deterministically

(1) is unfortunate but expected. (2) should be banned.

> The point is that your typical Scala app will run a lot of Java code, as the stack gets thicker and thicker as the app grows, and will mutate Throwable objects, whether we like it or note, therefore object Err extends Throwable is always dangerous, irregardless of what the IO implementation does.

Actually, it works just fine with ZIO. Because ZIO doesn't mutate errors or wrap them in other errors (it doesn't even know the type of the errors). Not just ZIO either, by the way. Cats IO *used to* not do this either. 

We are talking about reverting a change that degraded guarantees. 

> The MonadError laws are well defined, but if you think there's any law that cats.effect.IO is violating, then it needs to be added to the cats-laws project.

Look at this implementation of equality for `Throwable`:

```scala
      def eqv(x: Throwable, y: Throwable): Boolean = {
        // All exceptions are non-terminating and given exceptions
        // aren't values (being mutable, they implement reference
        // equality), then we can't really test them reliably,
        // especially due to race conditions or outside logic
        // that wraps them (e.g. ExecutionException)
        (x ne null) == (y ne null)
      }
```

Pretending all pairs of `Throwable` of matching nullability are equal is not remotely correct.

**Arrays are mutable too. We don't consider two arrays equal if they share the same nullability.**

> It's totally fine that ZIO has a different philosophy however cats.effect.IO is built to handle unexpected exceptions thrown in Java code, so this always works:

If that's your design goal, it fails non-deterministically. 

A user who throws `FileNotFoundException` may unexpectedly have to catch `CompositeException`; if they were relying on catching the same error type that they threw, say, in a partial function defined by a match block, this will lead to throwing a match error, which further obfuscates the problem, all the way up the call stack. 

So a user does not know whether to catch `FileNotFoundException` or `CompositeException`, because it depends on runtime information (e.g. finalizer throwing an exception). 

It is not possible to program in a principled fashion if the errors you catch are different than the errors you threw, **especially when they change non-deterministically based on runtime behavior**.

The current behavior of Cats IO is clearly broken because (1) a purely functional library is mutating `Throwable` values passed into the library (which is both dangerous and extremely surprising), (2) it's non-deterministically impossible to catch the same error that you throw, and (3) behavior differs on Scala.js versus JVM to such an extent that code which works in one may easily fail in the other.

These are serious, major issues that can be addressed easily and quickly without changing the fundamental Cats IO error model.

dwijnand:
Looks like there's some work to be done to reach a consensus.  Once that's done I'd be happy to try and help make the necessary changes.  Meanwhile, I'll close out this initial proposal.

alexandru:
@dwijnand sorry for the discussion on your PR.

I don't think you should close it. If we want to avoid `addSuppressed`, the only problem happens in `bracket`, so what to do if both `use` and `release` are throwing.

An alternative would be to do this:

1. if `use` throws, but `release` completes normally, then signal the error thrown in `use`
2. if `use` completes normally and `release` throws, then signal the error thrown in `release`
3. if both throw ... signal the error in `use` and log the error from `release`

Lets see what the others are saying about it.

SystemFw:
I don't have a strong opinion, except for saying that the PR shouldn't be closed :)
The discussion is good, and there's a good chance a closed PR with no issue iwill be neglected.

dwijnand:
> @dwijnand sorry for the discussion on your PR.

No apologies necessary, discussions happen.  Opened #452 to track the issue, and allow for the discussion to happen.  I like your proposal, it feels natural and seems to correct.  If I get a chance during the break I'll try to PR it.

Issue #199, opened 2018-05-06
`deferFuture` either on `Async` or the `Async` package object.


jmcardon:
Current interop with future is:
```scala
def fromFuture[A](iof: IO[Future[A]]): IO[A] =
    iof.flatMap { f =>
      IO.async { cb =>
        
    }
```

I propose something like: (Edited)

```scala
object Async {
//...
def deferFuture[F[_], A](f: => Future[A])(implicit F: Async[F]) = {
  F.async {
     f.onComplete(r => cb(r match {
          case Success(a) => Right(a)
          case Failure(e) => Left(e)
        }))(immediate)
      }
  } 
}
```

Reasons:
- This is a common operation for people.
- (Minor reason) `fromFuture` is 3 extra allocations and evals used right: Pure + Suspend (defer) + Bind + Async. This would only be `Async`.
- `fromFuture` allows you to construct non-RT values with `IO.pure` or whatever. While you can go wrong with `deferFuture` at least it is explicitly lazy.

LukaJCB:
Related: https://github.com/typelevel/cats-effect/issues/78

jmcardon:
I edited the proposal to have it, like `shift`, a convenience method in the companion object. It doesn't hurt to have it just be a convenience method, without removing `IO.fromFuture`.

#78 's thread points to the fact that `=>` isn't exactly capturing effects, which is fair, but both `fromFuture` and `deferFuture` can be misused, just `fromFuture` has _more_ ways to misuse it with `IO.pure(fut)` (where that `fut` is performing side effects). 

Worst case and enough people disagree, then an alleycats-like module for cats-effect is where this can live.

johnynek:
I like it. And I actually like it on the trait because many Task implementations might be close to `() => Future[A]` and they could optimize this method.

johnynek:
While we are at it, it is not clear to me why there is not `F[Future[A]] => F[A]` on the trait, `object Async` or the syntax.

jmcardon:
@johnynek having it on the trait means having to define laws for it, which is where it being a convenience method wins out in my mind. That's all. I'm sure we can write laws for this without much issue.

johnynek:
I think they are pretty natural generalizations of the delay laws, but maybe you are right that the opportunity for optimization is not great enough to merit the added complexity.

SystemFw:
[...]

PR #453, opened 2018-12-18
fromFuture

sideeffffect:
I'm sorry, I don't have any experience with `MiMa` but some tests are failing with `Binary compatibility check failed!` even though I'm only introducing new stuff.
how can I fix this?
I'd appreciate any help or pointer :)

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Async.scala
+++ b/core/shared/src/main/scala/cats/effect/Async.scala
@@ -21,12 +21,12 @@ import simulacrum._
 import cats.implicits._
 import cats.data._
 import cats.effect.IO.{Delay, Pure, RaiseError}
-import cats.effect.concurrent.{Ref, Deferred}
+import cats.effect.concurrent.{Deferred, Ref}
 import cats.effect.internals.{Callback, IORunLoop}
 
 import scala.annotation.implicitNotFound
 import scala.concurrent.ExecutionContext
-import scala.util.Either
+import scala.util.{Either, Failure, Success}
 
 /**
  * A monad that can describe asynchronous or synchronous computations
@@ -318,6 +318,43 @@ object Async {
           .rethrow
       }
     }
+
+  /**
+    * Constructs an `Async` which evaluates the given `Future` and
+    * produces the result (or failure).
+    *
+    * Because `Future` eagerly evaluates, as well as because it
+    * memoizes, this function takes its parameter as an `Async`,
+    * which could be lazily evaluated.  If this laziness is
+    * appropriately threaded back to the definition site of the
+    * `Future`, it ensures that the computation is fully managed by
+    * `Async` and thus referentially transparent.
+    *
+    * Example:
+    *
+    * {{{
+    *   // Lazy evaluation, equivalent with by-name params
+    *   Async.fromFuture(F.delay(startRunningFuture()))
+    *
+    *   // Eager evaluation, for pure futures
+    *   Async.fromFuture(F.pure(startRunningFuture()))
+    * }}}
+    */
+  def fromFuture[F[_], A](af: F[scala.concurrent.Future[A]])(implicit F: Async[F], cs: ContextShift[F]): F[A] = {
+    F.flatMap(af) { f =>
+      f.value match {
+        case Some(r) =>
+          F.fromTry(r)
+        case _ =>
+          cs.shift *> F.async { cb =>
+            f.onComplete {
+              case Success(v) => cb(Right(v))
+              case Failure(e) => cb(Left(e))
+            } (internals.TrampolineEC.immediate)
+          }
+      }
+    }
+  }
 
   /**
    * [[Async]] instance built for `cats.data.EitherT` values initialized
--- a/core/shared/src/main/scala/cats/effect/syntax/AllCatsEffectSyntax.scala
+++ b/core/shared/src/main/scala/cats/effect/syntax/AllCatsEffectSyntax.scala
@@ -18,6 +18,7 @@ package cats.effect.syntax
 
 trait AllCatsEffectSyntax
   extends BracketSyntax
+  with AsyncSyntax
   with ConcurrentSyntax
   with EffectSyntax
   with ConcurrentEffectSyntax
--- a/core/shared/src/main/scala/cats/effect/syntax/AsyncSyntax.scala
+++ b/core/shared/src/main/scala/cats/effect/syntax/AsyncSyntax.scala
@@ -1,0 +1,86 @@
+/*
+ * Copyright (c) 2017-2018 The Typelevel Cats-effect Project Developers
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.syntax
+
+import cats.effect.{Async, ContextShift}
+
+import scala.concurrent.Future
+
+trait AsyncSyntax extends Async.ToAsyncOps {
+
+  implicit def catsEffectSyntaxAsyncObj[F[_]](F: Async[F]): AsyncObjOps[F] =
+    new AsyncObjOps[F](F)
+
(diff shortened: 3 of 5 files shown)


alexandru:
Hi @sideeffffect, thanks for contributing.

Unfortunately we've talked about this before and the current consensus is to not add such a function on the `Async` type class. We've talked about it before, see: 

https://github.com/typelevel/cats-effect/issues/199

Some points:

1. we definitely cannot add it to the `Async` trait, for one because it breaks binary compatibility on Scala 2.11, but also because there isn't much potential for `F[_]`-specific optimizations
2. there isn't just one `Future`, there are many, including Java's `Future` and `CompletableFuture` and Twitter's `Future` ... I'm a little biased here, but in [Monix's](https://monix.io/) `monix-catnap` I introduced `FutureLift`, as I believe a different project should handle this concern

On your implementation:

1. having functions that require an implicit `ExecutionContext` is a problem, we don't want those, because in Cats-Effect we try to keep signatures as pure as possible; you'd need to use `ContextShift`
2. `() => Future[A]` is a problem too, because technically speaking only `F[_] : Sync` data types should suspend side effects and functions don't implement `Sync` and are supposed to be pure ... in your own code, that works if you know what you're doing, it also works in more pragmatic libraries, but not in Cats-Effect because Cats-Effect is a foundation

That said this is a common request, so it warrants careful consideration of whether we should add a `fromFuture` conversion on the `Async` companion object or not.

sideeffffect:
Hi @alexandru, thanks for such a quick response!

> consensus is to not add such a function on the Async type class

> we definitely cannot add it to the Async trait

I've added it as a syntax. does that address your concern(s)?

> there isn't just one Future

there's one that's in the Scala standard library :)
but I agree with your point. would you guys be open to introducing (possibly multiple) interoperability project(s), each for different *Future-like* class? I could create such interop project for `scala.concurrent.Future`

> implicit ExecutionContext

passing of the `ExecutionContext` is always explicit -- is it an issue even in such a case?

> `() => Future[A]`

I'm very open to removing this. should I get rid of that?

sideeffffect:
still failing because of binary compatibility, even though I implemented this as extension methods -- I probably don't understand how MiMa, could somebody please advise me how implement this? thanks :)

sideeffffect:
implemented on `Async` companion object

alexandru:
Holidays are coming, so expect latency in replies 🙂

Given that this was discussed before, and rejected, I've requested the review of other maintainers — hey, do you guys feel like we need a `fromFuture` on the `Async` companion object or not?

Have you changed your minds since the last time we talked about it in #199?

kubukoz:
I just want to point out that this is a commonly asked for feature on gitter (and we've had to implement it at least a few times in our company). Maybe it's worth adding something like FutureLift to a cats-effect-extras (or similar) module if people are using this so often?

alexandru:
@sideeffffect 

> would you guys be open to introducing (possibly multiple) interoperability project(s), each for different Future-like class? I could create such interop project for scala.concurrent.Future

As I mentioned, there is already a [FutureLift](https://monix.io/api/3.0/monix/catnap/FutureLift.html) type class in [monix-catnap](https://monix.io/docs/3x/#monix-catnap).

The reason for why I built a type class and added it in Monix is because Monix also provides a `CancelableFuture[A] <: Future[A]` in its `monix-execution` sub-project. The `Future` in Scala is an open interface that can be extended and so I did, see [CancelableFuture](https://monix.io/api/3.0/monix/execution/CancelableFuture.html). 

The proposed `Async.fromFuture(cancelableFutureRef)` wouldn't be able to work in the same way. Or otherwise it isn't generic.

---

@kubukoz 

> I just want to point out that this is a commonly asked for feature on gitter (and we've had to implement it at least a few times in our company). Maybe it's worth adding something like FutureLift to a cats-effect-extras (or similar) module if people are using this so often?

I agree that because it's a commonly asked feature, we need to entertain the possibility of adding it.

However a `cats-effect-extras` with a `FutureLift` in it already exists and it's called `monix-catnap`. It's a small sub-project, it does not bring Monix's `Task` or `Observable` in your classpath.

alexandru:
@kubukoz btw the idea of a `cats-effect-extras` was shot down multiple times, because in practice it ends up being a repository of junk that people don't really want to commit to.

So either something is in Cats-Effect, or it's a third-party project that's not Cats-Effect, or in other words we either commit to it, or we don't, no half measures.

djspiewak:
I’m definitely in favor of it being something in catnip, not in cats-effect itself. It’s always easy to add things, hard to remove them, and cats-effect is really really central to the ecosystem.

I think it’s also important to recognize that, theoretically, it should be possible to have an Async implementation which is able to better implement fromFuture than the direct approach of running to the boundary. Obviously IO can’t do much better, and I don’t think that’s a problem, but maybe a different implementation is possible which biases in favor of making Future conversions fast. We shouldn’t rule that out.

So having something in a third party project like catnip is good and useful, and having the concrete conversion on IO is essential, but I wouldn’t be in favor of adding the conversion in any form directly into the abstract hierarchy.

rossabaker:
The only misgiving I have about monix-catnap is that its release depends on cats-effect, and we'll have no solution between a (new crossbuild || breaking) cats-effect and the release of a compatible monix.  I expect this pain to be both infrequent and brief.

I have implemented this a few times, enough to want this in cats-effect-core in due time.  But the approaches still have enough variance (e.g., `FutureLift` or direct), I am content to let this incubate in monix-catnap and elsewhere until we are prepared to chisel it into stone.  As Daniel notes, this repo must be extremely conservative given its place in the ecosystem.

sideeffffect:
the proposal to create small independent `cats-effect-interop-*` packages was not accepted (#455 was closed)
would you guys then reconsider adding `fromFuture` to the `cats-effect-core` package as @rossabaker suggested?

johnynek:
Personally, I'm pretty -1 on the the prescriptivism approach we often take. Everyone interacting with Future code will want this, I think we should add either `Async.fromFuture` as we have done here, or `Async.delayFromFuture[A](f: => Future[A])(implicit F: Async[F]): F[A]`.

We have `delay` already. We allow users to lift effectful code into a pure wrapper and trust that they learn how to do this properly. I see no reason to exclude Future, a standard type, from this.

sideeffffect:
1. I also think thunks are confusing/potentially dangerous gunfoots (at least in this context), so let's not discuss this any further
1. I would much prefer the `cats-effect-interop-*` packages solution that doesn't clutter `cats-effect-core` with helper methods, but others don't see it as a good solution
1. I think adding this to `cats-effect-core` is the second best option I know of
1. I (and I think many others) want to convert `Future`s in *polymorphic* code -- that code should be oblivious if it will end up as cats-effect IO, Monix Task or ZIO

djspiewak:
@sideeffffect 

> I (and I think many others) want to convert Futures in polymorphic code

I'm interested in further elaboration on this point. The only case that I can think of where this applies would be library code, where you're wrapping some sort of `Future`-producing library and publishing *another* library, which itself will be consumed in different contexts. Also remember that `LiftIO` makes this possible even in polymorphic contexts, though it's definitely a bit bulkier.

Perhaps some of my perspective is biased, since my only encounters with `Future` is in application code, where I generally have a concrete effect type.

SystemFw:
It's quite common to not have a concrete type in application code. All my apps are polymorphic in `F` until `main`, for example (increased parametricity, interaction with final tagless code, some Kleisli tricks here and there).

I'm not going to comment on this specific issue though because I keep changing my mind and ultimately I don't particularly care which way we go here.

sideeffffect:
@djspiewak as @SystemFw said -- even in an app, concrete implementation is decided only at the very `main` method, otherwise everything is polymorphic, but somewhere inside I need to interop with a legacy `Furure`-producing library

kubukoz:
As a vote in favor of adding this, I've seen so many implementations of this that (turns out) weren't complete due to not doing a `shift` after the callback that it'd be really really nice to have a standard way of doing this out of the box. And many users will agree that having this isn't enough of a reason to add another dependency (on `monix-catnap`).

alexandru:
@kubukoz nobody claimed this is trivial, quite the contrary, the difficulty in getting it right has been an argument for keeping it out of Cats-Effect.

Also Cats-Effect’s goals is to be minimal because it is a foundation. Good or bad, there are certainly tradeoffs, but we’ve been encouraging people to build adjacent projects that we promote in the project’s README.

djspiewak:
I'm going to play fiat here and close this for the time being, just to break the impasse. Feel free to direct your complaints to me if you disagree. We can revisit this in the future (no pun intended). For the time being, I think the following is a perfectly reasonable approach (at the cost of specializing the thunking effect):

```scala
def someFutureThing: Future[A] = ???

def imapolymorphicfunction[F[_]: Async] =
  IO.fromFuture(IO(someFutureThing)).to[F]
```

PR #458, opened 2018-12-26
Avoid Throwable#addSuppressed & CompositeException

### LINKED ISSUE #452: Reconsider Throwable addSuppressed usage
See https://github.com/typelevel/cats-effect/pull/449 for an initial proposal and the exchange of thoughts and ideas around it.

dwijnand:
Log as a last resort.

Fixes #452

### DIFF
--- a/core/shared/src/test/scala/cats/effect/internals/TestUtils.scala
+++ b/core/shared/src/test/scala/cats/effect/internals/TestUtils.scala
@@ -16,7 +16,8 @@
 
 package cats.effect.internals
 
-import java.io.{ByteArrayOutputStream, PrintStream}
+import java.io.{ByteArrayOutputStream, OutputStream, PrintStream}
+
 import scala.util.control.NonFatal
 
 /**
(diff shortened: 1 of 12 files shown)


dwijnand:
In the interests of not logging exceptions in tests I tried to implement `ReleaseRecover#recover` with a new `RaiseErrors` constructor for `IO`, but I got stuck uncertain where between `IOPlatform.unsafeResync`, `IORunLoop.step`, and `IORunLoop.loop` I needed to make the changes, so I just left it as a TODO for someone else.

Let me know if there's anything I can do that would help to approve this.

laws/shared/src/test/scala/cats/effect/IOTests.scala
```diff
@@ -652,6 +652,42 @@ class IOTests extends BaseTestsSuite {
...
+  test("bracket signals the error in use and logs the error from release") {
+    val e1 = new RuntimeException("error in use")
+    val e2 = new RuntimeException("error in release: expected to be logged during test")
+
+    val r = IO.unit.bracket(_ => IO.raiseError(e1))(_ => IO.raiseError(e2))
+      .attempt
+      .unsafeRunSync()
+
+    r shouldEqual Left(e1)
+    // TODO: assert e2 is logged
```
alexandru: You can test this in two ways:

1. override `System.err`, like here: https://github.com/typelevel/cats-effect/blob/v1.1.0/core/shared/src/test/scala/cats/effect/internals/TestUtils.scala#L55 (this should work with Scala.js too)
2. execute that logic on top of a thread with your own [UncaughtExceptionHandler](https://docs.oracle.com/javase/7/docs/api/java/lang/Thread.UncaughtExceptionHandler.html), this only works on the JVM

core/shared/src/test/scala/cats/effect/internals/TestUtils.scala
```diff
@@ -52,9 +53,17 @@ trait TestUtils {
   /**
    * Catches `System.err` output, for testing purposes.
    */
-  def catchSystemErr(thunk: => Unit): String = synchronized {
-    val oldErr = System.err
+  def catchSystemErr(thunk: => Unit): String = {
     val outStream = new ByteArrayOutputStream()
+    catchSystemErrInto(outStream)(thunk)
+    outStream.toString("utf-8")
```
rossabaker: Nit: I usually refer to `StandardCharsets.UTF_8` just to have a constant, but I won't hold up progress over it.

PR #462, opened 2018-12-31
not create instance of TimeoutException if no real need

t3hnar:


t3hnar:
or should I better change signature 

```
def timeoutTo[F[_], A](fa: F[A], duration: FiniteDuration, fallback: F[A])
```
to
```
def timeoutTo[F[_], A](fa: F[A], duration: FiniteDuration, fallback: => F[A])
```

rossabaker:
Thanks!

As for the change, is it to avoid the overhead of filling in the stack trace of an exception that will not be thrown on the happy path?  I haven't used this idiom before, nor benchmarked it, but it _seems_ reasonable.

Cats and cats-effect generally do not implement laziness with by-name parameters except in types and typeclasses that represent laziness.  It would also break compatibility.  I would not change the signature, as suggested in the second comment.

t3hnar:
@rossabaker 

> is it to avoid the overhead of filling in the stack trace of an exception that will not be thrown on the happy path? 

yes, creating exception is an expensive operation.

alexandru:
How expensive? Did you notice actual problems? Do you have a benchmark?

The advantage of working with `IO` is that values can be reused, so something like `IO(1).timeout(1.second)` can be cached.

Also the disadvantage of your solution is that the stack trace won't show the call site of the `timeout` operation.

alexandru:
I might also add that `timeout` is going to get used on expensive `IO` values for which the cost of creating an exception is probably irrelevant.

t3hnar:
>  so something like IO(1).timeout(1.second) can be cached.

I would say that this is rather a rare case.

rossabaker:
> Yes, many times I faced issues with creating unnecessary exception and making the application slow.

I have too, in CPU-bound code.  But I'm skeptical of it being noticeable in this context.  The benchmark on `IO.unit` represents a worst case scenario that seems very far removed from the network calls that people  timeout in practice, and I'm not convinced it's worth the damage that would be done to the stack traces.

mpilquist:
I'm :+1: on this change. A good motivating example is timing out dequeue operations on a queue.

t3hnar:
Done. @alexandru thx for the hint.

PR #464, opened 2019-01-03
add `mapK` onto `Resource`

andimiller:
* Implement it by recursing over the structure to apply the transform
* This may not be stack-safe, but it's no less stack safe than `use` so I can't tell

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Resource.scala
+++ b/core/shared/src/main/scala/cats/effect/Resource.scala
@@ -388,7 +388,6 @@ object Resource extends ResourceInstances {
   final case class Suspend[F[_], A](
     resource: F[Resource[F, A]])
     extends Resource[F, A]
-
 }
 
 private[effect] abstract class ResourceInstances extends ResourceInstances0 {
--- a/core/shared/src/main/scala/cats/effect/Resource.scala
+++ b/core/shared/src/main/scala/cats/effect/Resource.scala
@@ -148,6 +148,20 @@ sealed abstract class Resource[F[_], A] {
     flatMap(a => Resource.pure[F, B](f(a)))
 
   /**
+    * Given a natural transformation from `F` to `G`, transforms this
+    * Resource from effect `F` to effect `G`.
+    */
+  def mapK[G[_]](f: F ~> G)(implicit B: Bracket[F, Throwable], D: Defer[G], G: Applicative[G]): Resource[G, A] =
+    this match {
+      case Allocate(resource) =>
+        Allocate(f(resource).map { case (a, r) => (a, r.andThen(u => f(u))) })
+      case Bind(source, f0) =>
+        Bind(Suspend(D.defer(G.pure(source.mapK(f)))), f0.andThen(_.mapK(f)))
+      case Suspend(resource) =>
+        Suspend(f(resource).map(_.mapK(f)))
+    }
+
+  /**
     * Given a `Resource`, possibly built by composing multiple
     * `Resource`s monadically, returns the acquired resource, as well
     * as an action that runs all the finalizers for releasing it.
@@ -374,6 +388,7 @@ object Resource extends ResourceInstances {
   final case class Suspend[F[_], A](
     resource: F[Resource[F, A]])
     extends Resource[F, A]
+
 }
 
 private[effect] abstract class ResourceInstances extends ResourceInstances0 {
--- a/laws/shared/src/test/scala/cats/effect/ResourceTests.scala
+++ b/laws/shared/src/test/scala/cats/effect/ResourceTests.scala
@@ -17,6 +17,9 @@
 package cats
 package effect
 
+import java.util.concurrent.atomic.AtomicBoolean
+
+import cats.data.Kleisli
 import cats.effect.laws.discipline.arbitrary._
 import cats.kernel.laws.discipline.MonoidTests
 import cats.laws._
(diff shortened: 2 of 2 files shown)


core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -104,31 +104,47 @@ sealed abstract class Resource[F[_], A] {
...
       loop(current, stack)
 
     // Interpreter that knows how to evaluate a Resource data structure;
     // Maintains its own stack for dealing with Bind chains
-    @tailrec def loop(current: Resource[F, Any], stack: List[Any => Resource[F, Any]]): F[Any] = {
+    @tailrec def loop(current: Resource[F, Any], stack: List[Any => Resource[F, Any]]): F[(Any, ExitCase[Throwable] => F[Unit])] = {
       current match {
         case Allocate(resource) =>
-          F.bracketCase(resource) { case (a, _) =>
+          resource.flatMap { case (a, release) =>
```
SystemFw: Having `flatMap` instead of `bracket` here seems suspicious, what's the rationale for it?

SystemFw: Ah, already stale

andimiller: yeah spotted that one and moved the match outside it, the `flatMap` was left over from how it used to work

andimiller:
Looks like this doesn't preserve the release behaviour across effects yet, will give it a think and see if I can work out where it does that

andimiller:
This simpler implementation correctly sequences acquisition and release, but doesn't preserve any custom exitcode handlers in the release you're transforming

core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -147,6 +147,13 @@ sealed abstract class Resource[F[_], A] {
   def map[B](f: A => B)(implicit F: Applicative[F]): Resource[F, B] =
     flatMap(a => Resource.pure[F, B](f(a)))
 
+  /**
+    * Given a natural transformation from `F` to `G`, transforms this
+    * Resource from effect `F` to effect `G`.
+    */
+  def mapK[G[_]: Applicative](f: F ~> G)(implicit B: Bracket[F, Throwable]): Resource[G, A] =
+    Resource.apply(f(allocated.map{ case (a, r) => (a, f(r))}))
```
SystemFw: This is not safe. In case `map` is implemented in terms of `flatMap`, then `allocated.flatMap` can be interrupted on the `flatMap` and leak. Never use `allocated` without `bracket`, basically

SystemFw: The other problem with this strategy (I feel), is that it runs the whole `Resource` into `F`, and then translates the result into `G`, whereas I feel it should be run in `G`

andimiller: to run each of the individual steps in `G` I'd need to add a new `MapK` node into the ADT, but I think this then breaks all of the current execution styles, since they assume a constant `F` throughout, I haven't found a nice way to do that yet, without separating it into two ADTs

SystemFw: > to run each of the individual steps in G I'd need to add a new MapK node into the ADT, 

I can't see why this is the case

SystemFw:
> but doesn't preserve any custom exitcode handlers in the release you're transforming

Is there a test for this behaviour?

andimiller:
> > but doesn't preserve any custom exitcode handlers in the release you're transforming
> 
> Is there a test for this behaviour?

I've just added it, and it fails as I expected

andimiller:
and it passes if I add `allocatedCase` and use that to implement `mapK`

core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -152,7 +152,7 @@ sealed abstract class Resource[F[_], A] {
     * Resource from effect `F` to effect `G`.
     */
   def mapK[G[_]: Applicative](f: F ~> G)(implicit B: Bracket[F, Throwable]): Resource[G, A] =
-    Resource.apply(f(allocated.map{ case (a, r) => (a, f(r))}))
+    Resource.applyCase(f(B.uncancelable(allocatedCase.map{ case (a, r) => (a, r.andThen(f.apply))})))
```
SystemFw: Remember when I told you that `allocated.map` was unsafe here because it could be interrupted? I was entirely wrong.
It can't be interrupted because that `F` is put into a `Resource.Allocated` case, which is ultimately (in `use`) passed as the `acquire` action in `bracket`, which is not interruptible.

The good news is that you don't need `B.uncancelable` here, the bad news is that (with or without an explicit `uncancelable`) now the _whole_ target resource is not cancelable, whereas the source resource was, which is a problem.

core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -211,6 +218,47 @@ sealed abstract class Resource[F[_], A] {
     }
   }
 
+   def allocatedCase(implicit F: Bracket[F, Throwable]): F[(A, ExitCase[Throwable] => F[Unit])] = {
```
alexandru: We already have `allocated`. Can't we express `allocated` in terms of `allocatedCase`?

andimiller:
I've opened the issue for stack-safety of `Resource.use` in #469

SystemFw:
Given comments in #469 , I'm :+1: on merging this PR as is.
It would be good to get a second review.

PR #466, opened 2019-01-07
Update Resource documentation

dqdinh:
For string interpolation, `$$` should be `$`


dqdinh:
Apologies again - I was just going through the tutorial and reading the source code examples - copied and pasted the source code and the $$ caused the output of the examples to not work. Didn't know that ScalaDoc $$ is the escaping of $. Sorry for the white noise. Thanks again for writing the tutorial -- Its a great way to learn how to use cats-effect!

PR #470, opened 2019-01-10
Do some doc fixing

ceedubs:
This started out as me just attempting to provide a link to the
"Acquiring and releasing `Resource`s" section of the tutorial from with
tin the `Resource` data type doc. But then I went down a bit of a rabbit
hole...

There was a Scala 2.12.6 vs 2.12.7 mismatch between `build.sbt` and
`.travis.yml`. This meant that the build hasn't actually been compiling
the docs recently. So I fixed a few spots that weren't compiling in the
docs (though I'm not sure whether I fixed them in the preferred way). I
went ahead and attempted to make the scala version check for the docs
build a bit more lenient to hopefully avoid a similar issue in the
future. Admittedly my bash skills are lacking, so we should probably
check that it works as expected in the Travis build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ script:
   - export SBT_PROFILE=$COVERAGE
   - sbt ++$TRAVIS_SCALA_VERSION $COVERAGE ci
   - |
-    if [ "$TRAVIS_SCALA_VERSION" = "2.12.7"]; then
+    if [[ "$TRAVIS_SCALA_VERSION" =~ "2\.12\." ]]; then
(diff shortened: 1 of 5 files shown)


ceedubs:
Come to think of it, I don't think that the scala version mismatch would have actually caused the build to not compile the docs, since it's going off of `TRAVIS_SCALA_VERSION` and not the `scalaVersion` in the `build.sbt`. Hmm so why weren't the docs compiling?

ceedubs:
@alexandru any thoughts on https://github.com/typelevel/cats-effect/pull/470#issuecomment-453170788 ?

PR #472, opened 2019-01-13
Make IO.Par's applicative commutative

kubukoz:
Apparently MiMa claims this is binary incompatible, but maybe it's a false positive (the trait is `private[effect]` so maybe it only matters for implementations), so I submitted it anyway.

Please let me know what you think about making IO.Par's applicative commutative (I checked it before and it passed the laws... apparently)

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -730,9 +730,29 @@ private[effect] abstract class IOParallelNewtype
    * for more details.
    */
   object Par extends IONewtype
+
+  private[effect] def ioParCommutativeApplicative(implicit cs: ContextShift[IO]): CommutativeApplicative[IO.Par] =
+    new CommutativeApplicative[IO.Par] {
+      import IO.Par.{unwrap, apply => par}
+
+      final override def pure[A](x: A): IO.Par[A] =
+        par(IO.pure(x))
+      final override def map2[A, B, Z](fa: IO.Par[A], fb: IO.Par[B])(f: (A, B) => Z): IO.Par[Z] =
+        par(IOParMap(cs, unwrap(fa), unwrap(fb))(f))
+      final override def ap[A, B](ff: IO.Par[A => B])(fa: IO.Par[A]): IO.Par[B] =
+        map2(ff, fa)(_(_))
+      final override def product[A, B](fa: IO.Par[A], fb: IO.Par[B]): IO.Par[(A, B)] =
+        map2(fa, fb)((_, _))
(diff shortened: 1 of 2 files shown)


alexandru:
This may be problematic. `IOInstances` may be `private[effect]` however what's public is `object IO` which inherits it.

That said I thought that in a relationship like `B <: A` if you upgrade a function return type from `A` to `B` it should still be binary compatible. So I wonder why it doesn't work here.

alexandru:
Maybe Mima complains because the upgrade is in `IOInstances` for a non-final method that can be overridden.

kubukoz:
Maybe... I'll look into it. Is the general idea OK?

alexandru:
I think so, but we need an answer for the Mima issue.

Not sure what we can do to find out. Maybe do a simulation with such an update attempted on just an `object`.

kubukoz:
Alright, finally got around to it. Looks like it is indeed incompatible.

Here's what I did (just to make sure I did the right thing):

`IO.scala`:
```scala
object IO extends IOInstances

trait Animal
class Dog extends Animal

trait IOInstances {
  def instance: Animal = new Dog
}
```

`Usage.scala`:
```scala
object Usage extends App {
  val a: Animal = IO.instance
  println(a)
}
```

```
$ scalac IO.scala
$ scalac Usage.scala
$ scala Usage
< Dog@7fbe847c
```

Then I changed the line in IOInstances:

```diff
-  def instance: Animal = new Dog
+  def instance: Dog = new Dog
```

Recompiled `IO.scala`, ran Usage again:

```scala
java.lang.NoSuchMethodError: IO$.instance()LAnimal;
	at Usage$.delayedEndpoint$Usage$1(User.scala:2)
	at Usage$delayedInit$body.apply(User.scala:1)
	at scala.Function0.apply$mcV$sp(Function0.scala:39)
	at scala.Function0.apply$mcV$sp$(Function0.scala:39)
	at scala.runtime.AbstractFunction0.apply$mcV$sp(AbstractFunction0.scala:17)
	at scala.App.$anonfun$main$1$adapted(App.scala:80)
	at scala.collection.immutable.List.foreach(List.scala:392)
	at scala.App.main(App.scala:80)
	at scala.App.main$(App.scala:78)
	at Usage$.main(User.scala:1)
	at Usage.main(User.scala)
```

I suppose we would need to add a new method that returns a more specific type (`CommutativeApplicative`) and do some prioritization tricks to get it to work.

kubukoz:
I submitted a workaround in the latest commit, basically moving the instance as explicit to the `IOParallelNewtype` trait and accessing it by implicits in the appropriately prioritized traits (at least this seems to be working when I check MiMa locally and see if the implicits are accessible)

PR #473, opened 2019-01-15
Add "repeated asyncF evaluation not memoized" law

kubukoz:
I thought it would be reasonable to check that.

It's easy to break this law, for example by removing the `suspend` call here: https://github.com/kubukoz/slick-effect/blob/3af6c4ab8950f42ebbcaaed0fc5621c503bb258a/src/main/scala/slickeffect/DBIOAsync.scala#L27. That would memoize the promise, and it would only be completed once.

### DIFF
--- a/laws/shared/src/main/scala/cats/effect/laws/AsyncLaws.scala
+++ b/laws/shared/src/main/scala/cats/effect/laws/AsyncLaws.scala
@@ -37,9 +37,22 @@ trait AsyncLaws[F[_]] extends SyncLaws[F] {
   def repeatedAsyncEvaluationNotMemoized[A](a: A, f: A => A) = {
     var cur = a
 
-    val change: F[Unit] = F async { cb =>
+    val change: F[Unit] = F.async { cb =>
       cur = f(cur)
       cb(Right(()))
+    }
+
+    val read: F[A] = F.delay(cur)
+
+    change *> change *> read <-> F.pure(f(f(a)))
+  }
+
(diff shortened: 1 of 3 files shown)


laws/shared/src/main/scala/cats/effect/laws/AsyncLaws.scala
```diff
@@ -47,6 +47,19 @@ trait AsyncLaws[F[_]] extends SyncLaws[F] {
...
+    var cur = a
+
+    val change: F[Unit] = F.asyncF { cb =>
+      cur = f(cur)
+      F.delay(cb(Right(())))
+    }
+
+    val read: F[A] = F.delay(cur)
+
+    change *> change *> read <-> F.pure(f(f(a)))
```
alexandru: I see you copied the implementation of `repeatedSyncEvaluationNotMemoized`. That's fine, but what I don't like is that `var` being exposed.

What I would do is to suspend the left hand side, otherwise the value returned by `repeatedAsyncFEvaluationNotMemoized` isn't pure.

kubukoz: Isn't the value returned also impure after suspending the left side? I think we'd need to suspend the part where the var is created, too (so basically the whole function).

alexandru: Yes, that's what I meant, suspend the `var`.

PR #479, opened 2019-01-25
Get rid of vars in SyncLaws and AsyncLaws

### LINKED ISSUE #231: Uncaptured vars in laws
[`SyncLaws#repeatedSyncEvaluationNotMemoized`](https://github.com/typelevel/cats-effect/blob/master/laws/shared/src/main/scala/cats/effect/laws/SyncLaws.scala#L78) creates a `var`, which is then modified by the `F[_]` which is returned. Creating the var is not inside a `delay` or `suspend`. This makes the returned `F[_]` runnable only once (or rather, the second run will reuse the already modified `var`) which results in correct code seeming to violate the law.

Other laws which seem suspect (although I didn't tested them explicitly):

- `SyncLaws`:
  - `bracketReleaseCalledForError`
  - `unsequencedDelayIsNoop` (although here it probably doesn't matter)
  - `bindSuspendsEvaluation`
  - `mapSuspendsEvaluation`
- `AsyncLaws`:
  - `repeatedAsyncEvaluationNotMemoized`
  - `repeatedCallbackIgnored`

daron666:
Fixes #231

### DIFF
--- a/laws/shared/src/main/scala/cats/effect/laws/AsyncLaws.scala
+++ b/laws/shared/src/main/scala/cats/effect/laws/AsyncLaws.scala
@@ -19,6 +19,7 @@ package effect
 package laws
 
 import java.util.concurrent.atomic.AtomicReference
+import java.util.function.UnaryOperator
 
 import cats.effect.ExitCase.{Completed, Error}
 import cats.effect.concurrent.{Deferred, Ref}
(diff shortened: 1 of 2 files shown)


daron666:
Ok, got your point. So then just close #231.

daron666:
@alexandru I'll try to find something else to do here 😄

alexandru:
@daron666 I think we have laws in which the vars aren’t suspended. You might want to fix those by suspending them.

PR #481, opened 2019-01-27
Patch for tutorial, to prevent blocking in cancellation

lrodero:
There is a catch on current tutorial code: the TCP server tries to `.cancel` a fiber that is based on `bracketCase`. Since CE 1.1.0 this cancellation will block the caller until the bracket finalizes, however the TCP server needs that cancellation to return immediately. As the tutorial is based on 1.0.0, it needs to be updated.

This patch modifies the tutorial code as needed by running the cancellation in its own fiber (`.cancel.start`), following a suggestion from Alex Nedelcu in gitter. It also introduces some text explaining why we need to do that. Finally it includes an unrelated typo fix.

The final result can be seen at https://lrodero.github.io/cats-effect/tutorial/tutorial.html

### DIFF
--- a/site/src/main/tut/tutorial/tutorial.md
+++ b/site/src/main/tut/tutorial/tutorial.md
@@ -833,7 +833,7 @@ call to `fiber.cancel`) is also run in a separate fiber to prevent being
 blocked by it. This is not always needed, but the cancellation of actions
 defined by `bracket` or `bracketCase` will wait until all finalizers (release
 stage of bracket) are finished. The `F` created by our `serve` function is
-debined based on `bracketCase`, so if the action is blocked at any bracket stage
+defined based on `bracketCase`, so if the action is blocked at any bracket stage
 (acquisition, usage or release), then the cancel call will be blocked too. And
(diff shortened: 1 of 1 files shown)


site/src/main/tut/tutorial/tutorial.md
```diff
@@ -828,7 +828,16 @@ that message is received.
 
 Let's first define a new method `server` that instantiates the flag, runs the
 `serve` method in its own fiber and waits on the flag to be set. Only when
-the flag is set the server fiber will be canceled.
+the flag is set the server fiber will be canceled. The cancellation itself (the
+call to `fiber.cancel`) is also run in a separate fiber to prevent being
+blocked by it. This is not always needed, but the cancellation of actions
+defined by `bracket` or `bracketCase` will wait until all finalizers (release
+stage of bracket) are finished. The `F` created by our `serve` function is
+debined based on `bracketCase`, so if the action is blocked at any bracket stage
```
rossabaker: Typo:

```suggestion
defined based on `bracketCase`, so if the action is blocked at any bracket stage
```

lrodero:
Typo fix added :+1: !

PR #483, opened 2019-01-29
Updated tutorial.md with a fix for non-terminating echo server.

deniszjukow:


lrodero:
This is already dealt with by PR [#481]( https://github.com/typelevel/cats-effect/pull/481).

rossabaker:
Thank you for the PR.  I agree, this is already in #481.

PR #490, opened 2019-02-27
Add Resource.evalTap

nigredo-tori:
A cross between `flatTap` and `evalMap`. Use case - an action returning a temporary resource with custom setup. For me this pattern pops up all the time when writing logic that uses temporary files (create a temporary file, write some data to it, yield the file).

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Resource.scala
+++ b/core/shared/src/main/scala/cats/effect/Resource.scala
@@ -231,6 +231,13 @@ sealed abstract class Resource[F[_], A] {
     */
   def evalMap[B](f: A => F[B])(implicit F: Applicative[F]): Resource[F, B] =
     this.flatMap(a => Resource.liftF(f(a)))
+
+  /**
+    * Applies an effectful transformation to the allocated resource. Like a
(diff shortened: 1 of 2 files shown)


kubukoz:
Not sure if it's worth the effort - consecutive `evalTap` invocations like `evalMap(f).evalMap(g).use(h)` are pretty much like `use(a => f(a) *> g(a) *> h(a))`.

nigredo-tori:
>  consecutive evalTap invocations like `evalMap(f).evalMap(g).use(h)` are pretty much like `use(a => f(a) *> g(a) *> h(a))`

This works when you immediately do `.use`, but `.evalTap` is useful for more complex cases (like traversing a collection in `Resource`, or returning a prepared `Resource` from a function).

mpilquist:
:+1: from me -- fs2 has `evalTap` as well:

```scala
// On Stream[F, O]
def evalTap[F2[x] >: F[x]: Functor](f: O => F2[Unit]): Stream[F2, O] =
```

djspiewak:
This is merged into `series/1.x`, but due to a snafu on my end, we didn't update the base *before* the merge. So this will eventually land in master, but only after the 1.3 release (soon™).

PR #491, opened 2019-02-27
Add LiftIO instance for Resource

nigredo-tori:
I'm not sure there is a nice way to build a `Bracket` instance, but there is an obvious, and rather useful, `LiftIO`.


nigredo-tori:
@rossabaker, what do you mean? `LiftIO` has no laws, AFAICT.

Issue #465, opened 2019-01-04
Add `LiftIO.liftK`


Daenyth:
`object LiftIO` could have this for convenience:
```scala
def liftK[F[_]: LiftIO]: IO ~> F = λ[IO ~> F](_.to[F])
```

Daenyth:
There could also be an `Effect.liftK` for `F ~> IO`

Daenyth:
I could PR this in since it seems like there's support. Are there any kind of unit tests that are useful here? I'm not really sure what is important.

nigredo-tori:
Someone should probably close this.

PR #492, opened 2019-02-28
Add LiftIO.liftK, Effect.toIOK

nigredo-tori:
As proposed by @Daenyth in #465.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Effect.scala
+++ b/core/shared/src/main/scala/cats/effect/Effect.scala
@@ -73,6 +73,11 @@ object Effect {
     }
 
   /**
+    * [[Effect.toIO]] as a natural transformation.
+    */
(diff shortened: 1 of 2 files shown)


nigredo-tori:
> Wouldn't mind seeing the corresponding one for `Effect`

Done. Called it `toIOK` to minimize confusion.

rossabaker:
But I would love to see a test if we can write one that doesn't just parrot the implementation.

nigredo-tori:
> But I would love to see a test if we can write one that doesn't just parrot the implementation.

I'm not sure what that should check, then, since the implementation is just building a `FunctionK`. I can't think of anything other than `ioa.to[F] <-> LiftIO.liftK[F].apply(ioa)`, and that tests kindprojector more than anything.

rossabaker:
I can't think of an interesting test either.  We'll see if someone else does, but it has my approval.

PR #496, opened 2019-03-19
Make Resource.liftF interruptible

SystemFw:
Currently `liftF` on `Resource` is implemented as an `Allocated` node with a no-op release, so 
`Resource.liftF(fa)` produces `Resource.Allocated(fa, _ => F.unit)`. 
During interpretation, `Allocated` nodes get translated into `bracket` calls, so `liftF` ends up producing `bracket(fa)(...)(_ => F.unit)`, which is no longer a no-op because `fa` is made uninterruptible by the fact that it's being passed as a `bracket.acquire` action.

----------

As a consequence, the following law is broken:
```
Resource.liftF(fa).use(IO.pure) <-> fa
```

and what ends up holding is
```
Resource.liftF(fa).use(IO.pure) <-> fa.uncancelable
```
 (This wasn't caught by our existing property tests for this law)

----------

On a practical level, this caused a bug in fs2 where compiling a Stream down to a  Resource made the Stream uninterruptible

```
 Resource
     .makeCase(CompileScope.newRoot[F])((scope, ec) => scope.close(ec).rethrow)
     .flatMap { scope =>
          Resource.liftF {
              // the main stream loop is now uninterruptible because it's in liftF
                F.delay(init())
                  .flatMap(i => Algebra.compile(s.get, scope, i)(foldChunk))
                  .map(finalize)
            }
        }
```

This PR fixes this issue by implementing `liftF` as a `Suspend` node instead

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Resource.scala
+++ b/core/shared/src/main/scala/cats/effect/Resource.scala
@@ -315,7 +315,7 @@ object Resource extends ResourceInstances {
    * @param fa the value to lift into a resource
    */
   def liftF[F[_], A](fa: F[A])(implicit F: Applicative[F]): Resource[F, A] =
-    make(fa)(_ => F.unit)
+    Resource.suspend(fa.map(a => Resource.pure(a)))
 
   /**
     * Creates a [[Resource]] by wrapping a Java
--- a/laws/shared/src/test/scala/cats/effect/ResourceTests.scala
+++ b/laws/shared/src/test/scala/cats/effect/ResourceTests.scala
@@ -122,21 +122,6 @@ class ResourceTests extends BaseTestsSuite {
     }
   }
 
-  testAsync("evalMap with cancellation <-> IO.never") { implicit ec =>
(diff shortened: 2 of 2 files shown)


kubukoz:
Damn, just noticed this and apparently I inserted a lie in #516, will have to fix that

core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -315,7 +315,7 @@ object Resource extends ResourceInstances {
    * @param fa the value to lift into a resource
    */
   def liftF[F[_], A](fa: F[A])(implicit F: Applicative[F]): Resource[F, A] =
-    make(fa)(_ => F.unit)
+    Resource.suspend(fa.map(a => Resource.pure(a)))
```
kubukoz: Only now do I realize how smart this is. You avoided adding a new node and support from the use loop with this simple trick (`suspend(fa.map)`) - great work as always Fabio @systemfw!

PR #262, opened 2018-06-03
Add the Continual type class

alexandru:
Fixes #242 — adds a `Continual` type class.

This proposal isn't necessarily for 1.0. Being an add-on, we can release it afterwards, although you know where my heart is at this point 🙂 If there's consensus (hopefully), we could schedule it for a 1.1

```scala
trait Continual[F[_]] {
  // Restriction via composition instead of inheritance
  implicit val sync: Sync[F]

  def continual[A](fa: F[A]): F[A]

  def onCancelRaiseError[A](fa: F[A], e: Throwable): F[A] =
    sync.guaranteeCase(fa) {
      case Canceled => sync.raiseError(e)
      case _ => sync.unit
    }
}
```

Via `continual`, the run-loop of implementing data types are forced to uphold this law, if `Concurrent`:

```scala
  def continualCancelationModel[A](a: A)
    (implicit C: Concurrent[F]) = {

    val lh = for {
      start <- Deferred.uncancelable[F, Unit]
      wasCanceled <- C.liftIO(Deferred.uncancelable[IO, A])
      // Relevant bit here:
      task = start.get >> C.cancelable[A](_ => wasCanceled.complete(a))
      fiber <- C.start(task)
      _ <- fiber.cancel
      _ <- start.complete(())
      r <- C.liftIO(wasCanceled.get)
    } yield r

    F.continual(lh) <-> C.pure(a)
  }
```

Or in other words, whenever you have:

```scala
for (t1 <- task1; t2 <- task2) yield ()
```

If `task1` does not get cancelled or finish in error, then `task2` is guaranteed to at least start.

The second requirement for the type class is for a `onCancelRaiseError` to be described as being equivalent to a `guaranteeCase` invocation, which adds restrictions on what `bracket` can do with errors thrown in `release` — `bracket` (and `guaranteeCase`) are now forced to re-throw errors thrown in  `release` and to reset the "internal cancelation flag".

So in summary:

1. the "*continual*" model is specifying that the underlying task should evaluate without cancelable `flatMaps`
2. specifies an `onCancelRaiseError` operation derived from `bracket`, that places restrictions on how `bracket` deals with errors in `release`

Important to note — this type class is an add-on to the type class hierarchy, which is why it is using composition instead of inheritance.

## Raison D'être

Reasons:

[...]

PR #498, opened 2019-03-21
add parTraverseN and parSequenceN

sideeffffect:
/cc @jendakol @augi

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Async.scala
+++ b/core/shared/src/main/scala/cats/effect/Async.scala
@@ -21,7 +21,7 @@ import simulacrum._
 import cats.implicits._
 import cats.data._
 import cats.effect.IO.{Delay, Pure, RaiseError}
-import cats.effect.concurrent.{Ref, Deferred}
+import cats.effect.concurrent.{Deferred, Ref, Semaphore}
 import cats.effect.internals.{Callback, IORunLoop}
 
 import scala.annotation.implicitNotFound
@@ -318,6 +318,26 @@ object Async {
           .rethrow
       }
     }
+
+  /**
+    * Like `Parallel.parTraverse`, but limits the degree of parallelism.
+    */
+  def parTraverseN[T[_]: Traverse, M[_], F[_], A, B](n: Long)(ta: T[A])(f: A => M[B])(implicit M: Async[M], P: Parallel[M, F]): M[T[B]] =
+    for {
+      semaphore <- Semaphore.uncancelable(n)(M)
+      tb <- ta.parTraverse { a =>
+          semaphore.withPermit(f(a))
+        }
+    } yield tb
+
+  /**
+    * Like `Parallel.parSequence`, but limits the degree of parallelism.
+    */
+  def parSequenceN[T[_]: Traverse, M[_], F[_], A](n: Long)(tma: T[M[A]])(implicit M: Async[M], P: Parallel[M, F]): M[T[A]] =
+    for {
+      semaphore <- Semaphore.uncancelable(n)(M)
+      mta <- tma.map(semaphore.withPermit).parSequence
+    } yield mta
 
   /**
    * [[Async]] instance built for `cats.data.EitherT` values initialized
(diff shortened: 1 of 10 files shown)


sideeffffect:
Is there a way to work around the issue
```scala
[error]  * abstract method catsEffectSyntaxAsyncObj(cats.effect.Async)cats.effect.Async in trait cats.effect.syntax.AsyncSyntax is inherited by class AllCatsEffectSyntax in current version.
[error]    filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("cats.effect.syntax.AllCatsEffectSyntax.catsEffectSyntaxAsyncObj")
[error]  * abstract method toAsyncOps(java.lang.Object,cats.effect.Async)cats.effect.Async#Ops in trait cats.effect.Async#ToAsyncOps is inherited by class AllCatsEffectSyntax in current version.
[error]    filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("cats.effect.syntax.AllCatsEffectSyntax.toAsyncOps")
[error] java.lang.RuntimeException: cats-effect: Binary compatibility check failed!
```
that happens only with Scala 2.11?

core/shared/src/main/scala/cats/effect/Async.scala
```diff
@@ -319,6 +319,26 @@ object Async {
       }
     }
 
+  /**
+    * Like `Parallel.parTraverse`, but limits the degree of parallelism.
+    */
+  def parTraverseN[T[_]: Traverse, M[_], F[_], A, B](n: Long)(ta: T[A])(f: A => M[B])(implicit M: Async[M], P: Parallel[M, F]): M[T[B]] =
+    for {
+      semaphore <- Semaphore.uncancelable(n)(M)
```
SystemFw: Why are you using `uncancelable` here? That strikes me as a bug

sideeffffect: I can work with mere `Async` and don't need `Concurrent`. Would you suggest to change that to `Semaphore.apply`?

SystemFw: Yes

sideeffffect:
@SystemFw any tips for solving the problem with Scala 2.11?

djspiewak:
@sideeffffect It can't *really* be resolved on Scala 2.11. What you're doing here is you're adding functions to a trait (`AllCatsEffectSyntax`) which previously existed. Anyone inheriting from that trait is going to get that function. The problem with this is that, in Scala 2.11 (or more specifically, prior to Java 8), traits were encoded through explicit function delegation to a nested class (which contained the definition) in the implementing subtypes. This is problematic because the implementing subtypes *cannot* delegate to the nested class on these new functions (such as `parTraverseN`) since it didn't exist when they were compiled! (since they were compiled against previous versions of cats-effect) So any code which calls `parTraverseN` would then get an `AbstractMethodError`. This is what MiMa is detecting.

You basically cannot add syntax like this in the 1.x timeline. Or rather, you could add `AsyncSyntax`, but you can't add it to the main trait. The good news is that 2.x is right around the corner. The bad news is we really don't want to change *too* much, so I'm not sure if there's time for this to make it.

rossabaker:
Is it any worse for compatibility than the changes proposed in #262?

djspiewak:
@rossabaker Do you mean #519? Broadly-speaking no, it should be about the same if I fully comprehend both PRs.

djspiewak:
@sideeffffect I would recommend waiting for #546 to land (which flushes out mima), then merge with master (please don't rebase open PRs) and see if that resolves the issues.

djspiewak:
@sideeffffect Any update on this? We're going to be in compatibility-freeze very shortly.

sideeffffect:
@djspiewak rebased on the newest master
OK to merge?

djspiewak:
@sideeffffect Reminder: as I said, do not rebase PRs. It's in the contributing guidelines and everything. :-)

I'll take a final look at things and merge (hopefully later today). Thanks for all your work!

sideeffffect:
@djspiewak *facepalm* sorry, I totally missed that
thanks for looking at it!

PR #374, opened 2018-09-24
Properly propogate Writer and State effects in bracketCase

LukaJCB:
Should fix #371.

With this change:
```scala
scala> WriterT(IO(1 -> "asd"))
res0: cats.data.WriterT[cats.effect.IO,Int,String] = WriterT(IO$2102705681)

scala> Sync[WriterT[IO, Int, ?]].bracket(res0)(s => WriterT(IO(2 -> "2")))(s => WriterT(IO( (7, ()) )))
res1: cats.data.WriterT[cats.effect.IO,Int,String] = WriterT(IO$1115955315)

scala> res1.run
res2: cats.effect.IO[(Int, String)] = IO$1115955315

scala> res2.unsafeRunSync
res2: (Int, String) = (10,2)
```

Without this change:

```scala
scala> WriterT(IO(1 -> "asd"))
res0: cats.data.WriterT[cats.effect.IO,Int,String] = WriterT(IO$62720203982)

scala> Sync[WriterT[IO, Int, ?]].bracket(res0)(s => WriterT(IO(2 -> "2")))(s => WriterT(IO( (7, ()) )))
res1: cats.data.WriterT[cats.effect.IO,Int,String] = WriterT(IO$31215753453)

scala> res1.run
res2: cats.effect.IO[(Int, String)] = IO$297120624

scala> res2.unsafeRunSync
res2: (Int, String) = (3,2)
```

PR #500, opened 2019-03-25
Remove unnecessary code from Sync[OptionT]

lukasz-golebiewski:


SystemFw:
Thanks for your contribution :)

The code you have removed is not unnecessary, its purpose is to ensure that the transformer effect in the `release` action of a `bracket` actually takes place

https://github.com/typelevel/cats-effect/pull/374/commits/7ca92e1b19245d26dc550a12e62e488378da063b

https://github.com/typelevel/cats-effect/pull/374

lukasz-golebiewski:
I see, thanks for the clarification. Is this tested anywhere?

SystemFw:
I suspect not, in which case adding a test for this scenario (based off the comments on the PR I linked) would be a welcome addition

PR #516, opened 2019-04-26
Clear up Resource docs

### LINKED ISSUE #336: Resource docs are outdated
In https://typelevel.org/cats-effect/datatypes/resource.html

`def allocate: F[(A, F[Unit])]` isn't present in the current definition of Resource:

https://github.com/typelevel/cats-effect/blob/master/core/shared/src/main/scala/cats/effect/Resource.scala#L96

kubukoz:
Closes #336

### DIFF
--- a/site/src/main/tut/datatypes/resource.md
+++ b/site/src/main/tut/datatypes/resource.md
@@ -31,7 +31,7 @@ val greet: String => IO[Unit] = x => IO(println("Hello " ++ x))
 Resource.liftF(IO.pure("World")).use(greet).unsafeRunSync
 ```
 
-Note that resource acquisition cannot be canceled, so if you lift a long-running `F[A]` to `Resource` and cancel the `use` call while it runs,
+Note that resource acquisition cannot be canceled, so if you lift a long running `F[A]` into `Resource` and cancel the `use` call while it runs,
 the program will wait until that `F[A]` completes.
 
 For example, if you do this:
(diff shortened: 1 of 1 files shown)


site/src/main/tut/datatypes/resource.md
```diff
@@ -32,8 +29,22 @@ import cats.implicits._
 val greet: String => IO[Unit] = x => IO(println("Hello " ++ x))
 
 Resource.liftF(IO.pure("World")).use(greet).unsafeRunSync
+```
+
+Note that resource acquisition cannot be canceled, so if you lift a long-running `F[A]` to `Resource` and cancel the `use` call while it runs,
```
rossabaker:
```suggestion
Note that resource acquisition cannot be canceled, so if you lift a long running `F[A]` into `Resource` and cancel the `use` call while it runs,
```

site/src/main/tut/datatypes/resource.md
```diff
@@ -32,8 +29,22 @@ import cats.implicits._
...
+the program will wait until that `F[A]` completes.
+
+For example, if you do this:
 
 ```
+//latch to make sure we cancel the `use` after it's actually started
```
rossabaker: There's nothing clever we can do to make this a tut shed, because the tut would never terminate, right?

kubukoz: I think so :/

djspiewak:
Merge with master?

rossabaker:
If we decide to do a 1.3 in addition to a 2.0 milestone, this should target `series/1.x`.

PR #518, opened 2019-04-27
Switch to xenial and openjdk8

rossabaker:
[openjdk11 builds are broken](https://travis-ci.community/t/install-of-openjdk11-is-failing-again/3061) again on trusty. xenial has it preinstalled, and [trusty is EOL](https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment) on April 30. Let's switch now.

More breaking news: oraclejdk8 doesn't install on xenial, so let's try openjdk8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: xenial
 sudo: false
 language: scala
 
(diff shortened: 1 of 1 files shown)


rossabaker:
So xenial doesn't have oraclejdk8 and trusty doesn't have openjdk11 :fire:

PR #537, opened 2019-05-21
Concurrent.memoize is well behaved under cancelation

### LINKED ISSUE #530: Cancelling the memo'd result of `Concurrent.memoize`
It appears that if you attempt to run a memo'd effect after some other call has canceled it the second attempt hangs indefinitely.  I could not find any issue, or documentation about what is expected to happen in this case.

Reproduction:
```
val ef = Timer[IO].sleep(2.seconds) // Long running IO
val mEF = Concurrent.memoize(ef)
val memoed = mEF.unsafeRunSync  // Get out the memoized version
val fiber = memoed.start
fiber.flatMap(_.cancel).unsafeRunSync

memoed.unsafeRunSync //Hangs indefinitely
```

I'm not sure if this is a bug that needs fixing, or just a opportunity for documentation improvement.

The cause of this specific issue is the deferred never completes: https://github.com/typelevel/cats-effect/blob/706b40f8f87dcf7779fd5926267138b9f2b9ee2e/core/shared/src/main/scala/cats/effect/Concurrent.scala#L368

SystemFw:
Fixes #530

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Concurrent.scala
+++ b/core/shared/src/main/scala/cats/effect/Concurrent.scala
@@ -354,66 +354,64 @@ object Concurrent {
     }
 
   /**
-    * Lazily memoizes `f`. For every time the returned `F[F[A]]` is
-    * bound, the effect `f` will be performed at most once (when the
-     * inner `F[A]` is bound the first time).
+    * Lazily memoizes `f`. Assuming no cancelation happens, the effect
+    * `f` will be performed at most once for every time the returned
+    * `F[F[A]]` is bound (when the inner `F[A]` is bound the first
+    * time).
     *
-    * Note: `start` can be used for eager memoization.
-    * 
-    * add extra semantics description for cancelation
-    * note that you can't really do much if `f` is canceled externally
+    * If you try to cancel an inner `F[A]`, `f` is only interrupted if
+    * there are no other active subscribers, whereas if there are, `f`
+    * keeps running in the background.
+    *
+    * If `f` is successfully canceled, the next time an inner `F[A]`
+    * is bound `f` will be restarted again. Note that this can mean
+    * the effects of `f` happen more than once.
+    *
+    * You can look at `Async.memoize` for a version of this function
+    * which does not allow cancelation.
     */
   def memoize[F[_], A](f: F[A])(implicit F: Concurrent[F]): F[F[A]] = {
     sealed trait State
     case class Subs(n: Int) extends State
     case object Done extends State
 
-    case class Fetch(
-      state: State,
-      v: Deferred[F, Either[Throwable, A]],
-      stop: Deferred[F, F[Unit]]
-    )
+    case class Fetch(state: State, v: Deferred[F, Either[Throwable, A]], stop: Deferred[F, F[Unit]])
 
-    def state =
-      Ref[F].of(Option.empty[Fetch])
+    Ref[F].of(Option.empty[Fetch]).map { state =>
+      (Deferred[F, Either[Throwable, A]] product Deferred[F, F[Unit]]).flatMap {
+        case (v, stop) =>
+          def endState(ec: ExitCase[Throwable]) =
+            state.modify {
+              case None =>
+                throw new AssertionError("unreachable")
+              case s @ Some(Fetch(Done, _, _)) =>
+                s -> F.unit
+              case Some(Fetch(Subs(n), v, stop)) =>
+                if (ec == ExitCase.Canceled && n == 1)
+                  None -> stop.get.flatten
+                else if (ec == ExitCase.Canceled)
+                  Fetch(Subs(n - 1), v, stop).some -> F.unit
+                else
+                  Fetch(Done, v, stop).some -> F.unit
+            }.flatten
 
-    def deferreds =
-      (Deferred[F, Either[Throwable, A]], Deferred[F, F[Unit]]).tupled
+          def fetch =
+            f.attempt
+              .flatMap(v.complete)
+              .start
+              .flatMap(fiber => stop.complete(fiber.cancel))
 
-    state.map { state =>
-      deferreds.flatMap { case (v, stop) =>
-        def endState(ec: ExitCase[Throwable]) =
-          state.modify {
-            case None =>
-              ??? // unreachable
-            case s @ Some(Fetch(Done, _, _)) =>
-              s -> F.unit
-            case Some(Fetch(Subs(n), v, stop)) =>
-              if (ec == ExitCase.Canceled && n == 1)
-                None -> stop.get.flatten
-              else if (ec == ExitCase.Canceled)
-                Fetch(Subs(n - 1), v, stop).some -> F.unit
-              else
-                Fetch(Done, v, stop).some -> F.unit
-          }.flatten
-
-        def fetch =
-          f.attempt
-            .flatMap(v.complete)
-            .start
-            .flatMap(fiber => stop.complete(fiber.cancel))
-
-        state
-          .modify {
-            case s @ Some(Fetch(Done, v, _)) =>
-              s -> v.get
-            case Some(Fetch(Subs(n), v, stop)) =>
-              Fetch(Subs(n + 1), v, stop).some -> v.get.guaranteeCase(endState)
-            case None =>
-              Fetch(Subs(1), v, stop).some -> fetch.bracketCase(_ => v.get) { case (_, ec) => endState(ec) }
-          }
-          .flatten
-          .rethrow
+          state
+            .modify {
+              case s @ Some(Fetch(Done, v, _)) =>
+                s -> v.get
+              case Some(Fetch(Subs(n), v, stop)) =>
+                Fetch(Subs(n + 1), v, stop).some -> v.get.guaranteeCase(endState)
+              case None =>
+                Fetch(Subs(1), v, stop).some -> fetch.bracketCase(_ => v.get) { case (_, ec) => endState(ec) }
+            }
+            .flatten
+            .rethrow
       }
     }
   }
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/effect/Concurrent.scala
```diff
@@ -353,26 +354,67 @@ object Concurrent {
     }
 
   /**
-    * Lazily memoizes `f`. For every time the returned `F[F[A]]` is
-    * bound, the effect `f` will be performed at most once (when the
-    * inner `F[A]` is bound the first time).
+    * Lazily memoizes `f`. Assuming no cancelation happens, the effect
```
rossabaker: Standardized spelling

```suggestion
    * Lazily memoizes `f`. Assuming no cancellation happens, the effect
```

core/shared/src/main/scala/cats/effect/Concurrent.scala
```diff
@@ -353,26 +354,67 @@ object Concurrent {
...
+    * If you try to cancel an inner `F[A]`, `f` is only interrupted if
+    * there are no other active subscribers, whereas if there are, `f`
+    * keeps running in the background.
+    *
+    * If `f` is successfully canceled, the next time an inner `F[A]`
+    * is bound `f` will be restarted again. Note that this can mean
+    * the effects of `f` happen more than once.
+    *
+    * You can look at `Async.memoize` for a version of this function
+    * which does not allow cancelation.
```
rossabaker:
```suggestion
    * which does not allow cancellation.
```

core/shared/src/main/scala/cats/effect/Concurrent.scala
```diff
@@ -353,26 +354,67 @@ object Concurrent {
...
-        ref
-          .modify {
-            case None =>
-              Some(d) -> f.attempt.flatTap(d.complete)
-            case s @ Some(other) =>
-              s -> other.get
-          }
-          .flatten
-          .rethrow
+  def memoize[F[_], A](f: F[A])(implicit F: Concurrent[F]): F[F[A]] = {
```
rossabaker: I wish we'd called this `fa` but I don't care enough to do the `@deprecatedName` dance.

laws/shared/src/test/scala/cats/effect/MemoizeTests.scala
```diff
@@ -92,4 +92,115 @@ class MemoizeTests extends BaseTestsSuite {
...
+    val prog = for {
+      completed <- Ref[IO].of(false)
+      action = IO.sleep(300.millis) >> completed.set(true)
+      memoized <- Concurrent.memoize(action)
+      fiber1 <- memoized.start
+      _ <- IO.sleep(100.millis)
+      fiber2 <- memoized.start
+      _ <- IO.sleep(100.millis)
+      _ <- fiber1.cancel
+      _ <- fiber2.cancel
```
rossabaker: Is the difference between this and (2) the order of these cancellations?

laws/shared/src/test/scala/cats/effect/MemoizeTests.scala
```diff
@@ -92,4 +92,115 @@ class MemoizeTests extends BaseTestsSuite {
...
+      _ <- IO.sleep(100.millis)
+      _ <- fiber.cancel
+      _ <- memoized.timeout(1.second)
+      v1 <- started.get
+      v2 <- completed.get
+    } yield v1 -> v2
+
+    val result = prog.unsafeToFuture()
+    ec.tick(500.millis)
+    result.value shouldBe Some(Success(2 -> 1))
```
rossabaker: Style nit: I prefer to use `->` for associations, not for something you just happen to have two of.

```suggestion
    result.value shouldBe Some(Success((2 , 1)))
```

core/shared/src/main/scala/cats/effect/Concurrent.scala
```diff
@@ -353,26 +354,67 @@ object Concurrent {
...
+
+    case class Fetch(state: State, v: Deferred[F, Either[Throwable, A]], stop: Deferred[F, F[Unit]])
+
+    Ref[F].of(Option.empty[Fetch]).map { state =>
+      (Deferred[F, Either[Throwable, A]] product Deferred[F, F[Unit]]).flatMap {
+        case (v, stop) =>
+          def endState(ec: ExitCase[Throwable]) =
+            state.modify {
+              case None =>
+                throw new AssertionError("unreachable")
```
rossabaker: I agree this is unreachable, but is it unrecoverable?  Would it corrupt things (further) to return `None -> F.unit` and log a loud warning?

core/shared/src/main/scala/cats/effect/Concurrent.scala
djspiewak: *twitches*

core/shared/src/main/scala/cats/effect/Concurrent.scala
djspiewak: Any opinions on whether or not we should consider such things valid for our fast-and-loose 2.0 compatibility? I doubt anyone is calling `memoize` with by-name parameters, so in practice this is almost certainly source compatible, but I'm very open to being convinced otherwise.

core/shared/src/main/scala/cats/effect/Concurrent.scala
djspiewak: I'm leery about effect types logging. I guess we do it already though in a couple other places?

laws/shared/src/test/scala/cats/effect/MemoizeTests.scala
```diff
@@ -92,4 +92,115 @@ class MemoizeTests extends BaseTestsSuite {
...
+    implicit val timer = ec.timer[IO]
+
+    val prog = for {
+      completed <- Ref[IO].of(false)
+      action = IO.sleep(200.millis) >> completed.set(true)
+      memoized <- Concurrent.memoize(action)
+      fiber <- memoized.start
+      _ <- IO.sleep(100.millis)
+      _ <- fiber.cancel
+      _ <- IO.sleep(300.millis)
```
djspiewak: Rather than sleeping, can we use `Deferred[Unit]` here? It will yield a much more robust enforced sequencing mechanism.

laws/shared/src/test/scala/cats/effect/MemoizeTests.scala
```diff
@@ -92,4 +92,115 @@ class MemoizeTests extends BaseTestsSuite {
...
+      completed <- Ref[IO].of(false)
+      action = IO.sleep(300.millis) >> completed.set(true)
+      memoized <- Concurrent.memoize(action)
+      fiber1 <- memoized.start
+      _ <- IO.sleep(100.millis)
+      fiber2 <- memoized.start
+      _ <- IO.sleep(100.millis)
+      _ <- fiber2.cancel
+      _ <- fiber1.cancel
+      _ <- IO.sleep(400.millis)
```
djspiewak: …and here

laws/shared/src/test/scala/cats/effect/MemoizeTests.scala
```diff
@@ -92,4 +92,115 @@ class MemoizeTests extends BaseTestsSuite {
...
+      action = IO.sleep(300.millis) >> completed.set(true)
+      memoized <- Concurrent.memoize(action)
+      fiber1 <- memoized.start
+      _ <- IO.sleep(100.millis)
+      fiber2 <- memoized.start
+      _ <- IO.sleep(100.millis)
+      _ <- fiber1.cancel
+      _ <- fiber2.cancel
+      _ <- IO.sleep(400.millis)
+      res <- completed.get
```
djspiewak: …and here

laws/shared/src/test/scala/cats/effect/MemoizeTests.scala
```diff
@@ -92,4 +92,115 @@ class MemoizeTests extends BaseTestsSuite {
...
+    } yield res
+
+    val result = prog.unsafeToFuture()
+    ec.tick(600.millis)
+    result.value shouldBe Some(Success(false))
+  }
+
+    testAsync("Memoized effects can be canceled when there are no other active subscribers (3)") { implicit ec =>
```
djspiewak: Indentation

laws/shared/src/test/scala/cats/effect/MemoizeTests.scala
```diff
@@ -92,4 +92,115 @@ class MemoizeTests extends BaseTestsSuite {
...
+   implicit val cs = ec.contextShift[IO]
+   implicit val timer = ec.timer[IO]
+   
+   val prog = for {
+     condition <- Deferred[IO, Unit]
+     action = IO.sleep(200.millis) >> condition.complete(())
+     memoized <- Concurrent.memoize(action)
+     fiber1 <- memoized.start
+     _ <- IO.sleep(50.millis)
+     fiber2 <- memoized.start
```
djspiewak: …and here

core/shared/src/main/scala/cats/effect/Concurrent.scala
rossabaker: :us: :fireworks: :statue_of_liberty:

core/shared/src/main/scala/cats/effect/Concurrent.scala
rossabaker: I'm okay changing it, but if we do, we should scan for others.

I was going to be polite and throw in a `@deprecatedName`, but Scala 2.13 deprecates the parameters to Scala 2.12, rendering it mostly useful to Scala itself. :man_shrugging:

core/shared/src/main/scala/cats/effect/Concurrent.scala
rossabaker: I'm leery about effect types logging but I'm leerier about applications crashing with fatal errors for cases that we know how to recover from.

laws/shared/src/test/scala/cats/effect/MemoizeTests.scala
SystemFw: Yes, in one case the fiber canceling is the same that started the computation, whereas in the other there is a sort of "transfer" of the "ownership" of being the last subscriber remaining. A previous implementation I had (which was a bit simpler) worked in the former but not latter case

SystemFw:
For posterity: the tests with `sleep` are actually deterministic because of `TextContext` stepping things through,  and easier to set up than a bunch of "happens before" through Deferred

PR #540, opened 2019-05-24
Use sbt-tpolecat, fix ensuing warnings and bug

rossabaker:
There was a subtle issue in local development, where the root project's default Scala version is 2.13, preventing a handful of compiler flags that we enable on 2.12 from being set.  Travis CI enforces them, but it's unnecessary inconsistency.

In the process of cleaning that up, I got more aggressive and turned on the sbt-tpolecat set of flags.  This identified a bug in JS IOApp exit code on signals, and an unused variable in the laws.


build.sbt
```diff
@@ -68,6 +60,9 @@ val commonSettings = Seq(
   scalacOptions in (Compile, doc) ++=
     Opts.doc.title("cats-effect"),
 
+  scalacOptions in Test += "-Yrangepos",
```
rossabaker: This is not set by sbt-tpolecat.  I don't know why we care, but I preserved it.

build.sbt
```diff
@@ -68,6 +60,9 @@ val commonSettings = Seq(
...
+  scalacOptions in Test ~= (_.filterNot(Set("-Wvalue-discard", "-Ywarn-value-discard"))),
```
rossabaker: There are a couple hundred warnings in the tests without this, because we blur the line between `Assertion` and `Unit`.  I think this is low value in tests, but we might find an unrun IO or something.  Pausing to see what the committee thinks.

build.sbt
```diff
@@ -35,19 +35,11 @@ val DisciplineVersion = "0.11.2-M1"
...
-  crossScalaVersions := Seq("2.11.12", "2.12.8", "2.13.0-RC1"),
+scalaVersion in ThisBuild := "2.12.8"
 
-  scalacOptions ++= PartialFunction.condOpt(CrossVersion.partialVersion(scalaVersion.value)) {
-    case Some((2, n)) if n >= 13 =>
-      Seq(
-        "-Ymacro-annotations"
-      )
-  }.toList.flatten,
+crossScalaVersions in ThisBuild := Seq("2.11.12", "2.12.8", "2.13.0-RC1")
```
rossabaker: We don't need these lines because we have sbt-travisci, but then we want to move our "default" Scala version to the bottom of .travis.yml.  Alternatively, this is replicating everything we get from sbt-travisci, and we could eliminate that plugin.

We need it in `ThisBuild` or else sbt-travisci stomps on it for the root project, which is how we started down this road in the first place.

core/js/src/main/scala/cats/effect/internals/IOAppPlatform.scala
```diff
@@ -82,7 +82,7 @@ private[effect] object IOAppPlatform {
     def handler(code: Int) = () =>
       fiber.cancel.unsafeRunAsync { result =>
         result.swap.foreach(Logger.reportFailure)
-        IO(sys.exit(code + 128))
+        sys.exit(code + 128)
```
rossabaker: This was the bug: killing a JS IOApp resulted in success, not the traditional signal+128.

core/shared/src/main/scala/cats/effect/syntax/BracketSyntax.scala
```diff
@@ -20,9 +20,11 @@ import cats.effect.{Bracket, ExitCase}
 
 
 trait BracketSyntax {
-  // Bracket instance here is required to ensure correct inference for E
-  implicit def catsEffectSyntaxBracket[F[_], A, E](fa: F[A])(implicit bracket: Bracket[F, E]): BracketOps[F, E, A] =
+  implicit def catsEffectSyntaxBracket[F[_], A, E](fa: F[A])(implicit bracket: Bracket[F, E]): BracketOps[F, E, A] = {
+    // Bracket instance here is required to ensure correct inference for E
```
rossabaker: The tests compiled without this, though I doubt we test this syntax much, and I'm inclined to trust the comment.  Removing it would also be a binary breaking change in a module we're staying conservative on for 2.0.

laws/jvm/src/test/scala/cats/effect/ResourceJVMTests.scala
```diff
@@ -21,15 +21,15 @@ import javax.security.auth.Destroyable
 
 class ResourceJVMTests extends BaseTestsSuite {
   test("resource from Destroyable is auto destroyed") {
+    var destroyed = false
```
rossabaker: This pattern, which I've done in a few other places, is to work around a warning about reflection.  We would be sad if this appeared in a test that made multiple instances of the anonymous class.

laws/shared/src/main/scala/cats/effect/laws/ConcurrentEffectLaws.scala
```diff
@@ -33,7 +33,7 @@ trait ConcurrentEffectLaws[F[_]] extends ConcurrentLaws[F] with EffectLaws[F] {
     fa1 <-> fa2
   }
 
-  def runCancelableIsSynchronous[A](fa: F[A]) = {
+  def runCancelableIsSynchronous[A] = {
```
rossabaker: Breaking change to the law, but it doesn't depend on an `F[A]`.

project/plugins.sbt
```diff
@@ -4,7 +4,7 @@ addSbtPlugin("com.typesafe"       % "sbt-mima-plugin"         % "0.3.0")
 addSbtPlugin("com.typesafe.sbt"   % "sbt-git"                 % "1.0.0")
 addSbtPlugin("de.heikoseeberger"  % "sbt-header"              % "5.2.0")
 addSbtPlugin("io.get-coursier"    % "sbt-coursier"            % "1.0.0")
-addSbtPlugin("org.lyranthe.sbt"   % "partial-unification"     % "1.1.2")
+addSbtPlugin("io.github.davidgregory084" % "sbt-tpolecat"     % "0.1.6")
```
rossabaker: We're overdue a Grand Realignment Of The Percent Operators.  I'll throw it in here if people want it.

build.sbt
DavidGregory084: Ah, I think this is needed for Metals :+1:

core/shared/src/main/scala/cats/effect/ConcurrentEffect.scala
```diff
@@ -93,7 +93,10 @@ object ConcurrentEffect {
 
     override def runCancelable[A](fa: EitherT[F, Throwable, A])
       (cb: Either[Throwable, A] => IO[Unit]): SyncIO[CancelToken[EitherT[F, Throwable, ?]]] =
-      F.runCancelable(fa.value)(cb.compose(_.right.flatMap(x => x))).map(EitherT.liftF(_)(F))
+      F.runCancelable(fa.value)(cb.compose(_ match {
+        case Right(x) => x
+        case left => left.asInstanceOf[Either[Throwable, A]]
```
rossabaker: This was necessary because of the deprecation of `_.right` in 2.13.  Alternatives:

* Importing `cats.implicits._` triggers an unused import warning in 2.12
* Turn off fatal warnings in 2.13.  I'd rather not.
* An internal cats-effect syntax, but there are only four of these.  
* Import `cats.implicits._` and gratuitously assign one of its members so it's used on all Scala versions.

I dislike all these options, but there are only four of them, so I inlined them.  We can clean this up when we drop Scala 2.11.

rossabaker:
I think [this TimerTest](https://travis-ci.org/typelevel/cats-effect/jobs/537052431#L8712) is an intermittent flicker. :frowning:

build.sbt
djspiewak: I would just rather move the default to the bottom in the yaml definition and delete this line.

core/js/src/main/scala/cats/effect/internals/IOAppPlatform.scala
djspiewak: warn-on-discard should just be the default. It's literally *never* a semantic that you want.

core/shared/src/main/scala/cats/effect/syntax/BracketSyntax.scala
djspiewak: Did they compile on 2.11?

I've actually been bitten by this spurious constraint a couple times. It's definitely odd. I wouldn't mind removing it for 2.0 if we're sure the inference works out, but I'm not convinced it will.

core/shared/src/main/scala/cats/effect/ConcurrentEffect.scala
djspiewak: Yeah I hate this particular situation. The juggle of `Either` shenanigans between 2.11, 2.12, and 2.13 is the *worst* if you want fatal warnings, and we do.

djspiewak:
@rossabaker How intermittent do you think the flicker is? I tend to prefer to be extremely putative with intermittent CI failures. I'd rather not have a test at all in that case (ideally with an issue to fix it later) than have an unreliable CI.

core/shared/src/main/scala/cats/effect/syntax/BracketSyntax.scala
rossabaker: It does, and is used in `SyntaxTests`.  But I'm not sure that covers whatever case made someone add this in the first place.

core/shared/src/main/scala/cats/effect/ConcurrentEffect.scala
rossabaker: Took the old fs2 approach by putting syntax in the root package object.  It's not considered unused in 2.12 and 2.13, and it's available where needed in 2.11.  We can delete scala211EitherSyntax when Scala 2.11 support is dropped and not have hacks dispersed throughout the code.

Issue #545, opened 2019-05-26
Several laws use Future and IO


djspiewak:
I would rather the laws not depend on either of these (particularly since one implies the other). This is probably a blocker for #480, and perhaps other 3.0 things. It's also worth noting that some of the use of `IO` in these places makes pretty significant assumptions about the underlying `F`'s execution model. For example:

```scala
cancel  <- F.liftIO(F.runCancelable(never)(_ => IO.unit).toIO)
```

I'm pretty sure this is effectively just calling `start`, but obliquely via `runAsync`.

PR #546, opened 2019-05-26
Enforce shifting within IO.fromFuture

### LINKED ISSUE #444: Unexpected execution context after calling IO.fromFuture
I was surprised to find that after calling `IO.fromFuture`, computation is *sometimes* continued on the `ExecutionContext` that the `Future` was run on.

```scala
import cats.effect._
import scala.concurrent._
import scala.concurrent.ExecutionContext.global

// Slow Future
IO.fromFuture(IO(Future(Thread.sleep(500))(global)))
  .flatMap(_ => IO.delay(println(Thread.currentThread.getName)))
  .unsafeRunSync()

// Fast Future
IO.fromFuture(IO(Future(())(global)))
  .flatMap(_ => IO.delay(println(Thread.currentThread.getName)))
  .unsafeRunSync()
```

The above prints something like:
```
scala-execution-context-global-14
main
```

I looked through the commits, and there used to [be a notice concerning this](https://github.com/typelevel/cats-effect/commit/ba6e4510d82f25ba5d6c24df9910bbc919511f84#diff-c01f002766a28f49dfe20d6b1f0afd50R747), but it was removed at some point.

I feel that the current behavior is a bit scary. I ran into it while using Slick, which maintains its own `ExecutionContext` for blocking operations. I was surprised when I suddenly saw some of my code executing there. I got around it by always shifting back after calling `IO.fromFuture`, but it took a bit of digging to find out what was happening. 

I was wondering if this is the intended behavior. If it is, maybe the original notice should be restored?

djspiewak:
This is probably the most serious bit of compatibility breakage we're going to introduce in 2.0, since it can actually break source compatibility by forcing people to thread a `ContextShift[IO]` where one was not previously necessary. With that said, any cases where people *weren't* already `shift`ing following a `fromFuture` were probably bugs, so I'm more or less okay with that.

I hate the fact that I had to use `IOFromFuture` in the laws, but I'm not sure it's fundamentally worse than the fact that we were already using `IO` in the laws. I filed #545 to track that, as it will need to be excised in 3.0.

Fixes #444

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -157,7 +157,7 @@ val mimaSettings = Seq(
   mimaPreviousArtifacts := {
     CrossVersion.partialVersion(scalaVersion.value) match {
       case Some((2, 13)) => Set.empty
-      case _ => Set(organization.value %% name.value % "1.0.0")
+      case _ => Set.empty   // TODO put 2.0.0 here
     }
   },
   mimaBinaryIssueFilters ++= {
@@ -175,10 +175,7 @@ val mimaSettings = Seq(
       // Laws - https://github.com/typelevel/cats-effect/pull/473
       exclude[ReversedMissingMethodProblem]("cats.effect.laws.AsyncLaws.repeatedAsyncFEvaluationNotMemoized"),
       exclude[ReversedMissingMethodProblem]("cats.effect.laws.BracketLaws.bracketPropagatesTransformerEffects"),
-      exclude[ReversedMissingMethodProblem]("cats.effect.laws.discipline.BracketTests.bracketTrans"),
-      exclude[ReversedMissingMethodProblem]("cats.effect.Bracket.onCancel"),
-      exclude[ReversedMissingMethodProblem]("cats.effect.Concurrent.continual"),
-      exclude[ReversedMissingMethodProblem]("cats.effect.Concurrent#Ops.continual")
+      exclude[ReversedMissingMethodProblem]("cats.effect.laws.discipline.BracketTests.bracketTrans")
     )
   })
 
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -1200,8 +1200,8 @@ object IO extends IOInstances {
    *
    * @see [[IO#unsafeToFuture]]
    */
-  def fromFuture[A](iof: IO[Future[A]]): IO[A] =
-    iof.flatMap(IOFromFuture.apply)
+  def fromFuture[A](iof: IO[Future[A]])(implicit cs: ContextShift[IO]): IO[A] =
+    iof.flatMap(IOFromFuture.apply).flatMap(a => cs.shift.map(_ => a))
 
   /**
    * Lifts an `Either[Throwable, A]` into the `IO[A]` context, raising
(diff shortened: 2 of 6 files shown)


build.sbt
```diff
@@ -157,7 +157,7 @@ val mimaSettings = Seq(
   mimaPreviousArtifacts := {
     CrossVersion.partialVersion(scalaVersion.value) match {
       case Some((2, 13)) => Set.empty
-      case _ => Set(organization.value %% name.value % "1.0.0")
+      case _ => Set.empty   // TODO put 2.0.0 here
```
rossabaker: This opens the floodgates., though arguably so does this PR.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1200,8 +1200,8 @@ object IO extends IOInstances {
    *
    * @see [[IO#unsafeToFuture]]
    */
-  def fromFuture[A](iof: IO[Future[A]]): IO[A] =
-    iof.flatMap(IOFromFuture.apply)
+  def fromFuture[A](iof: IO[Future[A]])(implicit cs: ContextShift[IO]): IO[A] =
```
rossabaker: I'm trying to think of a new name for this function, so we can deprecate this and preserve our guarantees. `liftFuture`?

build.sbt
djspiewak: It does open the floodgates. We had kind of already decided what would and would not be breaking though, so I'm okay taking the training wheels off. It's good to be skittish about it though.

core/shared/src/main/scala/cats/effect/IO.scala
djspiewak: `liftFuture` isn't bad. The thing I'm worried about though is exactly what I wrote in the description: anyone who is using `IO.fromFuture` who doesn't *already* have a `ContextShift[IO]` in scope (for use in an immediate follow-up `shift`) has a bug that they haven't uncovered. This change, breaking as it is, takes that silent bug and makes it a compile error.

rossabaker: I tend to assume `-deprecation` is on and fatal, though neither is the default.  I guess we can rationalize this as fixing a buggy signature.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1200,8 +1200,8 @@ object IO extends IOInstances {
...
+    iof.flatMap(IOFromFuture.apply).flatMap(a => cs.shift.map(_ => a))
```
kubukoz: @djspiewak have you thought about using `guarantee` instead of `flatMap`? In the current shape it won't shift after failures, will it?

SystemFw: `guarantee` does not work because you cannot access `a`. This is a use case where `continual` would be useful

SystemFw: or maybe just make the thing `uncancelable` given that Futures are not cancelable anyway, that would be even better

djspiewak: Actually `guarantee` works because it's already a `tap`, so we don't really need the `a`.

PR #554, opened 2019-06-07
Add Invariant instance for Ref

kubukoz:


djspiewak:
Very cool.

kubukoz:
To be honest, we could do with a couple tests of the methods other than `get` after calling `imap`. I'll try to deliver some time soon.

Issue #321, opened 2018-08-27
Cats Effect 3.0 Design Discussion


jdegoes:
For Cats Effect 2.0, I'd propose the following:

1. Refocus Cats Effect 2.0 on enabling interop via a common set of type classes, together with any essential, core building blocks that can be expressed in terms of these type classes (e.g. `Ref`, etc.).
2. Move `IO` into `cats-io`, with a separate (but initially overlapping) set of maintainers, and its own independent release schedule. Those so interested can enhance it with other utilities like linebacker, scheduling, etc. Importantly, `cats-io` would depend on `cats-effect`, not the other way around.
3. Redesign the hierarchy based on everything we've learned from 1.x. I've had discussions with many people on this and can summarize them if useful.
4. Support bifunctor and monofunctor type classes (with an `EitherT` adapter?), in synchronous and asynchronous variations (@mpilquist has convincing use cases for synchronous variants), in a clean and minimalist hierarchy that works across Monix Task, Cats IO, and ZIO. The bifunctor type classes should be truly bifunctor (`F[_, _]`).

These steps would, in my opinion, take what's good about Cats Effects and make it better, while reducing maintenance burden, supporting the whole FP ecosystem, and solving well-known problems with the current design.

djspiewak:
> 1. Refocus Cats Effect 2.0 on enabling interop via a common set of type classes, together with any essential, core building blocks that can be expressed in terms of these type classes (e.g. Ref, etc.).
> 
> 2. Move IO into cats-io, with a separate (but initially overlapping) set of maintainers, and its own independent release schedule. Those so interested can enhance it with other utilities like linebacker, scheduling, etc. Importantly, cats-io would depend on cats-effect, not the other way around.

These are kind of the same point, so I'll address them together.

There were two original goals behind marrying the typeclass hierarchy and the `IO` implementation. The first was simply convenience, thus easing adoption and lowering as many barriers as possible that might be in the way of people who were then using `scalaz.concurrent.Task`. The second reasoning was a fast-and-loose parametricity "rooting" concept.

The idea was simple: `_root_.cats.effect.IO` is the one magic type which represents a container for side-effects, and all other types which purport to have independent mechanisms for managing side-effects express such by showing isomorphism with `cats.effect.IO`. This is extremely loose reasoning, but the thought was that it is better than nothing (which is to say, no nominative characterization of what it means to be a side-effect-y datatype).

To be clear, that reasoning made its appearance in the original form of `Effect`:

```scala
[...]

PR #556, opened 2019-06-08
Added Blocker, a new type for blocking execution contexts

### LINKED ISSUE #555: Newtype blocking execution contexts
I'd really like a newtype of `ExecutionContext` that indicates the context is to be used for blocking tasks. Additionally, the newtype could provide good default constructors and convenience methods. I'm roughly thinking of something like this:

```scala
package cats.effect

import java.util.concurrent.{ExecutorService, Executors}

final class Blocker private (context: ExecutionContext) {

  def delay[F[_], A](thunk: => A)(implicit F: Sync[F], cs: ContextShift[F]): F[A] =
    block(F.delay(thunk))

  def block[F[_], A](fa: F[A])(implicit cs: ContextShift[F]): F[A] =
    cs.evalOn(context)(fa)
}

object Blocker {

  def apply[F[_]](implicit F: Sync[F]): Resource[F, Blocker] =
    fromExecutorService(F.delay(Executors.newCachedThreadPool()))

  def fromExecutorService[F[_]](makeExecutorService: F[ExecutorService])(
      implicit F: Sync[F]): Resource[F, Blocker] =
    Resource
      .make(makeExecutorService)(ec => F.delay { ec.shutdownNow(); () })
      .map(es => unsafeFromExecutionContext(ExecutionContext.fromExecutorService(es)))

  def unsafeFromExecutionContext(ec: ExecutionContext): Blocker = new Blocker(ec)
}
```

I believe this type would help prevent folks from using the wrong execution context for blocking operations. I'm somewhat in favor of not including `unsafeFromExecutionContext` at all, though I can see how that would be annoying.

It would be really nice if this could get in to the 2.0.0 release. It might change or be removed for 3.0 but that's okay -- I see it as a companion to the current `ContextShift` type class.

`Blocker` definitely treads on the ground covered by linebacker, though I think it's better to keep the newtype unaware of the type constructor you might use with it.

Thoughts? If there's interest, I can prepare a PR.

mpilquist:
Fixes #555

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Blocker.scala
+++ b/core/shared/src/main/scala/cats/effect/Blocker.scala
@@ -17,8 +17,10 @@
 package cats
 package effect
 
+import scala.collection.JavaConverters._
 import scala.concurrent.ExecutionContext
 import java.util.concurrent.{ExecutorService, Executors, ThreadFactory}
+import cats.data.NonEmptyList
 
 /**
  * An execution context that is safe to use for blocking operations.
@@ -59,18 +61,28 @@ object Blocker {
    * Creates a blocker backed by the `ExecutorService` returned by the
    * supplied task. The executor service is shut down upon finalization
    * of the returned resource.
+   * 
+   * If there are pending tasks in the thread pool at time the returned
+   * `Blocker` is finalized, the finalizer fails with a `Blocker.OutstandingTasksAtShutdown` 
+   * exception.
    */
-  def fromExecutorService[F[_]](makeExecutorService: F[ExecutorService])(
-      implicit F: Sync[F]): Resource[F, Blocker] =
-    Resource
-      .make(makeExecutorService)(ec => F.delay { ec.shutdownNow(); () })
-      .map(es => unsafeFromExecutorService(es))
+  def fromExecutorService[F[_]](makeExecutorService: F[ExecutorService])(implicit F: Sync[F]): Resource[F, Blocker] =
+    Resource.make(makeExecutorService) { ec =>
+        val tasks = F.delay {
+          val tasks = ec.shutdownNow().asScala.toList
+          NonEmptyList.fromList(tasks)
+        }
+        F.flatMap(tasks) {
+          case Some(t) => F.raiseError(new OutstandingTasksAtShutdown(t))
+          case None => F.unit
+        }
+    }.map(es => wrapExecutorService(es))
 
   /**
    * Creates a blocker that delegates to the supplied executor service.
    */
-  def unsafeFromExecutorService(es: ExecutorService): Blocker =
-    unsafeFromExecutionContext(ExecutionContext.fromExecutorService(es))
+  def wrapExecutorService(es: ExecutorService): Blocker =
+    wrapExecutionContext(ExecutionContext.fromExecutorService(es))
 
   /**
    * Creates a blocker that delegates to the supplied execution context.
@@ -78,5 +90,8 @@ object Blocker {
    * This must not be used with general purpose contexts like
    * `scala.concurrent.ExecutionContext.Implicits.global'.
    */
-  def unsafeFromExecutionContext(ec: ExecutionContext): Blocker = new Blocker(ec)
+  def wrapExecutionContext(ec: ExecutionContext): Blocker = new Blocker(ec)
+
+  /** Thrown if there are tasks queued in the thread pool at the time a `Blocker` is finalized. */
+  final class OutstandingTasksAtShutdown(val tasks: NonEmptyList[Runnable]) extends IllegalStateException("There were outstanding tasks at time of shutdown of the thread pool")
 }
--- a/core/shared/src/main/scala/cats/effect/Blocker.scala
+++ b/core/shared/src/main/scala/cats/effect/Blocker.scala
@@ -1,0 +1,76 @@
+/*
+ * Copyright (c) 2017-2019 The Typelevel Cats-effect Project Developers
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats
+package effect
+
+import scala.concurrent.ExecutionContext
+import java.util.concurrent.{ExecutorService, Executors, ThreadFactory}
+
+/**
+ * An execution context that is safe to use for blocking operations.
+ *
+ * Used in conjunction with [[ContextShift]], this type allows us to write functions
+ * that require a special `ExecutionContext` for evaluation, while discouraging the
+ * use of a shared, general purpose pool (e.g. the global context).
+ *
+ * Instances of this class should *not* be passed implicitly.
+ */
+final class Blocker private (val context: ExecutionContext) {
+
+  /**
+   * Like `Sync#delay` but the supplied thunk is evaluated on the blocking
+   * execution context.
+   */
+  def delay[F[_], A](thunk: => A)(implicit F: Sync[F], cs: ContextShift[F]): F[A] =
+    eval(F.delay(thunk))
+
+  /**
+   * Evaluates the supplied task on the blocking execution context via `evalOn`.
+   */
+  def eval[F[_], A](fa: F[A])(implicit cs: ContextShift[F]): F[A] =
+    cs.evalOn(context)(fa)
+}
+
+object Blocker {
+
+  /**
+   * Creates a blocker that is backed by a cached thread pool.
+   */
+  def apply[F[_]](implicit F: Sync[F]): Resource[F, Blocker] =
+    fromExecutorService(F.delay(Executors.newCachedThreadPool(new ThreadFactory {
+      def newThread(r: Runnable) = new Thread(r, "cats-effect-blocker")
+    })))
+
+  /**
+   * Creates a blocker backed by the `ExecutorService` returned by the
+   * supplied task. The executor service is shut down upon finalization
+   * of the returned resource.
+   */
+  def fromExecutorService[F[_]](makeExecutorService: F[ExecutorService])(
+      implicit F: Sync[F]): Resource[F, Blocker] =
+    Resource
+      .make(makeExecutorService)(ec => F.delay { ec.shutdownNow(); () })
+      .map(es => unsafeFromExecutionContext(ExecutionContext.fromExecutorService(es)))
+
+  /**
+   * Creates a block that delegates to the supplied execution context.
+   * 
+   * This must not be used with general purpose contexts like
+   * `scala.concurrent.ExecutionContext.Implicits.global'.
+   */
+  def unsafeFromExecutionContext(ec: ExecutionContext): Blocker = new Blocker(ec)
+}
--- a/core/shared/src/main/scala/cats/effect/Blocker.scala
+++ b/core/shared/src/main/scala/cats/effect/Blocker.scala
@@ -43,7 +43,7 @@ final class Blocker private (val blockingContext: ExecutionContext) {
    * Evaluates the supplied task on the blocking execution context via `evalOn`.
    */
   def evalOn[F[_], A](fa: F[A])(implicit cs: ContextShift[F]): F[A] =
-    cs.evalOn(blockingContext)(fa)
+    cs.evalOn(this)(fa)
 }
 
 object Blocker {
--- a/core/shared/src/main/scala/cats/effect/Blocker.scala
+++ b/core/shared/src/main/scala/cats/effect/Blocker.scala
@@ -17,7 +17,6 @@
 package cats
 package effect
 
-import scala.collection.JavaConverters._
 import scala.concurrent.ExecutionContext
 import java.util.concurrent.{ExecutorService, Executors, ThreadFactory}
 import cats.data.NonEmptyList
--- a/core/shared/src/main/scala/cats/effect/Blocker.scala
+++ b/core/shared/src/main/scala/cats/effect/Blocker.scala
@@ -64,10 +64,16 @@ object Blocker {
       implicit F: Sync[F]): Resource[F, Blocker] =
     Resource
       .make(makeExecutorService)(ec => F.delay { ec.shutdownNow(); () })
-      .map(es => unsafeFromExecutionContext(ExecutionContext.fromExecutorService(es)))
+      .map(es => unsafeFromExecutorService(es))
 
   /**
-   * Creates a block that delegates to the supplied execution context.
+   * Creates a blocker that delegates to the supplied executor service.
+   */
+  def unsafeFromExecutorService(es: ExecutorService): Blocker =
+    unsafeFromExecutionContext(ExecutionContext.fromExecutorService(es))
+
+  /**
+   * Creates a blocker that delegates to the supplied execution context.
    * 
    * This must not be used with general purpose contexts like
    * `scala.concurrent.ExecutionContext.Implicits.global'.
--- a/core/shared/src/main/scala/cats/effect/Blocker.scala
+++ b/core/shared/src/main/scala/cats/effect/Blocker.scala
@@ -31,20 +31,20 @@ import cats.data.NonEmptyList
  *
  * Instances of this class should *not* be passed implicitly.
  */
-final class Blocker private (val context: ExecutionContext) {
+final class Blocker private (val blockingContext: ExecutionContext) {
 
   /**
    * Like `Sync#delay` but the supplied thunk is evaluated on the blocking
    * execution context.
    */
   def delay[F[_], A](thunk: => A)(implicit F: Sync[F], cs: ContextShift[F]): F[A] =
-    eval(F.delay(thunk))
+    evalOn(F.delay(thunk))
 
   /**
    * Evaluates the supplied task on the blocking execution context via `evalOn`.
    */
-  def eval[F[_], A](fa: F[A])(implicit cs: ContextShift[F]): F[A] =
-    cs.evalOn(context)(fa)
+  def evalOn[F[_], A](fa: F[A])(implicit cs: ContextShift[F]): F[A] =
+    cs.evalOn(blockingContext)(fa)
 }
 
 object Blocker {
@@ -54,7 +54,11 @@ object Blocker {
    */
   def apply[F[_]](implicit F: Sync[F]): Resource[F, Blocker] =
     fromExecutorService(F.delay(Executors.newCachedThreadPool(new ThreadFactory {
-      def newThread(r: Runnable) = new Thread(r, "cats-effect-blocker")
+      def newThread(r: Runnable) = {
+        val t = new Thread(r, "cats-effect-blocker")
+        t.setDaemon(true)
+        t
+      }
     })))
 
   /**
@@ -69,20 +73,24 @@ object Blocker {
   def fromExecutorService[F[_]](makeExecutorService: F[ExecutorService])(implicit F: Sync[F]): Resource[F, Blocker] =
     Resource.make(makeExecutorService) { ec =>
         val tasks = F.delay {
-          val tasks = ec.shutdownNow().asScala.toList
-          NonEmptyList.fromList(tasks)
+          val tasks = ec.shutdownNow()
+          val b = List.newBuilder[Runnable]
+          val itr = tasks.iterator
+          while (itr.hasNext)
+            b += itr.next
+          NonEmptyList.fromList(b.result)
         }
         F.flatMap(tasks) {
           case Some(t) => F.raiseError(new OutstandingTasksAtShutdown(t))
           case None => F.unit
         }
-    }.map(es => wrapExecutorService(es))
+    }.map(es => liftExecutorService(es))
 
   /**
    * Creates a blocker that delegates to the supplied executor service.
    */
-  def wrapExecutorService(es: ExecutorService): Blocker =
-    wrapExecutionContext(ExecutionContext.fromExecutorService(es))
+  def liftExecutorService(es: ExecutorService): Blocker =
+    liftExecutionContext(ExecutionContext.fromExecutorService(es))
 
   /**
    * Creates a blocker that delegates to the supplied execution context.
@@ -90,7 +98,7 @@ object Blocker {
    * This must not be used with general purpose contexts like
    * `scala.concurrent.ExecutionContext.Implicits.global'.
    */
-  def wrapExecutionContext(ec: ExecutionContext): Blocker = new Blocker(ec)
+  def liftExecutionContext(ec: ExecutionContext): Blocker = new Blocker(ec)
 
   /** Thrown if there are tasks queued in the thread pool at the time a `Blocker` is finalized. */
   final class OutstandingTasksAtShutdown(val tasks: NonEmptyList[Runnable]) extends IllegalStateException("There were outstanding tasks at time of shutdown of the thread pool")
--- a/core/shared/src/main/scala/cats/effect/Blocker.scala
+++ b/core/shared/src/main/scala/cats/effect/Blocker.scala
@@ -18,8 +18,8 @@ package cats
 package effect
 
 import scala.concurrent.ExecutionContext
-import java.util.concurrent.{ExecutorService, Executors, ThreadFactory}
-import cats.data.NonEmptyList
+
+import cats.effect.internals.BlockerPlatform
 
 /**
  * An execution context that is safe to use for blocking operations.
@@ -46,50 +46,7 @@ final class Blocker private (val blockingContext: ExecutionContext) {
     cs.evalOn(this)(fa)
 }
 
-object Blocker {
-
-  /**
-   * Creates a blocker that is backed by a cached thread pool.
-   */
-  def apply[F[_]](implicit F: Sync[F]): Resource[F, Blocker] =
-    fromExecutorService(F.delay(Executors.newCachedThreadPool(new ThreadFactory {
-      def newThread(r: Runnable) = {
-        val t = new Thread(r, "cats-effect-blocker")
-        t.setDaemon(true)
-        t
-      }
-    })))
-
-  /**
-   * Creates a blocker backed by the `ExecutorService` returned by the
-   * supplied task. The executor service is shut down upon finalization
-   * of the returned resource.
-   * 
-   * If there are pending tasks in the thread pool at time the returned
-   * `Blocker` is finalized, the finalizer fails with a `Blocker.OutstandingTasksAtShutdown` 
-   * exception.
-   */
-  def fromExecutorService[F[_]](makeExecutorService: F[ExecutorService])(implicit F: Sync[F]): Resource[F, Blocker] =
-    Resource.make(makeExecutorService) { ec =>
-        val tasks = F.delay {
-          val tasks = ec.shutdownNow()
-          val b = List.newBuilder[Runnable]
-          val itr = tasks.iterator
-          while (itr.hasNext)
-            b += itr.next
-          NonEmptyList.fromList(b.result)
-        }
-        F.flatMap(tasks) {
-          case Some(t) => F.raiseError(new OutstandingTasksAtShutdown(t))
-          case None => F.unit
-        }
-    }.map(es => liftExecutorService(es))
-
-  /**
-   * Creates a blocker that delegates to the supplied executor service.
-   */
-  def liftExecutorService(es: ExecutorService): Blocker =
-    liftExecutionContext(ExecutionContext.fromExecutorService(es))
+object Blocker extends BlockerPlatform {
 
   /**
    * Creates a blocker that delegates to the supplied execution context.
@@ -98,7 +55,4 @@ object Blocker {
    * `scala.concurrent.ExecutionContext.Implicits.global'.
    */
   def liftExecutionContext(ec: ExecutionContext): Blocker = new Blocker(ec)
-
-  /** Thrown if there are tasks queued in the thread pool at the time a `Blocker` is finalized. */
-  final class OutstandingTasksAtShutdown(val tasks: NonEmptyList[Runnable]) extends IllegalStateException("There were outstanding tasks at time of shutdown of the thread pool")
 }
--- a/core/shared/src/main/scala/cats/effect/ContextShift.scala
+++ b/core/shared/src/main/scala/cats/effect/ContextShift.scala
@@ -68,10 +68,9 @@ trait ContextShift[F[_]] {
    * back to the default execution environment of `F` at the completion of `fa`,
    * regardless of success or failure.
    *
-   * The primary use case for this method is executing blocking code on a
-   * dedicated execution context, though it can also be used with libraries
-   * that require execution on a specific execution context. For the blocking
-   * use case, consider using the `evalOn(blocker)` overload instead.
+   * The primary use case for this method is executing code on a
+   * specific execution context. To execute blocking code, consider using
+   * the `evalOn(blocker)` overload instead.
    *
    * @param ec Execution context where the evaluation has to be scheduled
    * @param fa  Computation to evaluate using `ec`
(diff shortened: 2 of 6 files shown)


mpilquist:
Any opinions on an alternative constructor that uses a semaphore to guard
calls to `eval`? Like `BoundedLinebacker`

rossabaker:
With `BoundedLinebacker`, tasks just queue up, without bound or timeout?  I can see how it elegantly protects the blocking resource, but it introduces failure scenarios of its own  I think it's neat, but am mildly :-1: since there is a wide open design space for backpressure, and changing our APIs is expensive.

rossabaker:
But :+1: on `Blocker`.  There are a lot of ways to use it, but having it to use would be nice.

core/shared/src/main/scala/cats/effect/Blocker.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  /**
+   * Creates a blocker backed by the `ExecutorService` returned by the
+   * supplied task. The executor service is shut down upon finalization
+   * of the returned resource.
+   */
+  def fromExecutorService[F[_]](makeExecutorService: F[ExecutorService])(
+      implicit F: Sync[F]): Resource[F, Blocker] =
+    Resource
+      .make(makeExecutorService)(ec => F.delay { ec.shutdownNow(); () })
```
rossabaker: Note that this is a graceless shutdown.  It may be useful to have a version that takes a timeout and awaits termination until the timeout.  And since there are various things we might want to do with the returned list of runnables, I think `unsafeFromExecutionContext` carries its weight.

mpilquist: Yeah, that was intentional in this case. The pool *should* be empty at the point of finalization or it's arguably a bug in the caller's code. Also, we don't have a safe way to shutdown with a timeout unless we were to spawn a thread dedicated to shutdown.

mpilquist:
Here's what FS2 looks like if using `Blocker`: https://github.com/mpilquist/fs2/compare/topic/unblock...topic/unblock-plus-blocker

SystemFw:
> Here's what FS2 looks like if using Blocker:

I like that. As a bonus, places where you genuinely need an `Ec` not for blocking (e.g. `BlazeBuilder`) now are clearly different

core/shared/src/main/scala/cats/effect/Blocker.scala
rossabaker: I'm squeamish about the discard of the return value of `shutdownNow`.  I'm thinking a non-empty list of unrrun runnables should maybe be raised into an error.

Shutdown with a timeout can be done without a new thread by the `awaitTermination` call, which would, uh, block off the `Blocker`.  It's true that there's no async listener.  I'm coming around to it as a bad idea.

mpilquist: Right, `awaitTermination` would block a thread in the main CPU bound thread pool, which is problematic. Raising an error if task list is non-empty is probably a good idea. Think that should be controllable via a flag or just always the behavior?

mpilquist:
Any opinions on the names of `unsafeFromExecutionContext` and `unsafeFromExecutorService` versus `wrapExecutionContext` and `wrapExecutorService` or `liftExecutionContext` and `liftExecutorService`?

core/shared/src/main/scala/cats/effect/Blocker.scala
rossabaker: I think always. It's easily recoverable, but users can't do anything about it unless we surface it. Something that extends IllegalStateException with a NonEmptyList of Runnables feels right to me.

djspiewak:
I think `lift` or `liftExecutionContext` are fine names. I don't think it merits the `unsafe` prefix, since the unsafety was technically what the user did to create the "bare" EC in the first place.

core/shared/src/main/scala/cats/effect/Blocker.scala
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * An execution context that is safe to use for blocking operations.
+ *
+ * Used in conjunction with [[ContextShift]], this type allows us to write functions
+ * that require a special `ExecutionContext` for evaluation, while discouraging the
+ * use of a shared, general purpose pool (e.g. the global context).
+ *
+ * Instances of this class should *not* be passed implicitly.
+ */
+final class Blocker private (val context: ExecutionContext) {
```
djspiewak: *This*, however, might merit the word "unsafe". Perhaps `unsafeContext`? The point being that you want to discourage people from unwrapping it directly unless they really know what they're doing.

core/shared/src/main/scala/cats/effect/Blocker.scala
```diff
@@ -0,0 +1,97 @@
...
+   * Like `Sync#delay` but the supplied thunk is evaluated on the blocking
+   * execution context.
+   */
+  def delay[F[_], A](thunk: => A)(implicit F: Sync[F], cs: ContextShift[F]): F[A] =
+    eval(F.delay(thunk))
+
+  /**
+   * Evaluates the supplied task on the blocking execution context via `evalOn`.
+   */
+  def eval[F[_], A](fa: F[A])(implicit cs: ContextShift[F]): F[A] =
```
djspiewak: I kinda prefer the name `evalOn`, just to avoid having `eval` *and* `evalOn` which both do the same thing.

core/shared/src/main/scala/cats/effect/Blocker.scala
```diff
@@ -0,0 +1,97 @@
...
+   * exception.
+   */
+  def fromExecutorService[F[_]](makeExecutorService: F[ExecutorService])(implicit F: Sync[F]): Resource[F, Blocker] =
+    Resource.make(makeExecutorService) { ec =>
+        val tasks = F.delay {
+          val tasks = ec.shutdownNow().asScala.toList
+          NonEmptyList.fromList(tasks)
+        }
+        F.flatMap(tasks) {
+          case Some(t) => F.raiseError(new OutstandingTasksAtShutdown(t))
```
djspiewak: We honestly might want to make the threads daemon, because there's a rather ugly situation that can arise here where someone has a stuck thread, the `Resource` scope closes, this error is thrown, the main thread halts, and then the program… hangs.

core/shared/src/main/scala/cats/effect/Blocker.scala
```diff
@@ -0,0 +1,97 @@
...
+  def wrapExecutorService(es: ExecutorService): Blocker =
+    wrapExecutionContext(ExecutionContext.fromExecutorService(es))
+
+  /**
+   * Creates a blocker that delegates to the supplied execution context.
+   * 
+   * This must not be used with general purpose contexts like
+   * `scala.concurrent.ExecutionContext.Implicits.global'.
+   */
+  def wrapExecutionContext(ec: ExecutionContext): Blocker = new Blocker(ec)
```
djspiewak: `wrap` is also fine. :-)

core/shared/src/main/scala/cats/effect/Blocker.scala
mpilquist: How about `blockingContext`? I want to avoid making it difficult to integrate with APIs that take a plain `ExecutionContext` with an API note saying it should support blocking tasks (e.g. our very own `ContextShift`).

mpilquist:
All review comments are addressed -- let me know if I missed anything.

core/shared/src/main/scala/cats/effect/Blocker.scala
```diff
@@ -0,0 +1,105 @@
...
+   * If there are pending tasks in the thread pool at time the returned
+   * `Blocker` is finalized, the finalizer fails with a `Blocker.OutstandingTasksAtShutdown` 
+   * exception.
+   */
+  def fromExecutorService[F[_]](makeExecutorService: F[ExecutorService])(implicit F: Sync[F]): Resource[F, Blocker] =
+    Resource.make(makeExecutorService) { ec =>
+        val tasks = F.delay {
+          val tasks = ec.shutdownNow()
+          val b = List.newBuilder[Runnable]
+          val itr = tasks.iterator
```
mpilquist: `JavaConverters` is deprecated in 2.13 and I don't want to complicate the build with 2.12/2.13 source folders over 1 conversion from a Java to Scala list.

core/shared/src/main/scala/cats/effect/Blocker.scala
djspiewak: That's fair. Do you think we should also add an extra function to `ContextShift` which takes a `Blocker` and deprecate the `ExecutionContext` one? It feels weird to have two not-well-integrated parts of the API.

mpilquist: I thought so yeah, but then didn't do it to avoid breaking compatibility. But since this is a 2.0.0 release...

mpilquist: OK pushed a change -- I did *NOT* deprecate the EC variant of evalOn. See the ScalaDoc as to why.

jdegoes:
@mpilquist 

👍 for attempting to deal with a common scenario for async effect systems

👎 for contaminating the Scala.js API with something that only makes sense for and runs on the JVM

👎 for overstepping into effect-specific territory (neither Monix nor ZIO need this, so it's really a Cats IO-specific feature) and, within the Cats ecosystem, encroaching into `linebacker` territory

👎 for being too volatile an area to rapidly codify into 2.0 (the issue was created 3 days ago), especially without substantial input or discussion from contributors to 3rd party effect types

Overall, 👎 . It would be better IMO to rally around linebacker for the duration of Cats 2.0, for example, which is free to specialize on the JVM and evolve independently from Cats Effect; or to rethink this in a way that makes sense for Scala.js and has buy-in from Monix, ZIO, and other effect communities.

mpilquist:
@jdegoes It contaminates the Scala.js API and steps in to effect specific territory no more so than `ContextShift#evalOn`. We're stuck with `evalOn` for now and this only improves usability of the existing design.

ChristopherDavenport:
The point of cats is a set of abstractions that others can share and buy-in to if they would like to.

3rd party effect types are free to engage in discussion, but the progress of the library is independent of any implementations.

core/shared/src/main/scala/cats/effect/Blocker.scala
```diff
@@ -0,0 +1,104 @@
...
+  /**
+   * Creates a blocker that delegates to the supplied execution context.
+   * 
+   * This must not be used with general purpose contexts like
+   * `scala.concurrent.ExecutionContext.Implicits.global'.
+   */
+  def liftExecutionContext(ec: ExecutionContext): Blocker = new Blocker(ec)
+
+  /** Thrown if there are tasks queued in the thread pool at the time a `Blocker` is finalized. */
+  final class OutstandingTasksAtShutdown(val tasks: NonEmptyList[Runnable]) extends IllegalStateException("There were outstanding tasks at time of shutdown of the thread pool")
```
rossabaker: [GitHub ate my review, retrying]

I like raising an error when `Blocker` closes with outstanding tasks, but should we let the global compute pool in `IOApp` exit successfully in the same scenario?  A difference is that this is recoverable and that's not.  I don't know what we would do with the NEL in `IOApp` other than exit non-zero and maybe log the (probably uselessly `toString`ed) runnables.

I bring it up here, wondering whether this should be nested or has other uses.  But even if we change the behavior of `IOApp`, it's hard to see that the exception is useful there.

djspiewak:
@jdegoes 

> for contaminating the Scala.js API with something that only makes sense for and runs on the JVM

It's harmless on Scala.js. This is far from the first API which only makes sense on the JVM, and it will not be the last.

> for overstepping into effect-specific territory (neither Monix nor ZIO need this, so it's really a Cats IO-specific feature) and, within the Cats ecosystem, encroaching into linebacker territory

Cats Effect is… also a specific effect, whether you wish to accept it or not, and whether you agree with that situation or not. It is what it is right now.

I do see value in providing a newtype for a blocking effect context, even in a more generic ecosystem-wide context, but I'm viewing this through a less broad lens at the moment. It provides sufficient value even from that standpoint.

> for being too volatile an area to rapidly codify into 2.0 (the issue was created 3 days ago), especially without substantial input or discussion from contributors to 3rd party effect types

As you said, this is broadly immaterial to the only two major third-party effect types. @alexandru has been too busy to weigh in on things, and your discussion is… well, exactly what you wrote.

This is something which exists in quasar and has for years, it exists in almost every major cats-effect end-user project in some form, and the pattern is well-codified. We're not talking about a general abstraction here, we're talking about a tool to make it easier for users to do the right thing in a very common and very important use-case. For better or worse, people are generally very dependency-averse (unfortunately limiting linebacker's appeal), and they also look to cats-effect to codify the "right way" to do things. This moves us a step closer in that regard.

I'm well aware that a lot of this machinery only exists because of the lack of auto-shifting on `async`. That's a discussion for another issue. For the time being, this is something that has matriculated outside cats-effect for a *long* time and has proven sufficiently indispensable that I find it compelling.

Believe me, I'm very very sensitive to last-minute instability. I'm not particularly happy with shuffling more stuff into 2.0, but this is worth it, and it's sufficiently isolated of a change that I'm not as concerned with the implications, particularly as we will be turning our focus to 3.0 in the medium and long-term.

jdegoes:
>  It contaminates the Scala.js API and steps in to effect specific territory no more so than ContextShift#evalOn. We're stuck with evalOn for now and this only improves usability of the existing design.

That's a fair point, although `evalOn` works with "identity semantics" (there is only one execution context on Javascript), whereas this one will not. 

> 3rd party effect types are free to engage in discussion, but the progress of the library is independent of any implementations.

I will just [quote](https://github.com/typelevel/cats-effect/issues/321#issuecomment-424129260) @mpilquist on the matter:

> 3 is interesting as it has disadvantages as well (your bias point from earlier) -- the existence proof can convince us something's a good idea and then months later we find that it wasn't as it doesn't work for ZIO or Task.

If the purpose of Cats Effect has shifted to be a set of type classes around Cats IO, rather than an interop standard for different effect types&mdash;of which IO is only one example&mdash;then I'm sure a lot of people in the community would like to know that.

> It's harmless on Scala.js. This is far from the first API which only makes sense on the JVM, and it will not be the last.

It contaminates the Scala.js API with irrelevant and misleading clutter and will blowup at runtime.

Why not just move it over into `jvm`? We can do that with extension method and preserve the same API.

> Cats Effect is… also a specific effect, whether you wish to accept it or not, and whether you agree with that situation or not. It is what it is right now.

More precisely, Cats Effect _has_ a specific effect (_Cats IO_). This issue is evidence of further contamination of the type class hierarchy by the specific design characteristics of Cats IO.

> Believe me, I'm very very sensitive to last-minute instability. 

Indeed, the issue was created 3 days ago and is going to ship into a rushed 2.0 that will be with us for a very long period of time.

That said, I'm sympathetic to the idea of this being a lesser of two evils scenario...

mpilquist:
> That's a fair point, although evalOn works with "identity semantics" (there is only one execution context on Javascript), whereas this one will not.

Sure it will, just pass `Blocker.wrapExecutionContext(global)` on Scala.js.

> It contaminates the Scala.js API with irrelevant and misleading clutter and will blowup at runtime.

You mean specifically the constructor that creates a thread pool? I'm OK with moving that to JVM only.

> Why not just move it over into jvm? We can do that with extension method and preserve the same API.

FS2 has a couple of places in the core project that took a `blockingExecutionContext` to wrap things like printing to the console. Passing `global` there is harmless so I'd like to see this stay in both JVM and JS.

Again, this isn't a new type class -- there's already `ContextShift` instances for Monix (and presumably ZIO) so I don't think the last-minute instability comments are applicable.

mpilquist:
I moved the JVM specific constructors to JVM source directory: https://github.com/typelevel/cats-effect/pull/556/commits/57edf531f6b0e0527f44a3d370530d3f4d30dc10

djspiewak:
> That's a fair point, although evalOn works with "identity semantics" (there is only one execution context on Javascript), whereas this one will not.

That's up to whoever instantiates `Blocker`, as Mike points out.

> If the purpose of Cats Effect has shifted to be a set of type classes around Cats IO, rather than an interop standard for different effect types—of which IO is only one example—then I'm sure a lot of people in the community would like to know that.

This is not a typeclass.

The discussion of the conflation of the two concerns (a concrete datatype and a set of more abstract typeclasses) is a real one with a lot of nuance and merit, and IMO should be continued on the cats-effect 3 issue. We are where we are today, however.

> It contaminates the Scala.js API with irrelevant and misleading clutter and will blowup at runtime.

There is no runtime blow-up. I'll buy the clutter argument, but there's no need for hyperbole.

> Why not just move it over into jvm? We can do that with extension method and preserve the same API.

Conditionally-compiled APIs is one of the worst things *ever*, particularly with top-level types. Conditionally-available functions are a little better, since for the most part you can just avoid calling them, but making `Blocker` *the type* a conditionally-extant entity will have a viral effect on every cross-built in the ecosystem.

I think that perhaps wrapping the `ExecutorService` should be conditionally-compiled, particularly as it is an introduction rule, but that's about the extent of what I think is reasonable.

> More precisely, Cats Effect has a specific effect (Cats IO). This issue is evidence of further contamination of the type class hierarchy by the specific design characteristics of Cats IO.

Oh it would be… if it were in the typeclass hierarchy. Which it is not. Which you know. Please keep this discussion focused on the merits of this change, rather than on broader philosophical points. As I said, I'm interested in the discussion you're referencing, and I think it should be continued… on the cats-effect 3 discussion issue. Not here.

> Indeed, the issue was created 3 days ago and is going to ship into a rushed 2.0 that will be with us for a very long period of time.

Cats Effect 2 is an interim step. Cats Effect 3 will break many of these things, likely including `Blocker`, and will be undertaken with commensurate care. "With us for a very long time" is perhaps true to a degree, due to the expected long-tail on series/2.x, but as this is part of the concrete API rather than the polymorphic, I don't feel the repercussions are particularly relevant to effect implementors, only middleware and end-users, both of whom have voiced support.

The insinuation of carelessness is also unwelcome, here and everywhere.

If I had heard significant objection to `Blocker`'s inclusion in 2.0, I would have put my foot down. Instead I heard significant support, both here and elsewhere. If I had felt that this completely came out of no where without any conceptual proving in the wild, I would have not considered it. Instead, numerous people (including myself) came forward and noted that we had all rolled essentially the same utility internally in our various projects. This ticks off the last of the boxes, and satisfied me that this had matriculated sufficiently to merit inclusion in the concrete API (I would not have supported such a last-minute change to the abstractions, regardless of any proofing).

jdegoes:
> You mean specifically the constructor that creates a thread pool? I'm OK with moving that to JVM only.

Yes, exactly.

> I moved the JVM specific constructors to JVM source directory:

I am happy to see that. 👍 

> FS2 has a couple of places in the core project that took a blockingExecutionContext to wrap things like printing to the console. Passing global there is harmless so I'd like to see this stay in both JVM and JS.

I can see the convenience, although shouldn't that be an implementation detail of a `Console[F]`? Anytime blocking/non-blocking semantics end up inside an API that can be used cross-platform, it's probably a sign there's a missing type class somewhere.

> There is no runtime blow-up. I'll buy the clutter argument, but there's no need for hyperbole.

No hyperbole intended. By "blow up", I mean there were methods in the pull request that would throw an exception if invoked on Scala.js.

We have all the tools necessary to restrict JVM-only features to the JVM, and IMO we should use them (and I am happy to see 5c073ef).

> Oh it would be… if it were in the typeclass hierarchy. Which it is not. Which you know.

Whether type class or not, this is a chunk of functionality whose value is intrinsically tied to use of the concrete Cats IO data type, and I bring it up because adding _even more_ such IO-specific functionality into Cats Effect should be explicitly recognized as a design-smell.

> The insinuation of carelessness is also unwelcome, here and everywhere.

I never said it was careless—I said it was rushed, which is not the same as careless. In the past, changes this large have taken more time and have involved input from Alex, myself, or both, and have considered more aspects of the ecosystem (such as Scala.js compatibility, utility for Monix and ZIO, etc.).

I feel if I had not seen this today (more or less chance!), it would already be merged in, without separation of JVM-only functionality, and this would have had a lasting impact on the API.

SystemFw:
> Whether type class or not, this is a chunk of functionality whose value is intrinsically tied to use of the concrete Cats IO data type,

You seem to conflate "ZIO does not need it" with "it's needed only for cats IO", forgetting polymorphic code, which is a primary design concern for cats-effect

jdegoes:
@SystemFw 

IMO a true effect-polymorphic solution would introduce a `blocking` combinator as part of a type class, which is a design possibility unexplored here and which I would have loved to see considered as an alternative. It's also more JS friendly.

SystemFw:
> IMO a true effect-polymorphic solution would introduce a blocking combinator as part of a type class, 

Sure, but if you consider a bit of context you see that is just an improvement over current `evalOn`  , which makes it strictly better for 2.0.

I'm sure `blocking` can be considered for 3.0, even though when we did discuss it in the past the decision to have a single Ec for all blocking task in an app debatable at least, and therefore one could argue that `blocking` is just a special case of `Blocker`

jdegoes:
> Sure, but if you consider a bit of context you see that is just an improvement over current evalOn , which makes it strictly better for 2.0.

Well, this is precisely the kind of conversation I would have liked to have, or at least see someone else have. 😄 

The following design would be truly effect polymorphic (i.e. no assumptions about implementation) and of use to non-IO types, as well as fully implementable for Cats IO:

```scala
trait Blocking[F[_]] {
  def blocking[A](fa: F[A]): F[A]
}
```

It also has better ergonomics and is easier to constrain to the JVM.

> a single Ec for all blocking tasks in an app seemed debatable at least

Since we are literally talking about a JVM-only thread pool that has unbounded capacity, I don't think that argument has merit in the current context.

mpilquist:
> Whether type class or not, this is a chunk of functionality whose value is intrinsically tied to use of the concrete Cats IO data type, and I bring it up because adding even more such IO-specific functionality into Cats Effect should be explicitly recognized as a design-smell.

> IMO a true effect-polymorphic solution would introduce a blocking combinator as part of a type class, which is a design possibility unexplored here and which I would have loved to see considered as an alternative. It's also more JS friendly.

`Blocker` improves the usability of the *existing* abstractions. Replacing `ContextShift` with a blocking combinator / first class support in a type class is on the table for CE 3 but not CE 2.

> trait Blocking[F[_]] {
  def blocking[A](fa: F[A]): F[A]
}

This is a new type class, which would require all effect libraries to opt-in. This is beyond the scope of 2.0 IMO.

> I never said it was careless—I said it was rushed, which is not the same as careless. In the past, changes this large have taken more time and have involved input from Alex, myself, or both, and have considered more aspects of the ecosystem (such as Scala.js compatibility, utility for Monix and ZIO, etc.).

Deadlines tend to produce PRs and this is by no means even close to the largest PR to hit just prior to a deadline. I'd describe this as a *small* PR, adding a newtype around `ExecutionContext` along with some convenience constructors & combinators.

More importantly, cats-effect issues tend to be bike shed more than any project I've seen. I worry that we are driving away contributors and current maintainers with these conversations.

jdegoes:
> Blocker improves the usability of the existing abstractions. Replacing ContextShift with a blocking combinator / first class support in a type class is on the table for CE 3 but not CE 2.

I don't want to replace `ContextShift` in 2.0, just I'd prefer to avoid adding machinery to it, and I'd be happy to support the `Blocking` type class in both Monix and ZIO, leaving only Cats IO, which given the infrastructure in this PR, would be pretty easy to support as well.

That said, fair enough!

> Note you're focused on the apply constructor but forgetting that evalOn is used for other things as well (e.g. working with a library that requires all calls to occur from a special library managed thread pool).

This is exactly the kind of discussion I want to have. 😄 

I'd point out that `blocking` does not replace the need for `evalOn`, it only replaces the #1 use case for this functionality, in a way that doesn't assume any implementation details.

> Deadlines tend to produce PRs and this is by no means even close to the largest PR to hit just prior to a deadline.

I agree, and I saw similar urgency prior to CE 1.0 and _appreciated_ the pushbacks.

> More importantly, cats-effect issues tend to be bike shed more than any project I've seen. I worry that we are driving away contributors and current maintainers with these conversations.

Although you would expect and hope that in a widely depended upon library, changes would involve discussion and take some time to work through. 

But you may be right about the bike-shedding, and I would propose the following:

Stop bike-shedding at the request of any Cats Effect contributor when any pull request / issue has:

 - Buy-in from at least `2` distinct end-users of distinct libraries that depend on cats-effect
 - Buy-in from distinct contributors to at least `2` different libraries that depend on cats-effect
 - Buy-in from distinct contributors to at least `2` different effect data types that support cats-effect

The stake holders are the end-users, the library maintainers, and the effect maintainers, and if you have a little consensus in each category, it's probably a safe enough change to push through without additional bike-shedding.

djspiewak:
> We have all the tools necessary to restrict JVM-only features to the JVM, and IMO we should use them (and I am happy to see 5c073ef).

While I think that is true to a degree, it can also cause some significant heartache in a lot of cases when not done carefully, particularly when employed on top-level types. In this case, simply moving the constructor is certainly the right approach.

> Whether type class or not, this is a chunk of functionality whose value is intrinsically tied to use of the concrete Cats IO data type, and I bring it up because adding even more such IO-specific functionality into Cats Effect should be explicitly recognized as a design-smell.

You continue to see Cats Effect as "abstraction primary, concrete secondary", and thus you view any API changes to Cats Effect which are not *literally part of `IO`* to be something that is targeted at the broader abstraction surface area. Again, this is a conversation for 3.0 planning, but I want to be very specific about calling out that mindset. You have a misconception about what Cats Effect *is*, based solely on your perspective as an external implementor and user of the API, and severely distorted by your needs from that API and your design goals, subjective and objective. This is all to be expected, but please be conscious of it as we move forward with 3.0 design discussions, or it will keep arising as a point of contention over and over and over.

To be clear, I'm not begrudging you your biases. I certainly have my own. What I am begrudging is your assumption that everyone shares yours.

> In the past, changes this large have taken more time and have involved input from Alex, myself, or both, and have considered more aspects of the ecosystem (such as Scala.js compatibility, utility for Monix and ZIO, etc.).

I certainly considered Scala.js while reviewing this, though perhaps not with the thoroughness that I should have. Monix and ZIO always have a seat at this table (as does everyone, really), but in this case the input is less relevant. Alex is also extremely busy right now.

> I feel if I had not seen this today (more or less chance!), it would already be merged in, without separation of JVM-only functionality, and this would have had a lasting impact on the API.

Probably not merged today, but if you hadn't seen it at all and no one else had raised the point about the thread pool constructor, then yes it likely would have been merged in that form. And that is precisely why I engage in these sorts of conversations: deep down, at the end of the day, some value will come out of it.

> IMO a true effect-polymorphic solution would introduce a blocking combinator as part of a type class, which is a design possibility unexplored here and which I would have loved to see considered as an alternative. It's also more JS friendly.

Drifting into cats-effect 3 land, but unless datatypes have an intended implementation of `blocking` which does *not* directly boil down to `evalOn(blockingEC)` (assuming compositional `evalOn` semantics and the JVM), then I'm strongly in favor of saying that `blocking` just isn't a primitive. But it's at least an interesting design question.

> Since we are literally talking about a JVM-only thread pool that has unbounded capacity, I don't think that argument has merit in the current context.

I can think of at least three cases *within as many years* where I have needed non-singleton pools for blocking tasks. At least one of them you were personally witness to. It's very common, particularly when interacting with older APIs written for the Java Enterprise Application Server era.

> This is a new type class, which would require all effect libraries to opt-in. This is beyond the scope of 2.0 IMO.

It is.

> More importantly, cats-effect issues tend to be bike shed more than any project I've seen. I worry that we are driving away contributors and current maintainers with these conversations.

This. Very much this. While I don't mind a lot of reasoned debate and careful consideration around even minor things like naming on a project which is as central as cats-effect, there is certainly a line that we need to be careful of.

> Stop bike-shedding at the request of any Cats Effect contributor when any pull request / issue has: …

I more or less follow this process already when policing this community (granted, from which I took a long hiatus). The exception that I draw is that, on certain matters I lend considerably more weight to implementors of datatypes, on certain others I lend considerably more weight to middleware consumers (polymorphic libraries), and on a third set I lend more weight to end-user considerations (downstream consumers of polymorphic libraries and datatypes). Cats Effect's role is to mediate between all three of these, and some issues lend considerably more credence to one stakeholder over another.

But in general, yes, I think that if something has overwhelming and universal buy-in of the type you describe, then it is beyond a shadow of a doubt.

jdegoes:
> You have a misconception about what Cats Effect is, based solely on your perspective as an external implementor and user of the API

My biases, whatever they may be, don't invalidate my technical arguments that `Blocker` is overly specialized to Cats IO, rather than being truly _effect polymorphic_ and _effect type agnostic_, like the proposed `Blocking` type class (which makes no assumptions regarding how the functionality is provided).

In fact, if at any point we find ourselves discussing other people's "biases" and "misconceptions" (which are _personal judgments_), rather than discussing technical issues, then I'd say we've veered off the happy path and should discontinue the conversation.

core/shared/src/main/scala/cats/effect/Blocker.scala
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * An execution context that is safe to use for blocking operations.
+ *
+ * Used in conjunction with [[ContextShift]], this type allows us to write functions
+ * that require a special `ExecutionContext` for evaluation, while discouraging the
+ * use of a shared, general purpose pool (e.g. the global context).
+ *
+ * Instances of this class should *not* be passed implicitly.
+ */
+final class Blocker private (val blockingContext: ExecutionContext) {
```
kubukoz: Why not a value class?

core/shared/src/main/scala/cats/effect/ContextShift.scala
```diff
@@ -50,13 +50,28 @@ trait ContextShift[F[_]] {
...
    * back to the default execution environment of `F` at the completion of `fa`,
    * regardless of success or failure.
    *
    * The primary use case for this method is executing blocking code on a
    * dedicated execution context.
    *
+   * @param blocker blocker where the evaluation has to be scheduled
+   * @param fa  Computation to evaluate using `blocker`
+   */
+  def evalOn[A](blocker: Blocker)(fa: F[A]): F[A] =
```
kubukoz: overloading is still something I'd try to avoid as long as possible... If this was `blockOn`, it'd probably make Blocker eligible for a value class, no? Plus more explicit than just an overload

core/shared/src/main/scala/cats/effect/Blocker.scala
mpilquist: No reason really - I tend to avoid value classes unless the performance gain is significant. In this case, it's a single extra allocation per app in most cases. :)

kubukoz:
Has anyone mentioned the possibility of taking the required type classes (Sync, ContextShift [yes I know it's not a type class]) at construction time, instead of at call sites of methods? That'd make Blocker[F] a thing and possibly reduce the amount of Sync necessary to use it in many places. And it wouldn't push any responsibility on the implementations like Monix and ZIO either.

core/shared/src/main/scala/cats/effect/Blocker.scala
kubukoz: Good point, OTOH it'd make sense if old-style evalOn was removed. But since it's not being removed... :)

mpilquist:
@kubukoz That was intentional, as it causes problems when using multiple effects -- e.g. transformers plus a base effect type. The resource itself (the pool) doesn't care which `F` is used to submit tasks to it, so the current encoding is more flexible.

core/shared/src/main/scala/cats/effect/Blocker.scala
mpilquist: See #560 btw

kubukoz:
@mpilquist I see, so how about dual effects like in creation of a Ref and friends?

mpilquist:
Same issue still -- e.g. imagine a Skunk app that uses a `Blocker` to talk to Postgres via TCP but also uses the same `Blocker` when writing files to disk in `IO` effect. Or similarly, an http4s app that does blocking IO in handling responses but also file IO outside request handling.

kubukoz:
`mapK` then? You could have a single underlying EC and create a copy with a different effect just by mapping between them (from IO to Kleisli or from IO to whatever Skunk needs)

mpilquist:
Then you run in to an issue with `mapK[G[_]](u: F ~> G)` needing an implicit `Sync[G]` and `ContextShift[G]`, which means it's not really `mapK` (this same issue exists in fs2 -- see `translate` vs `translateInterrupt`).

mpilquist:
BTW, you also run in to the issue of changing which implicits have to be passed to various use sites -- something `Blocking` also shares. That is, with `Blocker` not depending on an `F`, you still need the same `Sync[F]` and `ContextShift[F]` at the call site. If `Blocker` was parameterized by `F`, many uses would change from needing those two type classes + an `ExecutionContext` to needing just a `Blocker[F]`. I think that's a big enough change to discount it for 2.0.

PR #558, opened 2019-06-09
Remove logical/native threads distinction from Concurrency Basics doc

Avasil:
In the hindsight I think it was confusing and/or false


site/src/main/tut/concurrency/basics.md
```diff
@@ -242,7 +234,9 @@ prog.unsafeRunSync()
 
 Now it will keep printing both `1` and `2` but neither `11` nor `22`. What changed?
 Those thread pools are independent and interleave because of thread scheduling done by the operating system.
-Let's get it right:
+Basically the thread pool decides which task gets a thread to run but OS decided what is actually running on the CPU.
```
djspiewak: Grammar:

```suggestion
Basically, the thread pool decides which task gets a thread to run but the OS decides what is actually evaluating on the CPU.
```

PR #560, opened 2019-06-11
Renamed ContextShift#evalOn(Blocker) to blockOn and changed Blocker to a value class

mpilquist:
Review comments from @kubukoz addressed

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Blocker.scala
+++ b/core/shared/src/main/scala/cats/effect/Blocker.scala
@@ -40,7 +40,7 @@ final class Blocker private (val blockingContext: ExecutionContext) extends AnyV
     blockOn(F.delay(thunk))
 
   /**
-   * Evaluates the supplied task on the blocking execution context via `evalOn`.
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/effect/Blocker.scala
```diff
@@ -30,20 +30,20 @@ import cats.effect.internals.BlockerPlatform
  *
  * Instances of this class should *not* be passed implicitly.
  */
-final class Blocker private (val blockingContext: ExecutionContext) {
+final class Blocker private (val blockingContext: ExecutionContext) extends AnyVal {
```
kubukoz: No strong opinion on this one :)

mpilquist: No strong reason not to either, and one benefit is improved binary compatibility (stuff that used to take `ExecutionContext` and now takes `Blocker` will be BC but not source compatible).

kubukoz: Yeah, that makes sense.

PR #565, opened 2019-06-14
Add `apply` as an alias for `delay` - fixes #517

### LINKED ISSUE #517: Proposal: Add `apply` function to Sync as an alias for `delay`
Back in the eons of the mist… I used the `apply` function on the `IO` object to signify "capture of a synchronous effect" because this case is, in my experience, far and away the most common thing. This API design had the effect of taking the old idiom of `Task.delay(` and reducing it it to `IO(`, a reduction of 8 characters on something done almost incessantly.

What I *didn't* anticipate is that everyone would just use the typeclasses and almost never `IO` directly! :-D So in practice, we've replaced `Task.delay` with `Sync[F].delay`, which is actually *worse* from a syntactic standpoint. Yes, I'm picking nits.

I'd like to toss out there as a minor proposal simply adding `apply` to `Sync` as an alias for `delay`. This would mirror `IO` more closely and simplify a lot of common cases. For example:

```scala
def putStrLn[F[_]](line: String)(implicit F: Sync[F]): F[Unit] =
  F.delay(println(line))

// becomes

def putStrLn[F[_]](line: String)(implicit F: Sync[F]): F[Unit] =
  F(println(line))
```

Unfortunately, there is a bit of a gotcha to it which prevents it from being *quite* as ubiquitous as we might want:

```scala
def putStrLn[F[_]](line: String)(implicit F: Sync[F]): F[Unit] =
  Sync[F](println(line)) // doesn't work!
```

This won't work since the parentheses are actually interpreted here as being the parameters of the `Sync` *object* `apply` function, which is to say, a value of type `Sync[F]`. There's also the (remote) possibility of a user accidentally doing something really awful if the body of their effect is coercible to `Sync[F]`.

So anyway, this is small potatoes, and I can see reasons not to do it, but it's been bugging me for a while so I thought I'd throw it out there.

TomTriple:


djspiewak:
I'm going to recuse myself from this one. :-) Other contributors should review and 👍 / 👎

SystemFw:
I'm mildly :-1: but I'm not going to bikeshed it to the death if people want it.

My reasoning is that in a polymorphic context if you use `F {  thing }` you have no context from the method name (`apply` is very neutral)`, nor from the type name (`IO { ....}` is informative).
The counterpoint would be doing `Sync[F] {...}` vs `Sync[F].delay`, which is why I don't feel super strongly about it.

ChristopherDavenport:
I commented in greater detail on this in the other issue, I'm personally 👎 on this one.

SystemFw:
I suspect there is a strong correlation between users of tagless interfaces vs direct wrapping of side-effects with `delay` and :+1: / :-1: on this PR. I fall in the former camp (and I suspect so does Chris), so it would be good to hear from someone with a different style

LukaJCB:
> The counterpoint would be doing `Sync[F] {...}` vs `Sync[F].delay`, which is why I don't feel super strongly about it.

The former won't work without any extra syntax tricks, as it will expect the `Sync[F]` instance instead of the lazily evaluated body. So you'll only be able to call it as `Sync[F].apply {...}`

SystemFw:
@LukaJCB obviously you're right, good point

TomTriple:
Ok, the need to call it with `Sync[F].apply` defeats the reason for adding it ;)

PR #566, opened 2019-06-14
Restore binary compat in core on 2.12

kubukoz:
Brings back `fromFuture` in the old shape, but deprecated (only for documentation reasons, since the method can't legally be called anyway) and package-private.

Hopefully this can end up in 2.0.0.


kubukoz:
@LukaJCB ;)

kubukoz:
Looks like tests don't compile because the method is now ambiguous in the `cats.effect` package. I'll try a more restricting `private` that's still binary compatible...

kubukoz:
Also, because of `onCancel` and friends this will still not fix BC in 2.11, but at least we can save 2.12. I think it's worth it.

PR #567, opened 2019-06-14
align scaladoc with examples from the official docs for IO.shift - fixes #559

### LINKED ISSUE #559: IO.shift(ExecutionContext) documentation code doesn't pass the ExecutionContext
For the signature of `IO.shift` that takes an explicit `ExecutionContext` the code examples in the ScalaDoc documentation call `shift` without passing an `ExecutionContext`.  This looks like a C&P error from the other signature.

TomTriple:
Fixes #559

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -880,9 +880,9 @@ private[effect] abstract class IOInstances extends IOLowPriorityInstances {
  *
  *         {{{
  *           import cats.effect.{IO, ContextShift}
- *           import scala.concurrent.ExecutionContext.Implicits.global
  *
- *           val contextShift = IO.contextShift(global)
+ *           val ec = ExecutionContext.fromExecutor(Executors.newFixedThreadPool(...))
+ *           val contextShift = IO.contextShift(ec)
  *         }}}
  *
  *         For example we can introduce an asynchronous
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -862,17 +862,51 @@ private[effect] abstract class IOInstances extends IOLowPriorityInstances {
 }
 
 /**
- * @define shiftDesc For example we can introduce an asynchronous
+ * @define shiftDesc Note there are 2 overloads of the `IO.shift` function:
+ *
+ *         - One that takes a `ContextShift` that manages the
+ *           thread-pool used to trigger async boundaries.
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -862,17 +862,51 @@ private[effect] abstract class IOInstances extends IOLowPriorityInstances {
...
  *         }}}
  *
  *         Or using Cats syntax:
  *         {{{
- *           import cats.syntax.all._
+ *           import cats.syntax.apply._
  *
  *           IO.shift *> task
+ *           // equivalent to
+ *           implicitly[ContextShift[IO]].shift *> task
```
```suggestion
 *           ContextShift[IO].shift *> task
```

Since 1.1, we have an implicit summoner for `ContextShift` on the companion object, and I would consider its usage to be better practice than `implicitly`.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -884,6 +918,8 @@ private[effect] abstract class IOInstances extends IOLowPriorityInstances {
  *         Or using Cats syntax:
  *         {{{
  *           task <* IO.shift
+ *           // equivalent to
+ *          task <* implicitly[ContextShift[IO]].shift
```
```suggestion
 *          task <* ContextShift[IO].shift
```

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -862,17 +862,51 @@ private[effect] abstract class IOInstances extends IOLowPriorityInstances {
...
+ *         use.
+ *
+ *         By default, Cats Effect can provide instance of
+ *         `ContextShift[IO]` that manages thread-pools, but
+ *         only if there’s an `ExecutionContext` in scope or
+ *         if `IOApp` is used:
+ *
+ *         {{{
+ *           import cats.effect.{IO, ContextShift}
+ *           import scala.concurrent.ExecutionContext.Implicits.global
```
djspiewak: Should we be encouraging people to use `global` like this? I'm sure this kind of thing exists elsewhere in the docs, but I'm leery of it particularly given the problems we've seen and the fact that `IOApp` no longer uses it.

core/shared/src/main/scala/cats/effect/IO.scala
TomTriple: Ah, yes I forgot about that. I'll send a PR that addresses its usage in the docs also.

core/shared/src/main/scala/cats/effect/IO.scala
TomTriple: I changed the docs by only outlining its basic usage - that at least removes the usage of `global`.

PR #568, opened 2019-06-18
Make default scheduler remove cancelled tasks.

avakhrenev:
Should prevent leaking timeouts.


avakhrenev:
@rossabaker right, thank you.

Issue #405, opened 2018-11-05
Add an Alternative instance for Fiber/IO.Par


LukaJCB:
`Alternative` basically allows us to combine two `F` into `F[(A, B)]]` or `F[Either[A, B]]` along with identities for each.
This maps perfectly onto racing (where `IO.never` is the identity) and running in parallel (where `IO.unit` is the identity) 
An instance for `IO.Par` could allow us to make use of existing abstractions using the `SemigroupK` family. I.e. racing a list of IOs could be as simple `list.foldMapK(f)`.

LukaJCB:
Potentially I think the same can be done for `Fiber` :)

LukaJCB:
Seems like Haskell already runs with this idea in its `Alternative` instance for `Concurrently`

catostrophe:
`race` depends on scheduling hence introduces indeterminism. How can we prove `SemigroupK` associativity law?

catostrophe:
I implemented `SemigroupK[IO.Par]`/`MonoidK[IO.Par]` and the associativity law test failed:
```
[info] - IO.Par.monoidK.monoidK left identity
[info] - IO.Par.monoidK.monoidK right identity
[info] - IO.Par.monoidK.semigroupK associative *** FAILED ***
[info]   GeneratorDrivenPropertyCheckFailedException was thrown during property evaluation.
[info]    (BaseTestsSuite.scala:44)
[info]     Falsified after 1 successful property evaluations.
[info]     Location: (BaseTestsSuite.scala:44)
[info]     Occurred when passed generated values (
[info]       arg0 = IO(1936284316),
[info]       arg1 = <function1>,
[info]       arg2 = IO$1181956722
[info]     )
[info]     Label of failing property:
[info]       Expected: <function1>
[info]   Received: <function1>
```

catostrophe:
Should we add a lawless instance in favor of convenience (e.g. for having `foldMapK`)?

LukaJCB:
I wonder why it wouldn't be associative, conceptually it seems to me like `race` should definitely be associative 🤔

catostrophe:
The choice depends on which task finishes first, and it's unpredictable and not reproducible in general. In contrast, `parApplicative` guarantees the same result irregardless of the execution speed & order.

catostrophe:
[...]

PR #576, opened 2019-06-28
Add an Alternative instance for Fiber

marcin-rzeznicki:
See issue #405 

This instance implements the notion of `|+|` as a parallel race between two of its sides. Whoever returns successfully (ie. without error) first wins. `empty` is chosen to be `raiseError(Empty)` where `Empty` is an unique exception subclass defined in the `Fiber` companion. This decision, IMO, prevents some issues that arise if `empty` was to be defined as `never`. First, it preserves absorption laws ie `ff <*> empty <-> empty` . Second, it preserves monoid identity laws ie. `<fail> |+| empty <-> <fail>` and `empty |+| <fail> <-> <fail>` (in case of `never`, `empty` consumes the failing computation).

To test the respective laws, I needed to change how `TestContext` behaves. Unfortunately, in the presence of both non-deterministic _combine_ and semiringal `Alternative` laws, this implementation might seem unlawful (please check the comments). But, if one eliminated non-determinism for the purpose of tests, it could be, at least partially, checked. For this to happen, I introduced the _detministic_ flag which makes `TestContext`:
- execute tasks without any randomness
- schedule new tasks in LIFO order (wrt to the clock)

This setting makes all races predictable (ie. first submitted task will always win). Currently it is used only to test `Alternative` laws.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Fiber.scala
+++ b/core/shared/src/main/scala/cats/effect/Fiber.scala
@@ -19,7 +19,7 @@ package cats.effect
 import cats.syntax.apply._
 import cats.syntax.applicative._
 import cats.syntax.flatMap._
-import cats.{Applicative, Apply, Monoid, Semigroup, ~>}
+import cats.{Alternative, Applicative, Apply, Monoid, Semigroup, ~>}
 
 /**
  * `Fiber` represents the (pure) result of an [[Async]] data type (e.g. [[IO]])
@@ -98,7 +98,11 @@ object Fiber extends FiberInstances {
 
 private[effect] abstract class FiberInstances extends FiberLowPriorityInstances {
 
-  implicit def fiberApplicative[F[_]](implicit F: Concurrent[F]): Applicative[Fiber[F, ?]] = new Applicative[Fiber[F, ?]] {
+  private case object Empty extends RuntimeException("Empty")
+
+  implicit def fiberAlternative[F[_]](implicit F: Concurrent[F]): Alternative[Fiber[F, ?]] = new Alternative[Fiber[F, ?]] {
+    private def cancelBoth[A,  B](f1: Fiber[F, A], f2: Fiber[F, B]) = F.map2(f1.cancel, f2.cancel)((_, _) => ())
+
     final override def pure[A](x: A): Fiber[F, A] =
       Fiber(F.pure(x), F.unit)
     final override def ap[A, B](ff: Fiber[F, A => B])(fa: Fiber[F, A]): Fiber[F, B] =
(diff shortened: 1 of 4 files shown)


marcin-rzeznicki:
Oh, I see it fails on `mima` - apparently I shouldn't have changed the method name. I am not well versed when it comes to binary compatibility. Should I keep the name `applicativeFiber` even though it implements `Alternative`?

djspiewak:
@marcin-rzeznicki To get around bincompat, perhaps you could use the new method name and have the old one simply forward to the new (non-`implicit`)?

@alexandru Can you take a look at this? You're much more qualified than I to judge the merits of changes to `TestContext`.

marcin-rzeznicki:
@djspiewak Ok, I'll experiment, thanks. I am currently on a short vacation, so some time next week, if that's ok with you

marcin-rzeznicki:
Ok, I got to perform this strange dance of creating two methods and accompanying class, but now it seems to be working. May I ask for a review?

marcin-rzeznicki:
Ping

djspiewak:
This is amazingly great! Thank you so much for doing this, @marcin-rzeznicki!

marcin-rzeznicki:
Pleasure is all mine. Thanks for all the support and discussions.

alexandru:
Guys, I've just noticed this PR and I'm worried about this:

> To test the respective laws, I needed to change how TestContext behaves. Unfortunately, in the presence of both non-deterministic combine and semiringal Alternative laws, this implementation might seem unlawful (please check the comments). But, if one eliminated non-determinism for the purpose of tests, it could be, at least partially, checked.

I'm not familiar with what `Alternative` does, but the randomness in `TestContext` simulates a real world `ExecutionContext`.

In other words, if the tests now pass, it does not mean the implementation is lawful. All it means is that you're testing with conditions that will not happen in real production code.

I don't think that flag has any place in `TestContext`. 

From my point of view, if the tests fail with a `TestContext` that introduces randomness in the selection of tasks, then either the code or the tests are probably broken.

jdegoes:
Racing joins is definitely not be lawful. Said instance should be deleted.

marcin-rzeznicki:
@alexandru The issue is a bit more subtle and was debated in the respective issue. Please read the comments there. TL/DR version from my POV is: The laws for Alternative demand that its instances form a semiring which excludes any non-deterministic implementation of Alternative (not just this one). This seems to be a serious drawback because non-deterministic instances are very useful in context of this type-class - some say that these laws are over-specified. OTOH if one could somehow test these laws as sets rather than values  ie. set of observed values is the same for both sides of a law, then the issue would be gone. At the moment we can do neither. So the least one can do is to make sure that a non-deterministic computation is sane in deterministic context which is why this change was needed and useful. Obviously, one should not over-use it.

PR #578, opened 2019-06-30
Get rid of N.B. in documentation

CucumisSativus:
It took me a longer while to realise what `N.B.` means in the docs. IMO `important` is more readable and expresses the intent in a cleaner way


djspiewak:
"NB" is definitely one of those terms that's ubiquitous in some circles and completely unheard of in others. 👍 for avoiding jargon wherever possible.

PR #586, opened 2019-07-19
Fix typo in bracketCase doc

julien-truffaut:


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -501,7 +501,10 @@ sealed abstract class IO[+A] extends internals.IOBinaryCompat[A] {
    * corrupted data.
    *
    * For those cases you might want to do synchronization (e.g. usage
-   * of locks and semaphores) and you might want to use [[bracketCase]],
+   * of locks and semaphores) and you might want to use [[
+   
+   
+   ]],
```
rossabaker: Is this a mistake?  The other chunk looks good.

julien-truffaut: yes, I don't know where does it come from. I will remove it

PR #588, opened 2019-07-23
MonoidK instance for IO.Par

marcin-rzeznicki:
Similar to #576


alexandru:
Here too, please see my comment from here:
https://github.com/typelevel/cats-effect/pull/576#issuecomment-515716766

TL;DR — if `TestContext` needs to be deterministic, then either the implementation or the tests are broken.

In real world code there's no such thing as a deterministic `ExecutionContext`.

PR #589, opened 2019-07-24
sbt microsites 0.7.27

amohrland:
Resolves the error in https://github.com/typelevel/cats-effect/pull/570, which is caused by trailing comments in tut sheds as described [here](https://github.com/tpolecat/tut/issues/243#issuecomment-438288206). Adding a trailing `()` fixes it, though it looks slightly odd.


djspiewak:
Yay!

PR #591, opened 2019-07-25
Add Async.fromFuture (#563)

### LINKED ISSUE #563: Polymorphic fromFuture and ContextShift
As noted in #453, there is a simple solution to have `fromFuture` for a type parameter `F[_]`:
```scala
IO.fromFuture(IO(someFutureThing)).to[F]
```

Until recently, a `LiftIO[F]` was enough for this. But now, due to #546, a `ContextShift[IO]` is also necessary. Note that that is `ContextShift[IO]` and _not_ `ContextShift[F]`. Which makes it significantly more inconvenient. Also note, that while #546 made `fromFuture` safer, an also safe and polymorphic version was already simple before (if `ContextShift[F]` is available):
```scala
IO.fromFuture(IO(someFutureThing)).to[F].guarantee(ContextShift[F].shift)
```
Now this is not possible without a `ContextShift[IO]`.

I see two possible solutions:
- have a polymorphic `fromFuture` (e.g., like in #453), which takes a `ContextShift[F]`; or
- have a function to create a `ContextShift[IO]` from a `ContextShift[F]` (but `Effect[F]` seems to be necessary for this).

Daenyth:
Currently lifting `F[Future[A]]` to `F[A]` requires `ContextShift[IO]`. This remedies that by exposing the same logic used by `IO.fromFuture` for any `Async`

Closes #563

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Async.scala
+++ b/core/shared/src/main/scala/cats/effect/Async.scala
@@ -23,10 +23,11 @@ import cats.data._
 import cats.effect.IO.{Delay, Pure, RaiseError}
 import cats.effect.concurrent.{Deferred, Ref, Semaphore}
 import cats.effect.internals.{Callback, IORunLoop}
+import cats.effect.internals.TrampolineEC.immediate
 
 import scala.annotation.implicitNotFound
-import scala.concurrent.ExecutionContext
-import scala.util.Either
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Either, Failure, Success}
 
 /**
  * A monad that can describe asynchronous or synchronous computations
@@ -275,6 +276,22 @@ object Async {
         def run(): Unit = cb(Callback.rightUnit)
       })
     }
+
+  def fromFuture[F[_], A](fa: F[Future[A]])(implicit F: Async[F], cs: ContextShift[F]): F[A] =
+    F.guarantee(fa.flatMap(f => f.value match {
+      case Some(result) =>
+        result match {
+          case Success(a) => F.pure(a)
+          case Failure(e) => F.raiseError[A](e)
+        }
+      case _ =>
+        F.async[A] { cb =>
+          f.onComplete(r => cb(r match {
+            case Success(a) => Right(a)
+            case Failure(e) => Left(e)
+          }))(immediate)
+        }
+    }))(cs.shift)
 
   /**
    * Lifts any `IO` value into any data type implementing [[Async]].


Daenyth:
I'm not sure why the change I made fails to compile - any suggestions?

core/shared/src/main/scala/cats/effect/Async.scala
```diff
@@ -17,16 +17,16 @@
 package cats
 package effect
 
-import simulacrum._
```
Avasil: @Daenyth My guess is this one is responsible for build failure

Daenyth: Damnit intellij...

core/shared/src/main/scala/cats/effect/Async.scala
```diff
@@ -276,6 +277,22 @@ object Async {
       })
     }
 
+  def fromFuture[F[_], A](fa: F[Future[A]])(implicit F: Async[F], cs: ContextShift[F]): F[A] =
+    F.guarantee(fa.flatMap(f => f.value match {
```
Daenyth: One thing I wasn't sure of - IO wraps the `f => ...` bit in a concrete class (presumably to reduce lambda allocations?), `IOFromFuture`. Should this do something similar? Does it help at all, since it couldn't be an `object` and would need the `new FFromFuture[F]` construction anyway?

djspiewak: I think this is fine as-is. I'm actually not 100% sure why `IOFromFuture` (and friends) are a thing. It doesn't actually reduce allocations to my knowledge, since there needs to be a synthetic wrapper to put it on the heap *anyway*, so it seems to be more a code organization thing. @alexandru could explain better, for sure.

djspiewak:
@Daenyth Would you like to rebase away the "rebaseme" commits? I'm happy to merge this as-is, but just wanted to give you a chance to clean it up if that's what you wanted to do. :-)

Daenyth:
Yes, thanks. I saw the contributing note about "please do not rebase open PR branches" (which is my normal workflow).

Daenyth:
Ready to merge once CI is done.

djspiewak:
> I saw the contributing note about "please do not rebase open PR branches" (which is my normal workflow).

I appreciate your conscientiousness! I generally am opposed to public rebasing (hence the contributors note) because it prevents collaboration, but since I saw the commits and I figured it was unlikely that someone else was basing on this branch in the short term, I thought it would be worth bending the rules just this once. :-)

Thanks really so much for the contribution!

core/shared/src/main/scala/cats/effect/Async.scala
alexandru: Yes, it's a code organization thing. I hate big files.

There was also a problem with `.class` files getting to big, a problem on Android that could be avoided if the implementation of the functions was moved in other classes, so for big data types like `IO`, `Task`, `Observable`, etc, I tend to prefer that multi-line functions to be in their own file.

Issue #604, opened 2019-08-16
Make IOApp#main final?


kevinmeredith:
What's the reason that [IOApp#main](https://github.com/typelevel/cats-effect/blob/v1.4.0/core/shared/src/main/scala/cats/effect/IOApp.scala#L66) isn't `final`? My incomplete understanding is that users of `IOApp` are expected to override `def run(args: List[String]): IO[ExitCode]`, but then `main` is an implementation detail, i.e. not to be modified by the `extend`-ing class/objejct.

```scala
@ import $ivy.`org.typelevel::cats-effect:1.4.0` 
https://repo1.maven.org/maven2/org/typelevel/cats-effect_2.12/1.4.0/cats-effect_2.12-1.4.0-sources.jar
  100.0% [##########] 131.4 KiB (1.3 MiB / s)
import $ivy.$                                 

@ class BadMain extends cats.effect.IOApp { 
    override def run(xs: List[String]): cats.effect.IO[cats.effect.ExitCode] = ???
    override def main(xs: Array[String]): Unit = ???
  } 
defined class BadMain
```

Should `cats.effect.IOApp#main` be `final` to avoid this potential issue?

djspiewak:
Yes. It should be. PR! 😃

alexandru:
Note that making `main` final will break binary compatibility.

kevinmeredith:
I'm going to close this per @alexandru's explanation, https://github.com/typelevel/cats-effect/pull/605#issuecomment-522124710.

PR #605, opened 2019-08-16
Mark IOApp#main as final.

kevinmeredith:
Addresses https://github.com/typelevel/cats-effect/issues/604.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOApp.scala
+++ b/core/shared/src/main/scala/cats/effect/IOApp.scala
@@ -63,7 +63,7 @@ trait IOApp {
    * The main method that runs the `IO` returned by [[run]] and exits
    * the app with the resulting code on completion.
    */
-  def main(args: Array[String]): Unit =
+  final def main(args: Array[String]): Unit =
     IOAppPlatform.main(args, Eval.later(contextShift), Eval.later(timer))(run)
 
   /**


kevinmeredith:
Without looking in depth at the failed builds, my guess is that a flaky test was to blame. Anyway, I'll look in-depth if they fail again.

djspiewak:
`IOAppCompanionPlatform.scala:108`

Apparently we override `main`? For… reasons?

kevinmeredith:
> IOAppCompanionPlatform.scala:108
> Apparently we override main? For… reasons?

Given what you noted, @djspiewak, in https://github.com/typelevel/cats-effect/blob/master/core/jvm/src/main/scala/cats/effect/internals/IOAppCompanionPlatform.scala#L102-L122, does that mean it's incorrect to mark `def main` as `final`?

alexandru:
Ah, you can't make `main` final because indeed we've got `IOApp.WithContext`.

`IOApp.WithContext` is a helper meant to provide a way to override the `ExecutionContext` via a `Resource`. The normal `IOApp` uses a thread-pool for daemonic threads, whereas `WithContext` can use a non-daemonic thread-pool.

So no, you can't make it final.

alexandru:
Also this change should fail the Mima checks too for binary compatibility, because turning a method into a final one does break bin compat.

kevinmeredith:
I'm closing per @alexandru's explanation, https://github.com/typelevel/cats-effect/pull/605#issuecomment-522124710.

rossabaker:
I'm gradually becoming convinced that daemonic main pools are not also demonic. `WithContext` is something I'd like to revisit in CE3.

alexandru:
I think we need to revisit `IOApp` as a whole. I'm bothered that `IOApp` objects are side-effectful.

`main` should have the thread-pool injected, instead of keeping a reference around as a class member.

djspiewak:
> main should have the thread-pool injected, instead of keeping a reference around as a class member.

I mean, does it really matter though? `IOApp` is *the* end of the world, barring sbt shenanigans. What advantage would be gained from injecting the thread pool rather than just effectfully creating it?

alexandru:
`main` would be easier to test, being just an `IO` value. Right now the `IO` value from `main` captures shared state and is not pure. Works for practical purposes, but it's also setting a bad example.

Although if we adopt Monix's design for injecting the `ExecutionContext` in `unsafeRun`, the problem would take care of itself, but not for `Timer` (Monix has `Scheduler` which can also schedule stuff with delays, so `Timer` is also pure).

djspiewak:
> main would be easier to test, being just an IO value. Right now the IO value from main captures shared state and is not pure. Works for practical purposes, but it's also setting a bad example.

I think `main` is pretty much as easy/hard to test regardless. We can't really *do* anything with the `IO` value other than run it, which of course is just as impure as the `ExecutionContext` it closes over.

We're definitely going to adopt Monix's context reader design in CE3. There are just so many advantages and very very few reasons not to. Whether or not we include the timer with that is a different question. I tend to think not, given that just having a single timer per system seems to be optimal in literally every use case.

SystemFw:
> I tend to think not, given that just having a single timer per system seems to be optimal in literally every use case.

injecting `TestContext` for tests through implicits is suboptimal imho

PR #612, opened 2019-08-28
Fix bracketReleaseIsCalledOnCompletedOrError

neko-kai:
In bracketReleaseIsCalledOnCompletedOrError, account for the fact that `fa` action might spawn internal forks and may not be equal to F.pure by TestContext

### DIFF
--- a/laws/shared/src/main/scala/cats/effect/laws/AsyncLaws.scala
+++ b/laws/shared/src/main/scala/cats/effect/laws/AsyncLaws.scala
@@ -97,7 +97,7 @@ trait AsyncLaws[F[_]] extends SyncLaws[F] {
       // to run and set the promise.
       F.asyncF[Unit](cb => F.delay(cb(Right(()))) *> br.attempt.as(())) *> promise.get
     }
-    lh <-> F.pure(b)
+    lh <-> fa.attempt.as(b)
   }
 }
 


neko-kai:
@djspiewak 
The cause of mismatches seems to be not auto-forking, but rather deliberate forking in `Gen`erated IO's - https://github.com/zio/zio/blob/062024e3e9abdf92e36fa138a1eafee9eba5433f/core-tests/shared/src/test/scala/zio/GenIO.scala#L110

Given that there may be multiple nested `race`/`par`s in the generated IO, I think this test cannot in general assert equivalence wrt `F.pure` (or we must stop generating forking IOs).

I'm quite befuddled what to do next, tbh, maybe Alex could help since he added this law (according to git)?
/cc @alexandru

djspiewak:
> Given that there may be multiple nested race/pars in the generated IO, I think this test cannot in general assert equivalence wrt F.pure (or we must stop generating forking IOs).

Or at the very least, we have to rely on a very strange and limited definition of equality. Neither is a great alternative.

`.attempt.as` is definitely not the way to go, since this swallows errors and I suspect only works downstream in ZIO because of coincidences in how those constructs are being evaluated. I'm not really sure. In theory, your generator is doing the right thing, but it's throwing away some information when it does so. Specifically, if I understand you correctly, it is constructing forking `IO`s which are guaranteed to be externally idempotent, but that guarantee is lost by the time it winds its way over to the `Equal` instance. Thus, in theory, this is something that your `Equal` instance should be accounting for, though I really don't have any good ideas as to *how* this could be accomplished.

I don't like this law *at all*, but I'm really hesitant to remove it in Cats 2.0. So… not sure the best approach.

neko-kai:
@djspiewak 

> .attempt.as is definitely not the way to go, since this swallows errors and I suspect only works downstream in ZIO because of coincidences in how those constructs are being evaluated.

`.attempt.as` here just mirrors the `fa.attempt.as` on the left hand side (in `br.attempt`), so it appears logical to me...

> Specifically, if I understand you correctly, it is constructing forking IOs which are guaranteed to be externally idempotent, but that guarantee is lost by the time it winds its way over to the Equal instance.

Well, they cannot be equivalent by TestContext.tick() to F.pure – given multiple nested races they may require more than one tick from executor to evaluate – this is only a problem with this specific law, so I suppose it's the only law currently that has generated `fa: F[A]` on only one side of the equation. Note that the problem here – although I might be completely misunderstanding how TestContext works – is just that `fa` is only on the left-hand side, causing the sides to have a nearly guaranteed mismatch and inequality on TestContext `tick`s.

djspiewak:
> .attempt.as here just mirrors the fa.attempt.as on the left hand side (in br.attempt), so it appears logical to me...

Ahhhh right I'm dumb. You're correct so long as `complete` doesn't raise an error, which it shouldn't. I hate this law… It's so complicated and tangled.

> Well, they cannot be equivalent by TestContext.tick() to F.pure – given multiple nested races they may require more than one tick from executor to evaluate – this is only a problem with this specific law, so I suppose it's the only law currently that has generated fa: F[A] on only one side of the equation. Note that the problem here – although I might be completely misunderstanding how TestContext works – is just that fa is only on the left-hand side, causing the sides to have a nearly guaranteed mismatch and inequality on TestContext ticks.

I think what it comes down to is the `Equal` here knows both too much and too little about the internals of the effect under test. It knows too much because it knows that one requires extra ticks, but it also knows too little because it doesn't know that the actions involved in those ticks eliminate each other to identity. There probably isn't a nice solution to this.

I'm 👍 on this if we can figure out why the inductive RWST instance broke.

alexandru:
@djspiewak 

> I don't like this law at all

The laws for `Bracket` got messy — but we have to test somehow that `release` gets triggered, otherwise the operation is lawless.

The reason for why the law is the way it is is because if you do something like this:

```scala
for {
  promise <- Ref[F].of(false)
  _ <- F.guaranteeCase(fa) { case Completed => promise.set(true) }
  r <- promise.get
} yield r
```

It may not seem apparent at first, but this basically assumes `IO`. And that's a pretty bad assumption to make, unless there is another law that specifies `IO`-like behavior.

A streaming data type for example, or something like `Resource`, will postpone that completion handler until _after_ the `promise.get` operation. In other words the `flatMap` operation cannot be relied on for specifying a certain order of execution in this case. We don't have an operation that says _do this after whatever `release` handle was specified before_.

FYI my `Observable` [implements Bracket](https://github.com/monix/monix/blob/452194b/monix-reactive/shared/src/main/scala/monix/reactive/Observable.scala#L5813) for real.

So keeping this fact in mind, that `flatMap` no longer works to test this, the only other way I could think of was to do a "*fire and forget*". This is where this super weird line comes in:

```scala
F.asyncF[Unit](cb => F.delay(cb(Right(()))) *> br.attempt.as(()))
```

I'm abusing `asyncF` here to trigger our task and eventually the release. And this code should work even if the implementation of `asyncF` does not fork.

Although there is the small assumption that either the implementation of `asyncF` will fork, or that the `release` of bracket will execute before `promise.get`. This assumption isn't necessarily true, but it's a better assumption to have than that of `F[_]` being `IO` or `MonadTransformer[IO]`.

In the laws of `Monad` there are no laws that assume IO-like behavior for `flatMap` which is what makes `Monad` so freaking awesome. We should strive to do the same.

laws/shared/src/main/scala/cats/effect/laws/AsyncLaws.scala
```diff
@@ -97,7 +97,7 @@ trait AsyncLaws[F[_]] extends SyncLaws[F] {
       // to run and set the promise.
       F.asyncF[Unit](cb => F.delay(cb(Right(()))) *> br.attempt.as(())) *> promise.get
     }
-    lh <-> F.pure(b)
+    lh <-> fa.attempt.as(b)
```
alexandru: This would unfortunately not work as was originally intended.

The left-hand side (`lh`) is supposed to, in all circumstances for `F[_]`, signal a single value.

`fa.attempt.as(b)` on the other hand can be an empty stream, or a stream of multiple values.

In other words it is introducing the assumption that I've been trying to prevent.

alexandru (commented):
Not sure how the change helps, I haven't read your entire conversation yet, however I believe this change will break compatibility ...

alexandru:
> Although there is the small assumption that either the implementation of `asyncF` will fork, or that the `release` of bracket will execute before `promise.get`

@neko-kai Going back to what I said before, I believe this could be your problem, right?

---

If you guys have any idea of how to test that `release` works without assuming `IO.flatMap`, I would be delighted to see this implemented in a better way.

Maybe we could simply do a double `bracket` :-)

alexandru:
Wait, I don't know why I haven't seen it before — the problem might be that we're going through `Deferred`, but that's totally unnecessary:

```scala
val lh = F.asyncF[Unit] { cb =>
  F.bracketCase(F.delay(promise)) { _ =>
    fa
  } {
    case (r, Completed | Error(_)) => F.delay(cb(b))
    case _ => F.unit
  }
}
```

neko-kai:
@alexandru
I'll try that, thanks! :+1:

Note however that the principal mismatch remains - once `fa` is executed on the left, it might take more ticks to complete than 1 tick of F.pure on the right - I think this is the issue.

Also, this law is in AsyncLaws, if monix Observable implements only Bracket it may not stress that law at all! Async F's seem very IO-like to me already - I wouldn't know how to implement asyncF for a Resource/Stream-like monad - but I may be completely wrong.

alexandru:
> it might take more ticks to complete than 1 tick of F.pure on the right

That shouldn't matter.

What matters, and what might be happening, is the order in which those ticks execute — we also deliberately introduced non-determinism in `TestContext` precisely to highlight cases such as this one.

alexandru:
Did a quick test and I think this might be good, but you have to confirm it — try it with this:

```scala
  def bracketReleaseIsCalledOnCompletedOrError[A, B](fa: F[A], b: B) = {
    val lh = F.asyncF[B] { cb =>
      F.bracketCase(F.pure(cb)) { _ =>
        fa.as(())
      } {
        case (cb, Completed | Error(_)) => F.delay(cb(Right(b)))
        case _ =>  F.unit
      }
    }
    lh <-> F.pure(b)
  }
```

Note that in Monix at least, if the task in `asyncF` fails, we log the error even if the task is already completed due to the invocation of that callback — so depending on your implementation, this might generate some logging noise that you might want to silence.

We cannot silence the errors from this code unfortunately. Because doing `attempt` on that would mean that `case Error(_)` never happens.

So I guess this is the last remaining caveat, which I don't see as much of a big deal.

alexandru:
@neko-kai 

> Also, this law is in AsyncLaws, if monix Observable implements only Bracket it may not stress that law at all!

Indeed, but I tested it manually and I think I have it in my own unit tests.

I can implement `asyncF` for `Observable`, not a big deal, would go something along the lines of ...

```scala
def asyncF[A](k: (Either[Throwable, A] => Unit) => Observable[Unit]): Observable[A] =
  Observable.unsafeCreate { out =>
    val task = k {
      case Right(a) => 
        out.onNext(a)
        out.onComplete()
      case Left(e) => 
        out.onError(e)
    }
    task.subscribe(Subscriber.empty(out.scheduler))
  }
```

The only reason I haven't implemented `Async` is that I don't have a stack safe `flatMap` implementation. No other Rx implementation that I know of has it, because it's pretty messy to implement. But I might give it a shot in Monix 4.0.0.

So it is my goal to have an `Async[Observable]` :-)

neko-kai:
@alexandru 
Ok, trying your implementation (https://github.com/typelevel/cats-effect/pull/614) at https://github.com/zio/interop-cats/pull/48
It's way cleaner, thank you! 
It works as long as I forbid generation of `race`/`par` so far, I think it should work in general.

I can also see from it that ZIO's current `asyncF` is unlawful wrt error propagation, this can be observed in your test (which is probably a bug in ZIO) specifically for all the laws I opened in https://github.com/typelevel/cats-effect/pull/615

PR #614, opened 2019-08-29
Fix AsyncLaws.bracketReleaseIsCalledOnCompletedOrError

alexandru:
Another attempt at fixing #612.


alexandru:
This might be good, but will pass it on to @neko-kai in #612, as this needs confirmation that it fixes his use-case.

PR #615, opened 2019-08-30
Add asyncF law that body of `k` CANNOT affect continuation

neko-kai:
Add async laws that specify that content of register function CANNOT affect continuation (`asyncF raiseError is never` and `asyncF ignored callback is never`)

Also add async & asyncF laws for catching throws, which I think are important to have since these are FFI (as much as `delay` ?)

Note, ZIO fails all of the new laws, while cats-effect IO passes all of them, which is why I think it's important to have them since leaving them underspecified can produce deadlocks in downstream code

### DIFF
--- a/laws/shared/src/main/scala/cats/effect/laws/AsyncLaws.scala
+++ b/laws/shared/src/main/scala/cats/effect/laws/AsyncLaws.scala
@@ -81,6 +81,18 @@ trait AsyncLaws[F[_]] extends SyncLaws[F] {
   def neverIsDerivedFromAsync[A] =
     F.never[A] <-> F.async[A]( _ => ())
 
+  def asyncFRaiseErrorIsNever[A](e: Throwable) =
+    F.never[A] <-> F.asyncF[A](_ => F.raiseError(e))
+
+  def asyncFIgnoredCallbackIsNever[A](fa: F[Unit]) =
+    F.never[A] <-> F.asyncF[A](_ => fa)
+
+  def asyncThrowIsRaiseError[A](e: Throwable) =
+    F.async[A](_ => throw e) <-> F.raiseError(e)
+
+  def asyncFThrowIsRaiseError[A](e: Throwable) =
+    F.asyncF[A](_ => throw e) <-> F.raiseError(e)
+
   def asyncCanBeDerivedFromAsyncF[A](k: (Either[Throwable, A] => Unit) => Unit) =
     F.async(k) <-> F.asyncF(cb => F.delay(k(cb)))
 
--- a/laws/shared/src/main/scala/cats/effect/laws/AsyncLaws.scala
+++ b/laws/shared/src/main/scala/cats/effect/laws/AsyncLaws.scala
@@ -87,6 +87,9 @@ trait AsyncLaws[F[_]] extends SyncLaws[F] {
   def asyncFIgnoredCallbackIsNever[A](fa: F[Unit]) =
     F.never[A] <-> F.asyncF[A](_ => fa)
 
+  def asyncFTerminationIsOptional[A](a: A) =
+    F.asyncF[A](k => F.delay(k(Right(a))) *> F.never) <-> F.pure(a)
+
   def asyncThrowIsRaiseError[A](e: Throwable) =
     F.async[A](_ => throw e) <-> F.raiseError(e)
 
--- a/laws/shared/src/main/scala/cats/effect/laws/discipline/AsyncTests.scala
+++ b/laws/shared/src/main/scala/cats/effect/laws/discipline/AsyncTests.scala
@@ -76,7 +76,7 @@ trait AsyncTests[F[_]] extends SyncTests[F] {
           default ++ Seq(
             "never is derived from async" -> Prop.lzy(laws.neverIsDerivedFromAsync[A]),
             "asyncF raiseError is never" -> forAll(laws.asyncFRaiseErrorIsNever[A] _),
-            "asyncF ignored callback is never" -> forAll(laws.asyncFIgnoredCallbackIsNever[A] _),
+            "asyncF ignored callback is never" -> forAll(laws.asyncFIgnoredCallbackIsNever[A] _)
           )
         else
           default
--- a/laws/shared/src/main/scala/cats/effect/laws/discipline/AsyncTests.scala
+++ b/laws/shared/src/main/scala/cats/effect/laws/discipline/AsyncTests.scala
@@ -67,7 +67,8 @@ trait AsyncTests[F[_]] extends SyncTests[F] {
           "async can be derived from asyncF" -> forAll(laws.asyncCanBeDerivedFromAsyncF[A] _),
           "bracket release is called on Completed or Error" -> forAll(laws.bracketReleaseIsCalledOnCompletedOrError[A, B] _),
           "async throw is raiseError" -> forAll(laws.asyncThrowIsRaiseError[A] _),
-          "asyncF throw is raiseError" -> forAll(laws.asyncFThrowIsRaiseError[A] _)
+          "asyncF throw is raiseError" -> forAll(laws.asyncFThrowIsRaiseError[A] _),
+          "don't wait for asyncF to complete" -> forAll(laws.asyncFTerminationIsOptional[A] _)
         )
 
         // Activating the tests that detect non-termination only if allowed by Params,
--- a/laws/shared/src/main/scala/cats/effect/laws/discipline/AsyncTests.scala
+++ b/laws/shared/src/main/scala/cats/effect/laws/discipline/AsyncTests.scala
@@ -65,13 +65,18 @@ trait AsyncTests[F[_]] extends SyncTests[F] {
           "repeated asyncF evaluation not memoized" -> forAll(laws.repeatedAsyncFEvaluationNotMemoized[A] _),
           "propagate errors through bind (async)" -> forAll(laws.propagateErrorsThroughBindAsync[A] _),
           "async can be derived from asyncF" -> forAll(laws.asyncCanBeDerivedFromAsyncF[A] _),
-          "bracket release is called on Completed or Error" -> forAll(laws.bracketReleaseIsCalledOnCompletedOrError[A, B] _))
+          "bracket release is called on Completed or Error" -> forAll(laws.bracketReleaseIsCalledOnCompletedOrError[A, B] _),
+          "async throw is raiseError" -> forAll(laws.asyncThrowIsRaiseError[A] _),
+          "asyncF throw is raiseError" -> forAll(laws.asyncFThrowIsRaiseError[A] _)
+        )
 
         // Activating the tests that detect non-termination only if allowed by Params,
         // because such tests might not be reasonable depending on evaluation model
         if (params.allowNonTerminationLaws)
           default ++ Seq(
-            "never is derived from async" -> Prop.lzy(laws.neverIsDerivedFromAsync[A])
+            "never is derived from async" -> Prop.lzy(laws.neverIsDerivedFromAsync[A]),
+            "asyncF raiseError is never" -> forAll(laws.asyncFRaiseErrorIsNever[A] _),
+            "asyncF ignored callback is never" -> forAll(laws.asyncFIgnoredCallbackIsNever[A] _),
           )
         else
           default
--- a/laws/shared/src/test/scala/cats/effect/LTask.scala
+++ b/laws/shared/src/test/scala/cats/effect/LTask.scala
@@ -24,6 +24,7 @@ import cats.syntax.flatMap._
 import cats.syntax.functor._
 
 import scala.concurrent.{ExecutionContext, ExecutionException, Future, Promise}
+import scala.util.control.NonFatal
 import scala.util.{Either, Left, Right}
 
 /**
@@ -64,15 +65,23 @@ object LTask {
       def async[A](k: (Either[Throwable, A] => Unit) => Unit): LTask[A] =
         LTask { _ =>
           val p = Promise[A]()
-          k(r => p.tryComplete(Conversions.toTry(r)))
-          p.future
+          try {
+            k(r => p.tryComplete(Conversions.toTry(r)))
+            p.future
+          } catch {
+            case NonFatal(e) => Future.failed(e)
+          }
         }
 
       def asyncF[A](k: (Either[Throwable, A] => Unit) => LTask[Unit]): LTask[A] =
         LTask { implicit ec =>
           val p = Promise[A]()
-          k(r => p.tryComplete(Conversions.toTry(r))).run(ec)
-          p.future
+          try {
+            k(r => p.tryComplete(Conversions.toTry(r))).run(ec)
+            p.future
+          } catch {
+            case NonFatal(e) => Future.failed(e)
+          }
         }
 
       def runAsync[A](fa: LTask[A])(cb: Either[Throwable, A] => IO[Unit]): SyncIO[Unit] =


iravid:
To me, it seems that there's a semantic mismatch between these two:
```scala
  def asyncFRaiseErrorIsNever[A](e: Throwable) =
    F.never[A] <-> F.asyncF[A](_ => F.raiseError(e))

  def asyncFThrowIsRaiseError[A](e: Throwable) =
    F.asyncF[A](_ => throw e) <-> F.raiseError(e
```

I would expect the failure of the registration function to always result in failed value - whether it is because of a thrown exception, or a raised error.

neko-kai:
@iravid 
That would introduce a race condition – by the time the task returns with a failure the `k` continuation is already out in the wild. Actually the same applies to `throw`, but here an assumption can be made that a side-effectful callback will be passed around after `F[Unit]` is constructed instead of as a side-effect during construction.

iravid:
That is true, but why is this different than calling the callback more than once? What's the behavior there?

What makes sense to me is for the platform to report an unhandled error for double-calling the callback, and in this case as well.

neko-kai:
@iravid 
The behavior is that the overall result of computation will change between cats-effect and ZIO from a trailing failure if the failure is reached before callback – in ZIO case the failure will overtake, but in cats-effect the failure will be logged and ignored and callback result will overtake (which tbh I think is correct as it leaves only one channel for the computation to continue)

iravid:
Ok, I'm fine with that behavior for consistency. I guess we should change back the registration in ZIO to `ZIO[R, Nothing, _]`.

What are we going to do if the registration function in ZIO returns a `ZIO.dieMessage`?

neko-kai:
@iravid 
Probably just `.run` and report everything via `reportFailure` ala this impl – https://github.com/zio/interop-cats/blob/d26973d0fcebc1dfe621423f91c4f81c7f7da65d/interop-cats/shared/src/main/scala/zio/interop/cats.scala#L221

djspiewak:
We used to have laws like these, but Monix failed them so we pulled them out and left it unspecified. I generally agree that it *should* be specified, but we couldn't agree on behavior.

I strongly agree with @iravid that we should be consistent between `throw` and `raiseError`. In fact, I'm pretty sure we *need* to given the fact that we claim `async` is derivable from `asyncF` using `delay`, and `delay(throw t) <-> raiseError(t)`.

I really *want* `asyncF(_ => raiseError(e)) <-> raiseError(e)`. It does introduce a race condition but it's a race which is deterministically resolved in one direction, but still a race. I would also accept `never` as a semantic here, but that *forces* downstream implementors to have side-channel reporting for such errors, and I want to get away from that kind of thing, not introduce more of it. I'd really like to hear @alexandru's thoughts on this one.

So tldr, I *really* like specifying this behavior in the laws, and I think you're quite correct that it needs to be constrained to avoid people relying on undefined semantics. The laws do need to be consistent with each other though, and with `SyncLaws`, so we need to pick one semantic or another. I prefer equating to `raiseError(e)` rather than to `never`, to avoid side-channels, but if that's not realistic to implement in certain run loop strategies, then I could be content with `never`.

neko-kai:
@djspiewak
> I really want asyncF(_ => raiseError(e)) <-> raiseError(e). It does introduce a race condition but it's a race which is deterministically resolved in one direction
I don't think it is deterministic at all! The action may be arbitrarily long and the failure may come at an arbitrary point before, during or after the callback is called. Note that the callback may be released early, but the action may continue for an arbitrary period of time.

neko-kai:
If we had a law specifying that the setup action must return before the callback is followed it could be deterministic (failure always taking priority), but I don't think that's how it works currently at all though.

djspiewak:
> I don't think it is deterministic at all! The action may be arbitrarily long and the failure may come at an arbitrary point before, during or after the callback is called. Note that the callback may be released early, but the action may continue for an arbitrary period of time.

I mean, it's still deterministic unless the user creates a race. For example:

```scala
F asyncF { cb =>
  F.delay(new Thread { override def run = cb(Right(())) }.start) *> F.raiseError(e)
}
```

The user has *created* a race here. There isn't an inherent race without something like this.

We already have an enforced semantic which is almost identical to this, namely that repeated evaluation is ignored. Here's another race condition which is *already* required by law:

```scala
F asyncF { cb =>
  F.delay(new Thread { override def run = cb(Right(true)) }.start) *>
    F.delay(new Thread { override def run = cb(Right(false)) }.start)
}
```

This is non-deterministic whether it will return `true` or `false`. This simply requires a write-once variable inside of the implementation. It basically says that `F.raiseError(e)` in the `asyncF` body is equivalent to `F.delay(cb(Left(e)))`, which I think is intuitive. Whether that wins the race or not is not deterministic. I don't think it's horrible to report that error on a side channel if it loses, but there's already a similar problem associated with repeated callbacks, so whatever solution is chosen there should be applied here.

neko-kai:
@djspiewak
But the entire point of asyncF is to supply the callback to someone external who would call it later, I.e. all useful uses of asyncF create a race, the only difference with your example is that external thread likely already exists, not created in asyncF

djspiewak:
What's the difference between my two examples though? Remember the second one is *already* enforced by law. If the first one is a problem, then so is the second.

neko-kai:
@djspiewak 
They seem different since they're at different levels of evaluation, failure in the effect that passes the callback is not the same as the failure directly signalled through the callback (and not the same as failure to construct the effect in the first place)
Either way, changing this would require changing cats IO, monix and that https://github.com/typelevel/cats-effect/pull/612 / https://github.com/typelevel/cats-effect/pull/614 law – since the _presence_ of this race condition in ZIO causes it to fail currently...

djspiewak:
> They seem different since they're at different levels of evaluation, failure in the effect that passes the callback is not the same as the failure directly signalled through the callback (and not the same as failure to construct the effect in the first place)

I really don't think they're at different levels of evaluation at all. Basically, I don't see why there should be a difference between these two constructs:

```scala
asyncF { cb =>
  F.delay(cb(Left(e1))) *> F.raiseError(e2)
}

asyncF { cb =>
  F.delay(cb(Left(e1))) *> F.delay(cb(Left(e2)))
}
```

Additionally, let's consider things in light of existing laws. The relevant entries, all pasted below for convenience:

```scala
  def delayThrowIsRaiseError[A](e: Throwable) =
    F.delay[A](throw e) <-> F.raiseError(e)

  def asyncLeftIsRaiseError[A](e: Throwable) =
    F.async[A](_(Left(e))) <-> F.raiseError(e)

  def repeatedAsyncEvaluationNotMemoized[A](a: A, f: A => A) = F.suspend {
    var cur = a

    val change: F[Unit] = F.async { cb =>
      cur = f(cur)
      cb(Right(()))
    }

    val read: F[A] = F.delay(cur)

    change *> change *> read
  } <-> F.pure(f(f(a)))

  def asyncCanBeDerivedFromAsyncF[A](k: (Either[Throwable, A] => Unit) => Unit) =
    F.async(k) <-> F.asyncF(cb => F.delay(k(cb)))
```

`repeatedAsyncEvaluationNotMemoized` is extremely difficult to work with in these terms (one of the problems with the current law set), but I think we can roughly agree on what it means: first result wins when you have repeated callbacks, and all others are ignored.

Anyway, what we're trying to do is reason about the following expressions:

- `F.asyncF[A](_ => F.raiseError(e))`
- `F.async[A](_ => throw e)`

You're adding laws which equate the former to `never` and the latter to `raiseError`, but by the laws I pasted above, they're the same thing:

- `F.asyncF[A](_ => F.raiseError(e)) <-> F.asyncF[A](_ => F.delay(throw e))` (by `delayThrowIsRaiseError`)
- `F.asyncF[A](_ => F.delay(throw e)) <-> F.async[A](_ => throw e)` (by a slight generalization of `asyncCanBeDerivedFromAsyncF`)

So they most definitely need to produce the same result. Now, the real question we need to answer is whether or not the following holds:

```scala
F.async(_(Left(e))) <-> F.async(_ => throw e)
```

If this is true, then everything becomes `raiseError`. If this is false, then everything becomes `never`. (side note: I can see wiggle room on your `asyncFThrowIsRaiseError` law, which seems reasonable for implementations like `IO` which catch in pure code, but perhaps less reasonable for implementations like ZIO which do not)

I tend to think that the above equivalence *should* hold, and throwing in the body of `async` (and ignoring the callback) should be equivalent to invoking the callback with `Left`. This immediately implies the same must hold for `asyncF` and `raiseError`. It also means it would also hold with first-wins semantics for any situation where the callback is invoked *and* an error is produced, as is already the case by law for non-error results.

alexandru:
This law hasn't been true for Monix, until 3.0.0-RC4 I believe:

```scala
F.asyncF[A](_ => throw e) <-> F.raiseError(e)
```

In order to guarantee such a law, you need a synchronized callback. This is because you need to preserve the contract:

1. the given callback can only be called at most once
2. if you give that callback to user code, then you don't know what the user code did with it, unless you synchronize it

Due to recent changes those callbacks are now synchronized in Monix as well, because users do break that contract — it's a funny thing, because on the JVM nothing is thread safe unless it's mentioned in the documentation and yet people end up calling that callback concurrently — I blame Scala's `Promise` for that.

I also like to look at prior art — and most specifically at [Reactive Streams](http://www.reactive-streams.org/), which has been adopted in `java.util.concurrent.Flow`, making it the most popular API for signaling results asynchronously. What I like about the Reactive Streams specification is that it is making demands from the *producer* to be thread safe and not from the implementation.

By contract you shouldn't call `onComplete` or `onError` multiple times concurrently and implementations that do it are broken.

I'm also not super comfortable with the current behavior of Monix and what is suggested here, not only because we would demand synchronized callbacks from the implementations, but also because the behavior is now inconsistent:

```scala
F.asyncF { cb =>
  F.delay(cb(Right(())) *> F.raiseError(e)
}
```

In Monix this gets logged — since you can't call that callback twice and it was already called. You can't specify this logging behavior via laws though.

However both approaches — logging versus signaling the error if possible (`raiseError` versus `never`), are perfectly valid in my book. And I don't see why we should impose on implementations a specific behavior here.

And that's because we are talking of broken code:

```scala
F.asyncF(_ => throw e)

F.asyncF(_ => F.raiseError(e))
```

In both cases, this is broken code that shouldn't happen in generic code or otherwise. That we do anything about it, it's only to inform the user that something is wrong with their code, but if you specify behavior via laws, then that behavior will be relied upon, making such code valid.

So I would leave it as currently is, unspecified.

neko-kai:
Note that I've added another law `asyncFTerminationIsOptional`, which I also think is important since I nearly merged a change that would break it – and deviate from expected behavior of cats-effect without knowing it or being caught by other laws. (travis build error is unrelated)

@djspiewak 
> by the laws I pasted above, they're the same thing:

I agree with this reasoning, although we could also divorce `async` and `asyncF` and just let asyncF hold the current rules. All I want is consistency for better interop.

> side note: I can see wiggle room on your asyncFThrowIsRaiseError law, which seems reasonable for implementations like IO which catch in pure code, but perhaps less reasonable for implementations like ZIO which do not

All implementations must catch in FFI code IMHO. More so, I support specifying catching behavior in ALL functions – which would outlaw a non-Throwable ZIO Monad instance, but at least it would ensure a higher level of interoperability.

@alexandru 

> And that's because we are talking of broken code: In both cases, this is broken code that shouldn't happen in generic code or otherwise.

IMHO these are all valid ways to return from an FFI call and ought to be covered, if anything, at least to ensure consistency with by `async<->delay<->asyncF` equivalence as per Daniel above. Also, code that doesn't happen just plain does, and when it does, it spells implementation-dependent asynchronous deadlocks that are near impossible to debug which is bad for users, for 3rd party libraries and for maintainers.

IMHO, possible deviations from cats-effect IO should be specified if not as mandatory laws, at least as an additional set of laws such that:
1. Third-parties know which behaviors are implementation-dependent and may avoid them
2. Maintainers know the axis they're "allowed" to vary on "safely", beside the somewhat skimpy set of laws currently

djspiewak:
@neko-kai 

> I agree with this reasoning, although we could also divorce async and asyncF and just let asyncF hold the current rules. All I want is consistency for better interop.

Agreed. As for divorcing the two, I'm a little skeptical of that. I would honestly rather bring them closer together (e.g. default-implementing `async` in terms of `asyncF` and `delay`).

> All implementations must catch in FFI code IMHO. More so, I support specifying catching behavior in ALL functions – which would outlaw a non-Throwable ZIO Monad instance, but at least it would ensure a higher level of interoperability.

Agreed. I'm just not sure if the pure definition of the effect returned to `asyncF` is actually FFI code. I'm also very much in favor of catching in all functions due the nature of concurrent programming on the JVM, but yeah it definitely causes problems for non-`Throwable` ZIO, and really bifunctor IO in general.

I'm not sure whether or not specifying that behavior is important. We've sort of punted on it over the years, since I can see a strong argument (grounded in an interpretation of the functor laws) that catching in ostensibly-pure functions like `map` is unlawful. So far, leaving it unspecified hasn't caused anybody any real problems, but it's possible. I certainly don't consider the argument closed by any means.

> IMHO these are all valid ways to return from an FFI call and ought to be covered, if anything, at least to ensure consistency by async<->delay<->asyncF equivalence as per Daniel above. Also, code that doesn't happen just plain does, and when it does, it spells implementation-dependent asynchronous deadlocks that are near impossible to debug which is bad for users, for 3rd party libraries and for maintainers.

This is my biggest concern about finding the balance here, for sure. We're basically trying to thread the needle between over-specifying implementations and accidentally opening doors for frameworks relying on implementation details and creating hard-to-diagnose bugs. Definitely specifying what should happen in case of errors thrown in the async registry function is a good candidate here and I think you're 100% correct we need to be explicit about it. It's just a question of what it should all mean.

> IMHO, possible deviations from cats-effect IO should be specified if not as mandatory laws, at least as an additional, separate and non-mandatory set of laws such that:

I'm generally 👎 on optional laws, but open to being convinced. I feel like behavior is either mandated or it isn't. Anything that isn't mandated is open season. Similarly, if you're the author of some code which is *using* Cats Effect's typeclasses, you really shouldn't be relying on anything that isn't specified in the laws, any more than you should be relying on behavior of some random `Monad` that isn't specified by the monad laws. It's just a much harder balance for us to find because our domain is so much more complex.

djspiewak:
@alexandru 

> In order to guarantee such a law, you need a synchronized callback. This is because you need to preserve the contract:

That's true of the double-callback law, too. If you're ignoring repeated callbacks, then you should be able to use that same mechanism to ignore errors coming from the registration function.

> I also like to look at prior art — and most specifically at Reactive Streams, which has been adopted in java.util.concurrent.Flow, making it the most popular API for signaling results asynchronously. What I like about the Reactive Streams specification is that it is making demands from the producer to be thread safe and not from the implementation.

Worth noting that reactive streams is a push model, while we're a pull model (due to laziness), so we have to invert some of their implications. Making demands from the producer makes sense for a push model, while making demands from the consumer makes more sense for a pull model. At least, that seems intuitive to me.

> In Monix this gets logged — since you can't call that callback twice and it was already called. You can't specify this logging behavior via laws though.

What happens in the following?

```scala
F.asyncF { cb =>
  F.delay(cb(Right(())) *> F.delay(cb(Left(e)))
}
```

IMO, those two should be consistent. There's absolutely nothing you can do with the error in *either* case except side-channel or sinkhole it. Side channels (like `printStackTrace`) are preferable to just silently swallowing, but they're morally about the same thing.

> In both cases, this is broken code that shouldn't happen in generic code or otherwise. That we do anything about it, it's only to inform the user that something is wrong with their code, but if you specify behavior via laws, then that behavior will be relied upon, making such code valid.

I mean, it's broken code, but I'm with @neko-kai that leaving it unspecified opens us up to people relying on implementation-specific details by accident. For the same reason that we specify that double callbacks are ignored (which is also broken code), I think we should specify what happens in this case.

PR #622, opened 2019-09-05
Add SyncEffect typeclass

LukaJCB:
This PR adds a new typeclass for running sync-only effect types. It uses a trick I learned from @djspiewak, so credit to him!


core/shared/src/main/scala/cats/effect/SyncEffect.scala
```diff
@@ -0,0 +1,82 @@
...
+object SyncEffect {
+
+  /**
+    * [[SyncEffect.to]] as a natural transformation.
+    */
+  def toK[F[_], G[_]](implicit F: SyncEffect[F], G: Sync[G]): F ~> G = new (F ~> G) {
+    def apply[A](fa: F[A]): G[A] = F.to[G, A](fa)
+  }
+
+  def toSyncIO[F[_], A](fa: F[A])(implicit F: SyncEffect[F]): SyncIO[A] =
```
alexandru: We should separate the type classes from the `IO` implementations we have in Cats-Effect.

In Cats-Effect 3.0 I'd like them in two separate sub-projects at the very least and this would be a circular dependency that's no longer possible.

core/shared/src/main/scala/cats/effect/SyncEffect.scala
```diff
@@ -0,0 +1,82 @@
...
+ * A monad that can suspend side effects into the `F` context and
+ * that supports lazy evaluation of these effects.
+ */
+@typeclass
+trait SyncEffect[F[_]] extends Sync[F] {
+
+  /**
+   * Convert to any other type that implements `Sync`.
+   */
+  def to[G[_], A](fa: F[A])(implicit G: Sync[G]): G[A]
```
alexandru: This is very cool. One interesting detail here is that `F[_]` cannot be `IO`, because `Sync` is not powerful enough to interpret an `IO` without `unsafeRunSync`.

For beginners it's hard to see though. Also this description is at least incomplete:

```
 * A monad that can suspend side effects into the `F` context and
 * that supports lazy evaluation of these effects.
...
 * Convert to any other type that implements `Sync`.
```

We are talking about effect types that can only be evaluated synchronously and I think that needs to be explicitly mentioned.

core/shared/src/main/scala/cats/effect/SyncIO.scala
```diff
@@ -375,12 +375,15 @@ object SyncIO extends SyncIOInstances {
...
 private[effect] abstract class SyncIOInstances extends SyncIOLowPriorityInstances {
-  implicit val syncIoSync: Sync[SyncIO] = new Sync[SyncIO] with StackSafeMonad[SyncIO] {
+  implicit val syncIoSync: SyncEffect[SyncIO] = new SyncEffect[SyncIO] with StackSafeMonad[SyncIO] {
     final override def pure[A](a: A): SyncIO[A] =
       SyncIO.pure(a)
     final override def unit: SyncIO[Unit] =
       SyncIO.unit
 
+    final override def to[G[_], A](fa: SyncIO[A])(implicit G: Sync[G]): G[A] =
+      G.delay(fa.unsafeRunSync())
```
neko-kai: This would mess with e.g. interruption and auto-forking on the target `G`. Should `liftIO` be used instead?

alexandru: Can you provide more details of how it can mess with interruption?

Also `SyncIO` is meant to be executed synchronously, no interruptions, no auto-forking and in this case the conversion works, you're just treating it like any other side effect.

You mean you wanted to interpret it somehow such that `SyncIO` `flatMap`s become `flatMap`s on `G`? 

This might be possible, by building an interpreter for `SyncIO` here, however currently `SyncIO` is based on `IO`, so that's not possible currently, but in the future, sure.

core/shared/src/main/scala/cats/effect/SyncEffect.scala
kubukoz: Plus, its a very quick implementation that's not much more troublesome than this convenience method :)

core/shared/src/main/scala/cats/effect/SyncIO.scala
neko-kai: Interruption happens between flatMaps so just stuffing a big side-effect into delay will prevent it.

> Also SyncIO is meant to be executed synchronously, no interruptions, no auto-forking and in this case the conversion works, you're just treating it like any other side effect.

> You mean you wanted to interpret it somehow such that SyncIO flatMaps become flatMaps on G? This might be possible, by building an interpreter for SyncIO here, however currently SyncIO is based on IO, so that's not possible currently, but in the future, sure.

Well, if it's meant to be treated as any other side effect then interpreting into G would be non-sensical, so we first have to decide on one of the semantics – does SyncEffect mean just running it or does it mean reinterpretation – that preserves properties of the target monad, but also requires a SyncEffect monad to have non-naive bracket model that can still work when interruption is possible? 
IMHO varying the semantic – letting some monads just delay, while others reinterpret – is unsafe, but it's just my opinion.

alexandru: > does SyncEffect mean just running it or does it mean reinterpretation

Does it really matter?

Should be up to the one implementing the instance. Ideally it should be a reinterpretation. However just the ability to execute `F[_]` is useful. Consider that:

1. Reinterpreting `F[_]` as `G[_]` requires knowledge of `F[_]`'s internals — you would have to look at internal state to unwrap `flatMap` calls, in order to build `G[_]` ... and due to SyncIO's implementation, right now we can't do it, since its internals don't reflect its capabilities
2. We can't specify a reinterpretation by law — how would you describe it?
3. For many purposes reinterpretation isn't relevant, but I agree it is nice when available ... consider that most uses of `SyncIO` in common practice is for initializing resources (e.g. file handles)

Therefore I would say it's up to the implementation.

If the author can provide a `SyncEffect[SyncIO]` instance that can reinterpret `F[_]`, then awesome, otherwise we'll take what we can get.

Do you agree?

alexandru: Although now that I think of it, this is also an efficiency issue.

Reinterpreting `SyncIO` into a Monix `Coeval` doesn't make much sense as you'd gain nothing by it 🤔

LukaJCB:
I got around bin compat by making the original method non-private but deprecated. :) If there's a better way please lmk

core/shared/src/main/scala/cats/effect/SyncIO.scala
LukaJCB: To me, this type class is a lot more about running than reinterpreting, it gives me a way to run code that's abstract in `F[_]` that I know to be sync only. In that way it's pretty much just like the other `*Effect` type classes. :)

neko-kai: I think I've had an incorrect model of where I'd like to use this typeclass, i.e. enforce a split of an application into parts that can use Concurrent primitives (and can deadlock) and parts that can't (non-termination of which is at least easily debuggable). But, SyncIO and all SyncEffect implementors can't use asynchronous IO, they can't play a delimiting role in most apps and should, ideally, be used exclusively for CPU-bound computations – which means a lack of interruption and fairness is probably not that dangerous for valid uses of `SyncEffect`.

Then I think it would be appropriate to mandate, recommend or at least make it clear that the user must be prepared to deal with running as the implementation.

> We can't specify a reinterpretation by law — how would you describe it?

Well, If we mandate running insread, we can copy a bunch of laws from bracket and delay.

core/shared/src/main/scala/cats/effect/SyncEffect.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  /**
+   * [[SyncEffect.to]] as a natural transformation.
+   */
+  def toK[G[_]](implicit G: Sync[G]): F ~> G = new (F ~> G) {
```
djspiewak: Because of the nicer type inference here, I suspect that `toK` is the more useful function for almost everyone. Do you think that merits a more first-class naming? I would almost thing that `to` should return the natural transformation and there shouldn't be a two type parameter alternative at all, but I'm not 100% sure.

core/shared/src/main/scala/cats/effect/SyncEffect.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  private[effect] trait WriterTSyncEffect[F[_], L] extends SyncEffect[WriterT[F, L, ?]]
+    with Sync.WriterTSync[F, L] {
+
+    protected def F: SyncEffect[F]
+    protected def L: Monoid[L]
+
+    def to[G[_], A](fa: WriterT[F, L, A])(implicit G: Sync[G]): G[A] =
+      F.to(F.map(fa.run)(_._2))
+  }
```
djspiewak: I think this is missing `RWST` instances.

laws/shared/src/main/scala/cats/effect/laws/SyncEffectLaws.scala
```diff
@@ -0,0 +1,35 @@
...
+package effect
+package laws
+
+import cats.laws._
+
+trait SyncEffectLaws[F[_]] extends SyncLaws[F] {
+  implicit def F: SyncEffect[F]
+
+  def toSyncIOAndBackIsIdentity[A](fa: F[A]) = {
+    SyncEffect[SyncIO].to[F, A](F.to[SyncIO, A](fa)) <-> fa
```
djspiewak: About the only reasonable law we can have here, tbh

core/shared/src/main/scala/cats/effect/SyncEffect.scala
neko-kai: Personally, I'd avoid either `to` or `toK` here, because, assuming a purely pragmatic perspective, most of the time this transformation will/should embed `F` as a single, uncancelable, synchronous effect node. A name like `runSync` or `embedSync` could at least hint to the reader that something "Syncy" / potentially uncancelable might happen here.

core/shared/src/main/scala/cats/effect/SyncEffect.scala
LukaJCB: I don't think we can have `RWST` instances for the same reason we can't have `Kleisli` and `StateT` instances. It would require us to run the inner functions. I can't think of a way to go from e.g. `Kleisli[SyncIO, A, B]` to `SyncIO[B]` without an `A`.

core/shared/src/main/scala/cats/effect/SyncEffect.scala
LukaJCB: `runSync` is probably a good idea 👍

LukaJCB:
CI is failing due to binary incompatability on 2.11. `cats-core` has relaxed this requirement on 2.11 and it probably makes sense to do the same thing for `cats-effect`. I'm going to open an issue.

LukaJCB:
Blocked by #667

LukaJCB:
Any objections to merge this? Already got 2 sign-offs, but it's been a while.

djspiewak:
@LukaJCB No objection at all, but it does need a reformat first!

LukaJCB:
Done :)

PR #635, opened 2019-09-12
[naming things] CE3: Rename `yielding` to `cede`

kubukoz:
So long, `yielding`...

### DIFF
--- a/playground.scala
+++ b/playground.scala
@@ -168,7 +168,7 @@ object playground {
           }
         } yield back
 
-        def yielding: PureIO[E, Unit] = unit
+        def surrender: PureIO[E, Unit] = unit
 
         def flatMap[A, B](fa: PureIO[E, A])(f: A => PureIO[E, B]): PureIO[E, B] =
           fa.flatMap(f)
--- a/playground.scala
+++ b/playground.scala
@@ -168,7 +168,7 @@ object playground {
           }
         } yield back
 
-        def surrender: PureIO[E, Unit] = unit
+        def cede: PureIO[E, Unit] = unit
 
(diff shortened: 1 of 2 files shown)


djspiewak:
FWIW, I think `surrender` is a better name in and of itself, but `yield` has such clear precedent and connotation within this space that I think `yielding` might be better overall.

alexandru:
We should name it `yield` instead of `yielding`.

Precedent: [Thread.yield()](https://docs.oracle.com/javase/7/docs/api/java/lang/Thread.html#yield())

alexandru:
> We should name it yield instead of yielding

Replying to myself, `yield` is unfortunately a Scala keyword, so we probably can't use it.

tplaskowski:
Have you considered `cede` as an alternative?  
It' a synonym for `yield`, and it's shorter than `surrender`. Of course, if you still want to change from `yielding`.

milessabin:
Yay! Bikeshedding time!

I quite like `cede`.

Seeing as we're passing the parcel or maybe skipping a turn ... `pass`?

djspiewak:
`cede` is really super-nice, and unlike `yielding`, it is correctly conjugated.

rossabaker:
:+1: to `cede`

fsvehla:
How about `yieldNow`?

djspiewak:
@fsvehla That works as well. It's kind of a padding non-word added onto `yield` to avoid the keyword problem.

djspiewak:
Executive decision! Let's go with `cede` for now. If you feel strongly opposed to this, feel free to flame me on twitter (or here) and make your case. :-) Nothing is set in stone!

kubukoz:
@djspiewak here you go :)

PR #636, opened 2019-09-15
Cleanup and SyncEffect sanity

djspiewak:
I realized that `Sync` is kind of unrelated to `Safe`, so we only need one variant of `SyncEffect` in the hierarchy. This PR also replaces `PartiallyApplied` with `~>` (per @smarter's astute suggestion) in the two `Effect`s, though we can't do this for `Managed` because the kinds don't line up.


LukaJCB:
Hey! It was my suggestion! 😂

djspiewak:
> Hey! It was my suggestion! 

Oh, nuts! Sorry gave credit to the wrong person! 😭 I wish we had a few more arrow shapes, tbh.

Issue #623, opened 2019-09-06
Setup Scalafmt


LukaJCB:

iRevive:
Hi all,

I took scalafmt configuration from the cats project.

<details>
<summary>.scalafmt.conf</summary>

```
version = 2.0.1
align.openParenCallSite = true
align.openParenDefnSite = true
maxColumn = 120
continuationIndent.defnSite = 2
assumeStandardLibraryStripMargin = true
danglingParentheses = true
rewrite.rules = [AvoidInfix, SortImports, RedundantBraces, RedundantParens, SortModifiers]
docstrings = JavaDoc
```
</details>

There are two major cases that do not match to the scalafmt rules:
1) ScalaDoc-like comments.
I would say that almost half of all comments are written in ScalaDoc style.

2) Modifiers order.
The project widely uses `private[effect] final class`, while scalafmt enforces `final private[effect] class`.

Should I change the configuration or we can move forward with the original one?

PR #637, opened 2019-09-16
Setup scalafmt

iRevive:
This PR will close https://github.com/typelevel/cats-effect/issues/623.

I took scalafmt configuration from the cats project. The only difference is `style = defaultWithAlign`.  

There are two major cases (besides minor alignments) that do not match to the scalafmt rules:

1) ScalaDoc-like comments.
I would say that almost half of all comments are written in ScalaDoc style.

2) Modifiers order.
The project widely uses `private[effect] final class`, while scalafmt enforces `final private[effect] class`.

Should I change the configuration or the current version is fine?


kubukoz:
@LukaJCB can you take a look? This PR is likely to go stale quite often :)

.scalafmt.conf
```diff
@@ -0,0 +1,10 @@
+version = 2.0.1
+style = defaultWithAlign
```
LukaJCB: Okay, I'll let the bikeshedding commence, I don't like this automatic alignment, it creates noisy diffs when changing a single line and I think the benefit to readabilty is fairly marginal. I prefer `align=none`.

SystemFw: Big :+1: to `align=none`

iRevive: Good point. But should dependencies be aligned or not?

Not aligned:
```
addSbtPlugin("com.dwijnand" % "sbt-travisci" % "1.2.0")
addSbtPlugin("io.crashbox" % "sbt-gpg" % "0.2.0")
addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.6.0")
```

Aligned:
```
addSbtPlugin("com.dwijnand" % "sbt-travisci"    % "1.2.0")
addSbtPlugin("io.crashbox"  % "sbt-gpg"         % "0.2.0")
addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.6.0")
```

I can add alignment only for % token.

LukaJCB: I'd personally be fine with that, but have no strong opinion on this one.

kubukoz: I like auto alignment. Whitespace can be ignored from diffs so that you won't see the difference propagate to other lines if you change just one anyway.

iRevive: I've changed `align` to `none`. Can you take a look @LukaJCB?

iRevive:
@djspiewak @SystemFw @kubukoz can you have a look, please?

.scalafmt.conf
rossabaker: Looks like it's decided, but weighing in for future debates to come on other repos:

> Whitespace can be ignored from diffs so that you won't see the difference propagate to other lines if you change just one anyway.

This is true for viewing, but it still increases the merge conflict surface and makes reviewing history harder.

larsrh:
Thanks @iRevive!

PR #638, opened 2019-09-16
Don't create async boundaries for SyncIO#bracket

### LINKED ISSUE #529: IO#unsafeRunSync erroneously fails on JavaScript
Consider this program:

```scala
def safe: Unit = IO.unit.guarantee(IO.unit).unsafeRunSync
def broken = IO.unit.bracket(_ => IO(safe))(_ => IO.unit) 
```

Calling `safe` on Scala.js correctly results in a `Unit` value. However, calling `safe` from within  another nested `IO` program (e.g. `broken` here), results in the following error:

```
[info]   java.lang.UnsupportedOperationException: cannot synchronously await result on JavaScript; use runAsync or unsafeRunAsync
[info]   at java.lang.UnsupportedOperationException.fillInStackTrace(/Users/mpilquist/Development/oss/fs2/core/js/target/scala-2.12/fs2-core-test-fastopt.js:76968:14)
[info]   at java.lang.UnsupportedOperationException.<init>(/Users/mpilquist/Development/oss/fs2/core/js/target/scala-2.12/fs2-core-test-fastopt.js:77090:10)
[info]   at java.lang.UnsupportedOperationException.<init>(/Users/mpilquist/Development/oss/fs2/core/js/target/scala-2.12/fs2-core-test-fastopt.js:132545:58)
[info]   at cats.effect.internals.IOPlatform$.unsafeResync(/Users/mpilquist/Development/oss/fs2/core/js/target/scala-2.12/fs2-core-test-fastopt.js:7111:53)
[info]   at cats.effect.IO$Async.unsafeRunTimed(/Users/mpilquist/Development/oss/fs2/core/js/target/scala-2.12/fs2-core-test-fastopt.js:47886:52)
[info]   at cats.effect.IO$Async.unsafeRunSync(/Users/mpilquist/Development/oss/fs2/core/js/target/scala-2.12/fs2-core-test-fastopt.js:47843:15)
[info]   at fs2.StreamSpec.safe$1(/Users/mpilquist/Development/oss/fs2/core/js/target/scala-2.12/fs2-core-test-fastopt.js:227516:106)
[info]   at {anonymous}()(/Users/mpilquist/Development/oss/fs2/core/js/target/scala-2.12/fs2-core-test-fastopt.js:208627:23)
[info]   at scala.scalajs.runtime.AnonFunction0.apply(/Users/mpilquist/Development/oss/fs2/core/js/target/scala-2.12/fs2-core-test-fastopt.js:97704:23)
[info]   at cats.effect.internals.IORunLoop$.cats$effect$internals$IORunLoop$$loop(/Users/mpilquist/Development/oss/fs2/core/js/target/scala-2.12/fs2-core-test-fastopt.js:7376:25)
```

I believe the cause is related to the internal trampoline used when evaluating the outer program.

LukaJCB:
Should fix #529 

Before: 
```scala
scala> def safe: Unit = SyncIO.unit.guarantee(SyncIO.unit).unsafeRunSync
safe: Unit

scala> def broken = SyncIO.unit.bracket(_ => SyncIO(safe))(_ => SyncIO.unit)
broken: cats.effect.SyncIO[Unit]

scala> broken.unsafeRunSync
java.lang.UnsupportedOperationException: cannot synchronously await result on JavaScript; use runAsync or unsafeRunAsync
...
```

After:
```scala
scala> def safe: Unit = SyncIO.unit.guarantee(SyncIO.unit).unsafeRunSync
safe: Unit

scala> def broken = SyncIO.unit.bracket(_ => SyncIO(safe))(_ => SyncIO.unit)
broken: cats.effect.SyncIO[Unit]

scala> broken.unsafeRunSync

scala>
```

### DIFF
--- a/core/shared/src/main/scala/cats/effect/SyncIO.scala
+++ b/core/shared/src/main/scala/cats/effect/SyncIO.scala
@@ -32,7 +32,10 @@ import scala.annotation.unchecked.uncheckedVariance
  * async part of the computation is run and doing so on Scala.js
  * throws an exception upon encountering an async boundary.
  */
-final class SyncIO[+A] private (val toIO: IO[A]) {
+final class SyncIO[+A] private (private val io: IO[A]) {
+
+  def toIO: IO[A] =
+    IO(io.unsafeRunSync())
 
   /**
    * Produces the result by running the encapsulated effects as impure
@@ -45,7 +48,7 @@ final class SyncIO[+A] private (val toIO: IO[A]) {
    * performs side effects and throws exceptions. You should ideally
    * only call this function *once*, at the very end of your program.
    */
-  def unsafeRunSync(): A = toIO.unsafeRunSync
+  def unsafeRunSync(): A = io.unsafeRunSync
 
   /**
    * Functor map on `SyncIO`. Given a mapping function, it transforms the
@@ -54,7 +57,7 @@ final class SyncIO[+A] private (val toIO: IO[A]) {
    * Any exceptions thrown within the function will be caught and
    * sequenced in to the result `SyncIO[B]`.
    */
-  def map[B](f: A => B): SyncIO[B] = new SyncIO(toIO.map(f))
+  def map[B](f: A => B): SyncIO[B] = new SyncIO(io.map(f))
 
   /**
    * Monadic bind on `SyncIO`, used for sequentially composing two `SyncIO`
@@ -68,7 +71,7 @@ final class SyncIO[+A] private (val toIO: IO[A]) {
    * Any exceptions thrown within the function will be caught and
    * sequenced in to the result `SyncIO[B].
    */
-  def flatMap[B](f: A => SyncIO[B]): SyncIO[B] = new SyncIO(toIO.flatMap(a => f(a).toIO))
+  def flatMap[B](f: A => SyncIO[B]): SyncIO[B] = new SyncIO(io.flatMap(a => f(a).io))
 
   /**
    * Materializes any sequenced exceptions into value space, where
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/cats/effect/SyncIO.scala
```diff
@@ -175,7 +175,11 @@ final class SyncIO[+A] private (val toIO: IO[A]) {
    *        along with the result of `use` (error or successful result)
    */
   def bracketCase[B](use: A => SyncIO[B])(release: (A, ExitCase[Throwable]) => SyncIO[Unit]): SyncIO[B] =
-    new SyncIO(toIO.bracketCase(a => use(a).toIO)((a, ec) => release(a, ec).toIO))
+    flatMap { a => use(a).attempt.flatMap {
```
alexandru: You might want to use `redeemWith` instead of `attempt.flatMap`.

LukaJCB: Good call 👍

alexandru:
I think we might need to change that `toIO` conversion as well 😕

alexandru:
I'm afraid that we're now forced to do this:

```scala
def toIO[A](fa: SyncIO[A]): IO[A] = IO.delay(fa.unsafeRunSync)
```

Which we'd probably do anyway, with a custom implementation of `SyncIO`.

LukaJCB:
Hmm, will that be binary compatible?

alexandru:
Or we could make it `uncancelable` on conversion.

What do you think?

alexandru:
I think we can change it — the constructor is private and we could define our own `toIO`, as a `def` maybe?

IDK, see what Mima tells you.

LukaJCB:
Mima seems fine with it :)

mpilquist:
Looks good to me, good find!

PR #639, opened 2019-09-17
Restore MiMa checks

kubukoz:


build.sbt
```diff
@@ -159,9 +159,10 @@ val commonSettings = Seq(
 val mimaSettings = Seq(
   mimaPreviousArtifacts := {
     CrossVersion.partialVersion(scalaVersion.value) match {
-      case Some((2, 13)) => Set.empty
-      case Some((2, 12)) => Set(organization.value %% name.value % "1.0.0")
-      case _             => Set.empty
+      //2.11 has some incompatibilities in core in 2.0.0, 2.13 didn't have a release before
```
rossabaker: I think we're going to want to drop 2.11, but that can be a separate discussion.

kubukoz: Yep! We can do that next.

PR #640, opened 2019-09-17
`start` as Resource - `background`

### LINKED ISSUE #628: Add sibling of `start` that returns a Resource
I think we can all agree fibers are hard to work with and probably what users shouldn't use 90% of the time. There are plenty of valid usecases at this moment though, for example starting an IO in background and cancelling it in case the application wraps up, which can be implemented with something like `Resource(fa.start.fproduct(_.cancel))`.

We should support that usecase and others and expose a method that'd look roughly like this:

```
def startResource: Resource[F, Fiber[A]]
//or
//the nested F is `join` without access to the finalizer
def startResource: Resource[F, F[A]]
```

in Concurrent or (as syntax in ConcurrentOps && as method on IO).

Name suggestions welcome - I'd go for `background` or `startResource`. Once we've figured out the appropriate shape and name I can provide an implementation.

Gitter microdiscussions:

- https://gitter.im/typelevel/cats-effect?at=5d755e9a460a6f5a16eb7751
- https://gitter.im/typelevel/cats-effect?at=5d762ae43b1e5e5df17c8aa0
- https://gitter.im/typelevel/cats-effect?at=5d765279d5c7bd5794608c5d

kubukoz:
This includes several commits from #639 which will be properly cleaned up once the other PR is closed.

I'm not too attached to the name, but it carries a nice meaning that I believe will be friendly to users.

Closes #628

### DIFF
--- a/core/shared/src/main/scala/cats/effect/syntax/ConcurrentSyntax.scala
+++ b/core/shared/src/main/scala/cats/effect/syntax/ConcurrentSyntax.scala
@@ -61,14 +61,16 @@ final class ConcurrentOps[F[_], A](val self: F[A]) extends AnyVal {
    * This will ensure we never close the server resource unless somebody cancels the whole `application` action.
    *
    * If at some point of using the resource you want to wait for the result of the background action,
-   * you can do so by sequencing the value inside the resource.
+   * you can do so by sequencing the value inside the resource (it's equivalent to `join` on `Fiber`).
    *
-   * This will start the background process, wait 5 seconds, and cancel it in case it takes longer than that.
-   * Note: this pattern is exactly what [[cats.effect.Concurrent#timeout]] does in a more concise and less error-prone way.
+   * This will start the background process, run another action, and wait for the result of the background process:
    *
    * {{{
-   *   longProcess.background.use(await => IO.sleep(5.seconds) *> await)
+   *   longProcess.background.use(await => anotherProcess *> await)
    * }}}
+   *
+   * In case the result of such an action is canceled, both processes will receive cancelation signals.
+   * The same result can be achieved by using `anotherProcess &> longProcess` with the Parallel type class syntax.
   */
   def background(implicit F: Concurrent[F]): Resource[F, F[A]] = Resource.make(F.start(self))(_.cancel).map(_.join)
 }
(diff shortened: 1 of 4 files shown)


kubukoz:
Finally green! @djspiewak @systemfw WDYT?

core/shared/src/main/scala/cats/effect/syntax/ConcurrentSyntax.scala
```diff
@@ -36,6 +37,41 @@ final class ConcurrentOps[F[_], A](val self: F[A]) extends AnyVal {
...
+   *
+   * If at some point of using the resource you want to wait for the result of the background action,
+   * you can do so by sequencing the value inside the resource.
+   *
+   * This will start the background process, wait 5 seconds, and cancel it in case it takes longer than that.
+   * Note: this pattern is exactly what [[cats.effect.Concurrent#timeout]] does in a more concise and less error-prone way.
+   *
+   * {{{
+   *   longProcess.background.use(await => IO.sleep(5.seconds) *> await)
+   * }}}
```
djspiewak: Are you sure that's what this does? I'm pretty sure it starts a background process, then waits 5 seconds and cancels it. If it's already completed, the result is just lost.

I think the correct implementation here would be as follows:

```scala
longProcess.background.use(await => race(sleep(5.seconds), await)).map(_.toOption)
```

Though note this still isn't *quite* the same as `timeout`, since it returns an `F[Option[A]]` rather than sequencing a `TimeoutException`.

djspiewak:
I like this function *a lot*. It's so much nicer than manually juggling `start`.

core/shared/src/main/scala/cats/effect/syntax/ConcurrentSyntax.scala
kubukoz: I think my comment was wrong, but so was yours! It actually doesn't cancel, just waits for the process forever, after 5 seconds of sleeping...

kubukoz: I'll come up with a better example.

kubukoz: Actually, my dumb October brain was wrong again, your version looks correct now... I'll get back to this sometime soon.

LukaJCB:
@kubukoz needs a scalafmt run :)

kubukoz:
Oh nice, we have scalafmt! I'll rebase and format.

kubukoz:
That was a big diff... 😂

kubukoz:
Ready when you are :)

kubukoz:
If only CI would pass.

kubukoz:
@djspiewak okay, seems to be good now!

PR #661, opened 2019-10-03
bump test dependencies, remove transitive dependencies

larsrh:


build.sbt
```diff
@@ -27,11 +27,7 @@ ThisBuild / startYear := Some(2017)
 val CompileTime = config("CompileTime").hide
 val SimulacrumVersion = "1.0.0"
 val CatsVersion = "2.0.0"
-val ScalaTestVersion = "3.1.0-SNAP13"
-val ScalaTestPlusScalaCheckVersion = "1.0.0-SNAP8"
-val ScalaCheckVersion = "1.14.2"
-val DisciplineCoreVersion = "1.0.1"
-val DisciplineScalatestVersion = "1.0.0-M1"
+val DisciplineScalatestVersion = "1.0.0-RC1"
```
djspiewak: Why remove the transitives? Generally I prefer to keep as an explicit dependency any type which is directly referenced within source code, but I admit it's not a hard and fast rule.

larsrh:
It's more practical when trying to keep binary-compatible versions aligned. Also it makes scala-steward's job easier, which in turn makes our lives easier 😉

larsrh:
I'm not seeing any continued opposition here, so I'll be bold and merge this :wink:

PR #662, opened 2019-10-04
Replace tut with mdoc

iRevive:
This PR closes https://github.com/typelevel/cats-effect/issues/580.


djspiewak:
This is immense! Thank you!

PR #665, opened 2019-10-07
fix Ref.unsafe example code

xuwei-k:


core/shared/src/main/scala/cats/effect/concurrent/Ref.scala
```diff
@@ -181,9 +181,9 @@ object Ref {
    *   import cats.effect.concurrent.Ref
    *
    *   class Counter private () {
-   *     private val count = Ref.unsafe[IO](0)
+   *     private val count = Ref.unsafe[IO, Int](0)
```
xuwei-k: https://github.com/typelevel/cats-effect/issues/666 😢

djspiewak:
I don't think we can fix this by using the partially-applied type params trick while preserving binary compatibility, so probably the best fix is what you did here as well as removing the inaccurate paragraph mentioned in #666. This really kinda sucks because this function *should* be done using the partially-applied trick, but it's just… not. 😢

PR #682, opened 2019-10-23
Remove ambiguity in Clock derivation

bplommer:
This makes it possible to derive `Clock` in a monad transformer from `Timer` in the underlying effect, e.g. `Clock[OptionT[F, *]]` from `Timer[F]`.

Currently this is impossible, because the derivation can go either via `Timer[OptionT[F, *]]` or via `Clock[F]`. This change prioritises the former, removing the ambiguity.


djspiewak:
Nice catch!

PR #685, opened 2019-10-27
Fix incorrect phrase in "Unsafe" Operations part

mlrv:
`Pretty much we have been using some "unsafe" operations in the previous example` -> `We have been using some "unsafe" operations pretty much everywhere in the previous examples`


rossabaker:
Thanks!

PR #689, opened 2019-11-05
two simple syntax for `Resource`

kailuowang:
I submit this PR to gauge interest. Personally I felt that the need to write a lot of `Resource.liftF` is a bit cumbersome, but this syntax, TBH isn't a huge improvement either.


rossabaker:
Mildly :-1:. I like to be conservative in a foundation library, and I don't see enough gain in concision to justify the magic here.

kailuowang:
I think I am convinced there isn't enough value in this to justify the extra code.

PR #690, opened 2019-11-05
Add Blocker.blockOnK

nigredo-tori:
This adds a `ContextShift.evalOnK` alternative for `Blocker`.  Unlike `evalOnK`, we can put this directly on the `Blocker` where it makes the most sense.


rossabaker:
[MiMa is not running](https://travis-ci.org/typelevel/cats-effect/jobs/607567521#L9894-L9895), and I think this is going to break on 2.11.  Before merging, we should:

1. Make sure it's binary compatible.
2. Consider dropping 2.11. The fact that cats did makes a strong case.

After that, looks good.

nigredo-tori:
>  I think this is going to break on 2.11

Could you elaborate? I thought adding a method to a `final` class was safe?

PR #706, opened 2019-11-26
Bracket: Do not evaluate "use" if "acquire" is canceled

Avasil:
If `IO` is canceled during `acquire` in bracket, `use` is still evaluated. 
If `use` is cancelable then it will be canceled so to me it looks like an oversight.

Before the change:

`IO(println("A")) >> IO.cancelBoundary >> IO(println("B"))` will print "A"
`IO(println("A")) >> IO(println("B"))` will print "A" and then "B" unless it is canceled on flatmap.

After the change nothing should be printed at all.

See the gitter question: https://gitter.im/typelevel/cats-effect?at=5ddbe546f3ea522f26667d78

```scala
def acquire: IO[String] = {
  IO(logger.info("Resource acquiring...")) >> timer.sleep(5.seconds) >> IO {
    logger.info("Resource acquired")
    "some resource"
  }
}

def release(resource: String): IO[Unit] = {
  IO(logger.info(s"Releasing $resource"))
}

val action = acquire.bracket { r =>
  IO(logger.info(s"Using $r"))
}(release)

val result = action.timeoutTo(100.millis, IO {
  logger.info("Resource acquiring timed out")
})

IO(logger.info("Started")) >> result.as(ExitCode.Success)

21:25:33.589 Started
21:25:33.608 Resource acquiring...
21:25:38.609 Resource acquired
21:25:38.613 Using some resource
21:25:38.615 Releasing some resource
21:25:38.617 Resource acquiring timed out
```
cc @LMNet (if it doesn't turn out to be intended I'll fix it in Monix as well!)

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -8,3 +8,4 @@ assumeStandardLibraryStripMargin = true
 danglingParentheses = true
 rewrite.rules = [AvoidInfix, SortImports, RedundantBraces, RedundantParens, SortModifiers]
 docstrings = JavaDoc
+lineEndings=preserve
--- a/laws/shared/src/test/scala/cats/effect/IOTests.scala
+++ b/laws/shared/src/test/scala/cats/effect/IOTests.scala
@@ -751,6 +751,26 @@ class IOTests extends BaseTestsSuite {
     effect shouldBe 1
   }
 
+  testAsync("bracket does not evaluate use on cancel") { implicit ec =>
+    implicit val contextShift = ec.contextShift[IO]
+    implicit val timer = ec.timer[IO]
+
+    var use = false
+    var release = false
+
+    val task = IO
+      .sleep(2.second)
+      .bracket(_ => IO { use = true })(_ => IO { release = true })
+      .timeoutTo[Unit](1.second, IO.never)
+
+    val f = task.unsafeToFuture()
+    ec.tick(2.second)
+
+    f.value shouldBe None
+    use shouldBe false
+    release shouldBe true
+  }
+
   test("unsafeRunSync works for IO.cancelBoundary") {
     val io = IO.cancelBoundary *> IO(1)
     io.unsafeRunSync() shouldBe 1
--- a/core/shared/src/main/scala/cats/effect/internals/IOBracket.scala
+++ b/core/shared/src/main/scala/cats/effect/internals/IOBracket.scala
@@ -73,16 +73,21 @@ private[effect] object IOBracket {
     def run(): Unit = result match {
       case Right(a) =>
         val frame = new BracketReleaseFrame[A, B](a, release)
-        val onNext = {
-          val fb = try use(a)
-          catch { case NonFatal(e) => IO.raiseError(e) }
-          fb.flatMap(frame)
-        }
+
         // Registering our cancelable token ensures that in case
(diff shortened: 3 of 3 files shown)


.scalafmt.conf
```diff
@@ -7,4 +7,5 @@ continuationIndent.defnSite = 2
 assumeStandardLibraryStripMargin = true
 danglingParentheses = true
 rewrite.rules = [AvoidInfix, SortImports, RedundantBraces, RedundantParens, SortModifiers]
-docstrings = JavaDoc
\ No newline at end of file
+docstrings = JavaDoc
+lineEndings=preserve
```
djspiewak: Doesn't this cause windows line endings to leak in? Shouldn't we leave this off?

laws/shared/src/test/scala/cats/effect/IOTests.scala
```diff
@@ -751,6 +751,26 @@ class IOTests extends BaseTestsSuite {
...
+      .bracket(_ => IO { use = true })(_ => IO { release = true })
+      .timeoutTo[Unit](1.second, IO.never)
+
+    val f = task.unsafeToFuture()
+    ec.tick(2.second)
+
+    f.value shouldBe None
+    use shouldBe false
+    release shouldBe true
+  }
```
djspiewak: Is this something which should be codified by laws?

.scalafmt.conf
Avasil: I think it's the opposite, I did the PR from Windows and initially scalafmt marked all of the files as changed

djspiewak: I think we want that, though. Or rather, I'd really like it to be somewhat standardized. Git can also do this with a setting I constantly forget about, but basically all of our line endings *should* be Unix-style, otherwise diffs just turn into madness the moment someone on a different platform makes a minor change and saves the file.

laws/shared/src/test/scala/cats/effect/IOTests.scala
Avasil: I'm not sure what's the criteria but I came up with this: (fails before the change)

```scala
  def bracketUseDoesNotEvaluateOnCancel(implicit EQ: Eq[F[(Boolean, Boolean)]]) = {
    var use = false
    var release = false

    val fa: F[(Boolean, Boolean)] = for {
      pa <- Deferred[F, Unit]
      pb <- Deferred[F, Unit]
      pc <- Deferred[F, Unit]
      task = F.bracket(pa.complete(()) >> pb.get)(_ => F.delay { use = true})(_ => F.delay { release = true } >> pc.complete(()))
      fiber <- F.start(task)
      _ <- pa.get
      _ <- F.start(fiber.cancel)
      _ <- pb.complete(())
      _ <- pc.get

    } yield (use, release)

    EQ.eqv(fa, F.pure((false, true)))
  }
```

It is not a property based test so I return `Boolean` and then wrap it in `Prop`:
```scala
"use of bracket is not evaluated when cancel is called during acquire" -> Prop(laws.bracketUseDoesNotEvaluateOnCancel)
```

djspiewak: Meh. We'll revisit this in CE3. :-) It has better mechanisms for expressing this type of thing deterministically.

Issue #710, opened 2019-12-02
Blocker platform no indexed thread names


Slakah:
[`BlockerPlatform`](https://github.com/typelevel/cats-effect/blob/master/core/jvm/src/main/scala/cats/effect/internals/BlockerPlatform.scala) looks to be inconsistent with the [`PoolUtils`](https://github.com/typelevel/cats-effect/blob/master/core/jvm/src/main/scala/cats/effect/internals/PoolUtils.scala#L38) thread naming. In that threads in the same pool are uniquely identifiable for compute, but not for blocker.

i.e.
`ioapp-compute-0/ioapp-compute-1/ioapp-compute-2` vs `cats-effect-blocker/cats-effect-blocker/cats-effect-blocker`.

Some valid use cases for being able to identify a individual thread by name would be that many loggers will output the executing thread, with this metadata being used to determine which thread a piece of code is executing on.

djspiewak:
Strong agree that this should be fixed.

Slakah:
Implementation can be found with PR #722 .

Slakah:
Fixed by #722

PR #722, opened 2019-12-09
#710 Indexed blocker thread names JVM

Slakah:
Implemented #710

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/internals/BlockerPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/internals/BlockerPlatform.scala
@@ -23,14 +23,15 @@ import java.util.concurrent.atomic.AtomicInteger
 import cats.data.NonEmptyList
 
 private[effect] trait BlockerPlatform {
+  private val threadCtr = new AtomicInteger(0)
+
   /**
    * Creates a blocker that is backed by a cached thread pool.
    */
   def apply[F[_]](implicit F: Sync[F]): Resource[F, Blocker] =
     fromExecutorService(F.delay(Executors.newCachedThreadPool(new ThreadFactory {
-      val ctr = new AtomicInteger(0)
       def newThread(r: Runnable) = {
-        val t = new Thread(r, s"cats-effect-blocker-${ctr.getAndIncrement()}")
+        val t = new Thread(r, s"cats-effect-blocker-${threadCtr.getAndIncrement()}")
         t.setDaemon(true)
         t
       }
--- a/core/jvm/src/main/scala/cats/effect/internals/BlockerPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/internals/BlockerPlatform.scala
@@ -19,6 +19,7 @@ package internals
 
 import scala.concurrent.ExecutionContext
 import java.util.concurrent.{ExecutorService, Executors, ThreadFactory}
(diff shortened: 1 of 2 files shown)


core/jvm/src/main/scala/cats/effect/internals/PoolUtils.scala
```diff
@@ -34,8 +34,7 @@ private[internals] object PoolUtils {
       new ThreadFactory {
         val ctr = new AtomicInteger(0)
         def newThread(r: Runnable): Thread = {
-          val back = new Thread(r)
-          back.setName(s"ioapp-compute-${ctr.getAndIncrement()}")
+          val back = new Thread(r, s"ioapp-compute-${ctr.getAndIncrement()}")
```
Slakah: Inlined the thread name argument as per the `BlockerPlatform` implementation.

Can revert this change if requested.

core/jvm/src/main/scala/cats/effect/internals/BlockerPlatform.scala
```diff
@@ -27,8 +28,9 @@ private[effect] trait BlockerPlatform {
    */
   def apply[F[_]](implicit F: Sync[F]): Resource[F, Blocker] =
     fromExecutorService(F.delay(Executors.newCachedThreadPool(new ThreadFactory {
+      val ctr = new AtomicInteger(0)
```
djspiewak: I actually wonder if it's worth putting the counter in a global location. The problem is that, unlike the global pool, people can create as many `Blocker` instances as they like. It's more conventional to only have one, but very much possible to have a ton of them.

core/jvm/src/main/scala/cats/effect/internals/PoolUtils.scala
djspiewak: Seems better this way to me!

core/jvm/src/main/scala/cats/effect/internals/BlockerPlatform.scala
Slakah: Yeah, very good spot. Will make the change now.

core/jvm/src/main/scala/cats/effect/internals/BlockerPlatform.scala
```diff
@@ -19,16 +19,20 @@ package internals
 
 import scala.concurrent.ExecutionContext
 import java.util.concurrent.{ExecutorService, Executors, ThreadFactory}
+import java.util.concurrent.atomic.AtomicInteger
 import cats.data.NonEmptyList
 
 private[effect] trait BlockerPlatform {
+
+  private val threadCtr = new AtomicInteger(0)
```
Slakah: As per https://github.com/typelevel/cats-effect/pull/722#discussion_r355536484, this does somewhat assume that `BlockPlatform` will continue to only be extended by a single object `Blocker`.

The other option would be to have an object to hold the `AtomicInteger` reference, maybe:
```scala
private object BlockerPlatform {
  val threadCtr = new AtomicInteger(0)
}
```
It felt a touch too involved to do that.

rossabaker: Only one subtype is a safe assumption, and if it's wrong, duplicate thread names aren't tragic and can be fixed without binary breakage.

PR #723, opened 2019-12-09
WIP: implement tracing

LukaJCB:


core/shared/src/main/scala/cats/effect/internals/IORunLoop.scala
```diff
@@ -87,23 +91,31 @@ private[effect] object IORunLoop {
             unboxed = thunk().asInstanceOf[AnyRef]
             hasUnboxed = true
             currentIO = null
+            val st = new Throwable().getStackTrace().toList.take(2)
```
djspiewak: Pretty sure this is fundamentally the wrong place for this code. :-( You need to capture the stack location impurely in the `flatMap`/`map`/`apply`/`suspend`/`async` methods, and then that trace element has to be attached to the `Bind`/etc node. Capturing the stack trace *here* means you're just tracing the interpreter.

core/shared/src/main/scala/cats/effect/internals/IORunLoop.scala
```diff
@@ -87,23 +91,31 @@ private[effect] object IORunLoop {
...
+            currentTrace = st ++ currentTrace
+            RaiseError(ex, Trace(currentTrace))
+          }
 
-        case RaiseError(ex) =>
+        case RaiseError(ex, trace) =>
           findErrorHandler(bFirst, bRest) match {
             case null =>
-              cb(Left(ex))
+              cb(Left(new TracedException(trace, ex)))
```
djspiewak: I think we want to hold off on wrapping in `TracedException` until things are actually re-thrown (e.g. in `unsafeRunSync`).

RaasAhsan:
Any progress on this work? I'm willing to pick up from here and continue since I think it will be invaluable.

Avasil:
@RaasAhsan It would be awesome! I'm not sure how far Luka got but one day I expressed interest in attempting this as well and @djspiewak provided me with some background information, maybe you will find it useful:

https://gist.github.com/djspiewak/c0e731eefd0cc8c82bbbfd9197d08474

https://gitter.im/typelevel/cats-effect?at=5d9e65f7e8de6f3ca06c73d7

Unfortunately I was caught up with personal life and didn't do any progress myself

RaasAhsan:
Thanks @Avasil ! I'll get caught up with that material and see if I can put something together.

LukaJCB:
Closing this in favor of #854

PR #727, opened 2019-12-13
Cancel should always wait for the finalizer

### LINKED ISSUE #721: IO's `Fiber#cancel` doesn't wait for finalizers if they already started running
In the following code:

```scala
val finalizer = IO.sleep(5.seconds) *> IO(println("finalized"))

val exhibitA = IO.unit.guarantee(finalizer).timeout(1.second)

val exhibitB = IO.never.guarantee(finalizer).timeout(1.second)
```

Sequencing `exhibitA` will complete after one second, without waiting for the finalizer to complete (it'll keep running in the background).

Sequencing `exhibitB` will wait one second, start the finalizer, wait for it to finish and only then continue.

I think this is a bug, as I was under the impression that `cancel` always waits for the finalizers.

I'd love to help fix this, if it's possible with the current implementation of IO (given some help).

ZIO handles the case properly but Monix Task also behaves like `cats.effect.IO`.

Tested on 2.0.0

Avasil:
Fixes #721

Not 100% sure about the test, I'm counting on final suggestions from @djspiewak and @kubukoz and we could include it in the laws as well

The problem was happening when `release` has been already called by other action

### DIFF
--- a/core/shared/src/main/scala/cats/effect/internals/IOBracket.scala
+++ b/core/shared/src/main/scala/cats/effect/internals/IOBracket.scala
@@ -19,7 +19,8 @@ package cats.effect.internals
 import cats.effect.IO.ContextSwitch
 import cats.effect.{CancelToken, ExitCase, IO}
 import cats.effect.internals.TrampolineEC.immediate
-import scala.concurrent.ExecutionContext
+
+import scala.concurrent.{ExecutionContext, Promise}
 import scala.util.control.NonFatal
 import java.util.concurrent.atomic.AtomicBoolean
 
@@ -127,15 +128,16 @@ private[effect] object IOBracket {
     // Guard used for thread-safety, to ensure the idempotency
     // of the release; otherwise `release` can be called twice
     private[this] val waitsForResult = new AtomicBoolean(true)
+    private[this] val p: Promise[Unit] = Promise()
 
     def release(c: ExitCase[Throwable]): CancelToken[IO]
 
     private def applyRelease(e: ExitCase[Throwable]): IO[Unit] =
       IO.suspend {
         if (waitsForResult.compareAndSet(true, false))
-          release(e)
+          release(e).map(_ => p.success(()))
         else
-          IO.unit
+          IOFromFuture.apply(p.future)
       }
 
     final val cancel: CancelToken[IO] =
--- a/core/shared/src/main/scala/cats/effect/internals/IOBracket.scala
+++ b/core/shared/src/main/scala/cats/effect/internals/IOBracket.scala
@@ -135,7 +135,7 @@ private[effect] object IOBracket {
     private def applyRelease(e: ExitCase[Throwable]): IO[Unit] =
       IO.suspend {
         if (waitsForResult.compareAndSet(true, false))
-          release(e).map(_ => p.success(()))
+          release(e).redeemWith(ex => IO(p.success(())).flatMap(_ => IO.raiseError(ex)), _ => IO(p.success(())))
         else
           IOFromFuture.apply(p.future)
       }
--- a/core/shared/src/main/scala/cats/effect/internals/IOConnection.scala
+++ b/core/shared/src/main/scala/cats/effect/internals/IOConnection.scala
@@ -19,6 +19,7 @@ package internals
 
 import java.util.concurrent.atomic.AtomicReference
 import scala.annotation.tailrec
+import scala.concurrent.Promise
 
 /**
  * INTERNAL API — Represents a composite of functions
@@ -70,16 +71,6 @@ sealed abstract private[effect] class IOConnection {
    * @return the cancelable reference that was removed.
    */
   def pop(): CancelToken[IO]
-
-  /**
-   * Tries to reset an `IOConnection`, from a cancelled state,
-   * back to a pristine state, but only if possible.
-   *
-   * Returns `true` on success, or `false` if there was a race
-   * condition (i.e. the connection wasn't cancelled) or if
-   * the type of the connection cannot be reactivated.
-   */
-  def tryReactivate(): Boolean
 }
 
 private[effect] object IOConnection {
@@ -99,19 +90,21 @@ private[effect] object IOConnection {
     def isCanceled: Boolean = false
     def push(token: CancelToken[IO]): Unit = ()
     def pop(): CancelToken[IO] = IO.unit
-    def tryReactivate(): Boolean = true
     def pushPair(lh: IOConnection, rh: IOConnection): Unit = ()
   }
 
   final private class Impl extends IOConnection {
     private[this] val state = new AtomicReference(List.empty[CancelToken[IO]])
+    private[this] val p: Promise[Unit] = Promise()
 
     val cancel = IO.suspend {
       state.getAndSet(null) match {
-        case null | Nil =>
-          IO.unit
+        case Nil  => IO.unit
+        case null => IOFromFuture(p.future)
         case list =>
-          CancelUtils.cancelAll(list.iterator)
+          CancelUtils
+            .cancelAll(list.iterator)
+            .redeemWith(ex => IO(p.success(())).flatMap(_ => IO.raiseError(ex)), _ => IO(p.success(())))
       }
     }
 
@@ -137,8 +130,5 @@ private[effect] object IOConnection {
           if (!state.compareAndSet(current, xs)) pop()
           else x
       }
-
-    def tryReactivate(): Boolean =
-      state.compareAndSet(null, Nil)
   }
 }
--- a/laws/shared/src/test/scala/cats/effect/IOTests.scala
+++ b/laws/shared/src/test/scala/cats/effect/IOTests.scala
@@ -1084,12 +1084,12 @@ class IOTests extends BaseTestsSuite {
     f.value shouldBe Some(Success(42))
   }
 
-  testAsync("cancel should wait for the release finalizer on success") { implicit ec =>
+  testAsync("cancel should wait for already started finalizers on success") { implicit ec =>
     implicit val contextShift = ec.contextShift[IO]
 
     val fa = for {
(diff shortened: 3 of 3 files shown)


ChristopherDavenport:
What happens with this when you cancel `Async.never`?

Avasil:
@ChristopherDavenport 
> What happens with this when you cancel `Async.never`?

Hmm, thanks to your question I noticed it doesn't wait for `use` finalizers if something calls them earlier. I have a fix for that as well but it's a bit awkward and I don't fully understand it. I will try to handle it over the weekend and search for more corner cases

core/shared/src/main/scala/cats/effect/internals/IOBracket.scala
```diff
@@ -127,15 +128,16 @@ private[effect] object IOBracket {
...
     private[this] val waitsForResult = new AtomicBoolean(true)
+    private[this] val p: Promise[Unit] = Promise()
 
     def release(c: ExitCase[Throwable]): CancelToken[IO]
 
     private def applyRelease(e: ExitCase[Throwable]): IO[Unit] =
       IO.suspend {
         if (waitsForResult.compareAndSet(true, false))
-          release(e)
+          release(e).map(_ => p.success(()))
```
kubukoz: In theory finalizers can fail so I'd also check for that somehow...

Avasil: Good catch, I will handle it and test for that

Avasil:
I think it was a false alarm and the solution might be legit, after few hours I've realized I was expecting to wait for finalizer of completely disconnected fiber :sweat_smile:

Although more test cases ideas are very welcome!

laws/shared/src/test/scala/cats/effect/IOTests.scala
```diff
@@ -1056,6 +1056,69 @@ class IOTests extends BaseTestsSuite {
 
     f.value shouldBe Some(Success(42))
   }
+
+  testAsync("cancel should wait for the release finalizer on success") { implicit ec =>
```
kubukoz: I'd rename it to make it more explicit that the finalizer has to start before the cancelation...

laws/shared/src/test/scala/cats/effect/IOTests.scala
```diff
@@ -1056,6 +1056,69 @@ class IOTests extends BaseTestsSuite {
...
+    implicit val contextShift = ec.contextShift[IO]
+
+    val fa = for {
+      pa <- Deferred[IO, Unit]
+      fiber <- IO.unit.bracket(_ => pa.complete(()))(_ => IO.never).start
```
kubukoz: I'd do `IO.unit.bracket(_ => IO.unit)(_ => pa.complete(()) *> IO.never).start` so that we knwo that the finalizer started before we cancel.

In fact, that's `IO.unit.guarantee(pa.complete(()) *> IO.never).start`.

laws/shared/src/test/scala/cats/effect/IOTests.scala
```diff
@@ -1056,6 +1056,69 @@ class IOTests extends BaseTestsSuite {
...
+    } yield ()
+
+    val f = fa.unsafeToFuture()
+
+    ec.tick()
+
+    f.value shouldBe None
+  }
+
+  testAsync("cancel should wait for the release finalizer on failure") { implicit ec =>
```
kubukoz: I'm confused by this test, the failure is in the finalizer and not in the `use` section. If the finalizer fails, there's no "waiting" anymore, just as in the success case.

kubukoz:
I'd make sure that these are true:

- if the acquire section has started, we must wait for it (and then the finalizer) to complete
- if the use section has started, we must wait for the finalizer to complete
- if the finalizer has started, we must wait for it to complete

For any finalizer (failing, or successful, or never terminating), and any use section.

Also, `uncancelable` should also wait for the finalizer - I'm not sure if this PR fixes this, so it'd be nice to check and possibly write tests.

Avasil:
Thanks for more tests ideas @kubukoz - I discovered an issue when there is a double cancelation and introduced similar change to `IOConnection`. I didn't cover all of the combinations in unit tests but I think the most important ones are there. I can include more if you think it would be valuable.

> Also, uncancelable should also wait for the finalizer - I'm not sure if this PR fixes this, so it'd be nice to check and possibly write tests.

No, it's a separate issue. The current implementation makes it no-op on purpose. I think it's doable to correct it, I have an idea

laws/shared/src/test/scala/cats/effect/IOTests.scala
Avasil: I added it after changing `map` to `redeemWith` according to your suggestion so it's more like a regression test. It would fail with `map` because the promise would never be completed.

laws/shared/src/test/scala/cats/effect/IOTests.scala
Avasil: Good point, bracket is redundant here, the essential code is shared

laws/shared/src/test/scala/cats/effect/IOTests.scala
Avasil: Good idea, I renamed it to "cancel should wait for already started finalizers on success"

core/shared/src/main/scala/cats/effect/internals/IOConnection.scala
```diff
@@ -70,16 +71,6 @@ sealed abstract private[effect] class IOConnection {
...
-
-  /**
-   * Tries to reset an `IOConnection`, from a cancelled state,
-   * back to a pristine state, but only if possible.
-   *
-   * Returns `true` on success, or `false` if there was a race
-   * condition (i.e. the connection wasn't cancelled) or if
-   * the type of the connection cannot be reactivated.
-   */
-  def tryReactivate(): Boolean
```
Avasil: This method is unused and the code is private.
IIRC it was used by `onCancelRaiseError` which was removed at some point.

It introduced a lot of race conditions if someone were to call `tryReactivate` when cancel is in progress and complicated implementation so I just removed it. If we need to bring it back then we can add proper support.

djspiewak:
This all seems quite good to me, but I defer to @kubukoz for final 👍 on merge! I really appreciate your work on this, @Avasil.

To answer your question in the OP, I'd say that this is *definitely* something that needs to be covered by the laws, but unfortunately write now writing laws which cover this in a sane way is, annoying. We can address it in a follow-up PR.

kubukoz:
So... I compared the 2.0.0 behavior to this, with this code:

```scala
object Demo extends IOApp {
  def putStrLn(s: String) = IO(println(s))

    (Deferred[IO, Unit], Deferred[IO, Unit], Ref[IO].of(false))
      .tupled
      .flatMap {
        case (started, canceled, completed) =>
          IO.unit
            .guarantee(
              started.complete(()) *> putStrLn("started.complete(())") *>
                canceled.get *> putStrLn("canceled.get") *>
                completed.set(true) *> putStrLn("completed.set(true)")
            )
            .start
            .flatMap(
              started.get *> putStrLn("started.get") *>
                _.cancel *> putStrLn("_.cancel") *>
                canceled.complete(()) *> putStrLn("canceled.complete(())") *>
                completed.get <* putStrLn("completed.get")
            )
      }
      .map(_.toString)
      .flatMap(putStrLn(_))
      .as(ExitCode.Success)
}
```

2.0.0 outputs:
```
started.complete(())
started.get
_.cancel
canceled.complete(())
canceled.get
completed.set(true)
completed.get
false
```
(and completes)

2.0.0-81bbbbd-SNAPSHOT outputs:
```
started.complete(())
started.get
```

and hangs.

laws/shared/src/test/scala/cats/effect/IOTests.scala
```diff
@@ -1083,6 +1083,141 @@ class IOTests extends BaseTestsSuite {
 
     f.value shouldBe Some(Success(42))
   }
+
+  testAsync("cancel should wait for already started finalizers on success") { implicit ec =>
+    implicit val contextShift = ec.contextShift[IO]
+
+    val fa = for {
+      pa <- Deferred[IO, Unit]
+      fiber <- IO.unit.guarantee(pa.complete(()) >> IO.never).start
```
kubukoz: I feel like this test doesn't actually confirm that `cancel` waits for this particular action (IO.never), but that it doesn't terminate in the case the finalizer doesn't terminate.

How about this:

```scala
testAsync("cancel should wait for already started finalizers on success") { implicit ec =>
  implicit val contextShift = ec.contextShift[IO]
  implicit val timer = ec.timer[IO]

  val fa = for {
    finalizerStarted <- Deferred[IO, Unit]
    finalizerFinish <- Ref.of[IO, Boolean](false)

    finalizer = finalizerStarted.complete(()) >> IO.sleep(5.seconds) *> finalizerFinish.set(true)
    fiber <- IO.unit.guarantee(finalizer).start

    _ <- finalizerStarted.get
    _ <- fiber.cancel
    result <- finalizerFinish.get
  } yield result

  val f = fa.unsafeToFuture()

  ec.tick(5.seconds)

  f.value shouldBe Some(Success(true))
}
```

Avasil: True, at some points I had a deadlock and the test was passing x)

I changed the test to something similar to the failure case - so basically yours but without `Ref`, I feel like it is redundant with `ec.tick()` + `ec.tick(x.second)`

Avasil:
Thank you @kubukoz and @djspiewak for feedback!

laws/shared/src/test/scala/cats/effect/IOTests.scala
kubukoz: Probably, yeah :)

kubukoz:
All good from my side! @djspiewak, can you pull the trigger?

PR #729, opened 2019-12-16
Add Align instances for IO, SyncIO and IO.Par

LukaJCB:


LukaJCB:
Build failure due to ` gem update --system` requiring a yes/no prompt. I have no idea why this could have happened

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -742,6 +743,24 @@ abstract private[effect] class IOParallelNewtype extends internals.IOTimerRef wi
...
 
+  private[effect] def ioParAlign(implicit cs: ContextShift[IO]): Align[IO.Par] = new Align[IO.Par] {
+    import IO.Par.{unwrap, apply => par}
+
+    def align[A, B](fa: IO.Par[A], fb: IO.Par[B]): IO.Par[Ior[A, B]] = alignWith(fa, fb)(identity)
+
+    override def alignWith[A, B, C](fa: IO.Par[A], fb: IO.Par[B])(f: Ior[A, B] => C): IO.Par[C] =
+      par(for {
+        fibA <- unwrap(fa).start
+        fibB <- unwrap(fb).start
```
ChristopherDavenport: Pretty sure these two lines means we can leak fibA

LukaJCB: How so? aren't errors thrown at `join` and not at `start`?

ChristopherDavenport: What if the final IO is cancelled and propagated at this flatMap invocation.

SystemFw: Shouldn't it be possible to just use `mapN` here? like `(fa.attempt, fb.attempt).mapN(twoEithersToIor)`

.travis.yml
```diff
@@ -16,8 +16,8 @@ matrix:
       env: COVERAGE=
 
 install:
-  - rvm use 2.6.0 --install --fuzzy
-  - gem update --system
+  - rvm use 2.6.5 --install --fuzzy
+  - gem update --system --force
```
ChristopherDavenport:
```suggestion
  - yes | gem update --system --force
```

LukaJCB: I hate that I have to deal with this and went into the rabbit hole of figuring out why this is happening, and tried some of the solutions I found over there, but alas, I hope this will just work.

build.sbt
```diff
@@ -26,7 +26,7 @@ ThisBuild / startYear := Some(2017)
 
 val CompileTime = config("CompileTime").hide
 val SimulacrumVersion = "1.0.0"
-val CatsVersion = "2.0.0"
+val CatsVersion = "2.1.0-RC3"
```
gvolpe: Can this be updated to `2.1.0`?

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -742,6 +743,21 @@ abstract private[effect] class IOParallelNewtype extends internals.IOTimerRef wi
    */
   object Par extends IONewtype
 
+  private[effect] def ioParAlign(implicit cs: ContextShift[IO]): Align[IO.Par] = new Align[IO.Par] {
```
djspiewak: Curiosity: why is this being organized in this fashion? I mean, `implicit val`s being declared as such, while the `def`s are declared private and non-implicit and just delegated by the implicit version?

LukaJCB: I assumed it had something to do with binary compability, I took a look at the git blame and didn't think about it too much, but it's probably not necessary here.

PR #311, opened 2018-08-17
Fix #308: Make IO auto-cancelable

alexandru:
Changes:

- makes `IO` auto-cancelable *after* `Async` boundaries
- `IO.shift` subsumes `IO.cancelBoundary`
- removes `onCancelRaiseError`

In order to support these changes:

- fixes `IOApp` for the JVM
- fixes `IO.bracket`
- fixes `Applicative[Fiber[F, ?]]`
- fixes tests

NOTE that I discovered a bug by working on this in `Applicative[Fiber[F, ?]]`, an instance that was broken.

Please ignore:

- `Semaphore`, `Ref`, `MVar` suffered some comments refactoring, added some notes on what "uncancelable" means

PR #730, opened 2019-12-17
Update docs about cancellation to reflect latest implementation

qingwei91:
Hi, I wish to update the docs to reflect the current implementation, but I am not sure how much implementation details to expose?

Another question: I notice the microsite does not work out of the box locally as the microsite project uses a non-default `baseUrl`, is this something we expect?

### DIFF
--- a/site/src/main/mdoc/datatypes/io.md
+++ b/site/src/main/mdoc/datatypes/io.md
@@ -369,15 +369,28 @@ asynchronous boundaries. It can be achieved in the following way:
   those operations is also possible to cancel. It includes, but is not limited to
   waiting on `Mvar.take`, `Mvar.put` and `Deferred.get`.
   
-  We should also note that `flatMap` chains are not auto-cancelable,
-  they are cancellable only if the `flatMap` chain happens *after*
-  an asynchronous boundary mentioned above. After an asynchronous
-  boundary, cancellation checks are performed on every N `flatMap`.
-  Asynchronous boundaries are important for fairness and it's not
-  reasonable to expect interruption in its' absence. With `IO`,
-  fairness needs to be managed explicitly, the protocol being easy
-  to follow and predictable in a WYSIWYG fashion. Try to avoid very
-  long, or never-ending loops without it.
+  We should also note that `flatMap` chains are only cancellable only if
+  the chain happens *after* an asynchronous boundary.
+  After an asynchronous boundary, cancellation checks are performed on every N `flatMap`.
+
+  Here is an example,
+  ```scala mdoc:reset:silent
+  import cats.effect.IO
+
+  def retryUntilRight[A, B](io: IO[Either[A, B]): IO[B] = {
+    io.flatMap {
+      case Right(b) => IO.pure(b)
+      case Left(a) => retryUntilRight(io)
+    }
+  }
+
+  // non-terminating IO that is NOT cancelable
+  val notCancellable: IO[Int] = retryUntilRight[A, B](IO(Left(0)))
+
+  // non-terminating IO that is cancelable because there is an
+  // async boundary created by IO.shift before `flatMap` chain
+  val cancellable: IO[Int] = IO.shift *> retryUntilRight[A, B](IO(Left(0)))
+  ```
   
 2. `IO` tasks that are cancelable, usually become non-terminating on
    `cancel`
--- a/site/src/main/mdoc/datatypes/io.md
+++ b/site/src/main/mdoc/datatypes/io.md
@@ -372,6 +372,7 @@ asynchronous boundaries. It can be achieved in the following way:
   We should also note that `flatMap` chains are only cancelable only if
   the chain happens *after* an asynchronous boundary.
   After an asynchronous boundary, cancellation checks are performed on every N `flatMap`.
+  The value of `N` is hardcoded to 512.
 
   Here is an example,
   ```scala mdoc:reset:silent
--- a/site/src/main/mdoc/datatypes/io.md
+++ b/site/src/main/mdoc/datatypes/io.md
@@ -371,7 +371,7 @@ asynchronous boundaries. It can be achieved in the following way:
   
   We should also note that `flatMap` chains are not auto-cancelable,
   they are cancellable only if the `flatMap` chain happens *after*
-  an asynchronous boundary mentioned above, after an asynchronous
+  an asynchronous boundary mentioned above. After an asynchronous
   boundary, cancellation checks are performed on every N `flatMap`.
   Asynchronous boundaries are important for fairness and it's not
   reasonable to expect interruption in its' absence. With `IO`,
--- a/site/src/main/mdoc/datatypes/io.md
+++ b/site/src/main/mdoc/datatypes/io.md
@@ -369,7 +369,7 @@ asynchronous boundaries. It can be achieved in the following way:
   those operations is also possible to cancel. It includes, but is not limited to
   waiting on `Mvar.take`, `Mvar.put` and `Deferred.get`.
   
-  We should also note that `flatMap` chains are only cancellable only if
+  We should also note that `flatMap` chains are only cancelable only if
   the chain happens *after* an asynchronous boundary.
   After an asynchronous boundary, cancellation checks are performed on every N `flatMap`.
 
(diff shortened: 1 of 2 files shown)


site/src/main/mdoc/datatypes/io.md
```diff
@@ -369,11 +369,15 @@ asynchronous boundaries. It can be achieved in the following way:
...
-  We should also note that `flatMap` chains are not auto-cancelable. Asynchronous
-  boundaries are important for fairness and it's not reasonable to expect interruption
-  in its' absence. With `IO`, fairness needs to be managed explicitly, the protocol being
-  easy to follow and predictable in a WYSIWYG fashion. Try to avoid very long, 
-  or never-ending loops without it.
+  We should also note that `flatMap` chains are not auto-cancelable,
+  they are cancellable only if the `flatMap` chain happens *after*
+  an asynchronous boundary mentioned above. After an asynchronous
+  boundary, cancellation checks are performed on every N `flatMap`.
+  Asynchronous boundaries are important for fairness and it's not
```
qingwei91: I am not sure if this part is still needed, AFAIK, the new implementation still only allow cancellation after async boundary, but it is not neccesary for it to happen right after async boundary

site/src/main/mdoc/datatypes/io.md
```diff
@@ -369,11 +369,15 @@ asynchronous boundaries. It can be achieved in the following way:
   those operations is also possible to cancel. It includes, but is not limited to
   waiting on `Mvar.take`, `Mvar.put` and `Deferred.get`.
   
-  We should also note that `flatMap` chains are not auto-cancelable. Asynchronous
-  boundaries are important for fairness and it's not reasonable to expect interruption
-  in its' absence. With `IO`, fairness needs to be managed explicitly, the protocol being
-  easy to follow and predictable in a WYSIWYG fashion. Try to avoid very long, 
-  or never-ending loops without it.
+  We should also note that `flatMap` chains are not auto-cancelable,
```
Avasil: `auto-cancelable` is not really a known term or standard so I feel like we could skip this part of the sentence and just explain that `IO` checks for cancelation only per N `flatMap` and async boundaries.

site/src/main/mdoc/datatypes/io.md
Avasil: Yeah, it seems to be a bit out of place. I think if we talk about fairness then we should have a small section for it somewhere

site/src/main/mdoc/datatypes/io.md
qingwei91: 👍 

I've remove that sentence: https://github.com/typelevel/cats-effect/pull/730/commits/fbcdbf64849a7fd41d680c5b2ee8d55eede6c210#diff-41df762a62b991556e1f624ca666281dR372

site/src/main/mdoc/datatypes/io.md
qingwei91: I removed the whole paragraph about fairness as suggested, is this ok? or would it be considered as a regression?

I am happy to create a section for it, but I am not familiar with the subject yet, do you know which part of the code should I read to have a better grasp?

If we are okay for this removal, shall I create an issue for adding a dedicated section for it?

kubukoz:
I was under the impression it's possible to cancel a series of flatMaps without any async boundaries at all (I believe that happens in the scastie https://scastie.scala-lang.org/kGCrZ9LcQbeUp8WzlLIvjw). Also, probably an easier example than all this left/right checking would be `foreverM` or a simple recursive `val rec: IO[Unit] = IO.unit >> rec`

qingwei91:
Hi @kubukoz 

If you are referring to the cancellation by `program.timeoutTo`, I believe it is implemented using `start` method which creates an async boundary, which makes it cancellable

kubukoz:
@qingwei91 following that thinking, it's impossible to cancel an IO without an async boundary. So I don't think we need a distinction between flatMap chains that have one and those that don't.

site/src/main/mdoc/datatypes/io.md
```diff
@@ -369,11 +369,28 @@ asynchronous boundaries. It can be achieved in the following way:
...
   waiting on `Mvar.take`, `Mvar.put` and `Deferred.get`.
   
-  We should also note that `flatMap` chains are not auto-cancelable. Asynchronous
-  boundaries are important for fairness and it's not reasonable to expect interruption
-  in its' absence. With `IO`, fairness needs to be managed explicitly, the protocol being
-  easy to follow and predictable in a WYSIWYG fashion. Try to avoid very long, 
-  or never-ending loops without it.
+  We should also note that `flatMap` chains are only cancelable only if
+  the chain happens *after* an asynchronous boundary.
+  After an asynchronous boundary, cancellation checks are performed on every N `flatMap`.
```
kubukoz: Worth mentioning that N is 512 at the time of writing (maxAutoCancelableBatchSize in `IORunLoop`), but could change in the future

qingwei91:
> @qingwei91 following that thinking, it's impossible to cancel an IO without an async boundary. So I don't think we need a distinction between flatMap chains that have one and those that don't.

Hmm, from my understanding, I think there is a subtle difference, in old implementation (which is fairly old now), flatMap chain is not cancellable at all, but cancellation can happen at async boundary, for example

```scala

// Async boundary followed by `flatMap chain without Async Boundary`
IO.shift >>
flatMap(
  flatMap(
    flatMap(
      ... // 1000 more flatMap
    )
  )
)
```

If cancellation happen during the flatMap chain, ie. `after` IO.shift, then it is **NOT** cancellable with old implementation, but it is cancellable in new implementation, because the chain happens AFTER an async boundary. In old implementation, it is only cancellable if the cancellation happen before the async boundary, as that is where the check for cancellation status happen.

I realize I wasn't being clear in my previous reply

site/src/main/mdoc/datatypes/io.md
qingwei91: Yep, can do, I wasn't sure how much implementation details to expose when I wrote it

kubukoz:
Sure, that sounds correct. Old implementation aside, as far as I'm aware now we don't need any async boundaries within our IO at all to have cancelable flatMaps. It doesn't matter whether you put as shift before or after the interruption, the chain can get canceled at pretty much every step that's not within `bracket` or similar.

My point being: because we don't need an async boundary for cancelation to stop our IO at some point, I don't think this is correct:

> We should also note that `flatMap` chains are only cancelable only if the chain happens *after* an asynchronous boundary.

They are just cancelable. After an async boundary the limit is reset (I assume - didn't look for this one really) and you have N more flatMaps till the cancelation is checked, but that sentence implies you can't cancel a flatMap chain before an async boundary. 

And if we think about every IO as having an async boundary "implicitly" inserted at the beginning (if it's not started with `unsafeRunSync()`, it will have one and won't run on the main thread - it also can' be canceled if `unsafeRunSync`ed), then the sentence makes even less sense because you're always after an async boundary.

qingwei91:
> as far as I'm aware now we don't need any async boundaries within our IO at all to have cancelable flatMaps

Hmm, I got the opposite from this PR: https://github.com/typelevel/cats-effect/pull/311, and looking into the code, it seems it is still the case

Following the call path from `unsafeRunTimed` into `step` in `IORunLoop.scala`, we can see that the cancellation check happens here https://github.com/typelevel/cats-effect/blob/master/core/shared/src/main/scala/cats/effect/internals/IORunLoop.scala#L154, inside a private method called `loop`, and this method is only invoked by `step` if there's an `Async` or `ContextSwitch`

Does it make sense? I am not really familiar with the code base, so I might have make some mistakes, I'll try to write some code to show what I meant

site/src/main/mdoc/datatypes/io.md
djspiewak: In this case, I think it's definitely worth exposing the implementation detail.

site/src/main/mdoc/datatypes/io.md
```diff
@@ -369,11 +369,28 @@ asynchronous boundaries. It can be achieved in the following way:
...
+    }
+  }
+
+  // non-terminating IO that is NOT cancelable
+  val notCancelable: IO[Int] = retryUntilRight[A, B](IO(Left(0)))
+
+  // non-terminating IO that is cancelable because there is an
+  // async boundary created by IO.shift before `flatMap` chain
+  val cancelable: IO[Int] = IO.shift *> retryUntilRight[A, B](IO(Left(0)))
+  ```
```
djspiewak: This is a really great pair of examples that nicely illustrates how subtle this can be.

PR #731, opened 2019-12-17
Fix #617: Make Resource covariant in F and A

### LINKED ISSUE #617: Resource should be covariant
It's a concrete datatype. To that end, it should be covariant. It's only abstractions which should be invariant.

neko-kai:
Following the discussion in #617 this simply makes Resource covariant. Binary compatibility is not affected as non-generic parts of the method signatures are unchanged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,8 @@ matrix:
       env: COVERAGE=
 
 install:
-  - rvm use 2.6.0 --install --fuzzy
-  - gem update --system
+  - rvm use 2.6.5 --install --fuzzy
+  - yes | gem update --system --force
   - gem install sass
   - gem install jekyll -v 3.2.1
 
--- a/build.sbt
+++ b/build.sbt
@@ -168,7 +168,15 @@ val mimaSettings = Seq(
       exclude[DirectMissingMethodProblem]("cats.effect.IO.fromFuture"),
       // Incompatible signatures should not cause linking problems.
       exclude[IncompatibleSignatureProblem]("cats.effect.IO.ioParallel"),
-      exclude[IncompatibleSignatureProblem]("cats.effect.IOInstances.ioParallel")
+      exclude[IncompatibleSignatureProblem]("cats.effect.IOInstances.ioParallel"),
+      // Signature changes to make Resource covariant, should not cause linking problems. - https://github.com/typelevel/cats-effect/pull/731
+      exclude[IncompatibleSignatureProblem]("cats.effect.Resource.use"),
+      exclude[IncompatibleSignatureProblem]("cats.effect.Resource.flatMap"),
+      exclude[IncompatibleSignatureProblem]("cats.effect.Resource.map"),
+      exclude[IncompatibleSignatureProblem]("cats.effect.Resource.mapK"),
+      exclude[IncompatibleSignatureProblem]("cats.effect.Resource.allocated"),
+      exclude[IncompatibleSignatureProblem]("cats.effect.Resource.evalMap"),
+      exclude[IncompatibleSignatureProblem]("cats.effect.Resource.evalTap")
     )
   }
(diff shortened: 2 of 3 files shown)


neko-kai:
Strange Travis errors, not related to sbt:
```
bundler's executable "bundle" conflicts with /home/travis/.rvm/rubies/ruby-2.6.0/bin/bundle

Overwrite the executable? [yN]  

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

The build has been terminated
```
Maybe the build needs restarting?

rossabaker:
That's yet another gift from our Ruby dependency.  It's fixed in #729.

neko-kai:
@rossabaker Thanks for pointing that out!

neko-kai:
MiMa failures seem to be false positives, https://github.com/lightbend/mima/issues/363

rossabaker:
Note the Ruby change will conflict with #733, which I think is a better solution, but it will be an easy merge for whoever goes second.

LukaJCB:
Do we know how this will affect type inference? I can see some extra type annotations in the changes that weren't needed before.

neko-kai:
@LukaJCB 
Seems like `F` parameter infers worse in `Resource.pure` when `import cats.implicits._` is present, because many high-priority candidates for Applicative cause 'ambiguous implicits' error:

```
Error:(468, 37) ambiguous implicit values:
 both method catsStdInstancesForEither in trait EitherInstances of type [A]=> cats.MonadError[[β$0$]scala.util.Either[A,β$0$],A] with cats.Traverse[[β$1$]scala.util.Either[A,β$1$]]
 and value catsStdBimonadForFunction0 in trait Function0Instances of type => cats.Bimonad[Function0]
 match expected type cats.Applicative[F]
      case Right(a) => Resource.pure(a)
```

`fs2` avoids that in Stream.apply by returning in essence `Stream[Nothing, A]`, but Resource's `pure` is not free, so we can't do the same.

Personally, I use neither `cats.implicits` nor `Resource.pure`, and I expect that aside of `.pure` the inference should be on par with fs2.

djspiewak:
This is awesome, thank you so much! I want to sort out the false positives on mima so we don't break the master build. I'm not *super* concerned about the source compatibility issue with `pure`; it's unfortunate, but not the end of the world, and maybe fixable later before release.

joroKr21:
Hmm that specific problem (type inference of `pure`) seems to be solved by scala/scala#8582, although for reasons other than implicit scope. Tried quickly in the REPL:
```scala
scala> class Applicative[F[_]]
class Applicative

scala> implicit val apList: Applicative[List] = new Applicative[List]
val apList: Applicative[List] = Applicative@2ad7bd26

scala> implicit val apVec: Applicative[Vector] = new Applicative[Vector]
val apVec: Applicative[Vector] = Applicative@149d7cc6

scala> class Resource[+F[_], +A]
class Resource

scala> def resource[F[_], A](implicit ap: Applicative[F]): Resource[F, A] = new Resource
def resource[F[_], A](implicit ap: Applicative[F]): Resource[F,A]

scala> val x: Resource[List, Int] = resource
val x: Resource[List,Int] = Resource@33f81280
```

I wonder what that means about enhancing lower bounds (but probably not safe in Scala 2 due to LUBs).

neko-kai:
@joroKr21 
That's great, another quality of life boost for variance connoiseurs brought on by your patch, thanks! 🙏

neko-kai:
@djspiewak I've added exclusions for MiMa, the build passes now.

djspiewak:
Sorry for the long delay on this. Thanks for all your hard work, @neko-kai!

Issue #681, opened 2019-10-23
Cats Effect 3.0 Interruption Model Proposal


SystemFw:
I'm writing this proposal to describe the interruption model I'd like to see for cats-effect 3.0.
I'm only going to be focusing on semantics, things like the place in the hierarchy should be discussed elsewhere.
The main goals of this proposal are to lower the skill level needed to write interruption- and finalisation- safe code, and to preserve as much compositionality as possible in a problem that at some level defies it.
I'm going to try setting enough context to justify the decisions in this proposal, if you'd like to skip to the core of it, go to the `Combinators and Semantics` section.

## The problem

The notion of _safety_ in concurrent code is not univocal, and often revolves around two separate concepts: _finalisation_ and _interruption_.
Finalisation is needed to ensure that certain sections of the code will run in any scenario: for example to close resources, or restore some data to a valid state.
Interruption is needed to avoid using resources once the calling code has decided they are no longer needed, to prevent deadlock by offering timeouts, and to compositionally express several patterns (e.g. do this for 10 seconds, then start doing that).
Unfortunately, these two concepts are at odds at a _fundamental level_ :
- Interruption means "No matter what you are doing, stop"
- Finalisation means "No matter what you say, I must keep going"

From a theoretical point of view, there is no perfect compromise: by prioritising resource safety at all costs, you lose all interruption (this is current `uncancelable`), and by prioritising interruption at
all costs, you lose resource safety (this is your standard `flatMap`).
However, _I will posit that for our users, loss of resource safety is more dangerous than loss of interruption_, and therefore we shall keep things as interruptible as possible, but prioritise resource safety when breaking a tie. You will hopefully see exactly what I mean by this, further down in the proposal.

## Current situation

I think it's safe to say that currently writing resource safe code that preserves interruption is a matter for experts, for two reasons:

- The first is that users need to be familiar with `guaranteeCase`, `bracketCase`, `continual`,  `uncancelable`, and the interaction of those with `join` and `cancel`.  Each operation has subtly  different semantics when it comes to  interruption and they are all absolutely necessary in some scenarios.

- The second is more fundamental: writing interruption-aware code means basically writing code that restores some properties when  interruption happens. How and what needs to be restored generally depends on _where_  interruption happened, and _there is no way of knowing that_.
In practice this often means having to do complicated reasoning over  a piece of state and figure out how to write code which can always restore it to a "steady" state, which can get very complex.

## Principles

[...]

PR #738, opened 2020-01-02
Rewrite PureIO on top of coop

### LINKED ISSUE #643: CE3: ExitCase.Completed
The `ExitCase` in the CE 3 proposal is this:

```scala
sealed trait ExitCase[+E, +A]

object ExitCase {
  final case class Completed[+A](a: A) extends ExitCase[Nothing, A]
  final case class Errored[+E](e: E) extends ExitCase[E, Nothing]
  case object Canceled extends ExitCase[Nothing, Nothing]
}
```

And I worry specifically about this case:

```scala
final case class Completed[+A](a: A) extends ExitCase[Nothing, A]
```

The reason for why the current `Completed` does not have `a: A` in it is for it to be able to be used:

1. for data types that don't "terminate" with exactly one result, since there's absolutely no reason to have that restriction — you've got `flatMap` if you want to trigger stuff in case of normal termination, you don't need `bracket` to act as a `flatMap` as well
2. even for data types that can complete with exactly one result (e.g. `EitherT[IO, E, *]`, the `Completed` case can be used to signal something other than successful completion ... if `Error` signals a `Throwable`, then `Completed` does NOT necessarily signal success, it could simply be the result of an `E` error that happened (in the context of `EitherT`) ... which is why we can implement `Sync[EitherT[F, E, *]]` for any `E`

This was already discussed on the PR for `Bracket` and I admit it's very noisy so hard to spot — but @LukaJCB should be aware of the problem as well:

https://github.com/typelevel/cats-effect/pull/113#issuecomment-360386552

The CE 3 proposal says this:

> The first two of the ExitCase cases have corresponding constructors which can build effects that produce to those results:
>
>    Completed ==> pure
>    Errored ==> raiseError

This view is not useful and is where we disagree. I don't see `Completed` as being the counterpart to `pure` or `Errored` as the counterpart to `raiseError`.

I view `Completed` as being `Unit` (aka termination) and `Error | Canceled` as being `Nothing`. No more, no less.

djspiewak:
Mega-change of doom! This PR does a couple of things:

- Adds an implementation of @SystemFw's `mask`/`poll` combinator (called `uncancelable` for now) as described in #681 
  + Random note: I chose to keep `bracketCase` primitive, despite the slight overlap with `uncancelable`. This is mostly because of inversion of control between `Bracket` and `Concurrent`. There is no way (that I'm aware of) to define `bracketCase` in `Bracket` in terms of `onCase` and `guarantee` without making it aware of cancelability. And if we *don't* make it aware of cancelability, then we can define it, but we would need to *override* it in every `Concurrent with Bracket` implementation, which seems very annoying. So instead I made `bracketCase` primitive while `guarantee` and `onCase` are derived from it, and we'll have a law which codifies the relationship between `uncancelable` and the `bracket`.
- Pushes `F[A]` into the `Completed` case of `ExitCase`. This has the interesting side-effect of making `ExitCase` kind of behave like a monad transformer, but only when `F` forms a `Traverse`. When it *doesn't* form a `Traverse`, `ExitCase` is only `Applicative`! Fixes #643
- Rewrite `PureIO` as `PureConc` (names are hard) on top of https://github.com/djspiewak/coop Eventually, if this works, I'll move coop under the Typelevel org

The last point is an important one. `PureConc` is now the following:

```scala
type PureConc[E, A] =
  Kleisli[                    // 1.
    ThreadT[                  // 2.
      Kleisli[                // 3.
        ExitCase[Id, E, ?],   // 4.
        FiberCtx[E],
        ?],
      ?],
    MVar.Universe,
    A]
```

The type seems imposing but it's actually relatively easy to deconstruct. Taking it one stage at a time:

1. The ability to use `MVar` from coop
2. Cooperative multithreading implemented with `FreeT`
3. A lexically-scoped context for the currently running fiber
4. Final results produced by an effect

The nesting is actually somewhat arbitrary: we could have put `ThreadT` on the outside, and even merged the two `Kleisli`s together into one if we wanted to.

Laws are currently not passing, presumably because of bugs. I'm working on that, and in the process working on making `ThreadT` structures a bit more introspectable for debugging purposes, but this is good enough to get some review eyes on it. Also getting it into the upstream branch allows everyone else to pitch in and help out!


djspiewak:
@SystemFw @LukaJCB Any thoughts on this? (singling you out since you self-assigned the reviews 😄; I'd love feedback on this from absolutely anyone who is interested in giving it) I ask because I'm getting close to a follow-up change which would supersede this PR which has working laws and things.

SystemFw:
I'm going to just offer two high level thoughts for lack of more time:
- let's merge the kleislis if possible. I don't know this isn't a big deal semantically but the whole purpose of this model is simplicity

> Random note: I chose to keep bracketCase primitive, despite the slight overlap with uncancelable. 

This is an interesting point, I can't make up my mind but it really irks me that `bracketCase` is primitive, it's a very hefty primitive if you think about it (in terms of the assumptions it makes about cancelation of acquire, for example), and I can tell you for a fact that in the real IO has caused quite a lot of issue when using eq reasoning to simplify some code to analyse buggy behaviour. 
Basically the point is that substituting `onCancel/guarantee` with their definitions in terms of `bracket` makes the code _harder_ to understand.

src/main/scala/ce3/concurrent.scala
```diff
@@ -16,38 +16,42 @@
...
+  def uncancelable[A](body: (F ~> F) => F[A]): F[A]
 
   // produces an effect which is already canceled (and doesn't introduce an async boundary)
-  def canceled[A]: F[A]
+  // this is effectively a way for a fiber to commit suicide and yield back to its parent
+  // The fallback value is produced if the effect is sequenced into a block in which
+  // cancelation is suppressed.
+  //
+  // uncancelable(_ => canceled(a)) <-> pure(a)
+  // (canceled(a) >> never).start.void <-> pure(a).start.flatMap(_.cancel)
```
LukaJCB: I really like this 👍

src/main/scala/ce3/concurrent.scala
```diff
@@ -16,38 +16,42 @@
...
+  def canceled[A](fallback: A): F[A]
```
LukaJCB: Maybe this should be by-name?

djspiewak: `pure` isn't. The idea here is basically that you're inserting a cancelation if not masked. `fallback` is what you get if you *are* masked. So it's exactly the same as the following:

```scala
pure(fallback).start.flatMap(_.cancel)
```

…with the guarantee that the `cancel` hits the bind either immediately before or after the `pure`. Making the argument by-name would mean that `canceled` is more analogous to `delay`, with the tighter bound that the cancel is checked *before* it runs. That seems like unnecessary added complexity.

djspiewak:
@SystemFw 

> let's merge the kleislis if possible. I don't know this isn't a big deal semantically but the whole purpose of this model is simplicity

The main reason I didn't is because I basically need to prism around and interpret them independently. The inner `Kleisli` gets interpreted before *anything* happens, and then has to get re-instantiated a few times later on (in results) due to the recursion in the types. The outer `Kleisli` is interpreted exactly once, in a relatively straightforward fashion. Additionally, there is no mechanism to materialize an `ApplicativeAsk` for some type which has a prism projecting it within some other type.

So it would just create a ton of boilerplate throughout the implementation, and would definitely be somewhat tricky in the interpreter due to the split instantiations. Not at all impossible, but is that *better* than simply having it in two distinct layers?

SystemFw:
I think I'm convinced yeah

djspiewak:
Splitting this out…

> This is an interesting point, I can't make up my mind but it really irks me that bracketCase is primitive, it's a very hefty primitive if you think about it (in terms of the assumptions it makes about cancelation of acquire, for example), and I can tell you for a fact that in the real IO has caused quite a lot of issue when using eq reasoning to simplify some code to analyse buggy behaviour.
>
> Basically the point is that substituting onCancel/guarantee with their definitions in terms of bracket makes the code harder to understand.

I agree that the code becomes harder to understand in this case, but I literally just can't get past the fact that `bracket` cannot be validly defined in terms of other primitives until you get to `Concurrent`. If we want to say that `bracket` is not primitive (which, for any `Concurrent`, it certainly doesn't need to be), then `Bracket` as a typeclass can no longer exist.

As an example, think about trying to write laws for `Bracket`, assuming it's just `guarantee` and `onCancel`, without having a `Concurrent`. You really can't. For starters, `onCancel` doesn't really have meaning at all. You could work with the slightly more general `onCase`, but even still it's awkward. `guarantee` becomes *entirely* lawless. You can only define this stuff in `Concurrent`.

Once you get down into `Concurrent`, of course, many things become relatively easy, but if `Bracket` is going to be a separate typeclass, then it needs to have *meaning* unto itself. The problem I have with eliminating it as a separate thing is that we almost immediately run into issues with `Region`, and it becomes unclear if we can even meaningfully define that class. Losing `Region` would in turn mean that we cannot generalize the hierarchy to the class of dynamic monadic regions, which would be frustrating. Like, I don't think that's a dealbreaker, but it's a pretty serious consequence for something as simple as "`bracket` is primitive".

In `Functor`, `map` is defined to be primitive. Once you get to `Applicative`, though, `map` is redefined in terms of `pure` and `ap` and further restricted with a new set of laws. I don't think `bracket` is really any different in this case.

Note that the code complexity issues shouldn't affect `IO`, since `IO` is perfectly free to… not define `bracket` to be primitive. `bracket` is primitive *in the calculus*, but not necessarily in the implementations. In fact, I'm strongly tempted to revise `bracket` in `PureConc` to be defined solely in terms of `uncancelable` and `onCase`, and then reimplement `onCase` to manipulate the finalizer stack.

SystemFw:
Your considerations on the hierarchy are why I'm torn yeah
I think the problem with `IO` is that actually they are both defined independently. 
I think I could settle with your proposal of changing it in `pureConc`. Like, it still irks me, but the considerations on the hierarchy are strong

djspiewak:
> I think the problem with IO is that actually they are both defined independently.

Agreed. And I think that's a smell unless it's being done for performance reasons. (and even then, I'm tempted to say "don't", since `bracket` is hardly something that happens as frequently as other combinators, like `flatMap`)

> I think I could settle with your proposal of changing it in pureConc. Like, it still irks me, but the considerations on the hierarchy are strong

I'll change it in `PureConc`. I think there's no reason *it* should have `bracket` as its own thing.

src/main/scala/ce3/concurrent.scala
LukaJCB: I understand your sentiment, but I'm not really sure it has to be analogous to `delay` which is an FFI entry point and has totally different semantics. In this case it'd be simply about evaluation, staying as lazy as possible.

djspiewak: Cats very consistently avoids these kinds of things though in favor of `Eval`, except when almost straight-up mandatory (e.g. `ifM`). It also breaks the `pure`-related laws, since `pure` itself is eager. If you can show me a compelling example, I'm open to being convinced (a la `ifM`). :-) Absent a really compelling rationale for laziness here though, I'm strongly leaning towards keeping it eager, at the very least to maintain equivalence with `pure`.

ChristopherDavenport:
Based on the above, am I appropriately understanding that perhaps Bracket being merged with Concurrent may be appropriate?

djspiewak:
> Based on the above, am I appropriately understanding that perhaps Bracket being merged with Concurrent may be appropriate?

If we did that, then `bracket` could cease to be a primitive, but then we would also lose the ability to abstract over dynamic monadic regions. Basically we would put ourselves back into the same boat as CE2, where `Resource` and `Stream` are these weird things that can't be `Sync` or `Concurrent` or anything.

PR #739, opened 2020-01-02
MVar: add "tryRead" fixes #732

### LINKED ISSUE #732: add `tryRead` to `MVar`
Currently `MVar` offers `read`, `put`, `take` as well as `tryPut` and `tryTake`. 

It seems to me, that an atomic non-blocking `tryRead` is missing.

kubum:
Motivation:

I found an open good first issue #732 while reading about `MVar`. This branch contains an attempt to implement it.

Modification:

- add `tryRead: F[Option[A]]` similar to [Control-Concurrent-MVar#tryRead](http://hackage.haskell.org/package/base-4.12.0.0/docs/Control-Concurrent-MVar.html#v:tryReadMVar)
- add a test and a line into the docs

I wonder does it do what is it intended to do? I am happy to amend it if I've missed something, suggestions are welcome. Thank you!

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -29,6 +29,7 @@ val CompileTime = config("CompileTime").hide
 val SimulacrumVersion = "1.0.0"
 val CatsVersion = "2.1.1"
 val DisciplineScalatestVersion = "1.0.1"
+val SilencerVersion = "1.6.0"
 val customScalaJSVersion = Option(System.getenv("SCALAJS_VERSION"))
 
 addCommandAlias("ci", ";scalafmtSbtCheck ;scalafmtCheckAll ;test ;mimaReportBinaryIssues; doc")
@@ -239,6 +240,11 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
       "org.typelevel" %%% "cats-laws" % CatsVersion % Test,
       "org.typelevel" %%% "discipline-scalatest" % DisciplineScalatestVersion % Test
     ),
+    libraryDependencies ++= Seq(
+      compilerPlugin(("com.github.ghik" % "silencer-plugin" % SilencerVersion).cross(CrossVersion.full)),
+      ("com.github.ghik" % "silencer-lib" % SilencerVersion % CompileTime).cross(CrossVersion.full),
+      ("com.github.ghik" % "silencer-lib" % SilencerVersion % Test).cross(CrossVersion.full)
+    ),
     libraryDependencies ++= {
       CrossVersion.partialVersion(scalaVersion.value) match {
         case Some((2, v)) if v <= 12 =>
@@ -329,9 +335,6 @@ lazy val siteSettings = Seq(
       "home",
       Map("permalink" -> "/", "title" -> "Home", "section" -> "home", "position" -> "0")
     )
-  ),
-  micrositeConfigYaml := ConfigYml(
-    yamlCustomProperties = Map("plugins" -> List("jekyll-relative-links"))
   ),
   micrositeCompilingDocsTool := WithMdoc,
   mdocIn := (sourceDirectory in Compile).value / "mdoc",
--- a/core/shared/src/main/scala/cats/effect/concurrent/MVar.scala
+++ b/core/shared/src/main/scala/cats/effect/concurrent/MVar.scala
@@ -17,13 +17,12 @@
 package cats.effect
 package concurrent
 
-import cats.effect.concurrent.MVar.TransformedMVar
+import cats.effect.concurrent.MVar.{TransformedMVar, TransformedMVar2}
 import cats.effect.internals.{MVarAsync, MVarConcurrent}
 import cats.~>
 
 /**
- * A mutable location, that is either empty or contains
- * a value of type `A`.
+ * @define mVarDescription A mutable location, that is either empty or contains a value of type `A`.
  *
  * It has the following fundamental atomic operations:
  *
@@ -37,6 +36,8 @@ import cats.~>
  *  - [[read]] which reads the current value without touching it,
  *    assuming there is one, or otherwise it waits until a value
  *    is made available via `put`
+ *  - `tryRead` returns a variable if it exists. Implemented in the successor [[MVar2]]
+ *  - `swap` takes a value, replaces it and returns the taken value. Implemented in the successor [[MVar2]]
  *  - [[isEmpty]] returns true if currently empty
  *
  * The `MVar` is appropriate for building synchronization
@@ -52,8 +53,12 @@ import cats.~>
  * Inspired by `Control.Concurrent.MVar` from Haskell and
  * by `scalaz.concurrent.MVar`.
  */
-abstract class MVar[F[_], A] {
+sealed private[concurrent] trait MVarDocumentation extends Any {}
 
+/**
+ * $mVarDescription
+ */
+abstract class MVar[F[_], A] extends MVarDocumentation {
   /**
    * Returns `true` if the `MVar` is empty and can receive a `put`, or
    * `false` otherwise.
@@ -120,9 +125,39 @@ abstract class MVar[F[_], A] {
     new TransformedMVar(this, f)
 }
 
+/**
+ * $mVarDescription
+ *
+ * The `MVar2` is the successor of `MVar` with [[tryRead]] and [[swap]]. It was implemented separately only to maintain
+ * binary compatibility with `MVar`.
+ */
+abstract class MVar2[F[_], A] extends MVar[F, A] {
+  /**
+   * Replaces a value in MVar and returns the old value.
+
+   * @param newValue is a new value
+   * @return the value taken
+   */
+  def swap(newValue: A): F[A]
+
+  /**
+   * Returns the value without waiting or modifying.
+   *
+   * This operation is atomic.
+   *
+   * @return an Option holding the current value, None means it was empty
+   */
+  def tryRead: F[Option[A]]
+
+  /**
+   * Modify the context `F` using transformation `f`.
+   */
+  override def mapK[G[_]](f: F ~> G): MVar2[G, A] =
+    new TransformedMVar2(this, f)
+}
+
 /** Builders for [[MVar]]. */
 object MVar {
-
   /**
    * Builds an [[MVar]] value for `F` data types that are [[Concurrent]].
    *
@@ -156,7 +191,7 @@ object MVar {
    * @param F is a [[Concurrent]] constraint, needed in order to
    *        describe cancelable operations
    */
-  def empty[F[_], A](implicit F: Concurrent[F]): F[MVar[F, A]] =
+  def empty[F[_], A](implicit F: Concurrent[F]): F[MVar2[F, A]] =
     F.delay(MVarConcurrent.empty)
 
   /**
@@ -172,7 +207,7 @@ object MVar {
    *
    * @see [[empty]] for creating cancelable MVars
    */
-  def uncancelableEmpty[F[_], A](implicit F: Async[F]): F[MVar[F, A]] =
+  def uncancelableEmpty[F[_], A](implicit F: Async[F]): F[MVar2[F, A]] =
     F.delay(MVarAsync.empty)
 
   /**
@@ -187,7 +222,7 @@ object MVar {
    * @param F is a [[Concurrent]] constraint, needed in order to
    *        describe cancelable operations
    */
-  def of[F[_], A](initial: A)(implicit F: Concurrent[F]): F[MVar[F, A]] =
+  def of[F[_], A](initial: A)(implicit F: Concurrent[F]): F[MVar2[F, A]] =
     F.delay(MVarConcurrent(initial))
 
   /**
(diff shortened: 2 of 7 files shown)


djspiewak:
The implementation looks fantastic here! Thank you! The problem (and the reason the build failed) is binary compatibility. I should have thought of that earlier, but basically it stems from the fact that `MVar` is a `trait` which can be implemented by third-parties. I wonder if there's any way around this?

rossabaker:
Deprecate and rename `MVar`? Problem is, in addition to Cats-Effect, it's already familiar as `MVar` to Haskell and Monix folks.  And I don't have a suggestion for a better name.

kubum:
@djspiewak Thanks for looking into it! I haven't checked the implementation of `MVar` [in Monix](https://github.com/monix/monix/blob/d82e2718a1d1f8bd6aa8277d1ac837c504dc3d36/monix-catnap/shared/src/main/scala/monix/catnap/MVar.scala#L127). Surprisingly, it already has `tryRead` implemented. 

If we want to leave the current `trait` untouched, we can create a new trait. Should the interface closely match `Control-Concurrent-MVar`? If so, we need to add more features such as `swap`.

What do you think?

djspiewak:
We could in theory define a new trait, `MVar2` or something like that, which has `tryRead` and `swap` and other things that the current one is missing and inherit from `MVar`. This would mean that the current `MVar` constructors would continue to be binary compatible but would actually produce `MVar2` (a subtype). Implementations like Monix would also be able to seamlessly swap to `MVar2` without breaking their own compatibility guarantees. In CE3 (when next we break compatibility), we can squish them back together.

I'm not sure if there is another meaningful approach here. The name thing is kind of annoying, and maybe we can come up with something better than `MVar2` as a name, but I'm not sure what else to do other than just punting on adding functionality here. But in truth, I don't see how punting on adding the functionality is any better than adding `MVar2`, since anyone who doesn't need the function will just ignore it and treat it like `MVar` anyway.

rossabaker:
:100: to Daniel's last point about `MVar2` being no worse.  I think we should do it that way and explain ourselves in the MVar doc.

djspiewak:
> I think we should do it that way and explain ourselves in the MVar doc.

Agreed. @kubum would you like to take a crack at restructuring this change to use the `MVar2` suggestion? We should probably also take this opportunity to provide the rest of the missing functionality, such as `swap`.

kubum:
> Agreed. @kubum would you like to take a crack at restructuring this change to use the `MVar2` suggestion? We should probably also take this opportunity to provide the rest of the missing functionality, such as `swap`.

Sounds good to me, thank you! I'll do it and ping you in the discussion for another review if you wouldn't mind.

djspiewak:
> Sounds good to me, thank you! I'll do it and ping you in the discussion for another review if you wouldn't mind.

Sounds perfect!

kubum:
@djspiewak thanks! I updated the PR. Could let me know is it something you had in mind? I couldn't add `@deprecated` to `MVar`, because it is in use when extending `MVar2` and it fails the build.

rossabaker:
The [silencer plugin](https://github.com/ghik/silencer) can help with that deprecation issue.

kubum:
@rossabaker thanks for the link! I wasn't sure if adding a dependency was appropriate. Just pushed a separate commit with the silencer. Have a look, and please let me know what you think. I am happy to keep it or remove it. 

I added silencer under a separate `libraryDependencies` declaration to separate from the important dependencies. However, maybe it is unnecessary.

rossabaker:
Yes, I think silencer is an appropriate addition, and the only way we can get the experience we want without turning off fatal warnings or breaking binary compatibility.

It looks like we also need to silence some references in tests.

kubum:
@rossabaker Well spotted, I just updated the PR. What do you think? Thank you!

core/shared/src/main/scala/cats/effect/concurrent/MVar.scala
```diff
@@ -52,6 +54,7 @@ import cats.~>
  * Inspired by `Control.Concurrent.MVar` from Haskell and
  * by `scalaz.concurrent.MVar`.
  */
+@deprecated("`MVar` is now deprecated in favour of a new generation `MVar2` with `tryRead` support", "2.1.0")
```
rossabaker: The version has progressed since we started talking about this:

```suggestion
@deprecated("`MVar` is now deprecated in favour of a new generation `MVar2` with `tryRead` support", "2.2.0")
```

site/src/main/mdoc/concurrency/mvar.md
```diff
@@ -20,6 +20,7 @@ abstract class MVar[F[_], A] {
 
   def tryPut(a: A): F[Boolean]
   def tryTake: F[Option[A]]
+  def tryRead: F[Option[A]]
```
rossabaker: We should look at changing references in this doc to `MVar2`, now that `MVar` is deprecated.  Lines 5, 6, and 12, at least.  We might also add a brief note about how it's not called `MVar` anymore.

core/shared/src/main/scala/cats/effect/concurrent/MVar.scala
kubum: Awesome! Updated

site/src/main/mdoc/concurrency/mvar.md
kubum: * Line 5 – Points to the file, and it contains all the interface. I guess it has to remain the same, or what do you think?
* Line 6 - Changed
* Line 12 - Changed

I have updated the declaration on the Line 16 too.

kubum:
> Thanks! The code looks good to me. Docs need a little more mopup.
> 
> I milestoned this as 2.2.0 for proper semver, but the version number is negotiable.
> 
> If we still want to add `swap`, we should do it here, or be sure it gets done before we release again, or else we'll have an `MVar3` and be doing this dance again. Are any other methods missing?

Thanks for the review @rossabaker! I added "swap" with tests too. Please have a look and let me know what you think.

build.sbt
```diff
@@ -238,6 +239,10 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
       "org.typelevel" %%% "cats-laws" % CatsVersion % Test,
       "org.typelevel" %%% "discipline-scalatest" % DisciplineScalatestVersion % Test
     ),
+    libraryDependencies ++= Seq(
+      compilerPlugin(("com.github.ghik" % "silencer-plugin" % SilencerVersion).cross(CrossVersion.full)),
+      ("com.github.ghik" % "silencer-lib" % SilencerVersion % Provided).cross(CrossVersion.full)
```
alexandru: I don't like provided libs, because they end up in the generated `pom.xml` anyway. Would be cool to have a setup that filters these out.

I'm using something like this:

```scala
/**
  * Filter out dependencies from the generated `pom.xml`.
  *
  * E.g. to exclude Scoverage:
  * {{{
  *   filterOutDependencyFromGeneratedPomXml("groupId" -> "org\\.scoverage".r)
  * }}}
  *
  * Or to exclude based on both `groupId` and `artifactId`:
  * {{{
  *   filterOutDependencyFromGeneratedPomXml("groupId" -> "io\\.estatico".r, "artifactId" -> "newtype".r)
  * }}}
  */
def filterOutDependencyFromGeneratedPomXml(conditions: (String, Regex)*) = {
  def shouldExclude(e: Elem) =
    e.label == "dependency" && {
      conditions.forall { case (key, regex) =>
        e.child.exists(child => child.label == key && regex.findFirstIn(child.text).isDefined)
      }
    }

  if (conditions.isEmpty) Nil else {
    Seq(
      pomPostProcess := { (node: xml.Node) =>
        new RuleTransformer(new RewriteRule {
          override def transform(node: xml.Node): Seq[xml.Node] = node match {
            case e: Elem if shouldExclude(e) => Nil
            case _ => Seq(node)
          }
        }).transform(node).head
      },
    )
  }
}
```

core/shared/src/main/scala/cats/effect/concurrent/MVar.scala
```diff
@@ -119,6 +123,36 @@ abstract class MVar[F[_], A] {
     new TransformedMVar(this, f)
 }
 
+/**
+ * The `MVar2` is the successor of `MVar` with [[tryRead]] and [[swap]]. It was implemented separately only to maintain
+ * binary compatibility with `MVar`.
```
alexandru: We want to provide good ScalaDocs. I wonder what we can do here, because people are going to be looking at the description of `MVar2`, which has this sentence that has technical implementation details and with no link to `MVar`. And then if they'll take a look at the ScalaDoc for `MVar`, they'll see it as being deprecated.

build.sbt
kubum: Ah, I didn't know "provided" still keeps it in the `pom.xml`. Thank you, this snipper is amazing. I added it to a separate file and called it to exclude both: "scoverage" and "silencer". What do you think?

core/shared/src/main/scala/cats/effect/concurrent/MVar.scala
kubum: Good point. It seems there are two ways:

- Reference both. The downside of it is that tools (IDE helpers) won't give the relevant docs immediately. 
- Copy docs from `MVar` to `MVar2`. The downside is maintaining the two copies.

Are there something else? What do you think?

alexandru: You can define reusable documentation via `@define` in ScalaDoc.

And definitions get inherited, so if you `@define` something on `MVar`, it should then be available in `MVar2`. And a trick would be to do something like ...

```scala
/** 
  * @define mvarDescription An MVar is a mutable location that can be empty or 
  *         contain a value, asynchronously blocking reads when empty and 
  *         blocking writes when full.
  * 
  *         Use-cases:
  *          - bla bla ...
  */
private[concurrent] trait MVarDocs extends Any {}

/** 
  * $mvarDescription
  */
abstract class MVar[F[_], A] extends MVarDocs

/** 
  * $mvarDescription
  */
abstract class MVar2[F[_], A] extends MVar[F, A]
```

It's just a template value, so you can add your warning before or after it.

build.sbt
alexandru: 👍

core/shared/src/main/scala/cats/effect/concurrent/MVar.scala
kubum: Ah, it is a great suggestion, thank you. I just added it. 

I have tried to avoid the redundant `trait`, but scaladocs kept crashing with expansion errors.

build.sbt
djspiewak: I think we already have something a bit like this with the `CompileOnly` scope, or something like that. I can't remember. We used it for simulacrum.

djspiewak:
Ping @rossabaker? I defer to you on all of this. :-)

djspiewak:
Actually I was thinking of this: https://github.com/typelevel/cats-effect/blob/master/build.sbt#L28 So in other words, if you swap `Provided` for `CompileTime`, it should be neatly stripped from the pom.

kubum:
> Actually I was thinking of this: https://github.com/typelevel/cats-effect/blob/master/build.sbt#L28 So in other words, if you swap `Provided` for `CompileTime`, it should be neatly stripped from the pom.

Hi @djspiewak, I wonder if you know the answer, but when I swap `Provided` for `CompileTime` it refuses to run `sbt ci`:

```
sbt:root> ci
[info] Checking 2 Scala sources...
[success] Total time: 3 s, completed 22-Mar-2020 16:39:32
[success] Total time: 0 s, completed 22-Mar-2020 16:39:32
[info] Compiling 18 Scala sources to /Users/andreyfadeyev/Sites/github/cats-effect/core/js/target/scala-2.12/test-classes ...
[info] Compiling 19 Scala sources to /Users/andreyfadeyev/Sites/github/cats-effect/core/jvm/target/scala-2.12/test-classes ...
[error] `silencer-plugin` was enabled but the @silent annotation was not found on classpath - have you added `silencer-lib` as a library dependency?
[error] one error found
[error] `silencer-plugin` was enabled but the @silent annotation was not found on classpath - have you added `silencer-lib` as a library dependency?
[error] one error found
[error] (coreJVM / Test / compileIncremental) Compilation failed
[error] (coreJS / Test / compileIncremental) Compilation failed
```

but `sbt compile` works:

```
sbt:root> compile
[success] Total time: 0 s, completed 22-Mar-2020 16:39:07
```

Do you know a way to make `CompileTime` to work with the silencer?

core/shared/src/main/scala/cats/effect/concurrent/MVar.scala
```diff
@@ -17,42 +17,51 @@
...
  *
- * Given its asynchronous, non-blocking nature, it can be used on
- * top of Javascript as well.
+ *                         Given its asynchronous, non-blocking nature, it can be used on
+ *                         top of Javascript as well.
  *
- * Inspired by `Control.Concurrent.MVar` from Haskell and
- * by `scalaz.concurrent.MVar`.
+ *                         Inspired by `Control.Concurrent.MVar` from Haskell and
+ *                         by `scalaz.concurrent.MVar`.
```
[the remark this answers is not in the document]
kubum: Ah, I thought it needs the indentation, but I was wrong. Just tested it that it renders it well without it. I am going to change it.

kubum:
Just updated the doc and rebased. I couldn't make `CompileTime` to work with the silencer, left the rest as it is. Please let me know what you think should be improved.

rossabaker:
I had the same problem in http4s last night trying to use `CompileTime` on silencer.

site/src/main/mdoc/concurrency/mvar.md
rossabaker: Sorry, I thought I'd commented already.  Yes, that looks great.

rossabaker:
When it's in `CompileTime`, it's not accessible to the tests.  I made it work by also adding it to `Test` scope:

```scala
    libraryDependencies ++= Seq(
      compilerPlugin(("com.github.ghik" % "silencer-plugin" % SilencerVersion).cross(CrossVersion.full)),
      ("com.github.ghik" % "silencer-lib" % SilencerVersion % CompileTime).cross(CrossVersion.full),
      ("com.github.ghik" % "silencer-lib" % SilencerVersion % Test).cross(CrossVersion.full)
    ),
```

For whatever it's worth, `SilencerVersion` can now be `1.6.0`.

@kubum has done great work on this, and I'd like to see us get it merged ASAP.  I'll add a follow-up ticket to get the other relevant methods from `MVar` in before 2.2.0 so we don't get stuck with `MVar3`.

kubum:
> When it's in `CompileTime`, it's not accessible to the tests. I made it work by also adding it to `Test` scope:
> 
> ```scala
>     libraryDependencies ++= Seq(
>       compilerPlugin(("com.github.ghik" % "silencer-plugin" % SilencerVersion).cross(CrossVersion.full)),
>       ("com.github.ghik" % "silencer-lib" % SilencerVersion % CompileTime).cross(CrossVersion.full),
>       ("com.github.ghik" % "silencer-lib" % SilencerVersion % Test).cross(CrossVersion.full)
>     ),
> ```
> 
> For whatever it's worth, `SilencerVersion` can now be `1.6.0`.

@rossabaker Your idea worked fantastic! I just removed the last commit with the pom filter, bumped silencer to 1.6.0 and followed the suggested way to include both in test and compile time. Thank you!

djspiewak:
Terribly sorry for missing this! I didn't notice the approval. 🎉

djspiewak:
Thank you so much for following this through, @kubum!

PR #742, opened 2020-01-08
Parallel resource

SystemFw:
First prototype for parallel resource acquisition.
Please only look at the implementation, I intend to expose this as a `Parallel` instance rather than the `parZip` function. 
Needs proper testing as well

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Resource.scala
+++ b/core/shared/src/main/scala/cats/effect/Resource.scala
@@ -449,6 +449,34 @@ object Resource extends ResourceInstances with ResourcePlatform {
    * resource value.
    */
   final case class Suspend[F[_], A](resource: F[Resource[F, A]]) extends Resource[F, A]
+
+  /**
+    * Newtype encoding for a `Resource` datatype that has a `cats.Applicative`
+    * capable of doing parallel processing in `ap` and `map2`, needed
+    * for implementing `cats.Parallel`.
+    *
+    * Helpers are provided for converting back and forth in `Par.apply`
+    * for wrapping any `IO` value and `Par.unwrap` for unwrapping.
+    *
+    * The encoding is based on the "newtypes" project by
+    * Alexander Konovalov, chosen because it's devoid of boxing issues and
+    * a good choice until opaque types will land in Scala.
+    * [[https://github.com/alexknvl/newtypes alexknvl/newtypes]].
+    *
+    */
+  type Par[+F[_], +A] = Par.Type[F, A]
+
+  object Par {
+    type Base
+    trait Tag extends Any
+    type Type[+F[_], +A] <: Base with Tag
+
+    def apply[F[_], A](fa: Resource[F, A]): Type[F, A] =
+      fa.asInstanceOf[Type[F, A]]
+
+    def unwrap[F[_], A](fa: Type[F, A]): Resource[F, A] =
+      fa.asInstanceOf[Resource[F, A]]
+  }
 }
 
 abstract private[effect] class ResourceInstances extends ResourceInstances0 {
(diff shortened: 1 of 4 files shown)


SystemFw:
Mmm, I wrote this on a version without covariance, need to do some changes

core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -133,6 +133,99 @@ sealed abstract class Resource[+F[_], +A] {
...
+    * B: close 1
+    * res5: (Int, Int) = (3,3)
+    *
+    **/
+
+  def parZip[G[x] >: F[x]: Sync: Parallel, B](
+      other: Resource[G, B]
+    ): Resource[G, (A, B)] = {
+
+    def store = cats.effect.concurrent.Ref[G].of(Sync[G].unit -> Sync[G].unit)
```
kubukoz: `val`?

core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -133,6 +133,99 @@ sealed abstract class Resource[+F[_], +A] {
...
+    type Store = (G[Unit] => G[Unit]) => G[Unit]
+
+    Resource.make(store)(_.get.flatMap(_.parTupled.void)).evalMap { store =>
+      val leftStore: Store = f => store.update(_.leftMap(f))
+      val rightStore: Store = f => store.update(_.map(f))
+
+      (this.stored(leftStore), other.stored(rightStore)).parTupled
```
kubukoz: I think a comment saying explicitly that *this* is where we actually allocate might help (especially since allocation comes up in the code later than cleanup, which might cause confusion)

kubukoz:
I'd say we could leave `parZip` as a direct method as well. It won't hurt compatibility and massively help discoverability (in fact, I believe we should have a couple parallel methods under IO too)

core/shared/src/main/scala/cats/effect/Resource.scala
SystemFw: Yeah I can add a small comment detailing the main idea behind the implementation as well, if it's deemed useful

laws/shared/src/main/scala/cats/effect/laws/discipline/arbitrary.scala
```diff
@@ -121,6 +121,11 @@ object arbitrary {
                                                            AFU: Arbitrary[F[Unit]]): Arbitrary[Resource[F, A]] =
     Arbitrary(Gen.delay(genResource[F, A]))
 
+    implicit def catsEffectLawsArbitraryForResourceParallel[F[_], A](implicit F: Applicative[F],
```
kubukoz: Maybe this could take the `Arbitrary[Resource[F,A]]` instance instead of Applicative and friends?

core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -133,6 +130,64 @@ sealed abstract class Resource[+F[_], +A] {
...
+   * reverse order of acquisition.
+   *
+   * Note that `Resource` also comes with a `cats.Parallel` instance
+   * that offers more convenient access to the same functionality as
+   * `parZip`, for example via `parMapN`:
+   *
+   * {{{
+   *   def mkResource(name: String) = {
+   *     val acquire =
+   *       IO(scala.util.Random.nextInt(1000).millis) *>
```
kubukoz:
```suggestion
   *       IO(scala.util.Random.nextInt(1000).millis).flatMap(IO.sleep) *>
```

laws/shared/src/main/scala/cats/effect/laws/util/TestInstances.scala
```diff
@@ -99,6 +99,18 @@ trait TestInstances {
...
+  /**
+   * Defines equality for `Resource.Par`.  Two resources are deemed
+   * equivalent if they allocate an equivalent resource.  Cleanup,
+   * which is run purely for effect, is not considered.
+   */
+  implicit def eqResourcePar[F[_], A](implicit E: Eq[F[A]], F: Bracket[F, Throwable]): Eq[Resource.Par[F, A]] =
+    new Eq[Resource.Par[F, A]] {
+      import Resource.Par.unwrap
+      def eqv(x: Resource.Par[F, A], y: Resource.Par[F, A]): Boolean =
+        eqResource[F, A].eqv(unwrap(x), unwrap(y))
```
kubukoz: Could probably do the same trick as with Arbitrary using `Eq[Resource[F, A]]`

laws/shared/src/test/scala/cats/effect/ResourceTests.scala
```diff
@@ -276,4 +293,188 @@ class ResourceTests extends BaseTestsSuite {
...
+      IO { leftReleasing = true } >> wait >> IO { leftReleased = true }
+    }
+    val rhs = Resource.make(wait >> IO { rightAllocated = true }) { _ =>
+      IO { rightReleasing = true } >> wait >> IO { rightReleased = true }
+    }
+
+    (lhs, rhs).parTupled.use(_ => wait).unsafeToFuture()
+
+    // after 1 second:
+    //  both resources have allocated (concurrency, serially it would happen after 2 seconds)
```
kubukoz:
```suggestion
    //  both resources have allocated (concurrently, serially it would happen after 2 seconds)
```

SystemFw:
I didn't manage to address the final few comments in time (like the scaladoc one). I'll open a follow-up PR and merge on CI pass since it's going to be trivial

PR #743, opened 2020-01-09
Add Odin to related projects table

sergeykolbasov:
https://github.com/valskalla/odin/

>Odin library enables functional approach to logging in Scala applications with reasoning and performance as the top priorities.
>
>- Each effect is suspended within the polymorphic F[_]
>- Context is a first-class citizen. Logger is structured by default, no more TheadLocal MDCs
>- Programmatically configurable. Scala is the perfect language for describing configs
>- Position tracing implemented with macro instead of reflection considerably boosts the performance
>- Own performant logger backends for console and log files
>- Composable loggers to bring different loggers together with Monoid[Logger[F]]
>- Backend for SLF4J API
>
> Standing on the shoulders of cats-effect type classes, Odin abstracts away from concrete effect types,  allowing users to decide what they feel comfortable with: IO, ZIO, monix.Task, ReaderT etc.

I think the main difference from [log4cats](https://github.com/ChristopherDavenport/log4cats/) is the own backend without any slf4j stuff, a structured logger by default and macro-derived logging position as a part of the interface instead of implementation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,7 @@ These are some of the projects that provide high-level functions on top of `cats
 | [Log4cats](https://christopherdavenport.github.io/log4cats/) | Functional Logging |
 | [Cats STM](https://timwspence.github.io/cats-stm/) | Software Transactional Memory for Cats Effect |
 | [Mau](https://github.com/kailuowang/mau) | A tiny library for an auto polling `Ref` |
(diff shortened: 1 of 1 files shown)


SystemFw:
Thanks, always good to see people build on top of cats-effect :)

PR #753, opened 2020-01-21
Support ScalaJS 1.0.x

cquiroz:
This is an attempt to support ScalaJS 1.0.0-RC2 with cats effect. Besides the obvious change on the plugin I updated `IOTimer`.

Compiling it with 1.0.0-RC2 throws a compilation error and I think it may be better to use `scala.scalajs.js.timers` instead. IMHO the code is cleaner this way

An obvious downside is that build times double as I run ci with both scala.js versions

I'm not sure how publication is done. perhaps it needs to be done manually for 1.0.0-RC2


.travis.yml
```diff
@@ -25,6 +25,7 @@ install:
 script:
   - export SBT_PROFILE=$COVERAGE
   - sbt ++$TRAVIS_SCALA_VERSION $COVERAGE ci
+  - SCALAJS_VERSION=1.0.0-RC2 sbt ++$TRAVIS_SCALA_VERSION $COVERAGE ci
```
djspiewak: I think we should do this using a Travis build matrix, so that the two jobs are split, similar to what we do for Scala versions.

cquiroz:
The script may try to publish the JVM side twice, would that work given we do sonatypeReleaseAll only at the end?

cquiroz:
A related question about travis. it seems coverage is never used, Should it be removed?

djspiewak:
> The script may try to publish the JVM side twice, would that work given we do sonatypeReleaseAll only at the end?

Ah yes, that's a good point. Hmm. Is there a way we can *only* publish the JS artifacts within the for loop? And only publish the JVM at the end, as you said?

> A related question about travis. it seems coverage is never used, Should it be removed?

I admit I haven't actually paid attention to it, but isn't it checked in PR builds?

cquiroz:
There maybe a way to only publish once suggested by `sbt-ci-release`.
I sent a separate PR about coverage

djspiewak:
sbt-ci-release would have the same problem as any local publication script, and would have the *added* problem of key management, which is why I don't like using it for OSS builds.

cquiroz:
I rebased and incorporated the comments. Now it will not publish the jvm side when doing scala.js 1.0 (That may need to change once 0.6 is dropped)
I added a `release.sh` script and it seems fine when I try locally

.travis.yml
```diff
@@ -21,6 +33,7 @@ install:
 
 script:
   - sbt ++$TRAVIS_SCALA_VERSION ci
+  - SCALAJS_VERSION=1.0.0-RC2 sbt ++$TRAVIS_SCALA_VERSION ci
```
djspiewak: I think this can be removed now that you have the matrix

cquiroz:
I changed it as requested and also rebased onto master

djspiewak:
Thank you so much for doing all the work on this! I really appreciate it!

cquiroz:
Great, thanks for your help. Do you think it could be released on 1.0.0-RC2? Or maybe wait for the next release?

djspiewak:
@cquiroz I'm working on getting a 2.1.0 release out today or tomorrow

build.sbt
```diff
@@ -252,6 +253,9 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
   .jvmConfigure(_.settings(mimaSettings))
   .jsConfigure(_.enablePlugins(AutomateHeaderPlugin))
   .jsConfigure(_.settings(scalaJSSettings))
+  .jvmSettings(
+    skip.in(publish) := customScalaJSVersion.forall(_.startsWith("1.0"))
+  )
```
SethTisue: having trouble understanding the logic here. did you mean `exists` instead of `forall`? (why disable JVM publishing when `SCALAJS_VERSION` isn't set at all?)

PR #757, opened 2020-01-22
Some small changes for Dotty compatibility

travisbrown:
This is a set of changes that prepare for Dotty cross-building but are reasonable to do now:

* I've replaced kind-projector's `?` with `*`, [which will be supported in Dotty soon](https://github.com/lampepfl/dotty/pull/7775).
* Dotty is pickier about parentheses (which also helps readability and consistency in my view).
* Dotty doesn't support eta-expansion syntax for converting by-name parameters to `Function0`.
* Dotty warns on `final` on case objects, and it doesn't do anything on Scala 2.
* Dotty doesn't have `do-while`, and recommends the rewrite I've done here.
* In a few places we rely on Scala 2 value discarding, which I've changed to either `.void` or `; ()`.
* One test extends `BeforeAndAfterAll` but doesn't use any of its methods, and it was causing problems with my local Dotty build of ScalaTest, so I just removed it here.

After these changes (and #756) a few other things are necessary to get tests passing on Dotty (see [my branch here](https://github.com/travisbrown/cats-effect/commits/topic/dotty-experiment) for a complete list of the additional changes that are needed).

* Right now you have to publish Dotty locally from [my kind-projector branch](https://github.com/lampepfl/dotty/pull/7775).
* You also still need to publish ScalaTest locally if you want to run the tests.
* You have to use my [Simulacrum Scalafix rules](https://github.com/travisbrown/simulacrum-scalafix) to expand `@typeclass`.
* You have to use another of those Scalafix rules to expand kind-projector's polymorphic lambda value syntax, which isn't supported in my Dotty branch.
* There's a [bug with private constructors in Dotty](https://github.com/lampepfl/dotty/issues/8064) that you have to work around.
* There's a possible bug involving [type aliases](https://github.com/lampepfl/dotty/issues/8067) that also needs working around.
* There are a few places where Dotty can't infer types that Scala 2, or where you need to cast to `Any` (see [my `topic/dotty-experiment` branch](https://github.com/travisbrown/cats-effect/commits/topic/dotty-experiment) for details).

The good news is that if you make all of those changes and publish Dotty and ScalaTest locally, all tests pass on Dotty.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Bracket.scala
+++ b/core/shared/src/main/scala/cats/effect/Bracket.scala
@@ -197,7 +197,7 @@ object ExitCase {
   /**
    * An [[ExitCase]] signaling completion in failure.
    */
-  case class Error[+E](e: E) extends ExitCase[E]
+  final case class Error[+E](e: E) extends ExitCase[E]
 
   /**
    * An [[ExitCase]] signaling that the action was aborted.
--- a/core/shared/src/main/scala/cats/effect/Bracket.scala
+++ b/core/shared/src/main/scala/cats/effect/Bracket.scala
@@ -192,12 +192,12 @@ object ExitCase {
    * outcome for the user, but it does for the purposes of the
    * `bracket` operation.
    */
-  final case object Completed extends ExitCase[Nothing]
+  case object Completed extends ExitCase[Nothing]
 
   /**
    * An [[ExitCase]] signaling completion in failure.
    */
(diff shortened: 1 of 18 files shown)


neko-kai:
> Dotty warns on `final` on case classes and objects

Do you mean just on case objects? There's no reason to warn on final case classes, if there's a warning, it's an upstream bug.

travisbrown:
@neko-kai Oh, right, I went past those warnings too quickly. I've added `final` back for `Error`.

PR #762, opened 2020-01-28
Dotty cross-building [WIP]

travisbrown:
This PR is waiting on the same things as [the Cats one](https://github.com/typelevel/cats/pull/3269), and a couple of others:

- [ ] A Dotty-friendly release of [sbt-tpolecat](https://github.com/DavidGregory084/sbt-tpolecat).
- [x] A Dotty release without [this bug](https://github.com/lampepfl/dotty/issues/8064) (which is fixed in master, but there haven't been nightlies for a few days).
- [x] Migrate laws.
- [ ] There's an issue with ScalaTest's `BeforeAndAfter` on Dotty that's currently causing some tests to fail.


sloshy:
FYI looks to me like `sbt-tpolecat` is now Dotty-friendly as of #797

travisbrown:
@sloshy Nice! I'll push an update of this branch tomorrow.

PR #764, opened 2020-01-29
Dotty-friendly changes for the laws module

travisbrown:
This is a follow-up to #757 for the laws module. The biggest part of the change is due to the tests relying on the fact that Scala 2 won't consider an implicit value in implicit search in its own definition if it doesn't have a type annotation. So for example this appears dozens of times:

```scala
implicit val cs = ec.contextShift[IO]
```

The `contextShift` method takes an `Async[IO]`, and there are a couple of implicit methods on `IO` that can provide this: `ioConcurrentEffect`, the higher-priority instance that's available when there's a `ContextShift` instance in scope for `IO`, and `ioEffect`, a lower-priority instances with no constraints.

When `cs` has no type annotation, `ioEffect` is chosen, since in effect the compiler doesn't yet know it's defining an implicit `ContextShift[IO]` during the implicit search for an `Async[IO]`. If you add the type annotation, it tries to use itself in that search, and you end up with a forward reference compiler error.

Dotty requires return types on implicits, which means we can't rely on this weird Scala 2 behaviour if we want to cross-build. I've worked around this by providing a concrete `ioContextShift` version of `contextShift` in the `TestContext` helper in laws.

There are some smaller changes in the other commits that are necessary for Dotty and also improve the code on Scala 2 in my view (a few other type annotations, consistent parentheses, explicit value discarding).

I can't run a few of the test classes on Dotty because of an issue with ScalaTest's `BeforeAndAfter`, but all tests are building and most are passing on Dotty in this [WIP PR](https://github.com/typelevel/cats-effect/pull/762) that includes the commits here.

### DIFF
--- a/laws/js/src/test/scala/cats/effect/internals/IOContextShiftTests.scala
+++ b/laws/js/src/test/scala/cats/effect/internals/IOContextShiftTests.scala
@@ -16,7 +16,7 @@
 
 package cats.effect.internals
 
-import cats.effect.{BaseTestsSuite, IO}
+import cats.effect.{BaseTestsSuite, ContextShift, IO}
 import scala.util.Success
 
 class IOContextShiftTests extends BaseTestsSuite {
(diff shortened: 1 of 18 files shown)


djspiewak:
IMO, all `implicit`s aside from lambda parameters should have explicit types *anyway*, so this is a good change even absent Dotty's change in semantics.

PR #769, opened 2020-01-31
[Website docs] Fix home page path

calvellido:
Sets `permalink` property for root index page.

The new default use of `mdoc` in `sbt-microsites` led to some slight changes when using extra `.md` files. This change lets the site to serve again the root page correctly.


rossabaker:
Thanks.  It looks right.  I'm going to try to republish with this to fix the site.

PR #780, opened 2020-02-07
Fix bug when double cancelation would hang second cancel

### LINKED ISSUE #779: Fiber cancellation is no longer idempotent
In cats-effect 2.0.0, calling cancel multiple times on a fiber worked. In 2.1.0, any call to `cancel` besides the first results in a hang.

```scala
object Main extends App {
  implicit val csIO: ContextShift[IO] = IO.contextShift(scala.concurrent.ExecutionContext.Implicits.global)
  implicit val timerIO: Timer[IO] = IO.timer(scala.concurrent.ExecutionContext.Implicits.global)
  val t = IO(println("pre-sleep")) *> IO.sleep(2.seconds) *> IO(println("post-sleep"))
  t.start.flatMap { f =>
    IO(println("joining")) *>
      f.join *>
      IO(println("joined")) *>
      f.cancel *>
      IO(println("cancelled")) *> 
      f.cancel *> // This hangs in cats-effect 2.1.0 but works in 2.0.0
      IO(println("cancelled again"))
  }.unsafeRunSync
}
```

Avasil:
Fixes #779

### DIFF
--- a/core/shared/src/main/scala/cats/effect/internals/IOConnection.scala
+++ b/core/shared/src/main/scala/cats/effect/internals/IOConnection.scala
@@ -99,7 +99,7 @@ private[effect] object IOConnection {
 
     val cancel = IO.suspend {
       state.getAndSet(null) match {
-        case Nil  => IO.unit
+        case Nil  => IO { p.success(()); () }
         case null => IOFromFuture(p.future)
         case list =>
           CancelUtils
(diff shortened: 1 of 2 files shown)


mpilquist:
Awesome!

rossabaker:
The MiMa errors are related to #770, which just needs a second approval.

djspiewak:
Amazing work!

PR #785, opened 2020-02-12
Using blocker in Resorce.fromAutoClosable

BalmungSan:
Yesterday @mpilquist suggested in the gitter channel that most classes that implement `AutoClosable` are blocking in its construction and `close`. As such, those operations should be called using a **Blocker**. Thus, this makes `Resource.fromAutoClosable` conceptually _"broken"_.
This PR attempts to solve that problem.

I have a couple of questions:

+ Do you all agree that this should be the expected behavior or not?
+ ~I believe not all `AutoClosable` are blocking, should we ignore that minority just for sake of consistency? or maybe we should leave both alternatives?~
+ ~Assuming we decide to get rid of the old method, what should we do with it? It has to be preserved for binary compatibility, but it shouldn't be exposed for use anymore. I think we could deprecate it. However, @kubukoz suggests that it could be made `private` and that will preserve binary compatibility.~
+ ~The current implementation has the possibility of calling `blockOn` over acquire two times. If someone constructs it using the blocker itself. I do not know if that has any consequence or if we can just ignore it. In any case, I was thinking, why not use a _by-name_ parameter for it instead?~
+ ~I am not 100% happy with the test. First, I do not like the idea of using the same `ec` for the blocker, but the other option that will work for JS & JVM is global, but being a test I guess I am just overthinking. The other thing that bothers me, is if there is a way to check if `close` was executed inside the blocker.~
+ ~Maybe we should use `java.nio.file.Files.newBufferedReader` and `Path` in the documentation?~

### DIFF
--- a/laws/shared/src/test/scala/cats/effect/ResourceTests.scala
+++ b/laws/shared/src/test/scala/cats/effect/ResourceTests.scala
@@ -92,18 +92,24 @@ class ResourceTests extends BaseTestsSuite {
     }
   }
 
-  test("resource from AutoCloseable is auto closed") {
+  testAsync("resource from AutoCloseable is auto closed") { implicit ec =>
+    implicit val ctx: ContextShift[IO] = ec.ioContextShift
+
+    val blocker = Blocker.liftExecutionContext(ec)
+
     var closed = false
     val autoCloseable = new AutoCloseable {
       override def close(): Unit = closed = true
     }
 
     val result = Resource
-      .fromAutoCloseable(IO(autoCloseable))
+      .fromAutoCloseable(blocker)(IO(autoCloseable))
       .use(_ => IO.pure("Hello world"))
-      .unsafeRunSync()
+      .unsafeToFuture()
 
-    result shouldBe "Hello world"
+    ec.tick()
+
+    result.value shouldBe Some(Success("Hello world"))
     closed shouldBe true
   }
 
--- a/laws/shared/src/test/scala/cats/effect/ResourceTests.scala
+++ b/laws/shared/src/test/scala/cats/effect/ResourceTests.scala
@@ -20,6 +20,7 @@ package effect
 import cats.data.Kleisli
 import cats.effect.concurrent.Deferred
 import cats.effect.laws.discipline.arbitrary._
+import cats.effect.laws.util.TestContext
 import cats.implicits._
 import cats.kernel.laws.discipline.MonoidTests
 import cats.laws._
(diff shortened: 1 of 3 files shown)


kubukoz:
As a general comment, I think you'll very rarely have resources that implement Closeable / Autocloseable and don't block, but since we already have a method for that we could just let it be. I'd consider renaming the new one with a postfix "Blocking".

alexandru:
I don't agree that `AutoCloseable` implies blocking I/O in `acquire` or `release`, even if this is often used for stuff coming out of `java.io`. 

But having an alternative that uses a `Blocker` might be OK.

laws/shared/src/test/scala/cats/effect/ResourceTests.scala
```diff
@@ -92,18 +92,24 @@ class ResourceTests extends BaseTestsSuite {
     }
   }
 
-  test("resource from AutoCloseable is auto closed") {
+  testAsync("resource from AutoCloseable is auto closed") { implicit ec =>
+    implicit val ctx: ContextShift[IO] = ec.ioContextShift
+
+    val blocker = Blocker.liftExecutionContext(ec)
```
alexandru: You're not testing if the execution actually happens on your `Blocker`.

I suggest initializing your `Blocker` from a different `TestContext`, an `ec2` that you initialize yourself. And then you check that you need an `ec2.tick()` in order for `acquire` and `release` to execute.

BalmungSan: Yeah, that was one of my questions, I wasn't sure how can I test that, but really I wanted to.

Let me check if I understand your suggestion. Thanks!

laws/shared/src/test/scala/cats/effect/ResourceTests.scala
```diff
@@ -107,6 +108,45 @@ class ResourceTests extends BaseTestsSuite {
...
+    // Check that acquire ran inside the blocking context.
+    ec.tick()
+    acquired shouldBe false
+    blockingEc.tickOne()
+    acquired shouldBe true
+
+    // Check that close was called and ran inside the blocking context.
+    ec.tick()
+    closed shouldBe false
+    blockingEc.tickOne()
```
alexandru: This test is a little fragile due to relying on `tickOne`. If you want to improve it further, you can avoid usage of `tickOne` by introducing a break in `use`. And you can do so for example by triggering an async boundary in it.

```scala
val result = Resource
  .fromAutoCloseableBlocking(blocker)(acquire)
  .use(_ => ctx.shift *> IO("Hello world"))
  .unsafeToFuture()
```

Thus executing `use` itself will require an `ec.tick()` and then you can afford to convert `blockingEc.tickOne()` into a regular `blockingEc.tick()`, which should make the test more future proof so to speak.

But in this form it's OK-ish too.

BalmungSan:
@alexandru Now that you mentioned the docs. What do you think about using `java.nio.file.Files.newBufferedReader` and `java.nio.file.Path` instead of `java.io.File`?

-----

Also, I still have this question:

The current implementation has the possibility of calling blockOn over acquire two times. If someone constructs it using the blocker itself. I do not know if that has any consequence or if we can just ignore it.

alexandru:
@BalmungSan 

> @alexandru Now that you mentioned the docs. What do you think about using java.nio.file.Files.newBufferedReader and java.nio.file.Path instead of java.io.File?

I think `java.io` is still far more common than `java.nio`. Leave it, unless you have good reasons.

> The current implementation has the possibility of calling blockOn over acquire two times. If someone constructs it using the blocker itself.

Indeed. Don't worry about it — if someone is using the blocker directly, then they know what they are doing and should probably use the simple version.

laws/shared/src/test/scala/cats/effect/ResourceTests.scala
BalmungSan: Actually the `shift` wasn't necessary. Because I suppose because `blockOn` should do it.

BalmungSan:
Ok, I think this PR is ready.
Unless someone has another suggestion.

djspiewak:
Worth noting that I do very commonly see people using `Path` rather than `File` these days, particularly since it's so easy to convert between them. Definitely agree that `java.io` is still way more common in general though.

PR #786, opened 2020-02-13
Create `IOResourceApp` as an alternative to `IOApp`

agustafson:
Create an `IOResourceApp` as an alternative to `IOApp`, implemented by providing the method:
```scala
def run(args: List[String]): Resource[IO, ExitCode]
```

The `IOApp` seems to shut down by doing the equivalent of a `kill -9` by running `Fiber.cancel`.

Given that I (and probably quite a few other people) use `Resource`s to create and run their services, it seemed to make more sense to use the `Resource.allocated` method to start a server and add a system shutdown hook using the `F[Unit]` shutdown function returned from `allocated`. This allows us to return a `Resource[IO, ExitCode]` from the `main` method rather than a `IO[ExitCode]`, and also negate the need to run `.use(_ => IO.never)` to convert a `Resource[IO, _]` to `IO[ExitCode]` in `IOApp`.

I've done a little testing with a local app and it seems to behave in a sane way, in that it shuts down `Resource`s when the app is shutdown.

I realise the js code might be very shonky/completely unusable, but I've done as best as I can to replicate (read "blatantly copy-and-paste") from the `IOAppPlatform` implementations, and the tests pass so 🤷‍♂

More than happy to hear any feedback. If things look vaguely sane it would make sense to move the duplicated code I've copied from `IOApp*`

### DIFF
--- a/core/js/src/main/scala/cats/effect/internals/IOResourceAppPlatform.scala
+++ b/core/js/src/main/scala/cats/effect/internals/IOResourceAppPlatform.scala
@@ -1,0 +1,116 @@
+/*
+ * Copyright (c) 2017-2019 The Typelevel Cats-effect Project Developers
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats
+package effect
+package internals
+
+import cats.implicits._
+import scala.concurrent.duration._
+import scala.scalajs.js
+
+private[effect] object IOResourceAppPlatform {
+  def main(args: Array[String], cs: Eval[ContextShift[IO]], timer: Eval[Timer[IO]])(
+    run: List[String] => Resource[IO, ExitCode]
+  ): Unit =
+    mainProcess(args, cs, timer)(run).unsafeRunAsync {
+      case Left(t) =>
+        Logger.reportFailure(t)
+        setExitCode(ExitCode.Error.code)
+      case Right(code) =>
+        setExitCode(code.code)
+    }
+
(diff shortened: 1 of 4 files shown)


agustafson:
I've looked at the js build errors but can't see how the code is any different from `IOApp` ... 🤔
Any ideas/advice welcome.

alexandru:
Hi @agustafson,

Thanks for your PR — before reading my comment further, please keep in mind that we need to have stability in mind — any new interface or function that we add is basically a promise that the interface or function being added will be supported for a long time. This makes adding new features to foundational libraries, like Cats-Effect, difficult to say the least.

---

> The IOApp seems to shut down by doing the equivalent of a kill -9 by running Fiber.cancel.

Minor nitpick, but `kill -9` is a `SIGKILL` that force kills the process, so no cancellation is executed. You're probably thinking of `SIGTERM` (i.e. `kill` without a signal name or `kill -15`).

---

> This allows us to return a `Resource[IO, ExitCode]` from the main method rather than a `IO[ExitCode]`, and also negate the need to run `.use(_ => IO.never)` to convert a `Resource[IO, _]` to `IO[ExitCode]` in IOApp.

There are several problems with this:

1. behavior is surprising because it is non-terminating; non-termination should be an explicit user choice, not an implicit one
2. the result of that method is in fact not `IO[ExitCode]`, but rather `IO[Nothing]` and hence the result of your own `main`, which is `Resource[IO, ExitCode]` is in fact `Resource[IO, _]`, because the `ExitCode` is never used so it's irrelevant

Also `IO.pure` is a perfectly valid option here, instead of `IO.never`. Even if it means that the computation described by `Resource` cannot be cancelled. Another reason for why the use of `IO.never` is confusing, because it's not the only or even the most obvious choice one can make.

---

So assuming you did not mean `IO.pure`, what's the difference between your code and this implementation?

```scala
trait IOResourceApp extends IOApp {

  def runResource(args: List[String]): Resource[IO, _]

  override final def run(args: List[String]): IO[Nothing] =
    runResource(args).use(_ => IO.never)
}
```

There shouldn't be any differences and from your description I don't understand why the need for the custom internals.

And if there are no differences, adding an entirely new interface for describing a one liner is problematic.

agustafson:
Hi @alexandru - thanks for replying so quickly and apologies for the delay in my response.

All of what you've said makes good sense, and thanks for taking the time to explain it.

The main drivers for me were that it felt more natural to provide a thing that could be used and then shutdown rather than something that was run to termination, but I can see why that doesn't make sense. I had hoped that it might avoid the problems that occur when shutting down a http4s server, when there are a lot of error logs as the server shuts down. However, looking at the code in `IOApp` and `IO` the server _should_ shut down more cleanly anyway.

I'll close the ticket now.

PR #788, opened 2020-02-15
Add nix-shell

rossabaker:
Motivation: I'm sick and tired of incompatible jekyll projects across my system and not being able to test pull requests locally.  Cats has similar.

Solution: add a nix shell that provides all the dependencies we need to build the project:
* sbt (obviously)
* git (for sbt-git)
* node (for the Scala.JS build)
* jekyll (for the microsite)

Use: `nix-shell` from this directory drops a user into a fully-configured development environment with a predictable version of Java, Node, and Jekyll.


aeons:
Could the gemset.nix file live in the nix folder for less clutter?

alexandru:
I know you like Nix, however for the rest of us it is just another tool we have to learn.  

Jekyll is a Ruby tool, the standard solution in the Ruby community for managing dependencies is [Bundler](https://bundler.io/). The actual problem is that [sbt-microsites](https://github.com/47deg/sbt-microsites), via [sbt-site](https://github.com/sbt/sbt-site/), is not using best practices when invoking Jekyll.

Therefore instead of fixing this by bringing a chainsaw to the party, somebody should fix this line here:

https://github.com/sbt/sbt-site/blob/v1.4.0/src/main/scala/com/typesafe/sbt/site/jekyll/JekyllPlugin.scala

That line should invoke `bundle && bundle exec jekyll` instead of just `jekyll`.

rossabaker:
It does look terrible, but the status quo is terrible.  What this offers is a clone-and-build solution for the tiny, but growing, subset of users who have adopted it, and can be ignored by those who haven't.  (Does the Cats one bother anybody?)  I work on this project on three laptops, and we have to set up tools for three languages on each to test this project.  Annoying already, but the real problem is that the environment is not even consistent across related projects: some require Java 11 and some require Java 8 and I have one that needs to be tested on a specific patch release of Java 11.  With this tooling, I don't need to think about that, and as more projects adopt it, the network effect grows and less people have to think about that.

I didn't figure this would be popular or accepted, and I'll probably do what I continue to do on every other project: maintain configuration on a cloud drive and curse when it doesn't build and copy it ad hoc to my checkout and locally git ignore it until it breaks again.

rossabaker:
Okay, it's now 95% lighter and self-contained.  I hope this makes it palatable to those who don't care for it, because is still a big win for those who do.

* Bundix is removed in favor of a jeykll wrapper similar to @alexandru's proposal for sbt-site.  We lose a little of nix's caching that way, but that wasn't ever the goal, and it's back to one source of truth.  A plausible alternative would be a shell hook to run bundle once.

* niv is dropped in favor of an inline call to fetchTarball.  We lose a nice command line interface for upgrading the pin, but that doesn't happen, and now it's much less scary.

PR #794, opened 2020-02-22
Add getAndUpdate, updateAndGet to Ref

kubukoz:
Leaving `setAndGet(a: A): F[A]` out because it's useless (the parameter you pass is the return value).

### DIFF
--- a/core/shared/src/main/scala/cats/effect/concurrent/Ref.scala
+++ b/core/shared/src/main/scala/cats/effect/concurrent/Ref.scala
@@ -59,9 +59,27 @@ abstract class Ref[F[_], A] {
   def set(a: A): F[Unit]
 
   /**
+   * Updates the current value using `f` and returns the value that was updated.
+   *
+   * In case of retries caused by concurrent modifications,
+   * the returned value will be the last one before a successful update.
+   */
+  def getAndUpdate(f: A => A): F[A] = modify { a =>
+    (f(a), a)
+  }
+
+  /**
    * Replaces the current value with `a`, returning the previous value.
    */
-  def getAndSet(a: A): F[A]
+  def getAndSet(a: A): F[A] = getAndUpdate(_ => a)
+
+  /**
+   * Updates the current value using `f`, and returns the updated value.
+   */
+  def updateAndGet(f: A => A): F[A] = modify { a =>
+    val newA = f(a)
+    (newA, newA)
+  }
 
   /**
    * Obtains a snapshot of the current value, and a setter for updating it.
@@ -220,7 +238,9 @@ object Ref {
 
     def set(a: A): F[Unit] = F.delay(ar.set(a))
 
-    def getAndSet(a: A): F[A] = F.delay(ar.getAndSet(a))
+    override def getAndSet(a: A): F[A] = F.delay(ar.getAndSet(a))
+    override def updateAndGet(f: A => A): F[A] = F.delay(ar.updateAndGet(f(_)))
+    override def getAndUpdate(f: A => A): F[A] = F.delay(ar.getAndUpdate(f(_)))
 
     def access: F[(A, A => F[Boolean])] = F.delay {
       val snapshot = ar.get
@@ -239,8 +259,7 @@ object Ref {
       else None
     }
 
-    def update(f: A => A): F[Unit] =
-      modify(a => (f(a), ()))
+    def update(f: A => A): F[Unit] = updateAndGet(f).void
 
     def modify[B](f: A => (A, B)): F[B] = {
       @tailrec
--- a/core/shared/src/main/scala/cats/effect/concurrent/Ref.scala
+++ b/core/shared/src/main/scala/cats/effect/concurrent/Ref.scala
@@ -239,8 +239,6 @@ object Ref {
     def set(a: A): F[Unit] = F.delay(ar.set(a))
 
     override def getAndSet(a: A): F[A] = F.delay(ar.getAndSet(a))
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/effect/concurrent/Ref.scala
```diff
@@ -239,8 +259,7 @@ object Ref {
       else None
     }
 
-    def update(f: A => A): F[Unit] =
-      modify(a => (f(a), ()))
+    def update(f: A => A): F[Unit] = updateAndGet(f).void
```
kubukoz: This one I'm not so sure about, but I suppose using AtomicRef's `updateAndGet` might be more efficient than spinning manually. I'll do some benchmarking.

kubukoz: It seems to perform roughly the same. Did a quick benchmark (`sbt benchmarksNext/jmh:run -o ../results/#{date}/#{name}-Next.txt  -i 5 -wi 1 -f 2 -t 1 cats.effect.benchmarks.RefBenchmark`) and results are similar, except the native implementation has larger errors.

```scala

import org.openjdk.jmh.annotations.BenchmarkMode
import org.openjdk.jmh.annotations.State
import org.openjdk.jmh.annotations.OutputTimeUnit
import org.openjdk.jmh.annotations.Benchmark
import org.openjdk.jmh.annotations.Scope
import org.openjdk.jmh.annotations.Mode
import java.util.concurrent.TimeUnit
import java.util.concurrent.atomic.AtomicReference
import scala.annotation.tailrec
import cats.effect.Sync
import cats.effect.IO
import org.openjdk.jmh.annotations.Param
import cats.implicits._

@State(Scope.Thread)
@BenchmarkMode(Array(Mode.Throughput))
@OutputTimeUnit(TimeUnit.SECONDS)
class RefBenchmark {
  implicit val cs = IO.contextShift(ExecutionContext.global)

  @Param(Array("10000"))
  var size: Int = _

  class Ref[F[_], A](ar: AtomicReference[A])(implicit F: Sync[F]) {
    def modify[B](f: A => (A, B)): F[B] = {
      @tailrec
      def spin: B = {
        val c = ar.get
        val (u, b) = f(c)
        if (!ar.compareAndSet(c, u)) spin
        else b
      }
      F.delay(spin)
    }
  }

  @Benchmark
  def withModifySequential() = {
    val ar = new AtomicReference(0)

    val ref = new Ref[IO, Int](ar)

    ref.modify(a => (a + 1, ())).replicateA(size).unsafeRunSync()
  }

  @Benchmark
  def nativeSequential() = {
    val ar = new AtomicReference(0)

    IO(ar.updateAndGet(_ + 1)).replicateA(size).unsafeRunSync()
  }

  @Benchmark
  def withModifyParallel() = {
    val ar = new AtomicReference(0)

    val ref = new Ref[IO, Int](ar)

    List.fill(size)(ref.modify(a => (a + 1, ()))).parSequence_.unsafeRunSync()
  }

  @Benchmark
  def nativeParallel() = {
    val ar = new AtomicReference(0)

    List.fill(size)(IO(ar.updateAndGet(_ + 1))).parSequence_.unsafeRunSync()
  }
}
```

```
Benchmark                          (size)   Mode  Cnt     Score    Error  Units
RefBenchmark.withModifySequential   10000  thrpt   10  1396.445 ± 14.894  ops/s
RefBenchmark.nativeSequential       10000  thrpt   10  1377.636 ± 77.142  ops/s
RefBenchmark.withModifyParallel     10000  thrpt   10    54.985 ±  2.888  ops/s
RefBenchmark.nativeParallel         10000  thrpt   10    54.521 ±  5.284  ops/s
```

kubukoz:
Apparently there's no `updateAndGet` nor `getAndUpdate` in ScalaJS's AtomicReference. I don't want to have a platform-specific Ref, so I'll probably update the code not to use these at all.

SystemFw:
I'm a bit undecided on this one, the names are almost as long as the implementation, so I wonder if they can be left out altogether. Won't stand in the way if someone wants to merge though

kubukoz:
Sure, the implementation looks short here, but it gets tedious when you have something bigger in your update:

```scala
ref.modify { a =>
  val newState = {
    //logic
    //logic
    //logic
    //logic
  }
  (newState, newState)
}
```

vs

```scala
ref.updateAndGet { a =>
    //logic
    //logic
    //logic
    //logic
}
```

Let's see what others think :)

vlovgr:
Just as a data point, I've written `updateAndGet` a few times, e.g. in fs2-kafka [here](https://github.com/fd4s/fs2-kafka/blob/ec9021f5d283681e6a06f9d9d47cc92da7bc8a68/modules/core/src/main/scala/fs2/kafka/internal/syntax.scala#L36-L44).

rossabaker:
I think they're comfortable to people coming from `AtomicRef`, which makes me like them more than most convenience methods.

core/shared/src/main/scala/cats/effect/concurrent/Ref.scala
djspiewak: Doesn't surprise me that it performs about the same. C2 will trivially inline one into the other, even with default settings.

djspiewak:
Even though the implementations are quite trivial, I like the more semantic names. It makes client code more readable, and relies less on "pattern matching" to see certain things. When using various monads, we almost never write `fa.flatMap(x => f(x).pure)`, we write `fa.map(f)`.

kubukoz:
Exactly - it's also why `flatTap` and friends are useful :) thanks for the merge

PR #799, opened 2020-02-24
Fix `continual` hanging when canceled really fast

### LINKED ISSUE #793: `continual` hangs if canceled before `use` in its internal`bracket`
```scala
import cats.effect._
import cats.implicits._
import cats.effect.concurrent._
import java.{util => ju}

object Demo extends IOApp {

  def run(args: List[String]): IO[ExitCode] = {
    (IO(ju.UUID.randomUUID()), Deferred[IO, Unit]).tupled.flatMap {
      case (id, latch) =>
        val action: IO[Unit] = latch.complete(()) *> IO.never

        Concurrent.continual(action)(_ => IO.unit).start.flatMap(latch.get *> _.cancel) *>
          IO(println(s"foo $id"))
    }
  }.foreverM.as(ExitCode.Success)
}
```

This will run for a while, but eventually it'll print its final `foo` and never progress further.

I suppose this might be caused by a combination of `continual` being based on `start` and nested `start`s not getting cancelation propagated to them. But the most concerning part of it all is that it takes a couple hundred / couple thousand repeats to actually hang the program.

kubukoz:
Fixes #793

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Concurrent.scala
+++ b/core/shared/src/main/scala/cats/effect/Concurrent.scala
@@ -621,7 +621,7 @@ object Concurrent {
             // We complete the promise because it's uncancelable - there needs to be *something* in it after this `bracketCase`.
             r.complete(Left(new Exception("Continual fiber cancelled") with NoStackTrace))
         case _ => F.unit
-      }.attempt *> r.get.rethrow.guaranteeCase(e => F.delay(println(e)))
+      }.attempt *> r.get.rethrow
     }
   }
 
--- a/core/shared/src/main/scala/cats/effect/Concurrent.scala
+++ b/core/shared/src/main/scala/cats/effect/Concurrent.scala
@@ -605,7 +605,6 @@ object Concurrent {
    */
   def continual[F[_], A, B](fa: F[A])(f: Either[Throwable, A] => F[B])(implicit F: Concurrent[F]): F[B] = {
     import cats.effect.implicits._
-    import scala.util.control.NoStackTrace
 
     Deferred.uncancelable[F, Either[Throwable, B]].flatMap { r =>
       fa.start.bracketCase { fiber =>
@@ -617,9 +616,7 @@ object Concurrent {
         case (fiber, ExitCase.Canceled) =>
           // This cancel has to be here, and not in the `guaranteeCase` above, as the `use` of `bracketCase` might not be evaluated.
           // See https://github.com/typelevel/cats-effect/issues/793
-          fiber.cancel *>
-            // We complete the promise because it's uncancelable - there needs to be *something* in it after this `bracketCase`.
-            r.complete(Left(new Exception("Continual fiber cancelled") with NoStackTrace))
+          fiber.cancel
         case _ => F.unit
       }.attempt *> r.get.rethrow
     }
--- a/core/shared/src/main/scala/cats/effect/Concurrent.scala
+++ b/core/shared/src/main/scala/cats/effect/Concurrent.scala
@@ -608,17 +608,20 @@ object Concurrent {
     import scala.util.control.NoStackTrace
 
     Deferred.uncancelable[F, Either[Throwable, B]].flatMap { r =>
-      fa.start
-        .bracket { fiber =>
-          fiber.join.guaranteeCase {
-            case ExitCase.Completed | ExitCase.Error(_) =>
-              fiber.join.attempt.flatMap(f).attempt.flatMap(r.complete)
-            case _ =>
-              fiber.cancel >>
-                r.complete(Left(new Exception("Continual fiber cancelled") with NoStackTrace))
-          }.attempt
-        }(_ => r.get.void)
-        .flatMap(_ => r.get.rethrow)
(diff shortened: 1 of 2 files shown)


laws/shared/src/test/scala/cats/effect/ConcurrentContinualTests.scala
```diff
@@ -20,8 +20,26 @@ import cats.effect.concurrent.Ref
...
+
+    val test = {
+      Deferred[IO, Unit]
+        .flatMap { started =>
+          (started.complete(()) *> IO.never: IO[Unit])
+            .continual(_ => IO.unit)
+            .start
+            .flatMap(started.get *> _.cancel)
+        }
+    }.replicateA(10000).as(true)
```
kubukoz: I haven't found a reliable way to reproduce the original issue without putting in arbitrary numbers. It can be anywhere from 1 to 7000... but the test runs very quickly and reliably hanged when the fix wasn't there.

kubukoz:
The build failed due to `unsafeRunSync` being used on ScalaJS, but I wasn't able to reproduce the original issue on `testContext`. I'll split out the previously-hanging test into a new AsyncFunSpec.

kubukoz:
Okay, get this - I moved from a 4-core machine to a 6-core one and the test hangs at ONE iteration 😂 😂

core/shared/src/main/scala/cats/effect/Concurrent.scala
```diff
@@ -608,17 +608,20 @@ object Concurrent {
...
+          case ExitCase.Completed | ExitCase.Error(_) => fiber.join.attempt.flatMap(f).attempt.flatMap(r.complete)
+          case _                                      => F.unit //will be canceled in release of enclosing bracketCase
+        }
+      } {
+        case (fiber, ExitCase.Canceled) =>
+          // This cancel has to be here, and not in the `guaranteeCase` above, as the `use` of `bracketCase` might not be evaluated.
+          // See https://github.com/typelevel/cats-effect/issues/793
+          fiber.cancel *>
+            // We complete the promise because it's uncancelable - there needs to be *something* in it after this `bracketCase`.
+            r.complete(Left(new Exception("Continual fiber cancelled") with NoStackTrace))
```
SystemFw: if this `bracketCase` is canceled, doesn't that mean that we never get to run `r.get` anyway?

kubukoz: oh, it probably does, yeah. I'll test just to make sure.

kubukoz: indeed, we never reach `*>` if it's canceled. Good catch!

SystemFw:
I like this implementation better anyway (in addition to the bug fix :) ). Let me ponder on it just a bit more then I'll merge

SystemFw:
I'm personally cool with merging now, should we wait for a second review?

kubukoz:
I think it's okay to merge but would appreciate someone else looking at this with a fresh pair of eyes before releasing :)

core/shared/src/main/scala/cats/effect/Concurrent.scala
```diff
@@ -605,20 +605,20 @@ object Concurrent {
...
+          case ExitCase.Completed | ExitCase.Error(_) => fiber.join.attempt.flatMap(f).attempt.flatMap(r.complete)
+          case _                                      => F.unit //will be canceled in release of enclosing bracketCase
+        }
+      } {
+        case (fiber, ExitCase.Canceled) =>
+          // This cancel has to be here, and not in the `guaranteeCase` above, as the `use` of `bracketCase` might not be evaluated.
+          // See https://github.com/typelevel/cats-effect/issues/793
+          fiber.cancel
+        case _ => F.unit
+      }.attempt *> r.get.rethrow
```
Avasil: I feel like we could remove this `attempt` if we can make the following assumption for generic code:

AFAIK `F#start` can only fail with something like `RejectedExecutionException` before creating the fiber for cats IO and Monix Task. It returns `UIO` for ZIO and Monix BIO so `attempt` won't do anything there

kubukoz: Removing this `attempt` causes this test to fail:

```scala

  testAsync("Concurrent.continual behaves like handleErrorWith on errors") { implicit ec =>
    implicit val cs: ContextShift[IO] = ec.ioContextShift

    val ex = new Exception("boom")
    val task = IO.raiseError[Unit](ex).continual(n => IO(n.swap.toOption.get))

    val f = task.unsafeToFuture()
    ec.tick()
    f.value shouldBe Some(Success(ex))
  }

```

I think it's because if `fiber` fails, `fiber.join.guaranteeCase` will also fail and we won't get the result from the Deferred (which would be a `Left[Throwable]`, so a valid `Either[Throwable, B]`), but a failed effect.

Avasil: Ah, right, I've forgotten about `fa` 😅

kubukoz:
Okay, let's merge :)

PR #804, opened 2020-03-01
Add a bunch of useful functions directly on IO

kubukoz:
One of the factors in attracting new users is discoverability. I think some of the methods that really show the power of the IO data type are really hard to discover for newcomers, for example:

- run these two actions in parallel
- schedule this IO for later

This PR adds a couple of methods:
- right and left shark (seriously, it's really good to make them very visible!)
- `&>`, `<&`, `parProduct`
- `delayBy`
- `as` (with a lazy parameter, which is an improvement over the functor syntax anyway)
- `io.option` (catching throwables to options), `IO.none[A]: IO[Option[A]]`, `IO.fromOption`

I also noticed the methods we already have aren't really sorted, so as another PR I'd like to order them alphabetically (there were no symbolic methods so I put the operators at the end).

I'm aware this will likely spark a discussion around how rich we want the API to be. If you consider some of these redundant, or have ideas for other methods we could add directly to IO - I'm all ears!

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -1362,6 +1362,9 @@ object IO extends IOInstances {
       case Left(err) => raiseError(err)
     }
 
+  /**
+   * Lifts an `Option[A]` into the `IO[A]` context, raising the throwable if the option is empty.
+   */
   def fromOption[A](option: Option[A])(orElse: => Throwable): IO[A] = option match {
     case None        => raiseError(orElse)
     case Some(value) => pure(value)
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -663,7 +663,7 @@ sealed abstract class IO[+A] extends internals.IOBinaryCompat[A] {
   /**
    * Replaces failures in this IO with [[scala.None]].
    */
-  def option: IO[Option[A]] = map(Some(_)).handleErrorWith(_ => IO.none)
+  def option: IO[Option[A]] = redeem(_ => None, Some(_))
 
   /**
    * Handle any error, potentially recovering from it, by mapping it to another
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -679,8 +679,8 @@ sealed abstract class IO[+A] extends internals.IOBinaryCompat[A] {
    * If [[parProduct]] is canceled, both actions are canceled.
    * Failure in either of the IOs will cancel the other one.
    *  */
-  def parProduct[B](another: IO[B])(implicit cs: ContextShift[IO]): IO[(A, B)] =
-    IO.Par.unwrap(IO.parApplicative.product(IO.Par(this), IO.Par(another)))
+  def parProduct[B](another: IO[B])(implicit p: NonEmptyParallel[IO]): IO[(A, B)] =
+    p.sequential(p.apply.product(p.parallel(this), p.parallel(another)))
 
   /**
    * Returns a new value that transforms the result of the source,
@@ -771,14 +771,14 @@ sealed abstract class IO[+A] extends internals.IOBinaryCompat[A] {
    * Failure in either of the IOs will cancel the other one.
    * If the whole computation is canceled, both actions are also canceled.
    * */
-  def &>[B](another: IO[B])(implicit cs: ContextShift[IO]): IO[B] =
-    IO.Par.unwrap(IO.parApplicative.productR(IO.Par(this))(IO.Par(another)))
+  def &>[B](another: IO[B])(implicit p: NonEmptyParallel[IO]): IO[B] =
+    p.parProductR(this)(another)
 
   /**
    * Like [[&>]], but keeps the result of the source.
    * */
-  def <&[B](another: IO[B])(implicit cs: ContextShift[IO]): IO[A] =
-    another &> this
+  def <&[B](another: IO[B])(implicit p: NonEmptyParallel[IO]): IO[A] =
+    p.parProductL(this)(another)
 }
 
 abstract private[effect] class IOParallelNewtype extends internals.IOTimerRef with internals.IOCompanionBinaryCompat {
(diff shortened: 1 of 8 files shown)


rossabaker:
I'm mildly :+1: on the richer IO interface.  I think experienced users tend to forget the initial bewilderment we felt with implicit syntax.

I'm mildly :-1: on reordering the source.  I think discoverability comes most from autocomplete, and secondarily from scaladoc, both of which will generally be ordered.

djspiewak:
I'm strongly 👍 on the richer `IO` interface. I don't at all mind redundancy with syntax enrichments we can otherwise get from Cats, so long as we're comfortable with the associated maintenance burden (usually not high). Implementing tracing properly requires this sort of thing *anyway*, so I see no reason to not start eating the elephant now.

Regarding alphabetization… I'm kind of meh on it. It has the potential to create unnecessary merge conflicts and other weirdness, and the benefits are suspect. Most editors (even non-IDEs) provide some mechanism for viewing and searching structural members. As an example (Sublime Text):



Emacs and Vim are the obvious outliers here.

Either way though, I'm not sure it's worth the potential disruption. Open to being convinced. :-) If others are in favor then I have no objections.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -720,6 +745,38 @@ sealed abstract class IO[+A] extends internals.IOBinaryCompat[A] {
...
+  /*
+   * Ignores the result of this IO.
+   */
+  def void: IO[Unit] = map(_ => ())
+
+  /**
+   * Run the current IO, then run [[another]], keeping its result.
+   * The result of [[this]] is ignored.
+   * */
+  def *>[B](another: IO[B]): IO[B] = flatMap(_ => another)
```
djspiewak: Is it worth also adding `>>`? It annoys me that one is lazy and the other isn't, but breaking with Cats on this would be a nightmare from a usability standpoint.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1298,6 +1360,11 @@ object IO extends IOInstances {
       case Left(err) => raiseError(err)
     }
 
+  def fromOption[A](orElse: => Throwable)(option: Option[A]): IO[A] = option match {
```
djspiewak: Any particular reason for this parameter order? It feels more intuitive to me the other way around. FWIW, `ApplicativeError` has the parameters in the other order, and in the same parameter block.

kubukoz: No particular reason, although now I see we could provide `fromOptionK(orElse: => Throwable): Option ~> IO` and have consistent parameter ordering.

I'm slightly biased towards having a separate parameter block for thunks, so I'd keep it separate.

kubukoz:
Fair enough, ordering is probably more useful for contributing than using (easier to find a place for new methods). However, although looking for all symbols will result in a sorted list, some people (myself included) like to explore an API by going to the source directly - there's some value in having that sorted, I think. But I won't push for it as I understand the consequences (mangled history) :)

core/shared/src/main/scala/cats/effect/IO.scala
kubukoz: I don't have a strong preference - whatever others think. @rossabaker?

Avasil: I would rather have `>>` only over `*>` which is common point of confusion

djspiewak: Except we don't have a `<<` :-(

Agreed it's a super-common point of confusion, made far worse by the fact that one is lazy and the other isn't.

core/shared/src/main/scala/cats/effect/IO.scala
djspiewak: Yeah I'm *tentatively* fine with keeping it separate, depending on how important we feel consistency is. I would flip the order though.

SystemFw:
I think that when relevant things should take `Concurrent[IO]`, and not `ContextShift[IO]` , which is an implementation detail, doesn't carry when you abstract to `F`, and we've already agreed to change

core/shared/src/main/scala/cats/effect/IO.scala
kubukoz: I think for new users we could just as well advise using `*> IO.suspend(...)` instead of `>>`. I suggest we wait and see how many people complain about the lack of `>>` in the class :)

core/shared/src/main/scala/cats/effect/IO.scala
kubukoz: Sounds good

core/shared/src/main/scala/cats/effect/IO.scala
djspiewak: > I suggest we wait and see how many people complain about the lack of >> in the class :)

I'm okay with this approach. Lazily-evaluated API surface area.

kubukoz:
I think the Parallel instance takes and uses ContextShift directly, so we can't really go with Concurrent :/

djspiewak:
> I think the Parallel instance takes and uses ContextShift directly

😭

kubukoz:
Alright, that was a brain fart. We can just take a `NonEmptyParallel` implicit instead :)

satorg:
Hi guys.
Sorry, just have noticed this PR.
My question is about the new `IO.as` method - it takes a lazy parameter while the corresponding methods from `Functor` and `FunctorOps` take a regular one.
I wonder what was the reasoning for changing its behavior?
Is it somehow better comparing to the other `as`-es?
My concern is that it may introduce a discrepancy between different variants of `as`.
Is it supposed to change the other `as` methods to take their parameter lazily in the future?

kubukoz:
@satorg it's a good question. 

The main reason for using laziness is that maybe you don't want to calculate the value of the expression if the IO you're doing `.as` on fails, is canceled or is just non-terminating.

Now I realize there are perfectly valid use cases for when you want to pass a precomputed value in `as` - for example, if it's the same value in multiple executions of the same program, or you just want it precomputed  - but it's still possible by assigning a `val` or `lazy val` before the expression using that value.

If you'd wanted to go the other way and we had an `as` that takes an eager parameter, you'd be forced to use `map(_ => x)`.

Changing `as` in Functor(ops) might be possible in a backwards-compatible way if we used a trick similar to #835, but I'm not sure that'll be desirable, for the same reason - people might prefer it eager. Especially with collection types, an `as` would result in multiple evaluations of the same expression, which in a pure context should yield the same results, so the implementations would have to optimize for that with a lazy val. I'm not sure it's worth it.

satorg:
@kubukoz got it, thanks for the explanations. This all makes sense.

But why not to give that lazy `as` a slightly different name then? `delayAs` for example?
Or make it something like `asEval(a: Eval[A])` which would allow to choose how to control the evaluation?

PR #810, opened 2020-03-06
WIP: allocatedCase

kubukoz:
Calling the finalizer returned by `Resource#allocated` always results in the Completed case.

This defines an alternative that allows specifying the exit case, making `allocated` dual to the `Allocate` node.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Resource.scala
+++ b/core/shared/src/main/scala/cats/effect/Resource.scala
@@ -243,23 +243,26 @@ sealed abstract class Resource[+F[_], +A] {
    * resource.
    *
    */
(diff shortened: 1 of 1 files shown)


rossabaker (commented):
Is there a reason we wouldn't prefer to use this over `allocated`?  Anyone using this API is far into the weeds anyway, and it seems more correct.  Should we also consider deprecating `allocated`?

kubukoz:
Not sure. Sounds like a good idea :)

I'll need to update the tests (making sure all nodes get the correct exit case if it's not Completed) and we'll see how it goes... but I definitely want to have `allocatedCase` of this shape.

PR #812, opened 2020-03-08
Remove sys.exit calls on JS runtime

### LINKED ISSUE #809: IOApp linker error with ScalaJS 1.0.0
While updating a project from ScalaJS 0.6.32 to 1.0.0 I encountered a linker error for IOApp:

> [error] Referring to non-existent method static java.lang.System.exit(int)void
[error]   called from scala.sys.package$.exit(int)nothing
[error]   called from private cats.effect.internals.IOAppPlatform$.setExitCode(int)void
[error]   called from private cats.effect.internals.IOAppPlatform$.$anonfun$main$2(scala.util.Either)void
[error]   called from cats.effect.internals.IOAppPlatform$.main([java.lang.String,cats.Eval,cats.Eval,scala.Function1)void
[error]   called from cats.effect.IOApp.main([java.lang.String)void
[error]   called from test.TestApp$.main([java.lang.String)void
[error]   called from static test.TestApp.main([java.lang.String)void
[error]   called from core module module initializers
[error] involving instantiated classes:
[error]   scala.sys.package$
[error]   cats.effect.internals.IOAppPlatform$
[error]   test.TestApp$
[error] There were linking errors

The problem boils down to the removal of java.lang.System.exit from ScalaJS with [#3520](https://github.com/scala-js/scala-js/pull/3520).

I created a minimal sample project that shows the problem [JSIOApp.tar.gz](https://github.com/typelevel/cats-effect/files/4294837/JSIOApp.tar.gz). The fastOptJS command fails with the aforementioned error.

Used library versions are cats-core 2.1.1 and cats-effect 2.1.2.

rossabaker:
Fixes #809 

Disclaimer: I don't use scala.js, haven't written serious JavaScript in a decade, and am utterly unqualified to be working on this side of the house.

### DIFF
--- a/core/js/src/main/scala/cats/effect/internals/IOAppPlatform.scala
+++ b/core/js/src/main/scala/cats/effect/internals/IOAppPlatform.scala
@@ -39,21 +39,13 @@ private[effect] object IOAppPlatform {
   }
 
   /**
-   * Sets the exit code with `process.exitCode = code` for runtimes
-   * that support it.  This allows a graceful shutdown with a specific
-   * exit code.
-   *
-   * If the call is not supported, does a `sys.exit(code)` on any
-   * non-zero exit code.
+   * Sets the exit code with `process.exitCode = code`.  This allows a graceful
+   * shutdown with a specific exit code.
    *
    * @see https://nodejs.org/api/process.html#process_process_exitcode
    **/
   private def setExitCode(code: Int): Unit =
-    try js.Dynamic.global.process.exitCode = code
-    catch {
-      case _: js.JavaScriptException =>
-        if (code != 0) sys.exit(code)
-    }
+    js.Dynamic.global.process.exitCode = code
 
   def mainFiber(args: Array[String], contextShift: Eval[ContextShift[IO]], timer: Eval[Timer[IO]])(
     run: List[String] => IO[ExitCode]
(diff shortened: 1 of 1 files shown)


core/js/src/main/scala/cats/effect/internals/IOAppPlatform.scala
```diff
@@ -39,21 +39,13 @@ private[effect] object IOAppPlatform {
...
    *
    * @see https://nodejs.org/api/process.html#process_process_exitcode
    **/
   private def setExitCode(code: Int): Unit =
-    try js.Dynamic.global.process.exitCode = code
-    catch {
-      case _: js.JavaScriptException =>
-        if (code != 0) sys.exit(code)
-    }
+    js.Dynamic.global.process.exitCode = code
```
rossabaker: I have found this documented as a standard in Node, and tested it in Node.  I can no longer remember which runtime inspired the catch block, but since the fallback has been eliminated, I guess I'm content to crash with a JavaScriptException on the off chance someone is using `IOApp` and not running on `node`. :man_shrugging:

core/js/src/main/scala/cats/effect/internals/IOAppPlatform.scala
```diff
@@ -89,7 +81,7 @@ private[effect] object IOAppPlatform {
       () =>
         fiber.cancel.unsafeRunAsync { result =>
           result.swap.foreach(Logger.reportFailure)
-          sys.exit(code + 128)
+          setExitCode(code + 128)
```
rossabaker: Contrary to @djspiewak's comment, this is referenced.  Without it, the process exits with 0 when killed.  I tested using `setExitCode` and still get the expected 143.

Issue #806, opened 2020-03-04
Missing inductive typeclass derivations


djspiewak:
For example, we have no implicit on `Bracket`'s companion for materializing a `Bracket[OptionT[F, *], E]` given a `Bracket[F, E]`, which we 100% should have. Similarly for `EitherT` and a bunch of others. The only one present right now derives `Kleisli`, which is useful but we should have the others, too.

This has been a problem on a few other typeclasses as well. At the very least, `Bracket` needs every derivation that `Sync` has, and similarly `ConcurrentEffect` needs everything that `Effect` has (since they are identical in power). Etc etc

ybasket:
I had a look and see why `Kleisli` is the only monad transformer with an instance in `Bracket`'s companion. All other monad transformers implement their `bracketCase` using a `Ref` which requires `Sync`. The reason seems to be the `bracketPropagatesTransformerEffects` law, especially in the `release` - if that returns `None` (for `OptionT`), this should propagate to the final result of the `bracketCase` call. Similar for the other transformers. I don't see any obvious way to achieve the same behavior without a `Ref`.

We could still add some proxies to `Bracket` for the cases where a `Sync` is available:
```scala
object Bracket {
  ...
  implicit def catsOptionTBracket[F[_]](implicit ev: Sync[F]): Bracket[OptionT[F, *], Throwable] =
    Sync.catsOptionTSync[F]
}
```
This would solve most of the problem (except for non-`Throwable` brackets and situations where no `Sync` is in scope) of bracket syntax not being available due to no `Bracket` being in the implicit scope, but it feels odd to have this link `Bracket` -> `Sync`. WDYT?

djspiewak:
Not gonna lie, that law took me quite a bit of digestion time to comprehend.

It makes sense that this is a bit problematic. This honestly makes me wonder if `bracket` is even a particularly good signature, given the problems of defining instances inductively. At any rate… The forwarders seem like a halfway-decent compromise, though I'm concerned that they might create a bit of an implicit mess. I guess the best thing to do would be to try it and see? I believe you're right that your proposed forwarder should solve the syntax problem.

kiranbayram:
This seems to resolved, shouldn't it be closed?

PR #814, opened 2020-03-10
Add implicit proxies for transformers in Bracket

ybasket:
This enables the smooth usage of `Bracket` syntax on monad transformers in cases where a `Sync` instance is present. See #806 for background


djspiewak:
I'll release a snapshot which contains these changes and we can try it in various contexts.

PR #827, opened 2020-04-01
Add Ref.lens constructor

jwojnowski:
I find `Ref.lens` useful in tests, where a single `Ref` aggregating state of multiple mocks can be transformed into a few "focused" `Refs`, ready for use in the mocks.

I'm not sure, though, if it should be a part of `cats-effect` or rather some small satellite library, so I'll leave it up to you to decide :wink:

### DIFF
--- a/core/shared/src/main/scala/cats/effect/concurrent/Ref.scala
+++ b/core/shared/src/main/scala/cats/effect/concurrent/Ref.scala
@@ -226,6 +226,23 @@ object Ref {
    */
   def unsafe[F[_], A](a: A)(implicit F: Sync[F]): Ref[F, A] = new SyncRef[F, A](new AtomicReference[A](a))
 
+  /**
+   * Creates an instance focused on a component of another Ref's value.
+   * Delegates every get and modification to underlying Ref, so both instances are always in sync.
+   *
+   * Example:
+   *
+   * {{{
+   *   case class Foo(bar: String, baz: Int)
+   *
+   *   val refA: Ref[IO, Foo] = ???
+   *   val refB: Ref[IO, String] =
+   *     Ref.lens[IO, Foo, String](refA)(_.bar, (foo: Foo) => (bar: String) => foo.copy(bar = bar))
+   * }}}
+   * */
+  def lens[F[_], A, B <: AnyRef](ref: Ref[F, A])(get: A => B, set: A => B => A)(implicit F: Functor[F]): Ref[F, B] =
+    new LensRef[F, A, B](ref)(get, set)
+
   final class ApplyBuilders[F[_]](val F: Sync[F]) extends AnyVal {
 
     /**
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/effect/concurrent/Ref.scala
```diff
@@ -303,6 +320,67 @@ object Ref {
...
+      tryModify(a => f(a).value)
+    }
+
+    override def modifyState[C](state: State[B, C]): F[C] = {
+      val f = state.runF.value
+      modify(a => f(a).value)
+    }
+
+    override def access: F[(B, B => F[Boolean])] =
+      F.map(underlying.get) { snapshotA =>
```
kubukoz, line 367: If we're going to keep the semantics (writes to A that don't touch B don't break access's setter), we'll need to make this pure:

```scala
override val access: F[(B, B => F[Boolean])] =
  F.flatMap(underlying.get) { snapshotA =>
    val snapshotB = lensGet(snapshotA)
    val setter = Sync[F].delay {
      val hasBeenCalled = new AtomicBoolean(false)

      (b: B) => {
        F.flatMap(Sync[F].delay(hasBeenCalled.compareAndSet(false, true))) { hasBeenCalled =>
          F.map(underlying.tryModify { a =>
            if (hasBeenCalled && (lensGet(a) eq snapshotB))
              (lensSet(a)(b), true)
            else
              (a, false)
          })(_.getOrElse(false))

        }
      }
    }
    setter.tupleLeft(snapshotB)
  }
```

of course that requires increasing the constraints to Sync too

kubukoz:
Some discussion on Gitter about the semantics (changes to A that don't involve changes in B): https://gitter.im/typelevel/cats-effect-dev?at=5e8b0e65d021c45cc7c995a6

rkrzewski:
I was looking at https://github.com/oleg-py/meow-mtl recently and I believe it covers your use case. It will both derive a `MonadState[F, A]` for a `Ref[F, A]` and also derive `MonadState[F, Part]` for a `MonadState[F, Whole]` whenever a `Lens[Whole, Part]` exists.

Of course `meow-mtl` involves a fair bit of magic based on even more magic in `shapeless` so I absolutely see merit in your lightweight solution!

kubukoz:
@rkrzewski we're using meow-mtl as well, the problem is that `MonadState` doesn't offer a method like `modify` from `Ref` :( and I haven't been able to get it done in https://github.com/typelevel/cats-mtl/pull/120 so far.

kubukoz:
@djspiewak @SystemFw thoughts? I think this belongs here, but I remember you had different opinions on it

SystemFw:
On balance, I like it and I'd like to see it included

core/shared/src/main/scala/cats/effect/concurrent/Ref.scala
```diff
@@ -226,6 +226,23 @@ object Ref {
    */
   def unsafe[F[_], A](a: A)(implicit F: Sync[F]): Ref[F, A] = new SyncRef[F, A](new AtomicReference[A](a))
 
+  /**
+   * Creates an instance focused on a component of another Ref's value.
```
kubukoz, line 230: I think we should briefly mention that updates in the parent that don't change the child won't disable the setter in the child's `access`.

kubukoz:
Hey @djspiewak, any chance we can release this soon?

PR #833, opened 2020-04-14
Add warnings to docs about IOApp cancelation issues within sbt.

### LINKED ISSUE #831: Add documentation about IOApp not terminating in sbt without fork
> IOApp shutdown uses jvm shutdown hooks, sbt doesn't terminate the jvm without fork (because it uses the same jvm for itself + your app)

(from https://gitter.im/typelevel/cats-effect?at=5e908d1a6823cb38acc1b5ae)

Add some documentation where IOApp + sbt is mentioned:
- https://typelevel.org/cats-effect/datatypes/ioapp.html#cancelation-and-safe-resource-release
- https://typelevel.org/cats-effect/tutorial/tutorial.html#copy-program-code

Perhaps a runtime check in IOApp to detect if the program is running under sbt? I don't know if this is either possible or desired.

arosien:
Fixes #831.

### DIFF
--- a/site/src/main/mdoc/datatypes/ioapp.md
+++ b/site/src/main/mdoc/datatypes/ioapp.md
@@ -86,10 +86,10 @@ interruption signal that can be caught, then the `IO` app will cancel
 and safely release any resources.
 
 *WARNING*: If you run your `IOApp` program from sbt, you may observe
-cancelation and resource releasing is *not* happening. This is due to
+cancellation and resource releasing is *not* happening. This is due to
 sbt, by default, running programs in the _same_ JVM as sbt, so when
 your program is canceled sbt avoids stopping its own JVM. To properly
-allow cancelation, ensure your progam is [forked into its own JVM](https://www.scala-sbt.org/1.x/docs/Forking.html#Enable+forking)
+allow cancellation, ensure your progam is [forked into its own JVM](https://www.scala-sbt.org/1.x/docs/Forking.html#Enable+forking)
 via a setting like `fork := true` in your sbt configuration.
 
 For example:
--- a/site/src/main/mdoc/datatypes/ioapp.md
+++ b/site/src/main/mdoc/datatypes/ioapp.md
@@ -85,6 +85,13 @@ This means that when `IOApp` receives a `SIGABORT`, `SIGINT` or another
 interruption signal that can be caught, then the `IO` app will cancel
 and safely release any resources.
 
+*WARNING*: If you run your `IOApp` program from sbt, you may observe
+cancelation and resource releasing is *not* happening. This is due to
+sbt, by default, running programs in the _same_ JVM as sbt, so when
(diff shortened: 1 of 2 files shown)


site/src/main/mdoc/datatypes/ioapp.md
```diff
@@ -85,6 +85,13 @@ This means that when `IOApp` receives a `SIGABORT`, `SIGINT` or another
 interruption signal that can be caught, then the `IO` app will cancel
 and safely release any resources.
 
+*WARNING*: If you run your `IOApp` program from sbt, you may observe
+cancelation and resource releasing is *not* happening. This is due to
+sbt, by default, running programs in the _same_ JVM as sbt, so when
+your program is canceled sbt avoids stopping its own JVM. To properly
+allow cancelation, ensure your progam is [forked into its own JVM](https://www.scala-sbt.org/1.x/docs/Forking.html#Enable+forking)
```
rossabaker, line 92: Nit: we standardized on this spelling of cancellation.

```suggestion
cancellation and resource releasing is *not* happening. This is due to
sbt, by default, running programs in the _same_ JVM as sbt, so when
your program is canceled sbt avoids stopping its own JVM. To properly
allow cancellation, ensure your progam is [forked into its own JVM](https://www.scala-sbt.org/1.x/docs/Forking.html#Enable+forking)
```

site/src/main/mdoc/tutorial/tutorial.md
```diff
@@ -466,6 +466,11 @@ run, still resources will be properly released. But recall what we commented
 before: this is because `use` returns `IO` instances that are cancelable, in
 contrast our `transfer` function is not cancelable.
 
+WARNING: To properly test cancelation, You should also ensure that
+`fork := true` is set in the sbt configuration, otherwise sbt will
+intercept the cancelation because it will be running the program
```
rossabaker, line 471:
```suggestion
WARNING: To properly test cancellation, you should also ensure that
`fork := true` is set in the sbt configuration, otherwise sbt will
intercept the cancellation because it will be running the program
```

site/src/main/mdoc/tutorial/tutorial.md
```diff
@@ -466,6 +466,11 @@ run, still resources will be properly released. But recall what we commented
 before: this is because `use` returns `IO` instances that are cancelable, in
 contrast our `transfer` function is not cancelable.
 
+WARNING: To properly test cancelation, You should also ensure that
+`fork := true` is set in the sbt configuration, otherwise sbt will
```
rossabaker, line 470: I like how you linked instructions for this in the other one.  Perhaps here, too.

site/src/main/mdoc/datatypes/ioapp.md
arosien: 🙀

PR #835, opened 2020-04-15
Remove Bracket constraint from Resource.mapK

kubukoz:
As far as I could see, in the current form it's only used to satisfy the same constraint when calling `mapK` recursively.

I added an overload and made the old one private, so code outside of CE shouldn't be hit with any incompatibility. I had to merge the argument lists to reduce the ambiguity locally, but it compiles to the same signature in JVM bytecode, so we should be good.


djspiewak:
LGTM!

kubukoz:
Would you mind making a release? It's been some time and we got some new stuff :)

Issue #828, opened 2020-04-04
Fiber identity


RaasAhsan:
It seems like Cats Effect today doesn't support any notion of fiber identity, at least via the public API. This has several interesting use cases:
1. Fiber logging: There is a disconnect between existing JVM logging libraries and fiber-based runtimes (or more generally, asynchronous code). The loggers typically output the ID of the thread on which it is running, but oftentimes that name is useless since fiber execution can take place on potentially many threads. Being able to aggregate the logs produced by a fiber could be pretty powerful.
2. Fiber-local state: which would be useful in situations where you want to pass along state for dealing with cross-cutting concerns, like passing along a trace ID that is exchanged in a chain of microservice calls. There is an argument that it suffers from similar problems as thread-local storage.
3. Fiber ancestry: fiber forking generates a directed acyclic graph where there are directed edges between a parent and its children. Unsure if there are any compelling use cases for this but it could provide for some interesting analysis.
4. Fiber tracing: There's already an issue open for this and it might not strictly fall under fiber identity, but I could see some overlap here in tracking state.

Some other effect types already support fiber identity to some degree, so maybe there's room for some effect class here, but I don't really know how much abstraction can be captured.

Does any of this make sense and does it even fall under the scope of Cats Effect IO or its classes? Would be interested in fleshing it out more if so.

djspiewak:
Sorry for taking forever to loop back to this…

Broadly speaking, I agree with you wholeheartedly. Having some notion of fiber identity is important for a lot of reasons, but particularly for things like fiber locals and such. Doing this in a sane way is immensely complicated though and I haven't had time to think deeply about what it should look like. There's also a whole series of real risks when you do something like that (see: all the abuse `ThreadLocal` gets). Also how to do it in a way which is amenable to abstraction.

But in general I do think this is very much a problem with tackling. Some of the things I'm thinking about:

- How does fiber identity relate to its ancestry tree? (note that it *is* a tree, not a DAG, unless you consider `join` to be an inbound edge, in which case it could actually be a fully cyclic graph) The answer to this question also dictates how a hypothetical `FiberLocal` would behave in the presence of `start`
- What kind of control and information does `self` fiber introspection give? Should you be able to get a reference to the `Fiber` instance? If so then you would be able to do things like `join` or `cancel` on yourself (the former being a deadlock, the latter likely leading to deadlocks indirectly in CE2).
[...]

PR #836, opened 2020-04-17
WIP: Fiber local state PoC

RaasAhsan:
This a PoC for supporting fiber-local state inside `cats.effect.IO`. I began to implement some of the ideas I laid out in #828 . I didn't get much feedback in the original issue, so hopefully there can be some more discussion here. I'd certainly like to hear whether or not this even belongs in this library.

## Changes
- Added a new `IO` primitive operation, `FiberLocal`, which ideally encapsulates all fiber-local behavior
- Support `FiberLocal` instruction in the `IO` interpreter.
- Added a new type `FiberRef` that leverages the new `IO` op to implement fiber local state. Unfortunately this is specific to `IO`.

## Usage
```scala
def printLine[A: Show](msg: A): IO[Unit] =
    IO.delay(println(msg.show))

  def program: IO[Unit] =
    for {
      ref <- FiberRef.of(5)
      a1  <- ref.get
      _   <- printLine(a1)
      _   <- ref.set(10)
      a2  <- ref.get
      _   <- printLine(a2)
    } yield ()
```

## TODO/Ideas
- [ ] Tests
- [ ] Docs
- [ ] Cleanup
- [ ] Java ThreadLocal integration?
- [ ] think about forking behavior

Adding onto some of the ideas in #828, you could start to implement some very basic, runtime non-invasive, application tracing. You could implement a userspace `Tracer` type with a `tracepoint` method that captures slices of stack traces and holds it in fiber-local state

cc @djspiewak @alexandru

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -27,8 +27,6 @@ import scala.util.control.NonFatal
 import scala.util.{Failure, Left, Right, Success, Try}
 import cats.data.Ior
 
-import scala.collection.mutable
-
 /**
  * A pure abstraction representing the intention to perform a
  * side effect, where the result of that side effect may be obtained
@@ -1625,5 +1623,5 @@ object IO extends IOInstances {
       Pure(Left(e))
   }
 
-  final private[effect] case class FiberLocal[+A](k: mutable.HashMap[FiberRefId, AnyRef] => A) extends IO[A]
+  private[effect] case object Introspect extends IO[FiberState]
 }
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -1610,6 +1610,8 @@ object IO extends IOInstances {
     restore: (A, Throwable, IOConnection, IOConnection) => IOConnection
   ) extends IO[A]
 
+  private[effect] case object Introspect extends IO[FiberLocals]
+
   /* -=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-= */
 
   /**
@@ -1623,5 +1625,4 @@ object IO extends IOInstances {
       Pure(Left(e))
   }
 
-  private[effect] case object Introspect extends IO[FiberState]
 }
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -27,6 +27,8 @@ import scala.util.control.NonFatal
 import scala.util.{Failure, Left, Right, Success, Try}
 import cats.data.Ior
 
+import scala.collection.mutable
+
 /**
  * A pure abstraction representing the intention to perform a
  * side effect, where the result of that side effect may be obtained
@@ -1622,4 +1624,6 @@ object IO extends IOInstances {
     override def recover(e: Throwable) =
       Pure(Left(e))
   }
+
+  final private[effect] case class FiberLocal[+A](k: mutable.HashMap[FiberRefId, AnyRef] => A) extends IO[A]
 }
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -1201,7 +1201,7 @@ object IO extends IOInstances {
    * @see [[asyncF]] and [[cancelable]]
    */
   def async[A](k: (Either[Throwable, A] => Unit) => Unit): IO[A] =
-    Async { (_, cb) =>
+    Async { (_, _, cb) =>
       val cb2 = Callback.asyncIdempotent(null, cb)
       try k(cb2)
       catch { case NonFatal(t) => cb2(Left(t)) }
@@ -1232,7 +1232,7 @@ object IO extends IOInstances {
    * @see [[async]] and [[cancelable]]
    */
   def asyncF[A](k: (Either[Throwable, A] => Unit) => IO[Unit]): IO[A] =
-    Async { (conn, cb) =>
+    Async { (conn, _, cb) =>
       // Must create new connection, otherwise we can have a race
       // condition b/t the bind continuation and `startCancelable` below
       val conn2 = IOConnection()
@@ -1285,7 +1285,7 @@ object IO extends IOInstances {
    *      the underlying cancelation model
    */
   def cancelable[A](k: (Either[Throwable, A] => Unit) => CancelToken[IO]): IO[A] =
-    Async { (conn, cb) =>
+    Async { (conn, _, cb) =>
       val cb2 = Callback.asyncIdempotent(conn, cb)
       val ref = ForwardCancelable()
       conn.push(ref.cancel)
@@ -1473,7 +1473,7 @@ object IO extends IOInstances {
    * }}}
    */
   val cancelBoundary: IO[Unit] = {
-    val start: Start[Unit] = (_, cb) => cb(Callback.rightUnit)
+    val start: Start[Unit] = (_, _, cb) => cb(Callback.rightUnit)
     Async(start, trampolineAfter = true)
   }
 
@@ -1593,7 +1593,7 @@ object IO extends IOInstances {
    *        signal downstream
    */
   final private[effect] case class Async[+A](
-    k: (IOConnection, Either[Throwable, A] => Unit) => Unit,
+    k: (IOConnection, IOContext, Either[Throwable, A] => Unit) => Unit,
     trampolineAfter: Boolean = false
   ) extends IO[A]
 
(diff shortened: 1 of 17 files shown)


djspiewak:
Oooooh! 👀

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1622,4 +1624,6 @@ object IO extends IOInstances {
     override def recover(e: Throwable) =
       Pure(Left(e))
   }
+
+  final private[effect] case class FiberLocal[+A](k: mutable.HashMap[FiberRefId, AnyRef] => A) extends IO[A]
```
RaasAhsan, line 1628: This probably doesn't need to be mapping function. It could return the state directly.
Using a `mutable.HashMap` might also have implications on thread safety

RaasAhsan:
An alternative implementation I had in mind was to give the responsibility of storage to each `FiberRef`. So each `FiberRef` would keep a reference to some `ConcurrentHashMap`-like data structure that indexes values by a unique fiber ID. The benefit of this approach is that it only requires the fiber identity to be supported by the runtime. 

On the other hand, it creates a more complex public API where users have to interact with fiber-local state via `Resource`, since state isn't automatically garbage collected when a fiber exits anymore. But I think we also lose the ability to do any useful state manipulation on forks and joins.

core/shared/src/main/scala/cats/effect/IO.scala
RaasAhsan: Maybe `Introspect` is a better name

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1593,7 +1593,7 @@ object IO extends IOInstances {
    *        signal downstream
    */
   final private[effect] case class Async[+A](
-    k: (IOConnection, Either[Throwable, A] => Unit) => Unit,
+    k: (IOConnection, IOContext, Either[Throwable, A] => Unit) => Unit,
```
RaasAhsan, line 1596: I ended up having to supply the `IOContext` through the `Async` constructor so it could be threaded through multiple asynchronous operations that are logically linked to the same fiber (`IOBracket` implementation). Not sure if this totally makes sense, but it eliminates the `Introspect` constructor for now.

Ideally we could pass state via method parameters, but that would be a more invasive change. I'm also not happy that `IOContext` has to be thread-safe now. But this implementation works.

RaasAhsan: I believe #681 and a revise of the `bracket` implementation would fix this.

core/shared/src/main/scala/cats/effect/internals/IOContext.scala
```diff
@@ -0,0 +1,47 @@
...
+
+import java.util.concurrent.ConcurrentHashMap
+
+import cats.effect.FiberRef
+
+/**
+ * Represents the state of execution for a
+ * single fiber.
+ */
+final private[effect] class IOContext {
```
RaasAhsan, line 27: `IOContext` encapsulates state around the execution of a logical fiber. Its a thread-safe implementation for now, but ideally it doesn't have to be since it should only ever be accessed by one thread at a time.

core/shared/src/main/scala/cats/effect/internals/IOFiberRef.scala
```diff
@@ -0,0 +1,38 @@
...
+
+import cats.effect.{FiberRef, IO}
+
+private[effect] object IOFiberRef {
+
+  def get[A](ref: FiberRef[A]): IO[Option[A]] =
+    IO.Async { (_, ctx, cb) =>
+      val k = ref.asInstanceOf[FiberRef[Any]]
+      val v = ctx.getLocal(k).map(_.asInstanceOf[A])
+      cb(Right(v))
```
kubukoz, line 27: I think we'll need to come up with some mechanism to ensure more type safety, but we can leave that till we've decided on the semantics :)

core/shared/src/main/scala/cats/effect/internals/IOContext.scala
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  def getLocal(key: FiberRef[Any]): Option[Any] =
+    Option(locals.get(key))
+
+}
+
+object IOContext {
+
+  def apply(): IOContext =
```
kubukoz, line 44: This breaks referential transparency so probably `unsafeApply` would be more apt

core/shared/src/main/scala/cats/effect/internals/IOContext.scala
kubukoz: I would assume it's okay to not care about thread safety in this case. But I might be missing something.

core/shared/src/main/scala/cats/effect/internals/IOContext.scala
```diff
@@ -0,0 +1,47 @@
...
+
+import cats.effect.FiberRef
+
+/**
+ * Represents the state of execution for a
+ * single fiber.
+ */
+final private[effect] class IOContext {
+
+  // TODO: lazy?
```
kubukoz, line 29: Why?

kubukoz (commented):
I'm really excited about this!

As for the behavior in forking, I think we could make it customizable per ref. Some things you might want to pass to children in a copy of the context (e.g. trace/span ID) but some would be local to each fiber (I can't think of an example right now but I'm sure someone will have reasons).

kubukoz:
@RaasAhsan could you add an example somewhere in the code? It'd be useful to see how it might look like in practice.

core/shared/src/main/scala/cats/effect/internals/IOFiberRef.scala
RaasAhsan: Yeah, I agree. Fortunately, `IOFiberRef` and `IOContext` are intended to be internal APIs, so end users can't access it. The implementations of `FiberRef` and `IOFiberRef` guarantee that you can never perform an invalid cast, since the object reference itself is used as a key.

core/shared/src/main/scala/cats/effect/internals/IOContext.scala
RaasAhsan: This is intended to be an internal API, similar to `IOConnection`, which is implemented similarly. Probably need to add that as a comment

core/shared/src/main/scala/cats/effect/internals/IOContext.scala
RaasAhsan: The issue is that `IOContext` needs to be threaded through multiple run-loops because of the `IOBracket` implementation, which also leverages the `Async` constructor. The "secondary" run-loops can manipulate fiber-local state on another thread (!!), and those changes need to propagate back to the original run-loop after the async callback is invoked. However, the current API doesn't lend itself to passing state to and back without some significant refactors. So the compromise for now is to make `IOContext` thread-safe so multiple threads/run-loops can access and manipulate that state safely.

core/shared/src/main/scala/cats/effect/internals/IOContext.scala
RaasAhsan: It was just a thought but probably not necessary 😅

RaasAhsan:
> As for the behavior in forking, I think we could make it customizable per ref. Some things you might want to pass to children in a copy of the context (e.g. trace/span ID) but some would be local to each fiber (I can't think of an example right now but I'm sure someone will have reasons).

I agree, I think that would be the best way to unify these to find some sort of abstraction. Some variants I've thought of:
1. Copy on merge - pass along a copy state to children. the parent and children can continue manipulating that state without either observing the others changes
2. Copy on merge + merge on fork - pass along children to state, supply a merge function to combine them again on join (ZIO semantics)
3. Shared reference - parents and children manipulate the same ref. almost like a `Ref` but its scoped to lineage of a fiber (Monix semantics)
4. Isolated reference - no copying on merge, no merge on fork. this is basically the PR implementation right now.

Its worth pointing out that #2 is a special case of #1 where the merge function merely drops the joined fiber's state value.

> could you add an example somewhere in the code? It'd be useful to see how it might look like in practice.
Added one in the `coreJVM` scope. let me know if that works for you

oleg-py:
As a main proponent for "monix semantics" in monix, I want to point out that we still have this as a law:
https://github.com/typelevel/cats-effect/blob/series/2.x/laws/shared/src/main/scala/cats/effect/laws/ConcurrentLaws.scala#L92
```scala
  def startJoinIsIdentity[A](fa: F[A]) =
    F.start(fa).flatMap(_.join) <-> fa
```

And you can observe the violation for `val fa = aFiberRef.set("foo")`

If we want to keep the law (and that's a question), then the only reasonable behavior on join would be to actually replace the values from joined fiber into the current one, with no special logic. It's not enough for practical reasons though, so we have `TaskLocal.isolate` that makes a snapshot of a state and restores it afterwards, so composed with `start` it makes _all_ fiber-local state have copy on fork + discard on join semantics.

It's a bit of a pain though, and configurable refs could reduce it, but not entirely. For example, monix `TaskLocal` is used in [quill](https://github.com/getquill/quill) to keep the current transaction, and whether you want sharing or isolated reference depends on whether you are going to join the child fibers in the parent or they outlive it

RaasAhsan:
I'm going to close this because CE3 will reach stable soon and we have a much simpler implementation for it in #1393

PR #837, opened 2020-04-17
Tutorial webpage : fix link to IO data type URL

nailyk:
In the https://typelevel.org/cats-effect/tutorial/tutorial.html webpage, a link redirecting to IO data type documentation is wrong (due to an issue on benbalter/jekyll-relative-links#61 )


nailyk:
@rossabaker i have just found out it's actually a current open issue on jekyll, which does not handle line-wrapped links (see https://github.com/benbalter/jekyll-relative-links/issues/61).
In the meantime the issue is fixed, i guess we still prefer the link is working on the website ? (or should i close this ?).
Thanks.

rossabaker:
Ah, nice find on the wrapping!

Issue #838, opened 2020-04-18
IOApp yields console errors when executed in browser


Taig:
Probably related to / introduced by #809 

**build.sbt**
```scala
enablePlugins(ScalaJSPlugin)

scalaVersion := "2.13.1"

libraryDependencies += "org.typelevel" %%% "cats-effect" % "2.1.3"

scalaJSUseMainModuleInitializer := true
```

**App.scala**
```scala
import cats.effect.{ExitCode, IO, IOApp}

object App extends IOApp {
  override def run(args: List[String]): IO[ExitCode] = IO(ExitCode.Success)
}
```

**index.html**
```html
<!DOCTYPE html>
<html lang="en">
<body>
<script src="target/scala-2.13/ce-test-fastopt.js"></script>
</body>
</html>
```

**Console output**
`scala.scalajs.js.JavaScriptException: ReferenceError: process is not defined`

![Bildschirmfoto 2020-04-18 um 12 24 29](https://user-images.githubusercontent.com/567857/79635198-d9951300-816f-11ea-9369-d73eb2018357.png)

rossabaker:
Yes, this almost certainly caused by [this](https://github.com/typelevel/cats-effect/pull/812#discussion_r389333325).

I am very ignorant of scala.js (and indeed, JS itself!).  What is the use case for `IOApp` inside the browser?  And is there any sense of exiting?  Any way to communicate an exit code?

Taig:
I don't know how/if the exit code is handled in the JS environment. The use case for me was simply the habit of wiring my app together in `IOApp`, not having to worry about `ContextShift` and `Timer` instances. I don't mind rolling my own `main` method for now and would offer to take a stab at this issue in the coming days.

rossabaker:
I think we could capture the `JavaScriptException` like we used to, but the question is what to do about it, since we can no longer exit.  Probably also worth thinking about how it operates in scala.js's other supported environments: there are more than node and the browser, though I don't know how prevalent they are.

PR #842, opened 2020-04-20
Avoid exception when using IOApp in browsers.

rpiaggio:
Fixes https://github.com/typelevel/cats-effect/issues/838 by checking if `global.process` exists before attempting to set the `exitCode`.

### DIFF
--- a/core/js/src/main/scala/cats/effect/internals/IOAppPlatform.scala
+++ b/core/js/src/main/scala/cats/effect/internals/IOAppPlatform.scala
@@ -43,13 +43,16 @@ private[effect] object IOAppPlatform {
    * that support it.  This allows a graceful shutdown with a specific
    * exit code.	
    *	
-   * If the call is not supported, no action is performed.
+   * If the call is not supported and the exit code is not Success,
+   * then it is logged.
    *
    * @see https://nodejs.org/api/process.html#process_process_exitcode
    **/
   private def setExitCode(code: Int): Unit =
     if (js.typeOf(js.Dynamic.global.process) != "undefined")
       js.Dynamic.global.process.exitCode = code
+    else if (code != ExitCode.Success.code)
+      Logger.reportFailure(new RuntimeException(s"Non-zero exit code: [$code]"))
 
   def mainFiber(args: Array[String], contextShift: Eval[ContextShift[IO]], timer: Eval[Timer[IO]])(
     run: List[String] => IO[ExitCode]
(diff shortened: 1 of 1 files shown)


rpiaggio:
I guess we could throw an exception if `ExitCode != Success`?

core/js/src/main/scala/cats/effect/internals/IOAppPlatform.scala
```diff
@@ -85,7 +92,7 @@ private[effect] object IOAppPlatform {
         }
 
     IO {
-      if (!js.isUndefined(js.Dynamic.global.process)) {
+      if (js.typeOf(js.Dynamic.global.process) != "undefined") {
```
rpiaggio, line 95: This change follows the "Testing whether a global variable exists" advice in https://www.scala-js.org/doc/interoperability/global-scope.html, so that it works with Scalajs 1.x.

rossabaker: I wonder if there are at least constants for this `"undefined"`.  I agree per docs that it's the right thing to do, but on first glance, it's worse code. Certainly not a blocker.

rpiaggio: I completely agree with you. There doesn't seem to be a constant for this. From what I've seen, it's common practice in JS to use strings as constants. I find this easier to digest by thinking of them as "singleton type constants" instead of strings ;).

Also, I don't know if this is the right place to document this, but the logging of an error code won't work with Scala.js 1.x in the browser if you are using webpack (which IMO is a fairly common practice) with default configuration. This is because webpack by default introduces a shim to `process` that Scala.js 1.x sees in `global`, but it didn't in 0.6. Things will work as expected if this explicitly disabled (`node: false`) as detailed in: https://webpack.js.org/configuration/node/

alexandru: Doing `typeof identifier === "undefined"` is certainly standard JavaScript practice. Other possible values are `"object"`, `"number"`, `"string"`. TypeScript can also use these checks to refine the typing.

So yes, it's standard practice and totally fine. If there are constants, they should get exposed by ScalaJS, but going with a string is acceptable.

rossabaker:
Do we need an issue to track what @rpiaggio mentioned about webpack above, or do we now have sensible defaults?  I have such a poor understanding of Scala.js deployment models, I'm unqualified to judge myself.

PR #844, opened 2020-04-21
Fix resolution of Clock implicits

### LINKED ISSUE #784: Clock[OptionT[F, *]] leads to implicit derivation issues
Currently attempting to consume this middleware in http4s, [Caching.`no-store`](https://github.com/http4s/http4s/blob/master/server/src/main/scala/org/http4s/server/middleware/Caching.scala#L24) and the below is what we derive when using `HttpRoutes`.

Would be nice to not need to explicitly pick a derivation approach.

```
ambiguous implicit values:
 both method deriveOptionT in trait LowPriorityImplicits of type [F[_]](implicit F: cats.Functor[F], implicit clock: cats.effect.Clock[F])cats.effect.Clock[[β$2$]cats.data.OptionT[F,β$2$]]
 and method extractFromTimer in object Clock of type [F[_]](implicit timer: cats.effect.Timer[F])cats.effect.Clock[F]
 match expected type cats.effect.Clock[[A]cats.data.OptionT[F,A]]
```

joroKr21:
The default from Timer is the least specific.
But it was defined in a subtype and hence ambiguous.
To fix that move it to a supertype.

In an ideal world (Scala 2.13) we would not need any low priority traits.
But a bug in Scala 2.12 causes it to turn a blind eye to higher-kinded specificity.

Fixes #784

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Clock.scala
+++ b/core/shared/src/main/scala/cats/effect/Clock.scala
@@ -139,13 +139,6 @@ object Clock extends LowPriorityImplicits {
         F.delay(unit.convert(System.nanoTime(), NANOSECONDS))
     }
 
-  /**
-   * Default implicit instance — given there's an implicit [[Timer]]
-   * in scope, extracts a [[Clock]] instance from it.
-   */
-  implicit def extractFromTimer[F[_]](implicit timer: Timer[F]): Clock[F] =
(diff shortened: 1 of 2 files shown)


djspiewak:
I'm amazed this actually solves it! Like honestly really surprised. Nice work!

PR #845, opened 2020-04-24
Scala 2.13.2

joroKr21:


rossabaker:
Does metals work on 2.13.2 yet? I usually hold off on Scala bumps for that.

rossabaker:
Though I don't think it's the default (not near computer to check), so it probably doesn't matter much.

alexandru:
@rossabaker sorry, I thought it isn't a problematic change, seeing the tests pass.

I'm not using Metals, IDK.

If you want we can revert the change.

alexandru:
@rossabaker btw, this particular change is only related to what Travis uses, doesn't affect local dev.

rossabaker:
Nope, I withdraw my concern.  I've been burned on other projects by merging Scala bumps too quickly because Metals takes a few days to catch up, but you're right about it being CI-only.

PR #847, opened 2020-04-26
Ref#modifyOr and Ref#updateOr

RaasAhsan:
Add combinators that allow for short-circuit behavior in CAS retry loops. This is useful when you want to perform an atomic state transition under specific conditions, but if those conditions aren't met, terminate early so no extra contention is placed on the atomic.

Maybe there's room for expressing some of the existing combinators in terms of these new ones. I could also see an `updateMaybe` or `modifyMaybe` that just take a `A => Option[A]`.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/concurrent/Ref.scala
+++ b/core/shared/src/main/scala/cats/effect/concurrent/Ref.scala
@@ -120,9 +120,19 @@ abstract class Ref[F[_], A] {
   def update(f: A => A): F[Unit]
 
   /**
+   * Like `update`, but can terminate early with an error value without retrying.
+   */
+  def updateOr[E](f: A => Either[E, A]): F[Option[E]]
+
+  /**
    * Like `tryModify` but does not complete until the update has been successfully made.
    */
   def modify[B](f: A => (A, B)): F[B]
+
+  /**
+   * Like `modify`, but can terminate early with an error value without retrying.
+   */
+  def modifyOr[E, B](f: A => Either[E, (A, B)]): F[Either[E, B]]
 
   /**
    * Update the value of this ref with a state computation.
@@ -264,6 +274,9 @@ object Ref {
       (f(a), ())
     }
 
+    def updateOr[E](f: A => Either[E, A]): F[Option[E]] =
+      modifyOr(a => f(a).map((_, ()))).map(_.left.toOption)
+
     def modify[B](f: A => (A, B)): F[B] = {
       @tailrec
       def spin: B = {
@@ -271,6 +284,21 @@ object Ref {
         val (u, b) = f(c)
         if (!ar.compareAndSet(c, u)) spin
         else b
+      }
+      F.delay(spin)
+    }
+
+    def modifyOr[E, B](f: A => Either[E, (A, B)]): F[Either[E, B]] = {
+      @tailrec
+      def spin: Either[E, B] = {
+        val c = ar.get
+        f(c) match {
+          case Right((u, b)) => {
+            if (!ar.compareAndSet(c, u)) spin
+            else Right(b)
+          }
+          case Left(e) => Left(e)
+        }
       }
       F.delay(spin)
     }
@@ -295,7 +323,9 @@ object Ref {
     override def tryUpdate(f: A => A): G[Boolean] = trans(underlying.tryUpdate(f))
     override def tryModify[B](f: A => (A, B)): G[Option[B]] = trans(underlying.tryModify(f))
     override def update(f: A => A): G[Unit] = trans(underlying.update(f))
+    override def updateOr[E](f: A => Either[E, A]): G[Option[E]] = trans(underlying.updateOr(f))
     override def modify[B](f: A => (A, B)): G[B] = trans(underlying.modify(f))
+    override def modifyOr[E, B](f: A => Either[E, (A, B)]): G[Either[E, B]] = trans(underlying.modifyOr(f))
     override def tryModifyState[B](state: State[A, B]): G[Option[B]] = trans(underlying.tryModifyState(state))
     override def modifyState[B](state: State[A, B]): G[B] = trans(underlying.modifyState(state))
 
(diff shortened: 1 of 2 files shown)


alexandru:
This is pretty cool, except it breaks binary compatibility 😑

SystemFw:
I think you can express what you want with `Ref.access` https://github.com/typelevel/cats-effect/blob/master/core/shared/src/main/scala/cats/effect/concurrent/Ref.scala#L96

SystemFw:
```
def failingUpdate(f: A => Either[E, A]) : F[Either[E, A]] =
  yourRef.access.flatMap { case (a, set) =>
      f(a) match {
         case err @ Left(_) => e.pure[F]
         case ok @ Right(a) => set(a).ifM(ok.pure[F], failingUpdate(f))
     }
  }
```

Something like this

RaasAhsan:
@alexandru  I don't really get that one. Adding a trait method breaks backwards binary compatibility? 

@SystemFw great catch! Do you think interleaving a bind between the read and the CAS operation would have any material impact on performance or evaluation? Either way, I think having some variants of these functions available would be good, pending these compatibility issues 😓

alexandru:
@RaasAhsan 

> Adding a trait method breaks backwards binary compatibility?

Yes, if it's abstract and not just binary compatibility, but source compatibility too. Prior to Scala 2.12 it broke binary compatibility even if it had a default implementation — nowadays default implementations are often based on Java 8's default methods implementations for interfaces, so it has gotten better.

You can check with `sbt mimaReportBinaryIssues` btw, it's probably why the tests are failing in Travis. It's not perfect, we ignore issues for classes or methods that are "private" (even though we sometimes shouldn't), but in this case you're adding a method to a public trait.

So something like this can't go in until we decide to release a major version that can break binary compatibility. And given the foundational nature of the library, it might take some time.

RaasAhsan:
Thanks, that makes sense, and also sounds a little bit inconvenient :) Unfortunately, I don't think we can even write a default method here with my implementation. With @SystemFw implementation, we could write a default method with a `FlatMap` constraint or introduce methods into the companion object, but I don't think we want to go that direction either.

Is there a process for keeping track of breaking changes we may want to make in the future? just keep the PR open?

alexandru:
We have a label named "*Breaking Change*" that might help.

kubukoz:
@RaasAhsan 
I would suggest going with an implicit class and extension methods in the companion object. Not sure how others feel about that though.

RaasAhsan:
I think that's a reasonable compromise. If its ever planned to breaking release, could push it into the trait. Deferring to others though

kubukoz:
3.0 is still far from done so it'll take a while before a breaking release :)

RaasAhsan:
Things look good here now.

core/shared/src/main/scala/cats/effect/concurrent/Ref.scala
```diff
@@ -286,6 +288,44 @@ object Ref {
     }
   }
 
+  implicit class RefOps[F[_], A](val ref: Ref[F, A]) extends AnyVal {
```
kubukoz, line 291: Let's make the constructor private to the enclosing object and make the `ref` field private. Also, I'd suggest a final class.

core/shared/src/main/scala/cats/effect/concurrent/Ref.scala
```diff
@@ -286,6 +288,44 @@ object Ref {
...
+
+    /**
+     * Like `modify`, but can terminate early with an error value without retrying.
+     * Returns a value or error based on whether the update short-circuited.
+     */
+    def modifyOr[E, B](f: A => Either[E, (A, B)])(implicit F: Monad[F]): F[Either[E, B]] =
+      ref.access.flatMap {
+        case (a, set) =>
+          f(a) match {
+            case Right((a, b)) => set(a).ifM(F.pure(Right(b)), modifyOr(f))
```
kubukoz, line 322: Please use named parameters in ifM, it really helps understand which branch is which :)

core/shared/src/test/scala/cats/effect/concurrent/RefTests.scala
```diff
@@ -136,6 +136,101 @@ class RefTests extends AsyncFunSuite with Matchers {
     run(op.map(_ shouldBe false).void)
   }
 
+  test("updateMaybe - successful") {
+    val op = for {
+      r <- Ref[IO].of(0)
+      result <- r.updateMaybe {
+        case 0 => Some(1)
+        case _ => None
```
kubukoz, line 144: Just nitpicking, we can probably make all these functions constant (without a pattern match) - we only ever call them in one way.

core/shared/src/main/scala/cats/effect/concurrent/Ref.scala
RaasAhsan: > constructor private to the enclosing object

could you elaborate on this? if i'm understanding correctly, if the class or constructor is declared private, callers outside the object/package couldn't use it

kubukoz: Hah, yeah, scratch that. I forgot it's only reasonable if you have an implicit def separate from the class.

kubukoz:
@SystemFw @alexandru what do you think?

RaasAhsan:
Thinking back on this, I think another way to circumvent the compatibility issue is to provide the current methods in `RefOps` as the default implementation in `Ref` and override it with the more efficient implementation in `SyncRef`.

djspiewak:
> Thinking back on this, I think another way to circumvent the compatibility issue is to provide the current methods in RefOps as the default implementation in Ref and override it with the more efficient implementation in SyncRef.

I *think* this would work, since we're dropping 2.11 support. Would you like me to hold off on merging so you can explore this?

RaasAhsan:
@djspiewak Ah the reason we can't do this is that the default implementation itself needs a `Monad` constraint which would be inappropriate for the trait. So I think it's OK as is, and can consider introducing into the interface in CE3

core/shared/src/main/scala/cats/effect/concurrent/Ref.scala
```diff
@@ -286,6 +289,44 @@ object Ref {
     }
   }
 
+  implicit final class RefOps[F[_], A](private val ref: Ref[F, A]) extends AnyVal {
+
+    /**
+     * Like `update`, but can terminate early without retrying.
+     * Returns whether or not the update short-circuited.
+     */
+    def updateMaybe(f: A => Option[A])(implicit F: Monad[F]): F[Boolean] =
```
johnynek, line 298: I'm concerned about returning `true` to mean the update *didn't* happen, if I am reading this correctly. `Ref.tryUpdate(f)` returns `true` to mean the update *did* happen. Both methods use the name `update` but use the Boolean in opposite directions to signal if an update took place.

I know there are two different causes here:
1. in tryUpdate, false means there was a race and we lost the race.
2. in updateMaybe, false means after 0 or more retries f eventually returned `None`

Since we haven't published a final version yet, it seems the horse isn't out of the barn yet.

I would say, it would maybe be nice to have:
```scala
sealed abstract class Modified(val success: Boolean)
object Modified {
  final case object Failure extends Modified(false)
  final case object Success extends Modified(true)
}
```
Then we could use `F[Modified]` here to be more clear about what `true` means.

PR #849, opened 2020-04-29
Implemented very silly workaround for #807

### LINKED ISSUE #807: Significant change in Resource's map/flatMap type with regards to variance
The type of the methods `map` and `flatMap` in `Resource` has changed from version `2.0.0` to `2.1.0`.
Before, the types looked like this:
```scala
def map[B](f: A => B)...: Resource[F, B] = ???
def flatMap[B](f: A => Resource[F, B]): Resource[F, B] = ???
```

Now, the types looks like this:
```scala
def map[G[x] >: F[x], B](f: A => B)...: Resource[G, B] = ???
def flatMap[G[x] >: F[x], B](f: A => Resource[G, B]): Resource[G, B] = ???
```

This change broke some code, which used used co-variance of the parameter of `F`, as in `F[+ _]`. For example:
```scala

class Base()
case class Derived() extends Base

case class Resource[F[_], A](use: F[A])(implicit F: Sync[F]) {
// the signatures before
//    def map[B](f: A => B)(implicit F: Functor[F]): Resource[F, B] = ???
//    def flatMap[B](f: A => Resource[F, B])(implicit F: Monad[F]): Resource[F, B] = ???
  def map[G[x] >: F[x], B](f: A => B)(implicit F: Functor[G]): Resource[G, B] = ???
  def flatMap[G[x] >: F[x], B](f: A => Resource[G, B])(implicit F: Monad[G]): Resource[G, B] = ???
}

def fBase[F[+ _]](implicit F: Sync[F]): F[Either[Base, Nothing]] = F.delay(Left(new Base()))
def fDerived[F[+ _]](implicit F: Sync[F]): F[Either[Derived, Nothing]] = F.delay(Left(Derived()))

def f[F[+ _]](implicit F: Sync[F]): F[Either[Base, Nothing]] =
  if (true) {
    fBase
  } else {
//      fDerived.map(identity[Either[Base, Nothing]]) // without F[+ _], one has to resort to hacks like this
    fDerived
  }

def g[F[+ _]](implicit F: Sync[F]): Resource[F, Either[Base, Nothing]] = {
  for {
    x <- Resource(f)
  } yield x
}
```

This now fails with
```
!I F (Could not find an instance of Functor for [x]F[x]): Functor[F[?]]
```

I tried to fiddle with the types a bit to find out if I could fix it, but didn't get anywhere. Do you guys have any idea how to change either the program, or the signature of `Resource`'s methods (while still keeping the benefits of the move to `G[x] >: F[x]`, which I presume is widening).
Thanks! :pray:

djspiewak:
Would also appreciate a look from @neko-kai. I think this works around the issue. It's a little annoying, but I wasn't able to find anything more elegant, and the nice thing is that it doesn't touch binary compatibility. I *believe* it should also be fully source-compatible with all existing call-sites that weren't already broken.

Fixes #807 (I think)

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Resource.scala
+++ b/core/shared/src/main/scala/cats/effect/Resource.scala
@@ -138,7 +138,7 @@ sealed abstract class Resource[+F[_], +A] {
    * @param f the function to apply to the allocated resource
    * @return the result of applying [F] to
    */
-  def use[G[x] >: F[x], B](f: A => G[B])(implicit F: Bracket[G, Throwable]): G[B] =
+  def use[G[x] >: F[x], B](f: A => G[B])(implicit F: Bracket[G[?], Throwable]): G[B] =
     fold[G, B](f, identity)
 
   /**
@@ -169,8 +169,8 @@ sealed abstract class Resource[+F[_], +A] {
    *
    **/
   def parZip[G[x] >: F[x]: Sync: Parallel, B](
-    that: Resource[G, B]
-  ): Resource[G, (A, B)] = {
+    that: Resource[G[?], B]
+  ): Resource[G[?], (A, B)] = {
     type Update = (G[Unit] => G[Unit]) => G[Unit]
 
     def allocate[C](r: Resource[G, C], storeFinalizer: Update): G[C] =
@@ -192,7 +192,7 @@ sealed abstract class Resource[+F[_], +A] {
    * Implementation for the `flatMap` operation, as described via the
    * `cats.Monad` type class.
    */
-  def flatMap[G[x] >: F[x], B](f: A => Resource[G, B]): Resource[G, B] =
+  def flatMap[G[x] >: F[x], B](f: A => Resource[G[?], B]): Resource[G[?], B] =
     Bind(this, f)
 
   /**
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -169,8 +169,8 @@ sealed abstract class Resource[+F[_], +A] {
    *
    **/
   def parZip[G[x] >: F[x]: Sync: Parallel, B](
-    that: Resource[G, B]
-  ): Resource[G, (A, B)] = {
+    that: Resource[G[?], B]
```
djspiewak, line 172: I'm not sure this one is necessary, btw.

neko-kai:
Seems like it works 🤷

SystemFw:
Could you explain what it does, I'm slightly at a loss :)

djspiewak:
@SystemFw It seems like, when it comes to variant constructors, `F` and `[x]F[x]` are not *entirely* equal inside of scalac. So what my workaround does is I replaced `F` with `[x]F[x]` in the type signatures. This is invisible 99.9% of the time, but in this case it appears to entirely fix the inference.

SystemFw:
Weird. Should we use the new kind projector syntax with *?

djspiewak:
@SystemFw I was going to implement that change across the whole project soon. For now, I just went with what was already there.

SystemFw:
Ah! Makes sense

PR #852, opened 2020-04-30
add IO[Future[A]]#liftFutureIO syntax

enzief:
Nice little syntax to do `IO(mkFuture).liftFutureIO` instead of `IO.fromFuture(IO(mkFuture))`, because there are too many parentheses in the latter.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -1627,4 +1627,11 @@ object IO extends IOInstances {
     override def recover(e: Throwable) =
       Pure(Left(e))
   }
+
+  /**
+   * Nice little syntax to do `IO(mkFuture).liftFutureIO` instead of `IO.fromFuture(IO(mkFuture))`.
+   */
+  implicit final class FutureOps[A](private val iof: IO[Future[A]]) extends AnyVal {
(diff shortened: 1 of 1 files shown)


alexandru:
IMO `IO.fromFuture(IO(mkFuture))` is more intuitive and discoverable than `IO(mkFuture).liftFutureIO`. This is because it is consistent with `fromTry` and `fromEither`. 

Also method syntax is useful when it can fit in a chain of transformations on some source value and I don`t see where `liftFutureIO` fits, because I don't see a case for a `IO[Future[A]]` being produced by anything other than `IO.apply`.

Therefore I'm personally against it. We do use OOP, some of us love it and its syntax, but sometimes simple functions are totally OK.

alexandru:
Btw, it's awesome you're attempting a contribution 🤗

enzief:
I guess it depends on personal taste. I usually chain `IO` methods following `IO.fromFuture()` (, `.void` for eg.). It's more readable to me personally instead of jumping my eyes backward from inner parentheses. Also `.liftFutureIO` makes you follow the logical steps from left to right, which makes it it's easier since I don't have to explain to new comers why there is `IO` inside `IO.fromFuture`.

rossabaker:
I agree with @alexandru's reasoning, but I'm a LISP fan.  Excess parens don't frighten me!

If we were to proceed with something like this, I'd consider generalizing it to syntax on anything `Async`.

enzief:
I don't want more imports so I think we're not proceeding. Thanks for the review.

PR #853, opened 2020-05-02
WIP: CE3 Resource with Region and tests

kubukoz:


src/test/scala/ce3/laws/ResourceSpec.scala
```diff
@@ -0,0 +1,78 @@
...
+        Resource,
+        PureConc[Int, *],
+        Outcome[PureConc[Int, *], Int, *],
+        Int
+      ]
+
+      val lawz: testz.laws.type =
+        testz.laws
+
+      implicit val cogenLawsCase: Cogen[lawz.F.Case[_]] =
```
kubukoz, line 45: @djspiewak can you give me a hand with this instance? I'm not sure why it has to have this placeholder... as well as whether it's possible to get this instance to typecheck without casting or some extra hackery.

src/test/scala/ce3/laws/ResourceSpec.scala
```diff
@@ -0,0 +1,78 @@
...
+        s"${result.show} !== ${expect.show}"
+      )
+
+  implicit def arbResource[F[_], E, A]: Arbitrary[Resource[F, A]] = ???
+  implicit def cogenResource[F[_], E, A]: Cogen[Resource[F, A]] = ???
+
+  implicit def eqResource[F[_]: Bracket[*[_], E], E, A](
+      implicit eqFUnit: Eq[F[Unit]]
+  ): Eq[Resource[F, A]] =
+    Eq.by(_.used) //todo pass real function?
```
kubukoz, line 76: I'm looking into using `ExhaustiveCheck[A => F[B]]` for some `B` for this.

src/main/scala/ce3/Resource.scala
```diff
@@ -0,0 +1,294 @@
...
+  //     acquire.map(a => (a, (e: ExitCase[Throwable]) => release(a, e)))
+  //   )
+
+  /**
+    * Lifts a pure value into a resource. The resource has a no-op release.
+    *
+    * @param a the value to lift into a resource
+    */
+  def pure[F[_], A, E](a: A)(implicit F: Bracket[F, E]): Resource[F, A] =
+    Allocate[F, F.Case, A](F.pure((a, (_: F.Case[Any]) => F.unit)))
```
kubukoz: I really didn't want to have a Bracket here but I'd need a new node in the ADT. I had one (LiftF) but it proved to replicate a reasonable amount of functionality from `Allocate` so I kept if away for now. Everything using Resource will always have a bracketful effect anyway, so maybe it's not a huge problem.

src/test/scala/ce3/laws/ResourceSpec.scala
kubukoz: I figured I could narrow a `Cogen[Case[Unit]]` to `Cogen[Case[_]]` by `.contramap(_.void)`, but I'm not too comfortable with that Unit.

src/test/scala/ce3/laws/RegionTests.scala
```diff
@@ -56,35 +56,40 @@ trait RegionTests[R[_[_], _], F[_], E] extends MonadErrorTests[R[F, ?], E] {
...
-      : RegionTests[R, F, E] =
+      B: Bracket[F, E] { type Case[A] = Case0[A] }
+  ): RegionTests[R, F, E] {
+    val laws: RegionLaws[R, F, E] {
+      val F: F0.type
+    }
+  } =
     new RegionTests[R, F, E] {
-      val laws = RegionLaws[R, F, Case0, E]
+      val laws: RegionLaws[R, F, E] {
```
kubukoz, line 91: I had to make the types more specific so that the compiler could see that `tests.laws.F.Case` is indeed `Outcome` (when F=PureConc).

src/test/scala/ce3/laws/ResourceSpec.scala
```diff
@@ -0,0 +1,145 @@
...
+  // Completed(Some(hello))
+  // │
+  // ├ Notify 0x6404F418
+  // ├ Notify 0x6404F418
+  // ├ Notify 0x6404F418
+  // ├ Notify 0x6404F418
+  // ├ Notify 0x6404F418
+  // ├ Notify 0x2794EAB6
+  // ╰ Pure Completed(hello)
+  //todo: shouldn't this raise an exception in the background? How does PureConc handle failures in finalizers?
```
kubukoz, line 141: I can see the law specifies that errors in the release are ignored, so how are we going to do that for IO? Shouldn't we at least report the error?

kubukoz: @djspiewak

src/test/scala/ce3/laws/ResourceSpec.scala
kubukoz: Okay, apparently the wildcard messes up implicit resolution in ways I've never seen before, so I had to add a bunch of explicit parameters for now. The tests still run but it's super annoying and makes everything that much dirtier.

src/test/scala/ce3/laws/ResourceSpec.scala
kubukoz: Yep, it's way too slow that way. It'd be nice to at least have `ExhaustiveCheck[F[Unit]]` though, so that we check failures and self-cancelation in the function passed to `use`.

src/test/scala/ce3/laws/ResourceSpec.scala
djspiewak: IMO real implementors (such as `IO`) should probably report ignored errors on a side-channel. `IO` does this today, in fact (IIRC it logs).

src/main/scala/ce3/Resource.scala
```diff
@@ -0,0 +1,320 @@
...
+    *      with its finalizer tuple in two parameters
+    *
+    * @tparam F the effect type in which the resource is acquired and released
+    * @tparam A the type of the resource
+    * @param resource an effect that returns a tuple of a resource and
+    *        an effectful function to release it
+    */
+  def applyCase[F[_], Case[_], E, A](
+      resource: F[(A, Case[_] => F[Unit])]
+  )(implicit F: Bracket.Aux[F, E, Case]): Resource[F, A] =
```
djspiewak, line 113: Ah it's interesting that this ends up being polymorphic in `Case`. I had originally assumed that `Resource` would hard-code to `Outcome`, but it makes some sense that it doesn't have to, since it's just a free `Region`.

src/main/scala/ce3/Resource.scala
```diff
@@ -0,0 +1,320 @@
...
+    * @tparam F the effect type in which the resource is acquired and released
+    * @tparam A the type of the resource
+    * @param acquire a function to effectfully acquire a resource
+    * @param release a function to effectfully release the resource returned by `acquire`
+    */
+  def makeCase[F[_], Case[_], E, A](
+      acquire: F[A]
+  )(
+      release: (A, Case[_]) => F[Unit]
+  )(implicit F: Bracket.Aux[F, E, Case]): Resource[F, A] =
```
djspiewak, line 164: Something I want to be pretty sensitive to as we start building out more and more things on top of the CE3 typeclasses is how the ergonomics of *using* the classes actually feels. What I mean by that is, how annoying is it to deal with the case polymorphism of `Bracket`? Or the fact that `Concurrent` doesn't pick a resource management strategy? Etc etc

src/test/scala/ce3/laws/Generators.scala
```diff
@@ -26,14 +26,56 @@ object Generators {
 
   private[this] def F[E] = ConcurrentBracket[PureConc[E, ?], E]
 
-  def genPureConc[E: Arbitrary: Cogen, A: Arbitrary: Cogen](depth: Int): Gen[PureConc[E, A]] = {
+  def genResource[F[_]: Bracket.Aux[*[_], E, Case], Case[_], E, A: Arbitrary: Cogen](
```
djspiewak, line 29: If we merge this together with your "generic generators" PR, do some of these generators just go away?

src/test/scala/ce3/laws/Generators.scala
```diff
@@ -26,14 +26,56 @@ object Generators {
...
+      : Gen[Resource[F, A]] =
+    Arbitrary.arbitrary[A].map(Resource.pure[F, E, A](_))
+
+  def genSuspendResource[F[_]: Bracket[*[_], E], E, A](
+      implicit arbEffect: Arbitrary[F[Resource[F, A]]]
+  ): Gen[Resource[F, A]] =
+    arbEffect.arbitrary.map(Resource.suspend)
+
+  def genApplyResource[F[_]: Bracket.Aux[*[_], E, Case], Case[_], E, A](
+      implicit arbFA: Arbitrary[F[(A, Case[_] => F[Unit])]]
```
djspiewak, line 55: It's actually probably better to use the `implicit` syntax for the `Bracket.Aux` instance here, rather than using `: `. The reason being that the type inference on `Case` is going to be sensitive to the parameter order within the `implicit` block, so it's better to be explicit about it.

src/test/scala/ce3/laws/RegionTests.scala
djspiewak: Makes sense. I had to do something similar when working on `BracketLaws`.

src/test/scala/ce3/laws/ResourceSpec.scala
```diff
@@ -0,0 +1,176 @@
...
+          arbPureConc[Int, (A, Outcome[F, Int, _] => F[Unit])](
+            i[Arbitrary[Int]],
+            Cogen[Int],
+            i[Arbitrary[(A, Outcome[F, Int, _] => F[Unit])]],
+            Cogen.tuple2[A, Outcome[F, Int, _] => F[Unit]](
+              Cogen[A],
+              cogenFinalizer[A]
+            )
+          )
+        )
```
djspiewak, line 91: Ouch. I wonder why it is that this doesn't infer nicely?

src/main/scala/ce3/Resource.scala
djspiewak: Yeah I think it's fine. Like, in theory it would be kinda nice to *not* require `Bracket` here, but it shouldn't make too much of a practical difference. If we feel strongly about it, we can smooth it out later.

src/test/scala/ce3/laws/ResourceSpec.scala
djspiewak: Yeah I'd like to improve the exhaustiveness of some of these tests. There are certainly a few areas where it's a little wimpy.

src/test/scala/ce3/laws/ResourceSpec.scala
djspiewak: > I figured I could narrow a Cogen[Case[Unit]] to Cogen[Case[_]] by .contramap(_.void), but I'm not too comfortable with that Unit.

So this is actually sound, just like the types imply it is, because the wildcard guarantees that no one can use the `A` from inside the `Case[A]`, meaning that it can be unit and you're fine.

Wildcards though… do… tend to mess things up. If that's really what the blocker is then we could use skolems, but I think we probably don't need it in this case. What instance exactly is failing to resolve in the current formulation?

kubukoz: The arbitrary/cogen instances for the finalizers in the `makeCase` / `allocatedCase` constructors, I think. Pretty much the stuff that's passed manually now...

src/main/scala/ce3/Resource.scala
kubukoz: I think it'd be really, really nice if we could require Bracket when constructing resources, but avoid it in `use`. If you're building a resourc e, chances are you already have Sync (not all the time but most of the time - in the CE3 hierarchy). If you're just using one, you might not have that (e.g. you have a Session resource from skunk)

src/test/scala/ce3/laws/ResourceSpec.scala
kubukoz: It does! So we want to keep that out of the specification, right?

Also, I feel like `bracket(noop)(use)(_ => raise) <-> fa` prohibits the implementation from doing anything else. Unless we remove the side channel from the definition of equality, which I'm probably fine with...

kubukoz: this probably won't have an effect on Resource's implementation though. It's just some work done by Bracket.

src/main/scala/ce3/Resource.scala
kubukoz: Right, for Concurrent brackets it'll be Outcome, but given that the new Bracket doesn't require cancelability, Case might just be Either for types that only go up to Bracket or Sync (with no Concurrent).

src/main/scala/ce3/Resource.scala
kubukoz: WDYM about strategy?

src/test/scala/ce3/laws/Generators.scala
kubukoz: Yes!

src/test/scala/ce3/laws/Generators.scala
kubukoz: Agreed

src/test/scala/ce3/laws/ResourceSpec.scala
kubukoz: I blame `Outcome[F, Int, _] => F[Unit]`. Also, I still need to `delay` the arbitrary of `F[RF[A]]`, because it's going to depend on the enclosing def. There's a deeper discussion...

I think we should consider getting a new type of the sort `type DeriveArb[F[_], A] = Arbitrary[A] => Arbitrary[F[A]]`, and require a single `DeriveArb` with arbitraries of the `A`s. In signatures like `RegionTests#region`, which look like this:

```scala
def region[A: Arbitrary: Eq, B: Arbitrary: Eq, C: Arbitrary: Eq](
    implicit
      ArbRFA: Arbitrary[R[F, A]],
      ArbFA: Arbitrary[F[A]],
      ArbRFB: Arbitrary[R[F, B]],
      ArbFB: Arbitrary[F[B]],
      ArbRFC: Arbitrary[R[F, C]],
      ArbFC: Arbitrary[F[C]],
      ArbRFU: Arbitrary[R[F, Unit]],
      ArbFU: Arbitrary[F[Unit]],
      ArbRFAtoB: Arbitrary[R[F, A => B]],
      ArbRFBtoC: Arbitrary[R[F, B => C]],
      ArbE: Arbitrary[E],
      CogenA: Cogen[A],
      CogenB: Cogen[B],
      CogenC: Cogen[C],
      CogenE: Cogen[E],
      CogenCase: Cogen[laws.F.Case[_]],
      EqRFA: Eq[R[F, A]],
      EqRFB: Eq[R[F, B]],
      EqRFC: Eq[R[F, C]],
      EqE: Eq[E],
      EqRFEitherEU: Eq[R[F, Either[E, Unit]]],
      EqRFEitherEA: Eq[R[F, Either[E, A]]],
      EqRFABC: Eq[R[F, (A, B, C)]],
      EqRFInt: Eq[R[F, Int]],
      EqRFUnit: Eq[R[F, Unit]],
      iso: Isomorphisms[R[F, ?]])
      : RuleSet
```

we could get rid of many parameters, and have better errors if any are missing (because fewer derivations will be involved). But as I said, it's a long discussion :)

src/test/scala/ce3/laws/ResourceSpec.scala
djspiewak: Hmm. I'll check this out and try to tease apart what's happening. You're probably right that it's the wildcard's fault.

src/main/scala/ce3/Resource.scala
djspiewak: > I think it'd be really, really nice if we could require Bracket when constructing resources, but avoid it in use.

I actually go the other way: I'd rather require it in `use` and *not* require it during construction. :-) Idiomatically, we tend to prefer to avoid code that captures instances and holds them for later utilization unless absolutely necessary. As an example, if you think about a "Cats-style" red/black tree, you wouldn't want `Tree.empty[A]` to require an `Order[A]`, because… you don't need it!

On a more practical note, something with `Resource` that would be useful is constructing a resource with an `F` that doesn't form a `Bracket` (conveniently), then later on `mapK`ing it to some `G` which forms a `Bracket`, then calling `use`. That kind of thing is made impossible if you capture the `Bracket` early. So just as a best-practice, deferring the capture as "late" as possible is ideal.

src/test/scala/ce3/laws/ResourceSpec.scala
djspiewak: > It does! So we want to keep that out of the specification, right?

Yep! We can't really specify side-channels anyway, because… they're side. I want to leave implementors free to do whatever they want here.

> Also, I feel like bracket(noop)(use)(_ => raise) <-> fa prohibits the implementation from doing anything else. Unless we remove the side channel from the definition of equality, which I'm probably fine with...

I think the side-channel is just outside the definition of equality.

src/main/scala/ce3/Resource.scala
djspiewak: Very very interesting!

src/main/scala/ce3/Resource.scala
djspiewak: Oh just the fact that `Concurrent` doesn't inherit from either `Bracket` or `Region`, and instead you have to do `Concurrent with Bracket` (or use a type alias) whenever you want a *specific* specialization. Just want to make it clear that the jury is still out on whether or not that's a good idea, and if it's painful for us to use any of these encodings, it'll definitely be painful for users.

src/test/scala/ce3/laws/ResourceSpec.scala
djspiewak: Yeah I think that might be a good idea. The error messages are getting insane on some of this stuff.

src/main/scala/ce3/Resource.scala
kubukoz: `you wouldn't want Tree.empty[A] to require an Order[A], because… you don't need it!`

Sure... until you try to combine two trees with different instances :) 

Are there that many situations in which you won't have Bracket? Would it be a stretch to say "if you want resource safety, you need Bracket" and require an instance when building these resources?

BTW. I'm wondering if we could abuse variance a bit here and use `Any` instead of `F.Case` - then we'd only need Applicative. I'll see if I can make Allocate contravariant in the case...

djspiewak: > Sure... until you try to combine two trees with different instances :)

Indeed, if you have multiple `Order[A]` in scope, then `Tree[A]` starts to become a bit insane. I have an old scratch project from ages ago where I worked on resolving that problem, but in the end I decided to just tilt other windmills. :-) Notably, it's a problem regardless of whether you capture the `Order[A]` in `Tree`'s constructor or in `empty`, it's just that you can sometimes detect the situation (usually by using instance equality) when you capture it earlier.

> Are there that many situations in which you won't have Bracket? Would it be a stretch to say "if you want resource safety, you need Bracket" and require an instance when building these resources?

Not sure. I'm just thinking more in terms of generality. It makes the API more flexible by allowing some valid scenarios that wouldn't be allowed otherwise.

> BTW. I'm wondering if we could abuse variance a bit here and use Any instead of F.Case - then we'd only need Applicative. I'll see if I can make Allocate contravariant in the case...

Hmm! I'm curious if that would work…

src/main/scala/ce3/Resource.scala
kubukoz: Yeah, we'll see how it all turns out. Most likely when we have an IO and can start porting small projects built on polymorphic effects...

kubukoz:
I hope rebases on draft PRs are fine...

src/main/scala/ce3/Resource.scala
kubukoz:
```
final case class Allocate[F[_], -Case[_], A](
      resource: F[(A, Case[_] => F[Unit])]
  ) extends Resource[F, A]
```

Apparently this won't compile because Case is in an invariant position (because of the `_`). Changing `_` to a concrete type fixes that, as well as `-Case[-_]` and `-Case[+_]` do. I'll see which one hurts the least.

kubukoz: oops, that was just Scala lying to me.

kubukoz: I realized Case isn't part of the type Allocate is extending, so its variance didn't matter - I used Any and it just works.

kubukoz:
Apparently `region.liftF is open unit` is now failing... but I don't really see why yet. I might need some help...

kubukoz:
oh no, the tests seem to work now, even though I didn't change a thing...

kubukoz:
OK, failing seed: `nJYWTl2F1wRo5eCaadsilZxn0_v1_WWUEQ8ksDQeJ7P=`

kubukoz:
```
Failed:
- ce3.laws.ResourceSpec:
  * region.liftF is open unit - Falsified after 0 passed tests.
> Labels of failing property:Expected: Completed(None)
│
├ Notify 0x6AEDC5FE
├ Notify 0x62EA87A2
├ Fork
│ ╰╮
│  ├ Notify 0x62EA87A2
│  ├ Notify 0x62EA87A2
│  ├ Notify 0x7F23F371
│  ├ Fork
│  │ ╰╮
│  │  ├ Cede
│  │  ├ Notify 0x7BE51B5E
│  │  ╰ Done
│  │
│  ├ Cede
│  ├ Notify 0x62EA87A2
│  ├ Notify 0x62EA87A2
│  ├ Notify 0x62EA87A2
│  ├ Notify 0x62EA87A2
│  ├ Notify 0x4F51ABCA
│  ├ Fork
│  │ ╰╮
│  │  ├ Notify 0x277C8ED
│  │  ╰ Done
│  │
│  ├ Notify 0x62EA87A2
│  ├ Notify 0x62EA87A2
│  ├ Notify 0x14745109
│  ├ Notify 0x67C019E1
│  ╰ Done
│
├ Notify 0x33224451
├ Fork
│ ╰╮
│  ├ Notify 0x33224451
│  ├ Notify 0x33224451
│  ├ Notify 0x3AC2A239
│  ├ Fork
│  │ ╰╮
│  │  ├ Notify 0x5BC4401D
│  │  ╰ Done
│  │
│  ├ Notify 0x7C21BE66
│  ├ Fork
│  │ ╰╮
│  │  ├ Notify 0x53B1C029
│  │  ├ Fork
│  │  │ ╰╮
│  │  │  ├ Notify 0x191E4193
│  │  │  ├ Fork
│  │  │  │ ╰╮
│  │  │  │  ├ Notify 0xBFEF35F
│  │  │  │  ╰ Done
│  │  │  │
│  │  │  ├ Notify 0x60B9661E
│  │  │  ╰ Done
│  │  │
│  │  ├ Notify 0x5473AB5D
│  │  ╰ Done
│  │
│  ├ Notify 0x71FED860
│  ├ Notify 0x4E8BBE49
│  ├ Fork
│  │ ╰╮
│  │  ├ Notify 0x4E8BBE49
│  │  ├ Notify 0x4E8BBE49
│  │  ├ Notify 0x75342106
│  │  ├ Cede
│  │  ├ Notify 0x3F496FB5
│  │  ├ Fork
│  │  │ ╰╮
│  │  │  ├ Notify 0x3F496FB5
│  │  │  ├ Notify 0x3F496FB5
│  │  │  ├ Notify 0x6FFC29E4
│  │  │  ├ Notify 0x3F496FB5
│  │  │  ├ Notify 0x3F496FB5
│  │  │  ├ Notify 0x288046EA
│  │  │  ╰ Done
│  │  │
│  │  ├ Notify 0x64C77574
│  │  ├ Fork
│  │  │ ╰╮
│  │  │  ├ Notify 0x64C77574
│  │  │  ├ Notify 0x64C77574
│  │  │  ├ Notify 0x64C77574
│  │  │  ├ Notify 0x64C77574
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC
│  │  │  ├ Await 0x2D6040FC

...

...

...

...
Received: Errored(2147483647)
│
├ Notify 0x736FB394
├ Notify 0x759AFD9
├ Fork
│ ╰╮
│  ├ Notify 0x759AFD9
│  ├ Notify 0x759AFD9
│  ├ Notify 0x235EE1F4
│  ├ Fork
│  │ ╰╮
│  │  ├ Cede
│  │  ├ Notify 0x148AEC49
│  │  ╰ Done
│  │
│  ├ Cede
│  ├ Notify 0x759AFD9
│  ├ Notify 0x759AFD9
│  ├ Notify 0x759AFD9
│  ├ Notify 0x759AFD9
│  ├ Notify 0x19BF9AFA
│  ├ Fork
│  │ ╰╮
│  │  ├ Notify 0x4DFF8D7
│  │  ╰ Done
│  │
│  ├ Notify 0x759AFD9
│  ├ Notify 0x759AFD9
│  ├ Notify 0x77CD7CE7
│  ├ Notify 0x209BE22E
│  ╰ Done
│
├ Notify 0x1CDFBAF2
├ Fork
│ ╰╮
│  ├ Notify 0x1CDFBAF2
│  ├ Notify 0x1CDFBAF2
│  ├ Notify 0x16ACBC04
│  ├ Fork
│  │ ╰╮
│  │  ├ Notify 0x39406864
│  │  ╰ Done
│  │
│  ├ Notify 0x6ECE88C8
│  ├ Fork
│  │ ╰╮
│  │  ├ Notify 0x46C16C92
│  │  ├ Fork
│  │  │ ╰╮
│  │  │  ├ Notify 0x4B6ABBB0
│  │  │  ├ Fork
│  │  │  │ ╰╮
│  │  │  │  ├ Notify 0x3B06EC6F
│  │  │  │  ╰ Done
│  │  │  │
│  │  │  ├ Notify 0x33CD0FAE
│  │  │  ╰ Done
│  │  │
│  │  ├ Notify 0x3C979534
│  │  ╰ Done
│  │
│  ├ Notify 0x6EA54599
│  ├ Notify 0x1CDFBAF2
│  ├ Notify 0x2D0ED2FF
│  ├ Notify 0x6C58F40C
│  ├ Await 0x1CDFBAF2
│  ├ Await 0x1CDFBAF2
│  ├ Await 0x1CDFBAF2
│  ├ Await 0x1CDFBAF2
│  ├ Await 0x1CDFBAF2
│  ├ Await 0x1CDFBAF2
│  ├ Await 0x1CDFBAF2
│  ├ Await 0x1CDFBAF2
│  ├ Await 0x1CDFBAF2
│  ├ Await 0x1CDFBAF2
│  ├ Await 0x1CDFBAF2
│  ├ Await 0x1CDFBAF2
│  ├ Await 0x1CDFBAF2
│  ├ Await 0x1CDFBAF2
│  ├ Await 0x1CDFBAF2
│  ├ Await 0x1CDFBAF2
│  ├ Await 0x1CDFBAF2
│  ├ Await 0x1CDFBAF2
│  ├ Await 0x1CDFBAF2
│  ├ Await 0x1CDFBAF2
│  ├ Await 0x1CDFBAF2
│  ├ Await 0x1CDFBAF2

...

...> ARG_0: Kleisli(cats.data.Kleisli$$$Lambda$842/0x000000084044e840@20ff4ab5)
The seed is WqZK2Kv-5ENgEGysFFpWU0RjtapUrpUF9CuVsJ2ColJ=
```

djspiewak:
Sorry for not getting onto this. I didn't miss it. :-) Should be able to look at it shortly. Diagnosing these things is… really annoying.

djspiewak:
As a quick thing: if you add a `F[A] => Pretty` implicit to `RegionTests`, you should be able to get `ARG_0` to pretty print. In the meantime, this is really strange. `Completed(None)` means things deadlocked, but I can't see any reason why that would be happening here.

kubukoz:
I imagine there's a `never` in an uncancelable part somewhere, but didn't look closely yet... the suspicious thing is the (apparently never-ending) sequence of awaits on the same monitor.

Yeah, I added some pretties locally and saw some more

djspiewak:
> I imagine there's a never in an uncancelable part somewhere, but didn't look closely yet... the suspicious thing is the (apparently never-ending) sequence of awaits on the same monitor.

Oh you know what?

```scala
val fa = F.canceled >> F.raiseError(e)
```

That would do it. The difference here really is the cancelability. The law needs to be revised. This is actually quite tricky because, in `Region`/`Bracket`, we have no way to talk about cancelability. I need to give that some thought. If you're blocked on this, I would just comment out the law for now.

kubukoz:
Yeah, I can comment it out for now.

I probably can't do much more than the current implementation has (there's still a couple things I can add I think, like `allocated`), as we don't have Ref - that rules out `parZip`. Probably also applies to Concurrent. Now... it would be nice if we could have `Concurrent[Resource[F` without `Sync[F]`, but Ref implies Sync...

djspiewak:
> I probably can't do much more than the current implementation has (there's still a couple things I can add I think, like allocated), as we don't have Ref - that rules out parZip. Probably also applies to Concurrent. Now... it would be nice if we could have Concurrent[Resource[F without Sync[F], but Ref implies Sync...

Do we *need* `Ref` though? That's a big part of the experiment here: can we implement `Concurrent[Resource[F, ?], E]` for any `F` such that `ConcurrentBracket[F, E]`?

kubukoz:
I'll see, but I'd expect to have the same kind of problems (except with more of the problems) as Fabio did in #742...

djspiewak:
> I'll see, but I'd expect to have the same kind of problems (except with more of the problems) as Fabio did in #742...

That would be very, very interesting.

`Resource` is pretty close to just a free `Region`, so if it can't form lawful compositions of equivalent power, then we may be on the wrong track with `Region` in general.

djspiewak:
Does this help as a starting point? I didn't even typecheck anything, I'm just throwing some code at hte wall:

```scala
def start[A](rfa: Resource[F, A]): Resource[F, Fiber[Resource[F, ?], E, A]] = {
  val delegateF = F start {
    F.bracketCase(rfa.allocated)(x => x) {
      case ((_, fin), Outcome.Canceled | Outcome.Errored(_)) => fin
      case (_, Outcome.Completed(_)) => F.unit
    }
  }

  Resource.liftF(delegateF) map { delegate =>
    new Fiber[Resource[F, ?], E, A] {

      val cancel = Resource.liftF(delegate.cancel)

      val join =
        Resource liftF {
          delegate.join map {
            case oc @ Outcome.Canceled | Outcome.Errored(_) => oc
            case Outcome.Completed(allocF) => Outcome.Completed(Resource(allocF))
          }
        }
    }
  }
}
```

kubukoz:
It'll surely be useful! I'll give it a closer look.

SystemFw:
> F.bracketCase(rfa.allocated)(x => x)

You need to sort out cancellation for this one

djspiewak:
> You may need to sort out cancellation for this one, depending on the specific semantics of bracketCase in this branch

Oh you mean because it will make all the `liftF`s uncancelable? I wonder if we need an `allocatedPoll` which can take a `F ~> F` and weave it appropriately into the interpreter?

SystemFw:
`bracketCase(rfa.allocated)`

if `bracketCase` is behaving like in CE2 here, it will make the resource uncancelable yeah

djspiewak:
> if bracketCase is behaving like in CE2 here, it will make the resource uncancelable yeah

It is. So yeah, we need an `allocatedPoll` then. I *think* that will resolve it. Because `bracketMask` isn't difficult to implement and would give us access to the underlying `uncancelable`, which in turn could be passed along to be used when interpreting `liftF` nodes.

SystemFw:
> So yeah, we need an allocatedPoll then.

yeah, I would be curious to see how that looks like. In CE2 I used Ref to accumulate finalisers so that I didn't have to use `bracket` at all, it would be good to compare

SystemFw:
I get where you're going to though :)

PR #854, opened 2020-05-02
Fiber tracing: asynchronous stack tracing

RaasAhsan:
This PR contains a first pass for fiber tracing. Fiber tracing refers to the analysis of the execution graph produced by an `IO` program. The scope of this PR is to capture stack frames at well-defined points in execution and collect them into trace frames, which can later be printed out on demand. This is useful in both development and production settings where asynchronous programming is concerned because fiber traces are much more useful for debugging and easier to understand than stack traces. In the future, fiber tracing can be expanded to cover additional use cases.

I make several claims about performance here, which are mostly backed by local runs of `MapCallsBenchmark`. This offers soft upper bound on performance impact, but I need to run the other benchmarks and create some more comprehensive test suites so others can run and compare.

## Design constraints
There are several design constraints I wrote this implementation under, which forced some tradeoffs. We should probably clarify how important these are and whether we can relax them at all.
1. If tracing is globally disabled, there should be negligible impact to application performance. The current implementation sees a 0% performance degradation; the JIT can dynamically eliminate all consequent code paths at runtime. But it comes at a steep cost when tracing is enabled.
2. There should be optimal support for lexically scoped tracing i.e. tracing only a subtree of an `IO` program. This makes tracing possible and appealing in production. The current implementation uses `ThreadLocal`. This has no impact when tracing is turned off globally, but if it is, every code path incurs the cost of `ThreadLocal` accesses. @djspiewak has some interesting ideas here that might be able to exploit the memory model to achieve thread-local state with non-volatile variables.
3. Support different tracing modes. A fast tracing mode would build trace frames, produce surface-level detail from stack trace, and cache it all so that subsequent calls complete faster. A slower tracing mode would perform more detailed analysis. I haven't done too much work here yet but I imagine there's a lot possible.

## Limitations
Additionally, there were some limitations in the Scala language that forced me to make some code design decisions I'm not very happy about.
1. In Scala 2.12, [singleton object methods cannot be elided by the JIT](https://github.com/scala/scala-dev/issues/537). This was fixed in Scala 2.13.
2. In general, Scala cross-module singleton variable accesses cost a volatile read. I don't have a primary source for this but if you inspect the bytecode you will see it. The JIT can't optimize this out. 

## Implementation details
Several implementation details are described below. Hopefully we can have discussions around what the best approach is moving forward.
1. Global flag defined in Java scope. After class initialization, the HotSpot C2 compiler is aware of the values of all `static final` variables (or so called just-in-time constants). With that information, it can eliminate code paths that are consequent to those constants. As described above, Scala object-level fields can't take advantage of this, so we're defining the global tracing flag in Java instead. One issue here is package visibility.
2. Global flag conditionals. Originally I wanted to hide the conditionals behind an `IO.Builder` helper and `IOTracing` utility methods, but C2 can't inline those method calls in 2.12. So the conditionals are somewhat scattered around the codebase.
3. Trace frame capture. There were two approaches I identified here: either capture the trace frames as instance variables in each `IO` node, or introduce a new `IO` constructor that reifies an `IO` node with a trace frame. I took the latter approach because there's virtually no cost when tracing is disabled, and it doesn't pollute the constructor or interpreter logic for other `IO` instructions. However, there is a very steep cost (more than 30-40% IIRC) when tracing IS enabled because it effectively doubles the number of interior nodes in the `IO` tree being traced. However, the instance variable approach is much cheaper during tracing, but incurs around a 4-5% performance penalty when its disabled.
4. Trace frame cache. Trace frames are cached by lambda references as keys, so we can easily cache `Map`, `Bind`, and `Async` nodes.  I chose a global `ConcurrentHashMap` here, but we probably need to benchmark it more and compare it to alternatives like `ThreadLocal`. Also, this should probably be an LRU cache bounded in entry count. There is a snag with this caching approach where the same lambda reference could be used in multiple places:
```scala
val x: Int => Int = (x: Int) => x + 1
Option(3)
  .map(x)
  .map(x)
```
4. The Scala compiler won't generate a new lambda class for each invocation of `x`, but we still need some way to detect if we're looking at a lambda that is shared like this, otherwise caching could produce incorrect results. Maybe there's a way to compare the name of the lambda with reflection. Bytecode analysis is also another option here. But I personally haven't seen or used this pattern ever, so maybe we explicitly call it out as a limitation.
5. Async. The implementation of `IOBracket` leverages the `Async` constructor and invokes a new execution of the `IORunLoop`. Since two executions of the run-loop can then be associated with the same fiber, we need some way to move state between each execution. My approach is to create a thread-safe `IOContext` class that holds a buffer of traces and gets passed along in the `Async` constructor. Ideally it would be mutable (with an internal ring buffer) because there's a cost to the synchronization, but the alternative was to significantly refactor the run-loop. I believe some of the work done for CE3 will simplify the `bracket` implementation so only one run-loop is invoked per `IO`.
6. Trace frame calculation. The current implementation filters out specific class prefixes from stack frames and takes the first one remaining. Obviously this won't work with constructs like monad transformers, so we need to figure out a smarter algorithm here. Maybe taking sequences of frames that might be relevant. Or the user can specify a list of packages they're interested in. Also need to clean up the class/lambda names a little bit.
7. Lexically scoped tracing mode. `ThreadLocal` was the easy option here and the performance hit isn't terrible. Preserving `ThreadLocal` state across asynchronous boundaries is fairly straightforward: before invoking the continuation of an `Async`, save the current state. Restore that state when the continuation invokes the callback. Note that in the case of cancellation, we don't need to worry about resetting the `ThreadLocal` of that thread, because it will be reset when it starts a new run-loop or picks up a new task (which is re-entering the run loop from an asynchronous boundary).

## Questions
1. There are many constructor calls for `IO.Async` throughout the library. Should we trace them all?
2. What should trace frames look like?

## TODO
- [x] Frame cache: thread-local or a thread-safe global cache? LRU entry-count bounded?
- [ ] Proper trace frame calculation
- [x] Add appropriate tracing to all primitive operations
- [ ] Exception handling and `TracedException`
- [x] Shared lambda issue
- [x] Demangle class and anonfun names
- [ ] Post benchmarks and create a more comprehensive test suite others can run
- [x] Tests
- [ ] Documentation and comments
- [ ] Cleanup

## Example
```scala
def program2: IO[Unit] =
  for {
    _ <- IO.delay(println("7"))
    _ <- IO.delay(println("8"))
  } yield ()

def program: IO[Unit] =
  for {
    _ <- IO.delay(println("1"))
    _ <- IO.delay(println("2"))
    _ <- IO.shift
    _ <- IO.unit.bracket(_ =>
      IO.delay(println("3"))
        .flatMap(_ => program2)
    )(_ => IO.unit)
    _ <- IO.delay(println("4"))
    _ <- IO.delay(println("5"))
  } yield ()

override def run(args: List[String]): IO[ExitCode] =
  for {
    _ <- IO.suspend(program).rabbitTrace
    _ <- IO.delay("10")
    trace <- IO.backtrace
    _ <- IO.delay(trace.printTrace())
  } yield ExitCode.Success
```

produces as output...
```
1
2
3
7
8
4
5
IOTrace
	map at org.simpleapp.example.Example$.$anonfun$program$8 (Example.scala:42)
	bind at org.simpleapp.example.Example$.$anonfun$program$7 (Example.scala:41)
	map at org.simpleapp.example.Example$.$anonfun$program2$1 (Example.scala:29)
	bind at org.simpleapp.example.Example$.program2 (Example.scala:28)
	bind at org.simpleapp.example.Example$.$anonfun$program$4 (Example.scala:39)
	async at org.simpleapp.example.Example$.$anonfun$program$3 (Example.scala:40)
	bind at org.simpleapp.example.Example$.$anonfun$program$3 (Example.scala:37)
	bind at org.simpleapp.example.Example$.$anonfun$program$2 (Example.scala:36)
	bind at org.simpleapp.example.Example$.$anonfun$program$1 (Example.scala:35)
	bind at org.simpleapp.example.Example$.program (Example.scala:34)
```

This is still a work-in-progress, so please leave your feedback here :)


djspiewak:
This is amazing work! Even just the write-up is epic. I have lots of thoughts and I'll try to respond shortly, but just a quick hat-tip first. :-) Well done.

core/jvm/src/main/scala/cats/effect/internals/Example.scala
```diff
@@ -0,0 +1,53 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.simpleapp.example
+
+import cats.effect.{ExitCode, IO, IOApp}
+
+object Example extends IOApp {
```
LukaJCB, line 21: I love this example, but maybe it shouldn't be part of the `core` module?

kubukoz:
This looks awesome, looking forward to having this!

core/jvm/src/main/scala/cats/effect/internals/Example.scala
RaasAhsan: Yeah it will be removed at some point. Its a convenient place to put it to test locally and demonstrate usage

core/jvm/src/main/java/cats/effect/internals/TracingPlatformFast.java
```diff
@@ -0,0 +1,38 @@
...
+package cats.effect.internals;
+
+import java.util.Optional;
+
+/**
+ * Scala object field accesses cost a volatile read across modules.
+ * Since this flag is read during construction of IO nodes, we are opting to
+ * hold this flag in a Java class to bypass the volatile read.
+ */
+public class TracingPlatformFast {
```
djspiewak, line 26: `final`

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -100,17 +102,24 @@ sealed abstract class IO[+A] extends internals.IOBinaryCompat[A] {
...
+    if (tracingEnabled) {
+      // Don't perform map fusion when tracing is enabled.
+      // We may not actually have to do this
+      IOTracing(Map(this, f, 0), f)
+    } else {
+      this match {
+        case Map(source, g, index) =>
+          // Allowed to do fixed number of map operations fused before
+          // resetting the counter in order to avoid stack overflows;
+          // See `IOPlatform` for details on this maximum.
```
djspiewak, line 115: Nice solution to this problem. Also, we may be able to dump map fusion altogether. It needs more measuring in frameworks like Monix and fs2. Map fusion is one of those things that tends to look *amazing* in microbenchmarks and really doesn't make a measurable impact *at best* in practice (since it increases indirection and polymorphism). As an example, fs2 had map fusion in the 0.10 release, and it looked awesome in certain benchmarks, but in practical application benchmarks it actually *reduced* performance by over 2x.

So tldr, I want us to consider just removing fusion altogether, but we need more measurements in higher level applications to actually decide one way or another.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1305,6 +1348,13 @@ object IO extends IOInstances {
         ref.complete(IO.unit)
     }
 
+    if (tracingEnabled) {
+      IOTracing(nextIo, k)
```
djspiewak, line 1352: Any reason not to do the `tracingEnabled` check inside of the `IOTracing` constructor?

Also, as I read through this, I'm realizing just how many micro-optimizations we could add if we really wanted to. For example, *all* of these constructors are imposing memory barriers due to `object` usage. At the very least, caching the objects in private `val`s would improve inlining behavior.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1554,6 +1604,9 @@ object IO extends IOInstances {
   def contextShift(ec: ExecutionContext): ContextShift[IO] =
     IOContextShift(ec)
 
+  def backtrace: IO[IOTrace] =
```
djspiewak, line 1607: `val`

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1575,7 +1628,7 @@ object IO extends IOInstances {
   /** Corresponds to [[IO.map]]. */
   final private[effect] case class Map[E, +A](source: IO[E], f: E => A, index: Int) extends IO[A] with (E => IO[A]) {
     override def apply(value: E): IO[A] =
-      new Pure(f(value))
+      Pure(f(value))
```
djspiewak, line 1631: `new` is actually faster and probably should be preferred for internal code whenever possible

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1610,6 +1663,10 @@ object IO extends IOInstances {
     restore: (A, Throwable, IOConnection, IOConnection) => IOConnection
   ) extends IO[A]
 
+  final private[effect] case class Trace[A](source: IO[A], frame: TraceFrame) extends IO[A]
```
djspiewak, line 1666: So, architecturally, this means we have to *double* the number of allocations per `IO` run whenever tracing is enabled. I think we really need to put tracing in the existing nodes and just `null`/`None` it out whenever it's disabled.

core/shared/src/main/scala/cats/effect/internals/IOBracket.scala
```diff
@@ -40,17 +43,31 @@ private[effect] object IOBracket {
       if (!conn.isCanceled) {
         // Note `acquire` is uncancelable due to usage of `IORunLoop.start`
         // (in other words it is disconnected from our IOConnection)
-        IORunLoop.start[A](acquire, new BracketStart(use, release, conn, deferredRelease, cb))
+        val mode = if (tracingEnabled) {
+          IOTracing.getLocalTracingMode()
+        } else {
+          TracingMode.Disabled
```
djspiewak, line 49: `Disabled` should be in a private val somewhere in `IOBracket` to avoid the double barrier hit here. Basically something like:

```scala
private[this] val Disabled = TracingMode.Disabled
```

Helps the inliner along quite a bit.

core/shared/src/main/scala/cats/effect/internals/IOBracket.scala
```diff
@@ -40,17 +43,31 @@ private[effect] object IOBracket {
...
+        }
+        IORunLoop.restart[A](acquire, ctx, mode, new BracketStart(use, release, conn, ctx, mode, deferredRelease, cb))
```
djspiewak, line 51: Dumb question: `start` was replaced with `restart`?

core/shared/src/main/scala/cats/effect/internals/IOContext.scala
```diff
@@ -0,0 +1,52 @@
...
+ * IOContext holds state related to the execution of an IO and
+ * should be threaded across multiple invocations of the run-loop
+ * for the same fiber.
+ */
+final private[effect] class IOContext private () {
+
+  // We have to use a volatile here because of IOBracket implementation
+  // and how it invokes a new run-loop "asynchronously."
+  // Ideally we could use a mutable, ring buffer here.
+  @volatile var frames: Vector[TraceFrame] = Vector.empty
```
djspiewak, line 31: Definitely need to use a different structure here. `Vector` is fine for now, but we'll want to use a projected fixed size `Array[TraceFrame]` (i.e. the mutable ring buffer concept).

core/shared/src/main/scala/cats/effect/internals/IORunLoop.scala
```diff
@@ -17,28 +17,54 @@
...
+import cats.effect.tracing.TracingMode
+import cats.effect.internals.TracingPlatformFast.tracingEnabled
+
 import scala.util.control.NonFatal
 
 private[effect] object IORunLoop {
   private type Current = IO[Any]
   private type Bind = Any => IO[Any]
   private type CallStack = ArrayStack[Bind]
+  // TODO: replace with a mutable ring buffer
```
djspiewak, line 30: Misplaced TODO?

core/shared/src/main/scala/cats/effect/internals/IOTracing.scala
```diff
@@ -0,0 +1,102 @@
...
+      val old = localTracingMode.get()
+      localTracingMode.set(mode)
+
+      // Rethrow any exceptions that `source` produces after resettubg nide,
+      // In the event of cancellation, the mode will be reset when the
+      // thread grabs a new task to run (via Async).
+      source.attempt.flatMap { e =>
+        localTracingMode.set(old)
+        IO.fromEither(e)
+      }
```
djspiewak, line 44: Faster variant:

```scala
source.onError(_ => IO(localTracingMode.set(old)).flatMap(a => IO { localTracingMode.set(old); a })
```

core/shared/src/main/scala/cats/effect/internals/IOTracing.scala
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+  def getLocalTracingMode(): TracingMode =
+    localTracingMode.get()
+
+  def setLocalTracingMode(mode: TracingMode): Unit =
+    localTracingMode.set(mode)
+
+  private def buildCachedFrame(source: IO[Any], lambdaRef: AnyRef): TraceFrame = {
+    val cachedFr = frameCache.get(lambdaRef)
```
djspiewak, line 54: We most definitely want to cache based on `lambdaRef.getClass` rather than `lambdaRef` itself, otherwise we'll almost never have hits.

core/shared/src/main/scala/cats/effect/internals/IOTracing.scala
```diff
@@ -0,0 +1,102 @@
...
+      val fr = buildFrame(source)
+      frameCache.put(lambdaRef, fr)
+      fr
+    } else {
+      cachedFr
+    }
+  }
+
+  private def buildFrame(source: IO[Any]): TraceFrame = {
+    // TODO: proper trace calculation
```
djspiewak, line 65: :-)

core/shared/src/main/scala/cats/effect/tracing/IOTrace.scala
```diff
@@ -0,0 +1,30 @@
...
+package cats.effect.tracing
+
+final case class IOTrace(frames: Vector[TraceFrame]) {
+
+  def printTrace(): Unit = {
+    System.err.println("IOTrace")
+    frames.foreach { f =>
+      f.line.foreach { l =>
+        System.err.println(s"\t${f.op} at ${l.className}.${l.methodName} (${l.fileName}:${l.lineNumber})")
+      }
```
djspiewak, line 26: BTW, from an appearance standpoint, I really like the ascii art I did here for fiber tracing: https://github.com/djspiewak/coop/blob/master/src/main/scala/coop/ThreadT.scala#L95 That's just me bikeshedding though. :-)

core/shared/src/main/scala/cats/effect/IO.scala
neko-kai: case class `apply` is inlined to constructor calls even with scalac inliner turned off, if this isn't custom apply the object call is likely not happening

djspiewak:
Some thoughts:

> If tracing is globally disabled, there should be negligible impact to application performance. The current implementation sees a 0% performance degradation; the JIT can dynamically eliminate all consequent code paths at runtime. But it comes at a steep cost when tracing is enabled.

IMO, global disabling of tracing should result in a 0% degradation. Local disabling should result in at worst something like a 1-2% degradation, while fully-enabled maybe something like 3-4%. At least those are kind of the numbers I'm thinking. Obviously less is better, but this feels relatively attainable.

> In general, Scala cross-module singleton variable accesses cost a volatile read. I don't have a primary source for this but if you inspect the bytecode you will see it. The JIT can't optimize this out.

I think the bytecode *is* a primary source. :-)

> Global flag defined in Java scope. After class initialization, the HotSpot C2 compiler is aware of the values of all static final variables (or so called just-in-time constants). With that information, it can eliminate code paths that are consequent to those constants. As described above, Scala object-level fields can't take advantage of this, so we're defining the global tracing flag in Java instead. One issue here is package visibility.

I don't think it *hurts* to make it publicly visible if we have to. It's in the `internals` package and I'm perfectly happy to say it's out of scope for bincompat.

> Global flag conditionals. Originally I wanted to hide the conditionals behind an IO.Builder helper and IOTracing utility methods, but C2 can't inline those method calls in 2.12. So the conditionals are somewhat scattered around the codebase.

Fascinating. I wonder if that's due to the inlining limits. Also that answers one of my review questions. We could get around this by using a macro in `IOTracing`.

> Trace frame capture. There were two approaches I identified here: either capture the trace frames as instance variables in each IO node, or introduce a new IO constructor that reifies an IO node with a trace frame. I took the latter approach because there's virtually no cost when tracing is disabled, and it doesn't pollute the constructor or interpreter logic for other IO instructions. However, there is a very steep cost (more than 30-40% IIRC) when tracing IS enabled because it effectively doubles the number of interior nodes in the IO tree being traced. However, the instance variable approach is much cheaper during tracing, but incurs around a 4-5% performance penalty when its disabled.

Ah more of my review comments answered!

Where does the cost come from when the tracing is disabled? Just the larger object header size?

> Trace frame cache. Trace frames are cached by lambda references as keys, so we can easily cache Map, Bind, and Async nodes. I chose a global ConcurrentHashMap here, but we probably need to benchmark it more and compare it to alternatives like ThreadLocal. Also, this should probably be an LRU cache bounded in entry count. There is a snag with this caching approach where the same lambda reference could be used in multiple places:

`ConcurrentHashMap` is *very* fast for read-biased workloads. Not like, perfectly optimal, but pretty good. It's certainly going to be faster than `ThreadLocal`.

Also I think we can actually not worry about the LRU part of this if we cache based on lambda *class* rather than lambda instance, since the number of distinct cache entries will be strictly bounded by the number of classes, which is static (and relatively small).

Also regarding your example, that seems completely fine to me. We know tracing is going to suffer from these kinds of limitations.

> The Scala compiler won't generate a new lambda class for each invocation of x, but we still need some way to detect if we're looking at a lambda that is shared like this, otherwise caching could produce incorrect results. Maybe there's a way to compare the name of the lambda with reflection. Bytecode analysis is also another option here. But I personally haven't seen or used this pattern ever, so maybe we explicitly call it out as a limitation.

We should explicitly call it out, but IMO it's completely fine. Tracing is an approximation, really, intended to give *some* information where, at present, we have none.

> Async. The implementation of IOBracket leverages the Async constructor and invokes a new execution of the IORunLoop. Since two executions of the run-loop can then be associated with the same fiber, we need some way to move state between each execution. My approach is to create a thread-safe IOContext class that holds a buffer of traces and gets passed along in the Async constructor. Ideally it would be mutable (with an internal ring buffer) because there's a cost to the synchronization, but the alternative was to significantly refactor the run-loop. I believe some of the work done for CE3 will simplify the bracket implementation so only one run-loop is invoked per IO.

Ah yeah this is rough. I really don't like the fact that bracket is leveraging `Async`, tbh. That definitely imposes a significant penalty on things. Thread safety in general imposes a high penalty, and particularly for a case like this where we know it *should* be synchronous… :-/ We probably need to resolve this.

> Trace frame calculation. The current implementation filters out specific class prefixes from stack frames and takes the first one remaining. Obviously this won't work with constructs like monad transformers, so we need to figure out a smarter algorithm here. Maybe taking sequences of frames that might be relevant. Or the user can specify a list of packages they're interested in. Also need to clean up the class/lambda names a little bit.

Monad transformers in general are going to cause a problem for rabbit tracing, since the lambdas will basically all come from the transformer. Slug tracing obviously doesn't have this problem since it bypasses the cache altogether, but then you take the performance hit.

At any rate, we're basically looking at heuristics either way. I think I had some random ideas on how we might do this in the spec gist, but ultimately nothing is *really* going to be fully general. I have no objections to starting with the class filtering approach and seeing how it works.

> Lexically scoped tracing mode. ThreadLocal was the easy option here and the performance hit isn't terrible. Preserving ThreadLocal state across asynchronous boundaries is fairly straightforward: before invoking the continuation of an Async, save the current state. Restore that state when the continuation invokes the callback. Note that in the case of cancellation, we don't need to worry about resetting the ThreadLocal of that thread, because it will be reset when it starts a new run-loop or picks up a new task (which is re-entering the run loop from an asynchronous boundary).

This is fairly nice, and my ideas around exploiting the memory model basically just make the thread local approach *faster*. So, this is the right place to start with implementation, IMO.

> There are many constructor calls for IO.Async throughout the library. Should we trace them all?

There are a *ton* of them, and IMO none of the internal ones should be traced. Tracing is meant to show *user*-created structure, not implementation details.

> What should trace frames look like?

See coop. :-)

RaasAhsan:
> Fascinating. I wonder if that's due to the inlining limits. Also that answers one of my review questions. We could get around this by using a macro in IOTracing.

Before 2.12, the singleton static instance was declared `final`, so method calls could be inlined. If you check out the link in the limitations section, it was made non-final in 2.12 to maintain compatibility with JDK 9 class files. But it's since been fixed in 2.13. Will look into the macro approach.

> Where does the cost come from when the tracing is disabled? Just the larger object header size?

Yeah, exactly. I'm certain the cost will be greater for the simpler `IO` constructors like `Pure` and `Delay`.

> Also I think we can actually not worry about the LRU part of this if we cache based on lambda class rather than lambda instance, since the number of distinct cache entries will be strictly bounded by the number of classes, which is static (and relatively small).

Ah, this makes a lot of sense. I thought the lambda reference is always a singleton, but this obviously isn't the case, especially if it closes over other variables in scope. 👍  to caching the class.

> Ah yeah this is rough. I really don't like the fact that bracket is leveraging Async, tbh. ... We probably need to resolve this.

I came across a comment in the codebase about leveraging `ContextSwitch` to implement `bracket`. I haven't looked into feasibility, but do you think this is something worth pursuing? Another alternative is to allow the `Async` continuation to pass back state, but in such a way that it's not exposed to the public API. It would be a bit of internal refactoring.

> Monad transformers in general are going to cause a problem for rabbit tracing, since the lambdas will basically all come from the transformer. 

This makes me question how much value rabbit tracing will provide. I think it's fair to say that if you're using transformers, its likely that it permeates throughout your entire codebase (this certainly has been the case for me). If the transformer lambdas/classes are cached, tracing is likely to produce incorrect results, at least past the immediate frame. 

I wonder if we should refocus the scope of rabbit tracing away from stack-based tracing and just expose run-time information (e.g. instructions, forks/joins, async boundaries, yielding, etc.). Slug mode could focus more on stack-based tracing, and we could offer options around caching and trace frame calculation that users can customize to their preference.

Of course, the user would have to be well-aware of what the limitations of those options are before turning them on, but I think that's reasonable since they are options. But the bottom line would be that the default, out-of-the-box behavior works correctly.

> There are a ton of them, and IMO none of the internal ones should be traced. Tracing is meant to show user-created structure, not implementation details.

👍

core/shared/src/main/scala/cats/effect/internals/IOBracket.scala
RaasAhsan: Good call. I've been looking for things like this throughout the codebase to help optimize a bit.

core/shared/src/main/scala/cats/effect/IO.scala
RaasAhsan: You called it out in your comment, but there's a performance hit due to the larger object header size. It's going to look worse for simpler `IO` nodes, but for `Map` there was a 4-5% degradation when tracing is disabled. Do you think this is acceptable?

core/shared/src/main/scala/cats/effect/IO.scala
RaasAhsan: Ahh that's right, this goes through an `apply` method if I'm not mistaken :) . I'll try to find other stragglers

core/shared/src/main/scala/cats/effect/IO.scala
RaasAhsan: Mentioned it in my comment, but only in 2.12, singleton method calls can't be inlined.

core/shared/src/main/scala/cats/effect/tracing/IOTrace.scala
RaasAhsan: You happen to have an example of what that looks like ?

core/shared/src/main/scala/cats/effect/internals/IOTracing.scala
Avasil: Probably even better to use `redeem` or `redeemWith`

RaasAhsan: Unrelated: has an `onSuccess` or even an `onComplete` combinator, with similar semantics to `onError`, ever been considered? It would make use cases like these or "logging actions" nicer to write out. I haven't found anything like it yet though.

Avasil: There is `flatTap` which is like `onSuccess` but it's coming from Cats syntax

RaasAhsan: Missed your first comment :). I like the idea of `redeemWith`, that pushes only one frame on the bind stack.

core/shared/src/main/scala/cats/effect/IO.scala
djspiewak: I think @neko-kai's point is that case class `apply` is special magic (which I completely forgot about). I can confirm that the bytecode should be identical to using `new`.

core/shared/src/main/scala/cats/effect/IO.scala
djspiewak: Hmm, let me ponder this…

djspiewak: It's fascinating to me that adding a pair of pointers to the object header reduces performance by 4-5%. That in and of itself hints that allocation costs are dominating here.

Given that we want the *default* mode of operation to have rabbit tracing enabled, I feel like we may have to simply accept the 4-5% performance hit when disabled. I'm not sure if there's any way around this given the approach that we're taking. There may be ways we can reduce some of the pre-existing allocation burden though, which would result in winning back those losses. For now though, a 4-5% disabled perf hit seems more acceptable than the 2x allocations hit when enabled.

core/shared/src/main/scala/cats/effect/tracing/IOTrace.scala
djspiewak: Dug a bit and apparently I wrote up an example here: https://github.com/djspiewak/coop/releases/tag/v0.2.1

djspiewak:
> I came across a comment in the codebase about leveraging ContextSwitch to implement bracket. I haven't looked into feasibility, but do you think this is something worth pursuing? Another alternative is to allow the Async continuation to pass back state, but in such a way that it's not exposed to the public API. It would be a bit of internal refactoring.

It just surprises me that we need those approaches at all. I guess this is betraying my lack of understanding of the current runloop, but `bracket` should in theory be implementable by maintaining a (possibly null) stack of finalizers, along with a boolean state indicating whether or not the current region is masked (or something isomorphic to this). I'm not sure I understand why either `Async` or `ContextShift` are strictly required.

> This makes me question how much value rabbit tracing will provide. I think it's fair to say that if you're using transformers, its likely that it permeates throughout your entire codebase (this certainly has been the case for me). If the transformer lambdas/classes are cached, tracing is likely to produce incorrect results, at least past the immediate frame.

Tracing in general is definitely lacking in value in a lot of common scenarios. I think the real advantage here is in providing *any* information, where currently there is none.

What I'm currently thinking about in terms of use-case for rabbit tracing is someone gets a production crash and/or a production livelock and they want to start diagnostics. They're going to turn on slug tracing locally, but they need to have *some* idea of how to reproduce the situation in the first place. Tracing at least gives some hints. (as a note on the livelock case, being able to dump the traces of all the fibers would be awesome, basically giving us the analogue of a thread dump)

I don't think there's any particular way of getting around the fact that, in a lot of cases, the trace will just show "yet more `OptionT.flatMap`" and/or things like the fs2 interpreter internals. But that's still at least *some* information. We could refocus more on map/yield/etc, but that's basically what rabbit tracing is already showing, just with the addition of a (possibly misleading) stack frame.

I feel like, as with all lossy introspection tools (think: heap dump analysis), people eventually learn the ins and outs of when it's helpful and what the hints usually mean in which context, and they learn to interpret the patterns in ways that the tool authors didn't anticipate. All of which is to say I think more information is better.

So while I agree with you that fast tracing (regardless of implementation details) is *generally* not going to be completely accurate in any application with interpreter stack layers (monad transformers, streaming stuff like fs2, etc), I do think there's still quite a bit of value in it just on the basis of giving you *something* to hang your hat on.

core/shared/src/main/scala/cats/effect/internals/IOTracing.scala
alexandru: @RaasAhsan that's what it was meant for, in many cases doing `attempt.flatMap` is wasteful.

alexandru:
> It just surprises me that we need those approaches at all. I guess this is betraying my lack of understanding of the current runloop, but bracket should in theory be implementable by maintaining a (possibly null) stack of finalizers, along with a boolean state indicating whether or not the current region is masked (or something isomorphic to this). I'm not sure I understand why either `Async` or `ContextShift` are strictly required.

@djspiewak `bracket` is currently implemented via `Async`. The stack you're thinking is in [IOConnection](https://github.com/typelevel/cats-effect/blob/master/core/shared/src/main/scala/cats/effect/internals/IOConnection.scala), but note that it corresponds to the current design.

The current implementation has limitations, because it's optimised for the current model. Some of the optimisations I've made are meant to avoid memory allocations and therefore are terrible, sorry about that.

I've cloned the CE 3 branch and will start experimenting with an expanded run-loop design that works with the new concurrency primitives.

---

@RaasAhsan I wouldn't go into building more things on top of `ContextSwitch` for now, it's an awful optimisation that's very error prone.

RaasAhsan:
I've updated the PR to address the `Bracket` / `Async` issue. Copying from Gitter...

> We already have a way to pass state back from a run-loop (the callback). So all we really need now is a way to capture the context we want to pass back in the IO, and zip it up with whatever the acquire is returning.

So we don't need to revise `bracket` anymore for the time being. I'll add some benchmarks to see what the hit for passing along context is.

core/jvm/src/main/java/cats/effect/internals/TracingPlatformFast.java
```diff
@@ -0,0 +1,54 @@
...
+        .filter(x -> !x.isEmpty())
+        .map(x -> Boolean.valueOf(x))
+        .orElse(true);
+
+    /**
+     * The number of trace lines to retain during tracing. If more trace
+     * lines are produced, then the oldest trace lines will be discarded.
+     * Automatically rounded up to the nearest power of 2.
+     */
+    public static final int maxTraceFrameSize = Optional.ofNullable(System.getProperty("cats.effect.tracing.maxTraceFrameSize"))
```
RaasAhsan, line 43: This probably doesn't need to be in Java but keeping it here for now

core/shared/src/main/scala/cats/effect/internals/IOContext.scala
```diff
@@ -0,0 +1,47 @@
...
+
+  def pushFrame(fr: TraceFrame): Unit = {
+    val a = frames.push(fr)
+    if (a != null) {
+      omitted += 1
+    }
+  }
+
+  def getTrace: IOTrace =
+    IOTrace(frames.toList.toVector, omitted)
```
RaasAhsan, line 40: Need a better way to access the contents of the RingBuffer. ArrayStack uses an iterator so may just use that here as well

core/shared/src/main/scala/cats/effect/tracing/IOTrace.scala
RaasAhsan: I'm liking how this looks. We could indent on binds and de-indent after the bind continuation completes, that would give a pretty cool visual of the execution tree.

i'm also wondering if there's anything we could render around start/join/cancellations.

RaasAhsan:
Alright, there's light at the end of the tunnel here. We've made several changes to the design, so I'll go over those:

1. **No lexical tracing mode**. Instead, there is a global tracing mode configured by a system property. The main reason for this is that it simplifies API and implementation (no ThreadLocals or non-volatile magic) and speeds up rabbit mode tracing. There are three modes: 
    * **DISABLED**: No traces are captured for any `IO` constructors. There is drop in performance here (2 to 3%) now because of a tradeoff that makes rabbit mode much faster. Nothing is accumulated in the trace buffer for a fiber.
    * **RABBIT**: Traces are captured for `flatMap`, `map`, and the three `async` variants. Traces are cached in a global CHM. When an `IO` is in an active tracing collection region (denoted by `IO.traced`), all collected traces in that region will be accumulated in the trace buffer. Need to do more performance testing here, but if no accumulation is happening, there is a 15% to 20% performance hit here.
    * **SLUG**: Traces are captured for all `RABBIT-traced nodes, as well as `pure`, `suspend` and `delay`. No caching of traces is performed, so the benchmarks show a 99% hit or so :)
2. **IOContext publication**. I don't believe we need to do anything special for guaranteeing publication of `IOContext` across `Async` or `IOBracket` anymore. Since the `IOContext` reference is held in `RestartCallback` and `BracketStart`, any updates to the context should be passed along with the callback until it's invoked. In other words, you can always establish a happens-before relationship between a write to and a read from the context *on the same fiber*, even across asynchronous boundaries and trampolines, since its piggybacking on the callback. Hopefully someone with a better understanding of the memory model can back me up on this :)
3. `IO.Trace`. We introduced a new constructor in the `IO` encoding that captures a trace frame for a child node. The intention behind this is to keep rabbit tracing fast: the `Trace` node slows down the run-loop tremendously (it effectively doubles the size of the tree), so for the `IO` constructors whose traces can be cached (`Map`, `Bind`, `Async`), track those directly in the constructor. As mentioned earlier, there is a minuscule performance penalty for this when tracing is disabled.

## Example
```scala
def print(msg: String): IO[Unit] =
  IO.delay(println(msg))

def program2: IO[Unit] =
  for {
    _ <- print("3")
    _ <- print("4")
  } yield ()

def program: IO[Unit] =
  for {
    _ <- print("1")
    _ <- print("2")
    _ <- IO.shift
    _ <- IO.async[Int](cb => cb(Right(32)))
    _ <- IO.unit.bracket(_ => program2)(_ => IO.unit)
    _ <- print("5")
  } yield ()

override def run(args: List[String]): IO[ExitCode] =
  for {
    _ <- IO.suspend(program).traced
    trace <- IO.backtrace
    _ <- trace.prettyPrint
    _ <- trace.compactPrint
  } yield ExitCode.Success
```
[Output](https://gist.github.com/RaasAhsan/33e2ad5e30fb5d578839561e13c39be6).

## Next steps
I'm going to start on tests, benchmarks and docs next, but looking for more feedback on the following items:
* What information can we collect and render for each mode?
* Does `compactPrint` and `prettyPrint` make sense for both modes?
* What other combinators should we trace for each mode? We can go pretty crazy in slug mode and capture them all if we wanted.
* Is there other useful information we can trace just from the run-loop? For example, in rabbit mode, we can still trace `pure`/`delay`/`suspend` in the run-loop and insert an entry (with the resulting value!). However we just can't print out a source file location.
* Other customizable tracing/rendering options?
We certainly don't have to get everything in on the first try, but looking to others to see what makes sense to have in the first iteration.

RaasAhsan:
>  Slug is all about giving users the most possible information we can. We might need to increase the size of the circular buffer though if we're adding extra intermediate frames, otherwise a slug trace might end up having less information than a rabbit trace.

👍 

I think a neat way to do this without polluting every single `IO` combinator is to infer what combinator was invoked by looking at the stack trace, which is exactly the kind of analysis slug tracing should let you do.

> The problem is that we can't toString values blindly because it can be slow (on large immutable structures), non-terminating (on BigDecimal), and/or side-effecting (on URL).

Great call out, should be fairly easy to implement. I wonder if this should be enabled/disabled by a system-level property, since it would arguably also incur a performance hit though should be very tiny.

> I wonder if there's any way we can make some of this instrumentation pluggable without compromising performance? I don't see a way off hand but it's worth some thought.

For rendering, we were discussing that the user could just consume the `IOTrace` themselves by converting to JSON/graphviz/etc. Pluggable instrumentation will theoretically be more costly, but C2's class hierarchy analysis can also optimize and inline dynamic dispatches for a class that only has one subclass loaded. Another thing to benchmark and dig into the native code for.

Agreed with the rest of your feedback, will push changes tomorrow.

core/js/src/main/scala/cats/effect/internals/Tracing.scala
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.internals
+
+object Tracing {
```
RaasAhsan, line 19: Will probably just merge this with `TracingPlatform` for both the jvm and js builds. The original intention was to have a separate `Tracing` source file in the shared sources so there aren't syntax errors in `IO.scala`, but scalaJS can't compile java, so it caused a linking error

core/js/src/main/scala/cats/effect/internals/Tracing.scala
```diff
@@ -0,0 +1,25 @@
...
+  val isRabbitTracing: Boolean = false
+
+  val isSlugTracing: Boolean = false
```
RaasAhsan, line 22: Not familiar with scalajs, is there a canonical way to load configuration from the environment?

alexandru: We could load stuff from `window` or `global`, the problem is the name of the global object differs:

- `window` in the browser
- `global` in Node.js
- [globalThis](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) in service workers

We could have a `catsEffectSettings` settings, set as a dictionary on these references, but you'll have to check all of them.

alexandru: Note that Scala.js might automatically link to the global `this` already, under a common name — read their docs or ask on their channel.

9rb:
Thanks a ton for this work. This will help developer mental health a lot and save many developer hours 👍

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1575,12 +1647,12 @@ object IO extends IOInstances {
   final private[effect] case class Suspend[+A](thunk: () => IO[A]) extends IO[A]
 
   /** Corresponds to [[IO.flatMap]]. */
-  final private[effect] case class Bind[E, +A](source: IO[E], f: E => IO[A]) extends IO[A]
+  final private[effect] case class Bind[E, +A](source: IO[E], f: E => IO[A], trace: AnyRef) extends IO[A]
```
djspiewak, line 1650: It's worth commenting why we're using `AnyRef` here rather than `Trace`. Oh also, out of curiosity, have you measured whether the performance impact from using `Trace` vs `AnyRef` is specific to C2? More specifically, does it also have the same impact on GraalVM?

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1598,10 +1670,15 @@ object IO extends IOInstances {
...
-    k: (IOConnection, Either[Throwable, A] => Unit) => Unit,
-    trampolineAfter: Boolean = false
+    k: (IOConnection, IOContext, Either[Throwable, A] => Unit) => Unit,
+    trampolineAfter: Boolean = false,
+    trace: AnyRef = null
   ) extends IO[A]
 
+  final private[effect] case class Trace[A](source: IO[A], trace: StackTraceFrame) extends IO[A]
+
+  final private[effect] case class SetTracing(collectStackTraces: Boolean) extends IO[Unit]
```
djspiewak, line 1680: Do we still need this?

core/shared/src/main/scala/cats/effect/internals/IOTracing.scala
```diff
@@ -0,0 +1,78 @@
...
+    buildFrame(traceTag)
+
+  def cached(traceTag: TraceTag, clazz: Class[_]): StackTraceFrame =
+    buildCachedFrame(traceTag, clazz)
+
+  def traced[A](source: IO[A]): IO[A] =
+    resetTrace *>
+      Bind(incrementCollection,
+           (_: Unit) => Bind(source, DecrementTraceCollection.asInstanceOf[A => IO[A]], null),
+           null)
```
djspiewak, line 40: Needs a scalafmt run. :-)

core/shared/src/main/scala/cats/effect/internals/IOTracing.scala
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  def buildFrame(traceTag: TraceTag): StackTraceFrame =
+    StackTraceFrame(traceTag, new Throwable())
+
+  private[this] val incrementCollection: IO[Unit] = SetTracing(true)
+
+  private[this] val decrementCollection: IO[Unit] = SetTracing(false)
+
+  private object DecrementTraceCollection extends IOFrame[Any, IO[Any]] {
```
djspiewak, line 60: `private[this]`

core/shared/src/main/scala/cats/effect/internals/RingBuffer.scala
```diff
@@ -0,0 +1,77 @@
...
+  private[this] val mask = length - 1
+
+  // TODO: this can be an expensive allocation
+  // either construct it lazily or expand it on-demand
+  private[this] val array: Array[AnyRef] = new Array(length)
+  private[this] var writeIndex: Int = 0
+  private[this] var readIndex: Int = 0
+
+  def push(a: A): A = {
+    val wi = writeIndex & mask
```
djspiewak, line 38: I'm ashamed to say that I stared at this pretty hard for a while before I realized that it was just `% length` in disguise.

core/shared/src/main/scala/cats/effect/internals/RingBuffer.scala
```diff
@@ -0,0 +1,77 @@
...
+    if (writeIndex == readIndex + length) {
+      val old = array(wi)
+      array(wi) = a
+      // TODO: overflow at int.maxvalue?
+      writeIndex = writeIndex + 1
+      readIndex = readIndex + 1
+      old.asInstanceOf[A]
```
djspiewak, line 45: The buffer will very quickly fill up, then this will be the only case you'll ever hit. I honestly wonder if it's just worth removing the conditional to improve branch prediction. The `else` is effectively a fastpath for the first `length` frames, since arrays are pre-initialized to `null`, so the `if` branch is always correct.

core/shared/src/main/scala/cats/effect/internals/RingBuffer.scala
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  def capacity: Int =
+    length
+
+  def isEmpty: Boolean =
+    readIndex == writeIndex
+
+  // TODO: expose this as an iterator instead?
+  def toList: List[A] =
```
djspiewak, line 60: Is this just used for rendering? If so then I think `List` is fine. If it's used for anything that requires performance, then exposing the array buffer directly would be better.

core/shared/src/main/scala/cats/effect/internals/RingBuffer.scala
```diff
@@ -0,0 +1,77 @@
...
+
+}
+
+object RingBuffer {
+
+  // TODO: bounds check at int.maxvalue ?
+  private def nextPowerOfTwo(i: Int): Int = {
+    var n = 1
+    while (n < i) {
+      n = n * 2
```
djspiewak, line 72: ⛳ `n *= 2`

site/src/main/mdoc/guides/tracing.md
```diff
@@ -0,0 +1,176 @@
+---
+layout: docsplus
+title:  "Tracing"
+position: 2
+---
```
djspiewak, line 5: !!! You documented it?!?! Wow!

site/src/main/mdoc/guides/tracing.md
```diff
@@ -0,0 +1,176 @@
...
+an application.
+5. Tree rendering. By collecting a trace of all `IO` operations, a pretty tree
+or graph can be rendered to visualize fiber execution.
+6. Fiber identity. Fibers, like threads, are unique and can therefore assume an
+identity. If user code can observe fiber identity, powerful observability tools
+can be built on top of it. For example, another shortcoming of asynchronous
+code is that it becomes tedious to correlate log messages across asynchronous
+boundaries (thread IDs aren't very useful). With fiber identity, log messages
+produced by a single fiber can be associated with a unique, stable identifier.
+7. Fiber ancestry graph. If fibers can assume an identity, an ancestry graph
```
djspiewak, line 69: This should be SUPER easy to do and will help a lot, IMO. Follow-up PRs.

Also this list is great.

site/src/main/mdoc/guides/tracing.md
```diff
@@ -0,0 +1,176 @@
...
+6. Fiber identity. Fibers, like threads, are unique and can therefore assume an
+identity. If user code can observe fiber identity, powerful observability tools
+can be built on top of it. For example, another shortcoming of asynchronous
+code is that it becomes tedious to correlate log messages across asynchronous
+boundaries (thread IDs aren't very useful). With fiber identity, log messages
+produced by a single fiber can be associated with a unique, stable identifier.
+7. Fiber ancestry graph. If fibers can assume an identity, an ancestry graph 
+can be formed, where nodes are fibers and edges represent a fork/join
+relationship.
+8. Asynchronous deadlock detection. Even when working with asynchronously
```
djspiewak, line 72: This is an important one. We should honestly consider having `IOApp` register a signal trap so that people can dump a current fiber trace with `kill`, and probably also register an `MBean` with the same. Of course, for that to actually help with deadlocks, we would need a) fiber identity, and b) fiber state (similar to when thread dumps show deadlocks) indicating that a fiber is in an `Async` awaiting its continuation (I guess this might be as simple as just "the top of the trace is an `async` node").

site/src/main/mdoc/guides/tracing.md
```diff
@@ -0,0 +1,176 @@
...
+blocking code, fiber deadlocks aren't impossible. Being able to detect
+deadlocks or infer when a deadlock can happen makes writing concurrent code
+much easier.
+9. Live fiber trace dumps. Similar to JVM thread dumps, the execution status 
+and trace information of all fibers in an application can be extracted for 
+debugging purposes.
+10. Monad transformer analysis.
```
djspiewak, line 79: This is basically just "better heuristics for stack frame extraction" and it's definitely going to be interesting.

site/src/main/mdoc/guides/tracing.md
```diff
@@ -0,0 +1,176 @@
...
+that subprogram completes, use the `backtrace` combinator to collect the full
+fiber trace.
+
+```scala
+for {
+  _     <- program.traced
+  trace <- IO.backtrace
+  _     <- trace.compactPrint
+} yield ()
+```
```
djspiewak, line 102: Does this actually work?

site/src/main/mdoc/guides/tracing.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Asynchronous stack tracing
+The stack tracing mode of an application is configured by the system property
+`cats.effect.tracing.mode`. There are three stack tracing modes: `DISABLED`,
+`CACHED` and `FULL`.
+
```
djspiewak, line 108: I think this should be merged with the **Enable Tracing** section above.

site/src/main/mdoc/guides/tracing.md
```diff
@@ -0,0 +1,176 @@
...
+#### DISABLED
+No tracing is instrumented by the program. This is the default stack tracing 
+mode.
```
djspiewak, line 111: Do we think that Cached tracing is fast enough we can enable it by default?

site/src/main/mdoc/guides/tracing.md
```diff
@@ -0,0 +1,176 @@
...
+When cached stack tracing is enabled, a stack trace is captured and cached for
+every `map`, `flatMap` and `async` call in a program. 
+
+The stack trace cache is keyed by the lambda class reference, so cached tracing
+may produce inaccurate fiber traces under several scenarios:
+1. Monad transformer composition
+2. A named function is supplied to `map`, `async` or `flatMap` at multiple
+call-sites
+
+When no collection is performed, we measured less than an 18% performance hit
```
djspiewak, line 123: Out of curiosity, do we have an idea of where the majority of the impact lies? Is this just the `ConcurrentHashMap` lookup?

site/src/main/mdoc/guides/tracing.md
```diff
@@ -0,0 +1,176 @@
...
+for a completely synchronous `IO` program, so it will most likely be negligible
+for any program that performs any sort of I/O.
+
+This is the recommended mode to run in production environments.
```
djspiewak, line 127: IMO if it's recommended for production environments (which I agree with), then it should be the default.

benchmarks/shared/src/main/scala/cats/effect/benchmarks/DeepBindBenchmark.scala
```diff
@@ -54,26 +54,26 @@ class DeepBindBenchmark {
...
+//  @Benchmark
+//  def delay(): Int = {
+//    def loop(i: Int): IO[Int] =
+//      for {
+//        j <- IO(i)
+//        _ <- if (j > size) IO(j) else loop(j + 1)
+//      } yield j
+//
+//    loop(0).unsafeRunSync()
+//  }
```
djspiewak, line 66: FYI, JMH allows you to invoke individual benchmarks when it is run. For example: `jmh:run -i 3 -wi 3 -f 1 -t 1 cats.effect.benchmarks.ShallowBindBenchmark.pure` will *just* run the `pure` benchmark. So you shouldn't need to comment things out.

core/js/src/main/scala/cats/effect/internals/Tracing.scala
djspiewak: Serious question: does tracing work properly on ScalaJS at all? I don't actually know how `Throwable` and stack trace works on that platform.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -598,8 +608,10 @@ sealed abstract class IO[+A] extends internals.IOBinaryCompat[A] {
    *        release, along with the result of `use`
    *        (cancellation, error or successful result)
    */
-  def bracketCase[B](use: A => IO[B])(release: (A, ExitCase[Throwable]) => IO[Unit]): IO[B] =
-    IOBracket(this)(use)(release)
+  def bracketCase[B](use: A => IO[B])(release: (A, ExitCase[Throwable]) => IO[Unit]): IO[B] = {
+    val nextIo = IOBracket(this)(use)(release)
+    nextIo
```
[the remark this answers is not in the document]
RaasAhsan: We can actually trace `bracket` and `bracketCase`, using `use` and/or `release` as the key. I had it in here but removed it at some point since its a derived combinator. Might be worth including though

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -779,6 +791,9 @@ sealed abstract class IO[+A] extends internals.IOBinaryCompat[A] {
    * */
   def <&[B](another: IO[B])(implicit p: NonEmptyParallel[IO]): IO[A] =
     p.parProductL(this)(another)
+
+  def traced: IO[A] =
+    IOTracing.traced(this)
```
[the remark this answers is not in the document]
RaasAhsan: So, tracing collection involves two separate mechanisms. The first is collecting the stack traces at the call-site, where we create a `Throwable` and cache it if we're in the cached mode. The second is buffering them into the fiber-local ring. The global flag controls the first, so if its enabled, the collection happens globally for all combinators. The `traced` combinator controls the second, so you can easily specify the region of your program you're actually interested in tracing.

It's designed that way because the buffering has a performance cost, and users may or may not actually want the entire trace. The awkward part about this is that it doesn't really make much sense to have cached tracing on by default if you're not actually interested in tracing. I don't feel too strongly about this, but if you think this makes for a simpler API and the performance hit of buffering isn't too detrimental, we can remove `traced`.

core/shared/src/main/scala/cats/effect/IO.scala
RaasAhsan: Good point, I'll check Graal out

core/shared/src/main/scala/cats/effect/IO.scala
RaasAhsan: We need this in conjunction with the `traced` combinator. Specifically, if two `traced` calls are nested, we don't want the completion of the inner one to stop collection while the outer `traced` is still active.

If we remove `traced` and decide to buffer all stack traces, we can remove this. (see my reply to your comment about `traced`)

core/shared/src/main/scala/cats/effect/internals/RingBuffer.scala
RaasAhsan: Yeah, pretty much just rendering for now.

site/src/main/mdoc/guides/tracing.md
RaasAhsan: If `program` is defined it will :) I haven't actually run the `mdoc` process locally set, so I have to go back and fix that.

site/src/main/mdoc/guides/tracing.md
RaasAhsan: The idea was that the global tracing flag will control *all* tracing features collectively. Stack tracking is merely one application of tracing; someone might just want thread tracking and not stack tracing. Each section was going to describe how to enable the particular feature (whether it be via a system property or a combinator), but the global tracing flag is necessary for any kind of tracing. 

Now that I think about it, maybe we don't need an ultimate global tracing flag, and each tracing feature can be controlled independently. Let me try to remember why I decided to put it in.

site/src/main/mdoc/guides/tracing.md
RaasAhsan: See my comments above wrt `traced`. If we decide to remove it, I think we can turn it on by default

site/src/main/mdoc/guides/tracing.md
RaasAhsan: The 18% comes just from the `CHM` without buffering into the ring. When you're actually buffering into the ring, it's going to be more. IIRC it was in the 20-35% range.

benchmarks/shared/src/main/scala/cats/effect/benchmarks/DeepBindBenchmark.scala
RaasAhsan: I think I was commenting them out because I was only interested in `pure` and `delay` at the time. Can we use wildcards as well (like `sbt testOnly`)

core/shared/src/main/scala/cats/effect/IO.scala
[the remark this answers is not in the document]
djspiewak: How much of a performance hit does buffering impose? I definitely think that separating the semantics in this fashion is a bit confusing.

site/src/main/mdoc/guides/tracing.md
djspiewak: I wonder if there's a cheaper alternative to CHM… :-(

benchmarks/shared/src/main/scala/cats/effect/benchmarks/DeepBindBenchmark.scala
djspiewak: Yep!

site/src/main/mdoc/guides/tracing.md
RaasAhsan: Some alternatives I want to check out:
- `AtomicReferenceArray`
- `ThreadLocal`
- Faster `ThreadLocal` via a custom `Thread`
- global, non-volatile hashmap or array

core/shared/src/main/scala/cats/effect/IO.scala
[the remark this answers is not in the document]
RaasAhsan: Less than the CHM, and maybe there are some optimizations we can do around the ring buffer to make it better. I'll run some benchmarks, and if we see less than a 25% hit, I think we just get rid of `traced`. The impact to real applications would be far less severe given how frequent asynchronous boundaries happen

site/src/main/mdoc/guides/tracing.md
djspiewak: > global, non-volatile hashmap or array

I was kind of thinking about that. Like, cache misses early on aren't as painful as the steady-state cost of lookups on the fully-populated cache.

core/js/src/main/scala/cats/effect/internals/Tracing.scala
RaasAhsan: The scalajs test runs always fail for me so I can't really tell. it does compile though

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1116,8 +1131,14 @@ object IO extends IOInstances {
    * Any exceptions thrown by the effect will be caught and sequenced
    * into the `IO`.
    */
-  def delay[A](body: => A): IO[A] =
-    Delay(() => body)
+  def delay[A](body: => A): IO[A] = {
+    val nextIo = Delay(() => body)
+    if (isFullStackTracing) {
+      IOTracing.decorated(nextIo, TraceTag.Delay)
```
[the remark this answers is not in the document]
RaasAhsan: Awesome, they're both just `Function0`. :)

RaasAhsan:
I finally got around to running some proper benchmarks for this. Some context:
1. Map fusion was removed.
2. Delay and Suspend aren't being traced in cached mode. It's going to cause a perf hit for at least cached tracing and maybe disabled tracing, so I'm going to think about it more. 
3. The `traced` combinator that marks a region of code for trace accumulated was removed. This means that all stack traces that are captured will be pushed to the ring buffer. Arguably this makes things easier to benchmark and a simpler API, but there's a heavier cost for it.

The first column is the baseline; I just ran the benchmarks on `master`. The second is this PR with cached tracing. The third is this PR with disabled tracing. I didn't bother running it for full tracing because it's going to be horrendous either way. Numbers measure throughput (ops/sec), so higher number is better. Apologies for not making it clear who won :)

| Benchmark | master | cached tracing | disabled tracing   |
| --------- | ------ | -------------- | ------------------ |
| AsyncBenchmark.async | 119572.326 ± 413.003 | 79073.289 ± 2699.042 | 113184.161 ± 1329.851 |
| AsyncBenchmark.bracket | 29696.470 ± 506.753 | 23979.161 ± 379.111 | 28690.440 ± 543.955 |
| AsyncBenchmark.cancelBoundary | 149274.089 ± 1308.698 | 93234.194 ± 1462.561 | 117542.862 ± 1911.121 |
| AsyncBenchmark.cancelable | 76047.130 ± 1530.196 | 57926.565 ± 588.724 | 76606.612 ± 1249.202 |
| AsyncBenchmark.parMap2 | 5770.955 ± 424.211 | 3658.221 ± 419.610 | 3805.028 ± 1452.939 |
| AsyncBenchmark.race | 54208.527 ± 1046.410 | 47548.477 ± 351.995 | 50205.045 ± 1203.067 |
| AsyncBenchmark.racePair | 52858.627 ± 1348.588 | 46234.207 ± 401.834 | 49184.446 ± 1434.859 |
| AsyncBenchmark.start | 4613.861 ± 66.900 | 3938.927 ± 117.411 | 4235.612 ± 125.023 |
| AsyncBenchmark.uncancelable | 335959.872 ± 8007.988 | 215468.003 ± 55895.984 | 310715.990 ± 3577.594 |
| AttemptBenchmark.errorRaised | 2214.245 ± 30.976 | 1071.959 ± 37.659 | 2082.647 ± 39.953 |
| AttemptBenchmark.happyPath | 2093.844 ± 14.947 | 1687.826 ± 28.531 | 2103.024 ± 20.174 |
| DeepBindBenchmark.async | 341.085 ± 25.915 | 252.261 ± 13.372 | 334.080 ± 40.588 |
| DeepBindBenchmark.delay | 6769.888 ± 197.760 | 3865.844 ± 62.744 | 6589.597 ± 191.756 |
| DeepBindBenchmark.pure | 7486.475 ± 306.902 | 4166.002 ± 51.511 | 7350.192 ± 87.553 |
| ECBenchmark.app | 30.129 ± 0.134 | 24.933 ± 0.172 | 29.192 ± 0.256 |
| ECBenchmark.appWithCtx | 12.329 ± 0.540 | 8.804 ± 1.090 | 11.539 ± 1.177 |
| HandleErrorBenchmark.errorRaised | 2356.960 ± 13.261 | 1354.339 ± 20.241 | 2117.004 ± 21.843 |
| HandleErrorBenchmark.happyPath | 3264.041 ± 225.132 | 2450.450 ± 287.435 | 3211.480 ± 30.413 |
| MapCallsBenchmark.batch120 | 12016.605 ± 85.183 | 3764.943 ± 109.195 | 5975.144 ± 59.635 |
| MapCallsBenchmark.batch30 | 15716.561 ± 239.801 | 2956.037 ± 33.107 | 5927.286 ± 41.393 |
| MapCallsBenchmark.one | 5853.308 ± 28.972 | 5886166.788 ± 31165.093 | 46695239.188 ± 411206.001 |
| MapStreamBenchmark.batch120 | 3097.527 ± 54.732 | 1587.907 ± 22.619 | 2404.530 ± 45.621 |
| MapStreamBenchmark.batch30 | 1332.489 ± 15.626 | 632.612 ± 17.989 | 946.424 ± 18.011 |
| MapStreamBenchmark.one | 1447.147 ± 11.316 | 879.340 ± 25.421 | 1454.559 ± 18.920 |
| ShallowBindBenchmark.pure | 5998.937 ± 82.625 | 3022.161 ± 51.800 | 5996.171 ± 108.693 |
| ShallowBindBenchmark.async | 117.344 ± 7.443 | 87.645 ± 5.915 | 119.520 ± 11.301 |
| ShallowBindBenchmark.delay | 5008.061 ± 21.284 | 2668.011 ± 19.410 | 4919.652 ± 62.336 |

The good news is that disabled tracing is for the most part right behind `master`. This is mostly coming from the extra allocation in the `Map`, `FlatMap` and `Async` constructors.

The bad news is that cached tracing is hurting pretty bad. I'm seeing up to 50% degradations compared to `master` in some benchmarks. In order of weight, the hit is coming from the CHM, then the ring buffer operations, and lastly the extra allocations. Personally, cached tracing would add barely any overhead to any of the applications that I've run, but I'm sure some people would be pretty sensitive to it. My real concern is when we start to introduce other forms of tracing, then these numbers really start to add up. At the end of the day, I think people will just have to benchmark their apps.

Maybe this doesn't block us from making a RC, but after I address all the other PR comments, I'm going to look into whether we can make any tradeoffs with the CHM, and if the ring buffer can be made faster.

core/shared/src/main/scala/cats/effect/IO.scala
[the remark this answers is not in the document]
RaasAhsan: I'm having some hesitations about introducing `Delay` and `Suspend` to cached tracing. See the benchmark results below.

djspiewak:
Yeah that's a pretty serious hit in the cached case. I think we can optimize the caching more. Removing the read barriers will help for sure. I would also imagine we can optimize the table itself somewhat, depending on how sparse it's ending up being. We should also sanity-check `equals` and `hashCode` on `Class` to make sure they aren't doing anything inane (I suspect they aren't).

RaasAhsan:
I totally missed 3 memory barriers in `IOContext`, so I removed those. I also replaced `TraceTag` with integer tags, since we can't really inline them in the `IO` class without allocating or defining in Java. Honestly we can probably remove the tags altogether since we should be able to infer them from the stack trace. The `IOTracing` reference also can't be inlined in the `IO` class without allocation, but it isn't very costly either (in Scala 2.13 it can be inlined).

I'm omitting disabled tracing from this table.

| Benchmark | master | new cached tracing | old cached tracing   |
| --------- | ------ | -------------- | ------------------ |
| AsyncBenchmark.async | 119572.326 | 93693.168 | 79073.289 |
| AsyncBenchmark.bracket | 29696.470 | 27175.376 | 23979.161 |
| AsyncBenchmark.cancelBoundary | 149274.089 | 108202.295 | 93234.194 |
| AsyncBenchmark.cancelable | 76047.130 | 64941.074 | 57926.565 |
| AsyncBenchmark.parMap2 | 5770.955 | 3727.669 | 3658.221 |
| AsyncBenchmark.race | 54208.527 | 47109.698 | 47548.477 |
| AsyncBenchmark.racePair | 52858.627 | 46779.455 | 46234.207 |
| AsyncBenchmark.start | 4613.861 | 3996.665 | 3938.927 |
| AsyncBenchmark.uncancelable | 335959.872 | 251950.982 | 215468.003 |
| AttemptBenchmark.errorRaised | 2214.245 | 1403.077 | 1071.959 |
| AttemptBenchmark.happyPath | 2093.844 | 2120.155 | 1687.826 |
| DeepBindBenchmark.async | 341.085 | 284.715 | 252.261 |
| DeepBindBenchmark.delay | 6769.888 | 4824.286 | 3865.844 |
| DeepBindBenchmark.pure | 7486.475 | 5297.557 | 4166.002 |
| ECBenchmark.app | 30.129 | 25.984 | 24.933 |
| ECBenchmark.appWithCtx | 12.329 | 9.334 | 8.804 |
| HandleErrorBenchmark.errorRaised | 2356.960 | 1748.884 | 1354.339 |
| HandleErrorBenchmark.happyPath | 3264.041 | 2801.483 | 2450.450 |
| MapCallsBenchmark.batch120 | 12016.605 | 4622.263 | 3764.943 |
| MapCallsBenchmark.batch30 | 15716.561 | 3647.625 | 2956.037 |
| MapCallsBenchmark.one | 5853.308 | 6261480.811 | 5886166.788 |
| MapStreamBenchmark.batch120 | 3097.527 | 2048.182 | 1587.907 |
| MapStreamBenchmark.batch30 | 1332.489 | 760.250 | 632.612 |
| MapStreamBenchmark.one | 1447.147 | 1122.805 | 879.340 |
| ShallowBindBenchmark.pure | 5998.937 | 4122.696 | 3022.161 |
| ShallowBindBenchmark.async | 117.344 | 96.441 | 87.645 |
| ShallowBindBenchmark.delay | 5008.061 | 3319.394 | 2668.011 |

**A lot** better than before. Next up is the CHM

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -100,18 +101,18 @@ sealed abstract class IO[+A] extends internals.IOBinaryCompat[A] {
...
-        // Allowed to do fixed number of map operations fused before
-        // resetting the counter in order to avoid stack overflows;
-        // See `IOPlatform` for details on this maximum.
-        if (index != fusionMaxStackDepth) Map(source, g.andThen(f), index + 1)
-        else Map(this, f, 0)
-      case _ =>
-        Map(this, f, 0)
+  final def map[B](f: A => B): IO[B] = {
+    val trace = if (isCachedStackTracing) {
+      IOTracing.cached(TraceTag.Map, f.getClass)
```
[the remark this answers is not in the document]
RaasAhsan: `IOTracing` doesn't actually hit a memory barrier, it's just a `static non-final` variable, so it can't be inlined in 2.12, but in 2.13 it can be.

Removed TraceTags, we should be able to avoid the tags altogether when we start inspecting the stack trace

core/shared/src/main/scala/cats/effect/IO.scala
RaasAhsan: So I can't really tell what the situation is with the graal compiler. There is a high variance in throughput between individual benchmark runs (anywhere from 6500 to 6800), so I can't make any useful conclusions. The `AnyRef` still consistently gives better performance on C2 though

site/src/main/mdoc/guides/tracing.md
RaasAhsan: I'm thinking we publish the RC with caching enabled and ask people to test their applications. Depending on the response, we can make a more informed decision about this later? of course this is still pending how optimized we can get the cached mode

core/js/src/main/scala/cats/effect/internals/Tracing.scala
RaasAhsan: What do you guys think about just supporting tracing on JVM for the RC and coming back to it later? I'm honestly not sure if tracing even works on scalaJS but theoretically the only change we need is adding a configuration flag to enable it. Not really too high of a priority for me right now

RaasAhsan:
@djspiewak I think I'm happy with the API now, feel free to pull it and try it out sometime. Before we merge, is it worth closing this PR, moving to a new branch and reorganizing the commits?

djspiewak:
> Before we merge, is it worth closing this PR, moving to a new branch and reorganizing the commits?

I actually rather like having a comprehensive history, so I'm in favor of keeping it as-is, but I know other people have strong opinions on this stuff. :-)

Overall, I'm 👍 for merging it now! There's definitely more to be done, but we can't boil the ocean in a single PR. Let's get it in master, get it into people's hands, and see how it behaves. (build has to be fixed first though)

LukaJCB:
Thanks so much! 😍

PR #855, opened 2020-05-02
CE3: Refactor racePair: extract common code for running one side

kubukoz:


src/main/scala/ce3/playground.scala
```diff
@@ -258,6 +258,51 @@ object playground {
       def never[A]: PureConc[E, A] =
         Thread.done[A]
 
+      private def startOne[Result](
+        ctx: FiberCtx[E],
+        errorReg: E => PureConc[E, Unit],
```
kubukoz, line 263: These three functions look like a fold on an Outcome so maybe we could pass a function from Outcome to PureConc.

src/main/scala/ce3/playground.scala
```diff
@@ -258,6 +258,47 @@ object playground {
...
+      // Using the partially applied pattern to defer the choice of L/R
+      final class StartOnePartiallyApplied[Result](
+        parentMasks: List[MaskId],
+        foldResult: Outcome[Id, E, Result] => PureConc[E, Unit]
+      ) {
+        // we play careful tricks here to forward the masks on from the parent to the child
+        // this is necessary because start drops masks
+        def apply[L, OtherFiber](
+          that: PureConc[E, L],
+          getOtherFiber: PureConc[E, OtherFiber]
```
kubukoz, line 275: I should probably add a comment so that `OtherFiber` isn't confused with the result of the other fiber (it's really the whole fiber)

djspiewak:
Very nice indeed. It's embarrassing that `racePair` had so much code duplication.

PR #856, opened 2020-05-02
CE3: Reusable generators

kubukoz:
This is just an idea, but I noticed all of the generators only use the constructors / methods provided by the type class, the only thing that needs something more (`run`) being the `Cogen` instance.

Pushing this further would involve splitting the class into generators for monads, monaderrors, brackets, concurrents... which might or might not be easy to encode in Scala, but if it's possible - might give us a lot of niceness (e.g. generators for the Concurrent part of Resource, when it's available) for a very low cost.

Maybe we could use an encoding similar to what we do with laws extending from each other.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -26,6 +26,10 @@ val CatsVersion = "2.1.1"
 
 addCompilerPlugin("com.olegpy" %% "better-monadic-for" % "0.3.0")
 
+scalacOptions ++= Seq(
+  "-Xcheckinit"
+)
+
 libraryDependencies ++= Seq(
   "org.typelevel" %% "cats-core" % CatsVersion,
   "org.typelevel" %% "cats-free" % CatsVersion,
--- a/src/main/scala/ce3/Outcome.scala
+++ b/src/main/scala/ce3/Outcome.scala
@@ -37,6 +37,7 @@ private[ce3] trait LowPriorityImplicits {
   implicit def applicativeError[F[_]: Applicative, E]: ApplicativeError[Outcome[F, E, ?], E] =
     new ExitCaseApplicativeError[F, E]
 
+  //todo needs renaming
   protected class ExitCaseApplicativeError[F[_]: Applicative, E] extends ApplicativeError[Outcome[F, E, ?], E] {
 
     def pure[A](x: A): Outcome[F, E, A] = Completed(x.pure[F])
--- a/src/test/scala/ce3/laws/Generators.scala
+++ b/src/test/scala/ce3/laws/Generators.scala
@@ -20,91 +20,184 @@ import cats.implicits._
 
 import org.scalacheck.{Arbitrary, Cogen, Gen}, Arbitrary.arbitrary
 import cats.~>
+import cats.Monad
+import cats.Applicative
+import cats.Functor
+import cats.MonadError
+import cats.Traverse
 
-class Generators[F[_], E: Arbitrary: Cogen](
-  run: F ~> Outcome[Option, E, *]
-)(implicit F: ConcurrentBracket[F, E]) {
-  import Generators._
+final case class ArbitraryCogen[A](arbitrary: Arbitrary[A], cogen: Cogen[A])
 
-  def genF[A: Arbitrary: Cogen](depth: Int): Gen[F[A]] = {
-    if (depth > 10) {
-      Gen.frequency(
-        1 -> genPure[A],
-        1 -> genRaiseError[A],
-        1 -> genCanceled[A],
-        1 -> genCede.flatMap(pc => arbitrary[A].map(pc.as(_))),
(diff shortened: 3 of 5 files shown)


djspiewak:
Wow! I like this very very much. This should eventually be part of a `cats-effect-testkit` project, or something like that.

kubukoz:
I agree!

kubukoz:
I'll see if I can come up with a nice way to split the class into pieces corresponding to specific type classes.

kubukoz:
That went surprisingly well. I had to come up with a new abstraction (GenK), but in general the idea seems to hold.

Outstanding problems I discovered so far:
- diamond problem (probably naming each of the generators in the lists will be good enough to ensure we don't have the same generator N times for N shared ancestors)

src/test/scala/ce3/laws/Generators.scala
```diff
@@ -18,97 +18,179 @@ package ce3
...
-        1 -> Gen.delay(genHandleErrorWith[E, A](depth)),
-        1 -> Gen.delay(genNever),
-        1 -> Gen.delay(genRacePair[E, A](depth)),
-        1 -> Gen.delay(genStart[E, A](depth)),
-        1 -> Gen.delay(genJoin[E, A](depth)),
-        1 -> Gen.delay(genFlatMap[E, A](depth)))
+  //All generators possible at depth [[depth]]
+  private def gen[A: Arbitrary: Cogen](depth: Int): Gen[F[A]] = {
+    val genK: GenK[F] = new GenK[F] {
+      def apply[B: Arbitrary: Cogen]: Gen[F[B]] = Gen.delay(gen(depth + 1))
```
kubukoz, line 47: One relatively huge problem is that if we ever need any more instances than Arbitrary / Cogen to generate a value of `F[A]`, we'll have to change all of this. But somehow I don't see that happening, at least not in Cats / CE type classes, as everything is usually polymorphic and the only implicits we need are for the `F` itself. Which is still something that doesn't usually happen in type classes but functions on concrete types.

kubukoz:
@djspiewak take a look if you can, I think it's working pretty well (modulo duplicate generators, which I'll probably work around by naming each generator, much like we do for law tests).

kubukoz:
I think it's okay to merge this in the current form, it'll unblock some other work and we can always improve later. So... ready for review!

build.sbt
```diff
@@ -26,6 +26,10 @@ val CatsVersion = "2.1.1"
 
 addCompilerPlugin("com.olegpy" %% "better-monadic-for" % "0.3.0")
 
+scalacOptions ++= Seq(
+  "-Xcheckinit"
```
djspiewak, line 30: TIL. Also that's awesome

src/main/scala/ce3/Outcome.scala
```diff
@@ -37,6 +37,7 @@ private[ce3] trait LowPriorityImplicits {
   implicit def applicativeError[F[_]: Applicative, E]: ApplicativeError[Outcome[F, E, ?], E] =
     new ExitCaseApplicativeError[F, E]
 
+  //todo needs renaming
```
djspiewak, line 40: My bad…

src/test/scala/ce3/laws/Generators.scala
```diff
@@ -18,97 +18,180 @@ package ce3
...
-
 import org.scalacheck.{Arbitrary, Cogen, Gen}, Arbitrary.arbitrary
+import cats.~>
+import cats.Monad
+import cats.Applicative
+import cats.Functor
+import cats.ApplicativeError
+import scala.collection.immutable.SortedMap
+
+trait GenK[F[_]] {
```
djspiewak, line 29: 👍 I wish scalacheck had this

src/test/scala/ce3/laws/Generators.scala
```diff
@@ -18,97 +18,180 @@ package ce3
...
-      fa <- genPureConc[E, A](depth + 1)
-      f <- arbitrary[E => PureConc[E, A]]
-    } yield F[E].handleErrorWith(fa)(f)
+      fa <- deeper[A]
+      f <- Gen.function1[A, F[A]](deeper[A])
+    } yield fa.flatMap(f)
   }
+}
+
+trait ApplicativeErrorGenerators[F[_], E] extends ApplicativeGenerators[F] {
```
djspiewak, line 104: I'm starting to feel like this should be its own library. :-) I've wanted this type of functionality a *lot*, not just in the context of CE.

kubukoz: It should, let's just make this a test bed first and see what we might be missing :)

djspiewak: Sounds perfect!

Issue #324, opened 2018-08-28
Improve code coverage for cats.effect.concurrent


alexandru:
We've got some function implementations not being tested in `Ref` and `Semaphore` and we have to cover them:

- [Ref](https://codecov.io/gh/typelevel/cats-effect/src/master/core/shared/src/main/scala/cats/effect/concurrent/Ref.scala)
- [Semaphore](https://codecov.io/gh/typelevel/cats-effect/tree/master/core/shared/src/main/scala/cats/effect/concurrent)

SystemFw:
WDYT about a `good first issue` label for things like this?

satansk:
I will add tests for `Semaphore` :)

CucumisSativus:
@satansk are you still working on it? If not I will be super happy to take it over form you, if you are still working on it please let me know - i'll find other issue to solve :)

satansk:
@CucumisSativus it's yours :)

CucumisSativus:
Looks like `MVar` is missing quite a lot of tests, I'll take this possibility and write tem

PR #857, opened 2020-05-02
Add two tests for Semaphore for no available permits cases #324

biscout24:


core/shared/src/main/scala/cats/effect/concurrent/Semaphore.scala
```diff
@@ -217,15 +220,9 @@ object Semaphore {
...
-                now match {
-                  case Left(_) => false
-                  case Right(n) =>
-                    previous match {
-                      case Left(_)  => false
-                      case Right(m) => n != m
-                    }
-                }
+              case (_, Left(_))         => false
+              case (Right(m), Right(n)) => n != m
```
kubukoz, line 224: If you put this case first, you can wildcard others as `case _ => false`.

core/shared/src/test/scala/cats/effect/concurrent/SemaphoreTests.scala
```diff
@@ -89,7 +105,7 @@ class SemaphoreTests extends AsyncFunSuite with Matchers with EitherValues {
         .map(_ shouldBe 1)
     }
 
-    test(s"$label - available with no available permits") {
+    test(s"$label - available with 0 available permits") {
```
kubukoz, line 108: what's the difference between 0 and no permits? (I'm referencing the test you added above)

core/shared/src/test/scala/cats/effect/concurrent/SemaphoreTests.scala
```diff
@@ -41,6 +41,22 @@ class SemaphoreTests extends AsyncFunSuite with Matchers with EitherValues {
...
     }
 
+    test(s"$label - available with no available permits") {
+      val n = 20
+      sc(20)
+        .flatMap { s =>
+          for {
+            _ <- (0 until n).toList.traverse(_ => s.acquire).void
+            _ <- s.acquire.start
+            x <- (IO.shift *> s.available).start
```
kubukoz, line 51: I think you don't need to shift as `start` will do that for you.

core/shared/src/test/scala/cats/effect/concurrent/SemaphoreTests.scala
```diff
@@ -41,6 +41,22 @@ class SemaphoreTests extends AsyncFunSuite with Matchers with EitherValues {
         .map(_ shouldBe 0)
     }
 
+    test(s"$label - available with no available permits") {
+      val n = 20
+      sc(20)
+        .flatMap { s =>
+          for {
+            _ <- (0 until n).toList.traverse(_ => s.acquire).void
```
kubukoz, line 49: This could be `s.acquire.replicateA(n).void` (not that you need the `void`, since you're ignoring the result)

core/shared/src/test/scala/cats/effect/concurrent/SemaphoreTests.scala
```diff
@@ -41,6 +41,22 @@ class SemaphoreTests extends AsyncFunSuite with Matchers with EitherValues {
...
+            _ <- s.acquire.start
```
kubukoz, line 50: Is `start` necessary here? We have no guarantees about whether it actually completes or not before `s.available` below.

core/shared/src/test/scala/cats/effect/concurrent/SemaphoreTests.scala
```diff
@@ -41,6 +41,22 @@ class SemaphoreTests extends AsyncFunSuite with Matchers with EitherValues {
...
+            x <- (IO.shift *> s.available).start
+            t <- x.join
```
kubukoz, line 52: You're joining after a `start` immediately, so maybe replace the two lines with just `x <- IO.shift *> s.available` (or even ditch the shift, unless it has a specific purpose)

core/shared/src/test/scala/cats/effect/concurrent/SemaphoreTests.scala
```diff
@@ -156,7 +187,7 @@ class SemaphoreTests extends AsyncFunSuite with Matchers with EitherValues {
 
   tests("concurrent", n => Semaphore[IO](n))
 
-  test("concurrent - acquire does not leak permits upon cancelation") {
+  test("concurrent - acquire does not leak permits upon cancellation") {
```
kubukoz, line 190: I think cats settled on the other spelling for now :)

core/shared/src/test/scala/cats/effect/concurrent/SemaphoreTests.scala
```diff
@@ -130,6 +146,21 @@ class SemaphoreTests extends AsyncFunSuite with Matchers with EitherValues {
     }
 
     test(s"$label - count with no available permits") {
+      val n: Long = 8
+      sc(n)
+        .flatMap { s =>
+          for {
+            _ <- s.acquireN(n).void
+            _ <- (s.acquireN(n).void).start
```
kubukoz, line 154: This `start` could result in a race condition, so `s.count` below will sometimes return `0` instead of `-n` (since this `acquireN` won't complete in time). Similar case to one of my previous comments.

core/shared/src/main/scala/cats/effect/concurrent/Semaphore.scala
biscout24: I would like to show last case explicitly to figure-out how-to test it. I'm not sure that the last `map` necessary. Please check next commit

core/shared/src/test/scala/cats/effect/concurrent/SemaphoreTests.scala
biscout24: The previous one tests the `Left(_)` case form https://github.com/typelevel/cats-effect/blob/master/core/shared/src/main/scala/cats/effect/concurrent/Semaphore.scala#L280

So having 0 permits leads to `Right` state AFAIU, which this test case suppose to cover

I'm referring to https://codecov.io/gh/typelevel/cats-effect/src/master/core/shared/src/main/scala/cats/effect/concurrent/Semaphore.scala

core/shared/src/test/scala/cats/effect/concurrent/SemaphoreTests.scala
biscout24: very good point. I need to find the way how-to put semaphore into `Left` state and to not lead to race conditions...

core/shared/src/test/scala/cats/effect/concurrent/SemaphoreTests.scala
biscout24: I'm looking for a way to put semaphore into `Left` state and call `acquire` in other thread w/o race conditions. Do you have any idea?

core/shared/src/test/scala/cats/effect/concurrent/SemaphoreTests.scala
biscout24: @kubukoz , I think I have found solution for that. Could you check latest state and idea with `lockSemaphore` function?

core/shared/src/test/scala/cats/effect/concurrent/SemaphoreTests.scala
biscout24: wdyt about latest version?

biscout24: @kubukoz , could you take another look at the idea with `lockSemaphore` function? I would like to put semaphore in locked state and avoid race conditions here. WDYT?

kubukoz: Sorry, look me a while. I'll look at it when I get back to a screen :)

core/shared/src/test/scala/cats/effect/concurrent/SemaphoreTests.scala
```diff
@@ -155,6 +200,8 @@ class SemaphoreTests extends AsyncFunSuite with Matchers with EitherValues {
   }
 
   tests("concurrent", n => Semaphore[IO](n))
+  tests("concurrent in", n => Semaphore.in[IO, IO](n))
+  tests(s"concurrent imapK", n => Semaphore[IO](n).map(_.imapK[IO](Effect.toIOK, Effect.toIOK)))
```
kubukoz, line 204: unnecessary `s` here and below :)

core/shared/src/test/scala/cats/effect/concurrent/SemaphoreTests.scala
```diff
@@ -41,6 +57,21 @@ class SemaphoreTests extends AsyncFunSuite with Matchers with EitherValues {
         .map(_ shouldBe 0)
     }
 
+    test(s"$label - available with no available permits") {
+      val n = 20
+      sc(20)
```
kubukoz, line 62: Probably want to use `n` here

core/shared/src/test/scala/cats/effect/concurrent/SemaphoreTests.scala
```diff
@@ -41,6 +57,21 @@ class SemaphoreTests extends AsyncFunSuite with Matchers with EitherValues {
...
+        .flatMap { s =>
+          for {
+            _ <- s.acquire.replicateA(n).void
```
kubukoz, line 65: redundant void

core/shared/src/test/scala/cats/effect/concurrent/SemaphoreTests.scala
```diff
@@ -30,7 +30,23 @@ class SemaphoreTests extends AsyncFunSuite with Matchers with EitherValues {
   implicit val cs: ContextShift[IO] = IO.contextShift(executionContext)
   implicit val timer: Timer[IO] = IO.timer(executionContext)
 
+  private def lockSemaphore(n: Long, s: Semaphore[IO]): IO[Boolean] = {
```
kubukoz, line 33: I want to understand this method's purpose so that we can try to simplify it (and reduce some of the potential flakiness we have now due to `start` and `timeout`).

AFAICS it concurrently:
- runs an acquire
- loops `count` until there are `n` pending permits.

If that's the case, we can change it so that it concurrently:
- runs an acquire
- loops `count` until there's a negative amount of permits

and then we'll require that amount to be `n`, otherwise fail fast (instead of looping indefinitely with a timeout).

That would look like this:

```scala
private def lockSemaphore(n: Long, s: Semaphore[IO]): IO[Boolean] = {
  s.acquireN(n).background.use(_ => s.count.iterateUntil(_ < 0)).map(n === _)
}
```

No timeouts, so the only non-deterministic thing is how many `s.count` will need to be evaluated. But that's to be expected, as we can't know that upfront...

biscout24: Hey @kubukoz , thanks a lot for the feedback. Nice idea about `background` and `iterateUntil`, I have learned something new 👍 

I have tried that, but because of  `background` calls `cancel`, the result would be released semaphore. 

I reused your idea and the way `background` is implemented. Could you have another look 🙏 ?

biscout24:
@kubukoz , thanks for your feedback. I have lost somehow a notification form GH, sorry for the late reply. I have applied your ideas.

biscout24:
And it doesn't work for scalaJS 🤔

core/shared/src/test/scala/cats/effect/concurrent/SemaphoreTests.scala
kubukoz: I think we can return the count from `lockSemaphore` and cancel the acquire, otherwise it's going to run forever :)

I'm not sure why it hangs on JS...

biscout24: Okey, I see the point now. Let me try to re-structure the test.

biscout24: We can't return the count, because the idea of the test to check what happens with Semaphore while it has no available permits. 

I have updated the test to ensure cleaning resources. WDYT, @kubukoz ?

core/shared/src/test/scala/cats/effect/concurrent/SemaphoreTests.scala
```diff
@@ -30,7 +30,17 @@ class SemaphoreTests extends AsyncFunSuite with Matchers with EitherValues {
   implicit val cs: ContextShift[IO] = IO.contextShift(executionContext)
   implicit val timer: Timer[IO] = IO.timer(executionContext)
 
+  private def withLock[T](n: Long, s: Semaphore[IO])(test: Long => IO[T]): IO[T] =
```
kubukoz, line 33: Almost perfect, but the shape of taking an `A => IO[B]` function is pretty much Resource. How about we customize the one we have already and return it for use at the call site:

```scala
s.acquireN(n).background *> Resource.liftF(cs.shift *> s.count.iterateUntil(_ < 0))
```

and then call `withLock(n, s).use(...)`?

To be honest, since the long will never change and there's no actual resource backing it up, we might as well just return the long from `withLock` in an IO, with no resource (having closed the resource started by `background

biscout24: nice idea about `Resource` as a type for `withLock` 👍  . Done.

> To be honest, since the long will never change and there's no actual resource backing it up, we might as well just return the long from withLock in an IO, with no resource (having closed the resource started by `background

Not sure that will work 🤔 . The idea of the test is to do something with semaphore, while it is locked and be sure, that is is locked. In case we return IO[Long], what would happen in `flatMap` over this `IO`? I assume, our `background` will be cleaned-up and original lock will be cancelled, and actually semaphore will not be in acquired state any longer. WDYT?

biscout24:
@kubukoz , would you have another minute to take a look?

core/shared/src/test/scala/cats/effect/concurrent/SemaphoreTests.scala
kubukoz: It won't be in the state any longer, but the IO you pass to `use` will have completed, so we'll already know the value of the `Long`.

kubukoz: In fact, the only thing that'll change from the current version is the amount of times you write `use` - these things are equivalent:

```scala
private def withLock[T](n: Long, s: Semaphore[IO]): Resource[IO, Long] =
  s.acquireN(n).background *> Resource.liftF(cs.shift *> s.count.iterateUntil(_ < 0))

withLock(1, s).use { t =>
  s.available.map(t -> _)
}

//is the same as

private def withLock[T](n: Long, s: Semaphore[IO]): IO[(Long, Long)] =
  (s.acquireN(n).background *> Resource.liftF(cs.shift *> s.count.iterateUntil(_ < 0))).use(t => s.available.map(t -> _))

withLock(1, s)
```

Additionally:

```scala
(a *> Resource.liftF(b)).use(c)

//== (definition of *>)

a.flatMap { _ => Resource.liftF(b) }.use(c)

//== (inlining use into bracket)

a.use { _ => Resource.liftF(b).use(c) }

//== (inlining definition of Resource.liftF)

a.use { _ => b.flatMap(c)) }
```

if you now supply:

- `a` as `s.acquireN(n).background`
- `b` as `cs.shift *> s.count.iterateUntil(_ < 0)`, and
- `c` as `t => s.available.map(t -> _)` (or `t => s.available.tupleLeft(t)`)

you get this:

```scala
private def withLock[T](n: Long, s: Semaphore[IO]): IO[(Long, Long)] =
  s.acquireN(n).background.use { _ =>
    cs.shift *>
      s.count.iterateUntil(_ < 0).flatMap(t => s.available.tupleLeft(t))
  }
```

or even:

```scala
private def withLock[T](n: Long, s: Semaphore[IO]): IO[(Long, Long)] =
  s.acquireN(n).background.use { _ =>
    cs.shift *>
      (s.count.iterateUntil(_ < 0), s.available).tupled
  }
```

that's all from compiling in the air, so there might be some minor issues, but the idea should hold...

kubukoz: I don't really need this to happen though, so if you don't want to make any more changes to this I'm happy to merge given another maintainer's approval :)

biscout24: If we put `s.available` into it, imho it will work. We currently have two usages:
```
            res <- withLock(n, s).use { t =>
              s.count.map(t -> _)
            }
```

and 
```
            res <- withLock(1, s).use { t =>
              s.available.map(t -> _)
            }
```

So I passed it as additional parameter.

biscout24: imho that is matter of readability of the test idea. We have more concise structure now, and at the same time we put some effort at the reader to understand that `withLock` 
a) puts semaphore into locked state
b) executes some other actions
Ok for me though ;)

PR #858, opened 2020-05-03
Make Eithers in tryAcquire coherent

kubukoz:
Thanks to @biscout24 for highlighting the issue.

The two Eithers (old state, new state) were passed in a tuple, which could make a reader think all 4 combinations of states are possible. As it turns out, we only case about two of them, and two others (newState=left, oldState=right; newState=right, oldState=left) are impossible.

This PR encodes that in the types by building a single Either as the value returned from `modify`.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/concurrent/Semaphore.scala
+++ b/core/shared/src/main/scala/cats/effect/concurrent/Semaphore.scala
@@ -215,7 +215,7 @@ object Semaphore {
                   val newValue = m - n
 
                   (Right(newValue), Right((m, newValue)))
-                case w => (w, w)
+                case _ => (old, old)
               }
 
               (newState, previousAndNow)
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/cats/effect/concurrent/Semaphore.scala
```diff
@@ -210,22 +210,19 @@ object Semaphore {
...
+                  val newValue = m - n
+
+                  (Right(newValue), Right((m, newValue)))
+                case _ => (old, old)
               }
-              (u, (old, u))
+
+              (newState, previousAndNow)
             }
             .map {
```
biscout24, line 223: I don't think we need second map. How about this
https://github.com/typelevel/cats-effect/pull/857/files#diff-60779fbe95602b12729d28efd91cddfcR216
?

kubukoz: Yeah, that looks better, let's do that.

PR #859, opened 2020-05-03
Removed [A] from Concurrent#canceled

djspiewak:
There was really no need for it. By curry-howard, `A => A` is **True**, which is to say, `Unit`. This simplifies the signatures and avoids the need to constantly write `canceled(())`.

### DIFF
--- a/src/test/scala/ce3/laws/Generators.scala
+++ b/src/test/scala/ce3/laws/Generators.scala
@@ -83,7 +83,7 @@ object Generators {
     genPureConc[E, A](depth).map(pc => F[E].uncancelable(_ => pc))
 
   def genCanceled[E, A: Arbitrary]: Gen[PureConc[E, A]] =
-    arbitrary[A].map(F[E].canceled(_))
+    arbitrary[A].map(F[E].canceled.as(_))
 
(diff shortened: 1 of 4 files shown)


src/test/scala/ce3/laws/Generators.scala
```diff
@@ -83,7 +83,7 @@ object Generators {
     genPureConc[E, A](depth).map(pc => F[E].uncancelable(_ => pc))
 
   def genCanceled[E, A: Arbitrary]: Gen[PureConc[E, A]] =
```
kubukoz, line 85: Wouldn't this also be with Unit?

kubukoz (approved):

src/test/scala/ce3/laws/Generators.scala
djspiewak: The generators all return `A` so they can be folded into the big `choose`. This is like, the one place where `canceled` producing `F[Unit]` was a bit weird, because the generator now produces `canceled.as(a)`.

kubukoz: gotchu. I still wish there was a Nothing in the signature, but I guess Unit isn't that bad 😅

PR #861, opened 2020-05-03
Made onCase somewhat nicer to work with

djspiewak:
This also makes it more consistent with `ApplicativeError#onError`.

### DIFF
--- a/src/main/scala/ce3/playground.scala
+++ b/src/main/scala/ce3/playground.scala
@@ -222,11 +222,15 @@ object playground {
         }
 
       def onCancel[A](fa: PureConc[E, A], body: PureConc[E, Unit]): PureConc[E, A] =
-        onCase(fa, body)(Outcome.Canceled ==)
+        onCase(fa) { case Outcome.Canceled => body }
 
-      override def onCase[A](fa: PureConc[E, A], body: PureConc[E, Unit])(p: Outcome[PureConc[E, ?], E, A] => Boolean): PureConc[E, A] = {
+      override def onCase[A](
+          fa: PureConc[E, A])(
+          pf: PartialFunction[Outcome[PureConc[E, ?], E, A], PureConc[E, Unit]])
+          : PureConc[E, A] = {
+
         def pbody(oc: Outcome[PureConc[E, ?], E, A]) =    // ...and Sherman
-          if (p(oc)) body.attempt.void else unit
+          pf.lift(oc).map(_.attempt.void).getOrElse(unit)
 
         val finalizer: Finalizer[E] =
           ec => uncancelable(_ => pbody(ec) >> withCtx(_.self.popFinalizer))
--- a/src/main/scala/ce3/safe.scala
+++ b/src/main/scala/ce3/safe.scala
@@ -41,8 +41,8 @@ trait Bracket[F[_], E] extends Safe[F, E] {
       : F[B] =
     bracketCase(acquire)(use)((a, _) => release(a))
 
-  def onCase[A](fa: F[A], body: F[Unit])(p: Case[A] => Boolean): F[A] =
-    bracketCase(unit)(_ => fa)((_, c) => if (p(c)) body else unit)
+  def onCase[A](fa: F[A])(pf: PartialFunction[Case[A], F[Unit]]): F[A] =
+    bracketCase(unit)(_ => fa)((_, c) => pf.lift(c).getOrElse(unit))
 }
 
 object Bracket {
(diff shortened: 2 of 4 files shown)


src/main/scala/ce3/safe.scala
```diff
@@ -41,8 +41,8 @@ trait Bracket[F[_], E] extends Safe[F, E] {
       : F[B] =
     bracketCase(acquire)(use)((a, _) => release(a))
 
-  def onCase[A](fa: F[A], body: F[Unit])(p: Case[A] => Boolean): F[A] =
-    bracketCase(unit)(_ => fa)((_, c) => if (p(c)) body else unit)
+  def onCase[A](fa: F[A])(pf: PartialFunction[Case[A], F[Unit]]): F[A] =
+    bracketCase(unit)(_ => fa)((_, c) => pf.lift(c).getOrElse(unit))
```
kubukoz, line 45:
```suggestion
    bracketCase(unit)(_ => fa)((_, c) => pf.applyOrElse(c, unit))
```

src/main/scala/ce3/playground.scala
```diff
@@ -222,11 +222,15 @@ object playground {
         }
 
       def onCancel[A](fa: PureConc[E, A], body: PureConc[E, Unit]): PureConc[E, A] =
```
kubukoz, line 224: Since Concurrent knows about Outcome, could `onCancel` be implemented there?

src/test/scala/ce3/laws/BracketLaws.scala
```diff
@@ -44,8 +44,8 @@ trait BracketLaws[F[_], E] extends MonadErrorLaws[F, E] {
   def bracketBodyIdentity[A](fa: F[A]) =
     F.bracketCase(F.unit)(_ => fa)((_, _) => F.unit) <-> fa
 
-  def onCaseDefinedByBracketCase[A](fa: F[A], body: F[Unit], p: F.Case[A] => Boolean) =
-    F.onCase(fa, body)(p) <-> F.bracketCase(F.unit)(_ => fa)((_, c) => if (p(c)) body else F.unit)
+  def onCaseDefinedByBracketCase[A](fa: F[A], pf: PartialFunction[F.Case[A], F[Unit]]) =
+    F.onCase(fa)(pf) <-> F.bracketCase(F.unit)(_ => fa)((_, c) => pf.lift(c).getOrElse(F.unit))
```
kubukoz, line 48: also `applyOrElse` (if the other comment makes sense)

src/main/scala/ce3/playground.scala
```diff
@@ -222,11 +222,15 @@ object playground {
...
-        onCase(fa, body)(Outcome.Canceled ==)
+        onCase(fa) { case Outcome.Canceled => body }
+
+      override def onCase[A](
```
kubukoz, line 227: Actually, why do we need to override it?

kubukoz: Ah, I see, it's used in bracket :)

src/main/scala/ce3/safe.scala
djspiewak: TIL

src/main/scala/ce3/playground.scala
djspiewak: Ah but `Concurrent` *doesn't* know about `Bracket`. It can be implemented as implicit syntax via the companion though.

src/main/scala/ce3/playground.scala
djspiewak: Yeah this is another manifestation of the fact that `Bracket` would just be merged with `Concurrent` if `Region` didn't exist.

djspiewak:
Fiddled around with `applyOrElse`. The type inference is actually horrible because it isn't two parameter blocks, so it ends up being more bulky (IMO) than the `lift` + `getOrElse` approach. It does avoid the added boxing, which is somewhat nice, but I think that's more a concern for later.

src/main/scala/ce3/playground.scala
kubukoz: ah, makes sense. ConcurrentBracket, then?

kubukoz:
It's still not userland code, right? Won't hurt that much. It's just a minor anyway, so I'm fine with keeping that for later :)

kubukoz (approved):

src/main/scala/ce3/playground.scala
djspiewak: Well, `ConcurrentBracket` is an alias though, not a real type. The problem (maybe a problem, anyway) is that the split goes all the way down the hierarchy until you get to `Effect`/`Managed`. So for example, `Async` doesn't know about `Bracket` *either*, which is why `ConcurrentBracket` can't be a class.

kubukoz: hmm, I find it a bit strange that we have ConcurrentBracketLaws then. I was almost sure it's a trait... but I think I see what you're saying.

PR #864, opened 2020-05-07
Update scaladoc for Ref.getAndUpdate

jgogstad:
Make it explicit that it's the previous value that is returned. To me at least, it's ambiguous whether "the value that was updated" refers to the new updated value or the previous one.

```
@ val r = Ref.unsafe[IO, Int](0)
r: Ref[IO, Int] = cats.effect.concurrent.Ref$SyncRef@11eed657

@ r.getAndUpdate(_ + 1).unsafeRunSync
res0: Int = 0

@
```


kubukoz:
I don't know who wrote it that way... 👀 

oh, yes. It was me.

Thanks!

PR #868, opened 2020-05-11
Split Clock out from Temporal

djspiewak:
Here's a proposal for resolving some of the issues @Daenyth raised in Gitter.


Daenyth (approved):
👍

PR #871, opened 2020-05-16
Revise Bracket CE2 #1

RaasAhsan:
## Motivation
An implementation detail of `IO.bracket` today is that it leverages `IO.Async` to guarantee non-interruptibility during the `acquire` action.

The implementation can be roughly described as follows:
1. With an `IO.Async` node, invoke a new run-loop for the `acquire` action whose connection is divorced from the active connection.
2. In the callback for the `acquire` run-loop, invoke a second run-loop for the `use` and `release` actions. Use the original connection so that cancellation is respected again.

Critically, the `acquire` action is executed in a run-loop from which it is infeasible to collect any state or context from (run-loop callbacks have the type signature `Either[Throwable, A] => Unit`). This represents a limitation for several applications like [fiber tracing](https://github.com/typelevel/cats-effect/pull/854) and [fiber-local state](https://github.com/typelevel/cats-effect/pull/836), where there is a need for a stable context that must be threaded along with the execution of a fiber. 

The current workaround is to pass a thread-safe `IOContext` class to which all run-loops have a reference, so that changes are consistently propagated across threads. This is rather unideal because the `IOContext` class is confined to a single fiber and is manipulated only by the thread the fiber is bound to. Synchronization is only necessary because the fiber may encounter asynchronous boundaries.

So the goal here is really to revise `bracket` so it's easier to thread state along the execution of a fiber.

## Design constraints
1. The public API *must* remain backwards compatible.
2. The new implementation *must* preserve CE2 semantics. Satisfying the CE2 laws is necessary, but it's crucial that the new implementation exhibits the exact same behavior as before. Proving this exhaustively may be non-trivial.

## Proposed solution
I followed the most obvious solution first. The idea is to introduce a new instruction to the `IO` encoding called `Bracket`. An internal concept of "guards" is introduced to facilitate uninterruptible execution in `acquire` actions.

For context, there are several points at which cancellation is observed and respected (I might have missed some):
1. auto-cancellation
2. exit of an asynchronous boundary
3. after `acquire` completes and before `use` begins

The main issue here is that `acquire` must execute to completion, so that it can finish registration of the `release` finalizer. Guards are used to help facilitate this. The behavior of guards is described as follows:
1. Before entering a bracket acquire region, set guard flag to true.
2. After exiting a bracket acquire region, set guard flag to false.
3. Whenever cancellation is checked, the guard flag is also checked. If a fiber observes itself to be cancelled, but the guard flag is on, ignore cancellation and continue. Note that cancellers would backpressure on the deferred finalizer that is registered before `acquire` executes.

Additionally, `acquire` actions are made `uncancelable`. This ensures that asynchronous operations can't be cancelled, so they must resume the run-loop to complete the `release` registration.

### 2.0.0
```
[info] Benchmark                   (size)   Mode  Cnt    Score   Error  Units
[info] BracketBenchmark.happyPath   10000  thrpt    5  469.289 ± 4.529  ops/s
```
### PR
```
[info] Benchmark                   (size)   Mode  Cnt    Score   Error  Units
[info] BracketBenchmark.happyPath   10000  thrpt    5  351.519 ± 4.413  ops/s
```

So there's a 25% performance degradation here. I haven't looked into it, but I'm certain that it's because several additional `IO` nodes are being inserted into the tree to facilitate `Bracket`. If we inlined some of that code, we could trim it down quite a bit.

## Alternative implementations
1. Make the async callbacks fit the signature we want. This might necessitate an additional layer of indirection that would look ugly, but I think it would work.
2. Leverage the Java memory model. In particular, the memory model specifies that writes to a non-volatile variable by one thread are visible to reads of that variable in another thread, as long as there is a write by the first thread/read by the second thread to a volatile variable interleaved between the non-volatile accesses. These memory semantics hold for atomics as well. The runtime would have to be programmed to ensure that the volatile is read from or written to whenever exiting or entering an asynchronous boundary respectively. Obviously, this approach is ridiculously hard to reason about.
3. Cats Effect 3 will feature a [brand new interruption model](https://github.com/typelevel/cats-effect/issues/681) that can express `bracket` in a way that satisfies our requirements here.
4. Use `Promise` in `IOBracket` to pass back context from the invoked run-loop. This approach definitely has the least risk and effort involved. Maybe I should have written out this one first 😅 

I'm particularly interested in option 4, will probably open a PR for it this weekend. Let me know if there are other approaches worth considering.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -599,7 +599,8 @@ sealed abstract class IO[+A] extends internals.IOBinaryCompat[A] {
    *        (cancellation, error or successful result)
    */
   def bracketCase[B](use: A => IO[B])(release: (A, ExitCase[Throwable]) => IO[Unit]): IO[B] =
-    IOBracket(this)(use)(release)
+    Bracket(this, use, release)
+//    cats.effect.internals.IOBracket(this)(use)(release)
 
   /**
    * Executes the given `finalizer` when the source is finished,
@@ -658,7 +659,7 @@ sealed abstract class IO[+A] extends internals.IOBinaryCompat[A] {
    * @see [[bracketCase]] for the more general operation
    */
   def guaranteeCase(finalizer: ExitCase[Throwable] => IO[Unit]): IO[A] =
-    IOBracket.guaranteeCase(this, finalizer)
+    unit.bracketCase(_ => this)((_, e) => finalizer(e))
 
   /**
    * Replaces failures in this IO with an empty Option.
@@ -1297,7 +1298,7 @@ object IO extends IOInstances {
       // Race condition test — no need to execute `k` if it was already cancelled,
       // ensures that fiber.cancel will always wait for the finalizer if `k`
       // is executed — note that `isCanceled` is visible here due to `push`
-      if (!conn.isCanceled)
+      if (!conn.isCanceled || conn.isGuarded)
         ref.complete(
           try k(cb2)
           catch {
@@ -1614,6 +1615,13 @@ object IO extends IOInstances {
     modify: IOConnection => IOConnection,
     restore: (A, Throwable, IOConnection, IOConnection) => IOConnection
   ) extends IO[A]
+
+  /** Corresponds to [[IO.bracketCase]] */
+  final private[effect] case class Bracket[A, B](
+    acquire: IO[A],
+    use: A => IO[B],
+    release: (A, ExitCase[Throwable]) => IO[Unit]
+  ) extends IO[B]
 
   /* -=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-= */
 
--- a/core/shared/src/main/scala/cats/effect/internals/IOBracket.scala
+++ b/core/shared/src/main/scala/cats/effect/internals/IOBracket.scala
@@ -33,27 +33,29 @@ private[effect] object IOBracket {
     release: (Any, ExitCase[Throwable]) => IO[Unit]
   ) extends IOFrame[Any, IO[Any]] {
 
-    override def apply(a: Any): IO[Any] = {
-      conn.unguard()
+    override def apply(a: Any): IO[Any] =
+      IO.suspend {
+        conn.unguard()
 
-      val frame = new BracketReleaseFrame(a, release)
-      deferredRelease.complete(frame.cancel)
+        val frame = new BracketReleaseFrame(a, release)
+        deferredRelease.complete(frame.cancel)
 
-      if (!conn.isCanceled || conn.isGuarded) {
-        val onUse =
-          try use(a)
-          catch { case NonFatal(ex) => RaiseError(ex) }
-        onUse.flatMap(frame)
-      } else {
-        IO.unit
+        if (!conn.isCanceled) {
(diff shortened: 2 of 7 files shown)


core/shared/src/main/scala/cats/effect/internals/IOBracket.scala
```diff
@@ -16,154 +16,85 @@
...
+            try use(a)
+            catch { case NonFatal(ex) => RaiseError(ex) }
+          onUse.flatMap(frame)
+        } else {
+          IO.never
         }
+      }
 
-      case error @ Left(_) =>
+    override def recover(e: Throwable): IO[Any] =
```
RaasAhsan, line 53: Should we check cancellation here?

core/shared/src/main/scala/cats/effect/internals/IOBracket.scala
```diff
@@ -16,154 +16,85 @@
...
+      }
 
-  final private class EnsureReleaseFrame[A](releaseFn: ExitCase[Throwable] => IO[Unit])
-      extends BaseReleaseFrame[Unit, A] {
-    def release(c: ExitCase[Throwable]): CancelToken[IO] =
-      releaseFn(c)
   }
 
-  abstract private class BaseReleaseFrame[A, B] extends IOFrame[B, IO[B]] {
+  final private[effect] class BracketReleaseFrame(a: Any, releaseFn: (Any, ExitCase[Throwable]) => IO[Unit])
```
RaasAhsan, line 62: Can probably recover some type parameters here

core/shared/src/main/scala/cats/effect/internals/IOBracket.scala
```diff
@@ -16,154 +16,85 @@
...
 
-    def release(c: ExitCase[Throwable]): CancelToken[IO]
+    def release(c: ExitCase[Throwable]): CancelToken[IO] =
+      releaseFn(a, c)
 
     private def applyRelease(e: ExitCase[Throwable]): IO[Unit] =
       IO.suspend {
-        if (waitsForResult.compareAndSet(true, false))
+        if (waitsForResult.compareAndSet(true, false)) {
           release(e)
```
RaasAhsan, line 75: have to think about it more, but should this be protected by a try/catch...? thinking that an exception is thrown before we arrive at the `IO[Unit]`.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -658,7 +659,7 @@ sealed abstract class IO[+A] extends internals.IOBinaryCompat[A] {
    * @see [[bracketCase]] for the more general operation
    */
   def guaranteeCase(finalizer: ExitCase[Throwable] => IO[Unit]): IO[A] =
-    IOBracket.guaranteeCase(this, finalizer)
+    unit.bracketCase(_ => this)((_, e) => finalizer(e))
```
RaasAhsan, line 662: Expressing `guaranteeCase` in terms of `bracketCase`...

RaasAhsan:
Posted in the gitter chat:
> Ok, after sleeping on it a little bit, I think the above PR is pretty unnecessary. We already have a way to pass state back from a run-loop (the callback). So all we really need now is a way to capture the context we want to pass back in the IO, and zip it up with whatever the acquire is returning. I'll leave the PR open for now but I'm going to update the tracing PR, should be able to relax the thread safety of IOContext

PR #872, opened 2020-05-16
Add some motivation for explicit Blocker

ronanM:
Some explanation found in gitter like this one:
https://gitter.im/typelevel/cats-effect?at=5d87a3962438b53a64fa7a7c


core/shared/src/main/scala/cats/effect/Blocker.scala
```diff
@@ -28,7 +28,8 @@ import cats.effect.internals.BlockerPlatform
  * that require a special `ExecutionContext` for evaluation, while discouraging the
  * use of a shared, general purpose pool (e.g. the global context).
  *
- * Instances of this class should *not* be passed implicitly.
+ * Instances of this class should *NOT* be passed implicitly, because it hold state
```
djspiewak, line 31:
```suggestion
 * Instances of this class should *NOT* be passed implicitly because they hold state
```

Minor correction.

PR #873, opened 2020-05-18
All the laws!

djspiewak:
I got tired of waiting and just implemented the rest of the law infrastructure. As an aside, I'm still very unhappy with the `F[Boolean] => Prop` implicit being used for the invariant laws. Also note that we can and probably should add more laws in several of these, I just wanted to get the basic infrastructure in place.

### DIFF
--- a/src/main/scala/ce3/SyncManaged.scala
+++ b/src/main/scala/ce3/SyncManaged.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Daniel Spiewak
+ * Copyright 2020 Typelevel
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/main/scala/ce3/SyncManaged.scala
+++ b/src/main/scala/ce3/SyncManaged.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Typelevel
+ * Copyright 2020 Daniel Spiewak
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,18 +19,14 @@ package ce3
 import cats.~>
 import cats.implicits._
 
-trait SyncEffect[F[_]] extends Sync[F] with Bracket[F, Throwable] {
+trait SyncManaged[R[_[_], _], F[_]] extends Sync[R[F, ?]] with Region[R, F, Throwable] {
   type Case[A] = Either[Throwable, A]
 
   def CaseInstance = catsStdInstancesForEither[Throwable]
 
-  def to[G[_]]: PartiallyApplied[G] =
-    new PartiallyApplied[G]
+  def to[S[_[_], _]]: PartiallyApplied[S]
 
-  def toK[G[_]](implicit G: Sync[G] with Bracket[G, Throwable]): F ~> G
-
-  final class PartiallyApplied[G[_]] {
-    def apply[A](fa: F[A])(implicit G: Sync[G] with Bracket[G, Throwable]): G[A] =
-      toK[G](G)(fa)
+  trait PartiallyApplied[S[_[_], _]] {
+    def apply[A](rfa: R[F, A])(implicit S: Sync[S[F, ?]] with Region[S, F, Throwable]): S[F, A]
   }
 }
--- a/src/main/scala/ce3/package.scala
+++ b/src/main/scala/ce3/package.scala
@@ -17,7 +17,10 @@
 import cats.data.Kleisli
 
 package object ce3 {
-  type ConcurrentE[F[_]] = Concurrent[F, Throwable]
+  type BracketThrow[F[_]] = Bracket[F, Throwable]
+  type RegionThrow[R[_[_], _], F[_]] = Region[R, F, Throwable]
+
+  type ConcurrentThrow[F[_]] = Concurrent[F, Throwable]
 
   type ConcurrentBracket[F[_], E] = Concurrent[F, E] with Bracket[F, E]
 
@@ -31,7 +34,7 @@ package object ce3 {
     def apply[R[_[_], _], F[_], E](implicit R: ConcurrentRegion[R, F, E]): ConcurrentRegion[R, F, E] = R
   }
 
-  type TemporalE[F[_]] = Temporal[F, Throwable]
+  type TemporalThrow[F[_]] = Temporal[F, Throwable]
 
   type TemporalBracket[F[_], E] = Temporal[F, E] with Bracket[F, E]
 
--- a/src/main/scala/ce3/package.scala
+++ b/src/main/scala/ce3/package.scala
@@ -45,5 +45,17 @@ package object ce3 {
     def apply[R[_[_], _], F[_], E](implicit R: TemporalRegion[R, F, E]): TemporalRegion[R, F, E] = R
   }
 
+  type AsyncBracket[F[_]] = Async[F] with Bracket[F, Throwable]
+
+  object AsyncBracket {
+    def apply[F[_]](implicit F: AsyncBracket[F]): AsyncBracket[F] = F
+  }
+
+  type AsyncRegion[R[_[_], _], F[_]] = Async[R[F, ?]] with Region[R, F, Throwable]
+
+  object AsyncRegion {
+    def apply[R[_[_], _], F[_]](implicit R: AsyncRegion[R, F]): AsyncRegion[R, F] = R
+  }
+
   type TimeT[F[_], A] = Kleisli[F, Time, A]
 }
--- a/src/test/scala/ce3/laws/AsyncLaws.scala
+++ b/src/test/scala/ce3/laws/AsyncLaws.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Daniel Spiewak
+ * Copyright 2020 Typelevel
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/test/scala/ce3/laws/AsyncLaws.scala
+++ b/src/test/scala/ce3/laws/AsyncLaws.scala
@@ -46,11 +46,23 @@ trait AsyncLaws[F[_]] extends TemporalLaws[F, Throwable] with SyncLaws[F] {
   def neverIsDerivedFromAsync[A] =
     F.never[A] <-> F.async(_ => F.pure(None))
 
-  def evalOnLocalIdentity(ec: ExecutionContext) =
-    F.evalOn(F.executionContext, ec) <-> F.pure(ec)
+  def executionContextCommutivity[A](fa: F[A]) =
+    (fa *> F.executionContext) <-> (F.executionContext <* fa)
 
-  def evalOnLocalScope(ec: ExecutionContext) =
-    F.evalOn(F.unit, ec) >> F.executionContext <-> F.executionContext
+  def evalOnLocalPure(ec: ExecutionContext) =
+    F.evalOn(F.executionContext, ec) <-> F.evalOn(F.pure(ec), ec)
+
+  def evalOnPureIdentity[A](a: A, ec: ExecutionContext) =
+    F.evalOn(F.pure(a), ec) <-> F.pure(a)
+
+  def evalOnRaiseErrorIdentity(e: Throwable, ec: ExecutionContext) =
+    F.evalOn(F.raiseError[Unit](e), ec) <-> F.raiseError[Unit](e)
+
+  def evalOnCanceledIdentity(ec: ExecutionContext) =
+    F.evalOn(F.canceled, ec) <-> F.canceled
+
+  def evalOnNeverIdentity(ec: ExecutionContext) =
+    F.evalOn(F.never[Unit], ec) <-> F.never[Unit]
 }
 
 object AsyncLaws {
--- a/src/test/scala/ce3/laws/AsyncLaws.scala
+++ b/src/test/scala/ce3/laws/AsyncLaws.scala
@@ -46,7 +46,7 @@ trait AsyncLaws[F[_]] extends TemporalLaws[F, Throwable] with SyncLaws[F] {
   def neverIsDerivedFromAsync[A] =
     F.never[A] <-> F.async(_ => F.pure(None))
 
-  def executionContextCommutivity[A](fa: F[A]) =
+  def executionContextCommutativity[A](fa: F[A]) =
     (fa *> F.executionContext) <-> (F.executionContext <* fa)
 
   def evalOnLocalPure(ec: ExecutionContext) =
--- a/src/test/scala/ce3/laws/AsyncLaws.scala
+++ b/src/test/scala/ce3/laws/AsyncLaws.scala
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2020 Daniel Spiewak
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ce3
+package laws
+
+import cats.implicits._
+
+import scala.concurrent.ExecutionContext
+import scala.util.{Left, Right}
+
+trait AsyncLaws[F[_]] extends TemporalLaws[F, Throwable] with SyncLaws[F] {
+  implicit val F: Async[F]
+
+  def asyncRightIsPure[A](a: A) =
+    F.async[A](k => F.delay(k(Right(a))).as(None)) <-> F.pure(a)
+
+  def asyncLeftIsRaiseError[A](e: Throwable) =
+    F.async[A](k => F.delay(k(Left(e))).as(None)) <-> F.raiseError(e)
+
+  def asyncRepeatedCallbackIgnored[A](a: A) =
+    F.async[A](k => F.delay(k(Right(a))) >> F.delay(k(Right(a))).as(None)) <-> F.pure(a)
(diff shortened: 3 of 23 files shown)


kubukoz:
You don't make small PRs, do you... 😅

djspiewak:
> You don't make small PRs, do you... 😅

I cheat because I make all the new files, so really you should say "sbt-header doesn't make small PRs". ;-) *Nearly* all of this PR is pure boilerplate.

src/test/scala/ce3/laws/AsyncLaws.scala
```diff
@@ -0,0 +1,59 @@
...
+  def asyncCancelTokenIsSequencedOnCancel(fu: F[Unit]) =
+    F.start(F.async[Unit](_ => F.pure(Some(fu)))).flatMap(_.cancel) <-> fu.attempt.void
+
+  def neverIsDerivedFromAsync[A] =
+    F.never[A] <-> F.async(_ => F.pure(None))
+
+  def evalOnLocalIdentity(ec: ExecutionContext) =
+    F.evalOn(F.executionContext, ec) <-> F.pure(ec)
+
+  def evalOnLocalScope(ec: ExecutionContext) =
```
kubukoz, line 52: How about a law for nesting evalOn explicitly? I know it'll probably be covered by the generators but it's probably worth checking

```scala
evalOn(a) {
  evalOn(b)(unit) *> get
} <-> a.pure
```

```scala
evalOn(a) {
  evalOn(b)(get)
} <-> b.pure
```

djspiewak: Ah, this is illustrative of how I've been thinking about laws (rightly or wrongly) and makes for a very good discussion. To wit...

I really want CE3's laws to be substitutions, since that's the only sane definition of equality anyway. There are some caveats that I accept around that (for example, we assume that all rewrites are occurring within a neutral reference frame w.r.t. cancelability, since nesting things within `uncancelable` changes the rules), but by and large I think we've mostly achieved it.

What I mean by this is that a rule like this is, IMO, a good one:

```scala
F.uncancelable(_(fa)) <-> fa
```

You can compose such rules together. It composes with itself, for example:

```scala
F.uncancelable(_(F.uncancelable(_(fa)))) <-> fa
```

And with other laws:

```scala
F.start(F.uncancelable(_(F.never))).flatMap(_.join) <-> F.never
```

…and still more laws:

```scala
F.start(F.uncancelable(_(F.never >> fa))).flatMap(_.join) <-> F.never
```

And so on and so on. Remember, we have the full set of functor hierarchy laws to work with, too!

The *ideal* would be for all possible programs that can be described using Cats Effect to be generated by arbitrarily composing laws into larger and larger programs.

We will never achieve this ideal.

In practice there are two major problems with this:

- **Context dependence**. I already mentioned, but we in all current laws we have to assume that we're operating within a cancelable scope. Technically, we should have an entirely different set of laws for operating in an *uncancelable* scope. At present, we fake this by just tossing `uncancelable` around things. This can *probably* be made reasonably rigorous by adding an implication law which has `F.uncancelable(_ => faf) === F.uncancelable(_ => fbf)` IMPLIES `faf <-> fbf`, but we can't encode implications in our current law framework.
  
  There's an additional problem of context dependence with `evalOn`, but made even worse because there is no "neutral" `ExecutionContext` that we can deterministically assert, other than just asking for it with `executionContext`. You'll notice that your proposed law runs afoul of this, because of `fa` is `executionContext`, then neither law holds.
- **Short-Circuiting**. This is the most annoying thing ever. We have to go to significant contortions to deal with the fact that we can have `cancelled` and `raiseError` (to say nothing of `never`) within our generated `F[A]` values. This is a double-edged sword, because it makes `bracketCase` lawfully describable (since we can assert things about the `release` being run by just remembering that any `F[Unit]` might be `F.never`), but it also means that we have to do wonky things like this: `F.start(F.pure(a)).flatMap(_.join) <-> F.pure(Outcome.Completed(F.pure(a)))`. I hate the `F.pure(a)` in there; I really want to use an `fa`, but I can't.

Both of these constrain the generality of the laws in ways that I don't like.

So anyway, looping back to your proposed laws. I like the idea of building a law which talks about nesting, and I very much want to do it, but I'd like it to be framed as a generative property if at all possible. In practice, I've been applying the generative principle to avoid "unit test-like laws" which have "some complicated program" on the left and "a.pure" on the right, which are ultimately just boiling down to an assertion about result. Instead, I've tried to focus on inductive, generative rewrites which can be composed with other rewrites to generate successively more complex programs.

kubukoz: I see... although in my proposed laws I didn't aim for composing things together, but actually checking that the nesting doesn't have any quirks like the context being leaked to the outside. But I can't even see anything like that in ApplicativeLocal laws, so maybe it's not the right approach. It makes sense if we're avoiding specific cases (unit test-like laws, as you called them).

kubukoz: > There's an additional problem of context dependence with evalOn, but made even worse because there is no "neutral" ExecutionContext that we can deterministically assert, other than just asking for it with executionContext. You'll notice that your proposed law runs afoul of this, because of *fa is executionContext*, then neither law holds.

You mean if it's `evalOn`? That probably makes more sense.

djspiewak: > I see... although in my proposed laws I didn't aim for composing things together, but actually checking that the nesting doesn't have any quirks like the context being leaked to the outside. But I can't even see anything like that in ApplicativeLocal laws, so maybe it's not the right approach. It makes sense if we're avoiding specific cases (unit test-like laws, as you called them).

I think I did a bad job describing my thought. Basically, when we look at a law, we should ask the following question: **How many programs does this law allow me to rewrite into a more principle form?** If the answer is just "one", where "one" corresponds to "the left hand side of the equals", then it's a bad law. To be clear, your proposed law certainly does not correspond to "one", but I think a lot of the current CE2 laws do.

> You mean if it's evalOn? That probably makes more sense.

Ah, no I actually mean `executionContext.flatMap(f)` for some `f`. Quick example:

```scala

val fa = executionContext flatMap {
  case `b` => fb1
  case _ => fb2
}

evalOn(a) {
  evalOn(b)(fa) *> get
} <-> fa *> a.pure
```

This no longer holds, because the `fa` evaluated *outside* of the `evalOn` blocks will resolve to `fb2`, while inside it resolves to `fb1`.

Maybe a better law would be something that relates `pure` and `executionContext` more directly, which brings the applicative laws into play. For example:

```scala
F.evalOn(fa >> F.executionContext.flatMap(f), ec) <-> F.evalOn(fa >> f(ec), ec)
```

This gives us arbitrary nesting, since `f` may be an additional `evalOn`. Finally, we add some more cases:

```scala
F.evalOn(F.pure(a), ec) <-> F.pure(a)

F.evalOn(F.raiseError(), ec) <-> F.raiseError(e)

F.evalOn(F.canceled, ec) <-> F.canceled

F.evalOn(F.never, ec) <-> F.never
```

And now everything is *mostly* tied together, since we have a way of reducing `executionContext` to another constructor within `evalOn`, *and* we have a way of reducing `evalOn` of one of those constructors to one of those constructors. Given sufficient generators, this also neatly forbids the current `IO` implementation of `async`.

djspiewak: Oh, here's a good one:

```scala
  def executionContextCommutivity[A](fa: F[A]) =
    (fa *> F.executionContext) <-> (F.executionContext <* fa)
```

This would mean that `evalOn` is constrained to *only* affect its local scope, and cannot leak.

kubukoz: I'll take it :)

src/main/scala/ce3/SyncEffect.scala
```diff
@@ -17,7 +17,20 @@
...
+import cats.implicits._
 
-trait SyncEffect[F[_]] extends Sync[F] {
-  def to[G[_]: Sync]: F ~> G
+trait SyncEffect[F[_]] extends Sync[F] with Bracket[F, Throwable] {
+  type Case[A] = Either[Throwable, A]
+
+  def CaseInstance = catsStdInstancesForEither[Throwable]
+
+  def to[G[_]]: PartiallyApplied[G] =
```
kubukoz, line 27: Wondering if these wouldn't be better as implicit syntax...

src/main/scala/ce3/SyncManaged.scala
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ce3
+
+import cats.~>
+import cats.implicits._
+
+trait SyncManaged[R[_[_], _], F[_]] extends Sync[R[F, ?]] with Region[R, F, Throwable] {
```
kubukoz, line 22: Why didn't we pick SyncRegion again? It'd match ConcurrentRegion.

src/main/scala/ce3/package.scala
```diff
@@ -45,5 +48,17 @@ package object ce3 {
     def apply[R[_[_], _], F[_], E](implicit R: TemporalRegion[R, F, E]): TemporalRegion[R, F, E] = R
   }
 
+  type AsyncBracket[F[_]] = Async[F] with Bracket[F, Throwable]
```
kubukoz, line 51: So Sync and Async don't know they can be brackets... interesting.

src/test/scala/ce3/laws/AsyncLaws.scala
```diff
@@ -0,0 +1,71 @@
...
+  def asyncCancelTokenIsUnsequencedOnError[A](e: Throwable, fu: F[Unit]) =
+    F.async[A](k => F.delay(k(Left(e))) >> F.pure(Some(fu))) <-> F.raiseError(e)
+
+  def asyncCancelTokenIsSequencedOnCancel(fu: F[Unit]) =
+    F.start(F.async[Unit](_ => F.pure(Some(fu)))).flatMap(_.cancel) <-> fu.attempt.void
+
+  def neverIsDerivedFromAsync[A] =
+    F.never[A] <-> F.async(_ => F.pure(None))
+
+  def executionContextCommutivity[A](fa: F[A]) =
```
kubukoz, line 49: Probably typoed name

src/test/scala/ce3/laws/AsyncRegionLaws.scala
```diff
@@ -0,0 +1,40 @@
...
+}
+
+object AsyncRegionLaws {
+  def apply[
+      R[_[_], _],
+      F[_]](
+    implicit
+      F0: Async[R[F, ?]] with Region[R, F, Throwable],
+      B0: Bracket.Aux[F, Throwable, Outcome[R[F, ?], Throwable, ?]])
+      : AsyncRegionLaws[R, F] =
```
kubukoz, line 35: I have a weird feeling this will need a more concrete type, eventually... but that's fine for now

src/test/scala/ce3/laws/AsyncRegionTests.scala
```diff
@@ -0,0 +1,111 @@
...
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration.FiniteDuration
+
+trait AsyncRegionTests[R[_[_], _], F[_]] extends AsyncTests[R[F, ?]] with TemporalRegionTests[R, F, Throwable] {
+
+  val laws: AsyncRegionLaws[R, F]
+
+  def asyncRegion[A: Arbitrary: Eq, B: Arbitrary: Eq, C: Arbitrary: Eq](tolerance: FiniteDuration)(
+    implicit
+      ArbRFA: Arbitrary[R[F, A]],
```
kubukoz, line 37: I really wonder how you're debugging diverging expansions in these...

src/test/scala/ce3/laws/ConcurrentRegionTests.scala
```diff
@@ -0,0 +1,103 @@
...
+      feuaPP: R[F, Either[Unit, A]] => Pretty,
+      fouPP: R[F, Outcome[R[F, ?], E, Unit]] => Pretty)
+      : RuleSet = {
+
+    new RuleSet {
+      val name = "concurrent (region)"
+      val bases = Nil
+      val parents = Seq(concurrent[A, B, C], region[A, B, C])
+
+      val props = Seq()
```
kubukoz, line 87: just wondering, what's the benefit of having rulesets that only inherit from parents, instead of calling two rulesets in a row?

src/test/scala/ce3/laws/ManagedTests.scala
```diff
@@ -0,0 +1,111 @@
...
+      EqFEitherUEitherEA: Eq[R[F, Either[Unit, Either[Throwable, A]]]],
+      EqFOutcomeEA: Eq[R[F, Outcome[R[F, ?], Throwable, A]]],
+      EqFOutcomeEU: Eq[R[F, Outcome[R[F, ?], Throwable, Unit]]],
+      EqFABC: Eq[R[F, (A, B, C)]],
+      EqFInt: Eq[R[F, Int]],
+      OrdFFD: Order[R[F, FiniteDuration]],
+      GroupFD: Group[R[F, FiniteDuration]],
+      exec: R[F, Boolean] => Prop,
+      iso: Isomorphisms[R[F, ?]],
+      faPP: R[F, A] => Pretty,
```
kubukoz, line 80: The implicit-parameters-to-implementation ratio is killing me

src/test/scala/ce3/laws/SyncLaws.scala
```diff
@@ -0,0 +1,77 @@
...
+
+trait SyncLaws[F[_]] extends MonadErrorLaws[F, Throwable] with ClockLaws[F] {
+
+  implicit val F: Sync[F]
+
+  def delayValueIsPure[A](a: A) =
+    F.delay(a) <-> F.pure(a)
+
+  def delayThrowIsRaiseError[A](e: Throwable) =
+    F.delay[A](throw e) <-> F.raiseError(e)
```
kubukoz, line 32: Nice to finally see this as a law...

src/main/scala/ce3/SyncEffect.scala
kubukoz: of course something will need to stay as an abstract method ;)

src/main/scala/ce3/SyncManaged.scala
kubukoz: right, so SyncManaged is a limited version of Managed (without Async)?

kubukoz (approved):
Just some minor things, awesome stuff!

src/main/scala/ce3/SyncEffect.scala
djspiewak: Could be syntax. It's not particularly discoverable that way though. Either way, we need a mechanism for reordering where the implicit parameter block is.

src/main/scala/ce3/SyncManaged.scala
djspiewak: Yep! Just like `SyncEffect`.

src/main/scala/ce3/package.scala
djspiewak: Yep! This is how we can have an `Async` for `Resource`.

src/test/scala/ce3/laws/AsyncLaws.scala
djspiewak: I hate the word "commutativity". I always get it confused with "commute" and subsequently "commutivity", which just isn't a word.

src/test/scala/ce3/laws/AsyncRegionLaws.scala
djspiewak: Oh you mean rather than the aliases? Technically we *could* build little shim traits which are concrete variants of things like `AsyncRegion`, but I'm not sure what we would gain.

src/test/scala/ce3/laws/AsyncRegionTests.scala
djspiewak: Ah! That I can answer: I ignore the error. It seems that NSC just defaults to "diverging implicit expansion" whenever it's trying to resolve a zillion things and any of them fails. So I debug the errors by manually materializing things that I expect might be problematic until I find one that doesn't materialize, mostly because I haven't defined it. I've only found one case where I *really* had to help the compiler along (the `Tolerance` implicit).

src/test/scala/ce3/laws/ConcurrentRegionTests.scala
djspiewak: So we have a place to add more later. We have some special laws in `ConcurrentBracketLaws`, so I would expect that, at some point, we would have something similar in `ConcurrentRegionLaws`.

src/test/scala/ce3/laws/ManagedTests.scala
djspiewak: It's getting crazy!

src/test/scala/ce3/laws/SyncLaws.scala
djspiewak: I think this is actually a law in CE2, but it's really really nice to finally get down to this level of the hierarchy. Feels like things are starting to come together!

src/test/scala/ce3/laws/AsyncRegionLaws.scala
kubukoz: I mostly meant using more specific types (aux pattern or with the type members provided), as sometimes we need to know these types outside to pass some typeclass instances (this happened in Resource with region laws)

src/test/scala/ce3/laws/AsyncRegionTests.scala
kubukoz: Yeah, that's pretty much how I do it... and it hurts!

src/test/scala/ce3/laws/SyncLaws.scala
kubukoz: oh, I might be mistaking this with the law about catching in `map` - which is not there and probably still won't.

src/test/scala/ce3/laws/ConcurrentRegionTests.scala
kubukoz: cool... not sure if I mentioned this yet, but if we have laws for these combinations, shouldn't they also be traits? Kind of like `Monad` which has no abstract methods other than the ones from `FlatMap[F]` and `Applicative[F]`.

PR #874, opened 2020-05-19
Restructure the whole build and things

djspiewak:
Also swaps to GitHub Actions. Attention, @kubukoz, this will definitely cause conflicts.


project/Common.scala
```diff
@@ -0,0 +1,28 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import sbt._, Keys._
+
+object Common extends AutoPlugin {
```
kubukoz, line 19: interesting, I never realized you could do that...

kubukoz:
this kind of conflicts I don't mind ;)

project/Common.scala
djspiewak: It's a lot nicer than the whole `commonSettings` pattern that everyone seems to use. Honestly, one-off autoplugins are really really underused in the sbt ecosystem.

PR #877, opened 2020-05-19
Sync for FreeT Eval

djspiewak:
Depends on #876 (for sanity)

Capturing side-effects in `Eval`! What could go wrong?!

It's a really uncanny feeling, writing these magic side-effect capturing thingies for pre-existing types. It's like I'm imbuing them with some sort of mystical power that they did not previously possess.

My next step in this vein is to add a `Temporal[FreeT[S, F, ?], E]` where `Temporal[F, E]`. Once that's done, the next-next step is `Async[ContT[F, Unit, ?]]` where `Temporal[F, Throwable]` *and* `Sync[F]`, at which point we should just magically for free get the following instance:

```scala
Async[
  ContT[
    Kleisli[
      FreeT[
        Eval,
        Kleisli[
          Kleisli[
            FreeT[
              ThreadF,
              Kleisli[
                Outcome[Id, Throwable, ?],
                FiberCtx[Throwable],
                ?],
              ?],
            MVar.Universe,
            ?],
          Time,
          ?],
        ?],
      ExecutionContext,
      ?],
    Unit,
    ?]]
```

Or, with everything realiased:

```scala
Async[ContAsync[FreeSync[TimeT[PureConc[Throwable, ?], ?], ?], ?]]
```

Which I think is very, very cool. This will also form an `Effect`.

### DIFF
--- a/laws/src/test/scala/cats/effect/FreeSyncGenerators.scala
+++ b/laws/src/test/scala/cats/effect/FreeSyncGenerators.scala
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2020 Daniel Spiewak
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ce3
+package laws
+
+import cats.{Eval, Monad, MonadError}
+import cats.free.FreeT
+
+import playground._
+
+import org.scalacheck.{Arbitrary, Cogen, Gen}
+import org.scalacheck.util.Pretty
+
+import scala.concurrent.duration.FiniteDuration
+import java.util.concurrent.TimeUnit
+
+object FreeSyncGenerators {
+  import OutcomeGenerators._
+
+  implicit def cogenFreeSync[F[_]: Monad, A: Cogen](implicit C: Cogen[F[A]]): Cogen[FreeT[Eval, F, A]] =
+    C.contramap(runSync(_))
+
+  def generators[F[_]](implicit F0: MonadError[F, Throwable]) =
+    new SyncGenerators[FreeT[Eval, F, ?]] {
+
+      val arbitraryE: Arbitrary[Throwable] =
+        Arbitrary(Arbitrary.arbitrary[Int].map(TestException(_)))
+
+      val cogenE: Cogen[Throwable] =
+        Cogen[Int].contramap(_.asInstanceOf[TestException].i)
+
+      val arbitraryFD: Arbitrary[FiniteDuration] = {
+        import TimeUnit._
+
+        val genTU = Gen.oneOf(NANOSECONDS, MICROSECONDS, MILLISECONDS, SECONDS, MINUTES, HOURS, DAYS)
+
+        Arbitrary {
+          genTU flatMap { u =>
+            Gen.posNum[Long].map(FiniteDuration(_, u))
+          }
+        }
+      }
+
+      val F: Sync[FreeT[Eval, F, ?]] =
+        syncForFreeT[F]
+    }
+
+  implicit def arbitraryFreeSync[F[_], A: Arbitrary: Cogen](
+      implicit F: MonadError[F, Throwable])
+      : Arbitrary[FreeT[Eval, F, A]] =
+    Arbitrary(generators[F].generators[A])
+}
+
+final case class TestException(i: Int) extends Exception
--- a/laws/src/test/scala/cats/effect/FreeSyncGenerators.scala
+++ b/laws/src/test/scala/cats/effect/FreeSyncGenerators.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Daniel Spiewak
+ * Copyright 2020 Typelevel
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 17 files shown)


laws/src/test/scala/cats/effect/FreeSyncGenerators.scala
```diff
@@ -0,0 +1,66 @@
...
+      val arbitraryE: Arbitrary[Throwable] =
+        Arbitrary(Arbitrary.arbitrary[Int].map(TestException(_)))
+
+      val cogenE: Cogen[Throwable] =
+        Cogen[Int].contramap(_.asInstanceOf[TestException].i)
+
+      val arbitraryFD: Arbitrary[FiniteDuration] = {
+        import TimeUnit._
+
+        val genTU = Gen.oneOf(NANOSECONDS, MICROSECONDS, MILLISECONDS, SECONDS, MINUTES, HOURS, DAYS)
```
kubukoz, line 47: I've seen this somewhere already, so I suppose we could extract it

laws/src/test/scala/cats/effect/FreeSyncSpec.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  implicit val eqThrowable: Eq[Throwable] =
+    Eq.fromUniversalEquals
+
+  implicit def exec(sbool: FreeEitherSync[Boolean]): Prop =
+    run(sbool).fold(Prop.exception(_), b => if (b) Prop.proved else Prop.falsified)
+
+  def beEqv[A: Eq: Show](expect: A): Matcher[A] = be_===[A](expect)
+
+  def be_===[A: Eq: Show](expect: A): Matcher[A] = (result: A) =>
```
kubukoz, line 48: Are these matchers actually used somewhere?

laws/src/test/scala/cats/effect/FreeSyncSpec.scala
```diff
@@ -0,0 +1,58 @@
...
+    (result === expect, s"${result.show} === ${expect.show}", s"${result.show} !== ${expect.show}")
+
+  Eq[Either[Throwable, Int]]
+
+  cats.Invariant[FreeEitherSync]
```
kubukoz, line 53:
```suggestion
```

laws/src/test/scala/cats/effect/Generators.scala
```diff
@@ -122,7 +124,36 @@ trait ApplicativeErrorGenerators[F[_], E] extends ApplicativeGenerators[F] {
...
+
+  private def genRealTime[A: Arbitrary] =
+    arbitrary[A].map(F.realTime.as(_))
+}
+
+trait SyncGenerators[F[_]] extends MonadErrorGenerators[F, Throwable] with ClockGenerators[F] {
+  implicit val F: Sync[F]
+
+  override protected def baseGen[A: Arbitrary: Cogen] =
+    ("delay" -> arbitrary[A].map(F.delay(_))) :: super.baseGen[A]
```
kubukoz, line 153: Don't you want to generate actual side effects? Although I don't know how exactly that would work, this is essentially `pure` and we already have that.

laws/src/test/scala/cats/effect/Generators.scala
```diff
@@ -202,6 +233,48 @@ trait ConcurrentGenerators[F[_], E] extends ApplicativeErrorGenerators[F, E] {
...
+      "evalOn" -> genEvalOn[A](deeper)) ++ super.recursiveGen[A](deeper)
+
+  private def genAsync[A: Arbitrary: Cogen](deeper: GenK[F]) =
+    for {
+      result <- arbitrary[Either[Throwable, A]]
+
+      fo <- deeper[Option[F[Unit]]](
+        Arbitrary(
+          Gen.option[F[Unit]](
+            deeper[Unit](Arbitrary(arbitrary[Unit]), Cogen.cogenUnit))),
```
kubukoz, line 267:
```suggestion
            deeper[Unit],
```

please tell me this just works ;)

laws/src/test/scala/cats/effect/Generators.scala
```diff
@@ -202,6 +233,48 @@ trait ConcurrentGenerators[F[_], E] extends ApplicativeErrorGenerators[F, E] {
...
+        Cogen.cogenOption(cogenFU))
+    } yield F.async[A](k => F.delay(k(result)) >> fo)
```
kubukoz, line 269: Unless I'm missing something, this completes immediately and `fo` never has a chance to be sequenced

laws/src/test/scala/cats/effect/FreeSyncSpec.scala
djspiewak: Nope. Just random pasta

laws/src/test/scala/cats/effect/Generators.scala
djspiewak: Well, I've been thinking about that a lot. I really *do* want to generate side-effects, but I kinda can't. Or rather, I *can* generate them, but I can't observe them.

We have this same problem with fibers. So when we generate a `start`… where does it go? Really what we want to be able to do is generate a DAG, where the `start` node forks the path and things come back together at `join` and everything in the middle is also generated (or alternatively, you could `start` and just ignore it). But… we can't really do that directly.

We have the same problem with side-effects: there's no point in generating a side-effect if you can't later come back and observe it. But that then requires generating a *pair* of effects, with stuff in between, and some shared mutable place that the side-effects can be represented without stepping on each other (or maybe *with* stepping on each other). It's kind of a mess.

So for now it's just the `pure` generator.

laws/src/test/scala/cats/effect/Generators.scala
djspiewak: I don't think it does! :-( I'll check though

laws/src/test/scala/cats/effect/Generators.scala
djspiewak: It does complete immediately, though that reminds me that we need to have a law which shows how that sequencing is intended to be handled (tldr, `fo` should still be sequenced).

PR #886, opened 2020-05-31
Eliminate boxing in IORunLoop.step

RaasAhsan:
In the current `IORunLoop.step` implementation, the compiler boxes `currentIO`, `bFirst`, `bRest` and `unboxed` in an `ObjectRef` because they are all mutable variables that are captured in a lambda. This degrades performance for certain programs, particularly all `SyncIO` programs and all `IO` programs before they encounter the first asynchronous boundary (that are run with `unsafeRunSync`).

I only had enough patience to run the `DeepBindBenchmark` and `ShallowBindBenchmark`, whose results are shown below. Benchmarks were run on a MacBook Pro 15" 2015. Most of the other benchmarks probably see improved performance as well, but I talk more about that later.

## 2.0.0 results
```
[info] Benchmark                   (size)   Mode  Cnt     Score    Error  Units
[info] ShallowBindBenchmark.async   10000  thrpt    5   120.309 ± 11.319  ops/s
[info] ShallowBindBenchmark.delay   10000  thrpt    5  3859.484 ± 25.209  ops/s
[info] ShallowBindBenchmark.pure    10000  thrpt    5  4612.369 ± 58.177  ops/s

[info] DeepBindBenchmark.async    3000  thrpt    5   336.482 ±  12.969  ops/s
[info] DeepBindBenchmark.delay    3000  thrpt    5  5312.525 ± 133.527  ops/s
[info] DeepBindBenchmark.pure     3000  thrpt    5  5748.930 ± 131.806  ops/s
```

## PR results
```
[info] Benchmark                   (size)   Mode  Cnt     Score     Error  Units
[info] ShallowBindBenchmark.async   10000  thrpt    5   119.277 ±  15.213  ops/s
[info] ShallowBindBenchmark.delay   10000  thrpt    5  5049.250 ±  51.853  ops/s
[info] ShallowBindBenchmark.pure    10000  thrpt    5  5696.337 ± 339.613  ops/s

[info] Benchmark                (size)   Mode  Cnt     Score    Error  Units
[info] DeepBindBenchmark.async    3000  thrpt    5   338.091 ± 34.497  ops/s
[info] DeepBindBenchmark.delay    3000  thrpt    5  6907.871 ± 70.006  ops/s
[info] DeepBindBenchmark.pure     3000  thrpt    5  7481.769 ± 95.016  ops/s
```

So there's quite a notable improvement, around 19-23% based on these benchmarks.

I realized that there's a bit of a blind spot in the benchmarks in master. `IORunLoop` includes two separate run-loop implementations, one that supports asynchronous evaluation and one that doesn't. Most of the benchmarks just invoke the synchronous run-loop, but in reality, most useful programs are shifted onto the asynchronous run-loop. Both run-loops likely observe different performance characteristics (the corresponding bytecode for each method are quite different). So it may be worth filling those gaps in to provide a more "realistic" benchmark.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/internals/IORunLoop.scala
+++ b/core/shared/src/main/scala/cats/effect/internals/IORunLoop.scala
@@ -224,24 +224,21 @@ private[effect] object IORunLoop {
           bFirst = bindNext.asInstanceOf[Bind]
           currentIO = fa
 
-        case Async(_, _) =>
+        case _ =>
           // Cannot inline the code of this method — as it would
           // box those vars in scala.runtime.ObjectRef!
-          return suspendAsync(currentIO.asInstanceOf[IO.Async[A]], bFirst, bRest)
-        case _ =>
-          return Async { (conn, cb) =>
-            loop(currentIO, conn, cb.asInstanceOf[Callback], null, bFirst, bRest)
-          }
+          return suspendAsync(currentIO.asInstanceOf[IO[A]], bFirst, bRest)
       }
 
       if (hasUnboxed) {
+        val _unboxed = unboxed
         popNextBind(bFirst, bRest) match {
           case null =>
-            return (if (currentIO ne null) currentIO else Pure(unboxed))
+            return (if (currentIO ne null) currentIO else Pure(_unboxed))
               .asInstanceOf[IO[A]]
           case bind =>
             currentIO =
-              try bind(unboxed)
+              try bind(_unboxed)
               catch { case NonFatal(ex) => RaiseError(ex) }
             hasUnboxed = false
             unboxed = null
@@ -255,10 +252,11 @@ private[effect] object IORunLoop {
     // $COVERAGE-ON$
   }
 
-  private def suspendAsync[A](currentIO: IO.Async[A], bFirst: Bind, bRest: CallStack): IO[A] =
-    // Hitting an async boundary means we have to stop, however
-    // if we had previous `flatMap` operations then we need to resume
-    // the loop with the collected stack
+  private def suspendAsync[A](currentIO: IO[A], bFirst: Bind, bRest: CallStack): IO[A] =
+    // Encountered an instruction that can't be interpreted synchronously,
+    // so suspend it in an Async node that can be invoked later.
+    // If we had previous `flatMap` operations then we need to resume
+    // the loop with the collected stack.
     if (bFirst != null || (bRest != null && !bRest.isEmpty))
       Async { (conn, cb) =>
         loop(currentIO, conn, cb.asInstanceOf[Callback], null, bFirst, bRest)
--- a/core/shared/src/main/scala/cats/effect/internals/IORunLoop.scala
+++ b/core/shared/src/main/scala/cats/effect/internals/IORunLoop.scala
@@ -231,18 +231,18 @@ private[effect] object IORunLoop {
       }
 
       if (hasUnboxed) {
-        val _unboxed = unboxed
         popNextBind(bFirst, bRest) match {
           case null =>
-            return (if (currentIO ne null) currentIO else Pure(_unboxed))
+            return (if (currentIO ne null) currentIO else Pure(unboxed))
               .asInstanceOf[IO[A]]
           case bind =>
-            currentIO =
-              try bind(_unboxed)
+            val fa =
+              try bind(unboxed)
               catch { case NonFatal(ex) => RaiseError(ex) }
             hasUnboxed = false
             unboxed = null
             bFirst = null
+            currentIO = fa
         }
       }
       true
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/cats/effect/internals/IORunLoop.scala
```diff
@@ -255,16 +252,12 @@ private[effect] object IORunLoop {
...
-    if (bFirst != null || (bRest != null && !bRest.isEmpty))
-      Async { (conn, cb) =>
-        loop(currentIO, conn, cb.asInstanceOf[Callback], null, bFirst, bRest)
-      }
-    else
-      currentIO
+  private def suspendAsync[A](currentIO: IO[A], bFirst: Bind, bRest: CallStack): IO[A] =
+    // Encountered an instruction that can't be interpreted synchronously,
+    // so suspend it in an Async node that can be invoked later.
+    Async { (conn, cb) =>
```
RaasAhsan, line 258: I removed the bind stack check so all asynchronous run-loop instructions are just wrapped in an `Async` node, but that means there may be an extra redundant `Async` node introduced.

Alternatively, we can create another method to suspend non-Async nodes (basically `ContextSwitch` right now), or just re-assign the mutable variables to an immutable value directly in the run-loop.

Avasil:
Nice spot, I did benchmarks when releasing the first version of `monix-bio` a while ago and differences in `IO` and `Task` flatMap seemed weird to me: https://github.com/monix/monix-bio/blob/master/benchmarks/results/10-12-2019.md so that could be it.

core/shared/src/main/scala/cats/effect/internals/IORunLoop.scala
```diff
@@ -224,24 +224,21 @@ private[effect] object IORunLoop {
...
-          return suspendAsync(currentIO.asInstanceOf[IO.Async[A]], bFirst, bRest)
-        case _ =>
-          return Async { (conn, cb) =>
-            loop(currentIO, conn, cb.asInstanceOf[Callback], null, bFirst, bRest)
-          }
+          return suspendAsync(currentIO.asInstanceOf[IO[A]], bFirst, bRest)
       }
 
       if (hasUnboxed) {
+        val _unboxed = unboxed
```
Avasil, line 234: Do you mind explaining how this line helps and how do you check it?

RaasAhsan: So if you take a look at the bytecode or the decompiled source, you'll notice that `unboxed` is ironically boxed in an `ObjectRef`. Unlike the others, it isn't being closed on in a lambda. Without any good reason, I originally thought that it was the `bind(unboxed)` call that was forcing the boxing, but it turns out that it's the `try` statement. So, mutable var references in a `try` block forces boxing, but for what reason I still don't understand yet.

RaasAhsan: This blog post describes the rationale in more detail: https://james-iry.blogspot.com/2013/01/scala-trycatch-lifting-proposal.html

Avasil: Thank you!

alexandru: Shit, I actually knew this could happen and I think I fixed it before, somewhere, maybe in Monix 🙂

Avasil: It's not fixed in Monix Task (although I recall hearing this specific issue from somewhere, perhaps Observable or Iterant):

https://github.com/monix/monix/blob/master/monix-eval/shared/src/main/scala/monix/eval/internal/TaskRunLoop.scala#L550

but the other "issue" might not be present

alexandru: @Avasil but that's not a closure. The line you highlighted, why would it cause problems?

Avasil: It just looks similar to the one changed here, maybe it's OK
I will check it later

RaasAhsan: Ah so the boxing only happens with try expressions since those are compiled to a separate method, but it sounds like regular try/catch block statements aren't.

RaasAhsan: This is kind of odd because the asynchronous run-loop uses a `try` expression as well but there's no boxing:
```scala
if (hasUnboxed) {
        popNextBind(bFirst, bRest) match {
          case null =>
            cb(Right(unboxed))
            return
          case bind =>
            val fa =
              try bind(unboxed)
              catch { case NonFatal(ex) => RaiseError(ex) }
            hasUnboxed = false
            unboxed = null
            bFirst = null
            currentIO = fa
        }
      }
```
The only difference is that its assigning the expression to a new `fa` rather than `currentIO`

SystemFw: > So, mutable var references in a try block forces boxing, but for what reason I still don't understand yet.

Good catch! Someone should write all this folklore somewhere :) 
Most aren't necessarily deterministic, but a compilation of hypotheses you need to check is still better than nothing

PR #891, opened 2020-06-09
Fix small mistake in Readme for ce3

mgibowski:
>  When ce3 was originally being designed, a decision was made [...] not automatically shift to a thread pool.

I believe the reference here is to Cats Effect and not ce3, right?


djspiewak:
Ah nuts! My mistake. Thank you!

PR #894, opened 2020-06-14
Inductive bracket instances

### LINKED ISSUE #860: CE 3: Attempting to Make Bracket More Composable
So there's a bit of a problem with `bracket` in master (and how it is currently defined in ce3): namely, inductive instances with zeros. The problem is nicely illustrated by the `Bracket` formed by `EitherT[F` where `F` forms a `Bracket`. Specifically, when we sequence the `release` action (which is of type `A => EitherT[F, E, Unit]` within the `Bracket[F`, we need to produce a function of type `A => F[Unit]`. The only way to do this is something like `_.andThen(_.value.void)`, which *works* but has the effect of dropping any `E` sequenced within the `EitherT`. This in turn is inconsistent with `ApplicativeError`, which is a problem.

In master, we resolve this by not having inductive instances for `Bracket`, and instead waiting until we can get to `Sync` and using a `Ref` to pass the error around. Which is… not great. I think we can do better.

If you look at `ApplicativeError`, you'll notice that the equivalent of `bracket` is `onError`, which in turn is implemented in terms of `handleErrorWith`, which has the signature `F[A] => (E => F[A]) => F[A]`. Critically, the handler action has the ability to pass a non-`Unit` value out into the context. When the handler simply wants to re-raise the error (as in the case of `onError`), then `void` is fine. However, the fact that `handleErrorWith` is the primitive means that inductive `ApplicativeError` instances have the ability to sequence state (in the case of `EitherT`, this would be the `Either`) from the handler back "outside" where it can be re-wrapped into the inductive structure.

Unfortunately, doing this with `bracket` is a lot more complicated, since we don't have any place to put state when we're in the `Outcome.Canceled` case. Nevertheless, I think the same trick can be applied:

```scala
  def bracketCase[A, B](
      acquire: F[A])(
      use: A => F[B])(
      release: (A, Case[B]) => F[B])
      : F[B]
```

Note that this also makes `bracket` considerably easier to test.

*However* the moment we do this, we must re-confront the problem of ignoring errors in handlers. In other words, what value should this have?

```scala
F.bracket(F.unit)(_ => F.unit)(_ => F.raiseError(e))
```

Should this be `F.unit`, or `F.raiseError(e)`? I'm starting to think the *correct* answer is the latter. This is the answer which is coherent with `handleErrorWith`, and it allows us to avoid all of this `Ref` juggling. Now, to be clear, errors would still get dropped in the event of cancelation, because there's simply no place for them to go (except an implementation-specific side-channel), but that's the nature of concurrent preemption.

Thoughts?

djspiewak:
Experimenting with inductive `Bracket` instances for `OptionT`.

Fixes #860

### DIFF
--- a/laws/src/test/scala/cats/effect/BracketSpec.scala
+++ b/laws/src/test/scala/cats/effect/BracketSpec.scala
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2020 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+
+import cats.data.OptionT
+import cats.effect.laws.BracketTests
+import cats.implicits._
+import cats.laws.discipline.arbitrary._
+
+import org.specs2.mutable.Specification
+
+import org.typelevel.discipline.specs2.mutable.Discipline
+
+class BracketSpec extends Specification with Discipline {
+
+  checkAll(
+    "Either[Int, *]",
+    BracketTests[Either[Int, *], Int].bracket[Int, Int, Int])
+
+  checkAll(
+    "OptionT[Either[Int, *], *]",
+    BracketTests[OptionT[Either[Int, *], *], Int].bracket[Int, Int, Int])
+}
--- a/core/src/main/scala/cats/effect/Sync.scala
+++ b/core/src/main/scala/cats/effect/Sync.scala
@@ -17,6 +17,11 @@
 package cats.effect
 
 import cats.{Defer, MonadError}
+import cats.data.OptionT
+
+import scala.concurrent.duration.FiniteDuration
+
+import java.time.Instant
 
 trait Sync[F[_]] extends MonadError[F, Throwable] with Clock[F] with Defer[F] {
   def delay[A](thunk: => A): F[A]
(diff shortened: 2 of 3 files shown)


cosmir17 (approved):
I attended Daniel's twitch session as 'ks1537'. I am not a member of the maintainers, so ignore my approval. It's a gesture of appreciation of Daniel's professional kindness of his knowledge sharing.

laws/src/test/scala/cats/effect/BracketSpec.scala
```diff
@@ -0,0 +1,37 @@
...
+
+class BracketSpec extends Specification with Discipline {
+
+  checkAll(
+    "Either[Int, *]",
+    BracketTests[Either[Int, *], Int].bracket[Int, Int, Int])
+
+  checkAll(
+    "OptionT[Either[Int, *], *]",
+    BracketTests[OptionT[Either[Int, *], *], Int].bracket[Int, Int, Int])
```
RaasAhsan, line 36: Is there more coming for an inductive `EitherT` instance?

djspiewak: At present, this is just a demonstration of what's possible. Further inductive instances will definitely be added.

As an aside, there are some problems with this approach. For example:

```scala
def foo[F[_], E, A](fa: F[A])(implicit F: Bracket[F, E]) =
  Bracket[OptionT[F, *], E].onCase(fa) { 
    case cs => ???
  }

def bar[F[_], E, A](fa: F[A])(implicit F: ConcurrentBracket[F, E]) =
  Bracket[OptionT[F, *], E].onCase(fa) { 
    case cs => ???
  }
```

The types of `cs` in `foo` and `bar` are not even remotely related. In `foo`, `cs` will have type `OptionT[F.Case, A]`, while in `bar`, `cs` will have type `Outcome[OptionT[F, *], E, A]`. I'm trying to decide whether or not users will be confused by this.

RaasAhsan: Interesting. Past comprehension, it's still possible for `F[_]` to be cancelled, even if you only have a `Bracket` class instance in scope, right? It seems like `Bracket` wouldn't let you *observe* cancellation, but it can still happen.

djspiewak: > It seems like Bracket wouldn't let you observe cancellation, but it can still happen.

Actually it'll still let you observe cancelation, it's just going to get wrapped up a bit differently. So to be clear, the `F.Case` existential type constructor is actually `Outcome[F, E, A]` if we assume that `foo` and `bar` can be called from the same call-site. So it's basically up to the type signatures to figure out whether or not you want the capability to observe cancelation *and* usefully act on it.

So specifically, you can constrain `Case` to be something:

```scala
def foo[F[_], C[_], E, A](fa: F[A])(implicit F: Bracket.Aux[F, E, C], C: ApplicativeError[C, E]) = ???
```

Okay, that's a bit of a gimme, since that same `ApplicativeError` instance will be available from `F.CaseInstance`, but you can further constrain things if you really want to. Like, for example:

```scala
def foo[F[_], C[_], E, A](fa: F[A])(implicit F: Bracket.Aux[F, E, C], C: ApplicativeError[C, E], filter: FunctorFilter[C]) = ???
```

Now `C` is very likely to characterize something a lot like `Outcome`, though we can't necessarily *observe* it since we have no way of extracting cases, given these signatures, but you see where I'm going with this. Basically, if you need to be able to do something with cancelation *specifically* in the body of `foo` (as opposed to just responding generically to an opaque existential `Case`), then you need to define what `Case` is in the constraints.

PR #896, opened 2020-06-16
Syntax classes

TimWSpence:


TimWSpence:
@djspiewak as we discussed :)

.scalafmt.conf
```diff
@@ -0,0 +1,12 @@
...
+align.openParenCallSite = true
+align.openParenDefnSite = true
+maxColumn = 120
+continuationIndent.defnSite = 2
+assumeStandardLibraryStripMargin = true
+danglingParentheses = true
+rewrite.rules = [AvoidInfix, SortImports, RedundantBraces, RedundantParens, SortModifiers]
+docstrings = JavaDoc
+lineEndings=preserve
+newlines.afterCurlyLambda = preserve
```
alexandru, line 12: Adding a `.scalafmt.conf` to this project would be nice, but not part of an unrelated PR.

Also if we add it, then it needs to be integrated in the build, so a reformatting of all files, plus agreement on the formatting style is needed.

djspiewak:
@alexandru We can't use Simulacrum anymore (in ce3) because of Dotty. This PR is specifically to the ce3 branch, which doesn't have any syntax classes.

core/src/main/scala/cats/effect/syntax/package.scala
```diff
@@ -16,10 +16,6 @@
 
 package cats.effect.syntax
 
-package object syntax {
-
-  object all extends AllSyntax
-  object async extends AsyncSyntax
-  object concurrent extends ConcurrentSyntax
+package object syntax extends AllSyntax {
```
djspiewak: Okay, I take it back, I think I like having the `all`/`async`/`concurrent`/etc split better than tossing everything into the package object. :-) Creating cycles via package object implicits just seems really scary, and we may as well stay consistent with ce2 when we *can*.

Beyond that though, I'm 👍 for merge.

LukaJCB:
We could think about using scalafix-simulacrum?

djspiewak:
I'm going to merge this now to avoid creating conflicts (particularly with #900). We can revise and address things along the way.

I do think that we should probably consider scalafix-simulacrum, and I think we should break the cyclic package object dependency, but one thing at a time. :-)

PR #900, opened 2020-06-20
ScalaJS Cross Build

### LINKED ISSUE #895: Replacement for Instant in realTime?
`java.time.Instant` doesn't exist on ScalaJS. The https://github.com/scala-js/scala-js-java-time project is intended to address this limitation, but a) it's another dependency, and b) it's not clear to me how mature it really is or if it's intended for this type of use-case (a foundational ecosystem library). Would appreciate some insight on this from @sjrd if he has a spare moment.

If we *don't* use `Instant`, we could move a little further afield. Obviously there's always `Long`, but if we're going to do that, we may as well just use `FiniteDuration`. @rossabaker had some strong opinions on this if I recall.

Floor is open. Anyone have thoughts or opinions on the appropriate signature here? To be clear, the *current* signature is as follows (which doesn't link on ScalaJS):

```scala
def realTime: F[Instant]
```

djspiewak:
Had to change `Instant` to `FiniteDuration` for compatibility. Also some upgrades to sbt-spiewak to make this possible. This PR is *definitely* going to cause merge conflicts out the wazoo, which kinda sucks. Heads up, @TimWSpence and @kubukoz.

Fixes #895


djspiewak:
As an FYI for those following along at home… My next planned (and probably final) bit of merge conflict generation will be to move the typeclasses out into their own project (probably something like cats-effect-kernel, following cats' convention). `IO` will live in core, which will depend upon kernel. The laws will similarly depend on kernel. The classes themselves will get forwarders for their types and companions in the `cats.effect` package object, which will live in core, keeping syntax and imports nice.

I'm not *entirely* sure where `TimeT` should live, but we can figure that out later.

core/shared/src/main/scala/cats/effect/TimeT.scala
```diff
@@ -164,7 +162,7 @@ object TimeT {
       Kleisli.ask[F, Time].map(_.now)
 
     val realTime =
-      pure(Instant.ofEpochMilli(0L))   // TODO is there anything better here*
+      pure(0.millis)   // TODO is there anything better here?
```
kubukoz, line 165: Don't believe for a second we're not going to make fun of the previous change here... 😂

djspiewak:
> That wasn't so bad (conflict-wise) :D

The really good one is coming soon! :-D

core/shared/src/main/scala/cats/effect/TimeT.scala
djspiewak: LOL!

PR #902, opened 2020-06-24
Improve docs for Fiber cancellation

bkirwi:
Inspired by: https://gitter.im/typelevel/cats-effect?at=5ed954c9778fad0b1348b6f6

I worry that strengthening the contract on `Fiber` itself is technically a breaking change, so I aimed for the `IO`-specific documentation.


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -377,6 +377,9 @@ sealed abstract class IO[+A] extends internals.IOBinaryCompat[A] {
    * consider in the example above what would happen if the first task
    * finishes in error. In that case the second task doesn't get canceled,
    * which creates a potential memory leak.
+   * 
+   * Cancelling the resulting `Fiber` will block until all finalizers
```
Avasil: I'd use a different wording, i.e. `Fiber#cancel will return after all finalizers have finished.` to avoid possible confusion that it blocks a thread

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -378,8 +378,8 @@ sealed abstract class IO[+A] extends internals.IOBinaryCompat[A] {
    * finishes in error. In that case the second task doesn't get canceled,
    * which creates a potential memory leak.
    * 
-   * Cancelling the resulting `Fiber` will block until all finalizers
-   * have run.
+   * Cancelling the resulting `Fiber` will run all finalizers
```
alexandru, line 392: Expression does not make it clear that implementation awaits all finalizers to finish before returning control.

Issue #820, opened 2020-03-22
Complete MVar2 interface


rossabaker:
After #739, we'll still be missing a few interesting functions present in the Haskell implementation:

* withMVar
* modifyMVar_
* modifyMVar

For consistent naming (no `MVar` suffix, avoid Scala keywords), I might consider naming these `using`, `modify_`, and `modify`, respectively.)

We should decide that we want these or reject them before releasing 2.2.0, because adding them later in the 2.x series would require the introduction of `MVar3`.

vasilmkd:
Hello, I'd like to work on this. Should I follow the Haskell semantics with regarding exception handling? Thanks.

alexandru:
The problem with `MVar` that we haven't foreseen is that the current `abstract class` doesn't allow adding new operations with default implementations. Adding methods with default implementations wouldn't be a problem for binary compatibility.

Instead of adding everything we ever wanted right now, maybe we should add the missing restriction:

```scala
abstract class MVar[F[_], A] {
  protected def F: Async[F]
}
```

This would allow us more flexibility in adding new operations on `MVar`.

vasilmkd:
I like that proposal. I'll wait for a more general agreement on the whole issue then.

rossabaker:
I think that's a good idea, but wouldn't it need to go on `MVar2` instead of `MVar`?  `MVar2` isn't locked down until 2.2.0, but `MVar` can't add anything abstract, even if it's just an instance to support more concrete methods later.

vasilmkd:
To me, it seems that it has to go on `MVar2`, yes.

Are default methods with `(implicit F: Async[F])` parameter list out of the question?

rossabaker:
Method-level `Async`s would be unpleasant with `MVarConcurrent`.  They should be coherent, and I suppose overrides that need `Concurrent` could ignore their method parameter and explicitly use the class constraint, but the abstract member seems a bit cleaner to me.  The cost is new operations have to go on `MVar2`.

vasilmkd:
Thanks for clearing it up. Should I proceed with a PR?

rossabaker:
:+1: That would be great. Thanks!

vasilmkd:
@rossabaker @alexandru I noticed that adding a `protected def F: Async[F]` is tough when implementing `mapK` for `MVar`. I don't think it's possible to obtain an `Async[G]` from an `Async[F]` and `F ~> G`. Thoughts?

rossabaker:
[...]

PR #912, opened 2020-07-06
Complete the MVar2 interface

vasilmkd:
Attempt at resolving #820.

The implementation aims to follow the semantics described [here](http://hackage.haskell.org/package/base-4.14.0.0/docs/Control-Concurrent-MVar.html#v:withMVar) and [here](https://books.google.mk/books?id=iSoRAAAAQBAJ&lpg=PT228&ots=FsTo4PEBXU&dq=haskell%20mask%20mvar&pg=PT229#v=onepage&q&f=false).

Please take a look at the tests and comment whether you think they are acceptable and/or sufficient.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/concurrent/MVar.scala
+++ b/core/shared/src/main/scala/cats/effect/concurrent/MVar.scala
@@ -124,6 +124,7 @@ abstract class MVar[F[_], A] extends MVarDocumentation {
   /**
    * Modify the context `F` using transformation `f`.
    */
+  @deprecated("`mapK` is deprecated in favor of `imapK` which supports the new invariant `MVar2` interface", "2.2.0")
   def mapK[G[_]](f: F ~> G): MVar[G, A] =
     new TransformedMVar(this, f)
 }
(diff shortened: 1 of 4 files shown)


SystemFw (commented):
Generally torn here. I think the implementation for `modify` is the best you can do in the CE2 model (short of trying to implement it inside MVar like Semaphore does), but it comes at quite a cost if you look at what `continual` does, and the resulting semantics are still more unsafe than `Semaphore.withPermit` in that a concurrent writer can still mean you deadlock. At least right now the absence of these functions prevents this.

Anyway, let's see what the others think :)

vasilmkd:
@SystemFw I agree completely. I did put your concern in the scaladoc for each of the new methods. It is however, the exact same unsafety in terms of deadlocks that is also present in Haskell's `MVar`, and there too, only the documentation is in place to warn the user.

Anyways, it was fun writing this PR, I got to explore the different cancelation semantics in CE as it currently stands. I'm also fine if none of this makes it into the final code.

rossabaker:
In a hurry tonight, so just a quick reaction for now: I'd say the deadlock situation in Haskell is somewhat less severe, because the runtime can detect threads deadlocked on MVar misuse and throw an exception.  It's bad in Haskell, but you get to clean up and see what went wrong, which is better than hanging.

This MVar incubated in Monix, and Monix still has its own, so I'm curious what @alexandru thinks.

I hope that I did not lead us astray by proposing these Haskell methods without thinking through their fit in Scala, but I'm glad you had fun either way.

vasilmkd:
However, if we do end up dropping this PR, `swap` in the `Concurrent` implementation is currently unsafe in terms of cancelation (cancelation can occur between reading and writing the `MVar`) which I believe is not desirable.

vasilmkd:
> Generally torn here. I think the implementation for `modify` is the best you can do in the CE2 model (short of trying to implement it inside MVar like Semaphore does), but it comes at quite a cost if you look at what `continual` does, and the resulting semantics are still more unsafe than `Semaphore.withPermit` in that a concurrent writer can still mean you deadlock. At least right now the absence of these functions prevents this.
> 
> Anyway, let's see what the others think :)

@SystemFw can you describe shortly what would the desirable semantics of a `Concurrent` `modify` be (w.r.t. cancelation)? When compared to `Semaphore`, the cleanup `releaseN` is still uncancelable (being executed as the release action of a `Bracket`), but that is unsafe for `MVar`.

Thus, if replacing the old value of the `MVar` should be cancelable in Scala (otherwise it would lead to deadlocks), why not just go with a simple imlementation similar to the current `swap`, where cancelation can occur anywhere.

Am I missing something? /cc @rossabaker

djspiewak:
@SystemFw @rossabaker Any thoughts on this? I'd like to unblock things. These honestly seem like reasonable functions, and they should be safe on ce3, so it's reasonably future-proof. Losing `mapK` is unfortunate, though not the end of the world.

SystemFw:
I'm still not a massive fan of them (not the implementation, which is good :) ), just on principle I think MVar is a bit of a trap, mitigated in haskell by a mechanism that we do not have.

That being said, the implementation is good (especially in ce3 when we can remove the weight of all the continuals), and other people like MVar, so we can merge

Also, apologies to @vasilmkd for taking so long to reply

vasilmkd:
Totally fine. Please do not feel the need to merge just because the code is already written. It should only be merged if it's right to do so. You guys certainly know much more about all of this.

djspiewak:
> Totally fine. Please do not feel the need to merge just because the code is already written. It should only be merged if it's right to do so. You guys certainly know much more about all of this.

This just mostly comes down to the fact that `MVar` is terrible. :-) I think we're all very interested in finding ways to steer people away from `MVar` towards safer and easier alternatives (`Ref`/`Deferred`/etc). I do think these functions are fine in so far as `MVar` goes, it's just that `MVar` is sadness.

At any rate… your work here is excellent! Thank you!

PR #914, opened 2020-07-07
Transformer instances for clock

TimWSpence:


TimWSpence:
For CE3. @djspiewak :)

build.sbt
```diff
@@ -54,7 +54,7 @@ Global / scmInfo := Some(
     url("https://github.com/typelevel/cats-effect"),
     "git@github.com:typelevel/cats-effect.git"))
 
-val CatsVersion = "2.1.1"
+val CatsVersion = "2.2.0-RC1"
```
djspiewak, line 57: 👍

kernel/shared/src/main/scala/cats/effect/kernel/Clock.scala
```diff
@@ -17,9 +17,13 @@
 package cats.effect.kernel
 
 import cats.Applicative
+import cats.data._
```
djspiewak, line 20: I much prefer specific imports, rather than underscores, since it's *very* easy to accidentally lose source compatibility with underscores. The only exception I make to this rule are things which contain implicits (e.g. `cats.implicits._`). Any names that I explicitly write though, I prefer to import just as explicitly, if only for documentation purposes.

laws/shared/src/test/scala/cats/effect/ClockSpec.scala
```diff
@@ -0,0 +1,147 @@
...
+import org.specs2.mutable._
+
+import org.typelevel.discipline.specs2.mutable.Discipline
+import cats.laws.discipline.{eq, ExhaustiveCheck, MiniInt}; import eq._
+
+class ClockSpec extends Specification with Discipline with ScalaCheck {
+  import FreeSyncGenerators._
+
+  implicit def prettyFromShow[A: Show](a: A): Pretty =
+    Pretty.prettyString(a.show)
```
djspiewak, line 41: We should definitely put this somewhere common, rather than copying it everywhere. :-)

testkit/shared/src/main/scala/cats/effect/testkit/pure.scala
```diff
@@ -178,7 +178,7 @@ object pure {
       S: Functor[S])
       : MonadError[FreeT[S, M, *], E] =
     new MonadError[FreeT[S, M, *], E] {
-      private val F = FreeT.catsFreeMonadErrorForFreeT[S, M, E]
+      private val F = FreeT.catsFreeMonadErrorForFreeT2[S, M, E]
```
djspiewak, line 181: Now that we're upgraded to the latest RC, we can actually drop the `catsFreeMonadErrorForFreeT2` jazz, because the one upstream has the fix merged. Task for another PR. :-)

PR #915, opened 2020-07-08
Update doc for unsafeRunTimed, to mention Concurrent.timeout

bplommer:


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -313,12 +313,13 @@ sealed abstract class IO[+A] extends internals.IOBinaryCompat[A] {
...
-   * implement timeouts, or anything similar. If you need that sort
-   * of functionality, you should be using a streaming library (like
-   * fs2 or Monix).
+   * This function should never appear in your mainline production code!
+   * If you want to implement timeouts, or anything similar, you should
+   * use `Concurrent.timeout` or `Concurrent.timeoutTo`, which implement
+   * a timeout ''within'' your `IO` program, while permitting proper
+   * resource cleanup and subsequent error handling. `unsafeRunTimed`
+   * should only be used for ''testing'', in cases where you want timeouts to
+   * cause immediate termination of the cats-effect runtime itself.
```
alexandru, line 322: I don't agree. 

While a similar effect can be achieved with `Concurrent.timeout`, they operate at different levels of the run-loop. 

`unsafeRunTimed` exists because doing `unsafeRunSync` on top of a thread that isn't meant to support blocking operations is very unsafe. In fact I would say that `unsafeRunSync` without a `timeout` parameter shouldn't have been added to the API IMO, the reason for why we have it is because we obviously needed to express synchronous operations and we didn't have `SyncIO` back then.

> while permitting proper resource cleanup and subsequent error handling

Sure, this is true, but you can say the same thing about `unsafeRunSync`. Ideally you'd use `IOApp`. But if not using `IOApp`, then `unsafeRunTimed` is probably better for async operations than `unsafeRunSync`.

alexandru: Ah, I now see the comment that it is replacing, sorry 😅

OK, it's a good idea to mention `Concurrent.timeout`, but maybe as a separate paragraph?

djspiewak: I actually think it's pretty reasonable to have it in this paragraph. It emphasizes that `unsafeRunTimed`'s sole purpose is to implement tests and the like, without getting into the resource management resources why this is the case. Users who find this function will *most* of the time actually be looking for `Concurrent#timeout`, so it doesn't seem that crazy to mention it.

PR #918, opened 2020-07-11
Revise Deferred, MVarConcurrent, LinkedMap

RaasAhsan:
@johnynek came on Gitter the other day and found that the `Deferred` implementation was consuming a surprising amount of CPU on an app (chat logs are [here](https://gitter.im/typelevel/cats-effect?at=5f0644d68342f4627402286a)). It turns out that the `hashCode` method for the `Id` class that `Deferred` uses to uniquely index callbacks in the `get` method is pretty costly.

The revised implementation here uses the `AtomicReference` state field to track an extra `Long` that serves as a monotonic counter to uniquely index `get` callbacks. `MVarConcurrent` has similar details in its implementation, so I also changed that one. Both `Deferred` and `MVarConcurrent` use a `LinkedMap` under the hood, but only ever uses `Long`s as keys. However since they're the only two clients of that class, I renamed the class to `LinkedLongMap` and specialized it to `Long`s, which avoids extra boxing.

## Sample application
```scala
import cats.effect.concurrent.Deferred

object App extends IOApp {
  def loop(n: Int): IO[Unit] =
    for {
      d <- Deferred[IO, Unit]
      _ <- d.get.start
      _ <- d.get.start
      _ <- d.get.start
      _ <- d.get.start
      _ <- d.get.start
      _  <- loop(n)
    } yield ()

  override def run(args: List[String]): IO[ExitCode] = for {
    _ <- IO(println("Starting"))
    _ <- loop(1000000)
  } yield ExitCode.Success
}
```

Below are some profiling results for both master and the PR. I ran some throughput benchmarks but these data structures are somewhat tedious to benchmark since they rely on concurrency and as soon as you start forking or synchronizing, the costs of allocations look very miniscule. FWIW the program above saw virtually no difference between the master and the PR, but somebody could probably write a better concurrent benchmark.

I also ran some allocation tracing and this PR seems to produce a lot less garbage than the previous implementation.

## master
image

## PR
image

### DIFF
--- a/core/shared/src/main/scala/cats/effect/concurrent/Deferred.scala
+++ b/core/shared/src/main/scala/cats/effect/concurrent/Deferred.scala
@@ -18,6 +18,7 @@ package cats
 package effect
 package concurrent
 
+import cats.implicits._
 import cats.effect.internals.{Callback, LinkedLongMap, TrampolineEC}
 import java.util.concurrent.atomic.AtomicReference
 
@@ -192,22 +193,23 @@ object Deferred {
       }
 
     private[this] def unsafeRegister(cb: Either[Throwable, A] => Unit): Long = {
-      var id: Long = 0
-
       @tailrec
-      def register(): Option[A] =
+      def register(): Either[Long, A] =
         ref.get match {
-          case State.Set(a) => Some(a)
+          case State.Set(a) => Right(a)
           case s @ State.Unset(waiting, nextId) =>
             val updated = State.Unset(waiting.updated(nextId, (a: A) => cb(Right(a))), nextId + 1)
-            if (ref.compareAndSet(s, updated)) {
-              id = nextId
-              None
-            } else register()
+            if (ref.compareAndSet(s, updated)) Left(nextId)
+            else register()
         }
 
-      register().foreach(a => cb(Right(a)))
-      id
+      register() match {
+        case Left(id) => id
+        case r @ Right(_) => {
+          cb(r.leftCast[Throwable])
+          0L
+        }
+      }
     }
 
     def complete(a: A): F[Unit] =
--- a/core/shared/src/main/scala/cats/effect/concurrent/Deferred.scala
+++ b/core/shared/src/main/scala/cats/effect/concurrent/Deferred.scala
@@ -18,7 +18,7 @@ package cats
 package effect
 package concurrent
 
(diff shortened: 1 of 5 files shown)


core/shared/src/main/scala/cats/effect/internals/LinkedLongMap.scala
```diff
@@ -25,9 +25,9 @@ import scala.collection.immutable.LongMap
  * traversed in the order they were inserted.  Alternative to `ListMap` that
  * has better asymptotic performance at the cost of more memory usage.
  */
-private[effect] class LinkedMap[K, +V](
-  val entries: Map[K, (V, Long)],
-  private[this] val insertionOrder: LongMap[K],
+private[effect] class LinkedLongMap[+V](
```
RaasAhsan, line 28: I feel like we could maybe remove this since the `Long` counter for both `Deferred` and `MVarConcurrent` are monotonic, but I'll have to think about it more

core/shared/src/main/scala/cats/effect/concurrent/Deferred.scala
```diff
@@ -188,22 +186,24 @@ object Deferred {
...
         }
       }
 
-    private[this] def unsafeRegister(cb: Either[Throwable, A] => Unit): Id = {
-      val id = new Id
+    private[this] def unsafeRegister(cb: Either[Throwable, A] => Unit): Long = {
+      var id: Long = 0
 
       @tailrec
       def register(): Option[A] =
```
johnynek, line 198: could return `Either[Long, A]` and not have the `var id`. That might be nice.

I *think* this var is safe, but with this code there are often threads so the reader has to remind themselves that we aren't in a volatile memory situation here.

core/shared/src/main/scala/cats/effect/internals/LinkedLongMap.scala
SystemFw: that was my first thought as well

core/shared/src/main/scala/cats/effect/concurrent/Deferred.scala
djspiewak: Worth noting that `var`s which are lambda lifted end up in a `Box`, which is (I believe) volatile.

PR #923, opened 2020-07-12
Reimplemented IOFiber callbacks with a thread-safe atomic stack

### LINKED ISSUE #920: IOFiber callbacks aren't thread-safe
The data structure is pretty dubious, given that `SafeArrayStack` can't really handle simultaneous `push`es from multiple threads. We really need a different structure here, one which allows truly concurrent modifications. Additionally, the structure is already not amenable to having a callback *removed*, which is problematic since it means that `f.join.start.flatMap(_.cancel)` is a memory leak.

djspiewak:
In order to avoid excess boxing, `CallbackStack` *extends* `AtomicReference`, with the reference cell pointing to the tail (possibly null). The callback itself is contained in a non-`volatile` field. All fibers are initialized with a `null` callback. This structure was chosen both because it allows fast, low-contention insert and traversal while also allowing for very fast contention-free item *removal* (by setting it to `null`). This fixes the memory leak on `join` cancelation, as well as avoiding the current problems with thread safety.

Fixes #920

### DIFF
--- a/core/shared/src/main/scala/cats/effect/CallbackQueue.scala
+++ b/core/shared/src/main/scala/cats/effect/CallbackQueue.scala
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2020 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


RaasAhsan (approved):
Very cool way to implement a queue w/o allocations

djspiewak:
I can't claim credit. :-) I believe it was @plokhotnyuk who first came up with this idea, which I beat my head against for a while, trying to understand, before describing [in this writeup](https://gist.github.com/djspiewak/671deab9d7ea027cdc42).

SystemFw:
oh I wish I had found that gist when I was documenting it in the fs2 Actor (a weekend spent understanding it), which incidentally is where I learned "ec submission == memory barrier"

djspiewak:
As an aside, I actually implemented a stack rather than a queue. I'm going to change the names so future us aren't confused.

PR #927, opened 2020-07-13
Remove skips in JVM publishing

rossabaker:
These caused a hiccup in the 2.2.0-RC1 release.  I believe they're an artifact of scala.js crossbuilding.  If they're not, we need another solution, because they now do surprising things on release and local publication.


djspiewak:
@rossabaker I *think* this is already reflected in master now that I pushed directly? I'm not sure why it doesn't show as conflicted.

PR #932, opened 2020-07-13
Saner runners

djspiewak:
Heads up, @TimWSpence!

Okay so we now have both realistic test runners *and* ticked test runners, and you can choose between them as relevant. You also no longer need to be strictly `sequential` except when doing laws testing. Arguably this should have just been in `BaseSpec` rather than `Runners`; I'm open to the suggestion that we should just merge these two. Anyway, all tests involving `IO` should fall into one of these two categories:

```scala
"my test case" in ticked { implicit ticker =>
  // ...
}
```

`ticked` tests are just normal tests, and they have access to `Eq[IO[A]]`, `Arbitrary[IO[A]]` and all that goodness. They do not run on real threads.

```scala
"my test case" in real {
  // ...
}
```

`real` tests are normal tests, except they need to return in `IO`. They don't have access to `Eq` or `Arbitrary` since they run the *real* `IO` on real threads and stuff. They all run with a timeout of ten seconds, so if you want to assert that something doesn't timeout, `.as(ok)` is generally pretty good. Otherwise, it's safe to just wrap normal specs `must` matchers in `IO { ... }`.

Note that `real` tests *do* work on ScalaJS, even when you do crazy stuff with time and `async` and stuff, and they do support parallel execution.


TimWSpence:
This looks great!! :)

PR #933, opened 2020-07-13
Port concurrent deferred

TimWSpence:
Note that this removes the `uncancelable` versions of `Deferred` as we no longer have the `Concurrent`/`Async` distinction where one supports cancellation and the other doesn't.


TimWSpence:
For CE3 @djspiewak :)

TimWSpence:
Also please don't judge that commit history 😂

djspiewak:
:-D I don't say this often, but it's probably worth squashing the wips away. I generally try to do that before I push.

TimWSpence:
@djspiewak I assumed you squash on merge? If not then yes, I definitely will!!

djspiewak:
> I assumed you squash on merge?

Nope. :-) Generally speaking, when something is in a PR, I consider it sacred and I won't rebase or squash those commits.

TimWSpence:
Then I will be a LOT more careful in future! 😂 I've squashed the WIP mess at least now

djspiewak:
I really need to get the import warnings working on 2.13 and Dotty… Sorry about that…

PR #938, opened 2020-07-15
Implement map2Eval for IO

johnynek:
map2Eval is often used when we want to recurse on the right hand side. The default implementation just calls `.map` on the Eval forcing the full evaluation to happen before we can return the `IO` instance.

We can instead just call `.value` once the `A` value from `fa` has been produced which can result in occasionally doing less work when `fa` is a failure, or never completes.


johnynek:
This should be green now. Mima failed because I added a `final` method.

Since I'm a binary compatibility conservative, I just removed `final` since I don't think it will hurt performance and I don't want to have to make a reasoned case about this binary breakage actually being safe.

Thank you for reviewing!

djspiewak:
I'm happy to make the reasoned case for you, but in either case, I agree that performance shouldn't be affected. Thanks for the PR!

PR #940, opened 2020-07-15
Port concurrent mvar

TimWSpence:
Like `Semaphore` and `Deferred` we remove the `uncancelable` versions as we no longer have the `Concurrent`/`Async` distinction.

This also merges `MVar` and `MVar2` as we don't have to maintain binary compatibility between CE2 and CE3


TimWSpence:
For CE3 @djspiewak. Dependent on https://github.com/typelevel/cats-effect/pull/939

kernel/shared/src/main/scala/cats/effect/kernel/package.scala
```diff
@@ -60,4 +60,6 @@ package object kernel {
   object AsyncRegion {
     def apply[R[_[_], _], F[_]](implicit R: AsyncRegion[R, F]): AsyncRegion[R, F] = R
   }
+
+  type CancelToken[F[_]] = F[Unit]
```
djspiewak: I don't think we need this unless we specifically *want* to bring it back. If we do want to bring it back, then it needs to be additionally mirrored in the `cats.effect` package object in **core**.

djspiewak (changes requested):

PR #941, opened 2020-07-15
Removed unneeded things that were ported over

djspiewak:
Attn @TimWSpence 

There were a number of things ported over as part of #933 which weren't needed. `LinkedMap` is still necessary, but I moved it to the `concurrent` package.


djspiewak:
Superseded by #939

PR #944, opened 2020-07-15
Implicit boilerplate

djspiewak:
This just finalizes some of the boilerplate templates for having implicit instances and syntax, following the cats pattern.

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/instances/ConcurrentInstances.scala
+++ b/kernel/shared/src/main/scala/cats/effect/instances/ConcurrentInstances.scala
@@ -14,10 +14,8 @@
  * limitations under the License.
  */
 
-package cats.effect
+package cats.effect.instances
 
-package object syntax {
-  object all extends AllSyntax
-  object async extends AsyncSyntax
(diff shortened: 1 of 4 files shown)


LukaJCB (approved):

kernel/shared/src/main/scala/cats/effect/instances/ConcurrentInstances.scala
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.instances
+
+trait ConcurrentInstances {
+  // TODO Parallel[F] given Concurrent[F] goes here!
```
LukaJCB, line 20: As an aside, I'm not fully convinced this is a good idea, but we can and should discuss this elsewhere

djspiewak: Gitter!

PR #945, opened 2020-07-16
Add combinators for more CE2 parity

RaasAhsan:
Quick PR for adding some more combinators to get CE2 parity. I was going to add `guaranteeCase` but it seems like we can't because `IO` is covariant in `A` whereas `Outcome` is invariant, and I think that causes a contravariance issue with `Function1`. Has that been identified as an issue?

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -37,9 +37,6 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   def >>[B](that: => IO[B]): IO[B] =
      flatMap(_ => that)
 
-  def <<[B](that: => IO[B]): IO[A] =
-    that.flatMap(_ => this)
-
   def as[B](b: B): IO[B] =
     map(_ => b)
 
@@ -284,7 +281,7 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
     Uncancelable(body)
 
   private[this] val _unit: IO[Unit] = Pure(())
-  val unit: IO[Unit] = _unit
+  def unit: IO[Unit] = _unit
 
   // utilities
 
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -16,12 +16,13 @@
 
 package cats.effect
 
-import cats.{~>, Monoid, /*Parallel,*/ Semigroup, Show, StackSafeMonad}
+import cats.{Eval, Monoid, Now, Semigroup, Show, StackSafeMonad, ~>}
 import cats.implicits._
 
 import scala.annotation.unchecked.uncheckedVariance
-import scala.concurrent.{ExecutionContext, Future, Promise}
+import scala.concurrent.{ExecutionContext, Future, Promise, TimeoutException}
 import scala.concurrent.duration._
+import scala.util.{Failure, Success, Try}
 
 sealed abstract class IO[+A] private () extends IOPlatform[A] {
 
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -36,6 +37,9 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   def >>[B](that: => IO[B]): IO[B] =
      flatMap(_ => that)
 
+  def <<[B](that: => IO[B]): IO[A] =
+    that.flatMap(_ => this)
```
LukaJCB: I think this should still perform `this` first, but then `map` to its result after `that` :)

djspiewak: It should! Also does cats have `<<`? I was under the impression that the only variant of `productL` in cats was `<*`, which is eager.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -262,7 +283,8 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
   def uncancelable[A](body: IO ~> IO => IO[A]): IO[A] =
     Uncancelable(body)
 
-  val unit: IO[Unit] = pure(())
+  private[this] val _unit: IO[Unit] = Pure(())
+  val unit: IO[Unit] = _unit
```
djspiewak: Does this actually result in improvements?

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -92,6 +96,18 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
...
+    IO.sleep(duration) *> this
+
+  def timeout[A2 >: A](duration: FiniteDuration): IO[A2] =
+    timeoutTo(duration, IO.raiseError(new TimeoutException(duration.toString)))
+
+  def timeoutTo[A2 >: A](duration: FiniteDuration, fallback: IO[A2]): IO[A2] =
+    race(IO.sleep(duration)).flatMap {
+      case Right(_) => fallback
+      case Left(value) => IO.pure(value)
+    }
```
djspiewak, line 106: I think these should also be copied to `Temporal`. I want to make sure we try to keep things as much in sync as we can.

djspiewak (approved):

core/shared/src/main/scala/cats/effect/IO.scala
LukaJCB: It definitely had it at some point, not sure if it's still around

djspiewak: Dash believes it no longer exists. Which I kinda get, since the only thing separating `*>` and `>>` is laziness, which makes a lot less sense in the `<<` case.

core/shared/src/main/scala/cats/effect/IO.scala
RaasAhsan: I think once they're in `Temporal`, we should also just be able to call out to instance for `IO`

core/shared/src/main/scala/cats/effect/IO.scala
RaasAhsan: oops, `unit` needs to be a `def`

core/shared/src/main/scala/cats/effect/IO.scala
RaasAhsan: Makes sense to remove here too then

core/shared/src/main/scala/cats/effect/IO.scala
djspiewak: Yes absolutely, though I like having copies of the combinators on `IO` itself for reasons for discoverability in documentation, IDEs, and error messages.

PR #952, opened 2020-07-17
Fiber.joinWith

RaasAhsan:
- [ ] Implement `to`
- [ ] Laws and tests

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -94,8 +94,32 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
 
   def racePair[B](that: IO[B]): IO[Either[
     (A @uncheckedVariance, Fiber[IO, Throwable, B]),
-    (Fiber[IO, Throwable, A @uncheckedVariance], B)]] =
-    IO.RacePair(this, that)
+    (Fiber[IO, Throwable, A @uncheckedVariance], B)]] = {
+    for {
+      f1 <- this.start
+      f2 <- that.start
+      oc <- f1.joinWith(f2)
+      r  <- oc match {
+        case Left(oca) => oca match {
+          case Outcome.Completed(fa) => fa.map(a => Left((a, f2)))
+          case Outcome.Errored(_) | Outcome.Canceled() => f2.join.flatMap {
+            case Outcome.Completed(fb) => fb.map(b => Right((f1, b)))
+            case Outcome.Errored(e) => IO.raiseError(e)
+            case Outcome.Canceled() => ???
+          }
+        }
+        case Right(ocb) => ocb match {
+          case Outcome.Completed(fb) => fb.map(b => Right((f1, b)))
+          case Outcome.Errored(_) | Outcome.Canceled() => f1.join.flatMap {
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -94,8 +94,32 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
...
+      f1 <- this.start
+      f2 <- that.start
+      oc <- f1.joinWith(f2)
+      r  <- oc match {
+        case Left(oca) => oca match {
+          case Outcome.Completed(fa) => fa.map(a => Left((a, f2)))
+          case Outcome.Errored(_) | Outcome.Canceled() => f2.join.flatMap {
+            case Outcome.Completed(fb) => fb.map(b => Right((f1, b)))
+            case Outcome.Errored(e) => IO.raiseError(e)
+            case Outcome.Canceled() => ???
```
RaasAhsan, line 108: I think we should self-cancel here? Depending on if the first outcome errored or cancelled

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -164,6 +164,14 @@ private[effect] final class IOFiber[A](name: String, scheduler: unsafe.Scheduler
       }
     }
 
+  def joinWith[B](that: Fiber[IO, Throwable, B]): IO[Either[Outcome[IO, Throwable, A], Outcome[IO, Throwable, B]]] =
```
RaasAhsan, line 167: Implementation uses only primitive operations. This is sort of forced upon us because we don't know anything about the implementation of `that`, but it means that `join` and `cancel` are technically the only `Fiber` primitives we need

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -164,6 +164,14 @@ private[effect] final class IOFiber[A](name: String, scheduler: unsafe.Scheduler
...
+    IO.async[Either[Outcome[IO, Throwable, A], Outcome[IO, Throwable, B]]] { cb =>
+      for {
+        fa <- join.flatMap(oc => IO(cb(Right(Left(oc))))).start
```
RaasAhsan, line 170: Invoking the callback inside an `IO()`... feels dirty but its the only way we can "register a listener" with what we have

I think calling *start* here is OK. If `this` is cancelled, then `join` here should return `Outcome.Cancelled`.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -164,6 +164,14 @@ private[effect] final class IOFiber[A](name: String, scheduler: unsafe.Scheduler
...
+        fb <- that.join.flatMap(oc => IO(cb(Right(Right(oc))))).start
+      } yield Some(fa.cancel *> fb.cancel)
```
RaasAhsan, line 172: If the supervisor fiber is cancelled, cancel both children fibers

Avasil: I think we should call both cancels concurrently
If `fa` finalizers take too long (or if `cancel` waits for the fiber to be stopped) then `fb` might not be canceled in time

PR #953, opened 2020-07-17
Exclusively use setTimeout for executor on JavaScript

djspiewak:
See scala-js/scala-js#4129 for some more details, but basically `Promises` do not actually yield, and thus cannot be used to back an `ExecutionContext`. This PR forces `setTimeout` as the backing mechanism, which does reliably yield back to the event loop. This in turn exposed some issues in the tests, since yielding is much, much slower than simply continuing. The tests *had* worked without yielding, but since we need to yield, they've slowed down. I resolved this by reducing the iteration count.


sjrd:
> since yielding is much, much slower than simply continuing

If you think it's slow in Node.js, wait until you try in browsers. *By spec*, browsers will bump the minimal delay from 0ms to 4ms after the 5th or so `setTimeout` in a chain (i.e., when the callback of `setTimeout` creates a new `setTimeout`, whose callback creates another one, etc.).

Reference: https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout#Reasons_for_delays_longer_than_specified

Such a change can drastically impact the performance of real-world applications.

djspiewak:
> If you think it's slow in Node.js, wait until you try in browsers. By spec, browsers will bump the minimal delay from 0ms to 4ms after the 5th or so setTimeout in a chain (i.e., when the callback of setTimeout creates a new setTimeout, whose callback creates another one, etc.).

Oh dear god…

The problem is that I don't think we can *avoid* this. Like, we don't actually yield *that* often outside of contrived scenarios, so it's probably somewhat okay, but the current implementation is just *wrong*. Like, it doesn't yield at all, and this in turn means that many concurrent programs are simply incorrect. A quick example of a program which is incorrect today is the following:

```scala
IO.cede.foreverM.start.flatMap(f => IO.cede >> f.cancel)
```

This program builds a fiber out of an infinite loop of yields. Literally the *only* thing the fiber does is yield control back to the system. Then, its parent fiber yields once, and then cancels the looping fiber. When using the promises executor, this program loops forever. This is obviously an extreme example, but it is quite archetypical of stuff that is very realistic. So… I'm not sure we have a choice. :-(

djspiewak:
BTW this is probably a faster approach: https://github.com/YuzuJS/setImmediate

PR #957, opened 2020-07-18
Align instance for parallelf

TimWSpence:


kernel/shared/src/main/scala/cats/effect/instances/ConcurrentInstances.scala
```diff
@@ -40,4 +42,40 @@ trait ConcurrentInstances {
         }
 
     }
+
+  implicit def applicativeForParallelF[F[_], E](
+      implicit F: Concurrent[F, E]): Applicative[ParallelF[F, *]] =
```
djspiewak, line 47: Random curiosity: since the compiler considers `ParallelF` to be a real type, does it similarly believe it has a companion? If so, then we can tuck this `Applicative` into that companion, which is generally better scoping.

laws/shared/src/test/scala/cats/effect/PureConcSpec.scala
```diff
@@ -45,4 +47,10 @@ class PureConcSpec extends Specification with Discipline with ScalaCheck {
     "PureConc",
     ConcurrentBracketTests[PureConc[Int, *], Int].concurrentBracket[Int, Int, Int]
   ) /*(Parameters(seed = Some(Seed.fromBase64("OjD4TDlPxwCr-K-gZb-xyBOGeWMKx210V24VVhsJBLI=").get)))*/
+
+  checkAll("PureConc", ParallelTests[PureConc[Int, *]].parallel[Int, Int])
```
djspiewak, line 50: 🎉

kernel/shared/src/main/scala/cats/effect/instances/ConcurrentInstances.scala
TimWSpence: No it does not 😭 I had it there originally and it took me a while to work out why my tests wouldn't compile! Maybe there is some formulation of newtypes that can fix this, I'm not sure!

laws/shared/src/test/scala/cats/effect/PureConcSpec.scala
TimWSpence: I know!!! Somewhat worrying that this disappeared somewhere.. I'm much more suspicious of rebased then I used to be!

djspiewak:
Good to merge once the conflicts are resolved!

PR #959, opened 2020-07-18
Build a better bracket

### LINKED ISSUE #860: CE 3: Attempting to Make Bracket More Composable

djspiewak:
I think @SystemFw would be interested in this.

This is a big change! I removed both `Region` and `Bracket`, as well as the whole parallel hierarchy machinery (as it is no longer required. `bracketCase` is now defined in terms of `bracketFull` (which takes a polling function), and `bracketFull` is defined by `Concurrent` in terms of `uncancelable`, `onCancel`, and more. `onCancel` is now the primitive, and it's *so* much nicer. The laws which govern it are really very natural, which makes sense since it's effectively the signal elimination rule for `canceled`, modulated by `uncancelable` itself.

This had the pleasing side-effect of making `IO` actually *faster* and simpler. I haven't rerun benchmarks, but all cases involving finalization should be considerably more efficient.

So this raises the natural question then: what about `Resource`? The whole point of `Region` was attempt to extend the hierarchy to dynamic regions like `Resource`, `Stream`, `Iterant`, `Managed`, and more. I still want to be able to do this, but I think it needs to be done slightly differently. My realization today was that all dynamic monadic regions are *also* static regions: it isn't a dual, it's a *generalization*. Let's walk through the thought process there.

If we no longer consider `bracket` to be primitive, then the defining characteristic which enables a `bracket`-like operation is really `handleErrorWith` and `flatMap`. Which is to say, `MonadError`. In fact, if you ignore cancelation entirely, you can define `bracket` for any `MonadError`! This is actually quite logical, and I think most of us have had this thought at some point.

Here's the problem: `Resource` forms a `MonadError`. So does `Stream`. Meaning that, again ignoring cancelation, these canonical examples of *dynamic* regions **are also static regions!** I find that fascinating. It's logical though: in theory, `bracket` is simply a "dynamic" monadic region with a single bind. Thus the isomorphism is complete: the free region (`Resource`) on series/2.x is defined in terms of `bracket`, and we just hand-wave defined `bracket` in terms of a region.

Once we've gotten to this point in our mental process, we need to figure out what cancelation is. I think the answer is that it's a form of monadic signal, similar to an error, but without elimination. In fact, if we only had `canceled` and `onCancel`, then it would look *remarkably* like `raiseError` and `onError`, which is rather nice. So in a sense, all `Concurrent`s are a particular type of monad which has a signal propagation mechanism. In the case of `Concurrent`, that signal is instantiated to `Outcome.Canceled`.

The trick though is that we have `uncancelable`, which affects the signal propagation in complicated and interesting ways. We *need* those complicated and interesting semantics, but I'm not 100% convinced that they generalize to other constructs in any way. At least, I can't think of any other sorts of "signaling monads" which have this kind of masking mechanism. Thus, rather than attempting to over-generalize, I've left `Concurrent` as defining this signaling mechanism *in addition* to defining concurrency itself.

But what about dynamic regions? Well, if `Concurrent` defines what it means to be a signaling monad, where that signal is cancelation, and if all structures which can represent dynamic regions are *also* capable of expressing a static region, then do we really need a separate abstraction? I'm not 100% sure, but if we do, it will be a little interesting. I have this sketch on another branch:

```scala
trait Region[F[_], E] extends MonadError[F, E] {

  def pushFinalizer(handler: Handler): F[Unit]
  def finalize: F[Unit]

  trait Handler {
    def completed[A](a: A): F[A]
    def errored[A](e: E): F[A]
  }
}
```

You could imagine extending this to handle cancelation as well. Try to look through the encoding though and notice what's happening here: `Handler` is basically just a pair of rank-2 functions mirroring `flatMap` and `handleErrorWith`. The types ensure that the only thing you're allowed to do in `errored` is either re-`raise` the `E`, or somehow produce a different sort of monadic zero (e.g. if your `F` is `OptionT`, then you can produce `none`).

Anyway, if you generalize this again with the concept of a signaling monad (which is defined by `Concurrent`), then you get something which literally represents `Resource`'s semantics. Is this worth building into an abstraction? I'm not sure. It's *interesting*. But given that we already showed that dynamic monadic regions *generalize* static ones, we've probably already indirectly achieved our original goal: `Concurrent[Resource[F, *], E]` should be definable even without a `Region` class. The problem was never the dynamism of the monadic region; the problem was `bracket` all along.

Fixes #860


SystemFw:
This is indeed very interesting. I've been thinking for a while that we don't need `bracket` as a primitive given that `uncancelable` can readily express it (as well as `cancelableBracket`, `continual`, and all the other things). The "signaling monad" and new `Region` is very cool though, especially given how complex the current `Case` encoding is (I've been dealing with it to port `Resource`, and it's proving to be nontrivial for some operations).

SystemFw (approved):
I'm very, very much in favour of this. The resulting hierarchy is just a lot more pleasant to look at, and although it may seem like Concurrent is a huge class, the number of _primitive_ operations is rather small and orthogonal. In fact I'm super tempted to just start building on top of this branch :)

djspiewak:
> I'm very, very much in favour of this. The resulting hierarchy is just a lot more pleasant to look at, and although it may seem like Concurrent is a huge class, the number of primitive operations is rather small and orthogonal. In fact I'm super tempted to just start building on top of this branch :)

Sounds like a vote of confidence to me. Let's merge it; what's the worst that could happen?

PR #960, opened 2020-07-19
Revise race pair

RaasAhsan:
- [x] Fix laws
- [x] Fix `Gen` instances

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -44,7 +44,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   def attempt: IO[Either[Throwable, A]] =
     map(Right(_)).handleErrorWith(t => IO.pure(Left(t)))
 
-  def bothOutcome[B](that: IO[B]): IO[(IOOutcome[A @uncheckedVariance], IOOutcome[B])] =
+  def bothOutcome[B](that: IO[B]): IO[(OutcomeIO[A @uncheckedVariance], OutcomeIO[B])] =
     racePair(that).flatMap {
       case Left((oc, f)) => f.join.map((oc, _))
       case Right((f, oc)) => f.join.map((_, oc))
@@ -79,7 +79,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   def bracket[B](use: A => IO[B])(release: A => IO[Unit]): IO[B] =
     bracketCase(use)((a, _) => release(a))
 
-  def bracketCase[B](use: A => IO[B])(release: (A, IOOutcome[B]) => IO[Unit]): IO[B] =
+  def bracketCase[B](use: A => IO[B])(release: (A, OutcomeIO[B]) => IO[Unit]): IO[B] =
     IO uncancelable { poll =>
       flatMap { a =>
         val finalized = poll(use(a)).onCancel(release(a, Outcome.Canceled()))
@@ -97,7 +97,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   def guarantee(finalizer: IO[Unit]): IO[A] =
     guaranteeCase(_ => finalizer)
 
-  def guaranteeCase(finalizer: IOOutcome[A @uncheckedVariance] => IO[Unit]): IO[A] =
+  def guaranteeCase(finalizer: OutcomeIO[A @uncheckedVariance] => IO[Unit]): IO[A] =
     onCase { case oc => finalizer(oc) }
 
   def handleErrorWith[B >: A](f: Throwable => IO[B]): IO[B] =
@@ -108,7 +108,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   def onCancel(fin: IO[Unit]): IO[A] =
     IO.OnCancel(this, fin)
 
-  def onCase(pf: PartialFunction[IOOutcome[A @uncheckedVariance], IO[Unit]]): IO[A] =
+  def onCase(pf: PartialFunction[OutcomeIO[A @uncheckedVariance], IO[Unit]]): IO[A] =
     IO uncancelable { poll =>
       val base = poll(this)
       val finalized = pf.lift(Outcome.Canceled()).map(base.onCancel(_)).getOrElse(base)
@@ -149,8 +149,8 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
     }
 
   def racePair[B](that: IO[B]): IO[Either[
-    (IOOutcome[A @uncheckedVariance], IOFiber[B]),
-    (IOFiber[A @uncheckedVariance], IOOutcome[B])]] =
+    (OutcomeIO[A @uncheckedVariance], FiberIO[B]),
+    (FiberIO[A @uncheckedVariance], OutcomeIO[B])]] =
     IO.RacePair(this, that)
 
   def delayBy(duration: FiniteDuration): IO[A] =
@@ -171,7 +171,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   def productR[B](that: IO[B]): IO[B] =
     flatMap(_ => that)
 
-  def start: IO[IOFiber[A @uncheckedVariance]] =
+  def start: IO[FiberIO[A @uncheckedVariance]] =
     IO.Start(this)
 
   def to[F[_]](implicit F: Effect[F]): F[A @uncheckedVariance] =
@@ -179,8 +179,8 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
     if (F eq IO.effectForIO) {
       asInstanceOf[F[A]]
     } else {
-      def fiberFrom[B](f: Fiber[F, Throwable, B]): IOFiber[B] =
-        new IOFiber[B] {
+      def fiberFrom[B](f: Fiber[F, Throwable, B]): FiberIO[B] =
+        new FiberIO[B] {
           val cancel = F.to[IO](f.cancel)
           val join = F.to[IO](f.join).map(_.mapK(F.toK[IO]))
         }
@@ -262,11 +262,11 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   }
 
   private[effect] def unsafeRunFiber(shift: Boolean)(cb: Either[Throwable, A] => Unit)(
-      implicit runtime: unsafe.IORuntime): FiberHandle[A @uncheckedVariance] = {
+      implicit runtime: unsafe.IORuntime): IOFiber[A @uncheckedVariance] = {
 
-    val fiber = new FiberHandle(
+    val fiber = new IOFiber(
       runtime.scheduler,
-      (oc: IOOutcome[A]) =>
+      (oc: OutcomeIO[A]) =>
         oc.fold((), e => cb(Left(e)), ioa => cb(Right(ioa.asInstanceOf[IO.Pure[A]].value))),
       0)
 
@@ -337,7 +337,7 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
   // utilities
 
-  def bothOutcome[A, B](left: IO[A], right: IO[B]): IO[(IOOutcome[A], IOOutcome[B])] =
+  def bothOutcome[A, B](left: IO[A], right: IO[B]): IO[(OutcomeIO[A], OutcomeIO[B])] =
     left.bothOutcome(right)
 
   def both[A, B](left: IO[A], right: IO[B]): IO[(A, B)] =
@@ -355,7 +355,7 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
   def racePair[A, B](
       left: IO[A],
-      right: IO[B]): IO[Either[(IOOutcome[A], IOFiber[B]), (IOFiber[A], IOOutcome[B])]] =
+      right: IO[B]): IO[Either[(OutcomeIO[A], FiberIO[B]), (FiberIO[A], OutcomeIO[B])]] =
     left.racePair(right)
 
   def toK[F[_]: Effect]: IO ~> F =
@@ -434,7 +434,7 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
       ioa.onCancel(fin)
 
     override def bracketCase[A, B](acquire: IO[A])(use: A => IO[B])(
-        release: (A, IOOutcome[B]) => IO[Unit]): IO[B] =
+        release: (A, OutcomeIO[B]) => IO[Unit]): IO[B] =
       acquire.bracketCase(use)(release)
 
     val monotonic: IO[FiniteDuration] = IO.monotonic
@@ -451,7 +451,7 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
     def racePair[A, B](
         fa: IO[A],
-        fb: IO[B]): IO[Either[(IOOutcome[A], IOFiber[B]), (IOFiber[A], IOOutcome[B])]] =
+        fb: IO[B]): IO[Either[(OutcomeIO[A], FiberIO[B]), (FiberIO[A], OutcomeIO[B])]] =
       fa.racePair(fb)
 
     override def race[A, B](fa: IO[A], fb: IO[B]): IO[Either[A, B]] =
@@ -460,7 +460,7 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
     override def both[A, B](fa: IO[A], fb: IO[B]): IO[(A, B)] =
       fa.both(fb)
 
-    def start[A](fa: IO[A]): IO[IOFiber[A]] =
+    def start[A](fa: IO[A]): IO[FiberIO[A]] =
       fa.start
 
     def uncancelable[A](body: IO ~> IO => IO[A]): IO[A] =
@@ -527,11 +527,11 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
   private[effect] case object Canceled extends IO[Unit] { def tag = 11 }
 
-  private[effect] final case class Start[A](ioa: IO[A]) extends IO[IOFiber[A]] {
+  private[effect] final case class Start[A](ioa: IO[A]) extends IO[FiberIO[A]] {
     def tag = 12
   }
   private[effect] final case class RacePair[A, B](ioa: IO[A], iob: IO[B])
-      extends IO[Either[(IOOutcome[A], IOFiber[B]), (IOFiber[A], IOOutcome[B])]] {
+      extends IO[Either[(OutcomeIO[A], FiberIO[B]), (FiberIO[A], OutcomeIO[B])]] {
 
     def tag = 13
   }
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -46,8 +46,24 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
 
   def both[B](that: IO[B]): IO[(A, B)] =
     racePair(that).flatMap {
-      case Left((a, f)) => f.joinAndEmbedNever.map((a, _))
-      case Right((f, b)) => f.joinAndEmbedNever.map((_, b))
+      case Left((oc, f)) => oc match {
+        case Outcome.Completed(fa) => f.join.flatMap {
+          case Outcome.Completed(fb) => fa.product(fb)
+          case Outcome.Errored(eb) => IO.raiseError(eb)
+          case Outcome.Canceled() => IO.canceled *> IO.never
+        }
+        case Outcome.Errored(ea) => IO.raiseError(ea)
+        case Outcome.Canceled() => IO.canceled *> IO.never
+      }
+      case Right((f, oc)) => oc match {
+        case Outcome.Completed(fb) => f.join.flatMap {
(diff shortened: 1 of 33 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -46,8 +46,24 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
 
   def both[B](that: IO[B]): IO[(A, B)] =
```
RaasAhsan, line 53: Seems like we could have a `bothOutcome` in a similar fashion to `raceOutcome`

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -101,13 +117,43 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
...
+        case Outcome.Completed(fa) => f.cancel *> fa.map(Left.apply)
+        case Outcome.Errored(ea) => f.join.flatMap {
+          case Outcome.Completed(fb) => fb.map(Right.apply)
+          case Outcome.Errored(eb) => IO.raiseError(eb)
+          case Outcome.Canceled() => IO.raiseError(ea)
+        }
+        case Outcome.Canceled() => f.join.flatMap {
+          case Outcome.Completed(fb) => fb.map(Right.apply)
+          case Outcome.Errored(eb) => IO.raiseError(eb)
+          case Outcome.Canceled() => IO.canceled *> IO.never
```
RaasAhsan: `IO.canceled *> IO.never` can deadlock if we're masked here

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -101,13 +121,50 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
 
   def race[B](that: IO[B]): IO[Either[A, B]] =
     racePair(that).flatMap {
-      case Left((a, f)) => f.cancel.as(Left(a))
-      case Right((f, b)) => f.cancel.as(Right(b))
+      case Left((oc, f)) =>
+        oc match {
+          case Outcome.Completed(fa) => f.cancel *> fa.map(Left.apply)
```
```suggestion
          case Outcome.Completed(fa) => f.cancel *> fa.map(Left(_))
```

And analogously for the others. I think it's just a bit cleaner.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -101,13 +121,50 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
...
+              case Outcome.Canceled() => IO.canceled *> IO.never
+            }
+        }
+      case Right((f, oc)) =>
+        oc match {
+          case Outcome.Completed(fb) => f.cancel *> fb.map(Right.apply)
+          case Outcome.Errored(eb) =>
+            f.join.flatMap {
+              case Outcome.Completed(fa) => fa.map(Left.apply)
+              case Outcome.Errored(ea) => IO.raiseError(ea)
```
djspiewak: BTW this is a slightly different semantic than before, where the *first* error would be emitted. I actually think we're better off just failing the race if one errors. I haven't checked on what ce2's semantics are here, but it just seems odd to me (in retrospect) that `race(raiseError(e), never)` would be a deadlock.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -101,13 +121,50 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
...
+              case Outcome.Canceled() => IO.canceled *> IO.never
+            }
+        }
+    }
+
+  def raceOutcome[B](that: IO[B])
+      : IO[Either[Outcome[IO, Throwable, A @uncheckedVariance], Outcome[IO, Throwable, B]]] =
```
djspiewak: Still not convinced this is actually useful. :-)

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -442,79 +438,30 @@ private[effect] final class IOFiber[A](name: String, scheduler: unsafe.Scheduler
...
-                        if (error != null) {
-                          // we were canceled, and the other errored, so use its error
-                          cb(Left(error))
-                        }
-                      }
-                    }
-                  }
+            val next = IO.async[Either[
+              (Outcome[IO, Throwable, Any], Fiber[IO, Throwable, Any]),
+              (Fiber[IO, Throwable, Any], Outcome[IO, Throwable, Any])]] { cb =>
```
djspiewak: We should like, make a type alias for this.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -442,79 +438,30 @@ private[effect] final class IOFiber[A](name: String, scheduler: unsafe.Scheduler
...
+                val ec = currentCtx
 
-                  execute(ec)(() => fiberA.run(cur.ioa, ec, masks))
-                  execute(ec)(() => fiberB.run(cur.iob, ec, masks))
+                execute(ec)(() => fiberA.run(cur.ioa, ec))
+                execute(ec)(() => fiberB.run(cur.iob, ec))
 
-                  Some(fiberA.cancel *> fiberB.cancel)
-                }
+                Some(fiberA.cancel *> fiberB.cancel)
```
djspiewak: Random thing that should probably be fixed… this should be run in parallel. We really want `IO.both(fiberA.cancel, fiberB.cancel).void`.

testkit/shared/src/main/scala/cats/effect/testkit/TimeT.scala
```diff
@@ -40,6 +40,11 @@ object TimeT {
   def liftF[F[_], A](fa: F[A]): TimeT[F, A] =
     Kleisli.liftF(fa)
 
+  def liftK[F[_], A]: F ~> TimeT[F, *] =
```
djspiewak: `A`?

djspiewak (changes requested):
My suggestion for the laws would be to remove most of the `race` laws and talk about `racePair` with a couple new ones instead. Also, to that end, `PureConc` needs to be updated.

testkit/shared/src/main/scala/cats/effect/testkit/TimeT.scala
RaasAhsan: I copied the type signature from `Kleisli.liftK` and didn't realize it was used 😅

core/shared/src/main/scala/cats/effect/IOFiber.scala
RaasAhsan: I was thinking about this too, maybe a specialized `Outcome` and `Fiber` for `IO` makes sense. When we're talking about either of those in the context of `IO`, `Throwable` is pretty much implied.

`IOOutcome`, `IOFiber` (this ones already taken but we can come up with better names for the implementation I think)

core/shared/src/main/scala/cats/effect/IO.scala
RaasAhsan: I tried to mimic the current CE3 semantics, it seems like the second error is reported https://github.com/typelevel/cats-effect/blob/series/3.x/core/shared/src/main/scala/cats/effect/IOFiber.scala#L484

core/shared/src/main/scala/cats/effect/IO.scala
RaasAhsan: I'm sure somebody could come up with a contrived example... but what do you think about `bothOutcome`?

core/shared/src/main/scala/cats/effect/IO.scala
djspiewak: Oh interesting! That wasn't intended. I think. I'd have to look at `PureConc`. In any case though, I think it's better to make the error report immediate; in other words, I think the old ce3 semantics are wrong.

core/shared/src/main/scala/cats/effect/IO.scala
djspiewak: `bothOutcome` feels a bit more useful. :-) If nothing else, `both(raiseError(e1), raiseError(e2))` is inherently weird, whereas it works fine with `bothOutcome`

RaasAhsan:
@djspiewak Updated the `PureConc.racePair` implementation, let me know if it looks right :)

The law for `racePair` aren't very elegant to write because of `Outcome` :) . My first attempt was something like this, but it doesn't pass exhaustively.
```scala
def racePairNeverIdentityLeft[A](fa: F[A]) = {
    val identity = F.racePair(F.never[Unit], fa).flatMap {
      case Left(_) => F.never[Outcome[F, E, A]]
      case Right((_, oc)) => F.pure(oc)
    }

    fa.map(a => Outcome.completed[F, E, A](F.pure(a))) <-> identity
  }
```

RaasAhsan:
Ah, well the concurrent laws for `PureConc` are failing which seems to suggest that the `racePair` implementation doesn't work.

```
[error] x concurrent.uncancelable race displaces canceled
[error]  Falsified after 0 passed tests.
[error]  > Labels of failing property:Expected: "Completed(None)\u000a│\u000a├ Notify 0x37D631D2\u000a├ Notify 0x37D631D2\u000a├ Notify 0x37D631D2\u000a╰ Done"
[error]  Received: "Completed(Some(()))\u000a│\u000a├ Notify 0x22CBF51E\u000a├ Notify 0x58164174\u000a├ Fork\u000a│ ╰╮\u000a│  ├ Notify 0x58164174\u000a│  ├ Notify 0x58164174\u000a│  ╰ Done\u000a│  \u000a├ Notify 0x73D34EA4\u000a├ Fork\u000a│ ╰╮\u000a│  ├ Notify 0x73D34EA4\u000a│  ├ Notify 0x73D34EA4\u000a│  ├ Notify 0x4E172C51\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a│  ├ Await 0x7C43D089\u000a\u000a...\u000a\u000a..."
[error]  The seed is P36o_rk8LlGX5TCkjlFJepb7OtUwwbENSNvQKwGZm_D=
[info] + concurrent.uncancelable race poll canceled identity (left)
[error] x concurrent.uncancelable race poll canceled identity (right)
[error]  Falsified after 0 passed tests.
[error]  > Labels of failing property:Expected: "Completed(None)\u000a│\u000a├ Notify 0xE638D54\u000a├ Notify 0xE638D54\u000a├ Notify 0xE638D54\u000a╰ Done"
[error]  Received: "Completed(Some(Left(1299570979)))\u000a│\u000a├ Notify 0x3AE912B2\u000a├ Notify 0x165C8DD4\u000a├ Fork\u000a│ ╰╮\u000a│  ├ Notify 0x165C8DD4\u000a│  ├ Notify 0x165C8DD4\u000a│  ├ Notify 0x4534522C\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a│  ├ Await 0x57AF7EF3\u000a\u000a...\u000a\u000a..."> ARG_0: "Completed(None)\u000a│\u000a├ Notify 0x4023AF0\u000a├ Notify 0x4023AF0\u000a├ Notify 0x4023AF0\u000a╰ Done"
[error]  The seed is ZhNu4goN-9O4XlJ2JWTzFrjI09c9Ih_XDQ0umO1IbEG=
```

I'm not really sure how to debug this thing without drowning a lot of time into figuring out how it really works, so let me know if you can spot what I looked over

djspiewak:
> Ah, well the concurrent laws for PureConc are failing which seems to suggest that the racePair implementation doesn't work.

Some of the old laws need to be removed! :-) There are a bunch of laws which claim that `uncancelable` should distribute over `racePair` and such. All of those need to go.

RaasAhsan:
@djspiewak I think all the `racePair` variants are going to have to wrapped in `uncancelable`, or else we might leak resources. Not sure if there are any reasonable regions we can poll inside. 

I want to get this branch merged as soon as fix these lingering `PureConc` issues, but I have it on my TODO to get to that afterwards. We should probably do a sweep of both `IO` and `Concurrent` machinery to see we're not missing any of these cases.

RaasAhsan:
I managed to isolate this failing `Align` test case for `PureConc`. It seems like the issue comes from either `both` or `racePair`. I've already made all the substitutions, but here's a reduced case that illustrate where things seem to go wrong:

```scala
// The following two are NOT the same when they participate in a `racePair`
fa.align(fb.map(identity))
fa.align(fb)

// If you make the appropriate substitutions, you end up here
val a = (ParallelF.value(ParallelF(fa).map(identity)).attempt, ParallelF.value(ParallelF(fb).map(identity)).attempt)
            .parMapN((ea, eb) => catsStdInstancesForEither.alignWith(ea, eb)(identity))
            .flatMap(C.fromEither)
// `ParallelF.value(ParallelF(fb).map(identity))` is the term of interest. that map call eventually makes its way to `both` and `racePair`, so something is wrong there
```

Here are the renderings of the different effects:
```scala
// fa
Completed(Some(1))
│
├ Notify 0x2E408AB9
├ Cede
├ Notify 0x15A9C7E3
╰ Pure Completed(1)

// fb
Completed(None)
│
├ Notify 0x3641C43B
├ Notify 0x3641C43B
├ Notify 0x3641C43B
╰ Done

// ParallelF.value(ParallelF(fa).map(identity).align(ParallelF(fb)))
Completed(Some(Ior.Left(1)))
│
├ Notify 0x1C152686
├ Notify 0x13BD612B
├ Fork
│ ╰╮
│  ├ Notify 0x13BD612B
│  ├ Notify 0x13BD612B
│  ├ Notify 0x5612A9E1
│  ├ Fork
│  │ ╰╮
│  │  ├ Notify 0x5612A9E1
│  │  ├ Notify 0x5612A9E1
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53
│  │  ├ Await 0x951FB53

...

...

...

// ParallelF.value(ParallelF(fa).map(identity).align(ParallelF(fb).map(identity)))
Completed(None)
│
├ Notify 0x20A079E1
├ Notify 0xCEE2D38
├ Fork
│ ╰╮
│  ├ Notify 0xCEE2D38
│  ├ Notify 0xCEE2D38
│  ├ Notify 0x2E102352
│  ├ Fork
│  │ ╰╮
│  │  ├ Notify 0x2E102352
│  │  ├ Notify 0x2E102352
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5
│  │  ├ Await 0x229EFF5

...

...

...
```

The failing seed for the `alignHomomorphism` test is `-ETrHGyCTzib58PxGaFyumO3n6Of8RjswGuqSDhQvnN=`.

djspiewak:
So, `fa` is something like `F.cede.as(1)`, while `fb` could be either `F.canceled` or `F.never`. Really nice minimization btw! I'll take a look at it in detail as soon as I can.

PR #962, opened 2020-07-19
Fixed cancelation semantics to block when canceling uncancelable

djspiewak:
I believe this is the only thing that makes sense, otherwise you lose finalizer backpressure in those cases.

### DIFF
--- a/core/shared/src/test/scala/cats/effect/IOSpec.scala
+++ b/core/shared/src/test/scala/cats/effect/IOSpec.scala
@@ -180,7 +180,7 @@ class IOSpec extends IOPlatformSpecification with Discipline with ScalaCheck wit
           _ <- f.cancel
         } yield ()
 
-        ioa must nonTerminate  // we're canceling an uncancelable never
+        ioa must nonTerminate // we're canceling an uncancelable never
         affected must beFalse
     }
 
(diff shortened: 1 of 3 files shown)


core/shared/src/test/scala/cats/effect/IOSpec.scala
```diff
@@ -528,6 +528,42 @@ class IOSpec extends IOPlatformSpecification with Discipline with ScalaCheck wit
       IO.cede.foreverM.start.flatMap(f => IO.sleep(50.millis) >> f.cancel).as(ok)
     }
 
+    "await uncancelable blocks in cancelation" in ticked { implicit ticker =>
+      var started = false
+
+      val markStarted = IO { started = true }
+      lazy val cedeUntilStarted: IO[Unit] =
+        IO(started).ifM(IO.unit, IO.cede >> cedeUntilStarted)
```
RaasAhsan, line 536: Poor man's `Deferred` 😂

RaasAhsan (approved):
Much easier to follow `cancel` now

core/shared/src/test/scala/cats/effect/IOSpec.scala
djspiewak: LOL exactly

PR #969, opened 2020-07-21
Implemented postMessage polyfill for setImmediate

djspiewak:
This is basically a polyfill for `setImmediate` on Firefox, Chrome, and other non-IE10+ browsers. The one modern case which isn't covered by this is web workers, which will have to wait until I can come up with a good way to test it.

### DIFF
--- a/core/js/src/main/scala/cats/effect/unsafe/PolyfillExecutionContext.scala
+++ b/core/js/src/main/scala/cats/effect/unsafe/PolyfillExecutionContext.scala
@@ -16,9 +16,10 @@
 
 package cats.effect.unsafe
 
-// import scala.collection.mutable
+import scala.collection.mutable
 import scala.concurrent.ExecutionContext
 import scala.scalajs.js
+import scala.util.Random
 
 /**
  * Based on https://github.com/YuzuJS/setImmediate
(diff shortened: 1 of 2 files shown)


core/js/src/main/scala/cats/effect/unsafe/PolyfillExecutionContext.scala
```diff
@@ -34,21 +35,86 @@ private[unsafe] object PolyfillExecutionContext extends ExecutionContext {
...
 
-      }*/
-      { k =>
-        js.Dynamic.global.setTimeout(k, 0)
         ()
       }
+
+      if (canUsePostMessage()) {
+        // postMessage is what we use for most modern browsers (when not in a webworker)
+        val messagePrefix = "setImmediate$" + Random.nextInt() + "$"
```
ritschwumm, line 90: what's the randomness good for here? what happens if `Random.nextInt()` returns the same value twice?

djspiewak: I just ported it directly from the JavaScript polyfill. The point to the randomness is to cheaply ensure uniqueness. This is only done once (not per yield), so it shouldn't matter much. We *need* a unique message prefix for the executor, otherwise we won't be able to efficiently rule out `postMessage` events that we ourselves didn't post.

ritschwumm: ok, so it's possible to cause problems, but quite unlikely. i think this deserves a comment in the code explaining why it's there and why it's unlikely to cause problems. sorry for being picky, but i've seen too many randomly generated "unique" identifiers where the code in question was actually executed a bit more often than expected.

djspiewak: Not nit-picky at all! This is good review feedback. I'll comment it.

PR #974, opened 2020-07-22
Protect racePair variants against fiber leaks with uncancelable

RaasAhsan:
`racePair` is a primitive operation that allocates a resource (fiber) that the caller is ultimately responsible for cleaning up properly. All `racePair` variants generally take the form `racePair(a, b).flatMap`. The presence of a bind continuation means that cancellation can be observed, and so in this case we could leak fibers. The solution is to wrap all these variants in `uncancelable` and `poll` during `join`s, ensuring that we cancel the remaining fiber on cancellation.

When we get around to writing documentation, we should make it clear that `racePair` is safe only in conjunction with `uncancelable`. I think it would be a good idea to capture commonly used `racePair` patterns so users don't need to think about cleanup. 

I was thinking about how to test all this cancellation behavior we're adding, and the most appealing answer to me is with `Deferred`. Ideally this goes in `IOSpec` but IIRC the `core` module doesn't rely on `concurrent`, but maybe the test scope could.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -45,35 +45,39 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
     map(Right(_)).handleErrorWith(t => IO.pure(Left(t)))
 
   def bothOutcome[B](that: IO[B]): IO[(OutcomeIO[A @uncheckedVariance], OutcomeIO[B])] =
-    racePair(that).flatMap {
-      case Left((oc, f)) => f.join.map((oc, _))
-      case Right((f, oc)) => f.join.map((_, oc))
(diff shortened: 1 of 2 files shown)


RaasAhsan:
`start` actually has the same problem as `racePair`, where it allocates a fiber and leaves clean up to the user. `Concurrent.background` is the safe version of `start` that handles cancellation for you. It might not be necessary but I wonder if we can do something similar for `racePair`.

djspiewak:
> start actually has the same problem as racePair, where it allocates a fiber and leaves clean up to the user. Concurrent.background is the safe version of start that handles cancellation for you. It might not be necessary but I wonder if we can do something similar for racePair.

Hmm. The type signature would be relatively straightforward. I think just the outer effect would be wrapped in `Resource`.

PR #977, opened 2020-07-22
Improve CommutativeApplicative instance for ParallelF

TimWSpence:


djspiewak (approved):
Needs scalafmt but otherwise looks great!

PR #979, opened 2020-07-22
Port LiftIO to CE3

### LINKED ISSUE #972: Add LiftIO to ce3
We should restore `LiftIO`, but in the **core** module and extending `Async`.

vasilmkd:
Resolves #972.

A little help with the dotty compilation issue... Please.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/LiftIO.scala
+++ b/core/shared/src/main/scala/cats/effect/LiftIO.scala
@@ -20,11 +20,6 @@ import cats.{~>, Applicative, FlatMap, Functor}
 import cats.data.{ContT, EitherT, IorT, Kleisli, OptionT, ReaderWriterStateT, StateT, WriterT}
 import cats.kernel.Monoid
 
-import scala.annotation.implicitNotFound
-
-@implicitNotFound("""Cannot find implicit value for LiftIO[${F}].
-Building this implicit value might depend on having an implicit
-s.c.ExecutionContext in scope, a Scheduler or some equivalent type.""")
 trait LiftIO[F[_]] {
   def liftIO[A](ioa: IO[A]): F[A]
 }
--- a/core/shared/src/main/scala/cats/effect/LiftIO.scala
+++ b/core/shared/src/main/scala/cats/effect/LiftIO.scala
@@ -1,0 +1,124 @@
+/*
+ * Copyright 2020 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+
+import cats.{~>, Applicative, Functor}
+import cats.data.{EitherT, IorT, Kleisli, OptionT, ReaderWriterStateT, StateT, WriterT}
+import cats.kernel.Monoid
+
+import scala.annotation.implicitNotFound
+
+@implicitNotFound("""Cannot find implicit value for LiftIO[${F}].
+Building this implicit value might depend on having an implicit
+s.c.ExecutionContext in scope, a Scheduler or some equivalent type.""")
+trait LiftIO[F[_]] {
+  def liftIO[A](ioa: IO[A]): F[A]
+}
+
+object LiftIO {
+
+  def apply[F[_]](implicit F: LiftIO[F]): F.type = F
+
+  /**
+   * [[LiftIO.liftIO]] as a natural transformation
+   */
+  def liftK[F[_]: LiftIO]: IO ~> F = λ[IO ~> F](LiftIO[F].liftIO(_))
+
+  /**
+   * [[LiftIO]] instance built for `cats.data.EitherT` values initialized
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/cats/effect/LiftIO.scala
```diff
@@ -0,0 +1,124 @@
...
+}
+
+object LiftIO {
+
+  def apply[F[_]](implicit F: LiftIO[F]): F.type = F
+
+  /**
+   * [[LiftIO.liftIO]] as a natural transformation
+   */
+  def liftK[F[_]: LiftIO]: IO ~> F = λ[IO ~> F](LiftIO[F].liftIO(_))
```
djspiewak: > A little help with the dotty compilation issue... Please.

The issue is that kind-projector doesn't exist on Dotty! I suppose I should technically remove it from the main project too, but I've been lazy.

The rewrite is pretty trivial though.

```suggestion
  def liftK[F[_]](implicit F: LiftIO[F]): IO ~> F = 
    new (IO ~> F) { def apply[A](fa: F[A]) = F.liftIO(fa) }
```

I also changed the instance to capture once, since it's marginally more efficient to do so.

It's rather annoying that we can't use type lambdas, but until Dotty improves its support for higher-rank type inference, we can't even shim it. :-(

vasilmkd: Yeah, this looks reasonable. I don't follow Dotty much, sorry.

djspiewak: No worries! We aren't using any of the fancy stuff, because we're cross building. So honestly it's easiest to think of Dotty in our case as "Scala with way fewer bugs and no macros or plugins".

mpilquist: @djspiewak FYI, Dotty supports `-Ykind-projector` with some limitations

djspiewak: @mpilquist Yeah, we're using it, but `-Ykind-projector` doesn't support the value-level aspects of kind-projector. I spent a bit of time trying to build a Dotty-specific shim using higher-rank types to get around this, but unfortunately it appears that higher-rank types don't currently infer at all in Dotty (@smarter wants to work on this when he gets time), so we're stuck using anonymous inner classes everywhere.

djspiewak (changes requested):
Looks really really good! I think it's just missing two instances though. First, you need an instance for `IO` (so, `LiftIO[IO]`), otherwise it's impossible to lift anything! :-) Second, it's worth tossing an instance in here for `ContT`, even though no one ever uses it.

core/shared/src/main/scala/cats/effect/LiftIO.scala
```diff
@@ -0,0 +1,143 @@
...
+
+import cats.{~>, Applicative, FlatMap, Functor}
+import cats.data.{ContT, EitherT, IorT, Kleisli, OptionT, ReaderWriterStateT, StateT, WriterT}
+import cats.kernel.Monoid
+
+import scala.annotation.implicitNotFound
+
+@implicitNotFound("""Cannot find implicit value for LiftIO[${F}].
+Building this implicit value might depend on having an implicit
+s.c.ExecutionContext in scope, a Scheduler or some equivalent type.""")
```
djspiewak: Oh I forgot: we don't need this anymore. I would just remove this annotation entirely.

core/shared/src/main/scala/cats/effect/LiftIO.scala
```diff
@@ -0,0 +1,143 @@
...
+    new LiftIO[ContT[F, R, *]] {
+      override def liftIO[A](ioa: IO[A]): ContT[F, R, A] =
+        ContT.liftF(F.liftIO(ioa))
+    }
+
+  /**
+   * [[LiftIO]] instance for [[IO]] values.
+   */
+  implicit val ioLiftIO: LiftIO[IO] =
+    new LiftIO[IO] { override def liftIO[A](ioa: IO[A]): IO[A] = ioa }
```
djspiewak, line 137: Just calling this out for discussion quickly: is it better to put this instance here, or on the `IO` companion?

djspiewak (changes requested):
Looking really good! Just one minor change and a general question for the room (that I don't know the answer to).

core/shared/src/main/scala/cats/effect/LiftIO.scala
vasilmkd: I feel like I'm being quizzed. 😆 If it remains in `LiftIO`, probably `LiftIO` companion object needs to be in scope. But when we work with `IO`, the companion object for `IO` is always in scope, so maybe it's better if it's on there.

core/shared/src/main/scala/cats/effect/LiftIO.scala
vasilmkd: Done.

core/shared/src/main/scala/cats/effect/LiftIO.scala
djspiewak: Well, in theory scalac should just look in both places. Specifically, if it's trying to resolve `LiftIO[IO]`, it will look on both `LiftIO`'s companion and on `IO`'s companion, because they're both a part of the type. Similarly for `LiftIO[OptionT[IO, *]]`. I *think* it looks on `LiftIO` first, so it might be marginally faster to compile this way, but I don't know if there are other disadvantages.

I was mostly asking in hope that someone else would chime in with a "you should do it this way!" since I really don't know the answer. :-) Let's go with what you implemented for now and we can change things if we don't like it!

vasilmkd: FWIW, all of the `cats.data` instances were already on `LiftIO` in CE2.

PR #980, opened 2020-07-23
Initial sketch of Sync.blocking

mpilquist:


mpilquist:
I have no idea why scalafmt is failing on CI - it doesn't fail locally for me.

djspiewak:
@mpilquist It's worth using the `prePR` action. Apparently `scalafmtAll` actually has some internal notion of caching which `clean` flushes. I dunno. It's annoying.

So on my local branch, I swapped out `IO.Delay` for the following:

```scala
  private[effect] final case class Suspend[+A](thunk: () => A, hintTag: Byte) extends IO[A] {
    def tag = 1
  }
```

I think it's worth benchmarking whether the increased expense of allocation on `Suspend` (vs `Delay`) is better or worse than any potential concerns which arise from having extra tags. What I'm worried about is specifically the `tableswitch` inside of `IOFiber#runLoop`. I don't *think* it will deoptimize, but there's a threshold in here somewhere which is tricky (and also not where everyone says it is), so it's worth measuring (I usually just run the *Bind* benchmarks for this kind of thing) whether the added `Blocking` node actually causes a drop-off in performance. If not, then it's definitely the way to go.

So regarding `Sync` itself, I was kind of thinking about something like this:

```scala
  def delay[A](thunk: => A): F[A] = suspend(thunk, Delay)

  def blocking[A](thunk: => A): F[A] = suspend(thunk, Blocking)

  def interruptible[A](loop: Boolean)(thunk: => A): F[A] =
    suspend(thunk, if (loop) InterruptibleMany else InterruptibleOnce)

  def suspend[A](thunk: => A, hint: Suspension): F[A]
```

`Suspension` is a terrible name, but the definition is obvious. Anyway, the idea here was simply to collapse it all together for the purposes of inductive instances as well as law definitions, since we know that all three of these functions need to behave according to the same laws.

I'm not entirely sure if this is a good idea or not, but it does make some things easier.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -288,10 +296,33 @@ private[effect] final class IOFiber[A](name: String, scheduler: unsafe.Scheduler
             runLoop(next, nextIteration)
 
           case 2 =>
+            val cur = cur0.asInstanceOf[Blocking[Any]]
+
+            execute(blockingEc) { () =>
```
djspiewak: BTW, we can actually skip the read barrier here and just do `blockingEc.execute`. We only need the read barrier when trampolining the `runLoop`.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -288,10 +296,33 @@ private[effect] final class IOFiber[A](name: String, scheduler: unsafe.Scheduler
...
+                  case NonFatal(t) => t
+                }
+
+              val next =
+                if (success)
+                  succeeded(r, 0)
+                else
+                  failed(r, 0)
+
+              execute(currentCtx) { () => runLoop(next, nextIteration) }
```
djspiewak: …like here! :-)

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -288,10 +296,33 @@ private[effect] final class IOFiber[A](name: String, scheduler: unsafe.Scheduler
...
+                  r
+                } catch {
+                  case NonFatal(t) => t
+                }
+
+              val next =
+                if (success)
+                  succeeded(r, 0)
+                else
+                  failed(r, 0)
```
djspiewak: This is a bit tricky, but it needs to be in the `currentCtx` executor, since this is where `Map`, `FlatMap`, and `Delay` are evaluated.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -288,10 +296,33 @@ private[effect] final class IOFiber[A](name: String, scheduler: unsafe.Scheduler
             runLoop(next, nextIteration)
 
           case 2 =>
+            val cur = cur0.asInstanceOf[Blocking[Any]]
+
+            execute(blockingEc) { () =>
+              var success = false
```
djspiewak: If you lambda lift a `var` but that `var` is never modified again, does it still get boxed? I *think* it does, since scalac doesn't really have much in the way of escape analysis, but I'm not sure. If it does, then we need to copy it to a `val` after the the `try`/`catch` and reference the *`val`* from the closure that we pass to the executor.

core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
```diff
@@ -33,13 +33,18 @@ avoid unintentionally degrading your application performance.
...
 
   override def toString: String = s"IORuntime($compute, $scheduler)"
 }
 
 object IORuntime extends IORuntimeCompanionPlatform {
-  def apply(compute: ExecutionContext, scheduler: Scheduler, shutdown: () => Unit): IORuntime =
-    new IORuntime(compute, scheduler, shutdown)
+  def apply(
+      compute: ExecutionContext,
+      blocking: ExecutionContext,
```
djspiewak, line 46: I did the same thing locally, but I think we'll probably need something different. I haven't thought it through completely. Basically, `interrupted()` needs to be able to get at the `Thread` (if it exists).

core/shared/src/main/scala/cats/effect/IOFiber.scala
RaasAhsan: We wouldn't need a read barrier on either `execute` then right? No state to publish before shifting, and the only state we need to keep after shifting back is `r`, which is already closed over

core/shared/src/main/scala/cats/effect/IOFiber.scala
RaasAhsan: I know that sometimes `try` gets compiled to a separate function, which causes boxing of `var`s, but I don't know what the precise conditions for that are, it's very poorly defined. This looks fine to me though since the `var` is inside the lambda, and the `try` expression is being assigned to a new `val`

SystemFw:
Small thought: we have this on `Resource`

```
def fromAutoCloseableBlocking[F[_]: Sync: ContextShift, A <: AutoCloseable](blocker: Blocker)(acquire: F[A]): Resource[F, A] =
   Resource.make(blocker.blockOn(acquire))(autoCloseable => blocker.delay(autoCloseable.close()))
```

Did we reach a conclusion yesterday? Will this take `ec` now?

mpilquist:
@SystemFw I think we should change that to:

```scala
def fromAutoCloseableBlocking[F[_]: Sync, A <: AutoCloseable](acquire: => A): Resource[F, A] =
   Resource.make(Sync[F].blocking(acquire))(autoCloseable => Sync[F].blocking(autoCloseable.close()))
```

t3hnar:
wow! this is nice!

I'm using https://github.com/evolution-gaming/cats-helper/blob/master/core/src/main/scala/com/evolutiongaming/catshelper/Blocking.scala#L38

and looks like I won't need any changes to my code :)

core/shared/src/main/scala/cats/effect/IOFiber.scala
djspiewak: Oh that's a good point. I couldn't quite remember why we were using execute. We basically need to do whatever Cede does.

core/shared/src/main/scala/cats/effect/IOFiber.scala
djspiewak: Well, I know the try here is fine. I'm more worried about the fact that we need the success var to be read inside of the lambda we send to the executor, because of the continuation logic.

core/shared/src/main/scala/cats/effect/IOFiber.scala
RaasAhsan: I think we want to use `execute()` when we want to insert a cancellation boundary. It looks like `Cede` just does `currentCtx.execute`, so no cancellation boundary there

core/shared/src/main/scala/cats/effect/IOFiber.scala
RaasAhsan: It will definitely be boxed in the new code segment. Before only `next` was being closed when trampolining. Now `success` is being closed over as well

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
```diff
@@ -23,7 +23,7 @@ import java.util.concurrent.atomic.AtomicInteger
 
 private[unsafe] abstract class IORuntimeCompanionPlatform { this: IORuntime.type =>
   def createDefaultComputeExecutionContext(
-      threadPrefix: String = "io-compute-"): (ExecutionContext, () => Unit) = {
+      threadPrefix: String = "io-compute"): (ExecutionContext, () => Unit) = {
```
djspiewak, line 26: Lol I really never caught the extra hyphen…

djspiewak (approved):
Still needs a quick benchmark run to make sure that we haven't messed up `tableswitch`. I'll try to do that today if no one else gets to it first.

djspiewak:
Ran the benchmarks:

**My Approach**

```
[info] Benchmark                   (size)   Mode  Cnt     Score      Error  Units
[info] DeepBindBenchmark.delay      10000  thrpt    3  6116.020 ± 1918.531  ops/s
[info] ShallowBindBenchmark.delay   10000  thrpt    3  5810.891 ±  568.632  ops/s
```

**This PR**

```
[info] Benchmark                   (size)   Mode  Cnt     Score      Error  Units
[info] DeepBindBenchmark.delay      10000  thrpt    3  6271.680 ± 1020.857  ops/s
[info] ShallowBindBenchmark.delay   10000  thrpt    3  5754.813 ±  353.173  ops/s
```

Within margin of error, basically. I could get a finer grained measurement if I ran them for longer, but overall it seems fine either way.

PR #984, opened 2020-07-23
Generalized Sync to arbitrary suspension hints

### LINKED ISSUE #958: Remove suspend in favor of defer
The `suspend` function (type signature `(=> F[A]) => F[A]`, so basically `flatten . delay`) dates back all the way to `scalaz.concurrent.Task`. Its primary contemporary usages surround a) smoothing migration of legacy imperative code which mixes functional and side-effects, and b) recursion of effectful functions. Cats Effect originally carried this function along, despite the obvious redundancy with `delay` + `flatten`.

Since then, the `Defer` typeclass was introduced in Cats, which has an *identical* type signature, but with confusingly different guarantees (since `Defer` doesn't talk at all about exceptions, and explicitly does not claim to correctly capture side-effects). The point of `Defer` is to abstract over lazy suspension. It's basically a lawfully guaranteed stack-safe `flatten`, which `Sync` *also* guarantees with vanilla `flatten`.

There's a ton of obvious redundancy here. `suspend` in Cats Effect is entirely indistinguishable from `defer` on `Sync`-forming types, and really `suspend` is just `defer` with exception handling and guarantees around non-memoization. I don't really see a need to stick with the larger API footprint; I would like to simply remove `suspend`.

Plenty of classes enrich existing functions with stricter laws at later points in the hierarchy. `ap` is more tightly constrained in `Monad`, for example. I see no reason *not* to do the same thing with `Defer`. Thus, any `Defer` which is also a `Sync` is further-constrained to follow the same semantics as `delay` + `flatten`, and thus `defer` and `suspend` become aliases for each other, and we can simply remove `suspend` entirely.

Opinions?

djspiewak:
This is a generalization of what @mpilquist did in #980, particularly restoring lawfulness of `blocking` and all related combinators. Also requesting eyes from @alexandru on this. Note that the semantic of `suspend` is that the `hint` is just that: a hint. Implementations are always free to ignore it, particularly if it is irrelevant to them. (I'll probably actually override the `IO` implementation on ScalaJS to just all point to `delay`, since there's no point in bouncing through `setImmediate` in any of them).

I still haven't implemented `interruptible`, and when I do it will mean some signature changes in `IORuntime`, but I need to think about exactly what that will be.

Fixes #958

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -203,7 +203,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
       this match {
         case IO.Pure(a) => F.pure(a)
         case IO.Delay(thunk) => F.delay(thunk())
-        case IO.Blocking(thunk) => F.blocking(thunk())
+        case IO.Blocking(hint, thunk) => F.suspend(hint)(thunk())
         case IO.Error(t) => F.raiseError(t)
         case IO.Async(k) => F.async(k.andThen(_.to[F].map(_.map(_.to[F]))))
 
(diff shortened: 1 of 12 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -522,8 +538,13 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
     override def toString: String = s"IO($value)"
   }
 
+  // we keep Delay as a separate case as a fast-path, since the added tags don't appear to confuse HotSpot (for reasons unknown)
   private[effect] final case class Delay[+A](thunk: () => A) extends IO[A] { def tag = 1 }
-  private[effect] final case class Blocking[+A](thunk: () => A) extends IO[A] { def tag = 2 }
+
+  private[effect] final case class Blocking[+A](hint: Sync.Type, thunk: () => A) extends IO[A] {
+    def tag = 2
```
daddykotex, line 545: What's the usage of `tag` (curious and I don't have access to source right now)

daddykotex: Ah nvm, I just saw it below. It's used in a pattern match when interpreting the IO

djspiewak: Yep! It gets compiled to a `tableswitch`, which is vastly faster than the corresponding `if`/`else` chain that you would normally get out of pattern matching.

Issue #987, opened 2020-07-23
Make subsequent cancels await the first one


djspiewak:
```scala
    "return immediately from double cancel of sleeping thing" in real {
      IO.never
        .onCancel(IO.sleep(1.hour))
        .start
        .flatMap(f => IO.racePair(f.cancel, f.cancel))
        .as(ok)
    }
```

This test should *fail*. It passes right now because one of the two `f.cancel`s will "lose" the race and complete immediately.

RaasAhsan:
For some more context, follow the discussion at https://gitter.im/typelevel/cats-effect-dev?at=5f1a0423c83e6242ed4e27dc. It goes on for quite awhile

PR #990, opened 2020-07-23
Interruptible Implementation (and bonus bugfixes)

djspiewak:
Implemented `interruptible`, which is cool. Also fixed some bugs, which is even cooler. Also unleashed a firestorm of confusion about semantics, which became #987 

Depends on #984


djspiewak:
I know what's happening here. It's really tricky and loosely related to #987. The problem is the *continuation* of an `Async` within an `Async` finalizer. Like, for example, the one within `interruptible`'s implementation. When an `Async` finalizer is itself an `Async` and the outer `Async` is canceled while suspended, the `Async` finalizer will run immediately. If that finalizer is suspended after registration fully completes, it will resume and then *immediately check its cancelation flag… which will be `true`, because its fiber was canceled. We have no particular flag for "finalizing", which I think we probably need.

This, btw, I think was the reason that I call `done` early: so that it can be checked in `Async`'s `cur.k` implementation. With that said, we don't *need* to check `outcome` there since we pass through the `suspended` write barrier (when we `resume()`), so we can just rely on `canceled`. So this is the relation to the other ticket.

But we really do need a `finalizing` state, since it changes the meaning of so many things, and inferring it is tricky.

djspiewak:
As an aside, I did some digging, and part of the reason we set `outcome` early is so that we can check it in `loop` and `asyncContinue`. When an `Async` finishes after cancelation starts (but before it completes), we don't want to fork the runloop. This is easily avoided by `suspended`. However, when cancelation has *already* completed… and then the `Async` completes… we need to prevent the runloop from being resurrected. This is not so easily accomplished, and even setting `masks = initMask` and allowing `canceled` to remain `true` doesn't seem to be enough. I think being more explicit about this particular scenario is what's needed.

djspiewak:
Depends on #998

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -330,28 +333,99 @@ private[effect] final class IOFiber[A](
...
+
+                  target <- IO.async_[Thread] { initCb =>
+                    blockingEc execute { () =>
+                      initCb(Right(Thread.currentThread()))
+
+                      try {
+                        nextCb(Right(cur.thunk()))
+                      } catch {
+                        case _: InterruptedException =>
+                          println("caught interrupt")
```
RaasAhsan: stray println

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -330,28 +333,99 @@ private[effect] final class IOFiber[A](
...
+                          }
+
+                        case NonFatal(t) =>
+                          nextCb(Left(t))
+                      }
+                    }
+                  }
 
-                runLoop(next, nextIteration)
+                  isSelf <- IO(target eq Thread.currentThread())
```
RaasAhsan: So the only way this could happen on the JVM is if we're using the same thread pool for both `compute` and `blocking`, correct? And if that's the case, we know for a fact that the blocking operation has completed, because the thread would have had to complete the operation above in order to reach this point, so there's nothing to interrupt anymore

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -330,28 +333,99 @@ private[effect] final class IOFiber[A](
...
+                } yield {
+                  if (isSelf) {
+                    None
+                  } else {
+                    Some {
+                      IO async { finCb =>
+                        val trigger = IO {
```
RaasAhsan: What happens if the blocking thread completes its operation, and it picks up another task, but before the finalizer is de-registered, we `cancel`, which runs the finalizer? We might be interrupting or killing a `Thread` that's running someone else's work now.

If we use `done` to communicate that the blocking operation is or isn't done, then I think we still have a race condition, where the finalizer observes that `done` is false, so it decides to `Thread.interrupt()`, but before it actually starts to interrupt, the thread finishes the blocking operation and picks up another task.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -330,28 +333,99 @@ private[effect] final class IOFiber[A](
...
+                        nextCb(Right(cur.thunk()))
+                      } catch {
+                        case _: InterruptedException =>
+                          println("caught interrupt")
+                          if (many) {
+                            done.set(true)
+                          } else {
+                            val cb0 = cb.get()
+                            if (cb0 != null) {
+                              cb0()
```
RaasAhsan: If I'm understanding correctly, this is how we backpressure cancellers on interruption (particularly when we `Thread.interrupt` only once)?

core/shared/src/main/scala/cats/effect/IOFiber.scala
djspiewak: Yep! This is like a random emergency check to make sure we don't interrupt ourself, since that's one situation in which things would dramatically fall apart, and also a situation which is relatively easy to check.

core/shared/src/main/scala/cats/effect/IOFiber.scala
djspiewak: This is a very *very* good point. I need to think about this, but you're 100% right.

core/shared/src/main/scala/cats/effect/IOFiber.scala
djspiewak: Yep!

djspiewak:
@RaasAhsan Ready for another look! I think I solved the race condition.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -339,8 +343,118 @@ private[effect] final class IOFiber[A](
...
+                      IO async { finCb =>
+                        val trigger = IO {
+                          if (!many) {
+                            cb.set(() => finCb(Right(())))
+                          }
+
+                          // if done is false, and we can't get the semaphore, it means
+                          // that the action hasn't *yet* started, so we busy-wait for it
+                          var break = false
+                          while (break && !done.get()) {
```
RaasAhsan: Is this while loop dead code? `false && x == false`

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -339,8 +343,118 @@ private[effect] final class IOFiber[A](
...
+
+                  canInterrupt <- IO(new juc.Semaphore(1))
+
+                  target <- IO.async_[Thread] { initCb =>
+                    blockingEc execute { () =>
+                      initCb(Right(Thread.currentThread()))
+
+                      try {
+                        canInterrupt.release(1)
+                        nextCb(Right(cur.thunk()))
```
RaasAhsan: Should we only run `cur.thunk()` in the interruptible region? Just so that nothing in `nextCb` is interrupted (I think the only possibility is a blocking `put` on the underlying task queue)

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -339,8 +343,118 @@ private[effect] final class IOFiber[A](
...
+               * 6. Action completed, finalizer unregistered
+               */
+
+              val many = cur.hint eq TypeInterruptibleMany
+              val next = IO.async[Any] { nextCb =>
+                for {
+                  done <- IO(new AtomicBoolean(false))
+                  cb <- IO(new AtomicReference[() => Unit](null))
+
+                  canInterrupt <- IO(new juc.Semaphore(1))
```
RaasAhsan: Do we want to start the semaphore at 0 permits? The execution I'm thinking about is, what if we interrupt before the blocking operation is run? `tryAcquire` would be able to acquire a permit

core/shared/src/main/scala/cats/effect/IOFiber.scala
djspiewak: …how did the tests not catch this

core/shared/src/main/scala/cats/effect/IOFiber.scala
djspiewak: Good call!

core/shared/src/main/scala/cats/effect/IOFiber.scala
djspiewak: Excellent point

djspiewak:
@RaasAhsan Ready for another look. I'm pretty sure it actually works correctly now, though it does lean somewhat on an implementation detail of `IO`, which is that we check for cancelation at the *beginning* of evaluating a node, rather than at the end, meaning that the finalizer always gets registered properly.

RaasAhsan:
I will take a look at this as soon as I can, been a little bit busy

djspiewak:
> I will take a look at this as soon as I can, been a little bit busy

Absolutely no worries!

core/jvm/src/main/scala/cats/effect/IOFiberPlatform.scala
```diff
@@ -0,0 +1,141 @@
...
+              }
+
+              // if done is false, and we can't get the semaphore, it means
+              // that the action hasn't *yet* started, so we busy-wait for it
+              var break = true
+              while (break && !done.get()) {
+                if (canInterrupt.tryAcquire()) {
+                  try {
+                    target.interrupt()
+                    break = false
```
RaasAhsan: Should we move this to the `finally` block, or before the `interrupt` call? `Thread.interrupt` can supposedly throw a `SecurityException` so this could in fact interrupt more than once.

core/jvm/src/main/scala/cats/effect/IOFiberPlatform.scala
```diff
@@ -0,0 +1,141 @@
...
+              }
+            }
+          }
+        }
+      } yield {
+        Some {
+          IO async { finCb =>
+            val trigger = IO {
+              if (!many) {
+                cb.set(() => finCb(Right(())))
```
RaasAhsan: I don't feel too strongly about it, but could avoid 2 allocations here if we just pass in `finCb` and allocate `Right(())` once similar to whats done in `IOFiber`

core/jvm/src/main/scala/cats/effect/IOFiberPlatform.scala
```diff
@@ -0,0 +1,141 @@
...
+     * 4. Action completed, finalizer registered
+     * 5. Action completed, finalizer firing
+     * 6. Action completed, finalizer unregistered
+     */
+
+    val many = cur.hint eq TypeInterruptibleMany
+
+    IO.async[Any] { nextCb =>
+      for {
+        done <- IO(new AtomicBoolean(false))
```
RaasAhsan, line 52: Is it worth allocating `done`, `cb` and `canInterrupt` once as instance variables?

core/jvm/src/main/scala/cats/effect/IOFiberPlatform.scala
```diff
@@ -0,0 +1,141 @@
...
+                    break = false
+                  } finally {
+                    canInterrupt.release()
+                  }
+                }
+              }
+            }
+
+            val repeat = if (many) {
+              IO blocking {
```
RaasAhsan: Should we be busy spinning on a blocking pool?

core/jvm/src/main/scala/cats/effect/IOFiberPlatform.scala
RaasAhsan: Okay, so what happens when the blocking operation above completes, but the the finalizer is not yet popped, and some actor tries to cancel, interrupting only once? The callback is set in `cb` but it would never be invoked, so the canceller is (asynchronously) blocked forever.

core/jvm/src/main/scala/cats/effect/IOFiberPlatform.scala
djspiewak: Hmm, that would certainly speed up `interruptible`, but at the cost of slowing down `start` and making fibers somewhat more heavyweight, even when `interruptible` isn't used. Is it worth it? `interruptible` doesn't seem like a particularly commonly-needed operation.

core/jvm/src/main/scala/cats/effect/IOFiberPlatform.scala
djspiewak: Hmm. Good point. This should be on the compute pool.

core/jvm/src/main/scala/cats/effect/IOFiberPlatform.scala
djspiewak: Nice catch! Fixed.

djspiewak:
@RaasAhsan Addressed review comments aside from the instance allocation of `done` etc, which merits further discussion. I lean *slightly* 👎 on that one.

core/jvm/src/main/scala/cats/effect/IOFiberPlatform.scala
```diff
@@ -0,0 +1,149 @@
...
+                      canInterrupt.release()
+                    }
+                  }
+                }
+
+                finCb(RightUnit)
+              }
+            } else {
+              IO {
+                if (done.get() && cb.get() != null) {
```
RaasAhsan, line 136: Apologies to do this to you again... :) But what if the blocking operation above succeeds and `done` is not yet set to `true`? This can happen even with the `Semaphore` protection because `cur.thunk()` can reset the interrupt status in response to the first interrupt, so `canInterrupt.acquire()` doesn't throw an `InterruptedException`.

I haven't thought too much about it, but maybe we can fix this if we `getAndSet` and invoke  `cb` in the `finally` block, before setting `done`? I'll wait to see what you come up with :)

djspiewak: I *believe* this case is actually handled! Note that the blocking operation is wrapped in a `try`/`finally` in which `done` is set to `true`. If the operation succeeds without being interrupted, then `cb` will be non-`null`, which will trip this conditional.

RaasAhsan: Here is the ordering that I'm concerned about:
1. At line 64, the target releases the `Semaphore` permit.
2. At line 65, the target runs the blocking operation.
3. Concurrently, the fiber is cancelled. `trigger` runs, which sets `cb`, acquires the `Semaphore` permit, interrupts the target once, and releases the permit. It breaks out of the loop.
4. The canceller runs `repeat`, and reaches line 136. At this point, `done` is set to `false` and `cb` is not null, so the conditional fails and the carrier thread for the canceller exits the `async`.
5. The target thread receives the interrupt signal during its blocking operation, but (this is the key part!) it resets the interrupt status, so the interrupt signal is effectively suppressed! `cur.thunk()` returns normally.
6. At line 69, the target thread acquires the `Semaphore` permit successfully because the interrupt status isn't set, and so `InterruptedException` isn't thrown.
7. At line 84, the target thread reaches the `finally` block and sets `done` to true, but at this point it's too late.
8. `cb` was never invoked, so the canceller is perpetually blocked.

Whenever I'm pointing out these bugs, I should probably describe the interleaving of effects that produces a race condition. It makes it much easier for both of us to follow but also easier to point out if I'm wrong.

djspiewak: > The canceller runs repeat, and reaches line 136. At this point, done is set to false and cb is not null, so the conditional fails and the carrier thread for the canceller exits the async.
>
> The target thread receives the interrupt signal during its blocking operation, but (this is the key part!) it resets the interrupt status, so the interrupt signal is effectively suppressed! cur.thunk() returns normally.

Ooooooooh this is a nice catch. I agree it's a problem. I wonder if we can solve it without introducing more state…

djspiewak: See latest push. I think I fixed it. I realized that we don't have to *only* check the callback on `InterruptedException`

RaasAhsan (approved):
Looks good to me! What an adventure that was

PR #992, opened 2020-07-24
Correct Parallel tests for Resource in the 2.x series

SystemFw:
This went undetected for some time


djspiewak:
LOL! I'm amazed this didn't get caught. Also I need to pull the RVM fix into series/2.x since now everyone is hitting it.

PR #996, opened 2020-07-24
Concurrent instances for monad transformers

TimWSpence:
Note that we have not defined instances for `StateT` or `ReaderWriterStateT` as they are fundamentally sequential and `Concurrent` is (oddly enough) fundamentally concurrent.

There's also no instance for `ContT` for the moment as there is currently no `MonadError` instance for it (undecided whether that is fundamental or not)

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/concurrent.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/concurrent.scala
@@ -243,7 +243,7 @@ object Concurrent {
 
     def forceR[A, B](fa: OptionT[F, A])(fb: OptionT[F, B]): OptionT[F, B] =
       OptionT(
-        fa.value.attempt >> fb.value
+        F.forceR(fa.value)(fb.value)
       )
 
     def pure[A](a: A): OptionT[F, A] = delegate.pure(a)
@@ -320,7 +320,7 @@ object Concurrent {
 
     def forceR[A, B](fa: EitherT[F, E0, A])(fb: EitherT[F, E0, B]): EitherT[F, E0, B] =
       EitherT(
-        fa.value.attempt >> fb.value
+        F.forceR(fa.value)(fb.value)
       )
 
     def pure[A](a: A): EitherT[F, E0, A] = delegate.pure(a)
@@ -397,7 +397,7 @@ object Concurrent {
 
     def forceR[A, B](fa: IorT[F, L, A])(fb: IorT[F, L, B]): IorT[F, L, B] =
       IorT(
-        fa.value.attempt >> fb.value
+        F.forceR(fa.value)(fb.value)
       )
 
     def pure[A](a: A): IorT[F, L, A] = delegate.pure(a)
@@ -474,7 +474,7 @@ object Concurrent {
     }
 
     def forceR[A, B](fa: Kleisli[F, R, A])(fb: Kleisli[F, R, B]): Kleisli[F, R, B] =
-      Kleisli(r => fa.run(r).attempt >> fb.run(r))
+      Kleisli(r => F.forceR(fa.run(r))(fb.run(r)))
 
     def pure[A](a: A): Kleisli[F, R, A] = delegate.pure(a)
 
@@ -553,7 +553,7 @@ object Concurrent {
 
     def forceR[A, B](fa: WriterT[F, L, A])(fb: WriterT[F, L, B]): WriterT[F, L, B] =
       WriterT(
-        fa.run.attempt >> fb.run
+        F.forceR(fa.run)(fb.run)
       )
 
     def pure[A](a: A): WriterT[F, L, A] = delegate.pure(a)
--- a/kernel/shared/src/main/scala/cats/effect/kernel/concurrent.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/concurrent.scala
@@ -158,10 +158,6 @@ object Concurrent {
 
   trait OptionTConcurrent[F[_], E] extends Concurrent[OptionT[F, *], E] {
 
-    val nat: F ~> OptionT[F, *] = new ~>[F, OptionT[F, *]] {
-      def apply[A](fa: F[A]): OptionT[F, A] = OptionT.liftF(fa)
-    }
-
     implicit protected def F: Concurrent[F, E]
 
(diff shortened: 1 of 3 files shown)


TimWSpence:
Currently blocked on reformulating
```
def canceledSequencesOnCancelInOrder(fin1: F[Unit], fin2: F[Unit]) =
    F.onCancel(F.onCancel(F.canceled, fin1), fin2) <-> (F.uncancelable(_ =>
      fin1.attempt >> fin2.attempt) >> F.canceled)
```
and
```
def uncancelableCanceledAssociatesRightOverFlatMap[A](a: A, f: A => F[Unit]) =
    F.uncancelable(_ => F.canceled.as(a).flatMap(f)) <-> (F.uncancelable(_ => f(a).attempt) >> F.canceled)
```
to support short-circuiting monad

kernel/shared/src/main/scala/cats/effect/kernel/concurrent.scala
```diff
@@ -157,4 +159,424 @@ trait Concurrent[F[_], E] extends MonadError[F, E] {
...
+      Kleisli { r =>
+        (F.racePair(fa.run(r), fb.run(r)).map {
+          case Left((oc, fib)) => Left((liftOutcome(oc), liftFiber(fib)))
+          case Right((fib, oc)) => Right((liftFiber(fib), liftOutcome(oc)))
+        })
+      }
+    }
+
+    def forceR[A, B](fa: Kleisli[F, R, A])(fb: Kleisli[F, R, B]): Kleisli[F, R, B] =
+      Kleisli(r => fa.run(r).attempt >> fb.run(r))
```
djspiewak: I think this needs to be `F.forceR(fa.run(r))(fb.run(r))`, otherwise `OptionT[OptionT[F, *], A]` will have the wrong semantics w.r.t. the inner effect.

kernel/shared/src/main/scala/cats/effect/kernel/concurrent.scala
```diff
@@ -157,4 +159,424 @@ trait Concurrent[F[_], E] extends MonadError[F, E] {
...
+        (Fiber[WriterT[F, L, *], E, A], Outcome[WriterT[F, L, *], E, B])]] = {
+      WriterT.liftF(F.racePair(fa.run, fb.run).map {
+        case Left((oc, fib)) => Left((liftOutcome(oc), liftFiber(fib)))
+        case Right((fib, oc)) => Right((liftFiber(fib), liftOutcome(oc)))
+      })
+    }
+
+    def forceR[A, B](fa: WriterT[F, L, A])(fb: WriterT[F, L, B]): WriterT[F, L, B] =
+      WriterT(
+        fa.run.attempt >> fb.run
```
djspiewak: Similar `forceR` issue here.

kernel/shared/src/main/scala/cats/effect/kernel/concurrent.scala
```diff
@@ -157,4 +159,424 @@ trait Concurrent[F[_], E] extends MonadError[F, E] {
...
+        (Fiber[EitherT[F, E0, *], E, A], Outcome[EitherT[F, E0, *], E, B])]] = {
+      EitherT.liftF(F.racePair(fa.value, fb.value).map {
+        case Left((oc, fib)) => Left((liftOutcome(oc), liftFiber(fib)))
+        case Right((fib, oc)) => Right((liftFiber(fib), liftOutcome(oc)))
+      })
+    }
+
+    def forceR[A, B](fa: EitherT[F, E0, A])(fb: EitherT[F, E0, B]): EitherT[F, E0, B] =
+      EitherT(
+        fa.value.attempt >> fb.value
```
djspiewak: …and here

kernel/shared/src/main/scala/cats/effect/kernel/concurrent.scala
```diff
@@ -157,4 +159,424 @@ trait Concurrent[F[_], E] extends MonadError[F, E] {
...
+
+    def liftFiber[A](fib: Fiber[F, E, Option[A]]): Fiber[OptionT[F, *], E, A] =
+      new Fiber[OptionT[F, *], E, A] {
+        def cancel: OptionT[F, Unit] = OptionT.liftF(fib.cancel)
+        def join: OptionT[F, Outcome[OptionT[F, *], E, A]] =
+          OptionT.liftF(fib.join.map(liftOutcome))
+      }
+  }
+
+  trait EitherTConcurrent[F[_], E0, E] extends Concurrent[EitherT[F, E0, *], E] {
```
djspiewak: Can we make all of these traits `private[kernel]` to avoid increasing the API footprint? Also it might be more reasonable to nest them under the appropriate object (`Concurrent`, in this case).

kernel/shared/src/main/scala/cats/effect/kernel/concurrent.scala
```diff
@@ -157,4 +159,424 @@ trait Concurrent[F[_], E] extends MonadError[F, E] {
...
+        (Fiber[OptionT[F, *], E, A], Outcome[OptionT[F, *], E, B])]] = {
+      OptionT.liftF(F.racePair(fa.value, fb.value).map {
+        case Left((oc, fib)) => Left((liftOutcome(oc), liftFiber(fib)))
+        case Right((fib, oc)) => Right((liftFiber(fib), liftOutcome(oc)))
+      })
+    }
+
+    def forceR[A, B](fa: OptionT[F, A])(fb: OptionT[F, B]): OptionT[F, B] =
+      OptionT(
+        fa.value.attempt >> fb.value
```
djspiewak: `forceR`

djspiewak (changes requested):

kernel/shared/src/main/scala/cats/effect/kernel/concurrent.scala
TimWSpence: Yes! Very good point!

kernel/shared/src/main/scala/cats/effect/kernel/concurrent.scala
TimWSpence: They are nested inside `Concurrent`? But yes, making them private is a very good idea, thanks!

PR #997, opened 2020-07-24
Transformer instances for temporal

TimWSpence:


TimWSpence:
Dependent on https://github.com/typelevel/cats-effect/pull/996 and on `TimeT` and hence the `Temporal` instance for `PureConc` being fixed

laws/shared/src/test/scala/cats/effect/PureConcSpec.scala
```diff
@@ -16,37 +16,212 @@
...
+    "Kleisli[PureConc]",
+    TemporalTests[Kleisli[TimeT[PureConc[Int, *], *], MiniInt, *], Int]
+      .temporal[Int, Int, Int](10.millis)
+  )
+
+  checkAll(
+    "WriterT[PureConc]",
+    TemporalTests[WriterT[TimeT[PureConc[Int, *], *], Int, *], Int]
+      .temporal[Int, Int, Int](10.millis)
+  )
```
djspiewak: Can we push all of these into separate files, similar to what I did for the other inductive `PureConc` specs? This might be a bit annoying given some of the implicit resolution hacks, which may have to be replicated. The reasoning behind getting all of these into their own files is that allows them to be run in parallel on ScalaJS, which drops the build times from somewhere around 30 minutes to somewhere closer to 8 minutes.

laws/shared/src/test/scala/cats/effect/PureConcSpec.scala
```diff
@@ -16,37 +16,212 @@
...
 
 //  checkAll(
 //    "ParallelF[PureConc]",
 //    AlignTests[ParallelF[PureConc[Int, *], *]].align[Int, Int, Int, Int])
 }
+
+//Push the priority of Kleisli instances down so we can explicitly summon more
+//specific instances to help 2.12 out - I think Kleisli[TimeT[PureConc[Int, *], *], MiniInt, *]
+//involves about 4 nested Kleisli's so the compiler just gives up
+trait LowPriorityInstances {
```
djspiewak: I really hate Scala 2.12… :-)

djspiewak (changes requested):
Minor (but unfortunately, annoying) request!

laws/shared/src/test/scala/cats/effect/PureConcSpec.scala
TimWSpence: 😭 But yes, that is definitely worth it!

laws/shared/src/test/scala/cats/effect/PureConcSpec.scala
TimWSpence: Interestingly this was no longer necessary when I split into different files. But yes, it sucks 😆

TimWSpence: Oh woops. I retract this. Just forgot to test locally with 2.12 😭

PR #998, opened 2020-07-25
Backpressured cancels

### LINKED ISSUE #987: Make subsequent cancels await the first one

RaasAhsan:
The main goal of this ticket was to fix #987 so that the first `cancel` on a fiber backpressures subsequent cancels. In order to do this, we had to fix a couple of other race conditions:
1. An `AsyncK` continuation would invoke one more iteration of the run-loop, even after `suspend`ing. This could lead to it invoking the finalizers concurrently to an external actor.
2. `AsyncK` checked for cancellation and then `suspend`ed in two different steps. This could lead to a race condition where if an external actor B attempts to cancel the fiber A *after* A reads the `canceled` flag, then it would be forced to `join` since A still owns the run-loop. A later suspends itself, but is cancelled, so everyone is blocked. The only way the finalizers would be run is through an `async` callback or if another external actor cancels B.
3. Finalizers can be async, and `cancel` doesn't backpressure on them, so it could return prematurely. Borrowed the fix from @djspiewak 's branch on #990 
4. Several other subtle ones I can't remember but would probably be worth documenting.

The change to backpressure subsequent cancellers is also rather subtle. `outcome` was previously used to determine whether a fiber had terminated or cancelled early, but that's not the case anymore. Instead we use `isCancelled` and `isUnmasked` to determine that.

I also tried to include some QOL changes and remove redundancies so that run-loop is easier to reason about. Some statements we can make about the run-loop now:
1. A fiber can terminate its own run-loop with a success, failure or cancellation.
2. A fiber can terminate another fiber's run-loop only with a cancellation outcome.
3. Only the fiber that owns a run-loop (not necessarily its own) can terminate that run-loop for as long as it holds ownership.
4. Once a run-loop is terminated, it can never be acquired again. (might have to confirm this one)
5. Once a fiber terminates with an outcome, that outcome cannot change.
6. If you can acquire a run-loop, it has not yet terminated. (contrapositive of 4)

We can probably make more statements like these. Should probably define some of this terminology better

Fixes #987

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -249,7 +249,7 @@ private[effect] final class IOFiber[A](
 
     val ec = currentCtx
 
-    if (!isCanceled() || !isUnmasked()) { // hard cancelation check, basically
+    if (!shouldFinalize()) {
       execute(ec) { () =>
         val next = e match {
           case Left(t) => failed(t, 0)
@@ -297,7 +297,7 @@ private[effect] final class IOFiber[A](
       iteration + 1
     }
 
-    if (isCanceled() && isUnmasked()) {
+    if (shouldFinalize()) {
       asyncCancel(null)
     } else {
       // println(s"<$name> looping on $cur0")
@@ -401,7 +401,7 @@ private[effect] final class IOFiber[A](
                     }
 
                     asyncContinue(state, e)
-                  } else if (!isCanceled() || !isUnmasked()) {
+                  } else if (!shouldFinalize()) {
                     loop()
                   }
 
@@ -591,6 +591,13 @@ private[effect] final class IOFiber[A](
   private def childMask: Int =
     initMask + 255
 
+  // We should attempt finalization if all of the following are true:
+  // 1) We own the runloop
+  // 2) We have been cancelled
+  // 3) We are unmasked
+  private def shouldFinalize(): Boolean =
+    isCanceled() && isUnmasked()
+
   // we use these forwarders because direct field access (private[this]) is faster
   private def isCanceled(): Boolean =
     canceled
@@ -624,11 +631,11 @@ private[effect] final class IOFiber[A](
   private def suspend(): Unit =
     suspended.set(true)
 
-  private def suspendAndCheckCancellation(): Unit = {
+  private def suspendWithFinalizationCheck(): Unit = {
     // full memory barrier
     suspended.compareAndSet(false, true)
     // race condition check: we may have been cancelled before we suspended
-    if (isCanceled() && isUnmasked()) {
+    if (shouldFinalize()) {
       // if we can acquire the run-loop, we can run the finalizers
       // otherwise somebody else picked it up and will run finalizers
       if (resume()) {
@@ -813,7 +820,7 @@ private object IOFiber {
                 popFinalizer()
                 asyncContinue(state, state.get().result)
               } else {
-                suspendAndCheckCancellation()
+                suspendWithFinalizationCheck()
               }
 
             case None =>
@@ -821,7 +828,7 @@ private object IOFiber {
                 // the callback was invoked before registration
                 asyncContinue(state, state.get().result)
               } else {
-                suspendAndCheckCancellation()
+                suspendWithFinalizationCheck()
               }
           }
         } else {
@@ -830,7 +837,7 @@ private object IOFiber {
             // the callback was invoked before registration
             asyncContinue(state, state.get().result)
           } else {
-            suspendAndCheckCancellation()
+            suspendWithFinalizationCheck()
           }
         }
 
@@ -846,7 +853,7 @@ private object IOFiber {
       val ec = popContext()
 
       // special cancelation check to ensure we don't unnecessarily fork the runloop here
-      if (!isCanceled() || !isUnmasked()) {
+      if (!shouldFinalize()) {
         execute(ec) { () =>
           if (success)
             runLoop(succeeded(result, 0), 0)
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -401,13 +401,8 @@ private[effect] final class IOFiber[A](
                     }
 
                     asyncContinue(state, e)
-                  } else {
-                    // race condition check: if we observe cancellation
-                    if (isCanceled() && isUnmasked()) {
-                      if (resume()) {}
-                    } else {
-                      loop()
-                    }
+                  } else if (!isCanceled() || !isUnmasked()) {
+                    loop()
                   }
 
                   // If we reach this point, it means that somebody else owns the run-loop
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -244,7 +244,9 @@ private[effect] final class IOFiber[A](
 
     val ec = currentCtx
 
-    if (!isCanceled()) { // hard cancelation check, basically
+    println(s"$name: Attempting to asyncContinue with ${isCanceled()}, ${masks} , ${initMask}")
+
+    if (!isCanceled() || !isUnmasked()) { // hard cancelation check, basically
       execute(ec) { () =>
         val next = e match {
           case Left(t) => failed(t, 0)
@@ -382,7 +384,7 @@ private[effect] final class IOFiber[A](
                 @tailrec
                 def loop(): Unit =
                   if (resume()) {
-                    if (!isCanceled()) { // double-check to see if we were canceled while suspended
+                    if (!isCanceled() || !isUnmasked()) { // double-check to see if we were canceled while suspended
                       if (old == AsyncStateRegisteredWithFinalizer) {
                         // we completed and were not canceled, so we pop the finalizer
                         // note that we're safe to do so since we own the runloop
@@ -391,7 +393,7 @@ private[effect] final class IOFiber[A](
 
                       asyncContinue(state, e)
                     }
-                  } else if (!isCanceled()) {
+                  } else if (!isCanceled() || !isUnmasked()) {
                     loop()
                   }
 
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -65,8 +65,6 @@ private[effect] final class IOFiber[A](
     extends FiberIO[A] {
   import IO._
 
-  val named = name
-
   // I would rather have these on the stack, but we can't because we sometimes need to relocate our runloop to another fiber
   private[this] var conts: ByteStack = _
 
@@ -142,12 +140,7 @@ private[effect] final class IOFiber[A](
           // if we have async finalizers, runLoop may return early
           IO.async_[Unit] { fin =>
 //            println(s"${name}: canceller started at ${Thread.currentThread().getName} + ${suspended.get()}")
-            if (hasFinalizers()) {
-              runFinalizers(fin)
-            } else {
-              done(_OutcomeCanceled.asInstanceOf[OutcomeIO[A]])
-              fin(Right(()))
-            }
+            asyncCancel(fin)
           }
         } else {
           // it was masked, so we need to wait for it to finish whatever it was doing and cancel itself
@@ -225,10 +218,12 @@ private[effect] final class IOFiber[A](
     // need to reset masks to 0 to terminate async callbacks
     // busy spinning in `loop`.
     masks = initMask
+    // full memory barrier to publish masks
+    suspended.set(false)
 
     // clear out literally everything to avoid any possible memory leaks
 
-    // conts may be null if the fiber was cancelled before it was cancelled
+    // conts may be null if the fiber was cancelled before it was started
     if (conts != null)
       conts.invalidate()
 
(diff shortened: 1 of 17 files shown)


core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -65,6 +65,8 @@ private[effect] final class IOFiber[A](
     extends FiberIO[A] {
   import IO._
 
+  val named = name
```
djspiewak: Stray debug logic?

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -130,33 +130,32 @@ private[effect] final class IOFiber[A](
...
-              masks += 1
-              runLoop(finalizers.pop(), 0)
+          // if we have async finalizers, runLoop may return early
+          IO.async_[Unit] { fin =>
+//            println(s"${name}: canceller started at ${Thread.currentThread().getName} + ${suspended.get()}")
+            if (hasFinalizers()) {
+              runFinalizers(fin)
+            } else {
+              done(_OutcomeCanceled.asInstanceOf[OutcomeIO[A]])
+              fin(Right(()))
```
djspiewak: This has a mild effect on fairness: the *first* caller of `cancel` will be continued *last*. I'm tempted to say we should invert the order of these calls.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -204,29 +203,58 @@ private[effect] final class IOFiber[A](
...
       callbacks.lazySet(null) // avoid leaks
     }
+
+    // need to reset masks to 0 to terminate async callbacks
+    // busy spinning in `loop`.
+    masks = initMask
+
+    // clear out literally everything to avoid any possible memory leaks
+
+    // conts may be null if the fiber was cancelled before it was cancelled
```
djspiewak: Actually `conts` will be `null` if the fiber was canceled before it was *started*. :-)

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -361,19 +402,23 @@ private[effect] final class IOFiber[A](
...
                   if (resume()) {
-                    if (outcome.get() == null) { // double-check to see if we were canceled while suspended
-                      if (old == AsyncStateRegisteredWithFinalizer) {
-                        // we completed and were not canceled, so we pop the finalizer
-                        // note that we're safe to do so since we own the runloop
-                        finalizers.pop()
-                      }
-
-                      asyncContinue(state, e)
+//                    if (!isCanceled() || !isUnmasked()) { // double-check to see if we were canceled while suspended
```
djspiewak: Should probably clean this up.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -361,19 +402,23 @@ private[effect] final class IOFiber[A](
...
+                      // we completed and were not canceled, so we pop the finalizer
+                      // note that we're safe to do so since we own the runloop
+                      finalizers.pop()
                     }
-                  } else if (outcome.get() == null) {
+
+                    asyncContinue(state, e)
+//                    } else {
+//                      runFinalizers(null)
+//                    }
```
djspiewak: Ditto here

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -361,19 +402,23 @@ private[effect] final class IOFiber[A](
...
+                  } else if (!isCanceled() || !isUnmasked()) {
                     loop()
                   }
 
+                // TODO: ELSE?
```
djspiewak: I guess this isn't needed? Is it worth writing a quick justification for each of our missing `else` cases?

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -450,13 +495,19 @@ private[effect] final class IOFiber[A](
...
+//              runLoop(
+//                null,
+//                nextIteration
+//              )
+            // we can't trust the cancelation check at the start of the loop
+            if (hasFinalizers()) {
+              runFinalizers(null)
+            } else {
+              done(_OutcomeCanceled.asInstanceOf[OutcomeIO[A]])
+            }
```
djspiewak: Note that this is parsing as an `else if`. Can we stick some curly braces around this (or remove the commented code) to make it a bit more clear?

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -772,27 +829,21 @@ private object IOFiber {
...
-                if (!state.compareAndSet(
-                    AsyncStateInitial,
-                    AsyncStateRegisteredWithFinalizer)) {
-                  // the callback was invoked before registration
-                  popFinalizer()
-                  asyncContinue(state, state.get().result)
-                } else {
-                  suspend()
-                }
+              // indicate the presence of the cancel token by pushing Right instead of Left
```
djspiewak: BTW this "right instead of left" thing is stale and should be removed. I forgot to delete it when I added `AsyncState`.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -817,14 +868,16 @@ private object IOFiber {
...
-            runLoop(succeeded(result, 0), 0)
-          else
-            runLoop(failed(result, 0), 0)
-        }
+//      if (!isCanceled() || !isUnmasked()) {
+      execute(ec) { () =>
+        if (success)
+          runLoop(succeeded(result, 0), 0)
+        else
+          runLoop(failed(result, 0), 0)
```
djspiewak: BTW I realized last night that the read barrier here is on the wrong thread: we put the barrier on the thread which *schedules* the callback with the executor, instead of putting it on the thread which owns the callback when run. We can just fix this in `execute`, but at the cost of an allocation.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -817,14 +868,16 @@ private object IOFiber {
...
 
       // special cancelation check to ensure we don't accidentally fork the runloop here
-      if (!isCanceled() || !isUnmasked()) {
-        execute(ec) { () =>
-          if (success)
-            runLoop(succeeded(result, 0), 0)
-          else
-            runLoop(failed(result, 0), 0)
-        }
+//      if (!isCanceled() || !isUnmasked()) {
```
djspiewak: More stale code

core/shared/src/test/scala/cats/effect/IOSpec.scala
```diff
@@ -560,6 +560,45 @@ class IOSpec extends IOPlatformSpecification with Discipline with ScalaCheck wit
...
+        }
+
+        IO.pure(Some(fin))
+      }
+
+      val test = target.start flatMap { f => IO(ticker.ctx.tickAll()) *> f.cancel }
+
+      test must completeAs(())
+      success must beTrue
+    }
```
djspiewak, line 600: I think it's worth also adding the following test case:

```scala
"never terminate when racing infinite cancels" in ticked { implicit ticker =>
  val test = IO.never.onCancel(IO.never).start flatMap { f => 
    IO.race(f.cancel, f.cancel)
  }

  test.void should nonTerminate
}
```

This gets a bit at part of what was changed, and also codifies in test-form one of the weirder consequences.

djspiewak (changes requested):

core/shared/src/main/scala/cats/effect/IOFiber.scala
RaasAhsan: yeah, couldn't access `name` before from the continuations so i used one of these

RaasAhsan: i probably just needed to declare them `val` in the constructor

core/shared/src/main/scala/cats/effect/IOFiber.scala
RaasAhsan: mistype!

core/shared/src/main/scala/cats/effect/IOFiber.scala
RaasAhsan: Going to delete this because the same check is performed in `asyncContinue`

core/shared/src/main/scala/cats/effect/IOFiber.scala
RaasAhsan: Bringing this one back

core/shared/src/main/scala/cats/effect/IOFiber.scala
RaasAhsan: I *think* it works because the lambda closes over a reference to `this` with respect to `IOFiber`. So if the thread running `execute` observes any changes through the memory barrier, then it those changes should also be reflected in the reference in the lambda.

core/shared/src/main/scala/cats/effect/IOFiber.scala
RaasAhsan: makes sense to bring this one back

core/shared/src/test/scala/cats/effect/IOSpec.scala
RaasAhsan: Needed to modify the test a little bit to ensure that the finalizer is registered before attempting cancellation.
```scala
 "never terminate when racing infinite cancels" in ticked { implicit ticker =>
      var started = false

      val markStarted = IO { started = true }
      lazy val cedeUntilStarted: IO[Unit] =
        IO(started).ifM(IO.unit, IO.cede >> cedeUntilStarted)

      val test = for {
        f <- (markStarted *> IO.never).onCancel(IO.never).start
        _ <- cedeUntilStarted
        _ <- IO.race(f.cancel, f.cancel)
      } yield ()

      test should nonTerminate
    }
```
I wonder if we should just use `Deferred` for asynchronous blocking

core/shared/src/main/scala/cats/effect/IOFiber.scala
RaasAhsan: There may a race condition here similar to what we had in `AsyncK`. In particular, if between the time we call `resume` and check `isCancelled`, the runloop is suspended while cancelled is set to true then we would be blocked.

I think it might not be an issue though, but here's my line of thinking. If `loop` can't `resume` the runloop, that means someone else owns the runloop. If the fiber gets cancelled between `resume()` and `isCancelled()`, then there are two cases:
1. Someone that isn't the canceller owns the runloop. If they happen to suspend, they will also check for cancellation via `suspendAndCheckCancellation`
2. Nobody owns the runloop, so the canceller acquires it, so they can run finalizers.

RaasAhsan: Yeah, I'm pretty convinced that this is a non-issue. If `loop` doesn't have the runloop to begin with and it sees cancellation, it doesn't have to do anything because somebody else will probably

core/shared/src/main/scala/cats/effect/IOFiber.scala
djspiewak: I'm not sure I follow that logic… The reference to `this` isn't volatile and is lifted into the lambda without barriers, so the thread which resumes the continuation will get access to `this.canceled` potentially without the value ever having been published to it, no?

core/shared/src/test/scala/cats/effect/IOSpec.scala
djspiewak: > I wonder if we should just use Deferred for asynchronous blocking

I'm okay with that. :-) I had been using the busy-wait mostly because it reduces the surface area of problems when trying to minimize stuff, but `Deferred` is far simpler to work with in tests.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -204,29 +196,60 @@ private[effect] final class IOFiber[A](
...
       callbacks(oc)
     } finally {
       callbacks.lazySet(null) // avoid leaks
     }
+
+    // need to reset masks to 0 to terminate async callbacks
+    // busy spinning in `loop`.
+    masks = initMask
+    // full memory barrier to publish masks
+    suspended.set(false)
```
djspiewak, line 222: `false` or `true`? Fibers start out as `suspended = true`, and we're pretty consistent about "`suspended = true` implies someone has `runLoop` somewhere".

djspiewak (changes requested):

core/shared/src/main/scala/cats/effect/IOFiber.scala
RaasAhsan: Can't we rely on the memory barriers inside the `ExecutionContext`? All the state that is visible to the thread running `ec.execute` should be visible to the thread that runs the submitted task.

It also makes sense to do the read barrier before the finalization check.

core/shared/src/main/scala/cats/effect/IOFiber.scala
RaasAhsan: This is merely a write memory barrier (the comment above it is wrong) to publish the write to `mask`, the value of `suspended` doesn't change at all. I want to avoid setting it back to `true` so that no one can acquire the runloop again (see point 4 in the PR description).

djspiewak: Ah I see! Rather than setting `suspended` to `true` when you're done with finalization, you just let it stay `false` since `outcome` is already set, so it shouldn't, ultimately, matter.

core/shared/src/main/scala/cats/effect/IOFiber.scala
djspiewak: Oh! Right. Transitivity. Okay yeah I think this works then. :-)

djspiewak (approved):
I really like the list of invariants in the OP. I think that we should move that into `IOFiber`'s scaladoc in a follow-up PR. It provides a more structured way of thinking about these cases.

Issue #995, opened 2020-07-24
Preallocate the lambda in Cede


djspiewak:
```scala
          // Cede
          case 18 =>
            currentCtx execute { () =>
              // println("continuing from cede ")

              runLoop(succeeded((), 0), nextIteration)
            }
```

We can actually preallocate this lambda if we ignore `nextIteration`, which we can do if we use `execute(...)` to insert a `readBarrier()`, which should make `Cede` noticeably quicker. There may be a few other places where we can do this. Any lambda which doesn't close over anything other than `this` can be preallocated. Additionally, anything which *does* close over state other than `this` but which is evaluated in a deterministic timeslice can be preallocated if we leverage the `objectState` trick. This is probably part of what we need to do to optimize `start` and `racePair`.

vasilmkd:
Issue addressed with #1004.

PR #999, opened 2020-07-25
Tailrec runLoop

vasilmkd:
Inspired by #995.

Not exactly preallocation but I imagine a `@tailrec` `runLoop` should amount to something.

The code keeps the previous semantics, it doesn't try to do anything clever.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -238,7 +238,33 @@ private[effect] final class IOFiber[A](
     }
   }
 
+  private[this] def blockingK(success: Boolean, result: Any, nextIteration: Int) =
+    new Runnable {
+      def run(): Unit = {
+        val next =
+          if (success)
+            succeeded(result, 0)
+          else
+            failed(result, 0)
+
+        runLoop(next, nextIteration)
+      }
+    }
+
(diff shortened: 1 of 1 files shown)


vasilmkd:
Upon examination of the bytecode, it seems that scalac was smart enough to compile `runLoop` as tail recursive even without the annotation all along (and in spite of simply adding the annotation, which complains about the closures that invoke `runLoop` again). I don't think this PR is worth it and it should be closed.

djspiewak:
> Upon examination of the bytecode, it seems that scalac was smart enough to compile runLoop as tail recursive even without the annotation all along (and in spite of simply adding the annotation, which complains about the closures that invoke runLoop again). I don't think this PR is worth it and it should be closed.

:-) The thought is appreciated though! It's definitely *very* important that `runLoop` is tail recursive – `IO` wouldn't be stack-safe if it wasn't! I do wish there were a way for us to say "this function needs to be tail recursive, except at one or two call-sites which we'll annotate". That would make it possible to make this kind of thing more explicit. As it stands though… scalac does identify tail recursion, with or without the annotation, and compiles it to a `goto`. The only thing the annotation does is add static checking to ensure that *all* call sites are tail recursive, which of course isn't the case here (though the ones which aren't in tail recursive position are trampolined via the executor, and so still stack-safe).

Issue #921, opened 2020-07-12
Improve performance of IO async, start, and race


djspiewak:
The concurrency operations on `IOFiber` are relatively slow, particularly compared to master. *Some* of that is a consequence of the fact that `async` in CE3 provides considerably stronger sequencing guarantees than `async` on master, but I still think we can be quite a bit faster. The performance difference is an order of magnitude under some circumstances, which is quite meaningful.

RaasAhsan:
I think a considerable portion of the slowdown comes from the internal shifting behavior of CE3 IO. Some ideas here that would certainly speed up `start` and `race`:
1. Add a shifting flag that is opt-out option. `start` and `race` would opt-out, since we're guaranteed to be on the default `ExecutonContext` as long as we don't expose this flag in the public API
2. Stack safety is now a concern. We can port the trampolining `ExecutionContext` implementation from Cats Effect 2, in conjunction with the opt-out flag.
3. Another idea to address stack safety is to keep track of a async depth counter. If we re-enter the run-loop from an async boundary and the depth counter exceeds some threshold, then do a shift regardless of what the shift flag is.

We can probably use this for `join` and `cancel` as well. Thoughts?

RaasAhsan:
I keep on mixing up Cats Effect 2 and 3 semantics. Cats Effect 2 implemented both `start` and `racePair` as `Async`, but CE3 only implements the latter as such. Plus, for `join` and `cancel`, we need to explicitly shift threads anyway since we might have several.

A ton of combinators were derived from `Async` in CE2, but I don't think that's the case any longer for CE3. The only compelling case here is `racePair`, but it may not be worth all the extra trampolining machinery to make it fast.

PR #1004, opened 2020-07-27
Preallocate closures in IO run loop

vasilmkd:
Another attempt at #995. Also related to #921.

All of the IO tests are passing with no issues. Opening as a draft because I'm just not sure if this is a good approach.

Thanks for taking a look.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -234,14 +234,14 @@ private[effect] final class IOFiber[A](
 
     finalizers.invalidate()
 
-    asyncContinueClosure.prepare(null)
-    blockingClosure.prepare(null, 0)
-    afterBlockingClosure.prepare(null, false, 0)
-    evalOnClosure.prepare(null, 0)
-    cedeClosure.prepare(0)
+    asyncContinueClosure = null
+    blockingClosure = null
+    afterBlockingClosure = null
+    evalOnClosure = null
+    cedeClosure = null
   }
 
-  private[this] object asyncContinueClosure extends Runnable {
+  private[this] final class AsyncContinueClosure extends Runnable {
     private[this] var either: Either[Throwable, Any] = _
 
     def prepare(e: Either[Throwable, Any]): Unit =
@@ -256,6 +256,8 @@ private[effect] final class IOFiber[A](
       runLoop(next, 0)
     }
   }
+
+  private[this] var asyncContinueClosure = new AsyncContinueClosure()
 
   /*
   4 possible cases for callback and cancellation:
@@ -299,7 +301,7 @@ private[effect] final class IOFiber[A](
     }
   }
 
-  private[this] object blockingClosure extends Runnable {
+  private[this] final class BlockingClosure extends Runnable {
     private[this] var cur: Blocking[Any] = _
     private[this] var nextIteration: Int = 0
 
@@ -324,7 +326,9 @@ private[effect] final class IOFiber[A](
     }
   }
 
-  private[this] object afterBlockingClosure extends Runnable {
+  private[this] var blockingClosure = new BlockingClosure()
+
+  private[this] final class AfterBlockingClosure extends Runnable {
     private[this] var result: Any = _
     private[this] var success: Boolean = false
     private[this] var nextIteration: Int = 0
@@ -341,7 +345,9 @@ private[effect] final class IOFiber[A](
     }
   }
 
-  private[this] object evalOnClosure extends Runnable {
+  private[this] var afterBlockingClosure = new AfterBlockingClosure()
+
+  private[this] final class EvalOnClosure extends Runnable {
     private[this] var ioa: IO[Any] = _
     private[this] var nextIteration: Int = 0
 
@@ -354,7 +360,9 @@ private[effect] final class IOFiber[A](
       runLoop(ioa, nextIteration)
   }
 
-  private[this] object cedeClosure extends Runnable {
+  private[this] var evalOnClosure = new EvalOnClosure()
+
+  private[this] final class CedeClosure extends Runnable {
     private[this] var nextIteration: Int = 0
 
     def prepare(ni: Int): Unit =
@@ -363,6 +371,8 @@ private[effect] final class IOFiber[A](
     def run(): Unit =
       runLoop(succeeded((), 0), nextIteration)
   }
+
+  private[this] var cedeClosure = new CedeClosure()
 
   // masks encoding: initMask => no masks, ++ => push, -- => pop
   private def runLoop(cur0: IO[Any], iteration: Int): Unit = {
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -233,6 +233,12 @@ private[effect] final class IOFiber[A](
     objectState.invalidate()
 
     finalizers.invalidate()
+
+    asyncContinueClosure.prepare(null)
+    blockingClosure.prepare(null, 0)
+    afterBlockingClosure.prepare(null, false, 0)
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -235,6 +235,19 @@ private[effect] final class IOFiber[A](
     finalizers.invalidate()
   }
 
+  private[this] object asyncContinueClosure extends Runnable {
+    var either: Either[Throwable, Any] = _
```
RaasAhsan: We should declare instance variables `private[this]`, otherwise we'll encounter memory barriers on reads and writes. So we'll have to define another method like:
```scala
def apply(ec: ExecutionContext, result: Either[Throwable, Any]): Unit
```

Same deal for all the `var`s in the other closures

vasilmkd: @RaasAhsan Do you mind taking another look? Thanks.

vasilmkd:
Should I clean up all closure instance object references in `done`?

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -233,6 +233,28 @@ private[effect] final class IOFiber[A](
...
     finalizers.invalidate()
+
+    asyncContinueClosure.prepare(null)
+    blockingClosure.prepare(null, 0)
+    afterBlockingClosure.prepare(null, false, 0)
+    evalOnClosure.prepare(null, 0)
+    cedeClosure.prepare(0)
+  }
+
+  private[this] object asyncContinueClosure extends Runnable {
```
djspiewak: I think it's worth using `val` rather than `object` so we avoid the memory barrier

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -233,6 +233,28 @@ private[effect] final class IOFiber[A](
     objectState.invalidate()
 
     finalizers.invalidate()
+
+    asyncContinueClosure.prepare(null)
+    blockingClosure.prepare(null, 0)
+    afterBlockingClosure.prepare(null, false, 0)
+    evalOnClosure.prepare(null, 0)
+    cedeClosure.prepare(0)
```
djspiewak: Do we actually need to pre-prepare them like this? Why not just let them all be uninitialized until used?

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -250,14 +272,8 @@ private[effect] final class IOFiber[A](
...
-      execute(ec) { () =>
-        val next = e match {
-          case Left(t) => failed(t, 0)
-          case Right(a) => succeeded(a, 0)
-        }
-
-        runLoop(next, 0) // we've definitely hit suspended as part of evaluating async
-      }
+      asyncContinueClosure.prepare(e)
+      execute(ec)(asyncContinueClosure)
```
djspiewak, line 269: I'm amazed this actually works, but it truly does. The logic here: for a given `IOFiber` instance, we're not going to be using more than one of these at a time, because the runloop *will be suspended*. So we really can just aggressively preallocate it.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -283,6 +299,71 @@ private[effect] final class IOFiber[A](
     }
   }
 
+  private[this] object blockingClosure extends Runnable {
```
djspiewak: Same concern about `object`

djspiewak (changes requested):
I honestly didn't think something like this would work, but seeing it actually written out, it seems very clever and functional! I do think we shouldn't be using `object`, but aside from that, this is wonderful.

core/shared/src/main/scala/cats/effect/IOFiber.scala
vasilmkd: It was `val`, but Dotty complained that the field is not part of the interface of `Runnable`. Should I create special classes?

core/shared/src/main/scala/cats/effect/IOFiber.scala
vasilmkd: I believe I put this in `done`, so this is not pre-preparing, it's actually cleanup.

core/shared/src/main/scala/cats/effect/IOFiber.scala
vasilmkd: That was my intuition as well.

core/shared/src/main/scala/cats/effect/IOFiber.scala
djspiewak: Oh, I see. Okay cool.

core/shared/src/main/scala/cats/effect/IOFiber.scala
djspiewak: I think it's worth creating special (final) inner classes.

core/shared/src/main/scala/cats/effect/IOFiber.scala
RaasAhsan: Alternatively, if we define these closures as `var`s, we can just set them to `null` to clean up all the internal state as well as the enclosing objects

core/shared/src/main/scala/cats/effect/IOFiber.scala
RaasAhsan: 👍  The CE2 run-loop exploited the same idea with `RestartCallback`

core/shared/src/main/scala/cats/effect/IOFiber.scala
vasilmkd: `private[this] var`?

RaasAhsan: Yup, the `private[this]` is crucial or else we encounter memory barriers

core/shared/src/main/scala/cats/effect/IOFiber.scala
vasilmkd: Should these be on the companion object or in the class itself?

RaasAhsan: Either works but the inner class is convenient because it closes over `this` to `IOFiber` and has access to private methods, otherwise you need to pass it in explicitly

vasilmkd: I figured this was the case. Implementing now.

core/shared/src/main/scala/cats/effect/IOFiber.scala
vasilmkd: Should I rename `prepare` to something else to make it more clear? `set`?

djspiewak: `prepare` seems relatively clear to me.

vasilmkd: It's even more clear now that I've made the closures to be variables themselves. The cleanup is now just nulling them out. No need to call prepare with nulls.

core/shared/src/main/scala/cats/effect/IOFiber.scala

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -257,6 +257,8 @@ private[effect] final class IOFiber[A](
     }
   }
 
+  private[this] var asyncContinueClosure = new AsyncContinueClosure()
```
djspiewak: Can you put all of these up towards the top of the file (ideally all together with a comment explaining their existence)? It just makes it a bit easier to find. Similarly, I'd prefer if the classes are down at the bottom of the file.

vasilmkd: Sounds good.

PR #1005, opened 2020-07-27
A fiber can be a runnable

vasilmkd:
Related to #921.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -196,7 +196,7 @@ private[effect] final class IOFiber[A](
     }
   }
 
-  private[effect] def run(cur: IO[Any], ec: ExecutionContext): Unit = {
+  private[effect] def exec(cur: IO[Any], ec: ExecutionContext): Unit = {
     conts = new ByteStack(16)
     pushCont(RunTerminusK)
 
@@ -493,7 +493,7 @@ private[effect] final class IOFiber[A](
             // println(s"<$name> spawning <$childName>")
 
             val ec = currentCtx
-            execute(ec)(() => fiber.run(cur.ioa, ec))
+            execute(ec)(() => fiber.exec(cur.ioa, ec))
 
             runLoop(succeeded(fiber, 0), nextIteration)
 
@@ -524,8 +524,8 @@ private[effect] final class IOFiber[A](
 
                     val ec = currentCtx
 
-                    execute(ec)(() => fiberA.run(cur.ioa, ec))
-                    execute(ec)(() => fiberB.run(cur.iob, ec))
+                    execute(ec)(() => fiberA.exec(cur.ioa, ec))
+                    execute(ec)(() => fiberB.exec(cur.iob, ec))
 
                     Some(fiberA.cancel.both(fiberB.cancel).void)
                   }
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -62,7 +62,8 @@ private[effect] final class IOFiber[A](
     scheduler: unsafe.Scheduler,
     blockingEc: ExecutionContext,
     initMask: Int)
-    extends FiberIO[A] {
+    extends FiberIO[A]
+    with Runnable {
   import IO._
 
   // I would rather have these on the stack, but we can't because we sometimes need to relocate our runloop to another fiber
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -534,9 +556,10 @@ private[effect] final class IOFiber[A](
                     fiberB.registerListener(oc => cb(Right(Right((fiberA, oc)))))
 
                     val ec = currentCtx
-
-                    execute(ec)(() => fiberA.run(cur.ioa, ec))
-                    execute(ec)(() => fiberB.run(cur.iob, ec))
+                    fiberA.prepare(cur.ioa, ec)
+                    fiberB.prepare(cur.iob, ec)
+                    execute(ec)(fiberA)
+                    execute(ec)(fiberB)
```
djspiewak, line 552: LOL! I'm laughing so hard right now…

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -499,12 +520,13 @@ private[effect] final class IOFiber[A](
             val childName = s"start-${childCount.getAndIncrement()}"
             val initMask2 = childMask
 
-            val fiber = new IOFiber(childName, scheduler, blockingEc, initMask2)
+            val fiber = new IOFiber[Any](childName, scheduler, blockingEc, initMask2)
```
djspiewak, line 513: Unnecessary change?

djspiewak (approved):
This is legit very clever. I'd like to hear if @RaasAhsan has any objections that I didn't notice, but it seems like a very smart way to cut down on the allocations in `start` and `race`

core/shared/src/main/scala/cats/effect/IOFiber.scala
vasilmkd: I meant to ask this. There are two `IOFiber[Any]` instances created in `racePair`. This one here is actually an `IOFiber[Nothing]`. Does it need changing? Otherwise, I'll drop it.

djspiewak: Oh, I didn't realize it was `IOFiber[Nothing]`. I'm cool with fixing that. The compiler sometimes freaks out around `Nothing`, so we may as well be explicit.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -187,6 +188,23 @@ private[effect] final class IOFiber[A](
...
+  private[this] var startIO: IO[Any] = _
+  private[this] var startEC: ExecutionContext = _
+
+  private def prepare(io: IO[Any], ec: ExecutionContext): Unit = {
+    startIO = io
+    startEC = ec
+  }
+
+  /**
+   * @note This method should not be used outside of the IO run loop under any circumstance.
```
RaasAhsan: I think it's fair to just declare the method straight up `private` then

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -187,6 +188,23 @@ private[effect] final class IOFiber[A](
...
+   */
+  def run(): Unit = {
+    run(startIO, startEC)
```
RaasAhsan: maybe call this `start` now to avoid ambiguity

RaasAhsan (approved):
Very nice catch!

core/shared/src/main/scala/cats/effect/IOFiber.scala
vasilmkd: That might not be possible due to the method coming from `Runnable`, but I'll try.

core/shared/src/main/scala/cats/effect/IOFiber.scala
vasilmkd: Unfortunately it must be called `run`.  Should I put `override` to make it more clear?

RaasAhsan: ah `run(startIO, startEC)` this one needs to be called `run` as well?

djspiewak: Let's rename that one. I'm not quite sure what to call it though.

vasilmkd: `exec`?

djspiewak: I'm cool with `exec`

PR #1010, opened 2020-07-28
update to Scala.js 1.1.1, remove CompileTime hack

larsrh:
Supersedes #908 

See also scala-js/scala-js#4122


build.sbt
```diff
@@ -259,13 +255,13 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
     name := "cats-effect",
     libraryDependencies ++= Seq(
       "org.typelevel" %%% "cats-core" % CatsVersion,
-      "org.typelevel" %%% "simulacrum" % SimulacrumVersion % CompileTime,
+      "org.typelevel" %%% "simulacrum" % SimulacrumVersion % "provided",
```
djspiewak, line 258: Doesn't this mean it gets added to the pom? Shouldn't we filter it out?

larsrh: It gets added to the POM, but it doesn't matter, because downstream dependencies will just ignore it.

(Alternatively, consider it to be an incentive to remove the simulacrum dependency altogether.)

djspiewak: > (Alternatively, consider it to be an incentive to remove the simulacrum dependency altogether.)

Technically we need to do this anyway.

larsrh: Then smash the approve button ^.^

PR #1011, opened 2020-07-28
It...works?

djspiewak:
This reenables the `TimeT` law tests, which apparently just straight up run. I didn't touch the timeout stuff yet; just wanted to get this in. I don't understand why this is working, given that I didn't really fix `racePair`.


djspiewak:
The build here is failing quite consistently on #1015, for reasons I really don't understand (since I didn't touch that codepath at all).

PR #1013, opened 2020-07-29
Remove IOCont and split continuation functions into success and failure

vasilmkd:
This might look like a big PR but it mostly shuffles code around.

How to approach this as a reviewer:
1. `IOCont` is deleted and the continuation functions have been moved to methods on `IOFiber`, as in #1001.
2. The continuation functions have been split up into `Success` and `Failure`, called from `succeeded` and `failed` respectively. The continuation functions are shorter, require fewer arguments to call and they contain no branching on `success`. The `Failure` continuation functions now take `Throwable` as the error and require no casting.
3. `evalOnSuccessK` and `evalOnFailureK` use a preallocated closure instead of allocating a new closure each time.
4. `AfterBlockingClosure` has been split into a `Success` and `Failure` case, which are reused in `Blocking[Any]` and `evalOnK`.
5. `private[this]` all the things.
6. Explicit `@tailrec` on `runLoop`.
7. Forwarder methods have been deleted.
8. The logic in `Delay[Any]` case and `Blocking[Any]` case has been simplified.

Thanks to @djspiewak for the discussion and for the original PR. I have straight up lifted some comments and the overall organization of the code from #1001. I haven't attempted any of the other optimizations from that PR.


core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -596,56 +597,17 @@ private[effect] final class IOFiber[A](
...
-  // allow for 255 masks before conflicting; 255 chosen because it is a familiar bound, and because it's evenly divides UnsignedInt.MaxValue
-  // this scheme gives us 16,843,009 (~2^24) potential derived fibers before masks can conflict
-  private def childMask: Int =
-    initMask + 255
-
-  // We should attempt finalization if all of the following are true:
-  // 1) We own the runloop
-  // 2) We have been cancelled
-  // 3) We are unmasked
-  private def shouldFinalize(): Boolean =
```
RaasAhsan, line 609: I'm highly in favor of keeping most or all of these, especially `shouldFinalize`. They help convey intent much better than their definitions.

vasilmkd: I restored `shouldFinalize` and `isUnmasked`. I think the other methods are clear enough.

djspiewak: I think it's worth rerunning the experiment that I did a while back, where I benchmarked `private[this]` vs the inlined variants. There's definitely an appreciable performance difference between `private` and `private[this]`, but it may be that `private[this]` inlines well enough that we don't need to worry about it. The functions *are* much more semantic.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -697,20 +674,20 @@ private[effect] final class IOFiber[A](
...
+      case 0 => cancelationLoopK()
+      case 1 => runTerminusFailureK(error)
+      case 2 => asyncFailureK(error, depth)
+      case 3 => evalOnFailureK(error)
+      // (case 4) will never continue to mapK
+      // (case 5) will never continue to flatMapK
+      case 6 => handleErrorWithK(error, depth)
+      case 7 => onCancelFailureK(error, depth)
+      case 8 => uncancelableFailureK(error, depth)
+      case 9 => unmaskFailureK(error, depth)
```
djspiewak: Ah! This is very tricky, but surprisingly, this will result in a significant loss of performance. The problem here is that HotSpot fast-paths `tableswitch`es with contiguous ranges, and you've removed `4` and `5`. An easy fix is to just restore the cases even though they'll never be hit, but a better fix is to move `MapK` and `FlatMapK` to either the start of the range (e.g. `0` and `1`) or the end of the range (e.g. `8` and `9`) and keep them out of this `match`, ensuring the range is still contiguous.

core/shared/src/main/scala/cats/effect/IOFiber.scala
vasilmkd: Should I restore all of them? How long does the benchmark take?

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -822,250 +805,213 @@ private[effect] final class IOFiber[A](
...
   }
 }
+
+private object IOFiber {
+  private val childCount = new AtomicInteger(0)
+
+  // prefetch
+  final private val OutcomeCanceled = Outcome.Canceled()
+  final private val OutcomeErrored = Outcome.Errored
+  final private val OutcomeCompleted = Outcome.Completed
```
djspiewak: I'm actually thinking that these particular prefetches aren't worth it, since we're just going to end up calling `apply`, which will eventually be inlined to the constructor. We can shortcut to the end of that whole optimization by *not* prefetching the companion values, since we don't need them, and calling `apply` directly on a case class constructor is compiled by scalac into just calling `new`.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -597,56 +598,27 @@ private[effect] final class IOFiber[A](
...
+  private[this] def suspend(): Unit =
     suspended.set(true)
 
-  private def suspendWithFinalizationCheck(): Unit = {
+  private[this] def suspendWithFinalizationCheck(): Unit = {
     // full memory barrier
     suspended.compareAndSet(false, true)
     // race condition check: we may have been cancelled before we suspended
-    if (shouldFinalize()) {
+    if (canceled && masks == initMask) {
```
djspiewak: Should go back to `shouldFinalize()` since that helper has been restored.

djspiewak (changes requested):

core/shared/src/main/scala/cats/effect/IOFiber.scala
vasilmkd: Ah, good catch. I missed this.

core/shared/src/main/scala/cats/effect/IOFiber.scala
djspiewak: The benchmarks are pretty quick. Like an hour with default configuration. I like to run the `Shallow` and `Deep` binds, and usually also `Attempt` and `HandleErrorWith`.

djspiewak: Oh, regarding which ones to restore… I think you already restored the most important one. As for the rest, I defer to @RaasAhsan.

core/shared/src/main/scala/cats/effect/IOFiber.scala

core/shared/src/main/scala/cats/effect/IOFiber.scala

core/shared/src/main/scala/cats/effect/IOFiber.scala

vasilmkd:
Do you mind restarting the CI? It failed during the initialization stage.

core/shared/src/main/scala/cats/effect/IOFiber.scala
RaasAhsan: I think these are good 👍

RaasAhsan:
I think merging `series/3.x` back in will help the build pass a bit, since we squashed 3 different bugs

djspiewak:
Looks like #1019 is a problem.

vasilmkd:
I had already rebased the branch.

Issue #1008, opened 2020-07-28
Weird issue seen once in DeferredJVM tests


djspiewak:
```
[info] DeferredJVMParallelism1Tests
Exception in thread "io-compute-5" java.lang.NullPointerException
  at cats.effect.ByteStack.checkAndGrow(ByteStack.scala:67)
  at cats.effect.ByteStack.push(ByteStack.scala:27)
  at cats.effect.IOFiber.cats$effect$IOFiber$$pushCont(IOFiber.scala:607)
  at cats.effect.IOFiber.run(IOFiber.scala:205)
  | => cat cats.effect.IOFiber.$anonfun$8$$anonfun$4$$anonfun$4(IOFiber.scala:534)
  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
  at java.lang.Thread.run(Thread.java:748)
```

Only saw it once on an orphaned branch. Recording for posterity.

djspiewak:
Related: 

```
Exception in thread "io-compute-cats.effect.concurrent.DeferredSpec-4" java.lang.NullPointerException
	at cats.effect.CallbackStack.apply(CallbackStack.scala:44)
	at cats.effect.IOFiber.cats$effect$IOFiber$$done(IOFiber.scala:225)
	at cats.effect.IOFiber.cats$effect$IOFiber$$asyncCancel(IOFiber.scala:294)
	at cats.effect.IOFiber.cats$effect$IOFiber$$runLoop(IOFiber.scala:313)
	at cats.effect.IOFiber.run(IOFiber.scala:212)
	at cats.effect.IOFiber.$anonfun$runLoop$8(IOFiber.scala:537)
  | => cat java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

Also… twice!

RaasAhsan:
I think both of these are solved with #1012 and #1014

PR #1014, opened 2020-07-29
Resume fiber before initializing in exec

RaasAhsan:
This PR addresses the other half of #1008 , particularly when `ByteStack` operations would throw a `NPE`. The race condition manifests when a fiber starts concurrently with its own cancellation. 

```scala
private[effect] def exec(cur: IO[Any], ec: ExecutionContext): Unit = {
    conts = new ByteStack(16)
    pushCont(RunTerminusK)

    ctxs = new ArrayStack[ExecutionContext](2)
    currentCtx = ec
    ctxs.push(ec)

    if (resume()) {
//      println(s"$name: starting at ${Thread.currentThread().getName} + ${suspended.get()}")
      runLoop(cur, 0)
    }
  }
```

It would initialize `conts` and then push the `RunTerminusK` continuation to it. However, between the execution of these two instructions, the fiber could be cancelled, setting `conts` to `null`. Same memory model rules apply here as in #1012 . We should try to acquire the runloop *before* running any initialization actions.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -216,15 +216,16 @@ private[effect] final class IOFiber[A](
   }
 
   private[effect] def exec(cur: IO[Any], ec: ExecutionContext): Unit = {
-    conts = new ByteStack(16)
-    pushCont(RunTerminusK)
-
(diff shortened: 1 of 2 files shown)


djspiewak:
Oh this is REALLY trippy. I didn't think of this one at all. Nice!

PR #1022, opened 2020-07-30
Reset cancel effect at done

RaasAhsan:
The intended semantics for cancellation dictate that the first canceller of a fiber should backpressure subsequent cancellers until finalization is complete. The `cancel` effect was previously reset right at the beginning of `cancel`, so if that write was propagated to another thread before it attempted cancellation, they would return immediately.


djspiewak:
Good call! I forgot about this when reviewing your first PR.

PR #1024, opened 2020-07-30
Preallocate right unit and reduce casts in mapK continuation

vasilmkd:
Simplified branching and removed casts from `mapK` continuation function.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -128,7 +128,7 @@ private[effect] final class IOFiber[A](
   private[this] val UnmaskK: Byte = 9
 
   // prefetch for Right(())
-  private[this] val rightUnit = IOFiber.RightUnit
+  private[this] val RightUnit = IOFiber.RightUnit
 
   // similar prefetch for Outcome
   private[this] val OutcomeCanceled = IOFiber.OutcomeCanceled.asInstanceOf[OutcomeIO[A]]
@@ -319,7 +319,7 @@ private[effect] final class IOFiber[A](
       runLoop(finalizers.pop(), 0)
     } else {
       if (cb != null)
-        cb(rightUnit)
+        cb(RightUnit)
 
       done(OutcomeCanceled)
     }
@@ -569,7 +569,7 @@ private[effect] final class IOFiber[A](
 
             val next = IO.async[Unit] { cb =>
               IO {
-                val cancel = scheduler.sleep(cur.delay, () => cb(rightUnit))
+                val cancel = scheduler.sleep(cur.delay, () => cb(RightUnit))
                 Some(IO(cancel.run()))
               }
             }
@@ -857,7 +857,7 @@ private[effect] final class IOFiber[A](
       // resume external canceller
       val cb = objectState.pop()
       if (cb != null) {
-        cb.asInstanceOf[Either[Throwable, Unit] => Unit](rightUnit)
+        cb.asInstanceOf[Either[Throwable, Unit] => Unit](RightUnit)
       }
       // resume joiners
       done(OutcomeCanceled)
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -127,6 +127,9 @@ private[effect] final class IOFiber[A](
   private[this] val UncancelableK: Byte = 8
   private[this] val UnmaskK: Byte = 9
 
+  // prefetch for Right(())
+  private[this] val rightUnit = IOFiber.RightUnit
+
   // similar prefetch for Outcome
   private[this] val OutcomeCanceled = IOFiber.OutcomeCanceled.asInstanceOf[OutcomeIO[A]]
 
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -127,6 +127,9 @@ private[effect] final class IOFiber[A](
   private[this] val UncancelableK: Byte = 8
   private[this] val UnmaskK: Byte = 9
 
+  // prefetch for Right(())
+  private[this] val rightUnit = IOFiber.rightUnit
```
djspiewak: Btw I generally prefer capitalized constants (like object names) in Scala. In this case: `RightUnit`. Minor nit. :-)

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -811,27 +814,12 @@ private[effect] final class IOFiber[A](
...
-      else
-        IO.Error(transformed.asInstanceOf[Throwable])
-    } else {
-      if (success)
-        succeeded(transformed, depth + 1)
-      else
-        failed(transformed.asInstanceOf[Throwable], depth + 1)
+    try {
+      val transformed = f(result)
+      if (depth > MaxStackDepth) IO.Pure(transformed) else succeeded(transformed, depth + 1)
```
djspiewak: This is much slower! :-( In my benchmarks, about 10-15%. The problem is subtle, but notice what `succeeded` is doing: this is what leads into subsequent `map`s. So if you just have a giant chain of `maps`, `succeeded` is where that nesting takes place (up until `MaxStackDepth`). The problem comes in with the fact that *all* of those `succeeded`s are individually wrapped in pointless `try`/`catch` blocks, because this particular invocation is *inside* the `try`. Meaning that if you have 500 `map`s, then the last one will be wrapped in 500 `try`s. The way it was previously written, you would only ever have a single layer of `try`/`catch`, which allows HotSpot to fast-path a lot of stuff.

This is basically the justification for how contorted it was. :-) I probably should have put this into a comment.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -1010,4 +998,5 @@ private object IOFiber {
 
   // prefetch
   final private val OutcomeCanceled = Outcome.Canceled()
+  final private val rightUnit = Right(())
```
djspiewak: Lovely optimization. Basically just easy low-hanging fruit and I didn't think of it at all. Nice work!

djspiewak (changes requested):
I love `RightUnit`, but the `mapK` rewrite is actually a significant deoptimization. :-(

Also, I'm really really sorry everyone keeps hitting these `PureConc` build failures. I've found the problem and I'm working on fix it.

vasilmkd:
Thanks for the helpful comments. Can you please double check `runLoop` case `Delay` and `BlockingClosure` body, I did something similar, just to make sure I didn't write what I shouldn't have.

Edit: Also `flatMapK` and `handleErrorWithK`.

core/shared/src/main/scala/cats/effect/IOFiber.scala

core/shared/src/main/scala/cats/effect/IOFiber.scala
vasilmkd: How about now? Also, please double check the other cases I mentioned. 😄

djspiewak: Looks excellent now! I do prefer this to the `Boolean` approach.

djspiewak:
> Thanks for the helpful comments. Can you please double check runLoop case Delay and BlockingClosure body, I did something similar, just to make sure I didn't write what I shouldn't have.

Looks like `Delay` has the same problem! Sorry I didn't catch this in the review of the inlining PR. `BlockingClosure` is fine, since you're just tossing it into an `ExecutionContext`, but there is a subtle difference, which is that any errors in communicating with the *pool itself* will now be caught and re-wrapped. I'm tentatively okay with that, since the behavior is undefined, but it has proven to be a bit weird in the past, so I would rather if a similar trick were played there.

`flatMapK` and `handleErrorWithK` are both fine, since the only continuation cases (their invocations of `failed`) are in the `catch` rather than in the body of the `try`.

vasilmkd:
Let's fix `BlockingClosure` too. What do you propose?

> Looks like `Delay` has the same problem! Sorry I didn't catch this in the review of the inlining PR.

That's my fault, I kind of snuck those changes in the PR.

djspiewak:
Something like this maybe?

```scala
val err: Throwable = null
val r = try {
  cur.thunk()
} catch {
  case NonFatal(t) => err = t
}

if (err == null) {
  afterBlockingSuccessfulClosure.prepare(r, nextIteration)
  currentCtx.execute(afterBlockingSuccessfulClosure)
} else {
  afterBlockingFailedClosure.prepare(t, nextIteration)
  currentCtx.execute(afterBlockingFailedClosure)
}
```

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -127,6 +127,9 @@ private[effect] final class IOFiber[A](
   private[this] val UncancelableK: Byte = 8
   private[this] val UnmaskK: Byte = 9
 
+  // prefetch for Right(())
+  private[this] val rightUnit = IOFiber.RightUnit
```
djspiewak: You changed the one in the companion, but not the one here. :-D

PR #1026, opened 2020-07-31
Side channel error reporting

### LINKED ISSUE #943: Make sure all the side-channel stuff is reported
There are several cases where errors (and even values) need to be side-channel reported because something bad happened. For example, in `IO.race(IO.raiseError(e1), IO.raiseError(e2))`, the loser needs to be reported on the `reportFailure` channel provided by the `ExecutionContext`. There are a ton of TODOs around `IOFiber` on this element, and it should be a simple matter of running through and just taking care of all of them.

RaasAhsan:
Closes #943 . Open question on async registration failures when the callback succeeds.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -119,19 +119,26 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   def onCancel(fin: IO[Unit]): IO[A] =
     IO.OnCancel(this, fin)
 
-  def onCase(pf: PartialFunction[OutcomeIO[A @uncheckedVariance], IO[Unit]]): IO[A] =
+  def onCase(pf: PartialFunction[OutcomeIO[A @uncheckedVariance], IO[Unit]]): IO[A] = {
+    def doOutcome(outcome: OutcomeIO[A]): IO[Unit] =
+      pf.lift(outcome)
+        .fold(IO.unit)(_.attempt.flatMap {
+          case Right(_) => IO.unit
+          case Left(e) => IO.executionContext.flatMap(ec => IO(ec.reportFailure(e)))
+        })
+
     IO uncancelable { poll =>
       val base = poll(this)
       val finalized = pf.lift(Outcome.Canceled()).map(base.onCancel(_)).getOrElse(base)
 
       finalized.attempt flatMap {
         case Left(e) =>
-          pf.lift(Outcome.Errored(e)).map(_.attempt).getOrElse(IO.unit) *> IO.raiseError(e)
-
+          doOutcome(Outcome.Errored(e)) *> IO.raiseError(e)
         case Right(a) =>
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -932,6 +936,7 @@ private[effect] final class IOFiber[A](
       // we got the error *after* the callback, but we have queueing semantics
       // therefore, side-channel the callback results
       // println(state.get())
+      // currentCtx.reportFailure(new AsyncRegistrationException(state.get()))
```
RaasAhsan: How should we report these?

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -83,16 +83,23 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   def bracket[B](use: A => IO[B])(release: A => IO[Unit]): IO[B] =
     bracketCase(use)((a, _) => release(a))
 
-  def bracketCase[B](use: A => IO[B])(release: (A, OutcomeIO[B]) => IO[Unit]): IO[B] =
+  def bracketCase[B](use: A => IO[B])(release: (A, OutcomeIO[B]) => IO[Unit]): IO[B] = {
+    def doRelease(a: A, outcome: OutcomeIO[B]): IO[Unit] =
+      release(a, outcome).attempt.flatMap {
+        case Right(_) => IO.unit
+        case Left(e) => IO.executionContext.flatMap(ec => IO(ec.reportFailure(e)))
+      }
```
```suggestion
      release(a, outcome) handleErrorWith { e =>
        IO.executionContext.flatMap(ec => IO(ec.reportFailure(e)))
      }
```

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -112,19 +119,26 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
...
     IO.OnCancel(this, fin)
 
-  def onCase(pf: PartialFunction[OutcomeIO[A @uncheckedVariance], IO[Unit]]): IO[A] =
+  def onCase(pf: PartialFunction[OutcomeIO[A @uncheckedVariance], IO[Unit]]): IO[A] = {
+    def doOutcome(outcome: OutcomeIO[A]): IO[Unit] =
+      pf.lift(outcome)
+        .fold(IO.unit)(_.attempt.flatMap {
+          case Right(_) => IO.unit
+          case Left(e) => IO.executionContext.flatMap(ec => IO(ec.reportFailure(e)))
+        })
```
djspiewak: Can use `handleErrorWith` here as well.

core/shared/src/main/scala/cats/effect/IOFiber.scala
djspiewak: I'm tentatively okay with memory-holing it. I'm honestly not sure. It's a weird situation. Do people care about hearing that they had orphaned values? They can't *get* the value and do anything about it.

djspiewak (changes requested):

Issue #1017, opened 2020-07-29
Allocated: New Typeclass Below Concurrent


djspiewak:
This one is a lot more annoying than `MonadCancel`, tbh. What I'm proposing would sit under `Concurrent` and adjacent to `Temporal`, with `Async` deriving from both. The core of this typeclass would simply be the ability to create a `Ref` and/or `Deferred`. Part of the thought process here is the fact that `Ref` and `Deferred` as so incredibly widely needed, and the only way to get them in the present hierarchy is to have an `Async`. This undermines one of the design goals of the hierarchy, which was allowing people to use safer classes (like `Concurrent`) which cannot capture side-effects. Additionally, there are examples of datatypes which trivially can implement `Ref` and `Deferred` but which cannot form an `Async` (e.g. `PureConc`).

Now, to be clear, I do think that much of the code which *today* relies on `Ref` and `Deferred` can probably be rewritten without loss of performance or clarity to use `Concurrent`. That isn't possible for *all* cases though (or possibly even *most*), and the rewriting is not always obvious. This is part of what drives the motivation for this proposed class.

The biggest problem with this sort of typeclass, aside from the fact that it introduces mutation directly into the kernel, is the fact that it's incredibly challenging to talk about laws and ensure that implementations are actually correct. We would need to sufficiently constrain implementations to a very tightly defined set of semantics, otherwise it is very likely that users of this typeclass (mostly middleware frameworks) would end up inadvertently assuming Cats Effect semantics, which effectively would turn this into a false abstraction. We would need to be very careful about this.

djspiewak:
Ongoing discussion: https://gitter.im/typelevel/cats-effect-dev?at=5f21855e27799668010c2b12

djspiewak:
The conclusion here seems to be incoherent "maker" typeclasses on the companions (e.g. `Ref.Mk`, `Semaphore.Mk`, etc), rather than trying to stuff something into the hierarchy. This also allows `Ref` etc to remain in the **concurrent** module.

djspiewak:
Restoring this. Some context: https://gitter.im/typelevel/cats-effect-dev?at=5f3c11c3ef8d4e243b482298

Tldr `Mk` isn't working. Let's try `Allocate`.

RaasAhsan:
[...]

Issue #1019, opened 2020-07-29
New and exciting PureConc laws failure


djspiewak:
```
2020-07-29T18:37:55.0946411Z [error] x concurrent.race never identity attempt (left)
2020-07-29T18:37:55.0946767Z [error]  Falsified after 17 passed tests.
[...]

PR #1028, opened 2020-07-31
wip: separate typeclasses for creating concurrent data structures

oleg-py:
Here's my vision for #1017.

This is using a variant with 2 type constructors as a main implementation, where 1-type-constructor is just an alias. That cuts on code duplication and establishes a rule where all (non-unsafe) construction is done only through the typeclass.

I'm experimenting a bit on how practical it's to cake them together if the app wants to avoid huge lists of constraints (e.g. you can do "unofficial" `class Allocated[F[_]](...) extends Ref.Mk[F] with Deferred.Mk[F] with Semaphore.Mk[F]` for your app if it's a common combo you're using together)

Every variant of construction (e.g. empty mvar / full mvar) has its own method. The methods are also named somewhat verbosely for now (I've taken the names from [tofu](https://github.com/TinkoffCreditSystems/tofu) for the most part).

Also I've made it so that Semaphore creation does not require FFI classes directly. Still need to add some docs.

/cc @Odomontois

### DIFF
--- a/concurrent/shared/src/main/scala/cats/effect/concurrent/Deferred.scala
+++ b/concurrent/shared/src/main/scala/cats/effect/concurrent/Deferred.scala
@@ -94,14 +94,14 @@ object Deferred {
   /**
    * Creates an unset promise. *
    */
-  def apply[F[_], A](implicit F: Async[F]): F[Deferred[F, A]] =
-    F.delay(unsafe[F, A])
+  def apply[F[_], A](implicit mk: Mk[F]): F[Deferred[F, A]] =
+    mk.deferred[A]
 
   /**
    * Creates an unset tryable promise. *
    */
-  def tryable[F[_], A](implicit F: Async[F]): F[TryableDeferred[F, A]] =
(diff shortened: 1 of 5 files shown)


djspiewak:
This looks entirely reasonable to me! Requesting some reviews to get people's opinions. Also the build failures are caused by #1019, not by anything you did. I'm working on it. :-/

djspiewak:
Btw @oleg-py if you merge series/3.x, the build failures should be gone.

djspiewak:
I'm 👍 on trying this out. Though, I do think we need to get some inductive instances in here for each of these classes to handle cases like this one:

```scala
def bar[F[_]: ConcurrentThrow: Ref.Mk](fa: F[A]) = ???

def foo[F[_]: ConcurrentThrow: Ref.Mk](fa: F[A]) = 
  bar(OptionT.liftF(fa))
```

We can't re-derive from `Async` because we don't have the `Async` constraint any longer, so this won't resolve.

mpilquist (approved):

PR #1030, opened 2020-07-31
Fix SemigroupK[Resource[F, *]].combineK 

### LINKED ISSUE #949: Idea: Change the behavior of SemigroupK.combineK for Resource
The current implementation of `SemigroupK.combineK` for **Resource** is the following:

```scala
def combineK[A](rx: Resource[F, A], ry: Resource[F, A]): Resource[F, A] =
  for {
    x <- rx
    y <- ry
    xy <- Resource.liftF(K.combineK(x.pure[F], y.pure[F]))
  } yield xy
```

This implies that both resources need to be successfully acquired and then it combines both of them using the **SemigroupK** instance of the effect of the resource.
This current behavior doesn't feel too useful for me _(I may be ignoring something)_ since it requires both **Resources** to be successfully acquired but then the common implementation of `combineK` for effects is to get the first one that was successful or the error of the first one.

------

I came across this while trying to implement the following use case:

I had three ways of getting some resource. And I wanted to try all these three in some order.
If the first one worked use that, if it didn't try with the second way and then the third one.

So, after checking the docs and seeing there was a **SemigroupK** instance for **Resource** I thought this would do work:

```scala
val r = r1 <+> r2 <+> r3
```

But, this ended in `r` being `r1` as long as all the three worked or failing with the exception of the first one to fail.
_(so, exactly the opposite of what I wanted)_

Hopefully, I could get what I wanted using `ApplicativeError.orElse` instead.
So, I wonder if there is really any reason or use case for this behavior.

----

Simple snippet to test the code:

```scala
import $ivy.`org.typelevel::cats-effect:2.1.4` 
  
import cats.implicits._
import cats.effect.implicits._
import cats.effect.{IO, Resource}

final case class Err(msg: String) extends Throwable(msg) 

def resource(id: Int, fail: Boolean = false): Resource[IO, Int] =
  Resource.liftF {
    if (fail) IO.raiseError(Err(s"Failed resource ${id}"))
    else IO.delay(println(s"Computing resource ${id}")).as(id)
  }

// You can play with the fail flag.
val r1 = resource(id = 1, fail = false)
val r2 = resource(id = 2, fail = false)
val r3 = resource(id = 3, fail = false)

// val r = r1 <+> r2 <+> r3
val r = r1 orElse r2 orElse r3

val program = r.use(x => IO.pure(x))
program.unsafeRunSync()
```

BalmungSan:
Changing the behavior of SemigroupK[Resource[F, *]].combineK to be consistent with MonadError[Resource[F, *], E].orElse

Fixes #949

-----

I am surprised I didn't have to change any test _(this is probably a sign that there was no reason for the previous behavior)_.
Should I add some tests for the current behavior?

BTW, I confirmed it worked as expected using the snippet posted in the original issue.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Resource.scala
+++ b/core/shared/src/main/scala/cats/effect/Resource.scala
@@ -561,11 +561,10 @@ abstract private[effect] class ResourceInstances0 {
       def F = F0
     }
 
-  implicit def catsEffectSemigroupKForResource[F[_], A](implicit F0: Monad[F],
-                                                        K0: SemigroupK[F]): ResourceSemigroupK[F] =
-    new ResourceSemigroupK[F] {
-      def F = F0
-      def K = K0
+  implicit def catsEffectSemigroupKForResource[F[_], E](implicit F0: MonadError[F, E]): SemigroupK[Resource[F, *]] =
+    new SemigroupK[Resource[F, *]] {
+      final override def combineK[A](a: Resource[F, A], b: Resource[F, A]): Resource[F, A] =
+        MonadError[Resource[F, *], E].handleErrorWith(a)(_ => b)
     }
 }
 
@@ -638,18 +637,6 @@ abstract private[effect] class ResourceSemigroup[F[_], A] extends Semigroup[Reso
       x <- rx
       y <- ry
     } yield A.combine(x, y)
-}
-
-abstract private[effect] class ResourceSemigroupK[F[_]] extends SemigroupK[Resource[F, *]] {
-  implicit protected def F: Monad[F]
-  implicit protected def K: SemigroupK[F]
-
-  def combineK[A](rx: Resource[F, A], ry: Resource[F, A]): Resource[F, A] =
-    for {
-      x <- rx
-      y <- ry
-      xy <- Resource.liftF(K.combineK(x.pure[F], y.pure[F]))
-    } yield xy
 }
 
 abstract private[effect] class ResourceLiftIO[F[_]] extends LiftIO[Resource[F, *]] {
--- a/core/shared/src/main/scala/cats/effect/Resource.scala
+++ b/core/shared/src/main/scala/cats/effect/Resource.scala
@@ -566,6 +566,27 @@ abstract private[effect] class ResourceInstances0 {
       final override def combineK[A](a: Resource[F, A], b: Resource[F, A]): Resource[F, A] =
         MonadError[Resource[F, *], E].handleErrorWith(a)(_ => b)
     }
+
+  // For binary compatibility.
+  @deprecated("Use the new implementaion that behaves like orElse", since = "2.2.0")
+  def catsEffectSemigroupKForResource[F[_], A](implicit F0: Monad[F],
+                                               K0: SemigroupK[F]): ResourceSemigroupK[F] =
+    new ResourceSemigroupK[F] {
+      def F = F0
+      def K = K0
+    }
+}
+
+@deprecated("Use the new implementaion that behaves like orElse", since = "2.2.0")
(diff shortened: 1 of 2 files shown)


BalmungSan:
It seems my implementation is not binary compatible.
I could preserve the old class, and still require a **SemigroupK** for `F` _(even if unneeded)_.
But still, the new implementation requires a **MonadError** for `F` instead of just a **Monad**... so, it wouldn't matter.

djspiewak:
Ah! So the trick to appease the MiMa gods here is to take the old method, remove the `implicit` modifier and make it `private[effect]`. It *compiles* to the same bytecode as what you have there now, but any call sites will recompile to use the new (correct) signature. It's definitely worth doing this, even though the old one was kind of broken and dumb, just to make sure we really hold up the "binary compatible" guarantee. :-)

BalmungSan:
@djspiewak oh good to know, I will make the change.
However, even if it is binary compatible. It is acceptable that using a different jar _(that was supposed to be binary compatible)_ will produce a different runtime behavior? Or in this case, are we considering this a bug fix?

BTW, what do you think about the tests?

djspiewak:
> However, even if it is binary compatible. It is acceptable that using a different jar (that was supposed to be binary compatible) will produce a different runtime behavior? Or in this case, are we considering this a bug fix?

Well, it's still going to produce the same runtime behavior. You *can't* call the new implementation from the old signature, so you just have to keep the old implementation around, but hidden behind the `private[effect]` signature. So that part should be okay.

If we *were* in the situation you described, I definitely would need to be sure it was a bugfix. :-)

> BTW, what do you think about the tests?

I think it's definitely worth adding some tests. It doesn't entirely surprise me that the behavior was previously untested, since I feel like anyone *writing* such a test would realize how silly the semantics were.

core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -561,14 +561,33 @@ abstract private[effect] class ResourceInstances0 {
...
-  implicit def catsEffectSemigroupKForResource[F[_], A](implicit F0: Monad[F],
-                                                        K0: SemigroupK[F]): ResourceSemigroupK[F] =
+  implicit def catsEffectSemigroupKForResource2[F[_], E](implicit F0: MonadError[F, E]): SemigroupK[Resource[F, *]] =
+    new SemigroupK[Resource[F, *]] {
+      final override def combineK[A](a: Resource[F, A], b: Resource[F, A]): Resource[F, A] =
+        MonadError[Resource[F, *], E].handleErrorWith(a)(_ => b)
+    }
+
+  // For binary compatibility.
+  @deprecated("Use the new implementaion that behaves like orElse", since = "2.2.0")
```
djspiewak: Minor pedantry: I would specify the new implementation more explicitly. Maybe something like this:

```suggestion
  @deprecated("use catsEffectSemigroupKForResource2, which behaves more like orElse (or implicitly via SemigroupK[Resource[F, *]])", since = "2.2.0")
```

Ish? Not sure.

laws/shared/src/test/scala/cats/effect/ResourceTests.scala
```diff
@@ -334,6 +334,15 @@ class ResourceTests extends BaseTestsSuite {
...
   }
 
+  test("combineK - should behave like orElse") {
+    check { (r1: Resource[IO, Int], r2: Resource[IO, Int]) =>
+      val lhs = r1.orElse(r2).use(IO.pure).attempt.unsafeRunSync()
+      val rhs = (r1 <+> r2).use(IO.pure).attempt.unsafeRunSync()
+
+      lhs <-> rhs
+    }
+  }
```
djspiewak, line 344: This is very nice. 👍

Issue #1015, opened 2020-07-29
New Nondeterminism in MVarSpec


djspiewak:
```
[info] MVarSpec
[error] x put is cancelable
[error]  Set(1, 2) != Set(1, 3) (MVarSpec.scala:49)
[error] Added (1)
[error] 2
[error] 
[error] Missing (1)
[error] 3
[error] 
[info] + take is cancelable
```

RaasAhsan:
Its worth pointing out that there is a definite source of non-determinism in the test:
```scala
"put is cancelable" in real {
  val op = for {
    mVar <- init(0)
    gate1 <- Deferred[IO, Unit]
    gate2 <- Deferred[IO, Unit]
    gate3 <- Deferred[IO, Unit]
    _ <- (mVar.put(1) *> gate1.complete(()) *> gate3.get *> mVar.put(3)).start
    p2 <- (gate1.get *> gate2.complete(()) *> mVar.put(2)).start
    _ <- mVar.take // Take the initial 0
    _ <- gate2.get
//!!!!!!!!
    _ <- IO.sleep(10.millis) // Give more chance for `p2` put to register 
//!!!!!!!!
    _ <- p2.cancel
    _ <- gate3.complete(())
    r1 <- mVar.take
    r3 <- mVar.take
  } yield Set(r1, r3)

  op.flatMap { res =>
    IO {
      res mustEqual Set(1, 3)
    }
  }
}
```
I've seen this once or twice before also. I think we should probably rewrite the test so that it isn't relying on time to wait for some invariant.

djspiewak:
That's a superb way of annotating it…

vasilmkd:
I tried fixing this but I'm out of ideas.

djspiewak:
It's really tricky. You basically need to somehow say "wait until you're inside of the `MVar#put` effect". :-/

vasilmkd:
Working on this again.

PR #1032, opened 2020-08-01
Add SyncIO

### LINKED ISSUE #924: Implement SyncIO
For now, we should just have `SyncIO` be a little newtype wrapper around `IO`, just so we can make forward progress. Long-term, it should be possible (and not hard) to write a synchronous-only runloop for `SyncIO` which is *much* faster than what is in `IO`, similar to what master does.

vasilmkd:
Resolves #924.


core/shared/src/main/scala/cats/effect/SyncIO.scala
```diff
@@ -0,0 +1,721 @@
...
+   * via `guarantee` in polymorphic code. Prefer `bracket`
+   * for the acquisition and release of resources.
+   *
+   * @see [[SyncIO#bracket]] for the more general operation
+   *
+   * @param finalizer `SyncIO` to be executed when this `SyncIO` is finished
+   * @return this `SyncIO` with registered `finalizer`
+   */
+  def guarantee(finalizer: SyncIO[Unit]): SyncIO[A] =
+    SyncIO.unit.bracket(_ => this)(_ => finalizer)
```
djspiewak: I still don't think this (or `bracket`) are very useful.

core/shared/src/main/scala/cats/effect/SyncIO.scala
```diff
@@ -0,0 +1,721 @@
...
+  /**
+   * Transforms this `SyncIO` into an [[IO]].
+   *
+   * @note This is a dangerous operation as it might end up executing blocking code on execution contexts which is not
+   *       designated for that purpose.
+   *
+   * @return an `IO` that evaluates to the result of this `SyncIO`
+   */
+  def toIO: IO[A] =
+    IO(unsafeRunSync())
```
djspiewak: This will produce slightly different results than `to[IO]`. I think it's worth dropping this function altogether.

core/shared/src/main/scala/cats/effect/SyncIO.scala
```diff
@@ -0,0 +1,721 @@
...
+  def void: SyncIO[Unit] =
+    map(_ => ())
+
+  override def toString(): String = "SyncIO(...)"
+
+  // unsafe
+
+  // mutable nullable state to avoid boxing of the result
+  private[this] var resultFailure: Throwable = _
+  private[this] var resultSuccess: A = _
```
djspiewak: I do not believe this is correct. `SyncIO` should not memoize. Also, what happens if this is run from multiple threads simultaneously?

core/shared/src/main/scala/cats/effect/SyncIO.scala
```diff
@@ -0,0 +1,721 @@
...
+   */
+  def fromTry[A](t: Try[A]): SyncIO[A] =
+    t.fold(raiseError, pure)
+
+  /**
+   * `SyncIO#to` as a natural transformation.
+   *
+   * @see [[SyncIO#to]]
+   */
+  def toK[F[_]: SyncEffect]: SyncIO ~> F =
```
djspiewak: I think this constraint should be `Sync`.

core/shared/src/test/scala/cats/effect/SyncIOSpec.scala
```diff
@@ -0,0 +1,190 @@
...
+import org.specs2.ScalaCheck
+
+import org.typelevel.discipline.specs2.mutable.Discipline
+
+class SyncIOSpec extends IOPlatformSpecification with Discipline with ScalaCheck with BaseSpec {
+
+  import SyncTypeGenerators._
+
+  // we just need this because of the laws testing, since the prop runs can interfere with each other
+  sequential
```
djspiewak: We probably don't need it for `SyncIO` though!

djspiewak (changes requested):
This is very exciting! Just a few things but overall it seems excellent.

core/shared/src/main/scala/cats/effect/SyncIO.scala
vasilmkd: It doesn't memoize, I used this only for signaling, then they are released. But you are right about running them from different threads. Fixing.

core/shared/src/test/scala/cats/effect/SyncIOSpec.scala
vasilmkd: Sorry, I copied this from `IOSpec`. Fixing.

core/shared/src/main/scala/cats/effect/SyncIO.scala
vasilmkd: I saw this in CE2. Deleting.

core/shared/src/main/scala/cats/effect/SyncIO.scala
vasilmkd: Removed both `bracket` and `guarantee` and the accompanying tests.

core/shared/src/main/scala/cats/effect/SyncIO.scala
vasilmkd: Fixed.

core/shared/src/test/scala/cats/effect/SyncIOSpec.scala
vasilmkd: Done.

core/shared/src/main/scala/cats/effect/SyncIO.scala
vasilmkd: Good catch. I copied this from `SyncEffect` directly.

core/shared/src/main/scala/cats/effect/SyncIO.scala
vasilmkd: Done.

vasilmkd:
CI error related to #1015.

PR #1036, opened 2020-08-02
Fix for uncancelable in IO#to

### LINKED ISSUE #993: Potential loss of semantics in IO#to
I'm not convinced this is sound:

```scala
            val poll2 = new (IO ~> IO) {
              def apply[B](ioa: IO[B]): IO[B] =
                F.to[IO](poll(ioa.to[F]))
            }
```

Specifically, the `F.to[IO]` bit. So the `poll` here is coming from `F.uncancelable`. Imagine, for example, if `F` is implementing `uncancelable` in the same fashion as `IO`, where there's a special constructor, `IO.Unmask`, which corresponds to it. This is not directly convertable, as we see a little later on in the function:

```scala
        case IO.Unmask(ioa, _) => ioa.to[F] // polling should be handled by F
```

When going from `IO` to `F`, we just ignore `Unmask`. So in other words, if you imagine a hypothetical `F` which has an internal implementation very much like `IO`'s, this function will not do the right thing! In fact, I'm relatively convinced that it doesn't do the right thing going from `IO` to `IO`, which ironically is law-tested, but the generators do not frequently create things which are sensitive to polling, and in any case our equality checkers cannot see it.

So tldr I think this is broken, and it *may* illustrate a broader brokenness with the concept of `Effect#to`. We almost certainly need to find a different way to implement this.

vasilmkd:
Resolves #993.

Opening as a draft to show that the tests pass for the actual implementation, instead of the fast path. Also, there's some code that needs to be changed, in particular, what to do about the now unused `Unmask` case.

Edit:
Most of the changes are formatting. Sorry about that.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -200,66 +200,75 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   def start: IO[FiberIO[A @uncheckedVariance]] =
     IO.Start(this)
 
-  def to[F[_]](implicit F: Effect[F]): F[A @uncheckedVariance] =
+  def to[F[_]](implicit F: Effect[F]): F[A @uncheckedVariance] = {
     // re-comment this fast-path to test the implementation with IO itself
-    if (F eq IO.effectForIO) {
-      asInstanceOf[F[A]]
-    } else {
-      def fiberFrom[B](f: Fiber[F, Throwable, B]): FiberIO[B] =
-        new FiberIO[B] {
-          val cancel = F.to[IO](f.cancel)
-          val join = F.to[IO](f.join).map(_.mapK(F.toK[IO]))
+    // if (F eq IO.effectForIO) {
+    // asInstanceOf[F[A]]
+    // } else {
+    def fiberFrom[B](f: Fiber[F, Throwable, B]): FiberIO[B] =
+      new FiberIO[B] {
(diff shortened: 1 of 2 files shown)


djspiewak:
Hmm, I like the idea of what you're doing here quite a bit, and this is *roughly* what I had in mind. Originally I was thinking that it could be done entirely with an inner function which maintained a mapping of which `poll` was which, but I think your approach here is better. It's unfortunate because it does require creating a new `IO` node, but fortunately that node doesn't impact the runtime since it will never appear in the interpreter.

I'm 👍 for moving forward with this.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -618,4 +627,8 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
   private[effect] final case class Unmask[+A](ioa: IO[A], id: Int) extends IO[A] {
     def tag = 19
   }
+
+  private[effect] final case class UnmaskF[F[_], A](ioa: IO[A], poll: F ~> F) extends IO[A] {
+    def tag = 20
```
djspiewak: I would pick a more obviously "don't try this at home" `tag`, like `-1`.

vasilmkd: Should I do the same for `Unmask` too?

vasilmkd: Done.

PR #1038, opened 2020-08-04
Replaced switch with (faster) jump

djspiewak:


vasilmkd:
For context:
```
[warn] -- [E115] Syntax Warning: /home/runner/work/cats-effect/cats-effect/core/shared/src/main/scala/cats/effect/SyncIO.scala:303:19 
[warn] 303 |      (k: @switch) match {
[warn]     |      ^
[warn]     |Could not emit switch for @switch annotated match since there are not enough cases
[warn] 304 |        // (case 0) will never continue to mapK
[warn] 305 |        // (case 1) will never continue to flatMapK
[warn] 306 |        case 2 => handleErrorWithK(error, depth)
[warn] 307 |        case 3 => SyncIO.Failure(error)
[warn] 308 |      }
```

PR #1039, opened 2020-08-04
Add UnsafeRun

### LINKED ISSUE #1020: Add cats.effect.unsafe.UnsafeRun to kernel
Blegh…

In some cases where you're embedded into unsafe code which is running *you* and where you don't already have a natural runtime to fall back on, you need something like `UnsafeRun`. A good example of this is http4s-blaze. `Effect` doesn't solve this problem because the best http4s could do is pick a target effect type (say, `IO`) and convert `F` to `IO`, but then it needs an `IORuntime` in order to actually run that `IO`, and if someone is running in Monix or ZIO, they aren't likely to have one that makes sense.

`UnsafeRun` solves this problem at the cost of being really dumb and annoying, but sometimes you need it. `IOApp` should provide an instance for `IO`, with the constructor defined on `IORuntime`.

```scala
package cats.effect.unsafe

// intentionally incoherent
trait UnsafeRun[F[_]] {
  def unsafeRunFutureCancelable[A](fa: F[A]): (Future[A], () => Future[Unit])
}
```

vasilmkd:
Resolves #1020.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
+++ b/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
@@ -14,9 +14,10 @@
  * limitations under the License.
  */
 
-package cats.effect.unsafe
+package cats.effect
+package unsafe
 
-import scala.concurrent.ExecutionContext
+import scala.concurrent.{ExecutionContext, Future, Promise}
 
 @annotation.implicitNotFound("""Could not find an implicit IORuntime.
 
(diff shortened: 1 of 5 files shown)


core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
```diff
@@ -37,6 +38,20 @@ final class IORuntime private (
     val scheduler: Scheduler,
     val shutdown: () => Unit) {
 
+  private implicit val self = this
+
+  val unsafeRunForIO: UnsafeRun[IO] =
+    new UnsafeRun[IO] {
+      def unsafeRunFutureCancelable[A](fa: IO[A]): (Future[A], () => Future[Unit]) = {
+        val p = Promise[A]()
+        val fiber = fa.unsafeRunFiber(true) {
```
djspiewak: We probably want `false` here

djspiewak (changes requested):

core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala

djspiewak (approved):

Issue #1025, opened 2020-07-31
Experiment with work-stealing pools


djspiewak:
Something that would be fun would be to experiment with a work-stealing *fixed* thread pool implementation. The current default pool (in both 2.x and 3.x) is a fixed pool, which is important, but the fixed thread pool in Java is non-work-stealing, meaning that individual threads can end up getting overloaded while others sit idle in certain workloads. Conversely, Java provides a work-stealing pool, but the implementation (to my knowledge! verify this please!) is just fork-join on the backend, meaning that it isn't fixed!

What we really want is a work-stealing pool with a strictly fixed worker count. If this doesn't exist in the JDK, then we should honestly just write one.

This is obviously a very experimental task! But it should be fun. Any PR should probably come accompanied by benchmarks showing the impact.

RaasAhsan:
> but the fixed thread pool in Java is non-work-stealing, meaning that individual threads can end up getting overloaded while others sit idle in certain workloads.

Could you describe those workloads in more detail and how a work stealing pool would help? Threads in a fixed thread pool executor generally submit and pull tasks from a shared task queue. The only scenarios I can think of that fit your description is (1) there isn't any work in the shared task queue or (2) you're mixing compute and I/O on the same pool, which is never a great idea to begin with.

I believe work stealing calls for a per-thread local task queue (not sure if the shared queue exists here) where tasks that are spawned from running tasks are queued onto the local task queue. The major benefit here is that threads avoid a great deal of contention when their local queue is non-empty. If any threads run out of work to do, they can try to steal work from another, or if they can't, pull off a shared queue if it exists.

I think work stealing has interesting implications for fiber-oriented workloads. It unlocks a sort of thread affinity, where fibers and their children have a tendency to stick to a single thread, which is great for cache locality.  I'm also interested in seeing if an implementation would include a shared task queue, which has definite consequences for fairness and throughput.

djspiewak:
> Could you describe those workloads in more detail and how a work stealing pool would help?

Ah this is a good point! Thinking about this a little more precisely than I did before…

*Because* there's a a shared task queue, individual workers can't get specifically overloaded, though some workers may certainly end up handling more actions than others. The trick is just that the amount of contention might be exceptionally high, particularly when workers often finish close together.

[...]

PR #1043, opened 2020-08-05
Use fiber as Runnable everywhere and remove preallocated closures

vasilmkd:
PR premise:
1. We didn't quite use `extends Runnable` to its full potential.
- The preallocated closures all did one thing: reschedule the `runLoop` for execution at a later time.
- If `def run(): Unit` was used to avoid allocation of a closure, why not use it to decide how to resume the whole fiber at a later time.
2. We were allocating 6 closure object for every fiber, regardless of the actual need for them.
- Each closure had to be `prepared`, which is an `invokevirtual` bytecode.

PR changes:
1. Remove all preallocated closures and inline their bodies into the class itself.
2. Use the `Runnable` interface to reschedule the fiber.
3. Decide how to resume the fiber by setting the `resumeTag: Byte` and check its value in `def run(): Unit`.
4. Slight cleanup of the constructors.

Desired outcome:
1. Fewer allocations per fiber creation (by removing the closures).
2. `invokespecial` and direct field access instead of `invokevirtual`.
3. Early cleanup of referenced mutable state.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -127,6 +127,15 @@ private final class IOFiber[A](
   private[this] val EvalOnR: Byte = 5
   private[this] val CedeR: Byte = 6
 
+  // mutable state for resuming the fiber in different states
+  private[this] var resumeTag: Byte = ExecR
+  private[this] var resumeNextIteration: Int = 0
+  private[this] var asyncContinueEither: Either[Throwable, Any] = _
+  private[this] var blockingCur: Blocking[Any] = _
+  private[this] var afterBlockingSuccessfulResult: Any = _
+  private[this] var afterBlockingFailedError: Throwable = _
+  private[this] var evalOnIOA: IO[Any] = _
+
   // prefetch for Right(())
   private[this] val RightUnit = IOFiber.RightUnit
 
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -57,11 +57,14 @@ import java.util.concurrent.atomic.{AtomicBoolean, AtomicInteger, AtomicReferenc
  * by the Executor read/write barriers, but their writes are
  * merely a fast-path and are not necessary for correctness.
  */
-private[effect] final class IOFiber[A](
+private final class IOFiber[A](
```
djspiewak, line 60: Wait, how is this instantiated in `IO` if it's `private`?

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -716,100 +681,108 @@ private[effect] final class IOFiber[A](
...
+      runLoop(startIO, 0)
     }
   }
 
-  private[this] final class AfterBlockingSuccessfulClosure extends Runnable {
-    private[this] var result: Any = _
-    private[this] var nextIteration: Int = 0
+  private[this] var resumeNextIteration: Int = 0
+
+  private[this] var asyncContinueEither: Either[Throwable, Any] = _
```
djspiewak: These should be towards the top of the class, rather than intermingled with the `def`s.

djspiewak (changes requested):
This is extremely clever. It actually pushes us quite strongly in the direction of #1025, which is a good thing IMO. The one change I would make is I would prefer if all the `var`s and `val`s remain at the top of the class, rather than intermingling with `def`s.

core/shared/src/main/scala/cats/effect/IOFiber.scala
vasilmkd: `private` for top level class and object definitions only means that the class/object is visible in its given package. `IO` is also in the `cats.effect` package, and therefore can see `IOFiber`. The constructor is not marked private and is available. I think `[effect]` in this case is superfluous. I guess you can use the feature to mark it available in a broader package, maybe `[cats]`.

core/shared/src/main/scala/cats/effect/IOFiber.scala

core/shared/src/main/scala/cats/effect/IOFiber.scala
djspiewak: TIL

PR #1044, opened 2020-08-06
IOFiber Constants Defined in Java (and another minor fix)

djspiewak:
This is a follow-up to #1043 which moves all of the byte constants for `IOFiber` into Java static fields for the JVM, and a Scala `object` for ScalaJS. In this way, we still get the advantages of fast direct access and inlining but without the added cost imposed on `IOFiber` allocation, which is starting to be something we optimize.

In that vein as well, I realized that we can just use a `@volatile` for `outcome`, since we only used `set`/`get` on the `AtomicReference`.

### DIFF
--- a/core/jvm/src/main/java/cats/effect/IOFiberConstants.java
+++ b/core/jvm/src/main/java/cats/effect/IOFiberConstants.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2020 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


core/jvm/src/main/java/cats/effect/IOFiberConstants.java
```diff
@@ -0,0 +1,44 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect;
+
+// defined in Java since Scala doesn't let us define static fields
+final class IOFiberConstants {
+
+  public static int MaxStackDepth = 512;
```
RaasAhsan: We need to declare all of these `final`, otherwise neither the Scala compiler nor the JIT compiler can inline

djspiewak: I realized this last night before heading to bed. Was hoping to fix it before anyone noticed! 😟 I don't know how Java works anymore…

PR #1059, opened 2020-08-07
Suppress normal async resumption after async finalizer suspension

### LINKED ISSUE #1045: ArrayIndexOutOfBoundsException observed when running `racePair` benchmark
```
Exception in thread "io-compute-4" java.lang.ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 16
[error]         at cats.effect.ArrayStack.pop(ArrayStack.scala:35)
[error]         at cats.effect.IOFiber.loop$1(IOFiber.scala:358)
[error]         at cats.effect.IOFiber.$anonfun$runLoop$1(IOFiber.scala:372)
[error]         at cats.effect.IOFiber.$anonfun$runLoop$1$adapted(IOFiber.scala:337)
[error]         at cats.effect.IOFiber.$anonfun$runLoop$5(IOFiber.scala:503)
[error]         at cats.effect.IOFiber.$anonfun$runLoop$5$adapted(IOFiber.scala:503)
[error]         at cats.effect.CallbackStack.apply(CallbackStack.scala:45)
[error]         at cats.effect.IOFiber.done(IOFiber.scala:195)
[error]         at cats.effect.IOFiber.runTerminusSuccessK(IOFiber.scala:820)
[error]         at cats.effect.IOFiber.succeeded(IOFiber.scala:596)
[error]         at cats.effect.IOFiber.runLoop(IOFiber.scala:303)
[error]         at cats.effect.IOFiber.execR(IOFiber.scala:699)
[error]         at cats.effect.IOFiber.run(IOFiber.scala:679)
[error]         at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
[error]         at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630)
[error]         at java.base/java.lang.Thread.run(Thread.java:832)
```

RaasAhsan:
Fix for another async/finalizer related bug. The race condition here occurs when an normal async callback resumes after finalization begins and suspends because of an async finalizer. @djspiewak summarizes the particular ordering of effects that produces this bug quite concisely in https://github.com/typelevel/cats-effect/issues/1045#issuecomment-670251464 so I won't repeat it here.

My attempt at the fix is to introduce another `IOFiber` non-volatile instance variable `finalizing`  that reflects whether finalization has begun already. After the async callback has been invoked and acquires the runloop, we check if the `finalizing` flag has changed since the `Async` node started. If it didn't, we can continue with the runloop. If it did, we suspend the runloop because it means that an async finalizer hasn't completed yet.

Note that the timely publication and visibility of `finalizing` is guaranteed by read and write barriers on `suspended`. I'm also thinking about whether `finalizing` can or should participate in the `shouldFinalize` invariant. My gut feeling says no since the purpose of `shouldFinalize` is to determine whether or not the conditions for finalization have been met, and *not* whether or not we should run them.

Fixes #1045 .

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -80,12 +80,12 @@ private final class IOFiber[A](
   private[this] var ctxs: ArrayStack[ExecutionContext] = _
 
   private[this] var canceled: Boolean = false
+  private[this] var masks: Int = initMask
+  private[this] var finalizing: Boolean = false
 
   // allow for 255 masks before conflicting; 255 chosen because it is a familiar bound, and because it's evenly divides UnsignedInt.MaxValue
   // this scheme gives us 16,843,009 (~2^24) potential derived fibers before masks can conflict
   private[this] val childMask: Int = initMask + 255
-
-  private[this] var masks: Int = initMask
 
   private[this] val finalizers = new ArrayStack[IO[Unit]](16)
 
@@ -247,6 +247,8 @@ private final class IOFiber[A](
 
   private[this] def asyncCancel(cb: Either[Throwable, Unit] => Unit): Unit = {
     //       println(s"<$name> running cancelation (finalizers.length = ${finalizers.unsafeIndex()})")
+    finalizing = true
+
     if (!finalizers.isEmpty()) {
       objectState.push(cb)
 
@@ -331,6 +333,10 @@ private final class IOFiber[A](
            */
           val state = new AtomicReference[AsyncState](AsyncStateInitial)
 
+          // Async callbacks may only resume if the finalization state
+          // remains the same after we re-acquire the runloop
+          val wasFinalizing = finalizing
+
           objectState.push(done)
           objectState.push(state)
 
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -331,6 +333,10 @@ private final class IOFiber[A](
            */
           val state = new AtomicReference[AsyncState](AsyncStateInitial)
 
+          // Async callbacks may only resume if the finalization state
+          // remains the same after we re-acquire the runloop
+          val wasFinalizing = finalizing
```
djspiewak, line 338: I completely didn't even think about this! Nice catch.

djspiewak (approved):
Excellent work. I really wanted to find a way to do this without introducing more state but I couldn't think of anything. The only other option I had was to like, speculatively inspect `objectState` in `asyncCancel` and swap out the finalizer status if it was found, but that would be super-hacky. This is more elegant.

PR #1060, opened 2020-08-07
Change SemigroupK instance for Resource to lift behaviour from underlying effect

bplommer:
Before #949, the `SemigroupK` instance for `Resource` had completely wrong behaviour. It now behaves as expected in the most common cases, but its behaviour is surprising in the case of underlying effects, such as `OptionT[IO, *]`, for which the `SemigroupK` instance behaves like the instance for `Option` (selecting the first non-empty `Option`) rather than the instance for `IO` (selecting the first successful `IO`).

This PR keeps the corrected behaviour from the previous PR, but generalises it by deriving the behaviour from that of the underlying effect.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Resource.scala
+++ b/core/shared/src/main/scala/cats/effect/Resource.scala
@@ -561,10 +561,11 @@ abstract private[effect] class ResourceInstances0 {
       def F = F0
     }
 
-  implicit def catsEffectSemigroupKForResource2[F[_], E](implicit F0: MonadError[F, E]): SemigroupK[Resource[F, *]] =
+  implicit def catsEffectSemigroupKForResource2[F[_]](implicit F0: Bracket[F, Throwable],
+                                                      K: SemigroupK[F]): SemigroupK[Resource[F, *]] =
     new SemigroupK[Resource[F, *]] {
       final override def combineK[A](a: Resource[F, A], b: Resource[F, A]): Resource[F, A] =
-        MonadError[Resource[F, *], E].handleErrorWith(a)(_ => b)
+        Resource(K.combineK(a.allocated, b.allocated))
     }
 
   // For binary compatibility.
--- a/core/shared/src/main/scala/cats/effect/Resource.scala
+++ b/core/shared/src/main/scala/cats/effect/Resource.scala
@@ -561,11 +561,19 @@ abstract private[effect] class ResourceInstances0 {
       def F = F0
     }
 
-  implicit def catsEffectSemigroupKForResource2[F[_]](implicit F0: Bracket[F, Throwable],
+  implicit def catsEffectSemigroupKForResource2[F[_]](implicit F0: Sync[F],
                                                       K: SemigroupK[F]): SemigroupK[Resource[F, *]] =
     new SemigroupK[Resource[F, *]] {
-      final override def combineK[A](a: Resource[F, A], b: Resource[F, A]): Resource[F, A] =
-        Resource(K.combineK(a.allocated, b.allocated))
+
+      final override def combineK[A](ra: Resource[F, A], rb: Resource[F, A]): Resource[F, A] =
+        Resource
+          .make(Ref[F].of(Sync[F].unit))(_.get.flatten)
+          .evalMap { finalizers =>
+            def allocate(r: Resource[F, A]): F[A] =
+              r.allocated.flatMap { case (a, release) => finalizers.update(_.guarantee(release)).as(a) }
+
+            K.combineK(allocate(ra), allocate(rb))
+          }
     }
 
   // For binary compatibility.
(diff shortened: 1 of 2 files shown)


bplommer:
Converted to draft because the release behaviour is incorrect

djspiewak (approved):
Nice catch! (NB: obviously the release behavior needs to be fixed and we need to be careful about uncancelability questions, but this is a good change overall)

bplommer:
Thanks @djspiewak - I think the latest commit corrects the release behaviour. What are the questions about uncancelability?

core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -561,10 +561,19 @@ abstract private[effect] class ResourceInstances0 {
       def F = F0
     }
 
-  implicit def catsEffectSemigroupKForResource2[F[_], E](implicit F0: MonadError[F, E]): SemigroupK[Resource[F, *]] =
+  implicit def catsEffectSemigroupKForResource2[F[_]](implicit F0: Sync[F],
```
djspiewak, line 578: This is a significant constraint to add to something like this. Why exactly do we ned the `Ref`-based implementation below?

djspiewak:
> What are the questions about uncancelability?

Just making sure that acquisition stays acquisition, lifts stay lifts, and finalizers stay finalizers. :-) `Resource` is sometimes a bit tricky since it can be difficult to ensure that the uncancelability boundaries remain in the right place. I haven't puzzled through this implementation just yet though so I can't speak to it one way or another.

core/shared/src/main/scala/cats/effect/Resource.scala
bplommer: > Why exactly do we ned the `Ref`-based implementation below?

I can't say for sure, but I couldn't come up with anything else so I ~stole~ adapted this from the `parTupled` implementation.

That said, I can't see a way to do this without capturing the finalizer through a side-effect. To do that we'd need to either (1) allocate the resource ourselves before invoking the underlying `combineK` or (2) get the finalizer from the result of the underlying `combineK`. (1) was the original approach and is incorrect because it allocated both resources whether or not the underlying `combineK` runs both effects; (2) was my first approach in this PR but it can't guarantee that the underlying `combineK` will produce a finalizer that releases all allocated resources. Maybe there's another way I didn't think of though?

bplommer: I guess it might be possible to do the same thing with `StateT` - I'll give it a try later.

bplommer: Nope, `StateT` has exactly the same problem - if the value gets discarded by the underlying instance, so does the state.

djspiewak:
Btw I haven't forgotten about this. Spent a little time staring at it last night to see if I had any blinding insights. :-) I'd really like to get rid of the `Sync` constraint if at all possible. This is ticketed for 2.2.0, so it's a priority to resolve.

bplommer:
The best I can think of is to weaken the constraint to `Defer` and `Bracket` by using `var` instead of `Ref` to store the finalisers, but I don't think that gains much?

core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -561,10 +561,19 @@ abstract private[effect] class ResourceInstances0 {
...
     new SemigroupK[Resource[F, *]] {
-      final override def combineK[A](a: Resource[F, A], b: Resource[F, A]): Resource[F, A] =
-        MonadError[Resource[F, *], E].handleErrorWith(a)(_ => b)
+
+      final override def combineK[A](ra: Resource[F, A], rb: Resource[F, A]): Resource[F, A] =
+        Resource
+          .make(Ref[F].of(Sync[F].unit))(_.get.flatten)
+          .evalMap { finalizers =>
+            def allocate(r: Resource[F, A]): F[A] =
+              r.allocated.flatMap { case (a, release) => finalizers.update(_.guarantee(release)).as(a) }
```
SystemFw: use `fold` here, `allocated.flatMap` is not safe

bplommer: Done - I had to make the implementation an instance method in order to access `fold`.

djspiewak:
I'm going to pull the trigger on this. :-) I seriously can't think of a way of doing it without `Ref`. That honestly bothers me quite a bit and makes me feel like we're missing something important, but I don't want to hold up important changes ticketed for 2.2 based on my uneasiness with the universe. :-) Thanks for staying on this!

SystemFw:
Btw, I've done the same changes in ce3 already. I've also spent some time thinking about these cases without Ref (including parZip), but there's something fundamental about the encoding of Resource that prevents it (the nature of Allocate).

djspiewak:
> Btw, I've done the same changes in ce3 already. I've also spent some time thinking about these cases without Ref (including parZip), but there's something fundamental about the encoding of Resource that prevents it (the nature of Allocate).

😢

bplommer:
> I seriously can't think of a way of doing it without `Ref`. That honestly bothers me quite a bit and makes me feel like we're missing something important

I feel like the problem is not so much with the use of `Ref` in itself as that it necessitates a `Sync` requirement. Maybe there should be typeclasses for instantiating Ref (and other concurrency primitives), so they can be used without exposing the ability to suspend arbitrary side effects.

Edit: I see CE3 already has this 🙂

PR #1061, opened 2020-08-07
Fix dubious iterations value for MapCallsBenchmark.one

vasilmkd:
This bit me when running benchmarks.


djspiewak:
Nice catch! This change should be mirrored into ce3 as well.

Issue #1027, opened 2020-07-31
Polling racePair


RaasAhsan:
`racePair` is inherently a leaky function since it returns a `Fiber` in its type signature. Because of that, we protect all invocations of `racePair` with `uncancelable` with the appropriate cancellations in place. For example, this is the definition of `raceOutcome`:
```scala
def raceOutcome[B](that: IO[B]): IO[Either[OutcomeIO[A @uncheckedVariance], OutcomeIO[B]]] =
  IO.uncancelable { _ =>
    racePair(that).flatMap {
      case Left((oc, f)) => f.cancel.as(Left(oc))
      case Right((f, oc)) => f.cancel.as(Right(oc))
    }
  }
```
Notice that in this definition (as well as `both`, `bothOutcome`, and `race`), `racePair` isn't wrapped in `poll`. It wasn't very obvious to me why we weren't doing this before, but why shouldn't we able to? If neither of the spawned fibers complete and we cancel, we should be able to cancel the entire operation. Additionally, `racePair` is internally defined in terms of `async` and registers a finalizer that cancels both fibers, but we don't really expose a library call that exercises that behavior.

Here's why we can't do it. Let's add a `poll` around `racePair` in the above definition:
```scala
def raceOutcome[B](that: IO[B]): IO[Either[OutcomeIO[A @uncheckedVariance], OutcomeIO[B]]] =
  IO.uncancelable { poll =>
    poll(racePair(that)).flatMap {
      case Left((oc, f)) => f.cancel.as(Left(oc))
      case Right((f, oc)) => f.cancel.as(Right(oc))
    }
  }
```
This looks like it should work, but this is in fact leaky as well. The problem lies in the operational semantics of `Async` and `Unmask` and how they interact. When `poll` is called, the argument is wrapped in an `Unmask` which drops the mask for the duration of the inner `IO`, allowing cancellation. After the inner `IO` completes, the mask is restored. On the other hand, when an `Async` completes, it pops its finalizer before re-entering the runloop. So if we happen to observe cancellation between the time the finalizer is popped and the time we restore the mask, we would end up leaking the `Fiber`.

The crux of the problem here is that the scope of unmasking by `poll` and the scope in which the async finalizer remains registered don't completely coincide, or at the very least, they don't share the same point of completion. In particular, the finalizer can be deregistered before we restore a mask.

I don't think this is a bug, but it certainly introduces some gaps where cancellation is conceptually appropriate but can't be finalized correctly.

[...]

PR #1065, opened 2020-08-08
Move the async cancelation check ahead of the finalizer pop

djspiewak:
This is related to #1027, and in fact closes the loophole noted by that issue. I didn't go through and modify `race`/`both` and friends to take advantage of this change though, and I'm leaving the issue open so we can discuss how this subtlety affects other datatypes and inductive implementations (via transformers).

It's possible that we simply define this to be the *correct* behavior for `IO`: namely, if the callback completes, you must immediately perform a cancelation check and use the `async` finalizer if it is available, and additionally that mask restoration in `poll` must be performed in an uncancelable continuation mechanism. That seems a little forcing of implementations though, and also very difficult to specify in laws.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -230,19 +230,11 @@ private final class IOFiber[A](
   3. Callback completes after cancelation and can't take over the runloop
   4. Callback completes after cancelation and after the finalizers have run, so it can take the runloop, but shouldn't
    */
-  private[this] def asyncContinue(
-      state: AtomicReference[AsyncState],
-      e: Either[Throwable, Any]): Unit = {
-    state.lazySet(AsyncStateInitial) // avoid leaks
-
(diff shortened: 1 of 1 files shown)


RaasAhsan:
Yeah, I think it makes sense that this is an implementation detail that `IO` is capable of offering. We can leave the `racePair` calls in the `Concurrent` implementation unpolled, and the instance for `IO` would just delegate.

RaasAhsan (approved):
Looks good to me! Would it be easy to write tests for all these cases w/ the ticked executor? I can take a look into doing that

djspiewak:
> Yeah, I think it makes sense that this is an implementation detail that IO is capable of offering. We can leave the racePair calls in the Concurrent implementation unpolled, and the instance for IO would just delegate.

The thing that concerns me a bit is this does mean that anyone using an inductive instance (such as `Kleisli`) will immediately end up in an unpolled case. I guess we could just override `race`/`both` etc in all of the inductive classes. That would probably resolve it.

PR #1066, opened 2020-08-09
Remove async done AtomicBoolean

vasilmkd:
Save an allocation on each `async` and be slightly kinder to the `objectState` stack.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -103,6 +103,7 @@ private final class IOFiber[A](
   private[this] val AsyncStateInitial = AsyncState.Initial
   private[this] val AsyncStateRegisteredNoFinalizer = AsyncState.RegisteredNoFinalizer
   private[this] val AsyncStateRegisteredWithFinalizer = AsyncState.RegisteredWithFinalizer
+  private[this] val AsyncStateFailure = AsyncState.Failure
 
   // mutable state for resuming the fiber in different states
   private[this] var resumeTag: Byte = ExecR
@@ -313,8 +314,6 @@ private final class IOFiber[A](
         case 4 =>
           val cur = cur0.asInstanceOf[Async[Any]]
 
-          val done = new AtomicBoolean()
-
           /*
            * The purpose of this buffer is to ensure that registration takes
            * primacy over callbacks in the event that registration produces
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -329,60 +328,53 @@ private final class IOFiber[A](
...
 
-              if (old != AsyncStateInitial) {
-                // registration already completed, we're good to go
-                loop()
-              }
+              // If we reach this point, it means that somebody else owns the run-loop
+              // and will handle cancellation.
             }
+            // println(s"<$name> callback run with $e")
+            val tag = state.getAndSet(AsyncState.Complete(e)).tag
```
RaasAhsan: If the async callback is invoked twice, wouldn't the second caller's result override the first? I think this might break the old semantics, particularly when they both complete before registration completes.

vasilmkd:
Slight overkill. But these changes can probably be upstreamed to Scala.js.

RaasAhsan:
If there's no other way around the `getAndAccumulate` shim, I personally prefer just rewriting the CAS loop, which I think would look something like this:
```scala
@tailrec
def stateLoop(): Unit = {
  val old = state.get()
  if (old.tag <= 2) {
    if (state.compareAndSet(old, AsyncState.Complete(e))) {
      if (old.tag == 1 || old.tag == 2) {
        loop(old.tag)
      }
    } else {
      stateLoop()
    }
  }
}
```

vasilmkd:
Benchmark results before this change:
```
Benchmark                         (size)   Mode  Cnt       Score      Error  Units
AsyncBenchmark.async                 100  thrpt   20   15797.058 ±  813.741  ops/s
AsyncBenchmark.bracket               100  thrpt   20   45266.004 ±  658.605  ops/s
AsyncBenchmark.cancelable            100  thrpt   20   16175.531 ±  447.670  ops/s
AsyncBenchmark.race                  100  thrpt   20    5096.527 ±   43.123  ops/s
AsyncBenchmark.racePair              100  thrpt   20    2254.237 ±   85.120  ops/s
AsyncBenchmark.start                 100  thrpt   20    3783.314 ±   26.295  ops/s
AsyncBenchmark.uncancelable          100  thrpt   20  213150.109 ± 5206.689  ops/s
AttemptBenchmark.errorRaised       10000  thrpt   20    1162.595 ±   13.677  ops/s
AttemptBenchmark.happyPath         10000  thrpt   20    1708.879 ±   12.740  ops/s
DeepBindBenchmark.async            10000  thrpt   20     265.326 ±    3.002  ops/s
DeepBindBenchmark.delay            10000  thrpt   20    3486.401 ±   11.317  ops/s
DeepBindBenchmark.pure             10000  thrpt   20    4337.455 ±    4.626  ops/s
HandleErrorBenchmark.errorRaised   10000  thrpt   20    1898.950 ±   19.212  ops/s
HandleErrorBenchmark.happyPath     10000  thrpt   20    2660.808 ±   12.576  ops/s
MapCallsBenchmark.batch120           N/A  thrpt   20    6344.165 ±    9.644  ops/s
MapCallsBenchmark.batch30            N/A  thrpt   20    3704.466 ±    3.058  ops/s
MapCallsBenchmark.one                N/A  thrpt   20     277.075 ±    1.398  ops/s
MapStreamBenchmark.batch120          N/A  thrpt   20    3563.282 ±   47.980  ops/s
MapStreamBenchmark.batch30           N/A  thrpt   20    1331.982 ±    2.729  ops/s
MapStreamBenchmark.one               N/A  thrpt   20    1391.153 ±   29.570  ops/s
ShallowBindBenchmark.async         10000  thrpt   20     258.141 ±   18.449  ops/s
ShallowBindBenchmark.delay         10000  thrpt   20    3963.032 ±    9.226  ops/s
ShallowBindBenchmark.pure          10000  thrpt   20    4026.195 ±  126.137  ops/s
```

Benchmark results after this change:
```
Benchmark                         (size)   Mode  Cnt       Score      Error  Units
AsyncBenchmark.async                 100  thrpt   20   18034.856 ±  103.538  ops/s
AsyncBenchmark.bracket               100  thrpt   20   48630.053 ±   89.389  ops/s
AsyncBenchmark.cancelable            100  thrpt   20   17812.308 ±  467.656  ops/s
AsyncBenchmark.race                  100  thrpt   20    6113.543 ±  468.343  ops/s
AsyncBenchmark.racePair              100  thrpt   20    2591.964 ±   68.527  ops/s
AsyncBenchmark.start                 100  thrpt   20    4795.263 ±  186.194  ops/s
AsyncBenchmark.uncancelable          100  thrpt   20  230044.214 ± 3263.898  ops/s
AttemptBenchmark.errorRaised       10000  thrpt   20    1254.625 ±   16.989  ops/s
AttemptBenchmark.happyPath         10000  thrpt   20    1827.529 ±   10.853  ops/s
DeepBindBenchmark.async            10000  thrpt   20     312.276 ±   15.670  ops/s
DeepBindBenchmark.delay            10000  thrpt   20    3666.530 ±    2.451  ops/s
DeepBindBenchmark.pure             10000  thrpt   20    4543.451 ±    3.598  ops/s
HandleErrorBenchmark.errorRaised   10000  thrpt   20    2040.571 ±   17.318  ops/s
HandleErrorBenchmark.happyPath     10000  thrpt   20    2846.865 ±   35.697  ops/s
MapCallsBenchmark.batch120           N/A  thrpt   20    5587.165 ±  256.708  ops/s
MapCallsBenchmark.batch30            N/A  thrpt   20    3806.722 ±   97.477  ops/s
MapCallsBenchmark.one                N/A  thrpt   20     280.575 ±    3.927  ops/s
MapStreamBenchmark.batch120          N/A  thrpt   20    3420.259 ±  129.198  ops/s
MapStreamBenchmark.batch30           N/A  thrpt   20    1377.132 ±   25.433  ops/s
MapStreamBenchmark.one               N/A  thrpt   20    1471.760 ±   65.889  ops/s
ShallowBindBenchmark.async         10000  thrpt   20     282.344 ±    3.866  ops/s
ShallowBindBenchmark.delay         10000  thrpt   20    4217.673 ±   18.566  ops/s
ShallowBindBenchmark.pure          10000  thrpt   20    5047.243 ±    7.341  ops/s
```

djspiewak:
This is a lovely change! I think we need to be a little careful though. We unset `state` later (on line 347, once the callback has fully resolved), which we do to avoid memory leaks. The problem with this is that we may have a repeated callback invocation long after the previous one has resolved, which I *think* would not be caught then. I think we probably need an additional `AsyncState` case which indicates that the callback has completed and doesn't hold any data, to avoid leaks, because setting back to `Initial` is problematic.

vasilmkd:
I should adapt `Failure` to `Done` and we should be alright.

vasilmkd:
@djspiewak Ready for another review.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -329,60 +328,69 @@ private final class IOFiber[A](
...
+             * set then `suspend()` should be right behind it *unless* we
+             * have been canceled. If we were canceled, then some other
+             * fiber is taking care of our finalizers and will have
+             * marked suspended as false, meaning that `canceled` will be set
+             * to true. Either way, we won't own the runloop.
+             */
+            @tailrec
+            def loop(old: Byte): Unit = {
+              if (resume()) {
+                state.lazySet(AsyncStateDone) // avoid leaks
```
djspiewak, line 347: As an aside, it would be worth writing a unit test for this. There's a law to this effect, but it doesn't comprehensively model the race conditions.

djspiewak (changes requested):
Just one quick test that exhibits the failure caught in review, but otherwise I think this is good to go.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -329,60 +328,69 @@ private final class IOFiber[A](
...
+            }
+            // println(s"<$name> callback run with $e")
+            val complete = AsyncState.Complete(e)
+
+            @tailrec
+            def stateLoop(): Unit = {
+              val old = state.get()
+              if (old.tag <= 2) {
+                if (state.compareAndSet(old, complete)) {
+                  val tag = old.tag
```
RaasAhsan: Is it worth moving this right after the declaration of `old`. just so there's only one heap access and the rest are stack accesses. it might get optimized but just to make sure

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -905,16 +912,16 @@ private final class IOFiber[A](
...
+    val old = state.getAndSet(AsyncState.Done)
+    if (!old.isInstanceOf[AsyncState.Complete]) {
       // if we get an error before the callback, then propagate
       failed(t, depth + 1)
     } else {
       // we got the error *after* the callback, but we have queueing semantics
       // so drop the results
 
-      state.lazySet(AsyncStateInitial) // avoid leaks
+      state.lazySet(AsyncStateDone) // avoid leaks
```
RaasAhsan: It looks like this is redundant since we're setting it right before the conditional.

core/shared/src/main/scala/cats/effect/IOFiber.scala
vasilmkd: Done.

core/shared/src/main/scala/cats/effect/IOFiber.scala

core/shared/src/main/scala/cats/effect/IOFiber.scala

PR #1069, opened 2020-08-09
Special case attempt in IO and SyncIO run loop

### LINKED ISSUE #1055: Port CE2 IO optimizations to CE3 IO
Cats Effect 2 IO introduced several optimizations that boosted performance of the runloop. Although it seems like Cats Effect 3 is already faster, we should consider whether any of those optimizations are worth porting. `redeem` and `attempt` requiring only one bind stack continuation is an important one that comes to mind for me.

vasilmkd:
Resolves #1055.

I checked the bytecode, we still keep the `tableswitch`. I have included the files as proof.
[iobytecode.txt](https://github.com/typelevel/cats-effect/files/5048251/iobytecode.txt)
[synciobytecode.txt](https://github.com/typelevel/cats-effect/files/5048252/synciobytecode.txt)

### DIFF
--- a/core/jvm/src/main/java/cats/effect/SyncIOConstants.java
+++ b/core/jvm/src/main/java/cats/effect/SyncIOConstants.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2020 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect;
+
+final class SyncIOConstants {
+
+    public static final int MaxStackDepth = 512;
+
+    public static final byte MapK = 0;
+    public static final byte FlatMapK = 1;
+    public static final byte HandleErrorWithK = 2;
+    public static final byte RunTerminusK = 3;
+    public static final byte AttemptK = 4;
+    public static final byte RedeemK = 5;
+    public static final byte RedeemWithK = 6;
+}
--- a/core/jvm/src/main/java/cats/effect/SyncIOConstants.java
+++ b/core/jvm/src/main/java/cats/effect/SyncIOConstants.java
@@ -25,6 +25,4 @@ final class SyncIOConstants {
   public static final byte HandleErrorWithK = 2;
   public static final byte RunTerminusK = 3;
   public static final byte AttemptK = 4;
-  public static final byte RedeemK = 5;
-  public static final byte RedeemWithK = 6;
 }
--- a/core/jvm/src/main/java/cats/effect/SyncIOConstants.java
+++ b/core/jvm/src/main/java/cats/effect/SyncIOConstants.java
@@ -18,13 +18,13 @@ package cats.effect;
 
 final class SyncIOConstants {
 
-    public static final int MaxStackDepth = 512;
+  public static final int MaxStackDepth = 512;
 
-    public static final byte MapK = 0;
-    public static final byte FlatMapK = 1;
-    public static final byte HandleErrorWithK = 2;
-    public static final byte RunTerminusK = 3;
-    public static final byte AttemptK = 4;
-    public static final byte RedeemK = 5;
-    public static final byte RedeemWithK = 6;
+  public static final byte MapK = 0;
+  public static final byte FlatMapK = 1;
+  public static final byte HandleErrorWithK = 2;
+  public static final byte RunTerminusK = 3;
+  public static final byte AttemptK = 4;
+  public static final byte RedeemK = 5;
+  public static final byte RedeemWithK = 6;
 }
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -42,7 +42,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
     map(_ => b)
 
   def attempt: IO[Either[Throwable, A]] =
-    map(Right(_)).handleErrorWith(t => IO.pure(Left(t)))
+    IO.Attempt(this)
 
   def bothOutcome[B](that: IO[B]): IO[(OutcomeIO[A @uncheckedVariance], OutcomeIO[B])] =
     IO.uncancelable { poll =>
@@ -269,6 +269,9 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
           val ioa = self.ioa.asInstanceOf[IO[A]]
           val poll = self.poll.asInstanceOf[Poll[F]]
           poll(ioa.to[F])
+
+        case self: IO.Attempt[A] =>
+          F.attempt(self.ioa.to[F]).asInstanceOf[F[A]]
       }
     }
 
@@ -587,7 +590,7 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
     def tag = 9
   }
 
-  private[effect] final case class OnCancel[A](ioa: IO[A], fin: IO[Unit]) extends IO[A] {
+  private[effect] final case class OnCancel[+A](ioa: IO[A], fin: IO[Unit]) extends IO[A] {
     def tag = 10
   }
 
@@ -620,8 +623,12 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
     def tag = 19
   }
 
+  private[effect] final case class Attempt[+A](ioa: IO[A]) extends IO[Either[Throwable, A]] {
+    def tag = 20
+  }
+
   // Not part of the run loop. Only used in the implementation of IO#to.
-  private[effect] final case class UnmaskTo[F[_], A](ioa: IO[A], poll: Poll[F]) extends IO[A] {
+  private[effect] final case class UnmaskTo[F[_], +A](ioa: IO[A], poll: F ~> F) extends IO[A] {
     def tag = -1
   }
 }
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -181,6 +181,9 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
     (FiberIO[A @uncheckedVariance], OutcomeIO[B])]] =
     IO.RacePair(this, that)
 
+  def redeem[B](recover: Throwable => B, map: A => B): IO[B] =
+    IO.Redeem(this, recover, map)
+
   def delayBy(duration: FiniteDuration): IO[A] =
     IO.sleep(duration) *> this
 
@@ -246,7 +249,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
         case self: IO.Start[_] =>
           F.start(self.ioa.to[F]).map(fiberFrom(_)).asInstanceOf[F[A]]
 
-        case self: IO.RacePair[a, b] =>
+        case self: IO.RacePair[_, _] =>
           val back = F.racePair(self.ioa.to[F], self.iob.to[F]) map { e =>
             e.bimap({ case (a, f) => (a, fiberFrom(f)) }, { case (f, b) => (fiberFrom(f), b) })
           }
@@ -264,14 +267,17 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
           // ioa.to[F] // polling should be handled by F
           sys.error("impossible")
 
+        case self: IO.Attempt[_] =>
+          F.attempt(self.ioa.to[F]).asInstanceOf[F[A]]
+
+        case IO.Redeem(ioa, recover, map) =>
+          F.redeem(ioa.to[F])(recover, map)
+
         case self: IO.UnmaskTo[_, _] =>
           // casts are safe because we only ever construct UnmaskF instances in this method
           val ioa = self.ioa.asInstanceOf[IO[A]]
           val poll = self.poll.asInstanceOf[Poll[F]]
           poll(ioa.to[F])
-
-        case self: IO.Attempt[A] =>
-          F.attempt(self.ioa.to[F]).asInstanceOf[F[A]]
       }
     }
 
@@ -539,6 +545,9 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
       IO.suspend(hint)(thunk)
 
     override def void[A](ioa: IO[A]): IO[Unit] = ioa.void
+
+    override def redeem[A, B](fa: IO[A])(recover: Throwable => B, f: A => B): IO[B] =
+      fa.redeem(recover, f)
   }
 
   implicit def effectForIO: Effect[IO] = _effectForIO
@@ -625,6 +634,14 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
   private[effect] final case class Attempt[+A](ioa: IO[A]) extends IO[Either[Throwable, A]] {
     def tag = 20
+  }
+
+  private[effect] final case class Redeem[A, +B](
+      ioa: IO[A],
+      recover: Throwable => B,
+      map: A => B)
+      extends IO[B] {
+    def tag = 21
   }
 
   // Not part of the run loop. Only used in the implementation of IO#to.
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -182,10 +182,10 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
     IO.RacePair(this, that)
 
   def redeem[B](recover: Throwable => B, map: A => B): IO[B] =
-    IO.Redeem(this, recover, map)
+    attempt.map(_.fold(recover, map))
 
   def redeemWith[B](recover: Throwable => IO[B], bind: A => IO[B]): IO[B] =
-    IO.RedeemWith(this, recover, bind)
+    attempt.flatMap(_.fold(recover, bind))
 
   def delayBy(duration: FiniteDuration): IO[A] =
     IO.sleep(duration) *> this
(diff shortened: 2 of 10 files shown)


djspiewak:
I think it's more important to check the timing. :-) The bytecode will remain tableswitch, but hotspot starts emitting less efficient assembly *for* `tableswitch` at some point.

vasilmkd:
No worries, I got it.

RaasAhsan:
I'm certain there will be a performance boost for the special cases, but the question is by how much. And does it hurt performance for the other combinators? I'd like to balance both of those with the growing sizes of the run-loop and the surface area for bugs. I'm reminded by how `guaranteeCase` is special cased in CE2 but exhibits different behavior than if it were implemented in terms of `bracketCase`.

vasilmkd:
Before:
```
Benchmark                         (size)   Mode  Cnt       Score      Error  Units
AsyncBenchmark.async                 100  thrpt   20   17306.762 ±  129.089  ops/s
AsyncBenchmark.bracket               100  thrpt   20   45306.318 ±  335.720  ops/s
AsyncBenchmark.cancelable            100  thrpt   20   17082.783 ±  385.100  ops/s
AsyncBenchmark.race                  100  thrpt   20    5786.621 ±  184.385  ops/s
AsyncBenchmark.racePair              100  thrpt   20    2471.910 ±   26.048  ops/s
AsyncBenchmark.start                 100  thrpt   20    4156.749 ±  459.760  ops/s
AsyncBenchmark.uncancelable          100  thrpt   20  212709.165 ± 3573.030  ops/s
AttemptBenchmark.errorRaised       10000  thrpt   20    1132.792 ±   20.359  ops/s
AttemptBenchmark.happyPath         10000  thrpt   20    1548.067 ±   58.161  ops/s
DeepBindBenchmark.async            10000  thrpt   20     301.051 ±    9.440  ops/s
DeepBindBenchmark.delay            10000  thrpt   20    3538.446 ±   19.084  ops/s
DeepBindBenchmark.pure             10000  thrpt   20    4284.279 ±   13.700  ops/s
HandleErrorBenchmark.errorRaised   10000  thrpt   20    1945.265 ±    1.895  ops/s
HandleErrorBenchmark.happyPath     10000  thrpt   20    2647.295 ±   18.309  ops/s
MapCallsBenchmark.batch120           N/A  thrpt   20    5546.021 ±  132.684  ops/s
MapCallsBenchmark.batch30            N/A  thrpt   20    3629.625 ±    8.017  ops/s
MapCallsBenchmark.one                N/A  thrpt   20     271.221 ±    1.839  ops/s
MapStreamBenchmark.batch120          N/A  thrpt   20    3764.633 ±   69.979  ops/s
MapStreamBenchmark.batch30           N/A  thrpt   20    1376.310 ±   55.688  ops/s
MapStreamBenchmark.one               N/A  thrpt   20    1483.375 ±   65.414  ops/s
ShallowBindBenchmark.async         10000  thrpt   20     246.260 ±    8.463  ops/s
ShallowBindBenchmark.delay         10000  thrpt   20    3997.139 ±    5.801  ops/s
ShallowBindBenchmark.pure          10000  thrpt   20    4784.682 ±    3.613  ops/s
```

djspiewak:
> I'm certain there will be a performance boost for the special cases, but the question is by how much. And does it hurt performance for the other combinators? I'd like to balance both of those with the growing sizes of the run-loop and the surface area for bugs. I'm reminded by how guaranteeCase is special cased in CE2 but exhibits different behavior than if it were implemented in terms of bracketCase.

Apparently I completely misread the HotSpot information, and the 17 threshold (which is now 8 in Java 1.8) is the threshold at which the *more* efficient assembly is generated. Below that threshold, it's a chain of conditional jumps; *above* the threshold, it's an array table. So in other words, the only thing we really need to worry about are cache hits on the indirection, which implies that the soft threshold (in terms of number of IO cases) is probably closer to 32 or 64.

None of this invalidates your second point about surface area. I think we really need to ask what cases are *worthwhile* to specialize in terms of runloop complexity. `attempt` seems worth it to me, since it's basically everywhere, but I've never actually seen `redeemWith` in the wild. Maybe that's just selection bias. I think it's at least worth examining the question though. Remember that a lot of these trivially derived combinators (like `redeemWith`) don't actually have law-based tests, meaning that we won't necessarily catch bugs without adding dedicated tests of our own.

vasilmkd:
For full up to date information:
https://github.com/openjdk/jdk/search?q=MinJumpTableSize&unscoped_q=MinJumpTableSize

This constant seems to be architecture specific.

vasilmkd:
I saw `redeem` and `redeemWith` specialized in CE2, that's why I included them.

Avasil:
> attempt seems worth it to me, since it's basically everywhere, but I've never actually seen redeemWith in the wild. Maybe that's just selection bias.

I've seen the opposite and `attempt` is rarely used alone - usually, it is followed by `map` or `flatMap` which is `redeem`. Using `transform(With)` is a good practice with `Future` too. And `redeem` is basically a `fold`.

If I were to choose, I'd implement `redeem` "natively" and use it in `attempt` implementation.

vasilmkd:
I agree, the `Throwable => Either[Throwable, Any]` and `Any => Either[Throwable, Any]` (`Left(_)` and `Right(_)`) can even be preallocated.

vasilmkd:
Before:
```
Benchmark                         (size)   Mode  Cnt       Score     Error  Units
AsyncBenchmark.async                 100  thrpt   20   16503.231 ± 366.925  ops/s
AsyncBenchmark.bracket               100  thrpt   20   45154.842 ± 370.696  ops/s
AsyncBenchmark.cancelable            100  thrpt   20   16542.374 ± 138.751  ops/s
AsyncBenchmark.race                  100  thrpt   20    5787.100 ± 192.888  ops/s
AsyncBenchmark.racePair              100  thrpt   20    2507.217 ±  33.028  ops/s
AsyncBenchmark.start                 100  thrpt   20    4236.338 ± 149.943  ops/s
AsyncBenchmark.uncancelable          100  thrpt   20  208284.574 ± 757.105  ops/s
AttemptBenchmark.errorRaised       10000  thrpt   20    1174.648 ±  16.118  ops/s
AttemptBenchmark.happyPath         10000  thrpt   20    1526.720 ±  65.244  ops/s
DeepBindBenchmark.async            10000  thrpt   20     264.309 ±   8.924  ops/s
DeepBindBenchmark.delay            10000  thrpt   20    3468.331 ±  14.560  ops/s
DeepBindBenchmark.pure             10000  thrpt   20    4343.592 ±   3.552  ops/s
HandleErrorBenchmark.errorRaised   10000  thrpt   20    1935.791 ±  27.261  ops/s
HandleErrorBenchmark.happyPath     10000  thrpt   20    2657.671 ±   7.120  ops/s
MapCallsBenchmark.batch120           N/A  thrpt   20    5985.075 ± 296.754  ops/s
MapCallsBenchmark.batch30            N/A  thrpt   20    3639.179 ±  14.137  ops/s
MapCallsBenchmark.one                N/A  thrpt   20     267.597 ±   4.481  ops/s
MapStreamBenchmark.batch120          N/A  thrpt   20    3741.862 ±  72.565  ops/s
MapStreamBenchmark.batch30           N/A  thrpt   20    1312.778 ±  17.716  ops/s
MapStreamBenchmark.one               N/A  thrpt   20    1437.523 ± 106.803  ops/s
ShallowBindBenchmark.async         10000  thrpt   20     252.081 ±   3.131  ops/s
ShallowBindBenchmark.delay         10000  thrpt   20    3981.937 ±   6.540  ops/s
ShallowBindBenchmark.pure          10000  thrpt   20    4738.045 ±  17.041  ops/s
```

After:
```
Benchmark                         (size)   Mode  Cnt       Score      Error  Units
AsyncBenchmark.async                 100  thrpt   20   16304.788 ±  445.369  ops/s
AsyncBenchmark.bracket               100  thrpt   20   45694.368 ±  159.485  ops/s
AsyncBenchmark.cancelable            100  thrpt   20   17185.895 ±  324.805  ops/s
AsyncBenchmark.race                  100  thrpt   20    5379.973 ±   45.180  ops/s
AsyncBenchmark.racePair              100  thrpt   20    2556.924 ±   68.430  ops/s
AsyncBenchmark.start                 100  thrpt   20    4222.219 ±  128.135  ops/s
AsyncBenchmark.uncancelable          100  thrpt   20  214506.321 ± 5815.579  ops/s
AttemptBenchmark.errorRaised       10000  thrpt   20    1637.429 ±   29.173  ops/s
AttemptBenchmark.happyPath         10000  thrpt   20    2386.265 ±   25.565  ops/s
DeepBindBenchmark.async            10000  thrpt   20     270.342 ±    5.995  ops/s
DeepBindBenchmark.delay            10000  thrpt   20    3673.515 ±    4.190  ops/s
DeepBindBenchmark.pure             10000  thrpt   20    4360.824 ±   64.771  ops/s
HandleErrorBenchmark.errorRaised   10000  thrpt   20    1930.234 ±   14.583  ops/s
HandleErrorBenchmark.happyPath     10000  thrpt   20    2626.928 ±   12.626  ops/s
MapCallsBenchmark.batch120           N/A  thrpt   20    5485.312 ±   78.199  ops/s
MapCallsBenchmark.batch30            N/A  thrpt   20    3936.995 ±   87.144  ops/s
MapCallsBenchmark.one                N/A  thrpt   20     268.833 ±    5.027  ops/s
MapStreamBenchmark.batch120          N/A  thrpt   20    3558.816 ±   86.411  ops/s
MapStreamBenchmark.batch30           N/A  thrpt   20    1292.747 ±    5.716  ops/s
MapStreamBenchmark.one               N/A  thrpt   20    1404.340 ±   34.666  ops/s
ShallowBindBenchmark.async         10000  thrpt   20     253.237 ±    2.951  ops/s
ShallowBindBenchmark.delay         10000  thrpt   20    3974.016 ±    6.087  ops/s
ShallowBindBenchmark.pure          10000  thrpt   20    4772.066 ±    5.039  ops/s
```

core/jvm/src/main/java/cats/effect/SyncIOConstants.java
```diff
@@ -0,0 +1,30 @@
...
+    public static final int MaxStackDepth = 512;
+
+    public static final byte MapK = 0;
+    public static final byte FlatMapK = 1;
+    public static final byte HandleErrorWithK = 2;
+    public static final byte RunTerminusK = 3;
+    public static final byte AttemptK = 4;
+    public static final byte RedeemK = 5;
+    public static final byte RedeemWithK = 6;
+}
```
djspiewak: Minor nit: the other Java source file is formatted with two-space indentation and we may as well be consistent. :-)

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -547,6 +559,9 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
       IO.suspend(hint)(thunk)
 
     override def void[A](ioa: IO[A]): IO[Unit] = ioa.void
+
+    override def redeem[A, B](fa: IO[A])(recover: Throwable => B, f: A => B): IO[B] =
```
djspiewak, line 549: Shouldn't we also override `redeemWith`?

core/shared/src/test/scala/cats/effect/IOSpec.scala
```diff
@@ -89,14 +89,61 @@ class IOSpec extends IOPlatformSpecification with Discipline with ScalaCheck wit
...
+      IO.raiseError[Unit](TestException)
+        .redeemWith(_ => throw ThrownException, _ => IO.pure(42))
+        .attempt must completeAs(Left(ThrownException))
+    }
+
+    "redeemWith catches exceptions thrown in bind function" in ticked { implicit ticker =>
+      case object ThrownException extends RuntimeException
+      IO.unit.redeem(_ => IO.pure(41), _ => throw ThrownException).attempt must completeAs(
+        Left(ThrownException))
+    }
```
djspiewak, line 165: I think it would be worth adding a scalacheck test. Something like:

```scala
in ticked { implicit ticker =>
  forAll { (ioa: IO[Int], f: Int => IO[Int], e: Throwable => IO[Int]) =>
    ioa.attempt.flatMap(_.bitraverse(f, e)) === ioa.redeemWith(f, e)
  }
}
```

And maybe something similar for `redeem` and `attempt`. Just because I don't trust unit tests on this sort of thing. :-D

core/shared/src/test/scala/cats/effect/SyncIOSpec.scala
```diff
@@ -72,6 +72,58 @@ class SyncIOSpec extends IOPlatformSpecification with Discipline with ScalaCheck
...
+        .attempt must completeAsSync(Left(ThrownException))
+    }
+
+    "redeemWith catches exceptions thrown in bind function" in {
+      case object ThrownException extends RuntimeException
+      SyncIO
+        .unit
+        .redeem(_ => SyncIO.pure(41), _ => throw ThrownException)
+        .attempt must completeAsSync(Left(ThrownException))
+    }
```
djspiewak, line 157: Similarly property tests here.

testkit/shared/src/main/scala/cats/effect/testkit/Generators.scala
```diff
@@ -119,7 +119,9 @@ trait ApplicativeErrorGenerators[F[_], E] extends ApplicativeGenerators[F] {
   override protected def recursiveGen[A](
       deeper: GenK[F])(implicit AA: Arbitrary[A], AC: Cogen[A]): List[(String, Gen[F[A]])] =
     List(
-      "handleErrorWith" -> genHandleErrorWith[A](deeper)(AA, AC)
+      "handleErrorWith" -> genHandleErrorWith[A](deeper)(AA, AC),
+      "attempt" -> genAttempt[Int, A](deeper),
+      "redeem" -> genRedeem[Int, A](deeper)
```
djspiewak: I don't think it's worth adding these to the abstract generators. Not every implementation will override them the way that we do. If we really want these, we can add them for `IO`/`SyncIO` explicitly in `Runners`

testkit/shared/src/main/scala/cats/effect/testkit/Generators.scala
```diff
@@ -130,12 +132,39 @@ trait ApplicativeErrorGenerators[F[_], E] extends ApplicativeGenerators[F] {
...
+      "redeemWith" -> genRedeemWith[Int, A](deeper)
+    ) ++ super.recursiveGen(deeper)(AA, AC)
+
+  private def genRedeemWith[A: Arbitrary: Cogen, B: Arbitrary: Cogen](
+      deeper: GenK[F]): Gen[F[B]] =
+    for {
+      fa <- deeper[A]
+      recover <- Gen.function1[E, F[B]](deeper[B])
+      bind <- Gen.function1[A, F[B]](deeper[B])
+    } yield F.redeemWith(fa)(recover, bind)
```
djspiewak: Similarly here: not worth adding these.

djspiewak (changes requested):
Okay I'm provisionally alright with overriding these. :-) Particularly since we know we aren't going to fall off a cliff in terms of tableswitch evaluation performance. The runloop complexity is something we really do need to be mindful of, but these are important functions.

core/jvm/src/main/java/cats/effect/SyncIOConstants.java
vasilmkd: Sorry, this was auto formatted.

core/shared/src/main/scala/cats/effect/IO.scala
vasilmkd: Good catch.

core/shared/src/test/scala/cats/effect/IOSpec.scala
vasilmkd: Done.

testkit/shared/src/main/scala/cats/effect/testkit/Generators.scala
vasilmkd: Removed.

core/shared/src/test/scala/cats/effect/SyncIOSpec.scala
vasilmkd: Done.

testkit/shared/src/main/scala/cats/effect/testkit/Generators.scala
vasilmkd: Removed.

djspiewak (approved):
Looks good to me! Any thoughts, @RaasAhsan? I'm still very convincible on the runloop complexity point. :-)

vasilmkd:
What is the most complex thing about the runloop in your opinion? I'd like to tackle it. Not in this PR obviously.

djspiewak:
> What is the most complex thing about the runloop in your opinion? I'd like to tackle it. Not in this PR obviously.

I actually think the runloop is quite nice now, but it's still a comparatively difficult thing to see through in terms of the encoding. Adding and maintaining functionality in the runloop is pretty objectively harder than adding and maintaining it in "purity land" at a higher level, which is a long-winded way of saying that every time we add a primitive algebraic case to `IO`, we increase the complexity of the runloop and we should be careful around that.

Particularly since your recent changes, I feel pretty good about the way the runloop works. There is *always* room for improvement, but I guess what I'm saying is that I don't feel anything is particularly acute right now. :-) The "runloop complexity point" was mostly just a general first-principles concern stemming from increased surface area of functionality.

RaasAhsan:
I'm still not entirely convinced that we need all 3 of these to be primitive. `redeemWith` is the least common denominator for `redeem` and `attempt`, so some benchmarks where the latter two are implemented in terms of `redeemWith` would shine some light on where the performance boost is coming from. `attempt` does see gains, but I have a feeling most of it is coming from its encoding with one continuation rather than two, but `redeem` and `redeemWith` enjoy the same benefit. There aren't any benchmarks for those yet so I can't really tell how impactful it is.

On runloop complexity, I completely agree that it's in a great spot right now, especially with all the refactoring and cleanup that's been done, and we should try to keep it that way. Every new constructor adds code to the runloop, and it being the impure beast that it is, we have to mentally track how every block of code interacts with another. I'd like to leverage FP principles as much as possible unless there's a solid justification for not doing so.

vasilmkd:
I'll add the benchmarks.

djspiewak:
> I'd like to leverage FP principles as much as possible unless there's a solid justification for not doing so.

There's also an element of dogfooding here: if the performance isn't good enough for us to allow derived operations to be implemented *in terms of* other operations, then shouldn't we correct that problem? Users don't have the option of making things primitive when the performance isn't good enough, so the performance *should be* good enough that derived combinators are acceptably fast.

Definitely seeing some comparative benchmarks would be helpful!

vasilmkd:
They'll be in a separate PR though, I need to first implement the combinators on IO. I'll probably split this one. Working on it.

vasilmkd:
Please take a look at the benchmark results in #1075. I hope they are clear enough to make a decision on how to continue here.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -587,7 +603,7 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
     def tag = 9
   }
 
-  private[effect] final case class OnCancel[A](ioa: IO[A], fin: IO[Unit]) extends IO[A] {
+  private[effect] final case class OnCancel[+A](ioa: IO[A], fin: IO[Unit]) extends IO[A] {
```
djspiewak, line 606: Oh lol

djspiewak (approved):

PR #1074, opened 2020-08-11
Add Poll

### LINKED ISSUE #1064: Introduce Poll as a nominal type for uncancelable's param
It's very difficult to look up scaladoc for `F ~> F`. We would be a lot better off with something like this in kernel:

```scala
trait Poll[F[_]] extends (F ~> F)
```

This gives us a place for documentation.

vasilmkd:
Resolves #1064.


kernel/shared/src/main/scala/cats/effect/kernel/concurrent.scala
```diff
@@ -32,14 +32,16 @@ trait Fiber[F[_], E, A] {
     joinAndEmbed(F.canceled *> F.never)
 }
 
+trait Poll[F[_]] extends (F ~> F)
```
djspiewak: Can you move this to its own file? Arguably `Fiber` should be moved as well, and then **concurrent.scala** can be capitalized to **Concurrent.scala**.

djspiewak (changes requested):
Very nice! Minor request but otherwise good to go

kernel/shared/src/main/scala/cats/effect/kernel/concurrent.scala
vasilmkd: Finally! 😆

kernel/shared/src/main/scala/cats/effect/kernel/concurrent.scala
```diff
@@ -21,19 +21,6 @@ import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
...
-  def joinAndEmbed(onCancel: F[A])(implicit F: Concurrent[F, E]): F[A] =
-    join.flatMap(_.fold(onCancel, F.raiseError(_), fa => fa))
-
-  def joinAndEmbedNever(implicit F: Concurrent[F, E]): F[A] =
-    joinAndEmbed(F.canceled *> F.never)
-}
-
-trait Poll[F[_]] extends (F ~> F)
-
 trait Concurrent[F[_], E] extends MonadError[F, E] {
```
djspiewak, line 24: Super-minor thing, but would you mind renaming this file to upper-case?

vasilmkd: I swear I did. Git failed me. Fixing.

Edit: Nevermind, Mac OS failed me.

vasilmkd: `git mv` to the rescue.

djspiewak: Yeah Mac and Windows are terrible for that kind of thing :-/

PR #1075, opened 2020-08-11
Add CE3 redeem benchmarks

vasilmkd:
Opening for posterity. Will post baseline scores, without any special cases.

CE3 master:
```
Benchmark                        (size)   Mode  Cnt     Score    Error  Units
AttemptBenchmark.errorRaised      10000  thrpt   20  1176.175 ± 19.742  ops/s
AttemptBenchmark.happyPath        10000  thrpt   20  1533.906 ± 31.811  ops/s
RedeemBenchmark.errorRaised       10000  thrpt   20  1005.665 ± 17.223  ops/s
RedeemBenchmark.happyPath         10000  thrpt   20  1403.133 ± 20.596  ops/s
RedeemWithBenchmark.errorRaised   10000  thrpt   20  1112.423 ±  3.353  ops/s
RedeemWithBenchmark.happyPath     10000  thrpt   20  1460.951 ±  2.383  ops/s
```

CE3 with `redeemWith` specialized in the run loop and `redeem` and `attempt` implemented in terms of it:
```
Benchmark                        (size)   Mode  Cnt     Score    Error  Units
AttemptBenchmark.errorRaised      10000  thrpt   20  1398.482 ± 45.226  ops/s
AttemptBenchmark.happyPath        10000  thrpt   20  1715.366 ± 13.471  ops/s
RedeemBenchmark.errorRaised       10000  thrpt   20  1263.061 ±  1.180  ops/s
RedeemBenchmark.happyPath         10000  thrpt   20  1664.181 ±  1.777  ops/s
RedeemWithBenchmark.errorRaised   10000  thrpt   20  2198.608 ± 11.373  ops/s
RedeemWithBenchmark.happyPath     10000  thrpt   20  3638.704 ± 56.893  ops/s
```

Further specializing `attempt`, `redeemWith` remains identical, `redeem` is still implemented in terms of `redeemWith`:
```
Benchmark                        (size)   Mode  Cnt     Score    Error  Units
AttemptBenchmark.errorRaised      10000  thrpt   20  1608.404 ± 33.204  ops/s
AttemptBenchmark.happyPath        10000  thrpt   20  2293.181 ± 91.073  ops/s
RedeemBenchmark.errorRaised       10000  thrpt   20  1324.089 ±  8.704  ops/s
RedeemBenchmark.happyPath         10000  thrpt   20  1648.077 ±  6.450  ops/s
RedeemWithBenchmark.errorRaised   10000  thrpt   20  2230.361 ± 28.927  ops/s
RedeemWithBenchmark.happyPath     10000  thrpt   20  3815.468 ± 68.948  ops/s
```

And finally, the super specialized version from #1069, where `attempt`, `redeem` and `redeemWith` are all part of the runloop:
```
Benchmark                        (size)   Mode  Cnt     Score    Error  Units
AttemptBenchmark.errorRaised      10000  thrpt   20  1651.755 ± 22.416  ops/s
AttemptBenchmark.happyPath        10000  thrpt   20  2385.165 ± 23.389  ops/s
RedeemBenchmark.errorRaised       10000  thrpt   20  1489.859 ±  4.968  ops/s
RedeemBenchmark.happyPath         10000  thrpt   20  1734.115 ± 24.102  ops/s
RedeemWithBenchmark.errorRaised   10000  thrpt   20  2261.896 ±  8.064  ops/s
RedeemWithBenchmark.happyPath     10000  thrpt   20  3631.001 ± 46.834  ops/s
```

IMO, the sweetspot is: specialized `attempt` and specialized `redeemWith`, with `redeem` being implemented in terms of `redeemWith`.

### DIFF
--- a/benchmarks/src/main/scala/cats/effect/benchmarks/RedeemBenchmark.scala
+++ b/benchmarks/src/main/scala/cats/effect/benchmarks/RedeemBenchmark.scala
@@ -18,7 +18,6 @@ package cats.effect.benchmarks
 
 import cats.effect.IO
 import cats.effect.unsafe.implicits.global
-import cats.implicits._
 
 import org.openjdk.jmh.annotations._
 
--- a/benchmarks/src/main/scala/cats/effect/benchmarks/RedeemBenchmark.scala
+++ b/benchmarks/src/main/scala/cats/effect/benchmarks/RedeemBenchmark.scala
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2020 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.benchmarks
+
+import cats.effect.IO
+import cats.effect.unsafe.implicits.global
+import cats.implicits._
+
+import org.openjdk.jmh.annotations._
+
+import java.util.concurrent.TimeUnit
+
+/**
+ * To do comparative benchmarks between versions:
+ *
+ *     benchmarks/run-benchmark RedeemBenchmark
+ *
+ * This will generate results in `benchmarks/results`.
+ *
+ * Or to run the benchmark from within sbt:
+ *
+ *     jmh:run -i 10 -wi 10 -f 2 -t 1 cats.effect.benchmarks.RedeemBenchmark
+ *
+ * Which means "10 iterations", "10 warm-up iterations", "2 forks", "1 thread".
+ * Please note that benchmarks should be usually executed at least in
+ * 10 iterations (as a rule of thumb), but more is better.
+ */
+@State(Scope.Thread)
+@BenchmarkMode(Array(Mode.Throughput))
+@OutputTimeUnit(TimeUnit.SECONDS)
+class RedeemBenchmark {
+
+  @Param(Array("10000"))
+  var size: Int = _
+
+  @Benchmark
+  def happyPath(): Int = {
+    val id = identity[Int] _
+
+    def loop(i: Int): IO[Int] =
+      if (i < size) IO.pure(i + 1).redeem(_ => 0, id).flatMap(loop)
+      else IO.pure(i)
+
+    loop(0).unsafeRunSync()
+  }
+
+  @Benchmark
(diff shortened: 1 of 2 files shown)


djspiewak:
I'd like to see the `AttemptBenchmark`, but it looks like `redeemWith` is the big win here! A little over a 2x improvement. `redeem` doesn't seem all that compelling, particularly since we can implement it with `redeemWith` and get about the same boost. Curious about `attempt`.

vasilmkd:
Attempt is running at the moment.

djspiewak:
Still pondering this, but… for what it's worth, I searched my entire source code subdirectory (which includes not just proprietary code but also the entire contents of http4s, fs2, cats, and many others). The *only* references to `redeem` or `redeemWith` were in cats-effect (in the `IO` definitions) and cats-core (syntax, typeclasses, and such). I'm really not convinced it's that common of a combinator. The fact that `redeem` doesn't show that much of a difference (~20%) between the version that's on **series/3.x** today and one which has a specialized `redeemWith` seems to indicate that while specializing `redeemWith` has a significant impact on `redeemWith` itself, its impact is muted when we look at things derived *from* `redeemWith`.

For the record, I tried the same search with `attempt`, and while the results are harder to gauge exactly because it's a common enough word to find in comments and other things, the result set wasn't even *close* to the same as `redeemWith`/`redeem`. I found hundreds of call sites across diverse codebases. It really seems like `attempt` is the more commonly-used combinator, even though it is technically slower than `redeemWith`.

So what I'm leaning towards is just specializing `attempt`, since the effect of specializing `redeemWith` seems limited and the complexity tradeoff is notable.

Avasil:
And how often `attempt` was used in combination with `map` or `flatMap`?
Or how often do you see `flatMap` + `errorHandleWith` ?

I think `attempt` is more well known but in many cases, it's a better practice to use `redeemWith` instead

vasilmkd:
Except that `redeemWith` is not exactly a substitute for `flatMap` + `handleErrorWith`.

Avasil:
Good point, it's only the case if `flatMap` doesn't raise errors

vasilmkd:
I agree in principle @Avasil , I too was/still am in favor of specializing `redeemWith` as well, but @djspiewak kind of convinced me that it might not be as used as I initially thought.

vasilmkd:
Awesome write up @djspiewak, thank you. I have incorporated the changes in #1069.

djspiewak:
> I think attempt is more well known but in many cases, it's a better practice to use redeemWith instead

I agree with this, but I was mostly measuring what people *do* rather than what they should do. I wish `redeemWith` were a little more widely used. I definitely wouldn't preclude adding back the optimization in `IO` if so! :-)

vasilmkd:
Do we want the benchmarks from this PR or can this be closed?

djspiewak:
> Do we want the benchmarks from this PR or can this be closed?

I think it's worth pulling them in. No reason *not* to measure performance on this stuff!

vasilmkd:
@djspiewak At your convenience. Dropped all changes, the PR now contains only the benchmarks.

PR #1077, opened 2020-08-12
Enhanced exceptions

RaasAhsan:
With asynchronous stack tracing implemented, we have the ability to make exceptions caught by the IO runloop much more useful and relevant than they were before. This PR introduces *enhanced exceptions*, which is a feature that augments stack traces of caught exceptions with information from an accumulated fiber trace. The feature is controlled by a system property `cats.effect.enhancedExceptions`, which is enabled by default and also requires asynchronous stack tracing to be enabled.

Stack traces before:
```
java.lang.Throwable: A runtime exception has occurred
	at org.simpleapp.examples.Main$.b(Main.scala:28)
	at org.simpleapp.examples.Main$.a(Main.scala:25)
	at org.simpleapp.examples.Main$.$anonfun$foo$11(Main.scala:37)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)
	at cats.effect.internals.IORunLoop$.cats$effect$internals$IORunLoop$$loop(IORunLoop.scala:103)
	at cats.effect.internals.IORunLoop$RestartCallback.signal(IORunLoop.scala:440)
	at cats.effect.internals.IORunLoop$RestartCallback.apply(IORunLoop.scala:461)
	at cats.effect.internals.IORunLoop$RestartCallback.apply(IORunLoop.scala:399)
	at cats.effect.internals.IOShift$Tick.run(IOShift.scala:36)
	at cats.effect.internals.PoolUtils$$anon$2$$anon$3.run(PoolUtils.scala:52)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

Stack traces after:
```
java.lang.Throwable: A runtime exception has occurred
	at org.simpleapp.examples.Main$.b(Main.scala:28)
	at org.simpleapp.examples.Main$.a(Main.scala:25)
	at org.simpleapp.examples.Main$.$anonfun$foo$11(Main.scala:37)
	at map @ org.simpleapp.examples.Main$.$anonfun$foo$10(Main.scala:37)
	at flatMap @ org.simpleapp.examples.Main$.$anonfun$foo$8(Main.scala:36)
	at flatMap @ org.simpleapp.examples.Main$.$anonfun$foo$6(Main.scala:35)
	at flatMap @ org.simpleapp.examples.Main$.$anonfun$foo$4(Main.scala:34)
	at flatMap @ org.simpleapp.examples.Main$.$anonfun$foo$2(Main.scala:33)
	at flatMap @ org.simpleapp.examples.Main$.foo(Main.scala:32)
	at flatMap @ org.simpleapp.examples.Main$.program(Main.scala:42)
	at as @ org.simpleapp.examples.Main$.run(Main.scala:48)
	at main$ @ org.simpleapp.examples.Main$.main(Main.scala:22)
```

The original idea for this feature called for a `TracedException` wrapper, but there were quite a few problems with that approach, one of which was deciding where to wrap exceptions. 

Relevant Gitter discussion: https://gitter.im/typelevel/cats-effect-dev?at=5f31e511e20413052e7c556e

TODO:
- [x] All my review comments
- [x] Benchmarks
- [ ] Should we sanitize the augmented stack traces like we do for `printFiberTrace`?

### DIFF
--- a/core/jvm/src/main/java/cats/effect/internals/TracingPlatform.java
+++ b/core/jvm/src/main/java/cats/effect/internals/TracingPlatform.java
@@ -36,7 +36,7 @@ public final class TracingPlatform {
      */
     private static final String stackTracingMode = Optional.ofNullable(System.getProperty("cats.effect.stackTracingMode"))
             .filter(x -> !x.isEmpty())
-            .orElse("cached");
+            .orElse("full");
 
     public static final boolean isCachedStackTracing = stackTracingMode.equalsIgnoreCase("cached");
 
@@ -59,5 +59,15 @@ public final class TracingPlatform {
             }
         })
         .orElse(16);
+
+    /**
+     * Sets the contextual exceptions flag, which controls whether or not the
+     * stack traces of IO exceptions are augmented to include async stack trace information.
+     * Stack tracing must be enabled in order to use this feature.
+     * This flag is enabled by default.
+     */
+    public static final boolean contextualExceptions = Optional.ofNullable(System.getProperty("cats.effect.contextualExceptions"))
+            .map(x -> Boolean.valueOf(x))
+            .orElse(true);
 
 }
--- a/core/jvm/src/main/java/cats/effect/internals/TracingPlatform.java
+++ b/core/jvm/src/main/java/cats/effect/internals/TracingPlatform.java
@@ -36,7 +36,7 @@ public final class TracingPlatform {
      */
     private static final String stackTracingMode = Optional.ofNullable(System.getProperty("cats.effect.stackTracingMode"))
             .filter(x -> !x.isEmpty())
-            .orElse("full");
+            .orElse("cached");
 
     public static final boolean isCachedStackTracing = stackTracingMode.equalsIgnoreCase("cached");
 
--- a/core/jvm/src/main/java/cats/effect/internals/TracingPlatform.java
+++ b/core/jvm/src/main/java/cats/effect/internals/TracingPlatform.java
@@ -58,6 +58,6 @@ public final class TracingPlatform {
                 return Optional.empty();
             }
         })
-        .orElse(128);
+        .orElse(16);
 
 }
--- a/core/jvm/src/main/java/cats/effect/internals/TracingPlatform.java
+++ b/core/jvm/src/main/java/cats/effect/internals/TracingPlatform.java
@@ -61,12 +61,12 @@ public final class TracingPlatform {
         .orElse(16);
 
     /**
-     * Sets the contextual exceptions flag, which controls whether or not the
+     * Sets the enhanced exceptions flag, which controls whether or not the
      * stack traces of IO exceptions are augmented to include async stack trace information.
      * Stack tracing must be enabled in order to use this feature.
      * This flag is enabled by default.
      */
-    public static final boolean contextualExceptions = Optional.ofNullable(System.getProperty("cats.effect.contextualExceptions"))
+    public static final boolean enhancedExceptions = Optional.ofNullable(System.getProperty("cats.effect.enhancedExceptions"))
             .map(x -> Boolean.valueOf(x))
             .orElse(true);
 
--- a/core/shared/src/main/scala/cats/effect/internals/IORunLoop.scala
+++ b/core/shared/src/main/scala/cats/effect/internals/IORunLoop.scala
@@ -19,7 +19,7 @@ package cats.effect.internals
 import cats.effect.IO
 import cats.effect.IO.{Async, Bind, ContextSwitch, Delay, Map, Pure, RaiseError, Suspend, Trace}
 import cats.effect.tracing.{IOEvent, IOTrace}
-import cats.effect.internals.TracingPlatform.{contextualExceptions, isStackTracing}
+import cats.effect.internals.TracingPlatform.{enhancedExceptions, isStackTracing}
 
 import scala.util.control.NonFatal
 
@@ -114,7 +114,7 @@ private[effect] object IORunLoop {
             catch { case NonFatal(ex) => RaiseError(ex) }
 
         case RaiseError(ex) =>
-          if (isStackTracing && contextualExceptions) {
+          if (isStackTracing && enhancedExceptions) {
             augmentException(ex, ctx)
           }
           findErrorHandler(bFirst, bRest) match {
@@ -247,7 +247,7 @@ private[effect] object IORunLoop {
             } catch { case NonFatal(ex) => RaiseError(ex) }
 
         case RaiseError(ex) =>
-          if (isStackTracing && contextualExceptions) {
+          if (isStackTracing && enhancedExceptions) {
             augmentException(ex, ctx)
           }
           findErrorHandler(bFirst, bRest) match {
@@ -371,21 +371,34 @@ private[effect] object IORunLoop {
   private def augmentException(ex: Throwable, ctx: IOContext): Unit = {
     val stackTrace = ex.getStackTrace
     if (!stackTrace.isEmpty) {
-      val prefix = IOTrace.dropRunLoopSuffix(stackTrace.toList)
-      val suffix = ctx
-        .getStackTraces()
-        // TODO: We should just store a List[StackTraceElement] in the cache
-        .flatMap(t => IOTrace.getOpAndCallSite(t.getStackTrace.toList))
-        .map {
-          case (methodSite, callSite) =>
-            new StackTraceElement(methodSite.getMethodName + " @ " + callSite.getClassName,
-                                  callSite.getMethodName,
-                                  callSite.getFileName,
-                                  callSite.getLineNumber)
-        }
-      ex.setStackTrace((prefix ++ suffix).toArray)
+      val augmented = stackTrace(stackTrace.length - 1) eq augmentationMarker
+      if (!augmented) {
+        val prefix = dropRunLoopFrames(stackTrace)
+        val suffix = ctx
+          .getStackTraces()
+          .flatMap(t => IOTrace.getOpAndCallSite(t.stackTrace))
+          .map {
+            case (methodSite, callSite) =>
+              new StackTraceElement(methodSite.getMethodName + " @ " + callSite.getClassName,
+                                    callSite.getMethodName,
+                                    callSite.getFileName,
+                                    callSite.getLineNumber)
+          }
+          .toArray
+        ex.setStackTrace(prefix ++ suffix ++ Array(augmentationMarker))
+      }
     }
   }
+
+  private def dropRunLoopFrames(frames: Array[StackTraceElement]): Array[StackTraceElement] =
+    frames.takeWhile(ste => !runLoopFilter.exists(ste.getClassName.startsWith(_)))
+
+  private[this] val runLoopFilter = List(
+    "cats.effect.",
+    "scala."
+  )
+
+  private[this] val augmentationMarker = new StackTraceElement("", "", "", 0)
 
   /**
    * A `RestartCallback` gets created only once, per [[startCancelable]]
--- a/core/shared/src/main/scala/cats/effect/internals/IORunLoop.scala
+++ b/core/shared/src/main/scala/cats/effect/internals/IORunLoop.scala
@@ -114,6 +114,35 @@ private[effect] object IORunLoop {
             catch { case NonFatal(ex) => RaiseError(ex) }
 
         case RaiseError(ex) =>
+          def getOpAndCallSite(frames: List[StackTraceElement]): Option[(StackTraceElement, StackTraceElement)] = {
+            val stackTraceFilter = List(
+              "cats.effect.",
+              "cats.",
+              "sbt.",
+              "java.",
+              "sun.",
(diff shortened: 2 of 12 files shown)


core/jvm/src/main/java/cats/effect/internals/TracingPlatform.java
```diff
@@ -58,6 +58,16 @@
...
-        .orElse(128);
+        .orElse(16);
+
+    /**
+     * Sets the contextual exceptions flag, which controls whether or not the
+     * stack traces of IO exceptions are augmented to include async stack trace information.
+     * Stack tracing must be enabled in order to use this feature.
+     * This flag is enabled by default.
+     */
+    public static final boolean contextualExceptions = Optional.ofNullable(System.getProperty("cats.effect.contextualExceptions"))
```
RaasAhsan: I'm down to calling this something else if there's a better name.

core/shared/src/main/scala/cats/effect/internals/IORunLoop.scala
```diff
@@ -357,6 +363,30 @@ private[effect] object IORunLoop {
...
+   * function will rewrite the stack trace of a captured exception
+   * to include the async stack trace.
+   */
+  private def augmentException(ex: Throwable, ctx: IOContext): Unit = {
+    val stackTrace = ex.getStackTrace
+    if (!stackTrace.isEmpty) {
+      val prefix = IOTrace.dropRunLoopSuffix(stackTrace.toList)
+      val suffix = ctx
+        .getStackTraces()
+        // TODO: We should just store a List[StackTraceElement] in the cache
```
RaasAhsan: I'm going to make a commit for this later. The idea is that `getStackTrace` calls `clone`, so we can avoid that.

core/shared/src/main/scala/cats/effect/internals/RingBuffer.scala
```diff
@@ -46,10 +46,11 @@ final private[internals] class RingBuffer[A <: AnyRef](size: Int) {
   def capacity: Int =
     length
 
+  // returns a list in reverse order of insertion
   def toList: List[A] = {
-    val end = index
-    val start = Math.max(end - length, 0)
-    (start until end).toList
+    val start = index - 1
```
RaasAhsan, line 51: Reversing the order of the elements that we get back from the ring buffer. This is going to be the default because generally traces are rendered in order of decreasing age.

core/shared/src/main/scala/cats/effect/tracing/IOTrace.scala
```diff
@@ -121,9 +109,33 @@ private[effect] object IOTrace {
         case (_, callSite) => !stackTraceFilter.exists(callSite.getClassName.startsWith(_))
       }
 
+  def dropRunLoopSuffix(frames: List[StackTraceElement]): List[StackTraceElement] =
```
RaasAhsan: `IOTrace` companion object seems like a weird place to put this

core/shared/src/main/scala/cats/effect/tracing/IOTrace.scala
```diff
@@ -33,11 +33,11 @@ final case class IOTrace(events: List[IOEvent], captured: Int, omitted: Int) {
     val Junction = "├"
     val Line = "│"
 
+    val acc0 = s"IOTrace: $captured frames captured\n"
```
RaasAhsan, line 36: I'm going to reorganize this code and come up with better names.

core/shared/src/main/scala/cats/effect/internals/RingBuffer.scala
RaasAhsan: This is going to break some tests that I haven't fixed yet

core/shared/src/main/scala/cats/effect/tracing/IOTrace.scala
```diff
@@ -62,56 +62,44 @@ final case class IOTrace(events: List[IOEvent], captured: Int, omitted: Int) {
...
                   .getOrElse("(...)")
               }
             }
             s" $junc $message"
         }
-        .mkString(acc0, "\n", "\n")
+        .mkString(acc0, "\n", "")
+
+      val acc2 = if (omitted > 0) {
+        acc1 + "\n " + TurnRight + s" ... ($omitted frames omitted)"
```
RaasAhsan, line 92: The number of omitted frames (if any) will now be rendered at the bottom of a trace, to keep consistency with Java stack traces

core/shared/src/main/scala/cats/effect/internals/IORunLoop.scala
```diff
@@ -114,6 +114,9 @@ private[effect] object IORunLoop {
             catch { case NonFatal(ex) => RaiseError(ex) }
 
         case RaiseError(ex) =>
+          if (isStackTracing && contextualExceptions) {
```
RaasAhsan: I'm still thinking about if this is the correct place to do this.

joroKr21:
Brilliant 👍

joroKr21:
How would this interact with `NoStackTrace`?

RaasAhsan:
@joroKr21 if `NoStackTrace` is mixed into a given `Throwable`, `t.getStackTrace` will be empty, so nothing will be added based on the current behavior. If it's more appropriate, we can change the implementation so the fiber trace that has been accumulated so far is included in the stack trace. That's worth discussing.

core/shared/src/main/scala/cats/effect/tracing/IOTrace.scala
djspiewak: Maybe it *should* be in `IORunLoop`, since it's technically part of the runloop's implementation details?

core/shared/src/main/scala/cats/effect/internals/IORunLoop.scala
djspiewak: I keep going back and forth. On one hand, doing it here means that every exception, even those explicitly `raiseError`ed and later `attempt`ed by users, gets decorated. This means that if users are doing something like `handleErrorWith(log.error(_))`, they'll see the augmented trace. This seems *very* valuable!

But conversely, this also means that we pay the penalty of rendering the trace every time we call `raiseError`, which is not a small performance hit I would assume. Maybe worth it? Probably? I'm not 100% sure.

RaasAhsan: I think we also have to be careful because augmented exceptions can be re-caught or re-lifted into the runloop. For example, `IO.raiseError(ex).attempt.rethrow`, so I think we might need to check if it's already been augmented. I'm hoping the benchmarks will paint a better picture, but worst case it's something users can disable.

core/shared/src/main/scala/cats/effect/internals/IORunLoop.scala
```diff
@@ -357,6 +363,43 @@ private[effect] object IORunLoop {
...
 
+  /**
+   * If stack tracing and contextual exceptions are enabled, this
+   * function will rewrite the stack trace of a captured exception
+   * to include the async stack trace.
+   */
+  private def augmentException(ex: Throwable, ctx: IOContext): Unit = {
+    val stackTrace = ex.getStackTrace
+    if (!stackTrace.isEmpty) {
+      val augmented = stackTrace(stackTrace.length - 1) eq augmentationMarker
```
RaasAhsan, line 374: So the idea here is that a reference equality is probably a lot faster than substring searches. After we augment an exception, we append the augmentation marker which can be used for comparison later. Notice that because it is declared `private`, callers can't access it and therefore they can't accidentally insert a marker.

The consequence of this is that printed stack traces show the marker at the end. I verified that we can't insert `null` as a marker.
```scala
java.lang.Throwable: hello world!
	at org.simpleapp.examples.Main$.b(Main.scala:29)
	at org.simpleapp.examples.Main$.a(Main.scala:26)
	at org.simpleapp.examples.Main$.$anonfun$foo$11(Main.scala:39)
	at main$ @ org.simpleapp.examples.Main$.main(Main.scala:23)
	at map @ org.simpleapp.examples.Main$.$anonfun$foo$10(Main.scala:39)
	at flatMap @ org.simpleapp.examples.Main$.$anonfun$foo$8(Main.scala:37)
	at flatMap @ org.simpleapp.examples.Main$.$anonfun$foo$6(Main.scala:36)
	at flatMap @ org.simpleapp.examples.Main$.$anonfun$foo$4(Main.scala:35)
	at flatMap @ org.simpleapp.examples.Main$.$anonfun$foo$2(Main.scala:34)
	at flatMap @ org.simpleapp.examples.Main$.foo(Main.scala:33)
	at flatMap @ org.simpleapp.examples.Main$.program(Main.scala:44)
	at as @ org.simpleapp.examples.Main$.run(Main.scala:50)
	at main$ @ org.simpleapp.examples.Main$.main(Main.scala:23)
	at .(:0)
```

We're also doing a lot of `Array` and `List` manipulation here. Without thinking about it more, I'm not sure what the most efficient approach is (do everything in `Array` or `List`, and when should we convert back to an array?).

I'll benchmark this tomorrow so we have a concrete answer.

core/shared/src/main/scala/cats/effect/internals/IORunLoop.scala
```diff
@@ -357,6 +363,43 @@ private[effect] object IORunLoop {
         }
     }
 
+  /**
+   * If stack tracing and contextual exceptions are enabled, this
+   * function will rewrite the stack trace of a captured exception
+   * to include the async stack trace.
+   */
+  private def augmentException(ex: Throwable, ctx: IOContext): Unit = {
+    val stackTrace = ex.getStackTrace
```
RaasAhsan, line 372: We will throw a NPE here if `ex` is null. I think that would break existing code that happen to rely on this behavior

core/shared/src/main/scala/cats/effect/internals/IORunLoop.scala
RaasAhsan: I just realized that users *can* access an instance of the market, through the stack trace itself! I don't think it'll be a problem either way because they would have to be deliberately manipulating stack traces for something to mess up. Same deal with checking ` @ ` probably.

djspiewak: > So the idea here is that a reference equality is probably a lot faster than substring searches. After we augment an exception, we append the augmentation marker which can be used for comparison later. Notice that because it is declared private, callers can't access it and therefore they can't accidentally insert a marker.

Hmm, I think this is worth benchmarking. In my experience, `indexOf` is *really* really fast, so I don't think there'll be that much of a difference, but I could be mistaken!

core/shared/src/main/scala/cats/effect/internals/IORunLoop.scala
djspiewak: Do people actually run `raiseError(null)`? I think I'm pretty comfortable telling those people that they just have to turn off augmented exceptions, because they don't deserve nice things. :-P

RaasAhsan:
Here is a benchmark comparison between the reference equality check, the `indexOf` check, cached tracing in master, and disabled tracing for reference.
```
indexOf
[info] Benchmark                     (size)   Mode  Cnt     Score    Error  Units
[info] AttemptBenchmark.errorRaised   10000  thrpt    5  1182.529 ± 14.167  ops/s
[info] AttemptBenchmark.happyPath     10000  thrpt    5  2222.477 ± 20.280  ops/s

reference equality
[info] Benchmark                     (size)   Mode  Cnt     Score    Error  Units
[info] AttemptBenchmark.errorRaised   10000  thrpt    5  1222.009 ±  4.819  ops/s
[info] AttemptBenchmark.happyPath     10000  thrpt    5  2165.725 ± 16.240  ops/s

cached tracing pre-enhanced exceptions
AttemptBenchmark.errorRaised | 1403.077
AttemptBenchmark.happyPath | 2120.155

disabled tracing
[info] Benchmark                     (size)   Mode  Cnt     Score    Error  Units
[info] AttemptBenchmark.errorRaised   10000  thrpt    5  2032.301 ± 17.058  ops/s
[info] AttemptBenchmark.happyPath     10000  thrpt    5  2313.851 ± 91.837  ops/s
```

The reference equality check is slightly faster, but not by any significant margin.  Your call @djspiewak ! :D Also, I'm positive we can improve the performance of the `Array` and `List` operations more, but I'm not sure whether we should consider it a blocker for this PR.

djspiewak:
> The reference equality check is slightly faster, but not by any significant margin. Your call @djspiewak ! :D Also, I'm positive we can improve the performance of the Array and List operations more, but I'm not sure whether we should consider it a blocker for this PR.

:-) Let's go with `indexOf` since it doesn't affect the user-visible output. I definitely think we can optimize this further, though it's worth keeping in mind that the `errorRaised` benchmark is a bit artificial. Let's land this and refine in follow-ups!

PR #1078, opened 2020-08-12
Resolve directory-package inconsistency in cats.effect.kernel.syntax

nikiforo:
Syntax objects are placed in cats-effects/kernel/shared/src/main/scala/cats/effect/kernel/syntax/ directory.

The path should corresponds to `cats.effect.kernel.syntax` package, but instead `package cats.effect.syntax` is used.


djspiewak (changes requested):
I quite like this change! And I do think this is better, but I think we should add a forwarder in the `cats.effect` package object, so that people can still import `cats.effect.syntax` and `cats.effect.implicits` and such. Also I believe that `instances` needs the same treatment!

djspiewak (changes requested):
Closer! I think this still isn't quite right, sorry. :-(

So what I *think* we should have is the following:

- [x] All the syntax and instances should be defined under the `cats.effect.kernel` package
- [ ] `cats.effect.kernel.implicits`, `cats.effect.kernel.instances`, and `cats.effect.kernel.syntax` should all exist in the **kernel** project
- [ ] `cats.effect.implicits`, `cats.effect.instances`, and `cats.effect.syntax` should *also* exist, but in the **core** package. They should be defined as objects extending the appropriate `kernel` things, which gives us the binary-compatible flexibility to add **core**-specific syntax/instances/etc down the line

Sorry for being unclear!

nikiforo:
Wow! Thanks a lot for such a thorough explanation!

Do we still need a `package object syntax` under `cats.effect.kernel` package? It's used in `SyntaxSpec` so I assume that it should be preserved.

djspiewak:
> Do we still need a package object syntax under cats.effect.kernel package? It's used in SyntaxSpec so I assume that it should be preserved.

Yep! I believe we do still need that. We probably also need a similar one in the `cats.effect` package.

core/shared/src/main/scala/cats/effect/instances/ConcurrentInstances.scala
```diff
@@ -0,0 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+package instances
+
+trait ConcurrentInstances extends kernel.instances.ConcurrentInstances
```
djspiewak: I don't think we actually need `ConcurrentInstances` to be mirrored here, though that's an interesting thought. We might add *other* instances, but basically the `all` and `concurrent` in `cats.effect.instances` can just directly refer to `kernel.instances.ConcurrentInstances`

core/shared/src/main/scala/cats/effect/instances/AllInstances.scala
```diff
@@ -15,5 +15,6 @@
  */
 
 package cats.effect
+package instances
 
-object implicits extends syntax.AllSyntax with instances.AllInstances
+trait AllInstances extends ConcurrentInstances
```
djspiewak: I think this can just directly extend `kernel.instances.ConcurrentInstances`, since we won't be adding new instances for `Concurrent` specifically in **core**

core/shared/src/main/scala/cats/effect/syntax/AllSyntax.scala
```diff
@@ -0,0 +1,25 @@
...
+
+package cats.effect
+package syntax
+
+trait AllSyntax
+    extends ConcurrentSyntax
+    with TemporalSyntax
+    with AsyncSyntax
+    with SyncEffectSyntax
+    with EffectSyntax
```
djspiewak: I think these should just extend the `kernel.syntax.*` traits, since we aren't adding new (e.g.) `AsyncSyntax` that is specific to **core**, but we might add new syntax for other things.

core/shared/src/main/scala/cats/effect/syntax/AsyncSyntax.scala
```diff
@@ -0,0 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+package syntax
+
+trait AsyncSyntax extends kernel.syntax.AsyncSyntax
```
djspiewak: Much like `ConcurrentInstances`, I don't think this is necessary here.

core/shared/src/main/scala/cats/effect/syntax/package.scala
```diff
@@ -0,0 +1,28 @@
...
+
+package object syntax {
+
+  object all extends AllSyntax
+
+  object concurrent extends ConcurrentSyntax
+  object temporal extends TemporalSyntax
+  object async extends AsyncSyntax
+  object syncEffect extends SyncEffectSyntax
+  object effect extends EffectSyntax
```
djspiewak: These should all just extend `kernel.syntax.*`

djspiewak (changes requested):
Almost there! There are some unnecessary identity forwarder traits in **core** now. I think removing those and having the objects directly inherit from the **kernel** versions is the right thing to do.

Basically what I'm thinking here is that it's very unlikely that **core** would want to introduce new syntax or instances for (say) `Temporal`. However, it's very possible that **core** may want to introduce new syntax or instances for **core**-specific things (like `IO`), and we need everything that's in **kernel** to be *forwarded* along to **core**, so that means we need the objects to inherit from the **kernel** traits, but we don't need **core**-specific variants of the **kernel** traits. Does that make some sense?

Sorry I'm doing a bad job of explaining this. This is a really great change! Really great. :-)

nikiforo:
Great thanks for the patience :-)

djspiewak (approved):
Outstanding! Thank you!

PR #1080, opened 2020-08-13
add clock extension method for creating java.time.Instant

jbwheatley:
I mainly end up converting clock's realTime to `Instant` and so define this extension method in my projects, so thought it might be useful to add it here if other people deal with `Instant` a lot

### DIFF
--- a/core/js/src/main/scala/cats/effect/internals/ClockPlatform.scala
+++ b/core/js/src/main/scala/cats/effect/internals/ClockPlatform.scala
@@ -1,0 +1,19 @@
+/*
+ * Copyright (c) 2017-2019 The Typelevel Cats-effect Project Developers
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


djspiewak:
I quite like this, but there's a little bit of a problem, which is that ScalaJS doesn't have `java.time` at all. So if anyone *calls* this function in cross-compiled code, not realizing the problems with it, the resulting JS sources will simply not link. In theory this might show up in their build though, so perhaps that isn't a problem.

Anyone else have strong opinions on this? I'm moderately 👍

vasilmkd:
Can it be jvm specific?

djspiewak:
> Can it be jvm specific?

It could be if we added a `ClockPlatform` trait. I'm usually a little reticent to do stuff like that, though CE3 does a fair amount of it, so… *shrug* :-) I would accept that.

jbwheatley:
Cool, I can add the `ClockPlatform` trait 👍

djspiewak (approved):
I quite like this. I'm going to leave it open for comments for a longer just to see what people say. Thank you!

bplommer:
I like this. Always wanted it but assumed it was decided not to add it because `Instant` is jvm-only.

PR #1081, opened 2020-08-13
CE 3 Resource

SystemFw:
Not really as polished as I'd like but as you've noticed I've been struggling to find proper time. Anyway the bulk of it is there, hopefully we can merge and polish in a couple of subsequent PR's. Should mostly unblock downstream.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/ResourcePlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/ResourcePlatform.scala
@@ -16,10 +16,11 @@
 
 package cats.effect
 
-
 import javax.security.auth.Destroyable
 
-/** JVM-specific Resource methods */
+/**
+ * JVM-specific Resource methods
+ */
 private[effect] trait ResourcePlatform {
 
   /**
@@ -43,6 +44,7 @@ private[effect] trait ResourcePlatform {
    * @tparam A the type of the destroyable resource
    * @return a Resource that will automatically destroy after use
    */
-  def fromDestroyable[F[_], A <: Destroyable](acquire: F[A])(implicit F: Sync[F]): Resource[F, A] =
+  def fromDestroyable[F[_], A <: Destroyable](acquire: F[A])(
+      implicit F: Sync[F]): Resource[F, A] =
     Resource.make(acquire)(destroyable => F.delay(destroyable.destroy()))
 }
--- a/core/jvm/src/main/scala/cats/effect/ResourcePlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/ResourcePlatform.scala
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2020 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+
+
+import javax.security.auth.Destroyable
+
+/** JVM-specific Resource methods */
+private[effect] trait ResourcePlatform {
+
+  /**
+   * Creates a [[Resource]] by wrapping a Java
+   * [[https://docs.oracle.com/javase/8/docs/api/javax/security/auth/Destroyable.html Destroyable]].
+   *
+   * Example:
+   * {{{
+   *   import java.security.KeyStore.PasswordProtection
+   *   import cats.effect._
+   *   import cats.implicits._
+   *
+   *   def passwordProtection[F[_]](getPassword: F[Array[Char]])(implicit F: Sync[F]): Resource[F, PasswordProtection] =
+   *     Resource.fromDestroyable(
+   *       getPassword.map(new PasswordProtection(_))
+   *     )
+   * }}}
+   * @param acquire The effect with the resource to acquire.
+   * @param F the effect type in which the resource was acquired and will be released
+   * @tparam F the type of the effect
+   * @tparam A the type of the destroyable resource
+   * @return a Resource that will automatically destroy after use
+   */
+  def fromDestroyable[F[_], A <: Destroyable](acquire: F[A])(implicit F: Sync[F]): Resource[F, A] =
+    Resource.make(acquire)(destroyable => F.delay(destroyable.destroy()))
+}
--- a/core/shared/src/main/scala/cats/effect/Resource.scala
+++ b/core/shared/src/main/scala/cats/effect/Resource.scala
@@ -678,45 +678,47 @@ object ExitCase {
 //     }
 // }
 
-// abstract private[effect] class ResourceMonadError[F[_], E] extends ResourceMonad[F] with MonadError[Resource[F, *], E] {
-//   import Resource.{Allocate, Bind, Suspend}
+abstract private[effect] class ResourceMonadError[F[_], E] extends ResourceMonad[F] with MonadError[Resource[F, *], E] {
+  import Resource.{Allocate, Bind, Suspend}
 
-//   implicit protected def F: MonadError[F, E]
+  implicit protected def F: MonadError[F, E]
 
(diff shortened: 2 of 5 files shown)


RaasAhsan:
> Anyway the bulk of it is there, hopefully we can merge and polish in a couple of subsequent PR's

I've been trying to push people to create issues lately, just so we don't lose track of that work, would be great if you can find time to do that!

SystemFw:
Will do yeah

core/jvm/src/main/scala/cats/effect/ResourcePlatform.scala
```diff
@@ -0,0 +1,48 @@
...
+   *     )
+   * }}}
+   * @param acquire The effect with the resource to acquire.
+   * @param F the effect type in which the resource was acquired and will be released
+   * @tparam F the type of the effect
+   * @tparam A the type of the destroyable resource
+   * @return a Resource that will automatically destroy after use
+   */
+  def fromDestroyable[F[_], A <: Destroyable](acquire: F[A])(implicit F: Sync[F]): Resource[F, A] =
+    Resource.make(acquire)(destroyable => F.delay(destroyable.destroy()))
```
djspiewak: 👍

core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -0,0 +1,811 @@
...
+   *     })
+   * }}}
+   * @param acquire The effect with the resource to acquire.
+   * @param F the effect type in which the resource was acquired and will be released
+   * @tparam F the type of the effect
+   * @tparam A the type of the autocloseable resource
+   * @return a Resource that will automatically close after use
+   */
+  def fromAutoCloseable[F[_], A <: AutoCloseable](acquire: F[A])(implicit F: Sync[F]): Resource[F, A] =
+    Resource.make(acquire)(autoCloseable => F.delay(autoCloseable.close()))
```
djspiewak: I take it that `AutoCloseable` is available on ScalaJS but `Destroyable` isn't? Also I think it might be worth straight-up using `F.blocking` here, since it's very unlikely to have a Java thing that's `AutoCloseable` that *isn't* blocking.

core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -0,0 +1,811 @@
...
+ * get canceled.
+ *
+ * The types of exit signals are:
+ *
+ *  - [[ExitCase$.Completed Completed]]: for successful completion
+ *  - [[ExitCase$.Error Error]]: for termination in failure
+ *  - [[ExitCase$.Canceled Canceled]]: for abortion
+ */
+sealed trait ExitCase extends Product with Serializable
+object ExitCase {
```
djspiewak: 😢 I still want to find a way to avoid this…

core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -0,0 +1,811 @@
...
+  // def fromAutoCloseableBlocking[F[_]: Sync: ContextShift, A <: AutoCloseable](
+  //   blocker: Blocker
+  // )(acquire: F[A]): Resource[F, A] =
+  //   Resource.make(blocker.blockOn(acquire))(autoCloseable => blocker.delay(autoCloseable.close()))
+
+  /**
+   * `Resource` data constructor that wraps an effect allocating a resource,
+   * along with its finalizers.
+   */
+  final case class Allocate[F[_], A](resource: F[(A, ExitCase => F[Unit])])
```
djspiewak, line 463: Can we make the algebraic cases `private`?

core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -0,0 +1,811 @@
...
+   * when it's at its `use` phase.
+   *
+   * Thus [[Bracket]] allows you to observe interruption conditions
+   * and act on them.
+   */
+  case object Canceled extends ExitCase
+}
+
+  @annotation.implicitNotFound("Cannot find an instance for Resource.Bracket. This normally means you need to add implicit evidence of Concurrent[F, Throwable]")
+  trait Bracket[F[_]] extends MonadError[F, Throwable] {
```
djspiewak: I'm really afraid people are going to start using this as a migration shim from `Bracket`.

core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -0,0 +1,811 @@
...
+    new ResourceMonoid[F, A] {
+      def A = A0
+      def F = F0
+    }
+
+  // implicit def catsEffectLiftIOForResource[F[_]](implicit F00: LiftIO[F], F10: Applicative[F]): LiftIO[Resource[F, *]] =
+  //   new ResourceLiftIO[F] {
+  //     def F0 = F00
+  //     def F1 = F10
+  //   }
```
djspiewak, line 631: `LiftIO` is in **core**, so we can technically have this.

core/shared/src/test/scala/cats/effect/ResourceSpec.scala
```diff
@@ -0,0 +1,466 @@
...
+  //   implicit val ticker = Ticker(TestContext())
+  //
+  //   checkAll(
+  //     "Resource[IO, *]",
+  //     SemigroupKTests[Resource[IO, *]].semigroupK[Int]
+  //   )
+  // }
+
+
+  // TODO fails some laws
```
djspiewak, line 436: `Par` is pretty much untestable, since `racePair` is not equal to itself.

djspiewak:
I pushed formatting fixes, but this still doesn't compile on Dotty for me. Some of the errors look suspiciously like the (legitimate!) soundness issues I ran into when trying to have a covariant `Outcome`. I haven't looked more closely yet though.

djspiewak:
Okay nvm. Apparently it's fine on Dotty, it just needed a little massaging. Ran into a few interesting bugs. Also, somewhat distressingly, the `G[*]` hack which is needed on Scala 2 to make things compile in the presence of covariant polymorphic constructors (e.g. `F[+_]`) *breaks everything* on Dotty. So that's great. We'll have to find a way around that.

core/shared/src/main/scala/cats/effect/Resource.scala
SystemFw: Yeah this was one of the TODOs, I will push a fix

core/shared/src/main/scala/cats/effect/Resource.scala
SystemFw: they are interpreted in fs2

core/shared/src/main/scala/cats/effect/Resource.scala
SystemFw: that's quite possible, but otoh I think 99%+ of the cases assume IO there, so they can use Concurrent. I was hoping the message would nudge them there

core/shared/src/main/scala/cats/effect/Resource.scala
SystemFw: yep, another TODO to fix then

core/shared/src/test/scala/cats/effect/ResourceSpec.scala
SystemFw: Should we remove the tests or the instance?

djspiewak: Can't believe I'm saying this, but the tests.

The problem is operational, not fundamental. Our definition of equality simply doesn't correctly capture `race`'s semantics, and any definition which *does* is undecidable. So… I'm at a bit of a loss as to how to do law testing on this stuff. You'll notice that we also cheat in `PureConc` and `IO` by removing `racePair` from the generators, for exactly the same reason.

core/shared/src/main/scala/cats/effect/Resource.scala
djspiewak: Why do you have to say such horrible things to me?

mpilquist:
Eager merging!

SystemFw:
Good call, I'll just open a followup with the fixes highlighted in the comments

core/shared/src/main/scala/cats/effect/Resource.scala
SystemFw: Btw, this was by design when resource was created. The fs2 interpretation is nothing crazy, but it's basically because Resource is the Free encoding of the "region"  abstraction (that we need for fs2 bracket)

djspiewak:
Don't forget to open issues! There are loads of people who would love to jump in and help, they just need a little guidance as to where it's needed.

SystemFw:
If I can't open another PR today (which I was in the middle of before the merge), I'll open them

PR #1084, opened 2020-08-14
Set up Docusaurus

RaasAhsan:
This PR is a proof-of-concept for trying out Docusaurus on the `series/2.x` branch. It's not a goal of this PR to restructure or rewrite the documentation site, so the general layout and content is retained from the old site.

Here is an image that illustrates what one of the pages looks like:
Screen Shot 2020-07-20 at 10 15 24 AM


RaasAhsan:
It's also worth looking at Docusaurus 2 and seeing if that serves our needs better.

RaasAhsan:
Most of this is complete on the `docs` branch and #1140. Remaining work is in #1142.

PR #1086, opened 2020-08-14
Async instances

TimWSpence:


TimWSpence:
This is currently blocked on deciding what to do about the following `Async` laws:
```
  def asyncRightIsSequencedPure[A](a: A, fu: F[Unit]) =
    F.async[A](k => F.delay(k(Right(a))) >> fu.as(None)) <-> (fu >> F.pure(a))

  def asyncLeftIsSequencedRaiseError[A](e: Throwable, fu: F[Unit]) =
    F.async[A](k => F.delay(k(Left(e))) >> fu.as(None)) <-> (fu >> F.raiseError(e))
```
The problem is that some monad transformers have an extra error/state channel and we have to suppress that on the LHS.
eg if `fu == OptionT.None` then we end up suppressing that error on the LHS whereas on the RHS we would never run `F.pure(a)` or `F.raiseError(e)`

TimWSpence:
One option is to replace the RHS with:
```
fu !> F.pure(a)
```
and
```
fu !> F.raiseError(e)
```
so that the laws match the current behaviour. This does mean that `async` will suppress errors and throw away the side channel (error/state) however, which is possibly not what users expect. Can we modify `async` to allow us to preserve this instead?

djspiewak:
Blocked on #1165. Once that lands, this can be landed.

djspiewak (approved):

Issue #1092, opened 2020-08-17
Add Bifunctor instance for Outcome


djspiewak:
It's kind of bizarre that I didn't do this already, since `Outcome` is trivially a bifunctor, but here we are. It should be part of the `monadError`/`applicativeError` instances (e.g. `MonadError with Bifunctor`) to avoid coherence issues.

vasilmkd:
This issue has been resolved by #1094.

PR #1094, opened 2020-08-18
Outcome[F,E,A] is a bifunctor

nikiforo:


kernel/shared/src/main/scala/cats/effect/kernel/Outcome.scala
```diff
@@ -129,14 +130,12 @@ object Outcome extends LowPriorityImplicits {
   implicit def monadError[F[_], E](
       implicit F: Monad[F],
       FT: Traverse[F]): MonadError[Outcome[F, E, *], E] =
-    new OutcomeApplicativeError[F, E]()(F) with MonadError[Outcome[F, E, *], E] {
+    new OutcomeApplicativeError[F, E]()(F)
+      with MonadError[Outcome[F, E, *], E]
+      with Bifunctor[λ[(α, β) => Outcome[F, α, β]]] {
```
djspiewak: TIL this compiles on Dotty…

But more relevantly, there's a simpler alternative:

```suggestion
      with Bifunctor[Outcome[F, *, *]] {
```

djspiewak (changes requested):
This is really great! However, it also has to be done in one other place: the `ApplicativeError` instance defined in the low priority trait. The `MonadError` instance is only relevant if the `F` forms a `Traverse`, while the `ApplicativeError` is always available. Both are `Bifunctor`s.

djspiewak:
Fixes #1092

djspiewak (approved):
🎉

PR #1095, opened 2020-08-18
Add fork join benchmark

vasilmkd:
Setting up a benchmark for #1025.

Please do not rebase on this branch. This is highly experimental code subject to change at any point. I'm continuously rebasing and squashing commits.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
@@ -22,19 +22,13 @@ import java.util.concurrent.Executors
 import java.util.concurrent.atomic.AtomicInteger
 
 private[unsafe] abstract class IORuntimeCompanionPlatform { this: IORuntime.type =>
+
   def createDefaultComputeExecutionContext(
-      threadPrefix: String = "io-compute"): (ExecutionContext, () => Unit) = {
-    val threadCount = new AtomicInteger(0)
-    val executor = Executors.newFixedThreadPool(
-      Runtime.getRuntime().availableProcessors(),
-      { (r: Runnable) =>
-        val t = new Thread(r)
-        t.setName(s"${threadPrefix}-${threadCount.getAndIncrement()}")
-        t.setDaemon(true)
-        t
-      }
-    )
-    (ExecutionContext.fromExecutor(executor), { () => executor.shutdown() })
+      self: => IORuntime,
+      threadPrefix: String = "io-compute"): (WorkStealingThreadPool, () => Unit) = {
+    val threadCount = Runtime.getRuntime().availableProcessors()
+    val threadPool = new WorkStealingThreadPool(threadCount, threadPrefix, self)
+    (threadPool, () => threadPool.shutdown())
   }
 
   def createDefaultBlockingExecutionContext(
@@ -60,10 +54,13 @@ private[unsafe] abstract class IORuntimeCompanionPlatform { this: IORuntime.type
     (Scheduler.fromScheduledExecutor(scheduler), { () => scheduler.shutdown() })
   }
 
-  lazy val global: IORuntime =
-    IORuntime(
-      createDefaultComputeExecutionContext()._1,
-      createDefaultBlockingExecutionContext()._1,
-      createDefaultScheduler()._1,
+  lazy val global: IORuntime = {
+    val blocking = createDefaultBlockingExecutionContext()._1
+    val scheduler = createDefaultScheduler()._1
+    new IORuntime(
+      createDefaultComputeExecutionContext(global)._1,
+      blocking,
+      scheduler,
       () => ())
+  }
 }
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -17,8 +17,8 @@
 package cats.effect
 
 import cats.{~>, Eval, Monoid, Now, Parallel, Semigroup, SemigroupK, Show, StackSafeMonad}
-import cats.implicits._
 import cats.effect.implicits._
+import cats.implicits._
 
 import scala.annotation.unchecked.uncheckedVariance
 import scala.concurrent.{ExecutionContext, Future, Promise, TimeoutException}
@@ -100,7 +100,8 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
     }
   }
 
-  def evalOn(ec: ExecutionContext): IO[A] = IO.EvalOn(this, ec)
+  def evalOn(ec: ExecutionContext): IO[A] =
+    IO.EvalOn(this, ec)
 
   def flatMap[B](f: A => IO[B]): IO[B] = IO.FlatMap(this, f)
 
@@ -326,7 +327,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
     )
 
     if (shift)
-      runtime.compute.execute(fiber)
+      runtime.compute.executeFiber(fiber)
     else
       fiber.run()
 
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -16,6 +16,7 @@
 
 package cats.effect
 
+import cats.effect.unsafe.{IOExecutionContext, WorkQueue}
 import cats.implicits._
 
 import scala.annotation.{switch, tailrec}
@@ -63,7 +64,7 @@ private final class IOFiber[A](
     initMask: Int,
     cb: OutcomeIO[A] => Unit,
     startIO: IO[A],
-    startEC: ExecutionContext)
+    startEC: IOExecutionContext)
     extends IOFiberPlatform[A]
     with FiberIO[A]
     with Runnable {
@@ -76,8 +77,13 @@ private final class IOFiber[A](
   private[this] val objectState = new ArrayStack[AnyRef](16)
 
   // fast-path to head
-  private[this] var currentCtx: ExecutionContext = _
-  private[this] var ctxs: ArrayStack[ExecutionContext] = _
+  private[this] var currentCtx: IOExecutionContext = _
+  private[this] var ctxs: ArrayStack[IOExecutionContext] = _
+
+  // Direct reference to the work queue of the worker thread that executes this fiber. Used for local rescheduling of
+  // the fiber on the same worker thread, which avoids contention with other fibers when going through a shared queue,
+  // which is typical for non work stealing executors.
+  private[effect] var queue: WorkQueue = _
 
   private[this] var canceled: Boolean = false
   private[this] var masks: Int = initMask
@@ -414,16 +420,18 @@ private final class IOFiber[A](
 
         // ReadEC
         case 5 =>
-          runLoop(succeeded(currentCtx, 0), nextIteration)
+          // supply the underlying Scala `ExecutionContext` which is needed for the public API of `IO.executionContext`
+          runLoop(succeeded(currentCtx.underlying, 0), nextIteration)
 
         case 6 =>
           val cur = cur0.asInstanceOf[EvalOn[Any]]
 
-          // fast-path when it's an identity transformation
-          if (cur.ec eq currentCtx) {
+          // fast-path when it's an identity transformation, but make sure to compare the Scala `ExecutionContext`s
+          // that are wrapped by the internal `IOExecutionContext`
+          if (cur.ec eq currentCtx.underlying) {
             runLoop(cur.ioa, nextIteration)
           } else {
-            val ec = cur.ec
+            val ec = IOExecutionContext.wrap(cur.ec)
             currentCtx = ec
             ctxs.push(ec)
             conts.push(EvalOnK)
@@ -461,7 +469,7 @@ private final class IOFiber[A](
         case 10 =>
           val cur = cur0.asInstanceOf[OnCancel[Any]]
 
-          finalizers.push(EvalOn(cur.fin, currentCtx))
+          finalizers.push(EvalOn(cur.fin, currentCtx.underlying))
           // println(s"pushed onto finalizers: length = ${finalizers.unsafeIndex()}")
 
           conts.push(OnCancelK)
@@ -570,7 +578,8 @@ private final class IOFiber[A](
         case 18 =>
           resumeTag = CedeR
           resumeNextIteration = nextIteration
-          currentCtx.execute(this)
+          // put this fiber at the back of the work queue of the current worker thread
+          currentCtx.reschedule(this, queue)
 
         case 19 =>
           val cur = cur0.asInstanceOf[UnmaskRunLoop[Any]]
@@ -621,7 +630,7 @@ private final class IOFiber[A](
   }
 
   // returns the *new* context, not the old
-  private[this] def popContext(): ExecutionContext = {
+  private[this] def popContext(): IOExecutionContext = {
     ctxs.pop()
     val ec = ctxs.peek()
     currentCtx = ec
@@ -682,11 +691,15 @@ private final class IOFiber[A](
     }
   }
 
-  private[this] def execute(ec: ExecutionContext)(action: Runnable): Unit = {
+  private[this] def execute(ec: IOExecutionContext)(fiber: IOFiber[_]): Unit = {
     readBarrier()
 
     try {
-      ec.execute(action)
+      // Schedule the current fiber for continued execution on a possibly different worker thread. This does not locally
+      // reschedule the current fiber on the same worker thread and is used when such rescheduling is unsafe (when
+      // executed from a different worker or external thread, e.g. when switching to the blocking `ExecutionContext`
+      // or an `ExecutionContext` provided in `evalOn`).
+      ec.executeFiber(fiber)
     } catch {
       case _: RejectedExecutionException =>
         () // swallow this exception, since it means we're being externally murdered, so we should just... drop the runloop
@@ -735,13 +748,19 @@ private final class IOFiber[A](
       conts = new ByteStack(16)
       conts.push(RunTerminusK)
 
-      ctxs = new ArrayStack[ExecutionContext](2)
+      ctxs = new ArrayStack[IOExecutionContext](2)
       currentCtx = startEC
       ctxs.push(startEC)
 
       runLoop(startIO, 0)
     }
   }
+
+  /**
+   * Needed for deciding whether manual execution of this fiber on the work stealing thread pool is needed.
+   */
+  private[effect] def isCeding: Boolean =
+    resumeTag == CedeR
 
   private[this] def asyncContinueR(): Unit = {
     val e = asyncContinueEither
@@ -767,12 +786,11 @@ private final class IOFiber[A](
     if (error == null) {
       resumeTag = AfterBlockingSuccessfulR
       afterBlockingSuccessfulResult = r
-      currentCtx.execute(this)
     } else {
       resumeTag = AfterBlockingFailedR
       afterBlockingFailedError = error
-      currentCtx.execute(this)
     }
+    currentCtx.executeFiber(this)
   }
 
   private[this] def afterBlockingSuccessfulR(): Unit = {
@@ -848,7 +866,7 @@ private final class IOFiber[A](
   }
 
   private[this] def cancelationLoopFailureK(t: Throwable): IO[Any] = {
-    currentCtx.reportFailure(t)
+    currentCtx.underlying.reportFailure(t)
 
     cancelationLoopSuccessK()
   }
--- a/core/shared/src/main/scala/cats/effect/unsafe/CircularBuffer.scala
+++ b/core/shared/src/main/scala/cats/effect/unsafe/CircularBuffer.scala
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2020 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+package unsafe
+
+/**
+ * Circular buffer implementation of the pseudocode presented in the paper "Dynamic Circular Work-Stealing Deque" by
+ * Chase and Lev, SPAA 2005 (https://www.dre.vanderbilt.edu/~schmidt/PDF/work-stealing-dequeue.pdf).
+ */
+private final class CircularBuffer(private[this] val logSize: Int) {
+  private[unsafe] val buffer: Array[IOFiber[_]] = new Array[IOFiber[_]](1 << logSize)
+
+  val size: Int = 1 << logSize
+
+  def get(index: Int): IOFiber[_] =
+    buffer(index % size)
+
+  def put(index: Int, fiber: IOFiber[_]): Unit =
+    buffer(index % size) = fiber
+
+  def grow(bottom: Int, top: Int): CircularBuffer = {
+    val buf = new CircularBuffer(logSize + 1)
+    val len = bottom - top
+    System.arraycopy(buffer, top, buf.buffer, top, len)
+    buf
+  }
+}
--- a/core/shared/src/main/scala/cats/effect/unsafe/WorkQueue.scala
+++ b/core/shared/src/main/scala/cats/effect/unsafe/WorkQueue.scala
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2020 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+package unsafe
+
+import java.util.concurrent.atomic.AtomicInteger
+
+/**
+ * Circular buffer implementation of the pseudocode presented in the paper "Dynamic Circular Work-Stealing Deque" by
+ * Chase and Lev, SPAA 2005 (https://www.dre.vanderbilt.edu/~schmidt/PDF/work-stealing-dequeue.pdf).
+ */
+private[effect] final class WorkQueue {
+  @volatile private[this] var bottom: Int = 0
+  private[this] val top: AtomicInteger = new AtomicInteger()
+  @volatile private[this] var buffer: CircularBuffer = new CircularBuffer(
+    WorkQueue.LogInitialSize)
+
+  private[this] def casTop(oldVal: Int, newVal: Int): Boolean =
+    top.compareAndSet(oldVal, newVal)
+
+  def push(fiber: IOFiber[_]): Unit = {
+    val b = bottom
+    val t = top.get()
+    var cb = buffer
+    val size = b - t
+    if (size >= cb.size - 1) {
+      cb = cb.grow(b, t)
+      buffer = cb
+    }
+    cb.put(b, fiber)
+    bottom += 1
+  }
+
+  def steal(): IOFiber[_] = {
+    val t = top.get()
+    val b = bottom
+    val cb = buffer
+    val size = b - t
+    if (size <= 0) null
+    else {
+      val fiber = cb.get(t)
+      if (!casTop(t, t + 1)) null
+      else fiber
+    }
+  }
+}
+
+private[effect] object WorkQueue {
+  final val LogInitialSize = 13
+}
--- a/core/shared/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
+++ b/core/shared/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
@@ -38,7 +38,7 @@ private[effect] final class WorkStealingThreadPool(
   private[this] val workerThreads: Array[WorkerThread] = new Array[WorkerThread](threadCount)
   private[this] val externalQueue: ConcurrentLinkedQueue[IOFiber[_]] =
     new ConcurrentLinkedQueue[IOFiber[_]]()
-  private[this] val parkedThreads: ConcurrentLinkedQueue[WorkerThread] =
+  private[unsafe] val parkedThreads: ConcurrentLinkedQueue[WorkerThread] =
     new ConcurrentLinkedQueue[WorkerThread]()
 
   {
@@ -74,10 +74,8 @@ private[effect] final class WorkStealingThreadPool(
     externalQueue.poll()
 
   private[unsafe] def park(thread: WorkerThread): Unit = {
-    if (!done) {
-      parkedThreads.offer(thread)
-      LockSupport.park(this)
-    }
+    parkedThreads.offer(thread)
+    LockSupport.park(this)
   }
 
   override protected[effect] def underlying: ExecutionContext = this
--- a/core/shared/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
+++ b/core/shared/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
@@ -1,0 +1,107 @@
+/*
+ * Copyright 2020 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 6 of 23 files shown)


djspiewak (approved):

vasilmkd:
Base performance.

```
Benchmark                               (size)   Mode  Cnt  Score   Error    Units
ForkJoinBenchmark.async                1000000  thrpt   20  0.969 ± 0.019  ops/min
ForkJoinBenchmark.asyncTooManyThreads  1000000  thrpt   20  1.069 ± 0.043  ops/min
```

djspiewak:
> ops/min

LOL!

vasilmkd:
Yeah, it's pretty huge. I set it to minutes, otherwise the nominal value is too low and hard to compare.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -570,7 +575,14 @@ private final class IOFiber[A](
...
           resumeTag = CedeR
           resumeNextIteration = nextIteration
-          currentCtx.execute(this)
+          // println("PUSHING")
+          // println(hashCode())
+          // println(getClass())
+          // println(queue)
+          // println(this)
+          queue.push(this)
+          // println("WTF")
```
djspiewak: 😂

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -76,8 +77,9 @@ private final class IOFiber[A](
   private[this] val objectState = new ArrayStack[AnyRef](16)
 
   // fast-path to head
-  private[this] var currentCtx: ExecutionContext = _
-  private[this] var ctxs: ArrayStack[ExecutionContext] = _
+  private[this] var currentCtx: IOExecutionContext = _
+  private[this] var ctxs: ArrayStack[IOExecutionContext] = _
+  private[effect] var queue: WorkQueue = _
```
djspiewak: So I really firmly believe that what we should actually do here is keep the `ExecutionContext`, but add an *additional* var here which is the `IOExecutionContext` or `WorkStealingPool` or whatever it is. Actually just having hte `WorkQueue` alone is maybe enough. Not sure. Need to read more.

Anyway, this will be `null` most of the time, but we do an `isInstanceOf[WorkStealingPool]` in the constructor and also in `EvalOn` below to populate it. Then we have one bit of branching in `cede` (and similar) which first tries the `WorkStealingPool`, and then falls back to the `ExecutionContext` if it has to. This should pipeline optimize quite well since it'll be the common case.

I could be wrong that this is ideal, btw :-)

Anyway, what this *allows* us to do is isolate this to just `IORuntime` on the JVM. So we don't have to worry about ScalaJS or `TestContext` or custom `IORuntime`s, and instead just make `WorkStealingPool` the default executor on the JVM.

core/shared/src/main/scala/cats/effect/unsafe/CircularBuffer.scala
```diff
@@ -0,0 +1,37 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+package unsafe
+
+private final class CircularBuffer(private[this] val logSize: Int) {
```
djspiewak, line 24: Very nice

core/shared/src/main/scala/cats/effect/unsafe/WorkQueue.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  def size: Int = bottom - top.get()
+
+  private[this] def casTop(oldVal: Int, newVal: Int): Boolean =
+    top.compareAndSet(oldVal, newVal)
+
+  def push(fiber: IOFiber[_]): Unit = {
+    // println("WTF")
+    val b = bottom
+    val t = top.get()
```
djspiewak, line 37: Random thought for future: I wish we could get away without this memory barrier.

core/shared/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -0,0 +1,55 @@
...
+        // println("stealing from external queue")
+        task = pool.stealFromExternalQueue()
+      }
+      if (task == null) {
+        // println("stealing from other work queue")
+        task = pool.stealFromOtherWorkQueue(index)
+      }
+      if (task == null) {
+        // println("yielding")
+        Thread.`yield`()
```
djspiewak: LOL

djspiewak (commented):
Okay so this is proof positive that we want to go in this direction. :-)

I left some thoughts on the general architecture but I left the code alone for now. It looks really really good. I think if it's *possible* for us to move the divergence point up into `IORuntime`, we'll gain a lot more VM-specific control over it, but if we can't do that… we'll find a way to make it work I'm sure. :-)

Let's do this.

core/shared/src/main/scala/cats/effect/unsafe/CircularBuffer.scala
vasilmkd: I will definitely add the correct attributions. This is an implementation of the pseudo code found here:
https://www.dre.vanderbilt.edu/~schmidt/PDF/work-stealing-dequeue.pdf

core/shared/src/main/scala/cats/effect/IOFiber.scala
vasilmkd: This alone is a 2x improvement. 🤣

vasilmkd:
The initial results are in:
```
Benchmark                               (size)   Mode  Cnt  Score   Error    Units
ForkJoinBenchmark.async                1000000  thrpt   20  5.208 ± 0.188  ops/min
ForkJoinBenchmark.asyncTooManyThreads  1000000  thrpt   20  0.596 ± 0.029  ops/min
```

The asyncTooManyThreads benchmark is about 2x slower, but I suspect this is due to the busy waiting of 256 threads which are being rescheduled all of the time. In a normal ExecutionContext, most of the threads would be parked on the blocking shared queue of the executor.

core/shared/src/main/scala/cats/effect/unsafe/CircularBuffer.scala
RaasAhsan: I like this, going to change the `traceBufferSize` to use that, eliminates a ton of edge cases

RaasAhsan:
This is looking awesome, great work! I'm going to take a closer look when I get the chance

vasilmkd:
After the simple parking of fibers:
```
Benchmark                               (size)   Mode  Cnt  Score   Error    Units
ForkJoinBenchmark.async                1000000  thrpt   20  4.653 ± 0.134  ops/min
ForkJoinBenchmark.asyncTooManyThreads  1000000  thrpt   20  5.114 ± 0.183  ops/min
```

The 256 thread benchmark is now under control and shows encouraging performance.

vasilmkd:
CE3 baseline results:
```
Benchmark                         (size)   Mode  Cnt       Score     Error  Units
AsyncBenchmark.async                 100  thrpt   20   16454.278 ± 863.843  ops/s
AsyncBenchmark.bracket               100  thrpt   20   44897.966 ± 549.965  ops/s
AsyncBenchmark.cancelable            100  thrpt   20   15969.151 ±  60.191  ops/s
AsyncBenchmark.race                  100  thrpt   20    5382.858 ± 213.934  ops/s
AsyncBenchmark.racePair              100  thrpt   20    2450.408 ±  84.304  ops/s
AsyncBenchmark.start                 100  thrpt   20    4255.892 ± 256.158  ops/s
AsyncBenchmark.uncancelable          100  thrpt   20  207215.765 ± 701.529  ops/s
AttemptBenchmark.errorRaised       10000  thrpt   20    1632.443 ±  36.812  ops/s
AttemptBenchmark.happyPath         10000  thrpt   20    2374.598 ±  29.518  ops/s
DeepBindBenchmark.async            10000  thrpt   20     276.334 ±  14.485  ops/s
DeepBindBenchmark.delay            10000  thrpt   20    3518.550 ± 111.233  ops/s
DeepBindBenchmark.pure             10000  thrpt   20    4386.944 ±  82.132  ops/s
HandleErrorBenchmark.errorRaised   10000  thrpt   20    1909.981 ±  12.314  ops/s
HandleErrorBenchmark.happyPath     10000  thrpt   20    2650.335 ±   2.747  ops/s
MapCallsBenchmark.batch120           N/A  thrpt   20    5666.345 ±  88.487  ops/s
MapCallsBenchmark.batch30            N/A  thrpt   20    3587.128 ±   3.479  ops/s
MapCallsBenchmark.one                N/A  thrpt   20     255.772 ±   1.634  ops/s
MapStreamBenchmark.batch120          N/A  thrpt   20    3611.146 ±  51.303  ops/s
MapStreamBenchmark.batch30           N/A  thrpt   20    1326.146 ±   8.293  ops/s
MapStreamBenchmark.one               N/A  thrpt   20    1374.579 ±  46.911  ops/s
RedeemBenchmark.errorRaised        10000  thrpt   20    1158.939 ±   2.564  ops/s
RedeemBenchmark.happyPath          10000  thrpt   20    1506.731 ±   2.002  ops/s
RedeemWithBenchmark.errorRaised    10000  thrpt   20    1544.749 ±   7.523  ops/s
RedeemWithBenchmark.happyPath      10000  thrpt   20    2109.408 ±   3.310  ops/s
ShallowBindBenchmark.async         10000  thrpt   20     280.945 ±  10.868  ops/s
ShallowBindBenchmark.delay         10000  thrpt   20    3972.251 ±   3.932  ops/s
ShallowBindBenchmark.pure          10000  thrpt   20    4173.640 ± 250.476  ops/s
```

Work stealing:
```
Benchmark                          (size)   Mode  Cnt       Score      Error    Units
AsyncBenchmark.async                  100  thrpt   20   16514.667 ±  136.115    ops/s
AsyncBenchmark.bracket                100  thrpt   20   45302.346 ±   67.824    ops/s
AsyncBenchmark.cancelable             100  thrpt   20   16303.206 ±  153.857    ops/s
AsyncBenchmark.race                   100  thrpt   20    2576.847 ±   28.267    ops/s
AsyncBenchmark.racePair               100  thrpt   20    5114.066 ±   30.373    ops/s
AsyncBenchmark.start                  100  thrpt   20    2315.641 ±   53.189    ops/s
AsyncBenchmark.uncancelable           100  thrpt   20  214851.929 ± 3421.282    ops/s
AttemptBenchmark.errorRaised        10000  thrpt   20    1633.883 ±    1.764    ops/s
AttemptBenchmark.happyPath          10000  thrpt   20    2418.851 ±   11.429    ops/s
DeepBindBenchmark.async             10000  thrpt   20    1252.555 ±    1.223    ops/s
DeepBindBenchmark.delay             10000  thrpt   20    3562.736 ±  101.650    ops/s
DeepBindBenchmark.pure              10000  thrpt   20    4291.482 ±    2.722    ops/s
ForkJoinBenchmark.async           1000000  thrpt   20       4.878 ±    0.307  ops/min
HandleErrorBenchmark.errorRaised    10000  thrpt   20    1899.809 ±    6.773    ops/s
HandleErrorBenchmark.happyPath      10000  thrpt   20    2657.838 ±    3.429    ops/s
MapCallsBenchmark.batch120            N/A  thrpt   20    5872.534 ±   94.591    ops/s
MapCallsBenchmark.batch30             N/A  thrpt   20    3665.925 ±   77.331    ops/s
MapCallsBenchmark.one                 N/A  thrpt   20     249.396 ±    2.036    ops/s
MapStreamBenchmark.batch120           N/A  thrpt   20    3751.336 ±   56.893    ops/s
MapStreamBenchmark.batch30            N/A  thrpt   20    1317.117 ±   21.515    ops/s
MapStreamBenchmark.one                N/A  thrpt   20    1446.186 ±    1.683    ops/s
RedeemBenchmark.errorRaised         10000  thrpt   20    1301.631 ±   34.655    ops/s
RedeemBenchmark.happyPath           10000  thrpt   20    1515.741 ±   11.885    ops/s
RedeemWithBenchmark.errorRaised     10000  thrpt   20    1512.488 ±   13.641    ops/s
RedeemWithBenchmark.happyPath       10000  thrpt   20    2129.847 ±   10.845    ops/s
ShallowBindBenchmark.async          10000  thrpt   20    1191.815 ±   17.255    ops/s
ShallowBindBenchmark.delay          10000  thrpt   20    4009.926 ±   28.383    ops/s
ShallowBindBenchmark.pure           10000  thrpt   20    4296.604 ±  417.161    ops/s
```

core/shared/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -0,0 +1,94 @@
...
+
+import java.util.concurrent.ConcurrentLinkedQueue
+import java.util.concurrent.locks.LockSupport
+
+private final class WorkStealingThreadPool(
+    scheduler: Scheduler,
+    blocking: ExecutionContext
+) extends IOExecutionContext
+    with ExecutionContext {
+  private[this] val size = Runtime.getRuntime().availableProcessors()
```
djspiewak: Shouldn't this be passed in as a parameter? That would allow us to continue testing the "too many threads" case in the benchmark.

core/shared/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -0,0 +1,94 @@
...
+  override protected[effect] def executeFiber(fiber: IOFiber[_]): Unit = {
+    val _ = externalQueue.offer(fiber)
+    LockSupport.unpark(parkedThreads.poll())
+  }
+
+  override protected[effect] def reschedule(fiber: IOFiber[_], queue: WorkQueue): Unit =
+    queue.push(fiber)
+
+  def execute(runnable: Runnable): Unit = {
+    val runtime = IORuntime(this, blocking, scheduler, () => ())
```
djspiewak: Shouldn't we allocate this just once and hold onto it? Or even better, we could actually be passed it during construction as a lazy parameter.

core/shared/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -0,0 +1,94 @@
...
+    val fiber = IO.delay(runnable.run()).unsafeRunFiber(true)(_ => ())(runtime)
+    val _ = externalQueue.offer(fiber)
+  }
+
+  def reportFailure(cause: Throwable): Unit = {
+    println(cause)
```
djspiewak: This should at least be `cause.printStackTrace()`

core/shared/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -0,0 +1,46 @@
...
+    while (!pool.done) {
+      var task: IOFiber[_] = stealFromLocalWorkQueue()
+      if (task == null) {
+        task = pool.stealFromExternalQueue()
+      }
+      if (task == null) {
+        task = pool.stealFromOtherWorkQueue(index)
+      }
+      if (task == null) {
+        pool.park(this)
```
djspiewak: There's a race condition here:

1. All other workers are busy
2. This worker attempts to steal and fails
3. A new task comes in on the external queue (this can't happen with the local queue)
4. The pool puts it into the external queue without awakening any workers (since none are parked)
5. This worker parks itself

The external task will remain outstanding until some other worker wakes up. Maybe that's okay, since the cost of making all of this atomic would be very high.

core/shared/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -0,0 +1,94 @@
...
+    LockSupport.unpark(parkedThreads.poll())
+  }
+
+  override protected[effect] def reschedule(fiber: IOFiber[_], queue: WorkQueue): Unit =
+    queue.push(fiber)
+
+  def execute(runnable: Runnable): Unit = {
+    val runtime = IORuntime(this, blocking, scheduler, () => ())
+    val fiber = IO.delay(runnable.run()).unsafeRunFiber(true)(_ => ())(runtime)
+    val _ = externalQueue.offer(fiber)
```
djspiewak: This won't actually wake up any workers. If a new task comes in and all the workers are parked, it will *never* be seen. I believe this should be `executeFiber(fiber)` instead.

core/shared/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -0,0 +1,94 @@
...
+    LockSupport.unpark(parkedThreads.poll())
+  }
+
+  override protected[effect] def reschedule(fiber: IOFiber[_], queue: WorkQueue): Unit =
+    queue.push(fiber)
+
+  def execute(runnable: Runnable): Unit = {
+    val runtime = IORuntime(this, blocking, scheduler, () => ())
+    val fiber = IO.delay(runnable.run()).unsafeRunFiber(true)(_ => ())(runtime)
```
djspiewak:
```suggestion
    val fiber = IO(runnable.run()).unsafeRunFiber(true)(_.fold(reportFailure(_), _ => ())(runtime)
```

Any uncaught exceptions need to be reported.

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
```diff
@@ -22,19 +22,19 @@ import java.util.concurrent.Executors
...
-        t.setName(s"${threadPrefix}-${threadCount.getAndIncrement()}")
-        t.setDaemon(true)
-        t
-      }
-    )
-    (ExecutionContext.fromExecutor(executor), { () => executor.shutdown() })
+      scheduler: Scheduler,
+      blocking: ExecutionContext): (WorkStealingThreadPool, () => Unit) = {
+    val threadPool = new WorkStealingThreadPool(scheduler, blocking)
+    (threadPool, { () => () })
```
djspiewak: I think `WorkStealingThreadPool` needs a `shutdown()` method which interrupts and disposes of the workers.

core/shared/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -0,0 +1,94 @@
...
+
+package cats.effect
+package unsafe
+
+import scala.concurrent.ExecutionContext
+
+import java.util.concurrent.ConcurrentLinkedQueue
+import java.util.concurrent.locks.LockSupport
+
+private final class WorkStealingThreadPool(
```
djspiewak: This type is leaked publicly via the `workStealing` method of the `IORuntime` object. I think it makes a lot more sense to keep this private, so I would rather remove `IORuntimeCompanion#workStealing` and just call `new IORuntime` directly.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -100,7 +101,8 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
     }
   }
 
-  def evalOn(ec: ExecutionContext): IO[A] = IO.EvalOn(this, ec)
+  def evalOn(ec: ExecutionContext): IO[A] =
+    IO.EvalOn(this, IOExecutionContext.wrap(ec))
```
djspiewak: I would prefer we do this wrapping in `IOFiber`. It would avoid the need to *unwrap* in `IO#to` below.

core/shared/src/main/scala/cats/effect/unsafe/WorkQueue.scala
```diff
@@ -0,0 +1,70 @@
...
+    var cb = buffer
+    val size = b - t
+    if (size >= cb.size - 1) {
+      cb = cb.grow(b, t)
+      buffer = cb
+    }
+    cb.put(b, fiber)
+    bottom += 1
```
djspiewak, line 45: NB this is the only consistent write barrier associated with `reschedule`

core/shared/src/main/scala/cats/effect/unsafe/WorkQueue.scala
```diff
@@ -0,0 +1,70 @@
...
+  }
+}
+
+private[effect] object WorkQueue {
+  final val LogInitialSize = 13
+
+  object Sentinel extends WorkQueue {
+    def push(fiber: IOFiber[_]): Unit = {}
+    def steal(): IOFiber[_] = null
+  }
```
djspiewak: I honestly can't decide if this hack is worth it. The only benchmarks in which we *don't* want to shift are ones which don't touch `cede` or `async`, and if we're not touching `cede` or `async`, then we don't need the sentinel in the first place and `null` should suffice.

core/shared/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  private[unsafe] def stealFromOtherWorkQueue(from: Int): IOFiber[_] = {
+    var i = 0
+    while (i < size) {
+      val index = (from + i + 1) % size
+      val res = workerThreads(index).queue.steal()
+      if (res != null) return res
+      i += 1
+    }
```
djspiewak: We should think carefully about the fairness implications of this scan. I'm guessing it's fine, but we need to sit down and puzzle it through.

core/shared/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -0,0 +1,94 @@
...
+package unsafe
+
+import scala.concurrent.ExecutionContext
+
+import java.util.concurrent.ConcurrentLinkedQueue
+import java.util.concurrent.locks.LockSupport
+
+private final class WorkStealingThreadPool(
+    scheduler: Scheduler,
+    blocking: ExecutionContext
```
djspiewak: Instead, we can have `self0: => IORuntime` along with an instance `var self: IORuntime = _`. Then, in `execute`, we populate `self` with `self0` if `self == null`. This goes along with another change in `IORuntimeCompanion`.

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
```diff
@@ -60,10 +60,13 @@ private[unsafe] abstract class IORuntimeCompanionPlatform { this: IORuntime.type
...
-      createDefaultBlockingExecutionContext()._1,
-      createDefaultScheduler()._1,
+  lazy val global: IORuntime = {
+    val blocking = createDefaultBlockingExecutionContext()._1
+    val scheduler = createDefaultScheduler()._1
+    IORuntime.workStealing(
+      createDefaultComputeExecutionContext(scheduler, blocking)._1,
+      blocking,
+      scheduler,
       () => ())
```
djspiewak:
```suggestion
    lazy val back: IORuntime = new IORuntime(
      createDefaultComputeExecutionContext(back)._1,
      blocking,
      scheduler,
      () => ())
    
    back
```

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
djspiewak:
```suggestion
      self: => IORuntime): (WorkStealingThreadPool, () => Unit) = {
    val threadPool = new WorkStealingThreadPool(self)
    (threadPool, { () => () })
```

core/shared/src/main/scala/cats/effect/unsafe/WorkQueue.scala
```diff
@@ -0,0 +1,70 @@
...
+private final class CircularBufferWorkQueue extends WorkQueue {
+  @volatile private[this] var bottom: Int = 0
+  private[this] val top: AtomicInteger = new AtomicInteger()
+  @volatile private[this] var buffer: CircularBuffer = new CircularBuffer(
+    WorkQueue.LogInitialSize)
+
+  private[this] def casTop(oldVal: Int, newVal: Int): Boolean =
+    top.compareAndSet(oldVal, newVal)
+
+  def push(fiber: IOFiber[_]): Unit = {
```
djspiewak, line 35: So, I'm *almost* 100% positive that it's safe to call `push` from whatever threads we want, so long as we don't call multiple `push`es *simultaneously*. As long as we guarantee synchronicity, it's fine.

djspiewak (changes requested):

core/shared/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
vasilmkd: Done.

core/shared/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
vasilmkd: Thanks, I have no idea what to put here.

core/shared/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
vasilmkd: Done.

core/shared/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
vasilmkd: Definitely. I had this change, but somehow I lost it. Good catch.

core/shared/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
vasilmkd: Thanks.

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
vasilmkd: Done.

core/shared/src/main/scala/cats/effect/unsafe/WorkerThread.scala
vasilmkd: I guess we'll be leaving this as is for now?

core/shared/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
vasilmkd: Done.

core/shared/src/main/scala/cats/effect/IO.scala
vasilmkd: That's nicer, I agree.

core/shared/src/main/scala/cats/effect/unsafe/WorkQueue.scala
vasilmkd: Removed. We'll be going for automatic shifting and no shifting only in synchronous benchmarks.

core/shared/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
vasilmkd: Agreed, I really don't like this.

core/shared/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
vasilmkd: Does a `lazy val self: IORuntime = self0` work?

core/shared/src/main/scala/cats/effect/unsafe/WorkQueue.scala
Avasil: Noob question: 
How do you guys reason about memory barriers? Is it intuition or is there any good tool for it? I suspect that you're not looking at assembly :D

core/shared/src/main/scala/cats/effect/unsafe/WorkQueue.scala
Avasil: Q: Are volatile writes ensuring that all operations above (buffer changes) are visible?

djspiewak: They do so long as the volatile in question (bottom) is read before the buffer is. But more relevantly, note that steal is the only function called from another thread. The buffer is read and pushed otherwise entirely locally.

core/shared/src/main/scala/cats/effect/unsafe/CircularBuffer.scala
```diff
@@ -0,0 +1,41 @@
...
+
+package cats.effect
+package unsafe
+
+/**
+ * Circular buffer implementation of the pseudocode presented in the paper "Dynamic Circular Work-Stealing Deque" by
+ * Chase and Lev, SPAA 2005 (https://www.dre.vanderbilt.edu/~schmidt/PDF/work-stealing-dequeue.pdf).
+ */
+private final class CircularBuffer(private[this] val logSize: Int) {
+  private[unsafe] val buffer: Array[IOFiber[_]] = new Array[IOFiber[_]](1 << logSize)
```
RaasAhsan, line 25: We can replace `1 << logSize` with `size`, moving `size` before this to avoid the forward reference.

Another cool trick we can exploit: since we know the buffer size is a power of 2, then `forall n. n % powerOf2 = n & (powerOf2 - 1)`. We can define `val mask = size - 1` and replace instances `index % size` with `index & mask`, which requires fewer ALU instructions to compute.

RaasAhsan: Also, can we confirm that accesses to `buffer` aren't protected by a memory barrier? I know for a fact that fields declared object-private aren't, but I'm not sure about regular private or package-private fields.

Regular private scope is also more restrictive here but still allows other instances of the class to access the field.

core/shared/src/main/scala/cats/effect/unsafe/WorkQueue.scala
```diff
@@ -0,0 +1,64 @@
...
+  def steal(): IOFiber[_] = {
+    val t = top.get()
+    val b = bottom
+    val cb = buffer
+    val size = b - t
+    if (size <= 0) null
+    else {
+      val fiber = cb.get(t)
+      if (!casTop(t, t + 1)) null
+      else fiber
```
RaasAhsan, line 57: We can just do the buffer get if the CAS operation succeeds

core/shared/src/main/scala/cats/effect/unsafe/WorkQueue.scala
```diff
@@ -0,0 +1,64 @@
...
+import java.util.concurrent.atomic.AtomicInteger
+
+/**
+ * Circular buffer implementation of the pseudocode presented in the paper "Dynamic Circular Work-Stealing Deque" by
+ * Chase and Lev, SPAA 2005 (https://www.dre.vanderbilt.edu/~schmidt/PDF/work-stealing-dequeue.pdf).
+ */
+private[effect] final class WorkQueue {
+  @volatile private[this] var bottom: Int = 0
+  private[this] val top: AtomicInteger = new AtomicInteger()
+  @volatile private[this] var buffer: CircularBuffer = new CircularBuffer(
```
RaasAhsan, line 29: I need to think about this more, but I believe we can make `buffer` non-volatile. All assignments to the `buffer` variable occur *before* a write to `bottom`, and all reads from the `buffer` variable occur *after* a read from `bottom`, so a happens-before relationship is established.

core/shared/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -0,0 +1,107 @@
...
+    while (i < threadCount) {
+      workerThreads(i).start()
+      i += 1
+    }
+  }
+
+  private[unsafe] def stealFromOtherWorkQueue(from: Int): IOFiber[_] = {
+    var i = 0
+    while (i < threadCount) {
+      val index = (from + i + 1) % threadCount
```
RaasAhsan, line 65: This might be immaterial, but it took me a bit to realize this is the same as `if (index != from)`

core/shared/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -0,0 +1,51 @@
...
+ * `IOFiber`s are scheduled for execution. This is the main difference to a fixed thread executor, in which all threads
+ * contend for work from a single shared queue.
+ */
+private final class WorkerThread(
+    private[unsafe] val index: Int,
+    private[this] val pool: WorkStealingThreadPool)
+    extends Thread {
+  private[effect] val queue: WorkQueue = new WorkQueue()
+
+  private[this] def stealFromLocalWorkQueue(): IOFiber[_] =
```
RaasAhsan, line 34: IMO `stealFromLocalWorkQueue` and `stealFromExternalQueue` don't really sound like stealing

vasilmkd:
@RaasAhsan thanks for the review. I will address all of these issues, they are really great finds. But first of all, I'd like to arrive at a better implementation for the `WorkerThread` run loop because the current one is really detrimental to `start` performance. The implementation is currently too simple.

core/shared/src/main/scala/cats/effect/unsafe/WorkQueue.scala
vasilmkd: Done.

core/shared/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
djspiewak: Still pretty concerned about doing a linear scan here.

PR #1101, opened 2020-08-19
Unsafe convergent cache for traces

### LINKED ISSUE #1076: Replace CHM in tracing cache with thread unsafe hash table
We don't actually need thread safety. We just need non-corruption in the event of concurrent writes, and we need barrier free optimized reads. We don't need to compute length, either. These properties can be achieved with a traditional collision-free hash table (where you grow on every collision) without any concurrent coordination or atomicity. In the event of a write conflict, one side loses data. However, it's just a cache, so the data will be recomputed.

The cache will be expected to converge very quickly to a state in which most of the program is indexed and cached. Once that happens, there are no further writes and thus reads should be barrier-free and require a minimum of instructions.

RaasAhsan:
This should be the last PR to have tracing ready for 2.2.0. The current version of tracing relies on `ConcurrentHashMap` to keep a cache of traces. As described in #1076, `ConcurrentHashMap` offers thread safety at the cost of synchronization via read barriers on the hot read path. Because the cache will eventually converge to an optimal set, it's OK if we don't have safe publication (different threads may not see the same cache), as long as the objects we *do* see are all safely initialized.  Safe initialization means that an object values initialized in a constructor are visible to threads who observe a shared reference to the object. This property is implicitly guaranteed with the usual synchronization primitives that offer thread safety.

The Java memory model offers another method of achieving safe initialization that is much cheaper than synchronization: final fields. The semantics of final fields dictate that as long as `this` doesn't escape during the constructor of an object and that its final fields are set *before* the constructor completes, then any thread who observes a shared reference of the initialized object will observe fully initialized values for its final fields, and that guarantee extends transitively.

We leverage these semantics to build a thread-unsafe, convergent cache that achieves a higher read throughput than `ConcurrentHashMap`.

TODO:
- [ ] Tests
- [ ] Benchmarks
- [ ] Detailed documentation about the memory model guarantees we are leveraging here
- [ ] Collisions beneath `mask`
- [ ] Max buffer sizes
- [ ] Rewrite the implementation in Scala, verifying field modifiers

Closes #1076.

### DIFF
--- a/core/jvm/src/main/java/cats/effect/internals/UnsafeConvergentCache.java
+++ b/core/jvm/src/main/java/cats/effect/internals/UnsafeConvergentCache.java
@@ -120,8 +120,9 @@ public final class UnsafeConvergentCache<K, V> {
                 }
             }
 
-            // This is crucial, array needs to be fully initialized and primed
-            // *before* Buffer is safely initialized.
+            // This is crucial, newArray needs to be fully initialized and primed
+            // *before* Buffer is safely initialized, otherwise the final field
+            // semantics won't be extended to newArray inside Buffer.
             return new Buffer(currentLogSize, newArray);
         }
     }
--- a/core/jvm/src/main/java/cats/effect/internals/UnsafeConvergentCache.java
+++ b/core/jvm/src/main/java/cats/effect/internals/UnsafeConvergentCache.java
@@ -1,0 +1,129 @@
+/*
+ * Copyright (c) 2017-2019 The Typelevel Cats-effect Project Developers
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.internals;
+
+@SuppressWarnings("unchecked")
+public final class UnsafeConvergentCache<K, V> {
+
+    // We need final-field semantics when:
+    // 1) Growing an array and replacing the reference
+    // 2) Inserting a new cache entry
+    private Buffer<K, V> buffer = new Buffer<>(10);
+
+    public void put(K k, V v) {
+        if (buffer.put(k, v) != null) {
+            // A collision occurred, grow the buffer and re-insert
+            // all elements.
+            buffer = buffer.grow();
+        }
+    }
+
+    public V get(K k) {
+        Node<K, V> entry = buffer.get(k);
+        if (entry != null) {
+            return entry.value;
+        } else {
+            return null;
+        }
+    }
+
+    private static final class Node<A, B> {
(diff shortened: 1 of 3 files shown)


core/jvm/src/main/java/cats/effect/internals/UnsafeConvergentCache.java
```diff
@@ -0,0 +1,130 @@
...
+            }
+            return old;
+        }
+
+        public Node<K, V> get(K k) {
+            int hash = k.hashCode() & this.mask;
+            return this.array[hash];
+        }
+
+        public Buffer grow() {
```
RaasAhsan, line 89: TODO: pass the key value pair that caused the original collision here to insert

core/jvm/src/main/java/cats/effect/internals/UnsafeConvergentCache.java
```diff
@@ -0,0 +1,130 @@
...
+
+        public Buffer(int logSize, Node<K, V>[] array) {
+            this.logSize = logSize;
+            this.size = 1 << logSize;
+            this.mask = this.size - 1;
+            this.array = array;
+        }
+
+        public Node<K, V> put(K k, V v) {
+            int hash = k.hashCode() & this.mask;
```
RaasAhsan, line 75: CHM and HM have another way of generating the hash, will look into that

retronym:
Storing strong references to the lambda classes as the key of the frame cache smells like a potential classloader leak when the lambda class is loaded from a classloader with a shorter lifecycle than cats-eval's classloader.

`java.lang.ClassValue` is designed to avoid such leaks. It's also designed to be extremely fast for lookups.

djspiewak:
> java.lang.ClassValue is designed to avoid such leaks. It's also designed to be extremely fast for lookups.

TIL!

RaasAhsan:
> java.lang.ClassValue is designed to avoid such leaks. It's also designed to be extremely fast for lookups.

Didn't know that either, thanks! We've also noticed that `Class.hashCode` is a pretty costly operation, would be awesome if this could make that cost cheaper

retronym:
> Didn't know that either, thanks! We've also noticed that Class.hashCode is a pretty costly operation, would be awesome if this could make that cost cheaper

That doesn't sound right to me, it statically resolves to `Object.hashCode`, which is a JIT treats as an intrinsic and amounts to a read from the identity hash code which is computed once and stored in the object header. Profiler's can sometimes give misleading answers here due to safe-point biasing problem, try [async-profiler ](https://github.com/jvm-profiling-tools/async-profiler) and/or Java Flight Recorder with `-XX:+DebugNonSafePoints`. It's also a good idea to setup a microbenchmark in JMH to investigate these questions (JMH now integrates with async-profiler and JFR with  `-prof async` and `-prof:jfr`)

RaasAhsan:
I'll take a look at those! We did a bit of digging and arrived at the same conclusion about the `hashCode` being cached in the object header. The aforementioned cost was relative to its absence. I have a JMH benchmark sitting around for a simpler version of what's in this PR, and keying the array by `Class.hashCode` as opposed to arbitrary integers lowered throughput by 50%, which I thought was odd too

RaasAhsan:
I forgot to mention, the likely implementation for this cache actually won't retain any class references. We're acknowledging that tracing is an imprecise approximation, so in order to speed up reads, we'll only keep one value per array index. This means that we're effectively ignoring collisions, and to lower collision probability, we can grow the size of the array.

RaasAhsan:
Going to do this on CE3

PR #1103, opened 2020-08-19
Add IO.whenA and related convenience methods

Daenyth:
Applicative#whenA is a very commonly useful method, but for codebases using
concrete `IO` instead of 'tagless' style, they need to invoke it as
`Applicative[IO].whenA`, which is harder to discover and harder to explain

To make that use case easier, add convenience methods:
- IO.when
- IO.unless
- IO.raiseWhen
- IO.raiseUnless

I chose the names `raiseWhen` and `raiseUnless` to be more discoverable via
autocompletion from `IO.raise<tab>`, as well as for the parallels to the other
methods

A hypothetical application diff for contrast:
```diff
-Applicative[IO].whenA(badThings)(IO.raiseError(new Exception("oh no")))
+IO.raiseWhen(badThings)(new Exception("oh no"))
```

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -1624,6 +1624,48 @@ object IO extends IOInstances {
       cb(Right(ctx.trace()))
     }
 
+  /**
+   * Returns the given argument if `cond` is true, otherwise `IO.Unit`
+   *
+   * @see [[IO.unlessA]] for the inverse
+   * @see [[IO.raiseWhen]] for conditionally raising an error
+   */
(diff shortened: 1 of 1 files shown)


djspiewak:
I am *massively* in favor of continuing to pull combinators explicitly onto the concrete `IO` for discoverability. Though, I really strongly believe that we should keep the names consistent with the Cats ones to avoid creating confusion. You're definitely right that `whenA` doesn't make much sense on `IO`, but it's consistent with the applicative variant, so I kind of want to keep it.

Daenyth:
@djspiewak Fixed

djspiewak (approved):
🎉

Issue #1096, opened 2020-08-18
Fix exhaustiveness issues in Resource on Dotty


djspiewak:
The warnings are legit. They should be fixed. Run `sbt ++0.26.0-RC1 coreJVM/compile` to reproduce.

TimWSpence:
After discussion with @djspiewak we think this should probably be closed for now as we don't believe the issues are fixable whilst keeping our current variance semantics (discussion of this [here](https://gitter.im/lampepfl/dotty?at=5f3d25b08b8d4f633e003717))

The pattern match warnings are an actual bug in Dotty which is being tracked [here](https://github.com/lampepfl/dotty/issues/9603)

You can see my attempt to fix the other warnings (runtime typechecks) [here](https://github.com/typelevel/cats-effect/pull/1105) . Unfortunately, these cause 2.12 to lose its mind:

```
[error] /home/runner/work/cats-effect/cats-effect/core/shared/src/main/scala/cats/effect/Resource.scala:114:18: constructor cannot be instantiated to expected type;
[error]  found   : cats.effect.Resource.Allocate[F,A]
[error]  required: cats.effect.Resource[?G1,Any] where type ?G1 <: G (this is a GADT skolem)
[error]         case a @ Allocate(_) => {
[error]                  ^
[error] /home/runner/work/cats-effect/cats-effect/core/shared/src/main/scala/cats/effect/Resource.scala:127:18: constructor cannot be instantiated to expected type;
[error]  found   : cats.effect.Resource.Bind[F,S,A]
[error]  required: cats.effect.Resource[?G2,Any] where type ?G2 <: G (this is a GADT skolem)
[error]         case b @ Bind(_, _) => {
[error]                  ^
[error] /home/runner/work/cats-effect/cats-effect/core/shared/src/main/scala/cats/effect/Resource.scala:131:18: constructor cannot be instantiated to expected type;
[error]  found   : cats.effect.Resource.Suspend[F,A]
[error]  required: cats.effect.Resource[?G3,Any] where type ?G3 <: G (this is a GADT skolem)
[error]         case s @ Suspend(_) => {
```

So it seems that if we use the current runtime typecheck style then Dotty issues a warning but if we use actual pattern matching then 2.12 explodes. Therefore I think we're stuck with dotty warnings while we support 2.12 (unless someone can suggest a significantly different encoding)

mpilquist:
[...]

PR #1105, opened 2020-08-19
1096 fix warnings

TimWSpence:
Fix runtime typecheck warnings.

Unfortunately this requires us to remove the tailrec annotations as 2.13 is buggy and thinks that it isn't tail-recursive. I've checked the bytecode and it looks ok though.


TimWSpence:
Abandoning. Reasoning is discussed on the issue ticket: https://github.com/typelevel/cats-effect/issues/1096

PR #1114, opened 2020-08-21
Replace Resource.ExitCase with a type alias to Outcome

### LINKED ISSUE #1110: Possibly eliminate Resource.ExitCase
In theory:

```scala
type ExitCase[E] = Outcome[Nothing, E, Nothing]
```

I would kinda like that to be the *actual* reality so we don't have two types floating around meaning almost the same thing.

joroKr21:
Closes #1110

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Resource.scala
+++ b/core/shared/src/main/scala/cats/effect/Resource.scala
@@ -17,7 +17,7 @@
 package cats.effect
 
 import cats._
-import cats.data.AndThen
+import cats.data.{AndThen, Const}
 import cats.effect.concurrent.Ref
 import cats.implicits._
 import cats.effect.implicits._
@@ -455,17 +455,16 @@ object Resource extends ResourceInstances with ResourcePlatform {
   final case class Suspend[F[_], A](resource: F[Resource[F, A]]) extends Resource[F, A]
 
   /**
-   * Type for signaling the exit condition of an effectful
-   * computation, that may either succeed, fail with an error or
-   * get canceled.
+   * Type of [[Outcome]] for signaling the exit condition of an effectful computation,
+   * that may either succeed, fail with an error or get canceled.
    *
    * The types of exit signals are:
    *
-   *  - [[ExitCase$.Completed Completed]]: for successful completion
-   *  - [[ExitCase$.Error Error]]: for termination in failure
-   *  - [[ExitCase$.Canceled Canceled]]: for abortion
+   *  - [[ExitCase.Completed Completed]]: for successful completion
+   *  - [[ExitCase.Errored Errored]]: for termination in failure
+   *  - [[ExitCase.Canceled Canceled]]: for abortion
    */
-  sealed trait ExitCase extends Product with Serializable
+  type ExitCase = Outcome[Const[Unit, *], Throwable, Nothing]
   object ExitCase {
 
     /**
@@ -478,12 +477,14 @@ object Resource extends ResourceInstances with ResourcePlatform {
      * outcome for the user, but it does for the purposes of the
      * `bracket` operation. <-- TODO still true?
      */
-    case object Completed extends ExitCase
+    type Canceled = Outcome.Canceled[Const[Unit, *], Throwable, Nothing]
+    val Canceled: Canceled = Outcome.Canceled()
 
     /**
      * An [[ExitCase]] signaling completion in failure.
      */
-    final case class Errored(e: Throwable) extends ExitCase
+    type Errored = Outcome.Errored[Const[Unit, *], Throwable, Nothing]
+    val Errored: Outcome.Errored.type = Outcome.Errored
 
     /**
      * An [[ExitCase]] signaling that the action was aborted.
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -454,49 +452,15 @@ object Resource extends ResourceInstances with ResourcePlatform {
...
-     * like [[IO]] and the task yielded by `bracket` gets canceled
-     * when it's at its `use` phase.
-     *
-     * Thus [[Bracket]] allows you to observe interruption conditions
-     * and act on them.
-     */
-    case object Canceled extends ExitCase
-  }
+  type Void[A] = Unit
+  val Void: Void[Nothing] = ()
```
djspiewak: Hmm, does this buy us anything over just `Const[Unit, *]`?

joroKr21: Good point, it will be another layer of boxing, but that is fine if we cache `Completed`

SystemFw:
I like `Const` better too.

How does pattern matching look like with this change? For example when writing a `makeCase`

djspiewak:
I like this quite a bit more than what's in series/3.x; can we get a resolution to this? Do you have further concerns/objections @SystemFw?

SystemFw:
I'm warming up to this, but I want to see how does it look for users when they pattern match using `makeCase`. If they end up having to fight with `a: Outcome[F, _, ...]` or things like that, then I think the extra prettiness inside cats-effect might not be worth it.

djspiewak:
> I'm warming up to this, but I want to see how does it look for users when they pattern match using makeCase. If they end up having to fight with a: Outcome[F, _, ...] or things like that, then I think the extra prettiness inside cats-effect might not be worth it.

The only weirdness would be when matching on `ExitCase.Completed`, in which case they'd probably want to do `Completed(_)` rather than deal with it. `Errored` and `Canceled` would work as expected.

SystemFw:
> in which case they'd probably want to do Completed(_) rather than deal with it

why do we want to add this artificial problem to our users? I believe it to be potentially worse than observing a bit of duplication 

Anyway, I don't feel that strongly about it, might be worth getting someone else's opinion as well :)

joroKr21:
> How does pattern matching look like with this change?

I will try to get to that next week

djspiewak:
Do we want to clean this up and squeeze it into M1? It still really annoys me that `ExitCase` is a totally separate thing.

bplommer:
I'm happy to finish this off (unless @joroKr21 wants to, of course).

bplommer:
On reflection - given the main typeclasses no longer have their error channel hard-coded to `Throwable`, why should `Resource`? Should there be a `GenResource` with a parametric error type?

djspiewak:
This is a good point. I think generalizing resource a bit is a good idea. GenResource follows the idiom we set elsewhere.

joroKr21:
@bplommer feel free to take over, I am swamped ATM 😅

core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -455,17 +455,16 @@ object Resource extends ResourceInstances with ResourcePlatform {
   final case class Suspend[F[_], A](resource: F[Resource[F, A]]) extends Resource[F, A]
 
   /**
-   * Type for signaling the exit condition of an effectful
-   * computation, that may either succeed, fail with an error or
-   * get canceled.
+   * Type of [[Outcome]] for signaling the exit condition of an effectful computation,
+   * that may either succeed, fail with an error or get canceled.
```
kubukoz, line 459:
```suggestion
   * that either succeeded, failed with an error or got canceled.
```

core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -495,7 +496,8 @@ object Resource extends ResourceInstances with ResourcePlatform {
      * Thus [[Bracket]] allows you to observe interruption conditions
      * and act on them.
      */
-    case object Canceled extends ExitCase
+    type Completed = Outcome.Completed[Const[Unit, *], Throwable, Nothing]
```
rkrzewski, line 499: I've noticed that `Canceled` and `Completed` switched places wrt old implementation, but their corresponding scaladocs did not.

djspiewak: That definitely seems like a mistake :-P

joroKr21:
> The only weirdness would be when matching on ExitCase.Completed, in which case they'd probably want to do Completed(_) rather than deal with it. Errored and Canceled would work as expected.

I'm afraid there is no way to convince the exhaustiveness checker that a pattern match like this would be exhaustive:
```scala
def toFoo(x: ExitCase): Int =
      x match {
        case Succeeded => 1
        case Errored(_) => 2
        case Canceled => 3
      }
```

So I think indeed that's not worth it.

joroKr21:
Should we add a `toOutcome` method on `ExitCase`?

Issue #946, opened 2020-07-16
Consider whether LinkedMap is still necessary


RaasAhsan:
As part of https://github.com/typelevel/cats-effect/pull/918, `Deferred` and `MVarConcurrent` were refactored so that callbacks are uniquely identified by a monotonically increasing long. This is in fact how `LinkedMap` tracks insertion order, so we may not actually need it anymore. If that's the case, we should remove it both on the 2.x series and the 3.x series (after https://github.com/typelevel/cats-effect/issues/942).

djspiewak:
This is an absolutely lovely observation, and it seems likely to be true!

SystemFw:
I'm on this btw

PR #1128, opened 2020-08-24
Deferred

### LINKED ISSUE #942: Port over the Deferred optimizations to CE3
@RaasAhsan made some changes to optimize `Deferred` in master, but those changes landed just *after* @TimWSpence pulled it out and copied it over to **ce3**. We should copy those changes over.

SystemFw:
Fixes #942 and #946 

Bunch of changes to Deferred:
- Remove TryableDeferred, which was only there because of compat
- Port the Long optimisation from ce2
- Go one step further and remove LinkedMap entirely

The latter can probably also be ported to CE2, let me know if you want me to do that.


SystemFw:
Oops, I broke MVar, don't merge yet

RaasAhsan:
> The latter can probably also be ported to CE2, let me know if you want me to do that.

It would be cool if we can fit that in before 2.2.0, not sure when @djspiewak is planning to cut that release though

SystemFw:
@RaasAhsan Let me see how much work is it to apply the same treatment to MVar (I suspect little), then I can port both quite quickly.

concurrent/shared/src/main/scala/cats/effect/concurrent/Deferred.scala
```diff
@@ -75,157 +76,175 @@ abstract class Deferred[F[_], A] {
...
+      def notifyReaders(readers: LongMap[A => Unit]): F[Unit] = {
+        // LongMap iterators return values in unsigned key order,
+        // which corresponds to the arrival order of readers since
+        // insertion is governed by a monotonically increasing id
+        val cursor = readers.valuesIterator
+        var acc = F.unit
+
+        while (cursor.hasNext) {
+          val next = cursor.next()
+          val task = F.map(F.start(F.delay(next(a))))(mapUnit)
```
RaasAhsan, line 215: I just noticed, but for `IO` the `start` is unnecessary since `next(a)` will always internally shift, but we can't rely on that fact in general :(

RaasAhsan (approved):
👍  Awesome work, way cleaner!

concurrent/shared/src/main/scala/cats/effect/concurrent/MVar.scala
```diff
@@ -646,8 +639,10 @@ private[effect] object MVarAsync {
    * `put` operations.
    */
   final private case class WaitForPut[A](
-      reads: LinkedMap[Id, Listener[A]],
```
RaasAhsan, line 649: Can we delete `LinkedMap` from the codebase now?

SystemFw: yeah just done it

SystemFw: I'm going to port to CE 2 straight away I think

concurrent/shared/src/main/scala/cats/effect/concurrent/Deferred.scala
djspiewak: Isn't `next` an `async` callback? If so, we can rely on internal shifting. It's in the laws, albeit indirectly.

RaasAhsan: Yeah `next` is an `async` callback, which law captures internal shifting?

djspiewak: https://github.com/typelevel/cats-effect/blob/series/3.x/laws/shared/src/main/scala/cats/effect/laws/AsyncLaws.scala#L47-L48

djspiewak: Also I'm realizing that law should be reformulated so as to be testable. :-/

RaasAhsan: ohh, that's pretty clever!

SystemFw: is that settled then, can I remove `start`? although I have to admit I'm having some trouble seeing how that law implies internal shifting (early morning for me)

djspiewak: @SystemFw What we *don't* do is tie the `executionContext` to `delay`, which is undecidable in general. However, the commutativity does guarantee that `F.executionContext` returns a consistent value regardless of what happens before or after it, and can only be affected by what is *around* it (i.e. `evalOn`). If we assume that `executionContext` represents the executor *actually* in use (which we have to assume; we can't generally assert it), then the commutativity law forces internal shifting if we instantiate `fa` with something which contains `async`.

PR #1133, opened 2020-08-25
Add onFinalize/onFinalizeCase methods to Resource

bplommer:


Daenyth (approved):

core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -212,7 +212,20 @@ sealed abstract class Resource[+F[_], +A] {
...
+    onFinalizeCase(_ => finalizer)
+
+  /**
+   * Like `onFinalize`, but the action performed depends on the exit case.
+   */
+  def onFinalizeCase[G[x] >: F[x]](f: ExitCase[Throwable] => G[Unit])(implicit F: Applicative[G]): Resource[G, A] =
+    Resource.makeCase(F.unit) { case (_, exitCase) => f(exitCase) }.flatMap[G, A](_ => this)
+
+  /**
+   * Given a natural transformation from `F` to `G`, transforms this)
```
ritschwumm: the `)` at the end of this line should not bet there, right?

bplommer: Correct - I wondered why the diff looked odd

djspiewak:
Thanks! Sorry it took me a while to loop back to this; it seems like a lovely addition to the API.

PR #1134, opened 2020-08-25
All things work stealing

vasilmkd:
Related to #1025 and #1095.

This is for the most part a port of the `tokio` runtime from Rust. I have added all proper attributions to every file that is heavily inspired by the `tokio` source code.

Although not necessary for understaing or reviewing this PR, still, a highly recommended read: https://tokio.rs/blog/2019-10-scheduler#the-next-generation-tokio-scheduler.

Note, please avoid rebasing on this branch, I believe there will be changes and I might want to rebase them away. Thank you.

I erred on the side of too many comments are better than no comments at all.

Preliminary benchmark results, I plan to run them again due to new changes and optimizations.

CE3 master:
```
Benchmark                                   (size)   Mode  Cnt       Score      Error  Units
AsyncBenchmark.async                           100  thrpt   20   16225.234 ±  395.071  ops/s
AsyncBenchmark.bracket                         100  thrpt   20   45362.819 ± 1125.382  ops/s
AsyncBenchmark.cancelable                      100  thrpt   20   16619.367 ±  109.049  ops/s
AsyncBenchmark.race                            100  thrpt   20    5544.669 ±   52.545  ops/s
AsyncBenchmark.racePair                        100  thrpt   20    2425.187 ±   66.205  ops/s
AsyncBenchmark.start                           100  thrpt   20    4259.642 ±   14.138  ops/s
AsyncBenchmark.uncancelable                    100  thrpt   20  213556.440 ± 5354.325  ops/s
AttemptBenchmark.errorRaised                 10000  thrpt   20    1648.350 ±    8.169  ops/s
AttemptBenchmark.happyPath                   10000  thrpt   20    2366.397 ±    3.078  ops/s
DeepBindBenchmark.async                      10000  thrpt   20     267.027 ±    0.484  ops/s
DeepBindBenchmark.delay                      10000  thrpt   20    3691.817 ±    4.206  ops/s
DeepBindBenchmark.pure                       10000  thrpt   20    4313.325 ±   34.245  ops/s
HandleErrorBenchmark.errorRaised             10000  thrpt   20    1905.917 ±   28.730  ops/s
HandleErrorBenchmark.happyPath               10000  thrpt   20    2668.715 ±   15.616  ops/s
MapCallsBenchmark.batch120                     N/A  thrpt   20    5876.335 ±   83.336  ops/s
MapCallsBenchmark.batch30                      N/A  thrpt   20    3649.645 ±   92.494  ops/s
MapCallsBenchmark.one                          N/A  thrpt   20     254.427 ±    2.567  ops/s
MapStreamBenchmark.batch120                    N/A  thrpt   20    3612.403 ±   52.848  ops/s
MapStreamBenchmark.batch30                     N/A  thrpt   20    1307.002 ±    7.202  ops/s
MapStreamBenchmark.one                         N/A  thrpt   20    1367.057 ±   33.906  ops/s
RedeemBenchmark.errorRaised                  10000  thrpt   20    1207.612 ±   45.724  ops/s
RedeemBenchmark.happyPath                    10000  thrpt   20    1526.144 ±    3.851  ops/s
RedeemWithBenchmark.errorRaised              10000  thrpt   20    1565.245 ±   11.768  ops/s
RedeemWithBenchmark.happyPath                10000  thrpt   20    2340.351 ±   52.876  ops/s
ShallowBindBenchmark.async                   10000  thrpt   20     281.658 ±   15.705  ops/s
ShallowBindBenchmark.delay                   10000  thrpt   20    3963.008 ±    3.245  ops/s
ShallowBindBenchmark.pure                    10000  thrpt   20    4028.397 ±  136.441  ops/s
WorkStealingBenchmark.async                1000000  thrpt   20       0.995 ±    0.019  ops/min
WorkStealingBenchmark.asyncTooManyThreads  1000000  thrpt   20       1.162 ±    0.068  ops/min
```

Work stealing:
```
Benchmark                                   (size)   Mode  Cnt       Score      Error    Units
AsyncBenchmark.async                           100  thrpt   20   22436.151 ±  217.206    ops/s
AsyncBenchmark.bracket                         100  thrpt   20   48127.640 ±  338.857    ops/s
AsyncBenchmark.cancelable                      100  thrpt   20   22838.937 ±  276.638    ops/s
AsyncBenchmark.race                            100  thrpt   20    4607.468 ±   26.697    ops/s
AsyncBenchmark.racePair                        100  thrpt   20   18638.108 ±  271.767    ops/s
AsyncBenchmark.start                           100  thrpt   20    7131.536 ±   56.862    ops/s
AsyncBenchmark.uncancelable                    100  thrpt   20  225983.305 ± 3854.823    ops/s
AttemptBenchmark.errorRaised                 10000  thrpt   20    1706.631 ±   20.437    ops/s
AttemptBenchmark.happyPath                   10000  thrpt   20    2528.153 ±   10.204    ops/s
DeepBindBenchmark.async                      10000  thrpt   20     506.963 ±   28.186    ops/s
DeepBindBenchmark.delay                      10000  thrpt   20    3619.885 ±   19.944    ops/s
DeepBindBenchmark.pure                       10000  thrpt   20    4582.860 ±   82.699    ops/s
HandleErrorBenchmark.errorRaised             10000  thrpt   20    2058.460 ±   18.188    ops/s
HandleErrorBenchmark.happyPath               10000  thrpt   20    2787.462 ±   36.151    ops/s
MapCallsBenchmark.batch120                     N/A  thrpt   20    5585.960 ±   26.981    ops/s
MapCallsBenchmark.batch30                      N/A  thrpt   20    3793.869 ±  181.826    ops/s
MapCallsBenchmark.one                          N/A  thrpt   20     267.421 ±    3.023    ops/s
MapStreamBenchmark.batch120                    N/A  thrpt   20    3630.350 ±   45.796    ops/s
MapStreamBenchmark.batch30                     N/A  thrpt   20    1322.036 ±   22.228    ops/s
MapStreamBenchmark.one                         N/A  thrpt   20    1461.211 ±   46.251    ops/s
RedeemBenchmark.errorRaised                  10000  thrpt   20    1305.533 ±   83.990    ops/s
RedeemBenchmark.happyPath                    10000  thrpt   20    1573.721 ±   15.368    ops/s
RedeemWithBenchmark.errorRaised              10000  thrpt   20    1594.318 ±   27.944    ops/s
RedeemWithBenchmark.happyPath                10000  thrpt   20    2312.903 ±  129.530    ops/s
ShallowBindBenchmark.async                   10000  thrpt   20     502.602 ±   10.863    ops/s
ShallowBindBenchmark.delay                   10000  thrpt   20    4140.454 ±   26.828    ops/s
ShallowBindBenchmark.pure                    10000  thrpt   20    4608.918 ±  335.367    ops/s
WorkStealingBenchmark.async                1000000  thrpt   20       4.883 ±    0.162  ops/min
WorkStealingBenchmark.asyncTooManyThreads  1000000  thrpt   20       7.045 ±    0.450  ops/min
```

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/IOFiberPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/IOFiberPlatform.scala
@@ -26,6 +26,10 @@ private[effect] abstract class IOFiberPlatform[A] { this: IOFiber[A] =>
 
   private[this] val TypeInterruptibleMany = Sync.Type.InterruptibleMany
 
+  // Allocation free linked queue nodes of `IOFiber[_]` objects.
+  // Represents the external work stealing thread pool queue.
+  private[effect] var next: IOFiber[_] = _
+
   protected final def interruptibleImpl(
       cur: IO.Blocking[Any],
       blockingEc: ExecutionContext): IO[Any] = {
--- a/core/jvm/src/main/scala/cats/effect/unsafe/ExternalQueue.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/ExternalQueue.scala
@@ -1,0 +1,180 @@
+/*
+ * Copyright 2020 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * This code is an adaptation of the `Inject` queue from the `tokio` runtime.
+ * The original source code in Rust is licensed under the MIT license and available
+ * at: https://docs.rs/crate/tokio/0.2.22/source/src/runtime/queue.rs.
+ *
+ * For the reasoning behind the design decisions of this code, please consult:
+ * https://tokio.rs/blog/2019-10-scheduler#the-next-generation-tokio-scheduler.
+ */
+
+package cats.effect
+package unsafe
+
+/**
+ * Multi producer, multi consumer linked queue of `IOFiber[_]` objects.
+ * This class cannot function without the `IOFiber#next` field, which is used
+ * to achieve allocation free linked queue semantics.
+ *
+ * New fibers scheduled by external threads end up at the back of this queue.
+ * When a local worker queue overflows (cannot accept new fibers), half of it
+ * gets spilled over to this queue.
+ */
+private final class ExternalQueue {
+  import WorkStealingQueueConstants._
+
+  // Lock that ensures exclusive access to the internal state of the linked queue.
+  private[this] val lock: AnyRef = new Object()
+
+  // Pointer to the least recently added fiber in the queue.
+  // The next to be dequeued when calling `dequeue`.
+  private[this] var head: IOFiber[_] = null
+
+  // Pointer to the most recently added fiber in the queue.
+  private[this] var tail: IOFiber[_] = null
+
+  // Tracks whether the queue has been shutdown.
+  private[this] var _shutdown: Boolean = false
+
+  // Number of enqueued fibers. Used as a fast-path to avoid unnecessary locking
+  // on the fast path. Can be accessed without holding the lock.
+  @volatile private[this] var len: Int = 0
+
+  /**
+   * Enqueues a fiber for later execution at the back of the queue.
+   */
+  def enqueue(fiber: IOFiber[_]): Unit = {
+    lock.synchronized {
+      if (_shutdown) {
+        // Do not accept new fibers if the queue has been shut down.
+        return
+      }
+
+      // Safe to mutate the internal state because we are holding the lock.
+      if (tail != null) {
+        // The queue is not empty, put the new fiber at the back of the queue.
+        tail.next = fiber
+      } else {
+        // The queue is empty, the new fiber becomes the head of the queue.
+        head = fiber
+      }
+
+      // Set the tail to point to the new fiber.
+      tail = fiber
+
+      // Should be changed for a `plain` get and `release` set.
+      // Requires the use of `VarHandles` (Java 9+) or Unsafe
+      // (not as portable and being phased out).
+      len += 1
+    }
+  }
+
+  /**
+   * Enqueues a linked list of fibers for later execution at the back of the queue.
+   */
+  def enqueueBatch(hd: IOFiber[_], tl: IOFiber[_]): Unit = {
+    lock.synchronized {
+      // Safe to mutate the internal state because we are holding the lock.
+      if (tail != null) {
+        // The queue is not empty, put the head of the fiber batch at the back of the queue.
+        tail.next = hd
+      } else {
+        // The queue is empty, the head of the fiber batch becomes the head of the queue.
+        head = hd
+      }
+
+      tail = tl
+
+      // Should be changed for a `plain` get and `release` set.
+      // Requires the use of `VarHandles` (Java 9+) or Unsafe
+      // (not as portable and being phased out).
+      len += BatchLength
+    }
+  }
+
+  /**
+   * Dequeues the least recently added fiber from the front of the queue for execution.
+   * Returns `null` if the queue is empty.
+   */
+  def dequeue(): IOFiber[_] = {
+    // Fast path, no locking.
+    if (isEmpty()) {
+      return null
+    }
+
+    lock.synchronized {
+      // Obtain the head of the queue.
+      val fiber = head
+
+      // The queue could have been emptied by the time we acquired the lock.
+      if (fiber == null) {
+        // Nothing to do, return.
+        return null
+      }
+
+      // Make the head of the queue point to the next fiber.
+      head = fiber.next
+
+      // If the new head is `null`, the queue is empty. Make sure the tail is consistent.
+      if (head == null) {
+        tail = null
+      }
+
+      // Unlink the fiber from the linked queue before returning.
+      fiber.next = null
+
+      // Should be changed for a `plain` get and `release` set.
+      // Requires the use of `VarHandles` (Java 9+) or Unsafe
+      // (not as portable and being phased out).
+      len -= 1
+
+      fiber
+    }
+  }
+
+  /**
+   * Returns true if there are no enqueued fibers.
+   */
+  def isEmpty(): Boolean =
+    len == 0
+
+  /**
+   * Shutdown, drain and unlink the queue. No more fibers can be enqueued after
+   * this call. Repeated calls have no effect.
+   */
+  def shutdown(): Unit = {
+    lock.synchronized {
+      if (_shutdown) {
+        // The queue has already been shutdown. Return.
+        return
+      }
+
+      // Unlink and drain the queue.
+      var fiber: IOFiber[_] = head
+      var next: IOFiber[_] = null
+      while (fiber != null) {
+        next = fiber.next
+        fiber.next = null
+        fiber = next
+      }
+
+      // Set the shutdown flag.
+      _shutdown = true
+    }
+  }
+}
--- a/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
@@ -22,22 +22,15 @@ import java.util.concurrent.Executors
 import java.util.concurrent.atomic.AtomicInteger
 
 private[unsafe] abstract class IORuntimeCompanionPlatform { this: IORuntime.type =>
-  def createDefaultComputeExecutionContext(
-      threadPrefix: String = "io-compute"): (ExecutionContext, () => Unit) = {
-    val threadCount = new AtomicInteger(0)
-    // lower-bound of 2 to prevent pathological deadlocks on virtual machines
-    val bound = math.max(2, Runtime.getRuntime().availableProcessors())
 
-    val executor = Executors.newFixedThreadPool(
-      bound,
-      { (r: Runnable) =>
-        val t = new Thread(r)
-        t.setName(s"${threadPrefix}-${threadCount.getAndIncrement()}")
-        t.setDaemon(true)
-        t
-      }
-    )
-    (ExecutionContext.fromExecutor(executor), { () => executor.shutdown() })
+  // The default compute thread pool on the JVM is now a work stealing thread pool.
+  def createDefaultComputeThreadPool(
+      self: => IORuntime,
+      threadPrefix: String = "io-compute"): (WorkStealingThreadPool, () => Unit) = {
+    val threads = math.max(2, Runtime.getRuntime().availableProcessors())
+    val threadPool =
+      new WorkStealingThreadPool(threads, threadPrefix, self)
+    (threadPool, { () => threadPool.shutdown() })
   }
 
   def createDefaultBlockingExecutionContext(
@@ -64,8 +57,8 @@ private[unsafe] abstract class IORuntimeCompanionPlatform { this: IORuntime.type
   }
 
   lazy val global: IORuntime =
-    IORuntime(
-      createDefaultComputeExecutionContext()._1,
+    new IORuntime(
+      createDefaultComputeThreadPool(global)._1,
       createDefaultBlockingExecutionContext()._1,
       createDefaultScheduler()._1,
       () => ())
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingQueue.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingQueue.scala
@@ -1,0 +1,487 @@
+/*
+ * Copyright 2020 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * This code is an adaptation of the `Local` queue from the `tokio` runtime.
+ * The original source code in Rust is licensed under the MIT license and available
+ * at: https://docs.rs/crate/tokio/0.2.22/source/src/runtime/queue.rs.
+ *
+ * For the reasoning behind the design decisions of this code, please consult:
+ * https://tokio.rs/blog/2019-10-scheduler#the-next-generation-tokio-scheduler.
+ */
+
+package cats.effect
+package unsafe
+
+import java.util.concurrent.atomic.AtomicInteger
+
+/**
+ * Fixed length, double ended, singe producer, multiple consumer queue local to a
+ * single `WorkerThread` that supports single threaded updates to the tail of the
+ * queue (to be executed **only** by the owner worker thread) and multi threaded
+ * updates to the head (local dequeueing or inter thread work stealing).
+ */
+private final class WorkStealingQueue {
+
+  import WorkStealingQueueConstants._
+
+  /**
+   * Concurrently updated by many threads.
+   *
+   * Contains two unsigned 16 bit values. The LSB bytes are the "real" head of the queue.
+   * The unsigned 16 bytes in the MSB are set by a stealer in process of stealing values.
+   * It represents the first value being stolen in the batch. Unsigned 16 bit integer is
+   * used in order to distinguish between `head == tail` and `head == tail - capacity`.
+   *
+   * When both unsigned 16 bit balues are the same, there is no active stealer.
+   *
+   * Tracking an in-progress stealer prevents a wrapping scenario.
+   */
+  private val head: AtomicInteger = new AtomicInteger()
+
+  /**
+   * Only updated by the owner worker thread, but read by many threads.
+   *
+   * Represents an unsigned 16 bit value.
+   */
+  @volatile private var tail: Int = 0
+
+  /**
+   * Holds the scheduled fibers.
+   */
+  private val buffer: Array[IOFiber[_]] = new Array(LocalQueueCapacity)
+
+  /**
+   * Returns true if there are no enqueued fibers.
+   */
+  def isEmpty(): Boolean = {
+    // Should be changed for `acquire` get operations.
+    // Requires the use of `VarHandles` (Java 9+) or Unsafe
+    // (not as portable and being phased out). Cannot use
+    // `getAcquire` and `setAcquire` either (also Java 9+).
+    val hd = lsb(head.get())
+    val tl = tail
+    hd == tl
+  }
+
+  /**
+   * Returns true if there are enqueued fibers available for stealing.
+   */
+  def isStealable(): Boolean =
+    !isEmpty()
+
+  /**
+   * Enqueues a fiber for execution at the back of this queue. Should
+   * **only** be called by the owner worker thread.
+   *
+   * There are three possible outcomes from the execution of this method:
+   * 1. There is enough free capacity in this queue, regardless if
+   *    another thread is concurrently stealing from this queue, in
+   *    which case the fiber will be pushed at the back of the queue.
+   * 2. There is not enough free capacity and some other thread is
+   *    concurrently stealing from this queue, in which case the fiber
+   *    will be enqueued on the `external` queue.
+   * 3. There is not enough free capacity in this queue and no other
+   *    thread is stealing from it, in which case, half of this queue,
+   *    including the new fiber will be spilled over and enqueued on
+   *    the `external` queue as a linked batch of fibers.
+   */
+  def enqueue(fiber: IOFiber[_], external: ExternalQueue): Unit = {
+    // Should be a `plain` get.
+    // Requires the use of `VarHandles` (Java 9+) or Unsafe
+    // (not as portable and being phased out).
+    val tl = tail
+
+    var cont = true
+    while (cont) {
+      // Should be an `acquire` get.
+      // Requires the use of `VarHandles` (Java 9+) or Unsafe
+      // (not as portable and being phased out).
+      val hd = head.get()
+
+      val steal = msb(hd) // Check if a thread is concurrently stealing from this queue.
+      val real = lsb(hd) // Obtain the real head of the queue.
+
+      if (unsignedShortSubtraction(tl, steal) < LocalQueueCapacity) {
+        // There is free capacity for the fiber, proceed to enqueue it.
+        cont = false
+      } else if (steal != real) {
+        // Another thread is concurrently stealing, so push the new
+        // fiber onto the external queue and return.
+        external.enqueue(fiber)
+        return
+      } else {
+        // There is no free capacity for the fiber and no one else is stealing from this queue.
+        // Overflow half of this queue and the new fiber into the external queue.
+        if (overflowToExternal(fiber, real, external)) {
+          return
+        }
+
+        // Failed to enqueue the back half of this queue on the external queue. Retrying.
+      }
+    }
+
+    // Enqueue the new fiber for later execution.
+
+    // Map the position to a slot index.
+    val idx = tl & CapacityMask
+    // Write the fiber to the slot.
+    buffer(idx) = fiber
+    // Make the fiber available.
+
+    // Should be a `release` set.
+    // Requires the use of `VarHandles` (Java 9+) or Unsafe
+    // (not as portable and being phased out).
+    tail = unsignedShortAddition(tl, 1)
+  }
+
+  private[this] def overflowToExternal(
+      fiber: IOFiber[_],
+      hd: Int,
+      external: ExternalQueue): Boolean = {
+    val prev = pack(hd, hd)
+
+    // Claim half of the fibers.
+    // We are claiming the fibers **before** reading them out of the buffer.
+    // This is safe because only the **current** thread is able to push new
+    // fibers.
+    val headPlusHalf = unsignedShortAddition(hd, HalfLocalQueueCapacity)
+    if (!head.compareAndSet(prev, pack(headPlusHalf, headPlusHalf))) {
+      // We failed to claim the fibers, losing the race. Return out of
+      // this function and try the full `enqueue` method again. The queue
+      // may not be full anymore.
+      return false
+    }
+
+    // We have successfully claimed the fibers. Continue with linking them
+    // and moving them to the external queue.
+
+    var i = 0 // loop counter
+    var j = 0 // one after the loop counter
+    var iIdx = 0 // mapped index
+    var jIdx = 0 // mapped next index
+    var next: IOFiber[_] = null // mutable fiber variable used for linking fibers in a batch
+    while (i < HalfLocalQueueCapacity) {
+      j = i + 1
+
+      // Map the loop counters to indices.
+      iIdx = unsignedShortAddition(i, hd) & CapacityMask
+      jIdx = unsignedShortAddition(j, hd) & CapacityMask
+
+      // Determine the next fiber to be linked in the batch.
+      next = if (j == HalfLocalQueueCapacity) {
+        // The last fiber in the local queue is being moved.
+        // Therefore, we should attach the new fiber after it.
+        fiber
+      } else {
+        // A fiber in the middle of the local queue is being moved.
+        buffer(jIdx)
+      }
+
+      // Create a linked list of fibers.
+      buffer(iIdx).next = next
+      i += 1
+    }
+
+    // Map the new head of the queue to an index in the queue.
+    val hdIdx = hd & CapacityMask
+
+    // Claim the head fiber of the linked batch.
+    val hdFiber = buffer(hdIdx)
+
+    // Push the fibers onto the external queue starting from the head fiber
+    // and ending with the new fiber.
+    external.enqueueBatch(hdFiber, fiber)
+    true
+  }
+
+  /**
+   * Dequeue a fiber from the local queue. Returns `null` if the queue is empty.
+   * Should **only** be called by the owner worker thread.
+   */
+  def dequeueLocally(): IOFiber[_] = {
+    // Should be an `acquire` get.
+    // Requires the use of `VarHandles` (Java 9+) or Unsafe
+    // (not as portable and being phased out).
+    var hd = head.get()
+    var idx = 0 // will contain the index of the fiber to be dequeued
+    var steal = 0 // will contain a concurrent stealer
+    var real = 0 // will contain the real head of the queue
+    var nextReal = 0 // one after the real head of the queue
+    var nextHead = 0 // will contain the next full head
+
+    // Should be a `plain` get.
+    // Requires the use of `VarHandles` (Java 9+) or Unsafe
+    // (not as portable and being phased out).
+    val tl = tail
+
+    var cont = true
+    while (cont) {
+      steal = msb(hd) // Check if a thread is concurrently stealing from this queue.
+      real = lsb(hd) // Obtain the real head of the queue.
+
+      if (real == tl) {
+        // The queue is empty. There is nothing to be dequeued. Return.
+        return null
+      }
+
+      nextReal = unsignedShortAddition(real, 1)
+
+      nextHead = if (steal == real) {
+        // There are no concurrent threads stealing from this queue. Both `steal` and `real`
+        // values should be updated.
+        pack(nextReal, nextReal)
+      } else {
+        // There is a thread concurrently stealing from this queue. Do not mess with its
+        // steal tag value, only update the real head.
+        pack(steal, nextReal)
+      }
+
+      // Attempt to claim a fiber.
+      if (head.compareAndSet(hd, nextHead)) {
+        // Successfully claimed the fiber to be dequeued.
+        // Map to its index and break out of the loop.
+        idx = real & CapacityMask
+        cont = false
+      } else {
+        // Failed to claim the fiber to be dequeued. Retry.
+
+        // Should be an `acquire` get.
+        // Requires the use of `VarHandles` (Java 9+) or Unsafe
+        // (not as portable and being phased out).
+        hd = head.get()
+      }
+    }
+
+    // Dequeue the claimed fiber.
+    buffer(idx)
+  }
+
+  /**
+   * Steal half of the enqueued fibers from this queue and place them into
+   * the destination `WorkStealingQueue`. Executed by a concurrent thread
+   * which owns `dst`. Returns the first fiber to be executed.
+   */
+  def stealInto(dst: WorkStealingQueue): IOFiber[_] = {
+    // Should be a `plain` get.
+    // Requires the use of `VarHandles` (Java 9+) or Unsafe
+    // (not as portable and being phased out).
+    val dstTail = dst.tail
+
+    // To the caller, `dst` may **look** empty but still have values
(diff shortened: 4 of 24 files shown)


core/jvm/src/main/scala/cats/effect/IOFiberPlatform.scala
```diff
@@ -26,6 +26,10 @@ private[effect] abstract class IOFiberPlatform[A] { this: IOFiber[A] =>
 
   private[this] val TypeInterruptibleMany = Sync.Type.InterruptibleMany
 
+  // Allocation free linked queue nodes of `IOFiber[_]` objects.
+  // Represents the external work stealing thread pool queue.
+  private[effect] var next: IOFiber[_] = _
```
djspiewak, line 31: You know, in retrospect, this is a really obvious idea. :-) Nice!

core/jvm/src/main/scala/cats/effect/unsafe/ExternalQueue.scala
```diff
@@ -0,0 +1,180 @@
...
+ */
+
+/*
+ * This code is an adaptation of the `Inject` queue from the `tokio` runtime.
+ * The original source code in Rust is licensed under the MIT license and available
+ * at: https://docs.rs/crate/tokio/0.2.22/source/src/runtime/queue.rs.
+ *
+ * For the reasoning behind the design decisions of this code, please consult:
+ * https://tokio.rs/blog/2019-10-scheduler#the-next-generation-tokio-scheduler.
+ */
```
djspiewak, line 24: 👍

core/jvm/src/main/scala/cats/effect/unsafe/ExternalQueue.scala
```diff
@@ -0,0 +1,180 @@
...
+ *
+ * New fibers scheduled by external threads end up at the back of this queue.
+ * When a local worker queue overflows (cannot accept new fibers), half of it
+ * gets spilled over to this queue.
+ */
+private final class ExternalQueue {
+  import WorkStealingQueueConstants._
+
+  // Lock that ensures exclusive access to the internal state of the linked queue.
+  private[this] val lock: AnyRef = new Object()
```
djspiewak, line 42: Is this faster than `ReentrantLock`?

core/jvm/src/main/scala/cats/effect/unsafe/ExternalQueue.scala
```diff
@@ -0,0 +1,180 @@
...
+
+      // Set the tail to point to the new fiber.
+      tail = fiber
+
+      // Should be changed for a `plain` get and `release` set.
+      // Requires the use of `VarHandles` (Java 9+) or Unsafe
+      // (not as portable and being phased out).
+      len += 1
+    }
+  }
```
djspiewak, line 85: I strongly suspect this would be much faster under contention without the lock. Would need to think about it, but should be doable with just three atomic references. We're only allocating these once ever, so the cost is relatively minimal.

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
```diff
@@ -22,22 +22,15 @@ import java.util.concurrent.Executors
...
-        t.setName(s"${threadPrefix}-${threadCount.getAndIncrement()}")
-        t.setDaemon(true)
-        t
-      }
-    )
-    (ExecutionContext.fromExecutor(executor), { () => executor.shutdown() })
+  // The default compute thread pool on the JVM is now a work stealing thread pool.
+  def createDefaultComputeThreadPool(
+      self: => IORuntime,
+      threadPrefix: String = "io-compute"): (WorkStealingThreadPool, () => Unit) = {
```
djspiewak, line 29: 👍

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingQueue.scala
```diff
@@ -0,0 +1,487 @@
...
+   * Unsigned 16 bit addition.
+   */
+  private[this] def unsignedShortAddition(x: Int, y: Int): Int =
+    lsb(x + y)
+
+  /**
+   * Unsigned 16 bit subtraction.
+   */
+  private[this] def unsignedShortSubtraction(x: Int, y: Int): Int =
+    lsb(x - y)
```
djspiewak, line 486: Can we just assume that we're not going to overflow? That would allow us to remove the extra mask.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -0,0 +1,358 @@
...
+  private[this] val externalQueue: ExternalQueue = new ExternalQueue()
+
+  // Represents two unsigned 16 bit integers.
+  // The 16 most significant bits track the number of active (unparked) worker threads.
+  // The 16 least significant bits track the number of worker threads that are searching
+  // for work to steal from other worker threads.
+  private[this] val state: AtomicInteger = new AtomicInteger(threadCount << UnparkShift)
+
+  // Lock that ensures exclusive access to the references of sleeping threads.
+  private[this] val lock: AnyRef = new Object()
```
djspiewak, line 74: Another potential use-case for `ReentrantLock`

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -0,0 +1,358 @@
...
+
+  // LIFO access to references of sleeping worker threads.
+  private[this] val sleepers: ArrayStack[WorkerThread] = new ArrayStack(threadCount)
```
djspiewak, line 77: Random note: `ArrayStack` isn't thread-safe.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -0,0 +1,358 @@
...
+
+  // Track currently sleeping threads by their indices.
+  private[this] val sleepingSet: Array[Boolean] = new Array[Boolean](threadCount)
+
+  // Full memory barrier for publishing the shutdown signal of each worker thread.
+  // That way each worker thread can avoid a volatile variable read on each iteration
+  // of the run loop.
+  @volatile private[this] var done: Boolean = false
```
djspiewak: I'm not sure it works this way. The problem is that the workers threads are not *themselves* reading `done`, are they? I *think* that might be okay though, since the only consequence is that shutdown is not immediately visible.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -0,0 +1,358 @@
...
+  }
+
+  /**
+   * Tries to steal work from another worker thread. This method does a linear search of the
+   * worker threads starting at a random index.
+   */
+  private[unsafe] def stealFromOtherWorkerThread(thread: WorkerThread): IOFiber[_] = {
+    var i = 0
+    // Use an uncontended source of randomness.
+    val from = ThreadLocalRandom.current().nextInt(threadCount)
```
djspiewak: Thread locals are very slow! Rather than using a thread local, I think it would be better to actually store a `Random` *on each `WorkerThread`*, and then use it here.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -0,0 +1,358 @@
...
+   * Searches for a parked thread to notify of arrived work.
+   */
+  private[this] def workerToNotify(): WorkerThread = {
+    if (!notifyShouldWakeup()) {
+      // Fast path, no locking, there are enough searching and/or running worker threads.
+      // No need to wake up more. Return.
+      return null
+    }
+
+    lock.synchronized {
```
djspiewak, line 172: The use of locks continues to scare me…

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -0,0 +1,275 @@
...
+  /**
+   * Obtain a fiber either from the local queue or the external queue,
+   * if the time has come to check for work from there. Returns `null`
+   * when no fiber is available.
+   */
+  private[this] def nextFiber(): IOFiber[_] = {
+    var fiber: IOFiber[_] = null
+
+    // Magic heuristic first seen in the golang runtime.
+    if (tick % 61 == 0) {
```
djspiewak: What's special about multiples of 61?

core/jvm/src/test/scala/cats/effect/concurrent/DeferredJVMSpec.scala
```diff
@@ -77,57 +79,31 @@ abstract class BaseDeferredJVMTests(parallelism: Int)
...
-        _ <- df.complete(())
-        _ <- IO(Thread.currentThread().getName mustEqual name)
-        _ <- fb.join
-      } yield ()
-
-      task.unsafeRunTimed(timeout).nonEmpty must beTrue
-    }
-
-    success
-  }
```
djspiewak, line 104: NB this test was removed because it's literally asserting things about scheduling details. It's a bad test.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -682,14 +683,36 @@ private final class IOFiber[A](
...
+  private[this] def execute(ec: ExecutionContext)(fiber: IOFiber[_]): Unit = {
     readBarrier()
 
-    try {
-      ec.execute(action)
-    } catch {
-      case _: RejectedExecutionException =>
-        () // swallow this exception, since it means we're being externally murdered, so we should just... drop the runloop
+    if (ec.isInstanceOf[WorkStealingThreadPool]) {
+      ec.asInstanceOf[WorkStealingThreadPool].executeFiber(fiber)
```
djspiewak, line 690: IIRC, HotSpot has some special logic to detect this particular pattern, so I imagine this will be very fast.

djspiewak:
My main concern right now is surrounding the use of `synchronized`. Everything else looks quite solid and clever!

core/jvm/src/main/scala/cats/effect/unsafe/ExternalQueue.scala
vasilmkd: I have never used a `ReentrantLock`. I'll change and benchmark.

core/jvm/src/main/scala/cats/effect/unsafe/ExternalQueue.scala
vasilmkd: I'm looking for help on this one. `tokio` uses locks in the same places as in this PR, I was mainly concerned with getting things working correctly first, then reiterating for optimizations after a review.

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
vasilmkd: Nice trick indeed. Thanks.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingQueue.scala
vasilmkd: I would argue that that is a premature optimization. I'm sure there's more performance to be gained elsewhere. It's also a bitwise operation, which has already been optimized to entropy.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
vasilmkd: I'll rewrite and benchmark.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
vasilmkd: It's only ever used under the lock.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
vasilmkd: So, before this, the `WorkerThread` run loop was a `while (!pool.done)` which I dislike as it's a volatile read. Now, each thread has a normal `done` variable (not volatile). The idea here is, after setting each of the individual `WorkerThread` non-volatile `done` variables to `true`, this would act as a volatile operation that would publish them, which should be picked up by the worker threads? I might be wrong though.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
vasilmkd: `ThreadLocalRandom` is used by `ForkJoinPool` in `openjdk`. Also the javadoc seemed fairly reasonable at the time. I can change it.
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadLocalRandom.html

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
vasilmkd: This is more or less directly translated from `tokio`.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
vasilmkd: It's a constant that provided the "right" throughput and latency for local vs. external queue polling. `tokio` references the `golang` runtime as the original inspiration.

core/shared/src/main/scala/cats/effect/IOFiber.scala
vasilmkd: Seems like it, it's much faster than the virtual dispatch solution from earlier.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
vasilmkd: I'm more in favor of using 64 and bitwise operations.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
djspiewak: > which should be picked up by the worker threads

Only if the workers themselves *read* from the `done` variable. Memory barriers aren't a gobal thing (at least, not guaranteed to be such). Instead, all writes performed prior to a write barrier on variable *x* are guaranteed to be visible to a thread who subsequently crosses a read barrier on variable *x*, but it doesn't say anything about some other variable.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
djspiewak: Ah! Didn't realize that was a thing. Doing the solution I proposed with `WorkerThread` is *definitely* faster; I'm surprised they never made that kind of a change.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
djspiewak: That's really, really, really interesting to me. Do they have any specific discussion about the use of locks rather than a lock-free encoding?

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
djspiewak: I mean, that seems… better… to me. :-) At the very least less arbitrary and faster.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
vasilmkd: Nothing that I could find.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
johnynek: be careful. 61 is prime. That is what is going on here I think.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
johnynek: I seem to remember seeing something about this in the past that lock free complex data structures often don't win. When you are in the contended state you are redoing a lot of work due to transaction failures, and in the uncontended case there is no win from concurrency anyway (because there is only one or two items).

I'd be reluctant to second guess their lock-based design without evidence across a variety of loads that there isn't a catastrophic case. I'd fear the wheels could really come off this under load if we went lock free: exactly when the contention increases that causes the throughput to decrease, which causes death-spiral-like behavior in some cases.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
johnynek: Though, I don't really know why that should matter if we are just looking every n-th time. It's not like we are mixing another entropy source in and trying to leverage the field of 61 elements.

That said, I'd be really surprised if performance was much better by doing `tick & 63 == 0`.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
djspiewak: Extremely valid points. And I've seen similar work on throughput of locks vs optimistic encodings under contention. As I recall, once the contention ratios start getting to be meaningful, locks do in fact win. Perhaps this is one of those cases. 

At the very least, locks on the JVM aren't worse than in Rust, and they chose them intentionally.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
djspiewak: Mask definitely pipelines a lot better than modulo. It's a good point that 61 is prime, but like you said, they aren't really using that in any way. It's just… ticking. And every 61st cycle they fire. This honestly strikes me as something that has been cargo-culted around without much justification, but perhaps that's naive.

core/jvm/src/main/scala/cats/effect/unsafe/ExternalQueue.scala
RaasAhsan: `ReentrantLock` implements `AbstractQueuedSynchronizer` which relies on thread parking and unparking to achieve mutual exclusion rather than monitors, so it should be a lot cheaper.

RaasAhsan: It would be cool if we could get rid of the lock altogether, but that's dangerous and wishful thinking :D

core/jvm/src/main/scala/cats/effect/IOFiberPlatform.scala
RaasAhsan: I wonder if there are any concerns around false sharing if we do this

core/jvm/src/main/scala/cats/effect/unsafe/ExternalQueue.scala
```diff
@@ -0,0 +1,180 @@
...
+   * Enqueues a fiber for later execution at the back of the queue.
+   */
+  def enqueue(fiber: IOFiber[_]): Unit = {
+    lock.synchronized {
+      if (_shutdown) {
+        // Do not accept new fibers if the queue has been shut down.
+        return
+      }
+
+      // Safe to mutate the internal state because we are holding the lock.
```
RaasAhsan, line 68: I think a lot of the comments here are redundant and follow clearly from local context

core/jvm/src/main/scala/cats/effect/unsafe/ExternalQueue.scala
```diff
@@ -0,0 +1,180 @@
...
+   */
+  def shutdown(): Unit = {
+    lock.synchronized {
+      if (_shutdown) {
+        // The queue has already been shutdown. Return.
+        return
+      }
+
+      // Unlink and drain the queue.
+      var fiber: IOFiber[_] = head
```
RaasAhsan, line 168: Is it sufficient to just set `head` to `null` here? From a GC standpoint, none of the elements in the queue would be reachable via `head`, so all elements would be eligible for collection as long as there aren't any other outstanding, reachable references to elements in the queue.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
RaasAhsan: Each platform/architecture has its own precise definitions around what a memory barrier is, what the effects of reading and writing to them are, what kind of optimizations a CPU has the liberty to take around them, and so on. Since we're running on the JVM, we can only reasonably rely on the guarantees of the Java memory model, which is what Daniel described. IIRC, the Java memory model doesn't even define what a memory barrier is, instead the memory semantics are defined in terms of reads and writes to volatile variables.

With that terminology, even if you publish the writes to the non-volatile `done`s by writing to this volatile `done`, the worker threads aren't actually guaranteed to see those writes to their individual `done`s *unless* it observes a write to the volatile `done`, which is only possible with a read from it.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
RaasAhsan: What's the rationale for picking a random index instead of something like `n + 1`, which you did in your previous PR? If we can assume that tasks are uniformly distributed among worker threads and that task completion times are normally distributed (I don't know how fair these assumptions are), it seems reasonable to say that every thread has the same probability of having work to steal. Maybe we'd see different results with different kinds of workloads.

Daniel brought up concerns around a linear search in the last PR, so I'm wondering if there are alternative solutions here. A constant-time solution that comes to mind is that each thread is responsible for stealing from the next k threads, where k << threadCount.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -0,0 +1,358 @@
...
+          // Successful steal. Return the next fiber to be executed.
+          return res
+        }
+      }
+
+      i += 1
+    }
+
+    // The worker thread could not steal any work. Fall back to checking the external queue.
+    externalQueue.dequeue()
```
RaasAhsan: `externalDequeue()` ?

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
RaasAhsan: I think it's worth commenting somewhere that the purpose of this is to avoid starvation of tasks on the external queue

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -0,0 +1,275 @@
...
+        // Try to steal work from other worker threads.
+        fiber = stealWork()
+      }
+
+      if (fiber == null) {
+        // No fiber has been stolen from any other worker thread.
+        // There is probably not enough work for every thread in
+        // the pool. It's time to park and await a notification
+        // when new work is submitted to the pool.
+        park()
```
RaasAhsan, line 271: It's worth pointing out that this can spuriously return, but I think that's fine since it doesn't violate any correctness invariant.

RaasAhsan (commented):
Finally got around to a first pass, things are looking hopeful :) I will admit that the amount of comments  is kind of overwhelming especially given that a great deal of them are redundant

2nd pass coming soon! i imagine there are going to be a ton because of the sheer amount of concurrent code involved

core/jvm/src/main/scala/cats/effect/unsafe/ExternalQueue.scala
vasilmkd: Unfortunately I couldn't get `ReentrantLock` to work. I did use it in the prescribed way of locking and unlocking in a `try/finally` block, but it deadlocks. This happens in `WorkStealingThreadPool#workerToNotify()`.

core/jvm/src/main/scala/cats/effect/unsafe/ExternalQueue.scala
vasilmkd: Yeah, I got carried away, maybe this class doesn't need comments.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
vasilmkd: There is a check in place for that. `transitionFromParked()` in the same class.

core/jvm/src/main/scala/cats/effect/unsafe/ExternalQueue.scala
vasilmkd: They shouldn't be reachable from anywhere else.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
vasilmkd: Fixed.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
vasilmkd: Done.

vasilmkd:
It looks like there's an issue with the configuration of GitHub Actions. I'll be running benchmarks and posting results later.

These benchmarks show the `volatile done` on `WorkStealingThreadPool` and 64 ticks to check the external queue. Personally I'd like to go for a compromise between checking the `volatile done` every `n` ticks.

New benchmark results:
```
Benchmark                                   (size)   Mode  Cnt       Score      Error    Units
AsyncBenchmark.async                           100  thrpt   20   22135.726 ±  273.204    ops/s
AsyncBenchmark.bracket                         100  thrpt   20   47822.108 ±  209.607    ops/s
AsyncBenchmark.cancelable                      100  thrpt   20   22043.520 ±  296.384    ops/s
AsyncBenchmark.race                            100  thrpt   20    4574.195 ±   49.872    ops/s
AsyncBenchmark.racePair                        100  thrpt   20   18103.404 ±  861.356    ops/s
AsyncBenchmark.start                           100  thrpt   20    6674.030 ±  193.112    ops/s
AsyncBenchmark.uncancelable                    100  thrpt   20  221844.829 ± 1710.424    ops/s
AttemptBenchmark.errorRaised                 10000  thrpt   20    1763.056 ±   11.005    ops/s
AttemptBenchmark.happyPath                   10000  thrpt   20    2511.296 ±   16.053    ops/s
DeepBindBenchmark.async                      10000  thrpt   20     493.671 ±   17.008    ops/s
DeepBindBenchmark.delay                      10000  thrpt   20    3797.610 ±   54.301    ops/s
DeepBindBenchmark.pure                       10000  thrpt   20    4616.667 ±   88.885    ops/s
HandleErrorBenchmark.errorRaised             10000  thrpt   20    1973.186 ±    3.473    ops/s
HandleErrorBenchmark.happyPath               10000  thrpt   20    2840.271 ±   28.494    ops/s
MapCallsBenchmark.batch120                     N/A  thrpt   20    5357.437 ±  106.185    ops/s
MapCallsBenchmark.batch30                      N/A  thrpt   20    3978.579 ±   42.870    ops/s
MapCallsBenchmark.one                          N/A  thrpt   20     265.222 ±    6.532    ops/s
MapStreamBenchmark.batch120                    N/A  thrpt   20    3567.266 ±   13.917    ops/s
MapStreamBenchmark.batch30                     N/A  thrpt   20    1391.849 ±   33.200    ops/s
MapStreamBenchmark.one                         N/A  thrpt   20    1502.303 ±    8.489    ops/s
RedeemBenchmark.errorRaised                  10000  thrpt   20    1241.739 ±   85.532    ops/s
RedeemBenchmark.happyPath                    10000  thrpt   20    1587.175 ±    3.653    ops/s
RedeemWithBenchmark.errorRaised              10000  thrpt   20    1599.029 ±   21.646    ops/s
RedeemWithBenchmark.happyPath                10000  thrpt   20    2239.390 ±    9.427    ops/s
ShallowBindBenchmark.async                   10000  thrpt   20     481.053 ±   10.821    ops/s
ShallowBindBenchmark.delay                   10000  thrpt   20    4150.804 ±   22.487    ops/s
ShallowBindBenchmark.pure                    10000  thrpt   20    5002.289 ±    3.475    ops/s
WorkStealingBenchmark.async                1000000  thrpt   20       4.649 ±    0.206  ops/min
WorkStealingBenchmark.asyncTooManyThreads  1000000  thrpt   20       7.299 ±    0.470  ops/min
```

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
vasilmkd: I think the rationale behind the linear search is the fact that at most half of the available threads are searching at any time, and any thread can be searching or not. Also, in case new work arrives, only a single thread is woken up to try to find the work. If we go with each thread is responsible for the next `k` threads, that work might be missed and that particular thread going through a large batch of fibers would be overbooked and remain without help.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
vasilmkd: They have a slightly nicer API surface in Rust. Mutexes are parameterized and they "wrap" the state that they protect. But otherwise yes, performance should be exactly the same.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
vasilmkd: I measured it, performance seems to be the same. There is a slight dip in the results, which I attribute to reading volatile state from the pool. Could we try reading the state every `n` iterations?

core/jvm/src/main/scala/cats/effect/IOFiberPlatform.scala
vasilmkd: This ties to using a non-locking `ExternalQueue`:

We could make `IOFiberPlatform` here extend `AtomicReference[IOFiber[_]]` so that we can go with an implementation similar to `java.util.concurrent.ConcurrentLinkedQueue` from the standard library.

Is anyone against measuring this?
@RaasAhsan @johnynek @djspiewak

vasilmkd:
Compiling a summary of the reviewers' current (completely valid) objections:

@djspiewak was mostly concerned about the `volatile done` on `WorkStealingThreadPool`. That part is fixed. Speaking privately, I believe they would like to see locks gone by trying the `extends AtomicReference[IOFiber[_]]` and making `ExternalQueue` use CAS instead of locking, similar to `java.util.concurrent.ConcurrentLinkedQueue`. Unfortunately, we don't seem to have a solution for the lock in `WorkStealingThreadPool`. I tried using a `ReentrantLock`, but it didn't go well. It was very late last night when I tried it, I might have made a mistake, I will try again, but I do not have high hopes that anything will change.

On the other hand, @RaasAhsan seems to think that making `IOFiber` inherently linkable in any way might create false sharing. Do you have an alternative that you'd like to see explored? My personal reasoning behind this was, we control the data type of the "task" type, so we should take advantage of every opportunity to choose and implement our own data structures that complement that task type, which gives us the ability to avoid allocating queue node objects, which I believe is a non-trivial difference when a huge number of fibers is created (which I also believe to be the main workload that this change is trying to support). Regarding unlinking the whole external queue on shutdown, I did that for my own peace of mind, I am not against simply unlinking the `head` and `tail` pointers, so that can be fixed if we do decide to keep the linking as an implementation detail of `IOFiber`.

Finally, @johnynek seems to be in favor of trusting the `tokio` implementors for now and stick with their locking implementation for now. Any other ideas or objections, I would love to hear them and experiment.

I'm sorry if I have misunderstood anyone's thoughts, wishes or ideas. I'm creating this comment to try to summarize and centralize the discussion a bit. To me, it seems that we have forked out many little comments without a clear direction on any of them.

Please correct any mistake I've made in this comment. I only wish to proceed the discussion in the most focused and distillable way forward, and have a sort of unified front. Thanks everyone for their time and comments. They are extremely appreciated.

djspiewak:
Btw I fixed the github actions issues if you merge with series/3.x

djspiewak:
> What's the rationale for picking a random index instead of something like n + 1, which you did in your previous PR? If we can assume that tasks are uniformly distributed among worker threads and that task completion times are normally distributed (I don't know how fair these assumptions are), it seems reasonable to say that every thread has the same probability of having work to steal. Maybe we'd see different results with different kinds of workloads.

@RaasAhsan I don't believe these are fair assumptions. I think it's more likely that we'll see normal task distributions since `start` and `racePair` attempt to reuse the same worker (I think?). Task loads tend to be pretty spiky in most applications, so I suspect that we'll see a general overload of certain threads. If we only steal within *k* threads of the hypothetically overloaded thread, then we're going to confine threads to that *region* of the mod ring.

In theory, every thread starting from its current location and linearly scanning forward around the ring achieves part of what we need: specifically, all threads have an equal probability of being stolen from assuming a perfectly uniform distribution of tasks. However, the other side of that equation is also important: given a thread *t*, what is the probability that it steals *from* some other thread *t'*? With a linear scan starting from the current index, that probability itself becomes a binomial distribution which decays with positive distance between *t* and *t'*. This in turn means that a single overloaded thread will overload its left-region and it will take a lot of theft to spread the work across the rest of the pool, which in turn goes back and violates our assumption of uniform distribution (since it's now binomial, matching the probability of original theft).

It's important to remember that *most* of our tasks are going to be submitted from within `IOFiber`, rather than extrinsic, meaning that most tasks enter the pool via theft rather than entering via the external queue (which should exhibit more uniform properties).

As a note, fixed-distance theft would make this problem worse, since suddenly the probability of a thread *t_n* stealing from some other thread *t_m* goes to zero if *m > n + k*, which we definitely don't want.

Using a random offset to each theft *should* fix the binomial distribution problem and restore the assumption of normal task distribution within the mod ring.

djspiewak:
> Speaking privately, I believe they would like to see locks gone by trying the extends AtomicReference[IOFiber[_]] and making ExternalQueue use CAS instead of locking, similar to java.util.concurrent.ConcurrentLinkedQueue. Unfortunately, we don't seem to have a solution for the lock in WorkStealingThreadPool

I'm pretty convinced we should *start* with the locking approach you're using in this PR, then experiment with CAS-based approaches later, ideally when we have even more contention benchmarks. @johnynek is 100% correct that optimistic concurrency falls behind pessimistic under contention, and we should set the empiricism bar high before jumping on the CAS train.

> I tried using a ReentrantLock, but it didn't go well. It was very late last night when I tried it, I might have made a mistake, I will try again, but I do not have high hopes that anything will change.

Let's get this into series/3.x *first*, and then try to use `ReentrantLock`. I still suspect it will produce a nice performance boost, and it *should* have the same semantics as `synchronized`, but it's not worth holding up this work to sort all that out. Let's land this sooner rather than later, then refine!

> My personal reasoning behind this was, we control the data type of the "task" type, so we should take advantage of every opportunity to choose and implement our own data structures that complement that task type, which gives us the ability to avoid allocating queue node objects, which I believe is a non-trivial difference when a huge number of fibers is created (which I also believe to be the main workload that this change is trying to support).

I agree with this. I think that we should probably take some time to carefully reason about the intrinsic linking, but on my first pass through it, things honestly seemed entirely sound. The locking saves us from a lot of weirdness, tbh.

> Finally, @johnynek seems to be in favor of trusting the tokio implementors for now and stick with their locking implementation for now. Any other ideas or objections, I would love to hear them and experiment.

I agree with this, outside of the "64 rather than 61" change, which just seems like… sanity.

> Please correct any mistake I've made in this comment. I only wish to proceed the discussion in the most focused and distillable way forward, and have a sort of unified front. Thanks everyone for their time and comments. They are extremely appreciated.

Thank you for taking the time to put this list together! This kind of thing makes it much, much easier for people to weigh-in in one place (here) and move things forward. You've done a tremendous job on this, and it feels close to being able to land in series/3.x. Would love to hear more from @RaasAhsan and @johnynek on the points I made here and in the previous comment, as well as any other replies to your list, but things are looking very good overall.

PR #1137, opened 2020-08-26
MonadCancel

### LINKED ISSUE #1016: MonadCancel: New Typeclass Above Concurrent
This would have `forceR`, `canceled`, `uncancelable`, and `onCancel`. The purpose to this is it generalizes things very slightly. More specifically, you can define this class for `StateT` and `RWST` (whereas you cannot define a sane `Concurrent` for these). This definition has value, too, since it allows `State`-using code to participate in cancelation semantics of some underlying `Concurrent` type, which is impossible at present.

In a sense, this would be the conceptual replacement for `Bracket`, and indeed, `bracket` would be a derived combinator on this typeclass.

djspiewak:
Conceptually, this is something like a direct replacement for CE2's `Bracket`, and I suspect any upgrade scalafix will assume such. The main difference is that cancelation *as a concept* is extracted from `Concurrent` and moved up into this class. There are two motivations for this change: practical and conceptual.

## Practical Motivation

`StateT` and `ReaderWriterStateT` (and also perhaps `ContT`; I haven't tried yet) form a lawful `MonadCancel`, but neither can form a lawful `Concurrent` (well, *possibly* lawful, but certainly not sane). The fact that they can't form a valid `Concurrent` is quite problematic, because it means that `StateT` cannot *participate* in cancelation-safe code, even when that code itself is fully sequential. This in turn all-but forces people to write unsafe code in areas involving these types, unsafety which would only be exposed when such code is in turn composed with other, `Concurrent`-forming base types.

All in all, it corrupts composability in a very subtle and dangerous way. Thus, feel that the conceptual weight of the added abstraction is justified. As an additional aside, there are many unforeseen third-party types which are isomorphic to `StateT`, and those types would have a similar problem. Thus, presenting this as a full abstraction opens the hierarchy to safe composition with those types, whereas we would otherwise have no options there.

## Conceptual Motivation

There's nothing about cancelation which is *fundamentally* concurrent, given that we have `canceled`. Technically, cancelation probably doesn't have that many practical applications absent concurrency, but that's an argument by lack of imagination. With this change, we can view the (simplified) functor hierarchy through the following lens:

- `Functor` – Contains thing(s) of type `A`
- `Applicative` – Static pairing of `F[A]`s
- `Monad` – Runtime sequentialization of `F[A]`s
- `MonadError` – Recoverable short-circuiting of sequential chains given a sentinel value
- `MonadCancel` – Unrecoverable (but maskable) short-circuiting of sequential chains without a sentinel
- `Concurrent` – Generalize sequential chains to sequential edge-restricted DAGs
- `Allocate` – Generalize chains to arbitrary graphs
- `Async` – Constructor capture of side-effects
- `Effect` – Functional effect type (`IO`-like)

I find this rather pleasing. It simplifies the definition of `Concurrent` down to specifically control-flow, analogous to how `Monad` is itself just a control flow mechanism (characterizing runtime sequentialization).

Fixes #1016

### DIFF
--- a/core/shared/src/main/scala/cats/effect/package.scala
+++ b/core/shared/src/main/scala/cats/effect/package.scala
@@ -23,6 +23,9 @@ package object effect {
   type Outcome[F[_], E, A] = cekernel.Outcome[F, E, A]
   val Outcome = cekernel.Outcome
 
+  type MonadCancel[F[_], E] = cekernel.MonadCancel[F, E]
+  val MonadCancel = cekernel.MonadCancel
+
   type Concurrent[F[_], E] = cekernel.Concurrent[F, E]
   val Concurrent = cekernel.Concurrent
 
@@ -47,6 +50,7 @@ package object effect {
   type Effect[F[_]] = cekernel.Effect[F]
   val Effect = cekernel.Effect
 
+  type MonadCancelThrow[F[_]] = cekernel.MonadCancelThrow[F]
   type ConcurrentThrow[F[_]] = cekernel.ConcurrentThrow[F]
   type TemporalThrow[F[_]] = cekernel.TemporalThrow[F]
 
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Concurrent.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Concurrent.scala
@@ -16,52 +16,14 @@
 
 package cats.effect.kernel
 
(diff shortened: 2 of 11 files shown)


djspiewak:
BTW, this PR increases our CI build times to about 16 minutes. That doesn't seem too bad, but it's definitely a bloating. Property testing RWST is *really* slow because of the the use of the exhaustive `Eq` on `Function2`.

PR #1138, opened 2020-08-27
IOApp Rework

ChristopherDavenport:
- Keeps IOApp consistent ce2 source
- IOMinApp drops all args and returns IO[Unit]
- Should IOMinApp be in its own file for easier visibility. (Its already being defined in a platform specific file so uncertain if that aids much)

### DIFF
--- a/core/js/src/main/scala/cats/effect/IOApp.scala
+++ b/core/js/src/main/scala/cats/effect/IOApp.scala
@@ -19,9 +19,17 @@ package cats.effect
 import scala.concurrent.duration._
 import scala.scalajs.js
 
-trait IOApp {
+trait IOApp extends IOFullApp {
+  def run: IO[Unit]
+  def runFull = run.as(0)
+}
 
-  def run(args: List[String]): IO[Int]
+trait IOFullApp {
+
+  private var margs : List[String] = List.empty
+  lazy val args = margs
+
+  def runFull: IO[Int]
 
   protected val runtime: unsafe.IORuntime = unsafe.IORuntime.global
 
@@ -44,7 +52,9 @@ trait IOApp {
       else
         args.toList
 
-    IO.race(run(argList), keepAlive)
+    margs = argList
+
+    IO.race(runFull, keepAlive)
       .unsafeRunAsync({
         case Left(t) => throw t
         case Right(Left(code)) => reportExitCode(code)
--- a/core/js/src/main/scala/cats/effect/IOApp.scala
+++ b/core/js/src/main/scala/cats/effect/IOApp.scala
@@ -26,7 +26,7 @@ trait IOApp extends IOFullApp {
 
 trait IOFullApp {
 
-  private var margs : List[String] = List.empty
+  private var margs: List[String] = List.empty
   lazy val args = margs
 
   def runFull: IO[Int]
--- a/core/js/src/main/scala/cats/effect/IOApp.scala
+++ b/core/js/src/main/scala/cats/effect/IOApp.scala
@@ -19,17 +19,14 @@ package cats.effect
 import scala.concurrent.duration._
 import scala.scalajs.js
 
-trait IOApp extends IOFullApp {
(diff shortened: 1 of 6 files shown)


core/js/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -19,9 +19,14 @@ package cats.effect
 import scala.concurrent.duration._
 import scala.scalajs.js
 
+trait IOMinApp extends IOApp {
+  def run: IO[Unit]
+  final def run(args: List[String]): IO[ExitCode] = run.as(ExitCode.Success)
+}
+
```
djspiewak: This should go into the **shared/** source directory, since it's all unified.

core/shared/src/main/scala/cats/effect/ExitCode.scala
```diff
@@ -0,0 +1,38 @@
...
+package cats.effect
+
+/**
+ * Represents the exit code of an application.
+ *
+ * `code` is constrained to a range from 0 to 255, inclusive.
+ */
+sealed abstract case class ExitCode private (code: Int)
+
+object ExitCode {
```
djspiewak, line 26: Is it worth adding something like this in here?

```scala
implicit def coerceU(iou: IO[Unit]): IO[ExitCode] = iou.as(Success)
```

mpilquist: At the risk of worrying too much about Scala 3.1 in the present, I'd suggest avoiding implicit conversions.

core/js/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -19,9 +19,14 @@ package cats.effect
 import scala.concurrent.duration._
 import scala.scalajs.js
 
+trait IOMinApp extends IOApp {
```
mpilquist: Love it but would prefer one of these names:
- `IOApplet` (my favorite - short, catchy, fun)
- `IOAppBasic`
- `IOAppLite`

I like keeping `IOApp` prefix to keep it sorted in ScalaDoc, directory listings, and auto-complete.

djspiewak: > IOApplet (my favorite - short, catchy, fun)

Has there been enough time passed for us to reclaim the "applet" name?

I agree the prefix is useful.

RaasAhsan: 👍  for the `IOApp` prefix and `IOApplet`. Some more ideas if that doesn't stick:
- `IOAppMini`
- `IOAppTiny`
- `IOAppSimple` (`SimpleIOApp` doesn't sound too bad either)

matil019: I would not like to have the name `IOApplet`. I work with some legacy systems in which Java applets are still / were recently in use. It is unfortunate, but `IOApplet` would confuse my less experienced team members. I'm fine with the other names :)

ChristopherDavenport: This was dropped on gitter and I kinda like it, credit to @oleg-py  

`IOApp_` which follows the cats pattern of `_` at then end returning unit which will perhaps aid in understanding.

SystemFw: @oleg-py suggests `IOApp.Simple`. I like that. Alternatively using objects like `simple.IOApp` or `tiny.IOApp`

ChristopherDavenport:
For `IOApp.Simple` the IO type in the source specific sources can't be found in `object IOApp` in shared. Replicating in platform specific code allows this to work.

core/js/src/main/scala/cats/effect/IOApp.scala
SystemFw: `IOApp.Lite` is another suggestion

RaasAhsan (approved):

PR #1140, opened 2020-08-27
2.x docs migration

RaasAhsan:
[ ] Delete microsite configuration


build.sbt
```diff
@@ -58,12 +47,12 @@ val SilencerVersion = "1.7.1"
 
 replaceCommandAlias(
   "ci",
-  "; project /; headerCheck ;scalafmtSbtCheck ;scalafmtCheckAll; clean; testIfRelevant; mimaReportBinaryIssuesIfRelevant; doc"
+  "; project /; headerCheck ;scalafmtSbtCheck ;scalafmtCheckAll; clean; testIfRelevant; mimaReportBinaryIssuesIfRelevant; doc; docs/mdoc"
```
djspiewak: Shouldn't we still run mdoc on build to ensure we don't break anything?

RaasAhsan: `docs/mdoc` does that right? and `sbt ci` is run on every build

djspiewak: Oh derp I read the diff backwards. Nvm!

djspiewak:
@RaasAhsan When you get a chance can you bring this up to date with `HEAD`?

RaasAhsan:
```
[error] (run-main-0) java.lang.ClassNotFoundException: mdoc.Main
[error] java.lang.ClassNotFoundException: mdoc.Main
[error] 	at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
```
What did i mess up

djspiewak:
…what

RaasAhsan:
It happens locally for Scala `3.0.0-M2` but not for `2.12` or `2.13`

djspiewak:
Oh, I think we should probably turn off mdoc for anything that isn't 2.13, at least for now. I *think* that's what other projects are doing at present?

RaasAhsan:
Seems like it's what we're doing on series/3.x: https://github.com/typelevel/cats-effect/blob/series/3.x/build.sbt#L69

PR #1145, opened 2020-08-29
Allocate

### LINKED ISSUE #1017: Allocated: New Typeclass Below Concurrent

RaasAhsan:
Closes #1017. This PR is a first pass at introducing an `Allocate` class into the typeclass hierarchy that sits below `Concurrent` and above `Temporal`. Opening the PR up early so people can see where things are headed.

## TODO
- [ ] Rename `Concurrent` to `Spawn` and `Allocated` to `Concurrent`. I may include this in a follow-up PR to reduce the scope of changes here.
- [ ] Add laws for `Allocate`. Starting points are covered in #1017.
- [ ] Address `Mk` for `Semaphore` and `MVar`
- [ ] Address `MkIn`. Should we define `refIn[G[_]]` and `deferredIn[G[_]]` on `Allocate`? Perhaps `ref` and `deferred` can be derived from those.
- [x] ~Move `SyncRef` and `AsyncDeferred` to `concurrent`. Note that this means `Ref.unsafe` and `Deferred.unsafe` isn't possible anymore, so callers will have to reference the implementation, or we expose it in `Allocated` (not a fan of that)~

### DIFF
--- a/concurrent/shared/src/main/scala/cats/effect/concurrent/package.scala
+++ b/concurrent/shared/src/main/scala/cats/effect/concurrent/package.scala
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2020 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+
+package object concurrent {
+  type Ref[F[_], A] = cats.effect.kernel.Ref[F, A]
+  type Deferred[F[_], A] = cats.effect.kernel.Deferred[F, A]
+
+  val Ref = cats.effect.kernel.Ref
+  val Deferred = cats.effect.kernel.Deferred
+}
--- a/core/shared/src/test/scala/cats/effect/concurrent/DeferredSpec.scala
+++ b/core/shared/src/test/scala/cats/effect/concurrent/DeferredSpec.scala
@@ -19,7 +19,6 @@ package effect
 package concurrent
 
 import cats.implicits._
-import cats.effect.kernel.{Deferred, Ref}
 
 import scala.concurrent.duration._
 
--- a/core/shared/src/test/scala/cats/effect/concurrent/DeferredSpec.scala
+++ b/core/shared/src/test/scala/cats/effect/concurrent/DeferredSpec.scala
@@ -19,6 +19,7 @@ package effect
 package concurrent
 
 import cats.implicits._
+import cats.effect.kernel.{Deferred, Ref}
 
 import scala.concurrent.duration._
 
--- a/core/shared/src/test/scala/cats/effect/concurrent/DeferredSpec.scala
+++ b/core/shared/src/test/scala/cats/effect/concurrent/DeferredSpec.scala
@@ -39,15 +39,11 @@ class DeferredSpec extends BaseSpec { outer =>
     }
 
     "complete is only successful once" in real {
-      val op = Deferred[IO, Int].flatMap { p =>
-        (p.complete(0) *> p.complete(1).attempt).product(p.get)
-      }
+      val op = Deferred[IO, Int].flatMap { p => p.complete(0) *> p.complete(1).product(p.get) }
 
       op.flatMap { res =>
         IO {
-          res must beLike {
-            case (Left(e), 0) => e must haveClass[IllegalStateException]
-          }
+          res must beEqualTo((false, 0))
         }
       }
     }
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Deferred.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Deferred.scala
@@ -16,13 +16,12 @@
 
 package cats
 package effect
-package concurrent
-
-import cats.effect.concurrent.Deferred.TransformedDeferred
-import cats.effect.kernel.{Async, Sync}
-import cats.syntax.all._
+package kernel
 
 import java.util.concurrent.atomic.AtomicReference
+
+import cats.effect.kernel.Deferred.TransformedDeferred
+import cats.syntax.all._
 
 import scala.annotation.tailrec
 import scala.collection.immutable.LongMap
@@ -236,7 +235,7 @@ object Deferred {
     }
   }
 
-  final private class TransformedDeferred[F[_], G[_], A](
+  final private[kernel] class TransformedDeferred[F[_], G[_], A](
       underlying: Deferred[F, A],
       trans: F ~> G)
       extends Deferred[G, A] {
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Deferred.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Deferred.scala
@@ -57,19 +57,14 @@ abstract class Deferred[F[_], A] {
 
   /**
    * If this `Deferred` is empty, sets the current value to `a`, and notifies
-   * any and all readers currently blocked on a `get`.
+   * any and all readers currently blocked on a `get`. Returns true.
    *
-   * If this `Deferred` has already been completed, the returned
-   * action immediately fails with an `IllegalStateException`. In the
-   * uncommon scenario where this behavior is problematic, you can
-   * handle failure explicitly using `attempt` or any other
-   * `ApplicativeError`/`MonadError` combinator on the returned
-   * action.
+   * If this `Deferred` has already been completed, returns false.
    *
    * Satisfies:
    *   `Deferred[F, A].flatMap(r => r.complete(a) *> r.get) == a.pure[F]`
    */
-  def complete(a: A): F[Unit]
+  def complete(a: A): F[Boolean]
 
   /**
    * Obtains the current value of the `Deferred`, or None if it hasn't completed.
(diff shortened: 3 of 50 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/Deferred.scala
```diff
@@ -92,8 +91,8 @@ object Deferred {
    * If you want to share one, pass it as an argument and `flatMap`
    * once.
    */
-  def apply[F[_], A](implicit mk: Mk[F]): F[Deferred[F, A]] =
-    mk.deferred[A]
+  def apply[F[_], A](implicit F: Allocate[F, _]): F[Deferred[F, A]] =
```
RaasAhsan, line 89: This works for `IO`, but I'm not sure how it would work for bifunctor implementations?

mpilquist:
When doing the renames (perhaps in separate PR like you said), are we going to rename the throw scheme too? E.g., `Concurrent[F] = ConcurrentE[F, Throwable]`

SystemFw:
> E.g., Concurrent[F] = ConcurrentE[F, Throwable]

That would be ideal imho

djspiewak:
I merged this with series/3.x and published as `3.0-8006968` if anyone wants to jump on it right away

SystemFw:
@RaasAhsan I think we still need `Ref.Mk`. `Ref[SyncIO]` is needed.

SystemFw:
Worth having a look at 
https://github.com/typelevel/fs2/blob/develop/core/shared/src/main/scala/fs2/Stream.scala#L4295-L4377
and
https://github.com/typelevel/fs2/blob/develop/core/shared/src/main/scala/fs2/tc.scala

to understand how this plays it out

djspiewak:
> I think we still need Ref.Mk. Ref[SyncIO] is needed.

*Needed*? Also do you concretely have `SyncIO`?

SystemFw:
> Needed? Also do you concretely have SyncIO?

yes and no. Easier to look at the code I linked to see why, but happy to talk about it further here or on Gitter ofc :)

EDIT: actually it's probably yes and yes
EDIT2: aargh, it's better to just look at the code, there are a few options. Regardless of fs2 though, I think `Ref[SyncIO]` is useful

djspiewak:
> EDIT2: aargh, it's better to just look at the code, there are a few options. Regardless of fs2 though, I think Ref[SyncIO] is useful

I agree with this, though I think it's achievable without keeping `Mk`. If you concretely know you have a specific type (`SyncIO`), then you also know you have a `Sync` for it. I don't see a problem with keeping a datatype constructor which retains the ability to construct a `Ref[F]` given `F: Sync`

SystemFw:
I think fs2's `CompileScope` is in the same place `Resource` is (which required `Resource.Bracket`), i.e. needing `Concurrent or SyncIO` as an abstraction. To be honest I do see potential changes there that could help, though it gets tricky. Genuinely not sure if that's cats-effect's concern though.

djspiewak:
> To be honest I do see potential changes there that could help, though it gets tricky. Genuinely not sure if that's cats-effect's concern though.

Not sure either. I agree that `MonadCancel | Sync` seems to be an increasingly common scenario (for some value of `MonadCancel`).

RaasAhsan:
Another idea that was floated in #1017 was to split `Ref` and `Deferred` into separate classes. `Deferred` only really makes sense at or below `Concurrent/Spawn`, but `Ref` more or less works anywhere. It lets us avoid having multiple ways to create a `Ref` depending on what constraints you have available, but I'm not sure that's worth another class

SystemFw:
I think the problem there is again multiplying constraints in most call sites.
We could do `Allocate extends Ref.Mk` to mitigate that, which is what I've done in my fs2 stub, but I don't know if that's the way to go yet.

RaasAhsan:
PR might end up taking a bit longer than expected to get merged. I don't think `PureConc` is in a position to provide an instance for `Allocated` right now since we can't create instances for `Ref` and `Deferred` over it. There's no built-in machinery for that and it also doesn't provide `Sync` or `Async` instances. We need a `Temporal` instance for `PureConc` to test a bunch of laws, but we can't create that without `Allocate`.

 I'm not very well-equipped to touch `PureConc`, so I think the quick solution might be to flip `Allocate` and `Temporal` so builds start passing and it unblocks FS2

SystemFw:
> PR might end up taking a bit longer than expected to get merged. 

Don't worry about it :)

> I don't think PureConc is in a position to provide an instance for Allocated right now since we can't create instances for Ref and Deferred over it

mm, I think PureConc can create MVars, so you should be able to create Refs and Deferreds out of that with a bit of effort 

> I'm not very well-equipped to touch PureConc, so I think the quick solution might be to flip Allocate and Temporal so builds start passing and it unblocks FS2

There's no rush. FS2 is unblocked in the short term (using its own versions), and the remaining questions like whether we need Ref.Mk require some thinking and work there first. I'd say to keep this open for now, and then if you can't/don't really want to work out the PureConc issue, I can pick it up after the `Promise` work

core/shared/src/test/scala/cats/effect/concurrent/DeferredSpec.scala
```diff
@@ -19,6 +19,7 @@ package effect
 package concurrent
 
 import cats.implicits._
+import cats.effect.kernel.{Deferred, Ref}
```
djspiewak: We should add forwarders in `concurrent` just to ease the migration path.

kernel/shared/src/main/scala/cats/effect/kernel/Deferred.scala
djspiewak: Should be fine for everything, afaict.

testkit/shared/src/main/scala/cats/effect/testkit/pure.scala
```diff
@@ -212,6 +211,68 @@ object pure {
       def never[A]: PureConc[E, A] =
         Thread.annotate("never")(Thread.done[A])
 
+      def ref[A](a: A): PureConc[E, Ref[PureConc[E, *], A]] =
```
RaasAhsan, line 214: Implementation for `Ref` and `Deferred` are here. I wanted to use `MVarPartiallyApplied` to avoid typing every call, but didn't manage to get it to work, need to take a closer look

SystemFw: So, unfortunately these implementations are unsafe in multiple ways, and after giving it more thought, I think they might be _fundamentally_ unsafe (talking about `Ref` mostly, `Deferred` has different semantics here but it's easily fixable).  I'm really thinking we should add `Ref+Deferred` to `PureConc` and remove `MVar` (`MVar` can safely written in terms of `Ref+Deferred`, but now I don't think the opposite is true, definitely not with a single `MVar` at least).

djspiewak: Honestly that might be the simplest approach. `MVar` is in coop, and the only reason I did it rather than `Ref` + `Deferred` was pure laziness: I wanted to write one thing rather than two.

concurrent/shared/src/main/scala/cats/effect/concurrent/package.scala
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+
+package object concurrent {
```
RaasAhsan, line 19: Added the type and static forwarders. Not sure if the `val`s make sense, but they seem to work. Worth noting that other classes inside the `concurrent` package have to use these

RaasAhsan:
Here's the current state of the PR:
1. The default implementations for `Ref` and `Deferred (`SyncRef` and `AsyncDeferred` respectively) remains in the `kernel` package. Users won't need to import a separate `concurrent` module to access them.
2. I decided to keep the `Allocate[PureConc]` instance written in terms of `MVar` in, so we can at least get started on writing laws and exercising them against both `PureConc` and `IO`. Eventually, `PureConc` should be augmented so that `Ref` and `Deferred` are primitive and `MVar` is derived.
3. We haven't reached consensus on the `Mk`/`MkIn` situation. I don't have a strong opinion here, but I find it weird to only have a `Ref.Mk[SyncIO]` (but not any of the others) for FS2. A disparity between different types having or not having a `Mk` or `MkIn` would also be odd and confusing to users. I'm leaving them in for now.

I can follow up this PR with the suggested name changes, and we can create an issue to talk about how we're going to write laws for this.

mpilquist:
Awesome, sounds good to me. I agree that next step should be renames. Once those are in, I can make a pass at fixing up fs2.

djspiewak:
Looks really good! I think we can follow-up with almost everything else. One change I'd like to get in this PR though would be to swap the return type of `Deferred#complete` from `F[Unit]` to `F[Boolean]`. We can't just synthesize an error, since we don't know what `E` is, so our best approach is to return `Boolean` and let users handle it.

PR #1149, opened 2020-09-01
Organize IOSpec into vaguely related sections

TimWSpence:


TimWSpence:
This is by no means the only grouping but I think it's a reasonably sensible one and at least makes `IOSpec` a good deal more manageable

djspiewak (approved):
Fantastic! I kind of want to organize the `platform` stuff too, rather than just lumping it into one category, but it's easy to iterate on this stuff. Thank you!

PR #1156, opened 2020-09-02
Hierarchy renaming

RaasAhsan:
- [x] Rename `Concurrent` to `Spawn`
- [x] Rename `Allocated` to `Concurrent`
~- [ ] Rename base classes to their `E` variants and add `Throw` aliases which fix the error type to `Throwable`~

I'm opening this PR before the third item is done because it's going to be pretty ugly

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Resource.scala
+++ b/core/shared/src/main/scala/cats/effect/Resource.scala
@@ -515,48 +515,61 @@ object Resource extends ResourceInstances with ResourcePlatform {
 
   trait Bracket0 {
     implicit def catsEffectResourceBracketForSyncEffect[F[_]](
-        implicit F: SyncEffect[F]): Bracket[F] =
-      new Bracket[F] {
-        def bracketCase[A, B](acquire: F[A])(use: A => F[B])(
-            release: (A, ExitCase) => F[Unit]): F[B] =
-          flatMap(acquire) { a =>
-            val handled = onError(use(a)) {
-              case e => void(attempt(release(a, ExitCase.Errored(e))))
-            }
-            flatMap(handled)(b => as(attempt(release(a, ExitCase.Completed)), b))
-          }
-
-        def pure[A](x: A): F[A] = F.pure(x)
-        def handleErrorWith[A](fa: F[A])(f: Throwable => F[A]): F[A] = F.handleErrorWith(fa)(f)
-        def raiseError[A](e: Throwable): F[A] = F.raiseError(e)
-        def flatMap[A, B](fa: F[A])(f: A => F[B]): F[B] = F.flatMap(fa)(f)
-        def tailRecM[A, B](a: A)(f: A => F[Either[A, B]]): F[B] = F.tailRecM(a)(f)
+        implicit F0: SyncEffect[F]): Bracket[F] =
+      new SyncBracket[F] {
+        implicit protected def F: Sync[F] = F0
       }
+
+    trait SyncBracket[F[_]] extends Bracket[F] {
+      implicit protected def F: Sync[F]
+
+      def bracketCase[A, B](acquire: F[A])(use: A => F[B])(
+          release: (A, ExitCase) => F[Unit]): F[B] =
+        flatMap(acquire) { a =>
+          val handled = onError(use(a)) {
+            case e => void(attempt(release(a, ExitCase.Errored(e))))
+          }
+          flatMap(handled)(b => as(attempt(release(a, ExitCase.Completed)), b))
+        }
+
+      def pure[A](x: A): F[A] = F.pure(x)
+      def handleErrorWith[A](fa: F[A])(f: Throwable => F[A]): F[A] = F.handleErrorWith(fa)(f)
+      def raiseError[A](e: Throwable): F[A] = F.raiseError(e)
+      def flatMap[A, B](fa: F[A])(f: A => F[B]): F[B] = F.flatMap(fa)(f)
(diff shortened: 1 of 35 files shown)


core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -176,7 +176,7 @@ sealed abstract class Resource[+F[_], +A] {
    *             .use(msg => IO(println(msg)))
    * }}}
    */
-  def parZip[G[x] >: F[x]: ConcurrentThrow: Ref.Mk, B](
+  def parZip[G[x] >: F[x]: SpawnThrow: Ref.Mk, B](
```
mpilquist: Upgrade `SpawnThrow: Ref.Mk` to `Concurrent`?

mpilquist:
Initial port of FS2 to the hierarchy in this PR: https://github.com/typelevel/fs2/compare/develop...mpilquist:topic/new-hierarchy?expand=1

I'm going to make another pass which removes all use of `Ref.Mk` from fs2, using an adhoc `Compilable[F]`.

Some minor things:
 - `Ref.of[F, A]` method on the companion currently requires `Concurrent[F, _]` but `Ref[F].of[A]` only requires `Sync`
 - `Deferred` needs a `completeOrError` to ease integration -- added this for now: https://github.com/typelevel/fs2/compare/develop...mpilquist:topic/new-hierarchy?expand=1#diff-b66575a3ef9a76c418eabd7ca0dd8ac4R55-R57

SystemFw:
Well that looks quite pleasant @mpilquist 😃 
I think in most cases `compileOrFail` is not needed and we can just `complete.void` , since `complete` will only fail if it's called twice on the same `Deferred`, which isn't the case in most code paths in that diff.

mpilquist:
@RaasAhsan I pushed 2 commits, 1 that adds forwarders for `Concurrent` and `ConcurrentThrow` to the `cats.effect` package and another that allows the `Resource.Bracket` classes to be reused without copy/paste.

RaasAhsan:
@mpilquist Yeah, that's a slightly unfortunate inconsistency. We still need a public API method to create a `Ref[SyncIO, A]` that isn't `new SyncRef` (it's declared `private` right now anyways) and the partially-applied constructor is the only way to do that at the moment. I would be ideal if we didn't have to call a different method based on whether we have an `Allocate` or `Sync`, so I'm wondering if there's some implicit magic we can do that can unambiguously produce a `Mk` from an `Allocated` or a `Sync`.

Also, I'm inclined to not push the new naming scheme as part of this PR, it's probably worth more discussion.

RaasAhsan:
So, it looks like this compiles:
```scala
object Ref {
  def create[F[_], A](a: A)(implicit mk: Make[F]): F[Ref[F, A]] = mk.refOf(a)

  trait Make[F[_]] {
    def refOf[A](a: A): F[Ref[F, A]]
  }

  object Make extends MakeInstances

  trait MakeInstances extends MakeLowPriorityInstances {
    implicit def allocatedInstance[F[_]](implicit F: Concurrent[F, _]): Make[F] =
      new Make[F] {
        override def refOf[A](a: A): F[Ref[F, A]] = F.ref(a)
      }
  }

  trait MakeLowPriorityInstances {
    implicit def syncInstance[F[_]](implicit F: Sync[F]): Make[F] =
      new Make[F] {
        override def refOf[A](a: A): F[Ref[F, A]] = F.delay(unsafe(a))
      }
  }

  def createAsync[F[_], A](a: A)(implicit F: Async[F]): F[Ref[F, A]] = create[F, A](a)
  
  def createSync[F[_], A](a: A)(implicit F: Sync[F]): F[Ref[F, A]] = create[F, A](a)

  def createConcurrent[F[_], A](a: A)(implicit F: Concurrent[F, _]): F[Ref[F, A]] = create[F, A](a)
}
```

RaasAhsan:
Removed `Mk` and `MkIn`. `Ref.Make` is sticking around, though I'm completely willing to change the name back to `Mk` or `Alloc` or something.

RaasAhsan:
`0.26.0` dotty compile is getting a stack overflow, not really sure what's causing that at this point, maybe the prioritized implicits?

joroKr21:
> so I'm wondering if there's some implicit magic we can do that can unambiguously produce a Mk from an Allocated or a Sync.

There is `OrElse` in shapeless. But I wonder, wouldn't overloading work?

RaasAhsan:
@joroKr21 Could you share an example that illustrates that? The solution right now is to prioritize the implicit `Make` by `Allocated` instances and then `Sync` instances. So `Ref.of` works with `Allocated`, `Sync`, `Async` (which extends `Sync` and `Allocated` transitively), and also with any combination of those in scope.

djspiewak:
@RaasAhsan `OrElse` is similarly biased. Glancing at [the implementation](https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/orelse.scala), it appears to be left biased. Either way though, I think directly encoding the disjunction here (rather than pulling in the dependency) is the right approach.

kernel/shared/src/main/scala/cats/effect/kernel/Ref.scala
```diff
@@ -146,6 +146,8 @@ abstract class Ref[F[_], A] {
 
 object Ref {
 
+  @annotation.implicitNotFound(
+    "Cannot find an instance for Ref.Make. Add implicit evidence of Concurrent[F, _], Sync[F] or Async[F] to scope to automatically derive one.")
```
djspiewak: I would just say `Concurrent` or `Sync`. Also, if you use `$F` rather than just `F`, the compiler will inject the *actual* type it was looking for.

Issue #1117, opened 2020-08-21
Special-Case IO#race/both


djspiewak:
They can be implemented in terms of racePair, but they have vastly more efficient implementations as primitives in the run loop.

RaasAhsan:
Any special casing will typically make for more efficient implementations, so benchmarks that show a considerable improvement is a good acceptance criteria here.

PR #1157, opened 2020-09-02
Special case race and both

TimWSpence:


TimWSpence:
https://github.com/typelevel/cats-effect/issues/1117

Creating this just to track it. Currently there are a few interesting test failures related to cancellation but we're waiting till the `Promise` stuff is done to see whether the problem is with this code or with relying on the old `async` cancellation behaviour.

PR #1164, opened 2020-09-06
Push per thread state to each thread and remove one source of contention

vasilmkd:
Guarding against spurious wakeups does not depend on other threads and should be pushed to a state local to the thread. As a bonus, no locking needs to be done, which should reduce contention when threads are waking up from sleep.

Furthermore, a spuriously woken up thread should not notify other workers of new work, because there is none.


vasilmkd:
The benchmarks deadlock...

vasilmkd:
```
Benchmark                                   (size)   Mode  Cnt  Score   Error    Units
WorkStealingBenchmark.async                1000000  thrpt   20  4.742 ± 0.227  ops/min
WorkStealingBenchmark.asyncTooManyThreads  1000000  thrpt   20  7.184 ± 0.338  ops/min
```

djspiewak:
Very clever!

Issue #1136, opened 2020-08-25
Memory leak in Deferred or IO.race


mpilquist:
Consider:

```scala
Deferred[IO, Int].flatMap { d =>
  def loop: IO[Unit] = IO.race(d.get, IO.cede >> IO.unit) >> loop
  loop
}
```

This leaks memory -- each call to `d.get` registers a callback in the internal `LongMap` and they never get removed as a result of losing the race. Bug may be in `IO.race` instead of `Deferred` -- perhaps not calling the unregistration action?

<img width="1048" alt="image" src="https://user-images.githubusercontent.com/118078/91236976-63994d80-e707-11ea-9d9f-718f3eae5337.png">

Result of running above through FS2's leak testing infrastructure:

```
Running leak test deferred
Heap:    4,464,560 total,      +42,296 since start,      +42,296 in last 1 second
Heap:    4,596,984 total,     +174,720 since start,     +132,424 in last 1 second
Heap:    4,679,464 total,     +257,200 since start,      +82,480 in last 1 second
Heap:    4,803,848 total,     +381,584 since start,     +124,384 in last 1 second
Heap:    4,880,024 total,     +457,760 since start,      +76,176 in last 1 second
Heap:    4,959,328 total,     +537,064 since start,      +79,304 in last 1 second
Heap:    5,032,096 total,     +609,832 since start,      +72,768 in last 1 second
Heap:    5,190,248 total,     +767,984 since start,     +158,152 in last 1 second
Heap:    5,264,224 total,     +841,960 since start,      +73,976 in last 1 second
Heap:    5,423,352 total,   +1,001,088 since start,     +159,128 in last 1 second
Heap:    5,542,248 total,   +1,119,984 since start,     +118,896 in last 1 second
Heap:    5,648,088 total,   +1,225,824 since start,     +105,840 in last 1 second
==> X fs2.MemoryLeakSpec.deferred  15.408s munit.FailException: Development/oss/fs2/core/jvm/src/it/scala/fs2/MemoryLeakSpec.scala:63 leak detected - heap dump: /var/folders/58/rc_15mds5vx9119_96_01wsc0000gp/T/fs2-leak-test-7474361769001304354.hprof
62:        case Right(Left(path)) =>
63:          fail(s"leak detected - heap dump: $path")
64:      }.unsafeRunSync()
    at munit.Assertions.fail(Assertions.scala:189)
    at munit.Assertions.fail$(Assertions.scala:14)
[...]

PR #1165, opened 2020-09-07
Simplify Async node

### LINKED ISSUE #1104: Implement Async#promise as a lower level primitive
```scala
trait Async[F[_]] extends Temporal[F] with Sync[F] {
  // ...
  def promise[A]: F[Promise[F, A]]

  def async[A](k: (Either[Throwable, A] => Unit) => F[Option[F[Unit]]]): F[A] =
    promise[A] flatMap { p => 
      k(_.fold(p.fail, p.complete)) flatMap {
        case Some(fin) => p.get.onCancel(fin)
        case None => p.get
      }
    }
}

trait Promise[F[_], A] {
  def fail(t: Throwable): Unit
  def complete(a: A): Unit
  def get: F[A]
}
```

This is a lower level primitive, similar to `Deferred` (but not identical!), which allows us to solve the problem of composable `Async` instances (e.g. #1086). Additionally, this gives us a lot more control over cancelation semantics and radically simplifies concrete implementations, which proves that it's the more primitive version of the concept.

SystemFw:
Fixes  #1104 #1136  #1227


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -705,6 +711,11 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
   private[effect] final case class Attempt[+A](ioa: IO[A]) extends IO[Either[Throwable, A]] {
     def tag = 20
   }
+  // TODO rename, move
+  private[effect] final case class Cont[A]() extends IO[(IO[A], (Either[Throwable, A] => Unit))] {
```
djspiewak: Should be an `object`. Also I would have expected this to outright replace `Async`.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -321,6 +325,33 @@ private final class IOFiber[A](
...
+          // If the callback finishes after registration, it owns the
+          // runloops (via async continue), and registration just
+          // terminates in suspendWithFinalizationCheck.
+          // if registration wins, it gets the result from the atomic ref
+          // and it continues, andThen the callback just terminates (stateLoop, when tag == 3)
+          // The rest of the code deals with publication details through
+          // `suspended`, and all the corner cases with finalisation
+          // which we aim to remove with the `promise` construction.
+          // Need to think about naming since resume is already used
+          // all over the place in the runloop.
```
djspiewak: I hope you're not planning on deleting this…

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -588,6 +627,81 @@ private final class IOFiber[A](
 
           conts.push(AttemptK)
           runLoop(cur.ioa, nextIteration)
+
+        // Cont TODO rename
+        case 21 =>
+          val state = new AtomicReference[ContState](ContState.Initial)
+
+          val cb: Either[Throwable, A] => Unit = { e =>
+            // should I do any checks on the old finalisation state like in async?
```
djspiewak: The only reason those checks exist is to solve the case where `async` is used *in a finalizer*. When an async callback is fired on a canceled fiber, we generally want to just eat the callback so we don't fork the runloop, but if we're in the process of evaluating finalizers, we *don't* want to eat the callback.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -588,6 +627,81 @@ private final class IOFiber[A](
...
+            @tailrec
+            def loop(): Unit = {
+              println(s"cb loop sees suspended ${suspended.get} on fiber $name")
+              if (resume("callback")) { // is this necessary to publish writes?
```
djspiewak: Yes, but also it's necessary to grab the runloop ownership. `suspend`/`resume` are sort of like a runloop semaphore protocol. They're responsible for ensuring that, if someone is running the fiber, there is only *one* such someone.

In this case, what you're doing here is reanimating evaluation of a `get`, so there was a suspension (when the `get` was hit) and now you're bringing it back up again.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -588,6 +627,81 @@ private final class IOFiber[A](
...
+          // we want this to be defer and not delay to be able to treat
+          // the result as `null: IO[A]` rather than `null: A`, since
+          // the `null` IO makes the runloop suspend
+          val get: IO[Any] = IO.defer {
+            if (!state.compareAndSet(ContState.Initial, ContState.Waiting)) {
+              // state was no longer Initial, so the callback has already been invoked
+              // and the state is Result
+              val result = state.get().result
+              // we leave the Result state unmodified so that `get` is idempotent
+              if (!shouldFinalize()) {
```
djspiewak: There's probably something awkward here w.r.t. `get` within an `onCancel`.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -602,11 +716,19 @@ private final class IOFiber[A](
   private[this] def isUnmasked(): Boolean =
     masks == initMask
 
-  private[this] def resume(): Boolean =
-    suspended.compareAndSet(true, false)
+  private[this] def resume(str: String = ""): Boolean = {
+    val v = suspended.compareAndSet(true, false)
+    if(str.nonEmpty) println(s"resume called by $str on fiber $name, suspended ${suspended.get}")
```
djspiewak: FYI, `str != ""` is a lot faster. `str.length > 0` might be faster-still.

djspiewak (commented):
Initial thoughts. I'll stare at the state machine a bit more. Off-hand, it doesn't look like it has enough states to actually work. Need to ponder.

core/shared/src/main/scala/cats/effect/IO.scala
SystemFw: oh sorry, this is just wip. It will replace `Async` once it's done, but I'm keeping both for now so that I can use `start/sleep` in tests and retain a touch of sanity.

core/shared/src/main/scala/cats/effect/IOFiber.scala
SystemFw: ahahah I was actually not sure if it was going to be useful for other people or not. But yeah I'm thinking that in the worst case scenario of this PR going nowhere, I'll still commit some of the comments.

core/shared/src/main/scala/cats/effect/IOFiber.scala
SystemFw: > the case where async is used in a finalizer

I have to say that this hadn't even remotely crossed my mind, thanks

core/shared/src/main/scala/cats/effect/IOFiber.scala
SystemFw: that's useful thanks, so far I haven't yet hit a case where the slow string check cause the races to disappear,  but it's bound to happen at some point I guess

core/shared/src/main/scala/cats/effect/IOFiber.scala
SystemFw: > . suspend/resume are sort of like a runloop semaphore protocol. They're responsible for ensuring that, if someone is running the fiber, there is only one such someone.

This bit I did get in general, although it breaks with `.get.start` as we've discussed. The reason I added the comment there though is that it seemed to me that `ContState` was going to be enough for that, although that's probably wrong

SystemFw:
> I'll stare at the state machine a bit more. Off-hand, it doesn't look like it has enough states to actually work. Need to ponder.

Please do. Also this construction does not push anything on `conts`, which feels dodgy (but the `async` protocol with respect to `conts` cannot be used here)

core/shared/src/main/scala/cats/effect/IOFiber.scala
djspiewak: I'm worried about things like this:

```scala
promise[A] flatMap { p =>
  p.get.start >> p.get.start
}
```

Like… good luck. You really would need `CallbackQueue` in order to do anything with the above. This is part of why I like the `CPST` approach better, if it works out the way it *seems* like it should.

djspiewak:
> Also this construction does not push anything on conts, which feels dodgy (but the async protocol with respect to conts cannot be used here)

That's actually expected. The only reason `Async` pushes something to `conts` is the fact that it has a *body* (the contents of the `async` block). Of course, `promise` doesn't, so you have no continuations to juggle.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -292,6 +292,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
           poll(ioa.to[F])
 
         case _ : IO.Cont[_] => ??? // TODO rename, translate to operation on Async once it's there
+        case _ : IO.Get[_] => sys.error("impossible")
```
djspiewak: I'm not sure that it is. This is probably similar to the `Unmask` situation

SystemFw: yeah actually I think we have a problem here

SystemFw: `cont.flatMap(_.get).to[F]` would probably crash, and it's not like `Unmask` either, because in that case we can express the operation as `poll`, I'm not sure we have an equivalent operation in any typeclass here, and putting `get` in `Async` is pretty weird since it's a standalone method, but an internal node which is a by-product of evaluating `Cont`.

djspiewak: I think the solution is similar to `Unmask`: you have to get the `get` from the delegate's CPS.

SystemFw: mm ok. I'll leave it for now and get back to it once I harden the external encoding

djspiewak: > cont.flatMap(_.get).to[F] would probably crash, and it's not like Unmask either, because in that case we can express the operation as poll, I'm not sure we have an equivalent operation in any typeclass here, and putting get in Async is pretty weird since it's a standalone method, but an internal node which is a by-product of evaluating Cont.

You're going to get it *from* the `cont` though. You're starting by interpreting `IO.Cont` to `F`, which you do by calling `F.cont`, which gives you a `get: F[A]`. Take that `get` and stuff it into *another* `IO` node and pass that to the `Cont` function. So you don't have to worry about ever seeing `IO.Get`, but you *will* see your fancy `Get` node that carries the `get: F[A]` value, similar to what we do with `poll` and `Unmask`.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -477,17 +510,19 @@ private final class IOFiber[A](
             def apply[B](ioa: IO[B]) = IO.UnmaskRunLoop(ioa, id)
           }
 
+          // The uncancelableK marker is used by `succeeded` and `failed`
+          // to unmask once body completes.
           conts.push(UncancelableK)
           runLoop(cur.body(poll), nextIteration)
 
         // Canceled
         case 12 =>
```
kubukoz, line 476: Just wondering, can these be replaced with final vals? Wouldn't they be inlined at compile time?

djspiewak: Nope! There are bugs in scalac related to this stuff. Generally-speaking, you can't get a `tableswitch` to generate if you use vals; you need to actually inline the constants manually.

djspiewak: (also, curiously, `@inline` appears to *sometimes* fix this issue, but not always; this is, to my knowledge, the only practically visible impact of that annotation)

kubukoz: eh. Thanks ;)

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -588,6 +627,78 @@ private final class IOFiber[A](
...
+            if (!shouldFinalize()) {
+              // println("get taking over")
+              asyncContinue(result, "get")
+            }
+          } else {
+            // callback has not been invoked yet
+            println(s"get suspending on fiber $name")
+            suspend("get") // async has a conditional suspend, why?
+            println(s"get sets suspended to ${suspended.get} on fiber $name")
+          }
```
djspiewak: I believe this is the origin of the canceled `never` deadlock: you aren't checking `canceled` here. The issue is the fact that cancelation could come in *during* the interpretation of this node, which means the canceler would fail to get the runloop and would thus suspend. Meanwhile, you go through this logic, never check cancelation, and just… suspend, meaning that no one runs the cancelation and things hang forever.

SystemFw: ah, I see.
Sounds like the answer to:
> // async has a conditional suspend, why?

and in retrospect, there is this comment in the relevant method too:
>     // race condition check: we may have been cancelled before we suspended

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -588,6 +627,79 @@ private final class IOFiber[A](
...
+              //println(s"cb loop sees suspended ${suspended.get} on fiber $name")
+              if (resume("callback")) { // is this necessary to publish writes?
+                if (!shouldFinalize()) {
+                //  println(s"callback taking over on fiber $name")
+                  asyncContinue(e, "callback")
+                }
+              } else if (!shouldFinalize()) {
+//                println(s"callback recurring on fiber $name")
+                loop()
+              } // else println("something wrong")
```
djspiewak: Here's your deadlock. The callback is `resume`ing the fiber, and it *does* need to finalize, but it doesn't have the ability to do so (here it just drops things on the floor). The scenario happens in the following case:

1. `get` suspends
2. The callback is fired and we pass through the `resume` barrier
3. Some other fiber cancels the `get`, and `cancel` cannot `resume`
4. The above block is hit, meaning that we have detected that we need to finalize, but no one ran the finalizers *for* us (and they can't, since we own the runloop), so we just die.

`Async` has quite a lot of complex conditional double-checking to avoid this particular scenario. :-)

core/shared/src/main/scala/cats/effect/IOFiber.scala
SystemFw: This fixed it

core/shared/src/main/scala/cats/effect/IOFiber.scala
SystemFw: After adding some tracing, deadlock doesn't happen on a winning callback path, it seems

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -588,6 +627,79 @@ private final class IOFiber[A](
...
+
+          if (!state.compareAndSet(ContState.Initial, ContState.Waiting)) {
+            // state was no longer Initial, so the callback has already been invoked
+            // and the state is Result
+            val result = state.get().result
+            // we leave the Result state unmodified so that `get` is idempotent
+            if (!shouldFinalize()) {
+              // println("get taking over")
+              asyncContinue(result, "get")
+            }
```
SystemFw: It's actually here, it hits the (invisible) `else`

core/shared/src/main/scala/cats/effect/IOFiber.scala
SystemFw: Ok I fixed the deadlock, which was a similar case but in `get`. This specific path isn't stressed by any tests, but I think you're right, basically on line 644 you need `else cancel`

core/shared/src/main/scala/cats/effect/IOFiber.scala
SystemFw: does this check apply to the callback in promise as well then?

SystemFw: it appears to not be needed: removing it when using `Async` makes a test consistently fail, and that same test consistently passes when using `Cont/promise`

kernel/shared/src/main/scala/cats/effect/kernel/Deferred.scala
```diff
@@ -175,11 +175,12 @@ object Deferred {
...
-              val id = addReader(awakeReader = resume)
-              val onCancel = F.delay(deleteReader(id))
+               val resume = (a: A) => cb(Right(a))
 
-              onCancel.some.pure[F]
+              F.delay(addReader(awakeReader = resume)).map { id =>
+                // if canceled
+                F.delay(deleteReader(id)).some
+              }
             }
```
SystemFw, line 170: I'm using `async` as we've agreed here, just wanted to double check we are ok with the extra allocations

core/js/src/main/scala/cats/effect/IOFiberConstants.scala
```diff
@@ -26,7 +26,7 @@ private[effect] object IOFiberConstants {
   val FlatMapK: Byte = 1
   val CancelationLoopK: Byte = 2
   val RunTerminusK: Byte = 3
-  val AsyncK: Byte = 4
+  // TODO reassign tags after AsyncK removal
```
djspiewak: FYI the tag reassignment is quite important. :-) `tableswitch` with consecutive values is treated differently by HotSpot than non-consecutive ones.

kernel/shared/src/main/scala/cats/effect/kernel/Deferred.scala
djspiewak: I think we can loop back and rework it if we decide that's worthwhile. There are definitely performance advantage to working with `cont` directly, but they might disappear when we fiddle with the API to be safer.

SystemFw: yeah fair enough. The side-effects in the old usage are a lot scarier imho now that `async` is not primitive, there are a lot more layers from there to execution

core/js/src/main/scala/cats/effect/IOFiberConstants.scala
SystemFw: oh wow. Btw once I get there, I really want to also cluster the operations differently, so switching order in `IOFiber` and `IO`. This might be annoying for people hacking on it so I'll check on gitter before starting, but I do think it will make `IOFiber` easier to digest.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -595,6 +513,146 @@ private final class IOFiber[A](
...
+          val cur = cur0.asInstanceOf[Get[Any]]
+          val state = cur.state
+
+            if (state.compareAndSet(ContStateInitial, ContStateWaiting)) {
+              // `state` was Initial, so `get` has arrived before the callback,
+              // needs to set to waiting and suspend: cb will resume with the result
+              // once that's ready
+
+              // full memory barrier
+              suspended.compareAndSet(false, true)
```
RaasAhsan: We were talking about this on Gitter earlier, but to reiterate why this was done (and to jog my memory about it). A CAS has the effect of both a volatile read and write, and so the non-volatile invariant extends here as well. Specifically, we want to observe the most up-to-date value of `cancelled`.

Additionally, this CAS should *always* succeed because the runloop is owned by the current thread.

core/shared/src/main/scala/cats/effect/IOFiber.scala
RaasAhsan: If we're running finalizers, that means `asyncCancel` was invoked at some point, which masks the runloop. If we're masked, `shouldFinalize()` evaluates to `false`, so we would proceed with `asyncContinue` in this case. In general, once we start finalization, subsequent calls to `shouldFinalize()` will always return `false`.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -595,6 +513,146 @@ private final class IOFiber[A](
...
+
+          val get: IO[Any] = IO.Get(state)
+          val cont = (get, cb)
+
+          runLoop(succeeded(cont, 0),  nextIteration)
+        case 22 =>
+          val cur = cur0.asInstanceOf[Get[Any]]
+          val state = cur.state
+
+            if (state.compareAndSet(ContStateInitial, ContStateWaiting)) {
```
RaasAhsan: extra indentation here?

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -595,6 +513,146 @@ private final class IOFiber[A](
...
+            }
+
+            stateLoop()
+          }
+
+          val get: IO[Any] = IO.Get(state)
+          val cont = (get, cb)
+
+          runLoop(succeeded(cont, 0),  nextIteration)
+        case 22 =>
```
RaasAhsan: 1. What is the effect of sequencing the same `get` multiple times? 
2. Do several invocations of `cont` interact with each other as expected with various interleavings of `get` and `resume`? 

I think its worth having test cases for these but feel free to open a follow up issue :)

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -595,6 +513,146 @@ private final class IOFiber[A](
...
+             * and is waiting for a result (i.e. it has suspended),
+             * acquire runloop to continue.
+             *
+             * If not, `cb` arrived first, so it just sets the result and die off.
+             *
+             * If `state` is `Result`, the callback has been already invoked, so no-op.
+             * (guards from double calls)
+             */
+            @tailrec
+            def stateLoop(): Unit = {
```
RaasAhsan, line 441: I was expecting otherwise, but it's interesting that this `stateLoop` stayed mostly intact from the old `Async`. I think that goes to say that old finalizer registration mechanism was just a more specific variant of what we have now

RaasAhsan (commented):
This is looking awesome, and much easier to understand!

core/shared/src/main/scala/cats/effect/IOFiber.scala
SystemFw: I'm going to expand that comment a bit to include some of this

core/shared/src/main/scala/cats/effect/IOFiber.scala
SystemFw: 1. Should be idempotent
2. I don't know, and this is worth checking to make sure there are no weird interactions. On one hand the number of interleavings is reduced by not being able to `start` `get`, but on the other I expect it to be fully possible to misuse this to deadlock (in a way that we cannot prevent). But still, worth checking that in sane usage we aren't messing up the runloop somehow

djspiewak:
I find it interesting that you seem to dislike line comments :-)

SystemFw:
> I find it interesting that you seem to dislike line comments :-)

I don't really. In this case I disliked having the debug printlns that I need to glance over interwoven with the stuff I need to read. So I started by adding `// LOG` to those lines, but then I moved to this commit. Not sure if I like the result or not (interlude: this is part of why I like working with granular commits and really avoid squashing on merge)

djspiewak:
> (interlude: this is part of why I like working with granular commits and really avoid squashing on merge)

Hard agree

kernel/shared/src/main/scala/cats/effect/kernel/internal/DefaultCont.scala
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+package cats.effect.kernel
+package internal
+
+import cats.syntax.all._
+import cats.arrow.FunctionK
+import java.util.concurrent.atomic.AtomicReference
+
+private[kernel] final class DefaultCont[F[_], A](implicit F: Async[F]) {
```
SystemFw: There's an interesting question as to how to test this

djspiewak: Maybe create an identity transformer which sits on top of `IO` but delegates to `IO`'s `async` rather than its `cont`?

SystemFw: and then run the entire IOSpec over that? that's fair enough, although ultimately it's still running on top of `IOCont`, which `IO.async` sits on top of

djspiewak: Yeah, but if we trust `IOCont`, then we trust its derived `async`

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -376,20 +364,28 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
...
-    async(cb => apply { k(cb); None })
+    effectForIO.async_(k)
 
   def canceled: IO[Unit] = Canceled
 
   def cede: IO[Unit] = Cede
 
+  /**
+   * This is a low-level API which is meant for implementors,
+   * please use `start`, `async` or `Deferred` instead,
```
RaasAhsan: Should we recommend `background` here instead

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -72,209 +74,135 @@ private final class IOFiber[A](
...
-    } else {
-      if (cb != null)
-        cb(RightUnit)
-
-      done(OutcomeCanceled)
-    }
-  }
-
-  // masks encoding: initMask => no masks, ++ => push, -- => pop
+  /* masks encoding: initMask => no masks, ++ => push, -- => pop */
```
RaasAhsan, line 193: is this supposed to be somewhere else

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -313,192 +275,267 @@ private final class IOFiber[A](
...
+            def loop(): Unit = {
+              // println(s"cb loop sees suspended ${suspended.get} on fiber $name")
+              /* try to take ownership of the runloop */
               if (resume()) {
-                state.lazySet(AsyncStateDone) // avoid leaks
-
-                // Race condition check:
-                // If finalization occurs and an async finalizer suspends the runloop,
-                // a window is created where a normal async resumes the runloop.
-                if (finalizing == wasFinalizing) {
```
RaasAhsan, line 370: This check was rather crucial: if a fiber is suspended in an asynchronous boundary and gets cancelled, the canceller will acquire the runloop and finalize. If one of the finalizers are asynchronous, the canceller enters an asynchronous boundary. At this point, if the callback from the *original* async completes, it's going to be able to acquire runloop, and try to apply the continuation on top of the stack to the result, which ultimately throws an exception.

If this isn't necessary anymore, we can remove the `finalizing` instance variable.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -313,192 +275,267 @@ private final class IOFiber[A](
...
-          val cur = cur0.asInstanceOf[Map[Any, Any]]
+          val get: IO[Any] = IOCont
+            .Get(state)
+            .onCancel(
+              /*
+               * If get gets canceled but the result hasn't been computed yet,
+               * restore the state to Initial to ensure a subsequent `Get` in
+               * a finalizer still works with the same logic.
+               */
+              IO(state.compareAndSet(ContStateWaiting, ContStateInitial)).void
```
RaasAhsan, line 469: ooh, this is a really good catch

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -567,42 +604,138 @@ private final class IOFiber[A](
...
+    } finally {
+      callbacks.lazySet(null) // avoid leaks
+    }
+
+    /*
+     * need to reset masks to 0 to terminate async callbacks
+     * busy spinning in `loop`.
+     */
+    masks = initMask
+    /* full memory barrier to publish masks */
```
RaasAhsan: I wrote this but it's wrong BTW, it's just a write barrier. 

Also, the reason this was added in: with the old `Async` state machine, after `Async` registered a finalizer and suspended, somebody else could `cancel` before the callback can acquire the runloop. In that scenario, the callback perpetually loops since the canceller owns the runloop now and masks before finalizing. 

After cancellation and finalizations completes, we reset `masks` which signals to the callback loop that the current fiber is both cancelled and completely unmasked, so somebody else handled cancellation. So the `!shouldFinalize()` check in `loop` fails, which terminates it

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -567,42 +604,138 @@ private final class IOFiber[A](
...
+    asyncContinueEither = null
+    blockingCur = null
+    afterBlockingSuccessfulResult = null
+    afterBlockingFailedError = null
+    evalOnIOA = null
+    resumeTag = DoneR
+  }
+
+  /*
+   4 possible cases for callback and cancellation:
```
RaasAhsan: I added this in awhile back, I think i came to the conclusion that this is wrong or just incomplete. But this behavior is better described by your new comments now :)

RaasAhsan (commented):
reviewed half of it, going to look at the rest tomorrow

core/shared/src/main/scala/cats/effect/IOFiber.scala
SystemFw: It was there before my changes as well. Not quite sure where to put it, it's a general point, which affects `childMask`, `initMask`, `masks`, `case 9` and `case 10` in the runloop (that's a candidate spot), `isUnmasked`, etc

core/shared/src/main/scala/cats/effect/IOFiber.scala
SystemFw: I need to think more about whether this is still needed or not. Please do the same :)
The reason I removed it is that with the old `async`, removing the check causes a test to consistently fail, and with `cont` (without the check), the same test consistently succeeds
https://github.com/typelevel/cats-effect/pull/1165#discussion_r492684017

SystemFw: If/when we all agree it's not needed, I'll go ahead and remove `finalizing`, thanks for pointing that out

core/shared/src/main/scala/cats/effect/IOFiber.scala
SystemFw: The solution here was surprisingly simple, I had gone down a really complicated avenue of double checking AtomicReferences before realising I could use `IO` itself and `onCancel`. This is the solution to "`get` within an `onCancel`" bug, ftr.

core/shared/src/main/scala/cats/effect/IOFiber.scala
SystemFw: What should I do here? Delete the full memory barrier comment for sure. Should I also add your comment above? is it still needed?

core/shared/src/main/scala/cats/effect/IOFiber.scala
SystemFw: Will delete then, cheers

core/shared/src/main/scala/cats/effect/IO.scala
SystemFw: yeah I will mention it there

djspiewak: Also oxford comma please :-)

core/shared/src/main/scala/cats/effect/IOFiber.scala
djspiewak: `"run the continuation of an async finalizer within async"` in `IOSpec` proves that this isn't necessary any longer, though it isn't immediately obvious to me *why* it is no longer necessary.

core/shared/src/main/scala/cats/effect/IO.scala
SystemFw: I actually love Oxford commas

core/shared/src/main/scala/cats/effect/IOFiber.scala
SystemFw: >  though it isn't immediately obvious to me why it is no longer necessary.

Exactly :)

core/shared/src/main/scala/cats/effect/IO.scala
djspiewak: They make lists so much less ambiguous

core/shared/src/main/scala/cats/effect/IOFiber.scala
RaasAhsan: Yeah, I believe it's still needed, I just wanted to make sure you and others are aware of why it's being done because it's not entirely clear (I had to spend 15 minutes recalling). Here's the race it fixes to be more clear:
1. Fiber A starts the `Cont` action.
2. Fiber A invokes `Cont.get`, which sets the `AtomicReference` to `ContStateWaiting`.
3. Before Fiber A suspends, the `Cont` action invokes the callback, and sets the `AtomicReference` to `ContStateResult`, and moves to `loop` to acquire the runloop, but it can't yet because fiber A hasn't suspended yet.
4. Fiber A suspends.
5. Fiber B cancels fiber A. Before the callback manages to acquire the runloop, B acquires the runloop, masks, finalization and terminates the fiber. At this point, assume we're not unmasking or publishing that state via `suspended`.
6. The callback for the `Cont` action is stuck in an infinite loop: it can't reacquire the runloop since it will never be released again, and the `!shouldFinalize()` will always pass because we're still masked.

So, by resetting `masks` and publishing that state, we can force that `!shouldFinalize` check to fail, which lets `loop` end.

SystemFw: Fantastic, I'll see if I can work that in a comment somewhere

core/shared/src/main/scala/cats/effect/IOFiber.scala
RaasAhsan: IIRC this was definitely a racey behavior that I observed for some test case over millions of invocations. So, I think there are two possible races here, one of which the finalizer on `Get` actually fixes:
1. Fiber A starts the `Cont` action.
2. Fiber A invokes `get`, suspending the runloop.
3. Fiber B cancels A, so it acquires A's runloop and starts finalization. The first finalizer is *guaranteed* to be the one registered over `Get`! So this resets the `AtomicReference` back to the `Initial` state.
4. B encounters an async finalizer for A, so it suspends the runloop. We will never encounter an async finalizer before the one over `Get`.
5. The callback from step 1 completes. It sees the `Initial` state so transitions to `Result` and exits out.

Here's an example of a race that it doesn't fix:
1. Fiber A starts the `Cont` action.
2. Fiber A calls `Get`, setting the `AtomicReference` to `Waiting`.
3. Before A suspends the runloop, the callback from step 1 completes. It sees the `Waiting` state, transitions to `Result`, and then tries to acquire the runloop in `loop`, which it can't at this point, so it loops.
4. A suspends the runloop.
5. Fiber B cancels A, acquiring the runloop. It begins finalization. The first finalizer resets the `AtomicReference` to `Initial`, as described above.
6. B encounters an async finalizer for A, so it suspends.
7. The callback from step 1 reacquires the runloop in `loop`. `!shouldFinalize()` evaluates to true because we are masked, so it moves onto `asyncContinue` which is where everything starts to break down.

I think that's right, but please double check that :D I wonder if there's a more obvious way to represent and compare against this fiber state, but it's not immediately obvious to me yet

djspiewak: I believe this is correct. It's a much more subtle and harder to hit case than the original one, but I believe the analysis is accurate that we don't handle it without checking that we're in the same finalization state as when we started.

As a note, we only need two finalization states because finalizers themselves are not cancelable. So we can only generate this scenario at a single level.

kernel/shared/src/main/scala/cats/effect/kernel/Cont.scala
```diff
@@ -0,0 +1,72 @@
...
+ * operations up to `MonadCancel` (`flatMap`, `onCancel`, `uncancelable`, etc).
+ *
+ * If you are an implementor, and you have an implementation of `async` but not `cont`,
+ * you can override `Async[F].async` with your implementation, and use `Async.defaultCont`
+ * to implement `Async[F].cont`.
+ */
+trait Cont[F[_], A] {
+  def apply[G[_]](
+      implicit
+      G: MonadCancel[G, Throwable]): (Either[Throwable, A] => Unit, G[A], F ~> G) => G[A]
```
djspiewak, line 71: Why not:

```scala
  def apply[G[_]](
      resume: Either[Throwable, A] => Unit,
      get: G[A],
      lift: F ~> G)(
      implicit G: MonadCancel[G, Throwable]): G[A]
```

More names and easier syntax at call sites.

SystemFw: > More names and easier syntax at call sites.

That's what I used to have, but actually the call site with that is _so_ much uglier

SystemFw: https://github.com/typelevel/cats-effect/pull/1165/commits/d2d23a2513cdb53b0ac573e132ae53c7e9c520c1#diff-8d42a1f7ffbea3d266c76d171c65d92c

have a look, note how many type signatures we can omit

SystemFw: Also I think the higher-rank type is clearer this way, although probably the only one who'll ever notice is me :P

djspiewak: Ah I see what you mean. Okay that's fair.

And yeah, I do think the higher-rank is clearer. That's actually the very first thing I noticed. :-) Had to stare at it a while to convince myself that it was actually okay to do what I was proposing. We can leave it as-is then.

kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
```diff
@@ -40,8 +53,21 @@ trait Async[F[_]] extends AsyncPlatform[F] with Sync[F] with Temporal[F] {
...
+   * provides an implementation of `cont` in terms of `async`.
+   * Note that if you use `defaultCont` you _have_ to override `async`.
+   */
+  def cont[A](body: Cont[F, A]): F[A]
 }
 
 object Async {
   def apply[F[_]](implicit F: Async[F]): F.type = F
+
+  def defaultCont[F[_]: Async, A](body: Cont[F, A]): F[A] =
```
djspiewak: If we make this `protected`, will subclasses of the companion trait be able to see it? That would be ideal, imo.

kernel/shared/src/main/scala/cats/effect/kernel/internal/DefaultCont.scala
```diff
@@ -0,0 +1,89 @@
...
+    loop()
+  }
+}
+private[kernel] object DefaultCont {
+  sealed trait State[A]
+  case class Initial[A]() extends State[A]
+  case class Value[A](v: Either[Throwable, A]) extends State[A]
+  case class Waiting[A](cb: Either[Throwable, A] => Unit) extends State[A]
+
+  def cont[F[_]: Async, A](body: Cont[F, A]): F[A] =
```
djspiewak: IMO it actually would be nicer to just inline this into `Async.defaultCont`. The types here don't serve any purpose other than redirecting the implementation to another file.

djspiewak (changes requested):

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -313,192 +275,285 @@ private final class IOFiber[A](
...
+              // println(s"cb loop sees suspended ${suspended.get} on fiber $name")
+              /* try to take ownership of the runloop */
               if (resume()) {
-                state.lazySet(AsyncStateDone) // avoid leaks
-
-                // Race condition check:
-                // If finalization occurs and an async finalizer suspends the runloop,
-                // a window is created where a normal async resumes the runloop.
-                if (finalizing == wasFinalizing) {
+                if (finalizing == wasFinalizing.get()) {
```
RaasAhsan, line 392: I *think* this looks good to me, will keep on thinking about it. Here's an alternative state coordination mechanism that makes somewhat more sense to me:
1. Renaming `wasFinalizing` to `waiting`. The purpose of this variable is to communicate to the callback whether the fiber is ready to resume from *this* callback specifically.
2. Add an extra statement to the `Get` finalizer that sets `waiting` to false.
3. Replace this conditional with `waiting.get()`. This basically indicates that the runloop suspended with the intention if waiting on this fiber. 
4. We can remove the `finalizing` field member.

If it's actually correct, I think this mechanism is slightly easier to understand, since it's completely independent of finalization. It's "if the runloop is waiting on me, then I can proceed" rather than "if the current runloop finalization state is the same as the state of whoever is currently waiting".

core/shared/src/test/scala/cats/effect/ContSpec.scala
```diff
@@ -0,0 +1,211 @@
...
+    val io = cont {
+      new Cont[IO, Int] {
+        def apply[F[_]: Cancelable] = { (resume, get, lift) =>
+          lift(IO(scheduler.sleep(10.millis, () => resume(Right(42))))) >> get >> get
+        }
+
+      }
+    }
+
+    val test = io.flatMap(r => IO(r mustEqual 42))
```
RaasAhsan, line 207: I think it would be nice to have tests that involve multiple `cont` calls with different orderings of `resume`/`get`. Let me know if I should just do it, would be a good opportunity to play around with this :)

kernel/shared/src/main/scala/cats/effect/kernel/internal/DefaultCont.scala
```diff
@@ -0,0 +1,89 @@
...
+
+  def get: F[A] =
+    F.defer {
+      state.get match {
+        case Value(v) => F.fromEither(v)
+        case Initial() =>
+          F.async { cb =>
+            val waiting = Waiting(cb)
+
+            def loop(): Unit =
```
RaasAhsan: `@tailrec` here just to be safe ?

kernel/shared/src/main/scala/cats/effect/kernel/internal/DefaultCont.scala
```diff
@@ -0,0 +1,89 @@
...
+           * - if a previous `get` has been interrupted and we are running as part of
+           *   its finalisers, the state would have been either restored to `Initial`
+           *   by the finaliser of that `get`, or set to `Value` by `resume`
+           */
+          sys.error("Impossible")
+      }
+    }
+
+  def resume(v: Either[Throwable, A]): Unit = {
+    def loop(): Unit =
```
RaasAhsan: `tailrec` here as well

core/shared/src/test/scala/cats/effect/ContSpec.scala
SystemFw: Go for it

kernel/shared/src/main/scala/cats/effect/kernel/internal/DefaultCont.scala
SystemFw: solved in the next commit, for some reason the diff doesn't show it

kernel/shared/src/main/scala/cats/effect/kernel/internal/DefaultCont.scala
SystemFw: yeah ok will do that

kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
SystemFw:
```
trait A
object A {
  protected def a = 1
}

trait B extends A {
  def b = A.a
}

method a in object A cannot be accessed in object A
[error]  Access to protected method a not permitted because
[error]  enclosing trait B is not a subclass of
[error]  object A where target is defined
[error]   def b = A.a
```

SystemFw: > method a in object A cannot be accessed in object A

it would be a good scala trivia to ask what code can generate this error lol

core/shared/src/main/scala/cats/effect/IOFiber.scala
SystemFw: I find the solution in the PR (and the one in the old async) a bit opaque.
I _think_ I prefer the direction with the `id` of the suspension, which acts as a sort of epoch number, because I think the specific problem is more accurately described as "we don't want a callback of a previous epoch to resume the runloop in a new epoch", with an orthogonal statement that says "a change of epoch happens during finalisation, in the presence of an async finaliser"

djspiewak: I think that @RaasAhsan's proposed solution is even *better* than the epoch number, actually, since we only need an epoch bit and his solution achieves it while removing the extra state from `IOFiber` itself. I really like that, but I think we should implement it in a follow-up PR.

The solution currently on the branch is definitely opaque, I just want to get this landed in a functional form so that we can unblock everything else. :-D

SystemFw: I'm fixing the conflicts as we speak, let's get it merged and then @RaasAhsan can add his solution and the additional tests in a follow up

PR #1168, opened 2020-09-07
Playing around with the CPST idea

djspiewak:
I only made changes in **kernel**. Just playing around with the abstraction a bit.

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
@@ -16,14 +16,31 @@
 
 package cats.effect.kernel
 
-import cats.implicits._
+import cats.syntax.all._
 
 import scala.concurrent.{ExecutionContext, Future}
 
 trait Async[F[_]] extends AsyncPlatform[F] with Sync[F] with Temporal[F, Throwable] {
 
+  type CPS[A]
+  def MonadCancelForCPS: MonadCancel[CPS, Throwable]
+  def liftToCPS[A](fa: F[A]): CPS[A]
+
+  def cps[A, B](body: (Either[Throwable, A] => Unit, CPS[A]) => CPS[B]): F[B]
+
   // returns an optional cancelation token
-  def async[A](k: (Either[Throwable, A] => Unit) => F[Option[F[Unit]]]): F[A]
+  def async[A](k: (Either[Throwable, A] => Unit) => F[Option[F[Unit]]]): F[A] =
+    cps[A, A] { (cb, get) =>
+      MonadCancelForCPS uncancelable { poll =>
+        MonadCancelForCPS.flatMap(liftToCPS(k(cb))) {
+          case Some(fin) =>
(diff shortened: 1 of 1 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
```diff
@@ -16,14 +16,31 @@
...
 package cats.effect.kernel
 
-import cats.implicits._
+import cats.syntax.all._
 
 import scala.concurrent.{ExecutionContext, Future}
 
 trait Async[F[_]] extends AsyncPlatform[F] with Sync[F] with Temporal[F, Throwable] {
 
+  type CPS[A]
```
djspiewak, line 25: Random note that I just realized: since we never have to provide a public concrete value for this type, we can just set it to… `F` itself in implementations. For example, `effectForIO` would be of type `Effect[IO]`, *without* any refinement to provide a concrete type for `CPS`. Internally, the concretization would be set as `type CPS[A] = IO[A]`

In other words, the existential only exists to get the compiler to give us a newtype, preventing unification between `F` and `CPS`, which in turn allows us to mask off some of `F`'s functionality (i.e. everything below `MonadCancel`).

RaasAhsan: > In other words, the existential only exists to get the compiler to give us a newtype, preventing unification between F and CPS

For this to work, implementations would need to explicitly type their implicit `Async` or `Effect` instance, otherwise the compiler *could* unify `CPS` and `F` right?

djspiewak: Nope. The opposite. The compiler won't unify CPS with anything unless given evidence to that effect. If implementors leave it entirely untyped, then CPS is simply a path dependent existential type.

RaasAhsan: Here's what I was thinking:
```scala
trait Read {
  type F[_]
  def readLine: F[String]
}

val impl/*: Read*/ = new Read {
  override type F[A] = IO[A]
  override def readLine: F[String] = IO(scala.io.StdIn.readLine())
}

val read: IO[String] = impl.readLine
```
If the type ascription is omitted, the compiler can infer the concrete type of the existential, otherwise it can't. Now obviously users wouldn't use `Async` instances by their assigned name; they would instead rely on implicit summoners which hides the existential, but I thought it was worth pointing out

djspiewak: Oh I see what you meant! I misunderstood and thought you meant the opposite. I agree with you.

SystemFw:
I think this is superseded by the current encoding now

djspiewak:
> I think this is superseded by the current encoding now

Also for posterity, the encoding I propose here has a very subtle issue exemplified by the following:

```scala
F.cps((_, get) => get.pure[F.CPS]).flatMap(leak => F.cps((_, _) => leak).start)
```

Hey look, I managed to get a `get` onto a different fiber!

The current encoding does not suffer from this issue.

PR #1173, opened 2020-09-09
Compile CE2 on dotty

### LINKED ISSUE #1166: Compile series/2.x on Dotty

vasilmkd:
Resolves #1166.

Depends on #1172.

Situation report:
I had to comment out one `ResourceSyntax` test, I have no idea how to compile it across all three compilers (2.12, 2.13 and dotty).

Dotty compiles with a couple of warnings (deprecation warnings on `MVar` which should be addressed by the dotty team soon, so that we can mute them) and some warnings in the `Resource` run loop, if I try to make dotty happy, Scala 2 complains about `@tailrec` and I didn't want anything to do with that.

All of the code compiles and all of the tests pass across all three versions locally.

And someone pretty please help me out with CI, I think it wants to create new workflows for dotty, IIUC. (Edit: I guess I did this?)

This PR is not that big, it contains all of the changes from #1172.


core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -629,17 +653,19 @@ abstract private[effect] class ResourceMonadError[F[_], E] extends ResourceMonad
...
-                 r match {
-                   case Left(error) => Resource.pure[F, Either[E, A]](Left(error))
-                   case Right(s)    => attempt(fs(s))
-                 })
+          Bind(
+            attempt(source),
+            (r: Either[E, Any]) =>
+              r match {
+                case Left(error) => Resource.pure[F, Either[E, A]](Left(error))
+                case Right(s)    => attempt(fs.asInstanceOf[Any => Resource[F, A]](s))
```
smarter: You can avoid the cast by porting https://github.com/typelevel/cats-effect/pull/1146 from ce3

vasilmkd: Done, thank you very much @smarter.

vasilmkd:
In CI, the 0.26.0 compiler fails with a `StackOverflowError` while the tests on 0.27.0-RC1 seem to complete but then there are non-sensical dependency resolution errors (looking for cats and cats-laws versions for dotty 0.27.0-RC1 which don't exist).

smarter:
Yeah there was a bug in 0.26 which was fixed in 0.27, ce3 dropped 0.26 from their build.

build.sbt
```diff
@@ -222,13 +229,29 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
...
-      compilerPlugin(("com.github.ghik" % "silencer-plugin" % SilencerVersion).cross(CrossVersion.full)),
-      ("com.github.ghik" % "silencer-lib" % SilencerVersion % "provided").cross(CrossVersion.full),
-      ("com.github.ghik" % "silencer-lib" % SilencerVersion % Test).cross(CrossVersion.full)
-    )
+    libraryDependencies ++= {
+      if (isDotty.value)
+        Seq(
+          // Only way to properly resolve this library
+          ("com.github.ghik" % "silencer-lib_2.13.3" % SilencerVersion % Provided)
+        )
```
djspiewak: TBH it surprises me that this works at all.

build.sbt
```diff
@@ -222,13 +229,29 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
...
+      else
+        Seq(
+          compilerPlugin(("com.github.ghik" % "silencer-plugin" % SilencerVersion).cross(CrossVersion.full)),
+          ("com.github.ghik" % "silencer-lib" % SilencerVersion % "provided").cross(CrossVersion.full),
+          ("com.github.ghik" % "silencer-lib" % SilencerVersion % Test).cross(CrossVersion.full)
+        )
+    }
   )
+  .settings(dottyLibrarySettings)
+  .settings(dottyJsSettings(ThisBuild / crossScalaVersions))
```
djspiewak: We can probably skip this! These settings actually straight-up remove ScalaJS for Dotty, which is no longer necessary as of 0.27

core/shared/src/main/scala/cats/effect/Resource.scala
```diff
@@ -169,8 +175,8 @@ sealed abstract class Resource[+F[_], +A] {
    *
    **/
   def parZip[G[x] >: F[x]: Sync: Parallel, B](
-    that: Resource[G[?], B]
-  ): Resource[G[?], (A, B)] = {
+    that: Resource[G, B]
+  ): Resource[G, (A, B)] = {
```
djspiewak: We need to make sure that this doesn't regress #807. I'm not sure if we have tests for that, but we should.

core/shared/src/test/scala/cats/effect/ResourceSyntax.scala
```diff
@@ -30,16 +30,21 @@ object Test {
...
+  }
 
-  // this one fails, but not the above
-  def g[F[+_]](implicit F: Sync[F]): Resource[F, Either[Base, Nothing]] =
-    Resource
-      .liftF[F, Either[Base, Nothing]](f[F])
-      .map(x => x)
-      .flatMap(x => Resource.pure[F, Either[Base, Nothing]](x))
-
+  // TODO: find a way to fix this that works for Scala 2.12, Scala 2.13 and dotty
```
djspiewak: Ah, here's the regression! I think the only answer is to duplicate `Resource` for now and have 2.x and 3.x versions, which sucks.

build.sbt
vasilmkd: I mean, it's almost the same as downloading the jar. This will break sooner or later, but should be fine for now.

build.sbt
vasilmkd: I just got 0.26.0 working, so I propose we keep this until next time.

vasilmkd: I shamelessly copied code from your plugin to work around this. 0.26.0 is paired with 2.13.3 for Scala.js. 0.27.0-RC1 works both on jvm and js.

vasilmkd: Hmmm, I got 0.26.0 to work, so I might need some help on how to tackle this now.

core/shared/src/test/scala/cats/effect/ResourceSyntax.scala
vasilmkd: I split it into 2.12 specific and 2.13+, since 2.13 doesn't have a problem with this.

core/shared/src/main/scala/cats/effect/Resource.scala
vasilmkd: Fixed by duplicating Resource into 2.12 specific and 2.13+ sources.

build.sbt
vasilmkd: I put a comment where I used the workaround I could come up with.

vasilmkd:
The PR became huge because I had to duplicate `Resource` for Scala 2.12. My work here is mostly done, I've fixed all of the comments.

djspiewak (approved):
🎉 Yay!

PR #1183, opened 2020-09-13
Refactor Resource interpreters to use (mostly) static typing

bplommer:


djspiewak:
This is very very cool! I'm actually astonished that you were able to get it to compile as-written on all three platforms; usually the tricks required to get 2.12 to accept something involving GADTs are (rightly) flagged as being unsound on Dotty. I wanna leave this open for a sec to get @SystemFw's take, but I'm very much 👍 on this change!

SystemFw:
> ! I'm actually astonished that you were able to get it to compile as-written on all three platforms; 

ahah that's exactly what I was thinking. Seems to work though, good job Ben!

vasilmkd:
We should do something similar for CE2. I had to do some simple changes to please dotty. Not to forget, we have two versions of Resource in CE2 now, one for Scala 2.12 and one for dotty.

PR #1190, opened 2020-09-14
Remove order instances now we have a new version of cats

TimWSpence:


djspiewak:
Good call! I completely forgot about this.

PR #1191, opened 2020-09-14
Fix traces spam in tests

vasilmkd:
`CatsEffectSuite` is the only relevant class for this PR.

### DIFF
--- a/core/shared/src/test/scala/cats/effect/CatsEffectSuite.scala
+++ b/core/shared/src/test/scala/cats/effect/CatsEffectSuite.scala
@@ -16,7 +16,9 @@
 
 package cats.effect
 
+import cats.effect.internals.TestUtils
 import munit.{DisciplineSuite, Location}
+import org.scalacheck.Prop
 
 /**
  * A stopgap solution until upstream dependencies (cats, cats-laws and discipline-munit for dotty Scala.js) are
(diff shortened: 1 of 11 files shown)


core/shared/src/test/scala/cats/effect/CatsEffectSuite.scala
```diff
@@ -26,6 +31,37 @@ import munit.{DisciplineSuite, Location}
...
+      lock.lock()
+      val oldErr = System.err
+      val errStream = new ByteArrayOutputStream()
+      val fakeErr = new PrintStream(errStream)
+      val passed = Prop.Result(Prop.True)
+
+      for {
+        _ <- Prop { _ => System.setErr(fakeErr); passed }
+        res <- body
+        _ <- Prop { _ => System.setErr(oldErr); fakeErr.close(); lock.unlock(); passed }
```
djspiewak: Hmm, I think we probably need to be a bit more imperative here with a `try`/`finally`, because this isn't going to correctly handle the case where `body` throws an exception.

vasilmkd: I believe `body` is a `org.scalacheck.Prop` and therefore shouldn't _throw_ exceptions.

vasilmkd: Nevermind, I get what you're saying, I might have another idea.

vasilmkd: Fixed, this is gone.

PR #1192, opened 2020-09-15
Gen hierarchy renaming

RaasAhsan:
- [x] Prefix `Concurrent`, `Temporal` and `Spawn` type classes with `Gen`
- [x] Remove `Throw` suffix from `ConcurrentThrow`, `TemporalThrow` and `SpawnThrow`
- [x] Add type alias `MonadThrow[F]` to `MonadError[F, Throwable]`

Made a best effort to remain as consistent as possible, let me know if I missed anything, or if some refactorings were too aggressive.


concurrent/shared/src/main/scala/cats/effect/concurrent/Semaphore.scala
```diff
@@ -134,7 +134,7 @@ object Semaphore {
   private type State[F[_]] = Either[Queue[(Long, Deferred[F, Unit])], Long]
 
   abstract private class AbstractSemaphore[F[_]](state: Ref[F, State[F]])(
-      implicit F: Spawn[F, Throwable])
+      implicit F: GenSpawn[F, Throwable])
```
djspiewak: `Spawn[F]`

core/shared/src/main/scala/cats/effect/instances/package.scala
```diff
@@ -20,5 +20,5 @@ package object instances {
 
   object all extends AllInstances
 
-  object concurrent extends kernel.instances.SpawnInstances
+  object concurrent extends kernel.instances.GenSpawnInstances
```
djspiewak: This should probably be `object spawn`

core/shared/src/main/scala/cats/effect/package.scala
```diff
@@ -53,10 +53,11 @@ package object effect {
...
   val Effect = cekernel.Effect
 
+  type MonadThrow[F[_]] = cekernel.MonadThrow[F]
   type MonadCancelThrow[F[_]] = cekernel.MonadCancelThrow[F]
-  type SpawnThrow[F[_]] = cekernel.SpawnThrow[F]
-  type ConcurrentThrow[F[_]] = cekernel.ConcurrentThrow[F]
-  type TemporalThrow[F[_]] = cekernel.TemporalThrow[F]
+  type Spawn[F[_]] = cekernel.Spawn[F]
+  type Concurrent[F[_]] = cekernel.Concurrent[F]
+  type Temporal[F[_]] = cekernel.Temporal[F]
```
djspiewak: We should also do value-level forwarders, so that we can use the summoning syntax. IMO, `val Spawn = GenSpawn` is sufficient

core/shared/src/main/scala/cats/effect/syntax/package.scala
```diff
@@ -20,8 +20,8 @@ package object syntax {
 
   object all extends AllSyntax
 
-  object spawn extends kernel.syntax.SpawnSyntax
-  object temporal extends kernel.syntax.TemporalSyntax
+  object spawn extends kernel.syntax.GenSpawnSyntax
```
djspiewak, line 23: We should add `concurrent` here

kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
```diff
@@ -20,7 +20,7 @@ import cats.implicits._
 
 import scala.concurrent.{ExecutionContext, Future}
 
-trait Async[F[_]] extends AsyncPlatform[F] with Sync[F] with Temporal[F, Throwable] {
+trait Async[F[_]] extends AsyncPlatform[F] with Sync[F] with GenTemporal[F, Throwable] {
```
djspiewak: `Temporal`

kernel/shared/src/main/scala/cats/effect/kernel/Fiber.scala
```diff
@@ -22,9 +22,9 @@ trait Fiber[F[_], E, A] {
   def cancel: F[Unit]
   def join: F[Outcome[F, E, A]]
 
-  def joinAndEmbed(onCancel: F[A])(implicit F: Spawn[F, E]): F[A] =
+  def joinAndEmbed(onCancel: F[A])(implicit F: GenSpawn[F, E]): F[A] =
```
djspiewak: We can loosen this to `MonadCancel`

kernel/shared/src/main/scala/cats/effect/kernel/package.scala
```diff
@@ -16,12 +16,15 @@
...
 package object kernel {
 
+  type MonadThrow[F[_]] = MonadError[F, Throwable]
   type MonadCancelThrow[F[_]] = MonadCancel[F, Throwable]
-  type SpawnThrow[F[_]] = Spawn[F, Throwable]
-  type TemporalThrow[F[_]] = Temporal[F, Throwable]
-  type ConcurrentThrow[F[_]] = Concurrent[F, Throwable]
+  type Spawn[F[_]] = GenSpawn[F, Throwable]
+  type Temporal[F[_]] = GenTemporal[F, Throwable]
+  type Concurrent[F[_]] = GenConcurrent[F, Throwable]
```
djspiewak: Needs value-level forwarders

kernel/shared/src/main/scala/cats/effect/kernel/syntax/GenTemporalSyntax.scala
```diff
@@ -16,32 +16,33 @@
...
-  def timeoutTo(duration: FiniteDuration, fallback: F[A])(implicit F: Temporal[F, E]): F[A] =
+  def timeoutTo(duration: FiniteDuration, fallback: F[A])(implicit F: GenTemporal[F, E]): F[A] =
     F.timeoutTo(wrapped, duration, fallback)
 }
 
 final class TemporalTimeoutOps[F[_], A](val wrapped: F[A]) extends AnyVal {
 
-  def timeout(duration: FiniteDuration)(implicit F: Temporal[F, _ >: TimeoutException]): F[A] =
+  def timeout(duration: FiniteDuration)(
+      implicit F: GenTemporal[F, _ >: TimeoutException]): F[A] =
```
djspiewak: Let's just make this `Temporal`

kernel/shared/src/main/scala/cats/effect/kernel/syntax/package.scala
```diff
@@ -20,8 +20,8 @@ package object syntax {
 
   object all extends AllSyntax
 
-  object spawn extends SpawnSyntax
-  object temporal extends TemporalSyntax
+  object spawn extends GenSpawnSyntax
+  object temporal extends GenTemporalSyntax
```
djspiewak, line 24: We need to add `concurrent`

mpilquist:
Looks great. One thing I noticed when converting FS2 to use this PR is that I really miss `Concurrent[F]` syntax for summoning an implicit instance.

djspiewak (changes requested):
👍 looks really good! Just some minor changes.

core/shared/src/main/scala/cats/effect/syntax/package.scala
RaasAhsan: We don't have any `Concurrent` syntax enrichment yet

kernel/shared/src/main/scala/cats/effect/kernel/syntax/package.scala
RaasAhsan: No syntax for `Concurrent` yet

core/shared/src/main/scala/cats/effect/syntax/package.scala
djspiewak: That's true, I just wanted to create a place for it. I can just do that in a follow-up PR.

mpilquist:
@djspiewak Can we get a release of this one?

djspiewak:
For posterity, already done: `3.0-c54025b`

PR #1194, opened 2020-09-15
Disallow null in IO pure, delay and defer

### LINKED ISSUE #1178: IO.defer(null) probably doesn't do what you expect
I haven't tried it, but I suspect that it hangs forever. It *should* be a `NullPointerException`. This is a scary surfacing of `IOFiber`'s internals.

vasilmkd:
Resolves #1178.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -277,9 +277,8 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
 
         case _: IO.Cede.type => F.cede.asInstanceOf[F[A]]
 
-        case IO.UnmaskRunLoop(_, _) =>
-          // Will never be executed. Case demanded for exhaustiveness.
-          // ioa.to[F] // polling should be handled by F
+        case IO.UnmaskRunLoop(_, _) | IO.BlockFiber =>
+          // Will never be executed. Cases demanded for exhaustiveness.
           sys.error("impossible")
 
         case self: IO.Attempt[_] =>
@@ -708,6 +707,10 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
   // Not part of the run loop. Only used in the implementation of IO#to.
   private[effect] final case class UnmaskTo[F[_], +A](ioa: IO[A], poll: Poll[F]) extends IO[A] {
+    def tag = -1
+  }
+
+  private[effect] case object BlockFiber extends IO[Nothing] {
     def tag = -1
   }
 }
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -110,7 +110,6 @@ private final class IOFiber[A](
 
   // mutable state for resuming the fiber in different states
   private[this] var resumeTag: Byte = ExecR
-  private[this] var resumeNextIteration: Int = 0
   private[this] var asyncContinueEither: Either[Throwable, Any] = _
   private[this] var blockingCur: Blocking[Any] = _
   private[this] var afterBlockingSuccessfulResult: Any = _
@@ -122,6 +121,9 @@ private final class IOFiber[A](
 
   // similar prefetch for Outcome
   private[this] val OutcomeCanceled = IOFiber.OutcomeCanceled.asInstanceOf[OutcomeIO[A]]
+
+  // similar prefetch for BlockFiber
+  private[this] val IOBlockFiber = IO.BlockFiber
 
   var cancel: IO[Unit] = IO uncancelable { _ =>
     IO defer {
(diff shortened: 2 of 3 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -387,7 +387,9 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
   private[this] val _never: IO[Nothing] = async(_ => pure(None))
   def never[A]: IO[A] = _never
 
-  def pure[A](value: A): IO[A] = Pure(value)
+  def pure[A](value: A): IO[A] =
+    if (value == null) raiseError(new NullPointerException())
+    else Pure(value)
```
djspiewak: I'm almost 100% positive that we need to support `IO.pure(null)`. I certainly can't see a reason why we would forbid it, and CE2 supports it.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -298,8 +298,9 @@ private final class IOFiber[A](
             }
 
           val next =
-            if (error == null) succeeded(r, 0)
-            else failed(error, 0)
+            if (error != null) failed(error, 0)
+            else if (r == null) failed(new NullPointerException(), 0)
+            else succeeded(r, 0)
```
djspiewak: Similarly, there's no reason to forbid `delay(null)`, and CE2 allows it.

core/shared/src/test/scala/cats/effect/IOSpec.scala
```diff
@@ -204,6 +204,27 @@ class IOSpec extends IOPlatformSpecification with Discipline with ScalaCheck wit
...
+        IO(null)
+          .attempt
+          .map(_.left.toOption.get.isInstanceOf[NullPointerException]) must completeAs(true)
+        IO.delay(null)
+          .attempt
+          .map(_.left.toOption.get.isInstanceOf[NullPointerException]) must completeAs(true)
+      }
+
+      "result in an NPE if deferring a null IO" in ticked { implicit ticker =>
+        IO.defer(null)
```
djspiewak, line 218: *Deferring* `null` is what we should forbid. Here's what it looks like when you try this in CE2:

```
@ cats.effect.IO.suspend(null).unsafeRunSync() 
scala.MatchError: null
  cats.effect.internals.IORunLoop$.cats$effect$internals$IORunLoop$$loop(IORunLoop.scala:83)
  cats.effect.internals.IORunLoop$.$anonfun$suspendAsync$1(IORunLoop.scala:315)
  cats.effect.internals.IORunLoop$.$anonfun$suspendAsync$1$adapted(IORunLoop.scala:314)
  cats.effect.internals.IORunLoop$RestartCallback.start(IORunLoop.scala:439)
  cats.effect.internals.IORunLoop$.cats$effect$internals$IORunLoop$$loop(IORunLoop.scala:155)
  cats.effect.internals.IORunLoop$.start(IORunLoop.scala:37)
  cats.effect.IO.unsafeRunAsync(IO.scala:274)
  cats.effect.internals.IOPlatform$.unsafeResync(IOPlatform.scala:39)
  cats.effect.IO.unsafeRunTimed(IO.scala:342)
  cats.effect.IO.unsafeRunSync(IO.scala:256)
  ammonite.$sess.cmd3$.<clinit>(cmd3.sc:1)
```

We can probably do a `NullPointerException` instead, but tldr we should check for it. I believe the correct place to look is in the `FlatMap` hander. Specifically, if the `f` continuation returns `null`, we should raise an NPE.

djspiewak (changes requested):

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -265,12 +267,19 @@ private final class IOFiber[A](
 
   // masks encoding: initMask => no masks, ++ => push, -- => pop
   @tailrec
-  private[this] def runLoop(cur0: IO[Any], iteration: Int): Unit = {
-    // cur0 will be null when we're semantically blocked
-    if (cur0 == null) {
+  private[this] def runLoop(cur: IO[Any], iteration: Int): Unit = {
+    // cur will be set to BlockFiber when we're semantically blocked
+    if (cur == IOBlockFiber) {
```
djspiewak: `eq` is faster

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -265,12 +267,19 @@ private final class IOFiber[A](
 
   // masks encoding: initMask => no masks, ++ => push, -- => pop
   @tailrec
-  private[this] def runLoop(cur0: IO[Any], iteration: Int): Unit = {
-    // cur0 will be null when we're semantically blocked
-    if (cur0 == null) {
+  private[this] def runLoop(cur: IO[Any], iteration: Int): Unit = {
```
djspiewak: Given that we use `cur` within the `match`, I'd kind of prefer some other name here. It definitely gets ugly though. Could just use `cur00` like we used to, but either way it's moderately annoying.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -265,12 +267,19 @@ private final class IOFiber[A](
...
+    // cur will be set to BlockFiber when we're semantically blocked
+    if (cur == IOBlockFiber) {
       return
     }
 
+    // Null IO, blow up
+    val cur0: IO[Any] = if (cur == null) {
+      Error(new NullPointerException())
```
djspiewak: We should check, but this might not get optimized correctly fwiw because of the indirection through the `val`. Better off doing `IO.Error(new NullPointerException())`

djspiewak (changes requested):

PR #1196, opened 2020-09-16
Remove unsound pattern matches from Resource

bplommer:
Fixes compiler warnings from Dotty (pointed out in https://github.com/typelevel/cats-effect/pull/1187#issuecomment-691920199). Unfortunately Scala 2 can't handle the pattern matches if they're written soundly with variables, so the only way I could find to do it was OOP-style using package-private instance methods on the Resource subclasses.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Resource.scala
+++ b/core/shared/src/main/scala/cats/effect/Resource.scala
@@ -103,6 +103,8 @@ sealed abstract class Resource[+F[_], +A] {
       onOutput: A => G[B],
       onRelease: G[Unit] => G[Unit]
   )(implicit G: Resource.Bracket[G]): G[B] = {
+    import Resource.Invariant
+
     sealed trait Stack[AA]
     case object Nil extends Stack[A]
     final case class Frame[AA, BB](head: AA => Resource[G, BB], tail: Stack[BB])
@@ -115,9 +117,9 @@ sealed abstract class Resource[+F[_], +A] {
     // Interpreter that knows how to evaluate a Resource data structure;
     // Maintains its own stack for dealing with Bind chains
     @tailrec def loop[C](current: Resource[G, C], stack: Stack[C]): G[B] =
-      current match {
-        case a: Allocate[G, C] =>
-          G.bracketCase(a.resource) {
+      current.invariant[G, C] match {
+        case Invariant.Allocate(resource) =>
+          G.bracketCase(resource) {
             case (a, _) =>
               stack match {
                 case Nil => onOutput(a)
@@ -127,10 +129,10 @@ sealed abstract class Resource[+F[_], +A] {
             case ((_, release), ec) =>
               onRelease(release(ec))
           }
-        case b: Bind[G, _, C] =>
-          loop(b.source, Frame(b.fs, stack))
-        case s: Suspend[G, C] =>
-          s.resource.flatMap(continue(_, stack))
+        case Invariant.Bind(source, fs) =>
+          loop(source, Frame(fs, stack))
+        case Invariant.Suspend(resource) =>
+          resource.flatMap(continue(_, stack))
       }
     loop(this, Nil)
   }
@@ -259,6 +261,8 @@ sealed abstract class Resource[+F[_], +A] {
    * resource.
    */
   def allocated[G[x] >: F[x], B >: A](implicit G: Resource.Bracket[G]): G[(B, G[Unit])] = {
+    import Resource.Invariant
+
     sealed trait Stack[AA]
     case object Nil extends Stack[B]
     final case class Frame[AA, BB](head: AA => Resource[G, BB], tail: Stack[BB])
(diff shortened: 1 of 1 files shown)


djspiewak (commented):
This is fantastic work, but doesn't it double the number of allocations required to interpret a `Resource`? Is that a penalty we're willing to accept?

bplommer:
> doesn't it double the number of allocations required to interpret a `Resource`? Is that a penalty we're willing to accept?

Yes, good point. I guess I just assumed that wasn't too major a concern, given that the interpreters haven't been particularly aggressively optimised. If it _is_ a serious concern, though, shouldn't we go the other way and change the stacks from a linked structure to a mutable buffer?

djspiewak:
> Yes, good point. I guess I just assumed that wasn't too major a concern, given that the interpreters haven't been particularly aggressively optimised. If it is a serious concern, though, shouldn't we go the other way and change the stacks from a linked structure to a mutable buffer?

TBH I'm not sure. This is definitely an area where I could use more opinions. :-)

smarter:
If you're not afraid to go full-OOP you could also try replacing the pattern match by calls to abstract methods in Resource with implementations in each subclasses, no idea if that makes sense here :).

bplommer:
> If you're not afraid to go full-OOP you could also try replacing the pattern match by calls to abstract methods in Resource with implementations in each subclasses, no idea if that makes sense here :).

That was what I tried first, but you lose the tail recursion 😞

SystemFw:
I'd definitely like to keep the pattern match, I think it really helps clarity here, and the code isn't trivial.

djspiewak:
> I'd definitely like to keep the pattern match, I think it really helps clarity here, and the code isn't trivial.

Do you think it's worth doubling up on the allocations in order to make the code easier to follow? (i.e. getting the typechecks)

smarter:
>  Unfortunately Scala 2 can't handle the pattern matches if they're written soundly with variables

What kind of errors do you get by the way?

SystemFw:
> Do you think it's worth doubling up on the allocations in order to make the code easier to follow? (i.e. getting the typechecks)

I haven't addressed that because I'm not sure. I suspect there are other things we want to tackle here, primarily stack safety, which might change whatever decision we take here anyway

bplommer:
Another option that seems to work, that would avoid runtime overhead, is to add `private[effect] sealed trait InvariantResource[F[_], +A] extends Resource[F, A]` to the type hierarchy and have the concrete types extend that - then the conversion can just be `def invariant[G[x] >: F[x]](implicit F: Functor[G]): Resource.Invariant[G, A] = asInstanceOf`. So you still have one use of `asInstanceOf`, but it's segregated away as the implementation of a method that makes a sound conversion.

bplommer:
> > Unfortunately Scala 2 can't handle the pattern matches if they're written soundly with variables
> 
> What kind of errors do you get by the way?
The following works in Dotty.
```scala
current match {
  case a: Allocate[_, _] =>
    G.bracketCase(a.resource) {
      case (a, _) =>
        stack match {
          case Nil => onOutput(a)
          case Frame(head, tail) => continue(head(a), tail)
        }
    } {
      case ((_, release), ec) =>
        onRelease(release(ec))
    }
  case b: Bind[_, _, C] =>
    loop(b.source, Frame(b.fs, stack))
  case s: Suspend[_, _] =>
    G.flatMap(s.resource)(continue(_, stack))
}
```
In 2.12 I get these errors:
```scala
[error] /Users/benplom/oss/cats-effect/core/shared/src/main/scala/cats/effect/Resource.scala:109:57: G takes no type parameters, expected: one
[error]     final case class Frame[AA, BB](head: AA => Resource[G, BB], tail: Stack[BB])
[error]                                                         ^
[error] /Users/benplom/oss/cats-effect/core/shared/src/main/scala/cats/effect/Resource.scala:122:18: constructor cannot be instantiated to expected type;
[error]  found   : (T1, T2)
[error]  required: Nothing
[error]             case (a, _) =>
[error]                  ^
[warn] /Users/benplom/oss/cats-effect/core/shared/src/main/scala/cats/effect/Resource.scala:121:37: dead code following this construct
[warn]           G.bracketCase(a.resource) {
[warn]                                     ^
[error] /Users/benplom/oss/cats-effect/core/shared/src/main/scala/cats/effect/Resource.scala:129:32: Any does not take parameters
[error]               onRelease(release(ec))
[error]              
```
And in 2.13 these:
```scala
[error] /Users/benplom/oss/cats-effect/core/shared/src/main/scala/cats/effect/Resource.scala:124:38: type mismatch;
[error]  found   : a.type (with underlying type _)
[error]  required: A
[error]                 case Nil => onOutput(a)
[error]                                      ^
[error] /Users/benplom/oss/cats-effect/core/shared/src/main/scala/cats/effect/Resource.scala:129:32: Any does not take parameters
[error]               onRelease(release(ec))
[error]                                ^
```

djspiewak:
> So you still have one use of asInstanceOf, but it's segregated away as the implementation of a method that makes a sound conversion.

This is incredibly smart.

bplommer:
> > So you still have one use of asInstanceOf, but it's segregated away as the implementation of a method that makes a sound conversion.
> 
> This is incredibly smart.

Thank you 😁 I updated the PR to use this approach - I think it's generally cleaner, and it removes the extra allocations. Thoughts?

djspiewak:
Amazing work. I would imagine this same trick would work for `fs2.Stream` if they chose to employ it. It reifies the argument that "`Functor` means this is safe". Fantastic.

PR #1197, opened 2020-09-16
Add Console

### LINKED ISSUE #1159: Console for IO
Printing to and reading from standard streams is a fairly pervasive operation, maybe less so in real codebases but certainly in numerous examples and gists in the wild. I've certainly seen and written a ridiculous amount of `IO(println("foo"))`. If anything but for better ergonomics and out-of-the-box experience , we should think about introducing some utility functions to cats-effect to serve that need.

I like throwing out lists of bulletpoints for every issue I create so I don't have to answer these questions, so here's another one:
* Should we target `IO` or `F`?
* What kind of API should we expose? I think `putStrLn` and `getStrLn` are pretty common operations to write, but not sure if there are more interesting operations to include. Another question is if these should be available in `IO` companion object or available as imports.
* `Show` integration
* Cats Effect positions itself at the base of an ecosystem and relies on the greater community to build out libraries that fulfill the needs of users. [https://github.com/profunktor/console4cats](console4cats) already fills the need in this space, so we would potentially be replacing that library (at least for Cats Effect 3).

vasilmkd:
We can iterate on this. I definitely couldn't please everyone. Could also serve as a place to continue the discussion.

Special thanks to @gvolpe for `console4cats`.

This still needs transformer instances. I will add them. I opened this so that we can have an early review.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -643,11 +643,6 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
     def tag = 20
   }
 
-  // Not part of the run loop. Only used in the implementation of IO#to.
-  private[effect] final case class UnmaskTo[F[_], +A](ioa: IO[A], poll: Poll[F]) extends IO[A] {
-    def tag = -1
-  }
-
   private[effect] case object BlockFiber extends IO[Nothing] {
     def tag = -1
   }


concurrent/shared/src/main/scala/cats/effect/concurrent/Console.scala
```diff
@@ -0,0 +1,121 @@
...
+   * @param err error channel
+   * @param out output channel
+   * @return Console[F[_]: Sync] instance
+   */
+  def make[F[_]: Sync](
+      in: BufferedReader,
+      err: PrintStream,
+      out: PrintStream
+  ): Console[F] =
+    new SyncConsole[F](in, err, out)
```
vasilmkd: I just adapted this code from `console4cats`, however I do think that we can provide a better user experience by creating it, I dare say in a `Resource`, which could then be used to close the readers and the streams at the end.

concurrent/shared/src/main/scala/cats/effect/concurrent/Console.scala
```diff
@@ -0,0 +1,121 @@
...
+      def readLine: G[String] = f(self.readLine)
+      def print[A: Show](a: A): G[Unit] = f(self.print(a))
+      def println[A: Show](a: A): G[Unit] = f(self.println(a))
+      def printError[A: Show](a: A): G[Unit] = f(self.printError(a))
+      def printlnError[A: Show](a: A): G[Unit] = f(self.printlnError(a))
+    }
+}
+
+object Console {
+  def apply[F[_]](implicit C: Console[F]): C.type = C
```
djspiewak: So the more I think about it, the more I'm convinced that this really shouldn't be a final tagless encoding. I think the `Console[F]` trait itself is fine, but I don't think it should *ever* be passed implicitly. It just doesn't make sense and there are so many drawbacks to it.

concurrent/shared/src/main/scala/cats/effect/concurrent/SyncConsole.scala
```diff
@@ -0,0 +1,44 @@
...
+
+import java.io.{BufferedReader, PrintStream}
+
+private class SyncConsole[F[_]: Sync](
+    in: BufferedReader,
+    err: PrintStream,
+    out: PrintStream
+) extends Console[F] {
+  def readLine: F[String] =
+    Sync[F].delay(in.readLine())
```
djspiewak: This should definitely be in `blocking` instead of `delay`, and probably in `interruptible`.

concurrent/shared/src/main/scala/cats/effect/concurrent/SyncConsole.scala
```diff
@@ -0,0 +1,44 @@
...
+    Sync[F].delay(out.print(a.show))
+
+  def println[A: Show](a: A): F[Unit] =
+    Sync[F].delay(out.println(a.show))
+
+  def printError[A: Show](a: A): F[Unit] =
+    Sync[F].delay(err.print(a.show))
+
+  def printlnError[A: Show](a: A): F[Unit] =
+    Sync[F].delay(err.println(a.show))
```
djspiewak: All of these should definitely be in `blocking` or possibly `interruptible`.

concurrent/shared/src/test/scala/cats/effect/concurrent/SyntaxSpec.scala
```diff
@@ -40,6 +40,7 @@ class SyntaxSpec extends Specification {
 
   def sync[F[_]](implicit F: Sync[F]) = {
     Ref.of[F, Int](0)
+    Console.stdio[F]
```
djspiewak: IMO this doesn't need to be in the syntax spec

concurrent/shared/src/test/scala/cats/effect/concurrent/SyntaxSpec.scala
```diff
@@ -49,4 +50,7 @@ class SyntaxSpec extends Specification {
 
   def semaphoreIsDeriveable[F[_]](implicit F: Concurrent[F]) =
     Semaphore[F](11)
+
+  def console[F[_]: Console] =
+    Console[F]
```
djspiewak: …same here

djspiewak (changes requested):
I think we should probably add @gvolpe to the list of developers in the build (is he already in there? is anybody? I haven't checked) in recognition of the fact that we're basically slurping his library directly into Cats Effect. :-)

concurrent/shared/src/main/scala/cats/effect/concurrent/SyncConsole.scala
vasilmkd: Yes, sorry, this PR needs more polish. I only opened it to have something that others can comment on with context and possibly see what works and what doesn't.

concurrent/shared/src/main/scala/cats/effect/concurrent/Console.scala
vasilmkd: Do you maybe have more concrete suggestions on how to proceed from here? Thanks.

vasilmkd:
What is a list of developers? I will gladly add him.

concurrent/shared/src/main/scala/cats/effect/concurrent/Console.scala
djspiewak: I would remove all the things that involve `implicit Console[F]`. Basically just distilling it down to a type that is passed explicitly.

djspiewak:
> What is a list of developers? I will gladly add him.

It's in the build and gets baked into the POM upon publication. I think the current one might be materialized implicitly by sbt-spiewak though, which would mean that it's probably very incomplete.

concurrent/shared/src/main/scala/cats/effect/concurrent/Console.scala
mpilquist: I’m not sure about this. I’d likely take Console as an implicit argument in fs2 and in apps.

djspiewak: It's stateful though. Among other things, remember that it's impossible to get an `implicit` value generated by a `for`-comprehension step without using better-monadic-for. Which like… obviously you can do (in Scala 2), but isn't that a decent red flag? What's wrong with passing it explicitly?

mpilquist: Why is it stateful?

Passing it manually is harder and more ceremonious than just taking a Sync constraint. This should be at least as easy as using Sync directly.

djspiewak: > Why is it stateful?
>
> Passing it manually […]

If it's not stateful, then why are you passing it at all? Just take a `Sync` constraint.

The state is the fact that it can be constructed to point at different input/output streams, and those streams might not actually be alive anymore. If there were only one global instance pointing to stdout/stderr/stdin, then the whole thing is actually really easy and doesn't need to be passed at all, implicitly or otherwise. If it's anything *other* than that, then passing it implicitly is dangerous and *constructing* the implicit in a scoped fashion is dodgy.

mpilquist: Ah, yes, let’s get rid of ability to customize streams.

djspiewak: I'm actually okay with that. :-) If we do that then this API becomes very very simple indeed, and we wouldn't need to pass around `Console` instances at all (since they would be effectively singleton anyway).

vasilmkd:
It seems that the header (typelevel licence) that was automatically updated deleted my original reference to `console4cats`. I just now noticed that and I have fixed it.

build.sbt
```diff
@@ -24,6 +24,14 @@ ThisBuild / baseVersion := "3.0"
 ThisBuild / organization := "org.typelevel"
 ThisBuild / organizationName := "Typelevel"
 
+ThisBuild / developers := List(
+  Developer("djspiewak", "Daniel Spiewak", "@djspiewak", url("https://github.com/djspiewak")),
+  
+  // Gabriel Volpe, for influencing the design of `cats.effect.concurrent.Console` through the development of the
+  // original `console4cats` (https://github.com/profunktor/console4cats).
+  Developer("gvolpe", "Gabriel Volpe", "@volpegabriel87", url("https://github.com/gvolpe"))
+)
```
djspiewak: Since we're adding this now, let's go ahead and get others in here as well. `git shortlog -s` tells me the following:

```
     5  Ben Plommer
     5  Chris Davenport
   478  Daniel Spiewak
    67  Fabio Labella
     1  Gavin Bisesi
     1  Guillaume Martres
    18  Jakub Kozłowski
    13  Michael Pilquist
     1  Michał Gibowski
     4  Oleg Pyzhcov
    86  Raas Ahsan
   135  Tim Spence
    62  Vasil Vasilev
    12  anikiforov
     5  mpilquist
```

Which, btw, gives us a pretty good idea who rebases with squashing and who doesn't. :-D

Off-hand, and in addition to what you have here:

- Fabio
- Jakob
- Michael
- Raas
- Tim
- Vasil
- Ben

I think the other changes have been relatively small in scope, but that's a subjective call with no clear answer. Feel free to disagree! I'm not fussed much about this either way.

concurrent/shared/src/main/scala/cats/effect/concurrent/Console.scala
```diff
@@ -0,0 +1,157 @@
...
+
+  /**
+   * Prints a value to the standard error output followed by a new line using the implicit `cast.Show` instance.
+   *
+   * @param a value to be printed to the standard error output
+   */
+  def printlnError[F[_]: Sync, A: Show](a: A): F[Unit] =
+    Sync[F].blocking(scala.Console.err.println(a.show))
+
+  final class ApplyBuilders[F[_]](val F: Sync[F]) extends AnyVal {
```
djspiewak: I would just change this to `Console[F]`. That seems more idiomatic to me, anyway. Basically, the `Console` *class* would exist solely to be a holder for the `F` type and its constraints.

concurrent/shared/src/main/scala/cats/effect/concurrent/Console.scala
```diff
@@ -0,0 +1,157 @@
...
+  def apply[F[_]](implicit F: Sync[F]): ApplyBuilders[F] =
+    new ApplyBuilders[F](F)
+
+  /**
+   * Reads a line as a string from the standard input.
+   *
+   * @return an IO effect that describes reading the user's input from the standard input as a string
+   */
+  def readLine[F[_]: Sync]: F[String] =
+    Sync[F].blocking(scala.Console.in.readLine())
```
djspiewak: This should be `interruptible(false)` rather than `blocking`. I think the rest are good though.

concurrent/shared/src/main/scala/cats/effect/concurrent/Console.scala
```diff
@@ -0,0 +1,157 @@
...
+    def printError[A](a: A)(implicit S: Show[A]): F[Unit] =
+      Console.printError(a)(F, S)
+
+    /**
+     * Prints a value to the standard error output followed by a new line using the implicit `cast.Show` instance.
+     *
+     * @param a value to be printed to the standard error output
+     */
+    def printlnError[A](a: A)(implicit S: Show[A]): F[Unit] =
+      Console.printlnError(a)(F, S)
```
djspiewak: If the `F: Sync[F]` above is made `implicit`, then you don't need to pass any of these explicitly.

concurrent/shared/src/main/scala/cats/effect/concurrent/Console.scala
```diff
@@ -0,0 +1,157 @@
...
+   * Allows access to `Console` functionality for `F` data types that are [[Sync]].
+   *
+   * This builder uses the
+   * [[https://typelevel.org/cats/guidelines.html#partially-applied-type-params Partially-Applied Type]]
+   * technique.
+   *
+   * For printing to the standard output:
+   * {{{
+   *   Console[F].print("Hello") <-> Console.print[F, String]("Hello")
+   *   Console[F].println("Hello") <-> Console.println[F, String]("Hello")
```
djspiewak: IMO, and this is a bit of a bikeshed, but I don't think we need to support both variants of the API. Let's just be a little more prescriptive:

- `Console[F].print("Hello")`
- `Console[F].println("Hello")`
- `Console[F].readLine`
- `Console[F].printErr("Hello")` (I like the shorter `Err`, but that's just my opinion)
- `Console[F].printlnErr("Hello")` (I like the shorter `Err`, but that's just my opinion)

djspiewak (changes requested):
I'm still thinking about the "override the streams" case…

concurrent/shared/src/main/scala/cats/effect/concurrent/Console.scala
```diff
@@ -0,0 +1,157 @@
...
+ */
+
+package cats.effect.concurrent
+
+import cats.Show
+import cats.effect.kernel.Sync
+import cats.syntax.all._
+
+/**
+ * Effect type agnostic `Console` with common methods to write to and read from the standard console.
```
kubukoz: I would mention that it's supposed to be used for simple console output exclusively, and anything more complex like logging should be done via a custom algebra (e.g. Logger from log4cats).

build.sbt
vasilmkd: Since this is now a task that requires compiling a list of people's names, GitHub accounts and Twitter handles, I'm going to do that in a separate issue/PR. Let's focus this current one on `Console`.

concurrent/shared/src/main/scala/cats/effect/concurrent/Console.scala
vasilmkd: Unfortunately that doesn't work with value classes. 😕

djspiewak: I wouldn't bother making it a value class. That probably doesn't buy very much in this instance.

djspiewak:
So on the subject of `setErr` and such… Also taking into account @kubukoz's excellent points…

I think that we should roll our own `readLine`. It's going to be a little annoying because we actually need to deal with `\n` vs `\r\n` and such, but it's not the end of the world. We can't reuse `BufferedReader` because that will result in missing some input depending on buffer quantization.

As for `println` and `print` and such, I think we just need to re-access them on each read:

```scala
def println[A: Show](a: A): F[Unit] = {
  val str = a.show
  F.blocking(System.out.println(str))
}
```

…and with that, we should support the setters.

PR #1199, opened 2020-09-16
Move Resource to kernel

### LINKED ISSUE #1109: Move Resource to kernel
This also enables us to add `Concurrent#background`.

vasilmkd:
Resolves #1109.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/LiftIO.scala
+++ b/core/shared/src/main/scala/cats/effect/LiftIO.scala
@@ -135,4 +135,23 @@ object LiftIO {
    */
   implicit val ioLiftIO: LiftIO[IO] =
     new LiftIO[IO] { override def liftIO[A](ioa: IO[A]): IO[A] = ioa }
+
+  /**
+   * [[LiftIO]] instance for [[Resource]] values.
+   */
+  implicit def catsEffectLiftIOForResource[F[_]](
+      implicit F00: LiftIO[F],
+      F10: Applicative[F]): LiftIO[Resource[F, *]] =
+    new ResourceLiftIO[F] {
+      def F0 = F00
+      def F1 = F10
+    }
+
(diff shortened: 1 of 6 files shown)


RaasAhsan (approved):

core/shared/src/main/scala/cats/effect/LiftIO.scala
```diff
@@ -135,4 +135,23 @@ object LiftIO {
...
+      def F1 = F10
+    }
+
+  abstract private class ResourceLiftIO[F[_]] extends LiftIO[Resource[F, *]] {
+    implicit protected def F0: LiftIO[F]
+    implicit protected def F1: Applicative[F]
+
+    def liftIO[A](ioa: IO[A]): Resource[F, A] =
+      Resource.liftF(F0.liftIO(ioa))
+  }
```
djspiewak, line 156: As a random note, I kind of prefer putting instances on datatype companions whenever *possible*. So in this case, the instance would be on `Resource`'s companion (similar to how `Effect` has an instance on `IO`'s companion). Not a big deal; doesn't block the merge. Also curious as to others' thoughts.

vasilmkd: That was the case. However, `LiftIO` doesn't exist in `kernel` and therefore it doesn't compile.

djspiewak: Oh! Derp. That's obvious in retrospect. Okay cool. :-)

vasilmkd: Rebase then?

djspiewak: Merge please :-) Rebasing public history is usually not ideal.

vasilmkd: Done.

PR #1202, opened 2020-09-17
Remove references to executed fibers

vasilmkd:
Fixes #1201.

cc @mpilquist 
https://github.com/typelevel/fs2/pull/2035

Verified by publishing `cats-effect` locally, updating the dependency in `fs2` and running the tests again.
```
sbt:fs2-reactive-streams> testOnly *Unicast*
[info] StreamUnicastPublisherSpec:
[TestNG] Running:
  Command line suite

[info] - optional_spec104_mustSignalOnErrorWhenFails
[info] - optional_spec105_emptyStreamMustTerminateBySignallingOnComplete
[info] - optional_spec111_maySupportMultiSubscribe
[info] - optional_spec111_multicast_mustProduceTheSameElementsInTheSameSequenceToAllOfItsSubscribersWhenRequestingManyUpfront
[info] - optional_spec111_multicast_mustProduceTheSameElementsInTheSameSequenceToAllOfItsSubscribersWhenRequestingManyUpfrontAndCompleteAsExpected
[info] - optional_spec111_multicast_mustProduceTheSameElementsInTheSameSequenceToAllOfItsSubscribersWhenRequestingOneByOne
[info] - optional_spec111_registeredSubscribersMustReceiveOnNextOrOnCompleteSignals
[info] - optional_spec309_requestNegativeNumberMaySignalIllegalArgumentExceptionWithSpecificMessage
[info] - required_createPublisher1MustProduceAStreamOfExactly1Element
[info] - required_createPublisher3MustProduceAStreamOfExactly3Elements
[info] - required_spec101_subscriptionRequestMustResultInTheCorrectNumberOfProducedElements
[info] - required_spec102_maySignalLessThanRequestedAndTerminateSubscription
[info] - required_spec105_mustSignalOnCompleteWhenFiniteStreamTerminates
[info] - required_spec107_mustNotEmitFurtherSignalsOnceOnCompleteHasBeenSignalled
[info] - required_spec109_mayRejectCallsToSubscribeIfPublisherIsUnableOrUnwillingToServeThemRejectionMustTriggerOnErrorAfterOnSubscribe
[info] - required_spec109_mustIssueOnSubscribeForNonNullSubscriber
[info] - required_spec109_subscribeThrowNPEOnNullSubscriber
[info] - required_spec302_mustAllowSynchronousRequestCallsFromOnNextAndOnSubscribe
[info] - required_spec303_mustNotAllowUnboundedRecursion
[...]

PR #1203, opened 2020-09-17
Prevent accidental resuming of a done fiber

vasilmkd:
Also helps to pad out the `resume` switch.

### DIFF
--- a/core/js/src/main/scala/cats/effect/IOFiberConstants.scala
+++ b/core/js/src/main/scala/cats/effect/IOFiberConstants.scala
@@ -42,4 +42,5 @@ private[effect] object IOFiberConstants {
   val AfterBlockingFailedR: Byte = 4
   val EvalOnR: Byte = 5
   val CedeR: Byte = 6
+  val DoneR: Byte = 7
 }
--- a/core/jvm/src/main/java/cats/effect/IOFiberConstants.java
+++ b/core/jvm/src/main/java/cats/effect/IOFiberConstants.java
@@ -42,4 +42,5 @@ final class IOFiberConstants {
   public static final byte AfterBlockingFailedR = 4;
   public static final byte EvalOnR = 5;
   public static final byte CedeR = 6;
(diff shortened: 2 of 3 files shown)


RaasAhsan:
Quick question: how come this is necessary? The fiber should terminate after returning from `runTerminusK` which leaves it in a state where it can never be acquired again since `suspended` is `false`. So, AFAICT there's no way for the fiber to be resubmitted to the executor again.

vasilmkd:
Right now, there is unpatched code in `WorkStealingThreadPool` (adressed by my other PR) that incorrectly enqueues a fiber produced by a runnable twice. If that fiber has an incorrect `resumeTag`, for example `CedeR` (I'm not debating whether this is a realistic example), running the fiber again though `run()` would execute potentially dangerous code. This nicely avoids all that. Feel free to close it if  you think otherwise.

Since `resumeTag` is mutable and we don't clear it after use, something like this isn't impossible.

vasilmkd:
@RaasAhsan what you described used to be true. Ever since we moved to `run()` as the only entry to the fiber, it doesn't always check the state of `suspended`, depending on the `resumeTag` value. Does that make sense?

RaasAhsan:
Got it, I think this is a reasonable precaution as a cleanup step then. If this is an indicator of a defect or leak in the runtime and/or scheduler, it may be worth throwing an exception or side channel reporting an error.

vasilmkd:
I don't think it is. I just see it as a regression due to the many refactorings. Also, we take care to cleanup all of the mutable state, this is I believe the only one we don't, so it's slightly more consistent. Another very minor thing, the openjdk JVM has a threshold of 8 cases, which cause it to produce the best tableswitch bytecode, so this in a way helps with that too.

vasilmkd:
And also the obvious bug (introduced originally by me) is being addressed with #1202.

djspiewak:
I'm okay with being extra-explicit here. It shouldn't result in a performance impact.

PR #1207, opened 2020-09-18
Mark MVarConcurrent tests as flaky

vasilmkd:


core/shared/src/test/scala/cats/effect/concurrent/MVarTests.scala
```diff
@@ -32,104 +31,105 @@ class MVarConcurrentTests extends BaseMVarTests {
...
-      _ <- IO.sleep(10.millis) // Give put callbacks a chance to register
-      _ <- p2.cancel
-      _ <- mVar.take
-      r1 <- mVar.take
-      r3 <- mVar.take
-    } yield Set(r1, r3)
+      v <- init(42)
+      gate <- Deferred.uncancelable[IO, Unit]
+      canceled <- Deferred[IO, Boolean]
+      fa <- (gate.complete(()) *> v.put(0)).guaranteeCase {
```
RaasAhsan: There is still some non-determinism in several interleavings, here's one of them: `fa` is cancelled after `gate.complete(())` but before `v.put(0)`, in which case we're not really exercising cancellation on `put`. I'm not really sure if this is better or worse than the sleeps.

core/shared/src/test/scala/cats/effect/concurrent/MVarTests.scala
```diff
@@ -32,104 +31,105 @@ class MVarConcurrentTests extends BaseMVarTests {
...
-      _ <- mVar.put(1)
-      _ <- mVar.put(3)
-      r1 <- t1.join
-      r3 <- t3.join
-    } yield Set(r1, r3)
+      gate <- Deferred.uncancelable[IO, Unit]
+      canceled <- Deferred[IO, Boolean]
+      v <- empty[Int]
+      fa <- v.take.start
+      fb <- (gate.complete(()) *> v.take).guaranteeCase {
```
RaasAhsan: Similar problem here as above, and I imagine for the rest of the tests as well. It's very difficult to test cancellation for `MVar` operations in the presence of autocancellation. The spawned fiber is telling the main fiber "I'm ready to be canceled now!" but it actually isn't because it hasn't started the operation of interest, and it needs to run a bind before getting there.

Perhaps in CE3 `uncancelable` and `poll` makes this completely deterministic

vasilmkd:
@RaasAhsan @djspiewak I'm aware of the possibility to not exercise the proper cancelation. However, these tests bit us on the dotty community build, there, they seem to _almost always_ fail for some reason. It seems running all tests for many projects creates some huge contention or something. We need a solution for that.

There is currently a `dotty-staging` fork of this project where these tests are outright disabled. That bothers me more than the possibility to not always exercise a logic which is non-deterministic by nature.

If you are still against this, let's compromise and return the `IO.sleep(500.millis)` but the gate needs to be in there as well, otherwise we risk failing a test, which I really do not like.

vasilmkd:
I've used https://scalameta.org/munit/docs/tests.html#tag-flaky-tests to mark the tests that were sporadically failing in the dotty community build. Basically, any test marked as `flaky` won't fail the whole build if it fails. This also applies for tests which timeout, something that was observed in the probably more resource constrained dotty community build execution environment.

If this is accepted, I'd like it to land into `dotty-staging`.

djspiewak (approved):
Even with CE3 cancelation semantics, I don't think there's a way to make these tests fully deterministic. I'll leave this open for a bit to hear @RaasAhsan's thoughts on the marker, but I'm in favor of merging.

RaasAhsan (approved):
👍

PR #1211, opened 2020-09-19
Remove unsafe cast from Resource#invariant

bplommer:
I hoped adding this type variable might remove the need for `InvariantResource` altogether, but it seems Scala 2.x can't directly unify `F0` with the type parameters of the `Resource` subtypes 😞


djspiewak:
Impressive!

PR #1212, opened 2020-09-20
Getting started on docs

djspiewak:
…and also fixed `MonadCancel`'s syntax, which didn't exist.


docs/typeclasses.md
```diff
@@ -0,0 +1,252 @@
...
+- State sharing between parallel processes
+- Interactions with time, including current time and `sleep`
+- Safe capture of side-effects which return values
+- Safe capture of side-effects which invoke a callback
+
+These are listed roughly in order of increasing power. Parallelism cannot be safely defined without some notion of resource safety and preemption. Interactions with time have no meaning if there is no notion of parallel evaluation, since it would be impossible to observe the passage of time. Capturing and controlling side-effects makes it possible to do *anything* simply by "cheating" and capturing something like `new Thread`, and thus it has the greatest power of all of these categories.
+
+Beyond the above, certain capabilities are *assumed* by Cats Effect but defined elsewhere (or rather, defined within [Cats](https://typelevel.org/cats)). Non-exhaustively, those capabilities are as follows:
+
+- Transforming the value inside an effect by `map`ping over it
```
RaasAhsan, line 16: I think it would be helpful to point out which classes these refer to

docs/typeclasses.md
```diff
@@ -0,0 +1,252 @@
+# Typeclasses
+
+The typeclass hierarchy within Cats Effect defines what it means, fundamentally, to *be* a functional effect. This is very similar to how we have a set of rules which tell us what it means to be a number, or what it means to have a `flatMap` method. Broadly, the rules for functional effects can be broken down into the following categories:
+
+- Resource safety and preemption
```
RaasAhsan: Is preemption taken to be a synonym for cancellation or interruption? I usually think of preemption to mean forcibly pausing a thread of execution such that it can be resumed later (this is consistent with OS literature), whereas cancellation/interruption is a cooperative mechanism for requesting termination without resumption.

I'd prefer if we settle on one of those three for consistency for the purposes of documentation

docs/typeclasses.md
```diff
@@ -0,0 +1,252 @@
...
+MonadCancel[F] uncancelable { outer =>
+  MonadCancel[F] uncancelable { inner =>
+    inner(fa)
+  }
+}
+```
+
+The `inner` poll eliminates the inner `uncancelable`, but the *outer* `uncancelable` still applies, meaning that this whole thing is equivalent to `MonadCancel[F].uncancelable(_ => fa)`.
+
+Note that polling for an *outer* block within an inner one has no effect whatsoever. For example:
```
RaasAhsan, line 95: `outer(inner(fa))` would undo both masks right?

docs/typeclasses.md
```diff
@@ -0,0 +1,252 @@
...
+- Transforming the value inside an effect by `map`ping over it
+- Putting a value *into* an effect
+- Composing multiple effectful computations together sequentially, such that each is dependent on the previous
+- Raising and handling errors
+
+Taken together, all of these capabilities define what it means to be an effect. Just as you can rely on the properties of integers when you perform basic mental arithmetic (e.g. you can assume that `1 + 2 + 3` is the same as `1 + 5`), so too can you rely on these powerful and general properties of *effects* to hold when you write complex programs. This allows you to understand and refactor your code based on rules and abstractions, rather than having to think about every possibly implementation detail and use-case. Additionally, it makes it possible for you and others to write very generic code which composes together making an absolute minimum of assumptions. This is the foundation of the Cats Effect ecosystem.
+
+## `MonadCancel`
+
+This typeclass extends `MonadError` (in Cats) and defines the meaning of resource safety and preemption (cancellation). Using it, we can define effects which safely acquire and release resources:
```
RaasAhsan: I think it would be good to have a sentence or paragraph that describes precisely what behavior cancellation/preemption has. things like how it's a cooperative mechanism etc. The rest of the section kind of assumes that the reader is familiar with it already

docs/typeclasses.md
```diff
@@ -0,0 +1,252 @@
...
+
+Self-cancellation is somewhat similar to raising an error with `raiseError` in that it will short-circuit evaluation and begin "popping" back up the stack until it hits a handler. Just as `raiseError` can be observed using the `onError` method, `canceled` can be observed using `onCancel`.
+
+The primary differences between self-cancellation and `raiseError` are two-fold. First, `uncancelable` suppresses `canceled` within its body (unless `poll`ed!), turning it into something equivalent to just `().pure[F]`. There is no analogue for this kind of functionality with errors. Second, if you sequence an error with `raiseError`, it's always possible to use `attempt` or `handleError` to *handle* the error and resume normal execution. No such functionality is available for cancellation.
+
+In other words, cancellation is *final*. It can be observed, but it cannot be *prevented*. This feature is exceptionally important for ensuring deterministic evaluation and avoiding deadlocks.
```
RaasAhsan: I like "effective" as an alternate to "final" :) "Prevented" is also kind of ambiguous. How about something like:

In other words, cancellation is *effective*; it cannot be undone. It can be suppressed, but once it is observed, it must be respected if there are no masks.

docs/typeclasses.md
djspiewak: Would it? I kind of wanted to leave that a bit vague here so that people can get the general shape of the hierarchy before worrying about memorizing names. Maybe that's a bad idea though.

docs/typeclasses.md
djspiewak: Yes. Perhaps that should be pointed out explicitly.

docs/typeclasses.md
djspiewak: Is it a cooperative mechanism though? It certainly is from the perspective of the *thread* layer (i.e. the implementation), but from the perspective of the fiber layer it's entirely preemptive.

RaasAhsan: Preemption sounds to me like "I am going to force you to stop execution immediately" rather than "Please stop execution whenever you see this message." I don't think this is the case at the fiber layer because a fiber won't necessarily observe a cancellation before the next logical instruction, but only at the next cancellation boundary, which could be up to 512 instructions later. If we checked cancellation with a barrier after *every* instruction I would be more inclined to agree

docs/typeclasses.md
RaasAhsan: Hmm then maybe it's not worth including this list at all, and just point directly to the Cats hierarchy?

PR #1213, opened 2020-09-20
Make UnsafeRun[IO] available anywhere there is an IORuntime

mpilquist:


vasilmkd (approved):

djspiewak:
Should we also remove the `implicit` `UnsafeRun` which is provided by `IOApp`, since I believe this new `implicit` will handle that materialization for us as well?

PR #1215, opened 2020-09-20
CE2 Resource: factor out duplication between versions, backport type safety improvements from CE3

bplommer:


vasilmkd:
Love this change. I just have a slight issue with the name of `ResourceCompat`. Personally, I'd go for a `BaseResource` or `ResourceLike` as it is more descriptive (again, to me). But feel free to ignore. This is more of a general question, is there a precedent for this kind of base class that works across versions?

bplommer:
> Personally, I'd go for a BaseResource or ResourceLike as it is more descriptive (again, to me).

I have no view on this so am happy to change it unless others object 🙂

> This is more of a general question, is there a precedent for this kind of base class that works across versions?

Not sure - I only remember seeing it done with different implementations for the same API, rather than different APIs for the same implementation.

bplommer:
Converting to draft because mima isn't happy with `InvariantResource#invariant` for some reason

bplommer:
I had to add a dummy implementation of `invariant` to `Resource`, as mima didn't like me adding an abstract method. Alternatively I could just move the real implementation from `InvariantResource` to `Resource`, but that would require `asInstanceOf` - thoughts?

djspiewak:
`Resource` is sealed (and all subtypes are non-abstract), so it's safe to add a mima exception for that particular warning.

Issue #1214, opened 2020-09-20
Add documentation page on testing with cats-effect


Daenyth:
To start with, it can index some of the third-party addons that offer compatibility with various test frameworks.

It can expand to include "best practices" and answer common questions (like how to use `Resource` with `beforeEach` style imperative hooks)

It can also have a section on `TestContext` usage

PR #1216, opened 2020-09-20
Add documentation page on testing with cats-effect

Daenyth:
Addressing #1214

### DIFF
--- a/site/src/main/mdoc/testing/index.md
+++ b/site/src/main/mdoc/testing/index.md
@@ -1,0 +1,86 @@
+---
+layout: docsplus
+title:  "Testing with cats-effect"
+position: 2
+---
+
+<nav role="navigation" id="toc"></nav>
+
+## Compatible libraries
+
+Relatively few libraries support cats-effect directly at this time. However, most (if not all) popular testing frameworks have libraries offering some level of integration
+
+- [cats-effect-testing](https://github.com/djspiewak/cats-effect-testing): Supports Scalatest, Specs2, munit, minitest, µTest, and scalacheck
+- [@Daenyth's IOSpec gist](https://gist.github.com/Daenyth/67575575b5c1acc1d6ea100aae05b3a9) for scalatest (library pending)
+- [munit-cats-effect](https://github.com/typelevel/munit-cats-effect)
+- [weaver-test](https://disneystreaming.github.io/weaver-test/): Supported natively
+
+## Property-based Testing
+
+### Scalacheck
+
+Scalacheck primarily supports properties in the shape `A => Assertion`.
+To support writing effectful properties with the shape `A => F[Assertion]`, you can use one of these tools:
+
+- [scalacheck-effect](https://github.com/typelevel/scalacheck-effect)
+- [cats-effect-testing](https://github.com/djspiewak/cats-effect-testing)
+
+You might also want to use [cats-scalacheck](https://christopherdavenport.github.io/cats-scalacheck/), which provides instances of `cats-core` typeclasses.
+
+## Best practices
+
+Avoid using `unsafe*` methods in tests, the same as you'd avoid them in "main" code.
+Writing tests to be structured the same way as "normal" code results in tests that are less likely to be flaky, act as executable documentation, and remain easy to read.
+
+Use a compatible framework's support for writing `IO[Assertion]` style tests.
+
+### Testing concurrency
+
+When possible, use `TestContext` as your `ExecutionContext`, `ContextShift`, and `Timer`. This is generally the case if your code uses `Timer` and `Clock` to perform time-related operations, and doesn't directly read or interact with the jvm System clock.
+
+To simulate time passing in the test, use `Timer[F].sleep(duration)`, which will defer to `TestContext`.
+
+Some reference material on this approach:
+
+- [TestContext api documentation](https://www.javadoc.io/doc/org.typelevel/cats-effect-laws_2.13/2.2.0/cats/effect/laws/util/TestContext.html) (includes examples and motivation)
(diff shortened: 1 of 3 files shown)


site/src/main/mdoc/testing/index.md
```diff
@@ -0,0 +1,86 @@
...
+layout: docsplus
+title:  "Testing with cats-effect"
+position: 2
+---
+
+<nav role="navigation" id="toc"></nav>
+
+## Compatible libraries
+
+Relatively few libraries support cats-effect directly at this time. However, most (if not all) popular testing frameworks have libraries offering some level of integration
```
RaasAhsan, line 11: The new Cats Effect documentation site is going to include some of this as well (https://github.com/typelevel/cats-effect/blob/docs/website/pages/en/ecosystem.js#L48), but it's probably worth including here as well

site/src/main/mdoc/testing/index.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Property-based Testing
+
+### Scalacheck
+
+Scalacheck primarily supports properties in the shape `A => Assertion`.
+To support writing effectful properties with the shape `A => F[Assertion]`, you can use one of these tools:
+
+- [scalacheck-effect](https://github.com/typelevel/scalacheck-effect)
+- [cats-effect-testing](https://github.com/djspiewak/cats-effect-testing)
```
djspiewak: I think cats-effect-testing actually doesn't support `A => F[Assertion]` properties, does it? I may be derping.

site/src/main/mdoc/testing/index.md
```diff
@@ -0,0 +1,86 @@
...
+## Best practices
+
+Avoid using `unsafe*` methods in tests, the same as you'd avoid them in "main" code.
+Writing tests to be structured the same way as "normal" code results in tests that are less likely to be flaky, act as executable documentation, and remain easy to read.
+
+Use a compatible framework's support for writing `IO[Assertion]` style tests.
+
+### Testing concurrency
+
+When possible, use `TestContext` as your `ExecutionContext`, `ContextShift`, and `Timer`. This is generally the case if your code uses `Timer` and `Clock` to perform time-related operations, and doesn't directly read or interact with the jvm System clock.
```
djspiewak: I'm actually not sure I agree with this. `TestContext` can mask a lot of subtle bugs. It's appropriate for some things, not for others.

djspiewak (changes requested):
This is *amazing*. Just a couple quick notes

site/src/main/mdoc/testing/index.md
Daenyth: The docs suggest that it does; https://github.com/djspiewak/cats-effect-testing#scalatest-scalacheck

```
  "My IO Code" - {

    "works with effect-full property-based testing" in {
      forAll { (l1: List[Int], l2: List[Int]) =>
        IO.delay(l1.size + l2.size shouldBe (l1 ::: l2).size)
      }
    }
```

I think from skimming the code it doesn't support *true async* properties (so, probably won't work on scala.js)

site/src/main/mdoc/testing/index.md
Daenyth: Hmm, ok. I can revise the wording here. I do want to point people at it - I think it's under-advertised

Do you have a suggestion I could use for new wording?

site/src/main/mdoc/testing/index.md
djspiewak: Ooooooooh I get why this is working. Okay this is a fair bullet point then. But yeah, it's not going to work on scala.js. :-)

site/src/main/mdoc/testing/index.md
djspiewak: Maybe something like this?

```markdown
In many test cases, `TestContext` should be used as an `ExecutionContext`, `ContextShift`, and `Timer`. This gives you very precise control over the sequencing and scheduling of evaluation, making it possible to deterministically replicate certain race conditions or even identify timing-related bugs without having to rely on non-deterministic thread ordering. `TestContext` also gives you control over *time* (as exposed by `Timer` and `Clock`), which makes it easy to write unit tests for time-related semantics without having to rely on `sleep`s or other hacks.

Be warned though that `TestContext` is a very artificial environment, and it can in turn mask bugs that a realistic executor would uncover. The most comprehensive test suites will often use a balance of each.
```

site/src/main/mdoc/testing/index.md
```diff
@@ -0,0 +1,86 @@
...
+Some reference material on this approach:
+
+- [TestContext api documentation](https://www.javadoc.io/doc/org.typelevel/cats-effect-laws_2.13/2.2.0/cats/effect/laws/util/TestContext.html) (includes examples and motivation)
+- [Time Traveling in Tests with Cats-Effect](https://blog.softwaremill.com/time-traveling-in-tests-with-cats-effect-b22084f6a89), by Krzysztof Ciesielski
+
+### Managing shared resources
+
+Sometimes you'll want to write a test that acquires a Resource before the suite and releases it after. For example, spinning up a database.
+
+For test frameworks that use imperative "hook"-style methods (such as scalatest's `BeforeAndAfterAll` mixin), you can use [`allocated`](https://typelevel.org/cats-effect/api/cats/effect/Resource.html#allocated[G[x]%3E:F[x],B%3E:A](implicitF:cats.effect.Bracket[G,Throwable]):G[(B,G[Unit])])
```
keynmol: Worth adding here weaver's resources as an alternative? As I think all the other frameworks use hook style

keynmol (approved):

PR #1219, opened 2020-09-20
Let munit help out with running the tests

vasilmkd:


core/shared/src/test/scala/cats/effect/CatsEffectSuite.scala
```diff
@@ -31,6 +31,9 @@ import org.scalacheck.Prop
 trait CatsEffectSuite extends DisciplineSuite with TestUtils {
   implicit protected val munitLocation: Location = implicitly
 
+  override def munitValueTransforms: List[ValueTransform] =
+    super.munitValueTransforms :+ new ValueTransform("IO", { case io: IO[_] => io.unsafeToFuture() })
```
djspiewak, line 35: TIL! Okay I'm very impressed with this.

djspiewak (approved):

core/shared/src/test/scala/cats/effect/CatsEffectSuite.scala
vasilmkd: This is available through `munit-cats-effect` too.

PR #1221, opened 2020-09-21
Refactor Semaphore for readability

bplommer:


concurrent/shared/src/main/scala/cats/effect/concurrent/Semaphore.scala
```diff
@@ -129,9 +129,13 @@ object Semaphore {
...
     else F.unit
 
+  private final case class Request[F[_]](n: Long, gate: Deferred[F, Unit])
+
   // A semaphore is either empty, and there are number of outstanding acquires (Left)
   // or it is non-empty, and there are n permits available (Right)
-  private type State[F[_]] = Either[Queue[(Long, Deferred[F, Unit])], Long]
+  private type State[F[_]] = Either[Queue[Request[F]], Long]
+
+  private final case class Promise[F[_]](await: F[Unit], release: F[Unit])
```
RaasAhsan: This doesn't really feel like a `Promise` to me (there were a ton of discussions about naming something `promise` which are still fresh in my mind :D ), maybe something more like `Acquire` or `Permit`?

RaasAhsan (approved):
One comment about naming, but this looks awesome! Named data types always makes these state machines way easier to follow

concurrent/shared/src/main/scala/cats/effect/concurrent/Semaphore.scala
bplommer: `Permit` doesn't seem right - when you get this you don't yet have a permit, just a way to manage one. Permits themselves don't have a first-class representation. `Acquire` feels vague. Maybe `PermitToken` or `Registration`?

SystemFw: `Permit` seems ok to me, since it will read like `permit.await`

bplommer: I'll change it to `Permit` since that seems to be the consensus 🙂

PR #1226, opened 2020-09-22
Added derived operations for UnsafeRun

mpilquist:
This may be controversial, as perhaps we're okay with `UnsafeRun` being a clunky experience. I figured I'd open this anyway and at least get the conversation on record.

Here's an example of using `UnsafeRun` in fs2: https://github.com/typelevel/fs2/pull/2045/files#diff-7e9f65865c6e0b05b7b601c674b1fc5dR86-R95


djspiewak:
I'm definitely 👎 on this. I really don't want `UnsafeRun` to be a convenient thing to use in any sense of the word. I *almost* didn't want to put the `apply` materializer on the companion! It should be a thing you reach for as a library author when you're desperate and absolutely nothing else will do.

With that said… it is more than a little annoying that everyone is going to have to do their own `Await.result(..., Duration.Inf)`. But then again… it does make it a lot more explicit what is going on and hints at why it's dangerous.

ChristopherDavenport:
I agree with Daniel that we want to make this difficult. The concern I have is fs2/blaze will all have to make their own version of this in every downstream. Regardless, we should document because library authors who don't interact with the impure side of the ecosystem may not know what they need to do to interact with the Future api.

mpilquist:
> The concern I have is fs2/blaze will all have to make their own version of this in every downstream.

Exactly this. :) In fs2, we have two places that need it -- one in fs2-io and one in fs2-reactive-streams, and no good place to put the syntax. Seems like this is one of those cases where our distaste for unsafe stuff is going to result in lots of copy/paste downstream. I'm curious if there's some balance here -- e.g., syntax but not part of `cats.effect.syntax._` import? no syntax but still derived operations?

djspiewak:
Specifically the piece missing is the *blocking* one, right?

mpilquist:
The fire and forget one would be nice too. :)

I don't care about the timed one.

djspiewak:
So the blocking one would definitely need to be in the jvm-only sources. Fire-and-forget doesn't seem to have *that* much value, since the existing `unsafeRunFutureCancelable` side-effects, so you can just… call it and then return `()`. :-)

I do think everything should have a timeout though. So I'd be okay with `UnsafeRun.unsafeRunTimedAndCancel` (or something similarly long) being included in the JVM companion. I think calling that with `Duration.Inf` isn't too onerous.

This is all definitely a balance. :-( I get the desire to avoid pasta everywhere, and `UnsafeRun` itself is a reasonable landing-place for it. I just don't want to make it *too* nice, and I feel like syntax enrichments definitely cross that line.

mpilquist:
OK, I removed the syntax and moved unsafeRunSync/unsafeRunTimed to the companion. I added a summoner though, as I don't want to force folks to pass the instance explicitly.

kernel/jvm/src/main/scala/cats/effect/unsafe/UnsafeRunPlatform.scala
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+package cats.effect.unsafe
+
+import scala.concurrent.{Await, Future, TimeoutException}
+import scala.concurrent.duration.Duration
+
+private[unsafe] trait UnsafeRunPlatform[F[_]] {
+  def unsafeRunFutureCancelable[A](fa: F[A]): (Future[A], () => Future[Unit])
```
djspiewak: You can remove this redundancy by adding a self-type: `this: UnsafeRun[F] =>`. I like doing that anyway with these kinds of platform classes so they don't end up getting reused.

Also `private` = `private[unsafe]`, surprisingly.

djspiewak (changes requested):
Minor bit

kernel/jvm/src/main/scala/cats/effect/unsafe/UnsafeRunPlatform.scala
mpilquist: Private results in:

```scala
cats-effect/kernel/shared/src/main/scala/cats/effect/unsafe/UnsafeRun.scala:21:31: private trait UnsafeRunPlatform escapes its defining scope as part of type cats.effect.unsafe.UnsafeRunPlatform[F]
[error] trait UnsafeRun[F[_]] extends UnsafeRunPlatform[F] {
[error]                               ^
[error] one error found
```

mpilquist:
Addressed comments

kernel/jvm/src/main/scala/cats/effect/unsafe/UnsafeRunPlatform.scala
djspiewak: So. Weird.

kernel/shared/src/main/scala/cats/effect/unsafe/UnsafeRun.scala
```diff
@@ -18,6 +18,15 @@ package cats.effect.unsafe
 
 import scala.concurrent.Future
 
-trait UnsafeRun[F[_]] {
+trait UnsafeRun[F[_]] extends UnsafeRunPlatform[F] {
   def unsafeRunFutureCancelable[A](fa: F[A]): (Future[A], () => Future[Unit])
+
+  def unsafeRunAsync[A](fa: F[A]): Unit = {
```
djspiewak: Shouldn't this be `unsafeRunAndForget`? `unsafeRunAsync` would imply taking a callback, no?

djspiewak (changes requested):
I missed a thing, sorry

kernel/shared/src/main/scala/cats/effect/unsafe/UnsafeRun.scala
mpilquist: Done :)

PR #1228, opened 2020-09-22
Factor out common code and properly organize laws tests

### LINKED ISSUE #1139: Reduce some of the copy/pasta in laws
The laws tests have a lot of copy/paste. That should be factored out a bit so we don't keep repeating the same things over and over.

vasilmkd:
- Also enable Id Outcome laws

Resolves #1139.

I didn't touch the duplicated implicit values which are put in place for Scala 2.12.


laws/shared/src/test/scala/cats/effect/laws/GenTemporalSpec.scala
```diff
@@ -16,7 +16,7 @@
 
 package cats
 package effect
-package concurrent
+package laws
```
djspiewak, line 19: Whoops!

laws/shared/src/test/scala/cats/effect/laws/IorTPureConcSpec.scala
```diff
@@ -15,37 +15,25 @@
  */
 
 package cats.effect
+package laws
```
djspiewak, line 18: Wow apparently there was a lot of this…

laws/shared/src/test/scala/cats/effect/laws/OutcomeSpec.scala
```diff
@@ -15,23 +15,40 @@
  */
 
 package cats.effect
+package laws
```
djspiewak: Minor nit, but I don't believe this test belongs in the `laws` package. `kernel` would be more appropriate

laws/shared/src/test/scala/cats/effect/BaseSpec.scala
```diff
@@ -16,13 +16,14 @@
...
+import cats.laws.discipline.{eq, ExhaustiveCheck, MiniInt}, eq._
+import cats.syntax.all._
 
 import org.scalacheck.{Arbitrary, Gen, Prop}
-
-import cats.laws.discipline.{eq, ExhaustiveCheck, MiniInt}; import eq._
+import org.scalacheck.util.Pretty
 
 // A dumping ground for random helpers for property tests
 trait BaseSpec {
```
djspiewak: Since this is a dumping ground, I would make it `private`

djspiewak (changes requested):
This is really nice! Just some minor remarks

PR #1234, opened 2020-09-23
Implement embed directly in Outcome

### LINKED ISSUE #1232: Copy embed machinery to Outcome
`Concurrent#background` is a little annoying right now because it returns `F[Outcome[F, E, A]]` in its body. This is consistent with `join`, so that makes sense, but `join` also has `joinAndEmbed` and related, and there's no analogue of this for `Outcome`. It might make sense to add `embed` as a method on `Outcome` to help ease this a bit. The `joinAndEmbed` machinery could then delegate to this method + `flatten`.

Open for discussion. :-) Could also have a different variant of `background`.

vasilmkd:
Resolves #1232.

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Fiber.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Fiber.scala
@@ -22,8 +22,8 @@ trait Fiber[F[_], E, A] {
   def cancel: F[Unit]
   def join: F[Outcome[F, E, A]]
 
-  def joinAndEmbed(onCancel: F[A])(implicit F: MonadCancel[F, E]): F[A] =
-    join.flatMap(_.fold(onCancel, F.raiseError(_), fa => fa))
+  def joinAndEmbed(onCancel: => F[A])(implicit F: MonadCancel[F, E]): F[A] =
+    join.flatMap(_.embed(onCancel))
 
   def joinAndEmbedNever(implicit F: GenSpawn[F, E]): F[A] =
     joinAndEmbed(F.never)
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Fiber.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Fiber.scala
@@ -22,7 +22,7 @@ trait Fiber[F[_], E, A] {
   def cancel: F[Unit]
   def join: F[Outcome[F, E, A]]
 
-  def joinAndEmbed(onCancel: => F[A])(implicit F: MonadCancel[F, E]): F[A] =
+  def joinAndEmbed(onCancel: F[A])(implicit F: MonadCancel[F, E]): F[A] =
     join.flatMap(_.embed(onCancel))
 
   def joinAndEmbedNever(implicit F: GenSpawn[F, E]): F[A] =
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
@@ -108,12 +108,6 @@ trait GenSpawn[F[_], E] extends MonadCancel[F, E] {
 
   def background[A](fa: F[A]): Resource[F, F[Outcome[F, E, A]]] =
     Resource.make(start(fa))(_.cancel)(this).map(_.join)(this)
-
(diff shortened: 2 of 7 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/Fiber.scala
```diff
@@ -22,8 +22,8 @@ trait Fiber[F[_], E, A] {
   def cancel: F[Unit]
   def join: F[Outcome[F, E, A]]
 
-  def joinAndEmbed(onCancel: F[A])(implicit F: MonadCancel[F, E]): F[A] =
-    join.flatMap(_.fold(onCancel, F.raiseError(_), fa => fa))
+  def joinAndEmbed(onCancel: => F[A])(implicit F: MonadCancel[F, E]): F[A] =
```
djspiewak: Why by-name?

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -108,6 +108,12 @@ trait GenSpawn[F[_], E] extends MonadCancel[F, E] {
 
   def background[A](fa: F[A]): Resource[F, F[Outcome[F, E, A]]] =
     Resource.make(start(fa))(_.cancel)(this).map(_.join)(this)
+
+  def backgroundAndEmbed[A](fa: F[A])(onCancel: => F[A]): Resource[F, F[A]] =
+    background(fa).map(flatMap(_)(_.embed(onCancel)(this)))(this)
+
+  def backgroundAndEmbedNever[A](fa: F[A]): Resource[F, F[A]] =
+    backgroundAndEmbed(fa)(never)
```
djspiewak: Discussion: is it worth having this, or should we just rely on people mapping over the contents? Like, `background(fa).evalMap(_.embedNever)` doesn't seem too onerous.

kernel/shared/src/main/scala/cats/effect/kernel/Outcome.scala
```diff
@@ -26,6 +26,12 @@ import scala.util.{Either, Left, Right}
 sealed trait Outcome[F[_], E, A] extends Product with Serializable {
   import Outcome._
 
+  def embed(onCancel: => F[A])(implicit F: MonadCancel[F, E]): F[A] =
```
djspiewak: Why by-name?

djspiewak (changes requested):

kernel/shared/src/main/scala/cats/effect/kernel/Fiber.scala
vasilmkd: `Outcome#fold` takes a by-name parameter. I would think the `onCancel` does not need to be fully built if that path is not taken? Is that a wrong assumption?

djspiewak: That's definitely a correct assumption. I think my thought on the `fold` signature is the fact that it's more general, and critically, people might recurse through it (think: just like `whenM` and `>>`). I can't really see people making `joinAndEmbed` the branch point in a conditional recursion, but maybe I'm wrong on that one.

vasilmkd: I'll revert to strict.

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
vasilmkd: I'm not a fan of the names, but I do like the simmetry with `join`.

kernel/shared/src/main/scala/cats/effect/kernel/Outcome.scala
vasilmkd: Reverted.

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
vasilmkd: I removed these methods.

kernel/shared/src/main/scala/cats/effect/kernel/Fiber.scala
vasilmkd: Reverted.

djspiewak:
🎉

PR #1239, opened 2020-09-25
MonadCancel scaladoc

RaasAhsan:
First PR for scaladoc-ing the hierarchy. I didn't bother including examples for now

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
@@ -31,43 +31,43 @@ import cats.data.{
 import cats.syntax.all._
 
 /**
- * A type class that characterizes monads that support cancellation, masking 
- * and finalization. [[MonadCancel]] instances must provide an instance for 
+ * A type class that characterizes monads that support cancellation, masking
+ * and finalization. [[MonadCancel]] instances must provide an instance for
  * [[MonadError]].
- * 
+ *
  * Cancellation refers to the act of requesting that the execution of a fiber be
  * terminated. [[MonadCancel]] exposes a means of self-cancellation, where a fiber
- * can request that its own execution be terminated. Self-cancellation is achieved 
- * via [[MonadCancel!,canceled canceled]].
- *  
- * Cancellation is vaguely similar to the short-circuit behavior introduced by [[MonadError]], 
+ * can request that its own execution be terminated. Self-cancellation is achieved
+ * via [[MonadCancel!.canceled canceled]].
+ *
+ * Cancellation is vaguely similar to the short-circuit behavior introduced by [[MonadError]],
  * but there are several key differences:
- * 
+ *
  *   1. Cancellation is effective; if it is observed it must be respected, and it cannot
  *      be reversed.
- *   1. Cancellation can be masked via [[MonadCancel!.uncancelable]]. This allows a fiber to 
- *      suppress cancellation for a period of time. If a fiber is cancelled while it is masked, 
+ *   1. Cancellation can be masked via [[MonadCancel!.uncancelable]]. This allows a fiber to
+ *      suppress cancellation for a period of time. If a fiber is cancelled while it is masked,
  *      it must respect cancellation whenever it reaches an unmasked state.
- *   1. [[GenSpawn]] introduces external cancellation, another cancellation mechanism by which fibers 
+ *   1. [[GenSpawn]] introduces external cancellation, another cancellation mechanism by which fibers
  *      can be cancelled by external parties. This can be likened to asynchronous exception throwing.
- * 
+ *
  * Finalization refers to the act of invoking finalizers, which are effects that are guaranteed to be
- * evaluated in the event of cancellation. Finalization is achieved via [[MonadCancel!.onCancel onCancel]], 
- * which registers a finalizer for the duration of some arbitrary effect. If a fiber is cancelled during 
+ * evaluated in the event of cancellation. Finalization is achieved via [[MonadCancel!.onCancel onCancel]],
+ * which registers a finalizer for the duration of some arbitrary effect. If a fiber is cancelled during
  * evaluation of that effect, the registered finalizer is guaranteed to be invoked before terminating.
- * 
+ *
  * The aforementioned concepts come together to unlock a powerful pattern for interacting with
- * effectful lifecycles: the bracket pattern. A lifecycle refers to a pair of actions, which are 
- * called the acquisition action and the release action respectively. The relationship between 
- * these two actions is that if the former completes successfully, then the latter is guaranteed 
+ * effectful lifecycles: the bracket pattern. A lifecycle refers to a pair of actions, which are
+ * called the acquisition action and the release action respectively. The relationship between
+ * these two actions is that if the former completes successfully, then the latter is guaranteed
  * to be evaluated eventually, even in the presence of exceptions and cancellation. While the
  * lifecycle is active, other work can be performed, but this invariant will always be respected.
- * 
- * The bracket pattern is a useful tool when dealing with resource lifecycles. Imagine an application that 
- * opens network connections to a database server to do work. If a task in the application is cancelled 
- * while it holds an open database connection, the connection would never be released or returned to a pool, 
+ *
+ * The bracket pattern is a useful tool when dealing with resource lifecycles. Imagine an application that
+ * opens network connections to a database server to do work. If a task in the application is cancelled
+ * while it holds an open database connection, the connection would never be released or returned to a pool,
  * causing a resource leak. The bracket pattern is analogous to the try-with-resources/finally pattern seen
- * in Java, See [[MonadCancel!.bracket bracket]], [[MonadCancel!.bracketCase bracketCase]] and 
+ * in Java, See [[MonadCancel!.bracket bracket]], [[MonadCancel!.bracketCase bracketCase]] and
  * [[MonadCancel!.bracketFull bracketFull]] for more details.
  */
 trait MonadCancel[F[_], E] extends MonadError[F, E] {
@@ -76,7 +76,7 @@ trait MonadCancel[F[_], E] extends MonadError[F, E] {
   def forceR[A, B](fa: F[A])(fb: F[B]): F[B]
 
   /**
-   * Masks cancellation while evaluating an effect, offering the ability to 
+   * Masks cancellation while evaluating an effect, offering the ability to
    * unmask within that effect. Unmasking is more commonly referred to as
    * "polling."
    */
@@ -93,7 +93,7 @@ trait MonadCancel[F[_], E] extends MonadError[F, E] {
    * during the evaluation of `fa`. If the evaluation of `fa` completes
    * without encountering a cancellation, the finalizer is unregistered
    * before proceeding.
-   * 
+   *
    * During finalization, all actively registered finalizers are invoked
    * in the opposite order of registration.
    *
@@ -105,15 +105,15 @@ trait MonadCancel[F[_], E] extends MonadError[F, E] {
   /**
    * Specifies an effect that is always invoked after evaluation of `fa`
    * completes, regardless of the outcome.
-   * 
+   *
    * This function can be thought of as a combination of `flatTap`,
    * `onError`, and [[MonadCancel!.onCancel onCancel]].
    *
    * @param fa The effect that is run after `fin` is registered.
    * @param fin The effect to run in the event of a cancellation or error.
-   * 
+   *
    * @see [[guaranteeCase]] for a more powerful variant
-   * 
+   *
    * @see [[Outcome]] for the various outcomes of evaluation
    */
   def guarantee[A](fa: F[A], fin: F[Unit]): F[A] =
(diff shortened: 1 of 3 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
```diff
@@ -30,34 +30,179 @@ import cats.data.{
 }
 import cats.syntax.all._
 
+/**
```
RaasAhsan, line 33: Any way to limit the line width here? I'd hate to make sure every line falls below 80 on my own, would be nice if that was automated by a formatter

kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
```diff
@@ -30,34 +30,179 @@ import cats.data.{
...
+   *        outcome.
+   * 
+   * @see [[bracketCase]] for a more powerful variant
+   * 
+   * @see [[Outcome]] for the various outcomes of evaluation
+   */
   def guaranteeCase[A](fa: F[A])(fin: Outcome[F, E, A] => F[Unit]): F[A] =
     bracketCase(unit)(_ => fa)((_, oc) => fin(oc))
 
+  /**
```
RaasAhsan, line 215: Most of the `bracket` ones are kind of copy-and-paste, not sure if this is the best way to deal with combinators with a ton of variants

kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
RaasAhsan: Following CE2, I sort of tried to cover the class from a high-level, leaving examples and more explanation to the site documentation. maybe there should be less detail here though

kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
```diff
@@ -30,34 +30,179 @@ import cats.data.{
...
   def forceR[A, B](fa: F[A])(fb: F[B]): F[B]
 
+  /**
+   * Masks cancellation while evaluating an effect, offering the ability to 
+   * unmask within that effect. Unmasking is more commonly referred to as
+   * "polling."
+   */
   def uncancelable[A](body: Poll[F] => F[A]): F[A]
 
-  // produces an effect which is already canceled (and doesn't introduce an async boundary)
```
RaasAhsan, line 40: I think it's reasonable for an implementation to use an async boundary here if they want

mpilquist (approved):

kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
```diff
@@ -30,34 +30,184 @@ import cats.data.{
 }
 import cats.syntax.all._
 
+/**
+ * A type class that characterizes monads that support cancellation, masking
+ * and finalization. [[MonadCancel]] instances must provide an instance for
```
djspiewak: Editorial nits:

- We usually use "typeclass" instead of "type class" just to distinguish from the OO thing
- "A typeclass **that**", but "monads **which**" (since the monads are non-specific)
- "masking," (I like the oxford comma since it creates fewer ambiguities in conjoined lists)

kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
```diff
@@ -30,34 +30,184 @@ import cats.data.{
...
+ * Cancellation refers to the act of requesting that the execution of a fiber be
+ * terminated. [[MonadCancel]] exposes a means of self-cancellation, where a fiber
+ * can request that its own execution be terminated. Self-cancellation is achieved
+ * via [[MonadCancel!.canceled canceled]].
+ *
+ * Cancellation is vaguely similar to the short-circuit behavior introduced by [[MonadError]],
+ * but there are several key differences:
+ *
+ *   1. Cancellation is effective; if it is observed it must be respected, and it cannot
+ *      be reversed.
```
djspiewak: IMO it's worth adding another sentence here noting the contrast with `handleError`, which allows you to "catch" an error and then continue normally, whereas there is no way to "catch" cancelation.

kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
```diff
@@ -30,34 +30,184 @@ import cats.data.{
 }
 import cats.syntax.all._
 
+/**
+ * A type class that characterizes monads that support cancellation, masking
+ * and finalization. [[MonadCancel]] instances must provide an instance for
+ * [[MonadError]].
+ *
+ * Cancellation refers to the act of requesting that the execution of a fiber be
+ * terminated. [[MonadCancel]] exposes a means of self-cancellation, where a fiber
```
djspiewak: I think it's worth clarifying here something like the following:

> In this context, "fiber" means "a chain of effects bound together by `flatMap`". So, `fa.flatMap(a => f(a)).flatMap(b => g(b))` is a fiber (or, more likely, one part of a longer fiber).

kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
```diff
@@ -30,34 +30,184 @@ import cats.data.{
...
+ * Cancellation is vaguely similar to the short-circuit behavior introduced by [[MonadError]],
+ * but there are several key differences:
+ *
+ *   1. Cancellation is effective; if it is observed it must be respected, and it cannot
+ *      be reversed.
+ *   1. Cancellation can be masked via [[MonadCancel!.uncancelable]]. This allows a fiber to
+ *      suppress cancellation for a period of time. If a fiber is cancelled while it is masked,
+ *      it must respect cancellation whenever it reaches an unmasked state.
+ *   1. [[GenSpawn]] introduces external cancellation, another cancellation mechanism by which fibers
+ *      can be cancelled by external parties. This can be likened to asynchronous exception throwing.
```
djspiewak: > This can be likened to asynchronous exception throwing.

Is this a helpful piece of information for non-Haskell developers?

kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
```diff
@@ -30,34 +30,184 @@ import cats.data.{
...
+ *
+ * Finalization refers to the act of invoking finalizers, which are effects that are guaranteed to be
+ * run in the event of cancellation. Finalization is achieved via [[MonadCancel!.onCancel onCancel]],
+ * which registers a finalizer for the duration of some arbitrary effect. If a fiber is cancelled during
+ * evaluation of that effect, the registered finalizer is guaranteed to be invoked before terminating.
```
djspiewak: Is it worth mentioning here that finalizers are invoked in nested order, with the inner-most finalizers running first? We could just put that on the `onCancel` doc.

kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
```diff
@@ -30,34 +30,184 @@ import cats.data.{
...
 
-  // analogous to productR, except discarding short-circuiting (and optionally some effect contexts) except for cancelation
+  /**
+   * Analogous to [[productR]], but suppresses short-circuiting behavior
+   * except for cancellation.
+   */
   def forceR[A, B](fa: F[A])(fb: F[B]): F[B]
 
+  /**
+   * Masks cancellation while evaluating an effect, offering the ability to
```
djspiewak: I like how you introduce the masking terminology without explicitly introducing it. That's really nice.

kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
```diff
@@ -30,34 +30,184 @@ import cats.data.{
...
   def canceled: F[Unit]
 
+  /**
+   * Registers a finalizer that is invoked if cancellation is observed
+   * during the evaluation of `fa`. If the evaluation of `fa` completes
+   * without encountering a cancellation, the finalizer is unregistered
+   * before proceeding.
+   *
+   * During finalization, all actively registered finalizers are invoked
+   * in the opposite order of registration.
```
djspiewak: Rather than saying "opposite order" I would say something like "innermost-first". We want to get people in the habit of thinking about their program semantics lexically, since those are the semantics we've very strictly enforced across the entire calculus.

kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
```diff
@@ -30,34 +30,184 @@ import cats.data.{
...
+   * @param fin The finalizer to register before evaluating `fa`.
+   */
   def onCancel[A](fa: F[A], fin: F[Unit]): F[A]
 
+  /**
+   * Specifies an effect that is always invoked after evaluation of `fa`
+   * completes, regardless of the outcome.
+   *
+   * This function can be thought of as a combination of `flatTap`,
+   * `onError`, and [[MonadCancel!.onCancel onCancel]].
```
djspiewak: Note to self: we need to get doclinks setup for cats and stuff.

kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
```diff
@@ -30,34 +30,184 @@ import cats.data.{
...
+   */
   def guarantee[A](fa: F[A], fin: F[Unit]): F[A] =
     guaranteeCase(fa)(_ => fin)
 
+  /**
+   * Specifies an effect that is always invoked after evaluation of `fa`
+   * completes, but depends on the outcome.
+   *
+   * This function can be thought of as a combination of `flatMap`,
+   * `onError`, and [[MonadCancel!.onCancel onCancel]].
```
djspiewak: Still `flatTap`.

kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
```diff
@@ -30,34 +30,184 @@ import cats.data.{
...
+   *
+   * @see [[bracketCase]] for a more powerful variant
+   *
+   * @see [[Outcome]] for the various outcomes of evaluation
+   */
   def guaranteeCase[A](fa: F[A])(fin: Outcome[F, E, A] => F[Unit]): F[A] =
     bracketCase(unit)(_ => fa)((_, oc) => fin(oc))
 
+  /**
+   * A pattern for interacting with effectful lifecycles.
```
djspiewak: I love that you're identifying `bracket` as a "pattern". That's perfect.

Do you think it's worth calling out that this is just implemented in terms of `uncancelable` and `onCancel` and friends? Specifically saying something like, if you need more customized bracketing behavior, you may want to use `uncancelable` and `onCancel` directly. Not sure.

djspiewak (changes requested):
Really really nice

kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
RaasAhsan: good point

kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
RaasAhsan: It actually felt a bit weird to talk about fibers in the context of `MonadCancel` since there is no notion of starting one, but I'll try to fit this in somewhere

laws/shared/src/main/scala/cats/effect/laws/MonadCancelLaws.scala
```diff
@@ -60,6 +60,12 @@ trait MonadCancelLaws[F[_], E] extends MonadErrorLaws[F, E] {
 
   def forceRCanceledShortCircuits[A](fa: F[A]) =
     F.forceR(F.canceled)(fa) <-> F.productR(F.canceled)(fa)
+
+  def uncancelableFinalizers[A](fin: F[Unit]) =
```
RaasAhsan, line 64: Added these two new laws, let me know if they even make sense and/or can be minimized more

kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
vasilmkd: > Any way to limit the line width here? I'd hate to make sure every line falls below 80 on my own, would be nice if that was automated by a formatter

Not automatically. I usually add a visual ruler guideline in the editor and make sure to follow it manually.

laws/shared/src/main/scala/cats/effect/laws/MonadCancelLaws.scala
```diff
@@ -60,6 +60,9 @@ trait MonadCancelLaws[F[_], E] extends MonadErrorLaws[F, E] {
...
+    F.onCancel(F.canceled, F.canceled >> fin) <-> F.onCancel(F.canceled, fin)
```
djspiewak: I would phrase this a little differently:

```scala
F.onCancel(fa, F.uncancelable(_ => fin)) <-> F.onCancel(fa, fin)
```

Comes down to the same thing, but is a bit more general of a formulation.

PR #1242, opened 2020-09-27
Override ref combinators for better performance (with benchmarks)

bplommer:
Prompted by the finding that [using Ref combinators in fs2 reduces benchmark performance](https://github.com/typelevel/fs2/pull/2050).

With these changes, `getAndUpdate` has about 6% higher throughput in the benchmark compared to `modify`, whereas in CE 2.2.0 it has about 13% lower throughput.

Benchmark results for this PR:
```
Benchmark                   Mode  Cnt     Score    Error  Units
RefBenchmark.getAndUpdate  thrpt   20  3403.336 ± 77.681  ops/s
RefBenchmark.modify        thrpt   20  3208.759 ± 30.875  ops/s
```

CE 2.2.0:
```
Benchmark                   Mode  Cnt     Score    Error  Units
RefBenchmark.getAndUpdate  thrpt   20  2709.840 ± 25.734  ops/s
RefBenchmark.modify        thrpt   20  3123.585 ± 37.992  ops/s
```

Edit: the differences are far smaller with Scala 2.13, but still there - see comment below

### DIFF
--- a/benchmarks/run-benchmark
+++ b/benchmarks/run-benchmark
@@ -19,7 +19,7 @@ date = Time.now.strftime("%Y-%m-%d")
 
 exec("mkdir -p benchmarks/results/#{date}")
 exec(
-    "sbt ++2.12.4 " +
+    "sbt ++2.12.12 " +
     "benchmarksPrev/clean benchmarksNext/clean " +
     "'benchmarksPrev/jmh:run -o ../results/#{date}/#{name}-Prev.txt  -i 10 -wi 10 -f 2 -t 1 cats.effect.benchmarks.#{name}' " +
     "'benchmarksNext/jmh:run -o ../results/#{date}/#{name}-Next.txt  -i 10 -wi 10 -f 2 -t 1 cats.effect.benchmarks.#{name}'"
--- a/benchmarks/shared/src/main/scala/cats/effect/benchmarks/RefBenchmark.scala
+++ b/benchmarks/shared/src/main/scala/cats/effect/benchmarks/RefBenchmark.scala
@@ -1,0 +1,78 @@
+/*
+ * Copyright (c) 2017-2019 The Typelevel Cats-effect Project Developers
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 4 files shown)


RaasAhsan:
Wow, I figured the derived methods would take somewhat of a hit but wasn't expecting it to be that bad. Does the performance hit ultimately come from the two extra allocations in `modify` (`B` and a tuple) that the `update` variants don't need?

Also, could you create a corresponding PR for Cats Effect 3 or at least open an issue for it? Don't want to lose these things :)

RaasAhsan (approved):

bplommer:
> Wow, I figured the derived methods would take somewhat of a hit but wasn't expecting it to be that bad. Does the performance hit ultimately come from the two extra allocations in `modify` (`B` and a tuple) that the `update` variants don't need?

Not sure, I haven't drilled down into the causes. I guess those would account for the benefit of non-derived `update` over `modify`, and the benefit of `modify` over derived `update` would be from not needing to instantiate an extra function.

It's slightly surprising that the benchmark improvement moving from derived `update` to `modify` is greater than that the improvement moving from `modify` to non-derived `update` - I guess that's to do with diminishing returns as the benchmark spends less time actually exercising `Ref`.

> Also, could you create a corresponding PR for Cats Effect 3 or at least open an issue for it?

Yep - https://github.com/typelevel/cats-effect/issues/1244. Will pick it up later if I have time.

bplommer:
Just ran the benchmarks with Scala 2.13 - the differences are far smaller (~4% penalty for derived `getAndUpdate`, ~4% boost for non-derived `getAndUpdate`), but I guess still worthwhile:

CE 2.2:
```
Benchmark                   Mode  Cnt     Score     Error  Units
RefBenchmark.getAndUpdate  thrpt   20  3028.729 ± 115.220  ops/s
RefBenchmark.modify        thrpt   20  3167.938 ±  71.214  ops/s
```
This PR:
```
Benchmark                   Mode  Cnt     Score    Error  Units
RefBenchmark.getAndUpdate  thrpt   20  3282.696 ± 31.261  ops/s
RefBenchmark.modify        thrpt   20  3164.669 ± 40.850  ops/s
```

djspiewak:
Remarkable!

PR #1250, opened 2020-10-01
Rename Outcome.Completed/ExitCase.Completed to Succeeded

bplommer:
This removes the ambiguity between "completed" meaning "completed successfully" and meaning "ran to completion". This can be a source of confusion, especially as it's used in the second sense e.g. in `fs2.Stream#onComplete`


djspiewak:
The more I think about this, the more I like it. :-) My initial reaction was that it was unnecessary, but I think you're correct that the rename is less ambiguous, particularly for someone without prior knowledge of `ExitCase.

PR #1251, opened 2020-10-02
GenSpawn and Fiber documentation

RaasAhsan:
This is still a work-in-progress, but I wanted to open it up early to get feedback. I have a general outline of how I want it to look, but concurrency is a very rigorous subject so I probably missed or got something wrong


kernel/shared/src/main/scala/cats/effect/kernel/Fiber.scala
```diff
@@ -18,8 +18,27 @@ package cats.effect.kernel
...
+ * @see [[GenSpawn]] documentation for more detailed information on
+ * concurrency of fibers.
+ */
 trait Fiber[F[_], E, A] {
+
+  /**
+   * Requests the cancellation of the fiber bound to this handle.
+   * 
+   * This function semantically blocks the caller until finalization of the 
+   * cancelee has completed.
```
djspiewak: Minor refinement: "This **effect** semantically blocks …"

I would also add here:

> If the cancelee is currently uncancelable, this effect will block until it becomes cancelable once again and all finalizers are complete. Cancelation is idempotent, and repeated calls have the same semantics as the first one: blocking until finalization is complete. Once the cancelee has been canceled and all finalizers have run, subsequent calls to `cancel` will immediately complete with `()`.
>
> Cancelation itself is uncancelable.

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -21,16 +21,156 @@ import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
...
+ * safely interact with eachother via three special functions. 
+ * [[GenSpawn!.start start]] spawns a fiber that executes concurrently with the 
+ * spawning fiber. [[Fiber!.join join]] semantically blocks the joining fiber 
+ * until the joinee fiber terminates, after which the outcome of the joinee is 
+ * returned. [[Fiber!.cancel cancel]] requests a fiber to abnormally terminate, 
+ * and semantically blocks the canceller until the cancellee has completed 
+ * finalization.
+ * 
+ * Just like threads, fibers can execute concurrently with respect to 
+ * eachother. This means that the effects of independent fibers may be
```
djspiewak: I believe "each other" is two words.

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -21,16 +21,156 @@ import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
...
+ * interleaved in an indeterminate fashion. This mode of concurrency is
```
djspiewak: > indeterminate

I think "nondeterministic" is a bit clearer.

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -21,16 +21,156 @@ import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
...
+ * returned. [[Fiber!.cancel cancel]] requests a fiber to abnormally terminate, 
+ * and semantically blocks the canceller until the cancellee has completed 
+ * finalization.
+ * 
+ * Just like threads, fibers can execute concurrently with respect to 
+ * eachother. This means that the effects of independent fibers may be
+ * interleaved in an indeterminate fashion. This mode of concurrency is 
+ * beneficial to modular program design: fibers that are described separately 
+ * can execute simultaneously without requiring programmers to explicitly
+ * yield back to the runtime system.
```
djspiewak: This isn't *strictly* true though. It's certainly not true at all on JavaScript. I love this paragraph though so I'm trying to think of how to make this more accurate without muddling the message.

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -21,16 +21,156 @@ import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
...
+ *   3. A1, B1, B2, A2
+ *   4. B1, B2, A1, A2
+ *   5. B1, A1, B2, A2
+ *   6. B1, A1, A2, B3
+ * 
+ * Notice how every execution preserves sequential consistency of the effects
+ * within each fiber: `A1` always prints before `A2`, and `B1` always prints 
+ * before `B2`. However, there are no guarantees around how the effects of 
+ * both fibers will be ordered with respect to eachother; it is entirely
+ * non-deterministic.
```
djspiewak: I don't think this needs the hyphen.

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -21,16 +21,156 @@ import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
...
+ * unlike self-cancellation, external cancellation need not ever be observed
+ * because of JVM memory model guarantees
+ * 
+ * 
+ */
 trait GenSpawn[F[_], E] extends MonadCancel[F, E] {
 
+  /**
+   * A low-level primitive for requesting concurrent evaluation of an effect.
+   * TODO: Common bad pattern start.flatmap(_.cancel)
```
djspiewak: IMO, `start` itself is a common bad pattern. :-D

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -21,16 +21,156 @@ import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
...
+   * 
+   * use background instead
+   */
   def start[A](fa: F[A]): F[Fiber[F, E, A]]
 
-  // produces an effect which never returns
+  /**
+   * An effect that never completes, causing the fiber to semantically block
+   * indefinitely. This is the purely functional equivalent of an infinite
```
djspiewak: > purely functional

I would say "purely functional, asynchronous equivalent"

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -21,16 +21,156 @@ import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
...
+   * the runtime system that it wishes to suspend itself with the intention of 
+   * resuming later, at the discretion of the scheduler. This is in contrast to 
+   * preemptive multitasking, in which threads of control are forcibly 
+   * suspended after a well-defined time slice. Preemptive and cooperative 
+   * multitasking are both features of runtime systems that influence the 
+   * fairness and throughput properties of an application.
+   * 
+   * Note that `cede` is merely a hint to the runtime system; implementations
+   * have the liberty to interpret this method to their liking as long as it
+   * obeys the respective laws. For example, a lawful implementation of this
```
djspiewak: I would say "a lawful (but unexpected) implementation of this function".

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -21,16 +21,156 @@ import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
...
+   * of the runtime system. This allows the carrier thread to resume execution 
+   * of another waiting fiber. 
+   * 
+   * `cede` is a means of cooperative multitasking by which a fiber signals to
+   * the runtime system that it wishes to suspend itself with the intention of 
+   * resuming later, at the discretion of the scheduler. This is in contrast to 
+   * preemptive multitasking, in which threads of control are forcibly 
+   * suspended after a well-defined time slice. Preemptive and cooperative 
+   * multitasking are both features of runtime systems that influence the 
+   * fairness and throughput properties of an application.
```
djspiewak: I think it's also worth noting here that preemptive and cooperative multitasking are not mutually exclusive. Most practical implementations incorporate a blend of the two, where cooperative behavior is possible in the event that a user *knows* that it's a particularly safe time to yield with minimal performance impact, but the runtime may still preempt if a fiber has run for a long time without yielding."

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -21,16 +21,156 @@ import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
...
+   * obeys the respective laws. For example, a lawful implementation of this 
+   * function is `F.unit`, in which case the fairness boundary is a no-op.
+   */
   def cede: F[Unit]
 
+  /**
+   * A low-level primitive for racing the evaluation of two arbitrary effects. 
+   * Returns the outcome of the winner and a handle to the fiber of the loser.
+   * 
+   * [[racePair]] is considered to be an unsafe function.
```
djspiewak: Worth noting *why* it's considered to be an unsafe function: the fact that `Fiber` appears in the result. This ties the concepts together with `start`. Also, I think we should be a little careful about the use of the word "unsafe". It may leak resources, but it's still referentially transparent. It isn't as unsafe as the `unsafe` unsafe things. :-)

Do you think it's worth noting here about how `racePair` (and races in general) cannot be *proven* to be lawfully referentially transparent in finite time, due to the fact that they introduce nondeterminism into the system? I think this is a really interesting factoid from a theory standpoint, though it has no impact whatsoever on users.

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
RaasAhsan: Which part of it is wrong: simultaneous execution or explicit yielding? I think the simultaneous execution part certainly isn't true for JS, so we say something like "appear to execute at the same time", but that's a little hand-wavy for the JVM too

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
RaasAhsan: Good idea, we could be more specific and call them "cancellation-unsafe" or maybe just call them "leaky"?

Is `racePair` not referentially transparent by definition? It's just a description of a computation, no less than `delay` or `realTime`. I guess I'm wondering if your emphasis on "proving lawful referentially transparency" is suggesting otherwise.

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
RaasAhsan: good point, we should just double down on using `background` :)

RaasAhsan:
Last step is to document the typeclass methods in more detail, then this should be ready for a more exhaustive review

RaasAhsan:
@SystemFw if you've got any spare time, requesting your review here for correctness, accuracy, and completeness :) It would be useful to take a look at the `MonadCancel` docs in this branch as well (it was merged in before, but I made some more changes here)

SystemFw:
I'll have a look today hopefully :)

kernel/shared/src/main/scala/cats/effect/kernel/Fiber.scala
```diff
@@ -18,13 +18,57 @@ package cats.effect.kernel
...
+ * concurrency of fibers.
+ */
 trait Fiber[F[_], E, A] {
+
+  /**
+   * Requests the cancellation of the fiber bound to this `Fiber` handle.
+   *
+   * This effect semantically blocks the caller until finalization of the
+   * cancellee has completed. This means that if the cancellee is currently
+   * masked, `cancel` will block until it is unmasked and finalized.
```
djspiewak: I believe we defined `cancel` to be outright-`uncancelable` (note to self: need to check to see if this is actually in the laws; it should be).

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -21,19 +21,276 @@ import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
...
+ *   3. A1, B1, B2, A2
+ *   4. B1, B2, A1, A2
+ *   5. B1, A1, B2, A2
+ *   6. B1, A1, A2, B3
+ *
+ * Notice how every execution preserves sequential consistency of the effects
+ * within each fiber: `A1` always prints before `A2`, and `B1` always prints
+ * before `B2`. However, there are no guarantees around how the effects of
+ * both fibers will be ordered with respect to each other; it is entirely
+ * nondeterministic.
```
djspiewak, line 77: This whole section is great

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -21,19 +21,276 @@ import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
...
+ *
+ * ==Cancellation==
+ *
+ * [[MonadCancel]] introduces a simple means of cancellation, particularly
+ * self-cancellation, where a fiber can request the abnormally terminatation of
```
djspiewak:
```suggestion
 * self-cancellation, where a fiber can request the abnormal terminatation of
```

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -21,19 +21,276 @@ import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
...
+ *   // After 10 seconds, cancel the fiber.
+ *   F.start(F.delay(println("A")).foreverM).flatMap { fiber =>
+ *     F.sleep(10.seconds) *> fiber.cancel
+ *   }
+ *
+ * }}}
+ *
+ * In the above example, imagine the spawning fiber is cancelled after it
+ * starts the printing fiber, but before the latter is cancelled. In this
+ * situation, the printing fiber is cancelled and will continue executing
```
djspiewak:
```suggestion
 * situation, the printing fiber is not cancelled and will continue executing
```

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -21,19 +21,276 @@ import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
...
+ * forever, contending with other fibers for system resources. This is known
+ * as a fiber leak, which is similar to a thread leak.
```
djspiewak: It's worth emphasizing again here that this leak happened because the only fiber which had a reference to the printer was canceled.

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -21,19 +21,276 @@ import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
...
+ *
+ * Resource leaks like this are unfavorable when writing applications. In
+ * the case of [[start]] and [[racePair]], it is recommended against using
```
djspiewak:
```suggestion
 * the case of [[start]] and [[racePair]], it is recommended not to use
```

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -21,19 +21,276 @@ import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
...
+ * these methods in favor of [[background]] and [[race]] respectively.
```
djspiewak:
```suggestion
 * these methods; instead, use [[background]] and [[race]] respectively.
```

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -21,19 +21,276 @@ import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
...
+ *
+ * The following example depicts a safer version of the `start` example
```
djspiewak:
```suggestion
 * The following example depicts a safer version of the [[start]] example
```

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -21,19 +21,276 @@ import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
...
+ *   // After 10 seconds, the resource scope exits so the fiber is cancelled.
+ *   F.background(F.delay(println("A")).foreverM).use { _ =>
+ *     F.sleep(10.seconds)
+ *   }
+ *
+ * }}}
+ *
+ * ==Scheduling==
+ *
+ * Fibers are commonly referred to as *lightweight threads* or *green threads*.
```
djspiewak: I do not believe asterisks italicize. Instead, I think we have to use `''`:

```suggestion
 * Fibers are commonly referred to as ''lightweight threads'' or ''green threads''.
```

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -21,19 +21,276 @@ import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
...
+ * This alludes to the nature by which fibers are schdeuled by runtime systems:
```
djspiewak:
```suggestion
 * This alludes to the nature by which fibers are scheduled by runtime systems:
```

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -21,19 +21,276 @@ import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
...
+ * many fibers are multiplexed onto one or more native threads.
+ *
+ * For applications running on the JVM, the scheduler typically manages a thread
+ * pool onto which fibers are scheduled. These fibers are executed
+ * simultaneously by the threads in the pool, achieving both concurrency and
+ * parallelism. For applications running on JavaScript platforms, all compute
+ * is restricted to a single worker thread, so multiple fibers must share that
+ * worker thread (dictated by fairness properties), achieving concurrency.
```
djspiewak:
```suggestion
 * worker thread (dictated by fairness properties), achieving concurrency without parallelism.
```

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -21,19 +21,276 @@ import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
...
+ * that influence the fairness and throughput properties of an application.
+ * These modes of scheduling are not necessarily mutually exclusive: a runtime
+ * system may incorporate a blend of the two, where users may yield at their
+ * own discretion, but the runtime may preempt a fiber if it has not yielded
+ * for some time.
+ *
+ * For more details on schedulers, refer to the following resources:
+ *
+ *   1. https://gist.github.com/djspiewak/3ac3f3f55a780e8ab6fa2ca87160ca40
+ *   1. https://gist.github.com/djspiewak/46b543800958cf61af6efa8e072bfd5c
```
djspiewak, line 189: LOL we really need to get this information onto the main CE website

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -21,19 +21,276 @@ import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
...
+ * resume  at some later time at the discretion of the scheduler. This is in
+ * contrast to preemptive multitasking, in which threads of control are forcibly
+ * yielded after a well-defined time slice.
+ *
+ * Preemptive and cooperative multitasking are both features of runtime systems
+ * that influence the fairness and throughput properties of an application.
+ * These modes of scheduling are not necessarily mutually exclusive: a runtime
+ * system may incorporate a blend of the two, where users may yield at their
+ * own discretion, but the runtime may preempt a fiber if it has not yielded
+ * for some time.
```
djspiewak: This paragraph is dope. So very clear.

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -21,19 +21,276 @@ import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
...
+ *   1. https://gist.github.com/djspiewak/46b543800958cf61af6efa8e072bfd5c
+ */
 trait GenSpawn[F[_], E] extends MonadCancel[F, E] {
 
+  /**
+   * A low-level primitive for starting the concurrent evaluation of a fiber.
+   * Returns a [[Fiber]] that can be used to wait for a fiber or cancel it.
+   *
+   * [[start]] is a cancellation-unsafe function; it is recommended to
+   * use the safer variant [[background]] to spawn fibers.
```
djspiewak:
```suggestion
   * use the safer variant, [[background]], to spawn fibers.
```

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -21,19 +21,276 @@ import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
...
+   * @param fa the effect for the fiber
+   *
+   * @see [[background]] for the safer, recommended variant
+   */
   def start[A](fa: F[A]): F[Fiber[F, E, A]]
 
-  // produces an effect which never returns
+  /**
+   * Returns a [[Resource]] that manages the concurrent execution of a fiber.
+   * The inner resource can be used to wait on the outcome of the underlying
```
djspiewak:
```suggestion
   * The inner effect can be used to wait on the outcome of the underlying
```

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -21,19 +21,276 @@ import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
...
+   * fiber; it is effectively a [[Fiber!.join join]].
+   *
+   * When the [[Resource]] goes out of scope or if the spawning fiber is
+   * cancelled, the underlying fiber is cancelled.
```
djspiewak: Worth adding a sentence here emphasizing that this resolves the leakage, since the outer fiber cannot be canceled without the inner fiber also being canceled.

Also, we should prefer the single `l` spelling, since that's what's used in the API.

Also also, how do you feel about "parent" and "child" for describing this kind of fiber relationship?

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -42,6 +299,26 @@ trait GenSpawn[F[_], E] extends MonadCancel[F, E] {
...
+   * The semantics of [[race]] are described by the following rules:
+   *
+   *   1. If the winner completes with [[Succeeded]], the race returns the
+   *   successful value. The loser is cancelled before returning.
+   *   2. If the winner completes with [[Errored]], the race raises the error.
+   *   The loser is cancelled before returning.
+   *   3. If the winner completes with [[Canceled]], the race returns the
+   *   result of the loser, consistent with the first two rules.
+   *   4. If both the winner and loser complete with [[Cancelled]], the race
+   *   is cancelled.
```
djspiewak: It's probably worth noting that if the race is canceled but masked by `uncancelable`, then the fiber will block indefinitely.

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -78,6 +366,30 @@ trait GenSpawn[F[_], E] extends MonadCancel[F, E] {
...
+   *   1. If the winner completes with [[Succeeded]], the race waits for the
+   *   loser to complete.
+   *   2. If the winner completes with [[Errored]], the race raises the error.
+   *   The loser is cancelled.
+   *   3. If the winner completes with [[Canceled]], the loser and the race
+   *   are cancelled as well.
+   *   4. If the loser completes with [[Succeeded]], the race returns the
+   *   successful value of both fibers.
+   *   5. If the loser completes with [[Errored]], the race returns the error.
+   *   6. If the loser completes with [[Canceled]], the race is cancelled.
```
djspiewak: Similarly worth noting here about `uncancelable` and `both`.

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
RaasAhsan: lol you caught the `abnormal` but not the `terminatation`

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
RaasAhsan: > Also, we should prefer the single l spelling, since that's what's used in the API.

Don't say these kinds of things, "cancelled" is the one and true spelling 😅

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
RaasAhsan: I just realized, i think we need to `poll` on all the `F.canceled` here

kernel/shared/src/main/scala/cats/effect/kernel/Fiber.scala
RaasAhsan:
```scala
def uncancelableCancelCancels =
    F.start(F.never[Unit]).flatMap(f => F.uncancelable(_ => f.cancel) >> f.join) <-> F.pure(
      Outcome.Canceled())
```

PR #1256, opened 2020-10-03
Removed leftover internal case after Effect was canned

djspiewak:
Forgot to clean this up. It's no longer needed since its sole purpose was to support `IO#to`


vasilmkd (approved):
Sad to see it go. Kidding.

PR #1257, opened 2020-10-04
Use syntax in typeclasses

bplommer:


djspiewak:
This makes things look much better, but it's dangerous because it introduces cycles! The problem is that the syntax is, itself, dependent on the classes. Both Scalaz and Cats ran into problems with cycle detection in the compiler when doing things like this, and so instead they tend to use the explicit syntax rather than the enrichments.

bplommer:
What if a given TC only uses syntax for its supertypes, rather than syntax for itself (so `GenSpawn` uses `MonadCancel` syntax, `MonadCancel` uses `MonadError` syntax - I don't think there should be any cycles then?

djspiewak:
> What if a given TC only uses syntax for its supertypes, rather than syntax for itself (so GenSpawn uses MonadCancel syntax, MonadCancel uses MonadError syntax - I don't think there should be any cycles then?

I *think* that avoids cycles, so long as you do the specific syntax import rather than the `all` one. (you can do `all` for Cats itself)

SystemFw:
I think a FAQ for this would be nice

djspiewak:
> I think a FAQ for this would be nice

There are quite a few dev FAQs that we should document, including wtf `def tag` is all about.

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -20,8 +20,10 @@ import cats.~>
 import cats.data.{EitherT, Ior, IorT, Kleisli, OptionT, WriterT}
 import cats.{Monoid, Semigroup}
 import cats.syntax.all._
+import cats.effect.kernel.syntax.monadCancel._
 
 trait GenSpawn[F[_], E] extends MonadCancel[F, E] {
+  implicit private def F: MonadCancel[F, E] = this
```
djspiewak: Oh one quick correction: can we make these `private[this]`? HotSpot is able to do slightly better with that scenario.

djspiewak (approved):

bplommer:
The CI failure seems to be unrelated to these changes:

```
Error:  x get within onCancel - 1
Error:   false -> false != true -> true (file:///home/runner/work/cats-effect/cats-effect/core/js/target/scala-0.27/cats-effect-test-fastopt.js:34311)
```

djspiewak:
Uh oh

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
ahjohannessen: @djspiewak I think `private[this]` is being dropped in [Scala 3](https://dotty.epfl.ch/docs/reference/dropped-features/this-qualifier.html)

PR #1258, opened 2020-10-04
Parameterize Resource error type - WIP

bplommer:
As discussed in #1114 - if the typeclasses are parametric in their error type, `Resource` probably should be too. Initially I'm making the changes in-place to keep the diff readable - the idea is to subsequently rename this to `GenResource` and make `Resource` a special case.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/LiftIO.scala
+++ b/core/shared/src/main/scala/cats/effect/LiftIO.scala
@@ -137,21 +137,21 @@ object LiftIO {
     new LiftIO[IO] { override def liftIO[A](ioa: IO[A]): IO[A] = ioa }
 
   /**
-   * [[LiftIO]] instance for [[Resource]] values.
+   * [[LiftIO]] instance for [[GenResource]] values.
    */
-  implicit def catsEffectLiftIOForResource[F[_], E](
+  implicit def catsEffectLiftIOForGenResource[F[_], E](
       implicit F00: LiftIO[F],
-      F10: Applicative[F]): LiftIO[Resource[F, E, *]] =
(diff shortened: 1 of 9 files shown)


SystemFw:
Not especially thrilled about this change (didn't manage to comment on the issue in time). It's not clear to me why we want to do it (and just aligning with the `Gen` typeclasses is not that strong), plus I'm afraid people will go for `Resource[IO, Error, A]`.

bplommer:
> It's not clear to me why we want to do it (and just aligning with the `Gen` typeclasses is not that strong)

My thinking was that `Resource` doesn't capture side-effects, so hard-coding the error type to `Throwable` is an arbitrary and unnecessary restriction. My understanding was that this is the same rationale as for the `Gen` typeclasses, but maybe there was a more specific rationale for those that doesn't apply here?

> plus I'm afraid people will go for `Resource[IO, Error, A]`.

Not sure I follow - generally there won't be a `Bracket[IO, Error]` instance available, so that (or rather the `use`/`allocated` call) won't compile.

SystemFw:
None of my objections to the PR are a hard technical problem, more of a UX/added complexity one.

That being said, if other people think there is a compelling case for this (e.g. using `Resource` with `EitherT`), I'm cool with that

bplommer:
> None of my objections to the PR are a hard technical problem, more of a UX/added complexity one.

My hope is that once this is renamed to `GenResource` and `Resource` is special-cased to `Throwable`, the API for `Resource` won't be any more complex than it is now. I think the naming should steer people towards using `Resource` rather than `GenResource` as a default.

PR #1261, opened 2020-10-04
Add Console

vasilmkd:
I messed something up when rebasing and pushing. I didn't mean to close the previous PR.

#1197


concurrent/shared/src/main/scala/cats/effect/concurrent/Console.scala
```diff
@@ -0,0 +1,180 @@
...
+ *   def myProgram[F[_]: Sync]: F[Unit] =
+ *     for {
+ *       _ <- Console[F].println("Please enter your name: ")
+ *       n <- Console[F].readLine
+ *       _ <- if (n.nonEmpty) Console[F].println(s"Hello $n!")
+ *            else Console[F].printlnErr("Name is empty!")
+ *     } yield ()
+ * }}}
+ */
+class Console[F[_]] private (val F: Sync[F]) extends AnyVal {
```
djspiewak: Should be `final`

concurrent/shared/src/main/scala/cats/effect/concurrent/Console.scala
```diff
@@ -0,0 +1,180 @@
...
+      loop()
+    }
+
+  /**
+   * Prints a value to the standard output using the implicit `cats.Show` instance.
+   *
+   * @param a value to be printed to the standard output
+   */
+  def print[A: Show](a: A): F[Unit] =
+    F.blocking(System.out.print(a.show))
```
djspiewak: Relatively minor niggle, but `a.show` should be run *outside* of the `blocking` pool.

```suggestion
  def print[A: Show](a: A): F[Unit] = {
    val text = a.show
    F.blocking(System.out.print(text))
  }
```

Same comment for the other `print`s.

concurrent/shared/src/main/scala/cats/effect/concurrent/Console.scala
```diff
@@ -0,0 +1,180 @@
...
+   */
+  def println[A: Show](a: A): F[Unit] =
+    F.blocking(System.out.println(a.show))
+
+  /**
+   * Prints a value to the standard error output using the implicit `cats.Show` instance.
+   *
+   * @param a value to be printed to the standard error output
+   */
+  def printErr[A: Show](a: A): F[Unit] =
```
djspiewak: Random idea: what if we called this `error` and the other `errorln`?

djspiewak (commented):
More thoughts incoming as I get into the tests

concurrent/shared/src/main/scala/cats/effect/concurrent/Console.scala
vasilmkd: Nice. Done.

concurrent/shared/src/main/scala/cats/effect/concurrent/Console.scala
vasilmkd: Done.

concurrent/shared/src/main/scala/cats/effect/concurrent/Console.scala
vasilmkd: I made the change.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -411,6 +412,23 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
...
+  def readLine: IO[String] =
+    Console[IO].readLine
+
+  /**
+   * Prints a value to the standard output followed by a new line using the implicit `cats.Show` instance.
+   *
+   * @param a value to be printed to the standard output
+   */
+  def println[A: Show](a: A): IO[Unit] =
+    Console[IO].println(a)
```
djspiewak: I think we should probably add the other forwarders as well: `print`, `error`, and `errorln`.

vasilmkd: I only added `IO.print`. `IO.error` might be really confusing to users. I also added scaladoc that points to the full `Console`.

Issue #1255, opened 2020-10-03
Resource.makeMask


SystemFw:
The design of `Resource` closely reflects the nature of `bracket` in CE2, in particular a `Resource.make` node translates to `bracket`. This is fine in a lot of cases, but it does force the `acquire` function to be uncancelable, which isn't great when using `Resource` with certain types of concurrent , e.g. `Resource.make(semaphore.acquire)(_ => semaphore.release)`.

This can be solved with ad-hoc techniques (see `Semaphore.withPermit` for an idea of how), but we can look for a more general solution that allows `make` to translate to both uninterruptible and interruptible `bracket`, by having it take a `Poll[F]`.

There is one significant caveat to this idea though: `Resource` is currently interpreted outside of cats-effect to integrate with other resource safety frameworks, which would now have to support both interruptible and uninterruptible acquires. A typical example would be `fs2`, and the work required to support the enhanced this is likely to be highly nontrivial.

djspiewak:
Just to add a little fuel to the fire, I would consider this to be something that we absolutely need for 3.0.0, particularly since the algebra is open and we can't easily do it later. It doesn't block M1, but it should be considered a blocker for any RCs.

PR #1268, opened 2020-10-05
Resource.makeMask

bplommer:
I'm not sure whether this is a full solution to #1255, but it works for the given example (canceling a semaphore acquire). Still needs docs and specific tests, but `Semaphore.withPermit` continues to pass all tests when reimplemented using this.

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -380,6 +380,11 @@ object Resource extends ResourceInstances with ResourcePlatform {
       implicit F: Functor[F]): Resource[F, A] =
     apply[F, A](acquire.map(a => a -> release(a)))
 
+  def makeMask[F[_]: Monad, A](awaitAndRelease: F[(F[A], F[Unit])]): Resource[F, A] =
+    Resource.make(awaitAndRelease) { case (_, release) => release }.evalMap {
+      case (awaitCancelable, _) => awaitCancelable
+    }
+
   /**
    * Creates a resource from an acquiring effect and a release function that can
    * discriminate between different [[ExitCase exit cases]].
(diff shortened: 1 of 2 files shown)


SystemFw:
This doesn't fix #1255 since it requires a special encoding downstream (the `acquireNInternal` one). It's also worth noting that most of the work in #1255 is downstream of cats-effect unfortunately.

That being said, I'd rework the PR to only touch Semaphore, and it's a great addition there :)

bplommer:
What would a solution that doesn't require a special encoding at the call site look like? (setting aside difficulties with implementation.)  

I mean - if we want to be able to clean up after a canceled allocation, we can't express how to do that using a cancelable effect that looks like `F[(A, F[Unit])]`, as no finalizer will be returned in the cancellation case. I'm probably missing something here, but `F[(F[A], F[Unit])]` is the simplest encoding I can think of.

SystemFw:
> I'm probably missing something here, but F[(F[A], F[Unit])] is the simplest encoding I can think of.

So, that encoding is not unreasonable by any means, and it's what ZIO went with after Pilquist discovered it in `Semaphore.acquireNInternal`.

>  if we want to be able to clean up after a canceled allocation, we can't express how to do that using a cancelable effect that looks like F[(A, F[Unit])], as no finalizer will be returned in the cancellation case.

My thinking here is that the `F` knows how to clean up after itself when canceled, but if it's not canceled, then it's up to us. Imagine solving the problem with `Semaphore`, but without having access to `acquireNInternal`, which you can do with the new `MonadCancel` capabilities.
In particular, through `uncancelable`, you can express uninterruptible/interruptible acquire, and indeed we do in `bracketFull`, but you lose that because `Resource` embeds `bracket`, rather than `bracketFull`. So `makeMask` should just embed that, it's actually a fairly small change here. The big problem is downstream.

bplommer:
Oh I see - thanks for explaining 🙂 I'll close this PR and open a new one for `Semaphore#Permit`.

Are the changes in CE #1255 something that would be helpful to make a PR for at this stage?

SystemFw:
> something that would be helpful to make a PR for at this stage

I'm genuinely not sure

PR #1276, opened 2020-10-07
Update the fiber yielding explanation in basics.md

slouc:
> we can decide when we yield to other threads from the same pool by calling shift

> Allowing different threads to advance their work is called fairness

These are a bit misleading. We actually yield to other pending *fibers*, not other threads. Also, fairness is allowing different *fibers* to advance their work, not different threads (at least in this context... fairness among threads is also a thing, but that's achieved on CPU level).

It might sound like nitpicking, but to me current version sounds as if, when `shift` is called, some or all of the busy threads might stop their work and replace their current fibers with some of the pending fibers. And that will not happen. When we `shift` some fiber F on some thread T, that means that T might get another fiber, and that F might continue running on a different thread (if any are available). But busy threads will not be disrupted.

### DIFF
--- a/site/src/main/mdoc/concurrency/basics.md
+++ b/site/src/main/mdoc/concurrency/basics.md
@@ -175,10 +175,15 @@ In other words, `IO` is executing synchronously until we call `IO.shift` or use 
 In terms of individual thread pools, we can actually treat `IO` like **green thread** with
 [cooperative multitasking](https://en.wikipedia.org/wiki/Cooperative_multitasking).  Instead of
 [preemption](https://en.wikipedia.org/wiki/Preemption_(computing)#PREEMPTIVE),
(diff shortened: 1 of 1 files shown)


Avasil:
nitpicking docs is great, feel free to do more of it. :D 

I've written most of this section a while ago, before I had a solid understanding and it could probably use a refactor if you'd like to contribute some of your article

slouc:
> nitpicking docs is great, feel free to do more of it. :D
> 
> I've written most of this section a while ago, before I had a solid understanding and it could probably use a refactor if you'd like to contribute some of your article

@Avasil It's a good text and I learned a lot from it. 👍 I might throw a PR like this every once in a while, but I don't think it needs any major rework.

PR #1277, opened 2020-10-07
Shutdown threadpools in global

TimWSpence:


mpilquist:
Why support shutting down the global pool? Seems dangerous to me, as folks can shut it down manually, before the application shuts down. Contrived example but consider manually running IOApp instances in a REPL.

TimWSpence:
> Why support shutting down the global pool? Seems dangerous to me, as folks can shut it down manually, before the application shuts down. Contrived example but consider manually running IOApp instances in a REPL.

@mpilquist That's a good question. Maybe we should remove `Runtime#shutdown` entirely instead. I just noticed that as it stands, `Runtime` does have a `shutdown` that we invoke from `IOApp` but it's a noop and when we construct the global `IORuntime` we created callbacks to shutdown the pools and then throw those callbacks away so the current behaviour seems kind of wrong either way.

mpilquist:
The rationale was that if you provide your own pool in IOApp then we'd shut it down but if you use the JVM wide shared instance, we wouldn't. That seems okay to me but open to opinions!

TimWSpence:
I don't honestly have a strong opinion about it so we can close it. It just seemed a bit strange to me - surely a user could just as easily shut their own custom pool down too early?  I expected `global` to be treated the same as any other `Runtime`. But maybe I'm thinking too much about `IOApp` and not other use-cases

TimWSpence:
Closing this. Many thanks to @mpilquist and @alexandru for their feedback - `global` should not be shutdown as it's dangerous if someone shuts it down prematurely and it's also morally dubious as resources should be closed by whoever creates them. Nothing really fits this role for `global` so we leave it to the JVM

PR #1279, opened 2020-10-07
Rename Poll to Demask

bplommer:
`Poll` is a pretty opaque name. My first thought was that it should be called `Unmask`, but that carries the obvious ambiguity between "remove a mask" (which it does) and "render maskless" (which it may or may not do). I think `Demask` avoids that ambiguity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,7 +214,7 @@ Laws are more than just a [TCK](https://en.wikipedia.org/wiki/Technology_Compati
 
 Notice how the laws are written in terms of the `<->` "rewrite" operator. This operator simply turns into a ScalaCheck property in practice, but in *theory* it should be read as something much more powerful: an equational substitution.
 
-The first law is saying that any time we see a program which contains the pattern `F.uncancelable(poll => poll(...))` for *any* value of the `...` subexpression, we can rewrite it as just the `...` subexpression itself. This kind of step-wise evaluation semantic is the foundation of all formal computational calculi, including Scala itself (as of Dotty, via the DOT calculus).
+The first law is saying that any time we see a program which contains the pattern `F.uncancelable(demask => demask(...))` for *any* value of the `...` subexpression, we can rewrite it as just the `...` subexpression itself. This kind of step-wise evaluation semantic is the foundation of all formal computational calculi, including Scala itself (as of Dotty, via the DOT calculus).
 
 Unfortunately, due to expressiveness limitations, not *all* programs can be reduced to a simple composition of these evaluation rules (laws). However, significantly more such programs can be reduced now than was possible with the laws framework in ce2, in which a single law very frequently reduced to a rewrite asserting one side was equal to `F.pure(...)` of some value, which is a rewrite of very limited applicability.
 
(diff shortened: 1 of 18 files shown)


djspiewak:
I initially read this as "damask", as in the type of fabric. :-P

Floor is open for comments! "Poll" is *super* opaque, though I'm not sure that "demask" is that much more clear, though I do think @RaasAhsan's scaladoc efforts have made the situation better since he consistently refers to `uncancelable` as "masking cancelation".

bplommer:
Yeah, `Demask` isn't exactly explanatory, but it feels more mnemonically helpful to me because it hooks into the "masking" imagery. I can imagine pulling off a mask, which may or may not have another one underneath.

RaasAhsan:
I've sort of convinced myself that "mask" and "unmask" sound appropriate after writing a ton of Scaladoc with it :) Ultimately I think people are going to go looking at the docs to understand what these functions do, no matter what they're named. For me it comes down to aesthetics, and I much rather prefer `uncancelable`/`poll`, `mask`/`poll`, or `mask`/`unmask` in that order. `uncancelable`/`demask` is still a bit weird to me, but maybe it'll settle in

bplommer:
@RaasAhsan what is it you like about `poll`? Honest question 🙂

bplommer:
Incidentally - I actually like the fact that `demask` sounds a bit weird - it raises the question "why not `unmask`", which has an answer that explains how nested cancelation/demasking works.

RaasAhsan:
`poll` is already well-defined in this space: the analog that has existed in Concurrent Haskell for many years is called `mask`/`poll`. Obviously a newcomer wouldn't know that, but we get to define what it means, just like everything else. FWIW we already use and borrow a lot of opaque terminology, like `MonadCancel.bracket`, but also the names of many Cats typeclasses and their functions.

I don't really perceive a difference between "unmask" and "demask", so if I were to choose between the two, I'd opt for the former because  it's how we've been talking about this stuff for awhile and it's worked pretty well. I'm not sure if a distinction between the two will mean much to users. I also don't think either of those pair well with `uncancelable` :)

bplommer:
> I'm not sure if a distinction between the two will mean much to users. I also don't think either of those pair well with uncancelable :)

You're probably right. Here's another idea: rename `uncancelable` to `maskCancel` or `maskCancelation`, rename `Poll` to `Unmask`, and add `def uncancelable[A](fa: F[A]) = maskCancel(_ => fa)`. Then the names go well together, *and* we get a simplified `uncancelable` that does exactly what it says and doesn't expose `Poll`.

We could also move `Poll`/`Unmask` to the companion object of `MonadCancel` to provide extra context, so it would be `MonadCancel.Poll`/`MonadCancel.Unmask`.

djspiewak:
The thing I like about "uncancelable" is it does exactly what it says it's going to do. If a newcomer is searching for answers and they find this one just by browsing scaladoc, they won't be led astray. The other half of it is just the hard part.

I like that "poll" is short and, just like demask, a little weird and so users are less likely to get the wrong idea from the name. "Demask" also has this nice property, though it's not as short and I'm going to misspell it every time.

I don't have super strong feelings on this one, though I do kinda like poll better. This is a good discussion and I hope more people will add their opinions!

SystemFw:
I still like `uncancelable/poll` better than any alternative so far

bplommer:
Seems the consensus is for the status quo - I'll close this tomorrow unless contrary views emerge.

vasilmkd:
I prefer `unmask/poll`.

PR #1280, opened 2020-10-07
Remove Defer constraint from Resource#mapK

bplommer:


joroKr21:
Hmm I wonder how is it possible that there is `Stream.resource` and `Stream.translate` without any constraints in fs2, but we need constraints for `Resource.mapK` :thinking:

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -227,15 +227,42 @@ sealed abstract class Resource[+F[_], +A] {
...
-  )(implicit D: Defer[H], G: Applicative[H]): Resource[H, A] =
-    this match {
-      case Allocate(resource) =>
-        Allocate(f(resource).map { case (a, r) => (a, r.andThen(u => f(u))) })
-      case Bind(source, f0) =>
-        Bind(Suspend(D.defer(G.pure(source.mapK(f)))), f0.andThen(_.mapK(f)))
-      case Suspend(resource) =>
-        Suspend(f(resource).map(_.mapK(f)))
-    }
+  )(implicit G: Applicative[H]): Resource[H, A] = {
```
joroKr21: It looks to me that you can weaken this to `Functor`.
There is also a choice to be made here - should the constraint be on `G` or `H`?
```suggestion
  )(implicit H: Functor[H]): Resource[H, A] = {
```

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -227,15 +227,42 @@ sealed abstract class Resource[+F[_], +A] {
...
+
+    sealed trait Stack[AA]
+    case object Nil extends Stack[A]
+    final case class Frame[AA, BB](head: AA => Resource[G, BB], tail: Stack[BB])
+        extends Stack[AA]
```
joroKr21, line 235: Is it possible to just use a `List[A => Resource[G, A]]`?

bplommer: Unfortunately no, because the type parameter changes each time. This is the same pattern as in `fold` and `allocated` - they used to use `List[Any => Resource[G, Any]]`, but that required a bunch of `asInstanceOf` invocations so we changed it to this.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
bplommer: > It looks to me that you can weaken this to `Functor`.

`Applicative` is needed for the `map` calls.

> There is also a choice to be made here - should the constraint be on `G` or `H`?

It needs to be on `H` for the `flatMap` calls - I don't think those can be done in `G`.

bplommer:
> Hmm I wonder how is it possible that there is `Stream.resource` and `Stream.translate` without any constraints in fs2, but we need constraints for `Resource.mapK` :thinking: 

They have completely different encodings - `Stream` uses a free monad. The trade-off is that Stream's encoding isn't part of its public API (so it can't be interpreted by user code like `Resource` can) and compiling it requires `Concurrent` (rather than `MonadCancel`).

(I'm sure others can explain this more accurately and in more depth than I can – I don't know much at all about fs2 internals 🙂)

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
bplommer: Oh - `flatMap` doesn't require any instance, so you're right, we could move the constraint from `H` to `G`. Do you see any particular benefit to having it on `G` rather than `H`?

bplommer:
More specifically - in `Resource` it's because `map` is encoded using `Bind`. I wonder if it's worth adding more primitive `Resource` subtypes to remove those constraints 🤔

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
joroKr21: > Applicative is needed for the map calls.

`Functor` gives you `map`

> Do you see any particular benefit to having it on G rather than H?

I don't have any arguments in either direction. Maybe that's just an arbitrary choice 🤔

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
joroKr21: Oh that's interesting. Still trying to wrap my head around it. But thanks for explaining 👍

SystemFw: > Still trying to wrap my head around it.

Imagine `resourceOfInt.flatMap(.. resourceOfString).flatMap(...resourceOfBoolean)`. 
To interpret it we need a stack of binds, which all have different types, although we do know the final result, which the type of the `Resource` we get as input. 
The easiest way to do it is the old `List[Any => Resource[F, Any]]` (which is used in `IO` as well): you have an untyped callstack, which is fine because:
- the user goes through a typed tree of `Resource` nodes, so they can't build anything wrong
- we know that we need to cast to `A` at the end of the interpretation process

Then it got changed to the type aligned sequence you see above: we know that the flatMaps have to be `A => F[B]`, `B => F[C]`, `C => F[D]`, etc, and that's encoded by the construction above (using the fact that `BB` is essentially existential there), and that avoids a bunch of casts to and from `Any => Resource[F, Any]`

joroKr21: Yeah I got that part about chaining `flatMap` calls, I just haven't convinced myself how the compiler is satisfied with that encoding and also wondering if there is a generic representation of a type-aligned sequence we can use.

SystemFw:
Yeah the general technique here is that you can always move a constraint from call site to interpretation side by adding a primitive node

bplommer:
> Yeah the general technique here is that you can always move a constraint from call site to interpretation side by adding a primitive node

Seems could do it in two stages - keep just the current three primitives publicly accessible, but add others that can be compiled down to those three. Then external interpreters don't need to deal with the extra nodes and the API stays stable, but we still get to remove the constraints from the call site.

SystemFw:
> Then external interpreters don't need to deal with the extra nodes 

why not? what happens when an external interpreter hits a `Map` node?

bplommer:
> > Then external interpreters don't need to deal with the extra nodes
> 
> why not? what happens when an external interpreter hits a `Map` node?

Probably easiest if I push a WIP PR to explain what I mean 🙂

bplommer:
Pushed as #1285

SystemFw:
I see. It makes sense but I wonder if it's worth over the simple changes required to support `LiftF` and `Pure` downstream. I would say it does not if `makeMask` goes ahead and changes are required anyway, and it's interesting if `makeMask` does not go ahead.

bplommer:
(incidentally, it wasn't straightforward to apply the same technique for `MapK` because of variance issues 😞)

bplommer:
> I see. It makes sense but I wonder if it's worth over the simple changes required to support `LiftF` and `Pure` downstream.

The main benefit I see is it adds the possibility to make further changes to the encoding in the future without breaking compatibility.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
bplommer: > `Functor` gives you `map`

You're right - my bad, I was thinking the `map` calls were on `Resource`.

djspiewak:
> I wonder if it's worth adding more primitive Resource subtypes to remove those constraints

I'm a little late to the party, but in my view, the answer to this is "yes". Obviously the burden is mostly borne by the fs2 devs, so I defer to them, but in general I prefer to see as few constraints on definitions as possible, with all the constraints being pushed to interpretation. This is a lot more composable in the long run, enabling unforeseen abstraction and use-cases that would be otherwise impossible.

PR #1281, opened 2020-10-07
Add Semaphore#permit returning Resource

bplommer:
Depends on #1280.

This adds an `Applicative` constraint to `imapK` but I can't see that being a problem.

### DIFF
--- a/core/shared/src/test/scala/cats/effect/std/SemaphoreSpec.scala
+++ b/core/shared/src/test/scala/cats/effect/std/SemaphoreSpec.scala
@@ -47,7 +47,7 @@ class SemaphoreSpec extends BaseSpec { outer =>
 
       op.flatMap { res =>
         IO {
-          res must beEqualTo(2: Long)
+          res must beEqualTo(2L)
         }
       }
     }
@@ -64,7 +64,7 @@ class SemaphoreSpec extends BaseSpec { outer =>
 
       op.flatMap { res =>
         IO {
-          res must beEqualTo(1: Long)
+          res must beEqualTo(1L)
         }
       }
     }
@@ -78,7 +78,7 @@ class SemaphoreSpec extends BaseSpec { outer =>
 
       op.flatMap { res =>
         IO {
-          res must beEqualTo(1: Long)
+          res must beEqualTo(1L)
         }
       }
     }
@@ -112,7 +112,7 @@ class SemaphoreSpec extends BaseSpec { outer =>
 
       op.flatMap { res =>
         IO {
-          res must beEqualTo(0: Long)
+          res must beEqualTo(0L)
         }
       }
     }
@@ -216,7 +216,7 @@ class SemaphoreSpec extends BaseSpec { outer =>
 
       op.flatMap { res =>
         IO {
-          res must beEqualTo(1: Long)
+          res must beEqualTo(1L)
         }
       }
     }
@@ -231,7 +231,7 @@ class SemaphoreSpec extends BaseSpec { outer =>
 
       op.flatMap { res =>
         IO {
-          res must beEqualTo(0: Long)
+          res must beEqualTo(0L)
         }
       }
     }
@@ -282,7 +282,7 @@ class SemaphoreSpec extends BaseSpec { outer =>
         case Outcome.Succeeded(ioa) =>
           ioa.flatMap { res =>
             IO {
-              res must beEqualTo(0: Long)
+              res must beEqualTo(0L)
             }
           }
         case _ => IO.pure(false must beTrue) //Is there a not a `const failure` matcher?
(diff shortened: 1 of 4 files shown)


core/shared/src/test/scala/cats/effect/std/SemaphoreSpec.scala
```diff
@@ -69,6 +69,19 @@ class SemaphoreSpec extends BaseSpec { outer =>
...
+    "permit.use does not leak fibers or permits upon cancelation" in real {
+      val op = Semaphore[IO](0L).flatMap { s =>
+        // Same as for `withPermit`
+        s.permit.use(_ => s.release).timeout(1.milli).attempt *> s.release *> IO.sleep(
+          10.millis) *> s.count
+      }
+
+      op.flatMap { res =>
+        IO {
+          res must beEqualTo(1: Long)
```
vasilmkd: Instead of `1: Long` you can write `1L`.

bplommer: Sure - I've changed it in the other places in that file too.

vasilmkd: It was mostly a nit and knowledge sharing, no worries 😄

bplommer:
It just occurred to me that if we remove `withPermit` then `imapK` can become `mapK`.

vasilmkd:
I am really hesitant about removing `withPermit`.

std/shared/src/main/scala/cats/effect/std/Semaphore.scala
```diff
@@ -249,6 +256,9 @@ object Semaphore {
         case Right(n) => n
       }
 
+    val permit: Resource[F, Unit] =
+      Resource.make(acquireNInternal(1))(_.release).evalMap(_.await)
```
djspiewak, line 254: Doesn't this have the cancelation issue? (i.e. isn't acquisition here uncancelable?)

SystemFw: `acquireNInternal` is Mike's idea of using `F[(F[A], F[Unit])]`, which solves the issue by splitting the state manipulation, which needs restoring but not interruption, from the waiting (which needs interruption but no restoring). It's so general than ZIO actually went with that `F[(F[A], F[Unit])]` for their Resource encoding, but it comes at the cost of having to complicate Semaphore's implementation.

djspiewak: Ahhh! Okay then, so `permit` works correctly, which makes me suspect that we simply don't need `withPermit` anymore, right?

SystemFw: `withPermit` can be replaced with `permit.use`, yes. The argument now is fully about weighting the `mapK` point in another discussion, with breaking code for users (there is precedent for this type of change in http4s client, which I have an argument for)

bplommer: We can also keep `withPermit` as a derived method defined in terms of `permit.use` - so it will only break user code if a `Resource.Bracket` instance isn't available.

djspiewak: I'm in favor of just removing `withPermit` entirely and going with `mapK` instead of `imapK`. Now is the time to clean stuff up. :-D

core/shared/src/test/scala/cats/effect/std/SemaphoreSpec.scala
```diff
@@ -47,28 +47,41 @@ class SemaphoreSpec extends BaseSpec { outer =>
...
       op.flatMap { res =>
         IO {
-          res must beEqualTo(2: Long)
+          res must beEqualTo(2L)
         }
       }
     }
 
-    "withPermit does not leak fibers or permits upon cancelation" in real {
+    "permit.use does not leak fibers or permits upon cancelation" in real {
```
djspiewak, line 55: I think this test is redundant now, right?

djspiewak (changes requested):

core/shared/src/test/scala/cats/effect/std/SemaphoreSpec.scala
bplommer: Why yes, yes it is 🤦‍♂️

bplommer:
Unrelated test failure:

```
[info]     timeout should
[info]       + succeed
[info]       + cancel a loop
Error:        x invoke finalizers on timed out things
Error:         the value is false (file:///home/runner/work/cats-effect/cats-effect/core/js/target/scala-2.13/cats-effect-test-fastopt.js:35155)
...
Error:  Failed tests:
Error:  	cats.effect.IOSpec
```

djspiewak:
My bad. Tracking in #1297

bplommer:
Should we add backport this to CE2 and deprecate `permit` there to help with the transition?

djspiewak:
I think that would be very worthwhile. Anything we can do to ease the transition is good.

bplommer:
On second thought, that wouldn't work - with CE2 `Resource` that would require adding new constraints to `imapK`, and we can't remove those constraints from `Resource` without breaking changes.

PR #1282, opened 2020-10-07
Add missing readLine encoding tests

### LINKED ISSUE #1263: Add unit tests for Console.readLine
Follow up to #1261. `readLine` is currently untested, and it's a relatively complex bit of functionality. We should add some unit tests based on canned pre-encoded files in various character sets. We should also ensure that both Windows and Unix line endings work.

At present, the above have been manually tested, but getting it automated should be considered a priority.

vasilmkd:
Resolves #1263.

I have canned files encoded for each of the `StandardCharsets` available on any JVM. Each file is also unique, with no files having characters which are unrepresentable in their encoding. This way we can be sure that these tests are an actual representation of a _user_ entering characters on their terminal, or in the worst case, opening files for which the real encoding is known ahead of time, as is the case in these tests.


core/jvm/src/test/scala/cats/effect/std/ConsoleJVMSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+class ConsoleJVMSpec extends BaseSpec {
+  sequential
+
+  private def fileContents(name: String, charset: Charset): IO[String] = {
+    val acquire =
+      IO.blocking(Source.fromResource(s"readline-test.$name.txt")(charset))
+    val release = (src: Source) => IO.blocking(src.close())
+    Resource
+      .make(acquire)(release)
+      .use(src => IO(src.getLines().mkString(System.lineSeparator())))
```
djspiewak: BTW this should be either `blocking` or (preferably) `interruptible`

core/jvm/src/test/scala/cats/effect/std/ConsoleJVMSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+
+import scala.io.Source
+
+class ConsoleJVMSpec extends BaseSpec {
+  sequential
+
+  private def fileContents(name: String, charset: Charset): IO[String] = {
+    val acquire =
+      IO.blocking(Source.fromResource(s"readline-test.$name.txt")(charset))
+    val release = (src: Source) => IO.blocking(src.close())
```
djspiewak: Perversely, this one can just be `apply` since I don't think `close()` ever blocks.

core/jvm/src/test/scala/cats/effect/std/ConsoleJVMSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+  private def readLines(name: String, charset: Charset): IO[String] = {
+    def inputStream(name: String): Resource[IO, InputStream] = {
+      val acquire =
+        IO.blocking {
+          Thread
+            .currentThread()
+            .getContextClassLoader()
+            .getResourceAsStream(s"readline-test.$name.txt")
+        }
+      val release = (in: InputStream) => IO.blocking(in.close())
```
djspiewak: `apply` again

Obviously none of this matters for tests, I'm just trying to get us to model good behavior for those who might read the tests looking for patterns. :-)

core/jvm/src/test/scala/cats/effect/std/ConsoleJVMSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+      Resource.make(acquire)(release)
+    }
+
+    def replaceStandardIn(in: InputStream): Resource[IO, Unit] = {
+      def replace(in: InputStream): IO[InputStream] =
+        for {
+          std <- IO(System.in)
+          _ <- IO.blocking(System.setIn(in))
```
djspiewak: This can just be `apply`

core/jvm/src/test/scala/cats/effect/std/ConsoleJVMSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+          Thread
+            .currentThread()
+            .getContextClassLoader()
+            .getResourceAsStream(s"readline-test.$name.txt")
+        }
+      val release = (in: InputStream) => IO.blocking(in.close())
+      Resource.make(acquire)(release)
+    }
+
+    def replaceStandardIn(in: InputStream): Resource[IO, Unit] = {
```
djspiewak, line 54: This is a nifty utility, honestly

core/jvm/src/test/scala/cats/effect/std/ConsoleJVMSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+      Resource.make(replace(in))(restore).void
+    }
+
+    val test = for {
+      in <- inputStream(name)
+      _ <- replaceStandardIn(in)
+    } yield ()
+
+    def loop(acc: List[String]): IO[String] =
+      Console[IO].readLineWithCharset(charset).flatMap {
```
djspiewak: Random bikeshedding: `readLine` or `readln`? Bad idea?

core/jvm/src/test/scala/cats/effect/std/ConsoleJVMSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+      }
+
+    test.use(_ => loop(Nil))
+  }
+
+  private def readLineTest(name: String, charset: Charset): IO[MatchResult[String]] =
+    for {
+      contents <- fileContents(name, charset)
+      lines <- readLines(name, charset)
+    } yield contents must beEqualTo(lines)
```
djspiewak: FYI `must` is side-effecting, so this should be rearranged slightly:

```suggestion
      lines <- readLines(name, charset)
      result <- IO(contents mustEqual lines)
    } yield result
```

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -414,14 +414,15 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
...
+   * that the end of the standard input stream has been reached.
    *
    * @see `cats.effect.std.Console#readLineWithCharset` for reading using a custom
    * `java.nio.charset.Charset`
    *
-   * @return an IO effect that describes reading the user's input from the standard input as a string
+   * @return an IO effect that describes reading the user's input from the standard input as an optional string
    */
-  def readLine: IO[String] =
+  def readLine: IO[Option[String]] =
```
djspiewak: Is the `Option` *really* necessary? It's worth noting that the *actual* standard input stream never ends; so in other words this is always going to be `Some` unless someone explicitly does `setIn` with a terminating stream (as the tests do). I'm trying to think about some sort of non-ugly solution to this…

std/shared/src/main/scala/cats/effect/std/Console.scala
```diff
@@ -103,10 +105,10 @@ final class Console[F[_]] private (val F: Sync[F]) extends AnyVal {
       }
 
       @tailrec
-      def loop(): String = {
+      def loop(): Option[String] = {
         val buffer = decodeNext()
         if (buffer == null) {
-          return builder.toString()
+          return Some(builder.toString()).filter(_.nonEmpty)
```
djspiewak: Can we restructure this so we don't need the `return`?

djspiewak (changes requested):

core/jvm/src/test/scala/cats/effect/std/ConsoleJVMSpec.scala
vasilmkd: Agreed. Will fix.

core/jvm/src/test/scala/cats/effect/std/ConsoleJVMSpec.scala
vasilmkd: Totally agree, thanks for looking out for this.

core/jvm/src/test/scala/cats/effect/std/ConsoleJVMSpec.scala
vasilmkd: `readLine` comes from Java/Scala standard libraries. I'm ok either way.

core/jvm/src/test/scala/cats/effect/std/ConsoleJVMSpec.scala
vasilmkd: Will fix, I always forget about this.

core/shared/src/main/scala/cats/effect/IO.scala
vasilmkd: Have to think of a way to signal EOF in tests.

core/jvm/src/test/scala/cats/effect/std/ConsoleJVMSpec.scala
djspiewak: I'm okay too, and I think `readLine` is quite clear. Also it matches `BufferedReader`, so that kind of seals it in my mind.

core/shared/src/main/scala/cats/effect/IO.scala
djspiewak: Seriously, maybe we should use an exception. It's really a very rare case that should only happen in tests.

core/jvm/src/test/scala/cats/effect/std/ConsoleJVMSpec.scala
vasilmkd: Fixed.

core/jvm/src/test/scala/cats/effect/std/ConsoleJVMSpec.scala
vasilmkd: Fixed.

core/jvm/src/test/scala/cats/effect/std/ConsoleJVMSpec.scala
vasilmkd: Fixed.

core/jvm/src/test/scala/cats/effect/std/ConsoleJVMSpec.scala
vasilmkd: Fixed.

core/jvm/src/test/scala/cats/effect/std/ConsoleJVMSpec.scala
vasilmkd: That was exactly my reasoning.

core/shared/src/main/scala/cats/effect/IO.scala
vasilmkd: Done.

std/shared/src/main/scala/cats/effect/std/Console.scala
vasilmkd: Done.

core/jvm/src/test/scala/cats/effect/std/ConsoleJVMSpec.scala
vasilmkd: Done.

PR #1283, opened 2020-10-08
Add ApplicativeThrow alias

fthomas:
This adds `ApplicativeThrow` which is an alias for `ApplicativeError[F, Throwable]`. It is not needed in this code base but a convenience for users of the library so that they don't have to define it themselves (like here https://github.com/scala-steward-org/scala-steward/pull/1646/files#diff-70c33485bfc9d88a649c21513414f088R28).


RaasAhsan (approved):

vasilmkd (approved):
I like it, it makes the library more immersive.

PR #1285, opened 2020-10-08
Remove call-site constraints from Resource methods

bplommer:
Removes typeclass requirements from `map`, `mapK`, `evalMap`, `pure`, `liftF`, `liftK`. Also removes second type parameter from `map`.

This makes a one-off change to the public API. There are extra primitive node types, but they are package-private - to interpret a `Resource` you first call `.preinterpret` to compile it down to the existing three types.

Of course, we could alternatively just add the `Pure`, `LiftF` and `MapK` nodes to the public API. `MapK` is the only one that isn't completely trivial to interpret. My main concern is whether this would be an exhaustive list or if we might want to add more in the future.

Supersedes #1280 if merged.

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
@@ -107,7 +107,7 @@ trait GenSpawn[F[_], E] extends MonadCancel[F, E] {
     }
 
   def background[A](fa: F[A]): Resource[F, F[Outcome[F, E, A]]] =
-    Resource.make(start(fa))(_.cancel)(this).map(_.join)(this)
+    Resource.make(start(fa))(_.cancel)(this).map(_.join)
 }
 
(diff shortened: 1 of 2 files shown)


djspiewak:
This is *fabulous*! `preinterpret` is such a good idea. I do have a quick question though: originally, `mapK` required a `Defer`, but `preinterpret` doesn't take one. Is that an oversight? Or is that simply something that is enabled by the fact that `preinterpret` is shallow (meaning that *callers* are responsible for ensuring stack-safety)?

bplommer:
> This is _fabulous_! `preinterpret` is such a good idea. I do have a quick question though: originally, `mapK` required a `Defer`, but `preinterpret` doesn't take one. Is that an oversight? Or is that simply something that is enabled by the fact that `preinterpret` is shallow (meaning that _callers_ are responsible for ensuring stack-safety)?

Yes - recursion only happens in the case of nested `MapK` nodes, and then it's tail recursion.

PR #1286, opened 2020-10-08
Replace tcp section with producer-consumer in tutorial

lrodero:
TCP echo section of tutorial is somehow large and confusing. And it deals a lot with fibers at a low-level (see use of `.start`) which is to be discouraged in CE. This PR replaces that section with a new one that explains concurrency in CE using the consumer-producer problem and concurrent queues. It is shorter and hopefully clearer than the previous one. It discourages using `.start` and `.join`, and brings examples of how to use `Semaphore`, `Ref`, `.shift` and `.uncancelable`. At the end of the section there are a couple of exercises that propose to implement a concurrent queue.

Other minor changes: CE version is upgraded to `2.2.0` and some links in the first section are introduced/fixed.

New version of tutorial can be read [here](https://lrodero.github.io/cats-effect/tutorial/tutorial.html). As before code is in [this repo](https://github.com/lrodero/cats-effect-tutorial/tree/replace_tcp_section_with_prodcons), including proposed implementations for the concurrent queue exercise for reference.

### DIFF
--- a/site/src/main/mdoc/tutorial/tutorial.md
+++ b/site/src/main/mdoc/tutorial/tutorial.md
@@ -712,7 +712,6 @@ We can now create a program that instantiates our `queueR` and runs both
 import cats.effect._
 import cats.effect.concurrent.Ref
 import cats.syntax.all._
-
 import collection.immutable.Queue
 
 object InefficientProducerConsumer extends IOApp {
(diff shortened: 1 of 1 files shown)


djspiewak:
🎉 Thank you for this!!

Issue #1217, opened 2020-09-20
Add basic concurrency constructs (Queue etc)


Daenyth:
> The concept here is that there are a bunch of concurrent utilities that we could have which are very, very commonly needed, but which aren't available. They should all be implemented on top of the Concurrent typeclass and in the concurrent module. Queue is an obvious one, but there are definitely others.

https://gitter.im/typelevel/cats-effect-dev?at=5f6797fb89b38d09213dd022

Utilities to add:
- [x] Queue
- [x] Dequeue
- [x] PriorityQueue
- [x] CountDownLatch
- [x] CyclicBarrier
- [ ] ReentrantLock

RaasAhsan:
Thoughts on `Lock` and `ReentrantLock` types? The main criticism around `MVar` is that it's exceptionally easy to run into a deadlock, but I think we can provide safer variant of it in the library focused on mutual exclusion. Obviously there's still room for deadlocks, but those are usually scoped at a higher semantic level. Reentrancy will also require some notion of fiber identity.

I believe @ChristopherDavenport also has a couple microlibraries that expose useful concurrent types like [MapRef](https://github.com/ChristopherDavenport/mapref) and [KeySemaphore](https://github.com/ChristopherDavenport/keysemaphore). He's very vocal about handing these things off to somebody else, so the question is what do we want to bring in :)

djspiewak:
`ReentrantLock` is particularly tricky to get right, and there are certainly cases where it's helpful. I'd be in favor of including it, particularly since we have the ability to make it resource-safe. Also it may be interesting to see if we can devise a concept of fiber identity which is more narrowly-defined such that this can be done in a generic way.

Avasil:
👍 for including some of Chris's microlibraries

From other suggestions, I would love to see `PriorityQueue` !

Daenyth:
@RaasAhsan I'm not super familiar with those apis. When I was skimming over the java.concurrent package before, I didn't think those were priorities due to being able to use a Semaphore with size 1 as a cheap lock. Do you think there's a good level of value adding those anyway? Or did I miss a more nuanced api design than it just being `lock`/`unlock` pairs?

RaasAhsan:
[...]

PR #1294, opened 2020-10-10
Queue

vasilmkd:
Part of #1217.

WIP. Opening for a sneak peek and early comments on the api direction.

The bounded variant is currently functioning properly and hopefully well tested. I have taken some inspiration from the tests in `monix-catnap` and I have added a proper attribution in the source code.

Nothing here should be considered final and everything is up for change, especially the naming.

This is my first stab at a functional data structure so please take the time to review it properly, there may be bugs or generally things I have misunderstood.

This still needs partially applied and `in` syntax.

Let me know what you think.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Queue.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Queue.scala
@@ -1,0 +1,331 @@
+/*
+ * Copyright 2020 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.std
+
+import cats.~>
+import cats.effect.kernel.{Concurrent, Deferred, Poll, Ref}
+import cats.effect.kernel.syntax.all._
+import cats.syntax.all._
+
+import scala.collection.immutable.{Queue => ScalaQueue}
+
+/**
+ * A purely functional, concurrent data structure which allows insertion and
+ * retrieval of elements of type `A` in a first-in-first-out (FIFO) manner.
+ *
+ * Depending on the type of queue constructed, the [[Queue#offer]] operation can
+ * block semantically until sufficient capacity in the queue becomes available.
+ *
+ * The [[Queue#take]] operation semantically blocks when the queue is empty.
+ *
+ * The [[Queue#tryOffer]] and [[Queue#tryTake]] allow for usecases which want to
+ * avoid semantically blocking a fiber.
+ */
+abstract class Queue[F[_], A] { self =>
+
+  /**
+   * Enqueues the given element at the back of the queue, possibly semantically
+   * blocking until sufficient capacity becomes available.
+   *
+   * @param a the element to be put at the back of the queue
+   */
+  def offer(a: A): F[Unit]
+
+  /**
+   * Attempts to enqueue the given element at the back of the queue without
+   * semantically blocking.
+   *
+   * @param a the element to be put at the back of the queue
+   * @return an effect that describes whether the enqueuing of the given
+   *         element succeeded without blocking
+   */
+  def tryOffer(a: A): F[Boolean]
+
+  /**
+   * Dequeues an element from the front of the queue, possibly semantically
+   * blocking until an element becomes available.
+   */
+  def take: F[A]
+
+  /**
+   * Attempts to dequeue an element from the front of the queue, if one is
+   * available without semantically blocking.
+   *
+   * @return an effect that describes whether the dequeueing of an element from
+   *         the queue succeeded without blocking, with `None` denoting that no
+   *         element was available
+   */
+  def tryTake: F[Option[A]]
+
+  /**
+   * Modifies the context in which this queue is executed using the natural
+   * transformation `f`.
+   *
+   * @return a queue in the new context obtained by mapping the current one
+   *         using `f`
+   */
+  def mapK[G[_]](f: F ~> G): Queue[G, A] =
+    new Queue[G, A] {
+      def offer(a: A): G[Unit] = f(self.offer(a))
+      def tryOffer(a: A): G[Boolean] = f(self.tryOffer(a))
+      val take: G[A] = f(self.take)
+      val tryTake: G[Option[A]] = f(self.tryTake)
+    }
+}
+
+object Queue {
+
+  /**
+   * Constructs an empty, bounded queue holding up to `capacity` elements for
+   * `F` data types that are [[Concurrent]]. When the queue is full (contains
+   * exactly `capacity` elements), every next [[Queue#offer]] will be
+   * backpressured (i.e. the [[Queue#offer]] blocks semantically).
+   *
+   * @param capacity the maximum capacity of the queue
+   * @return an empty, bounded queue
+   */
+  def bounded[F[_], A](capacity: Int)(implicit F: Concurrent[F]): F[Queue[F, A]] = {
+    assertNonNegative(capacity)
+    F.ref(State.empty[F, A]).map(new BoundedQueue(capacity, _))
+  }
+
+  /**
+   * Constructs a queue through which a single element can pass only in the case
+   * when there are at least one taking fiber and at least one offering fiber
+   * for `F` data types that are [[Concurrent]]. Both [[Queue#offer]] and
+   * [[Queue#take]] semantically block until there is a fiber executing the
+   * opposite action, at which point both fibers are freed.
+   *
+   * @return a synchronous queue
+   */
+  def synchronous[F[_], A](implicit F: Concurrent[F]): F[Queue[F, A]] =
+    bounded(0)
+
+  /**
+   * Constructs an empty, unbounded queue for `F` data types that are
+   * [[Concurrent]]. [[Queue#offer]] never blocks semantically, as there is
+   * always spare capacity in the queue.
+   *
+   * @return an empty, unbounded queue
+   */
+  def unbounded[F[_], A](implicit F: Concurrent[F]): F[Queue[F, A]] =
+    bounded(Int.MaxValue)
+
+  /**
+   * Constructs an empty, bounded, dropping queue holding up to `capacity`
+   * elements for `F` data types that are [[Concurrent]]. When the queue is full
+   * (contains exactly `capacity` elements), every next [[Queue#offer]] will be
+   * ignored, i.e. no other elements can be enqueued until there is sufficient
+   * capacity in the queue, and the offer effect itself will not semantically
+   * block.
+   *
+   * @param capacity the maximum capacity of the queue
+   * @return an empty, bounded, dropping queue
+   */
+  def dropping[F[_], A](capacity: Int)(implicit F: Concurrent[F]): F[Queue[F, A]] = {
+    assertPositive(capacity, "Dropping")
+    F.ref(State.empty[F, A]).map(new DroppingQueue(capacity, _))
+  }
+
+  /**
+   * Constructs an empty, bounded, circular buffer queue holding up to
+   * `capacity` elements for `F` data types that are [[Concurrent]]. The queue
+   * always keeps at most `capacity` number of elements, with the oldest
+   * element in the queue always being dropped in favor of a new elements
+   * arriving in the queue, and the offer effect itself will not semantically
+   * block.
+   *
+   * @param capacity the maximum capacity of the queue
+   * @return an empty, bounded, sliding queue
+   */
+  def circularBuffer[F[_], A](capacity: Int)(implicit F: Concurrent[F]): F[Queue[F, A]] = {
+    assertPositive(capacity, "CircularBuffer")
+    F.ref(State.empty[F, A]).map(new CircularBufferQueue(capacity, _))
+  }
+
+  private def assertNonNegative(capacity: Int): Unit =
+    require(capacity >= 0, s"Bounded queue capacity must be non-negative, was: $capacity")
+
+  private def assertPositive(capacity: Int, name: String): Unit =
+    require(capacity > 0, s"$name queue capacity must be positive, was: $capacity")
+
+  private sealed abstract class AbstractQueue[F[_], A](
+      capacity: Int,
+      state: Ref[F, State[F, A]]
+  )(implicit F: Concurrent[F])
+      extends Queue[F, A] {
+
+    protected def onOfferNoCapacity(
+        s: State[F, A],
+        a: A,
+        offerer: Deferred[F, Unit],
+        poll: Poll[F]
+    ): (State[F, A], F[Unit])
+
+    protected def onTryOfferNoCapacity(s: State[F, A], a: A): (State[F, A], F[Boolean])
+
+    def offer(a: A): F[Unit] =
+      F.deferred[Unit].flatMap { offerer =>
+        F.uncancelable { poll =>
+          state.modify {
+            case State(queue, size, takers, offerers) if takers.nonEmpty =>
+              val (taker, rest) = takers.dequeue
+              State(queue, size, rest, offerers) -> taker.complete(a).void
+
+            case State(queue, size, takers, offerers) if size < capacity =>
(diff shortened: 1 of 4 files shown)


SystemFw:
I need to take a deeper look, although it looks ok at a quick glance.
A couple of points:
- I don't think you need any of those `uncancelable { poll` blocks in this implementation. They seem to be no-ops to me.
- There is an interesting question on whether to implement bounded directly, or just by layering semaphore on top of an unbounded queue. The latter is not really possible to do safely with ce2, but you can in ce3 (and in that case you do need `uncancelable/poll`.

vasilmkd:
@SystemFw The `uncancelable` blocks aren't noops. We need to make sure that we don't get canceled when we are freeing another taker or offerer, and that can happen after the modify, but before the resulting effect is evaluated. This is then ensured by `.flatMap(poll(_))`, which preserves the cancelability of the effects but masks the bind. The effects that need to be `uncancelable` are also marked as such inside the `modify` blocks.

vasilmkd:
Regarding the direct implementation vs using a semaphore, I'm ok either way, whatever we decide on. I was planning on abstracting away some of the common logic when I add more strategies, so that's why I chose this route.

SystemFw:
> We need to make sure that we don't get canceled when we are freeing another taker or offerer, and that can happen after the modify, but before the resulting effect is evaluated

ah, that's true. It's still solvable by moving `cleanup` to encompass the whole effect (which is what I'm so used to in ce2), but I like this way better.

> The effects that need to be uncancelable are also marked as such inside the modify blocks.

if it works, I think it's prettier to put the `poll`  around the interruptible parts like `get`, instead of doing `uncancelable { poll( uncancelable) }`, if you know what I mean.

> I was planning on abstracting away some of the common logic when I add more strategies, so that's why I chose this route.

fair enough, let's revisit once you add those

vasilmkd:
> > The effects that need to be uncancelable are also marked as such inside the modify blocks.
> 
> if it works, I think it's prettier to put the `poll` around the interruptible parts like `get`, instead of doing `uncancelable { poll( uncancelable) }`, if you know what I mean.

I understand what you mean. However I don't think what you mentioned is exactly what's going on. In fact, the innermost `uncancelable` in your example is just a description, the effect is not actually sequenced there but only after the whole `modify` effect is done.

I'll try to illustrate as best as I can.
```
uncancelable { poll(modify) >>= (effectComputedInsideModify => poll(effectComputedInsideModify)) }
```
where `effectComputedInsideModify` may or may not be cancelable, depending on the current state (e.g. lifting a pure value can be canceled (or not executed at all in case of cancelation), but a `deferred.complete` must become uncancelable as it is core cancelation safety logic, with a third option being blocking on a deffered, waiting to be notified, which obviously must remain cancelable). What a sentence...

vasilmkd:
@SystemFw are you against this for example?

```scala
def unbounded[F[_], A](implicit F: Concurrent[F]): F[Queue[F, A]] = bounded(Int.MaxValue)
```

SystemFw:
>  However I don't think what you mentioned is exactly what's going on.

I'll post an example next to the code, I think it's easier to follow that way.

> are you against this for example?

No, that's fine. I think the other way simplifies the implementation of Queue, but if you're going to embed more strategies there it does make sense, so let's keep it as is

std/shared/src/main/scala/cats/effect/std/BoundedQueue.scala
```diff
@@ -0,0 +1,131 @@
...
+
+          case State(queue, size, takers, offerers) =>
+            val cleanup = state.update { s =>
+              s.copy(offerers = s.offerers.filter(_._2 ne offerer))
+            }
+            State(queue, size, takers, offerers.enqueue(a -> offerer)) ->
+              offerer.get.onCancel(cleanup)
+        }).flatMap(poll(_))
+      }
+    }
```
SystemFw: What I mean is 
```scala
F.uncancelable { poll => 
  F.deferred[Unit].flatMap { offerer =>
        state.modify {
            case ... =>
              ...
              State(...) -> taker.complete(a).void
 
            case ... =>
              State(...) -> F.unit
 
            case ... =>
              val cleanup = ...
              State(...) ->  poll(offerer.get).onCancel(cleanup)
         }.flatten
        }
     }

Comment (SystemFw) [reply std/shared/src/main/scala/cats/effect/std/BoundedQueue.scala]: Also note that `Ref.modify` cannot be canceled, so the `poll` there (in the original implementation) is not needed

Comment (vasilmkd) [reply std/shared/src/main/scala/cats/effect/std/BoundedQueue.scala]: The reason I had `uncancelable` come after the creation of the deferred is because that would allow for cancelation before anything is constructed, which I think is neat.

I didn't know about `modify`.

Comment (SystemFw) [reply std/shared/src/main/scala/cats/effect/std/BoundedQueue.scala]: > The reason I had uncancelable come after the creation of the deferred 

oh yeah that's fine

**Comment by vasilmkd** [comment]
Implementation and tests are done. I still need to add `in` syntax.

File: std/shared/src/main/scala/cats/effect/std/BoundedQueue.scala (hunk 1 above)
Comment (vasilmkd) [reply std/shared/src/main/scala/cats/effect/std/BoundedQueue.scala]: I have reinstated `poll` around `modify` as that would allow cancelation before `modify` is attempted, which I think is useful.

I have also kept `uncancelable` on the described effect in place because I think it helps reasoning, "we are describing an uncancelable effect which will be sequenced later".

The sequencing then does so again with a `poll`, which just means execute the next next effect with its preferred cancelation mode.

**Comment by bplommer** [comment]
It really bothers me that `Queue` and `Semaphore` require an effect with a `Throwable` channel just so they can validate arguments. I see a couple of ways to avoid this:

1. Have two versions, one for `Concurrent[F]` that raises an error on invalid arguments and one for `GenConcurrent[F, _]` that clamps the arguments to legal values (the second could have a constructor called something like `Queue.unchecked`)

2. Add an `Assert` typeclass that lets you make assertions without knowing the error type. A default instance would obviously be provided for `ApplicativeThrow`. (this seems more in the scope of cats than cats-effect).

File: std/shared/src/main/scala/cats/effect/std/BoundedQueue.scala (hunk 1 above)
Comment (SystemFw) [reply std/shared/src/main/scala/cats/effect/std/BoundedQueue.scala]: I find it harder to follow for no meaningful reason, tbh. There are now several levels of nesting of cancelability for the purpose of getting one more short lived operation to cancel in a very lucky interleaving (if at all)

Comment (vasilmkd) [reply std/shared/src/main/scala/cats/effect/std/BoundedQueue.scala]: I'll rework it. 😄

**Comment by djspiewak** [comment]
> It really bothers me that Queue and Semaphore require an effect with a Throwable channel just so they can validate arguments. I see a couple of ways to avoid this:

Another approach might be to validate arguments eagerly and `throw` as a side-effect, rather than sequencing the errors. In a sense, argument errors of this nature are more like assertion failures than runtime errors.

**Comment by bplommer** [comment]
> Another approach might be to validate arguments eagerly and `throw` as a side-effect, rather than sequencing the errors. In a sense, argument errors of this nature are more like assertion failures than runtime errors.

I don't think I'm comfortable with this. This isn't really an assertion failure from our point of view - only receiving valid arguments isn't an invariant we can control. And if we throw on invalid arguments, that means downstream library code will also do so unless it also validates the arguments - so our validation is either redundant, or breaks RT for anyone that depends directly or indirectly on us. (And where the dependency is indirect, the cause of invalid arguments could be a non-obvious logic error rather than a blatant mistake.)

I feel like we should really be using refined types for this - obviously we can't bring in `refined` as a dependency, but perhaps we could implement a minimal subset of its functionality for these cases.

**Comment by SystemFw** [comment]
> Another approach might be to validate arguments eagerly and throw as a side-effect, rather than sequencing the errors

This is the way it's worked in fs2 for years and no one has ever complained. I'm not a fan of complicated type level validation for arguments like this, which are static in the overwhelming majority of cases

**Comment by bplommer** [comment]
> This is the way it's worked in fs2 for years and no one has ever complained.

They have now 😛 but I'm happy to defer to fs2's precedent, given that any non-obvious possibility of failure would involve a fairly advanced use-case.

**Comment by vasilmkd** [comment]
Sorry for the hold up on this. I'm a bit swamped at the moment. I also want to take the time to experiment with a different queue collection implementation.

File: std/shared/src/main/scala/cats/effect/std/Queue.scala [hunk 2]
```diff
@@ -0,0 +1,333 @@
...
+            case State(queue, size, takers, offerers) if takers.nonEmpty =>
+              val (taker, rest) = takers.dequeue
+              State(queue, size, rest, offerers) -> taker.complete(a).uncancelable.void
+
+            case State(queue, size, takers, offerers) if size < capacity =>
+              State(queue.enqueue(a), size + 1, takers, offerers) -> F.unit
+
+            case s =>
+              onOfferNoCapacity(s, a, offerer)
+          }).flatMap(poll(_))
```
RaasAhsan: For these kinds of operations, I find it nice to call `flatten` here. We still need to preserve masking semantics, so we have to "invert" and specify the masks rather than the unmasks, which I also much prefer:
1. `poll` the specific effects that actually want to be unmasked
2. We don't need to explicitly mask the inner effects that want it anymore, since it will ultimately run in the scope of the `uncancelable`.

That avoids multiple nested `uncancelable` regions which can be tricky to follow sometimes.

std/shared/src/main/scala/cats/effect/std/Queue.scala
```diff
@@ -0,0 +1,333 @@
...
+        a: A,
+        offerer: Deferred[F, Unit]
+    ): (State[F, A], F[Unit])
+
+    protected def onTryOfferNoCapacity(s: State[F, A], a: A): (State[F, A], F[Boolean])
+
+    def offer(a: A): F[Unit] =
+      F.deferred[Unit].flatMap { offerer =>
+        F.uncancelable { poll =>
+          poll(state.modify {
```
RaasAhsan: On one hand, these outer `poll`s seem pretty unnecessary to me because `Ref#modify` is implemented as a single `delay`, and so the only time we could observe cancellation is right before it executes. 

On the other hand, the implementation is just that: an implementation. If `modify` were implemented as a sequence of binds, you could observe cancellation between effects. But for tight loops, it seems rather inconsequential if it's not there ,and there are some examples like `memoize` that don't bother polling the CAS op anyways (though we can have a discussion about whether that is right or wrong).

I'm generally in favor of removing these

std/shared/src/main/scala/cats/effect/std/Queue.scala
SystemFw: See comments here :) https://github.com/typelevel/cats-effect/pull/1294#discussion_r502839184

std/shared/src/main/scala/cats/effect/std/Queue.scala
SystemFw: I think we could make `Ref.modify` uncancelable by law btw

std/shared/src/main/scala/cats/effect/std/Queue.scala
RaasAhsan: I should look at all the outdated review comments :) Glad to hear we are in agreement though

std/shared/src/main/scala/cats/effect/std/Queue.scala
vasilmkd: I just wanted to write the most correct code possible. I didn't bother looking into each implementation. I will remove this though.

std/shared/src/main/scala/cats/effect/std/Queue.scala
vasilmkd: Yup, I will change this. Thanks for the comments both @SystemFw and @RaasAhsan 😄

vasilmkd:
In order to keep the scope of this PR somewhat under control, I will open it for review here. All 4 types of queues are fully functional and fairly well tested. Syntactic sugar can be added as follow up work. Regarding https://github.com/typelevel/fs2/issues/1551, this queue implementation doesn't seem to suffer from the same performance bottleneck. Further performance exploration can be done separately.

std/shared/src/main/scala/cats/effect/std/Queue.scala
```diff
@@ -0,0 +1,331 @@
...
+  /**
+   * Constructs a queue through which a single element can pass only in the case
+   * when there are at least one taking fiber and at least one offering fiber
+   * for `F` data types that are [[Concurrent]]. Both [[Queue#offer]] and
+   * [[Queue#take]] semantically block until there is a fiber executing the
+   * opposite action, at which point both fibers are freed.
+   *
+   * @return a synchronous queue
+   */
+  def synchronous[F[_], A](implicit F: Concurrent[F]): F[Queue[F, A]] =
```
djspiewak, line 115: What use-case does this address with cats effect specifically? I get why it's useful in fs2, but `F` isn't a stream.

std/shared/src/main/scala/cats/effect/std/Queue.scala
```diff
@@ -0,0 +1,331 @@
...
+        offerer: Deferred[F, Unit],
+        poll: Poll[F]
+    ): (State[F, A], F[Unit]) =
+      s -> F.unit
+
+    protected def onTryOfferNoCapacity(s: State[F, A], a: A): (State[F, A], F[Boolean]) =
+      s -> F.pure(false)
+  }
+
+  private final class CircularBufferQueue[F[_], A](capacity: Int, state: Ref[F, State[F, A]])(
```
djspiewak, line 298: I think it would be worth a separate implementation of (not backed by `ScalaQueue`) for this case, given that a) it can be specialized with vastly higher performance, and b) nearly all use-cases for circular buffers are sensitive to such performance concerns.

std/shared/src/main/scala/cats/effect/std/Queue.scala
```diff
@@ -0,0 +1,331 @@
...
+
+    protected def onTryOfferNoCapacity(s: State[F, A], a: A): (State[F, A], F[Boolean]) = {
+      val State(queue, size, takers, offerers) = s
+      val (_, rest) = queue.dequeue
+      State(rest.enqueue(a), size, takers, offerers) -> F.pure(true)
+    }
+  }
+
+  private final case class State[F[_], A](
+      queue: ScalaQueue[A],
```
djspiewak, line 321: Something we should think about from a performance standpoint: it might be worth inlining the bankers queue into this `State`, rather than delegating to `ScalaQueue`.

djspiewak (commented):
I definitely think we should continue iterating on performance things. We're obviously limited by `Concurrent`, but I think even within those bounds we should be able to make good things happen. At any rate, this is a very nice start.

std/shared/src/main/scala/cats/effect/std/Queue.scala
djspiewak: Worth noting that this is *exactly* equivalent to `MVar`, which we deleted. :-D

vasilmkd: So delete?

std/shared/src/main/scala/cats/effect/std/Queue.scala
vasilmkd: Yeah, I wanted to take a look at it. Is this now blocked on that, or can it be a part of a separate PR?

djspiewak: Definitely a separate PR

std/shared/src/main/scala/cats/effect/std/Queue.scala
vasilmkd: Also, not exactly equivalent. `MVar` is a bounded queue of size 1.

djspiewak: Isn't that what this is?

std/shared/src/main/scala/cats/effect/std/BoundedQueue.scala
vasilmkd: Done.

std/shared/src/main/scala/cats/effect/std/Queue.scala
vasilmkd: Done.

std/shared/src/main/scala/cats/effect/std/Queue.scala
vasilmkd: Done.

std/shared/src/main/scala/cats/effect/std/Queue.scala
vasilmkd: Will address in a separate PR.

std/shared/src/main/scala/cats/effect/std/Queue.scala
vasilmkd: I'd describe this as a handoff point, there is no progress being made (semantic blocking) until there's one offering fiber and one taking fiber, synchronously passing a value between them.

PR #1295, opened 2020-10-11
Reduce API surface

TimWSpence:


vasilmkd (approved):

djspiewak:
Oh derp I totally missed commenting on that. I meant to privatize it; thanks for the follow up!

TimWSpence:
> Oh derp I totally missed commenting on that. I meant to privatize it; thanks for the follow up!

Definitely my fault!! :)

PR #1300, opened 2020-10-11
Weaken typeclass constraints in Resource and Semaphore

bplommer:


kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -599,13 +599,14 @@ object Resource extends ResourceInstances with ResourcePlatform {
   }
 
   trait Bracket0 {
-    implicit def catsEffectResourceBracketForSync[F[_]](implicit F0: Sync[F]): Bracket[F] =
-      new SyncBracket[F] {
-        implicit protected def F: Sync[F] = F0
+    implicit def catsEffectResourceBracketForMonadThrow[F[_]](
+        implicit F0: MonadThrow[F]): Bracket[F] =
+      new MonadThrowBracket[F] {
+        implicit protected def F: MonadThrow[F] = F0
```
djspiewak, line 605: I think the purpose behind this was originally to ensure that it was completely unambiguous with `MonadCancel`. So the over-specificity is actually intentional.

bplommer: Doesn't the low-priority mix-in take care of this? And if not, wouldn't this problem remain if you have both `MonadCancel` and `Sync` available? I guess the answer is somehow no, but I think we should document that in a comment 🙂

djspiewak:
@SystemFw Any thoughts on this?

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
SystemFw: I'm open to changing this, but iirc the broad point was "include `SyncIO`, but not anything weaker", as well as avoiding any possibility whatsoever that a `MonadCancel` would resolve to the `MonadThrow`  instance

SystemFw:
Commented in the discussion. It was intentional but I'm open to counterarguments

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
bplommer: How about the class taking `MonadThrow` but the implicit taking `Sync`? Then it's not unduly difficult to manually instantiate for any potential advanced use-cases that might need it, but also not unduly easy to access it for novice misuse-cases.

djspiewak: @bplommer I like your suggestion. Leaves open the door for greater flexibility, while still steering people into a safer path.

bplommer:
Closing as outdated

PR #1303, opened 2020-10-12
First cut at Dispatcher

### LINKED ISSUE #1299: Implement an async dispatcher
The goal here is to solve the "soviet interop" case (where unsafe code runs you) without requiring `UnsafeRun`. In theory, this is doable with solely an `Async` constraint whenever we have the following kind of situation:

```
+-------------------+
| F[_]              |
|                   |
|  +-------------+  |
|  | Unsafe code |  |
|  |             |  |
|  |  +------+   |  |
|  |  | F[_] |   |  |
|  |  |      |   |  |
|  |  +------+   |  |
|  +-------------+  |
+-------------------+
```

So in other words, you have some effect context, `F[_]: Async`, which sets up some FFI region which needs the ability to unsafely run arbitrary other `F[A]` actions. A good example of this would be a hypothetical `cats-effect-netty` layer, where you would have some outer `F[_]` which instantiates the Netty pipelines and such, within which Netty would need the ability to convert arbitrary `F[ByteBuffer]` values into netty futures (i.e. unsafe running), and those `F[ByteBuffer]` values are themselves effects.

I've been calling this "the soviet enclave special-case". In theory, this should handle *all* "soviet interop" cases except for those which are truly at the end of the world, lacking an "outer" enclosing `F[_]`. For things that are actually the literal end of the world, I'm pretty comfortable just not abstracting, since libraries have no need for this kind of use-case and it only impacts users (who are in control of their specific effect type, and thus can use the appropriate runtime directly).

A quick PoC: https://gist.github.com/djspiewak/f761d0a804b5bb789abe834566239741

djspiewak:
The idea here is actually pretty obvious in retrospect: a non-blocking supervised fiber-based dispatcher for sequencing effects. State is being managed by a pair of `AtomicReference`s: one of which contains an optional callback (and also encodes a double-check state to resolve races), while the other contains a `LongMap` with the registered effects. The `dispatcher` fiber loops continuously, checking the state for new entries. If none are available, it creates an `async` node and registers the callback in the `latch` reference, double-checking the state to ensure that nothing was written to the `state` reference in the interim.

If any new work is encountered, the dispatcher creates a fiber for each registered effect (which must be of type `F[Unit]`) and places that fiber into a `Ref` which is folded into the outer `Resource` scope, ensuring that child fibers are appropriately cleaned up when the dispatcher is shut down (note: here's a good and unavoidable use of the `start` function). Cancelation is carefully checked here.

Note that any use of the `Runner` post-shutdown will result in deadlocks and memory leaks. I don't check for this at present.

Anyway, on the calling side, all effects of type `F[E]` (for some `E`) are wrapped in the appropriate machinery to populate a `Promise` upon completion. Additionally, a bit of extra machinery is constructed to ensure that the state is unregistered if the cancelation function is called, or if the effect has already been started, that cancelation is then passed along to the carrier fiber.

We don't actually need to fully encode a queue or anything like it since we can just spawn as many fibers as we want. When I realized this, I was able to greatly simplify the state machine. In abstract, it's really just an async latch which can be fired from impure code, together with a list of `F[Unit]` actions with callbacks to tie up cancelation.

I want to add a few more things to the respective `RunnerPlatform`s, but this is pretty much ready for review.

Resolves #1299


std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -0,0 +1,163 @@
...
+import scala.annotation.tailrec
+import scala.collection.immutable.LongMap
+import scala.concurrent.{Future, Promise}
+
+import java.util.concurrent.atomic.AtomicReference
+
+object Dispatcher extends DispatcherPlatform {
+
+  def apply[F[_]: Async, A](unsafe: Runner[F] => F[A]): Resource[F, A] = {
+    final case class State(begin: Long, end: Long, registry: LongMap[(F[Unit], F[Unit] => Unit)]) {
```
RaasAhsan: worth defining a `case class` for `(F[Unit], F[Unit] => Unit)`? I find these state machines much easier to grok with more names, and there are no extra allocations :)

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -0,0 +1,163 @@
...
+      // efficiency on the CAS
+      override def equals(that: Any) = this eq that.asInstanceOf[AnyRef]
```
RaasAhsan, line 58: neat trick, we're basically trading the structural equality of `case class` in return for referential equality! i think it's worth a more detailed comment

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -0,0 +1,163 @@
...
+      override def hashCode = System.identityHashCode(this)
+    }
+
+    val Open = () => ()
+    val Empty = State(0, 0, LongMap())
```
RaasAhsan: should we preallocate these in the enclosing object scope? i could see it having better locality on the stack though

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -0,0 +1,163 @@
...
+
+import scala.annotation.tailrec
+import scala.collection.immutable.LongMap
+import scala.concurrent.{Future, Promise}
+
+import java.util.concurrent.atomic.AtomicReference
+
+object Dispatcher extends DispatcherPlatform {
+
+  def apply[F[_]: Async, A](unsafe: Runner[F] => F[A]): Resource[F, A] = {
```
RaasAhsan: What do you think about having a variant of this where the user specifies the dispatching process as a `Resource`? 
```scala
def apply[F[_]: Async, A](unsafe: Runner[F] => Resource[F, A]): Resource[F, A]

def apply2[F[_]: Async, A](unsafe: Runner[F] => F[A]): Resource[F, A] = 
  apply(runner => Resource.liftF(unsafe(runner)))
```
Users can then hook into the cancellation model and specify how to cancel the dispatching process. I think this would be applicable in scenarios where you want to terminate a source from submitting more effects (http server, reactive subscriber, etc.)

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -0,0 +1,163 @@
...
+
+    val Open = () => ()
+    val Empty = State(0, 0, LongMap())
+
+    for {
+      latch <- Resource.liftF(Sync[F].delay(new AtomicReference[() => Unit]))
+      state <- Resource.liftF(Sync[F].delay(new AtomicReference[State](Empty)))
+
+      active <- Resource.make(Ref[F].of(Set[Fiber[F, Throwable, Unit]]())) { ref =>
+        ref.get.flatMap(_.toList.traverse_(_.cancel))
```
RaasAhsan: Can we run all these cancellations in parallel?

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -0,0 +1,163 @@
...
+      }
+
+      dispatcher = for {
+        _ <- Sync[F].delay(latch.set(null))   // reset to null
+        s <- Sync[F].delay(state.getAndSet(Empty))
+
+        State(begin, end, registry) = s
+        pairs = (begin until end).toList.flatMap(registry.get)
```
RaasAhsan: What's the point of the `until` here? It seems like we can just pull everything out of the `registry`, so we wouldn't even need `begin` anymore.

Also, I think we want `end` to be globally unique up to the dispatcher instance; right now it resets to `0` after processing a batch of effects. Two effects in two separate batches can have the same ID; if we invoke the `Future` cancel token, there's a possibility we remove an effect from a _future_ batch.

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -0,0 +1,163 @@
...
+              cb(Right(()))
+            }
+          }
+        } else {
+          MonadCancel[F] uncancelable { _ =>
+            for {
+              fibers <- pairs traverse {
+                case (action, f) =>
+                  for {
+                    fiberDef <- Deferred[F, Fiber[F, Throwable, Unit]]
```
RaasAhsan: This is pretty cool :) Pending my above comment about IDs, I think it would be slightly simpler if we identify each fiber by the ID of its corresponding effect; the `active` set would be a `LongMap[Fiber]`.

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -0,0 +1,163 @@
...
+
+                    enriched = action guarantee {
+                      fiberDef.get.flatMap(fiber => active.update(_ - fiber))
+                    }
+
+                    fiber <- enriched.start
+                    _ <- fiberDef.complete(fiber)
+                    _ <- Sync[F].delay(f(fiber.cancel))
```
RaasAhsan: I think there's a slight race here that's easily solved with a bit more coordination:
1. Fiber is started.
2. `() => Future[Unit]` cancel token is invoked. It reads the `cancelToken` to be `null`, so it doesn't run the cancellation.
3. We set the `cancelToken` on this line.

RaasAhsan:
Regarding where this lives, I feel like if this is going to be the spiritual successor to `ConcurrentEffect` and `UnsafeRun`, it should just go into `kernel`. If libraries like http4s and fs2 are likely going to be pulling this in, I'm sure they (and downstream libraries + users) would be happy to avoid a dependency on `std`.

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
djspiewak: Yeah I think it does deserve more of an explanation than I gave it. :-D

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
djspiewak: I'll move them into the object. They're going to live on the heap anyway so it probably doesn't make that much of a difference where the pointers are.

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
djspiewak: This is a really good point. I also wonder about other cases, like where someone wants the nested bit to return a `Stream`.

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
djspiewak: Good catch. That's worth doing.

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
djspiewak: > What's the point of the until here? It seems like we can just pull everything out of the registry, so we wouldn't even need begin anymore.

Hmm, I guess we can just say `.values` now that I think about it.

> Two effects in two separate batches can have the same ID; if we invoke the Future cancel token, there's a possibility we remove an effect from a future batch

Ah! I didn't think about this. You're correct. I'll think about how to address that, since I'm just being lazy with cancelation right now.

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
djspiewak: You still have to resolve the race between spawning and completing the fiber though. `Deferred` is pretty great for that; we could technically add some other mechanism in there (it would need to be something like `LongMap[Option[Fiber]]` and then we CAS), but I'm not sure that's better and it certainly would be more complex.

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
djspiewak: Ah, good point! I think we need to double-check *after* we set the `cancelToken` to see if we've been canceled. We simply don't have enough state to do that right now.

djspiewak:
> Regarding where this lives, I feel like if this is going to be the spiritual successor to ConcurrentEffect and UnsafeRun, it should just go into kernel. If libraries like http4s and fs2 are likely going to be pulling this in, I'm sure they (and downstream libraries + users) would be happy to avoid a dependency on std.

Possibly, but I kind of see the modules in the following way:

- **kernel** – Dependency of datatypes (like Monix)
- **std** – Dependency of libraries (like http4s/fs2)
- **core** – Dependency of users and library test suites

So from that standpoint, it's okay to stay in std, since everyone who wants it would already have this dependency.

mpilquist:
This is awesome.

kubukoz:
I'll take a look tomorrow ☺️

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
RaasAhsan: Assuming you agree with my comments about per-dispatcher unique effect IDs, I don't think there is a race between spawning and completing the fiber anymore. The race in question happens because the fiber needs to have a reference to itself (since we're using referential identity to identify fibers here). But we already have something that in theory uniquely identifies the fiber; it's just that ID associated with the effect in the registry. We know that ID well before the fiber is spawned.

I'm not seeing why we would need a `LongMap[Option[Fiber]]` rather than just a `LongMap[Fiber]`. After we spawn the batch of fibers, we add a new set of keys. The `guarantee` action simply removes the entry with the `Long` as a key. We'll need to use a CAS for those operations, but we already do that anyways (`active.update`). Is it important to distinguish between a fiber completing versus it not being in the active set?

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -0,0 +1,168 @@
...
+                        fiberDef.get.flatMap(fiber => active.update(_ - fiber))
+                      }
+
+                      fiber <- enriched.start
+                      _ <- fiberDef.complete(fiber)
+                      _ <- Sync[F].delay(f(fiber.cancel))
+                    } yield fiber
+                }
+
+                _ <- active.update(_ ++ fibers)
```
RaasAhsan: I think there is a race here that can manifest as a memory leak. Basically, what happens if a spawned fiber completes before we add it to the active set here. The fiber will remove nothing from the active set, and then we add it in later but it would never be removed. I see two solutions here:
1. A fiber adds itself to the active set before running the action. This is another appealing reason to use the `Long` ids.
2. We force all the fibers to wait on a `Deferred` that completes only after this active set has updated.

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -0,0 +1,168 @@
...
+
+    val Open = () => ()
+    val Empty = State(0, 0, LongMap())
+
+    for {
+      latch <- Resource.liftF(Sync[F].delay(new AtomicReference[() => Unit]))
+      state <- Resource.liftF(Sync[F].delay(new AtomicReference[State](Empty)))
+
+      active <- Resource.make(Ref[F].of(Set[Fiber[F, Throwable, Unit]]())) { ref =>
+        ref.get.flatMap(_.toList.traverse_(_.cancel))
```
RaasAhsan: You touched on this in the PR description, but calling it out explicitly so we don't forget: this can leak fibers if new ones are dispatched after we fetch the current active set

core/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
```diff
@@ -0,0 +1,88 @@
...
+
+package cats.effect
+package std
+
+import cats.effect.kernel.Deferred
+import cats.syntax.all._
+
+import scala.concurrent.duration._
+
+class DispatcherSpec extends BaseSpec {
```
kubukoz, line 25: Having some examples of this being used with a callback-driven API would be nice (not necessarily in the form of tests)

kubukoz (commented):
ok, I have no idea what's going on there so I can only look at it from a user's perspective 😅

core/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
djspiewak: Agreed. I think this needs some more documentation. Probably just scaladoc alone is sufficient.

djspiewak:
@RaasAhsan The state machine is ready for a second review!

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -0,0 +1,215 @@
...
+        _ <- F.delay(latch.set(null)) // reset to null
+
+        s <- F delay {
+          @tailrec
+          def loop(): State = {
+            val s = state.get()
+            if (!state.compareAndSet(s, s.copy(registry = s.registry.empty)))
+              loop()
+            else
+              s
```
RaasAhsan, line 86: I believe we need to return the _new_ committed state here, otherwise we will continuously spawn all the effects we've ever queued up over and over again

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -0,0 +1,215 @@
...
+              if (!latch.compareAndSet(null, () => cb(Right(())))) {
+                // state was changed between when we last set the latch and now; complete the callback immediately
+                cb(Right(()))
+              }
+            }
+          } else {
+            F uncancelable { _ =>
+              for {
+                // for catching race conditions where we finished before we were in the map
+                completed <- F.ref(LongMap[Unit]())
```
RaasAhsan, line 106: My only gripe with this is that it can turn into a very slight memory leak. Basically, `completed` won't be garbage collected until the _last_ fiber in a batch terminates.  I may be splitting hairs over this though, so feel free to rightfully ignore this :D

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -0,0 +1,215 @@
...
+            var cancelToken: () => Future[Unit] = null
+
+            @volatile
+            var canceled = false
+
+            def registerCancel(token: F[Unit]): Unit = {
+              cancelToken = () => unsafeToFuture(token)
+
+              // double-check to resolve race condition here
+              if (canceled) {
```
RaasAhsan: I think there's a benign race condition here, but I'm going to mention it anyways:
1. Unsafe cancel sets `canceled` to `true`.
2. `registerCancel` sets the cancel token here.
3. Unsafe cancel reads the `cancelToken` and runs it.
4. `registerCancel` reads `canceled` to be true and runs the token.

The cancel token is pretty much just a `fiber.cancel`, so I don't think we're violating correctness or leaking references, but it may just be worth fixing :)

RaasAhsan (commented):
A couple more super tiny comments :) I'm going to look at this more later, but 👍 for merging after these are addressed

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
RaasAhsan: Just kidding! If we did that we would never run anything

djspiewak:
This is released as `3.0-80f5cc5`

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
djspiewak: I think this is okay. It's a relatively small amount of state, and the only way to avoid this is to do `Ref[Ref[LongMap[Unit]]`, but the outer `Ref` still wouldn't be collected.

RaasAhsan (approved):

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
djspiewak: Sorry took me a little while to scrape together the brain cells to understand this. :-) So I actually thought about that case when I was writing this! I didn't bother trying to resolve it because we already know `cancel` to be idempotent, so I think we can leave it.

PR #1305, opened 2020-10-12
Fix race ordering so we check the stored outcome first

TimWSpence:
Currently we blindly invoke the `cb` with the result of a fiber, without checking to see if the other side of the race has already invoked it.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -698,49 +698,67 @@ private final class IOFiber[A](
 
                 fiberA registerListener { oc =>
                   val s = state.getAndSet(Some(oc))
-                  oc match {
-                    case Outcome.Succeeded(Pure(a)) =>
-                      finalizer.set(fiberB.cancel)
-                      cb(Right(Left(a)))
-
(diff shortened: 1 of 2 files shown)


TimWSpence:
Sorry, I was experimenting with this locally and https://github.com/typelevel/cats-effect/pull/1157 got merged before I had pushed the change.

RaasAhsan:
> Currently we blindly invoke the cb with the result of a fiber, without checking to see if the other side of the race has already invoked it.

Isn't this OK ? async continuations have at-most once completion semantics

TimWSpence:
> > Currently we blindly invoke the cb with the result of a fiber, without checking to see if the other side of the race has already invoked it.
> 
> Isn't this OK ? async continuations have at-most once completion semantics

Well it certainly seems to be 😂 But morally I think this change is correct / what you would expect the behaviour to be - when you complete, check if the other side completed before you do anything.

djspiewak:
I actually kind a prefer the blind invocation. `cb` already has to track this kind of state, so this change actually makes it *less* efficient (since the state checks are only really needed within the `Canceled()` case, but the changes make them happen in all branches).

TimWSpence:
Thanks @RaasAhsan and @djspiewak. I'll close this then

PR #1315, opened 2020-10-16
Remove Resource.Bracket in favor of MonadCancel

mpilquist:
This also defines a "no-op" `MonadCancel[SyncIO, Throwable]` instance, allowing `Resource[SyncIO, X]` instances despite `SyncIO` not supporting cancelation.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/SyncIO.scala
+++ b/core/shared/src/main/scala/cats/effect/SyncIO.scala
@@ -534,10 +534,10 @@ object SyncIO extends SyncIOLowPriorityImplicits {
       def canceled: SyncIO[Unit] = unit
 
       def forceR[A, B](fa: SyncIO[A])(fb: SyncIO[B]): SyncIO[B] =
-        productR(fa)(fb)
+        productR(fa.attempt)(fb)
 
       def onCancel[A](fa: SyncIO[A], fin: SyncIO[Unit]): SyncIO[A] =
-        fa <* fin
+        fa
 
       def uncancelable[A](body: Poll[SyncIO] => SyncIO[A]): SyncIO[A] =
         body(new Poll[SyncIO] { def apply[X](fx: SyncIO[X]) = fx })
--- a/core/shared/src/main/scala/cats/effect/SyncIO.scala
+++ b/core/shared/src/main/scala/cats/effect/SyncIO.scala
@@ -497,42 +497,53 @@ object SyncIO extends SyncIOLowPriorityImplicits {
     def empty: SyncIO[A] = pure(A.empty)
   }
 
-  private[this] val _syncForSyncIO: Sync[SyncIO] = new Sync[SyncIO]
-    with StackSafeMonad[SyncIO] {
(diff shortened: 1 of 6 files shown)


SystemFw:
🔥

SystemFw:
I would definitely like to see `Resource.Bracket` go. I don't think the SyncIO instance should be no-op though, I think we can have  a working one

mpilquist:
What do you mean by a working one? `SyncIO` doesn't support cancelation, so the instance here should be real.

SystemFw:
I would find it surprising if laws passed with that instance. I think `SyncIO` can support (self) cancellation, and the addition of MonadCancel means we can say that without having to support Concurrency. I'd take Resource.Bracket over a non lawful instance of MonadCancel for this I think (at least it's explicit in what it's trying to achieve), but I think we can make it all work

mpilquist:
Oh right, forgot about self cancelation.

SystemFw:
This probably means we need to reimplement SyncIO, but it shouldn't be too hard

core/shared/src/main/scala/cats/effect/SyncIO.scala
```diff
@@ -497,42 +497,53 @@ object SyncIO extends SyncIOLowPriorityImplicits {
...
 
-    override def redeemWith[A, B](
-        fa: SyncIO[A])(recover: Throwable => SyncIO[B], bind: A => SyncIO[B]): SyncIO[B] =
-      fa.redeemWith(recover, bind)
-  }
+      override def redeemWith[A, B](
+          fa: SyncIO[A])(recover: Throwable => SyncIO[B], bind: A => SyncIO[B]): SyncIO[B] =
+        fa.redeemWith(recover, bind)
+
+      def canceled: SyncIO[Unit] = unit
```
kubukoz: Should this and `monotonic` etc. be vals?

core/shared/src/test/scala/cats/effect/Runners.scala
```diff
@@ -190,7 +190,7 @@ trait Runners extends SpecificationLike with RunnersPlatform { outer =>
    */
   implicit def eqResource[F[_], A](
       implicit E: Eq[F[A]],
-      F: Resource.Bracket[F]): Eq[Resource[F, A]] =
+      F: MonadCancel[F, Throwable]): Eq[Resource[F, A]] =
     new Eq[Resource[F, A]] {
```
kubukoz, line 207: `Eq.by(_.use(F.pure))` maybe?

vasilmkd:
I also merged series/3.x into this branch, because it was quite behind, including the CI configuration.

vasilmkd:
There's some log spam, coming from `t.printStackTrace()` which report failures from finalizers, but we can take care of it in a follow up. We should probably adjust the Generators as well, to include canceled SyncIOs. Or maybe that comes out of the box (MonadCancelGenerators)?

mpilquist:
Bump on getting this one in for M3

SystemFw (approved):

PR #1316, opened 2020-10-16
Semaphore: throw rather than sequence IllegalArgumentException to weaken constraints

bplommer:
As discussed in https://github.com/typelevel/cats-effect/pull/1294#issuecomment-706933595 and responses


std/shared/src/main/scala/cats/effect/std/Semaphore.scala
```diff
@@ -158,91 +159,91 @@ object Semaphore {
...
-            case other => (other, false)
-          }
-      }
-
-    def releaseN(n: Long): F[Unit] =
-      assertNonNegative[F](n) *> {
-        if (n == 0) F.unit
-        else
+    def acquireNInternal(n: Long): F[Permit[F]] = {
+      requireNonNegative(n)
```
bplommer, line 161: It seems this diff is horribly mangled 😩 nothing has changed in the method definitions apart from making this change a bunch of times.

djspiewak (approved):
I'm in favor of this, but I'll leave it open for a little while to gather more thoughts, since it could be a somewhat controversial choice.

djspiewak:
Needs a merge with series/3, but otherwise I think we've left this open for comments long enough. :-)

PR #1318, opened 2020-10-17
Add some quality of life utilities

kubukoz:
Just some things I wish existed.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -301,6 +301,8 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
   // utilities
 
+  val stub: IO[Nothing] = raiseError(new NotImplementedError("This IO is not implemented"))
+
   def bothOutcome[A, B](left: IO[A], right: IO[B]): IO[(OutcomeIO[A], OutcomeIO[B])] =
     left.bothOutcome(right)
 
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -301,7 +301,10 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
   // utilities
 
-  val stub: IO[Nothing] = raiseError(new NotImplementedError("This IO is not implemented"))
+  private[this] val _stub: IO[Nothing] = raiseError(
+    new NotImplementedError("This IO is not implemented"))
+
+  def stub: IO[Nothing] = _stub
 
   def bothOutcome[A, B](left: IO[A], right: IO[B]): IO[(OutcomeIO[A], OutcomeIO[B])] =
     left.bothOutcome(right)
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -302,8 +302,8 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
   // utilities
 
   def stub: IO[Nothing] = {
-    val t = new NotImplementedError("this IO is not implemented"))
-    raiseError(t)
+    val e = new NotImplementedError("This IO is not implemented")
+    raiseError(e)
   }
 
   def bothOutcome[A, B](left: IO[A], right: IO[B]): IO[(OutcomeIO[A], OutcomeIO[B])] =
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -301,10 +301,10 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
   // utilities
 
-  private[this] val _stub: IO[Nothing] = raiseError(
-    new NotImplementedError("This IO is not implemented"))
-
-  def stub: IO[Nothing] = _stub
+  def stub: IO[Nothing] = {
+    val t = new NotImplementedError("this IO is not implemented"))
+    raiseError(t)
+  }
 
   def bothOutcome[A, B](left: IO[A], right: IO[B]): IO[(OutcomeIO[A], OutcomeIO[B])] =
     left.bothOutcome(right)
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
@@ -26,6 +26,12 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
   // (sleep(n) *> now) <-> now.map(_ + n + d) forSome { val d: Double }
   def sleep(time: FiniteDuration): F[Unit]
 
+  def delayBy[A](fa: F[A], time: FiniteDuration): F[A] =
+    productR(sleep(time))(fa)
+
+  def andWait[A](fa: F[A], time: FiniteDuration) =
+    productL(fa)(sleep(time))
+
   /**
    * Returns an effect that either completes with the result of the source within
    * the specified time `duration` or otherwise evaluates the `fallback`.
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -407,6 +407,11 @@ sealed abstract class Resource[+F[_], +A] {
     this.flatMap(a => Resource.liftF(f(a)).map(_ => a))
 
   /**
+   * Widens the effect type of this resource.
+   */
+  def covary[G[x] >: F[x]]: Resource[G, A] = this
+
+  /**
    * Converts this to an `InvariantResource` to facilitate pattern matches
    * that Scala 2 cannot otherwise handle correctly.
    */
@@ -495,6 +500,11 @@ object Resource extends ResourceInstances with ResourcePlatform {
     Pure(a)
 
   /**
+   * A resource with a no-op allocation and a no-op release.
+   */
+  val unit: Resource[Nothing, Unit] = pure(())
+
+  /**
    * Lifts an applicative into a resource. The resource has a no-op release.
    * Preserves interruptibility of `fa`.
    *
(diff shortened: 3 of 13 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -477,6 +503,21 @@ object Resource extends ResourceInstances with ResourcePlatform {
...
+  def onFinalize[F[_]: Applicative](release: F[Unit]): Resource[F, Unit] =
+    make(Applicative[F].unit)(_ => release)
+
+  /**
+   * Creates a resource that allocates immediately without any effects,
+   * but calls `release` when closing, providing the [[ExitCase the usage completed with]].
+   * Preserves interruptibility of `release`.
+   */
+  def onFinalizeCase[F[_]: Applicative](release: ExitCase => F[Unit]): Resource[F, Unit] =
+    makeCase(Applicative[F].unit)((_, exit) => release(exit))
```
RaasAhsan: There's an `OnFinalizeCase` constructor in the Resource algebra now, should this call out to that, on a empty resource i'm guessing?

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -477,6 +503,21 @@ object Resource extends ResourceInstances with ResourcePlatform {
   def liftF[F[_], A](fa: F[A]): Resource[F, A] =
     Resource.LiftF(fa)
 
+  /**
+   * Lifts a finalizer into a resource. The resource has a no-op allocation.
+   * Preserves interruptibility of `release`.
```
RaasAhsan: `release` is uncancelable by law

kubukoz: I was following the scaladoc from `liftF`, but I can remove it if it sounds redundant

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
kubukoz: Sure, I can reuse that

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
```diff
@@ -26,6 +26,12 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
   // (sleep(n) *> now) <-> now.map(_ + n + d) forSome { val d: Double }
   def sleep(time: FiniteDuration): F[Unit]
 
+  def delayBy[A](fa: F[A], time: FiniteDuration): F[A] =
+    productR(sleep(time))(fa)
```
djspiewak, line 30: Something that might be nice here: invert the order of the two parameters, since the sleep happens before the effect.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -214,6 +215,25 @@ sealed abstract class Resource[+F[_], +A] {
...
+  /**
+   * Allocates the resource and uses it to run the given Kleisli.
+   */
+  def useKleisli[G[x] >: F[x]: Resource.Bracket, B >: A, C](usage: Kleisli[G, B, C]): G[C] =
+    use(usage.run)
+
+  /**
+   * Creates a FunctionK that, when applied, will allocate the resource and use it to run the given Kleisli.
+   */
+  def useK[G[x] >: F[x]: Resource.Bracket, B >: A]: Kleisli[G, B, *] ~> G =
```
djspiewak: I think `useKleisliK` would be more consistent

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -214,6 +215,25 @@ sealed abstract class Resource[+F[_], +A] {
...
+  /**
+   * Allocates a resource and closes it immediately.
+   */
+  def used[G[x] >: F[x]](implicit G: Resource.Bracket[G]): G[Unit] = use(_ => G.unit)
+
+  /**
+   * Allocates the resource and uses it to run the given Kleisli.
+   */
+  def useKleisli[G[x] >: F[x]: Resource.Bracket, B >: A, C](usage: Kleisli[G, B, C]): G[C] =
+    use(usage.run)
```
djspiewak, line 227: Now that I think about it, this is a really good idea!

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -214,6 +215,25 @@ sealed abstract class Resource[+F[_], +A] {
   def useForever[G[x] >: F[x]](implicit G: Spawn[G]): G[Nothing] =
     use[G, Nothing](_ => G.never)
 
+  /**
+   * Allocates a resource and closes it immediately.
+   */
+  def used[G[x] >: F[x]](implicit G: Resource.Bracket[G]): G[Unit] = use(_ => G.unit)
```
djspiewak: IMO, `use_` would be a more consistent name (think: `traverse` vs `traverse_`).

kernel/shared/src/main/scala/cats/effect/kernel/syntax/ResourceSyntax.scala
```diff
@@ -0,0 +1,29 @@
...
+import cats.effect.kernel.Resource
+
+trait ResourceSyntax {
+  implicit def effectResourceOps[F[_], A](wrapped: F[A]): EffectResourceOps[F, A] =
+    new EffectResourceOps(wrapped)
+}
+
+final class EffectResourceOps[F[_], A] private[syntax] (private[syntax] val wrapped: F[A])
+    extends AnyVal {
+  def liftToResource: Resource[F, A] = Resource.liftF(wrapped)
```
djspiewak: `toResource` maybe?

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -301,6 +301,8 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
   // utilities
 
+  val stub: IO[Nothing] = raiseError(new NotImplementedError("This IO is not implemented"))
```
djspiewak: Should either be a `def` or should use the `private[this] val _stub` pattern we have elsewhere. Public `val`s on objects are *usually* less efficient than `def`s are because of the memory barriers.

djspiewak (commented):
Also I think the build failures are legitimate

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
kubukoz: I don't know, this matches the order you'd call it in on an effect: `print(...).delayBy(10.seconds)` (it's also how the method on IO is defined)

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
kubukoz: It reminds me of `modifyState` in Ref :)

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
kubukoz: fair, although I can speak from experience that people don't like yet another usecase of the underscore symbol...

core/shared/src/main/scala/cats/effect/IO.scala
kubukoz: ok, will change

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
djspiewak: lol fair

kubukoz:
I would make sure this is squashed when merging 😅

djspiewak:
> I would make sure this is squashed when merging

Nah. :-D

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -301,6 +301,11 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
   // utilities
 
+  private[this] val _stub: IO[Nothing] = raiseError(
+    new NotImplementedError("This IO is not implemented"))
+
+  def stub: IO[Nothing] = _stub
```
djspiewak: Just realized something interesting: this is going to throw a constant exception with a useless stack trace. We probably should do something like this:

```suggestion
  def stub: IO[Nothing] = {
    val t = new NotImplementedError("this IO is not implemented"))
    raiseError(t)
  }
```

djspiewak (changes requested):

core/shared/src/main/scala/cats/effect/IO.scala
kubukoz: yeah, I guess this isn't meant to be used in production code so we should rather focus on "experience" things like stack traces than perf.

btw. why the change to lowercase `this`? :D

kubukoz:
I was expecting it to be whitespace (formatting), but no! It was actually a paren.

kubukoz:
I think that'd be it for now... Unless there's anything else that needs attention here?

core/shared/src/main/scala/cats/effect/IO.scala
djspiewak: > btw. why the change to lowercase this? :D

IIRC, in most cases, exceptions have lower-case messages

kernel/shared/src/main/scala/cats/effect/kernel/syntax/ResourceSyntax.scala
```diff
@@ -0,0 +1,29 @@
...
+import cats.effect.kernel.Resource
+
+trait ResourceSyntax {
+  implicit def effectResourceOps[F[_], A](wrapped: F[A]): EffectResourceOps[F, A] =
+    new EffectResourceOps(wrapped)
+}
+
+final class EffectResourceOps[F[_], A] private[syntax] (private[syntax] val wrapped: F[A])
+    extends AnyVal {
+  def toResource: Resource[F, A] = Resource.liftF(wrapped)
```
bwiercinski, line 28: `to` is used when you transforming something eg.: `toList`, `toMap`, `toNel`, `toLong`, `toString`

where in `as` you changing the representation of the same thing: `asInstanceOf`, `asLeft`, `asRight`

`asResource`?


ps: my theory doesn't match with `Functor::as`

PR #1326, opened 2020-10-18
Port Hotswap from Fs2

### LINKED ISSUE #1325: Port Hotswap from FS2
`Hotswap` doesn't rely on `fs2.Stream` and it can be useful in the context of a normal `F`.

RaasAhsan:
Ported from https://github.com/typelevel/fs2/blob/60b57c242d5a1628826b9f14fcb92d60a47fccc2/core/shared/src/main/scala/fs2/Hotswap.scala .

Closes #1325

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
@@ -116,8 +116,8 @@ object Hotswap {
       new Hotswap[F, R] {
 
         override def swap(next: Resource[F, R]): F[R] =
-          F.uncancelable { _ =>
-            next.allocated[F, R].flatMap {
+          F.uncancelable { poll =>
+            poll(next.allocated[F, R]).flatMap {
               case (r, finalizer) =>
                 swapFinalizer(finalizer).as(r)
             }
--- a/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
@@ -20,16 +20,83 @@ import cats.syntax.all._
 import cats.effect.kernel.syntax.all._
 import cats.effect.kernel.{Concurrent, Ref, Resource}
 
+/**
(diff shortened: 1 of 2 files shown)


std/shared/src/main/scala/cats/effect/std/Hotswap.scala
```diff
@@ -0,0 +1,141 @@
...
+            }
+          }
+
+        override def clear: F[Unit] =
+          swapFinalizer(F.unit).uncancelable
+
+        private def swapFinalizer(next: F[Unit]): F[Unit] =
+          state.modify {
+            case Some(previous) =>
+              Some(next) -> previous.attempt.void
```
RaasAhsan: Slight change from the FS2 version -- swallowing errors on finalizers here (can't side channel since we aren't in `Async`, but maybe to the `defaultReporter`?) to remain consistent with `MonadCancel` semantics

RaasAhsan: actually, `Resource#allocated` does this already, removing

std/shared/src/main/scala/cats/effect/std/Hotswap.scala
```diff
@@ -0,0 +1,141 @@
...
+
+    def raise(message: String): F[Unit] =
+      F.raiseError[Unit](new RuntimeException(message))
+
+    Resource.make(initialize)(finalize).map { state =>
+      new Hotswap[F, R] {
+
+        override def swap(next: Resource[F, R]): F[R] =
+          F.uncancelable { _ =>
+            next.allocated[F, R].flatMap {
```
SystemFw: this needs a `poll`

SystemFw: I think somehow the `continual` in fs2 `main` got messed up in the port to `develop` (@mpilquist fyi)

SystemFw (changes requested):

PR #1328, opened 2020-10-18
Runloop cleanup

RaasAhsan:
saw some stuff to cleanup, the vast majority is formatting

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -132,7 +132,9 @@ private final class IOFiber[A](
   /* similar prefetch for EndFiber */
   private[this] val IOEndFiber = IO.EndFiber
 
-  def run(): Unit =
+  override def run(): Unit = {
+    // insert a read barrier after every async boundary
+    readBarrier()
     (resumeTag: @switch) match {
       case 0 => execR()
       case 1 => asyncContinueR()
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -132,7 +132,9 @@ private final class IOFiber[A](
   /* similar prefetch for EndFiber */
   private[this] val IOEndFiber = IO.EndFiber
 
-  def run(): Unit =
+  override def run(): Unit = {
+    // insert a read barrier after every async boundary
+    readBarrier()
```
RaasAhsan, line 137: moved a read barrier from before all async boundaries to after it

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -1022,13 +1026,11 @@ private final class IOFiber[A](
       case 6 => onCancelFailureK(error, depth)
       case 7 => uncancelableFailureK(error, depth)
       case 8 => unmaskFailureK(error, depth)
-      case 9 => succeeded(Left(error), depth + 1) // attemptK
+      case 9 => succeeded(Left(error), depth) // attemptK
```
RaasAhsan, line 1029: `succeeded` and `failed` are tail recursive with respect to this `attempt` case, so we don't actually need to increment the depth counter

PR #1333, opened 2020-10-19
Add cancellation state machine to Dispatcher

### LINKED ISSUE #1332: Dispatcher cancellation backpressure bug
If the unsafe cancel token is invoked, it should backpressure on finalization completion. There is an execution where it will return prematurely, specifically when cancellation occurs before the cancel token is registered.

RaasAhsan:
This PR fixes two bugs:
1. The one described in #1332 where the unsafe cancel token can return before finalization completes.
2. A benign race where a fiber cancel is called twice when the cancel token was only called once

Fixes #1332

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
@@ -25,6 +25,8 @@ import scala.collection.immutable.LongMap
 import scala.concurrent.{Future, Promise}
 
 import java.util.concurrent.atomic.{AtomicBoolean, AtomicReference}
+import scala.util.Success
+import scala.util.Failure
 
 sealed trait Dispatcher[F[_]] extends DispatcherPlatform[F] {
 
@@ -46,18 +48,12 @@ object Dispatcher {
   def apply[F[_]](implicit F: Async[F]): Resource[F, Dispatcher[F]] = {
     final case class Registration(action: F[Unit], prepareCancel: F[Unit] => Unit)
 
-    final case class State(end: Long, registry: LongMap[Registration]) {
+    final case class State(end: Long, registry: LongMap[Registration])
 
-      /*
-       * We know we never need structural equality on this type. We do, however,
-       * perform a compare-and-swap relatively frequently, and that operation
-       * delegates to the `equals` implementation. I like having the case class
-       * for pattern matching and general convenience, so we simply override the
-       * equals/hashCode implementation to use pointer equality.
-       */
-      override def equals(that: Any) = this eq that.asInstanceOf[AnyRef]
-      override def hashCode = System.identityHashCode(this)
-    }
+    sealed trait CancelState
+    case object CancelInit extends CancelState
+    final case class CancelledNoToken(promise: Promise[Unit]) extends CancelState
+    final case class CancelToken(cancelToken: () => Future[Unit]) extends CancelState
 
     val Empty = State(0, LongMap())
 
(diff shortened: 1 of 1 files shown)


std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -46,18 +48,12 @@ object Dispatcher {
   def apply[F[_]](implicit F: Async[F]): Resource[F, Dispatcher[F]] = {
     final case class Registration(action: F[Unit], prepareCancel: F[Unit] => Unit)
 
-    final case class State(end: Long, registry: LongMap[Registration]) {
-
-      /*
-       * We know we never need structural equality on this type. We do, however,
```
RaasAhsan, line 52: I checked out the javadocs again and `AtomicReference` will only ever perform referential equality. You can verify with this snippet:
```scala
@ val initialState = (2, 3)
initialState: (Int, Int) = (2, 3)

@ val state = new AtomicReference(initialState)
state: AtomicReference[(Int, Int)] = (2,3)

@ state.compareAndSet((2, 3), (5, 6))
res3: Boolean = false

@ state.compareAndSet(initialState, (5, 6))
res4: Boolean = true
```

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -25,6 +25,8 @@ import scala.collection.immutable.LongMap
 import scala.concurrent.{Future, Promise}
 
 import java.util.concurrent.atomic.{AtomicBoolean, AtomicReference}
+import scala.util.Success
+import scala.util.Failure
```
djspiewak: Random stupid preference:

```suggestion
import scala.util.{Failure, Success}
```

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
djspiewak: That makes it so much simpler! I did spend some time thinking about this and honestly wondered *how* it could be doing structural equality while keeping things atomic. I had a mental note to decompile the ASM and find out, but I guess it's easier if we just test it and find out that I was wrong. :-P

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -46,18 +48,12 @@ object Dispatcher {
...
-       * equals/hashCode implementation to use pointer equality.
-       */
-      override def equals(that: Any) = this eq that.asInstanceOf[AnyRef]
-      override def hashCode = System.identityHashCode(this)
-    }
+    final case class State(end: Long, registry: LongMap[Registration])
+
+    sealed trait CancelState
+    case object CancelInit extends CancelState
+    final case class CancelledNoToken(promise: Promise[Unit]) extends CancelState
```
djspiewak:
```suggestion
    final case class CanceledNoToken(promise: Promise[Unit]) extends CancelState
```

Just to be consistent with `F.canceled`

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -138,20 +134,34 @@ object Dispatcher {
...
+              state match {
+                case CancelInit =>
+                  if (!cancelState.compareAndSet(state, CancelToken(cancelToken))) {
+                    loop()
+                  }
+                case CancelledNoToken(promise) =>
+                  if (!cancelState.compareAndSet(state, CancelToken(cancelToken))) {
+                    loop()
+                  } else {
+                    import scala.concurrent.ExecutionContext.Implicits.global
```
djspiewak: IMO we would be better off using the parasitic executor, but it doesn't exist on Scala 2.12. I'm very leery about the consequences of introducing a random reference to the global pool here. At the *very* least, this could cause it to allocate threads, which otherwise won't happen at all in a well-behaved CE application.

Can we just capture the `ExecutionContext` in the scope using `F.executionContext`?

djspiewak (changes requested):

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
RaasAhsan: good call, and we only need to capture a reference to it once at the very beginning

PR #1339, opened 2020-10-19
Export Deferred and Ref in cats.effect

### LINKED ISSUE #1290: Ambiguous exports in CE3: Sync, etc.
I have

```scala
import cats.effect._
import cats.effect.kernel._
```

and get

```
reference to Sync is ambiguous;
it is imported twice in the same scope by
import cats.effect.kernel._
and import cats.effect._
```

I'm doing import cats.effect.kernel._ to get `Ref` into scope, and this is triggered from `def foo[F[_]: Sync] = ...`.

@SystemFw says:

> we reexport it, but that might be problematic
> we should either reexport Ref and Deferred too, and discourage the use of kernel, or remove reexports completely, I guess

vasilmkd:
Resolves #1290.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -30,7 +30,6 @@ import cats.{
 import cats.syntax.all._
 import cats.effect.std.Console
 import cats.effect.implicits._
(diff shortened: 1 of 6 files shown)


kubukoz (approved):
LGTM

PR #1340, opened 2020-10-19
Auto yielding

TimWSpence:
Auto-yield by inserting in `IO.cede` every specified numbere of iterations. The default is to check for cancellation every 512 iterations of the runloop and to auto-yield every 1024 but this is configurable in the `IORuntime`.

The behaviour of `global` is to check for the system properties `cats.effect.cancellation.check.threshold` and `cats.effect.auto.yield.threshold.multiplier` but to default to 512 and 2 respectively. The auto-yield threshold is the product of these.

### DIFF
--- a/core/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
+++ b/core/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
@@ -181,9 +181,9 @@ abstract class IOPlatformSpecification extends Specification with ScalaCheck wit
           _ <- IO.sleep(5.millis)
           //Only hope for the cancellation being run is auto-yielding
           _ <- fiber.cancel
-        } yield ()
+        } yield true
 
-        run.evalOn(ec).map { res => res mustEqual () }
+        run.evalOn(ec).map { res => res mustEqual true }
       }
 
     }
--- a/core/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
+++ b/core/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
@@ -173,7 +173,7 @@ abstract class IOPlatformSpecification extends Specification with ScalaCheck wit
 
         val ec = ExecutionContext.fromExecutorService(Executors.newSingleThreadExecutor())
 
-        val run = (for {
+        val run = for {
           //Run in a tight loop on single-threaded ec so only hope of
           //seeing cancellation status is auto-cede
           fiber <- forever.start
@@ -181,9 +181,9 @@ abstract class IOPlatformSpecification extends Specification with ScalaCheck wit
           _ <- IO.sleep(5.millis)
           //Only hope for the cancellation being run is auto-yielding
           _ <- fiber.cancel
-        } yield true).guarantee(IO(ec.shutdown()))
+        } yield true
 
-        run.evalOn(ec).flatMap { res => IO(res must beTrue) }
+        run.evalOn(ec).guarantee(IO(ec.shutdown())).flatMap { res => IO(res must beTrue) }
       }
 
     }
--- a/core/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
+++ b/core/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
@@ -168,6 +168,24 @@ abstract class IOPlatformSpecification extends Specification with ScalaCheck wit
         } yield ok
       }
 
+      "auto-cede" in real {
+        val forever = IO.unit.foreverM
+
+        val ec = ExecutionContext.fromExecutorService(Executors.newSingleThreadExecutor())
+
+        val run = for {
+          //Run in a tight loop on single-threaded ec so only hope of
+          //seeing cancellation status is auto-cede
+          fiber <- forever.start
+          //Allow the tight loop to be scheduled
+          _ <- IO.sleep(5.millis)
+          //Only hope for the cancellation being run is auto-yielding
+          _ <- fiber.cancel
+        } yield ()
+
+        run.evalOn(ec).map { res => res mustEqual () }
+      }
+
     }
   }
 }
--- a/core/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
+++ b/core/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
@@ -173,7 +173,7 @@ abstract class IOPlatformSpecification extends Specification with ScalaCheck wit
 
         val ec = ExecutionContext.fromExecutorService(Executors.newSingleThreadExecutor())
 
-        val run = for {
+        val run = (for {
           //Run in a tight loop on single-threaded ec so only hope of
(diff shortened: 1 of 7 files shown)


core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -67,7 +67,9 @@ private final class IOFiber[A](
     initMask: Int,
     cb: OutcomeIO[A] => Unit,
     startIO: IO[A],
-    startEC: ExecutionContext)
+    startEC: ExecutionContext,
```
RaasAhsan, line 70: Should we just start passing in the `IORuntime` into the constructor and adding `private[this]` fields for stuff inside it?

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -210,15 +214,21 @@ private final class IOFiber[A](
...
 
+    if ((nextIteration % cancellationCheckThreshold) == 0) {
+      //Ensure that we see cancellation
+      readBarrier()
+    }
+
     if (shouldFinalize()) {
       asyncCancel(null)
+    } else if ((nextIteration % autoYieldThreshold) == 0) {
+      runLoop(IO.cede >> _cur0, nextIteration)
```
RaasAhsan: So this is actually a `FlatMap(IO.cede, _ => _cur0`, I think there is a case where we're inserting two `cede`s?

Assume that `iterationThreshold` = `autoYieldThreshold` = 512. Let's say that in the current execution, `nextIteration` is 512, so we auto-yield here, and continue the runloop. When we evaluate the `FlatMap`, `iteration` will be 512 so `nextIteration` will be 513, and we run the `FlatMap`. On the _next_ loop, `iteration` will be 513, so `nextIteration` will be 0, so we add insert another `cede`. Am I running through that correctly?

core/shared/src/main/scala/cats/effect/IOFiber.scala
djspiewak: There are also other good reasons to do this

core/shared/src/main/scala/cats/effect/IOFiber.scala
TimWSpence: Good spot!! 👍

TimWSpence:
Also if anyone can tell me why git diff has lost its mind on `IOFiber` I would be very grateful :) The indentation changes are expected but not the rest

core/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
```diff
@@ -168,6 +168,24 @@ abstract class IOPlatformSpecification extends Specification with ScalaCheck wit
...
+          //Run in a tight loop on single-threaded ec so only hope of
+          //seeing cancellation status is auto-cede
+          fiber <- forever.start
+          //Allow the tight loop to be scheduled
+          _ <- IO.sleep(5.millis)
+          //Only hope for the cancellation being run is auto-yielding
+          _ <- fiber.cancel
+        } yield true
+
+        run.evalOn(ec).map { res => res mustEqual true }
```
djspiewak: Should `.guarantee(IO(ec.shutdown()))` on this. Also, `res mustEqual true` can be `res must beTrue` for nicer messages, and also is a side effect and should be wrapped in `IO(...)`

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -132,6 +133,10 @@ private final class IOFiber[A](
   /* similar prefetch for EndFiber */
   private[this] val IOEndFiber = IO.EndFiber
 
+  private[this] val cancellationCheckThreshold = runtime.config.cancellationCheckThreshold
+  private[this] val autoYieldThreshold = runtime.config.autoYieldThreshold
+  private[this] val iterationThreshold = autoYieldThreshold
```
djspiewak, line 138: We do need to check the performance implications of this. I think it would be worth running the benchmarks with this setup, then try swapping in a hard-coded integer value and see how different the results are. *Hopefully* the difference will be negligable.

core/shared/src/main/scala/cats/effect/IOFiber.scala
djspiewak: I also think this will be a lot easier (and cheaper!!) if we explicitly manipulate the continuation stack and `reschedule` the same as the implementation of `Cede`. We could probably extract that as common code in some tiny function somewhere without any performance impact.

core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
```diff
@@ -37,9 +37,30 @@ final class IORuntime private[effect] (
...
+    val config: IORuntimeConfig) {
 
-  override def toString: String = s"IORuntime($compute, $scheduler)"
+  override def toString: String = s"IORuntime($compute, $scheduler, $config)"
+}
+
+final case class IORuntimeConfig private (
+    val cancellationCheckThreshold: Int,
+    val autoYieldThreshold: Int
+)
```
djspiewak: Minor preference:

```suggestion
    val autoYieldThreshold: Int)
```

core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
```diff
@@ -37,9 +37,30 @@ final class IORuntime private[effect] (
...
+
+object IORuntimeConfig {
+  def apply(): IORuntimeConfig = new IORuntimeConfig(512, 1024)
+
+  def apply(cancellationCheckThreshold: Int, autoYieldThreshold: Int): IORuntimeConfig = {
+    if (autoYieldThreshold % cancellationCheckThreshold == 0)
+      new IORuntimeConfig(
+        cancellationCheckThreshold,
+        autoYieldThreshold
+      )
```
djspiewak:
```suggestion
        autoYieldThreshold)
```

djspiewak (changes requested):

core/shared/src/main/scala/cats/effect/IOFiber.scala
TimWSpence: You mean because it's not a constant and hence can't be inlined? I had wondered about this...

djspiewak:
I'm going to merge this so we can unblock M2, but we should follow-up on a) getting the benchmarks running, b) testing the relative performance impact of this whole change, and c) testing the relative performance impact of the way we're encoding the tuning parameters.

djspiewak:
I'm going to merge this to unblock M2, but we need to follow up on the remaining issues:

- Can improve performance of the autoyield by adding a special new `K` type rather than reusing `FlatMapK`. This special new type would avoid the `Function1` allocation
- Have to fix the benchmarks
- Test the relative performance impact of this change (relative to having it off)
- Test the relative performance impact of this constant encoding, rather than some alternative one where the JVM knows that the constants are always the same value forever and thus can inline more aggressively (like with `IOFiberConstants`)

PR #1349, opened 2020-10-22
Exit on fatal error

### LINKED ISSUE #1298: Make the blocking pool private within IORuntime
We want to leave the door open to, in the future, not use a thread pool.

### LINKED ISSUE #1327: Work stealing thread pool not parking
I was running a test application today in SBT where it doesn't fork the JVM. I noticed that after the application terminated, the SBT process was seeing 800% CPU utilization. I captured a thread dump and saw:
```
"io-compute-1" #197 daemon prio=5 os_prio=31 tid=0x00007feae741b800 nid=0x13d03 runnable [0x000070000fef4000]
   java.lang.Thread.State: RUNNABLE
        at sun.misc.Unsafe.park(Native Method)
        - parking to wait for  <0x00000006c7322960> (a cats.effect.unsafe.WorkStealingThreadPool)
        at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
        at cats.effect.unsafe.WorkerThread.park(WorkerThread.scala:166)
        at cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:268)

   Locked ownable synchronizers:
        - None
```
It seems like threads these threads are actually executing and _not_ parking, which is causing the high CPU. Compare with a regular executor:
```

"pool-17-thread-12" #188 prio=5 os_prio=31 tid=0x00007f9e72278000 nid=0xb803 waiting on condition [0x0000700010aa5000]
   java.lang.Thread.State: TIMED_WAITING (parking)
        at sun.misc.Unsafe.park(Native Method)
        - parking to wait for  <0x00000006c6c59170> (a java.util.concurrent.SynchronousQueue$TransferStack)
        at java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:215)
        at java.util.concurrent.SynchronousQueue$TransferStack.awaitFulfill(SynchronousQueue.java:460)
        at java.util.concurrent.SynchronousQueue$TransferStack.transfer(SynchronousQueue.java:362)
        at java.util.concurrent.SynchronousQueue.poll(SynchronousQueue.java:941)
        at java.util.concurrent.ThreadPoolExecutor.getTask(ThreadPoolExecutor.java:1073)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1134)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
```

### LINKED ISSUE #1355: What to do when fatal errors are thrown
The `IOFiber` runloop currently handles `NonFatal` exceptions.

Any exception not caught by this method continues down the call stack in the carrier thread loop. For example, in the work stealing pool, these are currently not handled, the carrier thread crashes and the application continues with 1 fewer threads (we currently do not have a mechanism of reporting this fatal error to the place where `unsafeRunSync()` was called, so no shutdown and vm exit logic can occur).

If you thought this is somehow better managed by the old Java fixed thread pool, you would be wrong. Running the following application inside sbt with a fixed IO compute thread pool completely hangs the process.

```scala
  def run: IO[Unit] = IO[Unit](throw new OutOfMemoryError("Boom!"))
}
```

None of this is ideal, as you can imagine. We need to arrive at an acceptable solution. Keep in mind that we should keep the experience somewhat consistent between our work stealing thread pool which we completely control and can add new functionality to, compared to a java executor or a custom `ExecutionContext`.

Interestingly, this is somewhat at odds with the final decision made in https://github.com/typelevel/cats-effect/pull/1277.

vasilmkd:
Resolves #1327.
Resolves #1355.
Resolves #1298.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
@@ -315,7 +315,7 @@ private[effect] final class WorkStealingThreadPool(
       }
 
       // `unsafeRunFiber(true)` will enqueue the fiber, no need to do it manually
-      IO(runnable.run()).unsafeRunFiber(_.fold(reportFailure(_), _ => ()))(self)
+      IO(runnable.run()).unsafeRunFiber(reportFailure, _ => ())(self)
       ()
     }
   }
@@ -339,8 +339,6 @@ private[effect] final class WorkStealingThreadPool(
     externalQueue.shutdown()
     // Send an interrupt signal to each of the worker threads.
     workerThreads.foreach(_.interrupt())
-    // Join all worker threads.
-    workerThreads.foreach(_.join())
     // Remove the references to the worker threads so that they can be cleaned up, including their worker queues.
     for (i <- 0 until workerThreads.length) {
       workerThreads(i) = null
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
@@ -161,7 +161,7 @@ private final class WorkerThread(
 
     // Only actually park if the pool has not been shut down. Otherwise, this
     // will break out of the run loop and end the thread.
-    while (!pool.done) {
+    while (!pool.done && !isInterrupted()) {
       // Park the thread until further notice.
       LockSupport.park(pool)
 
@@ -247,7 +247,7 @@ private final class WorkerThread(
     var fiber: IOFiber[_] = null
 
     // Loop until the pool has been shutdown.
-    while (!pool.done) {
+    while (!pool.done && !isInterrupted()) {
       tick += 1 // Count each iteration.
 
       // Try to obtain a fiber from the local queue or the external
--- a/core/jvm/src/test/scala/cats/effect/IOAppSpec.scala
+++ b/core/jvm/src/test/scala/cats/effect/IOAppSpec.scala
@@ -46,22 +46,13 @@ class IOAppSpec extends Specification {
       h.stdout() mustEqual expected.mkString("", System.lineSeparator(), System.lineSeparator())
     }
 
-    "exit on fatal error" in {
-      val h = java(FatalError, List.empty)
-      h.awaitStatus() mustEqual 1
-      h.stderr() must contain("Boom!")
-    }
-
-    "exit on fatal error with other unsafe runs" in {
-      val h = java(FatalErrorUnsafeRun, List.empty)
-      h.awaitStatus() mustEqual 1
-      h.stderr() must contain("Boom!")
-    }
-
     if (System.getProperty("os.name").toLowerCase.contains("windows")) {
       // The jvm cannot gracefully terminate processes on Windows, so this
       // test cannot be carried out properly. Same for testing IOApp in sbt.
       "run finalizers on TERM" in skipped(
+        "cannot observe graceful process termination on Windows")
+      "exit on fatal error" in skipped("cannot observe graceful process termination on Windows")
+      "exit on fatal error with other unsafe runs" in skipped(
         "cannot observe graceful process termination on Windows")
     } else {
       "run finalizers on TERM" in {
@@ -104,6 +95,18 @@ class IOAppSpec extends Specification {
           }
           readTest() must contain("canceled")
         }
+      }
+
+      "exit on fatal error" in {
+        val h = java(FatalError, List.empty)
+        h.awaitStatus() mustEqual 1
+        h.stderr() must contain("Boom!")
+      }
+
+      "exit on fatal error with other unsafe runs" in {
+        val h = java(FatalErrorUnsafeRun, List.empty)
+        h.awaitStatus() mustEqual 1
+        h.stderr() must contain("Boom!")
       }
     }
   }
@@ -171,7 +174,8 @@ package examples {
 
     def run(args: List[String]): IO[ExitCode] =
       for {
-        _ <- IO(IO(throw new OutOfMemoryError("Boom!")).start.void.unsafeRunSync())
+        _ <- (0 until 100).toList.traverse(_ => IO.blocking(IO.never.unsafeRunSync()).start)
+        _ <- IO.blocking(IO(throw new OutOfMemoryError("Boom!")).start.unsafeRunSync())
         _ <- IO.never[Unit]
       } yield ExitCode.Success
   }
--- a/core/jvm/src/test/scala/cats/effect/IOAppSpec.scala
+++ b/core/jvm/src/test/scala/cats/effect/IOAppSpec.scala
@@ -20,6 +20,7 @@ import cats.syntax.all._
 
 import org.specs2.mutable.Specification
 
+import scala.io.Source
 import scala.sys.process.{BasicIO, Process}
 
 import java.io.File
@@ -43,6 +44,18 @@ class IOAppSpec extends Specification {
       val h = java(Arguments, expected)
       h.awaitStatus() mustEqual 0
       h.stdout() mustEqual expected.mkString("", System.lineSeparator(), System.lineSeparator())
+    }
+
+    "exit on fatal error" in {
+      val h = java(FatalError, List.empty)
+      h.awaitStatus() mustEqual 1
+      h.stderr() must contain("Boom!")
+    }
+
+    "exit on fatal error with other unsafe runs" in {
+      val h = java(FatalErrorUnsafeRun, List.empty)
+      h.awaitStatus() mustEqual 1
+      h.stderr() must contain("Boom!")
     }
 
     if (System.getProperty("os.name").toLowerCase.contains("windows")) {
@@ -97,15 +110,20 @@ class IOAppSpec extends Specification {
 
   def java(proto: IOApp, args: List[String]): Handle = {
     val stdoutBuffer = new StringBuffer()
+    val stderrBuffer = new StringBuffer()
     val builder = Process(
       s"${JavaHome}/bin/java",
       List("-cp", ClassPath, proto.getClass.getName.replaceAll("\\$$", "")) ::: args)
-    val p = builder.run(BasicIO(false, stdoutBuffer, None))
+    val p = builder.run(BasicIO(false, stdoutBuffer, None).withError { in =>
+      val err = Source.fromInputStream(in).getLines().mkString(System.lineSeparator())
+      stderrBuffer.append(err)
+      ()
+    })
 
     new Handle {
       def awaitStatus() = p.exitValue()
       def term() = p.destroy() // TODO probably doesn't work
-      def stderr() = "" // TODO
+      def stderr() = stderrBuffer.toString
       def stdout() = stdoutBuffer.toString
     }
   }
@@ -141,5 +159,20 @@ package examples {
       (IO(println("Started")) >> IO.never)
         .onCancel(writeToFile("canceled", new File(args.head)))
         .as(ExitCode.Success)
+  }
+
+  object FatalError extends IOApp {
+    def run(args: List[String]): IO[ExitCode] =
+      IO(throw new OutOfMemoryError("Boom!")).as(ExitCode.Success)
+  }
+
+  object FatalErrorUnsafeRun extends IOApp {
+    import cats.effect.unsafe.implicits.global
+
+    def run(args: List[String]): IO[ExitCode] =
+      for {
+        _ <- IO(IO(throw new OutOfMemoryError("Boom!")).start.void.unsafeRunSync())
+        _ <- IO.never[Unit]
+      } yield ExitCode.Success
   }
 }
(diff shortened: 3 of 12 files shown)


core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -135,7 +135,14 @@ private final class WorkerThread(
     // Announce that this thread is no longer searching for work, if that was the case before.
     transitionFromSearching()
     // Execute the fiber.
-    fiber.run()
+    try fiber.run()
+    catch {
+      // Handle fatal errors.
+      case t: Throwable =>
+        pool.reportFailure(t)
+        fiber.errorCb(Left(t))
```
djspiewak: Oh! This is much nicer than what I was thinking of. The problem though is that we may have multiple separate fiber chains running on the same pool. Specifically, if someone is doing `unsafeRunSync()` multiple times against the same `IORuntime`. I think I'm kind of okay with all-but-one of them deadlocking though.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -135,7 +135,14 @@ private final class WorkerThread(
...
+        System.exit(1)
```
djspiewak: IMO we shouldn't exit. Let's allow the downstream callback to do that. However, we *should* completely shut down the pool.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -67,8 +67,9 @@ private final class IOFiber[A](
     initMask: Int,
     cb: OutcomeIO[A] => Unit,
     startIO: IO[A],
-    startEC: ExecutionContext)
-    extends IOFiberPlatform[A]
+    startEC: ExecutionContext,
+    val errorCb: Left[Throwable, Nothing] => Unit
```
djspiewak: This is just `Throwable => Unit`. Also should be `private[effect] val`. This is actually pretty important because we can't really allocate reliably after an OOM.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -222,7 +222,8 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
       (oc: OutcomeIO[A]) =>
         oc.fold((), e => cb(Left(e)), ioa => cb(Right(ioa.asInstanceOf[IO.Pure[A]].value))),
       this,
-      runtime.compute
+      runtime.compute,
+      cb
```
djspiewak: The wrapping in `Left` should be here, IMO, and in some cases (e.g. `unsafeRunSync()`) we can avoid it altogether by slightly revising the signature of `unsafeRunFiber` (to take a `Throwable => Unit` and an `A => Unit` instead of a single one with `Either`).

djspiewak (changes requested):

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -135,7 +135,14 @@ private final class WorkerThread(
     // Announce that this thread is no longer searching for work, if that was the case before.
     transitionFromSearching()
     // Execute the fiber.
-    fiber.run()
+    try fiber.run()
+    catch {
+      // Handle fatal errors.
+      case t: Throwable =>
```
RaasAhsan: This is catching all throwables (rather than just fatal `Error`s); is there another `try`/`catch` inside `fiber.run()` that catches nonfatal exceptions?

core/shared/src/main/scala/cats/effect/IOFiber.scala
RaasAhsan: Eventually it's going to be lifted into a `Left` anyways though, since that's the signature of `cb`. It seems kind of futile to catch these fatal errors :P

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
RaasAhsan: What are the behaviors of the default `Executors` with respect to fatal errors? I'm a bit in favor of keeping consistent with that behavior unless there's a really good reason not to

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
vasilmkd: Yes, the `IOFiber` runloop catches non fatal errors.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
vasilmkd: Downstream callback may not exit the VM, think about `unsafeRunSync()` from anywhere. You're right coupled with your previous comment about only one fiber chain being notified, this solution doesn't cut it.

core/shared/src/main/scala/cats/effect/IOFiber.scala
vasilmkd: We must do _something_, currently a fatal error crashes a carrier thread from the work stealing thread pool and the program continues as if nothing happened. Feel free to try it.

djspiewak: > Eventually it's going to be lifted into a Left anyways though, since that's the signature of cb. It seems kind of futile to catch these fatal errors :P

Doesn't have to be. We can avoid the lifting for `unsafeRunSync` for example.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
djspiewak: There also isn't a particularly clean "catch only fatal errors" mechanism, much less one I would trust when the VM is on fire because of one. Catching `Throwable` also has the advantage that we sweep up `InterruptedException`, which is currently incorrectly handled.

vasilmkd:
Resolves #1327.
Resolves #1355.
Resolves #1298.

This is now basically a different PR, but I guess it could be useful to keep the old comments.

I have used the following `IOApp` for testing, let me know if the results are satisfying.
```scala
package cats.effect

object CatsEffectApp extends IOApp.Simple {
  def run: IO[Unit] =
    for {
      _ <- IO.sleep(10.seconds)
      _ <- IO[Unit](throw new OutOfMemoryError("Boom!"))
    } yield ()
}
```

Console output:
Screen Shot 2020-10-25 at 02 10 22

JDK Mission Control screenshot during execution:
Screen Shot 2020-10-25 at 02 34 24

JDK Mission Control screenshot after execution:
Screen Shot 2020-10-25 at 02 34 49

Activity Monitor screenshot after execution (no spinning threads):
Screen Shot 2020-10-25 at 02 39 02

vasilmkd:
I also apologize in advance for implementing another questionable concurrent data structure.

I know how much you dislike locking, but please keep in mind that we are trying to give best effort reporting to all ends of the world (think `unsafeRunSync()` locations) while the JVM is in a very bad state due to having thrown a fatal error. `unsafeRunSync()` operations are fairly rare in practice, so they don't need high performance concurrency.

On the other hand, we absolutely need this hand rolled hashtable because we are trying to signal fatal errors without further allocations.

Hopefully these reasons are enough to convince you.

vasilmkd:
Special thanks to @djspiewak, I would not have been able to come close to this solution without your help.

djspiewak:
> unsafeRunSync() operations are fairly rare in practice, so they don't need high performance concurrency.

Exactly this. I think locking is actually okay here.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -315,7 +315,7 @@ private[effect] final class WorkStealingThreadPool(
       }
 
       // `unsafeRunFiber(true)` will enqueue the fiber, no need to do it manually
-      IO(runnable.run()).unsafeRunFiber(_.fold(reportFailure(_), _ => ()))(self)
+      IO(runnable.run()).unsafeRunFiber(reportFailure, _ => ())(self)
```
djspiewak, line 318: Amusingly, I completely forgot this was a source of unsafe running…

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -211,18 +211,29 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
...
 
-    val fiber = new IOFiber(
+    val fiber = new IOFiber[A](
       0,
-      (oc: OutcomeIO[A]) =>
-        oc.fold((), e => cb(Left(e)), ioa => cb(Right(ioa.asInstanceOf[IO.Pure[A]].value))),
+      oc =>
+        oc.fold(
+          (),
+          t => {
```
djspiewak: Very minor nit: I generally prefer `{ t =>` rather than `t => {` (since the former syntax works with paren-less function calls while the latter syntax does not)

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -135,15 +135,33 @@ private final class IOFiber[A](
...
+          val len = runtime.fiberErrorCbs.hashtable.length
+          while (idx < len) {
+            val cb = runtime.fiberErrorCbs.hashtable(idx)
+            if (cb != null) {
+              cb(t)
+            }
+            idx += 1
+          }
+        }
+        runtime.internalShutdown()
```
djspiewak: I think we have to shut down the runtime *before* we notify the callbacks, otherwise one could sneak in as we're notifying.

core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
```diff
@@ -34,12 +34,14 @@ avoid unintentionally degrading your application performance.
 """)
 // Constructor visible in the effect package for use in benchmarks.
 final class IORuntime private[effect] (
-    val compute: ExecutionContext,
+    private[effect] val compute: ExecutionContext,
```
djspiewak: I think that @SystemFw wanted these to be public. `blocking` is the one that really needs to be private; technically the other two can be public without a problem, I think.

core/jvm/src/test/scala/cats/effect/IOAppSpec.scala
```diff
@@ -142,4 +160,19 @@ package examples {
...
+    def run(args: List[String]): IO[ExitCode] =
+      IO(throw new OutOfMemoryError("Boom!")).as(ExitCode.Success)
+  }
+
+  object FatalErrorUnsafeRun extends IOApp {
+    import cats.effect.unsafe.implicits.global
+
+    def run(args: List[String]): IO[ExitCode] =
+      for {
+        _ <- IO(IO(throw new OutOfMemoryError("Boom!")).start.void.unsafeRunSync())
```
djspiewak: This test is probably more interesting if you add a few more lines like `_ <- IO(IO.never.unsafeRunSync()).start` to this for-comprehension.

core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
vasilmkd: Oh wow, I have made it exactly opposite to what's needed here https://github.com/typelevel/cats-effect/issues/1298, my bad...

core/shared/src/main/scala/cats/effect/IO.scala
vasilmkd: Unwanted syntax gone.

core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala

core/jvm/src/test/scala/cats/effect/IOAppSpec.scala
vasilmkd: Done.

core/shared/src/main/scala/cats/effect/IOFiber.scala
vasilmkd: Done. I made sure to handle interruption coming from the shutdown, hopefully this is how it should be done. Please check again.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -339,8 +339,6 @@ private[effect] final class WorkStealingThreadPool(
     externalQueue.shutdown()
     // Send an interrupt signal to each of the worker threads.
     workerThreads.foreach(_.interrupt())
-    // Join all worker threads.
-    workerThreads.foreach(_.join())
```
djspiewak, line 343: Out of curiosity, what are the implications of removing this joining?

djspiewak (approved):

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
vasilmkd: If attempted from inside the pool, from a worker thread, it will try to join itself and deadlock.

Issue #1291, opened 2020-10-09
Port parTraverseN, parSequenceN to CE3


arosien:
Currently missing.

djspiewak:
I'm going to tag this as a good first issue, since it should be relatively mechanical!

TimWSpence:
I had a quick look at this and it looks like we might want to move `Semaphore` to `kernel` as well for this. Discussion [here](https://gitter.im/typelevel/cats-effect-dev?at=5f9149dc270d004bcfe1d149)

Avasil:
`Semaphore` implementation is good enough to start but since we have a `Queue`, we could use an implementation similar to this one: https://github.com/monix/monix/blob/series/3.x/monix-eval/shared/src/main/scala/monix/eval/internal/TaskParSequenceN.scala

The problem with `Semaphore` implementation is that ALL fibers are started and then they wait asynchronously for a permit.
It wastes a lot of memory and it breaks if the list has too many elements IIRC

TimWSpence:
@Avasil nice, that could be good. But still has the same problem that `Queue` is currently in `std` rather than `kernel`. It seems like we might need some more of these abstractions to live in `kernel`

djspiewak:
The queue implementation is definitely a lot nicer, though the inversion of dependencies here is really annoying. I love having `Queue` (and `Semaphore`) in **std** instead of **kernel**, since they aren't primitives, but `parTraverseN` is compelling. Perhaps it would be best to start with an implementation on `IO` itself while we ponder how best to make it available within **kernel**.

TimWSpence:
> The queue implementation is definitely a lot nicer, though the inversion of dependencies here is really annoying. I love having `Queue` (and `Semaphore`) in **std** instead of **kernel**, since they aren't primitives, but `parTraverseN` is compelling. Perhaps it would be best to start with an implementation on `IO` itself while we ponder how best to make it available within **kernel**.

👍 I'll have a go at that today

PR #1354, opened 2020-10-23
parTraverseN and parSequenceN

TimWSpence:
Stop-gap solution while we figure out how we want to do this for any `Concurrent[F]`. This requires us to decide whether we want to move `Queue` or `Semaphore` into `kernel` - see https://github.com/typelevel/cats-effect/issues/1291 for more details


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -280,6 +281,24 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
...
   def never[A]: IO[A] = _never
 
+  /**
+   * Like `Parallel.parTraverse`, but limits the degree of parallelism.
+   */
+  def parTraverseN[T[_]: Traverse, A, B](n: Long)(ta: T[A])(f: A => IO[B]): IO[T[B]] =
+    for {
+      semaphore <- Semaphore[IO](n)
+      tb <- ta.parTraverse { a => semaphore.permit.use(_ => f(a)) }
+    } yield tb
```
djspiewak: I would prefer we use the `Queue` implementation (as @Avasil suggested) rather than the `Semaphore` variant.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -277,6 +278,53 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
...
+      pairs <- ta.traverse(a => Deferred[IO, B].map(p => (p, f(a))))
+      _ <- pairs.traverse_(queue.offer(_))
+      workers =
+        List
+          .fill(n) {
+            queue
+              .take
+              .flatMap {
+                case (p, task) =>
+                  task.redeemWith(
```
RaasAhsan: interesting question: should we spawn this task in another fiber so that self-cancellation doesn't interrupt a worker? then we can respond properly by terminating the whole operation

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -277,6 +278,53 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
...
+          error.get,
+          (pairs.map(_._1.get)).sequence
+        ).flatMap {
+          case Left(err) =>
+            IO.raiseError(err)
+
+          case Right(values) =>
+            IO.pure(values)
+        }
+      } { fibers => fibers.traverse(_.cancel).void }
```
RaasAhsan: We can cancel these fibers in parallel

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -277,6 +278,53 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
...
+   * Like `Parallel.parTraverse`, but limits the degree of parallelism.
+   *
+   * Based on https://github.com/monix/monix/blob/series/3.x/monix-eval/shared/src/main/scala/monix/eval/internal/TaskParSequenceN.scala
+   */
+  def parTraverseN[T[_]: Traverse, A, B](n: Int)(ta: T[A])(f: A => IO[B]): IO[T[B]] =
+    for {
+      error <- Deferred[IO, Throwable]
+      queue <- Queue.bounded[IO, (Deferred[IO, B], IO[B])](n)
+      pairs <- ta.traverse(a => Deferred[IO, B].map(p => (p, f(a))))
+      _ <- pairs.traverse_(queue.offer(_))
```
RaasAhsan: Should we `parTraverse` here to make things more nondeterministic?

SystemFw:
https://gitter.im/typelevel/cats-effect-dev?at=5f9838d1a0a3072b43aad96c

Some relevant discussion happening here

core/shared/src/main/scala/cats/effect/IO.scala
SystemFw: I think the whole point of this implementation is to not do that, otherwise you can use the implementation with `Semaphore`

SystemFw:
@TimWSpence if we get this to build, I will merge and do the changes on MiniSemaphore in a separate PR :)

djspiewak:
Firefox needs a more generous timeout, in my experience. Usually about twice as high. Btw you can run the Firefox tests locally by installing `geckodriver` (homebrew is sufficient for this if you're on a mac) and just running `ciFirefox`. Or alternatively, `; project coreJS; set useFirefox := true; test`

TimWSpence:
> Firefox needs a more generous timeout, in my experience. Usually about twice as high. Btw you can run the Firefox tests locally by installing `geckodriver` (homebrew is sufficient for this if you're on a mac) and just running `ciFirefox`. Or alternatively, `; project coreJS; set useFirefox := true; test`

Where is the timeout configured for firefox sorry? Or why has this not failed before?

TimWSpence:
> > Firefox needs a more generous timeout, in my experience. Usually about twice as high. Btw you can run the Firefox tests locally by installing `geckodriver` (homebrew is sufficient for this if you're on a mac) and just running `ciFirefox`. Or alternatively, `; project coreJS; set useFirefox := true; test`
> 
> Where is the timeout configured for firefox sorry? Or why has this not failed before?

Never mind - I realized it's probably because we now have property tests that run lots of iterations. I've moved them to a separate test suite with a larger timeout

PR #1361, opened 2020-10-26
Std Priority Queue

TimWSpence:
Priority queue for std, based on Okasaki's binomial heap.

Circular and dropping variants are not supported as the heap condition is not strong enough for us to be able to do this in < O(n).

### DIFF
--- a/core/shared/src/test/scala/cats/effect/std/PQueueSpec.scala
+++ b/core/shared/src/test/scala/cats/effect/std/PQueueSpec.scala
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2020 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * These tests have been inspired by and adapted from `monix-catnap`'s `ConcurrentQueueSuite`, available at
+ * https://github.com/monix/monix/blob/series/3.x/monix-catnap/shared/src/test/scala/monix/catnap/ConcurrentQueueSuite.scala.
+ */
+
+package cats.effect
+package std
+
+import cats.implicits._
+import cats.Order
+import cats.arrow.FunctionK
+import org.specs2.specification.core.Fragments
+
+import scala.collection.immutable.{Queue => ScalaQueue}
+import scala.concurrent.duration._
+
+class PQueueSpec extends BaseSpec {
+
+  implicit val orderForInt: Order[Int] = Order.fromLessThan((x, y) => x < y)
+
+  "PQueue" should {
+    boundedPQueueTests("PQueue", PQueue.bounded)
+    boundedPQueueTests("PQueue mapK", PQueue.bounded[IO, Int](_).map(_.mapK(FunctionK.id)))
+  }
+
+  private def boundedPQueueTests(
+      name: String,
+      constructor: Int => IO[PQueue[IO, Int]]): Fragments = {
+
+    s"$name - take when empty" in real {
+      for {
+        q <- constructor(0)
+        _ <- q.offer(1).start
+        v1 <- q.take
+        f <- q.take.start
+        _ <- q.offer(2)
+        v2 <- f.joinAndEmbedNever
+        r <- IO((v1 must beEqualTo(1)) and (v2 must beEqualTo(2)))
+      } yield r
+    }
+
+    s"$name - simple order check" in real {
+      for {
+        q <- constructor(10)
+        _ <- List(1, 3, 4, 2, 5).traverse(q.offer(_))
+        res <- List.fill(5)(q.take).sequence
+        r <- IO(res must beEqualTo(List(1, 2, 3, 4, 5)))
+      } yield r
+    }
+
+  }
+}
--- a/core/shared/src/test/scala/cats/effect/std/PQueueSpec.scala
+++ b/core/shared/src/test/scala/cats/effect/std/PQueueSpec.scala
@@ -30,7 +30,8 @@ import org.specs2.specification.core.Fragments
 import scala.collection.immutable.{Queue => ScalaQueue}
 import scala.concurrent.duration._
 
-class PQueueSpec extends BaseSpec {
+class BoundedPQueueSpec extends BaseSpec with PQueueTests {
+  sequential
 
   implicit val orderForInt: Order[Int] = Order.fromLessThan((x, y) => x < y)
 
@@ -43,7 +44,7 @@ class PQueueSpec extends BaseSpec {
       name: String,
       constructor: Int => IO[PQueue[IO, Int]]): Fragments = {
 
-    s"$name - take when empty" in real {
+    s"$name - demonstrate offer and take with zero capacity" in real {
       for {
         q <- constructor(0)
         _ <- q.offer(1).start
@@ -55,6 +56,54 @@ class PQueueSpec extends BaseSpec {
       } yield r
     }
 
+    s"$name - async take with zero capacity" in realWithRuntime { implicit rt =>
+      for {
+        q <- constructor(0)
+        _ <- q.offer(1).start
+        v1 <- q.take
+        _ <- IO(v1 must beEqualTo(1))
+        ff <- IO(q.take.unsafeToFuture()).start
+        f <- ff.joinAndEmbedNever
+        _ <- IO(f.value must beEqualTo(None))
+        _ <- q.offer(2)
+        v2 <- IO.fromFuture(IO.pure(f))
+        r <- IO(v2 must beEqualTo(2))
+      } yield r
+    }
+
+    s"$name - offer/take with zero capacity" in real {
+      val count = 1000
+
+      def producer(q: PQueue[IO, Int], n: Int): IO[Unit] =
+        if (n > 0) q.offer(count - n).flatMap(_ => producer(q, n - 1))
+        else IO.unit
+
+      def consumer(
+          q: PQueue[IO, Int],
+          n: Int,
+          acc: ScalaQueue[Int] = ScalaQueue.empty
+      ): IO[Long] =
+        if (n > 0)
+          q.take.flatMap { a => consumer(q, n - 1, acc.enqueue(a)) }
+        else
+          IO.pure(acc.foldLeft(0L)(_ + _))
+
+      for {
+        q <- constructor(0)
+        p <- producer(q, count).start
+        c <- consumer(q, count).start
+        _ <- p.join
+        v <- c.joinAndEmbedNever
+        r <- IO(v must beEqualTo(count.toLong * (count - 1) / 2))
+      } yield r
+    }
+
+    negativeCapacityConstructionTests(name, constructor)
+    tryOfferOnFullTests(name, constructor, false)
+    cancelableOfferTests(name, constructor)
+    tryOfferTryTakeTests(name, constructor)
+    commonTests(name, constructor)
+
     s"$name - simple order check" in real {
       for {
         q <- constructor(10)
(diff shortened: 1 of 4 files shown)


vasilmkd:
I honestly don't see a problem with not supporting dropping and circular, to me they don't make much sense on a priority queue. I may be wrong though.

TimWSpence:
> I honestly don't see a problem with not supporting dropping and circular, to me they don't make much sense on a priority queue. I may be wrong though.

Thanks @vasilmkd yeah I can't say I'm too bothered either. Just thought I'd call it out explicitly in case anyone was wondering why `Queue` supports them but `PQueue` doesn't

std/shared/src/main/scala/cats/effect/std/PQueue.scala
```diff
@@ -0,0 +1,333 @@
...
+        implicit val Ord = ord
+      }
+
+    /**
+     * Assumes trees is strictly monotonically increasing in rank
+     */
+    @tailrec
+    def insert[A: Order](tree: Tree[A], trees: List[Tree[A]]): List[Tree[A]] =
+      trees match {
+        case Nil => List(tree)
```
johnynek: ```tree :: Nil``` is marginally better than `List(tree)` since the latter uses varargs which allocate.

std/shared/src/main/scala/cats/effect/std/PQueue.scala
```diff
@@ -0,0 +1,333 @@
...
+    def link(other: Tree[A])(implicit Ord: Order[A]): Tree[A] = {
+      if (Ord.lteqv(value, other.value))
+        Tree(rank + 1, value, other :: children)
+      else Tree(rank + 1, other.value, this :: other.children)
+    }
+
+  }
+
+  private def assertNonNegative(capacity: Int): Unit =
+    require(capacity >= 0, s"Bounded queue capacity must be non-negative, was: $capacity")
```
johnynek, line 220: I think writing:
```scala
if (capacity < 0) throw new IllegalArgumentException(...)
else ()
```
is slightly better since you avoid any allocation except in the error case.

std/shared/src/main/scala/cats/effect/std/PQueue.scala
```diff
@@ -0,0 +1,333 @@
...
+import scala.collection.immutable.{Queue => ScalaQueue}
+
+/**
+ * A purely functional Priority Queue implementation based
+ * on a binomial heap (Okasaki)
+ *
+ * Assumes an `Order` instance is in scope for `A`
+ */
+
+abstract class PQueue[F[_], A] { self =>
```
johnynek, line 34: Side note: If we remove the `try` variants l *think* that if `F[_]: Semigroupal` then `PQueue[F, _]: Semigroupal` which could motivate separating this into two classes: one without try, and a subclass with try.

It is kind of interesting to imagine composing these queues as a kind of decompose and fan-out operation.

Also, I wonder if it makes sense to have two separate traits: Offer and Take. Take is covariant, Offer is contravariant. PQueue would extend both to become invariant, but very often you only need one side or the other in some code, so the variance could be useful for those applications: e.g. a consumer that loops reading from one Take and writing to a different Offer.

std/shared/src/main/scala/cats/effect/std/PQueue.scala
```diff
@@ -0,0 +1,333 @@
...
+   *   In fact, a binomial heap built from n elements has is a tree of rank i
+   *   iff there is a 1 in the ith digit of the binary representation of n
+   *   Consequently, the length of the list is <= 1 + log(n)
+   * - Each tree satisfies the heap property (the value at any node is greater
+   *   than that of its parent). This means that the smallest element of the
+   *   heap is found at one of the roots of the trees
+   * - The ranks of the children of a node are strictly monotonically decreasing
+   *   (in fact the rank of the ith child is r - i)
+   */
+  private[std] abstract case class BinomialHeap[A](trees: List[Tree[A]]) { self =>
```
johnynek: I personally think this should be in its own file and have its own scalacheck code since it is so easily tested. We can really hammer on this and make sure it matches a number of nice laws.

std/shared/src/main/scala/cats/effect/std/PQueue.scala
djspiewak: Ref can play the same trick with variance actually. I completely forgot about this trick; I used to use this everywhere years ago. It's very useful in many cases.

std/shared/src/main/scala/cats/effect/std/PQueue.scala
TimWSpence: I just took this from `Queue` but yes, probably

std/shared/src/main/scala/cats/effect/std/PQueue.scala
TimWSpence: Oh that's an interesting idea! I'll definitely have a go at it. Is it worth doing as a separate PR? The existing `Queue` would benefit from the same treatment and then this PR will then be quite big

std/shared/src/main/scala/cats/effect/std/PQueue.scala
TimWSpence: Actually, this is precisely what `require` does I think - it is lazy in the second argument.

std/shared/src/main/scala/cats/effect/std/PQueue.scala
```diff
@@ -0,0 +1,222 @@
...
+import cats.effect.kernel.{Concurrent, Deferred, Ref}
+import cats.effect.std.internal.BinomialHeap
+
+import scala.collection.immutable.{Queue => ScalaQueue}
+
+/**
+ * A purely functional Priority Queue implementation based
+ * on a binomial heap (Okasaki)
+ *
+ * Assumes an `Order` instance is in scope for `A`
```
SystemFw, line 31: I'm not sure about basing this interface on `Order`, vs just taking a priority on `offer`. It seems limiting in cases where you want a dynamic priority

std/shared/src/main/scala/cats/effect/std/internal/BinomialHeap.scala
```diff
@@ -0,0 +1,145 @@
...
+
+package cats.effect.std.internal
+
+import cats.Order
+
+import scala.annotation.tailrec
+
+/**
+ * A binomial heap is a list of trees maintaining the following invariants:
+ * - The list is strictly monotonically increasing in the rank of the trees
```
SystemFw, line 25: I think we should define `rank` precisely here

std/shared/src/main/scala/cats/effect/std/internal/BinomialHeap.scala
```diff
@@ -0,0 +1,145 @@
...
+ *   In fact, a binomial heap built from n elements has is a tree of rank i
+ *   iff there is a 1 in the ith digit of the binary representation of n
+ *   Consequently, the length of the list is <= 1 + log(n)
+ * - Each tree satisfies the heap property (the value at any node is greater
+ *   than that of its parent). This means that the smallest element of the
+ *   heap is found at one of the roots of the trees
+ * - The ranks of the children of a node are strictly monotonically decreasing
+ *   (in fact the rank of the ith child is r - i)
+ */
+private[std] abstract case class BinomialHeap[A](trees: List[BinomialTree[A]]) { self =>
```
SystemFw, line 35: When I wrote [this](https://github.com/SystemFw/upperbound/blob/master/src/main/scala/queues.scala) priority queue, I used `cats-collections` [Heap](https://github.com/typelevel/cats-collections/blob/master/core/src/main/scala/cats/collections/Heap.scala), which in turn is based on [this paper](https://arxiv.org/pdf/1312.4666.pdf%20Vladimir%20Kostyukov's%20paper)

The paper says:
> There are several purely functional implementations of heaps such as Leftist Heap, Binomial Heap (Okasaki, 1999) and Braun Tree (Braun and Rem, 1983), which are definitely the best choice as priority queues in a functional setting. However, there are other heaps around without proper functional implementa- tions. The simplest of them are standard binary heaps, which do not fit well into a functional environment since their reference implementation is based on mutable arrays.

But doesn't actually offer much in terms of evaluation. Any thoughts?

std/shared/src/main/scala/cats/effect/std/PQueue.scala
djspiewak: `require` still allocates though, because it has to allocate the thunk itself!

std/shared/src/main/scala/cats/effect/std/PQueue.scala
djspiewak: Definitely separate PR it. Let's keep it granular. Also 💯 that the existing queue can also benefit from this.

std/shared/src/main/scala/cats/effect/std/PQueue.scala
djspiewak: Is there a scenario where you want a dynamic priority and can't represent it by `Order`? I definitely don't think we want to allow for an effectful priority, since that would mean the structure of the heap would be unstable

core/shared/src/test/scala/cats/effect/std/PQueueSpec.scala
```diff
@@ -0,0 +1,333 @@
...
+  "PQueue" should {
+    boundedPQueueTests("PQueue", PQueue.bounded)
+    boundedPQueueTests("PQueue mapK", PQueue.bounded[IO, Int](_).map(_.mapK(FunctionK.id)))
+  }
+
+  private def boundedPQueueTests(
+      name: String,
+      constructor: Int => IO[PQueue[IO, Int]]): Fragments = {
+
+    s"$name - demonstrate offer and take with zero capacity" in real {
```
djspiewak, line 48: Random testing note: specs allows for better hierarchy here, so you could actually just do something like this:

```scala
name >> {
  "demonstrate offer and take with zero capacity" in real { ...
}
```

And so on, thus putting all of the tests for a single `name` under the same nesting.

std/shared/src/main/scala/cats/effect/std/PQueue.scala
SystemFw: I was thinking simply `offer(a: A, priority: Int)` tbh, in cases where e.g you want to re-enqueue at higher priority. It can still be done with a newtype though, so maybe this is moot

core/shared/src/test/scala/cats/effect/std/internal/BinomialHeapSpec.scala
```diff
@@ -0,0 +1,141 @@
...
+
+import org.scalacheck.{Arbitrary, Gen}, Arbitrary.arbitrary
+
+class BinomialHeapSpec extends Specification with ScalaCheck {
+
+  implicit val orderForInt: Order[Int] = Order.fromLessThan((x, y) => x < y)
+
+  "Binomial heap" should {
+
+    "dequeue by priority" in prop { elems: List[Int] =>
```
djspiewak, line 37: I'm actually surprised this compiles as-is on dotty. Something rather strange is probably going on.

```suggestion
    "dequeue by priority" in prop { (elems: List[Int]) =>
```

And similarly for all such lambdas.

djspiewak (commented):
This seems very nice! I think there are a few places within the `PQueue` implementation where we can eke out some more performance benefits, but this seems good enough to run with for the moment. @johnynek @SystemFw do either of you have any objections to merging this and iterating in follow-up PRs?

SystemFw:
I'm ok with merging :)

std/shared/src/main/scala/cats/effect/std/PQueue.scala
djspiewak: #1395

std/shared/src/main/scala/cats/effect/std/PQueue.scala
djspiewak: #1395

std/shared/src/main/scala/cats/effect/std/PQueue.scala
djspiewak: #1396

johnynek:
Yeah. Merge and iterate seems good.

std/shared/src/main/scala/cats/effect/std/PQueue.scala
TimWSpence: Yeah this seemed more general as you can always recover `priority: Int` with `(Int, A)` and `Order.by(_._1)`

std/shared/src/main/scala/cats/effect/std/internal/BinomialHeap.scala
TimWSpence: Yes, good point. Will raise a quick PR for this 👍

core/shared/src/test/scala/cats/effect/std/PQueueSpec.scala
TimWSpence: Oh nice! I have to admit that I shamelessly stole these tests from `QueueSpec` so we could fix this there as well

std/shared/src/main/scala/cats/effect/std/internal/BinomialHeap.scala
TimWSpence: Oh I actually didn't know about `cats-collections`. I did vaguely look into leftist heaps as well but Okasaki had a nice implementation of binomial heaps so I just went with that to be honest as the asymptotics were the same. I probably should have investigated a bit more whether the other options have lower constant factors

PR #1364, opened 2020-10-27
Added missing Show defaults for IO.print/ln

djspiewak:
I think these were missed when the defaults were added to `Console`.


vasilmkd:
Good catch. Thank you.

vasilmkd (approved):

Issue #1338, opened 2020-10-19
Add a function on Resource that turns it into an `F ~> F`


kubukoz:
It would make the resource wrap an effect in an usage, ignoring the value produced by the resource.

Possible signature:

```scala
trait Resource[+F[_], +A] { self =>
  //...

  //some constraint might be required on G
  def surround[G[x] >: F[x]]: G ~> G = new (G ~> G) { def apply[A](ga: G[A]): G[A] = self.use(_ => ga) }
}
```

SystemFw:
I like the idea but I think that specific signature is cumbersome to use, i.e
```
surround[IO].apply(IO(1))
```
vs
```
surround(IO(1))
```

kubukoz:
I wonder if `.surround.apply(IO(1))` would just work...

bplommer:
How about having `.surround` just take the thing to be surrounded (surroundee?) directly as an argument, and have `.surroundK` to return a `FunctionK`?

kubukoz:
Sounds good to me, that way should be even more ergonomic :)

pkowalcze:
Is it okay if I give this task a try?

RaasAhsan:
@pkowalcze Go to town :)

PR #1368, opened 2020-10-29
surround and surroundK added to Resource

pkowalcze:
As requested in https://github.com/typelevel/cats-effect/issues/1338 - added `surround` and `surroundK` methods on `Resource`

### DIFF
--- a/core/shared/src/test/scala/cats/effect/ResourceSpec.scala
+++ b/core/shared/src/test/scala/cats/effect/ResourceSpec.scala
@@ -470,6 +470,25 @@ class ResourceSpec extends BaseSpec with ScalaCheck with Discipline {
     //   }
     // }
 
+    "surround - should wrap an effect in a usage and ignore the value produced by resource" in ticked {
+      implicit ticker =>
+        val r = Resource.liftF(IO.pure(0))
+        val surroundee = IO(1)
+        val surrounded = r.surround(surroundee)
+
+        surrounded eqv surroundee
+    }
+
+    "surroundK - should wrap an effect in a usage, ignore the value produced by resource and return FunctionK" in ticked {
+      implicit ticker =>
+        val r = Resource.liftF(IO.pure(0))
+        val surroundee = IO(1)
+        val surround = r.surroundK
+        val surrounded = surround(surroundee)
+
+        surrounded eqv surroundee
+    }
+
   }
 
   {
--- a/core/shared/src/test/scala/cats/effect/ResourceSpec.scala
+++ b/core/shared/src/test/scala/cats/effect/ResourceSpec.scala
@@ -473,7 +473,7 @@ class ResourceSpec extends BaseSpec with ScalaCheck with Discipline {
     "surround - should wrap an effect in a usage and ignore the value produced by resource" in ticked {
       implicit ticker =>
         val r = Resource.liftF(IO.pure(0))
-        val surroundee = IO(1)
+        val surroundee = IO("hello")
         val surrounded = r.surround(surroundee)
 
         surrounded eqv surroundee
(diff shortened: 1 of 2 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -417,6 +417,20 @@ sealed abstract class Resource[+F[_], +A] {
...
+   */
+  def surround[G[x] >: F[x]: Resource.Bracket, B >: A](gb: G[B]): G[B] =
+    use(_ => gb)
+
+  /**
+   * Creates a FunctionK that wraps an effect in a usage when applied. Value produced by the resource is ignored.
+   */
+  def surroundK[G[x] >: F[x]: Resource.Bracket]: G ~> G =
+    new (G ~> G) {
+      override def apply[B](gb: G[B]): G[B] = use(_ => gb)
```
pkowalcze: I would like to base the `apply` implementation on `surround`, but unfortunately I'm running into variance issues:

```
[error]  found   : G[B]
[error]  required: G[Any]
[error] Note: B <: Any, but type G is invariant in type x.
[error] You may wish to define x as +x instead. (SLS 4.5)
[error]       override def apply[B](gb: G[B]): G[B] = surround(gb)
```

But `x` is in invariant position. Could you advise?

core/shared/src/test/scala/cats/effect/ResourceSpec.scala
```diff
@@ -470,6 +470,25 @@ class ResourceSpec extends BaseSpec with ScalaCheck with Discipline {
     //   }
     // }
 
+    "surround - should wrap an effect in a usage and ignore the value produced by resource" in ticked {
+      implicit ticker =>
```
pkowalcze, line 474: I'll welcome any suggestions for these tests. Should I go with property-based tests?

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
kubukoz: Would annotating the type parameters on the call to `surround` work?

kubukoz (approved):

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
kubukoz: Wow, turns out this was a good thing.

`surround` required the effect's result to be a supertype of the resource's type, whereas it should be completely irrelevant what the type is. That's why you couldn't use it in `surroundK` - `apply`'s `B` is unrelated to the resource's `A`.

kubukoz: I pushed a fix.

pkowalcze: Oh, that makes sense, thanks!

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -417,6 +417,20 @@ sealed abstract class Resource[+F[_], +A] {
    */
 
   private[effect] def invariant: Resource.InvariantResource[F0, A]
+
+  /**
+   * Wraps an effect in a usage and ignores the value produced by resource.
```
SystemFw: I would change this scaladoc to something like:
```
acquires the resource, runs `gb`, and closes the resource once `gb` terminates, fails or gets interrupted
```

SystemFw (commented):
I suggested a tiny scaladoc improvement, but it's otherwise good to go

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
pkowalcze: Sounds good :+1:  And for `surroundK` would the following be ok?

```
Creates a FunctionK that can run `gb` within a resource, which is then closed once `gb` terminates, fails or gets interrupted
```

pkowalcze: Updated the scaladoc

PR #1371, opened 2020-10-30
Work stealing queue tail does not need to be volatile

vasilmkd:
Playing around with memory barriers.


vasilmkd:
A thousand times this.
http://gee.cs.oswego.edu/dl/html/j9mm.html#rafencesec

djspiewak:
> A thousand times this.

This is a lovely reference honestly.

core/jvm/src/main/java/cats/effect/unsafe/Unsafe.java
```diff
@@ -0,0 +1,78 @@
...
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
+import java.lang.reflect.Field;
+
+final class Unsafe {
+  private static final MethodHandle AcquireFence;
+  private static final MethodHandle ReleaseFence;
+  private static final MethodHandle CompareAndSwapInt;
+  private static final MethodHandle ObjectFieldOffset;
```
djspiewak: This is extremely clever, but is there a particular reason that we need to use `invokedynamic` rather than simply calling `Unsafe` directly? When I've used `Unsafe` in the past, I needed to use reflection to actually get the *instance*, but once I had the instance I could call the members using conventional dispatch. Is this just to maintain compatibility with non-Sun JVMs?

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingQueue.scala
```diff
@@ -50,30 +48,40 @@ private final class WorkStealingQueue {
...
    */
   def isEmpty(): Boolean = {
-    // Should be changed for `acquire` get operations.
-    // Requires the use of `VarHandles` (Java 9+) or Unsafe
-    // (not as portable and being phased out). Cannot use
-    // `getAcquire` and `setAcquire` either (also Java 9+).
-    val hd = lsb(head.get())
+    val fullHead = head
     val tl = tail
+    Unsafe.acquireFence()
```
djspiewak, line 78: If we *don't* have an `Unsafe` instance, then we're going to be in trouble and the pool will actually just not work. There's no graceful fallback.

djspiewak (commented):
This is super-cool. I'm sort of wondering whether we should worry about non-Oracle JVMs. It's difficult to see how we could build a fallback without actually duplicating `WorkStealingPool` almost entirely.

core/jvm/src/main/java/cats/effect/unsafe/Unsafe.java
vasilmkd: You're probably right, this was first copied from Scala's runtime which invokes `VarHandle.releaseFence()`. I decided not to use them because we cannot use varhandles for compare and swap (need a varhandle instance in that case, which cannot be obtained on JDK8) so I just used Unsafe for everything. I will look into it.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingQueue.scala
vasilmkd: We can do a check in IORuntimePlatform, and we could fall back to a fixed pool executor?

vasilmkd:
I believe the concensus in the Java community is that `sun.misc.Unsafe` has become a de-facto Java API and that Oracle is begrudgingly supporting it.

https://blogs.oracle.com/javamagazine/the-unsafe-class-unsafe-at-any-speed

I think all modern Java implementations are forks of OpenJDK, including Oracle's, so we should be ok.

Should we add an openJ9 or some other vendor (amazon corretto) to the build (after I drop some of the Windows ones).

core/jvm/src/main/java/cats/effect/unsafe/Unsafe.java
vasilmkd: Also I was going to benchmark this, but never got around to it due to `race`.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingQueue.scala
vasilmkd: This is not actually the point where it will crash, it's actually in the class initialization where it computes the field offset.

vasilmkd: Ah yeah, maybe I forgot to stress this, failure to obtain the Unsafe instance results in an initialization error.

core/jvm/src/main/scala/cats/effect/unsafe/TreiberStack.scala
```diff
@@ -0,0 +1,47 @@
...
+ * limitations under the License.
+ */
+
+package cats.effect.unsafe
+
+import java.util.concurrent.atomic.AtomicReference
+
+private final class TreiberStack {
+  
+  private[this] val top: AtomicReference[WorkerThread] = new AtomicReference()
```
djspiewak, line 23: Since we already have `Unsafe` in the mix, we can actually do this without the `AtomicReference` if we want to, saving on allocations.

djspiewak (commented):
This is very clever. :-)

core/jvm/src/main/scala/cats/effect/unsafe/TreiberStack.scala
vasilmkd: Yeah, I'm actually looking into a different implementation altogether. This was just to get things running.

CremboC:
Just something to be aware when using `Unsafe` - it might have compatibility issues with Graal `native-image`. I think it depends on methods use though, and how exactly they're used.

vasilmkd:
Thanks! This is a highly experimental branch and will not be merged without being properly tested first.😄

vasilmkd:
Superseded by #1482.

PR #1375, opened 2020-10-31
use Dotty-native build of cats

larsrh:
This currently doesn't work:

* sjs1 artifacts cannot be resolved for Dotty (:eyes: @djspiewak)
* compile error in core (see below)

```
[info] compiling 78 Scala sources and 1 Java source to /home/lars/proj/typelevel/cats-effect/core/jvm/target/scala-0.27/classes ...
[error] undefined: new simulacrum.noop # 216316: TermRef(TypeRef(TermRef(ThisType(TypeRef(NoPrefix,module class <root>)),module simulacrum),noop),<init>) at sbt-api
[warn] sbt-api: Unhandled type class dotty.tools.dotc.core.Types$$anon$7 : <error undefined: new simulacrum.noop # 216316: TermRef(TypeRef(TermRef(ThisType(TypeRef(NoPrefix,module class <root>)),module simulacrum),noop),<init>) at sbt-api>

```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        scala: [0.26.0, 0.27.0-RC1, 2.12.12, 2.13.3]
+        scala: [0.27.0-RC1, 2.12.12, 2.13.3]
         java: [adopt@1.8, adopt@11]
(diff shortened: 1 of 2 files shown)


vasilmkd:
@larsrh Some of this could be my fault, I was desperately trying to get cats-effect to compile on Dotty 0.26.0 which didn't support ScalaJS, looking into this now.

PR #1379, opened 2020-11-02
On jvm IOApp clean exit, optionally print non-daemon threads

Daenyth:
Debugging when a jvm is being held open due to non-daemon threads is a royal
pain. Try to help users here (who might not even know this can happen) by
giving the option to debug, which we can put into a FAQ entry

I opted for making it configurable here so that folks who know their app is fine can disable it if they want to avoid us creating a Daemon thread near shutdown time.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/internals/IOAppPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/internals/IOAppPlatform.scala
@@ -27,7 +27,10 @@ private[effect] object IOAppPlatform {
       // Return naturally from main. This allows any non-daemon
       // threads to gracefully complete their work, and managed
       // environments to execute their own shutdown hooks.
-      ()
+      if (NonDaemonThreadLogger.isEnabled())
+        new NonDaemonThreadLogger().start()
+      else
+        ()
     } else {
       sys.exit(code)
     }
--- a/core/jvm/src/main/scala/cats/effect/internals/NonDaemonThreadLogger.scala
+++ b/core/jvm/src/main/scala/cats/effect/internals/NonDaemonThreadLogger.scala
@@ -32,7 +32,7 @@ private[internals] object NonDaemonThreadLogger {
   def sleepIntervalMillis: Long =
     Option(System.getProperty("cats.effect.logNonDaemonThreads.sleepIntervalMillis"))
       .flatMap(time => Either.catchOnly[NumberFormatException](time.toLong).toOption)
-      .getOrElse(1000L)
+      .getOrElse(10000L)
 }
 
 private[internals] class NonDaemonThreadLogger extends Thread("cats-effect-nondaemon-thread-logger") {
--- a/core/jvm/src/main/scala/cats/effect/internals/NonDaemonThreadLogger.scala
+++ b/core/jvm/src/main/scala/cats/effect/internals/NonDaemonThreadLogger.scala
@@ -1,0 +1,67 @@
+/*
+ * Copyright (c) 2017-2019 The Typelevel Cats-effect Project Developers
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


core/jvm/src/main/scala/cats/effect/internals/IOAppPlatform.scala
```diff
@@ -27,7 +27,10 @@ private[effect] object IOAppPlatform {
       // Return naturally from main. This allows any non-daemon
       // threads to gracefully complete their work, and managed
       // environments to execute their own shutdown hooks.
-      ()
+      if (Option(System.getProperty("cats.effect.logNonDaemonThreadsOnExit")).exists(_.equalsIgnoreCase("true")))
+        logNonDaemonThreads()
```
mpilquist: I think this logging should occur periodically -- e.g., maybe every 10s after completion of the main fiber, with the first log occurring after 10s, not immediately.

core/jvm/src/main/scala/cats/effect/internals/IOAppPlatform.scala
```diff
@@ -27,7 +27,10 @@ private[effect] object IOAppPlatform {
       // Return naturally from main. This allows any non-daemon
       // threads to gracefully complete their work, and managed
       // environments to execute their own shutdown hooks.
-      ()
+      if (Option(System.getProperty("cats.effect.logNonDaemonThreadsOnExit")).exists(_.equalsIgnoreCase("true")))
```
mpilquist: I'd rather default this to on -- it's a common gotcha and folks won't know to enable it without looking in FAQ or getting directed via chat.

core/jvm/src/main/scala/cats/effect/internals/IOAppPlatform.scala
```diff
@@ -63,4 +66,21 @@ private[effect] object IOAppPlatform {
       }
       ()
     }
+
+  private[this] def logNonDaemonThreads(): Unit = {
+    println("IOApp Debug: Checking for non-daemon Threads prior to exit")
```
mpilquist: Minor thing but I'd rather create 1 multi-line string and then make one call to print to standard error.

core/jvm/src/main/scala/cats/effect/internals/IOAppPlatform.scala
```diff
@@ -63,4 +66,21 @@ private[effect] object IOAppPlatform {
...
+    var foundThread = false
+
+    threads.forEach { t =>
+      if (!t.isDaemon()) {
+        foundThread = true
+        println(s" - ${t.getId}: ${t}")
+      }
+    }
+    if (!foundThread)
+      println(" - (no non-daemon threads running)")
```
mpilquist: No need to print at all if there are no active threads.

mpilquist (changes requested):

core/jvm/src/main/scala/cats/effect/internals/IOAppPlatform.scala
arosien: OTOH you don't want to introduce a 10s delay just to end the program.

mpilquist: Right, start a daemon thread that has a `while(true) { Thread.sleep(10000L); logActiveThreads() }`

Daenyth: Oh that's clever. I can see about doing that.

Daenyth:
@mpilquist Thanks for the feedback. I'll try to get that cleaned up this weekend. Feel free to poke me if I don't ;)

mpilquist (approved):

Daenyth:
The re-push was just the fix the "missing headers" CI failure. No changes aside from that

Daenyth:
- Fixed the sleep time being 1sec; made it 10.
- How do I do java collection compat across 2.12+2.13? It looks like collections-compat isn't in scope. Should I add it?

 https://github.com/typelevel/cats-effect/runs/1393002433#step:14:40

Daenyth:
I fixed the collections-compat part.

I think this should be ready for merge now.

I don't think it's a blocker, but I'm wondering how I could possibly write a unit test for this kind of behavior, though. It seems very hard to do. Any suggestions?

djspiewak:
Merged; thank you!

I don't think it needs unit testing. Probably possible, definitely annoying, limited value. :-)

PR #1382, opened 2020-11-03
Add helper for the equivalent of java.time.Instant.now for Clock/IO/SyncIO

jatcwang:
Not 100% happy with the name. suggestion welcome :)


core/js/src/main/scala/cats/effect/SyncIOCompanionPlatform.scala
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+import scalajs.js
+
+private[effect] trait SyncIOCompanionPlatform { this: SyncIO.type =>
+  def jsDateNow: SyncIO[js.Date] =
+    syncForSyncIO.map(syncForSyncIO.realTime)(d => new js.Date(d.toMillis.toDouble))
```
vasilmkd: `realTime` is defined on the `SyncIO` companion object and `map` on `SyncIO`, so you should be able to rewrite this as `realTime.map(d => new js.Date(d.toMillis.toDouble))`.

core/js/src/main/scala/cats/effect/SyncIOCompanionPlatform.scala
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+import scalajs.js
+
+private[effect] trait SyncIOCompanionPlatform { this: SyncIO.type =>
```
vasilmkd, line 21: Use `abstract class` instead of `trait` to match the other platform specific mixins.

core/js/src/main/scala/cats/effect/SyncIOCompanionPlatform.scala
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+import scalajs.js
```
vasilmkd, line 19: Nitpicking, we usually have imports separated by a line from the package definition.

core/jvm/src/main/scala/cats/effect/SyncIOCompanionPlatform.scala
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+import java.time.Instant
+
+private[effect] trait SyncIOCompanionPlatform { this: SyncIO.type =>
+  final def now: SyncIO[Instant] =
+    syncForSyncIO.map(syncForSyncIO.realTime)(d => Instant.ofEpochMilli(d.toMillis))
```
vasilmkd: `realTime.map(...)`

core/jvm/src/main/scala/cats/effect/SyncIOCompanionPlatform.scala
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+import java.time.Instant
+
+private[effect] trait SyncIOCompanionPlatform { this: SyncIO.type =>
```
vasilmkd, line 21: `abstract class` instead of `trait`.

kernel/jvm/src/main/scala/cats/effect/kernel/ClockPlatform.scala
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.kernel
+
+import java.time.Instant
+
+private[effect] trait ClockPlatform[F[_]] { self: Clock[F] =>
+  def instantNow: F[java.time.Instant] =
```
vasilmkd: No need for fully qualified `java.time.Instant`, just `Instant`.

vasilmkd (changes requested):
This is an idea, could we have a single name, e.g. `now` for all of these effects, but they all return a platform specific type (`Instant` on the JVM and `js.Date` on Scala.js)?

cc @djspiewak

vasilmkd:
@jatcwang There were some GitHub Actions changes in the meantime, please rebase your changes onto the current `series/3.x` or merge it in your branch to be able to pass the CI. :)

core/js/src/main/scala/cats/effect/SyncIOCompanionPlatform.scala
jatcwang: Do you know what that's done since a trait would suffice? I tried abstract class first and was running into compile errors (since it needs to be the first after `extends`)

core/js/src/main/scala/cats/effect/SyncIOCompanionPlatform.scala
jatcwang: nw changed

core/jvm/src/main/scala/cats/effect/SyncIOCompanionPlatform.scala
jatcwang: done

kernel/jvm/src/main/scala/cats/effect/kernel/ClockPlatform.scala
jatcwang: oops removed

core/js/src/main/scala/cats/effect/SyncIOCompanionPlatform.scala
jatcwang: fixed

jatcwang (commented):
@vasilmkd thanks for the review. Regarding `now` returning different type in different context: I'm not sure it feels a bit weird that you can type the same code and end up with a different type

vasilmkd:
This code is already platform specific (not portable and you cannot abstract over it) so there is no way to get different behavior/type by using the same code on different platforms. It was just an idea anyway. Congratulations on your first contribution. We'll get this merged soon enough, let's hear more suggestions on the naming.

core/js/src/main/scala/cats/effect/IOCompanionPlatform.scala
```diff
@@ -35,4 +37,6 @@ private[effect] abstract class IOCompanionPlatform { this: IO.type =>
 
   def fromPromise[A](iop: IO[Promise[A]]): IO[A] =
     asyncForIO.fromPromise(iop)
+
+  def jsDateNow: IO[js.Date] = asyncForIO.jsDateNow
```
djspiewak: I would say a better name for this would be `realTimeDate`

core/js/src/main/scala/cats/effect/SyncIOCompanionPlatform.scala
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+
+import scalajs.js
+
+private[effect] trait SyncIOCompanionPlatform { this: SyncIO.type =>
+  final def jsDateNow: SyncIO[js.Date] = realTime.map(d => new js.Date(d.toMillis.toDouble))
```
djspiewak: Ditto `realTimeDate`

core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
```diff
@@ -51,4 +52,6 @@ private[effect] abstract class IOCompanionPlatform { this: IO.type =>
         }
       }
     }
+
+  def instantNow: IO[Instant] = asyncForIO.instantNow
```
djspiewak: I would say `realTimeInstant` here.

core/jvm/src/main/scala/cats/effect/SyncIOCompanionPlatform.scala
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+
+import java.time.Instant
+
+private[effect] trait SyncIOCompanionPlatform { this: SyncIO.type =>
+  final def now: SyncIO[Instant] = realTime.map(d => Instant.ofEpochMilli(d.toMillis))
```
djspiewak: Also `realTimeInstant`

kernel/js/src/main/scala/cats/effect/kernel/ClockPlatform.scala
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package cats.effect.kernel
+
+import scalajs.js
+
+private[effect] trait ClockPlatform[F[_]] { self: Clock[F] =>
+
+  def jsDateNow: F[js.Date] = self.map(self.realTime)(d => new js.Date(d.toMillis.toDouble))
```
djspiewak: `realTimeDate`

kernel/jvm/src/main/scala/cats/effect/kernel/ClockPlatform.scala
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.kernel
+
+import java.time.Instant
+
+private[effect] trait ClockPlatform[F[_]] { self: Clock[F] =>
+  def instantNow: F[Instant] = {
```
djspiewak: `realTimeInstant`

kernel/jvm/src/main/scala/cats/effect/kernel/ClockPlatform.scala
```diff
@@ -0,0 +1,25 @@
...
+    self.map(self.realTime)(d => Instant.ofEpochMilli(d.toMillis))
```
djspiewak: Here's an interesting problem: if `Clock` no longer extends `Applicative`, then we can't actually do this!

djspiewak (changes requested):

djspiewak:
I really like where this is going! Also it's very much in line with ce2, which I believe has `realTimeInstant` or something like that.

core/js/src/main/scala/cats/effect/IOCompanionPlatform.scala
jatcwang: changed.

core/js/src/main/scala/cats/effect/SyncIOCompanionPlatform.scala
jatcwang: changed.

core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
jatcwang: changed.

core/jvm/src/main/scala/cats/effect/SyncIOCompanionPlatform.scala
jatcwang: changed.

kernel/js/src/main/scala/cats/effect/kernel/ClockPlatform.scala
jatcwang: changed.

kernel/jvm/src/main/scala/cats/effect/kernel/ClockPlatform.scala
jatcwang: changed.

kernel/jvm/src/main/scala/cats/effect/kernel/ClockPlatform.scala
jatcwang: Yeah I think I saw the conversation :sweat_smile: Makes sense to remove the constraint but will have to require an applicative instance for the def? I think that's fine in practical usage, though it's a bit weird as it leaks the implementation. lmk and I'll change it

jatcwang:
@djspiewak It was called instantNow :sweat_smile: but I think the common realTime* prefix is good for discoverability.

@vasilmkd thanks for the review. I have yet to make the change for trait -> abstract class as suggested as I'm keen on understanding why we're not using trait (Ensure a linear inheritance chain?)

djspiewak:
> I'm keen on understanding why we're not using trait (Ensure a linear inheritance chain?)

Two reasons, historically! First, class inheritance on the JVM is much faster than interface inheritance. Second, prior to the changes in Scala 2.12, traits imposed significantly more severe binary compatibility constraints than abstract classes. For example, it was impossible to add a new method *implementation* in a super-trait without breaking binary compatibility, whereas abstract classes do not have this issue since they didn't require forwarders in the subtypes. This problem is now resolved, though some other subtleties remain.

In general, you should prefer abstract classes over traits unless there's a strong reason to use traits.

jatcwang:
Thanks for the info Daniel. TIL.

I switched to abstract class but now we're getting issues from the likes of `clockForEitherT`. (See CI for more errors like this)

```
[error] kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala:118:18: illegal inheritance; superclass Object
[error]  is not a subclass of the superclass ClockPlatform
[error]  of the mixin trait OptionTClock
[error]       with Clock.OptionTClock[F] {             
```

So I think I can change  OptionTClock to be `abstract class` but that'll have a cascading effect since e.g. `OptionTTemporal` mixes in `Clock.OptionTClock` (we'll in turn need to change OptionTTemporal to be an abstract class and shuffle it around for any class that extends it).
I think this was the original reason why I didn't use abstract class. If you think it'll work I will give it a go - just want to double check before I jump into the black hole :)

djspiewak:
LOL I would just go back to the `trait`. I don't think anyone wants to diagnose those kinds of compile errors.

core/js/src/main/scala/cats/effect/SyncIOCompanionPlatform.scala
ritschwumm: is this behaving sensible for dates outside the JS date range as mentioned here https://github.com/scala-js/scala-js/issues/4131 ?

core/js/src/main/scala/cats/effect/SyncIOCompanionPlatform.scala
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+
+import scalajs.js
+
+private[effect] trait SyncIOCompanionPlatform { this: SyncIO.type =>
+  final def realTimeDate: SyncIO[js.Date] = realTime.map(d => new js.Date(d.toMillis.toDouble))
```
ritschwumm, line 22: is this behaving sensibly for dates outside the JS date range as mentioned here https://github.com/scala-js/scala-js/issues/4131 ?

jatcwang: It'll behave just like how js Date would behave ("Invalid Date"). The issue you linked is about the behaviour of java.util.Date#getTime which isn't what we're dealing with here.

core/jvm/src/main/scala/cats/effect/SyncIOCompanionPlatform.scala
djspiewak: Was this one missed?

djspiewak (changes requested):
Looks like just one last thing and this is ready! :tada:

core/jvm/src/main/scala/cats/effect/SyncIOCompanionPlatform.scala
jatcwang: Sorry this one fell out of of the rename train. Fixed now.

PR #1391, opened 2020-11-07
Scala 3.0.0-M1

djspiewak:
Also upgraded the Cats dependency, so we're pure-Dotty all the way on the JVM. Unfortunately, this also means I had to disable the ScalaJS builds. We're basically waiting on MUnit for that stuff, and it seems unlikely they will build on dottyjs until 3.0.0-M2.


mpilquist (approved):

larsrh:
The branching here is weird. GitHub tells me you're attempting to merge into the `topic/cats-dotty` branch.

djspiewak:
Whoops. Fixed

vasilmkd (approved):

vasilmkd:
Needs a `scalafmtSbt`.

PR #1392, opened 2020-11-08
Remove race and both specialization

### LINKED ISSUE #1380: AsyncBenchmark.race deadlocks
Running the benchmark for more than 1 iteration does not make any progress. Need to investigate why.

vasilmkd:
Opening to raise awareness of the problem with `racePair` and other race combinators.

Here I just removed the specialization of `both` and `race`, and that leads to failed tests and timeouts, related to cancelation.

Will eventually resolve #1380.

### DIFF
--- a/core/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
+++ b/core/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
@@ -103,7 +103,7 @@ abstract class IOPlatformSpecification extends Specification with ScalaCheck wit
           }
         }
 
-        task.replicateA(100000).as(ok)
+        task.replicateA(1000).as(ok)
       }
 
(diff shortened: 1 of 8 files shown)


djspiewak:
So yeah… this is around 10x faster than the implementation we had before, and unlike that implementation, doesn't appear to deadlock.

core/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
```diff
@@ -103,7 +103,7 @@ abstract class IOPlatformSpecification extends Specification with ScalaCheck wit
           }
         }
 
-        task.as(ok)
+        task.replicateA(100000).as(ok)
```
vasilmkd: Could this be slow in CI?

djspiewak: We're going to find out! It runs in 3 seconds on my laptop

PR #1394, opened 2020-11-08
Updated for Scala 3.0.0-M1

djspiewak:


core/shared/src/test/scala/cats/effect/IOSpec.scala
```diff
@@ -1009,6 +1009,7 @@ class IOSpec extends IOPlatformSpecification with Discipline with ScalaCheck wit
           for {
             ref <- Ref[IO].of(false)
             _ <- IO.sleep(100.millis).onCancel(ref.set(true)).timeoutTo(50.millis, IO.unit)
```
vasilmkd: I think it would be better to change this to `IO.never` instead of `sleep(100.millis)`.

core/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
```diff
@@ -110,8 +110,8 @@ class DispatcherSpec extends BaseSpec {
       }
 
       for {
-        _ <- rec.use(_ => IO.sleep(50.millis)).start
-        _ <- IO.sleep(100.millis) // scope should be closed by now
+        _ <- rec.use(_ => IO.sleep(250.millis)).start
+        _ <- IO.sleep(500.millis) // scope should be closed by now
```
vasilmkd:
```suggestion
        gate <- Deferred[IO, Unit]
        _ <- (rec.use(_ => IO.unit) *> gate.complete(())).start
        _ <- gate.get
```

larsrh: @djspiewak wrote that, so I'll let him decide

PR #1400, opened 2020-11-09
Std dequeue

TimWSpence:


djspiewak:
This is really nice! I like this addition.

PR #1401, opened 2020-11-09
On Windows run only latest stable Scala

vasilmkd:


build.sbt
```diff
@@ -80,6 +80,10 @@ ThisBuild / githubWorkflowBuild := Seq(
 ThisBuild / githubWorkflowBuildMatrixAdditions += "ci" -> List("ciJVM", "ciJS", "ciFirefox")
 
 ThisBuild / githubWorkflowBuildMatrixExclusions ++= {
+  val windowsFilters = (ThisBuild / githubWorkflowScalaVersions).value.filterNot(Set(Scala213)).map { scala =>
+    MatrixExclude(Map("scala" -> scala, "os" -> Windows))
+  }
```
djspiewak: Isn't it easier to do this as a matrix inclusion?

vasilmkd: TIL.

vasilmkd: A little help please:
https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#example-including-new-combinations

results in:
```java.lang.RuntimeException: inclusion key `os` was present in matrix, but value `windows-latest` was not in List(ubuntu-latest)```

File: build.sbt [hunk 2]
```diff
@@ -78,6 +78,10 @@ ThisBuild / githubWorkflowBuild := Seq(
     cond = Some(s"matrix.ci == 'ciJS' && matrix.os == '$PrimaryOS'")))
 
 ThisBuild / githubWorkflowBuildMatrixAdditions += "ci" -> List("ciJVM", "ciJS", "ciFirefox")
+ThisBuild / githubWorkflowBuildMatrixInclusions += MatrixInclude(
+  Map("scala" -> Scala213),
+  Map("os" -> Windows)
+)
```
djspiewak: The trick here is you have to include *all* keys. So you probably want the following:

```suggestion
ThisBuild / githubWorkflowBuildMatrixInclusions += MatrixInclude(
  Map("scala" -> Scala213, "java" -> PrimaryJava, "ci" -> "ciJVM"),
  Map("os" -> Windows)
)
```

djspiewak (changes requested):

PR #1404, opened 2020-11-10
Countdown latch for std

TimWSpence:


core/shared/src/test/scala/cats/effect/std/CountDownLatchSpec.scala
```diff
@@ -0,0 +1,134 @@
...
+import cats.arrow.FunctionK
+import cats.effect.kernel.Outcome.Canceled
+
+import org.specs2.specification.core.Fragments
+
+import scala.concurrent.duration._
+import java.util.concurrent.TimeoutException
+
+class CountDownLatchSpec extends BaseSpec {
+  sequential
```
SystemFw: why are these `sequential`?

std/shared/src/main/scala/cats/effect/std/CountDownLatch.scala
```diff
@@ -0,0 +1,93 @@
...
+  private[std] class ConcurrentCountDownLatch[F[_]](state: Ref[F, State[F]])(
+      implicit F: GenConcurrent[F, _])
+      extends CountDownLatch[F] {
+
+    override def release: F[Unit] =
+      state.modify {
+        case Awaiting(n, signal) =>
+          if (n > 1) (Awaiting(n - 1, signal), F.unit) else (Done(), signal.complete(()).void)
+        case d @ Done() => (d, F.unit)
+      }.flatten
```
SystemFw: let's make this `uncancelable`

std/shared/src/main/scala/cats/effect/std/CountDownLatch.scala
```diff
@@ -0,0 +1,93 @@
...
+
+package cats.effect.std
+
+import cats.~>
+import cats.syntax.all._
+import cats.effect.kernel.{Deferred, GenConcurrent, Ref}
+
+/**
+ * Concurrency abstraction that supports semantically blocking
+ * until n latches are released
```
SystemFw: Worth adding a comment about this being one-shot (in slightly less concise terms :) )

SystemFw (commented):
Looks good, small comments here and there

core/shared/src/test/scala/cats/effect/std/CountDownLatchSpec.scala
TimWSpence: Because I copied the header from `QueueSpec` 😂 I'm not actually sure why it needs `sequential` either?

std/shared/src/main/scala/cats/effect/std/CountDownLatch.scala
TimWSpence: Yep, good call 👍

core/shared/src/test/scala/cats/effect/std/CountDownLatchSpec.scala
SystemFw: yeah I would remove it unless someone else has a reason for it. I vaguely recall one place where it was actually needed, but then I think it spread through copy-paste

std/shared/src/main/scala/cats/effect/std/CountDownLatch.scala
TimWSpence: Oh good point! Just to check my understanding: if we observed cancellation after `modify` and before `flatten` then we would have transitioned to the `Done` state without ever unblocking the fibers by completing the `signal`?

SystemFw: exactly. You can solve it by adding an `onCancel` that checks the state yada-yada, but because `complete` is not a long-lived operation, it's overwhelmingly better to just make the whole thing `uncancelable` (since `modify` is already uncancelable anyway)

core/shared/src/test/scala/cats/effect/std/CountDownLatchSpec.scala
djspiewak: Only the laws tests need `sequential`. Everything else can be in parallel for maximum speeds. :-)

core/shared/src/test/scala/cats/effect/std/CountDownLatchSpec.scala
```diff
@@ -0,0 +1,133 @@
...
+    boundedQueueTests(
+      "CountDownLatch mapK",
+      CountDownLatch.apply[IO](_).map(_.mapK(FunctionK.id)))
+  }
+
+  private def boundedQueueTests(
+      name: String,
+      constructor: Int => IO[CountDownLatch[IO]]): Fragments = {
+
+    s"$name - raise an exception when constructed with negative initial latches" in real {
```
djspiewak: `name should {` :-)

djspiewak (changes requested):
Minor comment but otherwise good!

PR #1407, opened 2020-11-11
Cyclic barrier for std

TimWSpence:


std/shared/src/main/scala/cats/effect/std/CyclicBarrier.scala
```diff
@@ -0,0 +1,108 @@
...
+      F.deferred[Unit].flatMap { newSignal =>
+        F.uncancelable { poll =>
+          state.modify {
+            case State(capacity, current, ids, signal) =>
+              if (current < capacity - 1) {
+                val id = new Proxy()
+                val cleanup = state.update(s => {
+                  val newIds = s.ids.filter(_ != id)
+                  s.copy(current = newIds.size, ids = newIds)
+                })
```
SystemFw: I wonder if this machinery is actually necessary. What is the purpose of tracking the Set of IDs here?

TimWSpence: So you can see my initial attempt here (https://github.com/typelevel/cats-effect/pull/1407/commits/b31d5a486757f7793851814ec30e056b9c6e40b8) which just decrements `current` but I _think_ that has a race between `cleanup` and resetting - if a fiber that is `await`ing is cancelled but then before `cleanup` is run, another fiber `await`s and `current == capacity` then we'll reset to `current == 0` and then `cleanup` will set `current == -1`

If that makes any sense 😂 Basically if we reset after an `await`ing fiber is cancelled then we need the `cleanup` to effectively be a no-op

TimWSpence: And so we need some way to tell if the cancelled fiber is still considered registered with the cyclic barrier or whether it has reset in the meantime.

This was the simplest way I could think of but I'm open to suggestions!

TimWSpence: I should probably document this...

SystemFw: I think we could assign an `epoch` (an incrementing Long) to each cycle, and we only decrement in `cleanup` if the barrier is still in the current epoch

SystemFw: alternatively we could say `cleanup = max(0, size - 1)`, but I need to think it through and see if that actually works, the epoch idea should work though

TimWSpence: Oh interesting, yeah the `epoch` idea is nice 👍

TimWSpence: (and easy to reason about). I agree, `max` might also work but I'm not certain about it at first glance

SystemFw: can you try and isolate the problematic behaviour in a test though? (without the `epoch` change, and just decrementing)

TimWSpence: Yeah that's next on my to-do list 👍

SystemFw (approved):

PR #1409, opened 2020-11-12
Replaced memory barrier with something that's actually sound

djspiewak:
I haven't rewritten the comments yet, this is mostly just to test. I'll rewrite the comments before merge.

Heads up, @RaasAhsan! This is kind of what I was thinking. I thought about your ACK idea but couldn't quite understand what you meant by it, so here's just one proposal. The cost here is actually surprisingly minimal. There's an extra read barrier in `IOCont.Get` and an extra write barrier in `cancel`, but that's it. My understanding is that the JVM can sometimes coalesce barriers so it might not even be that significant of a cost.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -174,7 +174,7 @@ private final class IOFiber[A](
   var cancel: IO[Unit] = IO uncancelable { _ =>
     IO defer {
       canceled = true
-      barrier = true
+      writeBarrier()
 
       // println(s"${name}: attempting cancellation")
 
@@ -881,6 +881,9 @@ private final class IOFiber[A](
          */
       }
     }
+
+  private[this] def writeBarrier(): Unit =
+    barrier = true
 
   private[this] def readBarrier(): Unit = {
     val _ = barrier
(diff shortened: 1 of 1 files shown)


djspiewak:
Actually this doesn't really work for some reason (on arm).

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -439,13 +444,19 @@ private final class IOFiber[A](
...
                 }
-              } else if (!shouldFinalize()) {
-                /*
-                 * If we aren't canceled, loop on `suspended` to wait
-                 * until `get` has released ownership of the runloop.
-                 */
-                loop()
-              } /*
+              } else {
+                readBarrier()
```
RaasAhsan: I think this one might be unnecessary since there's always a read barrier on the `suspended` CAS, even if it fails. (there's still a happens-before edge between `canceled` and `suspended` on `cancel`)

RaasAhsan: ignore this

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -84,6 +84,10 @@ private final class IOFiber[A](
   private[this] var ctxs: ArrayStack[ExecutionContext] = _
 
   private[this] var canceled: Boolean = false
+
+  @volatile
+  private[this] var parity: Boolean = false
```
RaasAhsan, line 89: I tried out my idea and it works but it's a little bit awkward, so I think the dummy barrier (as an `AtomicBoolean` to force the read barrier) or the parity boolean is cleaner. The former is more appealing to me because we don't need two paragraphs to explain it and there isn't much of a performance hit, but it's your call :P

PR #1411, opened 2020-11-12
Implemented an artificial write barrier

djspiewak:
This is only required on architectures where CAS can under some circumstances be a read barrier and *not* a write barrier. For example, on ARM64 without LSE, CAS is non-primitive and, when it fails, will not publish the `canceled` write. This PR adds a check for architecture (at class-load time) and, when we don't *know* that it's safe to rely on CAS for publication, we explicitly add an extra publication. This isn't quite sufficient though because, in some circumstances, both the added publication and the `resume()` might fail, which would still result in a situation where `canceled` is unpublished. This is resolved by looping until we successfully publish.

Notably, this does not affect the efficiency of the *read* side, only the efficiency of the *write* side, and also only on troublesome architectures.

This is an alternative to #1411

### DIFF
--- a/core/jvm/src/main/java/cats/effect/IOFiberConstants.java
+++ b/core/jvm/src/main/java/cats/effect/IOFiberConstants.java
@@ -19,6 +19,9 @@ package cats.effect;
 // defined in Java since Scala doesn't let us define static fields
 final class IOFiberConstants {
 
+  // TODO there will likely be more in here, such as ARM ≥8.1
+  public static final boolean AggressiveCasSemantics = System.getProperty("os.arch").startsWith("x86");
+
   public static final int MaxStackDepth = 512;
 
   // continuation ids (should all be inlined)
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -173,6 +173,16 @@ private final class IOFiber[A](
 
       // println(s"${name}: attempting cancellation")
 
+      // artificial write barrier since resume() may fail
+      // certain architectures define a failed CAS to be a write
+      // barrier (notably x86), and we skip this logic on those
+      val prePublished = if (!AggressiveCasSemantics) {
+        val old = suspended.get()
+        suspended.compareAndSet(old, old)
+      } else {
+        true
+      }
+
       /* check to see if the target fiber is suspended */
       if (resume()) {
         /* ...it was! was it masked? */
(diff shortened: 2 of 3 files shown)


core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -173,6 +173,16 @@ private final class IOFiber[A](
 
       // println(s"${name}: attempting cancellation")
 
+      // artificial write barrier since resume() may fail
+      // certain architectures define a failed CAS to be a write
+      // barrier (notably x86), and we skip this logic on those
+      val prePublished = if (!AggressiveCasSemantics) {
+        val old = suspended.get()
+        suspended.compareAndSet(old, old)
```
RaasAhsan, line 181: Possible race condition: the write barrier on this CAS can be skipped if `suspended` changes from `old`. Similarly, the following `resume()` can fail too if the `suspended` isn't `true` at that point, so nothing would ever get published

core/jvm/src/main/java/cats/effect/IOFiberConstants.java
```diff
@@ -19,6 +19,9 @@
 // defined in Java since Scala doesn't let us define static fields
 final class IOFiberConstants {
 
+  // TODO there will likely be more in here, such as ARM ≥8.1
+  public static final boolean AggressiveCasSemantics = System.getProperty("os.arch").startsWith("x86");
```
RaasAhsan, line 23: I'm not a huge fan of this, mostly because this list is going to have to be exhaustive. I think it's more worthwhile to just respect the JMM

djspiewak: It doesn't have to be exhaustive though! It's a fast-path for architectures that we *know* are safe. The alternative semantics are safe on all arches.

core/shared/src/main/scala/cats/effect/IOFiber.scala
djspiewak: This is why we capture `prePublished`. So if `suspended` flips between `get` and `compareAndSet`, *and* the `resume()` fails, *and* we're not on an aggressive-memory architecture, we loop down below.

RaasAhsan: ahh got it, i missed that part

core/jvm/src/main/java/cats/effect/IOFiberConstants.java
RaasAhsan: fair enough, though i still think we should code against the spec and avoid platform/vm-specific code for as long as possible, unless it's absolutely crucial or the performance wins are more than microscopic. i'm bikeshedding at this point, but avoiding an extra barrier on a `cancel` and or at autocancellation points doesn't seem worth it tbh

SystemFw: > fair enough, though i still think we should code against the spec and avoid platform/vm-specific code for as long as possible, unless it's absolutely crucial or the performance wins are more than microscopic. I

100% this. Let's keep the big picture in mind, and the "don't do weird shit" principle, if at all possible

djspiewak: You guys are absolutely no fun at all. ;-)

PR #1413, opened 2020-11-13
Touchup to CyclicBarrier

RaasAhsan:
made things a bit nicer to read since formatting was a bit messed up

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/CyclicBarrier.scala
+++ b/std/shared/src/main/scala/cats/effect/std/CyclicBarrier.scala
@@ -106,10 +106,7 @@ object CyclicBarrier {
 
   }
 
-  private[std] case class State[F[_]](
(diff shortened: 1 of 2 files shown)


std/shared/src/main/scala/cats/effect/std/CyclicBarrier.scala
```diff
@@ -42,7 +42,7 @@ abstract class CyclicBarrier[F[_]] { self =>
   /*
    * The number of fibers required to trip the barrier
    */
-  def capacity: F[Int]
+  def remaining: F[Int]
```
RaasAhsan, line 45: i think `remaining` is a bit more useful to have, plus we should already know the `capacity` of the barrier because we created it

SystemFw (approved):
Cool, I have a CyclicBarrier PR coming today as well :)

Issue #1363, opened 2020-10-27
Identify and correct runtime issues on ARM


djspiewak:
@vasilmkd ran a test run under ARM and found several issues. We should fix those. We should also find a way to set up a runner for the build which allows us to test under ARM with regularity.

Oh also, to be clear, I think ARM64 should be the priority. Very few real systems run ARM these days, and those which do are pretty safely out of scope. Power-conscious server-side applications, tablets, and laptops (including the forthcoming Apple hardware) are all using the ARM64 architecture, and thus this should be our target.

djspiewak:
For the record, this is surprisingly useful: https://gist.github.com/humbertodias/6237f80df9a4bccf98be298057a82cf2

**Edit** I got more mileage out of this one, though: https://translatedcode.wordpress.com/2016/11/03/installing-debian-on-qemus-32-bit-arm-virt-board/

djspiewak:
As a bullet point, I was able to run the entire test suite (I just ran the JVM) on ARM64 and everything passed. Is it perhaps nondeterministic? I only had two cores on the instance I used.

vasilmkd:
I ran it on a 16 thread aws instance, and I could observe it more often, but they are definitely flaky.

`ContSpec`, `DefaultContSpec` and `DeferredJVMParallelization{1, 2, 4}Spec`.

djspiewak:
> I ran it on a 16 thread aws instance, and I could observe it more often, but they are definitely flaky.

I'll try a larger instance then

djspiewak:
Was able to reproduce the issues with a 16 core setup: https://gist.github.com/7a03365762cd9e361a39d25abb3151bf

djspiewak:
Further update: I've confirmed this is a memory visibility issue. Adding `@volatile` to `canceled` fixes all build errors.

djspiewak:
Here's a relevant SO question suggesting our use of memory barriers is valid: https://stackoverflow.com/questions/29846319/compareandset-memory-effects-of-unsuccessful-operations The test program btw does appear to terminate reliably on ARM64.

RaasAhsan:
Here's my take on this right now. We're making the following assumptions:
1. The ARM64 JVM implementation respects the JMM.
[...]

PR #1416, opened 2020-11-13
One-line fix for memory barrier issues!

djspiewak:
Fix for #1363 suggested by @viktorklang. I seriously had to write down a truth table to convince myself that it works, but it does! The nice thing about `getAndSet` is it's exactly as fast as `compareAndSet` on x86 (and *faster* on ARM), but also deterministically forms a write barrier.


SystemFw:
😮 

Incredible. Maybe worth a quick update to the comments as well

RaasAhsan (approved):
Genius

vasilmkd (approved):

vasilmkd:
Sorry, phones are still stupid.

PR #1417, opened 2020-11-14
Cyclic Barrier follow up

SystemFw:
Few changes:
- Changed the state machine to align with Countdown latch, I think it makes for slightly easier logic, see `awaitingNow == 0` vs `awaiting < capacity - 1`
- Changed the tests to never rely on inspecting the internal state of the barrier
- Removed awaiting and remaining

### DIFF
--- a/core/shared/src/test/scala/cats/effect/std/CyclicBarrierSpec.scala
+++ b/core/shared/src/test/scala/cats/effect/std/CyclicBarrierSpec.scala
@@ -85,11 +85,17 @@ class CyclicBarrierSpec extends BaseSpec {
 
     s"$name - clean up upon cancellation of await" in ticked { implicit ticker =>
       newBarrier(2).flatMap { barrier =>
-        // This should time out, so count goes back to 2
+        // This will time out, so count goes back to 2
         barrier.await.timeoutTo(1.second, IO.unit) >>
         // Therefore count goes only down to 1 when this awaits, and will block again
         barrier.await
       } must nonTerminate
+    }
+
+    s"$name - barrier of capacity 1 is a no op" in real {
+      newBarrier(1)
+        .flatMap(_.await)
+        .mustEqual(())
     }
 
     /*
@@ -100,13 +106,19 @@ class CyclicBarrierSpec extends BaseSpec {
     s"$name - race fiber cancel and barrier full" in real {
       val iterations = 100
 
-      val run = for {
-        barrier <- newBarrier(2)
-        f <- barrier.await.start
-        _ <- IO.race(barrier.await, f.cancel)
-        awaiting <- barrier.awaiting
-        res <- IO(awaiting must beGreaterThanOrEqualTo(0))
-      } yield res
+      val run = newBarrier(2).flatMap { barrier =>
+        barrier.await.start.flatMap { fiber =>
+          barrier.await.race(fiber.cancel).flatMap {
+            case Left(_) =>
+              // without the epoch check in CyclicBarrier,
+              // a late cancelation would increment the count
+              // after the barrier has already reset,
+              // causing this code to never terminate (test times out)
+              (barrier.await, barrier.await).parTupled.void
+            case Right(_) => IO.unit
+          }
+        }
+      }.mustEqual(())
 
       List.fill(iterations)(run).reduce(_ >> _)
     }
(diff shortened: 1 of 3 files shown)


SystemFw:
@TimWSpence FYI

std/shared/src/main/scala/cats/effect/std/CyclicBarrier.scala
```diff
@@ -59,57 +49,48 @@ abstract class CyclicBarrier[F[_]] { self =>
...
+      .map { state =>
+        new CyclicBarrier[F] {
+          val await: F[Unit] =
+            F.deferred[Unit].flatMap { gate =>
+              F.uncancelable { poll =>
+                state.modify {
+                  case State(awaiting, epoch, unblock) =>
+                    val awaitingNow = awaiting - 1
+
+                    if (awaitingNow == 0)
```
RaasAhsan: nice, this invariant is much clearer

core/shared/src/test/scala/cats/effect/std/CyclicBarrierSpec.scala
```diff
@@ -38,106 +33,62 @@ class CyclicBarrierSpec extends BaseSpec {
       CyclicBarrier.apply[IO](_).map(_.mapK(FunctionK.id)))
   }
 
-  private def cyclicBarrierTests(
-      name: String,
-      constructor: Int => IO[CyclicBarrier[IO]]): Fragments = {
-    s"$name - raise an exception when constructed with a negative capacity" in real {
-      val test = IO.defer(constructor(-1)).attempt
-      test.flatMap { res =>
+  implicit class Assertions[A](fa: IO[A]) {
```
RaasAhsan: this would be super useful in `BaseSpec`

core/shared/src/test/scala/cats/effect/std/CyclicBarrierSpec.scala
```diff
@@ -38,106 +33,62 @@ class CyclicBarrierSpec extends BaseSpec {
...
 
-    s"$name - await is blocking" in real {
-      for {
-        cb <- constructor(2)
-        r <- cb.await.timeout(5.millis).attempt
-        res <- IO(r must beLike {
-          case Left(e) => e must haveClass[TimeoutException]
-        })
-      } yield res
+    s"$name - await releases all fibers" in real {
```
RaasAhsan, line 54: some tests use `real` and others use `ticked`, but i can't really tell why :)

RaasAhsan (approved):
very nice! it's pretty cool how each test assumes the results of previous tests and builds upon them

core/shared/src/test/scala/cats/effect/std/CyclicBarrierSpec.scala
SystemFw: I'll move it

core/shared/src/test/scala/cats/effect/std/CyclicBarrierSpec.scala
SystemFw: The ones that use ticked need to assert nontermination or cancelation, and ticked let's you do that much more easily. I used real whenever I could though

RaasAhsan: nontermination makes sense, but how come cancellation is an issue? 

```
s"$name - await is cancelable" in ticked { implicit ticker =>
      for {	      newBarrier(2).flatMap(_.await).timeoutTo(1.second, IO.unit) must completeAs(())
```

i thought this was a good candidate for `real` but i'm probably missing something about how `TC` works

RaasAhsan: oh, is it because cancellation is more or less nondeterministic on the `sleep` when using `real`?

SystemFw: yeah, and `ticked` is deterministic and fast. With `real` you tradeoff slow test vs predictable test. Also these tests don't have any of the things that make `ticked` deadlock-prone (lots of fibers depending on each other) and untrustworthy (real memory barrier things, it's just `Ref + Deferred + uncancelable`)

PR #1430, opened 2020-11-18
Fix windows ci

### LINKED ISSUE #1421: The CI immediately passes on Windows without running the tests
https://github.com/typelevel/cats-effect/pull/1420/checks?check_run_id=1401656157#step:12:1

vasilmkd:
Resolves #1421.

@djspiewak Unfortunately doing matrix include did not run the CI properly on windows. :/
Switching to an explicit os in the matrix with exclusions works fine. I'm sorry.

Dependent on #1429.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
     name: Build and Test
     strategy:
       matrix:
-        os: [ubuntu-latest]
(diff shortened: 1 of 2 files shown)


djspiewak:
That is *definitely* a bug in GH Actions. :-( Unfortunate since this is a lot messier. Oh well.

PR #1437, opened 2020-11-23
Moved unsafe.implicits to shared

domaspoliakas:
Hello 👋 

All this does is move `cats.effect.unsafe.implicits` to be both JVM and JS vs. just being JVM. This is so that cross-platform code could import it for the JVM side, but would also compile on the JS side. At the moment importing it causes the JS side to not compile and one instead has to do `implicit val ioRuntime = IORuntime.global`, which is not a big deal of course but it still seems like a weird gotcha. And having it on the JS side is harmless anyway (as far as I can tell). 

Let me know if there's anything else that I need to do here to get the merge-ready.


djspiewak:
Nice catch; thank you!

PR #1442, opened 2020-11-25
Make AsyncPlatform a trait

joroKr21:
Otherwise we can't have user code extend another class and `Async`

### DIFF
--- a/kernel/js/src/main/scala/cats/effect/kernel/AsyncPlatform.scala
+++ b/kernel/js/src/main/scala/cats/effect/kernel/AsyncPlatform.scala
@@ -18,7 +18,7 @@ package cats.effect.kernel
 
 import scala.scalajs.js.{|, defined, Function1, JavaScriptException, Promise, Thenable}
 
-private[kernel] abstract class AsyncPlatform[F[_]] { this: Async[F] =>
+private[kernel] trait AsyncPlatform[F[_]] { this: Async[F] =>
 
   def fromPromise[A](iop: F[Promise[A]]): F[A] =
     flatMap(iop) { p =>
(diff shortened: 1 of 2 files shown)


djspiewak (approved):

vasilmkd:
I’m curious to know why exactly this is needed. Not against it in any way, just interested.

joroKr21:
> I’m curious to know why exactly this is needed. Not against it in any way, just interested.

Imagine I have a class that implements `GenTemporal`. Then I fix the error to `Throwable` and I can implement `Async`. I want to extend `MyGenTemporal` to reuse the code but I can't. And in general since the contract of a type class is "to be implemented" we shouldn't impose such a hard restriction as not being able to reuse other classes.

vasilmkd:
I get that completely. I'm curious why it is a technical limitation to extend a completely internal abstract class.

joroKr21:
> I get that completely. I'm curious why it is a technical limitation to extend a completely internal abstract class.

Ah you mean why the compiler won't let you? You can't extend two unrelated classes and that is a hard restriction. So a trait that extends a class forces any class extending that trait to also extend that class and taking the slot for any potential third class. I don't know if that sentence makes sense 😄 

But the error you get is something like this:
```scala
[error] ~/zio-interop-cats/interop-cats/shared/src/main/scala/zio/interop/cats.scala:206:91: illegal inheritance; superclass CatsTemporal
[error]  is not a subclass of the superclass AsyncPlatform
[error]  of the mixin trait Async
[error] private class CatsAsync[R <: Clock with Blocking] extends CatsTemporal[R, Throwable] with Async[RIO[R, *]] {
```

vasilmkd:
Very interesting. @djspiewak could we run into this issue again considering that all of our platform specific types are abstract classes?

djspiewak:
> could we run into this issue again considering that all of our platform specific types are abstract classes?

I think that all of our other ones are mixed into classes, so this wouldn't be an issue there, but we should check.

joroKr21:
> could we run into this issue again considering that all of our platform specific types are abstract classes?

I didn't find any other cases

PR #1453, opened 2020-11-27
Prepare upgrade to Scala 3 M2

mpilquist:


build.sbt
```diff
@@ -90,11 +90,7 @@ ThisBuild / githubWorkflowBuildMatrixExclusions ++= {
       MatrixExclude(Map("ci" -> ci, "java" -> java))
     }
 
-    val scalaFilters = crossScalaVersions.value.filterNot(_.startsWith("2.")) map { scala =>
```
mpilquist, line 93: I assume this was to avoid Scala.js+Dotty matrix entries, which are no longer needed.

mpilquist:
We need to fix this line of test-js.sh so parse 3.0.0-Mx versions: https://github.com/typelevel/cats-effect/blob/series/3.x/example/test-js.sh#L15

djspiewak:
Oh boy…

PR #1455, opened 2020-11-27
Raise exception when IOApp main fiber is canceled

### LINKED ISSUE #1451: Raise exception when main fiber is canceled
An application implemented by IO.canceled should result in an exception when run, since nontermination is simply weird and unnecessary behavior here.

wemrysi:
Ensures `IOApp` terminates with an exception if the main fiber is canceled.

Closes #1451

### DIFF
--- a/core/js/src/main/scala/cats/effect/IOApp.scala
+++ b/core/js/src/main/scala/cats/effect/IOApp.scala
@@ -42,7 +42,8 @@ trait IOApp {
       else
         args.toList
 
-    Spawn[IO].raceOutcome[ExitCode, Nothing](run(argList), keepAlive)
+    Spawn[IO]
+      .raceOutcome[ExitCode, Nothing](run(argList), keepAlive)
       .flatMap {
         case Left(Outcome.Canceled()) =>
           IO.raiseError(new RuntimeException("IOApp main fiber canceled"))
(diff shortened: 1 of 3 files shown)


djspiewak:
Very very nice! Thank you!

Random follow-up: any opinions on whether it's worth just pushing this fully into `unsafeRunFiber`? I think that might have some implications for `WorkStealingPool` and even `IOFiber` though.

wemrysi:
@djspiewak 
> whether it's worth just pushing this fully into unsafeRunFiber?

That seems reasonable, though we might want to add a run flavor that makes cancellation more explicit, maybe
```scala
def unsafeRunAsyncOutcome(cb: Outcome[Id, Throwable, A] => Unit)(implicit runtime: unsafe.IORuntime): Unit
```
and then `unsafeRunAsync` would raise an error in the cancellation case, to avoid deadlocks. Or if we don't mind changing the signature of `unsafeRunAsync`, we could expose the `Outcome` there.

djspiewak:
I like the new signature quite a lot! This might require adjusting unsafeRunFiber as well to have a third case in its fold (which avoids the boxing).

Unfortunately, I think we can't entirely avoid having a sentinel exception for the other unsafe functions, simply because those are going to get used, and we don't want the result to be deadlock (as it is now). But so long as we also have the Outcome based signature, people won't need to do horrible things like branch on the sentinel itself.

wemrysi:
> people won't need to do horrible things like branch on the sentinel itself

👍 That's definitely what I wanted to avoid.

PR #1457, opened 2020-11-28
Allow `IOApp` to access the default ExecutionContext

### LINKED ISSUE #748: Consider exposing IOApp's ExecutionContext as `protected`

Daenyth:
Currently it's not possible for users of `IOApp` to use the default
ExecutionContext that backs the ContextShift.

They need to either use `IOApp.WithContext`, or use two unrelated pools, or
override the default `ContextShift`.

This should remove some needless hoop-jumping, I think

Close #748

### DIFF
--- a/core/js/src/main/scala/cats/effect/internals/IOAppPlatform.scala
+++ b/core/js/src/main/scala/cats/effect/internals/IOAppPlatform.scala
@@ -21,6 +21,7 @@ package internals
 import cats.syntax.all._
 import scala.concurrent.duration._
 import scala.scalajs.js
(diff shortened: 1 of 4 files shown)


Daenyth:
> [error] /home/runner/work/cats-effect/cats-effect/core/shared/src/main/scala/cats/effect/IOApp.scala:105:3: Could not find any member to link for "ExecutionContext".

What's the correct way to fix that? I'm not super familiar with scaladoc

djspiewak:
Easiest way to fix it in the short term is to remove the reference and just use the name. Medium term though we should bring in unidoc or write the references ourselves.

PR #1458, opened 2020-11-28
More ParTraverseN 

SystemFw:
Several changes to parTraverseN

- Fixes a bug in MiniSemaphore, which prevented interruption during acquire
- Reverted to simplest implementation of parTraverseN
- Changed encoding of MiniSemaphore to take advantage of `uncancelable/poll`
- Strengthened tests

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
@@ -19,7 +19,7 @@ package cats.effect.kernel
 import cats.{Monoid, Semigroup, Traverse}
 import cats.syntax.all._
 import cats.effect.kernel.syntax.all._
-import  cats.effect.kernel.instances.spawn._
+import cats.effect.kernel.instances.spawn._
 
 import cats.data.{EitherT, IorT, Kleisli, OptionT, WriterT}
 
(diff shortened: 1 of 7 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
```diff
@@ -83,39 +85,21 @@ trait GenConcurrent[F[_], E] extends GenSpawn[F, E] {
...
-  def parSequenceN[T[_]: Traverse, A](n: Int)(
-      tma: T[F[A]])(implicit P: Parallel[F], ev: E <:< Throwable): F[T[A]] =
+  def parSequenceN[T[_]: Traverse, A](n: Int)(tma: T[F[A]]): F[T[A]] =
     parTraverseN(n)(tma)(identity)
 
   /**
    * Like `Parallel.parTraverse`, but limits the degree of parallelism.
+   * Note that the semantics of this operation aim to maximise
+   * fairness: when a spot to execute becomes available, every task
+   * has a chance to claim it, and not only the next `n` tasks in `ta`
```
RaasAhsan, line 95: The order in which tasks start is predetermined (effectively) before any of the tasks execute, as dictated by the order in which they claim a spot in the underlying queue, correct?

RaasAhsan (approved):

kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
SystemFw: Inside the semaphore, yeah

Issue #1460, opened 2020-11-28
Rely on Instant.now instead of System.currentTimeMillis() inside of Clock


sergeykolbasov:
Hello there

At the moment, `Clock[F].realTime` is platform-independent and relies on `System.currecurrentTimeMillis()`. It means, that the maximum precision for getting wall-clock time is limited to milliseconds.

I'd suggest having a JVM-specific implementation that relies on `java.time.Instant.now()` instead, as starting from Java 9 it's up to the nanoseconds precise time resolution.

An example, where such precision might matter, is OpenTracing API that operates in microseconds.

djspiewak:
I like this idea!

sergeykolbasov:
I can land a PR later if it's okay

djspiewak:
And I can merge said PR, if that's okay. 😉

SystemFw:
Shouldn't we add an additional op that returns Instant?

sergeykolbasov:
~@SystemFw I think it would mean some troubles with scala.js and lack of `Instant` type there~

It already does 🤔 , although it builds it from millis

SystemFw:
Yeah I'm thinking whether it's worth introducing a platform specific op.
But maybe not :)

RaasAhsan:
IIRC the trouble there was that it bloats the scala.js builds because it pulls in some pretty heavy dependencies ?

PR #1461, opened 2020-11-28
Use platform-dependent implementation of Clock.realTime

sergeykolbasov:
As discussed in #1460

### DIFF
--- a/core/js/src/main/scala/cats/effect/internals/DefaultClock.scala
+++ b/core/js/src/main/scala/cats/effect/internals/DefaultClock.scala
@@ -16,23 +16,11 @@
 
 package cats.effect.internals
 
-import java.time.Instant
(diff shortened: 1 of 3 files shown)


sergeykolbasov:
@djspiewak Hello, any thoughts on this?

djspiewak:
Sorry! This somehow fell off my radar in the shuffle. This looks very good! Would you be interested in looking into bringing these improvements to the `series/3.x` branch as well?

sergeykolbasov:
No problem.

Sure can do for 3.x, although last time I checked there was no actual implementation of clocks there. Is it supposed to be the same?

djspiewak:
Clock is defined as a typeclass, and the implementation is in IO and SyncIO (in CE3). 😃 Just a reorganization, not a removal.

PR #1462, opened 2020-11-29
Remove duplicate test from `RefSpec`

davidabrahams:
I believe this test is a copy-paste duplicate of the one right above it, so it should be safe to remove?


SystemFw (approved):

djspiewak:
Nice catch!

PR #1465, opened 2020-11-29
Implement Clock.applicative in Sync and GenTemporal

joroKr21:
I don't see a reason to keep it abstract. Followup #1419


djspiewak:
This was just an oversight on my part. Thanks!

PR #1466, opened 2020-11-29
Circular buffer for parked worker threads and removed lock

vasilmkd:
Corroborated by benchmark improvements

WorkStealingBenchmark.async before:
`10.660 +- 0.529`

WorkStealingBenchmark.async after:
`11.078 +- 0.218`

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/ConcurrentCircularBuffer.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/ConcurrentCircularBuffer.scala
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2020 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


core/jvm/src/main/scala/cats/effect/unsafe/ConcurrentCircularBuffer.scala
```diff
@@ -0,0 +1,69 @@
...
+    while ({
+      hd = head.get()
+      tl = tail.get()
+
+      if (hd == tl)
+        return null
+
+      idx = hd
+      !head.compareAndSet(hd, hd + 1)
+    }) ()
```
djspiewak, line 62: Officially the most hilarious `while` loop formulation I've ever seen. I know it's just encoding `do`/`while`, but still.

djspiewak (approved):
Very very nice

core/jvm/src/main/scala/cats/effect/unsafe/ConcurrentCircularBuffer.scala
vasilmkd: This is the recommended way to do it in the Dotty docs, just fyi. :)

PR #1467, opened 2020-11-30
Non-volatile sleeping flag

vasilmkd:
The flag need not be volatile because it will be published by the `LockSupport.park` and `LockSupport.unpark` operations, which form a full memory fence.

Also verified on ARM, thanks to @djspiewak.


RaasAhsan:
> The flag need not be volatile because it will be published by the LockSupport.park and LockSupport.unpark operations, which form a full memory fence.

Are there any documentation/code references for this ? `LockSupport` is a JDK-specific class so it has no special treatment in the memory model, and IIRC there is also no mention of memory semantics for it in the informal javadoc either

vasilmkd:
Thanks @RaasAhsan, you are right. Switching this PR to draft, I still need to find where the synchronization actually takes place.

PR #1470, opened 2020-11-30
Non-blocking external queue

vasilmkd:
It seems there's no reason to be too clever. This also reduces the footprint of `IOFiber`.

Benchmark results:

```
WorkStealingBenchmark.async before:
10.433 +- 0.474
```

```
WorkStealingBenchmark.async after:
11.193 +- 0.429
```


djspiewak:
It actually surprises me that this is so much faster. I wonder what witchcraft `ConcurrentLinkedQueue` is doing.

PR #1476, opened 2020-12-02
Fix migration warnings on Scala 3

vasilmkd:
Just for fun.


djspiewak:
I'm definitely in favor. :-)

PR #1477, opened 2020-12-03
Add Unique to Std

ChristopherDavenport:
Unique, I opted to make it `Concurrent` which should be safe. But I can see the argument for Sync.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Unique.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Unique.scala
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2020 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package cats
 package effect
 package std
--- a/std/shared/src/main/scala/cats/effect/std/Unique.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Unique.scala
@@ -19,7 +19,7 @@ package effect
 package std
 
 import cats.Hash
-import cats.effect.Sync
+import cats.effect.kernel._
 import cats.syntax.all._
 
 final class Unique private extends Serializable {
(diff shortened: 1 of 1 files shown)


SystemFw:
> Unique, I opted to make it Concurrent which should be safe

Nice, I was going to come out and propose Concurrent, just too useful :)

SystemFw (approved):

kubukoz:
Do we have any guarantee that an effect system won't optimize `unit.map(_ => a)` as `pure(a)`? (or just `pure(a).map(f)` as `pure(f(a))`)

std/shared/src/main/scala/cats/effect/std/Unique.scala
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * Alumnus of the Davenverse
+ */
+final class Unique private extends Serializable {
+  override def toString: String = s"Unique(${hashCode.toHexString})"
+}
+object Unique {
+  def apply[F[_]: Concurrent]: F[Unique] = Concurrent[F].unit.map(_ => new Unique)
+
+  def sync[F[_]: Sync]: F[Unique] = Sync[F].delay(new Unique)
```
kubukoz, line 45: I'd rather have `Unique.Mk` similar to the Ref/Deferred ones, then even the standard implementation can be with `Sync`.

SystemFw: The reason not to have it is the same why we avoided `Mk` for `Ref/Deferred` as the main way of abstraction. It's a very low level implementation detail that pollutes everything, and the main use case for `Unique` (at least for me) it's been using it for concurrent abstractions.

Also, if the only way to create `Unique` is `Sync`, fs2 cannot use it, basically, since we want to be able to compile under `Concurrent` only. It's definitely on the hacky side though, I can see that :)

kubukoz: Yeah, I see your point - if someone wants to make a Unique they'll likely need Concurrent too...

std/shared/src/main/scala/cats/effect/std/Unique.scala
```diff
@@ -0,0 +1,44 @@
...
+ * Unique is a unique identifier.
+ *
+ * The default constructor takes [[Concurrent]], so it can be used
+ * alongside Ref+Deferred, and is safe due to lazy map on anything
+ * with a concurrent instance.
+ *
+ * Alumnus of the Davenverse
+ */
+final class Unique private extends Serializable {
+  override def toString: String = s"Unique(${hashCode.toHexString})"
```
isomarcte, line 40: It might be worth noting in the Scaladoc that a `Unique` value is _only_ guaranteed to be unique in terms of the reference equality comparison. That is, two `Unique` values created with different effects can have to same `.hashCode` and `.toString`, even though they compare as non-equal.

```scala
scala> res7
val res17: cats.effect.std.Unique = Unique(2c0e55fe)

scala> res8
val res18: cats.effect.std.Unique = Unique(2c0e55fe)

scala> res7 == res8
val res19: Boolean = false

scala> res7.hashCode == res8.hashCode
val res20: Boolean = true

scala> res7.toString == res8.toString
val res21: Boolean = true
```

On JRE 15 with default settings I hit a collision for hashCode usually between 10K-100K invocations.

PR #1480, opened 2020-12-04
Trim some IOFiber mutable fat

vasilmkd:
Should result in a slightly slimmer fiber footprint.


djspiewak:
Very nice! I didn't think of using `objectState` for this but it's really obvious in retrospect :-D

vasilmkd:
Me neither, I facepalmed yesterday while looking for something else.

djspiewak:
LOL

PR #1484, opened 2020-12-06
Handle self-cancellation when running fibers to avoid deadlocks

wemrysi:
As discussed in #1455, we'd like to handle self-cancellation when running fibers to avoid deadlocks. This introduces `IO#unsafeRunAsyncOutcome` to explicitly expose the case where the running fiber the caller started was canceled. The existing `unsafeRun*` variants now raise a `CancellationException` if they are canceled, ensuring the callback is always invoked on termination.

A result of this change is that cancellation and non-termination are now distinguished in tests. This distinction resulted in a failure of [`GenSpawnLaws#uncancelableRaceDisplacesCanceled`](https://github.com/typelevel/cats-effect/blob/349fc13c6a6679ebd71168901874ee332e37ff5b/laws/shared/src/main/scala/cats/effect/laws/GenSpawnLaws.scala#L130) which, upon inspection, seems like it should be expecting `F.never` instead of `F.canceled`.

Looking at the [other `uncancelableRace*`](https://github.com/typelevel/cats-effect/blob/349fc13c6a6679ebd71168901874ee332e37ff5b/laws/shared/src/main/scala/cats/effect/laws/GenSpawnLaws.scala#L133) laws, they seem to imply the mask status of a fiber is inherited by fibers it spawns, is that the intended semantic? It doesn't appear to be the current semantic of `IO`, as illustrated by
```scala
var flag = 0
val flagger = IO.canceled.onCancel(IO { flag = -1 }).flatMap(_ => IO { flag = 1 })

// A
flag = 0
flagger.unsafeRunSync() // flag = -1

// B
flag = 0
IO.uncancelable(_ => flagger).unsafeRunSync() // flag = 1

// C
flag = 0
// unsafeRunTimed as both sides of the race are canceled which is defined to be non-termination when the race is masked
IO.uncancelable(_ => IO.race(flagger, IO.canceled)).unsafeRunTimed(1.second) // flag = -1

// D
flag = 0
// from uncancelableRacePollCanceledIdentityRight
IO.uncancelable(p => IO.race(flagger, p(IO.canceled))).unsafeRunTimed(1.second) // flag = -1
```
Generally, any time `fa` is canceled, the lhs of those laws will not terminate (by definition of `GenSpawn#race`) whereas the rhs will terminate as canceled (and prior to this PR, the tests considered those outcomes equal). Even if we take that into account via `unsafeRunTimed`, the laws stay that `D <-> B` when actually `D <-> A`.

Thoughts? Have I overlooked something or made poor assumptions?


djspiewak:
> and prior to this PR, the tests considered those outcomes equal

Wow, I didn't realize this. This is a rather significant oversight in the test harnessing.

Working back to your earlier points…

> Looking at the other uncancelableRace* laws, they seem to imply the mask status of a fiber is inherited by fibers it spawns, is that the intended semantic?

That's definitely not the intended semantic *anymore*. There was a time where it was what we were shooting for. `race` is now intended to simply be a mechanism for starting and simultaneously joining a pair of fibers, which is why it is immediately non-primitive within `IO`. It's possible though that this was an oversight caused by the fact that the tests just… never started failing.

djspiewak (approved):
This looks really really nice.

> seems like it should be expecting F.never instead of F.canceled.

You're correct! Documentation and implementation agree:

```scala
            case Outcome.Canceled() =>
              poll(f.join).onCancel(f.cancel).flatMap {
```

I believe that the laws should be corrected.

laws/shared/src/main/scala/cats/effect/laws/AsyncLaws.scala
```diff
@@ -26,11 +26,13 @@ import scala.util.{Left, Right}
 trait AsyncLaws[F[_]] extends GenTemporalLaws[F, Throwable] with SyncLaws[F] {
   implicit val F: Async[F]
 
-  def asyncRightIsSequencedPure[A](a: A, fu: F[Unit]) =
-    F.async[A](k => F.delay(k(Right(a))) >> fu.as(None)) <-> (fu >> F.pure(a))
+  def asyncRightIsUncancelableSequencedPure[A](a: A, fu: F[Unit]) =
```
wemrysi, line 29: My rationale for the `uncancelable` is that `fu` is sequenced when `async` is evaluated, but its cancellation status doesn't affect the outcome.

laws/shared/src/main/scala/cats/effect/laws/GenSpawnLaws.scala
```diff
@@ -127,16 +123,8 @@ trait GenSpawnLaws[F[_], E] extends MonadCancelLaws[F, E] {
   def neverDominatesOverFlatMap[A](fa: F[A]) =
     F.never >> fa <-> F.never[A]
 
-  def uncancelableRaceDisplacesCanceled =
-    F.uncancelable(_ => F.race(F.never[Unit], F.canceled)).void <-> F.canceled
-
-  def uncancelableRacePollCanceledIdentityLeft[A](fa: F[A]) =
```
wemrysi, line 133: These no longer seemed pertinent given the current desired semantics of `race`.

wemrysi:
@djspiewak Ready for another look. I've corrected or removed (justification in comments above) some of the laws that began failing due to distinguishing between non-termination and failure or to the presence of previously ungenerated constructs.

There are other failing laws/tests that I've commented out with a `FIXME` so that the build is clean, allowing this PR to be merged so others can benefit from the changes to the testing environment, if desired.

The remaining offenders are
* `IOSpec`: "round trip through s.c.Future"
* jvm/`IOPlatformSpecification`: "round trip through j.u.c.CompletableFuture"
* `GenSpawnTests`
  * "async.race canceled identity (left)"
  * "async.race canceled identity (right)"
* `MemoizeSpec`: "Concurrent.memoize and then flatten is identity"

the `Future` ones fail as we don't distinguish cancellation when converting to futures, though the tests seem like they'd still be worthwhile limited to uncancelable values (maybe we add some additional test that assert cancellation converts as an error). The others I haven't had a chance to investigate thoroughly yet.

djspiewak:
Sorry for the delay on getting to this! It's been a hell of a week… Churning through it now. Also *thank you* for taking this on!

laws/shared/src/main/scala/cats/effect/laws/AsyncLaws.scala
djspiewak: Actually I think it's cancelation status still affects the outcome. `fu = F.canceled` would still imply `F.canceled` on both sides. Does this actually produce a different test outcome?

laws/shared/src/main/scala/cats/effect/laws/GenSpawnLaws.scala
```diff
@@ -87,18 +87,14 @@ trait GenSpawnLaws[F[_], E] extends MonadCancelLaws[F, E] {
...
 
+  // FIXME fails when fa is IO.Uncancelable(...)
   def raceCanceledIdentityRight[A](fa: F[A]) =
     F.race(fa, F.canceled) <-> fa.map(_.asLeft[Unit])
 
-  def raceNeverIdentityLeft[A](fa: F[A]) =
-    F.race(F.never[Unit], fa) <-> fa.map(_.asRight[Unit])
-
-  def raceNeverIdentityRight[A](fa: F[A]) =
-    F.race(fa, F.never[Unit]) <-> fa.map(_.asLeft[Unit])
```
djspiewak, line 100: These actually seem like pretty important laws. I take it that they fall apart when `fa = F.canceled`?

laws/shared/src/main/scala/cats/effect/laws/GenSpawnLaws.scala
djspiewak: Agreed.

laws/shared/src/main/scala/cats/effect/laws/GenSpawnLaws.scala
```diff
@@ -127,16 +123,8 @@ trait GenSpawnLaws[F[_], E] extends MonadCancelLaws[F, E] {
...
-
-  def uncancelableRacePollCanceledIdentityLeft[A](fa: F[A]) =
-    F.uncancelable(p => F.race(p(F.canceled), fa)) <-> F.uncancelable(_ =>
-      fa.map(_.asRight[Unit]))
-
-  def uncancelableRacePollCanceledIdentityRight[A](fa: F[A]) =
-    F.uncancelable(p => F.race(fa, p(F.canceled))) <-> F.uncancelable(_ =>
-      fa.map(_.asLeft[Unit]))
+  def uncancelableRaceNotInherited =
+    F.uncancelable(_ => F.race(F.never[Unit], F.canceled)).void <-> F.never[Unit]
```
djspiewak, line 127: I'd rather be a bit more general here:

```scala
F.uncancelable(_ => F.race(fa, fb)) <-> F.race(fa, fb).onCancel(F.never)
```

project/plugins.sbt
```diff
@@ -1,6 +1,6 @@
 libraryDependencies += "org.scala-js" %% "scalajs-env-selenium" % "1.1.0"
 
-addSbtPlugin("com.codecommit"     % "sbt-spiewak-sonatype"     % "0.18.3")
+addSbtPlugin("com.codecommit"     % "sbt-spiewak-sonatype"     % "0.19.2")
```
djspiewak: 🎉 Not having scala steward on branches is pretty painful.

laws/shared/src/main/scala/cats/effect/laws/GenSpawnLaws.scala
wemrysi: Yeah, exactly.

laws/shared/src/main/scala/cats/effect/laws/AsyncLaws.scala
wemrysi: You're right, good call, looks like I misinterpreted the failures before. Running the original laws again shows they fail when one side doesn't terminate and the other ends up as `canceled`. I'll have to investigate that more.

laws/shared/src/main/scala/cats/effect/laws/GenSpawnLaws.scala
djspiewak: We can fix this by using `.onCancel(F.never)` again. That's not *quite* an identity, and it loses some generality, but it does allow us to talk about task completion and `race` once again.

mpilquist (approved):

djspiewak:
@wemrysi I think we're going to merge this aggressively (since it reveals a bunch of bugs which are currently invisible) and follow-up with further law refinements and fixes. Gucci?

wemrysi:
@djspiewak I was planning to fix the two that you called out in review later today, want to wait until those are done?

djspiewak:
@wemrysi I'm mostly thinking of the fact that we want to get another milestone out with support for 3.0.0-M3, and I'd rather fold your changes into it. Since the follow-ups are bug fixes (#1519) and law adjustments, it feels like we can release the milestone without them, but if you feel it's worth a bit of delay I'm cool with it unless @mpilquist rains fire on us.

wemrysi:
@djspiewak Ok, that's fine. I can address your comments in a followup PR.

laws/shared/src/main/scala/cats/effect/laws/AsyncLaws.scala
wemrysi: Looking into this a bit more, I think this change is actually correct if we want to keep the current semantic that the effect returned from the async callback
```scala
(Either[Throwable, A] => Unit) => F[Option[F[Unit]]]
```
is uncancelable (e.g. consider `fu = F.canceled >> F.never`).

I think my use of the term "outcome" in the OP caused confusion. `fu` does affect the resulting `Outcome`, but not the "outcome" in the sense that cancellation is suppressed.

PR #1488, opened 2020-12-08
Supervisor

RaasAhsan:
First pass at Supervisor. The approach here was to implement it with the usual composable concurrency patterns.

TODO:
- [x] Tests
- [x] Think about exposing cancellation
- [x] Address all the TODOs

I didn't spend too much time on it, but I started implementing a `Dispatcher` in terms of `Supervisor`. It's in the PR but still incomplete (pending whether we even decide to expose cancellation in `Supervisor`). Unfortunately, we still have to cross an impure boundary to get from `Future` back into `F`, which means we have to do the whole separate fiber + async switching trick again, which then means that we have to reimplement the whole queueing/cancellation mechanism again.

Using `Supervisor` will probably only save us ~15 lines of code (active fiber management), at the overhead of an extra dispatcher fiber (in addition to the supervisor fiber) plus the extra dispatcher queue (in addition to the supervisor queue). I'm sure microbenchmark performance would suffer a lot, so maybe it's a better idea just to keep these implementations separate

Alternatively, we could write `Supervisor` in terms of `Dispatcher` (or `Dispatcher` could subtype `Supervisor`) but this might be overkill since we'd have to switch to `Future` and then back to `F`.


std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -44,6 +44,173 @@ object Dispatcher {
 
   private[this] val Open = () => ()
 
+  def supervisor[F[_]](implicit F: Async[F]): Resource[F, Dispatcher[F]] = {
```
RaasAhsan: To help readers out... this is the implementation that uses `Supervisor` and the majority of it is the same as the default implementation

std/shared/src/main/scala/cats/effect/std/Supervisor.scala
```diff
@@ -0,0 +1,105 @@
...
+ */
+
+package cats.effect.std
+
+import cats.effect.kernel._
+import cats.effect.kernel.implicits._
+import cats.syntax.all._
+
+trait Supervisor[F[_], E] {
+  def supervise[A](fa: F[A]): F[F[Outcome[F, E, A]]]
```
RaasAhsan: I think we could just expose `Fiber` here

std/shared/src/main/scala/cats/effect/std/Supervisor.scala
```diff
@@ -0,0 +1,105 @@
...
+      activeRef <- Resource.make(F.ref(Set[Fiber[F, E, _]]())) { ref =>
+        ref.get.flatMap { fibers => fibers.toList.parTraverse_(_.cancel) }
+      }
+      aliveRef <- Resource.make(F.ref(true)) { ref =>
+        for {
+          _ <- ref.set(false)
+          state <- stateRef.get
+          // report canceled back for effects that weren't started
+          _ <- state.registrations.parTraverse_ {
+            case (_, join) => join.complete(F.pure(Outcome.canceled))
```
RaasAhsan: I'm wondering if we need to do something similar in `Dispatcher`. Right now, effects which are queued but never spawned because the `Dispatcher` fiber was cancelled will never terminate

std/shared/src/main/scala/cats/effect/std/Supervisor.scala
```diff
@@ -0,0 +1,140 @@
...
+        override def supervise[A](fa: F[A]): F[F[Outcome[F, Throwable, A]]] =
+          for {
+            alive <- aliveRef.get
+            _ <-
+              if (alive) F.unit
+              else F.raiseError(new IllegalStateException("supervisor has shutdown"))
+
+            token <- counterRef.updateAndGet(_ + 1)
+            joinDef <- Deferred[F, F[Outcome[F, Throwable, A]]]
+            reg = Registration(token, fa, joinDef)
```
RaasAhsan: we could avoid existentials here by completing the deferred in a `guaranteeCase`, but I thought it was creative :D

RaasAhsan: actually, there's a better case for existentials here: the `Supervisor` needs to be able to cancel tasks that were enqueued but not yet started, which involves sending back a `F.canceled` which is predicated over the result type. one way to get around is by explicitly passing in the canceled action in the registration

RaasAhsan:
Pushed a way simpler implementation, it's ridiculously simple... The problem was that I was basing this off `Dispatcher` but we have much stricter guarantees when we're confined to the IO runtime. I'll go back and see what `Dispatcher` looks like when using a `Supervisor` under the hood

Also, I think there might be a more general abstraction here -- basically, allocate a `Resource` to some enclosing scope that is already active. This may be somewhat similar to scopes in fs2, but confined to `IO` land

oleg-py:
Interestingly, I've talked about this and Scope constructs in my [exercise blog post](https://olegpy.com/resource-exercises/). The key difference between this implementation and what I suggested (based on [this article about Python](https://vorpus.org/blog/notes-on-structured-concurrency-or-go-statement-considered-harmful/)) is that the supervisor _joins_ fibers instead of cancelling them when it goes out of scope.

djspiewak:
@RaasAhsan An interesting follow-up thing to think about: should we make this `Supervisor[F, E]` so that we can use `GenConcurrent` instead?

RaasAhsan:
Yeah, I avoided it in this PR since it forces callers to explicitly
annotate E in most cases. We can't get away with an existential like we did
for Deferred since we are returning Fibers. We should probably figure this
out before stable since changing it will probably be breaking

PR #1490, opened 2020-12-09
Remove buggy circular buffer

vasilmkd:
I shouldn't be so careless when implementing highly complex data structures. The implementation of the circular buffer could allow the head to overtake the tail, which results in no worker threads being woken up ever again. This bit me in long running programs and benchmarks.

I apologize.


djspiewak:
I'm actually really surprised it was possible to overflow the circular buffer. After all, you're only ever going to have `threadCount` threads to worry about, so how exactly could you end up overtaking?

vasilmkd:
I've been working on too many things in parallel, I will probably give this another go at a later date. I only wanted to get this change in so that we can have a new definitely safe release if needed.

PR #1491, opened 2020-12-09
Uncancelable_

TimWSpence:
I thought this pattern was common enough to deserve a dedicated method?

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -313,6 +313,9 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
   def uncancelable[A](body: Poll[IO] => IO[A]): IO[A] =
     Uncancelable(body)
 
+  def uncancelable_[A](body: => IO[A]): IO[A] =
+    uncancelable(_ => body)
+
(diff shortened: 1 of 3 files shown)


RaasAhsan:
btw `MonadCancel` syntax includes an `uncancelable` method which behaves like this, should that one be renamed as well?

TimWSpence:
> btw `MonadCancel` syntax includes an `uncancelable` method which behaves like this, should that one be renamed as well?

Oh yep, good call! 👍

TimWSpence:
@RaasAhsan oh actually, I see what you mean. My gut is that `fa.uncancelable` looks better than `fa.uncancelable_` but I'm happy to be persuaded otherwise!

RaasAhsan:
@TimWSpence yeah I'm not really sure here either, but thought it was worth bringing up :)

TimWSpence:
@RaasAhsan we'll wait for someone else to weigh in then 😂 I realized I should have defined it directly on `IO` as well though so I've done that now

djspiewak:
IMO I prefer `fa.uncancelable` to `uncancelable_(fa)`. :-) I think whenever you don't want to use the `syntax`, `uncancelable(_ => fa)` isn't really that onerous.

PR #1495, opened 2020-12-10
Reintroduce IO.Par[A] as an alias in the IO companion object.

arosien:
Could help source compatibility issues with CE2.


djspiewak:
Good idea!

PR #1496, opened 2020-12-11
Scope

RaasAhsan:
A resource scoping data structure. This came out of the `Supervisor` PR (#1488) where I realized that we were building a scoping mechanism, but targeted towards `Fiber`. It would've been nice to use this in `Supervisor` but handling `Fiber`s correctly requires a bit more work to avoid leaks, though you can still use this with `background` (consider what the difference is). Alternatively, if we don't expose the `Fiber` handle in `Supervisor`, we _can_ use `Scope` + `background` to do this.

The idea here may be conceptually similar to FS2 scopes

TODO:
- [ ] Tests
- [ ] Better documentation

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Scope.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Scope.scala
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2020 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.std
+
+import cats.syntax.all._
+import cats.effect.kernel.implicits._
+import cats.effect.kernel.{Async, Resource}
+
+/**
+ * A [[Scope]] represents an effectful region in which resources can be 
+ * allocated or bound. Whenever this scope exits, all resources which were
+ * bound to it are released in the opposite order of allocation.
+ * 
+ * The following example illustrates the ordering of resource lifecycles when
+ * bound to a scope:
+ * 
+ * {{{
+ * 
+ *   Scope[IO].use { scope =>
+ *     for {
+ *       _ <- scope.allocate(Resource.make(IO.println("1"))(_ => IO.println("2")))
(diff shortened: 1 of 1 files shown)


djspiewak (commented):
This is super awesome. Also the documentation is much longer than the implementation. Also also, I think you can weaken the constraint to Concurrent.

std/shared/src/main/scala/cats/effect/std/Scope.scala
```diff
@@ -0,0 +1,86 @@
...
+   */
+  def allocate[A](resource: Resource[F, A]): F[A]
+}
+
+object Scope {
+
+  /**
+   * Create a scope in which resources can be allocated.
+   */
+  def apply[F[_]](implicit F: Async[F]): Resource[F, Scope[F]] =
```
djspiewak, line 68:
```suggestion
  def apply[F[_]](implicit F: Concurrent[F]): Resource[F, Scope[F]] =
```

std/shared/src/main/scala/cats/effect/std/Scope.scala
```diff
@@ -0,0 +1,86 @@
...
+    for {
+      stateRef <- Resource.make(F.ref[List[F[Unit]]](Nil)) { state =>
+        state.get.flatMap(_.sequence).void
```
djspiewak, line 71: Should we clear out the ref here to avoid memory leaks if someone has done something weird with their `F[_]` instances?

djspiewak (commented):
Interesting realization: this makes it *very* easy to do dangerous things:

```scala
Scope[IO] flatMap { s =>
  Resource.liftF(s.allocate(thingy).flatMap(t => ...).start)
}
```

I'm not sure that should stop us though. This sort of thing is really really useful, and `Resource` just has these kinds of holes.

std/shared/src/main/scala/cats/effect/std/Scope.scala
RaasAhsan: Yeah good call, I figured these would get GC'd whenever the `Scope` goes out of scope but it's better to be proactive about these things

diesalbla:
Looking at this Pull Request, I recalled that `fs2` has two notions around shared resources. 

- [`ScopedResource`](https://github.com/typelevel/fs2/blob/17eff83fe5f0d0068e626aa9e6590eba6d62655e/core/shared/src/main/scala/fs2/internal/ScopedResource.scala), a resource that can be used by several leases. It binds a resource (its finaliser) together with a counter of leases, such that the finaliser is only called once the last lease is released.

- [`Scope`](https://github.com/typelevel/fs2/blob/17eff83fe5f0d0068e626aa9e6590eba6d62655e/core/shared/src/main/scala/fs2/internal/Scope.scala#L81-L86), which represents a hierarchy of scopes, such that each Scope has a parent and children scopes, and a collection of resources that it shares with its ancestors and descendents. 

Perhaps, in order to reduce duplication, we could turn one of those into more general constructs, or use them on `cats-effect`? I would need to know how different the issues of resource scoping and supervision relations for `Fiber`, differ from those for Streams.

PR #1503, opened 2020-12-12
Remove SyncIO cancelation logic

### LINKED ISSUE #1501: Remove self-cancelation from SyncIO
We actually don't need it! `SyncIO` forms a lawful `MonadCancel` even without short-circuiting on `canceled`. All we need to do is define `canceled = unit`.

vasilmkd:
Resolves #1501.

### DIFF
--- a/core/shared/src/test/scala/cats/effect/SyncIOSpec.scala
+++ b/core/shared/src/test/scala/cats/effect/SyncIOSpec.scala
@@ -201,64 +201,20 @@ class SyncIOSpec
 
     "preserve monad right identity on uncancelable" in {
       val fa = MonadCancel[SyncIO].uncancelable(_ => MonadCancel[SyncIO].canceled)
-      fa.flatMap(SyncIO.pure(_)) must beCanceledSync
-      fa must beCanceledSync
+      fa.flatMap(SyncIO.pure(_)) must completeAsSync(())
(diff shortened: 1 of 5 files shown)


core/shared/src/test/scala/cats/effect/SyncIOSpec.scala
```diff
@@ -201,64 +201,20 @@ class SyncIOSpec
...
     }
 
     "cancel flatMap continuations following a canceled uncancelable block" in {
       MonadCancel[SyncIO]
         .uncancelable(_ => MonadCancel[SyncIO].canceled)
-        .flatMap(_ => SyncIO.pure(())) must beCanceledSync
+        .flatMap(_ => SyncIO.pure(())) must completeAsSync(())
     }
 
     "cancel map continuations following a canceled uncancelable block" in {
```
djspiewak, line 214: This test (and several others) now need some renaming, but otherwise looks good. :-)

djspiewak:
Thanks to the meddling of @wemrysi and our undiscovered missing generator cases, we now know that this change is going to be a bit more complex. So putting a hold on this until I can come back to the "existential root cancelation scope encoding" idea, which should solve this.

vasilmkd:
Folded into #1546.

PR #1509, opened 2020-12-16
fix inference for default show args

tpolecat:
Inference flows backwards with default args, so `implicit S: Show[A] = Show.fromString`, rather than inferring `showString[A]` it infers `A =:= Nothing` so you can't actually use the default arg. In any case providing a type arg fixes it.


vasilmkd (approved):
Thank you @tpolecat, I really should have added these tests the first time round.

PR #1526, opened 2020-12-20
Finalize, if necessary, when exiting an uncancelable region

### LINKED ISSUE #1525: Attempt to restore raceNeverIdentity{Left,Right} laws
As discussed in #1484, we should be able to restore the `GenSpawnLaws#raceNeverIdentity{Left,Right}` laws using `onCancel`.

wemrysi:
In `IOFiber#runLoop` we need to check if finalization is necessary after exiting an uncancelable region to ensure finalizers are run in the case where the rest of the call stack never requires reentering `runLoop`.

Consider, `F.onCancel(F.uncancelable(_ => F.canceled >> fa), cleanup)`. `cleanup` is not currently invoked as there is no node (`FlatMap`, `HandleErrorWith`, etc) between `onCancel` and `uncancelable` that would send us back into `runLoop`, the rest of the expression is evaluated completely within `IOFiber#succeeded` or `IOFiber#failed` (depending on `fa`).

This fixes four of the failing laws observed in #1484.

Closes #1525

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -1014,26 +1014,12 @@ private final class IOFiber[A](
   }
 
   private[this] def runTerminusSuccessK(result: Any): IO[Any] = {
-    val outcome: OutcomeIO[A] =
-      if (canceled)
-        /* this can happen if we don't check the canceled flag before completion */
-        OutcomeCanceled
-      else
-        Outcome.Succeeded(IO.pure(result.asInstanceOf[A]))
-
-    done(outcome)
+    done(Outcome.Succeeded(IO.pure(result.asInstanceOf[A])))
     IOEndFiber
   }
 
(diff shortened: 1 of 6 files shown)


djspiewak:
Holy shit I'm surprised we didn't catch this before. I remember working on this case specifically in `IO` a few months ago.

djspiewak (approved):
This seems correct to me, but would you mind taking your example and dropping it in `IOSpec` in some form? It's a little easier to work in terms of those tests when tracking down regressions in things like this.

RaasAhsan (approved):

vasilmkd (changes requested):
Marking as changes requested until the IOSpec is done, just to signal to other people this is not yet ready for merging. Great catch! 😄

djspiewak (approved):
Noice. Also I imagine scalafmt had a field day with both of those…

wemrysi:
Yeah, `scalafmt` and `scalac-3` were quarrelling, had to break them up.

PR #1533, opened 2020-12-22
Added experimental arm build

djspiewak:
Do not merge!


djspiewak:
Disappointing. It seems that QEMU doesn't reflect the ARM weak memory barrier semantics. So in order to really test ARM, we need a real instance rather than an emulator.

PR #1536, opened 2020-12-23
Handle uncaught errors in bracket

### LINKED ISSUE #1535: Uncaught error within bracketFull
We've got a bit of a problem in our default definition of `bracketFull` that's also causing problems for `Resource` and other projects downstream. The offending code:
```scala
def bracketFull[A, B](acquire: Poll[F] => F[A])(use: A => F[B])(
      release: (A, Outcome[F, E, B]) => F[Unit]): F[B] =
    uncancelable { poll =>
      val safeRelease: (A, Outcome[F, E, B]) => F[Unit] =
        (a, out) => uncancelable(_ => release(a, out))

      acquire(poll).flatMap { a =>
        val finalized = onCancel(poll(use(a)), safeRelease(a, Outcome.Canceled()))
        val handled = finalized.onError {
          case e => void(attempt(safeRelease(a, Outcome.Errored(e))))
        }
        handled.flatTap { b => safeRelease(a, Outcome.Succeeded(b.pure)).attempt }
      }
    }
```
A minimized example that illustrates how it's breaking:
```scala
IO.asyncForIO.bracketFull[Unit, Unit](_ => IO(println("A"))) { a =>
      sys.error("borked")
    } { case x => IO(println("B")) }
```
Running this program produces the following output:
```
[info] running testapp.ExampleFive
A
[error] (run-main-1a) java.lang.RuntimeException: borked
[error] java.lang.RuntimeException: borked
[error] 	at scala.sys.package$.error(package.scala:27)
[error] 	at testapp.ExampleFive$.$anonfun$run$3(TestApp.scala:32)
[error] 	at cats.effect.kernel.MonadCancel.$anonfun$bracketFull$4(MonadCancel.scala:352)
[error] 	at cats.effect.IOFiber.flatMapK(IOFiber.scala:987)
[error] 	at cats.effect.IOFiber.succeeded(IOFiber.scala:795)
[error] 	at cats.effect.IOFiber.runLoop(IOFiber.scala:304)
[error] 	at cats.effect.IOFiber.execR(IOFiber.scala:908)
[error] 	at cats.effect.IOFiber.run(IOFiber.scala:135)
[error] 	at cats.effect.unsafe.WorkerThread.runFiber(WorkerThread.scala:162)
[error] 	at cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:295)
[error] stack trace is suppressed; run last coreJVM / Compile / bgRun for the full output
[error] Nonzero exit code: 1
[error] (coreJVM / Compile / run) Nonzero exit code: 1
[error] Total time: 4 s, completed Dec 23, 2020 12:43:50 PM
```
As you can see, the message `B` is never printed, meaning that the finalizer is never being run! The root cause here is that the `use` clause of the `bracketFull` is raising an exception by throwing rather than lifting into `IO`. Take a look at our definition of `bracketFull` -- `use(a)` is _strictly evaluated_ before the error handler has a chance to be registered!

I don't really see a great solution here, but here some:
1. Make `poll` or `onCancel` lazy. This is completely arbitrary to me, and we might as well make everything lazy
2. Have `MonadCancel` extend `Defer` so we can lazily evaluate effects. The downside to this as many have pointed out is that it basically cheats `Sync` in.
3. Cheat lazy evaluation in by doing something like `F.unit.flatMap(x => poll(use(a)))`
4. Acknowledge that thrown exceptions are going to break things.
5. Make bracket primitive again...

I think 3 is looking like the most obvious solution here, so I'll go ahead and open a PR for that

RaasAhsan:
Fixed #1535

### DIFF
--- a/core/shared/src/test/scala/cats/effect/IOSpec.scala
+++ b/core/shared/src/test/scala/cats/effect/IOSpec.scala
@@ -188,7 +188,6 @@ class IOSpec extends IOPlatformSpecification with Discipline with ScalaCheck wit
           .handleErrorWith(_ => (throw TestException): IO[Unit])
           .attempt must completeAs(Left(TestException))
       }
-
     }
 
     "suspension of side effects" should {
@@ -944,6 +943,20 @@ class IOSpec extends IOPlatformSpecification with Discipline with ScalaCheck wit
       }
       // format: on
 
+      "finalize on uncaught errors in bracket use clauses" in ticked { implicit ticker =>
+        val test = for {
+          ref <- Ref[IO].of(false)
+          _ <-
+            IO.asyncForIO
+              .bracketFull[Unit, Unit](_ => IO.unit)(_ => sys.error("borked!")) {
+                case _ => ref.set(true)
+              }
+              .attempt
+          flag <- ref.get
+        } yield flag
+
+        test must completeAs(true)
+      }
     }
 
     "stack-safety" should {
--- a/kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
@@ -349,7 +349,9 @@ trait MonadCancel[F[_], E] extends MonadError[F, E] {
         (a, out) => uncancelable(_ => release(a, out))
 
       acquire(poll).flatMap { a =>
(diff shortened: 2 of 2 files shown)


djspiewak:
This is frustrating on several levels, but I don't have a better answer :-(

PR #1538, opened 2020-12-23
Supervisor and Dispatcher docs

RaasAhsan:
I think we could write a great deal more for both of these, but I'm trying to get a good baseline for all the important classes for the final 3.0.0 release


djspiewak (approved):
Holy crap those diagrams are amazing

PR #1541, opened 2020-12-25
Renaming functions in Fiber.scala

### LINKED ISSUE #1500: Consider renaming methods in Fiber
The current naming is descriptive, but a bit longwinded and maybe scary, I propose to replace

```scala
trait Fiber[F[_], E, A] {
   ....
  def join: F[Outcome[F, E, A]]
  def joinAndEmbed(onCancel: F[A])(implicit F: MonadCancel[F, E]): F[A] 
  def joinAndEmbedNever(implicit F: GenSpawn[F, E]): F[A]
}
```

with

```scala
trait Fiber[F[_], E, A] {
   ....
  def join: F[Outcome[F, E, A]]
  def joinWith(onCancel: F[A])(implicit F: MonadCancel[F, E]): F[A] 
  def joinWithNever(implicit F: GenSpawn[F, E]): F[A]
}
```

The work would involve renaming those methods, and changing the relevant tests, docs, and scaladoc. A pretty easy first contribution for someone :) (assuming the naming gets agreement)

jyoo980:
Should resolve #1500.

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Fiber.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Fiber.scala
@@ -60,7 +60,7 @@ trait Fiber[F[_], E, A] {
    * returned. If the fiber completes with [[Errored]], the error is raised.
    * If the fiber completes with [[Cancelled]], `onCancel` is run.
    */
-  def joinAndEmbed(onCancel: F[A])(implicit F: MonadCancel[F, E]): F[A] =
+  def joinWith(onCancel: F[A])(implicit F: MonadCancel[F, E]): F[A] =
     join.flatMap(_.embed(onCancel))
 
   /**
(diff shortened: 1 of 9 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/Fiber.scala
```diff
@@ -60,7 +60,7 @@ trait Fiber[F[_], E, A] {
    * returned. If the fiber completes with [[Errored]], the error is raised.
    * If the fiber completes with [[Cancelled]], `onCancel` is run.
    */
-  def joinAndEmbed(onCancel: F[A])(implicit F: MonadCancel[F, E]): F[A] =
+  def joinWith(onCancel: F[A])(implicit F: MonadCancel[F, E]): F[A] =
```
jyoo980, line 63: Any reason why this is public? It doesn't seem to be used anywhere else but as a delegated call from `joinWithNever`

vasilmkd: It looks like we do not have a use for it in the codebase, but someone else just might. 😄

vasilmkd (approved):
Congratulations on your first contribution!

djspiewak:
Thank you!

PR #1546, opened 2020-12-26
Configurable root cancelation scope in MonadCancel

djspiewak:
Burying the lede a bit… We now have `Sync <: MonadCancel`. This fixes issues encountered by @tpolecat, @oleg-py, and… pretty much everyone. It is no longer bizarre for `SyncIO` to extend `MonadCancel`.

This is achieved lawfully by the addition of a `rootCancelScope: CancelScope` property within `MonadCancel`, where `CancelScope` = `Uncancelable | Cancelable`. When `rootCancelScope = Uncancelable`, then all values `F[A]` are defined to behave as if they are wrapped in some outer `uncancelable` without an accessible `Poll`. `Spawn`s cannot be `Uncancelable` at the root, since this would unravel too much of the model, though if there's extremely strong demand for this we could explore it in future.

`MonadCancel.Uncancelable` provides some utility mixins which are helpful when defining `Uncancelable` instances. Note that `forceR` cannot be thus defined because of its semantics regarding compositional effect zeros, which is interesting since it implies that `MonadCancel` *is* fundamentally slightly stronger than `MonadError` even when `Uncancelable`.


SystemFw (approved):
Awesome!

mpilquist (approved):

PR #1549, opened 2020-12-26
Revise finalizer error semantics

### LINKED ISSUE #1494: Finalizer error semantics
There was a discussion the other day on Gitter about the semantics of finalizers in the presence of errors. Capturing the gist here so we don't forget about it.

The semantics of the following two effects were considered (current behavior is commented above each respective example):
```scala
// e1 is raised, e2 is side-channel reported
F.guaranteeCase(F.raiseError(e1), F.raiseError(e2))
// e2 is reported
F.guaranteeCase(F.unit, F.raiseError(e2))
```

The behavior of the first example is sensible and is consistent with Cats Effect 2 and Java's try-with-resources block (Java would track `e2` as a suppressed exception on `e1`).

For the second example, CE2's `bracket` (and by extension, `guarantee`) exhibits different semantics which is captured [in this test](https://github.com/typelevel/cats-effect/pull/458/files#diff-bfb6fd95750be02e2e0bbc72b99bda33bef009c81d600a00a6e8ff86fd1f6712R666). Errors in a finalizer are raised even when the user action succeeds. This is consistent with Java's try-with-resources block (we should confirm this).

For what it's worth, in both CE2 and CE3, if an error occurs in a cancellation, it is always reported.

I think I'm in favor of keeping consistency with CE2 and Java here, so raising the finalizer error when it occurs. One thing we'll want to verify is what happens when multiple finalizers fail -- is the first finalizer raised and subsequent ones reported/suppressed, or something else? I'm not sure how this change in semantics would interact with the laws yet, but I suppose we'll find out :)

RaasAhsan:
Fixes #1494 . The change is pretty simple: in the happy path, if we encounter an error when running the finalizer, just return that. If we're running the failing finalizer in the error path, we'll return the original error, like before.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -79,18 +79,22 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   def bracket[B](use: A => IO[B])(release: A => IO[Unit]): IO[B] =
     bracketCase(use)((a, _) => release(a))
 
-  def bracketCase[B](use: A => IO[B])(release: (A, OutcomeIO[B]) => IO[Unit]): IO[B] =
+  def bracketCase[B](use: A => IO[B])(release: (A, OutcomeIO[B]) => IO[Unit]): IO[B] = {
+    val safeRelease: (A, OutcomeIO[B]) => IO[Unit] =
+      (a, out) => IO.uncancelable(_ => release(a, out))
+
     IO.uncancelable { poll =>
       flatMap { a =>
-        val finalized = poll(use(a)).onCancel(release(a, Outcome.Canceled()))
+        val finalized = poll(use(a)).onCancel(safeRelease(a, Outcome.Canceled()))
         val handled = finalized.onError { e =>
-          release(a, Outcome.Errored(e)).handleErrorWith { t =>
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -80,16 +80,18 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
     bracketCase(use)((a, _) => release(a))
 
   def bracketCase[B](use: A => IO[B])(release: (A, OutcomeIO[B]) => IO[Unit]): IO[B] = {
-    def doRelease(a: A, outcome: OutcomeIO[B]): IO[Unit] =
-      release(a, outcome).handleErrorWith { t =>
-        IO.executionContext.flatMap(ec => IO(ec.reportFailure(t)))
-      }
+    val safeRelease: (A, OutcomeIO[B]) => IO[Unit] =
+      (a, out) => IO.uncancelable(_ => release(a, out))
```
djspiewak, line 85: Do we side-channel errors that are caught in `onCancel`?

RaasAhsan: yup, we do for the overrided `IO` method. this brings up an interesting question for the `MonadCancel` implementation though -- should we assume errors in the cancel path are swallowed, or should we proactively catch errors there?

PR #1552, opened 2020-12-28
Memoizing a canceled effect should result in canceled

### LINKED ISSUE #1519: Proper handling for self cancelation in memoize
`F.memoize(F.canceled)` deadlocks and… really shouldn't.

wemrysi:
Fixes #1519


kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
```diff
@@ -77,7 +89,7 @@ trait GenConcurrent[F[_], E] extends GenSpawn[F, E] {
               case st @ Done(value) =>
                 st -> F.pure(value)
             }.flatten
-          }.rethrow
+          }.flatMap(_.embed(F.canceled *> F.never))
```
wemrysi: I adopted prior art from `GenSpawn#race` where if the memoized effect results in cancellation and the attempt to get the memoized value is masked, the result is non-termination. Though, it'd seem more user-friendly to fail fast with an error in both cases. I'm happy to switch to that behavior unless there are objections.

kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
```diff
@@ -31,41 +31,53 @@ trait GenConcurrent[F[_], E] extends GenSpawn[F, E] {
...
-            }.flatten
-
-          def fetch: F[Either[E, A]] =
-            uncancelable { poll =>
-              for {
-                result0 <- poll(fa).attempt
-                // in some interleavings, there may be several racing fetches.
-                // always respect the first completion.
-                result <- state.modify {
-                  case st @ Done(result) => st -> result
```
wemrysi, line 62: Maybe I missed it, but I didn't see how this was possible. The tests also assert as much.

kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
djspiewak: I do agree it's more user-friendly to fail fast with an error, but given that we don't know `E`, I'm not sure how we could do that here. :-( This is an unfortunate hole in cancelation right now that I'm still thinking about. Haskell solves this problem by having `forkIO` inherit mask status, so we could just add a `startWithMasks` or something like that to try to resolve that problem, but it feels very ad hoc.

kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
RaasAhsan: IIRC it can happen under the following interleaving:
1. The first fetch is cancelled.
2. According to the state machine, we are in the `Start` state, so we start a second fetch.
3. The first fetch actually terminates before it observes cancellation, so it caches the result.
4. The second fetch terminates with a result, but if there is already one, we should use that. The result of `memoize` should be consistent or else we are breaking the contract

On the other hand, if the desired semantic is to cache the actual `Outcome` of the fiber, I'm not sure if this should even be possible since `cancel`ing the first `get` should just cache that `Outcome`. We should think about that more

kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
```diff
@@ -31,41 +31,53 @@ trait GenConcurrent[F[_], E] extends GenSpawn[F, E] {
...
   def deferred[A]: F[Deferred[F, A]]
 
+  /**
+   * Caches the outcome of `fa`.
+   *
+   * The returned inner effect, hence referred to as `get`, when sequenced, will
+   * evaluate `fa` and cache the outcome. If `get` is sequenced multiple times
+   * `fa` will only be evaluated once.
+   *
+   * If all `get`s are canceled prior to `fa` completing, it will be canceled
```
RaasAhsan, line 41: Is this still true ? If we're caching the `Outcome`, when the first fetch is cancelled, it would just write in `Canceled`

kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
wemrysi: Ah yes, still getting used to the parametric error type 😄

kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
wemrysi: I totally missed it, thanks for the explanation, will fix.

>  canceling the first get should just cache that Outcome

Are you suggesting that canceling the last remaining `get` just results in `Canceled` or that we allow evaluation of `fa` to finish, even when there is nothing actively awaiting, and cache the eventual `Outcome`?

RaasAhsan: I was looking at the PR implementation; it seems like if the first `fetch` is cancelled, `Done + Canceled` would be written in by the `guaranteeCase` clause. Correct me if I'm wrong there. I'm not really sure what the best semantics are here, evidenced by the ongoing conversation in Gitter :)

kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
wemrysi: My intent was to distinguish between the outcome of `fa` being `Canceled` and all the fetches being canceled. Where the former would cache `Canceled` and the latter would re-evaluate `fa` on the next fetch. However, given the interleaving you described above, this isn't always possible (without additional state).

djspiewak:
Spent some time thinking about this...

I'm of the opinion that, if someone does a `race` on a pair of `memoize`s, then the fact that they've engineered a race condition is literally on the tin. Leaving off that concern, the semantics would ideally be something like this:

- If unevaluated, then evaluate
- If already evaluating, then wait for evaluation to complete
- If evaluation completes with `Canceled`, then allow that to propagate through and release one of the blocked callers to attempt evaluation on *their* fiber

This implies a couple of things. First, it implies that we can't use a hidden inner fiber in the implementation. We aren't doing that anyway, so all good, but it's interesting that this is the implication. The reason that we can't is because we need to allow cancelation to propagate unimpeded from its origin in the current scope, which is only possible if we are not *relocating* that outcome from some other scope (i.e. a fiber that we `join`ed on).

Second, it implies that the state machine needs to be something like this:

```scala
final case class Completed[A](value: A)
final case class Errored(e: E)
final case class Evaluating(latch: Deferred[F, Unit])
final case class Unevaluated()
```

Or something like that. When the first evaluator gets to the effect, they atomically attempt to create the latch and set up `Evaluating`. If that fails, they just `get` on the winner's latch. When the latch releases, they simply loop on the read, re-doing the atomic creation if the state is still `Unevaluated`.

kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
wemrysi: If cancellation is observed before the fetch terminates, it shouldn't write anything as it'll have transitioned to `Start` which is now handled explicitly to ignore the outcome. However, this has other implications that I overlooked, like if another fetch is started in between when the cancellation state is updated and when the fiber actually cancels (transitioning us back to `Running`). This is the "needs more state" scenario I alluded to above as we'd need a flag or another state to transition to in order to serialize fetches.

wemrysi:
@djspiewak I'm 👍 on those semantics, I also like that it completely serializes evaluation.

djspiewak:
> I also like that it completely serializes evaluation.

Same :-D I think that's a lot saner from a user's perspective.

wemrysi:
Implemented the semantics proposed by @djspiewak .

djspiewak:
Sorry for the long delay on this, @wemrysi! My calendar currently looks a bit like the manual for the moon landing. Thanks for taking this on!

PR #1553, opened 2020-12-29
Add supertypes for concurrent types

### LINKED ISSUE #1396: Extract co- and contravariant parents from Queue, PQueue, Ref, and Deferred
Follow up on https://github.com/typelevel/cats-effect/pull/1361#discussion_r514490030

RaasAhsan:
Closes #1396 . I took the liberty to go ahead and do this one since http4s will benefit from the `Queue` ones

`Enqueue` and `Dequeue` were the predecessor names for `Queue` in fs2, however, there is already a `Dequeue` (double-ended queue) and also that naming convention doesn't extend well to `Ref`, `Deferred`, and `PQueue`. Not terribly happy with `Sink` and `Source` on everything but lmk if there's something more obvious (maybe `Take` and `Put`)

Also added some `Invariant` instances where it made sense, I think there might be some more opportunities with the parent types

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Ref.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Ref.scala
@@ -39,25 +39,7 @@ import scala.annotation.tailrec
  * The default implementation is nonblocking and lightweight, consisting essentially
  * of a purely functional wrapper over an `AtomicReference`.
  */
-abstract class Ref[F[_], A] {
-
-  /**
-   * Obtains the current value.
-   *
-   * Since `Ref` is always guaranteed to have a value, the returned action
-   * completes immediately after being bound.
-   */
-  def get: F[A]
-
-  /**
-   * Sets the current value to `a`.
-   *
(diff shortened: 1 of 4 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/Ref.scala
```diff
@@ -39,25 +39,7 @@ import scala.annotation.tailrec
...
-  /**
-   * Sets the current value to `a`.
-   *
-   * The returned action completes after the reference has been successfully set.
-   *
-   * Satisfies:
-   *   `r.set(fa) *> r.get == fa`
-   */
-  def set(a: A): F[Unit]
+abstract class Ref[F[_], A] extends RefSource[F, A] with RefSink[F, A] {
```
rkrzewski, line 42: `Invariant` instance is missing here. If this is the case where it does not make sense, I don't see why?

RaasAhsan: `Ref` has already got an `Invariant` instances surprisingly

PR #1556, opened 2020-12-30
Extract useful test instances from cats-effect-core test suite

mpilquist:
This PR does a few things in order to simplify testing downstream projects with ce3:
- Renames the existing cats-effect-testkit to cats-effect-kernel-testkit
- Introduces a new cats-effect-testkit
- Extracts useful instances (e.g. `Arbitrary[IO[A]]`, `Eq[IO[A]]`) in to a new trait in cats-effect-testkit

You can see these in use in fs2 here: https://github.com/typelevel/fs2/pull/2198/commits/f34abcf039b3a1e03d395bdfdd99fa386ef3ffac

So far, I've resisted the urge to editorialize. Before merging, we should make a few decisions though:
- What's a better name than `CatsEffectInstances`? Update: I renamed to `TestInstances`, which matches the CE2 name.
- Should we define aliases for things like `TestContext`? Update: I added an alias for `TestContext`.
- Should we provide an uber import that brings in all common instances? Take a look at the fs2 usage example -- note how we need a number of imports to check laws. Update: I mixed `SyncTypeGenerators`, `ParallelFGenerators`, and `OutcomeGenerators` in to `TestInstances`.


build.sbt
```diff
@@ -202,10 +202,29 @@ lazy val laws = crossProject(JSPlatform, JVMPlatform).in(file("laws"))
...
   .settings(
-    name := "cats-effect",
+    name := "cats-effect-testkit",
+    libraryDependencies ++= Seq(
+      "org.scalacheck" %%% "scalacheck" % ScalaCheckVersion))
 
+/**
+ * Unit tests for the core project, utilizing the support provided by testkit.
+ */
+lazy val tests = crossProject(JSPlatform, JVMPlatform).in(file("tests"))
```
djspiewak, line 224: I believe this should not be published.

djspiewak (commented):
Makes me 😢🐼, but probably no avoiding it.

rossabaker:
For what it's worth, http4s already has an `Eq[F[A]]` on the cats-effect-3 branch and I'd be happy to rip it out for `Eq[IO[A]]`.  We all know it's abominable and we all need it anyway.

djspiewak:
Took the liberty of publishing a snapshot off this branch: `3.0-42-83f5c56`

PR #1561, opened 2020-12-30
Don't terminate JVM forcibly on ExitCode.Success

### LINKED ISSUE #1560: CE3 IOApp always forcibly terminates JVM, even on ExitCode.Success
Compare:
CE2: https://github.com/typelevel/cats-effect/blob/series/2.x/core/jvm/src/main/scala/cats/effect/internals/IOAppPlatform.scala#L28
CE3: https://github.com/typelevel/cats-effect/blob/v3.0.0-M5/core/jvm/src/main/scala/cats/effect/IOApp.scala#L82

I've run into it because there's a direct invocation of `IOApp.main` from mdoc in Doobie documentation.
https://github.com/tpolecat/doobie/blame/master/modules/docs/src/main/mdoc/docs/14-Managing-Connections.md#L127
That's a pretty niche usecase, but I think preserving CE2 semantics would be a good idea as the comment about (non CE) worker threads and shutdown hooks indicates.

rkrzewski:
Fixes #1560

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/IOApp.scala
+++ b/core/jvm/src/main/scala/cats/effect/IOApp.scala
@@ -85,7 +85,14 @@ trait IOApp {
         // otherwise scheduler threads will accumulate over time.
         runtime.internalShutdown()
         runtime.shutdown()
-        System.exit(result.code)
+        if (result == ExitCode.Success) {
+          // Return naturally from main. This allows any non-daemon
+          // threads to gracefully complete their work, and managed
+          // environments to execute their own shutdown hooks.
+          ()
+        } else {
+          System.exit(result.code)
+        }
       }
     } catch {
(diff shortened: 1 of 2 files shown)


rkrzewski:
example/test-jvm.sh failed which sounds suspicious because it tests IOApp termination and the PR touches the same functionality, but consider this output:
```
[info] running cats.effect.example.Example 
started
Timed out waiting for 'started' in output
Error: Process completed with exit code 255.
```
Also, running `example/test-jvm.sh 2.13.3` locally failed for me once, on the first run where some compilation was performed with the same message as above and passed on 5 or so subsequent runs.
In conclusion I believe this test is flaky because 30s timeout is too tight.

djspiewak:
> In conclusion I believe this test is flaky because 30s timeout is too tight.

I believe you are correct.

PR #1565, opened 2020-12-31
Async[Resource]

### LINKED ISSUE #1551: Attempt to generalize Cont even further
It should be possible to move the callback value into an entirely existential context, giving us an even more general construct:

```scala
trait Cont[F[_], R] {
  def apply[G[_], K](
      implicit
      G: MonadCancel[G, Throwable]): (Either[Throwable, K] => Unit, G[K], F ~> G) => G[R]
}
```

djspiewak:
The dream!

No but seriously, this PR implements the hierarchy for `Resource[F, *]` given the appropriate instance for `F`. Note that `Spawn` cannot be implemented in isolation since we need a `Ref` to implement `start`, but `Concurrent` is more useful anyway. Additionally, we can only implement `MonadCancel[Resource[F, *], Throwable]` due to the fact that `allocated` requires exactly this. But even still, this is exciting.

One cool consequence: we get `Parallel[Resource[F, *]]` for free! Sort of. It has slightly different semantics from the old one.

The most complicated element of this PR is `start`/`join`, as you probably would expect. `start` is immensely tricky because you have to figure out when to run the finalizers of the constituent `Resource`: should you run them upon fiber completion, or should you allow them to be run within the `join`ing context, assuming it exists? The answer to this is obvious if the fiber is canceled or errored: you have to run them within the spawned fiber. However, if the fiber succeeds, we have the issue of dangling finalizers which have yet to be run, and we don't know whether or not we will ever be `join`ed. To make matters worse, we might actually be `join`ed multiple times. Or `cancel`ed (after completion!) and then `join`ed. It's kind of a mess.

The semantics @wemrysi and I designed for this basically boil down to the following:

1. If the scope containing the *fiber* is terminated, then trigger greedy finalization within the fiber itself
    a. If the fiber has already succeeded, immediately run the finalizers.
    b. If the fiber has yet to succeed (i.e. is still running), then set a flag such that the finalizers will be immediately run upon success (without waiting for `join`)
2. If the fiber is canceled or errored, finalize immediately
3. If the fiber succeeds, await `join` and (when `join`ed), extend the finalizers into the `join`ing scope
    a. The semantics of multiple `join`s are essentially undefined, but what will happen in practice is finalizers will be run multiple times
4. If a fiber is canceled *after* it succeeds but before it has been `join`ed, run the finalizers and rewrite the outcome to `Canceled()` (basically pretending the fiber was canceled *just* before it succeeded)

Semantic number 4 is a bit buggy right now and needs tests. Note that we considered a ref counting scheme to try to resolve **3.a**, but decided to err on the side of simplicity.

The motivating example here is we want `race` to work in compliance with what users would expect in a scenario where they're racing a pair of scarce resource acquisitions. In that scenario, users would reasonably expect that the winner of the race would be returned and remain open, while the loser would be closed and discarded. Notably, we don't want to special-case `race` just for `Resource`, so we need to make sure that the semantics of `start`/`join`/`cancel` are consistent with this goal modulo the existing and natural definition of `race` in `Spawn`. The described semantics achieve this.

Note: we may want to add a `Resource#flattenK` operation (`Resource[Resource[F, *], A] => Resource[F, A]`) to help users along with the `background` case.

One final bit of interest… @RaasAhsan had the very helpful intuition that `allocated` + `applyFull` gives us the ability to define *arbitrary* scoping semantics within the context of `Resource`. This intuition is what led to the (surprisingly obvious…) definition of `onCancel`, which is a theoretically interesting operation since it runs quite counter to `Resource`'s natural semantics. I think that we probably have more space to explore here, since if `onCancel` and `onFinalize` are *both* possible, then really any scoping semantic should be achievable.

Anyway! This needs a few more tests and some polish, but overall I think it's ready for a more serious set of eyes on it.

Also resolves #1551


kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -757,87 +760,327 @@ object Resource extends ResourceInstances with ResourcePlatform {
...
+    extends ResourceMonadError[F, Throwable]
+    with MonadCancel[Resource[F, *], Throwable] {
+  implicit protected def F: MonadCancel[F, Throwable]
+
+  def canceled: Resource[F, Unit] =
+    Resource.eval(F.canceled)
+
+  def forceR[A, B](fa: Resource[F, A])(fb: Resource[F, B]): Resource[F, B] =
+    Resource.applyFull(
+      _(fa.use_ !> fb.allocated).map(_.map(fin => (_: Resource.ExitCase) => fin)))
```
wemrysi: I think `fa.attempt.productR(fb)` would match the described semantics of `forceR` better. Currently the scope of `fa` is closed, even on success, prior to `fb`.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -757,87 +760,327 @@ object Resource extends ResourceInstances with ResourcePlatform {
...
+
+          case Succeeded(_) =>
+            val maybeFins = state modify { s =>
+              (
+                s.copy(completed = true),
+                Some(s.fin).filter(_ => s.runOnComplete && !s.canceled))
+            }
+
+            maybeFins.flatMap(_.getOrElse(F.unit))
+        }
```
wemrysi:
```suggestion
      val finalized = F uncancelable { poll =>
        poll(fa.allocated) flatMap {
          case (a, rel) =>
            state modify { s =>
              if (s.runOnComplete || s.canceled)
                (s.copy(completed = true), rel.as(None))
              else
                (s.copy(completed = true, fin = Some(rel)), F.pure(Some(a)))
            }
        }
      }
```

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -757,87 +760,327 @@ object Resource extends ResourceInstances with ResourcePlatform {
...
+        F.start(finalized) map { outer =>
+          val fiber = new Fiber[Resource[F, *], Throwable, A] {
+
+            def cancel =
+              Resource eval {
+                F uncancelable { poll =>
+                  // technically cancel is uncancelable, but separation of concerns and what not
+                  val maybeFins = poll(outer.cancel) *> {
+                    state modify { s =>
+                      (s.copy(canceled = true), Some(s.fin).filter(_ => s.completed))
```
wemrysi:
```suggestion
                      if (!s.joined)
                        (s.copy(canceled = true, fin = None), s.fin.traverse_(x => x))
                      else
                        (s.copy(canceled = true), F.unit)
```
How I think it'd look with `joined`.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
wemrysi: I think this is a bit easier to understand and we leverage the cleanup behavior already implied by `allocated`. We also structurally represent cancellation from here on via `None`, which I think is nice.

Unless I've confused things, I'm pretty sure we need to run the finalizer if `runOnComplete || canceled` to handle the race where the fiber completes after `cancel` has been called but before cancellation could be observed. In this case, we're guaranteeing that if the user calls `cancel` prior to the fiber terminating, it will be canceled. We could alternatively check if they've joined already and allow the "natural" outcome of the race, though it seems nice to implement their intent since we can.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -757,87 +760,327 @@ object Resource extends ResourceInstances with ResourcePlatform {
...
+   * 3. If the fiber succeeds, await joining and extend finalizers into joining scope
+   * 4. If a fiber is canceled *after* it completes, we explicitly snag the finalizers and run them, setting the result to Canceled()
+   *   a) Note that we are missing a check here! We need to disable this behavior if someone has already joined
+   */
+  def start[A](fa: Resource[F, A]): Resource[F, Fiber[Resource[F, *], Throwable, A]] = {
+    final case class State(
+        fin: F[Unit] = F.unit,
+        runOnComplete: Boolean = false,
+        completed: Boolean = false,
+        canceled: Boolean = false)
```
wemrysi:
```suggestion
    final case class State(
        fin: Option[F[Unit]] = None,
        runOnComplete: Boolean = false,
        canceled: Boolean = false,
        joined: Boolean = false)
```
If we use `allocated` below, we could represent the need to finalize as the presence of a finalizer. When we actually decide to finalize, we set `fin = None` allowing us to simplify the outer resource finalizer to just "run it if it's there".

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -757,87 +760,327 @@ object Resource extends ResourceInstances with ResourcePlatform {
...
+                    oc)
+                }
+              }
+          }
+
+          val finalizeEarly = {
+            val maybeFins = state modify { s =>
+              (
+                s.copy(runOnComplete = true),
+                Some(s.fin).filter(_ => s.completed && !s.canceled))
```
wemrysi: With the change to `State`, this could become
```suggestion
                s.fin.traverse_(x => x)
```

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -757,87 +760,327 @@ object Resource extends ResourceInstances with ResourcePlatform {
...
+                    }
+                  }
+
+                  maybeFins.flatMap(_.getOrElse(F.unit))
+                }
+              }
+
+            def join =
+              Resource eval {
+                val results = outer.join.map[Outcome[Resource[F, *], Throwable, A]] {
```
wemrysi: Given the above change to `finalized`, here we'd `flatMap` instead and branch on the `Option[A]` in the `Succeeded` case which forces us to return `Canceled` when `None` (since we won't have a `Throwable` or an `A`).

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -757,87 +760,327 @@ object Resource extends ResourceInstances with ResourcePlatform {
...
+                  case Canceled() => Canceled()
+                  case Errored(e) => Errored(e)
+
+                  case Succeeded(fp) =>
+                    Succeeded(Resource.make(fp)(_ => state.get.flatMap(_.fin)))
```
wemrysi: Also with the change to `State`, here we'd `state.modify`, setting `fin = None` before finalizing, preventing double finalization (which I think is currently possible via the outer `Resource` since we aren't clearing the finalizer here).

wemrysi (commented):
Damn this is complicated 😅 

Had a couple suggestions, primarily centered around representing the completion/finalization state in `start` structurally, which I think allows us to drop one of the booleans and simplify some of the implementation.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
djspiewak: Whether or not to close the scope of the left side in `forceR` I think is a decision we can make here. I'm not sure what's right or wrong. If we want to keep the scope open, then something like `fold` putting hte finalizers in a `Ref` and then still delegating to `forceR` on the `F` for the composition might be the right approach.

We definitely can't just use `.attempt.productR` since `F` might be `OptionT` or something like that.

wemrysi: Fair enough regarding `attempt`. Thinking about it in the context of the success case, being consistent there with `fa *> fb` seemed least surprising.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
wemrysi: This should also prevent multiple finalizations in the case of multiple joins.

djspiewak:
@wemrysi I like your suggested `State` change quite a bit! I'll incorporate that and push.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
djspiewak: Spent some more time thinking about this. I think the only consistent semantic is to close the left scope. The problem here is that the left side *might* error, and if it does, its finalizers will naturally be run. However, that error will be discarded (by the semantics of `forceR`) and the right will be run. In that case, the finalizers are forcibly run early. If the left side *succeeds* and we attempt to keep the scope open, then the result will be the same from the user's perspective (the value will be discarded and the right side sequenced), but now the scope is mysteriously open. So that's more than a bit weird.

I think it's a little simpler if we just say that the scope on the left always closes. This gives `forceR` the same semantics as `supersedeBy` from `Region` in some of the earlier CE3 prototypes, which I think is useful in its own right.

djspiewak:
Added some tests, because that seems generally useful. Also note the test for `race` on `Resource`, which is a case that I've wanted to encode for *years* and was legitimately one of the precipitating use-cases for a lot of CE3's design. The fact that this semantic directly falls out of the composed primitives, without any special-casing (it's just the generic `race` defined in `Concurrent`!), is incredibly satisfying.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
wemrysi: Hm, maybe my understanding of `forceR` is incomplete/misguided. How should one view `fa forceR fb` in relation to `fa *> fb` when `fa` and `fb` succeed (asking in a parametric context here, not specifically for `Resource`)? My current understanding, from the description of `forceR`, is that `fa forceR fb` and `fa *> fb` are equivalent when both arguments succeed.

Similarly, when I know I'm dealing with `Resource`, I'd expect `fa`'s scope to extend over `fb` in `fa *> fb` so I thought it seemed reasonable that `forceR` should behave accordingly.

In the case where `fa` errors, having it's finalizers run is also expected and consistent with `*>`, the only difference with `forceR` being that `fb` would still evaluate, but that's expected and the point of calling `forceR` in the first place.

djspiewak: > My current understanding, from the description of forceR, is that fa forceR fb and fa *> fb are equivalent when both arguments succeed.

This is an area that I think we can do better on actually *defining*. It's difficult to do generically though. Another area where this shows up is `StateT`. If I remember correctly, `StateT`'s `MonadCancel` instance discards state in one of the branches of `forceR` (I think the right?).

In a sense, `fa !> fb` is meant to mirror `(fa *> F.canceled).onCancel(fb)`, whatever that ends up being (also naively ignoring errors in `fa` in this case). This is obviously a strange situation, since the finalizer kind of lives in its own universe under *normal* circumstances (when canceled externally). Trying to figure out what this means for `Resource` is a little mind-bendy…

wemrysi (approved):
Sorry for the delay!

djspiewak:
Merging this in the interest of moving forward. I'm pretty sure that we should remove the magical `join` finalizer-forwarding semantic (meaning that we would entirely rely on the outer scope for finalization), but that can be a follow-up.

PR #1567, opened 2020-12-31
Fix syntax type inference

### LINKED ISSUE #1564: GuaranteeCase syntax doesn't seem to type infer in general
```scala
def foo[F[_], A](fa: F[A])(implicit F: MonadCancel[F, Throwable) =
  fa guaranteeCase {
    case Outcome.Canceled() => ...
    // ...
  }
```

In the above, the compiler will give an error like `missing parameter type for expanded function` or something similar related to type inference. Directly invoking the function on `MonadCancel` by way of `F.guaranteeCase` circumvents this problem entirely. We should try to find a way to smooth this over for users if possible.

Scala 2.13

joroKr21:
Split syntax based on the error type.
If we care about the error type, use an implicit to infer it.
Otherwise make the syntax methods free from the error type.

fixes #1564

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/syntax/GenConcurrentSyntax.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/syntax/GenConcurrentSyntax.scala
@@ -16,38 +16,43 @@
 
 package cats.effect.kernel.syntax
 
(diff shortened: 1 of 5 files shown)


djspiewak:
Ahhhhhhhhh! This makes all the sense. Thank you for taking this on!

PR #1572, opened 2021-01-01
revise IO.bracketCase and override IO.bracketFull

### LINKED ISSUE #1522: Add IO.bracketFull
The easy route here is to just delegate to `MonadCancel#bracketFull`

heka1024:
Resolve #1522.

As pointed in `MonadCancel`, `use` in `bracket` should lazily evaluate to avoid exceptions. So I changed `poll(use(a))` to `poll(IO.Unit >> use(a))`. And override `bracketFull` of `Async[IO]` instance using revised `bracketCase`.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -85,7 +85,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
 
     IO.uncancelable { poll =>
       flatMap { a =>
-        val finalized = poll(use(a)).onCancel(safeRelease(a, Outcome.Canceled()))
+        val finalized = poll(IO.unit >> use(a)).onCancel(safeRelease(a, Outcome.Canceled()))
         val handled = finalized.onError { e =>
           safeRelease(a, Outcome.Errored(e)).handleErrorWith { t =>
             IO.executionContext.flatMap(ec => IO(ec.reportFailure(t)))
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -538,6 +538,12 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
         release: (A, OutcomeIO[B]) => IO[Unit]): IO[B] =
       acquire.bracketCase(use)(release)
 
+    override def bracketFull[A, B](acquire: Poll[IO] => IO[A])(use: A => IO[B])(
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -538,6 +538,10 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
         release: (A, OutcomeIO[B]) => IO[Unit]): IO[B] =
       acquire.bracketCase(use)(release)
```
RaasAhsan: We can leave `bracketCase` un-overrided so it just uses the default implementation now

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -538,6 +538,10 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
...
 
+    override def bracketFull[A, B](acquire: Poll[IO] => IO[A])(use: A => IO[B])(
```
RaasAhsan, line 541: This should work, but it's a bit weird to me because `bracketFull` is the more general operation. I would define `bracketFull` in the `IO` companion object (with the side channel reporting details) and use that here. Then `bracketCase` doesn't need to be overrided anymore, and the `bracket*` methods on `IO` can just forward to the `Async[IO]` instance for consistency

heka1024: Thank you for your comment. move `bracketFull` in `IO` companion object and use in `IO` abstract class. And use it in `Async[IO]` instance. Can you review it again?

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -369,6 +355,24 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
...
+      (a, out) => IO.uncancelable(_ => release(a, out))
+
+    IO.uncancelable { poll =>
+      acquire(poll).flatMap { a =>
+        val finalized = poll(IO.unit >> use(a)).onCancel(safeRelease(a, Outcome.Canceled()))
+        val handled = finalized.onError { e =>
+          safeRelease(a, Outcome.Errored(e)).handleErrorWith { t =>
+            IO.executionContext.flatMap(ec => IO(ec.reportFailure(t)))
+          }
+        }
```
djspiewak, line 370: Random thought that doesn't need to be handled in this PR: shouldn't we just put this side-channeling behavior into `IO#onError` itself, rather than having it live in `bracketFull`?

djspiewak (approved):
Looks good to me. Thanks for taking this on!

PR #1583, opened 2021-01-07
Deferred with different effects for source and sink

durban:
This generalizes `Deferred` to allow different effects for its `DeferredSource` and `DeferredSink` parts. So instead of `Deferred[F, A]` we could have `Deferred2[F, G, A]`, which we can `get` in  `F` and `complete` in `G`.

This is a work in progress. If the general idea is acceptable, I'll work on at least the following:

- The names are terrible (`Deferred2`, `in2`, ...); I'm open to suggestions.
- Scaladoc may need changes.

### DIFF
--- a/kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/pure.scala
+++ b/kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/pure.scala
@@ -278,6 +278,8 @@ object pure {
           override def complete(a: A): PureConc[E, Boolean] = mVar.tryPut[PureConc[E, *]](a)
 
           override def tryGet: PureConc[E, Option[A]] = mVar.tryRead[PureConc[E, *]]
+
+          override def tryGetG: PureConc[E, Option[A]] = tryGet
         }
 
       def start[A](fa: PureConc[E, A]): PureConc[E, Fiber[PureConc[E, *], E, A]] =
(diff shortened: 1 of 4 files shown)


durban:
Also, I'm unsure about the following:

- What to do with `GenConcurrent#deferred`? I left it alone for now.
- Related to the previous point: should I create a `Deferred.Make` typeclass (similar to `Ref.Make`)?

SystemFw:
My 2c:
 I would personally prefer to keep deferred simple, as it's now part of our primitive building blocks due to its inclusion in the typeclass hierarchy.  If we want to introduce this abstraction, it could be a different thing, with a different set of use cases (e.g `InteropChannel`), and I would also explore if it can be implemented generically for any `Async`, given `Dispatcher`. 

Let's see what others think :)

durban:
There is no need for `Dispatcher`. It works for any `F[_] : Async` and `G[_] : Sync`.

SystemFw:
yeah sorry, I was thinking "see if it can be done without relying on effect capture", but that's a pointless requirement since `Dispatcher` will require `Async` anyway :)

djspiewak:
It's a cool generalization, but it definitely feels like a much higher degree of complexity. I think if this were just a separate thing in the std lib then I'd be in favor of making a generalized `Deferred2` with two effects with a specialized `Deferred` type alias for one, but given that it's part of the core axioms, I think it's better to not open that pandora's box.

Issue #1473, opened 2020-12-01
Add iteration combinators to IO


djspiewak:
- `foreverM`
- `untilM`
- `whileM`
- `iterateWhile`/`iterateUntil`
- others-that-i've-probably-forgotten

Should have the same signatures and semantics as their counterparts within cats core.

pkowalcze:
I'd like to give it a go, if it's ok with you

djspiewak:
Do it!!

TimWSpence:
@pkowalcze are you still planning to work on this? If so that's fantastic! :) Let me know if you need a hand with anything.
If you're not then I might pick this up in a couple of days.

pkowalcze:
@TimWSpence Very sorry for the delay, I'm going to get it done as soon as possible. Could you verify that I understood the intent of this task correctly? Is it to have something like:

```
def untilM[G[_], A](f: IO[A])(cond: => IO[Boolean])(implicit G: Alternative[G]): IO[G[A]] = f.untilM(cond)
```

in the `IO` object?

TimWSpence:
Hey @pkowalcze no worries at all!! :)  Yes, exactly - it's so we have convenient syntax like `IO.untilM` instead of `Monad[IO].untilM`

Note that we want to be sure that it has the same semantics as the methods defined on the monad instance for `IO`. I _think_ the best way to do that is
```
object IO {
  def untilM(...) = _asyncForIO.untilM(...)
  
   def foreverM(...) = _asyncForIO.foreverM(...)

  //etc
}
```
so the implementation on `IO` just forwards to the `Monad` instance. @djspiewak might have a better idea though

djspiewak:
I think it's better to make these instance methods, rather than putting them on the object. Basically, take the following example:

```scala
IO.println("Hello, World").foreverM
```

Right now, that example requires `import cats.syntax.all._`. Make it so that example (and others like it involving Cats iteration combiantors) works without requiring the import. That's the task!

PR #1589, opened 2021-01-10
Added iteration combinators from cats core to IO

pkowalcze:
As per https://github.com/typelevel/cats-effect/issues/1473 - added the following iteration combinators to IO:

* foreverM
* whileM / whileM_
* untilM / untilM_
* iterateWhile
* iterateUntil

### DIFF
--- a/tests/shared/src/test/scala/cats/effect/IOSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/IOSpec.scala
@@ -764,11 +764,11 @@ class IOSpec extends IOPlatformSpecification with Discipline with ScalaCheck wit
       }
 
       "reliably cancel infinite IO.unit(s)" in real {
-        IO.unit.foreverM[Unit].start.flatMap(f => IO.sleep(50.millis) >> f.cancel).as(ok)
+        IO.unit.foreverM.start.flatMap(f => IO.sleep(50.millis) >> f.cancel).as(ok)
       }
 
       "reliably cancel infinite IO.cede(s)" in real {
-        IO.cede.foreverM[Unit].start.flatMap(f => IO.sleep(50.millis) >> f.cancel).as(ok)
+        IO.cede.foreverM.start.flatMap(f => IO.sleep(50.millis) >> f.cancel).as(ok)
       }
 
       "await uncancelable blocks in cancelation" in ticked { implicit ticker =>
--- a/tests/shared/src/test/scala/cats/effect/IOSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/IOSpec.scala
@@ -764,11 +764,11 @@ class IOSpec extends IOPlatformSpecification with Discipline with ScalaCheck wit
       }
 
       "reliably cancel infinite IO.unit(s)" in real {
-        IO.unit.foreverM.start.flatMap(f => IO.sleep(50.millis) >> f.cancel).as(ok)
+        IO.unit.foreverM[Unit].start.flatMap(f => IO.sleep(50.millis) >> f.cancel).as(ok)
       }
 
       "reliably cancel infinite IO.cede(s)" in real {
-        IO.cede.foreverM.start.flatMap(f => IO.sleep(50.millis) >> f.cancel).as(ok)
+        IO.cede.foreverM[Unit].start.flatMap(f => IO.sleep(50.millis) >> f.cancel).as(ok)
       }
 
       "await uncancelable blocks in cancelation" in ticked { implicit ticker =>
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -248,7 +248,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
     fiber
   }
 
-  def foreverM: IO[A] = Monad[IO].foreverM(this)
+  def foreverM[B]: IO[B] = Monad[IO].foreverM[A, B](this)
 
   def whileM[G[_]: Alternative, B >: A](p: IO[Boolean]): IO[G[B]] =
     Monad[IO].whileM[G, B](p)(this)
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -17,9 +17,11 @@
 package cats.effect
 
 import cats.{
+  Alternative,
   Applicative,
   Eval,
   Id,
+  Monad,
   Monoid,
   Now,
   Parallel,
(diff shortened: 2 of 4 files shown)


tests/shared/src/test/scala/cats/effect/IOSpec.scala
```diff
@@ -764,11 +764,11 @@ class IOSpec extends IOPlatformSpecification with Discipline with ScalaCheck wit
       }
 
       "reliably cancel infinite IO.unit(s)" in real {
-        IO.unit.foreverM.start.flatMap(f => IO.sleep(50.millis) >> f.cancel).as(ok)
+        IO.unit.foreverM[Unit].start.flatMap(f => IO.sleep(50.millis) >> f.cancel).as(ok)
```
pkowalcze: Not sure how to fix type inference here, any tips?

djspiewak: First off, this PR looks really nice, thank you! Second, I think maybe we might need to define `FiberIO` to be an actual trait which is covariant in its `A`. I'm guessing that's what the issue is here.

kubukoz: I'm not sure we can make `FiberIO` covariant without changing `Fiber` (which we also can't do): apparently you can't extend a trait and add covariance:

```scala
@ trait Fib[A]
trait Fib[defined trait Fib

@ trait FibIO[+A] extends Fib[A]
cmd1.sc:1: covariant type A occurs in invariant position in type [+A]ammonite.$sess.cmd0.Fib[A] {

} of trait FibIO
trait FibIO[+A] extends Fib[A]
      ^
Compilation Failed
```

Unless you meant something else :)

Anyway, I'm not sure if we had this discussion before, but we could probably just make `foreverM` return `IO[Nothing]`. IO is covariant itself, so it'll automatically be an `IO[B]` for all `B` anyway.

It won't look very cool if you e.g. `flatMap` on that directly, as you'll get unreachable code, but it will work in this particular test, since `f` will be a `FiberIO[Nothing]`.

djspiewak: I think I was thinking of using unsafeVariance in the extends

kubukoz: sounds... unsafe 😅

pkowalcze: I tried the following as a quick test:

```
type FiberIO[+A] = Fiber[IO, Throwable, A @uncheckedVariance]
```

and it still can't infer the type correctly

djspiewak: What about the trait? What I'm thinking is that the type alias might not be enough to convince Scalac to work properly

kubukoz: Apparently no difference:

```diff
diff --git a/core/shared/src/main/scala/cats/effect/IO.scala b/core/shared/src/main/scala/cats/effect/IO.scala
index 535e1055e..3023ad032 100644
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -169,8 +169,8 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   def productR[B](that: IO[B]): IO[B] =
     flatMap(_ => that)
 
-  def start: IO[FiberIO[A @uncheckedVariance]] =
-    IO.Start(this)
+  def start: IO[FiberIO[A]] = ???
+    // IO.Start(this)
 
   def background: ResourceIO[IO[OutcomeIO[A @uncheckedVariance]]] =
     Spawn[IO].background(this)
@@ -366,7 +366,8 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
   def racePair[A, B](
       left: IO[A],
       right: IO[B]): IO[Either[(OutcomeIO[A], FiberIO[B]), (FiberIO[A], OutcomeIO[B])]] =
-    asyncForIO.racePair(left, right)
+    // asyncForIO.racePair(left, right)
+    ???
 
   def ref[A](a: A): IO[Ref[IO, A]] = IO(Ref.unsafe(a))
 
diff --git a/core/shared/src/main/scala/cats/effect/package.scala b/core/shared/src/main/scala/cats/effect/package.scala
index d549c896c..50966c86d 100644
--- a/core/shared/src/main/scala/cats/effect/package.scala
+++ b/core/shared/src/main/scala/cats/effect/package.scala
@@ -68,7 +68,9 @@ package object effect {
   val Resource = cekernel.Resource
 
   type OutcomeIO[A] = Outcome[IO, Throwable, A]
-  type FiberIO[A] = Fiber[IO, Throwable, A]
+  trait FiberIO[+A]
+      extends Fiber[IO, Throwable, A @scala.annotation.unchecked.uncheckedVariance] {}
+
   type ResourceIO[A] = Resource[IO, A]
 
   type Deferred[F[_], A] = cekernel.Deferred[F, A]
diff --git a/tests/shared/src/test/scala/cats/effect/IOSpec.scala b/tests/shared/src/test/scala/cats/effect/IOSpec.scala
index d49cfd5ab..41dfbbcce 100644
--- a/tests/shared/src/test/scala/cats/effect/IOSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/IOSpec.scala
@@ -764,11 +764,11 @@ class IOSpec extends IOPlatformSpecification with Discipline with ScalaCheck wit
       }
 
       "reliably cancel infinite IO.unit(s)" in real {
-        IO.unit.foreverM[Unit].start.flatMap(f => IO.sleep(50.millis) >> f.cancel).as(ok)
+        IO.unit.foreverM.start.flatMap(f => IO.sleep(50.millis) >> f.cancel).as(ok)
       }
 
       "reliably cancel infinite IO.cede(s)" in real {
-        IO.cede.foreverM[Unit].start.flatMap(f => IO.sleep(50.millis) >> f.cancel).as(ok)
+        IO.cede.foreverM.start.flatMap(f => IO.sleep(50.millis) >> f.cancel).as(ok)
       }
 
       "await uncancelable blocks in cancelation" in ticked { implicit ticker =>

```

```
[E] [E2] tests/shared/src/test/scala/cats/effect/IOSpec.scala:771:40
[E]      missing parameter type
[E]      L771:         IO.cede.foreverM.start.flatMap(f => IO.sleep(50.millis) >> f.cancel).as(ok)
[E]                                                   ^
[E] [E1] tests/shared/src/test/scala/cats/effect/IOSpec.scala:767:40
[E]      missing parameter type
[E]      L767:         IO.unit.foreverM.start.flatMap(f => IO.sleep(50.millis) >> f.cancel).as(ok)
[E]                                                   ^
```

djspiewak: Fuuuuuuuuuuuuuuu

Okay I'm tentatively okay with having foreverM return Nothing. It's not a great experience though because the only way to get a specific type recovered is to ascribe. I'll poke at this a bit later today, but my guess is that there are only bad options.

pkowalcze: Thank you both for looking into this! I've pushed the version with `foreverM` returning `Nothing` for now, but please let me know if you do find a more pleasant solution.

kubukoz:
I'll merge to avoid this becoming stale, @djspiewak I suppose we can tackle the Nothing when somebody complains? :D (there's still `widen` too)

djspiewak:
I'll file an issue so we don't forget to think about it. The inconsistency with `IO.never` is part of what's top of mind for me right now.

PR #1591, opened 2021-01-12
Docs

TimWSpence:
Docs on `Concurrent` and `Temporal`. Also port docs for `Ref` and `Deferred` from CE2 with the minimal set of changes to make them not wrong


djspiewak (approved):
Nice work! Thank you!

PR #1597, opened 2021-01-16
Properly publish done resume tag

vasilmkd:
I've seen this bug pop up when I was doing optimization work on fiber stealing logic.


djspiewak:
LOL whoops; nice catch

PR #1598, opened 2021-01-16
Sbt and plugin updates

vasilmkd:


project/plugins.sbt
```diff
@@ -1,8 +1,8 @@
 libraryDependencies += "org.scala-js" %% "scalajs-env-selenium" % "1.1.0"
 
 addSbtPlugin("com.codecommit"     % "sbt-spiewak-sonatype"     % "0.19.3")
-addSbtPlugin("org.scala-js"       % "sbt-scalajs"              % "1.3.0")
+addSbtPlugin("org.scala-js"       % "sbt-scalajs"              % "1.4.0")
```
djspiewak: We should actually try to use the *oldest* version of scalajs that we can support. So if we can still build on 1.3.1, we should do so.

djspiewak (changes requested):

project/plugins.sbt
vasilmkd: I had no idea, sorry. Thanks.

vasilmkd: Done.

PR #1599, opened 2021-01-16
Please both scalafmt and Scala 3

vasilmkd:
Fix compilation warnings on Scala 3.

### DIFF
--- a/tests/shared/src/test/scala/cats/effect/IOSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/IOSpec.scala
@@ -187,8 +187,8 @@ class IOSpec extends IOPlatformSpecification with Discipline with ScalaCheck wit
           .attempt must completeAs(Left(TestException))
       }
 
-      "raise first bracket release exception if use effect succeeded" in ticked {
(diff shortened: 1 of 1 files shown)


tests/shared/src/test/scala/cats/effect/IOSpec.scala
```diff
@@ -187,8 +187,8 @@ class IOSpec extends IOPlatformSpecification with Discipline with ScalaCheck wit
           .attempt must completeAs(Left(TestException))
       }
 
-      "raise first bracket release exception if use effect succeeded" in ticked {
-        implicit ticker =>
+      "raise first bracket release exception if use effect succeeded" in ticked(
+        implicit ticker => {
```
djspiewak, line 191: Wow I hate this… We'll go with it for now but I almost wish we were doing `// noformat` or something

vasilmkd: Yeah, scalafmt gets confused around the 100 character line length and resorts to stupid alignment which Scala 3 dislikes.

PR #1600, opened 2021-01-17
IOFiber extends AtomicBoolean

vasilmkd:
It has been said that operations in the run loop, apart from allocating new `IO` objects, are instantaneous. This should help out with both allocations and speed.

Benchmarks:
```
series/3.x:
AsyncBenchmark.async     100  thrpt   10  19675.613 ± 379.777  ops/s
```

```
this PR:
AsyncBenchmark.async     100  thrpt   10  20235.886 ± 252.208  ops/s
```


djspiewak:
Very very cool

PR #1610, opened 2021-01-19
Fix rootJS project

vasilmkd:


djspiewak:
Holy wtf I can't believe we missed that

PR #1612, opened 2021-01-20
Sketch of takeAll and offerAll

RaasAhsan:
This PR (potentially) implements `takeAll`, `tryTakeAll`, `offerAll`, and `tryOfferAll`, as an attempt to squeeze some extra performance out when dealing with batched workloads. In particular, enqueueing and dequeueing multiple elements encounters a memory barrier on every call. The initial implementations are merely placeholders for us to agree on semantics; there are more efficient ones that directly manipulate the internal state. There are a few thorns to get through first though

The biggest obstacle in this PR right now is... the signature for these methods. Here's my first attempt:
```scala
import scala.collection.immutable.{Queue => ScalaQueue}
trait Queue[F[_], A] {
  def takeAll: F[ScalaQueue[A]]
  def tryTakeAll: F[ScalaQueue[A]]
  def offerAll(as: ScalaQueue[A]): F[Unit]
  def tryOfferAll(as: ScalaQueue[A]): F[Boolean]
}
```
The choice of `ScalaQueue` here is completely arbitrary from a user's perspective. My intention is that these operations could just directly manipulate the internal state which is also represented with a `ScalaQueue`. However, this poses a problem: it's really unlikely that a user is going to have their elements residing in a `ScalaQueue`. That means it's really likely that they end up performing a conversion and copy... which might defeat the purpose of doing this anyway because allocations can be as expensive as the memory barriers we want to avoid. I think it's pretty much the same deal no matter what collection type we choose.

Furthermore, the choice of `ScalaQueue` only really makes sense for a subset of our `Queue` implementations. `Dequeue` is another subtype of `Queue` whose internal state is represented by a `List`, so we'd be converting and copying on both batched enqueue and dequeue operations. Not too great. The bottom line is that there's likely to be copying, at the call-site, internally, or even both (imagine we chose `Vector` in the signature but a user has a `List` in their hands). 

The answer to this problem is inching closer and closer towards `fs2.Chunk`. That being said, here are some solutions I've got in mind:
1. Just pick the most common collection type and use that, without any regard to performance. These would really just be convenience methods (`takeAll` in particular is a nice one).
2. Implement `Chunk`. We could either reimplement, implement as a separate library (which is pulled in by `std`), or even pull the fs2 one out, since it's not reliant on `Stream`.
3. Forget doing this in CE and focus efforts on the `fs2.Queue` instead

It's probably worth collecting empirical evidence around what the performance impact of these copying operations and memory barriers are, but I think it's going to largely be sensitive to workload and the specific types involved

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Queue.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Queue.scala
@@ -234,9 +234,9 @@ object Queue {
         .flatten
         .uncancelable
 
-    override val tryTakeAll: F[ScalaQueue[A]] = ???
-
     override val takeAll: F[ScalaQueue[A]] = ???
+    
+    override val tryTakeAll: F[ScalaQueue[A]] = ???
   }
 
   private final class BoundedQueue[F[_], A](capacity: Int, state: Ref[F, State[F, A]])(
@@ -321,10 +321,10 @@ object Queue {
             fa.take.map(f)
           override def tryTake: F[Option[B]] =
             fa.tryTake.map(_.map(f))
-          override def tryTakeAll: F[ScalaQueue[B]] =
-            fa.tryTakeAll.map(_.map(f))
           override def takeAll: F[ScalaQueue[B]] =
             fa.takeAll.map(_.map(f))
+          override def tryTakeAll: F[ScalaQueue[B]] =
+            fa.tryTakeAll.map(_.map(f))
         }
     }
 }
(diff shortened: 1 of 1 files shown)


RaasAhsan:
@mpilquist any thoughts would be great here

djspiewak:
> Just pick the most common collection type and use that, without any regard to performance. These would really just be convenience methods (takeAll in particular is a nice one).

Back in scalaz-stream days of yore, the equivalent of `takeAll` produced a `Seq[A]`, basically for exactly this reason. `Seq` is horrible, but I honestly wonder if it's the best option here.

With that said, the big win is in avoiding the memory barrier and effect boundary (which implies a lot more than just a barrier). If we have to copy the internal structure rather than just projecting on it, that's not necessarily too much of a hinderance.

With that said, my two cents:

```scala
trait Queue[F[_], A] {
  def takeAll: F[Seq[A]]      // yolo (should this be NonEmptySeq?)
  def tryTakeAll: F[Seq[A]]
  def offerAll[G[_]: Foldable](as: G[A]): F[Unit]
  def tryOfferAll[G[_]: Foldable](as: G[A]): F[Boolean]
}
```

I could also be persuaded that `List` is the right answer.

> Implement Chunk. We could either reimplement, implement as a separate library (which is pulled in by std), or even pull the fs2 one out, since it's not reliant on Stream.

Hard pass. :-( Fs2 has needed to evolve `Chunk` several times over the years, at times in response to internal implementation details shifting, and I think they really do need the freedom to do that.

> Forget doing this in CE and focus efforts on the fs2.Queue instead

Also viable. We're definitely inching pretty close to fs2 here. In particular, even at a fundamental level, `takeAll` has some interesting backpressure implications that are subtle and difficult to reason about with raw cats effect.

djspiewak:
Closing this to clean up some older PRs. I do still want something like this, though it's obviously a bit harder now.

PR #1614, opened 2021-01-20
Sink and source for Dequeue

TimWSpence:


djspiewak:
Excellent!

std/shared/src/main/scala/cats/effect/std/Dequeue.scala
```diff
@@ -16,109 +16,22 @@
...
-  /**
-   * Attempts to dequeue an element from the front of the dequeue, if one is
-   * available without semantically blocking.
-   *
-   * @return an effect that describes whether the dequeueing of an element from
-   *         the dequeue succeeded without blocking, with `None` denoting that no
-   *         element was available
-   */
-  def tryTakeFront: F[Option[A]]
+trait Dequeue[F[_], A] extends Queue[F, A] with DequeueSource[F, A] with DequeueSink[F, A] {
```
djspiewak, line 27: Can we define an `Invariant` for `Dequeue`?

PR #1616, opened 2021-01-21
Invariant for dequeue

TimWSpence:


TimWSpence:
@djspiewak as requested :)

vasilmkd (approved):

vasilmkd:
Are there laws for Invariant that we can use?

TimWSpence:
@vasilmkd yup https://github.com/typelevel/cats/blob/master/laws/src/main/scala/cats/laws/InvariantLaws.scala

I was just lazy because none of the other `{Co, Contra, In}variant` instances for the queues check the laws and it seems pretty obvious that they are satisfied 😂 

We probably should check them 😭

vasilmkd:
That could be a nice first issue.

TimWSpence:
> That could be a nice first issue.

Good call! :) I created https://github.com/typelevel/cats-effect/issues/1618 but can't tag it with first issue or ce3

PR #1622, opened 2021-01-25
thread model docs

TimWSpence:
Donate https://timwspence.github.io/blog/posts/2021-01-12-threading-best-practices-cats-effect.html
as documentation on the thread model


djspiewak:
Niiiiiiiiiiice! :-D

PR #1625, opened 2021-01-27
fromCompletableFuture: unwrap CompletionExceptions

amesgen:
`CompletionException`s often occur when a downstream CompletableFuture is `compleExceptionally`-ed.


vasilmkd (approved):
I was just about to write a comment with the suggestion. Really not sure why Scala 2.12 has trouble inferring that. Either way, thank you and congratulations on your first contribution!

djspiewak:
Awesome thank you so much!

Issue #1613, opened 2021-01-20
Add startOn combinator


RaasAhsan:
This is pretty much an alias of `evalOn` plus `start`. We should define it on `IO` and `Async`. Whether it should be `fa.evalOn(ec).start` or `fa.start.evalOn(ec)`, I'm not sure yet, but they do commute

`backgroundOn` might also be a good one

Wosin:
Will happy to work on that :)

PR #1632, opened 2021-01-29
Add `backgroundOn` and `startOn` combinators.

Wosin:
This PR introduces new combinators `backgroundOn` and `startOn` as described in https://github.com/typelevel/cats-effect/issues/1613.

Not sure if we need any additional tests, if so I will be happy to add them.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -86,6 +86,9 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
 
   def startOn(ec: ExecutionContext): IO[FiberIO[A @uncheckedVariance]] = evalOn(ec).start
 
+  def backgroundOn(ec: ExecutionContext): ResourceIO[IO[OutcomeIO[A @uncheckedVariance]]] =
+    evalOn(ec).background
+
   def flatMap[B](f: A => IO[B]): IO[B] = IO.FlatMap(this, f)
 
   def flatten[B](implicit ev: A <:< IO[B]): IO[B] = flatMap(ev)
@@ -536,6 +539,11 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
     def startOn[A](fa: IO[A], ec: ExecutionContext): IO[FiberIO[A]] =
       fa.startOn(ec)
+
+    override def backgroundOn[A](
+        fa: IO[A],
+        ec: ExecutionContext): Resource[IO, IO[Outcome[IO, Throwable, A]]] =
+      fa.backgroundOn(ec)
 
     val executionContext: IO[ExecutionContext] =
       IO.executionContext
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -84,10 +84,10 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
 
   def evalOn(ec: ExecutionContext): IO[A] = IO.EvalOn(this, ec)
 
-  def startOn(ec: ExecutionContext): IO[FiberIO[A @uncheckedVariance]] = evalOn(ec).start
+  def startOn(ec: ExecutionContext): IO[FiberIO[A @uncheckedVariance]] = start.evalOn(ec)
 
   def backgroundOn(ec: ExecutionContext): ResourceIO[IO[OutcomeIO[A @uncheckedVariance]]] =
-    evalOn(ec).background
+    background.map(f => f.evalOn(ec))
 
   def flatMap[B](f: A => IO[B]): IO[B] = IO.FlatMap(this, f)
 
@@ -536,14 +536,6 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
     def evalOn[A](fa: IO[A], ec: ExecutionContext): IO[A] =
       fa.evalOn(ec)
-
-    def startOn[A](fa: IO[A], ec: ExecutionContext): IO[FiberIO[A]] =
-      fa.startOn(ec)
-
-    override def backgroundOn[A](
-        fa: IO[A],
-        ec: ExecutionContext): Resource[IO, IO[Outcome[IO, Throwable, A]]] =
-      fa.backgroundOn(ec)
 
     val executionContext: IO[ExecutionContext] =
       IO.executionContext
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -87,7 +87,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   def startOn(ec: ExecutionContext): IO[FiberIO[A @uncheckedVariance]] = start.evalOn(ec)
 
   def backgroundOn(ec: ExecutionContext): ResourceIO[IO[OutcomeIO[A @uncheckedVariance]]] =
-    background.map(f => f.evalOn(ec))
+    Resource.make(startOn(ec))(_.cancel).map(_.join)
 
   def flatMap[B](f: A => IO[B]): IO[B] = IO.FlatMap(this, f)
 
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -84,6 +84,8 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
 
   def evalOn(ec: ExecutionContext): IO[A] = IO.EvalOn(this, ec)
 
+  def startOn(ec: ExecutionContext): IO[FiberIO[A @uncheckedVariance]] = evalOn(ec).start
+
   def flatMap[B](f: A => IO[B]): IO[B] = IO.FlatMap(this, f)
 
   def flatten[B](implicit ev: A <:< IO[B]): IO[B] = flatMap(ev)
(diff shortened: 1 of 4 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
```diff
@@ -50,6 +50,10 @@ trait Async[F[_]] extends AsyncPlatform[F] with Sync[F] with Temporal[F] {
   // evalOn(executionContext, ec) <-> pure(ec)
   def evalOn[A](fa: F[A], ec: ExecutionContext): F[A]
 
+  def startOn[A](fa: F[A], ec: ExecutionContext): F[Fiber[F, Throwable, A]]
+
+  def backgroundOn[A](fa: F[A], ec: ExecutionContext): Resource[F, F[Outcome[F, Throwable, A]]]
+
```
djspiewak: These don't actually have to be abstract, since they have a trivial definition in terms of `evalOn` and such. In fact, if you do that, then you can remove all of the transformer-specific implementations below

kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
```diff
@@ -203,6 +207,15 @@ object Async {
...
 
+    override def startOn[A](
+        fa: OptionT[F, A],
+        ec: ExecutionContext): OptionT[F, Fiber[OptionT[F, *], Throwable, A]] =
+      start(evalOn(fa, ec))
+
+    override def backgroundOn[A](fa: OptionT[F, A], ec: ExecutionContext)
+        : Resource[OptionT[F, *], OptionT[F, Outcome[OptionT[F, *], Throwable, A]]] =
+      background(evalOn(fa, ec))
+
```
djspiewak: Like this. All of these can go away if we give `startOn`/`backgroundOn` a default concrete implementation.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -1121,6 +1121,15 @@ abstract private[effect] class ResourceAsync[F[_]]
...
 
+  override def startOn[A](
+      fa: Resource[F, A],
+      ec: ExecutionContext): Resource[F, Fiber[Resource[F, *], Throwable, A]] =
+    start(evalOn(fa, ec))
+
+  override def backgroundOn[A](fa: Resource[F, A], ec: ExecutionContext)
+      : Resource[Resource[F, *], Resource[F, Outcome[Resource[F, *], Throwable, A]]] =
+    background(evalOn(fa, ec))
+
```
djspiewak: Also here

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -532,6 +537,14 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
...
       fa.evalOn(ec)
 
+    def startOn[A](fa: IO[A], ec: ExecutionContext): IO[FiberIO[A]] =
+      fa.startOn(ec)
+
+    override def backgroundOn[A](
+        fa: IO[A],
+        ec: ExecutionContext): Resource[IO, IO[Outcome[IO, Throwable, A]]] =
+      fa.backgroundOn(ec)
+
```
djspiewak: I think we actually don't need to override these, because the implementation is directly identical to what we would have in `IO`-proper.

kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
```diff
@@ -50,6 +50,10 @@ trait Async[F[_]] extends AsyncPlatform[F] with Sync[F] with Temporal[F] {
   // evalOn(executionContext, ec) <-> pure(ec)
   def evalOn[A](fa: F[A], ec: ExecutionContext): F[A]
 
+  def startOn[A](fa: F[A], ec: ExecutionContext): F[Fiber[F, Throwable, A]]
```
djspiewak: Also, can you add these to the `AsyncOps` type so that they're available using infix syntax? (e.g. `fa.startOn(ec)`)

djspiewak (changes requested):
Nice! Just a few changes but this looks very good

djspiewak:
I had another thought! I think we should do `.start.evalOn(ec)` for the implementation, rather than `.evalOn(ec).start`. The scenario I'm thinking of here is if we imagine that the *outer* EC is resource-constrained in some way, shifting to the new one before `start`ing and then shifting back is going to yield more intuitive results than the other way around, since it gets the scheduling work onto the new pool.

Wosin:
Sure, but I assume we should do exactly the same for `backgroundOn` am I right here ??

djspiewak:
> Sure, but I assume we should do exactly the same for backgroundOn am I right here ??

Yep!

Wosin:
I will be happy to rebase the commits since I've created quite a lotte of them, if that's ready to be merged.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -84,6 +84,11 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
 
   def evalOn(ec: ExecutionContext): IO[A] = IO.EvalOn(this, ec)
 
+  def startOn(ec: ExecutionContext): IO[FiberIO[A @uncheckedVariance]] = start.evalOn(ec)
+
+  def backgroundOn(ec: ExecutionContext): ResourceIO[IO[OutcomeIO[A @uncheckedVariance]]] =
+    background.map(f => f.evalOn(ec))
```
djspiewak: Ah! This is subtle but actually wrong. It will take the current `IO[A]` and run it *on the current pool* and then run the `join` on the specified pool. I believe you want the following.

```suggestion
    evalOn(ec).background
```

kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
```diff
@@ -50,6 +50,14 @@ trait Async[F[_]] extends AsyncPlatform[F] with Sync[F] with Temporal[F] {
   // evalOn(executionContext, ec) <-> pure(ec)
   def evalOn[A](fa: F[A], ec: ExecutionContext): F[A]
 
+  def startOn[A](fa: F[A], ec: ExecutionContext): F[Fiber[F, Throwable, A]] =
+    evalOn(start(fa), ec)
+
+  def backgroundOn[A](
+      fa: F[A],
+      ec: ExecutionContext): Resource[F, F[Outcome[F, Throwable, A]]] =
+    background(fa).map(f => evalOn(f, ec))
```
djspiewak: Same problem as above!

```suggestion
    background(evalOn(f, ec))
```

kernel/shared/src/main/scala/cats/effect/kernel/syntax/AsyncSyntax.scala
```diff
@@ -29,4 +29,11 @@ final class AsyncOps[F[_], A] private[syntax] (private[syntax] val wrapped: F[A]
     extends AnyVal {
   def evalOn(ec: ExecutionContext)(implicit F: Async[F]): F[A] =
     Async[F].evalOn(wrapped, ec)
+
+  def startOn(ec: ExecutionContext)(implicit F: Async[F]): F[Fiber[F, Throwable, A]] =
+    Async[F].startOn(wrapped, ec)
+
+  def backgroundOn(ec: ExecutionContext)(
+      implicit F: Async[F]): Resource[F, F[Outcome[F, Throwable, A]]] =
+    Async[F].backgroundOn(wrapped, ec)
```
djspiewak, line 38: :+1:

djspiewak (changes requested):
Hey! Sorry for my tardiness. It's been a week. :-) Some slight adjustments and I think this is good to go!

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -84,6 +84,11 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
 
   def evalOn(ec: ExecutionContext): IO[A] = IO.EvalOn(this, ec)
 
+  def startOn(ec: ExecutionContext): IO[FiberIO[A @uncheckedVariance]] = start.evalOn(ec)
```
kubukoz, line 87: This would probably need a similar change to what Daniel mentioned below:

```suggestion
  def startOn(ec: ExecutionContext): IO[FiberIO[A @uncheckedVariance]] = evalOn(ec).start
```

kubukoz:
> I had another thought! I think we should do .start.evalOn(ec) for the implementation, rather than .evalOn(ec).start. The scenario I'm thinking of here is if we imagine that the outer EC is resource-constrained in some way, shifting to the new one before starting and then shifting back is going to yield more intuitive results than the other way around, since it gets the scheduling work onto the new pool.

Now I'm lost :D in `backgroundOn`, you suggested to do `evalOn` first, then do `background` on that?

djspiewak:
Oh crap I think I inverted it. The first suggestion was right. My background thing was wrong. I'll review more carefully today to disambiguate

kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
djspiewak: Correction, I think this is a more accurate suggestion:

```suggestion
    Resource.make(evalOn(start, ec))(_.cancel).map(_.join)
```

Basically I don't think there's a way to reuse the implementation of `background` itself because of the inversion problem.

core/shared/src/main/scala/cats/effect/IO.scala
djspiewak: A more accurate suggestion:

```suggestion
    Resource.make(start.evalOn(ec))(_.cancel).map(_.join)
```

core/shared/src/main/scala/cats/effect/IO.scala
kubukoz: ignore ;)

core/shared/src/main/scala/cats/effect/IO.scala
Wosin: Okay, sounds cool. Will fix. Btw. I think we can use the `startOn` here.

kubukoz (approved):

djspiewak:
Thank you!

PR #1644, opened 2021-02-04
Docs

TimWSpence:
Docs for dispatcher, supervisor, resource and hotswap

### DIFF
--- a/docs/datatypes/dispatcher.md
+++ b/docs/datatypes/dispatcher.md
@@ -3,12 +3,16 @@ id: dispatcher
 title: dispatcher
 ---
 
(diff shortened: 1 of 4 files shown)


TimWSpence:
I'm not sure if we have an articulated philosophy on our documentation but the principle I've currently been following is that most things have very good, detailed scaladocs and so the role of the docs is to be more discursive and give a high-level understanding of how the concept fits into the bigger picture. You can then go to the scaladocs for the fine details, rather than duplicating them into the docs.

RaasAhsan (approved):
Looks great!

PR #1663, opened 2021-02-08
Renamed implicits.global to IORuntime.Implicits.global

djspiewak:
This is just for consistency with `Future`. It also has the pleasing side-benefit of making the call site EVEN UGLIER! Jk. Mostly.

The only purpose of this implicit is to ease migration from codebases which are predominantly using `Future` today. Thus, even while I subjectively prefer the aesthetics of `cats.effect.unsafe.implicits.global` much better, I have to conclude that `cats.effect.unsafe.IORuntime.Implicits.global` will be more familiar to that user segment.


mpilquist:
Big :-1: from me. We're trading consistency with cats (e.g. `import cats.implicits._`) for consistency with `Future`.

djspiewak:
Oooooooh! Bikeshed fight. Let's GOOOOO

Okay so my reasoning here is that people already bought into the Cats ecosystem probably won't ever see this *anyway*, since they'll use `IOApp` and `Dispatcher` and all will be well in the world. This feature is almost specifically and exclusively targeted at people who are on the periphery and haven't bought in yet.

Second, `cats.implicits._` is the idiom, yes, but it's also a dying idiom since Cats 2.2.0. `cats.syntax.all._` is the recommendation at this point, which makes things a bit orthogonal to this PR.

vasilmkd:
Love the label. Indifferent about the change.

kubukoz:
honestly I never remember there's `unsafe` in it. Not sure how users will feel, but I'd optimize for REPLiness... `cats.effect.yolo._` would do it for me, but I can see why this might be controversial 😅

djspiewak:
IMO consistency with cats (and our own `implicits._`) is probably more important here, as @mpilquist said. I'll close this.

Issue #1352, opened 2020-10-23
Rework AutoCedeK into FlattenK


djspiewak:
It occurred to me that `AutoCedeK` is really just a specialized variant of `FlattenK`, but the only specialization is the name. We should rename it, group it with `FlatMapK` and `MapK`, and special-case the `IO#flatten` operation, since that will result in a noticeable performance win in such scenarios (relative to running through `flatMap`).

This is a run loop task, but a very straightforward one (mostly just renaming stuff) and a decent-ish introduction to how the implementation works. Someone who hasn't worked on the run loop yet (but wants to!) please take this one!

stephen-lazaro:
If no one else has interest, I'd be interested in looking at this.

djspiewak:
Take it!

PR #1665, opened 2021-02-09
AutoCedeR

vasilmkd:
Auto cede is a preparation to relinquish the hold over the run loop, which makes it a _resumption_ mechanism, not a _continutation_. (Also it would have never been visible in `failed`). This occurred to me when I was looking at #1352. Unfortunately I do not see the resemblance between auto cede and a hypothetical `FlattenK`. `FlattenK` is literally a `FlatMap` with an `identity` bind, while `AutoCede` gives up control over the run loop.

I understand that `FlattenK` could be a nice optimization, but it would look closer to what we have already for `Attempt`.

Thoughts?


djspiewak:
Oh! Nice. This should be faster. Also I believe you (and @TimWSpence!) are correct that my initial intuition about being able to unify it with `flatten` was in err.

PR #1675, opened 2021-02-11
Reclaim unused fiber stack memory

vasilmkd:
In case the stacks grow too much, it would be good to keep them in line from time to time.

If a stack uses less than a quarter of the array capacity, the capacity is halved. This should avoid unnecessary allocations back and forth (thrashing) when operating close to the capacity.

The checks happen at async boundaries, which is often enough since we have auto yielding and should not be noticeable at all when it comes to performance.

### DIFF
--- a/core/js/src/main/scala/cats/effect/IOFiberConstants.scala
+++ b/core/js/src/main/scala/cats/effect/IOFiberConstants.scala
@@ -48,4 +48,10 @@ private[effect] object IOFiberConstants {
   val ContStateInitial: Int = 0
   val ContStateWaiting: Int = 1
   val ContStateResult: Int = 2
+
+  // Initial sizes of fiber stacks
+  val ContsStackInitSize: Int = 16
+  val ObjectStateStackInitSize: Int = 16
+  val FinalizersStackInitSize: Int = 16
+  val ECStackInitSize: Int = 2
 }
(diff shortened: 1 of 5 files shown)


djspiewak:
Very very very nice! My strong guess is that this is worth doing, but I think it deserves a quick round of benchmarks just as a sanity check. I think you're correct that doing it at async boundaries is fine and should make the branch prediction penalties insignificant.

vasilmkd:
Ok the benchmarks are really bad after this change.

Before:
```
Benchmark                     (size)   Mode  Cnt   Score   Error    Units
WorkStealingBenchmark.sched  1000000  thrpt   20  15.037 ± 1.585  ops/min
```

After:
```
Benchmark                     (size)   Mode  Cnt   Score   Error    Units
WorkStealingBenchmark.sched  1000000  thrpt   20  12.773 ± 2.003  ops/min
```

djspiewak:
😢

djspiewak:
Most (stable) fibers fall into one of two categories:

- cyclic ongoing patterns
- expiring one-time patterns

In the case of the latter, the fiber will terminate and die, so we really don't care about the stack long-term. In the case of the former, the patterns will repeat, so there's no particular need to clean things up. Think of it more like heap warming: the fiber that is cycling between high and low stack needs will eventually need the high water mark again.

The only other class of fiber pattern I can think of would be one which is leaking: it goes in cycles but consumes more each time. That will just eventually crash. Reclamation only makes sense for a fiber which is cyclic and decreasing: it hit some high water mark and now it's cycling and going down. I just don't see that as a common pattern really at all.

PR #1677, opened 2021-02-11
feat: utils for measuring effect time

### LINKED ISSUE #1672: Add utilities for measuring an effect's execution time
Something of the sort:

```scala
//in IO[A]

def timed: IO[(FiniteDuration, A)]

//in GenTemporal syntax

def timed: F[(FiniteDuration, A)]
```

After it lands in series/3.x, I think it would be a welcome addition to CE2 as well (for effects that have a `Clock[F]`, maybe - it would require a new syntax class).

patseev:
Closes #1672

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
@@ -71,14 +71,6 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
     val timeoutException = raiseError[A](ev(new TimeoutException(duration.toString)))
     timeoutTo(fa, duration, timeoutException)
   }
-
-  /**
-   * Returns an effect that completes with the result of the source together
-   * with the duration that it took to complete.
-   */
-  def timed[A](fa: F[A]): F[(FiniteDuration, A)] =
-    map3(realTime, fa, realTime)((startTime, a, endTime) => (endTime.minus(startTime), a))
-
 }
 
 object GenTemporal {
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
@@ -72,6 +72,13 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
     timeoutTo(fa, duration, timeoutException)
   }
 
+  /**
+   * Returns an effect that completes with the result of the source together
+   * with the duration that it took to complete.
+   */
+  def timed[A](fa: F[A]): F[(FiniteDuration, A)] =
+    map3(realTime, fa, realTime)((startTime, a, endTime) => (endTime.minus(startTime), a))
+
 }
 
 object GenTemporal {
--- a/kernel/shared/src/main/scala/cats/effect/kernel/syntax/ClockSyntax.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/syntax/ClockSyntax.scala
@@ -16,11 +16,17 @@
 
 package cats.effect.kernel.syntax
 
-trait AllSyntax
-    extends MonadCancelSyntax
-    with GenSpawnSyntax
-    with GenTemporalSyntax
-    with GenConcurrentSyntax
-    with AsyncSyntax
-    with ResourceSyntax
-    with ClockSyntax
+import cats.effect.kernel.Clock
+
+import scala.concurrent.duration.FiniteDuration
+
+trait ClockSyntax {
+  implicit def clockOps[F[_], A, E](wrapped: F[A]): ClockOps[F, A] =
+    new ClockOps(wrapped)
+}
+
+final class ClockOps[F[_], A] private[syntax] (private val wrapped: F[A]) extends AnyVal {
+
+  def timed(implicit F: Clock[F]): F[(FiniteDuration, A)] =
+    F.timed(wrapped)
+}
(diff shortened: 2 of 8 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
```diff
@@ -72,6 +72,13 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
     timeoutTo(fa, duration, timeoutException)
   }
 
+  /**
+   * Returns an effect that completes with the result of the source together
+   * with the duration that it took to complete.
+   */
+  def timed[A](fa: F[A]): F[(FiniteDuration, A)] =
+    map3(realTime, fa, realTime)((startTime, a, endTime) => (endTime.minus(startTime), a))
```
djspiewak: Close! We *definitely* want to use `monotonic` here. For starters, it's much faster and higher resolution than `realTime`, but additionally it doesn't suffer from re-sync issues. If someone changes the system time between `startTime` and `endTime`, for example, the result here will be inaccurate. `monotonic` doesn't have that problem

```suggestion
    map3(monotonic, fa, monotonic)((startTime, a, endTime) => (endTime.minus(startTime), a))
```

Also given that we're doing this with `Applicative`, I *think* it can just be on `Clock` instead!

djspiewak (changes requested):
This is fantastic. Just the one small change about `monotonic` vs `realTime`, and additionally I think it can be moved up to `Clock` since it doesn't rely on `sleep` or anything concurrent.

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
patseev: Done! Thank you for your feedback

patseev: Appears I broke something, I'll check a bit later

vasilmkd: You just need to do `scalafmtAll` in `sbt`.

kernel/shared/src/main/scala/cats/effect/kernel/syntax/ClockSyntax.scala
```diff
@@ -0,0 +1,33 @@
...
+  implicit def clockOps[F[_], A, E](wrapped: F[A])(implicit F: Clock[F]): ClockOps[F, A] =
+    new ClockOps(wrapped)
+}
+
+final class ClockOps[F[_], A] private[syntax] (private val wrapped: F[A])
+  extends AnyVal {
+
+  def timed(implicit F: Clock[F]): F[(FiniteDuration, A)] =
+    F.timed(wrapped)
+}
```
kubukoz: minor: I'd suggest adding a newline here (POSIX-compliant files have that, it helps in printing etc.)

kernel/shared/src/main/scala/cats/effect/kernel/syntax/ClockSyntax.scala
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package cats.effect.kernel.syntax
+
+import cats.effect.kernel.Clock
+
+import scala.concurrent.duration.FiniteDuration
+
+trait ClockSyntax {
+  implicit def clockOps[F[_], A, E](wrapped: F[A])(implicit F: Clock[F]): ClockOps[F, A] =
```
kubukoz: This constraint seems unused to me - we require another Clock at the callsite of `timed`. I would remove this one and keep the one on the method. If you saw this somewhere else in CE, please update these ones as well! (or create an issue to do so before 3.0 final) :)

kubukoz (changes requested):

kernel/shared/src/main/scala/cats/effect/kernel/syntax/ClockSyntax.scala
djspiewak: The constraint is redundant *but* it makes the type inference better. Cats uses the same pattern.

kubukoz: hmm. It also makes the extension methods less visible in IDEs (or the compiler really, since you'll get a missing symbol error instead of a missing implicit error), but I'll defer to you

patseev:
Made all requested changes. 
@djspiewak @kubukoz should I target 2.x with the same utils?
I suppose I won't be able to put it into Clock because there's no applicative there, but I could create similar `ClockSyntax` with implicit requirement of `Applicative` to make it work. What do you think?

kubukoz:
I would like to see that, but I'm not sure we should do that...

kernel/shared/src/main/scala/cats/effect/kernel/syntax/ClockSyntax.scala
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package cats.effect.kernel.syntax
+
+import cats.effect.kernel.Clock
+
+import scala.concurrent.duration.FiniteDuration
+
+trait ClockSyntax {
+  implicit def clockOps[F[_], A, E](wrapped: F[A]): ClockOps[F, A] =
```
joroKr21: Unused type parameter
```suggestion
  implicit def clockOps[F[_], A](wrapped: F[A]): ClockOps[F, A] =
```

patseev: thank you! missed it

kernel/shared/src/main/scala/cats/effect/kernel/syntax/ClockSyntax.scala
joroKr21: It makes sense if you want to infer the type of `E` from the typeclass instance but here it's not used.

kubukoz (approved):
I think we got it! @djspiewak I think your concerns were addressed, but feel free to blame me if anything is still missing

PR #1682, opened 2021-02-12
Null out reference to last stolen fiber

vasilmkd:
Not a bug, just a nice, consistent cleanup.

I observed this when stress testing `fs2-netty`.

When fibers are stolen, the last fiber in the chain of stolen fibers is directly executed. Here, I forgot to null out the reference to this fiber in the local queue which does the stealing. While this exact fiber is executed, a concurrent worker thread might come and steal all other fibers from this same queue. At this point, if there is no more work, all threads will go to sleep but this reference will remain set in the queue.

This is not a bug, safety is still guaranteed because this reference is technically "out" of the circular buffer. It's just that the memory for the already executed fiber will not be reclaimed until new work arrives in the pool, which might take a long time. During this time, that fiber is still strongly referenced in memory and the GC cannot reclaim it.


djspiewak:
Awesome catch!

Issue #1652, opened 2021-02-07
Add Resource#race


djspiewak:
Mostly we want to add this to the `Resource` class and companion object (similar to `IO`) for discoverability reasons. The implementation is already available on the `Concurrent` instance, and thus the methods added here would just delegate to that implementation, similar to what `IO` does.

patseev:
I'd like to take this one. Should I target 2.x or 3.x or both?

patseev:
To save time, here's a sketch of implementation. Do I understand correctly that we're racing allocations? Implementation is very similar to `Resource.both` but relies on `race` instead of `parTupled`

```
def race[B](that: Resource[F, B])(implicit F: Concurrent[F]): Resource[F, Either[A, B]] = {
    def allocate[C](r: Resource[F, C], storeFinalizer: (F[Unit] => F[Unit]) => F[Unit]): F[C] =
      r.fold(
        _.pure[F],
        release => storeFinalizer(MonadCancel[F, Throwable].guarantee(_, release))
      )

    val bothFinalizers = Ref.of[F, (F[Unit], F[Unit])](().pure[F] -> ().pure[F])

    Resource.make(bothFinalizers)(_.get.flatMap(_.parTupled).void).evalMap { store =>
      val leftStore = (f: F[Unit] => F[Unit]) => store.update(_.leftMap(f))
      val rightStore = (f: F[Unit] => F[Unit]) => store.update(t => (t._1, f(t._2)))

      F.race(allocate(this, leftStore), allocate(that, rightStore))
    }
  }
  ```

djspiewak:
> I'd like to take this one. Should I target 2.x or 3.x or both?

Just 3.x! :-) It would run into a bunch of problems on 2.x as far as I can tell.

Regarding the implementation, I think you can actually just delegate straight to the `Async` for `Resource` the same way that `IO` does with *its* `race` implementation. So really all that is needed here is a forwarder.

patseev:
Sweet. Then I propose I'll simplify the existing `both` via delegate as well

Issue #1576, opened 2021-01-02
Decide what we're going to do with Resource#parZip


djspiewak:
It's somewhat redundant with `both` (defined in `Concurrent`) now, though it is *probably* slightly faster. The real problem though is that `parZip` is actually *inconsistent* with `both` in that it runs finalizers in parallel, whereas `both` runs them sequentially.

Daenyth:
Why does `both` run finalizers in sequence? Can that be fixed?

SystemFw:
I kinda would have expected the Parallel newtypes to be CommutativeApplicative, and that, for Resource, order of finalisation is part of equivalence, which make sequential release on `both` weird.

Validated breaks the CommutativeApplicative assumption admittedly , but I think not having resources released in parallel actually has practical consequences and breaks the principle of least surprise

djspiewak:
I agree with the commutative applicative argument. It's also worth noting that Validated does work here so long as the inner E forms a commutative monoid. But at any rate…

What I don't see is how to make this happen with Resource without bespoke overrides, which make me nervous in this case. The problem is that we flatMap inside Fiber#joinWith, and we don't have any knowledge of the other fibers when we do so. Even if we _did_ have knowledge of the other fibers, it's not at all clear to me that we could get the finalizers to run in parallel with Resource specifically.

What is it about Resource in particular which gives rise to this property? In abstract, I mean. Are we not likely to see other effects with the same issue in Parallel?

djspiewak:
In light of everything with `EitherT`, I'm strongly inclined to say that `parZip` (now `both`) should just have the same semantics as the `Spawn[Resource[F, *]].both` implementation. Which is to say, sequential finalizers. Any strong objections?

SystemFw:
> Which is to say, sequential finalizers. Any strong objections?

it still sounds very very weird to me

djspiewak:
> it still sounds very very weird to me

I agree in a vacuum that parallel finalization makes *sense*, but I don't see a way that it can be derived from the more primitive semantics without totally special-casing this one specific scenario. It's basically like the whole `Parallel[EitherT[F, E` thing: one set of semantics arises naturally from base primitives, while the other set of semantics requires a bespoke specialized interpretation of the meaning of `ap` *in this context*, which seems like a red flag.

SystemFw:
there is another scarier possibility, which is that the primitive semantics of `Async[Resource` ultimately don't make sense, and we're looking at the first example of why.
[...]

PR #1683, opened 2021-02-13
addition of Resource#race

patseev:
Related issue: https://github.com/typelevel/cats-effect/issues/1652

Implements `Resource#race` as a delegate to `Async#race` for discoverability reasons.

I have two questions though:

- Tests. I've found that there are tests for `Async[Resource[F, *].race` already. Do I need to duplicate these tests in the Resource spec?
- I've tried to replace manual implementation of `Resource.both` to instead rely on delegation to `Concurrent#both` but tests started to fail. Does this hint that `Concurrent` implementation for Resource is not correct? Or maybe the semantics of the Resouce.both spec are invalid? I will revert this change and then will explore what's up with that.

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -270,8 +270,26 @@ sealed abstract class Resource[F[_], +A] {
    */
   def both[B](
       that: Resource[F, B]
-  )(implicit F: Concurrent[F]): Resource[F, (A, B)] =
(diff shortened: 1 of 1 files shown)


patseev:
Found your discussion about the difference in semantics of release.
https://github.com/typelevel/cats-effect/issues/1576
Now it makes sense. Will revert "both".

patseev:
Also, as per discussion in https://github.com/typelevel/cats-effect/issues/1576, I can include `Async[Resource]#both` override if that's something that's already agreed upon

djspiewak (approved):
Outstanding thank you!

PR #1685, opened 2021-02-14
Add g8 template snippet to readme

kubukoz:
I'm not sure this is the best way to do it, but it's a start - suggestions welcome!


djspiewak:
Btw haven't forgotten about this. Planning to come back to it this week

PR #1686, opened 2021-02-14
Scalafix migration for the update to 3.0.0

### LINKED ISSUE #1049: Create a migration scalafix
I have no idea how scalafix stuff works, but given it's limitations, I do suspect it can handle most of the 2 -> 3 migration in many cases. At least it's worth a shot!

fthomas:
This adds a Scalafix migration for the update to 3.0.0. It is still WIP because not all possible rewrites mentioned by Daniel in https://github.com/typelevel/cats-effect/issues/1049#issuecomment-777126404 are implemented. I'm opening this PR anyway so that people can better keep track of the status of the migration and maybe gather further rewrites.

DONE:

* [x] `Async#async() -> async_`
* [x] `Bracket -> MonadCancel`
* [x] `Bracket#uncancelable(a) -> MonadCancel#uncancelable(_ => a)`
* [x] `Bracket#guarantee(a)(b) -> MonadCancel#guarantee(a, b)`
* [x] `BracketThrow -> MonadCancelThrow`
* [x] `remove Blocker`
* [x] `Blocker[F] -> Resource.unit[F]`
* [x] `Blocker#delay -> Sync[F].blocking`
* [x] `remove implicit Concurrent[IO]`
* [x] `remove ContextShift`
* [x] `ContextShift#shift -> Spawn[F].cede`
* [x] `IO.async() -> async_`
* [x] `IO.suspend() -> defer`
* [x] `Resource#parZip() -> both`
* [x] `Resource.liftF() -> eval`
* [x] `Sync#suspend(). -> defer`
* [x] `Timer -> Temporal`
* [x] `concurrent.Deferred -> Deferred`
* [x] `concurrent.Ref -> Ref`
* [x] `concurrent.Semaphore -> std.Semaphore`
* [x] `remove implicit Parallel[F] if implicit Concurrent[F] + import cats.effect.implicits._` (https://github.com/typelevel/cats-effect/pull/1686#pullrequestreview-590070078)
* [x] `cats.effect.ApplicativeThrow -> cats.ApplicativeThrow` (#1788)
* [x] `cats.effect.MonadThrow -> cats.MonadThrow` (#1788)

TODO:

* [ ] `cancelableF -> async`, as well as the other variations of it
* [ ] `join -> joinAndEmbed`
* [ ] `ExitCase -> Outcome`
* [ ] `async(...).guarantee(shift) -> async(...)`
* [ ] There is some restructuring and renaming within the Outcome algebra vis-a-vis ExitCase. Completed vs Succeeded(_), for example

The migration can already be tested in your project by [addingt sbt-scalafix to the build](https://scalacenter.github.io/scalafix/docs/users/installation) and running the following commands:
```sh
echo "ThisBuild / scalacOptions += \"-P:semanticdb:synthetics:on\"" > scalafix.sbt

sbt ";scalafixEnable ;scalafixAll github:fthomas/cats-effect/v3_0_0?sha=topic/scalafix"
```

Once this is merged, I'll add it to Scala Steward so that it runs the migration on updates to 3.0.0.

Closes #1049.


scalafix/rules/src/main/scala/fix/v3_0_0.scala
```diff
@@ -0,0 +1,131 @@
...
+    val Bracket_guarantee_M = SymbolMatcher.exact("cats/effect/Bracket#guarantee().")
+    val Bracket_uncancelable_M = SymbolMatcher.exact("cats/effect/Bracket#uncancelable().")
+    val ContextShift_M = SymbolMatcher.normalized("cats/effect/ContextShift.")
+
+    Patch.replaceSymbols(
+      "cats/effect/Async#async()." -> "async_",
+      "cats/effect/package.BracketThrow." -> "cats/effect/MonadCancelThrow.",
+      "cats/effect/Bracket." -> "cats/effect/MonadCancel.",
+      "cats/effect/IO.async()." -> "async_",
+      "cats/effect/IO.suspend()." -> "defer",
```
djspiewak, line 31: Can also do `cats/effect/Sync.suspend().` to `defer`.

djspiewak (commented):
This is very very cool! Thank you for taking this on!

One thing I thought about while going through it… if someone has an implicit `Concurrent[F]` *and* an implicit `Parallel[F]` (a common scenario with CE2), we want to get rid of the `Parallel` and make sure they have `import cats.effect.implicits._`. If they *already* have the implicits import then failure to remove the redundant constraint will result in weird ambiguities (iirc?).

That one seems like it might be too complex for scalafix so we could just skip it.

scalafix/rules/src/main/scala/fix/v3_0_0.scala
fthomas: Done in 14fab138d3d7c735fa61ba546e2144222cb6a33a.

scalafix/v3_0_0/output/src/main/scala/fix/ContextShiftRewrites.scala
```diff
@@ -10,4 +12,6 @@ object ContextShiftRewrites {
   def f3[F[_]](implicit F: Sync[F]): Int = 0
 
   def f4[F[_]](implicit F1: Sync[F], F2: Sync[F]): Int = 0
```
kubukoz, line 14: random note, won't this introduce a conflict for `implicitly[Sync[F]]` inside `f4`?

fthomas: Sure, but this is not the important bit of this test case. If you look at the corresponding input file, you'll see that the scalafix removed a `ContextShift[F]` parameter in the middle of the parameter list and that is the only thing we're testing here. The types of `F1` and `F2` are irrelevant.

fthomas:
This is now ready for review.

I had trouble implementing some of the proposed rewrites (those that are not checked in the list above) but don't want to delay this PR further because of them.

scalafix/README.md
```diff
@@ -0,0 +1,18 @@
+# Scalafix rules for Cats Effect
+
+## How to use
+
+1. [Install the Scalafix sbt plugin](https://scalacenter.github.io/scalafix/docs/users/installation)
+1. Run the rules appropriate to your Cats Effect version (see below)
+
+## Migration to Cats Effect 3.0.0
```
kubukoz, line 8: we should probably mention you need to do this before upgrading CE3 itself

fthomas: Good suggestion - I amended the instructions. It would also be nice to link to the migration guide (#1687) once it is merged.

kubukoz: I'm going to link this in the migration guide too 😂

djspiewak:
This looks *absolutely* fantastic. Thank you for taking this on, Frank!

l-helix:
Hi @djspiewak @fthomas how can this be imported from the gradle-scalafix plugin? Thanks!

fthomas:
@alcortaj I don't know gradle-scalafix but with sbt and sbt-scalafix, you can run the migration with `sbt ";scalafixEnable ;scalafixAll github:typelevel/cats-effect/v3_0_0?sha=v3.0.0"`. Scalafix or sbt-scalafix will pull the migration directly from GitHub for this. Maybe this works with gradle-scalafix too? If gradle-scalafix requires the migration to be in JAR, you're out of luck because the Cats Effect rules are not published. Maybe publishing them yourself locally would work in this case.

PR #1687, opened 2021-02-15
Migration guide

### LINKED ISSUE #1048: Create a migration guide
We need to document all the changes we've made in operational terms, so that people know what new type class to use, what functions moved around, what semantics are new can old (especially this!), etc etc. Basically everything you need to know to upgrade to 3.0.

kubukoz:
~It's not done yet, but I think we should push something ASAP so we can help at least library authors to some extent.~ It's done.

Preview at https://cats-effect-mig.netlify.app/cats-effect/docs/migration-guide

Closes #1048


kubukoz:
note to self:

```scala
promise.complete(r).ensure(promiseAlreadyCompleted)(identity)
```

bplommer:
> note to self:
> 
> ```scala
> promise.complete(r).ensure(promiseAlreadyCompleted)(identity)
> ```

I wonder if `Deferred#complete` should return `F[Either[PromiseAlreadyCompleted.type, Unit]]` rather than `F[Boolean]`. Then it's more obvious that the return value represents a possibility of error, and with `MonadThrow` you can just do `.rethrow` rather than `.ensure(promiseAlreadyCompleted)(identity)`

kubukoz:
I'm not really sure why the change happened in the first place - failing if already completed seemed like a reasonable thing to do, with the option to get a boolean (`tryComplete`). This might break people's code if someone relies on the behavior of failing.

The change you're suggesting also makes sense, and with cached values of the `Either` we should be able to avoid allocations too.

bplommer:
> I'm not really sure why the change happened in the first place - failing if already completed seemed like a reasonable thing to do, with the option to get a boolean (`tryComplete`). This might break people's code if someone relies on the behavior of failing.

It's because `Deferred` is parametric in its error type in CE3, and there's no way to raise errors in a parametric type. I agree it's annoying though - I wonder if we could add some kind of `Fail` typeclass that allows failing without specifying the error value.

kubukoz:
I had a busy week and had to take time off for the end, but I plan on wrapping this up asap.

kubukoz:
todo: mention Backpressure isn't in yet: https://github.com/typelevel/cats-effect/pull/1695

kubukoz:
@djspiewak I'm more or less done with everything above `### Async <!-- todo -->`. Please take a look :)

README.md
```diff
@@ -43,6 +43,15 @@ libraryDependencies += "org.typelevel" %% "cats-effect-testkit" % "3.0.0-RC3" %
 
 Cats Effect provides backward binary compatibility within the 2.x and 3.x version lines, and both forward and backward compatibility within any major/minor line. This is analogous to the versioning scheme used by Cats itself, as well as other  major projects such as ScalaJS. Thus, any project depending upon Cats Effect 2.2.1 can be used with libraries compiled against Cats Effect 2.0.0 or 2.2.3, but *not* with libraries compiled against 2.3.0 or higher.
 
+Binary compatibility for a given set of *release candidates* (e.g. **3.0.0-RC1**, **3.0.0-RC2**, etc) is provided on a best-effort basis. It is the intention that no breaking changes are introduced within a release candidate cycle, but exceptions may be made for high priority changes.
```
kubukoz: bad rebase, I'll drop this paragraph

kubukoz:
note to self: remember to mention MVar https://app.element.io/#/room/#typelevel_cats-effect:gitter.im/$3YZ_zMzFbL9bNCkbIRBFZUNxK60tT3qMhhx3ucQ4fBM

docs/migration-guide.md
```diff
@@ -0,0 +1,705 @@
...
+
+```scala
+trait Fiber[F[_], E, A] {
+  def cancel: F[Unit]
+  def join: F[Outcome[F, E, A]]
+}
+```
+
+However, there are still some differences here: first of all, `join` doesn't just return `F[A]`, but the whole `Outcome` of it.
+This is also why `Fiber` got the extra type parameter `A`.
```
catostrophe: Maybe E?

kubukoz: true lol, I made a huge mistake here. Thanks!

docs/migration-guide.md
```diff
@@ -0,0 +1,705 @@
+---
+id: migration-guide
+title: Migration guide
+---
+
+## Summary
+
+Here is a general view of the steps you should take to migrate your application to Cats Effect 3:
```
wemrysi: "a general view" -> "an overview"?
```suggestion
Here is an overview of the steps you should take to migrate your application to Cats Effect 3:
```

docs/migration-guide.md
```diff
@@ -0,0 +1,705 @@
...
+This guide is meant for existing users of Cats Effect 2 who want to upgrade their applications
+to 3.0.0.
+
+> If you haven't used Cats Effect before and want to give it a try,
+> please follow the [getting started guide](./getting-started.md) instead!
+
+### 🤔 Need help?
+
+If any point of the migration turns out to be difficult and you feel like you need help, feel free to [explain your problem on Gitter](https://gitter.im/typelevel/cats-effect) and we will do our best to assist you.
+If you spot a mistake in the guide or the library itself, you can [report an issue on GitHub](https://github.com/typelevel/cats-effect/issues/new).
```
wemrysi: Maybe also nudge towards a PR?

docs/migration-guide.md
```diff
@@ -0,0 +1,705 @@
...
+  interface: LibraryInterface,
+  listener: String => F[Unit]
+): F[Unit] =
+  Dispatcher[F].use { dispatcher =>
+    Sync[F].delay {
+      interface.addListener(str => dispatcher.unsafeRunSync(listener(str)))
+      interface.run()
+    } *> Async[F].never[Unit].guarantee(Sync[F].delay(interface.shutdown()))
+  }
+```
```
wemrysi, line 481: Another dispatcher example illustrating the case where it isn't immediately `.use`d might be helpful to illustrate the lifecycle mentioned above. My concern is that the current example easily becomes a footgun if someone were to modify it for a case where they didn't immediately block with `.never`.

```scala
trait Consumer[A] {
  def onNext(a: A): Unit
}
```

``` scala
// CE2
def consumer[F[_]: Effect, A](handler: A => F[Unit]): Consumer[A] =
  new Consumer[A] {
    def onNext(a: A) = handler(a).toIO.unsafeRunSync()
  }
```

```scala
// CE3
def consumer[F[_]: Async, A](handler: A => F[Unit]): Resource[F, Consumer[A]] =
  Dispatcher[F] map { dispatcher =>
    new Consumer[A] {
      def onNext(a: A) = dispatcher.unsafeRunSync(handler(a))
    }
  }
```

```scala
// CE3 alternative
def consumer[F[_], A](dispatcher: Dispatcher[F], handler: A => F[Unit]): Consumer[A] =
  new Consumer[A] {
    def onNext(a: A) = dispatcher.unsafeRunSync(handler(a))
  }
```

wemrysi (commented):
This looks great, the detailed function-by-function comparisons are particularly helpful, nice work!

kubukoz:
@wemrysi I read your comments the other day and really liked them, especially the Dispatcher/Consumer part. I'll try to incorporate these suggestions later this week.

docs/migration-guide.md
kubukoz: adding in f86f4ddd5457fe14ecfcbd019c3a60bfb3f91949

docs/migration-guide.md
kubukoz: actually, I'll just use that. It's a much simpler example.

kubukoz:
And added.

kubukoz:
@djspiewak if I accidentally become the #1 committer after this is merged, it's on you :D 

image

kubukoz:
Looks like some mdoc issues on Windows, pointers welcome

djspiewak (approved):
This is IMMENSE. I'll fix Windows.

djspiewak:
@wemrysi Your `Dispatcher` example is *on point* and very recognizable to anyone who has done reactive-style things without functional effects. Nice work!

catostrophe:
Would it also be useful to add a demonstration of how the dispatcher captures the current context and escapes context modifications with `.local` or `.evalOn`?

kubukoz:
> Would it also be useful to add a demonstration of how the dispatcher captures the current context and escapes context modifications with `.local` or `.evalOn`?

In `Dispatcher` docs, probably. In the migration guide, I don't think it's that important:

- `local` didn't exist
- `evalOn` inside an effect ran with the dispatcher will still win over `evalOn` on creating the dispatcher

kubukoz:
I thought the issue was about relative paths, but that was probably a red herring, since other pages have `./` and they work...

kubukoz:
nope, wasn't it.

djspiewak:
I'm pretty sure what we're seeing here is a bug in mdoc. I'm fiddling with it to see if I can find a workaround.

djspiewak:
Fixed it. The issue is with the `mdoc:fail:reset` directive. Found by binary searching the file a bit. I'm working around it by manually inlining the mdoc results and just using a standard `scala` block.

kubukoz:
wouldn't have thought of that one... did you actually go and run it on Windows??

djspiewak:
> wouldn't have thought of that one... did you actually go and run it on Windows??

Yep. VMware for the win. Basically just tried deleting subsections of the file until I found the offending line.

kubukoz:
This is the way.

PR #1689, opened 2021-02-15
parallel finalization in Concurrent[Resource]#both

patseev:
Fixed the inconsistency between `Resource#both` and `Async[Resource]#both`.
Now both methods run finalizers in parallel. 
Related discussion: https://github.com/typelevel/cats-effect/issues/1576

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -1119,6 +1119,9 @@ abstract private[effect] class ResourceAsync[F[_]]
   override def never[A]: Resource[F, A] =
     Resource.eval(F.never[A])
 
+  override def both[A, B](fa: Resource[F, A], fb: Resource[F, B]): Resource[F, (A, B)] =
+    Resource.both(fa, fb)
+
   def cont[K, R](body: Cont[Resource[F, *], K, R]): Resource[F, R] =
     Resource applyFull { poll =>
       poll {
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -297,8 +297,8 @@ sealed abstract class Resource[F[_], +A] {
    */
   def race[B](
       that: Resource[F, B]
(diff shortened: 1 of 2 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -1119,6 +1119,9 @@ abstract private[effect] class ResourceAsync[F[_]]
   override def never[A]: Resource[F, A] =
     Resource.eval(F.never[A])
 
+  override def both[A, B](fa: Resource[F, A], fb: Resource[F, B]): Resource[F, (A, B)] =
```
kubukoz: This can probably go into ResourceConcurrent?

Also, I just saw https://github.com/typelevel/cats-effect/pull/1683/ and that looks like it should use `Concurrent[Resource[F, *]]` as well - that one has fewer constraints for `F` too.

kubukoz: Another thought... how does `race` handle finalizers? The default implementations of `both` and `bothOutcome` use `racePair`, so if we could implement that for Resource, we should get "the right thing" in `both`/`bothOutcome`/`race` for free. @SystemFw @djspiewak what do you think?

patseev: You're right about ResourceConcurrent.
I'm still struggling with the new hierarchy. Thank you! Will move it to a proper place.
And yeah, same applies to the race as well.

patseev: It closes the loser and maintains the winner

djspiewak: So the problem is that `racePair` doesn't have any special understanding of the way in which finalizers need to be composed. Ultimately, the "for free" parallelism here, or lack thereof, comes down to the semantics of `ap`, which underlies `tupled`, which underlies `both`. I'm actually relatively certain that it's impossible to get the "right thing" for free here, and it just has to be a bespoke special case. That's okay though so long as we're viewing it as a performance optimization and not as a semantic guarantee.

djspiewak (approved):

Issue #1497, opened 2020-12-11
Multiplexed Dispatcher


RaasAhsan:
`Dispatcher` hasn't yet been benchmarked but I had an idea that could reduce contention on the singular queue. Basically, the idea is to create another `Dispatcher` implementation which itself dispatches to one of N `Dispatcher`s in a roundrobin fashion. I'm having a hard time reasoning about how performance would change. On one hand, there is less contention and waiting in the queue, but on the other, batch sizes become smaller. The other concern is what would the ideal N be?

djspiewak:
I've done some general measurements of this in fs2-netty and the conclusion I've come to is: it's more than fast enough the way it is. :-) It's still worth benchmarking and playing around with this idea, so I'll leave this open, but the performance seems extremely good in practice.

PR #1692, opened 2021-02-16
Striped/multiplexed Dispatcher

### LINKED ISSUE #1344: Create a benchmark for Dispatcher
`Dispatcher` is going to have some additional overhead that `UnsafeRun` didn't have, so we should probably try to quantify what that overhead is, though I'm not entirely sure what a comparison would look like.

vasilmkd:
Resolves #1344.

Answers #1497.

# Idea:
  - `Runtime#availableProcessors()` number of Dispatcher fibers
    - Incoming `F`s choose a dispatcher at random, using `ThreadLocalRandom` which is not synchronized by default like `java.util.Random` is.
    - Latch (each Dispatcher fiber has a separate latch) mechanism is slightly modified to first check the state before executing `getAndSet`. We use this trick in the work stealing runtime and it seems to help a bit with contention.

# Benchmarks
## series/3.x
```
Benchmark                       (size)   Mode  Cnt   Score   Error    Units
DispatcherBenchmark.scheduling    1000  thrpt   20  31.591 ± 2.140  ops/min
```

## this PR
```
Benchmark                       (size)   Mode  Cnt   Score   Error    Units
DispatcherBenchmark.scheduling    1000  thrpt   20  74.832 ± 1.377  ops/min
```

These results are valid as of 1 commit before the last one...

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
@@ -85,22 +85,12 @@ object Dispatcher {
 
   private[this] val Cpus: Int = Runtime.getRuntime().availableProcessors()
 
-  def apply[F[_]](implicit F: Async[F]): Resource[F, Dispatcher[F]] =
-    for {
-      dispatchers <- (0 until Cpus).toList.traverse(_ => internal[F]).map(_.toArray)
-    } yield new Dispatcher[F] {
-      def unsafeToFutureCancelable[E](fe: F[E]): (Future[E], () => Future[Unit]) = {
-        val idx = ThreadLocalRandom.current().nextInt(Cpus)
-        dispatchers(idx).unsafeToFutureCancelable(fe)
-      }
-    }
-
   /**
    * Create a [[Dispatcher]] that can be used within a resource scope.
    * Once the resource scope exits, all active effects will be canceled, and
    * attempts to submit new effects will throw an exception.
    */
-  private[this] def internal[F[_]](implicit F: Async[F]): Resource[F, Dispatcher[F]] = {
+  def apply[F[_]](implicit F: Async[F]): Resource[F, Dispatcher[F]] = {
     final case class Registration(action: F[Unit], prepareCancel: F[Unit] => Unit)
 
     final case class State(end: Long, registry: LongMap[Registration])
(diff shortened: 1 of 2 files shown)


vasilmkd:
I'll follow this up with benchmarks from `fs2-netty`.

vasilmkd:
Real world results (`fs2-netty`, not bad IMO):

Results averaged over 5 runs of `tcpkali` benchmarks.

| Implementation | series/3.x | Unit | This PR | Unit | Improvement |
| :--- | ---: | :--- | ---: | :--- | ---: |
| Fs2IO | 2512.402 | Mbps | 2770.672 | Mbps | 10% |
| Fs2Netty | 4451.502 | Mbps | 4563.774 | Mbps | 2.5% |

RaasAhsan:
I had a simpler implementation in mind that involved no changes to the existing `Dispatcher` implementation. Basically, we just allocate N regular `Dispatcher` and define a new implementation of `Dispatcher` which points to those N. Then you just dispatch to one of the N with a round-robin strategy (or anything else) via a shared `AtomicInteger` or randomly or something

I have no idea that's going to fair with the changes here (the monomorphic call is now bimorphic) but it would be cool to see :)

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -87,64 +91,62 @@ object Dispatcher {
...
-              if (!latch.compareAndSet(null, () => cb(Right(())))) {
-                // state was changed between when we last set the latch and now; complete the callback immediately
-                cb(Right(()))
-              }
-            }
-          } else {
-            registry
-              .toList
-              .traverse_ {
-                case (id, Registration(action, prepareCancel)) =>
```
RaasAhsan, line 137: 👍  Oh yeah, I think these IDs lost their value after we incorporated `Supervisor` here

vasilmkd:
What call becomes bimorphic exactly? I'm interested to hear more.

RaasAhsan:
I was referring to the implementation I suggested above -- there are two `Dispatcher` implementations involved rather than one

vasilmkd:
@RaasAhsan I'll implement it now.

djspiewak:
Random curiosity (I'm going to review this more closely today): how much is the optimal number of fibers here driven by the semantics of the underlying thread pool? In a work-stealing implementation, exactly matching the number of workers is always going to be optimal in a construct like this, but it seems like this wouldn't be the case for a fixed pool. I wish there were a nice way for us to detect this kind of stuff and self-optimize a bit, rather than assuming CE's work-stealing mode.

vasilmkd:
Relevant comments also here:
https://gitter.im/typelevel/cats-effect-dev?at=603058646825b964d61e5222

vasilmkd:
@djspiewak I'm not sure this should have been merged in this state...

djspiewak:
YOLOLOLO

PR #1695, opened 2021-02-16
Add Backpressure / Supervisor

etspaceman:
This PR backports `Supervisor` from CE3 and adds a `Backpressure` interface, which allows users to supply a rate-limiting strategy to Effects.


vasilmkd:
Thank you for the contribution.

First of all, I might have missed some discussion somewhere so these comments may be totally irrelevant.

A question to the other maintainers, with CE3 being so close to release, was there a plan for backporting any of the new cool things to CE2, in particular `Supervisor`?

`Backpressure` seems interesting to me, and maybe it would be a nice addition to `cats.effect.std` instead in CE3. I would however, like to see a more generalized version, which allows for specifying a time interval with which more bandwidth-like constructs can be encoded in addition to just rate limiting.

Thoughts?

etspaceman:
> was there a plan for backporting any of the new cool things to CE2, in particular Supervisor?

To answer this specifically for why this was proposed, our team needed the `Supervisor` trait so we created this back-port in our own code-base.

vasilmkd:
I figured I was missing some context 😄

etspaceman:
> I would however, like to see a more generalized version, which allows for specifying a time interval with which more bandwidth-like constructs can be encoded in addition to just rate limiting.

I am open to ideas here. I'm not sure what you mean by a time interval, could you provide an example?

vasilmkd:
Something similar to `fs2`'s `metered`.

djspiewak:
I'm not sure Fs2's `metered` is all that similar here. Really the problem is that there are a lot of different ways you might want to encode backpressure, and this really just represents one very simple one. For anything really meaningfully interesting, moving over to Fs2 rather than trying to do it with "pure" cats effect is almost certainly the right approach.

More generally regarding CE2… I definitely don't mind stuff getting backported as people need it, though it's unlikely I would have enough time to personally do so.

kubukoz:
Maybe I'm missing something, but is Backpressure going to be in 3.x?

djspiewak:
> but is Backpressure going to be in 3.x?

I was thinking about that. It could be! It's a very common scenario, but also a scenario which is technically better suited to fs2.

At this point, it's too late to get anything into 3.0, but we should discuss whether it should be in 3.1

kubukoz:
Mmm. I'll make a note about this in The Thing That Just Won't Come Out Until Last Minute.

kubukoz:
> Mmm. I'll make a note about this in The Thing That Just Won't Come Out Until Last Minute.

Narrator: he didn't. This isn't in in the migration guide 😅

PR #1703, opened 2021-02-18
Added mergify config

djspiewak:


vasilmkd:
Is this file auto-generated or created manually? Is there a potential drift with regards to the GH Actions matrix?

djspiewak:
It's being done manually right now. I think generating it automatically makes a lot of sense, and there's some issue on sbt-github-actions to that point, but for now I just wanted to experiment with it.

rossabaker:
There is drift, but the fallout is that conditions aren't met and auto-merging stops.  It's a quiet but harmless failure mode.

PR #1709, opened 2021-02-18
Use Unique in Supervisor instead of custom Token

### LINKED ISSUE #1684: Use Unique in Supervisor instead of custom Token
`Supervisor.apply[F]` does `val token = new Token` on its own - we can probably reuse `Unique[F].unique` instead.

DennisVDB:
Closes #1684


djspiewak:
Nice! Thank you

PR #1716, opened 2021-02-20
Directly continue async execution in Get

vasilmkd:
Opening this as a draft because I'm not 100% confident that this is legal. Here are some benchmark results to make this PR harder to turn down however.

The reasoning is the following, when `IOCont.Get` is being executed in the runloop, if the async callback has already been completed, there is no reason to reschedule the runloop again.

## series/3.x
```
AsyncBenchmark.async     100  thrpt   20  10975.597 ± 184.580  ops/s
```

## This PR
```
AsyncBenchmark.async     100  thrpt   20  14625.968 ± 588.554  ops/s
```

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -419,8 +419,11 @@ private final class IOFiber[A](
                 // `wasFinalizing` is published
                 if (finalizing == state.wasFinalizing) {
                   if (!shouldFinalize()) {
-                    /* we weren't cancelled, so resume the runloop */
-                    asyncContinue(e)
+                    /* we weren't cancelled, so schedule the runloop for execution */
+                    val ec = currentCtx
+                    resumeTag = AsyncContinueR
+                    objectState.push(e)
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -573,7 +576,12 @@ private final class IOFiber[A](
 
             if (!shouldFinalize()) {
               /* we weren't cancelled, so resume the runloop */
-              asyncContinue(result)
+              val next = result match {
+                case Left(t) => failed(t, 0)
+                case Right(a) => succeeded(a, 0)
+              }
+
+              runLoop(next, nextIteration)
```
djspiewak, line 584: Oh! This is very smart. And also it's entirely legal. The case here is that the callback was satisfied prior to the `Get` being scheduled, so you have it as a guarantee that you're on the correct thread to begin with.

vasilmkd: Exactly my reasoning.

vasilmkd:
I guess it's not a draft anymore then.

Issue #1719, opened 2021-02-21
Replace assertions with an explicit error raising in std.Random


patseev:
Today I've asked in gitter for reasoning behind use of assertions and throwing exceptions in the open instead of using ApplicativeThrow or MonadThrow.

Examples of this are constructors of `CountDownLatch`, `CyclicBarrier` and other abstractions from `std`.

As it turns out, the primary reason is that we need to be polymorphic in error type. And it's a trade-off that is justified in fs2 originated abstractions where arguments are usually static.

However, this is not the case in `std.Random`, where arguments are not static. Default implementation relies on Sync, so we can rely on MonadThrow instead of using assertions.

PR #1720, opened 2021-02-21
Use raiseError instead of assertions in Random

patseev:
Related issue: https://github.com/typelevel/cats-effect/issues/1719

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Random.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Random.scala
@@ -210,71 +210,69 @@ object Random {
 
   private abstract class ScalaRandom[F[_]: Sync](f: F[SRandom]) extends Random[F] {
 
-    def betweenLong(minInclusive: Long, maxExclusive: Long): F[Long] = {
-      require(minInclusive < maxExclusive, "Invalid bounds")
-      val difference = maxExclusive - minInclusive
-      for {
-        out <-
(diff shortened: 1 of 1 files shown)


vasilmkd (approved):

std/shared/src/main/scala/cats/effect/std/Random.scala
```diff
@@ -382,5 +379,8 @@ object Random {
         r <- f
         out <- Sync[F].delay(r.shuffle(v))
       } yield out
+
+    private def require(condition: Boolean, errorMessage: => String): F[Unit] =
+      new IllegalArgumentException(errorMessage).raiseError[F, Unit].unlessA(condition)
```
RaasAhsan: There is a bit of a subtlety with `raiseError` syntax in conjunction with `MonadError[F, Throwable]`. In particular, that exception is always going to be allocated and force the population of a stack trace regardless of whether `condition` is true or not, which can be quite expensive. I think we should conditionally allocate the exception via a regular `if` statement, or the monadic `ifM` combinator

RaasAhsan (changes requested):

std/shared/src/main/scala/cats/effect/std/Random.scala
patseev: Fixed. TY for the remark!

djspiewak:
Thank you!

Issue #1723, opened 2021-02-22
Add IO#to from CE2


kubukoz:
Since IO is in the same module as LiftIO, we should still be able to do this:

```scala
//in trait IO (cats-effect 2.x)
  /**
   * Converts the source `IO` into any `F` type that implements
   * the [[LiftIO]] type class.
   */
  final def to[F[_]](implicit F: LiftIO[F]): F[A @uncheckedVariance]
```

Issue #1722, opened 2021-02-21
Add `Async.evalOnK`


kubukoz:
This would correspond to CE2's `ContextShift.evalOnK`.

Original:

```scala
//in object ContextShift
def evalOnK[F[_]](ec: ExecutionContext)(implicit cs: ContextShift[F]): F ~> F
```

CE3 proposal:

```scala
//in trait Async[F]
def evalOnK(ec: ExecutionContext): F ~> F
```

PR #1730, opened 2021-02-22
IO#to, IO#option, IO.none, IO.some, Async.evalOnK

patseev:
Bunch of small utility methods ported from CE2 + Async.evalOnK

https://github.com/typelevel/cats-effect/issues/1722
https://github.com/typelevel/cats-effect/issues/1723
https://github.com/typelevel/cats-effect/issues/1724
https://github.com/typelevel/cats-effect/issues/1728


djspiewak:
This is great, thank you!

Issue #1580, opened 2021-01-04
Investigate replacing ScheduledExecutorService with a HashedWheelTimer


mpilquist:
Netty, Akka, Blaze, and other libraries all use a variant of a hash wheel timer for scheduling lots of non-blocking tasks:

- [Netty's HashWheelTimer](https://github.com/liaokailin/netty/blob/master/common/src/main/java/io/netty/util/HashedWheelTimer.java)
- [Akka's LightArrayRevolverScheduler](https://github.com/akka/akka/blob/a4acf23d0541471ecf81b9ba5dd97c2e97670814/akka-actor/src/main/scala/akka/actor/LightArrayRevolverScheduler.scala)
- [Blaze's TickWheelExecutor](https://github.com/http4s/blaze/blob/main/core/src/main/scala/org/http4s/blaze/util/TickWheelExecutor.scala)

https://stackoverflow.com/questions/17276393/hashedwheeltimer-vs-scheduledthreadpoolexecutor-for-higher-performance

We should consider adopting something similar for CE3, measured against ScheduledExecutorService to see which is better. At the very least, we should ensure there are no API changes necessary if we later want to swap in a custom scheduler.

djspiewak:
Based on what I can find, this is definitely something that is worth doing. If we can get it in for 3.0, it would be ideal, but it's technically something that can land afterward if necessary.

TimWSpence:
This looks fun! Happy to have a stab at it at least

He-Pin:
I think this is really needed. I was fired an issue https://github.com/reactor/reactor-core/issues/2965 and switch to Akka just because of this.

armanbilge:
@He-Pin see the following discussions regarding plans for 3.5.0:
- https://github.com/typelevel/cats-effect/discussions/3101
- https://github.com/typelevel/cats-effect/discussions/3070

armanbilge:
Not a hashwheel timer, but still a replacement for `ScheduledExecutorService`.
- https://github.com/typelevel/cats-effect/pull/3219

He-Pin:
@armanbilge @djspiewak  I think this need renmaed to add a HashedWheelTimer to cats-effect

because the current integrated timer is O(log N), not O(1)

Issue #1433, opened 2020-11-19
Delays when using Stream.awakeEvery


adamw:
This is a follow-up to https://github.com/http4s/http4s/issues/3879

I've got the following simple program:

```scala
import cats.effect.{ExitCode, IO, IOApp}
import fs2._

import scala.concurrent.duration.DurationInt

object Test extends IOApp {
  override def run(args: List[String]): IO[ExitCode] = {
    Stream.awakeEvery[IO](1.seconds).map(d => println(s"Ping! ${d.toMillis}ms")).compile.drain.map(_ => ExitCode.Success)
  }
}
```

*sometimes* this produces results as follows:

```
Ping! 4987ms
Ping! 5992ms
Ping! 13921ms
```

This happens locally, and on travis & jenkins (though on CI I only observed it indirectly through http4s & websockets). The jenkins jobs run pods on top of a GCP-managed k8s. Locally a system restart (macos 11) helps, at least for some time; the CPU is quite idle. I've tried both java8 and java11.

Any hints where to look / how to debug? :)

djspiewak:
Follow up: I let this run on my laptop for about 5 hours, during which time it was under intermittent extreme load (enough to trigger thermal throttling), as well as longer downtime periods. I even put it to sleep at one point. The maximum tick differential was 1018 milliseconds. (macOS 11.0.1, GraalVM CE 20.3.0 targeting JDK 8, Scala 2.13.3, 16 physical threads over 8 cores)

adamw:
Thanks, I even didn't know that you can transfer issues ;)

I've got similar hardware: macos 11.0.1, 6-core i9, using either java `8.0.232-zulu`, `11.0.5-zulu` or `20.2.0.r11-grl` (from sdkman).

But I suppose this is neither a cats-effect nor even a scala issue - if it's an issue at all. Running this code using pure java:

```java
class Test {
  public static void main(String[] args) throws Exception {
    long last = System.currentTimeMillis();

    while(true) {
      Thread.sleep(1000);
      long now = System.currentTimeMillis();
      System.out.println(now - last);
      last = now;
    }
  }
}
```

while the build is running in the background, gives the same results (that is - large delays). So I suppose this is the way OS-level thread scheduling works. I'm still a bit surprised that everything else works normally (well, maybe a bit slower, but rather normal), so I'd expect the processes to be scheduled more fairly, but there's always something new to learn :)

[...]

PR #1731, opened 2021-02-22
Hashed wheel timer

TimWSpence:
WIP but opening a PR to seek some feedback. From my perspective there are at least 2 open questions:
1. Is it safe to depend on `System.currentTimeMillis()` given that this isn't guaranteed to be monotonic across threads. This is how Blaze does it but I can't convince myself that it's safe - if something is scheduled with a very short delay it seems possible to me that it could miss its chance to be inserted into the right bucket before that bucket is scheduled. More investigation needed
2. Is the benchmark sane? How can we make it better? And what should we set the resolution to to make it a representative test?


vasilmkd:
I'll take a look later tonight.

djspiewak:
It definitely sounds very very unlikely that `currentTimeMillis()` is safe. In fact, I'd be willing to bet this is a meaningful bug in Blaze

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
```diff
@@ -0,0 +1,246 @@
...
+
+  private val thread = new Thread("io-scheduler") {
+    override def run(): Unit = loop()
+  }
+
+  thread.setDaemon(true)
+  thread.setPriority(Thread.MAX_PRIORITY)
+  thread.start()
+
+  private def loop(): Unit = {
```
djspiewak: IMO this should be inlined into the thread `run` definition to avoid the shadowing of the `loop` name

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
```diff
@@ -0,0 +1,246 @@
...
+        //isn't guaranteed
+        if (curr < target) {
+          //TODO do we need to handle thread interrupted ex?
+          // println("sleeping")
+          Thread.sleep(target - curr)
+        }
+        loop(ticks)
+      } else {
+        //null out to avoid leaks
+        (0.until(wheelSize)).foreach { n => wheel(n) = null }
```
djspiewak: Worth doing this with a `while` loop since it's much faster.

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
```diff
@@ -0,0 +1,246 @@
...
+      }
+    }
+
+    loop((nowMillis() * invResolutionMillis).toLong)
+  }
+
+  @inline private def tsToBucketIdx(ts: Long): Int =
```
djspiewak, line 153: I would remove the `@inline` annotations since they don't really do anything in this case.

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
```diff
@@ -0,0 +1,246 @@
...
+    ((ts * invResolutionMillis).toLong % wheelSize).toInt
```
djspiewak, line 154: Maybe we should restrict `wheelSize` to powers of 2?

tests/shared/src/test/scala/cats/effect/Runners.scala
```diff
@@ -34,7 +34,7 @@ import scala.reflect.ClassTag
 
 trait Runners extends SpecificationLike with TestInstances with RunnersPlatform { outer =>
 
-  def executionTimeout = 10.seconds
+  def executionTimeout = 1000.seconds
```
djspiewak: That's quite a while. :-D

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
```diff
@@ -0,0 +1,246 @@
...
+import scala.util.control.NonFatal
+
+/*
+ * Hashed wheel timer based on George Varghese and Tony Lauck's paper <a
+ * href="http://cseweb.ucsd.edu/users/varghese/PAPERS/twheel.ps.Z"> 'Hashed and
+ * Hierarchical Timing Wheels: data structures to efficiently implement a timer
+ * facility' and the http4s implementation
+ * https://github.com/http4s/blaze/blob/b9305ccb2e04ee62bde6619d7e7e2d17ea645e37/core/src/main/scala/org/http4s/blaze/util/TickWheelExecutor.scala
+ */
+class HashedWheelTimerScheduler(wheelSize: Int, resolution: FiniteDuration) extends Scheduler {
```
djspiewak: Should be `private[unsafe]` and also `final`, btw

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
```diff
@@ -0,0 +1,246 @@
...
+        val curr = nowMillis()
+        val target = (ticks + 1) * resolutionMillis
+        //We don't blindly sleep for resolutionMillis - (curr - start)
+        //as we want to be self-normalizing to awake at the
+        //start of each time interval as Thread.sleep accuracy
+        //isn't guaranteed
+        if (curr < target) {
+          //TODO do we need to handle thread interrupted ex?
+          // println("sleeping")
+          Thread.sleep(target - curr)
```
djspiewak: It seems like this busy-waits even when nothing is enqueued. Should we park and wait for new registrations if everything is empty?

djspiewak (commented):
Looks really promising!

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
TimWSpence: So we can do bitshifting instead of modulo? If so then yes, I was thinking that too 👍

tests/shared/src/test/scala/cats/effect/Runners.scala
TimWSpence: 😂 Sorry, haven't cleaned up the tests yet. Running the timer tests through this takes quite a while due to 100 iterations of sleeping up to 1s. I just haven't created a separate execution path for them yet

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
TimWSpence: Yeah that's a good call! Is there a better way than just having some `signal` object that we `signal.wait()` and `signal.notify()` on? I _think_ that would require having a volatile (or stronger) counter so that requesting threads know if they need to `notify` (or they could just `notify` every time but I assume that would be much worse)

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
TimWSpence: Not really doing anything?

TimWSpence:
This is still WIP but I'm moving it out of draft to start running CI against it.

Also I've started https://gitter.im/http4s/http4s-dev?at=603cbc7a823b6654d276c641 to discuss the dependence on `currentTimeMillis` in Blaze

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
djspiewak: Not really doing anything anywhere. :-D `@inline` is almost useless.

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
djspiewak: I think @vasilmkd has relevant experience here. We should probably do the same thing the `WorkStealingThreadPool` is doing.

vasilmkd:
I haven't forgotten about this PR, I just wanted to read up on hashed wheel timer first, but I didn't find the proper chance yet.

TimWSpence:
> I haven't forgotten about this PR, I just wanted to read up on hashed wheel timer first, but I didn't find the proper chance yet.

No worries!! I actually haven't had any time to spend on this for a week or so but was just trying to pick up momentum again to try to get it in before final release

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
vasilmkd: We don't need to check for interruption in the scheduler at all because it's completely walled off from the world. We control what we submit to it (`IO.sleep`).

As for the signal, I will think about it.

vasilmkd: Ok, now I checked because I can never remember exactly what is available to outside users. We can probably combine the signal and the interruption check.

vasilmkd:
I did some research on `HashedWheelTimers` and I couldn't find any design which doesn't include busy waiting or `Thread.sleep`. I am honestly not comfortable including this in the runtime, I think we would be disrupting the compute thread pool by constantly rescheduling a thread for execution, to get a bit better resolution when using `IO.sleep`, which in my honest opinion is a wrong tradeoff to make. If we do design a signaling mechanism, I think we wouldn't fare much better than the JVMs `ScheduledExecutorService` (whose threads do sleep until notified). This is again just my own opinion, but unlike creating a custom compute pool where fibers are first class citizens and we specifically optimize for our use cases, here, the HashedWheelTimer does not bring anything to the runtime that would justify its non-trivial implementation and maintenance costs. As far as I know, Cats Effect makes no promises on the precision of `IO.sleep`, which is ultimately a futile exercise on non real-time operating systems (we have already seen reports of this such as #1433), and I believe this wouldn't be any better with the `HashedWheelTimer` (there is no guarantee that the thread will be scheduled in time to end the `sleep` operation and reschedule the fiber in the compute pool).

I would ideally like to see a mechanism which is fiber-first. To me, `HashedWheelTimer` does not seem like it. I'm sorry.

djspiewak:
I believe the advantage to HWT is less about timing precision and more about efficiency in scheduling and canceling. Neither of these are particularly efficient with the stock scheduler, which is why Netty, Akka, Blaze, etc etc all implement the same construct.

I do think that this is likely to be disruptive to the compute pool though. We should try to design some benchmark scenarios where we can see this one way or another.

djspiewak:
Also I'm pretty sure it's possible to implement HWT like structures without busy waiting. You would get the best of both worlds in theory: a ton of efficiency under high timer pressure (a scenario which will always compete for threading resources with the compute pool), and also a ton of efficiency under low timer contention (a scenario in which we want to have zero contention with compute).

It would necessitate parking until enqueue, then sleep for the relevant number of seconds. If a new enqueue comes in sooner, wake the thread early and force it to resleep for the new time. Always quantize to the wheel. When in doubt sleep until next tick after execution until the wheel is known to be empty, at which point we can just park again.

Also note that we could in theory communicate this pressure state to the work stealing queue, which could temporarily stop using one of its workers until the pressure abates. This gives us some dynamic awareness wherein we can self tune to timer heavy workloads

TimWSpence:
I think it's worth noting that HashedWHeelTimers are explicitly _worse_ for sleep accuracy but have lower CPU usage and do _fewer_ sleeps when there are actually tasks scheduled frequently. The Java `ScheduledExecutorService` essentially maintains a priority queue of scheduled tasks and sleeps till the next one is scheduled.
If we consider eg `List.range(0, 100).parTraverse(i => IO.sleep(i.millis) >> IO.pure(i))` the behaviour of the `ScheduledExecutorService` would be to sleep and awake every millisecond to schedule the next task (and each insert into the queue will be `O(log n)`). By contrast, the `HashedWheelTimer` would sleep once for 100 millis and then schedule all the tasks (and the insertion will be `O(1)` in each case)

So I guess (as Daniel essentially suggested above) we probably want some sort of hybrid approach where we use bucketing but park if there are no tasks scheduled and potentially sleep till we reach the next bucket that has a task in it (although that might actually have more overhead, not sure yet).

As it stands, the hashed approach adds an essentially constant overhead whereas the `ScheduledExeutorService` has virtually no overhead when there are no tasks scheduled but becomes pathologically bad in examples with lots of scheduling like the one above.

vasilmkd:
Those are some compelling arguments.

As an immediate pointer, you can take a look at `java.util.concurrent.locks.LockSupport`. It offers support for parking a thread indefinitely `park()` as well as for a timed period `parkNanos()`. I haven't used `parkNanos` yet, but it should do what it says on the box, and it should be better than `Thread.sleep`.

benchmarks/src/main/scala/cats/effect/benchmarks/HashedWheelTimerSchedulerBenchmark.scala
```diff
@@ -0,0 +1,88 @@
...
+ *     jmh:run -i 10 -wi 10 -f 2 -t 1 cats.effect.benchmarks.HashedWheelTimerSchedulerBenchmark
+ *
+ * Which means "10 iterations", "10 warm-up iterations", "2 forks", "1 thread".
+ * Please note that benchmarks should be usually executed at least in
+ * 10 iterations (as a rule of thumb), but more is better.
+ */
+@State(Scope.Thread)
+@BenchmarkMode(Array(Mode.Throughput))
+@OutputTimeUnit(TimeUnit.SECONDS)
+class HashedWheelTimerSchedulerBenchmark {
```
vasilmkd, line 47: For the benchmark, instead of `Blackhole.consumeCPU` I think it would be better if all fibers decremented some `AtomicInteger` (the sleeping part stays the same) and maybe we use a random sleeping value (bounded between two reasonable values, we can find those experimentally). That way, I believe we devise a situation where there is enough work for the scheduler thread, as well as contention between the compute pool threads. On the other hand, `Blackhole.consumeCPU` is I believe just a long `IO.delay`, which doesn't contend with other worker threads, just adds further delay in the execution of the benchmark.

vasilmkd: This is more what I have in mind, an `IO` version of this [benchmark](https://github.com/ifesdjeen/hashed-wheel-timer/blob/master/bench/src/com/ifesdjeen/timer/AbstractBenchmark.java), and a `CountDownLatch` instead of busy waiting.

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
vasilmkd: For inline to work, a compiler flag needs to be passed to `scalac` (I think it's `-optimise`), by default, `@inline` does nothing as the optimizer is [disabled](https://www.scala-lang.org/api/current/scala/inline.html).

The JVM is better left to its own devises and the JIT usually delivers better performance when compiling to native code than what `scalac` can optimize when compiling to bytecode.

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
vasilmkd: If we plan to keep `wheelSize` as not configurable by the user then yes, let's go for a power of 2.

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  thread.setDaemon(true)
+  thread.setPriority(Thread.MAX_PRIORITY)
+  thread.start()
+
+  @inline private def tsToBucketIdx(ts: Long): Int =
+    ((ts * invResolutionMillis).toLong % wheelSize).toInt
+
+  @inline private def ticksToBucketIdx(ticks: Long): Int =
```
vasilmkd, line 156: If you want to achieve what `@inline` would do, it's better to use `private[this]` (only on Scala 2, Scala 3 will do it automatically in the future). This way, the JVM will know that this method is only used in this class (no subclasses can call it) and it will most likely inline it in the call sites. We rely on this behavior extensively in `IOFiber.scala` and `LocalQueue.scala` (I would say these are good examples in our codebase 😄).

vasilmkd: When writing performance sensitive code it's actually a good idea to annotate all class members as `private[this]`. This helps the JVM inline things as much as possible. So, all `private` `val`s, `var`s and `def`s should become `private[this]`.

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
```diff
@@ -0,0 +1,250 @@
...
+
+  private val noopCancel: Runnable = () => ()
+
+  private case class TaskState(
+      task: Runnable,
+      scheduled: Long,
+      //It can happen that a Register and a Cancel are enqueued in the same
+      var canceled: Boolean = false,
+      var next: TaskState = null,
+      var previous: TaskState = null) {
```
vasilmkd, line 186: Is there a reason why this should be a doubly linked list? I tried searching through the code and I don't immediately see a use case for the backwards pointer.

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
```diff
@@ -0,0 +1,250 @@
...
+          val next = state.next
+          if (state.scheduled <= ts) {
+            state.unlink()
+            try {
+              // println("running")
+              state.task.run()
+            } catch {
+              case NonFatal(e) => println(s"Caught error $e in io timer")
+            }
+          } else {}
```
vasilmkd, line 229: `else` branch can be dropped here.

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
```diff
@@ -0,0 +1,250 @@
...
+        if (!pendingOps.compareAndSet(op, Cancel(task, op))) go()
+      }
+      go()
+    }
+
+  //TODO is this safe if resolution is < 1ms
+  //Probably ok as Thread.sleep only allows millis
+  //(or millis and nanos precision)
+  private val resolutionMillis: Long = resolution.toMillis
+  private val invResolutionMillis: Double = 1.0 / resolutionMillis
```
vasilmkd, line 88: If you want to experiment with `java.util.concurrent.locks.LockSupport.parkNanos(nanos)` ([docs](https://docs.oracle.com/en/java/javase/15/docs/api/java.base/java/util/concurrent/locks/LockSupport.html#parkNanos(long))), we can probably get more granular resolution that milliseconds.

I think it needs to be benchmarked in the end and see if such resolution is actually feasible on the hardware that we're running on.

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
```diff
@@ -0,0 +1,250 @@
...
+
+  private val wheel: Array[Bucket] = (0.until(wheelSize)).map(_ => new Bucket()).toArray
+
+  private val pendingOps: AtomicReference[Op] = new AtomicReference[Op](Noop)
+
+  //TODO would it be better if shutdown was another op to be inserted into the queue?
+  @volatile private var canceled = false
```
vasilmkd, line 95: Can we please rename this field, it's a bit confusing because tasks themselves are also canceled, while the scheduler is "shutdown".

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
```diff
@@ -0,0 +1,250 @@
...
+
+          go(0)
+
+          val curr = nowMillis()
+          val target = (ticks + 1) * resolutionMillis
+          //We don't blindly sleep for resolutionMillis - (curr - start)
+          //as we want to be self-normalizing to awake at the
+          //start of each time interval as Thread.sleep accuracy
+          //isn't guaranteed
+          if (curr < target) {
```
vasilmkd, line 128: This is quite clever.

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
```diff
@@ -0,0 +1,250 @@
...
+          go(next)
+        }
+      }
+
+      go(head)
+    }
+
+  }
+
+  private sealed trait Op
```
vasilmkd, line 239: Instead of holding a linked list of operations, which are currently used for registration and cancelation of a task, how about `TaskState extends AtomicBoolean(false)` (false means the task is not canceled). When a task is canceled (i.e. `cancelToken` simply calls `TaskState#set(true)` on the `TaskState` argument). And finally in `schedule`, before calling `run()`, you can check if the task has been canceled by doing `TaskState#get`.

This way we also get rid of the `compareAndSet` loop in `cancelToken` and `pendingOps` can be deleted, which reduces contention between the worker threads who will be competing to schedule and cancel tasks on the `HashedWheelTimerScheduler`.

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
```diff
@@ -0,0 +1,250 @@
...
+      scheduled: Long,
+      //It can happen that a Register and a Cancel are enqueued in the same
+      var canceled: Boolean = false,
+      var next: TaskState = null,
+      var previous: TaskState = null) {
+
+    def unlink(): Unit = {
+      if (previous != null) {
+        previous.next = next
+      } else {}
```
vasilmkd, line 191: Another empty `else` branch.

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
```diff
@@ -0,0 +1,250 @@
...
+      case Cancel(state, next) => {
+        state.unlink()
+        state.canceled = true
+        executeOps(next)
+      }
+    }
+
+  private val noopCancel: Runnable = () => ()
+
+  private case class TaskState(
```
vasilmkd, line 180: `TaskState` and `Bucket` can also be `private[this]`. If we're totally paranoid about performance, we can inline `TaskState` into bucket (basically not have a separate class which results in `invokevirtual` calls but make `Bucket`, or even `HashedWheelTimerScheduler` itself hold all of the logic for scheduling tasks in buckets).

djspiewak:
Is there a way we can park for some number of nanoseconds, but then some other thread can efficiently awake us early? Interrupt obviously works but I haven't benched it

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
```diff
@@ -0,0 +1,250 @@
...
+import scala.util.control.NonFatal
+
+/*
+ * Hashed wheel timer based on George Varghese and Tony Lauck's paper <a
+ * href="http://cseweb.ucsd.edu/users/varghese/PAPERS/twheel.ps.Z"> 'Hashed and
+ * Hierarchical Timing Wheels: data structures to efficiently implement a timer
+ * facility' and the http4s implementation
+ * https://github.com/http4s/blaze/blob/b9305ccb2e04ee62bde6619d7e7e2d17ea645e37/core/src/main/scala/org/http4s/blaze/util/TickWheelExecutor.scala
+ */
+private[unsafe] final class HashedWheelTimerScheduler(
```
vasilmkd, line 31: Crazy idea, because we know that our scheduler is single threaded, we can do `HashedWheelTimerScheduler extends Thread` and inline all code in `HashedWheelTimerScheduler`, resulting in inlineable `invokespecial` calls all over the place. 😄

vasilmkd:
> Is there a way we can park for some number of nanoseconds, but then some other thread can efficiently awake us early? Interrupt obviously works but I haven't benched it

`java.util.concurrent.locks.LockSupport.unpark(thread: Thread)` still unparks it (as efficiently as we do in the `WorkStealingThreadPool`). Of course, I assume that we're going to use it.

This is why I think we can encode `IO.sleep` in `WorkStealingThreadPool` itself, precisely using this `park/parkNanos/unpark` mechanism. I just still haven't figured out a way how this can be done if a user chooses their own compute pool and not go with the default. I'm sure it can be done, but now is not the time to pursue it.

vasilmkd:
@TimWSpence @djspiewak I apologize if my comment was yesterday was received as dismissive or in bad taste. That wasn't my intention. I probably wasn't in the best mood yesterday and I have some exams this week. I shouldn't have let my stress show itself. I apologize again.

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
vasilmkd: Ideally, the `schedule` method would only find the corresponding `bucket` in which the task should be enqueued. This bucket will need atomic operations for synchronizing on it, but the performance win comes from the fact that given a fairly spread out distribution of `sleep` operations, there is almost zero contention on each bucket. As it currently stands, `pendingOps` is a global source of contention on which _every_ schedule ends up.

TimWSpence:
> @TimWSpence @djspiewak I apologize if my comment was yesterday was received as dismissive or in bad taste. That wasn't my intention. I probably wasn't in the best mood yesterday and I have some exams this week. I shouldn't have let my stress show itself. I apologize again.

Hey no worries! You know a _lot_ more about JVM optimization than I do so I appreciate the input! All the best for the exams :)

And thanks for all your comments. I'll look through them in detail later 👍

vasilmkd:
I'm here to help. The comments are mainly about what we have learned from the development of our runtime. We should use the knowledge as much as possible.

TimWSpence:
Btw I'm currently de-prioritizing this in favour of stuff that _has_ to be done for release as this should not affect binary compatiblity if we release it later. Let me know if you think this is a bad call 😂

vasilmkd:
It's ok for this to land after 3.0.0 final. No worries.

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
djspiewak: That's an amazing idea

djspiewak:
Agreed this should move out past 3.0 final. I'm excited about it though! Also I spent a ton of time thinking about how we can implement the "park until work arrives, then do the hashwheel thing" bit that I referenced earlier. I'll try to write up soon.

djspiewak:
Linking to #1580

TimWSpence:
> Agreed this should move out past 3.0 final. I'm excited about it though! Also I spent a ton of time thinking about how we can implement the "park until work arrives, then do the hashwheel thing" bit that I referenced earlier. I'll try to write up soon.

Exciting!!! I've been intending to have a good think about it and have thus far failed to have time :sob:

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
RafalSumislawski: IMO the `wheelSize` and the `resolution` should be configurable.

If you need to schedule lots of tasks, and the delays are relatively long. You need a small `resolution` to evenly distribute the load, and a large `wheelSize` so that for mosts tasks `dealy < resolution * wheelSize`.  For example I've got an application which keeps around 2 million tasks with 5 minute delay scheduled on Akka's scheduler. That use case needs a `wheelSize` of 32768.

I've got nothing against restricting `wheelSize` to powers of two.

core/jvm/src/main/scala/cats/effect/unsafe/HashedWheelTimerScheduler.scala
```diff
@@ -0,0 +1,250 @@
...
+
+  private sealed trait Op
+  private case class Register(state: TaskState, next: Op) extends Op
+  private case class Cancel(state: TaskState, next: Op) extends Op
+  private case object Noop extends Op
+
+}
+
+object HashedWheelTimerScheduler {
+  val defaultResolution: FiniteDuration = 200.millis
```
RafalSumislawski, line 247: I think that's a too high number for a default of a general purpose timer.

Users would be surprised to learn that` IO.sleep(10.millis)` sleeps on average `110ms`.

For reference:
- Blaze can get away with using 100ms resolution, because it has a specific purpose: timeouts of network operations. And 100ms seems like a reasonable resolution in this context. Also, http4s blaze issues a warning if the resolution of the timer has large impact on the accuracy of the configured timeouts (https://github.com/http4s/http4s/blob/bd9a64d95a9d280f46fa0ce3ebd80f741ef96bb0/blaze-client/src/main/scala/org/http4s/client/blaze/BlazeClientBuilder.scala#L248).   
- Akka's scheduler is more general-purpose and it uses just 10ms resolution.

TimWSpence: Thanks, yeah these were pretty arbitrary values. Definitely it should be configurable 👍

PR #1732, opened 2021-02-22
CE2: add Resource.eval, deprecate liftF

bplommer:
Backporting this change to CE2 should help smoothen the transition, especially when maintaining multiple branches based on different CE versions.


fthomas:
I support this as author of the Scalafix migration to 3.0.0 (#1686). The migration contains a bunch of [simple replacements](https://github.com/typelevel/cats-effect/pull/1686/files#diff-573fd879d1d2011a3386f8c542452c83db213d0d1c4003e324c52716c6c7002fR24-R37). If some of them are backported to CE2, we could split the migration in two, one for 2.4.0 and the other for 3.0.0. My gut feeling is that the migration for 2.4.0 would generate changes that could be merged as-is (because they are only renames) while the 3.0.0 migration would in most cases require manual fixes and tweaks.

kubukoz:
I second @fthomas's idea, but we'd need to make sure migrations for deprecated symbols are in place (and defined in a way Scala Steward can find them) before we release such a 2.4.0 :)

djspiewak:
Let's do it! I agree with all of the reasoning stated here.

PR #1733, opened 2021-02-22
Fiber[Resource[F, *], E, A] join scope contained by outer

### LINKED ISSUE #1619: Remove join scope weirdness in Async[Resource]
The current implementation has a rather bizarre semantic, wherein if you `join` on a `Fiber`, the finalizer scope for the fiber is forwarded into the `join` scope, but *via* the `Outcome`. If you don't `join`, then the relevant finalizer scope is the one which arose originally from `start`.

In retrospect, I think this is a little counter-intuitive. The only situation under which `join` could terminate its scope before the outer scope is `eval(joinEmbed.use)` or something similar, meaning we don't actually get much in the way of expressiveness from this rather odd semantic, and the downsides are that `join` after `join` doesn't work reasonably, and the state machine is unnecessarily complex.

Instead of all of this, we should simply treat the `Fiber` itself as a resource (opened by `start`) and make *that* scope responsible for ensuring the inner finalizers are cleaned up.

wemrysi:
Refactors `Concurrent[Resource[F, *]].start` so the outer resource scope contains the fiber `join` scope, rather than being extended into it.

Closes #1619

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -970,20 +970,17 @@ abstract private[effect] class ResourceConcurrent[F[_]]
 
   /*
    * 1. If the scope containing the *fiber* terminates:
-   *   a) if the fiber is incomplete, run its inner finalizers when it complets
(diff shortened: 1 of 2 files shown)


djspiewak:
So much simpler! Thank you!

PR #1609, opened 2021-01-19
Fiber runtime 2.0

vasilmkd:
Don't mind me, I'm just dumping some code I've been working on for the last 3 months.

In all seriousness, thank you for being a great community and for letting me do crazy experiments like this. I really feel welcome in this project, it has absolutely changed my life for the better, and it's a pleasure to be able to say that I have worked with some of you here and call you all my peers.

The code is functional (functional as in working, otherwise it looks like C with Scala syntax) enough, feel free to comb through it, but beware, there are no comments. I'll make sure to change that over the coming period. Feel free to experiment with it if you want.

I'm marking it as a draft because there are still changes that will be coming, and I'm also known for squashing commits and force pushing all the time, so maybe it's not the best idea to rebase on this branch for now.

I'm negotiating to get a powerful VM from my previous university to benchmark this on in the coming days, so stay tuned for that.

For now, there are indications that we can expect >5% improvement vs the current runtime under full load (due to reduced thread contention), better single thread fiber performance (I'm yet to test this properly, but I've seen significantly better scores on some single threaded benchmarks).

There's initial support for `scala.concurrent.blocking`. I'll make some improvements as I clean up the code, but it's already working. It would be really helpful if we can test situations where the current runtime deadlocks.

Finally, I'd like to extend a special thank you to @djspiewak for surviving the constant bombardment of questions, random links and reading through 3 A.M. ideas, and for helping me with the testing.

Thank you all again, for everything.

PR #1734, opened 2021-02-24
The work stealing pool becomes a BlockContext

vasilmkd:
By popular demand, here at last.

This is probably my least favorite code I've written for Cats Effect.


core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -271,39 +268,22 @@ private[effect] final class WorkStealingThreadPool(
    * a worker thread. Otherwise falls back to scheduling on the external queue.
    */
   private[effect] def executeFiber(fiber: IOFiber[_]): Unit = {
-    if (Thread.currentThread().isInstanceOf[WorkerThread]) {
-      rescheduleFiberAndNotify(fiber)
+    val thread = Thread.currentThread()
+    if (thread.isInstanceOf[WorkerThread]) {
+      thread.asInstanceOf[WorkerThread].schedule(fiber)
+    } else if (thread.isInstanceOf[HelperThread]) {
+      thread.asInstanceOf[HelperThread].schedule(fiber)
```
djspiewak, line 275: Can we just make `WorkerThread` and `HelperThread` both extend some common superclass to remove one of these branches? Though I suppose that the implementations are distinct, so it might actually be better to have the branching exactly as you wrote it.

djspiewak (commented):
Very nice! Impressive work. Just leaving a quick comment but otherwise LGTM

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
vasilmkd: Back in the day in my original implementation of the WorkStealingPool, I had exactly that. I created our own type `IOExecutor` which both our `WorkStealingThreadPool` and any `ExecutionContext` could implement (precisely to avoid this branching). That _tanked_ performance, so I've been shying away from inheritance solutions since.

Furthermore, in #1609 I had an even less optimized version of this code, and we were still a lot faster than what's currently in `series/3.x`.

vasilmkd: Btw this exact branch is only a factor when we call `execute` on our `WorkStealingPool`, which is only at async boundaries originating from the outside world (i.e. already the slowest path).

vasilmkd: Also this https://shipilev.net/blog/2015/black-magic-method-dispatch/#_bimorphic_cases.

PR #1738, opened 2021-02-26
Scala and GraalVM version updates

vasilmkd:
Warming up the planet using the latest Scala and GraalVM.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,12 +22,12 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        scala: [3.0.0-M3, 3.0.0-RC1, 2.12.13, 2.13.4]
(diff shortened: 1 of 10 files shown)


vasilmkd:
@djspiewak, @RaasAhsan can you please check this PR very carefully. I'm especially curious about the unused functor instances for the `std` functor derivations, it seems the underlying `F[_]` does not explicitly need to be a `Functor` itself, but maybe this was an intentional constraint? If that was indeed the case, I apologize.

djspiewak (approved):
Incidentally this is breaking binary compatibility. Still worth it I think.

djspiewak:
I'm inclined to trust the compiler on this one. It does indeed seem that the constraints were unnecessary.

vasilmkd:
Oh yeah totally binary compatibility breaking, I forgot to mention...

Issue #1018, opened 2020-07-29
IOApp no longer shuts down its thread pools


djspiewak:
It uses global, which doesn't have a working `shutdown()`, meaning that `IOApp` actually can't shut its pools down. We should probably poke a hole in `IORuntime` for "unsafe shutdown" or something so that `IOApp` can actually do this, since it's probably the *only* place where we can safely shutdown global.

mpilquist:
Why does it need to be shutdown? They are all daemon threads.

djspiewak:
Basically just for build tools that don't play the same game that sbt does *now*. Sbt today uses `SecurityManager` hooks and side-effects to track every `Thread` created by processes, specifically so it can clean them up later. It didn't back in 0.13 though, and I don't know what other build tools do this. Any tool which "forks" by classloader trickery within a resident vm will need us to clean up the threads for them.

RaasAhsan:
Might be related, but Ctrl+C in SBT (with or without forks) doesn't shut down any of my apps like I remember it doing in CE2

djspiewak:
> Might be related, but Ctrl+C in SBT (with or without forks) doesn't shut down any of my apps like I remember it doing in CE2

Really? We actually test this in CI so I'm very surprised. :-)

vasilmkd:
So the status on this is the following (asuming fork := false):

1. The Ctrl+C signal is properly detected by IOApps, and they stop in accordance.
2. This does not shutdown the global IORuntime because its shutdown method is a noop.
3. Running an IOApp again starts a new global IORuntime, it doesn't reuse the old one, so threads keep being spawned for each IOApp invocation in sbt.

mpilquist:
Oh, that's no good... making shutdown a no-op was originally done to support calling `main(..)` from a REPL session. Not really a strong reason to keep that behavior as the default.

vasilmkd:
I'm never sure if this behavior is depedent on specific versions of sbt, which makes me vary of attempting anything...

djspiewak:
I *think* this is probably specific to sbt 1.4.x where x < 7. Cancelation in general didn't work on those versions. It's worth checking it on 1.4.7.

vasilmkd:
I checked this on 1.4.7

djspiewak:
Is this still broken?

djspiewak:
I'm going to say "no", it's not broken, though I've seen some odd stuff with timers (see #2251).

PR #1740, opened 2021-02-27
Consolidated IORuntime configuration and enabled access in IOApp itself

### LINKED ISSUE #1238: Find a nice way of configuring the compute pool in IOApp
Some applications will perform better with slightly different thread counts within the compute pool. Notably, @viktorklang has discussed how, in many of his tests, JVM applications behave better when the compute pool runs closer to 70% of available cores rather than the 100% we default to. Whether or not this is the case for any given application depends on a lot of factors, including how much data is being pushed through the caches, how good the cache utilization is, how much allocation is being done, and the exact implementation of hyperthreading available (or unavailable) on the host system. All in all, we can't really detect these factors, so it's better for us to, in the end, make it configurable.

This would be similar to the ce2 `IOApp.WithContext`, but maybe a little nicer. Note that, in order to do this properly, we're going to need to adjust `createDefaultComputeThreadPool` somewhat so that users can actually construct a work-stealing pool with a non-default thread count.

djspiewak:
Fixes #1238

I also removed `internalShutdown` and a few other nice bits of cleanup. Note that this doesn't move pool configuration into `IORuntimeConfig` since pools are configured adjacent to the `IORuntimeConfig` options.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
+++ b/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
@@ -39,36 +39,12 @@ final class IORuntime private[effect] (
     val scheduler: Scheduler,
     val shutdown: () => Unit,
     val config: IORuntimeConfig,
-    private[effect] val fiberErrorCbs: FiberErrorHashtable = new FiberErrorHashtable(16),
-    private[effect] val internalShutdown: () => Unit = () => ()
+    private[effect] val fiberErrorCbs: FiberErrorHashtable = new FiberErrorHashtable(16)
 ) {
   override def toString: String = s"IORuntime($compute, $scheduler, $config)"
 }
 
-final case class IORuntimeConfig private (
-    val cancellationCheckThreshold: Int,
-    val autoYieldThreshold: Int)
-
-object IORuntimeConfig {
-  def apply(): IORuntimeConfig = new IORuntimeConfig(512, 1024)
-
-  def apply(cancellationCheckThreshold: Int, autoYieldThreshold: Int): IORuntimeConfig = {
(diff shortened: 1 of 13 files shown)


core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
```diff
@@ -39,36 +39,12 @@ final class IORuntime private[effect] (
     val scheduler: Scheduler,
     val shutdown: () => Unit,
     val config: IORuntimeConfig,
-    private[effect] val fiberErrorCbs: FiberErrorHashtable = new FiberErrorHashtable(16),
-    private[effect] val internalShutdown: () => Unit = () => ()
```
RaasAhsan, line 43: How come we don't need this anymore ?

vasilmkd: @djspiewak Hmmmm, `internalShutdown` was being called when a fatal throwable has been raised somewhere in the application and was used to clean up things. Does `IOApp` now fold this into `shutdown`? What about the default runtime on the jvm, that one seems to me that it has a noop shutdown?

vasilmkd:
How do you think the comments in #1018 fit with this PR and can we do something about it? I sort of revitalized that discussion with some new findings.

core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
djspiewak: It's now being folded into the default shutdown. However, if `IOApp` does *not* initialize `global`, then the default shutdown is empty. So `internalShutdown` is no longer needed.

PR #1747, opened 2021-02-28
Generalize defaultCont

### LINKED ISSUE #1746: CE3 defaultCont can't implement cont
The type signatures are different:
```scala
 /*
   * NOTE: This is a very low level api, end users should use `async` instead.
   * See cats.effect.kernel.Cont for more detail.
   *
   * If you are an implementor, and you have `async`, `Async.defaultCont`
   * provides an implementation of `cont` in terms of `async`.
   * Note that if you use `defaultCont` you _have_ to override `async`.
   */
  def cont[K, R](body: Cont[F, K, R]): F[R]
```

```scala
object Async {
  def apply[F[_]](implicit F: Async[F]): F.type = F

  def defaultCont[F[_], A](body: Cont[F, A, A])(implicit F: Async[F]): F[A] = ???
}
```

joroKr21:
Fixes #1746

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
@@ -86,14 +86,14 @@ trait Async[F[_]] extends AsyncPlatform[F] with Sync[F] with Temporal[F] {
 object Async {
   def apply[F[_]](implicit F: Async[F]): F.type = F
 
-  def defaultCont[F[_], A](body: Cont[F, A, A])(implicit F: Async[F]): F[A] = {
+  def defaultCont[F[_], K, R](body: Cont[F, K, R])(implicit F: Async[F]): F[R] = {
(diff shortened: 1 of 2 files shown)


joroKr21:
What resolver / version do I have to use to get this?

djspiewak:
We don't auto publish on this repo. If you need a snapshot, one will need to be manually cut. I'm pretty much afk this week, but perhaps @mpilquist or @larsrh could make one quickly?

vasilmkd:
Do we have `javacOptions` "-target:1.8" in place to make sure we guard against broken releases like Cats Effect 2.3.2 was?

joroKr21:
Oh I thought every merge is published somewhere because I used something resembling a git sha version before that but maybe it was done manually as well. Would really appreciate a release but I'm swamped myself so I wouldn't say it's urgent. Thank you for the hard work! 🙇

PR #1752, opened 2021-03-01
CE2: deprecate Sync#suspend

bplommer:
In a similar vein to #1732 - the change from `suspend` to `defer` is a particular gotcha since CE3 has `suspend` with a different signature.

This is a bit of a weird one since it still requires implementers of `Sync` to implement the deprecated `suspend` method, but I don't think that should cause any actual problems.

I'll try to add the scalafix migration if people think this is a good idea.


djspiewak:
I think this is a good idea and I approve!

PR #1759, opened 2021-03-02
Port IOApp.Simple to CE2

### LINKED ISSUE #1721: Port IOApp.Simple to CE2
In series/3.x, a simpler interface for IOApp is provided, for the users who don't need args or exit codes:

```scala
object Main extends IOApp.Simple:
  def run: IO[Unit] = IO.println("hello")
```

I think it could be backported to CE2 for the poor souls who won't be able to migrate for a while for reasons.

alexandrustana:
This PR  fixes #1721 by porting the `Simple` trait  from CE3 to CE2

### DIFF
--- a/benchmarks/shared/src/main/scala/cats/effect/benchmarks/ECBenchmark.scala
+++ b/benchmarks/shared/src/main/scala/cats/effect/benchmarks/ECBenchmark.scala
@@ -25,19 +25,18 @@ import scala.concurrent.ExecutionContext
 @BenchmarkMode(Array(Mode.Throughput))
 @OutputTimeUnit(TimeUnit.SECONDS)
 class ECBenchmark {
-  trait Run { self: IOApp =>
+  trait Run { self: IOApp.Simple =>
     val size = 100000
-    def run(args: List[String]) = {
-      val _ = args
+    def run: IO[Unit] = {
       def loop(i: Int): IO[Int] =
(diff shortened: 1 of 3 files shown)


vasilmkd (approved):
Nice. Thank you.

djspiewak:
Very nice! Thank you!

benchmarks/shared/src/main/scala/cats/effect/benchmarks/ECBenchmark.scala
```diff
@@ -25,31 +25,30 @@ import scala.concurrent.ExecutionContext
...
+  private val ioApp = new IOApp.Simple with Run
   private val ioAppCtx = new IOApp.WithContext with Run {
     protected def executionContextResource: Resource[SyncIO, ExecutionContext] =
       Resource.eval(SyncIO.pure(ExecutionContext.Implicits.global))
   }
 
   @Benchmark
   def app(): Unit = {
-    val _ = ioApp.main(Array.empty)
+    val _ = ioApp.run
```
kubukoz, line 47: I *think* this change makes the benchmarks not actually run the app at all 😅

kubukoz: Following up in https://github.com/typelevel/cats-effect/pull/1769

PR #1760, opened 2021-03-03
Multi fiber dispatcher

vasilmkd:
Revive the multi fiber dispatcher effort. This is a more down to earth implementation of the dispatcher I created by mistake when experimenting with @RaasAhsan's idea of a multiplexed dispatcher, which ultimately performed the best. This implementation is exactly the same but without the cost of a bimorphic call. I believe this is worth it because the code complexity when compared to the old dispatcher is very similar, only slight details have been altered.

The idea is that there are `Runtime.getRuntime().availableProcessors()` number of CPUs, each of which is monitoring an array of atomic references each of which simulates a queue of registrations. An incoming action to be dispatched randomly chooses a dispatcher fiber and then again randomly chooses a queue from that dispatcher fiber. This results in minimal contention between the dispatched actions.

Benchmark results:

```
series/3.x
Benchmark                       (size)   Mode  Cnt   Score   Error    Units
DispatcherBenchmark.scheduling    1000  thrpt   20  30.591 ± 0.569  ops/min
```

```
This PR
Benchmark                       (size)   Mode  Cnt   Score   Error    Units
DispatcherBenchmark.scheduling    1000  thrpt   20  72.569 ± 2.093  ops/min
```

### DIFF
--- a/benchmarks/src/main/scala/cats/effect/benchmarks/DispatcherBenchmark.scala
+++ b/benchmarks/src/main/scala/cats/effect/benchmarks/DispatcherBenchmark.scala
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2020-2021 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.benchmarks
+
+import cats.effect.IO
+import cats.effect.std.Dispatcher
+import cats.effect.unsafe._
+import cats.syntax.all._
+
+import java.util.concurrent.TimeUnit
+import org.openjdk.jmh.annotations._
+
+/**
+ * To do comparative benchmarks between versions:
+ *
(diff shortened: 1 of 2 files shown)


RaasAhsan:
Is there a benchmark for the `Dispatcher` with just the decorator approach? I only called out the bimorphic call because I figured it was worth mentioning, but I have a feeling that things like that generally don't bear much weight against allocations and barriers. If the difference between the two is minuscule, then I'm more inclined to go with the decorator approach because I do think of this as an orthogonal concern and it would be great to keep the core `Dispatcher` as lean and as simple as possible.

On a separate note, I don't think it makes sense to always allocate M fibers for a `Dispatcher` because there are use cases where `Dispatcher` is used sequentially -- FS2 has several examples here (e.g. `InputStream` and `OutputStream` conversions). We _could_ do that, but it's a poor use of resources and would incur extra GC overhead where it's unnecessary. If we can offer clear guidance on when some particular configuration makes sense, I would opt for multiple constructors.

vasilmkd:
@RaasAhsan here are the benchmark results:
https://gitter.im/typelevel/cats-effect-dev?at=603058646825b964d61e5222

The very first result is my original implementation (not in this PR, but `N` fibers and `N` state slots).
The second result is the decorator approach, 1 dispatcher multiplexing on `N` single fiber dispatchers.
The third approach I made in error, `N` fibers each multiplexing to `N` other fibers.

In this PR, I have `N` fibers each with `N` state slots, which is the same as the final result, just without the `N^2` fibers.

The results are not directly comparable to the results in this PR, as there have been many changes since.

djspiewak:
I definitely wish we could be a bit more rigorous about the number of fibers here. It feels like userland code implemented in CE shouldn't have to care about the number of processors, but… this is sort of where we are, and this PR is a significant boost to a lot of common cases without being a massive detriment in sequential cases.

PR #1761, opened 2021-03-04
Add rewrite for Sync#suspend -> defer

bplommer:
Followup to #1752. Unfortunately the output doesn't compile, because it uses CE3. I guess we need a snapshot release to build the output against.


vasilmkd:
I'm sure a set of coordinated versions of the library will be published together. It should be fine.

bplommer:
> I'm sure a set of coordinated versions of the library will be published together. It should be fine.

Sure, but in the meantime this can't be merged without breaking the build.

fthomas:
I would comment the code that does not compile with CE3 and trust you that the scalafix does the right thing. We could update the build and tests later when or if there is a snapshot release.

fthomas (approved):

Issue #975, opened 2020-07-22
Port IO Tracing


djspiewak:
Title says it all! Tracing is in series/2.x now, and it would be absolutely lovely to get it into series/3.x

Issue #1618, opened 2021-01-22
Laws testing for queue functor instances


TimWSpence:
We have `{Functor, Invariant, Contravariant}` instances for `{Q, Deq, PQ}ueue` in `cats.effect.std` but they currently have no laws tests. We should correct this :)

patseev:
I'll try to add this during the next week

TimWSpence:
> I'll try to add this during the next week

Awesome, thanks!! :)

patseev:
Started fiddling with it, and already have a couple of questions. Maybe you guys could help me with direction.

First, for law-tests I need `Eq[Queue[F, A]]` instance. I suppose I can derive it by taking all elements from Q/Deq/PQueue and comparing the lists. Does that make sense here?

Second, I need a `Gen` instance for `Queue[F, A]` and I suppose that I want to generate real instances - Bounded, Unbounded, Dropping, CircullarBuffer. Creation of these is effectful (cuz of Ref) and I'm ending up with `Gen[IO[Queue[IO, A]]]` and don't know how to solve this.

TimWSpence:
> Started fiddling with it, and already have a couple of questions. Maybe you guys could help me with direction.
> 
> First, for law-tests I need `Eq[Queue[F, A]]` instance. I suppose I can derive it by taking all elements from Q/Deq/PQueue and comparing the lists. Does that make sense here?
> 
> Second, I need a `Gen` instance for `Queue[F, A]` and I suppose that I want to generate real instances - Bounded, Unbounded, Dropping, CircullarBuffer. Creation of these is effectful (cuz of Ref) and I'm ending up with `Gen[IO[Queue[IO, A]]]` and don't know how to solve this.

Excellent questions! I think for laws/property testing we at some point have to depend on `unsafeRun`. If you've looked at `TestInstances` you may have already seen that we have
```
  implicit def eqIOA[A: Eq](implicit ticker: Ticker): Eq[IO[A]] =
    Eq.by(unsafeRun(_))
```

so I think you can piggy-back on this to get
```
//loop with tryTake until you get None
def toList[A](q: Queue[IO, A]): IO[List[A]] = ???

implicit def eqForQueue[A: Eq]: Eq[Queue[IO, A]] = Eq.by(q => toList(q))
```

As for `Gen`, I think I'd be inclined to pull a similar trick:
```
[...]

PR #1763, opened 2021-03-07
Tracing on CE3

RaasAhsan:
apologies for this taking so long, but I think this is a decent start. Instead of copying over the various performance tricks we snuck in on CE2, I elected to write things out "nicely" so that we can benchmark and optimize later.

TODO:
- [ ] I don't think anyone made use of the `IO.traced` combinator in CE2 (enhanced exceptions are way better), so I'm going to avoid porting it for now. We can add it in later
- [ ] Figure out if/where we need to improve performance
- [ ] Add in the `Trace()` op for non-cached constructors
- [ ] Port tests

### DIFF
--- a/core/js/src/main/scala/cats/effect/IOCompanionPlatform.scala
+++ b/core/js/src/main/scala/cats/effect/IOCompanionPlatform.scala
@@ -23,16 +23,16 @@ import scala.scalajs.js.Promise
 private[effect] abstract class IOCompanionPlatform { this: IO.type =>
 
   def blocking[A](thunk: => A): IO[A] =
-    apply(thunk)
+    apply(thunk, null)
 
   def interruptible[A](many: Boolean)(thunk: => A): IO[A] = {
     val _ = many
-    apply(thunk)
+    apply(thunk, null)
   }
 
   def suspend[A](hint: Sync.Type)(thunk: => A): IO[A] = {
     val _ = hint
-    apply(thunk)
+    apply(thunk, null)
   }
 
   def fromPromise[A](iop: IO[Promise[A]]): IO[A] =
--- a/core/js/src/main/scala/cats/effect/TracingConstants.scala
+++ b/core/js/src/main/scala/cats/effect/TracingConstants.scala
@@ -1,0 +1,13 @@
+package cats.effect
+
+object TracingConstants {
+  final val isCachedStackTracing: Boolean = false
+
+  final val isFullStackTracing: Boolean = false
+
+  final val isStackTracing: Boolean = isFullStackTracing || isCachedStackTracing
+
+  final val traceBufferLogSize: Int = 4
+
+  final val enhancedExceptions: Boolean = false
+}
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -263,10 +263,17 @@ private final class IOFiber[A](
 
         case 1 =>
           val cur = cur0.asInstanceOf[Error]
+
+          if (isStackTracing && enhancedExceptions) {
+            augmentException(cur.t)
+          }
+
           runLoop(failed(cur.t, 0), nextIteration)
 
         case 2 =>
           val cur = cur0.asInstanceOf[Delay[Any]]
+
+          pushEvent(cur.event)
 
           var error: Throwable = null
           val r =
@@ -309,6 +316,8 @@ private final class IOFiber[A](
 
             if (error == null) succeeded(result, 0) else failed(error, 0)
           }
+
+          pushEvent(cur.event)
 
           (ioe.tag: @switch) match {
             case 0 =>
@@ -365,6 +374,8 @@ private final class IOFiber[A](
             catch {
               case NonFatal(t) => failed(t, 0)
             }
+
+          pushEvent(cur.event)
 
           (ioe.tag: @switch) match {
             case 0 =>
@@ -454,6 +465,8 @@ private final class IOFiber[A](
         case 9 =>
           val cur = cur0.asInstanceOf[HandleErrorWith[Any]]
 
+          pushEvent(cur.event)
+
           objectState.push(cur.f)
           conts.push(HandleErrorWithK)
 
@@ -484,6 +497,8 @@ private final class IOFiber[A](
 
         case 12 =>
           val cur = cur0.asInstanceOf[Uncancelable[Any]]
+
+          pushEvent(cur.event)
 
           masks += 1
           val id = masks
@@ -799,6 +814,8 @@ private final class IOFiber[A](
         case 20 =>
           val cur = cur0.asInstanceOf[Blocking[Any]]
           /* we know we're on the JVM here */
+
+          pushEvent(cur.event)
 
           if (cur.hint eq TypeBlocking) {
             resumeTag = BlockingR
@@ -1274,10 +1291,11 @@ private final class IOFiber[A](
     failed(t, depth + 1)
   }
 
-  private[this] def pushEvent(ev: IOEvent): Unit = {
-    captured += 1
-    if (events.push(ev) != null) omitted += 1
-  }
+  private[this] def pushEvent(ev: IOEvent): Unit =
+    if (ev != null) {
+      captured += 1
+      if (events.push(ev) != null) omitted += 1
+    }
 
   /**
    * If stack tracing and contextual exceptions are enabled, this
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -23,11 +23,11 @@ import cats.arrow.FunctionK
 import scala.annotation.{switch, tailrec}
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration._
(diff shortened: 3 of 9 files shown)


kubukoz:
Closes #975

vasilmkd:
I benched this PR on my machine (I merged in the latest changes from `series/3.x`) and here are the results:
```
series/3.x
DeepBindBenchmark.async   10000  thrpt   20  2251.625 ± 25.068  ops/s
```

```
tracing none
DeepBindBenchmark.async   10000  thrpt   20  2165.577 ± 11.885  ops/s
```

```
tracing full
Benchmark                (size)   Mode  Cnt  Score   Error  Units
DeepBindBenchmark.async   10000  thrpt   20  9.111 ± 0.151  ops/s
```

```
tracing cached
DeepBindBenchmark.async   10000  thrpt   20  1473.340 ± 15.144  ops/s
```

These benchmark results are overly optimistic as there are missing operations (`pushEvent`) from the hot path.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -106,6 +107,11 @@ private final class IOFiber[A](
 
   private[this] val callbacks = new CallbackStack[A](cb)
 
+  /* tracing data structures */
+  private[this] val events: RingBuffer[IOEvent] = new RingBuffer(traceBufferLogSize)
+  private[this] var captured: Int = 0
+  private[this] var omitted: Int = 0
```
vasilmkd, line 113: I believe`captured` and `omitted` are only updated but never used anywhere. Is this an oversight?

vasilmkd:
Any ideas about tracing `cont`?

core/js/src/main/scala/cats/effect/TracingConstants.scala
```diff
@@ -0,0 +1,13 @@
+package cats.effect
+
+object TracingConstants {
```
djspiewak, line 3: I think this should be `private`

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -257,11 +263,18 @@ private final class IOFiber[A](
 
         case 1 =>
           val cur = cur0.asInstanceOf[Error]
+
+          if (isStackTracing && enhancedExceptions) {
+            augmentException(cur.t)
+          }
```
djspiewak, line 269: I believe this is going to miss exceptions caught in `map` or `flatMap`, and probably also exceptions caught in `Delay`.

core/shared/src/main/scala/cats/effect/IOTracing.scala
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.concurrent.ConcurrentHashMap
+
+private[effect] object IOTracing {
+
+  /**
+   * Global cache for trace frames. Keys are references to lambda classes.
+   * Should converge to the working set of traces very quickly for hot code paths.
+   */
+  private[this] val frameCache: ConcurrentHashMap[Class[_], IOEvent] = new ConcurrentHashMap()
```
djspiewak, line 27: Realization: we can immediately do better than this by just using `java.util.HashMap` (which is faster than Scala's `HashMap`). The reason this would work is the fact that we won't cross a memory barrier mid-mutation, so new values may be lost, but no values will be *corrupted*.

core/js/src/main/scala/cats/effect/IOCompanionPlatform.scala
```diff
@@ -23,16 +23,16 @@ import scala.scalajs.js.Promise
...
   def interruptible[A](many: Boolean)(thunk: => A): IO[A] = {
     val _ = many
-    apply(thunk)
+    apply(thunk, null)
   }
 
   def suspend[A](hint: Sync.Type)(thunk: => A): IO[A] = {
     val _ = hint
-    apply(thunk)
+    apply(thunk, null)
```
djspiewak, line 35: Wait how does this compile?

djspiewak:
> Any ideas about tracing cont?

Indexing based on the `IOCont` class would probably work fairly well. It's likely though that in most cases we just want to trace `async`, since that's what most users are going to end up with.

core/js/src/main/scala/cats/effect/IOCompanionPlatform.scala
vasilmkd: It doesn't. The build does not go so far because of the missing license headers.

vasilmkd:
> > Any ideas about tracing cont?
> 
> Indexing based on the `IOCont` class would probably work fairly well. It's likely though that in most cases we just want to trace `async`, since that's what most users are going to end up with.

Yes, but we don't have an `async` constructor.

core/shared/src/main/scala/cats/effect/IOTracing.scala
vasilmkd: I don't know about that. Writing object references is not atomic with regards to the JMM.

core/shared/src/main/scala/cats/effect/IOFiber.scala
vasilmkd: Agreed, I observed this when testing. Since we treat the `Throwable` as a black box and only care to propagate it to `unsafe` land, we probably only need to do this in the callbacks to the outside world.

vasilmkd:
I think the tests are deadlocking on this PR and I believe is because of `getClass` on a `=> A` thunk. I tried removing them and then they execute.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -360,6 +375,8 @@ private final class IOFiber[A](
               case NonFatal(t) => failed(t, 0)
             }
 
+          pushEvent(cur.event)
+
           (ioe.tag: @switch) match {
```
vasilmkd, line 380: There are missing `pushEvents` for the `Delay` mini interpreters in `Map, FlatMap and Attempt`.

core/shared/src/main/scala/cats/effect/IOTracing.scala
vasilmkd: I tried this out. It works. The gains are miniscule, but I have an idea on how to proceed.

vasilmkd: There is however this ominous warning in the javadoc of `ju.HashMap`:
```
* <p><strong>Note that this implementation is not synchronized.</strong>
 * If multiple threads access a hash map concurrently, and at least one of
 * the threads modifies the map structurally, it <i>must</i> be
 * synchronized externally.  (A structural modification is any operation
 * that adds or deletes one or more mappings; merely changing the value
 * associated with a key that an instance already contains is not a
 * structural modification.)  This is typically accomplished by
 * synchronizing on some object that naturally encapsulates the map.
 *
 * If no such object exists, the map should be "wrapped" using the
 * {@link Collections#synchronizedMap Collections.synchronizedMap}
 * method.  This is best done at creation time, to prevent accidental
 * unsynchronized access to the map:<pre>
 *   Map m = Collections.synchronizedMap(new HashMap(...));</pre>
```

core/shared/src/main/scala/cats/effect/IOFiber.scala
vasilmkd: We should have this in `def runTerminusFailureK(t: Throwable)` right before calling `done(Outcome.Errored(t))` and in the `catch case t: Throwable =>` in `def run()`

core/shared/src/main/scala/cats/effect/IOTracing.scala
djspiewak: Another thing to try here is `ClassValue`, which is *supposed* to be the optimized way of doing this kind of thing.

vasilmkd: Scala.js supports it.

SethTisue: > Realization: we can immediately do better than this by just using java.util.HashMap (which is faster than Scala's HashMap)

Just checking, is this remark based on the Scala 2.≤12 `HashMap` or the 2.13 one? Because yes the one we'd had for donkey's years was a dog, but Stefan did a full reimplementation for 2.13 and it's fast now. "On par" with the Java ones, he says. Details: https://github.com/scala/scala/pull/7348

vasilmkd:
@RaasAhsan would you mind if I took over this PR?

RaasAhsan:
@vasilmkd My bandwidth is filled up by some other priorities so it's safe to say I won't be getting to this anytime soon :) Feel free to pick up from here or start from scratch again

vasilmkd:
Thank you.

yanns:
Any help needed here?

vasilmkd:
@yanns Thanks. Unfortunately I'm actively working on this and there are quite some changes to be made that require familiarity with the IO run loop.

vasilmkd:
@yanns I'm not sure if this would be your first contribution to Cats Effect, if I may interest you in some other issues that would be easier to tackle (#1618, #1908). I apologize if I assumed wrongly.

yanns:
@vasilmkd no problem. Thanks for taking care of this!

djspiewak:
Superseded by #2051

PR #1769, opened 2021-03-08
Restore ECBenchmark behavior

kubukoz:
This was changed in #1759 and I think we should still be using `main`.


kubukoz:
@djspiewak ping :P

djspiewak:
Nice catch!

PR #1771, opened 2021-03-09
Split typeclass docs

TimWSpence:


TimWSpence:
Sorry for the delay - things have been a bit hectic recently. I think this is a good start at least (apart from `clock.md` which is _very_ threadbare). All suggestions welcome!

djspiewak:
This is amazing thank you!

Issue #1741, opened 2021-02-27
Add support for UUIDs to cats.effect.std


mpilquist:
Christopher's [fuuid](https://github.com/davenverse/fuuid) project is great. I'd love to have this built in to CE3, where we're encouraging folks to use capability traits and NOT pass `Sync` constraints around.

At minimum, we'd need a capability trait:
```scala
trait UUIDGen[F[_]]:
  def randomUUID: F[UUID]

object UUIDGen:
  given [F[_]: Sync] UUIDGen[F] with
    def randomUUID: F[UUID] = Sync[F].delay(UUID.randomUUID)
```

Some additional design decisions:
- Should we have our own wrapper class like `cats.effect.std.UUID`, which elides the unsafe constructors and operations?
- If so, should we name that `UUID` or `FUUID` or something else?
- Do we skip things like literal macros and integrations with libraries?

Wosin:
Happy to take a look at that.

mpilquist:
@Wosin Awesome, I assigned to you. Feel free to ask any questions here or on cats-effect-dev on Gitter.

chuwy:
Just a random (pun not intended) thought on this - it would be great to have a more general `Random` capability rather than UUID-specific, like [rng](https://github.com/NICTA/rng/), but more TF-like.

Wosin:
Hey @mpilquist, just a few things to clarify :) 
Were any such wrappers introduced in the past ? IMHO, this shouldn't be needed, but happy to discuss here, also IMHO if we were to introduce one I think we should keep the naming consistent and just call it `UUID`.
As for the last question, does the original lib have the literal macros ? I haven't seen that in the docs, but I may be wrong here. Adding that shouldn't be too big problem though.

djspiewak:
@chuwy We already have `Random` in the std module. :-)

chuwy:
Thanks @djspiewak, it looks awesome. I urgently need to get on CE3 train.

I'm wondering though why we'd have a separate `UUIDGen` capability then, if it's just a subset of `Random`

djspiewak:
Technically `Random` is more complex than `UUID`, meanwhile `UUID` offers some guarantees that `Random` doesn't.

Wosin:
@mpilquist What do You think about this comment ? https://github.com/typelevel/cats-effect/pull/1772#discussion_r593868748

[...]

PR #1772, opened 2021-03-09
Implement UUIDGen capability trait.

Wosin:
This PR implements UUIDGen capability trait as discussed in https://github.com/typelevel/cats-effect/issues/1741. 

The implementation is as simple as possible, not sure if we need any tests for that, but if so I will be happy to add that.
The issue mentions introducing potential literal macro, IMHO it's not a very commonly used feature, so I've just decided to skip it but If we want to have it anyway, again happy to add.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/UUIDGen.scala
+++ b/std/shared/src/main/scala/cats/effect/std/UUIDGen.scala
@@ -1,16 +1,46 @@
+/*
+ * Copyright 2020-2021 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package cats.effect.std
 
 import cats.effect.kernel.Sync
 
 import java.util.UUID
 
+/**
+ * A purely functional UUID Generator
+ */
 trait UUIDGen[F[_]] {
+
+  /**
+   * Generates a UUID in a pseudorandom manner.
+   * @return randomly generated UUID
+   */
   def randomUUID: F[UUID]
-  def fromString(uuidString: String)
+
+  /**
+   * Creates a UUID from standard String representation.
+   * @param uuidString a standard String UUID representation
+   * @return A UUID with specified value
+   */
+  def fromString(uuidString: String): F[UUID]
 }
 
 object UUIDGen {
   def randomUUID[F[_]: Sync]: F[UUID] = Sync[F].delay(UUID.randomUUID())
-  def fromString[F[_]: Sync](uuidString: String): F[UUID] = Sync[F].delay(UUID.fromString(uuidString))
+  def fromString[F[_]: Sync](uuidString: String): F[UUID] =
+    Sync[F].delay(UUID.fromString(uuidString))
 }
-
--- a/std/shared/src/main/scala/cats/effect/std/UUIDGen.scala
+++ b/std/shared/src/main/scala/cats/effect/std/UUIDGen.scala
@@ -1,0 +1,16 @@
+package cats.effect.std
+
+import cats.effect.kernel.Sync
+
+import java.util.UUID
+
+trait UUIDGen[F[_]] {
+  def randomUUID: F[UUID]
+  def fromString(uuidString: String)
+}
+
+object UUIDGen {
+  def randomUUID[F[_]: Sync]: F[UUID] = Sync[F].delay(UUID.randomUUID())
+  def fromString[F[_]: Sync](uuidString: String): F[UUID] = Sync[F].delay(UUID.fromString(uuidString))
+}
+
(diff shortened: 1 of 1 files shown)


vasilmkd:
Make sure to run `compile` in `sbt`, it will generate the needed license header. 😄

vasilmkd:
I guess you're going to have to run `scalafmtAll` in `sbt` as well. While we're at it, do you mind adding a basic scaladoc to this file? You can look at the other modules in `cats/effect/std` for inspiration.

Wosin:
> I guess you're going to have to run `scalafmtAll` in `sbt` as well. While we're at it, do you mind adding a basic scaladoc to this file? You can look at the other modules in `cats/effect/std` for inspiration.

Yeah, something actually went wrong with my git, so it was left in some weird half-state :<

std/shared/src/main/scala/cats/effect/std/UUIDGen.scala
```diff
@@ -0,0 +1,46 @@
...
+  /**
+   * Creates a UUID from standard String representation.
+   * @param uuidString a standard String UUID representation
+   * @return A UUID with specified value
+   */
+  def fromString(uuidString: String): F[UUID]
+}
+
+object UUIDGen {
+  def randomUUID[F[_]: Sync]: F[UUID] = Sync[F].delay(UUID.randomUUID())
```
catostrophe: These 2 methods should require `F[_] : UUIDGen` bound.

And the companion object should have:
```scala
implicit def fromSync[F[_] : Sync]: UUIDGen[F] = ???
```

std/shared/src/main/scala/cats/effect/std/UUIDGen.scala
```diff
@@ -0,0 +1,46 @@
...
+  def fromString[F[_]: Sync](uuidString: String): F[UUID] =
+    Sync[F].delay(UUID.fromString(uuidString))
+}
```
catostrophe, line 45: Can we also add this to the companion?
```scala
  def randomString[F[_]: Functor: UUIDGen]: F[String] = randomUUID[F].map(_.toString)
```

std/shared/src/main/scala/cats/effect/std/UUIDGen.scala
```diff
@@ -0,0 +1,46 @@
...
+   * @return randomly generated UUID
+   */
+  def randomUUID: F[UUID]
+
+  /**
+   * Creates a UUID from standard String representation.
+   * @param uuidString a standard String UUID representation
+   * @return A UUID with specified value
+   */
+  def fromString(uuidString: String): F[UUID]
```
catostrophe: This trait is a UUID _generator_. Does this method reflect the purpose of the trait? And if so, why is there no method for making a UUID from a byte array?

catostrophe: So maybe we should rename the trait and add some methods?

I also don't like the `Gen` suffix because it's used in `kernel` (yet, as a prefix) for generic type classes like `GenConcurrent`

std/shared/src/main/scala/cats/effect/std/UUIDGen.scala
Wosin: Sure, sounds reasonable.

std/shared/src/main/scala/cats/effect/std/UUIDGen.scala
Wosin: Yeah, I was kinda basing on https://github.com/davenverse/fuuid here. I had an idea to implement the byte array version too, but decided to skip it for now.

IMHO, if we are doing this it should cover all basic use-cases so I think including at least a `fromString` would be a good idea, but happy to discuss this  :)

djspiewak: I tend to think this is a bit out of scope. At the very least, I would have expected it to return `Option[UUID]`.

djspiewak (changes requested):

std/shared/src/main/scala/cats/effect/std/UUIDGen.scala
Wosin: Ok, will drop it and stick to `random` and `randomString`

std/shared/src/main/scala/cats/effect/std/UUIDGen.scala
Wosin: Hey, is that what You meant ? I think I didn't fully understand what You meant .

std/shared/src/main/scala/cats/effect/std/UUIDGen.scala
```diff
@@ -0,0 +1,49 @@
...
+   * Generates a UUID in a pseudorandom manner.
+   * @return randomly generated UUID
+   */
+  def randomUUID: F[UUID]
+
+  /**
+   * Generates a UUID in a pseudorandom manner and returns String representation of this UUID.
+   * @return randomly generated UUID string
+   */
+  def randomString: F[String]
```
catostrophe: I meant that the trait would have only `def randomUUID: F[UUID]`.
While `randomString` would be a method on the object.

Like:
```scala
object UUIDGen {
  def apply[F[_]](implicit ev: UUIDGen[F]): UUIDGen[F] = ev

  implicit def fromSync[F[_]: Sync]: UUIDGen[F] = new UUIDGen[F] {
    override def randomUUID: F[UUID] = Sync[F].delay(UUID.randomUUID())
  }

  def randomUUID[F[_]: UUIDGen]: F[UUID] = UUIDGen[F].randomUUID
  def randomString[F[_]: UUIDGen: Functor]: F[String] = randomUUID[F].map(_.toString)
}
```

Or we can just forget about it at all :)

Wosin: Cool, done!

mpilquist:
Where does this one stand?

PR #1775, opened 2021-03-10
Port align instance for IO and SyncIO from CE2

### LINKED ISSUE #1726: Missing `Align[IO]` in CE3
I'm not sure if this is intentional, but CE2 IO has an `Align` instance, which doesn't seem to be there in CE3.

```scala
implicit val ioAlign: Align[IO] = new Align[IO] {
  def align[A, B](fa: IO[A], fb: IO[B]): IO[Ior[A, B]] =
    alignWith(fa, fb)(identity)

  override def alignWith[A, B, C](fa: IO[A], fb: IO[B])(f: Ior[A, B] => C): IO[C] =
    fa.redeemWith(
      t => fb.redeemWith(_ => IO.raiseError(t), b => IO.pure(f(Ior.right(b)))),
      a => fb.redeem(_ => f(Ior.left(a)), b => f(Ior.both(a, b)))
    )

  def functor: Functor[IO] = Functor[IO]
}
```

alexandrustana:
This PR aims to fix #1726

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -17,9 +17,11 @@
 package cats.effect
 
 import cats.{
+  Align,
   Alternative,
   Applicative,
   Eval,
+  Functor,
   Id,
   Monad,
   Monoid,
@@ -31,6 +33,7 @@ import cats.{
   StackSafeMonad,
   Traverse
 }
+import cats.data.Ior
 import cats.syntax.all._
 import cats.effect.instances.spawn
 import cats.effect.std.Console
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1227,6 +1230,22 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
       a.handleErrorWith(_ => b)
   }
 
+  implicit val alignForIO: Align[IO] =
```
djspiewak: It's worth following the pattern we have elsewhere: define `alignForIO` to be a `def` which delegates to `private[this] val _alignForIO`. I realize it's a bit arcane, but this actually produces performance benefits because it avoids the memory barriers associated with `val` on `object`.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1227,6 +1230,22 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
...
+    new IOAlign
+
+  protected class IOAlign extends Align[IO] {
```
djspiewak: Why `protected`?

core/shared/src/main/scala/cats/effect/SyncIO.scala
```diff
@@ -497,6 +498,23 @@ object SyncIO extends SyncIOCompanionPlatform with SyncIOLowPriorityImplicits {
     def empty: SyncIO[A] = pure(A.empty)
   }
 
+  implicit val alignForIO: Align[SyncIO] =
+    new SyncIOAlign
+
+  protected class SyncIOAlign extends Align[SyncIO] {
```
djspiewak: Same comments as with `IO`

djspiewak (changes requested):
Some minor comments, but otherwise looks very good!

core/shared/src/main/scala/cats/effect/IO.scala
alexandrustana: Ok, I'll switch to this implementation then.
Thanks for the tip :)

core/shared/src/main/scala/cats/effect/IO.scala
alexandrustana: While adding `alignForIO` I've used `semigroupKForIO` as a reference. Didn't really thought it through when choosing the access specifier.

djspiewak:
🎉

PR #1776, opened 2021-03-10
Return `Either` rather than `Boolean` from `Deferred#complete`

bplommer:
As discussed in #1687. This makes it more obvious that the return value of `Deferred#complete` can represent an error state, and (when `F` has `MonadThrow`) allows just using `.rethrow` to convert a failure value into a failed `F`.

### DIFF
--- a/kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/pure.scala
+++ b/kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/pure.scala
@@ -275,7 +275,12 @@ object pure {
         new Deferred[PureConc[E, *], A] {
           override def get: PureConc[E, A] = mVar.read[PureConc[E, *]]
 
-          override def complete(a: A): PureConc[E, Boolean] = mVar.tryPut[PureConc[E, *]](a)
+          override def complete(
+              a: A): PureConc[E, Either[DeferredSink.AlreadyCompleted, Unit]] =
+            mVar.tryPut[PureConc[E, *]](a).map {
+              case true => Right(())
+              case false => Left(DeferredSink.AlreadyCompleted)
+            }
 
           override def tryGet: PureConc[E, Option[A]] = mVar.tryRead[PureConc[E, *]]
         }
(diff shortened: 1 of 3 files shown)


bplommer:
This is failing on Scala 3.0.0-M3 with `java.lang.AssertionError: assertion failed: no companion AlreadyCompleted$ in dotty.tools.dotc.core.Scopes$MutableScope@4e4afe7d`. Seems to be a compiler bug, as it compiles fine on 3.0.0-RC1.

kubukoz (approved):

djspiewak:
Is this really better than `Boolean` though? Users can always just `.whenM` and `raiseError` on the results with their own error type, if that semantic is required. It's a bit less convenient than `rethrow`, but I'm also not convinced that error raising is *that* pervasively required of a semantic.

Also, not to bikeshed, but why `Either[Error, Unit]` rather than just `Option[Error]`?

kubukoz:
> Is this really better than Boolean though? Users can always just .whenM and raiseError on the results with their own error type, if that semantic is required. It's a bit less convenient than rethrow, but I'm also not convinced that error raising is that pervasively required of a semantic.

It's not just convenience, we're being more explicit about what true/false means. I don't even care about the error being a Throwable, but giving it a name (even a newtype over Boolean would do) is something I think is valuable.

> why Either[Error, Unit] rather than just Option[Error]?

Errors in an Option don't have the same meaning to me as Left (weird to have an error in the successful branch of the ADT, which "Some" is).

djspiewak:
I think it makes more sense to just do a straight enum algebra here rather than paying the penalty of the `Either` allocation on every `complete`, not to mention the fact that we're imposing an error type which may or may not be what the user wants.

I personally think that `Boolean` is reasonably sensible in this case, for the same reason that I think it's a sensible return result from `compareAndSwap`, but I'm open to the enum answer. Something like this, but with better names?

```scala
sealed trait Completion

object Completion {
  case object Successful extends Completion
  case object Failed extends Completion
}
```

One thing I *don't* like about this is we don't really use this idiom anywhere else in Cats Effect or in Cats. This also doesn't play particularly nicely with other combinators, such as `ifM` and such.

mpilquist:
I'm in favor of sticking with `F[Boolean]`

kubukoz:
We don't need to pay for the allocation, there are two values which we can cache forever (I think)... Anyway i don't have a strong opinion anymore.

PR #1778, opened 2021-03-11
Fix the numbers in the MonadCancel docs

davidabrahams:
Noticed a typo in the docs! Thank you as always for such a well documented and educational library :)


djspiewak:
Thank you!

PR #1788, opened 2021-03-16
Remove MonadThrow & ApplicativeThrow

bplommer:
These aliases are now part of cats, so having them in CE is redundant and a possible source of confusion.


rossabaker (approved):
This is a breaking change, but I think worthwhile before we have to drag it along for the life of CE3.

fthomas:
👍 The breakage could be mitigated by adding rewrites to the Scalafix migration (#1686).

Issue #1568, opened 2020-12-31
Add JMX beans for IOApp thread pool metrics


Daenyth:
Spinning this off from #1238 

I think metric collection is an important feature (it is for me, at the very least), so we should enable users to either enrich on their own beans, or build in ones that could be useful.

For example, this is what I'm using right now, building on a standard thread pool:

```scala
  trait ThreadPoolSamplerMBean {
    def getMinThreads: Int
    def getMaxThreads: Int
    def getActiveThreads: Int
    def getCurrentSize: Int
    def getQueuedTasks: Int
  }

  class ThreadPoolSampler(pool: ThreadPoolExecutor) extends ThreadPoolSamplerMBean {
    override def getActiveThreads: Int = pool.getActiveCount
    override def getMinThreads: Int = pool.getCorePoolSize
    override def getMaxThreads: Int = pool.getMaximumPoolSize
    override def getCurrentSize: Int = pool.getPoolSize
    override def getQueuedTasks: Int = pool.getQueue.size
  }
```

And then I register beans of that for both my main compute EC (backing ContextShift) as well as my Blocker EC.

For CE3, it would be *amazing* if we could go even further - build in knowledge of fibers in some way, like number of pending fibers, or number of concurrent fibers waiting for a thread. Worker pool depth, perhaps. I think there's a lot of options there.

djspiewak:
Wdyt, @vasilmkd ?

vasilmkd:
It should be doable. Is there an expectations that these metrics should be dynamic? I'm asking because the thread pool on the JVM uses `Runtime#availableProcessors()`, but they are configured to sleep if there is no work to go around. Again, it should be doable, there's just going to be some discussion as to what metrics are useful and how they should be reported. Otherwise, the the pool is fiber aware, so it can already give you extremely accurate metrics for fibers.

Daenyth:
> Is there an expectations that these metrics should be dynamic

Things like `getActiveThreads`, it's definitely important that it be at least *relatively* up to date. Whether that's updated at the time the method is polled, or independently on a schedule, or constantly updated in a field, doesn't matter so much to me at least

vasilmkd:
Cool, I'm working on a large rewrite of the whole pool which should reduce the code footprint a lot, I've just been busy with other things these weeks. I'd like to get that relatively stable and then I can look into adding configuration for this. It would be an awesome addition, even for us, as we continue to tune and upgrade the pool.

djspiewak:
[...]

PR #1789, opened 2021-03-17
Compute and local queue MBeans

vasilmkd:
A stab at #1568.

The MBean registration and deregistration logic is a bit of a mess. It was actually crashing the tests because of duplicate MBeans. I think I managed to find a somewhat nice solution.

@Daenyth Let me know what you think. This is my first ever exposure to MBeans, so any feedback is greatly appreciated.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/HelperThread.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/HelperThread.scala
@@ -58,6 +58,7 @@ import java.util.concurrent.atomic.{AtomicBoolean, AtomicInteger}
 private final class HelperThread(
     private[this] val threadPrefix: String,
     private[this] val blockingThreadCounter: AtomicInteger,
+    private[this] val activeHelperThreadGauge: AtomicInteger,
     private[this] val batched: ScalQueue[Array[IOFiber[_]]],
     private[this] val overflow: ScalQueue[IOFiber[_]],
     private[this] val pool: WorkStealingThreadPool)
@@ -129,6 +130,8 @@ private final class HelperThread(
    * for the [[HelperThread]] to exit its runloop and die.
    */
   override def run(): Unit = {
+    activeHelperThreadGauge.incrementAndGet()
+
     random = ThreadLocalRandom.current()
     val rnd = random
 
@@ -149,6 +152,7 @@ private final class HelperThread(
           // the `overflow` queue. This pathological case is not handled as it
           // is a case of uncontrolled blocking on a fixed thread pool, an
           // inherently careless and unsafe situation.
+          activeHelperThreadGauge.decrementAndGet()
           return
         } else {
           overflow.offerAll(batch, rnd)
@@ -157,6 +161,9 @@ private final class HelperThread(
         fiber.run()
       }
     }
+
+    activeHelperThreadGauge.decrementAndGet()
+    ()
   }
 
   /**
@@ -177,7 +184,13 @@ private final class HelperThread(
 
       // Spawn a new `HelperThread`.
       val helper =
-        new HelperThread(threadPrefix, blockingThreadCounter, batched, overflow, pool)
+        new HelperThread(
+          threadPrefix,
+          blockingThreadCounter,
+          activeHelperThreadGauge,
+          batched,
+          overflow,
+          pool)
       helper.start()
 
       // With another `HelperThread` started, it is time to execute the blocking
--- a/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
@@ -16,10 +16,15 @@
 
 package cats.effect.unsafe
 
+import cats.effect.unsafe.metrics.ComputePoolSampler
+
+import scala.collection.mutable
 import scala.concurrent.ExecutionContext
 
+import java.lang.management.ManagementFactory
 import java.util.concurrent.{Executors, ScheduledThreadPoolExecutor}
 import java.util.concurrent.atomic.AtomicInteger
+import javax.management.ObjectName
 
 private[unsafe] abstract class IORuntimeCompanionPlatform { this: IORuntime.type =>
 
@@ -30,7 +35,47 @@ private[unsafe] abstract class IORuntimeCompanionPlatform { this: IORuntime.type
       threadPrefix: String = "io-compute"): (WorkStealingThreadPool, () => Unit) = {
     val threadPool =
       new WorkStealingThreadPool(threads, threadPrefix, self)
-    (threadPool, { () => threadPool.shutdown() })
+
+    val unregisterMBeans =
+      if (MetricsConstants.metricsEnabled) {
+        val mBeanServer =
+          try ManagementFactory.getPlatformMBeanServer()
+          catch {
+            case t: Throwable =>
+              t.printStackTrace()
+              null
+          }
+
+        val registeredObjects = mutable.Set.empty[ObjectName]
+
+        if (mBeanServer ne null) {
+          val computePoolSamplerName =
+            new ObjectName("cats.effect.metrics:type=ComputePoolSampler")
+          val computePoolSampler = new ComputePoolSampler(threadPool)
+
+          try {
+            mBeanServer.registerMBean(computePoolSampler, computePoolSamplerName)
+            registeredObjects += computePoolSamplerName
+          } catch {
+            case t: Throwable =>
+              t.printStackTrace()
+          }
+        }
+
+        () => {
+          if (mBeanServer ne null) {
+            registeredObjects.foreach(mBeanServer.unregisterMBean)
+          }
+        }
+      } else () => ()
+
+    (
+      threadPool,
+      { () =>
+        unregisterMBeans()
+        threadPool.shutdown()
+      }
+    )
   }
 
   def createDefaultBlockingExecutionContext(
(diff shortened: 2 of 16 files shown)


vasilmkd:
Not sure if this is an issue with GraalVM and Java 8. I tested it locally on GraalVM and Java 11, it worked fine.

core/jvm/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -40,7 +40,10 @@ trait IOApp {
 
       IORuntime installGlobal {
         val (compute, compDown) =
-          IORuntime.createDefaultComputeThreadPool(runtime, threads = computeWorkerThreadCount)
+          IORuntime.createDefaultComputeThreadPool(
+            runtime,
+            threads = computeWorkerThreadCount,
+            registerMBeans = true)
```
Daenyth: 🤔 It may be helpful to users for us to make this toggle-able via a jvm property, so it can be controlled with a -Darg.

If you'd consider that out of scope for this PR, that's also sensible. If we decide we want that toggle, I could send it in later

core/jvm/src/main/scala/cats/effect/unsafe/HelperThread.scala
```diff
@@ -58,6 +58,7 @@ import java.util.concurrent.atomic.{AtomicBoolean, AtomicInteger}
 private[effect] final class HelperThread(
     private[this] val threadPrefix: String,
     private[this] val blockingThreadCounter: AtomicInteger,
```
Daenyth, line 60: ⛏️ Because we use an "increment, run, decrement" approach, I might consider renaming this;
```suggestion
    private[this] val blockingThreadGauge: AtomicInteger,
```

core/jvm/src/main/scala/cats/effect/unsafe/HelperThread.scala
```diff
@@ -139,6 +144,11 @@ private[effect] final class HelperThread(
         fiber.run()
       }
     }
+
+    // This helper thread is finishing and is no longer counted in the global
+    // counter.
+    blockingThreadCounter.decrementAndGet()
+    ()
```
Daenyth: ❓ Is `fiber.run()` permitted to throw exceptions? Do we want to wrap the `decrement` action in a `finally` block?

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
```diff
@@ -16,21 +16,62 @@
...
-      threadPrefix: String = "io-compute"): (WorkStealingThreadPool, () => Unit) = {
+      threadPrefix: String = "io-compute",
+      registerMBeans: Boolean = false,
+      mBeansPostfix: String = ""): (WorkStealingThreadPool, () => Unit) = {
     val threadPool =
       new WorkStealingThreadPool(threads, threadPrefix, self)
-    (threadPool, { () => threadPool.shutdown() })
+
+    val objectNames = mutable.Set.empty[ObjectName]
+    val mBeanServer = ManagementFactory.getPlatformMBeanServer()
```
Daenyth: 🤔 I'm not sure if the `get..` action can fail - I *think* it's permitted to throw.

I think "catch, print-log, and move on" might be the best approach in that case.

❓ Do we want the `get` call to be inside the `if (registerMBeans)` branch? Seems little point in the call here, if `registerMBeans` was false

core/jvm/src/main/scala/cats/effect/unsafe/LocalQueue.scala
```diff
@@ -207,6 +207,44 @@ private final class LocalQueue {
    */
   private[this] val overflowBuffer: ArrayList[IOFiber[_]] = new ArrayList(OverflowBatchSize)
 
+  /*
+   * What follows is a collection of counters exposed through the
+   * `cats.effect.unsafe.metrics.LocalQueueSamplerMBean` interface.
+   */
```
Daenyth: 👍 This is helpful

core/jvm/src/main/scala/cats/effect/unsafe/LocalQueue.scala
```diff
@@ -600,6 +643,20 @@ private final class LocalQueue {
    */
   def nonEmpty(): Boolean = !isEmpty()
 
+  /**
+   * Returns the number of currently enqueued fibers in this queue.
+   *
+   * @note This number is an approximation and may not be correct while
+   *       stealing is taking place.
```
Daenyth, line 682: 👍 This is a reasonable restriction. I wouldn't have a problem with that for apps I'm monitoring

core/jvm/src/main/scala/cats/effect/unsafe/metrics/ComputePoolSamplerMBean.scala
```diff
@@ -0,0 +1,80 @@
...
+ * An MBean interface for monitoring the IO work stealing compute pool.
+ */
+trait ComputePoolSamplerMBean {
+
+  /**
+   * Returns the number of worker threads in the compute pool.
+   *
+   * @return the number of worker threads in the compute pool
+   */
+  def getWorkerThreadCount: Int
```
Daenyth, line 30: Do nullary methods get compiled to normal jvm zero-arg methods with `()`?

Because the beans are a java-ish thing, we may want to enforce explicit `()` application for the methods in this interface.

```suggestion
  def getWorkerThreadCount(): Int
```

core/jvm/src/main/scala/cats/effect/IOApp.scala
vasilmkd: Feel free to extend this PR if you like. I don't mind. I would also like it to be a toggle. There's also the GraalVM 8 issue present at the moment. Maybe some of your suggestions can solve it.

core/jvm/src/main/scala/cats/effect/unsafe/HelperThread.scala
vasilmkd: Feel free to tackle this in the code.

core/jvm/src/main/scala/cats/effect/unsafe/HelperThread.scala
vasilmkd: No, all exceptions are caught inside the `fiber` runloop (this is in `IOFiber#run`).

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
vasilmkd: That's a good catch, thank you. This code definitely needs some cleanup.

core/jvm/src/main/scala/cats/effect/unsafe/LocalQueue.scala
vasilmkd: It's honestly inevitable, during a steal operation, some of the fibers might appear to "vanish", but it's just the implementation where it takes some time for them to be published in a different queue, so it really depends if you're unlucky enough to observe this anomaly.

vasilmkd: Actually I may be able to do something about it, I'll see what I can do.

core/jvm/src/main/scala/cats/effect/unsafe/metrics/ComputePoolSamplerMBean.scala
vasilmkd: I honestly followed your example in the linked issue 😅. I didn't give it too much thought. But it might be a good change.

Daenyth: Well, I suppose if what I pasted worked, then it just works, because that's what I have in prod :)

core/jvm/src/main/scala/cats/effect/unsafe/LocalQueue.scala
djspiewak: FWIW, I'm not worried about this at all. External observation of concurrent processes should not be expected to be deterministic.

core/jvm/src/main/scala/cats/effect/unsafe/HelperThread.scala
kubukoz: It won't be hit in the `if fiber eq null then return` case, no? Maybe it's best to still wrap it in `finally`?

vasilmkd: Oh you're right...

djspiewak:
Bumping this a bit

djspiewak:
Bump again :-) We can close if it's not going to get into a working state.

djspiewak:
Holy metrics…

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -103,6 +104,25 @@ private[effect] final class WorkStealingThreadPool(
...
+   * whose asynchronous callback has not been executed yet). Does not include
+   * fibers which have not started execution at all.
+   */
+  private[this] val suspendedFiberGauge: LongAdder = new LongAdder()
+
+  /**
+   * An atomic counter used for tracking the number of fibers started on this
+   * pool during its whole lifetime.
+   */
+  private[this] val lifetimeFiberGauge: LongAdder = new LongAdder()
```
vasilmkd, line 124: I know of a way to make this non-volatile.

vasilmkd:
So, I'm not sure if metrics should be on by default. If they indeed aren't how should we test this in the CI?

djspiewak:
I tend to think they should be on by default. Ideally it should be easy for someone to set up dashboards operationalizing a CE application without changing any other configurations.

PR #1794, opened 2021-03-17
Replace interruptible with blocking in Console.readLine

### LINKED ISSUE #1793: Console.readLine is not interruptible due to `InputStream#read`
The produced effect should not use `IO.interruptible` but `IO.blocking` instead.

`InputStream#read` is not always an interruptible action, so our previous assumption was incorrect.

Some more details[here](https://stackoverflow.com/questions/3843363/thread-interrupt-not-ending-blocking-call-on-input-stream-read), for anyone interested, but not necessary.

alexandrustana:
Fixes #1793


vasilmkd (approved):

djspiewak:
This looks good! Holding pending my comment on the issue though

vasilmkd:
Unfortunately, I may have been too quick to open the original discussion. Nonetheless, it was a fruitful discussion. Thank you for your contribution, but I'm going to have to close this PR without merging it.

PR #1799, opened 2021-03-21
Updated tutorial to new doc site + CE 3.0.0-RC2

lrodero:
Updated tutorial for CE `3.0.0-RC2`. Changes to note:

- The producer/consumer section does not use semaphores any longer. Instead it is coded using the same approach than in the implementation of CE3's `std/Queue.scala`.
- Some links had to be removed as they are not present (yet?) in the current version of the new website.
- Scattered syntax fixes added.
- Current `build.sbt` in the text points to `3.0.0-RC2`. As soon as final version is released I can prepare a small change so it points to `3.0.0`.

This new version of the tutorial can be read [here](https://lrodero.github.io/cats-effect/docs/tutorial).

Code samples used in the tutorial are available [here](https://github.com/lrodero/cats-effect-tutorial/tree/series/3.x).

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -1,0 +1,1220 @@
+---
+id: tutorial
+title: Tutorial
+---
+
+## Introduction
+
+This tutorial tries to help newcomers to cats-effect to get familiar with its
+main concepts by means of code examples, in a learn-by-doing fashion. Two small
+programs will be coded, each one in its own section. [The first
+one](#copyingfiles) copies the contents from one file to another, safely
+handling _resources_ and _cancellation_ in the process. That should help us to
+flex our muscles. [The second one](#producerconsumer) implements a solution to
+the producer-consumer problem to introduce cats-effect _fibers_.
+
+This tutorial assumes certain familiarity with functional programming. It is
+also a good idea to read cats-effect documentation prior to starting this
+tutorial, at least the [Getting Started page](getting-started.md).
+
+Please read this tutorial as training material, not as a best-practices
+document. As you gain more experience with cats-effect, probably you will find
+your own solutions to deal with the problems presented here. Also, bear in mind
+that using cats-effect for copying files or implementing basic concurrency
+patterns (such as the producer-consumer problem) is suitable for a 'getting
+things done' approach, but for more complex systems/settings/requirements you
+might want to take a look at [fs2](http://fs2.io) or [Monix](https://monix.io)
+to find powerful network and file abstractions that integrate with cats-effect.
+But that is beyond the purpose of this tutorial, which focuses solely on
+cats-effect.
+
+That said, let's go!
+
+## Setting things up
+
+This [Github
+repo](https://github.com/lrodero/cats-effect-tutorial/tree/series/3.x) includes
+all the software that will be developed during this tutorial (branch
+`series/3.x`). It uses `sbt` as the build tool. To ease coding, compiling and
+running the code snippets in this tutorial it is recommended to use the same
+`build.sbt`, or at least one with the same dependencies and compilation options:
+
+```scala
+name := "cats-effect-tutorial"
+
+version := "3.0.0-RC2"
+
+scalaVersion := "2.13.4"
+
+libraryDependencies += "org.typelevel" %% "cats-effect" % "3.0.0-RC2" withSources() withJavadoc()
+
+scalacOptions ++= Seq(
+  "-feature",
+  "-deprecation",
+  "-unchecked",
+  "-language:postfixOps"
+)
+```
+
+Also make sure that you use a recent version of `sbt`, at least `1.4.2`. You can
+set the `sbt` version in `project/build.properties` file:
+
+```scala
+sbt.version=1.4.2
+```
+
+Almost all code snippets in this tutorial can be pasted and compiled right in
+the scala console of the project defined above (or any project with similar
+settings).
+
+## <a name="copyingfiles"></a>Copying files - basic concepts, resource handling and cancellation
+
+Our goal is to create a program that copies files. First we will work on a
+function that carries such task, and then we will create a program that can be
+invoked from the shell and uses that function.
+
+First of all we must code the function that copies the content from a file to
+another file. The function takes the source and destination files as parameters.
+But this is functional programming! So invoking the function shall not copy
+anything, instead it will return an `IO` instance that encapsulates all the side
+effects involved (opening/closing files, reading/writing content), that way
+_purity_ is kept.  Only when that `IO` instance is evaluated all those
+side-effectful actions will be run. In our implementation the `IO` instance will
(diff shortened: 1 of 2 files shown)


docs/tutorial.md
```diff
@@ -0,0 +1,1220 @@
...
+  } yield (inStream, outStream)
+
+def copy(origin: File, destination: File): IO[Long] = {
+  for {
+    guard <- Semaphore[IO](1)
+    count <- inputOutputStreams(origin, destination, guard).use { case (in, out) => 
+      guard.permit.use { _ =>
+        transfer(in, out)
+      }
+    }
```
djspiewak, line 378: Would it be more clear to rewrite this in the following fashion?

```scala
  val rsrc = for {
    guard <- Resource.eval(Semaphore[IO](1))
    _ <- inputOutputStreams(origin, destination, guard)
    _ <- guard.permit
    count <- Resource.eval(transfer(in, out))
  } yield count

  rsrc.use(IO.pure(_))
```

docs/tutorial.md
```diff
@@ -0,0 +1,1220 @@
...
+  def copy(origin: File, destination: File): IO[Long] = ???
+
+  override def run(args: List[String]): IO[ExitCode] =
+    for {
+      _      <- if(args.length < 2) IO.raiseError(new IllegalArgumentException("Need origin and destination files"))
+                else IO.unit
+      orig = new File(args(0))
+      dest = new File(args(1))
+      count <- copy(orig, dest)
+      _     <- Console[IO].println(s"$count bytes copied from ${orig.getPath} to ${dest.getPath}")
```
djspiewak: `IO.println` also works. :-)

docs/tutorial.md
```diff
@@ -0,0 +1,1220 @@
...
+apply if the `copy` function is run from other modules that require its
+functionality. If the `IO` returned by this function is canceled while being
+run, still resources will be properly released. But recall what we commented
+before: this is because `use` returns `IO` instances that are cancelable, in
+contrast our `transfer` function is not cancelable.
+
+WARNING: To properly test cancelation, You should also ensure that
+`fork := true` is set in the sbt configuration, otherwise sbt will
+intercept the cancelation because it will be running the program
+in the same JVM as itself.
```
djspiewak: This warning should not be required any longer since `IOApp` detects this scenario and does the Right Thing™

docs/tutorial.md
```diff
@@ -0,0 +1,1220 @@
...
+import cats.syntax.all._
+import collection.immutable.Queue
+
+def consumer[F[_] : Sync: Console](queueR: Ref[F, Queue[Int]]): F[Unit] =
+  (for {
+    iO <- queueR.modify{ queue =>
+      queue.dequeueOption.fold((queue, Option.empty[Int])){case (i,queue) => (queue, Option(i))}
+    }
+    _ <- if(iO.exists(_ % 10000 == 0)) Console[F].println(s"Consumed ${iO.get} items") else Sync[F].unit
+  } yield ()) >> consumer(queueR)
```
djspiewak: Should we just recommend better-monadic-for so we can pull the `consumer(queueR)` into the `for`-comprehension?

djspiewak (commented):
Absolutely amazing! Thank you for tackling this! It was on my todo list, but my todo list is massive at this point…

Quick question: can we adjust the code snippets to build with mdoc so that they don't go stale?

lrodero:
Glad to help :) , sorry I didn't undertake this sooner.

No prob on using mdoc, but I have never used it before. Any modifier I should use? `compile-only` maybe?

djspiewak:
> No prob on using mdoc, but I have never used it before. Any modifier I should use? compile-only maybe?

I *think* so? My knowledge of mdoc rounds to zero. :-D @keynmol might have some suggestions

keynmol:
@lrodero given that the docs/ folder should'be already covered by mdoc, then yeah, changing all your snippets (apart from SBT ones) from

````
```scala
````

to 

````
```scala mdoc:compile-only
````

Will make sure they're compiled along with the code, but they won't be executed (which for bare IO means nothing without unsafeRunSync or using a special modifier ~which I totally was going to finish but never did~.

Couple of points:

1. Mdoc processes snippets linearly, so if you see any conflicts (`is already defined bla bla`), just add `:nest` to the list of modifiers
2. Ideally running the snippets would be cool, but for a tutorial it's actually more beneficial to see the self contained `IOApp` snippet, rather than just the individual bits of IO actions running

File: docs/tutorial.md (hunk 2 above)
Comment (lrodero) [reply docs/tutorial.md]: Yes! But there are `Console[IO].errorln` calls in the snippets while there is no `IO.errorln`. So for the sake of homogeneity I wasn't sure what to do here. I'll change it and explain that `IO.println`is just `Console[IO].println` underneath.

File: docs/tutorial.md (hunk 3 above)
Comment (lrodero) [reply docs/tutorial.md]: Cool! Removing the warning then :+1:

**Comment by keynmol** [comment]
And your magic SBT watch command for gradually updating the tutorial is:

```
sbt:root> docs/mdoc --in docs/tutorial.md --watch
```

(mdoc has its own watch mode for inputs)

docs/tutorial.md
lrodero: In fact I put the recursive call outside the `for` so it's easily spotted by the reader. Like, 'this is the main thing to do and then we start all over again'. But if you think it would be actually clearer putting the recursive call inside the `for` I'd be happy to change it.

docs/tutorial.md
lrodero: Well... it's undeniably more elegant, but I'm having trouble myself reading the code (sorry :) ), I guess other newcomers can find it also a bit daunting. Also, it doesn't compile... I think I get the idea but I'd need to refactor the code to make it work.

docs/tutorial.md
djspiewak: Totally fair!

docs/tutorial.md
djspiewak: It's kind of a stylistic thing. IMO I prefer having it inside, the same way it would be if you were writing something like this in an imperative style.

lrodero: No prob then, changing it :+1:

lrodero:
Thanks for the tips @keynmol ! Ok I think I got it. Two things to note:
1. I've addressed most of @djspiewak comments, save the first one as it requires a refactor that can lead to a rabbit hole of text rewrites :sweat_smile: 
2. `mdoc` works fine. But I'm not using it in a couple of snippets as anyway they are not supposed to compile.

docs/tutorial.md
lrodero: If you don't mind I'd rather let the code as it's now. I still think it will be easier for people with little or no experience with CE. Also, the change would require some substantial rewrite I think :/ .

PR #1800, opened 2021-03-21
Opened uncancelable region to the right

### LINKED ISSUE #1744: Uncancelable boundaries compose poorly
The motivating example here:

```scala
val fa = F.uncancelable(_ => fb) // where fb is something long but not infinite
F.uncancelable(poll => poll(fa) *> fc)
```

If we imagine canceling the above where cancellation is observed within `fb`, the cancellation will be suppressed until the end of the inner `uncancelable` (within `fa`) and then immediately become visible *before* the end of the `poll`, meaning that `fc` will not run despite the fact that there's no apparent "gap" between the end of the `uncancelable` and the end of the `poll`.

The *reason* this is visible is to abide by the functor laws. Specifically, consider the following pair of expressions:

```scala
fa.onCancel(fin)
fa.map(identity).onCancel(fin)
```

Per the functor laws, these must be equivalent. Thus, cancellation must be visible *at the end* of an `uncancelable` block just as it would be at the end of an `uncancelable` block + some identity `map` or `flatMap` action. (note: there's some wiggle room here due to the soft nature of external cancellation and the fact that we're allowed to let the signal remain unobserved for some time, but internal cancellation makes this brutally and immediately apparent)

Unfortunately, this kind of visibility makes all sorts of things harder or impossible to express. `allocated` on `Resource` becomes particularly dangerous, since you can't guarantee that you're able to get at the *results* of the `allocated` before some cancellation is made visible, and the results contain a finalizer which would not have been run within the `uncancelable` block (since the cancellation would not be visible at all until after such a block). This scenario commonly arises with `Resource.make` (where the `poll` is ignored). This probably means that `Async[Resource]` in its present form has some tiny bits of unsound finalizer semantics littered around its implementation.

In general, it's just not possible to trust, given an arbitrary expression `F[A]`, that one of the following *always* holds:

1. Results which are sensitive to leakage in the face of cancellation are wrapped such that any cancellation manifested *within* `fa` produces the appropriate cleanup
2. Cancellation which is unobserved within `fa` cannot be observed within some enclosing expression prior to that expression's first opportunity to handle it without loss

This is violated because `fa` in the above internally handles cancellation to the best of its ability (by wrapping its own critical region), but the outer expression is unable to handle cancellation before the point at which the results from `fa` are lost, which is precisely what `fa` was trying to avoid by wrapping itself in `uncancelable`.

I can think of a couple ways of addressing this issue:

- Violate the functor laws. I'm... not fond of this one, though as noted above, there is some interpretive wiggle room in those laws when it comes to external cancellation, which is already a hint and not a dictate. Still, I'm not *super* comfortable with substitutability being violated in such a blatant and easy-to-hit fashion. Our current violation of substitutability of the functor identity in the face of external cancellation requires 512 `map(identity)` calls in a row, which seems... improbable.
- Beef up `onCancel` in some fashion which allows it to detect the case where a value is produced despite cancellation. This runs afoul of the same functor law issue though in very short order, since we have to answer the question of what it means for an `uncancelable` to finish successfully, then get `map`ped once, then hit this hypothetical `onCancel`.
- Enhance `uncancelable` to optionally carry a "canceled despite completion" finalizer which would be passed the result (or presumably, the error). This at least seems theoretically tractable, though it probably re-raises the same composability issues as `bracketCase`. Also it's not clear it really solves the problem, since there would still be a gap wherein cancellation could be lost.
- Adopt an idiom of accepting a `Poll[F]` as a parameter on signatures which would normally have an inner `uncancelable`. So `Resource#allocated`, for example, would take a `Poll[F]` rather than having its own `uncancelable`. That's not necessarily safe though because it kind of depends on the "good will" of the caller
- Adopt an idiom of accepting a `Poll[F]` as a parameter and then *composing* it with an inner `Poll[F]`. In other words, instead of `uncancelable(poll => poll(uncancelable(poll => poll(fa))))` (where the inner `uncancelable` is likely in a different function), we would carrot people to do something like `uncancelable(outer => uncancelable(inner => outer(inner(fa))))`. This would be an appropriate step to take for any subexpression which produces results which must be considered critical (e.g. `racePair` is dodgy).

None of these are great. All of them involve resource leaks, and the one which has a special case in which resources *don't* leak achieves this by violating algebraic identities... and still leaks resources if people (reasonably) assume those identities hold. I think the *last* of them is probably the most achievable.

Btw this whole discussion is related to #1027

djspiewak:
Fixes #1744 

Requesting expedited reviews from @SystemFw, @wemrysi, @RaasAhsan, and @vasilmkd. Also anyone else who wants to pile on, please do so quickly. I would ideally like to get this merged today so we can release quickly.

This is a major semantic shift in several ways. Bottom line up front: **this change violates the functor laws**. If you're looking for a thing that Cats Effect is choosing to do in the name of "pragmatism over purity", this is it. We didn't take this decision lightly.

Just to provide a bit of context, consider the following construction:

```scala
uncancelable(_(uncancelable(fa) /* think carefully about this spot */))
```

In the encoding which is present in the current HEAD, the commented location in the expression above represents a very small gap in which cancelation may be observed but not suppressed. This in turn means that `poll` and `uncancelable` don't compose *safely*, and this has some quite profound impacts on everything from Fs2 to `Resource`. Ultimately, any function which wraps its contents in `uncancelable` and returns was *actually unsafe* because the caller couldn't close the gap.

This PR changes `uncancelable` to be an *open* region to the right. In other words, we no longer perform a cancelation check upon completion of the region. This is well exemplified by the new law:

```scala
uncancelable(_ => fa.onCancel(fin)) <-> uncancelable(_ => fa).onCancel(fin)
```

Which, transitively, also means the following:

```scala
uncancelable(_ => fa).onCancel(fin) <-> uncancelable(_ => fa)
```

How does this violate the functor laws? Well, imagine an identity `flatMap`:

```scala
uncancelable(_ => fa).flatMap(pure).onCancel(fin)
```

(by the monad laws, this is equivalent to `.map(identity)`, which is why this is about the functor laws specifically)

The above is *not* equivalent: the `onCancel` will sequence the finalizer if `fa` is canceled. If you think about it, this is fundamentally necessary if we want to allow auto-cancelable `flatMap`, but it's also a violation of the functor laws because identity mappings must be identity.

As it turns out, violating the functor laws has… some implications. Nearly all of this PR is just running through and trying to correct all of the problems which arise when you do something like that, though I did also fix a very subtle bug in `Resource#allocated` as well as implemented a more optimized `IO#guarantee`, which should help `Resource` performance a bit.

The hand-wavy justification here is that the violation of functor laws is only observable within confined runtime scopes. Like, for example, what we do in the tests. 😢 The requisite incantation to resurface the old semantics is as follows:

```scala
fa <-> fa.flatMap(pure).handleErrorWith(raiseError)
```

Obviously these two expressions *should* be equivalent, and if we ignore confined runtime scopes such as `unsafeRun`, they *are* indistinguishable. However, the latter expression will guarantee that cancelation is detected if `fa` was indeed canceled, whereas in the former expression, if `fa` is secretly `uncancelable(fa')`, you aren't guaranteed to observe the cancelation.

The one place where this gets *really* strange is `race`, because `race` actually represents a confined evaluation scope similar to `unsafeRun`. However, `race` *already* violates a whole series of laws, which is why we exclude it from the generators, so uh… I guess it's fine?

All in all, this should result in a semantic which is much safer and much more intuitive for users and the ecosystem, with the only serious casualty being our self-respect every time we have to look at what I did in this PR.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -363,7 +363,14 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
    *      between termination conditions
    */
   def guarantee(finalizer: IO[Unit]): IO[A] =
-    guaranteeCase(_ => finalizer)
+    // this is a little faster than the default implementation, which helps Resource
+    IO uncancelable { poll =>
+      val handled = finalizer handleErrorWith { t =>
+        IO.executionContext.flatMap(ec => IO(ec.reportFailure(t)))
+      }
+
+      poll(this).onCancel(finalizer).onError(_ => handled).flatTap(_ => finalizer)
+    }
 
   /**
    * Executes the given `finalizer` when the source is finished,
@@ -387,7 +394,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
    * @see [[guarantee]] for the simpler version
    */
   def guaranteeCase(finalizer: OutcomeIO[A @uncheckedVariance] => IO[Unit]): IO[A] =
-    onCase { case oc => finalizer(oc) }
+    IO.unit.bracketCase(_ => this)((_, oc) => finalizer(oc))
 
   /**
    * Handle any error, potentially recovering from it, by mapping it to another
@@ -415,26 +422,6 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
 
   def onCancel(fin: IO[Unit]): IO[A] =
     IO.OnCancel(this, fin)
-
-  def onCase(pf: PartialFunction[OutcomeIO[A @uncheckedVariance], IO[Unit]]): IO[A] = {
-    def doOutcome(outcome: OutcomeIO[A]): IO[Unit] =
-      pf.lift(outcome)
-        .fold(IO.unit)(_.handleErrorWith { t =>
-          IO.executionContext.flatMap(ec => IO(ec.reportFailure(t)))
-        })
-
-    IO uncancelable { poll =>
-      val base = poll(this)
-      val finalized = pf.lift(Outcome.Canceled()).map(base.onCancel).getOrElse(base)
-
-      finalized.attempt flatMap {
-        case Left(e) =>
-          doOutcome(Outcome.Errored(e)) *> IO.raiseError(e)
-        case Right(a) =>
-          doOutcome(Outcome.Succeeded(IO.pure(a))).as(a)
-      }
-    }
-  }
 
   def onError(f: Throwable => IO[Unit]): IO[A] =
     handleErrorWith(t => f(t).attempt *> IO.raiseError(t))
@@ -628,7 +615,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   def to[F[_]](implicit F: LiftIO[F]): F[A @uncheckedVariance] =
     F.liftIO(this)
 
-  override def toString: String = "IO(...)"
+  // override def toString: String = "IO(...)"
 
   // unsafe stuff
 
@@ -1260,6 +1247,12 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
     def pure[A](x: A): IO[A] =
       IO.pure(x)
 
+    override def guarantee[A](fa: IO[A], fin: IO[Unit]): IO[A] =
+      fa.guarantee(fin)
+
+    override def guaranteeCase[A](fa: IO[A])(fin: OutcomeIO[A] => IO[Unit]): IO[A] =
+      fa.guaranteeCase(fin)
+
     def handleErrorWith[A](fa: IO[A])(f: Throwable => IO[A]): IO[A] =
       fa.handleErrorWith(f)
 
@@ -1278,7 +1271,7 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
       ioa.onCancel(fin)
 
     override def bracketFull[A, B](acquire: Poll[IO] => IO[A])(use: A => IO[B])(
-        release: (A, Outcome[IO, Throwable, B]) => IO[Unit]): IO[B] =
+        release: (A, OutcomeIO[B]) => IO[Unit]): IO[B] =
       IO.bracketFull(acquire)(use)(release)
 
     val monotonic: IO[FiniteDuration] = IO.monotonic
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -615,7 +615,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   def to[F[_]](implicit F: LiftIO[F]): F[A @uncheckedVariance] =
     F.liftIO(this)
 
-  // override def toString: String = "IO(...)"
+  override def toString: String = "IO(...)"
 
   // unsafe stuff
 
--- a/laws/shared/src/main/scala/cats/effect/laws/GenSpawnLaws.scala
+++ b/laws/shared/src/main/scala/cats/effect/laws/GenSpawnLaws.scala
@@ -88,16 +88,24 @@ trait GenSpawnLaws[F[_], E] extends MonadCancelLaws[F, E] with UniqueLaws[F] {
   }
 
   def raceCanceledIdentityLeft[A](fa: F[A]) =
-    F.race(F.canceled, fa) <-> fa.map(_.asRight[Unit])
+    F.race(F.canceled, fa.flatMap(F.pure(_)).handleErrorWith(F.raiseError(_))) <-> fa.map(
+      _.asRight[Unit])
 
   def raceCanceledIdentityRight[A](fa: F[A]) =
-    F.race(fa, F.canceled) <-> fa.map(_.asLeft[Unit])
+    F.race(fa.flatMap(F.pure(_)).handleErrorWith(F.raiseError(_)), F.canceled) <-> fa.map(
+      _.asLeft[Unit])
 
   def raceNeverNoncanceledIdentityLeft[A](fa: F[A]) =
-    F.race(F.never[Unit], fa) <-> F.onCancel(fa.map(_.asRight[Unit]), F.never)
+    F.race(F.never[Unit], fa.flatMap(F.pure(_)).handleErrorWith(F.raiseError(_))) <-> F
+      .onCancel(
+        fa.flatMap(r => F.pure(r.asRight[Unit])).handleErrorWith(F.raiseError(_)),
+        F.never)
 
   def raceNeverNoncanceledIdentityRight[A](fa: F[A]) =
-    F.race(fa, F.never[Unit]) <-> F.onCancel(fa.map(_.asLeft[Unit]), F.never)
+    F.race(fa.flatMap(F.pure(_)).handleErrorWith(F.raiseError(_)), F.never[Unit]) <-> F
+      .onCancel(
+        fa.flatMap(r => F.pure(r.asLeft[Unit])).handleErrorWith(F.raiseError(_)),
+        F.never)
 
   // I really like these laws, since they relate cede to timing, but they're definitely nondeterministic
   /*def raceLeftCedeYields[A](a: A) =
--- a/laws/shared/src/main/scala/cats/effect/laws/MonadCancelLaws.scala
+++ b/laws/shared/src/main/scala/cats/effect/laws/MonadCancelLaws.scala
@@ -76,11 +76,6 @@ trait MonadCancelLaws[F[_], E] extends MonadErrorLaws[F, E] {
     F.onCancel(F.onCancel(F.canceled, fin1), fin2) <->
       F.forceR(F.uncancelable(_ => F.forceR(fin1)(fin2)))(F.canceled)
 
-  // the following set of three laws is formulated in this way to exclude F.never
-  // notably, `F.uncancelable(_ => F.canceled >> fa) >> F.never` results in cancelation
-  // however, `F.uncancelable(_ => fa) >> F.forceR(F.never)(F.canceled)` is nonterminating
-  // these are the semantics we want, but it also means the rewrite rule cannot be generalized
-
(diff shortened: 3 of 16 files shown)


laws/shared/src/main/scala/cats/effect/laws/MonadCancelLaws.scala
```diff
@@ -59,9 +76,14 @@ trait MonadCancelLaws[F[_], E] extends MonadErrorLaws[F, E] {
     F.onCancel(F.onCancel(F.canceled, fin1), fin2) <->
       F.forceR(F.uncancelable(_ => F.forceR(fin1)(fin2)))(F.canceled)
 
-  def uncancelableCanceledAssociatesRightOverFlatMap[A](a: A, f: A => F[Unit]) =
-    F.uncancelable(_ => F.canceled.as(a).flatMap(f)) <->
-      F.forceR(F.uncancelable(_ => f(a)))(F.canceled)
+  // the following set of three laws is formulated in this way to exclude F.never
+  // notably, `F.uncancelable(_ => F.canceled >> fa) >> F.never` results in cancelation
+  // however, `F.uncancelable(_ => fa) >> F.forceR(F.never)(F.canceled)` is nonterminating
+  // these are the semantics we want, but it also means the rewrite rule cannot be generalized
```
djspiewak: Oh whoops this is stale

testkit/shared/src/main/scala/cats/effect/testkit/TestInstances.scala
```diff
@@ -221,8 +221,11 @@ trait TestInstances extends ParallelFGenerators with OutcomeGenerators with Sync
     try {
       var results: Outcome[Option, Throwable, A] = Outcome.Succeeded(None)
 
-      ioa.unsafeRunAsyncOutcome { oc => results = oc.mapK(someK) }(
-        unsafe.IORuntime(ticker.ctx, ticker.ctx, scheduler, () => (), unsafe.IORuntimeConfig()))
+      ioa
+        .flatMap(IO.pure(_))
+        .handleErrorWith(IO.raiseError(_))
+        .unsafeRunAsyncOutcome { oc => results = oc.mapK(someK) }(unsafe
+          .IORuntime(ticker.ctx, ticker.ctx, scheduler, () => (), unsafe.IORuntimeConfig()))
```
vasilmkd, line 228: Looking to learn, is this something users (or even other library authors of libraries dependent on CE3) will ever need to worry about?

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -628,7 +615,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   def to[F[_]](implicit F: LiftIO[F]): F[A @uncheckedVariance] =
     F.liftIO(this)
 
-  override def toString: String = "IO(...)"
+  // override def toString: String = "IO(...)"
```
SystemFw: leftover?

testkit/shared/src/main/scala/cats/effect/testkit/TestInstances.scala
djspiewak: I don't *believe* anyone else will ever need to worry about this. The problem is we're effectively drawing a distinction between "canceled but we haven't seen it yet" vs "canceled and we *have* seen it". The identity `flatMap` + `handleErrorWith` trick turns the former scenario into the latter, and I believe that this is something that is unlikely to be a meaningful thing outside of Cats Effect itself. Everything else downstream is operating in at least a slightly higher level of abstraction, and remember it only takes one `flatMap` to sort this out.

laws/shared/src/main/scala/cats/effect/laws/MonadCancelLaws.scala
```diff
@@ -41,6 +41,23 @@ trait MonadCancelLaws[F[_], E] extends MonadErrorLaws[F, E] {
...
+   * which allows regional composition, since this avoids "gaps" in
+   * otherwise-safe code.
+   *
+   * The argument is that cancellation is a *hint* not a mandate. This
+   * holds for self-cancellation just as much as external-cancellation.
+   * Thus, laws about where the cancellation is visible are always going
+   * to be a bit off.
+   */
+  def onCancelAssociatesOverUncancelableBoundary[A](fa: F[A], fin: F[Unit]) =
+    F.uncancelable(_ => F.onCancel(fa, fin)) <-> F.onCancel(F.uncancelable(_ => fa), fin)
```
SystemFw, line 59: does this hold?
```
uncancelable(poll => fa(poll).onCancel(fin)) <-> uncancelable(poll => fa(poll)).onCancel(fin)
```

SystemFw (approved):

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -416,26 +423,6 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
...
-
-      finalized.attempt flatMap {
-        case Left(e) =>
-          doOutcome(Outcome.Errored(e)) *> IO.raiseError(e)
-        case Right(a) =>
-          doOutcome(Outcome.Succeeded(IO.pure(a))).as(a)
-      }
-    }
-  }
-
```
vasilmkd, line 438: I don't hate this being gone.

vasilmkd (approved):
I asked my questions. LGTM.

core/shared/src/main/scala/cats/effect/IO.scala
djspiewak: Yep

laws/shared/src/main/scala/cats/effect/laws/MonadCancelLaws.scala
djspiewak: Yes. If you see cancelation on the inside, you'll see it on the outside. If you don't see it on the inside, you won't see it on the outside.

core/shared/src/main/scala/cats/effect/IO.scala
djspiewak: Yeah its continued existence was definitely an oversight.

laws/shared/src/main/scala/cats/effect/laws/MonadCancelLaws.scala
SystemFw: right, as expected. Worth documenting it in its full form then, somewhere.

djspiewak: I want to do a full deep-dive discussion in the docs about cancelation in general

RaasAhsan:
Did the following not hold in the current encoding? Particularly thinking about when `fa = canceled`. I think external cancellation is generally a blind spot in the laws, but I think that's been eliminated in the PR?
```scala
uncancelable(_ => fa.onCancel(fin)) <-> uncancelable(_ => fa).onCancel(fin)
```

Also does this mean the laws are violated in the other direction as well?
```scala
uncancelable(_ => fa).onCancel(fin).map(identity) <-/-> uncancelable(_ => fa).onCancel(fin)
```

RaasAhsan (approved):

djspiewak:
> Did the following not hold in the current encoding? Particularly thinking about when fa = canceled. I think external cancellation is generally a blind spot in the laws, but I think that's been eliminated in the PR?

I think external cancelation is still a bit of a blind spot. To answer your question though, that particular law would *not* have held prior to this PR. If `fa = canceled`, the left side would have ignored `fin`, while the right side would have applied it. With this PR, neither side applies it.

> Also does this mean the laws are violated in the other direction as well?

Actually, no! Those two sides are equivalent both before and after this PR.

laws/shared/src/main/scala/cats/effect/laws/GenSpawnLaws.scala
```diff
@@ -88,16 +88,24 @@ trait GenSpawnLaws[F[_], E] extends MonadCancelLaws[F, E] with UniqueLaws[F] {
   }
 
   def raceCanceledIdentityLeft[A](fa: F[A]) =
-    F.race(F.canceled, fa) <-> fa.map(_.asRight[Unit])
+    F.race(F.canceled, fa.flatMap(F.pure(_)).handleErrorWith(F.raiseError(_))) <-> fa.map(
```
wemrysi, line 91: Moving surfacing of cancelation into the argument might make the intent of the laws clearer. Something like
```scala
final class CancelationSurfaced[F[_], A](fa: F[A]) // where the generator for CancellationSurfaced applies the flatMap -> handleError transformatiuon

def raceCanceledIdentifyLeft[A](fa: CancelationSurfaced[F, A]) = ...
```
especially as the particular sequence used to ensure cancelation is visible isn't unique in that function.

Arguably, if you're reasoning over the laws then you're also likely aware of these cancelation semantics and why the "identity" is applied, so maybe it's a feature, but you have to learn to look through it anyhow to see the intent of the law.

The new type would also give a place to document the effect on the laws.

tests/shared/src/test/scala/cats/effect/IOSpec.scala
```diff
@@ -62,13 +65,6 @@ class IOSpec extends IOPlatformSpecification with Discipline with ScalaCheck wit
         fa.flatMap(i => IO.pure(i)) must completeAs(42)
         fa must completeAs(42)
       }
-
-      "preserve monad right identity on uncancelable" in ticked { implicit ticker =>
```
wemrysi, line 66: 😢

wemrysi (commented):
A bit late to the party, but LGTM, glad to have a path forward 👍 Had a comment on how we might think about improving the clarity of the laws in light of the new semantics.

laws/shared/src/main/scala/cats/effect/laws/GenSpawnLaws.scala
djspiewak: Oh this is a really really good idea. It breaks bincompat so we have to do it now if we're going to do it.

tests/shared/src/test/scala/cats/effect/IOSpec.scala
djspiewak: Ikr? :-(

PR #1803, opened 2021-03-21
Do not report the cancellation exception in IOApp

### LINKED ISSUE #1796: Allow suppressing default IOApp cancellation behavior
For apps that serve as e.g. REPLs, being cancelled by `Ctrl+C` doesn't seem like abnormal behavior, so I think there should be a way to opt out of the current behavior of printing a stack trace of `CancellationException`.

I have worked around this by calling the default `main` myself and catching that exception: 

```scala
object Main0 {

  def main(args: Array[String]) =
    try new Main().main(args)
    catch {
      case _: CancellationException => ()
      case e: Throwable            => throw e
    }
}
```

but that obviously has the disadvantage of having to manually wrap the IOApp instance and relies on an implementation detail.

I would suggest something like adding a non-final `def handle(e: Exception): Unit` / `def handle(e: CancellationException): Unit` method in IOApp, throwing by default. Open to other suggestions for solving this :)

vasilmkd:
Resolves #1796.

### DIFF
--- a/core/js/src/main/scala/cats/effect/IOApp.scala
+++ b/core/js/src/main/scala/cats/effect/IOApp.scala
@@ -72,7 +72,14 @@ trait IOApp {
         case Right(_) => sys.error("impossible")
       }
       .unsafeRunAsync({
-        case Left(t) => throw t
+        case Left(t) =>
+          t match {
+            case _: CancellationException =>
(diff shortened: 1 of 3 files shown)


djspiewak (approved):

vasilmkd:
Please don't merge before the CI runs.

kubukoz (approved):

PR #1807, opened 2021-03-22
Document the design of Async#cont

TimWSpence:


docs/typeclasses/async.md
```diff
@@ -102,3 +102,118 @@ for {
...
+
+We could then define `async` in terms of `cont`
+
+```scala
+def async[A](k: (Either[Throwable, A] => Unit) => F[Option[F[Unit]]]): F[A] =
+  cont flatMap {
+    case (cb, fa) => 
+      k(cb).flatMap(_.map(fa.onCancel(_)).getOrElse(fa))
+  }
+```
```
SystemFw, line 184: you probably want to introduce `uncancelable` here first

docs/typeclasses/async.md
```diff
@@ -102,3 +102,118 @@ for {
...
+
+In this case, the error channel is propagated from `k(cb)` as we directly `flatMap`
+on the result.
+
+Fantastic! We're done, right? Well... not quite. The problem is that it is not safe
+to call concurrent operations such as `get.start`. We therefore need to employ
```
SystemFw, line 190: yeah this is hard to explain, but essentially the reason is that `get` represents "the rest of the runloop", which don't want to be resumed concurrently from multiple places

docs/typeclasses/async.md
TimWSpence: Yep thanks 👍 Just copied this from Gitter without thinking too much about it 😂 Although I did also consider removing `uncancelable` from both implementations as it's pretty irrelevant to what we're actually discussing here (could leave a note to say that we're ignoring cancellation). What do you think?

docs/typeclasses/async.md
```diff
@@ -102,3 +102,123 @@ for {
...
+## Here be dragons
+
+`Async` also defines a function called `cont`. This is _not_ intended to be
+called from user code. You also _absolutely do not_ need to understand it in
+order to use `Async`. The design is however very instructive for the curious
+reader. To start with, consider the definition of `async` again
+
+```scala
+trait Async[F[_]] {
+  def async[A](k: (Either[Throwable, A] => Unit) => F[Option[F[Unit]]]): F[A] = {
```
djspiewak: Random open curly? :-)

docs/typeclasses/async.md
```diff
@@ -102,3 +102,123 @@ for {
...
+This looks (vaguely) reasonable. However, there is a subtle problem lurking in
+there -  the use of `flatten`. The problem is that `OptionT` (and similarly
+`EitherT` and `IorT`) has an additional error channel (where `x = 
+delegrate.pure(None)`) that we are forced to discard when mapping to something
+of type `F[Option[F[Unit]]]`.
+
+This in fact breaks several of the `Async` laws:
+
+```scala
+def asyncRightIsSequencedPure[A](a: A, fu: F[Unit]) =
```
djspiewak, line 147: Needs indentation

docs/typeclasses/async.md
```diff
@@ -102,3 +102,123 @@ for {
...
+    F.async[A](k => F.delay(k(Right(a))) >> fu.as(None)) <-> (fu >> F.pure(a))
+
+  def asyncLeftIsSequencedRaiseError[A](e: Throwable, fu: F[Unit]) =
+    F.async[A](k => F.delay(k(Left(e))) >> fu.as(None)) <-> (fu >> F.raiseError(e))
+```
+
+In both cases if we have `fu = OptionT.none[F, A]` (the error channel for `OptionT`) then
+the LHS will suppress the error, whereas the RHS will not.
+
+A possible solution would be to un-CPS the computation by defining
```
djspiewak: Maybe link `CPS` to the [Wikipedia description](https://en.wikipedia.org/wiki/Continuation-passing_style)?

djspiewak (changes requested):
Looks really really nice! A couple changes but otherwise this is awesome

docs/typeclasses/async.md
```diff
@@ -102,3 +102,123 @@ for {
...
+      }
+    }
+  )
+}
+```
+
+This looks (vaguely) reasonable. However, there is a subtle problem lurking in
+there -  the use of `flatten`. The problem is that `OptionT` (and similarly
+`EitherT` and `IorT`) has an additional error channel (where `x = 
+delegrate.pure(None)`) that we are forced to discard when mapping to something
```
catostrophe: ~delegrate~ delegate

docs/typeclasses/async.md
TimWSpence: 🤦 thanks!

TimWSpence:
@djspiewak I _believe_ I've addressed all your comments. Apologies if I've missed any!

PR #1810, opened 2021-03-22
Remove `Resource.liftF`

kubukoz:
It was deprecated in 2.4.0 and there's a working migration for users moving to it - I think it can be removed at this point.


djspiewak:
I tend to agree; anyone else have strong opinions?

vasilmkd (approved):

kubukoz:
3 in favor, I'm doing it :D we can always bring it back.

PR #1811, opened 2021-03-22
Change to rename IOApp#config to runtimeConfig

vlovgr:
This change is because `config` is a fairly common name used in `IOApp`, since it's the point where you normally load and construct 'the world'. We've already seen a conflict in the following pull request: https://github.com/vlovgr/ciris/pull/439. There is also a Gitter discussion about it here: https://gitter.im/typelevel/cats-effect-dev?at=6058ebcc8478e061b9528068.


SystemFw (approved):
I think we should go ahead with this, following the principles of "delightful experience" and "minimise WTFs" for IOApp

djspiewak (approved):
Agree! Great find!

PR #1817, opened 2021-03-24
Backpressure Interface for CE3

etspaceman:
Created per the discussion on https://github.com/typelevel/cats-effect/pull/1695#issuecomment-804263305


std/shared/src/main/scala/cats/effect/std/Backpressure.scala
```diff
@@ -0,0 +1,89 @@
...
+
+  /**
+   * Creates an instance of Backpressure that can be used to rate limit effects
+   * @param strategy strategy to apply for this backpressure instance
+   * @param bound depth of the queue that the backpressure instance should manage
+   * @return a [[Backpressure]] instance
+   */
+  def apply[F[_]](
+      strategy: Strategy,
+      bound: Int
```
vasilmkd, line 57: Since this is passed to the underlying semaphore, which accepts `Long` arguments, should we maybe have this be a `Long` too?

etspaceman: Potentially, though the CE2 version uses an `Int`. Also, `Long` would be quite a large bound for what `Backpressure` is; `Long.MaxValue` would be a heck of a lot of concurrent effects 😅

vasilmkd: I agree. I was just raising a question on the API design. I'm fine either way.

vasilmkd:
Any objections here? I'd be happy to see this merged.

std/shared/src/main/scala/cats/effect/std/Backpressure.scala
```diff
@@ -0,0 +1,89 @@
...
+  )(implicit GC: GenConcurrent[F, Throwable]): F[Backpressure[F]] = {
+    require(bound > 0)
```
rossabaker, line 59: Would it be better to raise than throw here?  I'm surprised to find `require` in the code in a couple places, so there is precedent for this.

std/shared/src/main/scala/cats/effect/std/Backpressure.scala
```diff
@@ -0,0 +1,89 @@
...
+    Semaphore[F](bound.toLong).map(sem =>
+      strategy match {
+        case Strategy.Lossy =>
+          new Backpressure[F] {
+            override def metered[A](f: F[A]): F[Option[A]] =
+              sem
+                .tryAcquire
+                .bracket {
+                  case true => f.map(_.some)
+                  case false => none[A].pure[F]
```
rossabaker, line 69: Could cache this value so it doesn't need to be recreated on every backpressured call, at the expense of always creating it once.

std/shared/src/main/scala/cats/effect/std/Backpressure.scala
vasilmkd: Yeah... these instances exist because of the `GenConcurrent` constraint...

armanbilge (commented):
Cool!! Can we add syntax for this? Like I just proposed for `Supervisor` in #2068.

kubukoz:
@armanbilge if you're up for it, feel free to make the PR :) I'll merge for now to avoid delaying further (we're already past 3.2.0 though)

vasilmkd:
@djspiewak I guess the we're shipping this?

PR #1822, opened 2021-03-26
FiberRef

RaasAhsan:
Follow up to #1393

### DIFF
--- a/core/shared/src/main/scala/cats/effect/FiberRef.scala
+++ b/core/shared/src/main/scala/cats/effect/FiberRef.scala
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2020-2021 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+
+import cats.syntax.all._
+
+trait FiberRef[F[_], A] extends Ref[F, A] {
+
+  /**
+   * Divorces the current reference from parent fiber and
+   * sets a new reference for the duration of `fa` evaluation.
+   */
+  def locally(fa: F[A]): F[A]
+
+}
+
+object FiberRef {
+
+  def apply[A](default: A): IO[FiberRef[IO, A]] =
+    for {
+      ref <- Ref.of[IO, A](default)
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/effect/FiberRef.scala
```diff
@@ -0,0 +1,77 @@
...
+
+import cats.syntax.all._
+
+trait FiberRef[F[_], A] extends Ref[F, A] {
+
+  /**
+   * Divorces the current reference from parent fiber and
+   * sets a new reference for the duration of `fa` evaluation.
+   */
+  def locally(fa: F[A]): F[A]
```
oleg-py, line 27: shouldn't that be `def locally[B](fb: F[B]): F[B]`? with a new method level type-parameter?

kubukoz: should yeah

kubukoz: We've had a similar case in Resource once... [`0bd5660` (#1368)](https://github.com/typelevel/cats-effect/pull/1368/commits/0bd5660eec0432b8141f76af214468c45ffdb744)

core/shared/src/main/scala/cats/effect/FiberRef.scala
```diff
@@ -0,0 +1,77 @@
...
+
+object FiberRef {
+
+  def apply[A](default: A): IO[FiberRef[IO, A]] =
+    for {
+      ref <- Ref.of[IO, A](default)
+      local <- FiberLocal[Ref[IO, A]](ref)
+    } yield new FiberRef[IO, A] {
+      override def locally(fa: IO[A]): IO[A] = {
+        val acquire = local.get.product(Ref.of[IO, A](default)).flatTap {
```
kubukoz, line 39: Random thought, should this use `default` or the current value in the "outer" scope as the initial value instead?

Also, we can save a lot of typing by doing `local.get` first and flatMapping over that, but I don't know about the performance.

core/shared/src/main/scala/cats/effect/FiberRef.scala
kubukoz, line 27: I tried using locals (a copy of this PR) to get a separate scope for every test in a weaver suite, but:

- using a normal `IOLocal` is insufficient, as I want to share the scope between fibers forked _within_ a single test
- using a `FiberRef` like here works as long as I remember to isolate the scope. Problem is, I don't expose the locals because they're hidden behind some interface (i.e. a tagless algebra). A way to do `locally` with _all FiberRefs in existence_ would be really cool. So pretty much the `IO.clearLocals` you mentioned in #1393, but with knowledge of `FiberRef` (just clearing the local doesn't cut it for a Ref because it can be the same value if we're at the start of a test).

> with knowledge of `FiberRef`

this is probably the worst part.

vasilmkd:
Not sure what the status is of this PR. Is there a consensus on how to continue here?

Feel free to remove it from the milestone if you think it's not going to make it.

djspiewak:
Any consensus on this? It seems like `locally` does need some adjustment at the least.

PR #1823, opened 2021-03-26
Scaladoc all the things

TimWSpence:
This is by no means comprehensive or intended to be the final word in documentation but I had a bit of time so thought I'd give us at least an initial version of a bunch of the missing scaladoc


djspiewak (approved):
Thank you!

PR #1826, opened 2021-03-28
Remove unused UnsafeTimer

durban:
`UnsafeTimer` seems unused and unnecessary.


kubukoz (approved):
Good find. Seems like an oversight to me, if it doesn't break the build or docs seems safe to remove, worst case we can add it later and still keep compatibility.

kubukoz:
Thanks!

PR #1829, opened 2021-03-28
Fixed several extremely rare race conditions in runloop resumption

### LINKED ISSUE #1816: Cancelation deadlock in `IO.race`
Just as soon as #1814 was merged, the `AsyncBenchmark.race` benchmark is deadlocking. We had this exact same issue when we had a specialized implementation of `RacePair` and `Race`. Last time we chalked it up to a bad implementation of the specialized `Race` combinator, but now I'm suspecting that something fundamental is buggy in IO.

The only idea I can think of is that somehow fibers which are started but canceled before they've had the chance to run don't propagate their `Canceled` outcome to the outside world. If we look at the fix from today:
```scala
back <- onCancel(
  poll(result.get), {
    for {
      doneA <- deferred[Unit]
      doneB <- deferred[Unit]
      _ <- start(guarantee(fibA.cancel, doneA.complete(()).void)) // <- the guarantee is never called with an `Outcome`
      _ <- start(guarantee(fibB.cancel, doneB.complete(()).void)) // <- or here
      _ <- doneA.get
      _ <- doneB.get
    } yield ()
  }
)
```

I'm sorry if my intuition is wrong.

Wild speculation at 3:30 AM and I hope I'm wrong. What if the finalizer in guarantee is never even registered as a finalizer?

djspiewak:
Fixes #1816 

There's a lot in here, review carefully. All the credit goes to @vasilmkd and @RaasAhsan.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -663,7 +663,7 @@ private final class IOFiber[A](
             state.compareAndSet(ContStateWaiting, ContStateInitial)
             ()
           }
-          finalizers.push(EvalOn(fin, currentCtx))
+          finalizers.push(fin)
           conts.push(OnCancelK)
 
           if (state.compareAndSet(ContStateInitial, ContStateWaiting)) {
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -1068,9 +1068,11 @@ private final class IOFiber[A](
   ///////////////////////////////////////
 
   private[this] def execR(): Unit = {
-    if (resume()) {
-      // println(s"$name: starting at ${Thread.currentThread().getName} + ${suspended.get()}")
+    // println(s"$name: starting at ${Thread.currentThread().getName} + ${suspended.get()}")
 
+    if (canceled) {
+      done(OutcomeCanceled)
+    } else {
       conts = new ByteStack(16)
       conts.push(RunTerminusK)
 
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -566,24 +566,32 @@ private final class IOFiber[A](
                 // `resume()` is a volatile read of `suspended` through which
                 // `wasFinalizing` is published
                 if (finalizing == state.wasFinalizing) {
-                  if (!shouldFinalize()) {
-                    /* we weren't cancelled, so schedule the runloop for execution */
-                    val ec = currentCtx
-                    e match {
-                      case Left(t) =>
-                        resumeTag = AsyncContinueFailedR
-                        objectState.push(t)
-                      case Right(a) =>
-                        resumeTag = AsyncContinueSuccessfulR
-                        objectState.push(a.asInstanceOf[AnyRef])
+                  if (outcome == null) {
+                    if (!shouldFinalize()) {
+                      /* we weren't canceled or completed, so schedule the runloop for execution */
+                      val ec = currentCtx
+                      e match {
+                        case Left(t) =>
+                          resumeTag = AsyncContinueFailedR
+                          objectState.push(t)
+                        case Right(a) =>
+                          resumeTag = AsyncContinueSuccessfulR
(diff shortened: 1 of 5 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
```diff
@@ -104,41 +104,25 @@ trait GenConcurrent[F[_], E] extends GenSpawn[F, E] {
...
+            canA <- start(fibA.cancel)
+            canB <- start(fibB.cancel)
+
+            _ <- canA.join
+            _ <- canB.join
+          } yield ())
+      } yield back match {
+        case Left(oc) => Left((oc, fibB))
+        case Right(oc) => Right((fibA, oc))
+      }
```
vasilmkd, line 125: So much faster-er and lighter-weight-er-er.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -663,7 +663,7 @@ private final class IOFiber[A](
             state.compareAndSet(ContStateWaiting, ContStateInitial)
             ()
           }
-          finalizers.push(EvalOn(fin, currentCtx))
+          finalizers.push(fin)
```
vasilmkd, line 671: Really nice.

vasilmkd:
Doing another pass.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -567,7 +567,7 @@ private final class IOFiber[A](
                 // `wasFinalizing` is published
                 if (finalizing == state.wasFinalizing) {
                   if (!shouldFinalize()) {
```
vasilmkd, line 569: Just to confirm, `ContState` guards against very late, repeated async invocations, while if `ContState` itself is not completed, the fiber can't be, so the `outcome` check is not needed. Is my intuition correct?

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -592,15 +592,20 @@ private final class IOFiber[A](
...
+                 * ownership of the runloop.
                  */
                 loop()
-              } /*
-               * If we are canceled, just die off and let `cancel` or `get` win
-               * the race to `resume` and run the finalisers.
+              }
+
+              /*
+               * If we are canceled or completed or in hte process of finalizing
```
vasilmkd, line 606: Nitpicking: the instead of hte, and two lines below, finalizers instead of finalisers.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -703,7 +708,10 @@ private final class IOFiber[A](
                * finalisers.
                */
               if (resume()) {
-                asyncCancel(null)
+                if (shouldFinalize())
+                  asyncCancel(null)
+                else
+                  suspend()
```
vasilmkd, line 714: Critical fix right here. So glad we found this.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -741,7 +749,7 @@ private final class IOFiber[A](
               }
 
               runLoop(next, nextIteration)
-            } else {
+            } else if (outcome == null) {
```
vasilmkd, line 752: Question for future development. Is there any chance `outcome` can be non-volatile and published in terms of `suspended`?

vasilmkd: I'm guessing no because of the interaction with `CallbackStack`.

vasilmkd:
Running the fix on my machine for some time.

vasilmkd:
Running 3 forks of the fix in parallel.

core/shared/src/main/scala/cats/effect/IOFiber.scala
djspiewak: Also the fact that we set `suspended` to `false` in `done`, meaning that you can't `resume()` if `outcome` is non-null.

vasilmkd (approved):
2 hours running this, with 3 forks in parallel. I'm calling it a success. Monumental effort, I'm absolutely grateful for your time. @RaasAhsan @djspiewak

kubukoz:
I will remember the moment I saw this merge till the day I die.

PR #1830, opened 2021-03-28
Specialize FiberLocal to IOLocal for now

RaasAhsan:


core/shared/src/main/scala/cats/effect/IOLocal.scala
```diff
@@ -16,29 +16,29 @@
 
 package cats.effect
 
-trait FiberLocal[F[_], A] {
+sealed trait IOLocal[A] {
```
kubukoz, line 19: 👍 on sealed

core/shared/src/main/scala/cats/effect/IOLocal.scala
```diff
@@ -16,29 +16,29 @@
...
 
-  def get: F[A]
+  def get: IO[A]
 
-  def set(value: A): F[Unit]
+  def set(value: A): IO[Unit]
 
-  def clear: F[Unit]
+  def clear: IO[Unit]
```
kubukoz: I kinda feel like this should be `reset`. There's still going to be a value, just the default one.

kubukoz (approved):

core/shared/src/main/scala/cats/effect/IOLocal.scala
RaasAhsan: hmm the only weirdness with `reset` is that the underlying map entry is actually removed, but that's an implementation detail. TBH I only added `clear` because it was easy to implement, I wouldn't be averse to leaving it out for now

kubukoz: I don't have a strong opinion either way :) fine as it is

djspiewak (changes requested):
@kubukoz raises a really good point actually. I would rather we change the semantics to `reset` rather than `clear`, because removing the underlying entry means that you can end up getting exceptions out of `get` and it makes the whole thing fairly stateful. I would rather we not have a situation where you can get into an invalid state.

kubukoz:
Oh it actually raises exceptions? Then i definitely want the reset semantics.

kubukoz:
No idea why that was in italics (thanks mobile github).

Actually, there's a .getOrElse. i think we're good

djspiewak:
Ooooh yeah I see the `getOrElse` now. Okay less worried. Still would prefer `reset` as a name.

kubukoz:
:shipit:

djspiewak (approved):

PR #1832, opened 2021-03-28
IOFiber is not really a single writer

vasilmkd:
Hear me out on this one. I may have been too eager to introduce this originally. In the spirit of not making not-fully-informed decisions that we might end up regretting, I'd like the concurrent `lazySet`s to be reverted to `set`s. With `resume()` being a `getAndSet` (and thus `IOFiber` is **not** a single writer) operation with _strictly_ higher memory synchronization semantics than `lazySet` (even though I have never seen this fail, and it also works properly on ARM), I'd like to use concurrent writes which have the same store memory synchronization effects as `getAndSet` and let the VM take care of eliding memory fences where they can be elided.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -628,7 +628,7 @@ private final class IOFiber[A](
                 else {
                   state.result = e
                   // The winner has to publish the result.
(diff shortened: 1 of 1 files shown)


djspiewak:
I think this is actually okay, but in the spirit of being conservative we can back it out and return to it later.

PR #1836, opened 2021-03-29
Add IO#unsafeToFutureCancelable

vasilmkd:
Everything available on `Dispatcher` is now available on `IO` too.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -676,16 +676,45 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
    *
    * @see [[IO.fromFuture]]
    */
-  def unsafeToFuture()(implicit runtime: unsafe.IORuntime): Future[A] = {
+  def unsafeToFuture()(implicit runtime: unsafe.IORuntime): Future[A] =
+    unsafeToFutureCancelable()._1
+
+  /**
+   * Evaluates the effect and produces the result in a `Future`, along with a
+   * cancelation token that can be used to cancel the original effect.
+   *
+   * This is similar to `unsafeRunAsync` in that it evaluates the `IO`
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -676,17 +676,46 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
...
+  def unsafeToFutureCancelable()(
+      implicit runtime: unsafe.IORuntime): (Future[A], () => Future[Unit]) = {
     val p = Promise[A]()
 
-    unsafeRunAsync {
-      case Left(t) => p.failure(t)
-      case Right(a) => p.success(a)
-    }
+    val fiber = unsafeRunFiber(
+      p.failure(new CancellationException("Main fiber was canceled")),
```
kubukoz: Does it have to be the main one? Looks too concrete

vasilmkd: Sorry, for some reason my notifications didn't fire. It's what we use in other places too. Feel free to suggest anything else.

kubukoz: Looks like a good spot for fiber identity One Day... for now I guess I'd just skip `Main` as it could be confusing

vasilmkd: I agree! I'll push a change for this.

kubukoz (approved):

kubukoz (approved):

vasilmkd:
@kubukoz merging this PR now requires a minor version bump. Perhaps we should roll it back. 😔

kubukoz:
I guess it does... @djspiewak how do you feel about entering 3.1.x territory at this point?

djspiewak:
I'm okay with 3.1.x territory. We've got a number of open PRs right now which would require a minor bump. You need to update the `baseVersion` though.

PR #1841, opened 2021-03-29
Add Packlink.com as cats-effect user

juaneto:
We (Packlink) are using cats-effects as part of our stack


vasilmkd (approved):

djspiewak:
Perfect thank you!

PR #1845, opened 2021-03-29
Document how to build the site

### LINKED ISSUE #1820: Document how to build the site/docs
From Gitter:

```
djspiewak (Daniel Spiewak)
@matthughes @vasilmkd @arosien Btw to build the site:

Install Node (14.x ideally)
Install Yarn (whatever…)
git checkout docs
git submodule update
./build.sh host
Your browser will open at the end of this

vasilmkd (Vasil Vasilev)
thank you
djspiewak (Daniel Spiewak)
You can cd into versions/2.x and versions/3.x and do normal submodule things, up to and including making commits and pushing branches and stuff
```

gvolpe:
closes #1820

### DIFF
--- a/docs/typeclasses/monadcancel.md
+++ b/docs/typeclasses/monadcancel.md
@@ -18,7 +18,6 @@ by extending `MonadError` (in Cats) to provide the capability to guarantee the
 running of finalizers when a fiber is canceled. Using it, we can define effects
 which safely acquire and release resources:
 
-<!-- TODO enable mdoc for this -->
 ```scala
 import cats.effect._
 import cats.effect.syntax.all._
@@ -37,7 +36,7 @@ import cats.effect.syntax.all._
 import cats.syntax.all._
 
 def guarded[F[_], R, A, E](
-    s: Semaphore[F], 
+    s: Semaphore[F],
     alloc: F[R])(
     use: R => F[A])(
     release: R => F[Unit])(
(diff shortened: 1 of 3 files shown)


shell.nix
```diff
@@ -0,0 +1,48 @@
+{ jdk ? "jdk15" }:
+
+let
+  config = {
+    packageOverrides = p: rec {
+      java = p.${jdk};
+
+      sbt = p.sbt.overrideAttrs (
```
kubukoz, line 8: why don't we use the default sbt? Any specific version is handled by `build.properties` anyway and the launcher doesn't change enough to break anything

docs/typeclasses/monadcancel.md
```diff
@@ -18,7 +18,6 @@ by extending `MonadError` (in Cats) to provide the capability to guarantee the
 running of finalizers when a fiber is canceled. Using it, we can define effects
 which safely acquire and release resources:
 
-<!-- TODO enable mdoc for this -->
```
kubukoz, line 21: I think it's still a pending todo, can you check if there is (and optionally create) a ticket for this?

shell.nix
gvolpe: I usually keep them in sync. The next version `1.5.x` ships with support for Scala 3, there might be breaking changes. Just a matter of preference I guess :)

kubukoz: I'm using the 1.4.7 launcher I think, and 1.5.0 works flawlessly ;) in fact, I have never had an incompatibility in the launcher <-> version in build. Which is also the reason I have a global sbt installation instead of a nix shell for every project, but YMMV

here I was just aiming for a simpler build, but won't insist

kubukoz (approved):

docs/typeclasses/monadcancel.md
gvolpe: Done: https://github.com/typelevel/cats-effect/issues/1846 Can you add the proper labels?

kubukoz: yep yep

shell.nix
gvolpe: no problems, I can update it, not a big deal :) The important things are the rest of the deps

gvolpe: done :)

djspiewak (approved):
Thanks!

PR #1850, opened 2021-03-29
Add Random materializers for monad transformers

### LINKED ISSUE #1834: Add inductive materializations for Random
We should be able to materialize over monad transformers similarly to `Console`

alexandrustana:
This closes #1834

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Random.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Random.scala
@@ -20,7 +20,15 @@ package std
 
 import cats._
 import cats.syntax.all._
-import cats.data.{EitherT, IorT, Kleisli, OptionT, ReaderWriterStateT, StateT, WriterT}
+import cats.data.{
+  EitherT,
+  IndexedReaderWriterStateT,
+  IndexedStateT,
+  IorT,
+  Kleisli,
+  OptionT,
+  WriterT
+}
 import cats.effect.kernel._
 import scala.util.{Random => SRandom}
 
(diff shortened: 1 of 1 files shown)


std/shared/src/main/scala/cats/effect/std/Random.scala
```diff
@@ -140,6 +205,63 @@ object Random {
...
+   */
+  implicit def catsOptionTRandom[F[_]: Random: Functor]: Random[OptionT[F, *]] =
+    Random[F].mapK(OptionT.liftK)
+
+  /**
+   * [[Random]] instance built for `cats.data.StateT` values initialized with
+   * any `F` data type that also implements `Random`.
+   */
+  implicit def catsStateTRandom[F[_]: Random: Applicative, S]: Random[StateT[F, S, *]] =
+    Random[F].mapK(StateT.liftK)
```
djspiewak: This should probably be `IndexedStateT` just to be slightly more general

std/shared/src/main/scala/cats/effect/std/Random.scala
```diff
@@ -140,6 +205,63 @@ object Random {
...
+   * `F` data type that also implements `Random`.
+   */
+  implicit def catsIorTRandom[F[_]: Random: Functor, L]: Random[IorT[F, L, *]] =
+    Random[F].mapK(IorT.liftK)
+
+  /**
+   * [[Random]] instance built for `cats.data.ReaderWriterStateT` values
+   * initialized with any `F` data type that also implements `Random`.
+   */
+  implicit def catsReaderWriterStateTRandom[
```
djspiewak: Also `Indexed`, ideally

djspiewak (changes requested):
This is great! Just a couple super-minor changes. I'm thinking that the mima compatibility failures are spurious? Need to look at them a little closer and could use a second set of eyes.

wemrysi:
@djspiewak I agree the mima issues are spurious, looks like the PR branch is based on a `series/3.x` pre #1854.

alexandrustana:
@wemrysi you are right it's before #1854. I'll get the fix in when applying the requested changes

std/shared/src/main/scala/cats/effect/std/Random.scala
alexandrustana: I wasn't aware of the Indexed version, should be ok now 🙂

PR #1855, opened 2021-03-29
Updated scastie link to use Cats Effect 3.0.0

djspiewak:


vasilmkd:
I couldn't try the link fast enough. Approved.

PR #1860, opened 2021-03-30
Fix footer broken links

alexandrustana:
It seems that the i18n version of the links wasn't leading anywhere  😅


djspiewak:
Thank you!

PR #1866, opened 2021-03-31
Calculate child mask on demand

vasilmkd:
Right now `childMask` is a field, and it's probably not used as often. It's also very simple to derive, so there's no need for it to take 4 bytes in the object.


djspiewak:
Yeah I saw this the other day. I think it's almost certainly worth calculating it on demand rather than holding onto it. The calculation cost is extremely low while the memory cost is higher than it seems.

PR #1868, opened 2021-03-31
Merge the EndFiber branch with the tags

vasilmkd:
This is an experimental branch but I think the shenanigans I pulled here might point to a couple of bugs. I will comment on the relevant code segments.

The main idea behind this experiment is to get rid of the branching at the top of `runLoop`, particularly the one which checks if the current `IO` is `IOFiberEnd` and whether the runloop should be dropped. I used the tag of `IOFiberEnd` to fuse this branch with the tableswitch. However, this uncovered some potential issues with calling `asyncCancel` multiple times from the runloop itself.

@djspiewak @RaasAhsan please take a look at the comments and the questions I raised. Thank you in advance.

Slight warning, this branch does contain some code from #1869 because I was benchmarking all changes together. I hope this is not too confusing. Only the second commit is relevant for the discussion.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -134,6 +134,9 @@ private final class IOFiber[A](
         case 9 => ()
       }
     } catch {
+      case npe: NullPointerException =>
+        val next = IO.Error(npe)
+        runLoop(next, cancelationCheckThreshold, autoYieldThreshold)
       case t: Throwable =>
         Thread.interrupted()
         currentCtx.reportFailure(t)
@@ -203,24 +206,9 @@ private final class IOFiber[A](
   /* masks encoding: initMask => no masks, ++ => push, -- => pop */
   @tailrec
   private[this] def runLoop(
-      _cur0: IO[Any],
+      cur0: IO[Any],
       cancelationIterations: Int,
       autoCedeIterations: Int): Unit = {
-    /*
-     * `cur` will be set to `EndFiber` when the runloop needs to terminate,
-     * either because the entire IO is done, or because this branch is done
-     * and execution is continuing asynchronously in a different runloop invocation.
-     */
-    if (_cur0 eq IOEndFiber) {
-      return
-    }
-
-    /* Null IO, blow up but keep the failure within IO */
-    val cur0: IO[Any] = if (_cur0 == null) {
-      IO.Error(new NullPointerException())
-    } else {
-      _cur0
-    }
 
     var nextCancelation = cancelationIterations - 1
     var nextAutoCede = autoCedeIterations
@@ -232,7 +220,7 @@ private final class IOFiber[A](
       nextAutoCede -= nextCancelation
     }
 
-    if (shouldFinalize()) {
+    if (!finalizing && shouldFinalize()) {
       asyncCancel(null)
     } else if (autoCedeIterations <= 0) {
       resumeIO = cur0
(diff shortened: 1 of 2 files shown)


vasilmkd:
This doesn't quite work unfortunately.

core/shared/src/main/scala/cats/effect/unsafe/FiberErrorHashtable.scala
```diff
@@ -61,7 +61,7 @@ private[effect] final class FiberErrorHashtable(initialSize: Int) {
         } else {
           idx += 1
           idx &= mask
-          if (idx == init - 1) {
```
vasilmkd: This is definitely a bug, when `init = 0`, `idx` can never be `-1`, so this code spins forever. However! It is very interesting how I discovered this bug.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -813,15 +810,23 @@ private final class IOFiber[A](
...
 
         case 21 =>
           val cur = cur0.asInstanceOf[Local[Any]]
 
           val (nextLocalState, value) = cur.f(localState)
           localState = nextLocalState
-          runLoop(succeeded(value, 0), nextIteration)
+          runLoop(succeeded(value, 0), nextCancelation, nextAutoCede)
+
+        case 22 =>
```
vasilmkd: Because I moved the `IOFiberEnd` check _after_ the cancelation check, it suddenly became possible to call `asyncCancel` twice, even if it was called once already. However! This is not unique to this change. There is a possibility in finalizers which are long `flatMap` chains > 512 iterations, to call `asyncCancel` from the runloop twice.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -207,40 +210,28 @@ private final class IOFiber[A](
...
+    if (cancelationIterations <= 0) {
       // Ensure that we see cancelation.
       readBarrier()
+      nextCancelation = cancelationCheckThreshold
+      // automatic yielding threshold is always a multiple of the cancelation threshold
+      nextAutoCede -= nextCancelation
     }
 
-    if (shouldFinalize()) {
+    if (!finalizing && shouldFinalize()) {
```
vasilmkd, line 223: Because `asyncCancel` here (or in other places for that matter) is _not_ protected by checking `finalizing`. I have a genuine question here. Should `finalizing` be a part of `shouldFinalize()`, i.e.

```scala
private[this] def shouldFinalize(): Boolean =
    canceled && !finalizing && isUnmasked()
```

core/shared/src/main/scala/cats/effect/unsafe/FiberErrorHashtable.scala
vasilmkd: After reading the following comments, this particular comment is clear. `asyncCancel` was being called more than once, and deregistering the error callback for the fiber was being called more than once, searching for a callback that had already been deregistered from the hash table.

vasilmkd:
I rebased this branch to reduce the noise of other commits that have already been merged. The raised issues still stand.

Btw, this PR has a severe negative performance hit on `DeepBindBenchmark.pure`, resulting in 10% reduced performance.

core/shared/src/main/scala/cats/effect/IOFiber.scala
vasilmkd: I've convinced myself this is a non-issue because `isUnmasked()` will always return `false` when running finalizers because the fiber is permanently masked.

core/shared/src/main/scala/cats/effect/unsafe/FiberErrorHashtable.scala
vasilmkd: This was fixed.

core/shared/src/main/scala/cats/effect/IOFiber.scala
vasilmkd: This leads to repeated `done` invocations, so it's just incorrect behavior.

vasilmkd:
The performance is actually significantly worse with this change, around 10% worse on `DeepBindBenchmark.pure`.

PR #1869, opened 2021-03-31
Fix the cancelation check mistake

vasilmkd:
I had made a mistake when refactoring the modulo operation for checking cancelation. After 512 iterations, the check was done on each following iteration. I'm sorry.

In this PR, I have corrected that mistake by using two counters which count down to 0, one for cancelation checks which is reset to `cancelationCheckThreshold` each time a cancelation check is performed and another for automatic ceding, which doesn't change on every iteration but is instead decremented in chunks of `cancelationCheckThreshold` (because the automatic ceding threshold is a multiple of the cancelation check threshold defined by `IORuntimeConfig` by construction, and I mean this literally, by construction of an `IORuntimeConfig` object).

Furthermore, there are fewer field accesses to `cancelationCheckThreshold` and `autoYieldThreshold`, they are now done only on calling the `runLoop` and when doing the cancelation check. This has resulted in some marginal improvements.

Benchmark results:
```
series/3.x
Benchmark                (size)   Mode  Cnt     Score   Error  Units
DeepBindBenchmark.async   10000  thrpt   20  2235.321 ± 5.293  ops/s

Benchmark               (size)   Mode  Cnt     Score    Error  Units
DeepBindBenchmark.pure   10000  thrpt   20  8876.212 ± 31.860  ops/s
```

```
Benchmark                (size)   Mode  Cnt     Score   Error  Units
DeepBindBenchmark.async   10000  thrpt   20  2283.980 ± 8.280  ops/s

Benchmark               (size)   Mode  Cnt     Score    Error  Units
DeepBindBenchmark.pure   10000  thrpt   20  9019.904 ± 20.467  ops/s
```

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -207,7 +207,10 @@ private final class IOFiber[A](
 
   /* masks encoding: initMask => no masks, ++ => push, -- => pop */
   @tailrec
-  private[this] def runLoop(_cur0: IO[Any], iteration: Int): Unit = {
+  private[this] def runLoop(
+      _cur0: IO[Any],
+      cancelationIterations: Int,
+      autoCedeIterations: Int): Unit = {
     /*
      * `cur` will be set to `EndFiber` when the runloop needs to terminate,
      * either because the entire IO is done, or because this branch is done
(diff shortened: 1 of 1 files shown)


vasilmkd:
Do not merge as is, the code will not compile anymore because #1872 was merged.

djspiewak:
OMG LOL

djspiewak:
Okay sorry that's an amazingly hilarious performance jump. Don't worry at all about the bug, it happens.

vasilmkd:
I've heard a saying about some omelette and some eggs.

djspiewak:
Exactly

Issue #1345, opened 2020-10-20
Idea: Managed resource map


djspiewak:
```scala
trait Manager[F[_], K, V] {

  def +=(pair: (K, Resource[F, V])): F[Unit]

  def -=(key: K): F[Unit]

  def apply(key: K): F[Option[V]]

  def join(key: K): F[V]
}

object Manager {
  def apply[F[_]: Sync, K, V]: Resource[F, Manager[F, K, V]] = ???
}
```

The idea here is that it's often useful to have a concurrent map-like data structure which manages `Resource` lifecycles for you, indexed by some key (so you can retrieve the results). These resources can be submitted (via `+=`) and will be evaluated in background, with their value stored in the map once complete. Removing the key from the map would close the resource scope. Similarly, closing the resource scope on the `Manager` would close all constituent `Resource`s and cancel any outstanding computations.

SystemFw:
> and will be evaluated in background, with their value stored in the map once complete.

should this allow cancellation? What happens if the evaluation fails?

kubukoz:
> > and will be evaluated in background, with their value stored in the map once complete.
> 
> should this allow cancellation? What happens if the evaluation fails?

It would only be cancelable by `-=` or closing the Manager, I think. If it fails, I would probably fail `apply` and `join`'s effects. That is, if I understand `join` correctly (apply + getOrElse raise?)

RaasAhsan:
Should failing remove the entry from the map, or should that be pushed back to the user? I could see it trying to acquire the `Resource` again. 

I think this might be a case where we want to designate `V` as being thread-safe, since multiple fibers can trivially access the same resource, or should that be forbidden?

djspiewak:
> should this allow cancellation?

Yes

> What happens if the evaluation fails?

Right now, the interface doesn't have a good channel for that. The key/value pair would simply not be inserted into the map. Users could add their own side-channels by tossing an `onError` on the resource before insertion, or we could do something for them to allow for the reporting. Certainly I think that, if the user `join`s, it should be reported back to them. In fact, perhaps `join` should return an `Outcome`.

> I think this might be a case where we want to designate V as being thread-safe, since multiple fibers can trivially access the same resource, or should that be forbidden?

I would definitely assume `V` has to be thread-safe. This kind of structure doesn't make much sense otherwise, imo.

ChristopherDavenport:
This sounds a lot like keypool - https://github.com/ChristopherDavenport/keypool

[...]

PR #1870, opened 2021-04-01
Shared

RaasAhsan:
This came out of an FS2 PR where we ultimately decided to implement another solution, so I'm checking to see if there's any appetite for it here. `Shared` is another resource-oriented concurrent data structure that's used to share a single resource between multiple actors. It is reminiscent of memoization, but the main difference is that this is one-shot; after the last actor releases the resource, you can't acquire it anymore. The use case in FS2 was to transfer a resource that was allocated on one stream to another safely

I'll write some tests if we want to push it forward

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Shared.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Shared.scala
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2020-2021 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


std/shared/src/main/scala/cats/effect/std/Shared.scala
```diff
@@ -0,0 +1,62 @@
...
+object Shared {
+  def allocate[F[_], A](
+      resource: Resource[F, A]
+  )(implicit F: Concurrent[F]): Resource[F, (Shared[F, A], A)] = {
+    final case class State(value: A, finalizer: F[Unit], permits: Int) {
+      def addPermit: State = copy(permits = permits + 1)
+      def releasePermit: State = copy(permits = permits - 1)
+    }
+
+    MonadCancel[Resource[F, *]].uncancelable { poll =>
```
RaasAhsan, line 35: cool opportunity to use `MonadCancel` over `Resource` here

djspiewak:
I think this is cool, but it's also somewhat redundant with #1345 depending on how the latter is implemented, and a full map of resources would be much more general. I think we should see which approach shakes out as the more useful one and move forward from there.

PR #1873, opened 2021-04-01
Updated tutorial cats-effect dependency to 3.0.0

lrodero:
Updated tutorial cats-effect dependency to `3.0.0`

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -42,11 +42,11 @@ running the code snippets in this tutorial it is recommended to use the same
 ```scala
 name := "cats-effect-tutorial"
 
-version := "3.0.0-RC2"
+version := "3.0.0"
 
 scalaVersion := "2.13.4"
 
-libraryDependencies += "org.typelevel" %% "cats-effect" % "3.0.0-RC2" withSources() withJavadoc()
+libraryDependencies += "org.typelevel" %% "cats-effect" % "3.0.0" withSources() withJavadoc()
 
 scalacOptions ++= Seq(
   "-feature",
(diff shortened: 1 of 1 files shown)


vasilmkd:
Can you please change to Scala 2.13.5 and cats-effect 3.0.1? Thank you 😊

lrodero:
Done :slightly_smiling_face:

docs/tutorial.md
```diff
@@ -42,11 +42,11 @@ running the code snippets in this tutorial it is recommended to use the same
 ```scala
 name := "cats-effect-tutorial"
 
-version := "3.0.0-RC2"
+version := "3.0.1"
```
vasilmkd, line 45: Are we using this anywhere? This is a comment on the file, not on this change.

vasilmkd: Oh wait, this is the version string of the sbt project?

djspiewak (approved):

vasilmkd (approved):

docs/tutorial.md
lrodero: It is used in the [repo](https://github.com/lrodero/cats-effect-tutorial/tree/series/3.x) where the tutorial full code is kept. I think it is a good idea to keep the version there. Although, truly, it _should_ always be the same than the CE version being discussed in the tutorial. So I'm also fine if we want to remove it.

djspiewak: I think it's all cool to keep it :-) Makes sense to me

lrodero: Then we are cool :relaxed:

PR #1874, opened 2021-04-01
[Experiment] `Thread.onSpinWait()`

vasilmkd:
I just found out that Java 9+ has a JVM helper method for signaling spin waiting to the runtime.


vasilmkd:
I expected at least some minor performance improvement, but got nothing.

```
series/3.x
[info] Benchmark             (size)   Mode  Cnt      Score     Error  Units
AsyncBenchmark.async     100  thrpt   20  51179.177 ± 250.853  ops/s

This PR
AsyncBenchmark.async     100  thrpt   20  51105.159 ± 187.038  ops/s
```

My verdict on this is: not worth the complexity.

PR #1879, opened 2021-04-03
'Dealing with cancelation' section rewritten

### LINKED ISSUE #1878: Rewrite 'Dealing with cancelation' section of the tutorial
The '_Dealing with cancelation_' section uses semaphores to prevent files to be closed while data is being read/written. Thing is, they don't make sense there as files only will be closed _after_ the transfer function is done (either because is fully run or because it is canceled). Section should be fully rewritten to get rid of the semaphores mention there.

lrodero:
Removed any mention to semaphores in '_Dealing with cancelation_' section.  Small changes across the tutorial according with the new section contents, including updating the related exercise. Also opening and closing of file streams is done inside `IO.blocking` calls.

Modified version of tutorial is available [here](https://lrodero.github.io/cats-effect/docs/tutorial#dealing-with-cancelation).

Closes #1878

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -103,10 +103,15 @@ parameterized by the return type). Now, let's start implementing our function.
 First, we need to open two streams that will read and write file contents.
 
 ### Acquiring and releasing `Resource`s
-We consider opening a stream to be a side-effect action, so we have to
-encapsulate those actions in their own `IO` instances. For this, we will make
-use of cats-effect `Resource`, that allows us to create, use and then
-release resources in an orderly fashion. See this code:
+We consider opening a stream to be a side-effectful action, so we have to
+encapsulate those actions in their own `IO` instances. We can just embed the
+actions by calling `IO(action)`, but when dealing with input/output actions it
+is advised to use instead `IO.blocking(action)`. This way we help cats-effect to
+better plan how to assign threads to actions. We will return to this topic when
+we introduce _fibers_ later on in this tutorial.
+
+Also, we will make use of cats-effect `Resource`. It allows to orderly create,
+use and then release resources. See this code:
 
 ```scala mdoc:compile-only
 import cats.effect.{IO, Resource}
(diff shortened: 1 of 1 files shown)


docs/tutorial.md
```diff
@@ -103,27 +103,32 @@ parameterized by the return type). Now, let's start implementing our function.
 First, we need to open two streams that will read and write file contents.
 
 ### Acquiring and releasing `Resource`s
-We consider opening a stream to be a side-effect action, so we have to
-encapsulate those actions in their own `IO` instances. For this, we will make
-use of cats-effect `Resource`, that allows to orderly create, use and then
-release resources. See this code:
+We consider opening a stream to be a side-effectul action, so we have to
```
djspiewak:
```suggestion
We consider opening a stream to be a side-effectful action, so we have to
```

docs/tutorial.md
```diff
@@ -103,27 +103,32 @@ parameterized by the return type). Now, let's start implementing our function.
...
+use and then release resources. See this code:
 
 ```scala mdoc:compile-only
 import cats.effect.{IO, Resource}
 import java.io._ 
 
 def inputStream(f: File): Resource[IO, FileInputStream] =
   Resource.make {
-    IO(new FileInputStream(f))                         // build
+    IO.blocking(new FileInputStream(f))                         // build
```
djspiewak, line 122: Does this actually block?

docs/tutorial.md
```diff
@@ -307,95 +309,18 @@ cancelation in the next section.
...
-
-And that is it! We are done, now we can create a program that uses this
-`copy` function.
+Thankfully, `Resource` makes dealing with cancelation an easy task. If the `IO`
+inside a `Resource.use` is canceled, the release section of that resource is
+run. In our example this means the input and output streams will be properly
+closed. Also, cats-effect does not cancel code inside `IO.blocking` instances.
+In the case of our `transmit` function this means the execution would be
+interrupted only between two calls to `IO.blocking`. If we wanted the execution
+of the `IO` instance to be abandoned we should use `IO.interruptible` instead.
```
djspiewak: Maybe "abandoned" is the wrong word? It wouldn't be abandoned, since the `Thread#interrupt()` would be triggered and the cancelation would wait for that interruption to fully complete (or for the action itself to complete) before considering itself "canceled". In other words, `interruptible` makes `Thread#interrupt()` behave identically to `Fiber#cancel`, which is basically what you want.

djspiewak (commented):
Some minor comments but otherwise looks very good!

docs/tutorial.md
lrodero: Tbh I am not really sure... Instantiating `java.io.FileInputStream` opens the file, can we safely assume that is a 'fast' operation that will not block?

docs/tutorial.md
lrodero: What about this text:
_If we want the execution of an `IO` instance to be interrupted when canceled, without waiting for it to finish, we must instantiate it using `IO.interruptible`._
?

djspiewak: Sorry for the delay; this wording is great!

lrodero: Text replaced :slightly_smiling_face:

djspiewak:
Sorry for the delay! Thanks for this!

PR #1880, opened 2021-04-04
Use latest JDKs in CI

vasilmkd:
Jabba, the JDK downloader that we depend on through our Github Actions configuration has not been updated in quite some time, and hacks like this need to be done to get a hold of the latests JDK releases in any of the major versions (8, 11, 16, Graal).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,14 +26,14 @@ jobs:
         java:
           - adopt@1.8
           - adopt@1.11
-          - adopt@1.15
+          - adopt@1.16
           - graalvm-ce-java8@21.0
         ci: [ciJVM, ciJS, ciFirefox]
         exclude:
           - ci: ciJS
             java: adopt@1.11
           - ci: ciJS
-            java: adopt@1.15
+            java: adopt@1.16
           - ci: ciJS
             java: graalvm-ce-java8@21.0
           - os: windows-latest
@@ -47,7 +47,7 @@ jobs:
           - ci: ciFirefox
             java: adopt@1.11
           - ci: ciFirefox
-            java: adopt@1.15
+            java: adopt@1.16
           - ci: ciFirefox
             java: graalvm-ce-java8@21.0
           - os: windows-latest
@@ -58,7 +58,7 @@ jobs:
             scala: 2.12.13
           - os: windows-latest
             ci: ciFirefox
-          - java: adopt@1.15
+          - java: adopt@1.16
             os: windows-latest
     runs-on: ${{ matrix.os }}
     steps:
--- a/build.sbt
+++ b/build.sbt
@@ -52,7 +52,7 @@ ThisBuild / crossScalaVersions := Seq("3.0.0-RC2", "3.0.0-RC3", "2.12.13", Scala
 ThisBuild / githubWorkflowTargetBranches := Seq("series/3.x")
 
 val LTSJava = "adopt@1.11"
-val LatestJava = "adopt@1.15"
+val LatestJava = "adopt@1.16"
 val GraalVM8 = "graalvm-ce-java8@21.0"
 
 ThisBuild / githubWorkflowJavaVersions := Seq(ScalaJSJava, LTSJava, LatestJava, GraalVM8)


djspiewak:
@vasilmkd Is this ready?

vasilmkd:
It works, but the GH actions are now ugly, the JDK strings are too long and cut off. Is that an issue?

djspiewak:
> the JDK strings are too long and cut off. Is that an issue?

I'm okay with it as long as we can look at build failures and tell which case we're in. It's unfortunate, but Jabba really needs to be updated.

vasilmkd:
You can inspect the GH action task for setting up Java... not ideal. Maybe close it and we should look for an alternative?

djspiewak:
Yeah this isn't super-great. It's not intolerable though. Let's leave it open and hunt for another option. I think Jabba allows us to define an upstream definitions repo with an environment variable or something?

vasilmkd:
Does coursier support installation of JDKs?

Edit: It's still done through Jabba.

keynmol:
@vasilmkd yep it does: https://get-coursier.io/docs/cli-java.html#managed-jvms

vasilmkd:
@djspiewak Would you accept a PR to `sbt-github-actions` which adds a new key that is a `Map[String, String]` mapping human readable Java strings like `adopt@11` for presentation purposes to the URI of the download. The key would be used in the GH Actions frontend, and the value would be passed to jabba?

larsrh:
Jabba now has support for JDK 16:

```
$ coursier java --available | grep 1.16
openjdk:1.16.0
zulu:1.16.0-0
```

vasilmkd:
I see. But it's not from `AdoptOpenJDK`.

durban:
Apparently Coursier can also install a JVM by using its own index (or probably any index). The necessary option is `--jvm-index cs` (the default is `--jvm-index jabba`).

```
$ cs java --available --jvm-index cs | grep 1.16
adopt:1.16
adopt-debugimage:1.16
adopt-jre:1.16
adopt-testimage:1.16
```

vasilmkd:
Honestly this is what I've been looking for. Do you mind sharing a link to where this is documented? Thank you.

durban:
I don't think it's documented :-) I've found it in the source and experimented: https://github.com/coursier/coursier/blob/master/modules/jvm/src/main/scala/coursier/jvm/JvmIndex.scala. The `--jvm-index` option seems to accept "cs", "jabba", an URL, or I guess mvn coordinates (e.g., "io.get-coursier:jvm-index"). I don't know how the last one works exactly.

djspiewak:
Honestly seems like something we should try to handle in sbt-github-actions. I don't mind using Coursier instead of Jabba directly to handle that installation, it's just going to require some reworking of the core `build` preamble.

vasilmkd:
Are you sure it doesn't require building a new GH action for this? I'd actually be willing to try to experiment with this and write it in Scala.js.

djspiewak:
> Are you sure it doesn't require building a new GH action for this? I'd actually be willing to try to experiment with this and write it in Scala.js.

I mean, if you wanna try then go for it. :-D Everything that you can do in an action you can also do in a bit of bash dropped into   an `exec`. Either way it's possible. :-) Since sbt-github-actions is already automatically generating and managing the preamble, we don't technically need a *second* mechanism of abstraction, though it is arguably cleaner to do so.

vasilmkd:
But we do use the `olafurpg/setup-scala` Action to do it, don't we?

vasilmkd:
Looks like the jabba index has been [refreshed](https://github.com/shyiko/jabba/commit/ecdfcd24785a4ee8bbd4ecade08ca6ecce38b914) to include the latest JDKs.

SethTisue:
(not familiar with what particular needs this repo might have, but:)

can you use `actions/setup-java@v2` instead of setup-scala? all setup-scala gets you, to my knowledge, is nonstandard sbt launcher scripts

setup-java supports things like

```
    - uses: actions/setup-java@v2
      with:
        distribution: adopt
        java-version: ${{matrix.java}}
```

where `matrix.java` can be set to 8, 11, 16, etc.

vasilmkd:
Yeah, I've used `setup-java` in personal projects and ultimately it works the same, by delegating to the `jabba` tool.

We don't really have any special requirements. We "inherit" `setup-scala` because it is hard-coded in `sbt-github-actions`, for better or for worse. :)

vasilmkd:
AdoptOpenJDK 16 is finally available through Jabba. GraalVM 21.1 is also available but fails to install for some reason. I have opened an [issue](https://github.com/shyiko/jabba/issues/791) with Jabba.

vasilmkd:
I'm merging this because it only touches the build and it is important to build and test against the latest supported JDK.

PR #1889, opened 2021-04-06
Add dual-effect Queue constructors, expose methods to change effect type of `SyncRef` and `AsyncDeferred` without `mapK`

bplommer:
The motivation for this is to be able to allocate resources in the form `Resource[F, Thing[G]]` where the same state needs to be accessed in `G` by the `Thing` but in F for the `Resource`.

I'll add tests and typeclass instances if this is approved in principle, and do something similar for Ref.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -222,13 +222,12 @@ lazy val kernel = crossProject(JSPlatform, JVMPlatform)
     libraryDependencies += "org.specs2" %%% "specs2-core" % Specs2Version % Test)
   .settings(dottyLibrarySettings)
   .settings(libraryDependencies += "org.typelevel" %%% "cats-core" % CatsVersion)
-  .jsSettings(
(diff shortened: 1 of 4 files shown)


djspiewak:
Interesting! Will need to think about this a bit… I can see the utility though.

bplommer:
After playing around with this some more, I've added a confirmed use case for this - dual-effect Queue constructors.

bplommer:
Actually, the previous comment is misleading - the transform methods aren't needed for a dual-effect queue constructor. Anyway, I'll leave this PR as is for now and maybe open a separate one to just add the dual-effect constructor.

djspiewak:
Given that we can now easily convert between `SyncIO` and `IO`, is this really still necessary? I'm trying to try to wrap my head around where this benefits us vs something simpler like `mapK`.

PR #1893, opened 2021-04-09
Grammar tweaks

jducoeur:
Deliberately minimal adjustments to the first half of the tutorial, to make the English a little more idiomatic.  There shouldn't be any semantic changes, and I left the fun archaisms (eg, "Mark", "Heed") alone.


wemrysi (approved):
Thank you!

vasilmkd (approved):

PR #1894, opened 2021-04-09
Fix incorrect scheduling of fibers on the work stealing runtime in presence of parasitic execution contexts

vasilmkd:
Thank you @Baccata for the minimized reproduction and thank you @oleg-py for discovering the issue.

Gitter discussion that goes into more detail is available [here](https://gitter.im/typelevel/cats-effect-dev?at=60708532a2ac0d38e7a6e57d).

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -1062,22 +1062,16 @@ private final class IOFiber[A](
   }
 
   private[this] def rescheduleFiber(ec: ExecutionContext)(fiber: IOFiber[_]): Unit = {
-    val thread = Thread.currentThread()
-    if (thread.isInstanceOf[WorkerThread]) {
-      thread.asInstanceOf[WorkerThread].reschedule(fiber)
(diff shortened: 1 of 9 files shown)


core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -1062,22 +1062,16 @@ private final class IOFiber[A](
   }
 
   private[this] def rescheduleFiber(ec: ExecutionContext)(fiber: IOFiber[_]): Unit = {
-    val thread = Thread.currentThread()
-    if (thread.isInstanceOf[WorkerThread]) {
-      thread.asInstanceOf[WorkerThread].reschedule(fiber)
-    } else if (thread.isInstanceOf[HelperThread]) {
-      thread.asInstanceOf[HelperThread].schedule(fiber)
+    if (ec.isInstanceOf[WorkStealingThreadPool]) {
```
wemrysi, line 1065: That was subtle!

wemrysi (approved):
Nice find all, scheduling based on the EC rather than the thread makes sense 👍

PR #1903, opened 2021-04-13
add three methods to Outcome for easy testing

davidabrahams:
When writing unit-tests on Fibers, I wrote an extension method to allow myself to write:

```
assert(fiber.isCancelled)
```

rather than pattern-matching every time. I thought others might find this useful too!

Let me know if anyone has thoughts on the method names here. I wasn't sure about `isSuccess` vs `wasSuccessful` vs `isSucceeded`, etc, so happy to change this.

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Outcome.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Outcome.scala
@@ -45,6 +45,24 @@ sealed trait Outcome[F[_], E, A] extends Product with Serializable {
       case Errored(e) => Errored(e)
       case Succeeded(fa) => Succeeded(f(fa))
     }
+
+  def isSuccess: Boolean = this match {
+    case Canceled() => false
+    case Errored(_) => false
+    case Succeeded(_) => true
+  }
+
+  def isError: Boolean = this match {
+    case Canceled() => false
+    case Errored(_) => true
+    case Succeeded(_) => false
+  }
+
+  def isCancelled: Boolean = this match {
+    case Canceled() => true
+    case Errored(_) => false
+    case Succeeded(_) => false
+  }
 }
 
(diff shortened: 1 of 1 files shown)


kubukoz:
Mild :-1: as you can use `fold` with relatively little code (`fold(false, _=>false, _=>true)` for example), and this increases the API surface, but if others think it'd be good to have it, I'm okay with it :)

davidabrahams:
Fair enough! I found this useful, but I'm happy to keep this out of the cats-effect library to reduce surface area

kernel/shared/src/main/scala/cats/effect/kernel/Outcome.scala
```diff
@@ -45,6 +45,24 @@ sealed trait Outcome[F[_], E, A] extends Product with Serializable {
...
+    case Succeeded(_) => true
+  }
+
+  def isError: Boolean = this match {
+    case Canceled() => false
+    case Errored(_) => true
+    case Succeeded(_) => false
+  }
+
+  def isCancelled: Boolean = this match {
```
djspiewak: Minor note: We generally using the single `l` spelling consistently throughout the library. Also another thing to think about: this is a past-tense conjugation while the other two methods are present-tense. `isCancel` would probably be more consistent. Alternatively, given that the conjugation of the ADT cases themselves is past-tense, perhaps it would be better to have `isSucceeded`, `isErrored`, and `isCanceled`?

djspiewak (changes requested):
I agree that this is easily accomplished via `fold`, but I have no objection to the convenience functions!

kernel/shared/src/main/scala/cats/effect/kernel/Outcome.scala
djspiewak:
```suggestion
  def isCanceled: Boolean = this match {
```

PR #1921, opened 2021-04-21
Document cancelation boundaries

TimWSpence:


TimWSpence:
I was trying to justify 3 (boundary after `poll` is not a cancelation boundary) from the laws but failed to do so. What have I missed?

Also, I wasn't sure if this belonged in scaladoc or mdoc docs or both?

vasilmkd:
I like it a lot.

vasilmkd (approved):

vasilmkd:
I think we'll definitely need some more on the website but this is a great start.

TimWSpence:
> I think we'll definitely need some more on the website but this is a great start.

Yeah, do you think I should directly copy this stuff to the website? Or is it too much detail for that? Still find that balance hard to judge

djspiewak:
I quite like this, though one *downside* to it is the text in question is somewhat `IO`-specific. Other implementations of the typeclasses (notably, Monix) are likely to have different semantics. Would it be better to push this to the website or the `IO` scaladoc rather than putting it in `MonadCancel`? Otherwise looks very good!

TimWSpence:
> I quite like this, though one _downside_ to it is the text in question is somewhat `IO`-specific. Other implementations of the typeclasses (notably, Monix) are likely to have different semantics. Would it be better to push this to the website or the `IO` scaladoc rather than putting it in `MonadCancel`? Otherwise looks very good!

Yeah I was worried about this as well. I'll see if I can find a place to move it to. The uncancelable boundary at the end of `uncancelable` is guaranteed by the laws though, right? Is the uncancelable boudnary at the end of `poll(foo` (assuming cancelation isn't observed during `foo`) also guaranteed by law? I didn't manage to prove it

djspiewak:
> The uncancelable boundary at the end of uncancelable is guaranteed by the laws though, right? 

Meaning that `uncancelable` is guaranteed to *not* check cancelation at the termination? That is in fact guaranteed by laws:

```scala
  def onCancelAssociatesOverUncancelableBoundary[A](fa: F[A], fin: F[Unit]) =
    F.uncancelable(_ => F.onCancel(fa, fin)) <-> F.onCancel(F.uncancelable(_ => fa), fin)
```

> Is the uncancelable boudnary at the end of `poll(foo` (assuming cancelation isn't observed during foo) also guaranteed by law?

This one is a bit trickier. I *think* this is guaranteed by the above law composed with the following:

```scala
  def uncancelablePollIsIdentity[A](fa: F[A]) =
    F.uncancelable(_(fa)) <-> fa
```

It's tricky though. Both are definitely pervasively assumed properties in polymorphic code.

PR #1928, opened 2021-04-24
Try waking up a WorkerThread before spawning a HelperThread

vasilmkd:
At the moment, any `scala.concurrent.blocking` operation is handled too paranoiacally, with a `HelperThread` being spawned almost every time and the local queue of the blocked worker thread completely drained. This is not ideal for the situations where there are blocking operations which occur infrequently. In those situations, it is better to try to wake up another worker thread which would be able to steal the fibers of the blocked thread. That way, no new thread needs to be spawned, which is much faster and safer. If the blocking does propagate to other worker threads, or if a worker thread could not be woken up due to any reason, this degrades to the old implementation where a `HelperThread` is spawned in anticipation.


vasilmkd:
Please do not merge, I discovered an issue with `fs2` that may or may not be related to this change.

djspiewak:
This is really cool!

Issue #1898, opened 2021-04-11
Different exception handling on the work stealing runtime compared to a fixed thread executor


vasilmkd:
The work stealing runtime is _safer_ when it comes to handling exceptions through running `Runnable`s through `EC#execute` than when using a different `Executor`. Throwing exceptions on the worker threads of a fixed thread pool results in the death of those threads, and depending on the configuration of the `ThreadFactory`, those same threads may or may not be replaced by new ones. Furthermore, throwing fatal exceptions in a `Runnable` on any other thread pool _will not_ currently result in a graceful shutdown of an `IOApp`, because we have bespoke logic to handle this as part of the `IOFiber` runloop. Currently, this safety comes at a performance disadvantage for the work stealing runtime by having to suspend every submitted runnable in an `IO.delay`. I think I have a way of equalizing the exception handling between the work stealing thread pool and any other thread pool used as a compute pool for IO _and_ improving the performance of the work stealing runtime as a normal execution context for running `Runnable`s, but this essentially comes at a performance cost of other `ExecutionContext`s used as a compute pool.

Here is a sample application with a fixed thread pool that shows this behavior:
```scala
package cats.effect
package example

import cats.effect.unsafe.IORuntime

object Example extends IOApp.Simple {

  override def runtime: IORuntime = {
    import java.util.concurrent.atomic.AtomicInteger
    import java.util.concurrent.Executors
    import cats.effect.unsafe.IORuntimeConfig
    import cats.effect.unsafe.Scheduler
    import scala.concurrent.ExecutionContext

    val (blocking, blockDown) = {
      val threadCount = new AtomicInteger(0)
      val executor = Executors.newCachedThreadPool { (r: Runnable) =>
        val t = new Thread(r)
        t.setName(s"io-blocking-${threadCount.getAndIncrement()}")
        t.setDaemon(true)
        t
      }
      (ExecutionContext.fromExecutor(executor), () => executor.shutdown())
    }

    val (scheduler, schedDown) = {
      val executor = Executors.newSingleThreadScheduledExecutor { r =>
        val t = new Thread(r)
        t.setName("io-scheduler")
        t.setDaemon(true)
        t.setPriority(Thread.MAX_PRIORITY)
        t
      }
      (Scheduler.fromScheduledExecutor(executor), () => executor.shutdown())
    }

    val (compute, compDown) = {
      val threadCount = new AtomicInteger(0)
      val executor = Executors.newFixedThreadPool(
        8,
        { r =>
[...]

PR #1935, opened 2021-05-02
Introduce benchmark of scheduling Runnables on WorkStealingThreadPool

RafalSumislawski:
Many Scala code bases mix pure cats-effect code, with impure code. Such code bases may want to use `IORuntime#compute` as their `ExecutionContext` for scheduling Futures, or Runnables. I found performance of the `WorkStealingThreadPool` in `cats-effect` 3.1.0 to be sub-optimal in such usecases. Therfore I propose introduction of a micro-benchmark which can drive future improvements in this area.

The new micro-benchmark is called mimics `WorkStealingBenchmark.runnableScheduling` and it mimics the `WorkStealingBenchmark.schedulingBenchmark`, except instead of working with `Fibers` it works with `Runnables`.

### Fibers vs Runnables comparision (CE 3.1.0)

```
[info] Benchmark                          (size)   Mode  Cnt   Score   Error    Units
[info] WorkStealingBenchmark.scheduling  1000000  thrpt   20  34.384 ± 0.400  ops/min
34 * 1e10 = 3.4e11
```

```
[info] Benchmark                                  (size)   Mode  Cnt  Score   Error    Units
[info] WorkStealingBenchmark.runnableScheduling  1000000  thrpt   20  9.402 ± 0.031  ops/min
9.4 * 10e7 = 9.4e7
```

Please keep in mind that `scheduling` does `1e10` scheduling operations per `op` and the `runnableScheduling` does `1e7` scheduling operations per `op`. So in the latest stable release (3.1.0), scheduling Runnables is ~3600 times slower than scheduling Fibers. The difference is at least partially due to exceptionally good performance of Fiber scheduling in `WorkStealingThreadPool`. Nonetheless I think it's worth having a look at possible ways of reducing the gap. In fact @vasilmkd already made a huge leap in that direction with (#1913) improving the score in this benchmark 11.5 times.

### Improvements in series/3.x (#1913)

3.1.0
```
[info] Benchmark                                  (size)   Mode  Cnt  Score   Error    Units
[info] WorkStealingBenchmark.runnableScheduling  1000000  thrpt   20  9.402 ± 0.031  ops/min
9.4e7
```

series/3.x (2c00489aa68bb7c1fd9d83f1e7173125a5b01c6e)
```
[info] WorkStealingBenchmark.runnableScheduling  1000000  thrpt   20  108.812 ± 1.164  ops/min
1.1e9 operations
```

### DIFF
--- a/benchmarks/src/main/scala/cats/effect/benchmarks/WorkStealingBenchmark.scala
+++ b/benchmarks/src/main/scala/cats/effect/benchmarks/WorkStealingBenchmark.scala
@@ -118,6 +118,33 @@ class WorkStealingBenchmark {
     allocBenchmark
   }
 
+  def runnableSchedulingBenchmark(ec: ExecutionContext): Unit = {
+    val theSize = 10000
+    val countDown = new java.util.concurrent.CountDownLatch(theSize)
+
+    def run(j: Int): Unit = {
+      ec.execute { () =>
+        if (j > 1000) {
+          countDown.countDown()
+        } else {
+          run(j + 1)
+        }
+      }
+    }
+
+    (0 to theSize).foreach(_ => run(0))
+
+    countDown.await()
+  }
+
+  /**
+   * Demonstrates performance of WorkStealingThreadPool when executing Runnables (that includes Futures).
+   */
+  @Benchmark
+  def runnableScheduling(): Unit = {
+    runnableSchedulingBenchmark(cats.effect.unsafe.implicits.global.compute)
+  }
+
   lazy val manyThreadsRuntime: IORuntime = {
     val (blocking, blockDown) = {
       val threadCount = new AtomicInteger(0)


vasilmkd:
Oh wow... it seems I did something right.

A comment on the difference in performance, each Runnable is currently wrapped in an IOFiber, so this is most likely where most of the performance is lost.

I also opened #1898, which will definitely improve on this, but at the cost of slower scheduling of Runnables on another ExecutionContext which is not our own WorkStealingThreadPool. May I interest you in commenting on that potential change? This benchmark can definitely be used to measure the impact. Thank you so much.

vasilmkd:
Out of curiosity, what machine are you executing this on? I'm interested in the OS, JDK and CPU cores/threads configuration. Thank you. 😄

vasilmkd (approved):

RafalSumislawski:
> Oh wow... it seems I did something right.

Indeed. I noticed the problem with the single lock while profiling http4s and came here to further investigate, just to see that you already fixed it ;).

According to my measurements the wrapping in IOFiber is responsible for roughly half to the CPU-time cost of scheduling Runnables. The other half is still the `ThreadSafeHashtable`. I've got some ideas here, and I'm preparing a PR. I'll also have a look at #1898.

I'm running it on 
Ryzen 5600X (6C12T) at ~3700MHz. 
RAM clocked at 3600MHz. 
VM version: JDK 11.0.11, OpenJDK 64-Bit Server VM, 11.0.11+9-Ubuntu-0ubuntu2.20.04
OS: Ubuntu 20.04.2 LTS

vasilmkd:
That's awesome. I'm curious to see what you come up with.

Unfortunately the biggest problem with `ThreadSafeHashTable` is the fact that we need to traverse it without allocating any objects (think `Iterator` or some sort of `Map.Entry<K, V>` object), in the case that the JVM exits with a fatal error.

The code is intentionally simple, so that we keep a good balance of fairly decent performance (after the improvements) versus code complexity and maintainability.

vasilmkd:
> The other half is still the `ThreadSafeHashtable`.

This reminds me, if we do go for some of the ideas behind #1898, scheduling a `Runnable`, at least on the `WorkStealingThreadPool` will not go through `IO` anymore, completely circumventing the `ThreadSafeHashTable`.

vasilmkd:
Let's start things off, this PR has been marinating for some time now.

PR #1938, opened 2021-05-04
Reimplement how `Runnable` instances are executed on the `WorkStealingThreadPool`

vasilmkd:
Tangentially related to #1898.

@RafalSumislawski if you could run the Runnables scheduling on this branch, it would be great. Thank you in advance. In any case, curious to hear what you think.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
@@ -424,7 +424,10 @@ private[effect] final class WorkStealingThreadPool(
    * Preallocated fiber callback function for transforming
    * [[java.lang.Runnable]] values into [[cats.effect.IOFiber]] instances.
    */
-  private[this] val outcomeToUnit: OutcomeIO[Unit] => Unit = _ => ()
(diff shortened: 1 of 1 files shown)


RafalSumislawski:
This looks awesome performance-wise. It skips the `StripedHashtable` completly and achieves a 66% throughput improvement.

Baseline:
```
[info] WorkStealingBenchmark.runnableScheduling  1000000  thrpt   20  110.278 ± 1.948  ops/min
```

This PR
```
[info] WorkStealingBenchmark.runnableScheduling  1000000  thrpt   20  183.341 ± 0.358  ops/min
```

vasilmkd:
That's the idea, yes. Are you happy with this solution?

Also, as a final compromise on #1937, since this PR circumvents the `ThreadSafeHashtable`, let's go with a 2x CPUs multiplier for the number of tables, that way we get most of the performance and a small sized table for 99% of the `IOApp` users who will never need to use `unsafeRun` methods. What do you say?

@RafalSumislawski

RafalSumislawski:
Yes. I'm happy about this solution. And indeed, with Runnables going around the hash table I care less about the performance of the hash table. Nonetheless let's discuss that 2x multiplier in comments #1937 with some measurements.

djspiewak:
This is a very smart little optimization!

PR #1942, opened 2021-05-05
Welcome cats-effect-sync

### LINKED ISSUE #1861: Move SyncIO to std / separate module
fs2 uses SyncIO to implement pure streams, and since that's part of the core, it drags IO and related things with it. Given quite foundational position of fs2, that makes the idea of "not having c.e.IO on the classpath if you don't want it" that was advertised in several talks go out of the window very quickly.

Generally it seems that SyncIO has been most useful in scenarios where:
- you want a fast, stack-safe Defer/Sync/MonadCancel that you can unsafeRun without exposing, for pure/exceptional computations, maybe with local side-effects, so `Eval` alone doesn't fit, and stacking stuff on top of it is leaky, ugly and slow;
- you want to cheat with dual-effect constructors while keeping the second effect type abstract, e.g. `Deferred.in[SyncIO, F].unsafeRunSync()`;
- you want less powerful type than IO but don't want to use typeclasses for some reason,

all of which are quite different from the scenarios for `IO` (which is to be the main effect type for your app).

I know @SystemFw and @mpilquist think that IO shouldn't have been separated out in the first place, posting it more for the official response™ 😄

vasilmkd:
Resolves #1861.

Maybe the solution was staring us in the face all along.


rossabaker:
Is the idea that fs2 would depend on kernel and sync but not core then?

vasilmkd:
https://github.com/typelevel/fs2/pull/2387

kubukoz:
What are those!!!

kubukoz:
Get hype!

vasilmkd:
As nice as this PR is, it is still a binary incompatible change, it will require people to recompile their projects, even though source compatibility has been preserved. I'm waiting for further guidance. Will people care enough to let this one slide. I really do like the appeal of the `fs2` change personally.

mpilquist:
If we forego the package name change (or only introduce the new package as aliases to original definition), can this be made binary compatible?

vasilmkd:
@mpilquist That is binary compatible, but then `fs2` cannot drop the dependency on `core`.

vasilmkd:
@mpilquist I've incorporated your suggestions, this indeed looks better. Thank you.

kubukoz:
Why 😭😭😭

vasilmkd:
There's a great, productive discussion on the `cats-effect-dev` Discord channel, you're welcome to chime in.

PR #1943, opened 2021-05-06
Hide internal platform specific API

### LINKED ISSUE #1899: `IORuntimeConfigCompanionPlatform` type is visible
We forgot to hide this trait unfortunately.

vasilmkd:
Resolves #1899.


vasilmkd:
This is a really strange change. It turns out it is source _incompatible_ but binary _compatible_ because of Java class visibility 😆 being less granular than Scala.

PR #1945, opened 2021-05-06
IO#syncStep

### LINKED ISSUE #1862: Provide a way to unsafeRun IO synchronously without shifting until the async boundary
I'd like to have something like `def unsafeRunSyncStep(): Either[A, IO[A]]` on IO.

My particular scenario: in my react interop library, I'd like to get initial state for rendering from fs2 Stream as soon as possible as to not require two render passes if the state is available immediately (e.g. it's a constant stream). This is an optional optimization, but a nice one (e.g. lets library users not care about putting a placeholder for 1 frame if they're just using a `SignallingRef` or two)

This wasn't needed in CE2 IO since every run method blocked until the first boundary, but CE3 IO will forcibly auto-shift to the compute pool, meaning it can't get anything done during a first render call. Pretty sure the dispatcher can't be made helpful there either.

vasilmkd:
Resolves #1862.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -755,6 +755,32 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
     fiber
   }
 
+  def syncStep: SyncIO[Either[IO[A], A]] = {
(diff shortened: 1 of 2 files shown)


vasilmkd:
This is a wrong approach. I will try again.

PR #1947, opened 2021-05-08
`SyncIO#>>` parameter should be by-name

vasilmkd:
Unfortunately, this is a binary incompatible change...


djspiewak:
Ahhhh this is a concerning oversight. We should be able to avoid the binary breakage by overloading though. If we define a `private[effect] def >>(...)` eager version alongside the lazy version, it should satisfying the old binary linkage without polluting the public API.

djspiewak:
Ahhhh this is a concerning oversight. We should be able to avoid the binary breakage by overloading though. If we define a `private[effect] def >>(...)` eager version alongside the lazy version, it should satisfying the old binary linkage without polluting the public API.

vasilmkd:
That's very smart. Recompiled code against the next version will pick up the new signature, right?

djspiewak:
> That's very smart. Recompiled code against the next version will pick up the new signature, right?

Yep! It would only be old previously-linked code. This is kind of a general technique: when you want to break something, hide the old thing as package-private and, due to the granularity of Scala's access modifiers, you're basically getting the same ABI while still hiding the replaced entity from source resolution.

vasilmkd:
Awesome, thank you! I will push out the change shortly.

djspiewak (approved):

rossabaker (approved):

PR #1951, opened 2021-05-09
Add discord badge

paualarco:
It might be a good idea to link people to Discord now that Typelevel has its own server, the way the badge is created is by the server id `632277896739946517`, see the result:
Screen Shot 2021-05-09 at 17 59 39

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Cats Effect
 
-[![Gitter](https://img.shields.io/gitter/room/typelevel/cats-effect.svg)](https://gitter.im/typelevel/cats-effect) [![Latest version](https://index.scala-lang.org/typelevel/cats-effect/cats-effect/latest.svg?color=orange)](https://index.scala-lang.org/typelevel/cats-effect/cats-effect)
+[![Latest version](https://index.scala-lang.org/typelevel/cats-effect/cats-effect/latest.svg?color=orange)](https://index.scala-lang.org/typelevel/cats-effect/cats-effect)
 [![Discord](https://img.shields.io/discord/632277896739946517.svg?label=&logo=discord&logoColor=ffffff&color=404244&labelColor=6A7EC2)](https://discord.gg/vpEv3HJ)
 
 <img align="right" width="256px" height="256px" src="images/cats-effect-logo.png"/>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,7 @@
 # Cats Effect
 
 [![Gitter](https://img.shields.io/gitter/room/typelevel/cats-effect.svg)](https://gitter.im/typelevel/cats-effect) [![Latest version](https://index.scala-lang.org/typelevel/cats-effect/cats-effect/latest.svg?color=orange)](https://index.scala-lang.org/typelevel/cats-effect/cats-effect)
+[![Discord](https://img.shields.io/discord/632277896739946517.svg?label=&logo=discord&logoColor=ffffff&color=404244&labelColor=6A7EC2)](https://discord.gg/vpEv3HJ)
```
djspiewak: I would actually go a step further and remove Gitter altogether!

```suggestion
[![Discord](https://img.shields.io/discord/632277896739946517.svg?label=&logo=discord&logoColor=ffffff&color=404244&labelColor=6A7EC2)](https://discord.gg/vpEv3HJ) [![Latest version](https://index.scala-lang.org/typelevel/cats-effect/cats-effect/latest.svg?color=orange)](https://index.scala-lang.org/typelevel/cats-effect/cats-effect)
```

Also I didn't check; I assume this is a non-expiring direct link to the `#cats-effect` channel?

paualarco: I think it is not posible to link directly into a specific channel in Discord, it will redirect to the main Typelevel server and from there the user will have to go the the channel of its interest.

Also I believe this is a non expiring link, as I mentioned this is using the serverId, which should not change :)

djspiewak: > it will redirect to the main Typelevel server 

Ah interesting! This must be a server setting, perhaps related to the CoC intercept stuff that the server is doing. I know for a fact that it's possible to link to specific channels in Discord invites, but perhaps that just doesn't work in our case. Either way, nice. :-)

vasilmkd (approved):

README.md
paualarco: Wait, I think it is possible, currently the link is incorrect as I just copy pasted it from shields.io. 
Sorry abt that, let me address it...

PR #1952, opened 2021-05-09
Corrects discord link

paualarco:
Corrects the link in the discord badge that points to cats effect channel.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Cats Effect
 
 [![Latest version](https://index.scala-lang.org/typelevel/cats-effect/cats-effect/latest.svg?color=orange)](https://index.scala-lang.org/typelevel/cats-effect/cats-effect)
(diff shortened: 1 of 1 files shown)


paualarco:
Still I think this is not correct as it **links to the server**, not the channel. But the user needs first to go through the invite. And invites are temporary by default unless you change that. You, @djspiewak should be able to do that as administrator by going to the `Channel > Settings > Add new invite`, and from there you will be able to create one that will never expire (as in the image below)
Screen Shot 2021-05-09 at 19 46 40

paualarco:
Surprisingly I also had permissions to do that :D 
This is the right link: https://discord.gg/QNnHKHq5Ts

vasilmkd:
Confirming the link works correctly for me.

vasilmkd (approved):

PR #1958, opened 2021-05-11
IOApp.ResourceApp

RaasAhsan:
Two convenience traits that fully realize the notion that applications are resources. The two variants mirror `IOApp` and `IOApp.Simple` -- `IOApp.ResourceApp` and `IOApp.SimpleResourceApp` respectively. I could use some help with naming; `IOApp.Resource` is somewhat troublesome because it overloads `kernel.Resource` in scope. We could define it with the fully qualified package names though.

I also took the opportunity to split out an `IOAppPlatform` since the shared code is growing a bit


RaasAhsan:
Might have to undo the `Platform`, seems like it's made mima unhappy and I've never seen those issues before :)

RaasAhsan:
Another path we can take here is to create a `ResourceIOApp` or `ResourceApp` and add a `Simple` to the companion object

rossabaker:
Assuming IO is still baked in, I would not call it just `ResourceApp`.  `ResourceIOApp` is fine.

core/shared/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -0,0 +1,43 @@
...
+object IOApp {
+
+  trait Simple extends IOApp {
+    def run: IO[Unit]
+    final def run(args: List[String]): IO[ExitCode] = run.as(ExitCode.Success)
+  }
+
+  trait ResourceApp extends IOApp {
+    def runResource(args: List[String]): Resource[IO, ExitCode]
+    final def run(args: List[String]): IO[ExitCode] = runResource(args).use(IO.pure(_))
```
kubukoz: It feels weird doing `use(IO.pure(_))` when it's usually _the_ antipattern of using Resource.

What's the usecase for this? Most of what I've seen is doing a `Resource[IO, Anything]` and `.useForever`, as is usually the case for HTTP servers.

djspiewak: Yeah I almost always do `useForever` with `Resource`-based `IOApp`s. I think that would be the more useful default

djspiewak:
IMO let's just do `cats.effect.ResourceApp` and `cats.effect.ResourceApp.Simple`, with both basically wrapping `useForever`. For anyone who wants to do `.use(IO.pure(_))`, then can just use `IOApp` directly.

RaasAhsan:
I've actually written several applications recently that didn't call `useForever`, which was the main motivation for this. `useForever` being the default behavior seems rather opaque to me, so I wouldn't be surprised if we have confused users later who are wondering why their application isn't terminating or releasing it's resources after it performs its function. There could be similar confusion for the alternative though

Perhaps we can expose that as another pair of traits:  `ResourceApp.Forever` and `ResourceApp.ForeverSimple` ? I foresee some ambiguity in overloads but I think is a reasonable resolution for this

core/shared/src/main/scala/cats/effect/ResourceApp.scala
```diff
@@ -0,0 +1,52 @@
...
+  }
+}
+
+object ResourceApp {
+  trait Simple extends ResourceApp {
+    def run: Resource[IO, Unit]
+    final def run(args: List[String]): Resource[IO, ExitCode] = run.as(ExitCode.Success)
+  }
+
+  trait Forever { self =>
```
RaasAhsan, line 40: I don't really think it makes sense to have an `ExitCode` here... does it? also, you'll notice it doesn't extend `ResourceApp` for that reason, but also because we can't override `main` since it's declared `final` for now

core/shared/src/main/scala/cats/effect/ResourceApp.scala
```diff
@@ -0,0 +1,52 @@
...
+
+    final def main(args: Array[String]): Unit = {
+      val ioApp = new IOApp {
+        override def run(args: List[String]): IO[ExitCode] =
+          self.run(args).useForever
+      }
+
+      ioApp.main(args)
+    }
+  }
```
RaasAhsan, line 51: Does a `ForeverSimple` or `Forever.Simple` seem useful? It just wouldn't take any args but not sure how appealing that is if `Forever` doesn't need to return an `ExitCode` to begin with

aeons: I guess, if even just for consistency with the other ones. I would certainly use it, since we usually take arguments from env vars.

core/shared/src/main/scala/cats/effect/ResourceApp.scala
kubukoz: yeah, I'm good with Unit / Any

djspiewak (approved):
I like this as-is

djspiewak:
Nice work! I like this addition quite a lot

PR #1965, opened 2021-05-13
Fix unresolved links in 2.x docs

antonkw:


vasilmkd:
@antonkw I remembered how to test this.

In your branch:
```
sbt coreJVM/doc docs/mdoc
```

It seems that now the docs are even more confused. Removing the `.md` is not a solution unfortunately. Can you try again?

vasilmkd (changes requested):
Marking as changes needed because there are still warnings in the docs.

antonkw:
Will check with command you've shared.
I saw that original warnings disappeared, plus generated links started to direct me to right pages instead of 404.

vasilmkd:
It's the same set of commands used in the `build.sh` script. :)

alexandru:
Those links should be somehow checked by the build process. They were specified as `.md` because AFAIK Jekyll was able to do that.

vasilmkd:
@antonkw I want to thank you for your effort. I'm going to close this PR now, because I have identified and fixed the issue in #2005. It was a misconfiguration in the documentation build script and no changes to the links themselves were needed.

PR #1966, opened 2021-05-13
Scala 3.0.0 final [CE2]

larsrh:
sbt-spiewak depends on an outdated sbt-dotty version, which is why it needs to be added here


smarter:
> sbt-spiewak depends on an outdated sbt-dotty version

Could sbt-spiewak be updated to not depend on sbt-dotty anymore? /cc @djspiewak

larsrh:
Yes. I hope I can get around to send a PR but for now it's better to get the releases out.

PR #1967, opened 2021-05-13
Scala 3.0.0 final [CE3]

larsrh:
sbt-spiewak depends on an outdated sbt-dotty version, which is why it needs to be added here


example/test-js.sh
```diff
@@ -12,7 +12,7 @@ sbt ++$1 exampleJS/fastOptJS
 output=$(mktemp)
 expected=$(mktemp)
 
-cd example/js/target/scala-$(echo $1 | sed -E 's/\.[0-9]+$//')/
+cd example/js/target/scala-$(echo $1 | sed -E 's/^(2\.[0-9]+)\.[0-9]+$/$1/')/
```
larsrh: (ノಠ益ಠ)ノ彡┻━┻

PR #1977, opened 2021-05-17
Add code of conduct

kubukoz:
I realized it was missing in series/3.x, so I'm adding one - as suggested by Alex on discord I added @djspiewak too


rossabaker (approved):

vasilmkd (approved):

alexandru:
👍

@kubukoz you should add a `CONTRIBUTING.md` too 🙂

Didn't we have one? Here's [the one from Monix](https://github.com/monix/monix/blob/series/4.x/CONTRIBUTING.md) for inspiration.

alexandru:
Can be a separate PR 🙂

alexandru (approved):

aeons:
Maybe have discord references instead of gitter, if that is the new official communication channel?

djspiewak:
:+1: for Discord references (is that possible?) but overall this is excellent. I'll merge for now so we don't hold it up, and we should come back to it and adjust if we can figure out how to make those direct links.

PR #1980, opened 2021-05-18
Add explicit return type to IO.interruptible and IO.attempt

lorandszakacs:
Did a quick sweep of the repo for other such methods, didn't find any.

Maybe it might be worth adding the [ExplicitResultType](https://scalacenter.github.io/scalafix/docs/rules/ExplicitResultTypes.html) scalafix rule. And have it live under an optional sbt alias so builds don't wind up slowing down because of semanticdb.


vasilmkd (approved):
This is definitely an oversight, thank you.

vasilmkd:
Good, GitHub Actions recovered.

PR #1983, opened 2021-05-19
Replace silenceSystemErr hack with a generated silent exception

### LINKED ISSUE #418: Replace silenceSystemErr hack with a generated silent exception
I didn't know about `silenceSystemErr` when I was working on shims-effect, so I ended up finding a different solution which is, I think, a bit more elegant. Basically, I override the `Arbitrary[Throwable]` which is otherwise brought in from ScalaCheck with the following implementation:

```scala


  private[this] case object StaticException extends Exception {
    override def printStackTrace() = ()   // this is to stop the tests from spewing useless traces
  }

  // override built-in
  private[this] implicit def arbitraryThrowable: Arbitrary[Throwable] =
    Arbitrary(Gen.const(StaticException))
```

And then you don't have to silence anything. If you add this code to the scope in which `Arbitrary[IO[A]]` is defined, you shouldn't have to worry about any spurious exception output. Just a thought anyway.

vasilmkd:
Resolves #418.

What Daniel said in #418.

### DIFF
--- a/core/shared/src/test/scala/cats/effect/CatsEffectSuite.scala
+++ b/core/shared/src/test/scala/cats/effect/CatsEffectSuite.scala
@@ -17,8 +17,7 @@
 package cats.effect
 
 import cats.effect.internals.TestUtils
-import munit.{DisciplineSuite, Location}
-import org.scalacheck.Prop
+import munit.DisciplineSuite
 
 /**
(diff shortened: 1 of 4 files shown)


vasilmkd:
After inspecting the CI logs, the change seems to be working correctly.

vasilmkd:
The one tricky thing is that the `implicit def` needs to remain public, otherwise this stops working.

PR #2015, opened 2021-05-28
Fix Hotswap example in docs

bastewart:
Previously there were a few errors:
 - Missing for-comprehension block meaning the syntax was invalid
 - Overloaded names in scope meaning it would not compile
 - A `Ref` was "subtracted" from an int as the ref was not accessed
 - Not very important but I think the implementation would use the
   `0.log` filename twice!

### DIFF
--- a/docs/std/hotswap.md
+++ b/docs/std/hotswap.md
@@ -31,30 +31,35 @@ A rotating logger would then look something like this:
 def rotating(n: Int): Resource[IO, Logger[IO]] =
   Hotswap.create[IO, File].flatMap { hs =>
     def file(name: String): Resource[IO, File] = ???
-  
+
     Resource.eval {
       for {
         index <- Ref[IO].of(0)
         count <- Ref[IO].of(0)
         //Open the initial log file
         f <- hs.swap(file("0.log"))
-        log <- Ref[IO].of(f)
+        logFile <- Ref[IO].of(f)
       } yield new Logger[IO] {
         def log(msg: String): IO[Unit] =
-          if (msg.length() < n - count)
-            log.get.flatMap { file => write(file, msg) } 
-            _ <- count.update(_ + msg.length())
-          else
-            for {
-              //Reset the log length counter
-              _ <- count.set(msg.length())
-              //Increment the counter for the log file name
-              idx <- index.modify(n => (n+1, n))
-              //Close the old log file and open the new one
-              f <- hs.swap(file(s"${idx}.log"))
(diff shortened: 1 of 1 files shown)


docs/std/hotswap.md
```diff
@@ -31,30 +31,35 @@ A rotating logger would then look something like this:
...
-              //Close the old log file and open the new one
-              f <- hs.swap(file(s"${idx}.log"))
-              _ <- log.set(f)
-              _ <- write(f, msg)
-            } yield ()
+          count.get.flatMap { currentCount =>
+            if (msg.length() < n - currentCount)
+              for {
+                currentFile <- logFile.get
+                _ <- write(currentFile, msg)
```
bastewart, line 49: If someone were to copy and paste this example now the only error would be that this method, `write`, doesn't exist. I stubbed it out with `def write(file: File): IO[Unit] = ???` when checking this compiled. I could add that next to the stubbed `file` method above if anyone wanted?

rossabaker: That sounds wise to me.

bastewart: 👍 

Done in 8420bc5221

rossabaker (approved):
:+1: on green

bastewart:
> 👍 on green

I'm seeing "First-time contributors need a maintainer to approve running workflows." for the checks below. Might the CI need to be manually approved to run?

rossabaker:
Sorry, I approved it once, but it makes us keep approving it.  I guess you'll have to wait until your next PR to mine cryptocoins on GitHub's infra!

bastewart:
I've been rumbled!

docs/std/hotswap.md
```diff
@@ -31,30 +31,36 @@ A rotating logger would then look something like this:
 def rotating(n: Int): Resource[IO, Logger[IO]] =
   Hotswap.create[IO, File].flatMap { hs =>
     def file(name: String): Resource[IO, File] = ???
-  
+    def write(file: File): IO[Unit] = ???
```
rossabaker:
```suggestion
    def write(file: File, msg: String): IO[Unit] = ???
```

rossabaker (commented):
Sorry, found one more thing when I tried in a REPL.  (And some handwaving around what `Logger` is, but that's fine.)

docs/std/hotswap.md
bastewart: My bad, had put the wrong signature in my message and just copied that 🤦

PR #2019, opened 2021-05-31
Documentation of bracket is unclear?

balthz:
Looking at CE 3.1.1's unit test

https://github.com/typelevel/cats-effect/blob/ffcd1dbaf455bc3dc86dff7884375917738b0b08/tests/shared/src/test/scala/cats/effect/IOSpec.scala#L188-L198

I found the documentation of `bracket()` slightly confusing in its use of "signaling" and "side-channel reporting". I'm suggestion a rewording in the hope it makes things clearer. WDYT?


djspiewak (approved):
Excellent change!

vasilmkd (approved):

PR #2027, opened 2021-06-01
Parallel async/await implementation

Baccata:
Figured I'd contribute this, as I foresee people asking for it. Since the current async/await impl relies on a scala 2 macro, it is probably acceptable. 

* This translates things like `parallel { await(io1) + await(io2) }` into `Parallel.parMap2(io1, io2)((x1, x2) => x1 + x2)`
* Protects against using terms that would prevent such a transformation, like `parallel { val x = 1; await(io(x))}`


std/shared/src/main/scala-2/asyncAwait.scala
```diff
@@ -143,6 +164,68 @@ object AsyncAwaitDsl {
...
 
+  def parallelImpl[F[_], T](c: blackbox.Context)(body: c.Expr[T]): c.Expr[F[T]] =
+    parallelImpl0[F, T](c)(body.tree)
+
+  def parallelImpl0[F[_], T](c: blackbox.Context)(body: c.Tree): c.Expr[F[T]] = {
+    import c.universe._
+    if (!c.compilerSettings.contains("-Xasync")) {
+      c.abort(
+        c.macroApplication.pos,
+        "The parallel method requires the compiler option -Xasync (supported only by Scala 2.12.12+ / 2.13.3+)"
```
Baccata, line 175: This is un-necessary, but I figured it'd be good to ensure that both async and parallel methods are subject to the same requirements.

std/shared/src/main/scala-2/asyncAwait.scala
```diff
@@ -143,6 +164,68 @@ object AsyncAwaitDsl {
...
+        tempIdent
+      case (tt, api) => api.default(tt)
+    }
+
+    val (exprs, bindings) = bound.unzip
+    val callback = c.typecheck(q"(..$bindings) => $transformed ")
+    val parMapMethod = TermName("parMap" + exprs.size)
+    val res = if (exprs.size >= 2) {
+      q"""
+        _root_.cats.Parallel.$parMapMethod(..$exprs)($callback)(cats.effect.kernel.instances.spawn.parallelForGenSpawn(${c.prefix}._AsyncInstance))
```
Baccata: In reality, this whole thing could be provided for any Parallel instance. But using the same arguments that @djspiewak put forward when he decided to ruin `cats.Parallel` for me (:trollface:), the notion of parallelism is fundamentally tied to concurrency and it's cats-core fault for choosing a bad name. 

Also, from a UX perspective, I think it make sense to provide both sequential and parallel "async/await" constructs under the same object.

Baccata:
Lowercased the AsyncAwait.scala file here. I hadn't clocked that the rename had not registered in my previous PR, due to case-insensitive nature the macos filesystem.

djspiewak:
> Lowercased the AsyncAwait.scala file here. I hadn't clocked that the rename had not registered in my previous PR, due to case-insensitive nature the macos filesystem.

APFS fixes this! :-D

vasilmkd:
Are you running the case sensitive version? Because it's not the default and I've had this happen to me on vanilla APFS.

std/shared/src/main/scala-2/asyncAwait.scala
```diff
@@ -143,6 +164,71 @@ object AsyncAwaitDsl {
...
+        c.internal.setFlag(tempSym, (1L << 44).asInstanceOf[c.universe.FlagSet])
+        bound.append((fun, c.internal.valDef(tempSym)))
+        tempIdent
+      case (tt, api) => api.default(tt)
+    }
+
+    val (exprs, bindings) = bound.unzip
+    val callback =
+      c.typecheck(
+        q"(..$bindings) => ${c.prefix}._AsyncInstance.delay { val res = $transformed; res }")
```
Baccata: Working around an issue that prevents the writing of `parallel { x += 1 }`

Baccata:
Will move this to `cats-effect-cps`

PR #2040, opened 2021-06-04
Adds documentation for async/await

Baccata:
* Removes AsyncAwait implementation as it was moved to https://github.com/typelevel/cats-effect-cps 
* Amend documentation to point to that (still provide snippets to avoid forcing people to jump) 
* Removes mdoc tags from the snippets as it'd introduce a cyclic dependency (there may be ways to solve it with front-end magic or build magic if docs were packaged downstream in a standalone jar with an `mdoc` modifier or whatever, but don't think it needs to be solved right now) 
* Re-adds docs to sidebar


site-docs/sidebars.json
```diff
@@ -30,7 +30,8 @@
       "std/queue",
       "std/ref",
       "std/resource",
-      "std/supervisor"
+      "std/supervisor",
+      "std/async-await"
```
djspiewak, line 35: Let's hold off on putting it in the sidebar until after it has a release, to avoid confusing people looking at the site

docs/std/async-await.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Sequential async/await
+
+This construct works for any effect type that has an associated [Async](../typeclasses/async.md) instance (including but not limited to `IO`, `Resource`, and any combination of those with `EitherT`, `Kleisli`, ...).
+
+```scala mdoc:compile-only
+import cats.effect.IO
+import scala.concurrent.duration._
+
+object dsl extends cats.effect.std.AsyncAwaitDsl[IO]
```
djspiewak: Can we do an `import` to clean this up a bit? People are going to copy/paste and things should look nice. :-)

docs/std/async-await.md
```diff
@@ -0,0 +1,81 @@
+---
+id: async-await
+title: Async/await
```
djspiewak:
```suggestion
title: Async/Await
```

Also should we say like, "EXPERIMENTAL" or something like that? We really need to flag that this is something that we aren't promising compatibility on just yet.

djspiewak (changes requested):

docs/std/async-await.md
Baccata: cfb0fd6 . Better not use fullcaps as it'd draw attention to that section in the sidebars.

docs/std/async-await.md
Baccata: Yes Daniel, the shed can be blue if you'd like 😛  cfb0fd6

Seriously though, the fully qualified name was a conscious decision to optimise for copy/paste, since the pretty version is not necessarily the best in that regard.

site-docs/sidebars.json
Baccata: cfb0fd6

docs/std/async-await.md
```diff
@@ -0,0 +1,82 @@
...
+```scala
+val program : IO[Int] = for {
+  x1 <- io
+  x2 <- io
+} yield (x1 + x2)
+```
+
+### Known limitations
+
+`await` cannot be called from within local methods or lambdas (which prevents its use in `for` loops (that get translated to a `foreach` call)). This is due to a limitation in the Scala compiler.
```
Baccata: @retronym is this a hard limitation, or something that could be addressed later ? Also, is there any publicly available write-up (even a comment in a PR) that I could link to ? 

I know that Haoyi did manage to enable it in mill's Task macros (which I'm borrowing implementation details from in [this PR](https://github.com/typelevel/cats-effect/pull/2027/files#diff-4b84b4080312c550c4a02469f70a788b7c013d5edad4d7261561859a924a0dd8R170)), but in that context I assume it's always fine because tasks called in `await` cannot depend on anything that was defined in the `async` block. When that guarantee doesn't exist, is there some devilish detail that prevents the `-Xasync` rewrite from allowing calls to `await` within lambdas ?

djspiewak: > When that guarantee doesn't exist, is there some devilish detail that prevents the -Xasync rewrite from allowing calls to await within lambdas ?

How would it rewrite it? Think about something like this:

```scala
val xs: List[Int] = ???
async {
  xs.foreach(i => await(IO.println(i)))
}
```

The "correct" rewriting here based on manual inspection would be the following:

```scala
xs.traverse_(i => IO.println(i))
```

But like… how is `-Xasync` supposed to know that? What if I have some weird type that the compiler doesn't know anything about which exposes a higher-order function, and then I use `await` within a lambda I pass to that function? How would that work?

This is really the benefit to Kotlin's approach to this problem, which lifts it all into the type system and forces declaration at method boundaries.

docs/std/async-await.md
djspiewak: > Seriously though, the fully qualified name was a conscious decision to optimise for copy/paste, since the pretty version is not necessarily the best in that regard.

This is very true, but I'm guessing that anyone who is leaning on copy/paste here probably also has an editor which supports auto-import insertion. :-)

docs/std/async-await.md
Baccata: > The "correct" rewriting here based on manual inspection would be the following:

I mean sure, but if I take your argument further, the whole async blocks should be rewritten into a series of `flatMap` calls, because that's the "correct" way to compose calls sequentially. 

I'm not even sure how the `-XAsync` machinery rewrites the AST, but I guess it implements some kind of coroutine machinery under the hood, a bit like https://github.com/storm-enroute/coroutines used to. If that's the case, I'd like to understand what kind of problems happens when a yielding call happens within the lambda.

Baccata: I assume it's an ownership problem, as it'd be easy for an await to escape the scope of an async (at runtime) without this

site-docs/sidebars.json
vasilmkd: Wasn't there a release yesterday? I think you can add the sidebar link again. :)

vasilmkd: It's already done, sorry.

Baccata: np, should have updated the comment. Here's the commit that re-adds : cb7e0b1

vasilmkd (approved):

docs/std/async-await.md
```diff
@@ -0,0 +1,101 @@
...
+
+```scala
+import cats.effect.IO
+import cats.effect.cps._
+
+import scala.concurrent.duration._
+
+val io = IO.sleep(50.millis).as(1)
+
+val program : IO[Int] = async[IO] { io.await + io.await }
```
djspiewak:
```suggestion
val program: IO[Int] = async[IO] { io.await + io.await }
```

docs/std/async-await.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+Under the hood, the `async` block is rewritten into non-blocking code that calls onto [Dispatcher](./dispatcher.md) every time `await` is encountered. The end result is lifted into the effect type via callback.
+
+Semantically speaking, the `program` value above is equivalent to
+
+```scala
+val program : IO[Int] = for {
```
djspiewak:
```suggestion
val program: IO[Int] = for {
```

docs/std/async-await.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Known limitations
+
+`await` cannot be called from within local methods or lambdas (which prevents its use in `for` loops (that get translated to a `foreach` call)).
+
+```scala
+import cats.effect.IO
+import cats.effect.std.AsyncAwaitDsl
+
+val program : IO[Int] = async[IO] {
```
djspiewak:
```suggestion
val program: IO[Int] = async[IO] {
```

docs/std/async-await.md
```diff
@@ -0,0 +1,101 @@
...
+  var n = 0
+  for (i <- 1 to 3) (n += IO.pure(i).await ) // compile error
```
djspiewak:
```suggestion
  for (i <- 1 to 3) { n += IO.pure(i).await } // compile error
```

docs/std/async-await.md
```diff
@@ -0,0 +1,101 @@
...
+}
+``` 
+
+**However**, it is possible to call `await` within an imperative while loop:
+
+```scala
+import cats.effect.IO
+import cats.effect.cps._
+
+val program : IO[Int] = async[IO]{
```
djspiewak:
```suggestion
val program: IO[Int] = async[IO] {
```

docs/std/async-await.md
```diff
@@ -0,0 +1,101 @@
...
+  var n = 0
+  var i = 1
+
+  while(i <= 3){
```
djspiewak:
```suggestion
  while (i <= 3) {
```

docs/std/async-await.md
```diff
@@ -0,0 +1,101 @@
...
+
+Because it relies on experimental functionality from the compiler, cats-effec-cps ought to be considered experimental until upstream support stabilises, at which point it will be folded into Cats Effect itself.
+
+## Installation
+
+[![cats-effect-cps Scala version support](https://index.scala-lang.org/typelevel/cats-effect-cps/cats-effect-cps/latest-by-scala-version.svg)](https://index.scala-lang.org/typelevel/cats-effect-cps/cats-effect-cps)
+
+
+```scala
+scalacOptions += "-Xasync" // scala 2.12.12+ / 2.13.3+
```
djspiewak:
```suggestion
scalacOptions += "-Xasync" // scala 2.12.12+ / 2.13.3+ (not needed if on Scala 3)
```

docs/std/async-await.md
```diff
@@ -0,0 +1,101 @@
...
+} yield (x1 + x2)
+```
+
+### Known limitations
+
+`await` cannot be called from within local methods or lambdas (which prevents its use in `for` loops (that get translated to a `foreach` call)).
+
+```scala
+import cats.effect.IO
+import cats.effect.std.AsyncAwaitDsl
```
djspiewak:
```suggestion
```

docs/std/async-await.md
```diff
@@ -0,0 +1,101 @@
...
+  var n = 0
+  for (i <- 1 to 3) (n += IO.pure(i).await ) // compile error
+  n
+}
+```
+
+This constraint is implemented in the Scala compiler (not in cats-effect), for good reason : the Scala language does not provide the capability to guarantee that such methods/lambdas will only be called during the runtime lifecycle of the `async` block. These lambdas could indeed be passed to other constructs that would use them asynchronously, thus **escaping** the lifecycle of the `async` block :
+
+```scala
+async {
```
djspiewak:
```suggestion
async[IO] {
```

docs/std/async-await.md
```diff
@@ -0,0 +1,101 @@
...
+  // executes asynchronously on a separate thread, completes after 1 second.
+  scala.concurrent.Future(IO.sleep(1.second).await)
+  // returns instantly, closing the "async" scope
+  true
+}
+``` 
+
+**However**, it is possible to call `await` within an imperative while loop:
```
djspiewak:
```suggestion
**However**, it is possible to call `await` within an imperative `while` loop:
```

djspiewak (changes requested):
Lots of bikeshedding, but otherwise this looks great and it's pretty much ready to go!

PR #2041, opened 2021-06-05
Rewrite and improve `ThreadSafeHashtable`

vasilmkd:
I realized that copying the elements of the old array to the beginning of the new array messes up the hashing.

For this one, I dug deep into my Data Structures university course, I think the hash table is finally what it was honestly supposed to be from day 1.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -1324,6 +1324,9 @@ private final class IOFiber[A](
     currentCtx.reportFailure(t)
     runtime.shutdown()
 
+    // Make sure the shutdown did not interrupt this thread.
+    Thread.interrupted()
+
     var idx = 0
     val tables = runtime.fiberErrorCbs.tables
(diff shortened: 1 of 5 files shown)


vasilmkd:
I still have a couple more changes to make.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -1324,16 +1324,20 @@ private final class IOFiber[A](
     currentCtx.reportFailure(t)
     runtime.shutdown()
 
+    // Make sure the shutdown did not interrupt this thread.
+    Thread.interrupted()
+
```
djspiewak, line 1329: Sanity-check: do we need to re-interrupt the current thread if we clear the bit here?

vasilmkd: What do you mean exactly? I do the interruption at the end of this method.

djspiewak: Oh, gotcha. Okay then!

PR #2043, opened 2021-06-06
Run the CI on arm64

vasilmkd:


vasilmkd:
I'm having issues with my personal debit cards getting recognized on Oracle Cloud. I need to close this PR. We would need to set up a different account/runner.

PR #2046, opened 2021-06-06
Cattify cats-effect perks

Zelenya:


vasilmkd (approved):
This is oh so cool. Thank you!

djspiewak:
This looks super awesome! I think we might need to make the images a bit less padded though. Here's what it currently renders like:

Zelenya:
Recropped

vasilmkd:
I'll grab a screenshot.

vasilmkd:
I think the images are now good, but our text seems to be a bit off. Can we do some CSS magic here?
Screen Shot 2021-06-07 at 09 37 36

djspiewak:
Actually I think that's what it's supposed to look like. Or at the very least, I think it looks pretty solid :-)

PR #2047, opened 2021-06-06
Run mdoc on Scala 3

vasilmkd:
Confirming it works properly in the CI. I verified it by manually inspecting the logs.


djspiewak:
I think we also need a corresponding PR for `docs` which updates the `sbt` incantation

vasilmkd:
Hmmm, not completely sure what you're referring to. Do you mean we should build and distribute the docs with Scala 3? I wasn't aiming for that in this PR, mostly like a small stepping stone by making sure we were ready to make this change.

PR #2048, opened 2021-06-06
Load fields into local variables when using them multiple times

vasilmkd:
A couple of small details I missed yesterday.


djspiewak:
I actually wonder if this makes a difference once C2 has gotten to it?

vasilmkd:
All of `java.util.concurrent` is written this way, so not completely sure.

PR #2051, opened 2021-06-07
Tracing

vasilmkd:
Building on the work of @RaasAhsan and after several rounds of experimentation, I bring you the production ready initial support for tracing on CE3.

Features:
1. Tracing is available for the following `IO` constructors and combinators:
  - `IO.delay`
  - `IO#map`
  - `IO#flatMap`
  - `IO#handleErrorWith`
  - `IO.uncancelable`
  - `IO.blocking`

2. Noticeable performance improvements to _full_ stack tracing. In practice, this is a slow, tracing mode for debugging, unsuitable for use in production, but still results in an improved developer experience.
  - For those interested, stack trace information on the JVM is usually collected by constructing (but not throwing) new `Throwable` instances. This operation collects a stack trace of the current thread in a JVM specific, binary format. The `Throwable#getStackTrace()` call then parses and transforms this binary representation to the format we know and love from Java Exceptions. Both of these operations are costly. In CE2, we used to eagerly call the `Throwable#getStackTrace()` method on every tracing event, but it was in fact unnecessary. Now, the parsing of the binary format is done only on the necessary tracing information and only when this information is actually necessary for displaying, not earlier.

3. For the first time ever, Cats Effect brings tracing with improved application and memory safety on the modern JVM landscape. Tracing information is usually tied to references of specific `java.lang.Class[_]` objects. Keeping hard references to `Class` objects is generally unsafe on the JVM because it prevents the runtime from unloading those classes, should they become unused. This is not only a memory leak, but also a security issue. By using the `java.lang.ClassValue[_]` API, Cats Effect can be notified by the runtime when certain classes are unloaded, which in turn deletes the tracing information associated to them, freeing memory and improving application safety. This change should have an immediate impact on Cats Effect applications running in `sbt`, which uses complex class loading and unloading to control applications when executing `sbt run`. This change also brought performance improvements over the old `ConcurrentHashMap` based caching approach, due to relying on atomic operations in default JVM internal code, which is heavily optimized by the JVM itself. There are concrete plans to continue experimenting and improving upon this implementation.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/tracing/TracingEvent.scala
+++ b/core/shared/src/main/scala/cats/effect/tracing/TracingEvent.scala
@@ -19,5 +19,5 @@ package cats.effect.tracing
 private[effect] sealed trait TracingEvent
 
 private[effect] object TracingEvent {
-  final case class StackTrace(stackTrace: Throwable) extends TracingEvent
+  final class StackTrace extends Throwable with TracingEvent
 }
--- a/core/shared/src/main/scala/cats/effect/tracing/TracingEvent.scala
+++ b/core/shared/src/main/scala/cats/effect/tracing/TracingEvent.scala
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2020-2021 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.tracing
+
+private[effect] sealed trait TracingEvent
+
+private[effect] object TracingEvent {
+  final case class StackTrace(stackTrace: Throwable) extends TracingEvent
+}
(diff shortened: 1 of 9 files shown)


core/shared/src/main/scala/cats/effect/tracing/TracingEvent.scala
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.tracing
+
+private[effect] sealed trait TracingEvent
+
+private[effect] object TracingEvent {
+  final case class StackTrace(stackTrace: Throwable) extends TracingEvent
```
vasilmkd: Is there a reason for allocating these objects? I think we can inline the `Throwable` inside the `RingBuffer` directly, saving on allocations. If more types of `TracingEvent`s are indeed necessary in the future, we can always introduce this type. @djspiewak Thoughts?

vasilmkd: Or another alternative:
```scala
final case class StackTrace() extends Throwable with TracingEvent
```

djspiewak: I think the only reason is just declarativeness. @RaasAhsan expressed a desire to implement things cleanly first, then bring in the performance tricks later. I'm fine with unboxing it in this PR. The version you suggest above seems very good to me, though I would just make it a regular `class` rather than a `case class` since the equality and `hashCode` will be wrong.

djspiewak (approved):
Superb work, both of you! Holding off on the merge pending resolution of @vasilmkd's one comment.

djspiewak:
Closes #975
Closes #1076

core/shared/src/main/scala/cats/effect/tracing/TracingEvent.scala
vasilmkd: Yeah, on second thought that's the way to go, I just jotted that down when I remembered it.

Great, I will make the change later tonight.

vasilmkd:
I successfully forged Raas' author signature. I'm very happy about that. 😄

djspiewak:
> I successfully forged Raas' author signature.

How else shall we harness this power?!

vasilmkd:
Next time I push dubious code, I know what to do.

PR #2063, opened 2021-06-17
Add docs for Semaphore

### LINKED ISSUE #1927: Document Semaphore
I think this one slipped through the cracks.

armanbilge:
Fixes #1927 mostly by way of CE2 copypasta.


docs/std/semaphore.md
```diff
@@ -0,0 +1,81 @@
...
+  def acquire: F[Unit]
+  def release: F[Unit]
+  // ... and more
+}
+```
+
+## Semantic Blocking and Cancellation
+
+Semaphore does what we call "semantic" blocking, meaning that no actual threads are 
+being blocked while waiting to acquire a permit. Blocking acquires are cancelable.
```
armanbilge, line 22: Re cancellation: this is correct right? CE2 was different here.

djspiewak: It is correct! Also blocking acquires were *generally* cancelable in CE2 though, but it sometimes depended on how exactly you did the acquisition. In CE3 they're always cancelable.

djspiewak (changes requested):
Very nice!!! Thank you!

Would you mind adding this to the `sidebar.json` file? That will make it show up on the site.

armanbilge:
Np, this was an easy one :) fixed the sidebar, and also noticed the sidebar TODO. Still quite a bit to do? Adding a page for `IO` would be nice, the lack of docs for `IOLocal` came up on Discord but I wasn't sure where they should go.

djspiewak:
> Np, this was an easy one :) fixed the sidebar, and also noticed the sidebar TODO. Still quite a bit to do? Adding a page for IO would be nice, the lack of docs for IOLocal came up on Discord but I wasn't sure where they should go.

Not sure either; let's just write them first and then place them somewhere! :-)

LOTS of stuff to do. I have a somewhat outdated list in #1715, and more generally in the [📚 docs label](https://github.com/typelevel/cats-effect/labels/%3Abooks%3A%20docs)

PR #2065, opened 2021-06-18
Add tracing event to `IO.async`

vasilmkd:
Add tracing event to `IO.async`.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -1277,6 +1277,21 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
   private[this] val _asyncForIO: kernel.Async[IO] = new kernel.Async[IO]
     with StackSafeMonad[IO] {
 
+    override def async[A](k: (Either[Throwable, A] => Unit) => IO[Option[IO[Unit]]]): IO[A] = {
+      val body = new Cont[IO, A, A] {
+        def apply[G[_]](implicit G: MonadCancel[G, Throwable]) = { (resume, get, lift) =>
+          G.uncancelable { poll =>
+            lift(k(resume)) flatMap {
+              case Some(fin) => G.onCancel(poll(get), lift(fin))
+              case None => poll(get)
+            }
+          }
+        }
+      }
+
+      IOCont(body, Tracing.calculateTracingEvent(k.getClass))
+    }
+
     override def async_[A](k: (Either[Throwable, A] => Unit) => Unit): IO[A] = {
       val body = new Cont[IO, A, A] {
         def apply[G[_]](implicit G: MonadCancel[G, Throwable]) = { (resume, get, lift) =>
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -885,7 +885,7 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
    * depending on the use case
    */
   def cont[K, R](body: Cont[IO, K, R]): IO[R] =
-    IOCont[K, R](body)
+    IOCont[K, R](body, Tracing.calculateTracingEvent(body.getClass))
```
djspiewak, line 907: I think this is actually not going to be that helpful because it's going to trace the `async` implementation (which is generic), and not the specific *call site*. I think we probably need something a bit like what @RaasAhsan did in CE2 to support tracing of `>>` and similar.

djspiewak: Actually I think we can make it even simpler actually. Rather than `IO.async` delegating to `IO.cont`, it can simply create its own `IOCont` based on the `Class` of the registration function. Should solve the problem.

vasilmkd: Let me show you a trace.

vasilmkd:
```scala
package blah

import cats.effect._

object Example extends IOApp.Simple {

  def run: IO[Unit] =
    (IO(println("started")) >> IO.async_[Unit](_ => throw new RuntimeException())).onCancel(IO(println("canceled")))
}
```

```
Exception in thread "main" java.lang.RuntimeException
	at blah.Example$.$anonfun$run$3(Example.scala:24)
	at async_ @ blah.Example$.$anonfun$run$2(Example.scala:24)
	at apply @ blah.Example$.run(Example.scala:24)
	at >> @ blah.Example$.run(Example.scala:24)
	at run$ @ blah.Example$.run(Example.scala:21)
```

vasilmkd: Although you are right that this doesn't really work as the `body.getClass` is always the same key.

vasilmkd:
@djspiewak Please review carefully. I made a change in `cats.effect.kernel.Async` to factor out the common implementation of the `async` `Cont`, to avoid copy-pasting code in `IO.scala`.

core/shared/src/main/scala/cats/effect/IO.scala
vasilmkd: Also, you forget that Raas did write CE3's implementation. 😄

djspiewak:
I think copy/paste is still needed. I believe you forgot to add tracing to `IO.async`. What you did should catch `Async[IO].async`, but not `IO.async` itself.

I'm a little iffy on the `defaultAsyncContBody` thing. It's basically just a tiny bit of DRY but not much more, and we still end up copy/pasting in most cases. We definitely have to `override` `async` within `Async[IO]`, but we can just delegate it to `IO.async`, which is where the only copy/paste will be.

djspiewak:
Oh, I'm dumb. It delegates. I see.

vasilmkd:
Should I revert it and just copy paste the code?

vasilmkd:
I removed the DRY method.

PR #2068, opened 2021-06-20
Add syntax for Supervisor

armanbilge:
I am quite lazy. This PR adds syntax `F[A]#supervise` equivalent to `Supervisor[F]#supervise(fa: F[A])` which is available whenever an implicit `Supervisor[F]` is in scope. Not sure where this would go, so I created a syntax package in `std`.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/syntax/SupervisorSyntax.scala
+++ b/std/shared/src/main/scala/cats/effect/std/syntax/SupervisorSyntax.scala
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2020-2021 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


vasilmkd:
This looks good to me. Should we port it to CE2 as well?

vasilmkd:
Unless there are objections to passing `Supervisor` instances implicitly. Not sure if we have a firm decision on this.

armanbilge:
> Unless there are objections to passing Supervisor instances implicitly. Not sure if we have a firm decision on this.

Discussing on discord, @ChristopherDavenport made this suggestion:

> Or if you are in the mood to live dangerously `import supervisor.syntax._` in which case if the syntax is used it references that supervisors instance.

~Although there may be some trickiness regarding the import.~ Update: see this example
https://scastie.scala-lang.org/3gjiuAaaRoeEclAUuIcMJg

djspiewak:
I'm not really a fan of passing `Supervisor` implicitly, for the same reason I don't think it's a good idea to pass `Dispatcher` implicitly. Also neither of them are really set up for the "summoning" pattern. With that said, that doesn't really prevent the syntax, it would just mean that the caller would need to pass the supervisor explicitly: `fa.supervise(supervisor)`. Is that worthwhile? It does add to the discoverability, at least in theory.

vasilmkd:
Personally, I see `Dispatcher` and `Supervisor` as similar constructs to CE2's `Blocker`, which wasn't passed around implicitly. Maybe we should not go through with this change.

armanbilge:
Ok, sounds like this isn't a great idea after all. I'm happy to adapt my PR for @djspiewak's suggestion `fa.supervise(supervisor)`, otherwise we can close. Thanks for considering it though!

djspiewak:
Closing this for now; I think the revised syntax is worthwhile, but let's get it into its own PR!

PR #2071, opened 2021-06-21
Clarify Dispatcher instantiation strategy in docs.

arosien:


vasilmkd:
Ok, we finally need to rewrite this bit of information. `Dispatcher` now creates more than 1 fiber, the actual count is the number of CPUs (but this is not necessary information to convey to the user), and in a low concurrency environment most of those fibers would be nicely suspended doing nothing.

With that said, it is now equally viable to have a single instance as well as instances "close to the unsafe regions". Obviously, there exists a point where there can be too many unsafe regions in an app, so 1 global dispatcher would be more memory efficient.

vasilmkd:
@djspiewak Thoughts please.

djspiewak:
In general I still think this is pretty good advice. It certainly isn't *needed* for performance (either way works), but it emphasizes that the users can make whichever choice they need, depending on their API and implementation.

PR #2074, opened 2021-06-24
`bracketFull` for transformers

### LINKED ISSUE #2013: Override MonadCancel#bracketFull in monad transformers instances
[Context](https://gitter.im/typelevel/cats-effect-dev?at=60818cf4ae90f3684098c042)

In the following example, only "goodbye" gets printed : 

```scala
import cats.effect.IOApp
import cats.effect.IO
import cats.effect.std._
import cats.effect.kernel.syntax.all._
import cats.data.OptionT

  type Foo[A] = OptionT[IO, A]

  val prog: Foo[Int] = OptionT[IO, Int](IO.pure(None))

  def run: IO[Unit] = {
    prog
      .guarantee(Console[Foo].println("hit guarantee"))
      .value
      .productR(IO.println("goodbye"))
  }
}
```

This is caused by the generic implementation of [bracketFull](https://github.com/typelevel/cats-effect/blob/d9285906a0b8f3c1b902549cae2781ebc4b90ee7/kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala#L371) which doesn't take into consideration the algebraic "unhappy" cases of monad transformers.

vasilmkd:
Fixes #2013.

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
@@ -443,12 +443,22 @@ trait MonadCancel[F[_], E] extends MonadError[F, E] {
   def bracketFull[A, B](acquire: Poll[F] => F[A])(use: A => F[B])(
       release: (A, Outcome[F, E, B]) => F[Unit]): F[B] =
     uncancelable { poll =>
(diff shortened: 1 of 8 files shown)


vasilmkd:
This needs a rework.

vasilmkd:
Cherry picked into #2107.

PR #2083, opened 2021-06-26
Make Scaladoc example compile and modernize it

igstan:
I'm just reading through the source code of the library so I'm fixing such issues as I bump into them. In this case:

  * `release` takes the allocated resource as a param
  * Use `IO.println`
  * Use Cats Effect's `Random`

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -254,18 +254,27 @@ sealed abstract class Resource[F[_], +A] {
    * `both`, for example via `parMapN`:
    *
    * {{{
-   *   def mkResource(name: String) = {
-   *     val acquire =
-   *       IO(scala.util.Random.nextInt(1000).millis).flatMap(IO.sleep) *>
-   *       IO(println(s"Acquiring $$name")).as(name)
+   * import scala.concurrent.duration._
+   * import cats.effect.{IO, Resource}
+   * import cats.effect.std.Random
+   * import cats.syntax.all._
    *
-   *     val release = IO(println(s"Releasing $$name"))
-   *     Resource.make(acquire)(release)
-   *   }
+   * def mkResource(name: String) = {
+   *   val acquire = for {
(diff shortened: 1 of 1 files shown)


vasilmkd:
I like this. Should we also maybe add the required imports at the beginning of the code block? That way the examples can immediately be copy/pasted.

igstan:
@vasilmkd I've thought about that too, but I wasn't sure. Since you're thinking the same, then I'll add them. It certainly causes no issues.

vasilmkd:
@igstan Check the output in the GitHub actions, you can see some small things that you've missed that you can fix. 😄

igstan:
@vasilmkd oops, yes 😅 Fixing now.

vasilmkd:
No worries, I'm pretty sure you can run `doc` in `sbt` to run this check, before pushing (for a faster turnaround).

igstan:
I've amended the commit with a fix for the CI issue, plus the extra imports needed to run the example. This time I ran `+doc` locally, thanks :D

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -254,18 +254,27 @@ sealed abstract class Resource[F[_], +A] {
...
    *
    * {{{
-   *   def mkResource(name: String) = {
-   *     val acquire =
-   *       IO(scala.util.Random.nextInt(1000).millis).flatMap(IO.sleep) *>
-   *       IO(println(s"Acquiring $$name")).as(name)
+   * import scala.concurrent.duration._
+   * import cats.effect.{IO, Resource}
+   * import cats.effect.std.Random
+   * import cats.implicits._
```
vasilmkd:
```suggestion
   * import cats.syntax.all._
```

igstan: I've applied the change by amending the commit locally instead of using GitHub's UI. I didn't want to add an extra commit just for this.

vasilmkd (approved):

vasilmkd:
Looks great.

igstan:
Thank you, both!

PR #2084, opened 2021-06-26
Small, mostly cosmetics changes to Resource.both

igstan:
This is related in a way to my previous PR (https://github.com/typelevel/cats-effect/pull/2083) where I've just fixed the docs for `Resource.both`. I decided to make a new PR for the changes here because they touch on the method's implementation. These changes are not supposed to add or remove anything as far as functionality is concerned, but it seems to me that they make the code a little easier to read and they get rid of a leftover comment that concerns Dotty/Scala 3.

Let me know if this is kind of PR is annoying from a maintainer's point of view, since they might eat up your attention with small, irrelevant stuff.

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -274,12 +274,9 @@ sealed abstract class Resource[F[_], +A] {
     type Update = (F[Unit] => F[Unit]) => F[Unit]
 
     def allocate[C](r: Resource[F, C], storeFinalizer: Update): F[C] =
-      r.fold(
-        _.pure[F],
-        release => storeFinalizer(MonadCancel[F, Throwable].guarantee(_, release))
-      )
+      r.fold(_.pure[F], release => storeFinalizer(F.guarantee(_, release)))
 
-    val bothFinalizers = Ref.of[F, (F[Unit], F[Unit])](().pure[F] -> ().pure[F])
+    val bothFinalizers = F.ref(F.unit -> F.unit)
 
     Resource.make(bothFinalizers)(_.get.flatMap(_.parTupled).void).evalMap { store =>
       val leftStore: Update = f => store.update(_.leftMap(f))
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -279,12 +279,10 @@ sealed abstract class Resource[F[_], +A] {
     val bothFinalizers = F.ref(F.unit -> F.unit)
 
     Resource.make(bothFinalizers)(_.get.flatMap(_.parTupled).void).evalMap { store =>
-      val leftStore: Update = f => store.update(_.leftMap(f))
-      val rightStore: Update =
-        f =>
-          store.update(t => (t._1, f(t._2))) // _.map(f) doesn't work on 0.25.0 for some reason
+      val thisStore: Update = f => store.update(_.bimap(f, identity))
+      val thatStore: Update = f => store.update(_.bimap(identity, f))
 
-      (allocate(this, leftStore), allocate(that, rightStore)).parTupled
+      (allocate(this, thisStore), allocate(that, thatStore)).parTupled
     }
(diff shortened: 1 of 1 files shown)


vasilmkd:
If you're really bugged by the `map` comment, you can fix it by doing:
```scala
val rightStore: Update =
  f => store.update(Functor[(F[Unit], *)].map(_)(f))
```
instead of `bimap`.

If you're interested, this issue arises in Scala 3 because tuples now have a `map` method of their own, and the type signatures don't match.

igstan:
@vasilmkd thanks. I agree, that's a possible solution too. I've actually motivated my decision in the body of the commit message (can be seen if you click the ellipsis button next to the first line of the commit message). Now... I know this is pretty cosmetic, so either way it's fine by me.

vasilmkd:
If that's the case, please change it to my proposed syntax because we use it consistently throughout the codebase in other places. 😄

vasilmkd:
You know what, sorry. Forget about it. I like what you've done. 😃

igstan:
@vasilmkd sure. Any feedback about the other commit, where I'm using `F.ref` and `F.guarantee` instead of the other constructions?

vasilmkd:
I like the changes. I think the code looks cleaner now.

igstan:
Thanks, then I'll fix the formatting issue reported by the CI and it should be ready.

vasilmkd (approved):
We lost a few lines of code and I can't be happier.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -274,20 +274,15 @@ sealed abstract class Resource[F[_], +A] {
...
 
     def allocate[C](r: Resource[F, C], storeFinalizer: Update): F[C] =
-      r.fold(
-        _.pure[F],
-        release => storeFinalizer(MonadCancel[F, Throwable].guarantee(_, release))
-      )
+      r.fold(_.pure[F], release => storeFinalizer(F.guarantee(_, release)))
 
-    val bothFinalizers = Ref.of[F, (F[Unit], F[Unit])](().pure[F] -> ().pure[F])
+    val bothFinalizers = F.ref(F.unit -> F.unit)
```
rossabaker, line 279: Stylistically, I prefer `a -> b` for pairs that are key-value, and `(a, b)` for pairs of arbitrary order.  But it doesn't really matter.

rossabaker (approved):

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
vasilmkd: That's a really interesting point! 😄

igstan: First time I'm hearing about this line of (stylistic) reasoning and it kinda makes sense since there's a direction to that arrow... Here, I just kept the original choice, but in general I've been trying to avoid parentheses if possible.

rossabaker: Yeah, this PR is already a net win.  I don't want to block merging for my idiosyncratic style rule, but just thought I'd share to see whether anyone agrees. :smile:

vasilmkd: @igstan Up to you. Either push the change if you feel like it, or we can merge as is. 😃

igstan: I'm leaning on the side of pushing as is. I may change my mind in the future, but right now I just think that it's easier to visually parse code that uses `->` instead of commas. With commas I have a parsing "bump" in trying to figure out whether the call takes 2 arguments or a 2-tuple.

djspiewak: FWIW, I agree with @rossabaker's idiosyncratic rule and have converged to the exact same guideline over the years.

igstan: Now with two people against it I feel bad about having merged it like that 😄 I'll keep in mind next time.

PR #2093, opened 2021-06-30
Add `unsafeToPromise` and friends to `Dispatcher`

armanbilge:
Implemented in terms of `unsafeToFuture` for backwards-compatibility.

### DIFF
--- a/std/js/src/main/scala/cats/effect/std/DispatcherPlatform.scala
+++ b/std/js/src/main/scala/cats/effect/std/DispatcherPlatform.scala
@@ -16,4 +16,34 @@
 
 package cats.effect.std
 
-private[std] trait DispatcherPlatform[F[_]] { this: Dispatcher[F] => }
+import scala.scalajs.js
+import scala.scalajs.js.JSConverters._
+import scala.concurrent.ExecutionContext.parasitic
+
+private[std] trait DispatcherPlatform[F[_]] { this: Dispatcher[F] =>
+
+  /**
+   * Submits an effect to be executed, returning a `Promise` that holds the
+   * result of its evaluation, along with a cancelation token that can be
+   * used to cancel the original effect.
+   */
+  def unsafeToPromiseCancelable[A](fa: F[A]): (js.Promise[A], () => js.Promise[Unit]) = {
+    val (f, cancel) = unsafeToFutureCancelable(fa)
+    (f.toJSPromise(parasitic), () => cancel().toJSPromise(parasitic))
+  }
+
+  /**
+   * Submits an effect to be executed, returning a `Promise` that holds the
+   * result of its evaluation.
+   */
+  def unsafeToPromise[A](fa: F[A]): js.Promise[A] =
+    unsafeToPromiseCancelable(fa)._1
+
+  /**
+   * Submits an effect to be executed, returning a cancelation token that
+   * can be used to cancel it.
+   */
+  def unsafeRunCancelablePromise[A](fa: F[A]): () => js.Promise[Unit] =
+    unsafeToPromiseCancelable(fa)._2
+
+}
--- a/std/js/src/main/scala/cats/effect/std/DispatcherPlatform.scala
+++ b/std/js/src/main/scala/cats/effect/std/DispatcherPlatform.scala
@@ -16,40 +16,27 @@
 
 package cats.effect.std
 
-import scala.concurrent.ExecutionContext
-
-import scala.scalajs.js
-import scala.scalajs.js.JSConverters._
+import scala.scalajs.js.{|, Function1, JavaScriptException, Promise, Thenable}
 
 private[std] trait DispatcherPlatform[F[_]] { this: Dispatcher[F] =>
 
   /**
    * Submits an effect to be executed, returning a `Promise` that holds the
-   * result of its evaluation, along with a cancelation token that can be
-   * used to cancel the original effect.
-   */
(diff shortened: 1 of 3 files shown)


std/js/src/main/scala/cats/effect/std/DispatcherPlatform.scala
```diff
@@ -16,4 +16,34 @@
...
+private[std] trait DispatcherPlatform[F[_]] { this: Dispatcher[F] =>
+
+  /**
+   * Submits an effect to be executed, returning a `Promise` that holds the
+   * result of its evaluation, along with a cancelation token that can be
+   * used to cancel the original effect.
+   */
+  def unsafeToPromiseCancelable[A](fa: F[A]): (js.Promise[A], () => js.Promise[Unit]) = {
+    val (f, cancel) = unsafeToFutureCancelable(fa)
+    (f.toJSPromise(parasitic), () => cancel().toJSPromise(parasitic))
```
armanbilge: Relying on `ExecutionContext.parasitic` seemed to be the only escape hatch here.

std/js/src/main/scala/cats/effect/std/DispatcherPlatform.scala
```diff
@@ -16,4 +16,34 @@
...
+   * result of its evaluation.
+   */
+  def unsafeToPromise[A](fa: F[A]): js.Promise[A] =
+    unsafeToPromiseCancelable(fa)._1
+
+  /**
+   * Submits an effect to be executed, returning a cancelation token that
+   * can be used to cancel it.
+   */
+  def unsafeRunCancelablePromise[A](fa: F[A]): () => js.Promise[Unit] =
```
armanbilge: Wasn't sure how to name this, so it doesn't conflict with the `Future` variant.

std/js/src/main/scala/cats/effect/std/DispatcherPlatform.scala
armanbilge: Oops, no parasitic on 2.12 🙃 if this is deemed an acceptable solution, I'll follow up with a fix.

vasilmkd:
My thoughts on this.

I would love to have these methods, I'm sure they would be useful to someone.

Unfortunately, `Dispatcher` is a `trait`, which means its abstract API is frozen forever, and what's even more unfortunate (IMO) is that `scala.concurrent.Future` was chosen as the common unsafe API. In hindsight, I would have loved if `unsafeRunAsync` was the abstract method (if there's a reason why this could not ever be the case, I'd like to know it). If we had that, we could be able to implement everything else in terms of it (including JS and JVM interop).

Again, I may be talking nonsense here, I'd love to hear from the other maintainers.

With that said, given that the choices that have been made are permanent, I'd rather not add these methods, strictly because they require an `ExecutionContext` to work. IMO, if a user really needed them, I'm pretty sure they would be able to figure how to wire things correctly. I'm aware that this is not ideal.

However, I would also not oppose adding these methods, if the other maintainers can justify their existence in this form.

vasilmkd (commented):
My thoughts on this.

I would love to have these methods, I'm sure they would be useful to someone.

Unfortunately, `Dispatcher` is a `trait`, which means its abstract API is frozen forever, and what's even more unfortunate (IMO) is that `scala.concurrent.Future` was chosen as the common unsafe API. In hindsight, I would have loved if `unsafeRunAsync` was the abstract method (if there's a reason why this could not ever be the case, I'd like to know it). If we had that, we could be able to implement everything else in terms of it (including JS and JVM interop).

Again, I may be talking nonsense here, I'd love to hear from the other maintainers.

With that said, given that the choices that have been made are permanent, I'd rather not add these methods, strictly because they require an `ExecutionContext` to work. IMO, if a user really needed them, I'm pretty sure they would be able to figure how to wire things correctly. I'm aware that this is not ideal.

However, I would also not oppose adding these methods, if the other maintainers can justify their existence in this form.

armanbilge:
> With that said, given that the choices that have been made are permanent, I'd rather not add these methods, strictly because they require an ExecutionContext to work.

Sadly, I'm inclined to agree. `Dispatcher` by itself just doesn't have the power to pull this off cleanly.

Just to throw out an alternative: what if we implemented like
```scala
def unsafeToPromiseCancelable[A](fa: F[A])(implicit F: Async[F])
```
This would give us the power we need.

vasilmkd:
Can we `override` the `unsafeRunAsync` implementation in the default dispatcher and implement `unsafeToFuture` in terms of it?

djspiewak:
> Can we override the unsafeRunAsync implementation in the default dispatcher and implement unsafeToFuture in terms of it?

I'm trying to decide whether I want to expose it at all. :-D We can always make it `private[std]`.

armanbilge:
I'm glad this is happening! I'm with @vasilmkd, I definitely want to override `unsafeRunAsync`, no reason it has to go through `Future`.

djspiewak:
> no reason it has to go through Future

The main thinking here (not saying it's right, just expounding on the idea) is that `Future` is conceptually sufficient to cover all of these use-cases and semantically higher-level than a raw callback. You can always *recover* the raw callback by using tricks like `parasitic`. If you think about it, most frameworks actually work in terms of `Future` or similar, not in terms of callbacks.

armanbilge:
Ah, that reasoning does seem sound. Tbh I was just thinking performance 😆

djspiewak:
There's a tiny performance difference (one allocation), but it shouldn't be noticeable in 99% of cases. Also the Either is an allocation anyway, so it's not like we would avoid it with the current signatures.

vasilmkd (approved):

PR #2097, opened 2021-07-04
Use automatic unmanaged source directory handling

vasilmkd:


rossabaker (commented):
The net effect of this is we shed some custom SBT code, but have to maintain a second copy of `ResourceLike`?

vasilmkd:
I honestly don't care that much about this, I wanted to make the build as simple as possible and more approachable. Since this is now handled natively by `sbt-crossproject`, I thought why not. It also matches `fs2`.

rossabaker:
It's small enough, I think the new duplication is acceptable.  If it ever goes much further than this, I'd say there's a missing feature in crossproject!

benchmarks/shared/src/main/scala/cats/effect/benchmarks/ECBenchmark.scala
```diff
@@ -40,8 +41,10 @@ class ECBenchmark {
 
   private val ioApp = new IOApp with Run
   private val ioAppCtx = new IOApp.WithContext with Run {
+
+    @nowarn("cat=deprecation")
     protected def executionContextResource: Resource[SyncIO, ExecutionContext] =
-      Resource.eval(SyncIO.pure(ExecutionContext.Implicits.global))
+      Resource.liftF(SyncIO.pure(ExecutionContext.Implicits.global))
```
djspiewak, line 47: Why the change?

vasilmkd: `benchmarkPrev` uses an older version of CE, which has no `eval` defined. Should I update the dependency?

vasilmkd: I think it needs `2.4.x`, and the current dependency is `2.2.x`.

djspiewak: Nah all good

PR #2101, opened 2021-07-05
Override `map2Eval` for `IO`

### LINKED ISSUE #2095: [CE3] Kleisli Stack safety
This code is Ok with CE2, but fail with `StackOverflowError` in CE3.
```scala
import cats.effect.{ExitCode, IO, IOApp}
import cats.syntax.all._
import cats.data.Kleisli

object MainKleisli extends IOApp {

    val N = 10000
    var i = 0

    List.fill(N)(0).traverse_(_ => Kleisli.liftF(IO(i += 1))).run("Go...") *>
      IO(println(s"i: ${i}")) *>
      IO(assert(i === N, s"i: $i != $N")).as(ExitCode.Success)
  }
}
```

vasilmkd:
Resolves #2095.


vasilmkd:
`map2Eval` was overridden in CE2 in this exact way. Is this ok, or should this override live somewhere in the hierarchy? AFAIK, stack safety is required by any `MonadCancel` (and extensions) in the hierarchy?

What about transformers?

rossabaker (approved):

PR #2105, opened 2021-07-07
Add syntax for Supervisor

armanbilge:
Following up on #2068, add syntax for `fa.supervise(supervisor)`

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/syntax/package.scala
+++ b/std/shared/src/main/scala/cats/effect/std/syntax/package.scala
@@ -19,7 +19,7 @@ package cats.effect.std
 package object syntax {
 
   object all extends AllSyntax
-  
+
   object supervisor extends SupervisorSyntax
 
 }
--- a/std/shared/src/main/scala/cats/effect/std/syntax/package.scala
+++ b/std/shared/src/main/scala/cats/effect/std/syntax/package.scala
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2020-2021 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


std/shared/src/main/scala/cats/effect/std/syntax/package.scala
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.std
+
+package object syntax {
```
djspiewak, line 19: I think this also needs an `all` and an `AllSyntax`, following the boilerplate pattern of the other packages. Otherwise, looks good!

djspiewak (changes requested):

std/shared/src/main/scala/cats/effect/std/syntax/package.scala
armanbilge: Definitely. Do we want to make `all` in core extend from this too?

armanbilge: Hmm. In fact, shouldn't this `all` extend from kernel? To minimize imports

djspiewak: I think that `all` in **core** should extend from this for sure. I don't think this should extend from **kernel** though. If you're using **std**, you should be more a la carte. If you're using **core**, you just get the whole thing by default.

armanbilge: Done and done.

djspiewak:
Scalafmt :-)

armanbilge:
This was a weird one! I ran it like 100x, nothing, then tried `sbt clean` and it did its thing.

djspiewak:
`prePR` helps but doesn't cover everything. The scalafmt plugin is super weird. Restarting sbt and then running `prePR` usually does the trick.

armanbilge:
CI seems to be stuck checking workflows are up-to-date for an hour 🙃

PR #2107, opened 2021-07-07
Fix `MonadCancel#guaranteeCase` for transformers and harden with new laws

djspiewak:


djspiewak:
@vasilmkd I think this is kind of ready to go? I'd be fine merging it as-is, but if we want to clean up the commits at all, we can rebase things and push to a non-`wip/` branch and merge from there.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1151,22 +1151,21 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
...
   def bracketFull[A, B](acquire: Poll[IO] => IO[A])(use: A => IO[B])(
-      release: (A, OutcomeIO[B]) => IO[Unit]): IO[B] = {
-    val safeRelease: (A, OutcomeIO[B]) => IO[Unit] =
-      (a, out) => IO.uncancelable(_ => release(a, out))
-
+      release: (A, OutcomeIO[B]) => IO[Unit]): IO[B] =
     IO.uncancelable { poll =>
       acquire(poll).flatMap { a =>
-        val finalized = poll(IO.unit >> use(a)).onCancel(safeRelease(a, Outcome.Canceled()))
+        def fin(oc: OutcomeIO[B]) = IO.uncancelable(_ => release(a, oc))
```
vasilmkd, line 1157: Doesn't `onCancel` on the `finalized` value force this method to be evaluated at least once? Should we go for a `val`?

djspiewak: I think the `val` wouldn't change anything. We would still need multiple versions of it since we're passing different outcomes.

vasilmkd: You are right, I missed that detail.

vasilmkd (approved):

testkit/shared/src/main/scala/cats/effect/testkit/TestInstances.scala
```diff
@@ -24,11 +24,12 @@ import cats.effect.kernel.testkit.{
   OutcomeGenerators,
   ParallelFGenerators,
   SyncGenerators,
-  SyncTypeGenerators
+  SyncTypeGenerators,
+  TestInstances => KernelTestkitTestInstances
```
djspiewak, line 28: Out of curiosity, can we just have `TestInstances` extend `KernelTestkitTestInstances`?

djspiewak: Oh, it's because forwarders aren't bincompat. Got it.

Issue #1959, opened 2021-05-12
Restart scala.js runtime environment


cquiroz:
We build all our scala.js apps with cats-effect 3 starting with `IOApp` which is exported to the js runtime. e.g.
https://github.com/gemini-hlsw/explore/blob/fef6caaafdafa144b6a9ae04afad18990467e132/explore/src/main/scala/explore/Explore.scala#L61

This works fine for the regular browser case and node but has some restrictions when doing development on browser.
There we often want to use HMR (Hot module replacement) where the dev environent detects changes to the compiled scala.js output and reloads the js code without a browser reload.

In practice this means we call `IOApp.main` again, however the code below can only be ever called once.
https://github.com/typelevel/cats-effect/blob/series/3.x/core/js/src/main/scala/cats/effect/IOApp.scala#L33

Would it be possible to add a method that could be called on replacement to reset the environment such that another call to `IOApp.main` wouldn't fail?

djspiewak:
This is very interesting! How would you expect this to work exactly? Just an `IOApp#reset()` or something like that?

cquiroz:
Yes! a method on `IOApp` that could be called from the JS runtime

djspiewak:
Do you *just* want it exposed to JavaScript, or is this something you would call from Scala?

cquiroz:
I wouldn't expect to call it from java on my use case of HMR. And it would need to be a kind of `end-of-the-world` call returning `Unit` to be exported easily to js

djspiewak:
I think this is sufficiently addressed as of #2114. If you disagree, leave a comment and I'll reopen!

armanbilge:
So I wasn't around when this originally came up, but I think the right solution here is that browser applications that have bespoke needs just shouldn't use `IOApp`. Instead they can just call `unsafeRunAndForget()`.

`IOApp` works fine for a vanilla usecase, but any serious web framework has its own application class anyway. `IOApp` brings little to no value for browser applications.

Calico offers an `IOWebApp`:
https://github.com/armanbilge/calico/blob/bc9b5cb9b06708c0a2fc35cc66c593823445d8f3/calico/src/main/scala/calico/IOWebApp.scala#L25

Tyrian offers a `TyrianApp`:
https://github.com/PurpleKingdomGames/tyrian/blob/80a6b76944309a98048d97dc4709154195ac2bac/tyrian-io/src/main/scala/tyrian/TyrianApp.scala#L15

[...]

PR #2114, opened 2021-07-09
Added private functions for resetting the global runtime

djspiewak:
This is one attempt to (somewhat) address #1959. Would love some thoughts on it. I basically want to make the most narrowly-scoped thing possible. You could get at this by calling `IORuntime.asInstanceOf[{ def resetGlobal(): Unit }].resetGlobal()`.

### DIFF
--- a/core/js/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
+++ b/core/js/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
@@ -45,6 +45,9 @@ private[unsafe] abstract class IORuntimeCompanionPlatform { this: IORuntime.type
     _global = global
   }
 
+  private[effect] def resetGlobal(): Unit =
+    _global = null
+
   lazy val global: IORuntime = {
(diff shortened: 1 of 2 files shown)


vasilmkd:
Gotta love reflective calls.

cquiroz:
Thanks I'll give it a try, cool
Trick to allow calls to a private method

johnynek:
I'm confused how this interacts well with `lazy val global`? Isn't that fixed exactly once and changing `_global` won't change anything?

cquiroz:
FWIW this is working as I wanted for my use case

djspiewak:
> I'm confused how this interacts well with lazy val global? Isn't that fixed exactly once and changing _global won't change anything?

Bleh. You're correct. It doesn't interact at all. This basically only solves the problem from the perspective of `IOApp`.

johnynek:
mutation. Not even once!

;)

djspiewak:
> mutation. Not even once!

Sir, you came to the WRONG part of town! ;-)

PR #2115, opened 2021-07-10
WIP: Adding the Parallel instance for Eval

BalmungSan:
The other day in the **Discord** server there was a quick discussion on #general about if it would be better to use `Future` over `IO` for parallelizing **CPU** only work.
During the discussion, there was also evaluated the possibility of adding a `LazyFuture` type as just an alias / wrapper over `IO` or adding a `Parallel` instance for `Eval`.
The main argument about using something different than `IO` was that such type should imply that the computation should not produce _"side-effects"_ but just **CPU** work and parallelization.

This PR attempts to revive such discussion and see if people find this _(or a similar approach)_ useful enough to be included.

I decided to start with the `Parallel` for `Eval` route using `IO.Par` under the hood; I kind of like the simplicity of the change, but dislike that users need to import `cats.effect.implicits._` and not sure how the direct usage of `IORuntime` would interact if the user is also using `IO` for other parts of the application.
Rather I think it would be better to add another type `LazyFuture` that may be a simple wrapper over `IO` so users can keep composing the result if they want or just run it there.

Anyways, what do you all think?

-----

PS: I know I am missing tests and that the naming is not the best, I just wanted to have something working to show and start the discussion since I feel is very possible this won't be merged.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/instances/AllInstances.scala
+++ b/core/shared/src/main/scala/cats/effect/instances/AllInstances.scala
@@ -17,4 +17,4 @@
 package cats.effect
 package instances
 
-trait AllInstances extends kernel.instances.GenSpawnInstances
+trait AllInstances extends kernel.instances.GenSpawnInstances with ParEval
--- a/core/shared/src/main/scala/cats/effect/instances/ParEval.scala
+++ b/core/shared/src/main/scala/cats/effect/instances/ParEval.scala
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2020-2021 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.instances
+
+import cats.{Applicative, Eval, Monad, Parallel, ~>}
+import cats.arrow.FunctionK
+import cats.effect.IO
+import cats.effect.kernel.Par.ParallelF
+import cats.effect.unsafe.implicits.global
+
+trait ParEval {
+  implicit final val parEvalInstance: Parallel[Eval] =
+    new Parallel[Eval] {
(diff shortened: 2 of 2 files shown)


djspiewak:
There's some interesting philosophical discussion to be had here. My interpretation of `Parallel` is that it is mapping an `F[_]` into some other type space, `G[_]`, such that `F` and `G` are isomorphic in expressiveness, but the semantics of `ap` are different. A category theoretical intuition for this is that we're considering alternative endofunctor monoids for a given type constructor `F[_]`. The strong implication here is that the underlying "concept type" is the same, while the semantics vary around it.

This is most definitely not the case for `Eval` and `IO`, which are two fundamentally very differently capable type constructors. Thus, mapping `Eval` to `IO.Par` makes me quite nervous.

But with that said, mapping `Eval` to some `Eval.Par` which is an opaque type around `IO` seems a bit less weird. I can see this being a valid thing to do. *However*, there are two problems I can see here, one practical and one pragmatically predictive:

- `Eval` defines a function `Eval[A] => A`. This is problematic in any situation involving asynchronous callbacks because it necessitates locking the current thread. This is in fact exactly what *has* to happen in any underlying implementation behind something like `parMapN`, where two fibers complete independently and the downstream has to asynchronously await those completions. **In other words, `Parallel[Eval]` cannot be implemented without blocking.**
- By the same argument that gives the side-eye to `Future`… Virtually no one is doing parallel pure computation. How many people have ever used the `.par` method on the pre-2.13 Scala Collections? I'd wager you could count the number on one hand. And yet this is a considerably more ergonomic and intuitive interface for *pure* parallel computation.

What a lot of it comes down to is that very, very few processes are purely CPU bound and *not* already request- or data-parallel. In most applications, you get so much high-level parallelism (e.g. from concurrent requests) that it's not worth imposing artificial parallelism on CPU-bound things within your responses. Where parallelism starts to matter is when you're doing multiple independent I/O operations (like calling some upstream service in some sort of scatter/gather pattern). *Here* doing things in parallel is extremely valuable, but it's also not a situation which is CPU-bound, and it's (by definition) also not something that would be in `Eval`.

The only commonplace example I can think of where CPU-bound operations really need to be usefully run in parallel is linear algebra solvers (e.g. for ML runtimes). But there you almost certainly have alternative parallel evaluation strategies which are superior, such as GPUs, or native runtimes like BLAS. At the very least, wrapping anything at that level in the overhead of a monadic interface would be unacceptably costly.

And there we get to the root of the problem: if you're *really* CPU bound, then you're probably going to gain more by avoiding monads (`Eval` or `IO` alike) than you would gain by imposing artificial parallelism on your calculation. All of which is to say that I just don't see much practical use for `Parallel[Eval]` even if it *didn't* imply thread blockage. It does though, which makes it very dangerous on top of all of the above.

PR #2120, opened 2021-07-12
Document outcome

TimWSpence:
Document `Outcome` and in particular why `Outcome#Succeeded` contains a value of type `F[A]` rather than one of type `A`.


vasilmkd (approved):

djspiewak (approved):
This is great, thank you!

PR #2125, opened 2021-07-13
Add `suspended` constructor for Dispatcher

catostrophe:
This is to address the following peculiarities of Dispatcher:

1. Dispatcher is created as a resource and has some "infrastructure" under the hood. When the resource scope is closed, all the fibers launched by the dispatcher get canceled. This promotes users to create a long-lived or even global dispatcher with a wide lifespan.

2. At the same time, the dispatcher needs to capture the current context of F that is needed for its internals. Thus, a global dispatcher uses the context captured at the moment of creation and ignores local modifications. This applies both to the execution context and ReaderT's context. See examples: 
https://scastie.scala-lang.org/catostrophe/ZCa884n1SASE0kwLqRF8Qg
https://scastie.scala-lang.org/catostrophe/w4ON2L2SRFSCAs8QxZbUgA

The proposed change detaches the context of F from the resource of the dispatcher internals, so the dispatcher can be created as a global resource, but context capturing is suspended until the place of usage.

```scala
Dispatcher.suspended[F].use { dispF =>
  ... // local modifications to the EC or user context
  _ <- dispF.flatMap(disp => ...) // launch future in the current context
  ...
}
```

N.B. If multiple instances of Dispatcher are produced from the same `F[Dispatcher]`, they will share common internals but this should not break the whole mechanism. This needs confirmation from the authors though.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
@@ -103,7 +103,15 @@ object Dispatcher {
    * Once the resource scope exits, all active effects will be canceled, and
    * attempts to submit new effects will throw an exception.
    */
-  def apply[F[_]](implicit F: Async[F]): Resource[F, Dispatcher[F]] = {
+  def apply[F[_]](implicit F: Async[F]): Resource[F, Dispatcher[F]] =
+    suspended[F].evalMap(identity)
+
+  /**
+   * Create a [[Dispatcher]] within the context of F detached from the resource
+   * scope. This allows the dispatcher to capture the context at the place of
+   * usage rather than creation.
+   */
+  def suspended[F[_]](implicit F: Async[F]): Resource[F, F[Dispatcher[F]]] = {
     final case class Registration(action: F[Unit], prepareCancel: F[Unit] => Unit)
         extends AtomicBoolean(true)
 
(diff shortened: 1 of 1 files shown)


djspiewak:
Oh this is interesting…

djspiewak:
I'm pretty sure that the fact that the `Supervisor` is constructed in the outer `Resource` means that this doesn't work the way that you think it does. :-( It would be worth adding some tests to find out.

Thinking about this more deeply… Is there *that* much difference between this and just `Dispatcher[F].use(...)`? It's really not the end of the world if people create many of them, it's just somewhat more efficient if they share.

catostrophe:
Ouch, yes, @djspiewak you're absolutely right. I was too hasty. I realized it's the Supervisor who captures the context. So this solution just doesn't work as I expected.

Before closing this completely, I'll try to modify Supervisor and explore the possibility of suspended context capturing.

As for `Dispatcher[F].use(...)`, it may end up leaving the use scope before the launched fiber is completed, which I try to avoid by widening the lifespan of Dispatcher.

catostrophe:
I couldn't write a proper way to postpone the full context capturing (including ReaderT's) using only `Async` methods. Async might have something similar to `Unlift`, but it doesn't. I solved the issue with extra means outside of CE3. So closing this.

PR #2126, opened 2021-07-13
Reimplement `IO#guaranteeCase` and `IO.bracketFull`

vasilmkd:
Initially, I was going to say how reimplementing `IO#bracketFull` in terms of `IO#guaranteeCase` is clearer and uses less special cased code for guarding user code against misuse of the inherently non-lexical nature of the `bracketFull` constructor, but it turns out it is both simpler and faster.

`series/3.x`:
```
Benchmark               (size)   Mode  Cnt      Score     Error  Units
AsyncBenchmark.bracket     100  thrpt   20  29369.257 ± 492.607  ops/s
```

`This PR`:
```
Benchmark               (size)   Mode  Cnt      Score     Error  Units
AsyncBenchmark.bracket     100  thrpt   20  30321.272 ± 998.223  ops/s
```


djspiewak:
Fascinating!

PR #2128, opened 2021-07-14
Cats Effect 3 tracing docs

vasilmkd:


docs/tracing.md
```diff
@@ -0,0 +1,267 @@
...
+poluted with the implementation methods of the `IO` run-loop, trace information
+which is unnecessary and confusing for the end-user application.
+
+`IO` solves this problem by collecting a stack trace at various `IO` operations
+that a fiber executes during its lifetime, and knitting them together to produce
+a more coherent view of the fiber's execution path.
+
+However, fiber tracing isn't limited to just collecting and assembling
+meaningful stack traces. In fact, the first release of tracing support for the
+Cats Effect 3 release comes with the following features:
```
djspiewak:
```suggestion
meaningful stack traces:
```

I'm not quite sure what to do with the following section. It contains some really cool information, but it also reads a lot more like release notes than documentation. :-) Maybe that's okay.

docs/tracing.md
```diff
@@ -0,0 +1,267 @@
...
+For example, `void` and `as` are combinators that are derived from `map`, and
+should appear in the fiber trace rather than `map`. The same goes for `IO`
+constructors such as `async`, which is a complex composition of many lower level
+implementation details, none of which are present in the final stack trace, as
+they are not relevant to the user.
+3. **JVM and GC safe trace information cache**. The trace information caching
+mechanism of Cats Effect 3 has been completely rewritten to take advantage of
+the `java.lang.ClassValue` platform APIs, a JVM-native caching solution with
+several appealing properties:
+   - Class-loader safety -- for the first time ever, long-running Cats Effect
```
djspiewak:
```suggestion
   - Class-loader safety -- long-running Cats Effect
```

docs/tracing.md
```diff
@@ -0,0 +1,267 @@
...
+fiber.
+2. **Constructor/combinator inference**. Stack traces can be walked to determine
+what specific `IO` constructor or combinator was actually called by user code.
+For example, `void` and `as` are combinators that are derived from `map`, and
+should appear in the fiber trace rather than `map`. The same goes for `IO`
+constructors such as `async`, which is a complex composition of many lower level
+implementation details, none of which are present in the final stack trace, as
+they are not relevant to the user.
+3. **JVM and GC safe trace information cache**. The trace information caching
+mechanism of Cats Effect 3 has been completely rewritten to take advantage of
```
djspiewak:
```suggestion
mechanism of Cats Effect was completely rewritten in version 3.2 to take advantage of
```

docs/tracing.md
```diff
@@ -0,0 +1,267 @@
...
+the `java.lang.ClassValue` platform APIs, a JVM-native caching solution with
+several appealing properties:
+   - Class-loader safety -- for the first time ever, long-running Cats Effect
+   applications fully respect dynamic class-loading environments, promptly
+   releasing references to unloaded classes when instructed by the JVM runtime,
+   thus helping with GC.
+   - Lock-free caching data structures natively optimized by the JIT compiler
+   for maximum performance. During the development of the cached asynchronous
+   stack tracing feature for Cats Effect 3, it was measured that the switch to
```
djspiewak:
```suggestion
   stack tracing feature for Cats Effect, it was measured that the switch to
```

docs/tracing.md
```diff
@@ -0,0 +1,267 @@
...
+Cats Effect runtime, stripped of unnecessary internal implementation details and
+augmented using asynchronous stack tracing information gathered at runtime, such
+as information about traced `IO` constructors and combinators, before finally
+being presented to the user.
+
+### General note on stack tracing performance
+
+As a note of caution, fiber tracing generally introduces overhead to
+applications in the form of higher CPU usage, memory and GC pressure. Always
+remember to performance test your specific application with tracing enabled
```
djspiewak:
```suggestion
remember to performance test your specific application with and without tracing enabled
```

docs/tracing.md
```diff
@@ -0,0 +1,267 @@
...
+For example, to enable full stack tracing and a trace buffer of size 1024,
+specify the following system properties:
+```
+-Dcats.effect.stackTracingMode=full -Dcats.effect.traceBufferLogSize=10
+```
+
+## Stack tracing modes
+
+### `none`
+No tracing is instrumented by the program and thus incurs no performance impact.
```
djspiewak:
```suggestion
Tracing is fully disabled and all performance impacts are negated. Outside of some extremely small (~8-16 bytes) object footprint differences, this should be exactly identical to a hypothetical `IO` run-loop which lacks any support for tracing.
```

docs/tracing.md
```diff
@@ -0,0 +1,267 @@
...
+in synthetic benchmarks for tightly looping `IO` programs consisting of a
+combination of both synchronous and asynchronous combinators (in Cats Effect 3,
+all `IO` programs are asynchronous in nature). The actual performance impact of
+cached tracing will be much lower for any program that performs any sort of I/O.
+Again, we strongly recommend benchmarking applications that make use of tracing,
+in order to avoid surprises and to get a realistic picture of the specific
+impact on your application.
+
+This is the recommended stack tracing mode to run in most production
+applications and is enabled by default.
```
djspiewak:
```suggestion
It's also worth qualifying the above numbers slightly. A "30% performance degradation" in this case is defined relative to `IO`'s evaluation *without any support for tracing.* This performance bar is extremely high, meaning that the absolute (as opposed to relative) difference between "tracing" and "no tracing at all" is measured in micro- and nanoseconds in most cases. To put all of this in perspective, even with tracing, `IO`'s evaluation is still multiple orders of magnitude faster than the equivalent program written using Scala's `Future` when measured using the same highly-synthetic benchmarks. All of which is not to say that `Future` is slow, but rather that these benchmarks are attempting to magnify even the slightest differences in `IO`'s performance and are not representative of most real-world scenarios.

When tracing was first introduced in Cats Effect 2, its absolute performance impact was considerably higher than the impact of the rewritten implementation in Cats Effect 3.2. Despite this, even extremely performance-sensitive real world applications subjected to intensive benchmarking and stress testing did not observe differences in top-line metrics.

It is for this reason that `cached` tracing is on by default, and is the recommended mode even when running in production.
```

docs/tracing.md
```diff
@@ -0,0 +1,267 @@
...
+As with cached stack tracing, when full stack tracing is enabled, a stack trace
+is captured for most `IO` combinators and constructors.
+
+The difference comes in the fact that stack traces are collected on
+*every invocation*, so naturally, most programs will experience a significant
+performance hit. This mode is mainly useful for debugging in development
+environments.
+
+## Enhanced exceptions
+The stack trace of an exception caught by the Cats Effect 3 `IO` runtime looks
```
djspiewak:
```suggestion
Without tracing, the stack trace of an exception caught by the `IO` runtime often looks
```

docs/tracing.md
```diff
@@ -0,0 +1,267 @@
...
+`cats.effect.enhancedExceptions`. It is enabled by default.
+
+It can be disabled with the following configuration:
+```
+-Dcats.effect.enhancedExceptions=false
+```
+
+### Complete code
+Here is the code snippet that was used to generate the above examples:
+```scala
```
djspiewak:
```suggestion
```scala mdoc
```

djspiewak (changes requested):
This is really great! I made a few suggestions along the way to try to make it more ever-green (i.e. just as relevant to a reader years from now as it is today). Also not *entirely* sure to do with the section in the middle, but we can work on it.

vasilmkd:
I incorporated your changes in a single commit and reformatted the page.

vasilmkd:
I don't mind taking the middle part out and dropping it in the release notes.

djspiewak:
Eh, let's keep it :-D

PR #2130, opened 2021-07-15
Fix typo in tutorial.md

moleike:
It should say `delay` instead of `blocking`

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -409,8 +409,8 @@ import java.io._
 
 def transmit[F[_]: Sync](origin: InputStream, destination: OutputStream, buffer: Array[Byte], acc: Long): F[Long] =
   for {
-    amount <- Sync[F].blocking(origin.read(buffer, 0, buffer.length))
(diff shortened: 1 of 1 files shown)


vasilmkd:
I'm sorry, I do believe that the example is indeed correct as it is. `InputStream` and `OutputStream` are the Java blocking I/O APIs and `IO.blocking` accurately wrap the blocking side-effectful calls that are `InputStream#read` and `OutputStream#write`.

vasilmkd:
You can read more about it [here](https://typelevel.org/cats-effect/docs/typeclasses/sync). Unfortunately, we are still working on the `IO` specific documentation.

moleike:
My bad, I assumed it was `delay` based on the preceding paragraph:

> replacing any use of `IO` by calls to the `delay` and `pure` methods of the
`Sync[F]` instance:

Thanks for your quick reply, @vasilmkd!

vasilmkd:
That's a good observation. Do you mind changing this PR to revert the code change and instead fix the paragraph text? Thanks. 😄

vasilmkd:
Closing this one, feel free to follow up on what we discussed above in a new PR.

PR #2134, opened 2021-07-20
Add method forwarders to `Resource`

### LINKED ISSUE #2116: Add method forwarders to `Resource`
`Resource` implements `Async`, but does so in a purely "typeclassy" way. If you want to do something like `rsrc.start` (where `rsrc: Resource[F, A]`), you actually need to `import cats.effect.syntax.all._`! This isn't a great experience.

Instead, we should flip all of this around and follow the same pattern `IO` does: move the definitions of these methods to `Resource` itself as members (taking the appropriate implicit evidence for `F` individually), and then change the typeclass implementations to simply *forward* the call onto the appropriate member function. This is actually a pretty easy change, just a little tedious and time-consuming. It should help with the ergonomics quite considerably though.

alexandrustana:
This PR should fix #2116

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -493,6 +493,9 @@ sealed abstract class Resource[F[_], +A] {
       }
     }
 
+  def !>[B](that: Resource[F, B])(implicit F: MonadCancel[F, Throwable]): Resource[F, B] =
+    forceR(that)
+
   def onCancel(fin: Resource[F, Unit])(implicit F: MonadCancel[F, Throwable]): Resource[F, A] =
     Resource applyFull { poll =>
       poll(this.allocated).onCancel(fin.use_) map { p =>
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -1185,17 +1185,21 @@ private[effect] trait ResourceHOInstances7 {
 
 abstract private[effect] class ResourceFOInstances0 extends ResourceFOInstances1 {
   implicit def catsEffectMonoidForResource[F[_], A](
-      implicit A0: Monoid[A]): Monoid[Resource[F, A]] =
+      implicit F0: Monad[F],
+      A0: Monoid[A]): Monoid[Resource[F, A]] =
     new ResourceMonoid[F, A] {
       def A = A0
+      def F = F0
     }
 }
 
 abstract private[effect] class ResourceFOInstances1 {
   implicit def catsEffectSemigroupForResource[F[_], A](
-      implicit A0: Semigroup[A]): ResourceSemigroup[F, A] =
+      implicit F0: Monad[F],
+      A0: Semigroup[A]): ResourceSemigroup[F, A] =
     new ResourceSemigroup[F, A] {
       def A = A0
+      def F = F0
     }
 }
 
@@ -1300,15 +1304,17 @@ abstract private[effect] class ResourceAsync[F[_]]
 }
 
 abstract private[effect] class ResourceMonadError[F[_], E]
-    extends ResourceApplicativeError[F, E]
-    with ResourceMonad[F]
+    extends ResourceMonad[F]
+    with ResourceApplicativeError[F, E]
     with MonadError[Resource[F, *], E] {
+
   implicit protected def F: MonadError[F, E]
 }
 
-abstract private[effect] class ResourceApplicativeError[F[_], E]
+private[effect] trait ResourceApplicativeError[F[_], E]
     extends ResourceApplicative[F]
     with ApplicativeError[Resource[F, *], E] {
+
   implicit protected def F: ApplicativeError[F, E]
 
   override def attempt[A](fa: Resource[F, A]): Resource[F, Either[E, A]] =
@@ -1321,7 +1327,7 @@ abstract private[effect] class ResourceApplicativeError[F[_], E]
     Resource.raiseError[F, A, E](e)
 }
 
-private[effect] trait ResourceMonad[F[_]]
+abstract private[effect] class ResourceMonad[F[_]]
     extends ResourceApplicative[F]
     with Monad[Resource[F, *]]
     with StackSafeMonad[Resource[F, *]] {
@@ -1337,8 +1343,8 @@ abstract private[effect] class ResourceApplicative[F[_]] extends Applicative[Res
   def pure[A](a: A): Resource[F, A] =
     Resource.pure(a)
 
-  override def ap[A, B](ff: Resource[F, A => B])(fa: Resource[F, A]): Resource[F, B] = {
-    fa.flatMap(a => ff.map(f => f(a)))
+  def ap[A, B](ff: Resource[F, A => B])(fa: Resource[F, A]): Resource[F, B] = {
+    ff.flatMap(f => fa.map(a => f(a)))
   }
 }
 
@@ -1351,6 +1357,7 @@ abstract private[effect] class ResourceMonoid[F[_], A]
 }
 
 abstract private[effect] class ResourceSemigroup[F[_], A] extends Semigroup[Resource[F, A]] {
+  implicit protected def F: Monad[F]
   implicit protected def A: Semigroup[A]
 
   def combine(rx: Resource[F, A], ry: Resource[F, A]): Resource[F, A] =
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -981,8 +981,6 @@ object Resource extends ResourceFOInstances0 with ResourceHOInstances0 with Reso
   def raiseError[F[_], A, E](e: E)(implicit F: ApplicativeError[F, E]): Resource[F, A] =
     Resource.eval(F.raiseError[A](e))
 
-  def empty[F[_], A](implicit A: Monoid[A]): Resource[F, A] = Resource.pure[F, A](A.empty)
-
   /**
    * `Resource` data constructor that wraps an effect allocating a resource,
    * along with its finalizers.
@@ -1321,7 +1319,7 @@ abstract private[effect] class ResourceMonoid[F[_], A]
     with Monoid[Resource[F, A]] {
   implicit protected def A: Monoid[A]
 
-  def empty: Resource[F, A] = Resource.empty
+  def empty: Resource[F, A] = Resource.pure[F, A](A.empty)
 }
 
 abstract private[effect] class ResourceSemigroup[F[_], A] extends Semigroup[Resource[F, A]] {
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -21,8 +21,8 @@ import cats.data.Kleisli
 import cats.syntax.all._
 import cats.effect.kernel.instances.spawn
 import cats.effect.kernel.implicits._
-
 import scala.annotation.tailrec
+import scala.annotation.unchecked.uncheckedVariance
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration.FiniteDuration
 
@@ -473,6 +473,96 @@ sealed abstract class Resource[F[_], +A] {
     new (F ~> F) {
       override def apply[B](gb: F[B]): F[B] = surround(gb)
     }
+
+  def forceR[B](that: Resource[F, B])(implicit F: MonadCancel[F, Throwable]): Resource[F, B] =
+    Resource applyFull { poll =>
+      poll(this.use_ !> that.allocated) map { p =>
+        // map exists on tuple in Scala 3 and has the wrong type
+        Functor[(B, *)].map(p)(fin => (_: Resource.ExitCase) => fin)
+      }
+    }
+
+  def onCancel(fin: Resource[F, Unit])(implicit F: MonadCancel[F, Throwable]): Resource[F, A] =
+    Resource applyFull { poll =>
+      poll(this.allocated).onCancel(fin.use_) map { p =>
+        Functor[(A, *)].map(p)(fin => (_: Resource.ExitCase) => fin)
+      }
+    }
+
+  /*
+   * 1. If the scope containing the *fiber* terminates:
+   *   a) if the fiber is incomplete, run its inner finalizers when it completes
+   *   b) if the fiber is succeeded, run its finalizers
+   * 2. If the fiber is canceled or errored, finalize
+   * 3. If the fiber succeeds and .cancel won the race, finalize eagerly and
+   *    `join` results in `Canceled()`
+   * 4. If the fiber succeeds and .cancel lost the race or wasn't called,
+   *    finalize naturally when the containing scope ends, `join` returns
+   *    the value
+   */
+  def start(implicit F: GenConcurrent[F, Throwable]): Resource[F, Fiber[Resource[F, *], Throwable, A@uncheckedVariance]] = {
+    final case class State(fin: F[Unit] = F.unit, finalizeOnComplete: Boolean = false)
+
+    Resource {
+      import Outcome._
+
+      F.ref[State](State()) flatMap { state =>
+        val finalized: F[Option[A]] = F uncancelable { poll =>
+          val ff: F[F[Option[A]]] =
+            poll(this.allocated) flatMap { // intentionally short-circuits in case of cancelation or error
+              case (a, rel) =>
+                state modify { s =>
+                  if (s.finalizeOnComplete)
+                    (s, rel.attempt.as(None))
+                  else
+                    (s.copy(fin = rel), F.pure(Some(a)))
+                }
+            }
+
+          ff.flatten
+        }
+
+        F.start(finalized) map { outer =>
+          val fiber = new Fiber[Resource[F, *], Throwable, A] {
+
+            def cancel =
+              Resource eval {
+                F uncancelable { poll =>
+                  // technically cancel is uncancelable, but separation of concerns and what not
+                  poll(outer.cancel) *> state.update(_.copy(finalizeOnComplete = true))
+                }
+              }
+
+            def join =
+              Resource eval {
+                outer.join.flatMap[Outcome[Resource[F, *], Throwable, A]] {
+                  case Canceled() =>
+                    Outcome.canceled[Resource[F, *], Throwable, A].pure[F]
+
+                  case Errored(e) =>
+                    Outcome.errored[Resource[F, *], Throwable, A](e).pure[F]
+
+                  case Succeeded(fp) =>
+                    fp map {
+                      case Some(a) =>
+                        a.pure[Outcome[Resource[F, *], Throwable, *]]
+
+                      case None =>
+                        Outcome.canceled[Resource[F, *], Throwable, A]
+                    }
+                }
+              }
+          }
+
+          val finalizeOuter =
+            state.modify(s => (s.copy(finalizeOnComplete = true), s.fin)).flatten
+
+          (fiber, finalizeOuter)
+        }
+      }
+    }
+  }
+
 }
 
 object Resource extends ResourceFOInstances0 with ResourceHOInstances0 with ResourcePlatform {
@@ -682,6 +772,49 @@ object Resource extends ResourceFOInstances0 with ResourceHOInstances0 with Reso
       rfb: Resource[F, B]
   ): Resource[F, Either[A, B]] =
     rfa.race(rfb)
+
+  def canceled[F[_]](implicit F: MonadCancel[F, Throwable]): Resource[F, Unit] =
+    Resource.eval(F.canceled)
+
+  def forceR[F[_], A, B](fa: Resource[F, A])(fb: Resource[F, B])(implicit F: MonadCancel[F, Throwable]): Resource[F, B] =
+    fa.forceR(fb)
+
+  def onCancel[F[_], A](fa: Resource[F, A], fin: Resource[F, Unit])(implicit F: MonadCancel[F, Throwable]): Resource[F, A] =
+    fa.onCancel(fin)
+
+  def uncancelable[F[_], A](body: Poll[Resource[F, *]] => Resource[F, A])(implicit F: MonadCancel[F, Throwable]): Resource[F, A] =
+    Resource applyFull { poll =>
+      val inner = new Poll[Resource[F, *]] {
+        def apply[B](rfb: Resource[F, B]): Resource[F, B] =
+          Resource applyFull { innerPoll =>
+            innerPoll(poll(rfb.allocated)) map { p =>
+              Functor[(B, *)].map(p)(fin => (_: Resource.ExitCase) => fin)
+            }
+          }
+      }
+
+      body(inner).allocated map { p =>
+        Functor[(A, *)].map(p)(fin => (_: Resource.ExitCase) => fin)
+      }
+    }
+
+  def unique[F[_]](implicit F: Unique[F]): Resource[F, Unique.Token] =
+    Resource.eval(F.unique)
+
+  def never[F[_], A](implicit F: GenSpawn[F, Throwable]): Resource[F, A] =
+    Resource.eval(F.never[A])
+
+  def cede[F[_]](implicit F: GenSpawn[F, Throwable]): Resource[F, Unit] =
+    Resource.eval(F.cede)
+
+  def start[F[_], A](fa: Resource[F, A])(implicit F: GenConcurrent[F, Throwable]): Resource[F, Fiber[Resource[F, *], Throwable, A]] =
+    fa.start
+
+  def deferred[F[_], A](implicit F: GenConcurrent[F, Throwable]): Resource[F, Deferred[Resource[F, *], A]] =
+    Resource.eval(F.deferred[A]).map(_.mapK(Resource.liftK[F]))
+
+  def ref[F[_], A](a: A)(implicit F: GenConcurrent[F, Throwable]): Resource[F, Ref[Resource[F, *], A]] =
+    Resource.eval(F.ref(a)).map(_.mapK(Resource.liftK[F]))
 
   /**
    * Creates a [[Resource]] by wrapping a Java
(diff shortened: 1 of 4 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -484,6 +485,182 @@ sealed abstract class Resource[F[_], +A] {
...
+  def forceR[B](that: Resource[F, B])(implicit F: MonadCancel[F, Throwable]): Resource[F, B] =
+    Resource applyFull { poll =>
+      poll(this.use_ !> that.allocated) map { p =>
+        // map exists on tuple in Scala 3 and has the wrong type
+        Functor[(B, *)].map(p)(fin => (_: Resource.ExitCase) => fin)
+      }
+    }
+
+  def onCancel(fin: Resource[F, Unit])(
+      implicit F: MonadCancel[F, Throwable]): Resource[F, A @uncheckedVariance] =
```
djspiewak: Does this actually need `@uncheckedVariance`?

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -484,6 +485,182 @@ sealed abstract class Resource[F[_], +A] {
...
+    Resource applyFull { poll =>
+      poll(this.allocated).onCancel(fin.use_) map { p =>
+        Functor[(A @uncheckedVariance, *)].map(p)(fin => (_: Resource.ExitCase) => fin)
```
djspiewak, line 502: I'm guessing it's not needed here either

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -484,6 +485,182 @@ sealed abstract class Resource[F[_], +A] {
...
+
+        case Outcome.Errored(e) =>
+          fin(Outcome.Errored(e)).use_.handleError(_ => ())
+
+        case Outcome.Canceled() =>
+          fin(Outcome.Canceled()).use_
+      }
+
+      back.map(p =>
+        Functor[(A @uncheckedVariance, *)].map(p)(fu => (_: Resource.ExitCase) => fu))
```
djspiewak, line 524: Or here

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -484,6 +485,182 @@ sealed abstract class Resource[F[_], +A] {
...
+          val finalizeOuter =
+            state.modify(s => (s.copy(finalizeOnComplete = true), s.fin)).flatten
+
+          (fiber, finalizeOuter)
+        }
+      }
+    }
+  }
+
+  def evalOn(ec: ExecutionContext)(implicit F: Async[F]): Resource[F, A @uncheckedVariance] =
```
djspiewak: Or here

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -484,6 +485,182 @@ sealed abstract class Resource[F[_], +A] {
...
+    Resource applyFull { poll =>
+      poll(this.allocated).evalOn(ec) map { p =>
+        Functor[(A @uncheckedVariance, *)].map(p)(fin => (_: Resource.ExitCase) => fin)
```
djspiewak, line 616: Or here

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -484,6 +485,182 @@ sealed abstract class Resource[F[_], +A] {
...
+          case Right(s) => f(s).attempt
+        }
+      case p @ Pure(_) =>
+        Resource.pure(p.a.asRight)
+      case e @ Eval(_) =>
+        Resource.eval(e.fa.attempt)
+    }
+
+  def handleErrorWith[E](f: E => Resource[F, A @uncheckedVariance])(
+      implicit F: MonadError[F, E]): Resource[F, A] =
```
djspiewak: Actually, this is the compiler telling you that you need a different type here! In particular, the signature should be something like this:

```suggestion
  def handleErrorWith[B >: A, E](f: E => Resource[F, B])(implicit F: MonadError[F, E]): Resource[F, B] =
```

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -484,6 +485,182 @@ sealed abstract class Resource[F[_], +A] {
...
+    attempt.flatMap {
+      case Right(a) => Resource.pure(a)
+      case Left(e) => f(e)
+    }
+
+  def combine(that: Resource[F, A @uncheckedVariance])(
+      implicit A: Semigroup[A @uncheckedVariance]): Resource[F, A] =
```
djspiewak:
```suggestion
  def combine[B >: A](that: Resource[F, B])(implicit A: Semigroup[B]): Resource[F, B] =
```

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -484,6 +485,182 @@ sealed abstract class Resource[F[_], +A] {
...
+    for {
+      x <- this
+      y <- that
+    } yield A.combine(x, y)
+
+  def combineK(that: Resource[F, A @uncheckedVariance])(
```
djspiewak: Similar pattern to `combine`

djspiewak (changes requested):
Looks really good! Some changes are needed, but this is close

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
alexandrustana: Yeah, the compiler was pointing that the type variance wasn't ok.
I've switched to this approach (thank you for pointing it out) and removed the `@uncheckedVariance` occurrences

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -484,6 +484,178 @@ sealed abstract class Resource[F[_], +A] {
...
+  def forceR[B](that: Resource[F, B])(implicit F: MonadCancel[F, Throwable]): Resource[F, B] =
+    Resource applyFull { poll =>
+      poll(this.use_ !> that.allocated) map { p =>
+        // map exists on tuple in Scala 3 and has the wrong type
+        Functor[(B, *)].map(p)(fin => (_: Resource.ExitCase) => fin)
+      }
+    }
+
+  def onCancel[B >: A](fin: Resource[F, Unit])(
+      implicit F: MonadCancel[F, Throwable]): Resource[F, B] =
```
djspiewak: I think this one *shouldn't* need the context bounds.

```suggestion
  def onCancel(fin: Resource[F, Unit])(
      implicit F: MonadCancel[F, Throwable]): Resource[F, A] =
```

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -484,6 +484,178 @@ sealed abstract class Resource[F[_], +A] {
...
+    Resource applyFull { poll =>
+      poll(this.allocated).onCancel(fin.use_) map { p =>
+        Functor[(B, *)].map(p)(fin => (_: Resource.ExitCase) => fin)
+      }
+    }
+
+  def guaranteeCase[B >: A](fin: Outcome[Resource[F, *], Throwable, B] => Resource[F, Unit])(
+      implicit F: MonadCancel[F, Throwable]): Resource[F, B] =
```
djspiewak:
```suggestion
  def guaranteeCase(fin: Outcome[Resource[F, *], Throwable, A @uncheckedVariance] => Resource[F, Unit])(
      implicit F: MonadCancel[F, Throwable]): Resource[F, A] =
```

This is one case where I believe `@uncheckedVariance` is actually needed, for the same reason it is needed in `IO`.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -484,6 +484,178 @@ sealed abstract class Resource[F[_], +A] {
...
+   *   b) if the fiber is succeeded, run its finalizers
+   * 2. If the fiber is canceled or errored, finalize
+   * 3. If the fiber succeeds and .cancel won the race, finalize eagerly and
+   *    `join` results in `Canceled()`
+   * 4. If the fiber succeeds and .cancel lost the race or wasn't called,
+   *    finalize naturally when the containing scope ends, `join` returns
+   *    the value
+   */
+  def start[B >: A](
+      implicit F: Concurrent[F]): Resource[F, Fiber[Resource[F, *], Throwable, B]] = {
```
djspiewak:
```suggestion
  def start(
      implicit F: Concurrent[F]): Resource[F, Fiber[Resource[F, *], Throwable, A @uncheckedVariance]] = {
```

This one also needs `uncheckedVariance`. Basically this comes from the fact that `Fiber` and `Outcome` are both invariant, but their parameter is used in entirely covariant contexts when the type constructor is also covariant, meaning that `uncheckedVariance` is sound.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -484,6 +484,178 @@ sealed abstract class Resource[F[_], +A] {
...
+          val finalizeOuter =
+            state.modify(s => (s.copy(finalizeOnComplete = true), s.fin)).flatten
+
+          (fiber, finalizeOuter)
+        }
+      }
+    }
+  }
+
+  def evalOn[B >: A](ec: ExecutionContext)(implicit F: Async[F]): Resource[F, B] =
```
djspiewak:
```suggestion
  def evalOn(ec: ExecutionContext)(implicit F: Async[F]): Resource[F, A] =
```

Another unnecessary context bound.

djspiewak (changes requested):
Went a bit too far! :-) Some of these *do* need `uncheckedVariance`.

alexandrustana:
> Went a bit too far! :-) Some of these _do_ need `uncheckedVariance`.

Sorry for that 😓 

Seems that this type of functor invocation: `Functor[(A, *)].map(...)` 
also needs some kind of solution(`covariant type A occurs in invariant position in type [β$4$](A, β$4$) of type Λ$`). 
Maybe  `@uncheckedVariance` ?

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -484,6 +484,179 @@ sealed abstract class Resource[F[_], +A] {
...
+  }
+
+  def evalOn(ec: ExecutionContext)(implicit F: Async[F]): Resource[F, A] =
+    Resource applyFull { poll =>
+      poll(this.allocated).evalOn(ec) map { p =>
+        Functor[(A, *)].map(p)(fin => (_: Resource.ExitCase) => fin)
+      }
+    }
+
+  def attempt[E](implicit F: MonadError[F, E]): Resource[F, Either[E, A]] =
```
joroKr21:
```suggestion
  def attempt[E](implicit F: ApplicativeError[F, E]): Resource[F, Either[E, A]] =
```

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -484,6 +484,179 @@ sealed abstract class Resource[F[_], +A] {
...
+    this match {
+      case Allocate(resource) =>
+        Resource.applyFull { poll =>
+          resource(poll).attempt.map {
+            case Left(error) => (Left(error), (_: ExitCase) => F.unit)
+            case Right((a, release)) => (Right(a), release)
+          }
+        }
+      case Bind(source, f) =>
+        MonadError[Resource[F, *], E].attempt(source).flatMap {
```
joroKr21:
```suggestion
        source.attempt.flatMap {
```

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -484,6 +484,179 @@ sealed abstract class Resource[F[_], +A] {
...
+          case Right(s) => f(s).attempt
+        }
+      case p @ Pure(_) =>
+        Resource.pure(p.a.asRight)
+      case e @ Eval(_) =>
+        Resource.eval(e.fa.attempt)
+    }
+
+  def handleErrorWith[B >: A, E](f: E => Resource[F, B])(
+      implicit F: MonadError[F, E]): Resource[F, B] =
```
joroKr21:
```suggestion
      implicit F: ApplicativeError[F, E]): Resource[F, B] =
```

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -722,6 +895,95 @@ object Resource extends ResourceFOInstances0 with ResourceHOInstances0 with Reso
       implicit F: Sync[F]): Resource[F, A] =
     Resource.make(acquire)(autoCloseable => F.blocking(autoCloseable.close()))
 
+  def canceled[F[_]](implicit F: MonadCancel[F, Throwable]): Resource[F, Unit] =
```
joroKr21:
```suggestion
  def canceled[F[_]](implicit F: MonadCancel[F, _]): Resource[F, Unit] =
```

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -722,6 +895,95 @@ object Resource extends ResourceFOInstances0 with ResourceHOInstances0 with Reso
...
+
+      body(inner).allocated map { p =>
+        Functor[(A, *)].map(p)(fin => (_: Resource.ExitCase) => fin)
+      }
+    }
+
+  def unique[F[_]](implicit F: Unique[F]): Resource[F, Unique.Token] =
+    Resource.eval(F.unique)
+
+  def never[F[_], A](implicit F: GenSpawn[F, Throwable]): Resource[F, A] =
```
joroKr21:
```suggestion
  def never[F[_], A](implicit F: GenSpawn[F, _]): Resource[F, A] =
```

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -722,6 +895,95 @@ object Resource extends ResourceFOInstances0 with ResourceHOInstances0 with Reso
...
+    Resource.eval(F.never[A])
+
+  def cede[F[_]](implicit F: GenSpawn[F, Throwable]): Resource[F, Unit] =
```
joroKr21:
```suggestion
  def cede[F[_]](implicit F: GenSpawn[F, _]): Resource[F, Unit] =
```

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -722,6 +895,95 @@ object Resource extends ResourceFOInstances0 with ResourceHOInstances0 with Reso
...
+    Resource.eval(F.cede)
+
+  def deferred[F[_], A](
+      implicit F: GenConcurrent[F, Throwable]): Resource[F, Deferred[Resource[F, *], A]] =
```
joroKr21:
```suggestion
      implicit F: GenConcurrent[F, _]): Resource[F, Deferred[Resource[F, *], A]] =
```

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -722,6 +895,95 @@ object Resource extends ResourceFOInstances0 with ResourceHOInstances0 with Reso
...
+    Resource.eval(F.deferred[A]).map(_.mapK(Resource.liftK[F]))
+
+  def ref[F[_], A](a: A)(
+      implicit F: GenConcurrent[F, Throwable]): Resource[F, Ref[Resource[F, *], A]] =
```
joroKr21:
```suggestion
      implicit F: GenConcurrent[F, _]): Resource[F, Ref[Resource[F, *], A]] =
```

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -722,6 +895,95 @@ object Resource extends ResourceFOInstances0 with ResourceHOInstances0 with Reso
...
+    Resource.eval(F.monotonic)
+
+  def realTime[F[_]](implicit F: Clock[F]): Resource[F, FiniteDuration] =
+    Resource.eval(F.realTime)
+
+  def suspend[F[_], A](hint: Sync.Type)(thunk: => A)(implicit F: Sync[F]): Resource[F, A] =
+    Resource.eval(F.suspend(hint)(thunk))
+
+  def sleep[F[_]](time: FiniteDuration)(
+      implicit F: GenTemporal[F, Throwable]): Resource[F, Unit] =
```
joroKr21:
```suggestion
      implicit F: GenTemporal[F, _]): Resource[F, Unit] =
```

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -722,6 +895,95 @@ object Resource extends ResourceFOInstances0 with ResourceHOInstances0 with Reso
...
+            }
+          }
+        }
+      }
+    }
+
+  def executionContext[F[_]](implicit F: Async[F]): Resource[F, ExecutionContext] =
+    Resource.eval(F.executionContext)
+
+  def raiseError[F[_], A, E](e: E)(implicit F: MonadError[F, E]): Resource[F, A] =
```
joroKr21:
```suggestion
  def raiseError[F[_], A, E](e: E)(implicit F: ApplicativeError[F, E]): Resource[F, A] =
```

joroKr21 (approved):
Some constraints can be relaxed further

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
joroKr21: This is probably not stack safe though

djspiewak: It sucks but we probably have to `defer` around it. Honestly I wonder if `.allocated.attempt` would be better.

djspiewak:
> Seems that this type of functor invocation: Functor[(A, *)].map(...)
also needs some kind of solution(covariant type A occurs in invariant position in type [β$4$](A, β$4$) of type Λ$).
> Maybe @uncheckedVariance ?

Ah that's unfortunate. Yes, it needs `@uncheckedVariance`. I'm actually slightly surprised that it has this problem, but it's not the end of the world.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
alexandrustana: Using `allocated` would require having `MonadCancel` in scope.

alexandrustana:
> Some constraints can be relaxed further

Thank you for the suggestions, I've tried to relax constraints in the places you suggested.
This however generated the need for an `ApplicativeError` instance and having this enabled me to change the hierarchy a bit. (e.g. `ResourceMonadError` now extends `ResourceMonad` and also `ResourceApplicativeError`)

joroKr21:
> This however generated the need for an ApplicativeError instance and having this enabled me to change the hierarchy a bit. (e.g. ResourceMonadError now extends ResourceMonad and also ResourceApplicativeError)

Hmm I thought the constraints are on the `F` not on `Resource` - unless we nest them? 🤔

alexandrustana:
Unfortunately, the constraints are on `Resource`. Not sure what you mean with nesting 😅

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -1133,120 +1288,80 @@ abstract private[effect] class ResourceAsync[F[_]]
...
+    with Monad[Resource[F, *]]
     with StackSafeMonad[Resource[F, *]] {
-
   implicit protected def F: Monad[F]
 
+  def flatMap[A, B](fa: Resource[F, A])(f: A => Resource[F, B]): Resource[F, B] =
+    fa.flatMap(f)
+}
+
+abstract private[effect] class ResourceApplicative[F[_]] extends Applicative[Resource[F, *]] {
```
vasilmkd: Please revert this class, we have an implementation for `Applicative[Resource[F, *]]` because `Monad[Resource[F, *]]` extends it.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -1133,120 +1288,80 @@ abstract private[effect] class ResourceAsync[F[_]]
...
 abstract private[effect] class ResourceMonadError[F[_], E]
     extends ResourceMonad[F]
+    with ResourceApplicativeError[F, E]
     with MonadError[Resource[F, *], E] {
-  import Resource._
 
   implicit protected def F: MonadError[F, E]
+}
+
+private[effect] trait ResourceApplicativeError[F[_], E]
```
vasilmkd: Please also remove this class, we have it through `MonadError[Resource[F, *], E]`.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -1133,120 +1288,80 @@ abstract private[effect] class ResourceAsync[F[_]]
...
   def handleErrorWith[A](fa: Resource[F, A])(f: E => Resource[F, A]): Resource[F, A] =
-    attempt(fa).flatMap {
-      case Right(a) => Resource.pure(a)
-      case Left(e) => f(e)
-    }
+    fa.handleErrorWith(f)
 
   def raiseError[A](e: E): Resource[F, A] =
-    Resource.eval(F.raiseError[A](e))
+    Resource.raiseError[F, A, E](e)
```
vasilmkd, line 1304: Put these lines back in `ResourceMonadError` and also override `attempt` there.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -899,13 +1160,29 @@ private[effect] trait ResourceHOInstances4 extends ResourceHOInstances5 {
...
+}
+
+private[effect] trait ResourceHOInstances7 {
+  implicit def catsEffectApplicativeForResource[F[_]](
+      implicit F0: Applicative[F]): Applicative[Resource[F, *]] =
+    new ResourceApplicative[F] {
+      def F = F0
+    }
+}
+
```
vasilmkd: Also remove these 2 traits.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -484,6 +484,180 @@ sealed abstract class Resource[F[_], +A] {
...
+    this match {
+      case Allocate(resource) =>
+        Resource.applyFull { poll =>
+          resource(poll).attempt.map {
+            case Left(error) => (Left(error), (_: ExitCase) => F.unit)
+            case Right((a, release)) => (Right(a), release)
+          }
+        }
+      case Bind(source, f) =>
+        ApplicativeError[Resource[F, *], E].attempt(source).flatMap {
```
vasilmkd: Replace this line with `source.attempt.flatMap {` and you should be done. 😄

vasilmkd (changes requested):

joroKr21:
> Unfortunately, the constraints are on Resource. Not sure what you mean with nesting 😅

All the suggestions I made were constraints on `F[_]` 🤔 
I even checked that it compiles with just those changes.
By nesting I meant something like `Resource[Resource[F, *], A]` 😃

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
alexandrustana: I really thought that `attempt` would cause a `StackOverflow` that's why  I went with changing the hierarchy 😭

vasilmkd: That was already the case before. We might want to do something about it now, let me see.

alexandrustana:
I kind of went over the top with these changes, didn't think just using `source.attempt` would be enough. 
Thank you @joroKr21 and @vasilmkd for the review and the suggestions.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
vasilmkd: Indeed, this is stack unsafe, but the previous implementation is also stack unsafe. Let's think about how to fix that in another issue/PR, this one is big enough.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -484,6 +484,180 @@ sealed abstract class Resource[F[_], +A] {
     new (F ~> F) {
       override def apply[B](gb: F[B]): F[B] = surround(gb)
     }
+
+  def forceR[B](that: Resource[F, B])(implicit F: MonadCancel[F, Throwable]): Resource[F, B] =
+    Resource applyFull { poll =>
+      poll(this.use_ !> that.allocated) map { p =>
+        // map exists on tuple in Scala 3 and has the wrong type
+        Functor[(B, *)].map(p)(fin => (_: Resource.ExitCase) => fin)
```
vasilmkd, line 492: We have an `import Resource._` at the top of the class body, so you can drop reduce `Resource.ExitCase` to `ExitCase`. Same for all the other places.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -484,6 +484,180 @@ sealed abstract class Resource[F[_], +A] {
     new (F ~> F) {
       override def apply[B](gb: F[B]): F[B] = surround(gb)
     }
+
+  def forceR[B](that: Resource[F, B])(implicit F: MonadCancel[F, Throwable]): Resource[F, B] =
```
vasilmkd, line 488: @djspiewak We don't have this for IO.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -722,6 +896,93 @@ object Resource extends ResourceFOInstances0 with ResourceHOInstances0 with Reso
...
+      }
+    }
+
+  def executionContext[F[_]](implicit F: Async[F]): Resource[F, ExecutionContext] =
+    Resource.eval(F.executionContext)
+
+  def raiseError[F[_], A, E](e: E)(implicit F: ApplicativeError[F, E]): Resource[F, A] =
+    Resource.eval(F.raiseError[A](e))
+
+  def empty[F[_], A](implicit A: Monoid[A]): Resource[F, A] = Resource.pure[F, A](A.empty)
```
vasilmkd: @djspiewak I think this one is confusing. Thoughts? We don't have it for `IO`.

vasilmkd:
I fixed the stack-safety of `attempt` and added unit tests as well.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
djspiewak: We should! We should also have `!>`

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
djspiewak: I'm kind of on the fence about this one. I'm okay with it for both `Resource` and `IO`, but devoid of context it's also a slightly odd name. Maybe we should pass on it for now.

djspiewak (approved):
Approved modulo `empty`. I think I agree with @vasilmkd that we should pull that one out. Everything else looks great though, and this unblocks 3.2!

vasilmkd (changes requested):
Let me make the last needed changes official:

- [x] Add `!>` which is the symbolic name of `forceR`.
- [x] Remove `Resource.empty` and move the implementation to `Monoid[Resource[F, *], A]`

And we should be done. Great work! 🎉

djspiewak:
I made the pair of changes. I think that's all we needed!

vasilmkd (approved):

PR #2135, opened 2021-07-21
[DO NOT MERGE] Run CI on JDK 17

vasilmkd:
https://contributors.scala-lang.org/t/scala-community-build-on-jdk-17/5153

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,14 +26,14 @@ jobs:
         java:
           - adopt@1.8
           - adopt@1.11
-          - adopt@1.16
+          - zulu@1.17
           - graalvm-ce-java8@21.1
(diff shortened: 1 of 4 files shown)


vasilmkd:
https://github.com/dwijnand/sbt-extras/issues/326

djspiewak:
We may want to upgrade to the vanilla launcher.

vasilmkd:
There seems to be no need for that, I followed the workaround posted in `sbt-extras` and added an empty `.jvmopts` file for the `scalafix` project which was enough to suppress the default ones added by the script itself, so we're fine. I'm sure `sbt-extras` will receive some love by the time JDK 17 final is out.

With that being said, we're ready for JDK 17!😄

PR #2140, opened 2021-07-22
Added conceptual documentation and general introduction

djspiewak:


docs/concepts.md
```diff
@@ -0,0 +1,265 @@
+---
+id: concepts
+title: Concepts
+---
+
+Cats Effect introduces a large number of concepts which, while very general and individually simple, can be seem foreign and intimidating if you're new to the space. This also brings along with it a set of terminology for describing concepts and scenarios which can be significantly different from that which is used in other asynchronous frameworks (such as Akka or Vert.x). This page is an attempt to describe the Cats Effect runtime at a high level, introducing the concepts and terminology which is often assumed in discussions about the various details.
```
rintcius: can ~~be~~ seem foreign and intimidating

docs/concepts.md
```diff
@@ -0,0 +1,265 @@
...
...that all of the above libraries have significant differences and conceptual mismatches with Cats Effect. Tokio and Vert.x are probably the closest counterparts, though both are more vertically integrated (meaning they take a more *framework* oriented approach), while Cats Effect prescriptively defines itself as a library which enables a broad composable ecosystem. Additionally, in many places, Cats Effect defines features and functionality which simply doesn't exist in these ecosystems. As an example, neither Akka (via `Future`), Netty, nor Vert.x have any support for asynchronous cancelation (also known as "interruption"), meaning that basic functionality such as timeouts and concurrent error handling can result in resource leaks in a running application. Another example of functionality mismatch is the fiber-aware work-stealing runtime, which is present in Tokio and (to a lesser extent) Akka, but not in Netty or Vert.x. As a final example, asynchronous tracing is present to a limited degree in Vert.x and Akka, but absent from all other framework, neither of these frameworks provide a version of this functionality which is performant enough for production use (unlike Cats Effect).
```
rintcius: but absent from all other framework, neither of these frameworks
->
but absent from all other framework**s**, **and** neither of these frameworks

rintcius (commented):
Just some typos I saw while reading part of it.

docs/concepts.md
```diff
@@ -0,0 +1,265 @@
...
+
+## "Asynchronous" vs "Concurrent" vs "Parallel"
+
+It is quite common in most imperative language contexts (including Scala) to conflate these three terms. "Asynchronous" in particular is often taken to be synonymous with "parallel", when in fact all three have a very distinct definition which becomes significant in frameworks like Cats Effect.
+
+### Asynchronous
+
+In particular, "asynchronous" is the opposite of "synchronous", and it returns to a manner in which a given effect produces a value. Synchronous effects are defined using `apply` (also `delay` or `blocking` or `interruptible`) and produce their results using `return`, or alternatively raise errors using `throw`. These are the familiar "sequential" type of effects:
```
vasilmkd:
```suggestion
In particular, "asynchronous" is the opposite of "synchronous", and it pertains to a manner in which a given effect produces a value. Synchronous effects are defined using `apply` (also `delay` or `blocking` or `interruptible`) and produce their results using `return`, or alternatively raise errors using `throw`. These are the familiar "sequential" type of effects:
```

vasilmkd: I _think_ `returns` does not correctly convey the meaning of the sentence, but I'm also not sure if `pertains` is the right choice either.

docs/concepts.md
```diff
@@ -0,0 +1,265 @@
...
+
+It is critical to note that nothing in the definition of "asynchronous" implies "parallel" or "simultaneous", nor does it negate the meaning of "sequential" (remember: all fibers are sequences of effects). "Asynchronous" simply means "produces values or errors using a callback rather than `return`/`throw`". It is an implementation detail of an effect, managed by a fiber, rather than a larger fundamental pattern to be designed around.
+
+### Concurrent
+
+In the parlance of Cats Effect, "concurrent" generally refers to two or more actions which are defined to be independent in their control flow. It is the opposite of "sequential", or rather, "sequential" implies that something *cannot* be "concurrent". Critically, it is possible for things that are "concurrent" to evaluate sequentially if the underlying runtime decides this is optimal, whereas actions which are sequential will always be evaluated one after the other.
```
vasilmkd:
```suggestion
In the language of Cats Effect, "concurrent" generally refers to two or more actions which are defined to be independent in their control flow. It is the opposite of "sequential", or rather, "sequential" implies that something *cannot* be "concurrent". Critically, it is possible for things that are "concurrent" to evaluate sequentially if the underlying runtime decides this is optimal, whereas actions which are sequential will always be evaluated one after the other.
```

vasilmkd: I honestly had to google what `parlance` means. It's definitely a sign that my vocabulary is poor, but it might also be a sign that it does not read well in documentation. 😄

docs/concepts.md
```diff
@@ -0,0 +1,265 @@
...
+
+Concurrency is often conflated with asynchronous execution due to the fact that, in practice, the *implementation* of concurrency often relies upon some mechanism for asynchronous evaluation. But as noted above, asynchrony is just that: an implementation detail, and one which says nothing about concurrent vs sequential semantics.
```
vasilmkd, line 149: As a fun addition, we could add that literally every fiber in Cats Effect is executed asynchronously, even if it was written using only synchronous combinators. Just a thought. 😄

docs/concepts.md
```diff
@@ -0,0 +1,265 @@
...
+
+All concurrency in Cats Effect is implemented in terms of underlying primitives which create and manipulate fibers: `start` and `join`. These concurrency primitives are very similar to the equivalently-named operations on `Thread`, but as with most things in Cats Effect, they are considerably faster and safer.
+
+#### Structured Concurrency
+
+Formally-speaking, structured concurrency is a form of control flow in which all concurrent operations must form a closed hierarchy. Conceptually, it means that any operation which *forks* some actions to run concurrently must forcibly ensure that those actions are completed before moving forward. Furthermore, the *results* of a concurrent operation must only be made available upon its completion, and only to its parent in the hierarchy. `parTupled` above is a simple example of this: the `IO[(Response, Response)]` is unavailable as a result until *both* service calls have completed, and those responses are only accessible within the resulting tuple.
```
vasilmkd, line 163: Since _fork_ has not been used up to here, should we add a (`start` in Cats Effect) to maybe link the general concept to the actual implementation in CE?

docs/concepts.md
```diff
@@ -0,0 +1,265 @@
...
+
+Cats Effect has a large number of structured concurrency tools, most notably `parTupled`, `parMapN`, and `parTraverse`. Additionally, it offers a number of more robust structured concurrency operators such as `background`, `Supervisor`, and `Dispatcher`. It has also fostered an ecosystem wherein structured concurrency is the *rule* rather than the exception, particularly with the help of higher-level frameworks such as [Fs2](https://fs2.io). However, structured concurrency can be very limiting, and Cats Effect does not *prevent* unstructured concurrency when it is needed.
```
vasilmkd, line 165:
```suggestion
Cats Effect has a large number of structured concurrency tools, most notably `parTupled`, `parMapN`, and `parTraverse`. Additionally, it offers a number of more robust structured concurrency operators such as `background`, `cats.effect.std.Supervisor`, and `cats.effect.std.Dispatcher`. It has also fostered an ecosystem wherein structured concurrency is the *rule* rather than the exception, particularly with the help of higher-level frameworks such as [Fs2](https://fs2.io). However, structured concurrency can be very limiting, and Cats Effect does not *prevent* unstructured concurrency when it is needed.
```

docs/concepts.md
```diff
@@ -0,0 +1,265 @@
...
+  val printer: F[Unit] = Console[F].println(str)
+  (printer >> printer).as(str)
+}
+```
+
+In the above, `example` is an effectful function, and `printer` is an effect (as is `printer >> printer` and `(printer >> printer).as(str)`). The effect *type* here is `F`, which might be `IO`, but it also might be something more interesting! The caller of `example` is free to choose the effect at the call site, for example by writing something like `example[IO]("Hello, World")`, which in turn would return an `IO[String]`. Much of the Cats Effect ecosystem's composability is achieved using this technique under the surface.
+
+### Side-Effects
+
+When running a piece of code causes changes outside of just returning a value, we generally say that code "has side-effects". More intuitively, code where you care whether it runs more than once, and/or *when* it runs, almost always has side-effects. The classic example of this is `System.out.println`:
```
vasilmkd, line 235: Should we stick to idiomatic Scala and `println`?

vasilmkd (commented):
Amazing read. Thank you so much for this.

docs/concepts.md
djspiewak: "Pertains" is exactly what I was going for. What's funny is I re-read this paragraph twice before pushing and I think my brain auto-corrected the word each time.

docs/concepts.md
djspiewak: "Language" is probably a better word. :-) I think "parlance" isn't commonly known even among native speakers.

docs/concepts.md
djspiewak: :-D I thought about it, but I figured that would just confuse people. I want to bring readers to a place of not worrying about such details.

docs/concepts.md
djspiewak: I struggled a lot with this paragraph. "Fork" in particular is difficult, because it's hard to define the concept without being circular. I'm not sure the implementation really helps, but I do agree this needs re-wording.

docs/concepts.md
vasilmkd: Agreed.

docs/concepts.md
djspiewak: Counter-point: doesn't everyone have either an IDE, Metals, or something like Dash? Do we really need the fully-qualified names?

docs/concepts.md
djspiewak: I wanted to differentiate it explicitly from `IO.println`.

docs/concepts.md
vasilmkd: It's just that they cannot be found under the `cats.effect` package, unlike `Ref` and `Deferred` for example. It was just a thought, I don't really mind either way.

vasilmkd:
TIL, commit co-authoring. Handy.

djspiewak:
> TIL, commit co-authoring. Handy.

Says the guy who figured out how to co-author all of his tracing commits with @RaasAhsan…

docs/concepts.md
vasilmkd: I thought about that too. I agree.

vasilmkd:
I did not actually, they are all authored solely by @RaasAhsan. I only committed them (even though in reality I also wrote them, because they needed a lot of changes since the underlying code had moved forward so much).

vasilmkd:
I used `git commit --amend --author "Author Name <author@name.com>"`.

vasilmkd (approved):

PR #2143, opened 2021-07-24
Grow and shrink the callback hashtable

vasilmkd:
Recently I've done a lot of interview preparations and I was revising algorithms and data structures, so I figured why not fiddle with the custom hashtable that we have, make it slightly better.


djspiewak:
I was initially looking at this with an eye towards performance but, honestly, I'm not sure it's going to matter much. :-) In the normal case, there is probably only a single error callback. Additionally, we don't really hit this code path very often in that case anyway.

However, in the event that someone is doing a *lot* of `IO#unsafeRunSync` instead of using `Dispatcher`, this will definitely be meaningful. So tldr I think it's cool. :-)

PR #2147, opened 2021-07-24
Adjust tracing to use `IORuntimeConfig`

### LINKED ISSUE #2144: Add tracing configuration to `IORuntimeConfig`
Probably every configurable aspect except `cats.effect.stackTracingMode` which we need to be statically available at JVM startup.

djspiewak:
Fixes #2144

### DIFF
--- a/core/jvm/src/main/java/cats/effect/tracing/TracingConstants.java
+++ b/core/jvm/src/main/java/cats/effect/tracing/TracingConstants.java
@@ -24,7 +24,7 @@ class TracingConstants {
    * Sets stack tracing mode for a JVM process, which controls how much stack
    * trace information is captured. Acceptable values are: NONE, CACHED, FULL.
    */
-  private static final String stackTracingMode = Optional.ofNullable(System.getProperty("cats.effect.stackTracingMode"))
+  private static final String stackTracingMode = Optional.ofNullable(System.getProperty("cats.effect.tracing.mode"))
       .filter(x -> !x.isEmpty()).orElse("cached");
 
   static final boolean isCachedStackTracing = stackTracingMode.equalsIgnoreCase("cached");
--- a/core/jvm/src/main/java/cats/effect/tracing/TracingConstants.java
+++ b/core/jvm/src/main/java/cats/effect/tracing/TracingConstants.java
@@ -32,27 +32,4 @@ class TracingConstants {
   static final boolean isFullStackTracing = stackTracingMode.equalsIgnoreCase("full");
 
   static final boolean isStackTracing = isFullStackTracing || isCachedStackTracing;
-
-  /**
-   * The number of trace lines to retain during tracing. If more trace lines are
-   * produced, then the oldest trace lines will be discarded. Automatically
-   * rounded up to the nearest power of 2.
-   */
-  static final int traceBufferLogSize = Optional.ofNullable(System.getProperty("cats.effect.traceBufferLogSize"))
-      .filter(x -> !x.isEmpty()).flatMap(x -> {
-        try {
-          return Optional.of(Integer.valueOf(x));
-        } catch (Exception e) {
-          return Optional.empty();
-        }
-      }).orElse(4);
-
-  /**
-   * Sets the enhanced exceptions flag, which controls whether or not the stack
-   * traces of IO exceptions are augmented to include async stack trace
-   * information. Stack tracing must be enabled in order to use this feature. This
-   * flag is enabled by default.
-   */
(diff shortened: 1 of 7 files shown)


core/jvm/src/main/java/cats/effect/tracing/TracingConstants.java
```diff
@@ -24,35 +24,12 @@
    * Sets stack tracing mode for a JVM process, which controls how much stack
    * trace information is captured. Acceptable values are: NONE, CACHED, FULL.
    */
-  private static final String stackTracingMode = Optional.ofNullable(System.getProperty("cats.effect.stackTracingMode"))
+  private static final String stackTracingMode = Optional.ofNullable(System.getProperty("cats.effect.tracing.mode"))
       .filter(x -> !x.isEmpty()).orElse("cached");
```
vasilmkd, line 28: I did not know this previously, but the config parameters in `IORuntimeConfig` may fail with parsing errors. Should we do the same for `stackTracingMode` as well?

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeConfig.scala
```diff
@@ -19,15 +19,47 @@ package unsafe
...
+    val autoYieldThreshold: Int,
+    val enhancedExceptions: Boolean,
+    val traceBufferSize: Int) {
+
+  def this(cancelationCheckThreshold: Int, autoYieldThreshold: Int) =
+    this(
+      cancelationCheckThreshold,
+      autoYieldThreshold,
+      IORuntimeConfig.DefaultEnhancedExceptions,
+      IORuntimeConfig.DefaultTraceBufferSize)
```
vasilmkd, line 31: Yeah, I figured this would be the way to go.

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeConfig.scala
```diff
@@ -19,15 +19,57 @@ package unsafe
...
+
+  private[effect] val traceBufferLogSize: Int =
+    Math.round(Math.log(traceBufferSize.toDouble) / Math.log(2)).toInt
```
vasilmkd: Should we be using the `math` from `scala.math`?

vasilmkd (commented):
I like the changes to the names of the constants.

I guess this still needs the constructor/copy shenaningans for Scala.js.

core/jvm/src/main/java/cats/effect/tracing/TracingConstants.java
djspiewak: Technically we should probably fail a bit more gracefully in the config parameter cases. Wouldn't mind incorporating that fix into this PR

vasilmkd: Please do if you can.

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeConfig.scala
djspiewak: We shouldn't! When you have some time, you should run a disassembler (or just JMH) on `scala.math` invocations. The `java.lang.Math` functions are all intrinsics, while the `scala.math` functions are *not*, meaning that they're thousands of times faster than the `scala.math` equivalents.

vasilmkd: Awesome, thanks.

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeConfigCompanionPlatform.scala
```diff
@@ -17,16 +17,27 @@
...
 
+import scala.util.Try
+
 private[unsafe] abstract class IORuntimeConfigCompanionPlatform { this: IORuntimeConfig.type =>
 
+  // TODO make the cancelation and auto-yield properties have saner names
   protected final val Default: IORuntimeConfig = {
     val cancelationCheckThreshold =
-      System.getProperty("cats.effect.cancelation.check.threshold", "512").toInt
+      Try(System.getProperty("cats.effect.cancelation.check.threshold").toInt).getOrElse(512)
```
vasilmkd, line 27: My favorite stdlib monad.

djspiewak: "monad"

vasilmkd: Exactly.

vasilmkd (approved):

PR #2150, opened 2021-07-24
Add chrome to CI matrix

armanbilge:
Let's see what happens.


build.sbt
```diff
@@ -123,13 +126,20 @@ lazy val unidoc213 = taskKey[Seq[File]]("Run unidoc but only on Scala 2.13")
 lazy val useFirefoxEnv =
   settingKey[Boolean]("Use headless Firefox (via geckodriver) for running tests")
 Global / useFirefoxEnv := false
+lazy val useChromeEnv =
+  settingKey[Boolean]("Use headless Chrome (via geckodriver) for running tests")
+Global / useChromeEnv := false
```
djspiewak: Random bikeshedding: would it be better if we had a `seleniumEnv = settingKey[Option[Env]]` where `Env` is either `Firefox` or `Chrome`?

build.sbt
```diff
@@ -278,16 +292,21 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
       ProblemFilters.exclude[MissingClassProblem]("cats.effect.AsyncPropagateCancelation$"),
       // introduced by #1913, striped fiber callback hashtable, changes to package private code
       ProblemFilters.exclude[MissingClassProblem]("cats.effect.unsafe.FiberErrorHashtable"),
-      ProblemFilters.exclude[IncompatibleResultTypeProblem]("cats.effect.unsafe.IORuntime.fiberErrorCbs"),
+      ProblemFilters.exclude[IncompatibleResultTypeProblem](
+        "cats.effect.unsafe.IORuntime.fiberErrorCbs"),
```
djspiewak: Can we do the scalafmt in a separate commit?

build.sbt
armanbilge: Definitely!

build.sbt
armanbilge: Oops, sure.

build.sbt
armanbilge: Using a `String` seemed easier than defining `Env` (I tried this and SBT balked, I think maybe we need to add a package declaration to `build.sbt`?) although I would certainly prefer the latter.

.github/workflows/ci.yml
```diff
@@ -54,6 +54,18 @@ jobs:
...
+          - ci: ciChrome
+            java: adopt@1.16
+          - ci: ciChrome
+            java: graalvm-ce-java8@21.1
+          - os: windows-latest
+            scala: 3.0.1
+          - os: windows-latest
+            scala: 2.12.14
+          - os: windows-latest
+            ci: ciChrome
```
vasilmkd, line 68: I think we don't test Scala.js on Windows, so I vote for dropping these combinations.

build.sbt
```diff
@@ -1,3 +1,4 @@
+import org.openqa.selenium.remote.server.DriverFactory
```
vasilmkd: Can you please move this import below the license header and group it with the other ones.

build.sbt
```diff
@@ -121,19 +128,36 @@ ThisBuild / githubWorkflowBuildMatrixExclusions ++= Seq(
 
 lazy val unidoc213 = taskKey[Seq[File]]("Run unidoc but only on Scala 2.13")
 
-lazy val useFirefoxEnv =
-  settingKey[Boolean]("Use headless Firefox (via geckodriver) for running tests")
-Global / useFirefoxEnv := false
+lazy val useBrowserEnv =
+  settingKey[Option[String]]("Use headless browser (via geckodriver) for running tests")
```
vasilmkd: `geckodriver` is the Firefox browser driver. Just drop the `(via geckodriver)` comment.

vasilmkd (changes requested):
Looks good!

build.sbt
vasilmkd: Maybe sbt needs a Scala source file in `project` which can define the `Env` trait and then it can be imported in the main `build.sbt`?

armanbilge: I can try that. Btw I think `Browser` is preferable to `Env`?

build.sbt
armanbilge: Oops, yup. auto-import must have snuck that in there.

build.sbt
armanbilge: Didn't realize this :) makes sense

build.sbt
armanbilge: Or if we do `Env` then we shouldn't make it an `Option` and instead include `NodeJS` as a possible `Env`?

vasilmkd: That sounds cool.

build.sbt
vasilmkd: All good. 😄

djspiewak (dismissed):
Very nice! Needs conflict resolution but otherwise this is awesome

**Edit** Oh it's already resolved. Okay perfect. :-D

build.sbt
```diff
@@ -122,22 +131,37 @@ ThisBuild / githubWorkflowBuildMatrixExclusions ++= Seq(
...
-    options.setProfile(profile)
-    options.addArguments("-headless")
-    new SeleniumJSEnv(options)
-  } else {
-    old
+  useJSEnv.value match {
+    case NodeJS => old
+    case Firefox =>
+      val profile = new FirefoxProfile()
+      profile.setPreference("privacy.file_unique_origin", false)
```
armanbilge, line 146: We need to find the Chrome equivalent of this.

armanbilge:
Sorry don't merge yet!! Need to get webworker going on Chrome.

build.sbt
```diff
@@ -122,22 +131,38 @@ ThisBuild / githubWorkflowBuildMatrixExclusions ++= Seq(
...
+      val options = new FirefoxOptions()
+      options.setProfile(profile)
+      options.addArguments("-headless")
+      new SeleniumJSEnv(options)
+    case Chrome =>
+      val options = new ChromeOptions()
+      options.setHeadless(true)
+      options.addArguments("--allow-file-access-from-files")
```
armanbilge, line 154: that was easy!

armanbilge:
Would you accept a PR disabling `sbt --client`?

build.sbt
vasilmkd: https://dev.to/dengel29/loading-local-files-in-firefox-and-chrome-m9f

vasilmkd: https://www.selenium.dev/selenium/docs/api/java/org/openqa/selenium/chromium/ChromiumOptions.html#addArguments(java.util.List)

vasilmkd (approved):

vasilmkd:
Very cool work.

djspiewak (approved):

PR #2157, opened 2021-07-25
Kleisli `parTraverse` CE2

vasilmkd:


djspiewak:
The build is… green?

vasilmkd:
Check the second commit, but it's fake, it doesn't work with all derivations.

djspiewak (approved):

PR #2159, opened 2021-07-26
Parallel performance improvement

### LINKED ISSUE #2095: [CE3] Kleisli Stack safety

### LINKED ISSUE #2155: Improve performance of Parallel by overriding map2 to avoid tupling
This is a bit of low-hanging fruit, but we currently define `Parallel[F]` given `GenSpawn[F, _]` in terms of the following construction:

```scala
      final override def map2[A, B, Z](fa: ParallelF[F, A], fb: ParallelF[F, B])(
          f: (A, B) => Z): ParallelF[F, Z] =
        ParallelF(
          F.both(ParallelF.value(fa), ParallelF.value(fb)).map { case (a, b) => f(a, b) }
        )
```

That's cool but there's a lot of unnecessary boxing in here, both in `both` and in `racePair` underneath. We can actually do a lot better given `GenConcurrent[F, _]`, which will be the case quite a lot of the time in practical code. However, rather than creating some sort of sneaky implicit prioritization thing, I think the right approach here is probably to add a `mapBoth` function (feel free to bikeshed the name) to `GenSpawn`, then override it in `GenConcurrent` with a more efficient implementation in terms of `Fiber` and `Deferred`.

Overall this probably sounds a lot more intimidating of an enhancement than it actually is. Tldr, add the following to `GenSpawn`:

```scala
def mapBoth[A, B, C](fa: F[A], fb: F[B])(f: (A, B) => F[C]): F[C] = ???
```

Delegate the implementation to `both` in `GenSpawn`, but then override that implementation in `GenConcurrent` with something implemented in terms of `Fiber` and `Deferred`. The results should be a noticeable improvement in performance of the `par` operators.

manufacturist:
First PR. Be gentle 🙏 🎉  PR covering #2155:
- [x] ~mapBoth~
- [x] ~map2 in terms of mapBoth~
- [x] ~GenConcurrent.mapBoth override~
- [x] `map2` & `map2Eval` overrides (with `mapBoth` solution)
- [x] ~unit tests~ (No longer needed due to overriding `map2` & `map2Eval`)
- [x] benchmarks + profiling results

> Delegate the implementation to both in GenSpawn, but then override that implementation in GenConcurrent with something implemented in terms of Fiber and Deferred. The results should be a noticeable improvement in performance of the par operators.

In abstract, I need to ensure that two effects run in parallel, obtain their results and map over them.

Qs:
1. The `cancelled *> never` trick ensures that `never` will never run, therefore it is only used throughout the cats-effect code to help with matching the type signature where needed. Is this correct?
2. When joining a fiber in a construct such as `outcome <- poll(workFiber.join)`, is there any need for calling `onCancel` or `guarantee`? Is there the possibility of having the main fiber cancelled while the `workFiber` hasn't joined yet?  And if yes, then the mentioned constructs guards us against such scenarios?
3. Why would the solution need to be implemented in terms of fibers AND `Deferred`? Why isn't fibers enough, e.g.
```scala
  override def mapBoth[A, B, C](fa: F[A], fb: F[B])(f: (A, B) => F[C]): F[C] = {
    implicit val F: GenConcurrent[F, E] = this

    uncancelable { poll =>
      for {
        fiberA <- start(fa)
        fiberB <- start(fb)

        outcomeA <- poll(fiberA.join).guarantee(fiberA.cancel)
        outcomeB <- poll(fiberB.join).guarantee(fiberB.cancel)

        resultC <- (outcomeA, outcomeB) match {
          case (Outcome.Succeeded(fa), Outcome.Succeeded(fb)) =>
            fa.product(fb).flatMap { case (a, b) => f(a, b) }

          case (Outcome.Errored(e), _) => raiseError(e)
          case (_, Outcome.Errored(e)) => raiseError(e)
          case (Outcome.Canceled(), _) => F.canceled *> never
          case (_, Outcome.Canceled()) => F.canceled *> never
        }
      } yield resultC
     }
    }
```

### DIFF
--- a/benchmarks/src/main/scala/cats/effect/benchmarks/ParallelBenchmark.scala
+++ b/benchmarks/src/main/scala/cats/effect/benchmarks/ParallelBenchmark.scala
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2020-2021 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.benchmarks
+
+import cats.effect.IO
+import cats.effect.unsafe.implicits.global
+import cats.implicits.{catsSyntaxParallelTraverse1, toTraverseOps}
+
+import org.openjdk.jmh.annotations._
+import org.openjdk.jmh.infra.Blackhole
+
+import java.util.concurrent.TimeUnit
+
+/**
+ * To do comparative benchmarks between versions:
+ *
+ *     benchmarks/run-benchmark ParallelBenchmark
+ *
+ * This will generate results in `benchmarks/results`.
+ *
+ * Or to run the benchmark from within sbt:
+ *
+ *     jmh:run -i 10 -wi 10 -f 2 -t 4 cats.effect.benchmarks.ParallelBenchmark
+ *
+ * Which means "10 iterations", "10 warm-up iterations", "2 forks", "4 thread".
+ * Please note that benchmarks should be usually executed at least in
+ * 10 iterations (as a rule of thumb), but more is better.
+ */
+@State(Scope.Thread)
+@BenchmarkMode(Array(Mode.Throughput))
+@OutputTimeUnit(TimeUnit.SECONDS)
+class ParallelBenchmark {
+
+  @Param(Array("100", "1000", "10000"))
+  var size: Int = _
+
+  @Param(Array("100", "1000", "10000", "100000", "1000000"))
+  var cpuTokens: Long = _
+
+  @Benchmark
+  def parTraverse(): Unit =
+    1.to(size).toList.parTraverse(_ => IO(Blackhole.consumeCPU(cpuTokens))).void.unsafeRunSync()
+
+  @Benchmark
+  def traverse(): Unit =
+    1.to(size).toList.traverse(_ => IO(Blackhole.consumeCPU(cpuTokens))).void.unsafeRunSync()
+}
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
@@ -125,30 +125,6 @@ trait GenConcurrent[F[_], E] extends GenSpawn[F, E] {
       }
     }
   }
-
-  override def mapBoth[A, B, C](fa: F[A], fb: F[B])(f: (A, B) => F[C]): F[C] = {
-    implicit val F: GenConcurrent[F, E] = this
-
-    uncancelable { poll =>
-      for {
-        fiberA <- start(fa)
-        fiberB <- start(fb)
-
-        outcomeA <- poll(fiberA.join).onCancel(fiberB.cancel)
-        outcomeB <- poll(fiberB.join)
-
-        resultC <- (outcomeA, outcomeB) match {
-          case (Outcome.Succeeded(fa), Outcome.Succeeded(fb)) =>
-            fa.flatMap(a => fb.flatMap(b => f(a, b)))
-
-          case (Outcome.Errored(e), _) => raiseError(e)
-          case (_, Outcome.Errored(e)) => raiseError(e)
-          case (Outcome.Canceled(), _) => F.canceled *> never
-          case (_, Outcome.Canceled()) => F.canceled *> never
-        }
-      } yield resultC
-    }
-  }
 }
 
 object GenConcurrent {
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
@@ -131,21 +131,15 @@ trait GenConcurrent[F[_], E] extends GenSpawn[F, E] {
 
     uncancelable { poll =>
       for {
-        resultA <- deferred[Outcome[F, E, A]]
-        resultB <- deferred[Outcome[F, E, B]]
+        fiberA <- start(fa)
+        fiberB <- start(fb)
 
-        fiberA <- start(guaranteeCase(fa)(outcome => resultA.complete(outcome).void))
-        fiberB <- start(guaranteeCase(fb)(outcome => resultB.complete(outcome).void))
-
-        _ <- poll(fiberA.join)
-        _ <- poll(fiberB.join)
-
-        outcomeA <- poll(resultA.get)
-        outcomeB <- poll(resultB.get)
+        outcomeA <- poll(fiberA.join).onCancel(fiberB.cancel)
+        outcomeB <- poll(fiberB.join)
 
         resultC <- (outcomeA, outcomeB) match {
           case (Outcome.Succeeded(fa), Outcome.Succeeded(fb)) =>
-            fa.product(fb).flatMap { case (a, b) => f(a, b) }
+            fa.flatMap(a => fb.flatMap(b => f(a, b)))
 
           case (Outcome.Errored(e), _) => raiseError(e)
           case (_, Outcome.Errored(e)) => raiseError(e)
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
@@ -125,6 +125,36 @@ trait GenConcurrent[F[_], E] extends GenSpawn[F, E] {
       }
     }
   }
+
+  override def mapBoth[A, B, C](fa: F[A], fb: F[B])(f: (A, B) => F[C]): F[C] = {
+    implicit val F: GenConcurrent[F, E] = this
+
+    uncancelable { poll =>
+      for {
+        resultA <- deferred[Outcome[F, E, A]]
+        resultB <- deferred[Outcome[F, E, B]]
+
+        fiberA <- start(guaranteeCase(fa)(outcome => resultA.complete(outcome).void))
+        fiberB <- start(guaranteeCase(fb)(outcome => resultB.complete(outcome).void))
+
+        _ <- poll(fiberA.join)
+        _ <- poll(fiberB.join)
+
+        outcomeA <- poll(resultA.get)
+        outcomeB <- poll(resultB.get)
+
+        resultC <- (outcomeA, outcomeB) match {
+          case (Outcome.Succeeded(fa), Outcome.Succeeded(fb)) =>
+            fa.product(fb).flatMap { case (a, b) => f(a, b) }
+
+          case (Outcome.Errored(e), _) => raiseError(e)
+          case (_, Outcome.Errored(e)) => raiseError(e)
+          case (Outcome.Canceled(), _) => F.canceled *> never
+          case (_, Outcome.Canceled()) => F.canceled *> never
+        }
+      } yield resultC
+    }
+  }
 }
 
 object GenConcurrent {
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
@@ -434,26 +434,6 @@ trait GenSpawn[F[_], E] extends MonadCancel[F, E] with Unique[F] {
           }
       }
     }
-
-  def mapBoth[A, B, C](fa: F[A], fb: F[B])(f: (A, B) => C): F[C] =
-    uncancelable { poll =>
-      for {
-        fiberA <- start(fa)
-        fiberB <- start(fb)
-
-        a <- poll(fiberA.join).onCancel(fiberB.cancel).flatMap[A] {
-          case Outcome.Succeeded(fa) => fa
-          case Outcome.Errored(e) => fiberB.cancel *> raiseError(e)
-          case Outcome.Canceled() => fiberB.cancel *> never
-        }
-
-        c <- poll(fiberB.join).flatMap[C] {
-          case Outcome.Succeeded(fb) => fb.map(b => f(a, b))
-          case Outcome.Errored(e) => raiseError(e)
-          case Outcome.Canceled() => canceled *> never
-        }
-      } yield c
-    }
 }
 
 object GenSpawn {
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
@@ -435,8 +435,26 @@ trait GenSpawn[F[_], E] extends MonadCancel[F, E] with Unique[F] {
       }
     }
 
-  def mapBoth[A, B, C](fa: F[A], fb: F[B])(f: (A, B) => F[C]): F[C] =
-    both(fa, fb).flatMap { case (a, b) => f(a, b) }
+  def mapBoth[A, B, C](fa: F[A], fb: F[B])(f: (A, B) => C): F[C] =
+    uncancelable { poll =>
+      for {
+        fiberA <- start(fa)
+        fiberB <- start(fb)
+
+        outcomeA <- poll(fiberA.join).onCancel(fiberB.cancel)
+        outcomeB <- poll(fiberB.join)
+
+        resultC <- (outcomeA, outcomeB) match {
+          case (Outcome.Succeeded(fa), Outcome.Succeeded(fb)) =>
+            fa.flatMap(a => fb.map(b => f(a, b)))
+
+          case (Outcome.Errored(e), _) => raiseError(e)
+          case (_, Outcome.Errored(e)) => raiseError(e)
+          case (Outcome.Canceled(), _) => F.canceled *> never
+          case (_, Outcome.Canceled()) => F.canceled *> never
+        }
+      } yield resultC
+    }
 }
 
 object GenSpawn {
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
@@ -526,13 +526,6 @@ object GenSpawn {
     override def both[A, B](fa: OptionT[F, A], fb: OptionT[F, B]): OptionT[F, (A, B)] =
       OptionT(F.both(fa.value, fb.value).map(_.tupled))
 
-    override def mapBoth[A, B, C](fa: OptionT[F, A], fb: OptionT[F, B])(
-        f: (A, B) => OptionT[F, C]): OptionT[F, C] =
-      OptionT(F.mapBoth(fa.value, fb.value) {
-        case (Some(a), Some(b)) => f(a, b).value
-        case (_, _) => OptionT.none[F, C].value
-      })
-
     override def raceOutcome[A, B](fa: OptionT[F, A], fb: OptionT[F, B])
         : OptionT[F, Either[Outcome[OptionT[F, *], E, A], Outcome[OptionT[F, *], E, B]]] =
       OptionT.liftF(
@@ -596,14 +589,6 @@ object GenSpawn {
         fa: EitherT[F, E0, A],
         fb: EitherT[F, E0, B]): EitherT[F, E0, (A, B)] =
       EitherT(F.both(fa.value, fb.value).map(_.tupled))
-
-    override def mapBoth[A, B, C](fa: EitherT[F, E0, A], fb: EitherT[F, E0, B])(
-        f: (A, B) => EitherT[F, E0, C]): EitherT[F, E0, C] =
-      EitherT(F.mapBoth(fa.value, fb.value) {
-        case (Right(a), Right(b)) => f(a, b).value
-        case (Left(e), _) => EitherT.leftT[F, C](e).value
-        case (_, Left(e)) => EitherT.leftT[F, C](e).value
-      })
 
     override def raceOutcome[A, B](fa: EitherT[F, E0, A], fb: EitherT[F, E0, B]): EitherT[
       F,
@@ -670,19 +655,6 @@ object GenSpawn {
     override def both[A, B](fa: IorT[F, L, A], fb: IorT[F, L, B]): IorT[F, L, (A, B)] =
       IorT(F.both(fa.value, fb.value).map(_.tupled))
 
-    override def mapBoth[A, B, C](fa: IorT[F, L, A], fb: IorT[F, L, B])(
-        f: (A, B) => IorT[F, L, C]): IorT[F, L, C] =
-      IorT(F.mapBoth(fa.value, fb.value) {
-        case (Ior.Right(a), Ior.Right(b)) => f(a, b).value
-        case (Ior.Both(_, a), Ior.Both(_, b)) => f(a, b).value
-        case (Ior.Both(_, a), Ior.Right(b)) => f(a, b).value
-        case (Ior.Right(a), Ior.Both(_, b)) => f(a, b).value
-        case (Ior.Left(l), _) => IorT.leftT[F, C](l).value
-        case (_, Ior.Left(l)) => IorT.leftT[F, C](l).value
-        case (Ior.Both(l, _), _) => IorT.leftT[F, C](l).value
-        case (_, Ior.Both(l, _)) => IorT.leftT[F, C](l).value
-      })
-
     override def raceOutcome[A, B](fa: IorT[F, L, A], fb: IorT[F, L, B])
         : IorT[F, L, Either[Outcome[IorT[F, L, *], E, A], Outcome[IorT[F, L, *], E, B]]] =
       IorT.liftF(F.raceOutcome(fa.value, fb.value).map(_.bimap(liftOutcome(_), liftOutcome(_))))
@@ -744,10 +716,6 @@ object GenSpawn {
 
     override def both[A, B](fa: Kleisli[F, R, A], fb: Kleisli[F, R, B]): Kleisli[F, R, (A, B)] =
       Kleisli { r => F.both(fa.run(r), fb.run(r)) }
-
-    override def mapBoth[A, B, C](fa: Kleisli[F, R, A], fb: Kleisli[F, R, B])(
-        f: (A, B) => Kleisli[F, R, C]): Kleisli[F, R, C] =
-      Kleisli { r => F.mapBoth(fa.run(r), fb.run(r)) { case (a, b) => f(a, b)(r) } }
 
     override def raceOutcome[A, B](fa: Kleisli[F, R, A], fb: Kleisli[F, R, B]): Kleisli[
       F,
@@ -818,12 +786,6 @@ object GenSpawn {
 
     override def both[A, B](fa: WriterT[F, L, A], fb: WriterT[F, L, B]): WriterT[F, L, (A, B)] =
       WriterT(F.both(fa.run, fb.run).map { case ((l1, a), (l2, b)) => (l1 |+| l2) -> (a -> b) })
-
-    override def mapBoth[A, B, C](fa: WriterT[F, L, A], fb: WriterT[F, L, B])(
-        f: (A, B) => WriterT[F, L, C]): WriterT[F, L, C] =
-      WriterT(F.mapBoth(fa.run, fb.run) {
-        case ((l1, a), (l2, b)) => f(a, b).map((l1 |+| l2) -> _).value
-      })
 
     override def raceOutcome[A, B](fa: WriterT[F, L, A], fb: WriterT[F, L, B]): WriterT[
       F,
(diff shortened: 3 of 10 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -523,6 +526,13 @@ object GenSpawn {
     override def both[A, B](fa: OptionT[F, A], fb: OptionT[F, B]): OptionT[F, (A, B)] =
       OptionT(F.both(fa.value, fb.value).map(_.tupled))
 
+    override def mapBoth[A, B, C](fa: OptionT[F, A], fb: OptionT[F, B])(
```
djspiewak: I actually don't think we need to override `mapBoth` for the transformers! Which is nice because all these overrides are annoying.

kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
```diff
@@ -125,6 +125,36 @@ trait GenConcurrent[F[_], E] extends GenSpawn[F, E] {
...
+
+        _ <- poll(fiberA.join)
+        _ <- poll(fiberB.join)
+
+        outcomeA <- poll(resultA.get)
+        outcomeB <- poll(resultB.get)
+
+        resultC <- (outcomeA, outcomeB) match {
+          case (Outcome.Succeeded(fa), Outcome.Succeeded(fb)) =>
+            fa.product(fb).flatMap { case (a, b) => f(a, b) }
```
djspiewak: We can avoid the tuple allocation here by using `fa.flatMap(a => fb.flatMap(b => f(a, b)))` rather than using `product`.

djspiewak (commented):
This is great! Just a couple adjustments. Also, @vasilmkd do we have a `parTraverse` benchmark already? We should add one and check the difference.

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
vasilmkd: Is this tested in the laws? If so, let's drop it.

vasilmkd:
Unfortunately I don't think we have a `parTraverse` benchmark. It would be great to have a suite of benchmarks for `traverse` and `parTraverse`, maybe it can be done in a sepate PR so that we can test before and after. Sorry about this.

djspiewak:
> Unfortunately I don't think we have a parTraverse benchmark. It would be great to have a suite of benchmarks for traverse and parTraverse, maybe it can be done in a sepate PR so that we can test before and after. Sorry about this.

No worries! I was thinking of just doing it as part of this PR actually. @manufacturist are you comfortable working with JMH? Would you like to give it a try? Otherwise I (or Vasil) can get to it and push here.

vasilmkd:
Feel free to ask and to look into the `benchmarks` sources and copy/paste liberally.

djspiewak:
Missed the questions! Answers incoming…

> The `cancelled *> never` trick ensures that never will never run, therefore it is only used throughout the cats-effect code to help with matching the type signature where needed. Is this correct?

The problem here is a little more existential. If it was just a bit of type juggling I would be a lot more worried about it! Here's the issue:

```scala
F.uncancelable(_ => F.both(F.canceled.as(42), F.unit))
```

What should this produce?

The problem here is the type signature of `both` is too limited to properly represent this scenario. We can't produce a value of type `(Int, Unit)`, and we can't cancel ourselves (because of `uncancelable`), so we have to either error or spin forever. The error approach doesn't work either though because it might be caught within the `uncancelable` block, etc etc.

> When joining a fiber in a construct such as outcome <- poll(workFiber.join), is there any need for calling onCancel or guarantee? Is there the possibility of having the main fiber cancelled while the workFiber hasn't joined yet? And if yes, then the mentioned constructs guards us against such scenarios?

Precisely as you said. :-) Also I'm realizing that there's actually a leak here: it's possible that the parent fiber is canceled prior to the first `poll`, in which case it will show up there but only one of the child fibers will be canceled! We need something a little trickier using `guaranteeCase` to ensure that *both* fibers get canceled if either of them error or cancel.

> Why would the solution need to be implemented in terms of fibers AND Deferred? Why isn't fibers enough, e.g.

I believe you're correct! I'm pretty sure we don't need `Deferred` at all here.

manufacturist:
> > Unfortunately I don't think we have a parTraverse benchmark. It would be great to have a suite of benchmarks for traverse and parTraverse, maybe it can be done in a sepate PR so that we can test before and after. Sorry about this.
> 
> No worries! I was thinking of just doing it as part of this PR actually. @manufacturist are you comfortable working with JMH? Would you like to give it a try? Otherwise I (or Vasil) can get to it and push here.

I'll give it a try 😄

djspiewak:
@manufacturist If it helps, I literally always copy/paste from something pre-existing for JMH. I recommend just copying something like the `BothBenchmark`, removing all the cases, and writing a new one for `parTraverse`. Call it `ParallelBenchmark` or something like that. The `parTraverse` benchmark can probably be pretty simple, just like traverse over a list of size `n` (where `n: Int` is a `@Param` and you can set it to like, `"10000"` or something). Anyway, just create a list of size `n` and then `parTraverse` over it to an `IO` which uses `BlackHole.consumeCPU()` with some small (but not totally trivial number). It might be useful to also add an identical `traverse` benchmark in the same file. We would expect that `parTraverse` would be *k* times faster than `n`, where *k* is the number of physical threads on the underlying machine.

vasilmkd:
Maintainers caught admitting copy/pasting code all over the place.

djspiewak:
> Maintainers caught admitting copy/pasting code all over the place.

My favorite Sublime key combo is Cmd-Shift-D: it duplicates the current line. :-P

manufacturist:
Updated the `mapBoth` override to drop `Deffered` from the solution & removed overrides for `mapBoth` in instances. Added the benchmarks for `parTraverse` and `traverse` only with `val cpuTimeTokens = 10000`.

### Executing Machine

```
Processor Name:         Quad-Core Intel Core i7
Processor Speed:        2,2 GHz
Number of Processors:   1
Total Number of Cores:  4
```

All benchmarks executed with variations of `jmh:run -i X -wi Y -f 2 -t 4 cats.effect.benchmarks.ParallelBenchmark`

### CPU Time Tokens = 200

```
[info] Benchmark                                  (size)   Mode  Cnt      Score     Error  Units
[info] ParallelBenchmark.traverse100               10000  thrpt   15  37910.533 ± 398.023  ops/s
[info] ParallelBenchmark.traverse1000              10000  thrpt   15   4742.708 ± 207.744  ops/s
[info] ParallelBenchmark.traverse10000             10000  thrpt   15    506.790 ±  25.914  ops/s
[info] ParallelBenchmark.parTraverseBoth100        10000  thrpt   15   3632.091 ± 354.683  ops/s
[info] ParallelBenchmark.parTraverseBoth1000       10000  thrpt   15    435.991 ±   8.477  ops/s
[info] ParallelBenchmark.parTraverseBoth10000.     10000  thrpt   15     55.030 ±   0.788  ops/s
[info] ParallelBenchmark.parTraverseMapBoth100     10000  thrpt   10   7824.942 ± 111.829  ops/s
[info] ParallelBenchmark.parTraverseMapBoth1000    10000  thrpt   10   1049.031 ± 151.450  ops/s
[info] ParallelBenchmark.parTraverseMapBoth10000   10000  thrpt   10    108.418 ±  13.729  ops/s
```

### CPU Time Tokens = 2000 - `mapBoth` approach

```
[info] Benchmark                           (size)   Mode  Cnt     Score     Error  Units
[info] ParallelBenchmark.parTraverse100     10000  thrpt    5  7291.586 ± 621.810  ops/s
[info] ParallelBenchmark.parTraverse1000    10000  thrpt    5   729.836 ± 143.587  ops/s
[info] ParallelBenchmark.parTraverse10000   10000  thrpt    5    62.743 ±   3.068  ops/s
[info] ParallelBenchmark.traverse100        10000  thrpt    5  6977.727 ± 218.018  ops/s
[info] ParallelBenchmark.traverse1000       10000  thrpt    5   755.024 ±  26.890  ops/s
[info] ParallelBenchmark.traverse10000      10000  thrpt    5    72.412 ±  10.234  ops/s
```

### CPU Time Tokens = 5000 - `mapBoth` approach

```
[info] ParallelBenchmark.parTraverse100     10000  thrpt    6  4741.184 ± 173.918  ops/s
[info] ParallelBenchmark.parTraverse1000    10000  thrpt    6   476.457 ±  74.728  ops/s
[info] ParallelBenchmark.parTraverse10000   10000  thrpt    6    38.966 ±   2.960  ops/s
[info] ParallelBenchmark.traverse100        10000  thrpt    6  2945.512 ±  55.656  ops/s
[info] ParallelBenchmark.traverse1000       10000  thrpt    6   298.424 ±  25.320  ops/s
[info] ParallelBenchmark.traverse10000      10000  thrpt    6    31.545 ±   0.906  ops/s
```

### CPU Time Tokens = 10000 - `mapBoth` approach

```
[info] ParallelBenchmark.parTraverse100     10000  thrpt    6  2726.039 ±  19.494  ops/s
[info] ParallelBenchmark.parTraverse1000    10000  thrpt    6   283.107 ±   2.026  ops/s
[info] ParallelBenchmark.parTraverse10000   10000  thrpt    6    25.774 ±   4.185  ops/s
[info] ParallelBenchmark.traverse100        10000  thrpt    6  1496.449 ± 147.759  ops/s
[info] ParallelBenchmark.traverse1000       10000  thrpt    6   151.205 ±   3.623  ops/s
[info] ParallelBenchmark.traverse10000      10000  thrpt    6    15.209 ±   1.546  ops/s
```

<hr>

How I interpret the benchmarks:
* Non-CPU intensive tasks result in bad `parTraverse` performance due to an occuring `performance cost` from internal CE operating mechanisms (it's a black box for me 😅)
* CPU intensive tasks are an excellent candidate for `parTraverse`
* The `mapBoth` solution provides a better performance than the old one using `both` (`GenConcurrent.mapBoth` not overriden)

benchmarks/src/main/scala/cats/effect/benchmarks/ParallelBenchmark.scala
```diff
@@ -0,0 +1,69 @@
...
+
+import org.openjdk.jmh.annotations._
+import org.openjdk.jmh.infra.Blackhole
+
+import java.util.concurrent.TimeUnit
+
+/**
+ * To do comparative benchmarks between versions:
+ *
+ *     benchmarks/run-benchmark BothBenchmark
```
vasilmkd: Please replace all occurrences of `BothBenchmark` with `ParallelBenchmark`.

kernel/shared/src/main/scala/cats/effect/kernel/instances/GenSpawnInstances.scala
```diff
@@ -53,7 +53,7 @@ trait GenSpawnInstances {
       final override def map2[A, B, Z](fa: ParallelF[F, A], fb: ParallelF[F, B])(
           f: (A, B) => Z): ParallelF[F, Z] =
         ParallelF(
-          F.both(ParallelF.value(fa), ParallelF.value(fb)).map { case (a, b) => f(a, b) }
+          F.mapBoth(ParallelF.value(fa), ParallelF.value(fb)) { case (a, b) => F.pure(f(a, b)) }
```
vasilmkd: I think the `case` is now unnecessary and can be dropped.

kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
```diff
@@ -125,6 +125,30 @@ trait GenConcurrent[F[_], E] extends GenSpawn[F, E] {
...
+            fa.flatMap(a => fb.flatMap(b => f(a, b)))
+
+          case (Outcome.Errored(e), _) => raiseError(e)
+          case (_, Outcome.Errored(e)) => raiseError(e)
+          case (Outcome.Canceled(), _) => F.canceled *> never
+          case (_, Outcome.Canceled()) => F.canceled *> never
+        }
+      } yield resultC
+    }
+  }
```
vasilmkd: Now that there are no `GenConcurrent` constructs used, we should try to move this code in `GenSpawn`. This was an observation by @TimWSpence, thank you!

vasilmkd: I feel this block of code needs a good review from more pairs of eyes.

vasilmkd (changes requested):

kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
vasilmkd: That being said, if the logic does end up changing, this code may indeed need to live in `GenConcurrent`.

vasilmkd:
@manufacturist Are the benchmark results you posted only for the current branch which includes the changes in this PR?

manufacturist:
> @manufacturist Are the benchmark results you posted only for the current branch which includes the changes in this PR?

They are. The first batch (`CPU Time Tokens = 200`) also includes results for the old implementation of `map2`.

vasilmkd:
I would like to split the benchmark code out of this PR, open it as a separate PR and then when we merge it, rebase this PR and do a proper comparison for every case between `series/3.x` and this branch.

vasilmkd:
@manufacturist The original issue has been changed in light of new information. I suggest that we split out the benchmarks into a separate PR and push that through, and then we can work together on tackling this problem again and be a bit more focused this time round.

djspiewak:
I think it's okay to keep updating this PR in light of the revised requirements. It's good history. :-) Agree that splitting the benchmarks out into a separate PR (#2167) is a good move.

vasilmkd:
Can you drop the commit that adds the `ParallelBenchmark`?

manufacturist:
> Can you drop the commit that adds the `ParallelBenchmark`?

Will do so in ~12 hours.

djspiewak:
@manufacturist New plan! Let's skip the `mapBoth` in `GenSpawn` and instead just put that same implementation into `map2` for `commutativeApplicativeForParallelF`. In fact, if you copy/paste the implementation into `map2Eval` in the same class, it should *also* fix #2095. This is great. :-D

@vasilmkd gets all the credit for this realization btw.

vasilmkd:
Plan of attack that @manufacturist can follow tomorrow:

- [ ] Drop the `ParallelBenchmark` commit and rebase on `series/3.x`, you'll get it from there.
- [ ] Reimplement `map2` in `commutativeApplicativeForParallelF`.
- [ ] You can attempt to override `map2Eval` in `commutativeApplicativeForParallelF` too, and then I will push the unit tests that we have prepared for that particular bug, or you can choose to ignore this, your call, we can also add that code to the already open PR

manufacturist:
Will rerun the benchmarks later today to be able to compare them with the ones from #2167.

manufacturist:
Finished running the benchmarks. They can be compared with the ones from the [other PR](https://github.com/typelevel/cats-effect/pull/2167#issuecomment-888030397):  

```sbt
[info] Benchmark                      (cpuTokens)  (size)   Mode  Cnt      Score     Error  Units
[info] ParallelBenchmark.parTraverse          100     100  thrpt   20   9944.533 ± 211.074  ops/s
[info] ParallelBenchmark.parTraverse          100    1000  thrpt   20   1428.851 ±   9.781  ops/s
[info] ParallelBenchmark.parTraverse          100   10000  thrpt   20    139.946 ±   1.383  ops/s
[info] ParallelBenchmark.parTraverse         1000     100  thrpt   20   8184.030 ±  39.369  ops/s
[info] ParallelBenchmark.parTraverse         1000    1000  thrpt   20   1008.832 ±   2.568  ops/s
[info] ParallelBenchmark.parTraverse         1000   10000  thrpt   20     98.693 ±   0.729  ops/s
[info] ParallelBenchmark.parTraverse        10000     100  thrpt   20   2526.056 ±  29.223  ops/s
[info] ParallelBenchmark.parTraverse        10000    1000  thrpt   20    257.168 ±   2.535  ops/s
[info] ParallelBenchmark.parTraverse        10000   10000  thrpt   20     25.164 ±   0.491  ops/s
[info] ParallelBenchmark.parTraverse       100000     100  thrpt   20    305.078 ±   1.061  ops/s
[info] ParallelBenchmark.parTraverse       100000    1000  thrpt   20     30.497 ±   0.589  ops/s
[info] ParallelBenchmark.parTraverse       100000   10000  thrpt   20      3.006 ±   0.084  ops/s
[info] ParallelBenchmark.parTraverse      1000000     100  thrpt   20     31.570 ±   0.133  ops/s
[info] ParallelBenchmark.parTraverse      1000000    1000  thrpt   20      3.164 ±   0.017  ops/s
[info] ParallelBenchmark.parTraverse      1000000   10000  thrpt   20      0.330 ±   0.034  ops/s
[info] ParallelBenchmark.traverse             100     100  thrpt   20  61728.601 ± 694.063  ops/s
[info] ParallelBenchmark.traverse             100    1000  thrpt   20   7589.424 ±  84.756  ops/s
[info] ParallelBenchmark.traverse             100   10000  thrpt   20    873.365 ±   2.606  ops/s
[info] ParallelBenchmark.traverse            1000     100  thrpt   20  15077.902 ±  89.639  ops/s
[info] ParallelBenchmark.traverse            1000    1000  thrpt   20   1622.568 ±  10.988  ops/s
[info] ParallelBenchmark.traverse            1000   10000  thrpt   20    165.695 ±   1.015  ops/s
[info] ParallelBenchmark.traverse           10000     100  thrpt   20   1770.700 ±   0.867  ops/s
[info] ParallelBenchmark.traverse           10000    1000  thrpt   20    177.990 ±   0.240  ops/s
[info] ParallelBenchmark.traverse           10000   10000  thrpt   20     17.709 ±   0.125  ops/s
[info] ParallelBenchmark.traverse          100000     100  thrpt   20    179.605 ±   0.153  ops/s
[info] ParallelBenchmark.traverse          100000    1000  thrpt   20     17.886 ±   0.097  ops/s
[info] ParallelBenchmark.traverse          100000   10000  thrpt   20      1.737 ±   0.067  ops/s
[info] ParallelBenchmark.traverse         1000000     100  thrpt   20     17.896 ±   0.069  ops/s
[info] ParallelBenchmark.traverse         1000000    1000  thrpt   20      1.729 ±   0.075  ops/s
[info] ParallelBenchmark.traverse         1000000   10000  thrpt   20      0.176 ±   0.008  ops/s
```

The `traverse` ones (`cpuTokens = 100`) are slightly worse possibly due to the fact that I ran the benchmarks immediately after a work day, whereas for the other ones I restarted my machine beforehand (The other run took 4h, this one took 4h20m).

For my machine, the performance gain for `parTraverse` is visible up until the `10k cpuTokens / 10k size` benchmark. Afterwards it kinda syncs with the other batch of results. I would've been happy (and curious) to provide some benchmarks with 8 threads, however my machine is limited in this situation 🤷

vasilmkd:
Here are the runs from a 16 core cloud instance:

`series/3.x`:
```
Benchmark                      (cpuTokens)  (size)   Mode  Cnt     Score    Error  Units
ParallelBenchmark.parTraverse          100     100  thrpt   10  1022.928 ± 97.351  ops/s
ParallelBenchmark.parTraverse          100    1000  thrpt   10   590.857 ± 17.301  ops/s
ParallelBenchmark.parTraverse          100   10000  thrpt   10   114.312 ±  1.927  ops/s
ParallelBenchmark.parTraverse         1000     100  thrpt   10   920.307 ± 61.221  ops/s
ParallelBenchmark.parTraverse         1000    1000  thrpt   10   541.315 ±  8.762  ops/s
ParallelBenchmark.parTraverse         1000   10000  thrpt   10    91.450 ±  9.160  ops/s
ParallelBenchmark.parTraverse        10000     100  thrpt   10   627.563 ± 11.182  ops/s
ParallelBenchmark.parTraverse        10000    1000  thrpt   10   282.210 ±  4.189  ops/s
ParallelBenchmark.parTraverse        10000   10000  thrpt   10    36.335 ±  0.490  ops/s
ParallelBenchmark.parTraverse       100000     100  thrpt   10   364.892 ±  5.622  ops/s
ParallelBenchmark.parTraverse       100000    1000  thrpt   10    47.056 ±  0.736  ops/s
ParallelBenchmark.parTraverse       100000   10000  thrpt   10     4.995 ±  0.053  ops/s
ParallelBenchmark.parTraverse      1000000     100  thrpt   10    46.086 ±  0.445  ops/s
ParallelBenchmark.parTraverse      1000000    1000  thrpt   10     5.107 ±  0.044  ops/s
ParallelBenchmark.parTraverse      1000000   10000  thrpt   10     0.524 ±  0.005  ops/s
```

`This PR`:
```
Benchmark                      (cpuTokens)  (size)   Mode  Cnt     Score    Error  Units
ParallelBenchmark.parTraverse          100     100  thrpt   10  1923.358 ± 94.031  ops/s
ParallelBenchmark.parTraverse          100    1000  thrpt   10   802.741 ± 20.411  ops/s
ParallelBenchmark.parTraverse          100   10000  thrpt   10   213.009 ±  5.766  ops/s
ParallelBenchmark.parTraverse         1000     100  thrpt   10  1452.882 ± 55.123  ops/s
ParallelBenchmark.parTraverse         1000    1000  thrpt   10   735.565 ± 22.182  ops/s
ParallelBenchmark.parTraverse         1000   10000  thrpt   10   156.585 ±  5.453  ops/s
ParallelBenchmark.parTraverse        10000     100  thrpt   10   976.165 ± 59.770  ops/s
ParallelBenchmark.parTraverse        10000    1000  thrpt   10   316.931 ± 31.421  ops/s
ParallelBenchmark.parTraverse        10000   10000  thrpt   10    41.550 ±  2.564  ops/s
ParallelBenchmark.parTraverse       100000     100  thrpt   10   396.673 ±  6.149  ops/s
ParallelBenchmark.parTraverse       100000    1000  thrpt   10    48.691 ±  0.602  ops/s
ParallelBenchmark.parTraverse       100000   10000  thrpt   10     5.116 ±  0.065  ops/s
ParallelBenchmark.parTraverse      1000000     100  thrpt   10    46.660 ±  0.361  ops/s
ParallelBenchmark.parTraverse      1000000    1000  thrpt   10     5.093 ±  0.089  ops/s
ParallelBenchmark.parTraverse      1000000   10000  thrpt   10     0.528 ±  0.004  ops/s
```

I think this is a spectacular improvement and well worth the code duplication. Amazing work @manufacturist! You should be proud.

vasilmkd:
I will do another pass over the code and give my final approval later.

kernel/shared/src/main/scala/cats/effect/kernel/instances/GenSpawnInstances.scala
```diff
@@ -53,7 +53,49 @@ trait GenSpawnInstances {
       final override def map2[A, B, Z](fa: ParallelF[F, A], fb: ParallelF[F, B])(
           f: (A, B) => Z): ParallelF[F, Z] =
         ParallelF(
-          F.both(ParallelF.value(fa), ParallelF.value(fb)).map { case (a, b) => f(a, b) }
+          F.uncancelable { poll =>
+            for {
+              fiberA <- F.start(ParallelF.value(fa))
+              fiberB <- F.start(ParallelF.value(fb))
+
+              a <- F.onCancel(poll(fiberA.join), fiberB.cancel).flatMap[A] {
```
vasilmkd: Here's a scenario, we get canceled while trying to `join` the `fiberA`. However, this does not cancel `fiberA`.

Is this a valid concern or am I missing something?

kernel/shared/src/main/scala/cats/effect/kernel/instances/GenSpawnInstances.scala
```diff
@@ -53,7 +53,49 @@ trait GenSpawnInstances {
...
+                case Outcome.Succeeded(fa) => fa
+                case Outcome.Errored(e) => fiberB.cancel *> F.raiseError(e)
+                case Outcome.Canceled() => fiberB.cancel *> F.never
```
vasilmkd: I don't quite understand this interaction. I fear this will enter `F.never` here.

Should this line be: `fiberB.cancel *> F.canceled *> F.never` instead? I also don't quite understand how this all interacts with the outer `uncancelable`.

vasilmkd (commented):
@djspiewak Please take a close look when reviewing the changes. I'd love an explanation on a couple of confusion points I encountered in the review. Thanks.

kernel/shared/src/main/scala/cats/effect/kernel/instances/GenSpawnInstances.scala
djspiewak: This should indeed be `poll(fiberB.cancel *> F.canceled *> F.never)` (and analogously in the `b` case)

kernel/shared/src/main/scala/cats/effect/kernel/instances/GenSpawnInstances.scala
djspiewak: This is a valid concern! We need to cancel *both* of the fibers if either `join` is canceled. And in fact, we need to ensure that this cancelation is itself run in parallel, so the correct finalizer here would be `F.onCancel(poll(fiberA.join), F.both(fiberB.cancel, fiberA.cancel))`.

djspiewak (changes requested):
Just the changes noted and confirmed in @vasilmkd's review. Otherwise, this looks good! Very impressive work.

kernel/shared/src/main/scala/cats/effect/kernel/instances/GenSpawnInstances.scala
```diff
@@ -53,7 +53,53 @@ trait GenSpawnInstances {
...
+            for {
+              fiberA <- F.start(ParallelF.value(fa))
+              fiberB <- F.start(ParallelF.value(fb))
+
+              a <- F
+                .onCancel(poll(fiberA.join), F.both(fiberA.cancel, fiberB.cancel).void)
+                .flatMap[A] {
+                  case Outcome.Succeeded(fa) => fa
+                  case Outcome.Errored(e) => fiberB.cancel *> F.raiseError(e)
+                  case Outcome.Canceled() => poll(fiberB.cancel *> F.canceled *> F.never)
```
vasilmkd: @djspiewak I don't agree that this `poll` should be here. Can you provide an explanation why?

djspiewak: `cancel` is uncancelable, so it has no effect on that segment. The `poll` strictly applies to the `*> F.canceled *> F.never` bit. But of course, if we're canceled while we're trying to cancel ourselves, then clearly there's no problem.

I could see an argument for pushing the `poll` after the `fiberB.cancel`, just so we don't rely on `poll(uncancelable(...))` being gapless.

vasilmkd: The `poll(F.canceled *> F.never)` is needed because if this line were `F.canceled *> F.never`, `F.canceled` would be ignored due to the outer `uncancelable`?

djspiewak: Yep!

vasilmkd: Ok, I am still in favor of moving the `poll` after `fiberB.cancel`.

djspiewak: After thinking about it, so am I. I think this is probably the last change needed to land this PR!

djspiewak (approved):
Excellent work!

kernel/shared/src/main/scala/cats/effect/kernel/instances/GenSpawnInstances.scala
```diff
@@ -53,7 +53,53 @@ trait GenSpawnInstances {
...
         ParallelF(
-          F.both(ParallelF.value(fa), ParallelF.value(fb)).map { case (a, b) => f(a, b) }
+          F.uncancelable { poll =>
+            for {
+              fiberA <- F.start(ParallelF.value(fa))
+              fiberB <- F.start(ParallelF.value(fb))
+
+              a <- F
+                .onCancel(poll(fiberA.join), F.both(fiberA.cancel, fiberB.cancel).void)
+                .flatMap[A] {
```
vasilmkd, line 63: Is this type annotation `[A]` (and later `[Z]`) really necessary?

vasilmkd: It is necessary.

vasilmkd (approved):
Very impactful change @manufacturist! Amazing contribution. Thank you for your patience while we figured things out along the way. It's usually how great PRs come together. 😄

kernel/shared/src/main/scala/cats/effect/kernel/instances/GenSpawnInstances.scala
```diff
@@ -53,7 +53,53 @@ trait GenSpawnInstances {
       final override def map2[A, B, Z](fa: ParallelF[F, A], fb: ParallelF[F, B])(
```
johnynek, line 53: this code is exactly the same as below, except for the `Eval` wrapping, no?

Why not have one private method that returns `Now[ParallelF[F, Z]]` and in the map2 case pass `fa, Now(fb)`? I think the extra now wrap unwrap is likely trivial.

djspiewak: > I think the extra now wrap unwrap is likely trivial.

The unwrap is pretty trivial but the allocation is… less so. Particularly since you pay that penalty on a per-item basis.

PR #2163, opened 2021-07-27
Updates to Backpressure per initial PR comments

etspaceman:
CC @rossabaker @vasilmkd

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Backpressure.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Backpressure.scala
@@ -55,34 +55,31 @@ object Backpressure {
   def apply[F[_]](
       strategy: Strategy,
       bound: Int
-  )(implicit GC: GenConcurrent[F, Throwable]): F[Backpressure[F]] = {
+  )(implicit GC: GenConcurrent[F, _]): F[Backpressure[F]] = {
+    require(bound > 0)
     val notAcquired: F[None.type] = GC.pure(None)
 
-    GC.pure(bound > 0)
-      .ifM(
-        Semaphore[F](bound.toLong).map(sem =>
-          strategy match {
-            case Strategy.Lossy =>
-              new Backpressure[F] {
-                override def metered[A](f: F[A]): F[Option[A]] =
-                  sem
-                    .tryAcquire
-                    .bracket {
-                      case true => f.map(_.some)
-                      case false => notAcquired.widen[Option[A]]
-                    } {
-                      case true => sem.release
-                      case false => GC.unit
-                    }
-              }
-            case Strategy.Lossless =>
-              new Backpressure[F] {
-                override def metered[A](f: F[A]): F[Option[A]] =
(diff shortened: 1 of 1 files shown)


std/shared/src/main/scala/cats/effect/std/Backpressure.scala
```diff
@@ -56,29 +56,33 @@ object Backpressure {
       strategy: Strategy,
       bound: Int
   )(implicit GC: GenConcurrent[F, Throwable]): F[Backpressure[F]] = {
```
vasilmkd: Is there a possibility to relax the `GenConcurrent` constraint to be `GenConcurrent[F, _]` like the other types in `std`? That does mean that you would have to rever the `require` change.

vasilmkd (commented):
The diff is terrible for this change. 😆

vasilmkd:
I take issue with the fact that the previous pull request was merged without explicit approval by anyone. I feel like we collectively dropped the ball there.

std/shared/src/main/scala/cats/effect/std/Backpressure.scala
etspaceman: Done. `MonadCancellable[F, Throwable]` is a requirement of `Resource.use`. So I replaced the `sem.permit` call with a bracket around `sem.acquire`

djspiewak:
> I take issue with the fact that the previous pull request was merged without explicit approval by anyone. I feel like we collectively dropped the ball there.

We did, but that's mostly on me. I was kind of tacitly waiting for more feedback and @kubukoz broke the deadlock. We shouldn't make a habit of that kind of thing, but it wasn't the first step to go wrong.

Taking a step back to what I was pondering… my concern is that people will see this mechanism and think it's all they really need for backpressure, ignoring other higher-level tools like `Queue`. This is also a very general name (`Backpressure`) tailored at a very specific case (fiber supervision), which gives me a bit of pause. No one else seems to have strong opinions though so I could be very much over-indexing. :-)

vasilmkd:
We live and we learn. As previously discussed, the train originally left the station when we shipped this interface in CE2. I personally was not in favor of it, even for CE2, but I didn't voice my opinion more loudly, so I guess that's ultimately fair.

vasilmkd (approved):

kubukoz:
> > I take issue with the fact that the previous pull request was merged without explicit approval by anyone. I feel like we collectively dropped the ball there.
> 
> We did, but that's mostly on me. I was kind of tacitly waiting for more feedback and @kubukoz broke the deadlock. We shouldn't make a habit of that kind of thing, but it wasn't the first step to go wrong.

Sorry about that. It's really on me that I didn't check if every comment was resolved.

Can we enable required approvals in this repo? That should at least serve as a good hint that something is missing if there are no approvals.

djspiewak:
> Can we enable required approvals in this repo? That should at least serve as a good hint that something is missing if there are no approvals.

Done. I generally don't like that restriction because it's annoying. :-P But you're right that it would have caught this problem.

PR #2164, opened 2021-07-27
takeN/takeFrontN/takeBackN proposal for Queue/Dequeue

etspaceman:
When converting to CE3, I noticed that the CE3 Queue does not offer the same functionality that FS2 had with `dequeueChunk1`: https://github.com/typelevel/fs2/blob/series/2.5.x/core/shared/src/main/scala/fs2/concurrent/Queue.scala#L71

This is a proposal to add such functionality.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Queue.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Queue.scala
@@ -51,6 +51,7 @@ abstract class Queue[F[_], A] extends QueueSource[F, A] with QueueSink[F, A] { s
       def tryOffer(a: A): G[Boolean] = f(self.tryOffer(a))
       def size: G[Int] = f(self.size)
       val take: G[A] = f(self.take)
+      def takeN[H[_]: MonoidK: Applicative](n: Int): G[H[A]] = f(self.takeN[H](n))
       val tryTake: G[Option[A]] = f(self.tryTake)
     }
 }
@@ -211,6 +212,15 @@ object Queue {
         }
       }
 
+    def takeN[G[_]: MonoidK: Applicative](n: Int): F[G[A]] = {
+      require(n > 0)
+      F.tailRecM((MonoidK[G].empty[A], n)) {
+        case (acc, remaining) =>
+          if (remaining == 0) acc.asRight[(G[A], Int)].pure[F]
+          else take.map(x => (acc <+> x.pure[G], remaining - 1).asLeft[G[A]])
+      }
+    }
+
     val tryTake: F[Option[A]] =
       state
         .modify {
@@ -319,6 +329,8 @@ object Queue {
             fa.tryOffer(g(b))
           override def take: F[B] =
             fa.take.map(f)
+          override def takeN[G[_]: MonoidK: Applicative](n: Int): F[G[B]] =
+            fa.takeN[G](n).map(_.map(f))
           override def tryTake: F[Option[B]] =
             fa.tryTake.map(_.map(f))
           override def size: F[Int] =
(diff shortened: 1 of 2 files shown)


etspaceman:
Tests to come

std/shared/src/main/scala/cats/effect/std/Queue.scala
```diff
@@ -211,6 +212,15 @@ object Queue {
         }
       }
 
+    def takeN[G[_]: MonoidK: Applicative](n: Int): F[G[A]] = {
+      require(n > 0)
```
etspaceman, line 216: I tried doing a RaiseError with a RuntimeException but got the following:

```
[error] /Users/emeisel/GitHub/cats-effect/std/shared/src/main/scala/cats/effect/std/Queue.scala:220:20: type mismatch;
[error]  found   : RuntimeException
[error]  required: _$6
[error]       F.raiseError(new RuntimeException("N must be > 0 for takeN"))
[error]                    ^
[error] one error found
```

RaasAhsan: What was the typeclass constraint on `F` ? The CE3 hierarchy abstract over the error type now so raising an exception only works when `E <: Throwable`

RaasAhsan: Also I think there is precedent for using `require` anyway (check some of the other concurrent types) so it's not a big deal

std/shared/src/main/scala/cats/effect/std/Queue.scala
```diff
@@ -211,6 +212,15 @@ object Queue {
...
+      F.tailRecM((MonoidK[G].empty[A], n)) {
+        case (acc, remaining) =>
+          if (remaining == 0) acc.asRight[(G[A], Int)].pure[F]
+          else take.map(x => (acc <+> x.pure[G], remaining - 1).asLeft[G[A]])
```
etspaceman, line 220: I'm not sure how we feel about this one. Ideally we'd be able to take a chunk in one move. But @RaasAhsan mentioned we had some issues with that when designing the Queue. So this simply just calls `take` under the hood. If you have concurrent calls to this, you might have some odd behavior. Not sure if we want that.

std/shared/src/main/scala/cats/effect/std/Queue.scala
vasilmkd: There's precedent for `require`, but it has been limited to constructors of concurrent data types until now.

vasilmkd:
@djspiewak Decision time. Are you still in favor of breaking binary compatibility and redesign/add native support for these features on `Queue`, and possibly others, like `takeAll/offerAll`.

One more question, did the previous `fs2.Queue` have methods which take from the back of the queue?

djspiewak:
I think we need to do some benchmarking to sort out the value-add. I was also playing around with `Queue#takeN` locally, and as I was doing it, I took a look at `Stream.fromQueue` and realized that, by and large, it's actually already solving this problem, specifically because it is very conventional to have a `Queue[F, Chunk[A]]` rather than a `Queue[F, A]`. I *think* it's just a difference between multiple effect boundaries and just one; the backpressure win (which is to say, the negative feedback loop) is already present in practice.

I'm definitely open to breaking bincompat for this type of thing if we can convince ourselves that it's really buying us that much.

etspaceman:
> One more question, did the previous fs2.Queue have methods which take from the back of the queue?

I don't believe it did. I can remove that if desired.

vasilmkd:
No, it's fine, I was just curious. 😄

Swoorup:
Wonder if there should also be a way to do `tryTakeChunks` in a non-blocking away? If the queue is empty, return empty, I am using something like this atm. Thanks to @armanbilge 

```scala
def tryTakeChunk[F[_]: Concurrent](q: Queue[F, Int]) = 
  Stream.repeatEval(q.tryTake).unNoneTerminate.compile.toList
```

Issue #1112, opened 2020-08-20
Find a way to test webworkers and implement executor polyfill


djspiewak:
At present, the `ExecutionContext` polyfill for `setImmediate` specifically excludes webworkers, which means that we use `setTimeout` in that runtime. This is suboptimal, but the only reason we *haven't* solved this problem is because we don't have a good test harness which runs `IO` within a web worker. This is kind of an infrastructure problem. If we can come up with a way to have an automated harness in CI which runs `IOSpec` under a web worker, then we can implement the polyfill and close that last loophole.

PR #2165, opened 2021-07-27
Test `IO` on Web Workers

armanbilge:
Hopefully will address #1112. I think this is _really_ close, we just need to find the magical Firefox setting to make security lax enough to load the web worker despite cross-origin bla bla.

The approach is to create an `IOApp` which runs the `IOSpec`, `fastOptJS` it, then run a new test suite that loads it as a webworker.


djspiewak (approved):
Very impressive hackery!

Issue #1302, opened 2020-10-12
Experiment with scala native


djspiewak:
It would be really interesting to see what we can do on this front, now that Scala Native is on 2.12. We would need Cats to publish for it, though. This is a rather large and sprawling project, tbh, and I'm not sure if there's even a definitive payoff at the end, but it will be cool to try anyway!

djspiewak:
To note, the only async-anything support within scala-native comes via libuv and https://github.com/scala-native/scala-native-loop. We could technically do something like this, but it's an open question as to whether or not it would be worthwhile.

benhutchison:
🙌 If you come up with any experiments, I'd love to test them out

djspiewak:
This is definitely one of those things that will be a little interest-driven. 😃 Also if anyone wants to dive in and start playing with getting it working, the Cats Effect side of it would actually be very straightforward (for the most part, you just need to implement an ExecutionContext that works), and I would love to help out anyone interested in trying their hand at it.

benhutchison:
So I did some investigating and concluded that integrating with Scala Native & libuv may be currently possible, but it's bleeding edge and exceeds my personal bravery level. 

I think it might be prudent to wait for stabilization of:
- Native interop design changes, which are part-implemented ATM https://github.com/scala-native/scala-native/issues/897
- The libuv integration, eg the [changes](https://github.com/scala-native/scala-native-loop/pull/36) to migrate from 0.4.0-M2 to 0.4.0 have a lot of surface area, and that may continue as the native interop evolves

JD557:
Would it probably make sense to just publish the kernel for scala native for now?

While it migh make sense to wait before porting cats IO, having the typeclasses available would allow:
- Library authors to still cross-publish code with CE3 to scala-native
- Application authors to plug in a simpler IO monad and use only a subset of operations (e.g. it should be simple to write an IO monad that only implements Sync and the required typeclasses).
- ZIO supposedly (I think there are some caveats) already supports scala-native 0.4.0, so maybe one could use that to run CE3 libraries in scala-native while the Cats IO is not ready

WDYT?

benhutchison:
That makes sense to me. It opens the way for someone to have a go of
implementing an IO on SN.

Maintaining an IO for SN within CE is what I'd be wary of, as the
underlying libuv bindings might keep changing for a while yet ..

ScalaWilliam:
> Would it probably make sense to just publish the kernel for scala native for now?
> 
[...]

PR #2168, opened 2021-07-27
Cross-build kernel+std for Scala Native

armanbilge:
As proposed in https://github.com/typelevel/cats-effect/issues/1302#issuecomment-882643273. Note: there is currently no laws or testkit support, this will require coop and MTL to join us on Native (probably doable). ~Also there is no specs2 for Native, probably the best fix here is for CE to migrate to munit :(~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
           - adopt@1.11
           - adopt@1.16
           - graalvm-ce-java8@21.1
-        ci: [ciJVM, ciJS, ciFirefox]
+        ci: [ciJVM, ciJS, ciFirefox, ciNative]
         exclude:
           - ci: ciJS
             java: adopt@1.11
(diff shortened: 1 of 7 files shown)


djspiewak:
I think my main concern with this is the fact that Scala Native still doesn't support Scala 3, which in turn means that all of the build machinery becomes very… bespoke. For example, we can no longer run `+test`, or `+ci`. We see some of the annoyances with this on the Cats build, which are ameliorated by the `validate` aliases, but it still represents an ongoing maintenance pain.

I'm certainly open to being convinced, but I have a hard time believing that this maintenance pain is justified given the low current rate of adoption of Scala Native.

armanbilge:
I agree and think we should wait to cross at least until it's a little less painful (ahem Scala 3 :) and/or there's a runtime available. There's _just_ enough here that anyone who wants to play around with implementing a native runtime can fork my branch, publishLocal, and give it a go.

PR #2169, opened 2021-07-27
`IO.suspend` -> `IO.defer`

vasilmkd:


djspiewak (dismissed):

vasilmkd:
The CI unfortunately won't pass because CE `2.5.2` is not released yet.

djspiewak (approved):

PR #2180, opened 2021-07-31
Small CI optimizations

vasilmkd:


djspiewak (commented):
Actually we need these steps for it to be deterministic and self-contained. Removing the separate compilation steps actually makes the script fail sometimes because the timeout includes compiling.

vasilmkd:
That's unfortunate.

Can we put the exampleJVM/compile in ciJVM and exampleJS/compile in ciJS, right before they both finish?

vasilmkd:
I reworked the PR from scratch. I removed the `sbt showVersion` invocation in the `test-jvm.sh` script, I think it is unnecessary because we use `sbt` in prior tasks, so the launcher should already be available. I also put `exampleJVM/compile` at the end of `ciJVM`. I know these changes look funny, but they make up around 30 seconds per CI run, which adds up.

djspiewak (approved):

PR #2187, opened 2021-08-01
Playing with the memory of CI workers

vasilmkd:
Running `sbt ciFirefox/ciChrome` on this branch only executes `IOSpec`.

`IOSpec` on my machine takes around 48 seconds to execute on Chrome. It takes 2 minutes and 40 seconds to execute on Firefox.

Making every browser use `setTimeout` slows down the Chrome run to around 70 seconds, however the Firefox run stays within margin of error.

Either postMessage is not optimized on Firefox at all, or we have some slow section (which trips up Firefox) in the code. I'm pretty sure, but not 100% certain, that both Chrome and Firefox end up executing the same code (postMessage).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,6 +120,24 @@ jobs:
         shell: bash
         run: npm install
 
+      - name: Let sbt use all available memory
+        if: matrix.ci == 'ciJVM' && matrix.os != 'windows-latest'
+        shell: bash
+        run: |
+          cat > .jvmopts <<EOF
+          -Xmx7G
+          -XX:+UseG1GC
+          EOF
+
+      - name: Let browsers and node use more memory
+        if: matrix.ci != 'ciJVM'
+        shell: bash
+        run: |
+          cat > .jvmopts <<EOF
+          -Xmx3G
+          -XX:+UseG1GC
+          EOF
+
       - name: Check that workflows are up to date
         shell: bash
         run: sbt ++${{ matrix.scala }} githubWorkflowCheck
--- a/build.sbt
+++ b/build.sbt
@@ -78,6 +78,22 @@ ThisBuild / githubWorkflowBuildPreamble ++= Seq(
     List("npm install"),
     name = Some("Install jsdom"),
     cond = Some("matrix.ci == 'ciJSDOMNodeJS'")
+  ),
+  WorkflowStep.Run(
+    List(s"""cat > .jvmopts <<EOF
+            |-Xmx7G
+            |-XX:+UseG1GC
+            |EOF""".stripMargin),
+    name = Some("Let sbt use all available memory"),
+    cond = Some(s"matrix.ci == 'ciJVM' && matrix.os != '$Windows'")
+  ),
+  WorkflowStep.Run(
+    List(s"""cat > .jvmopts <<EOF
+            |-Xmx3G
+            |-XX:+UseG1GC
+            |EOF""".stripMargin),
+    name = Some("Let browsers and node use more memory"),
+    cond = Some("matrix.ci != 'ciJVM'")
   )
 )
 


core/js/src/main/scala/cats/effect/unsafe/PolyfillExecutionContext.scala
```diff
@@ -16,132 +16,18 @@
...
 /**
  * Based on https://github.com/YuzuJS/setImmediate
  */
 private[unsafe] object PolyfillExecutionContext extends ExecutionContext {
-  private[this] val Undefined = "undefined"
-
-  def execute(runnable: Runnable): Unit =
-    setImmediate(() => runnable.run())
+  def execute(runnable: Runnable): Unit = {
+    js.Dynamic.global.setTimeout(() => runnable.run(), 0)
```
japgolly: There's a properly typed version of this:

```scala
import scala.scalajs.js.timers._

setTimeout(0)(runnable.run())
```

armanbilge:
> I'm pretty sure, but not 100% certain, that both Chrome and Firefox end up executing the same code (postMessage)

Yes, I also mentioned this to @djspiewak yesterday that in our current tests we aren't checking if any of the fancy implementations are actually being used.

After thinking: what if in `PolyfillExecutionContext` we set a variable/flag that indicates which polyfill was selected?

djspiewak:
I did some poking at this. I can confirm that the Firefox polyfill is being used, though it has a bug which doesn't affect the tests which I'm fixing.

What seems to happen is the tests run relatively quickly and then Firefox just sits there, not reporting success, for almost the whole time. It makes me wonder if this is a bug in Selenium.

armanbilge:
> What seems to happen is the tests run relatively quickly and then Firefox just sits there, not reporting success, for almost the whole time.

Yes!! So it's not just me seeing this. I just observed this again on my http4s tests.

vasilmkd:
> I did some poking at this. I can confirm that the Firefox polyfill is being used, though it has a bug which doesn't affect the tests which I'm fixing.
> 
> What seems to happen is the tests run relatively quickly and then Firefox just sits there, not reporting success, for almost the whole time. It makes me wonder if this is a bug in Selenium.

How do you verify that this is happening? I'm not doubting you, I want to see it for myself.

djspiewak:
> How do you verify that this is happening? I'm not doubting you, I want to see it for myself.

I put a `println` in the `postMessage` hack, so I got a new line every time a yield returned. A bunch printed very quickly, then nothing for a long period of time, then the test suite finished and printed its results. That leads me to suspect that Firefox is just sitting idle.

vasilmkd:
@djspiewak Can you run the same setup with `ciChrome`? I'm curious to see if you will see the same gaps (obviously shorter because Chrome is faster).

I think that a big part of the suite is fully synchronous, and this corresponds to the gap.

vasilmkd:
The idea here is the following:

The GitHub Action runners have 2 CPU cores and 7 Gb of RAM.

The first commit uses the `all testsJS/test webWorkerTests/test` command to run the tests in parallel. This saves a few minutes because otherwise `webWorkerTests` are waiting to start because there is 1 slow spec from the `testsJS` run and the other CPU core is doing nothing.

The second commit plays with the `.jvmopts` file which `sbt` uses to manage its resource usage. By default, we have it set at 4 Gb. This is honestly starting to push it, even for my local development environment where I have increased this limit to 8 Gb for the Cats Effect project.

In similar fashion, I gave `sbt` on `ciJVM` access to potentially 6 Gb of heap memory, because there aren't other competing processes in the worker.

With `node` and the browsers however, I restricted `sbt` to 3 Gb of heap. This leaves 4 Gb of memory for the other processes which actually run the testing code.

Finally, because the workers are so small, the JVM was defaulting to the more primitive [serial GC](https://docs.oracle.com/en/java/javase/11/gctuning/available-collectors.html#GUID-C7B19628-27BA-4945-9004-EC0F08C76003). There are warnings printed by the JVM to use the G1GC and try increasing the heap size, so the second commit also forces the `G1GC` garbage collector.

I'm open to other suggestions. The worst offender is usually Scala.js on Firefox with Scala 3, which can take up to 29-32 minutes to execute fully. It now seems to be around 20-25 minutes with these changes (I honestly don't understand the variance).

@djspiewak Thoughts?

vasilmkd:
On the other hand of the spectrum, Chrome can run in less than 7 minutes with these changes...

armanbilge:
This is really great stuff, lots of inspiration here for improving the http4s CI as well.

armanbilge:
Btw I switched to G1GC in http4s and it seems to have fixed all the memory issues in the fastOptJS stage. Thanks for the tip!

vasilmkd:
Wow that's great news!!

.github/workflows/ci.yml
```diff
@@ -120,6 +120,24 @@ jobs:
         shell: bash
         run: npm install
 
+      - name: Let sbt use all available memory
+        if: matrix.ci == 'ciJVM' && matrix.os != 'windows-latest'
+        shell: bash
+        run: |
+          cat > .jvmopts <<EOF
+          -Xmx7G
+          -XX:+UseG1GC
```
djspiewak, line 129: IMO we could just put this flag into the default file :-D

.github/workflows/ci.yml
```diff
@@ -120,6 +120,24 @@ jobs:
         shell: bash
         run: npm install
 
+      - name: Let sbt use all available memory
+        if: matrix.ci == 'ciJVM' && matrix.os != 'windows-latest'
+        shell: bash
+        run: |
+          cat > .jvmopts <<EOF
+          -Xmx7G
```
djspiewak, line 128: A better way of doing this might be to change the sbt command to `$SBT`, which in turn we set to `sbt` in the build and then modify conditionally to include `-J-Xmx7g` (and similar). I can try to prototype that quickly. The problem with doing it this way is we're modifying the file, which in turn makes the directory dirty, which in turn will affect the versioning scheme. That's not a problem *today* but if we do fancier things with the build in future, we may have forgotten about this bit of twiddling, and the result may be a *very* hard-to-debug problem.

.github/workflows/ci.yml
vasilmkd: I have it in my local ignore list, but I use `8G`. I feel `7G` is a weird default for anybody. I'm against it.

.github/workflows/ci.yml
vasilmkd: Will this clash in any way with the `v13` update to `setup-scala` which brings the vanilla `sbt` launcher? I have a PR for that in your `sbt-github-actions` repo.

.github/workflows/ci.yml
vasilmkd: I don't want to impose weird defaults on newcomers, who will run `sbt` inside the Cats Effect directory and suddenly their build is hanging because we're asking for too much memory. `4G` works as a fine default IMO, it's really needed for our build.

djspiewak: Oh I meant the `+UseG1GC` flag in the default file :-)

.github/workflows/ci.yml
djspiewak: Should work just fine with the vanilla launcher

vasilmkd: Yup, I completely agree with this approach.

.github/workflows/ci.yml
vasilmkd: I'm not sure if `G1GC` is a recommended GC on JDK8.

djspiewak:
Bumping this for follow-up. Super low priority just want to make sure it doesn't get lost

vasilmkd:
Can we punt this off to an issue and close?

djspiewak:
> Can we punt this off to an issue and close?

Yep!

PR #2188, opened 2021-08-02
Fix JS feature detection so that it no longer crashes

japgolly:
An example of this crashing in practice can be seen in https://github.com/japgolly/scalajs-react/blob/v2.0.0-RC2/downstream-tests/js/src/test/resources/polyfill.js where I had to add a fake `importScripts` as a workaround. A `try`/`catch` is necessary.

### DIFF
--- a/core/js/src/main/scala/cats/effect/unsafe/PolyfillExecutionContext.scala
+++ b/core/js/src/main/scala/cats/effect/unsafe/PolyfillExecutionContext.scala
@@ -25,7 +25,6 @@ import scala.util.Random
  * Based on https://github.com/YuzuJS/setImmediate
  */
 private[unsafe] object PolyfillExecutionContext extends ExecutionContext {
-  private[this] val Undefined = "undefined"
 
   def execute(runnable: Runnable): Unit =
     setImmediate(() => runnable.run())
@@ -33,8 +32,15 @@ private[unsafe] object PolyfillExecutionContext extends ExecutionContext {
   def reportFailure(cause: Throwable): Unit =
     cause.printStackTrace()
 
+  private[this] def isAvailable(a: => js.Dynamic): Boolean =
+    try {
+      a.asInstanceOf[js.UndefOr[Any]].isDefined
+    } catch {
+      case _: Throwable => false
+    }
+
   private[this] val setImmediate: (() => Unit) => Unit = {
-    if (js.typeOf(js.Dynamic.global.setImmediate) == Undefined) {
+    if (!isAvailable(js.Dynamic.global.setImmediate)) {
       var nextHandle = 1
       val tasksByHandle = mutable.Map[Int, () => Unit]()
       var currentlyRunningATask = false
@@ -42,8 +48,7 @@ private[unsafe] object PolyfillExecutionContext extends ExecutionContext {
       def canUsePostMessage(): Boolean = {
         // The test against `importScripts` prevents this implementation from being installed inside a web worker,
         // where `global.postMessage` means something completely different and can't be used for this purpose.
-        if (js.typeOf(js.Dynamic.global.postMessage) != Undefined && js.typeOf(
-            js.Dynamic.global.importScripts) == Undefined) {
+        if (isAvailable(js.Dynamic.global.postMessage) && !isAvailable(js.Dynamic.global.importScripts)) {
           var postMessageIsAsynchronous = true
           val oldOnMessage = js.Dynamic.global.onmessage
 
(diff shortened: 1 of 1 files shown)


armanbilge:
Thanks for this! Our test suite currently passes on Firefox and Chrome, is there a way we can test for this in the future?

japgolly:
No worries! Yeah it should be reproducible by running the tests with `jsEnv := new org.scalajs.jsenv.jsdomnodejs.JSDOMNodeJSEnv`

armanbilge:
Hmm, isn't that an artificial environment? Does this actually occur in the browser?

japgolly:
When you say the tests pass on Firefox and Chrome, I presume you're not testing using Cats Effect on the web worker side? If you have the capacity it would be good to test that too because it's a different env that normal JS world.

japgolly:
> Hmm, isn't that an artificial environment?

Nah, it's a valid use-case to run Scala.JS from Node itself (eg. for AWS Lambdas).

armanbilge:
> I presume you're not testing using Cats Effect on the web worker side?

We are! I just added that #2165 actually.

armanbilge:
> Nah, it's a valid use-case to run Scala.JS from Node itself (eg. for AWS Lambdas).

Yes, well we also test in the Node.js environment for this reason :) Node.js, Chrome, and Firefox.

japgolly:
> Yes, well we also test in the Node.js environment for this reason

Oh you're already testing Node.js! Awesome! Hmmm, in that case maybe it's literally the jsdom interaction. But regardless, there are still two good reasons to adopt this (or a similar change):

1) Node with JSDOM is the standard way for unit testing the Scala.JS component of webapps
2) `js.typeOf(js.Dynamic.global.xxx)` is a direct copy of JS where polyfills do `typeof(xxx)` but it's an incorrect translation to Scala.JS. `js.Dynamic.global.xxx` will throw an exception in SJS where no exception would occur in JS.

armanbilge:
> Node with JSDOM is the standard way for unit testing the Scala.JS component of webapps

Sounds like maybe this should be added to the test environments.

djspiewak (approved):
I think you're probably right that this should be added to the test environments. The matrix grows ever wider…

armanbilge:
I added the JSDOM environment and did a before-after with @japgolly's commit in https://github.com/typelevel/cats-effect/pull/2190, unfortunately, it doesn't appear to fix the problem. TBH I'm not sure it's related to the execution context at all.

core/js/src/main/scala/cats/effect/unsafe/PolyfillExecutionContext.scala
```diff
@@ -25,25 +25,30 @@ import scala.util.Random
...
 
   def execute(runnable: Runnable): Unit =
     setImmediate(() => runnable.run())
 
   def reportFailure(cause: Throwable): Unit =
     cause.printStackTrace()
 
+  private[this] def isAvailable(a: => js.Dynamic): Boolean =
+    try {
+      a.asInstanceOf[js.UndefOr[Any]].isDefined
```
vasilmkd, line 37: What about `!js.isUndefined(a)`?

core/js/src/main/scala/cats/effect/unsafe/PolyfillExecutionContext.scala
```diff
@@ -25,25 +25,30 @@ import scala.util.Random
...
+    } catch {
+      case _: Throwable => false
+    }
```
vasilmkd, line 40: How about `!js.isUndefined(a)`? Is that the same?

vasilmkd (changes requested):

armanbilge:
Ok, I fixed this in #2191, @japgolly would you mind giving it a try?

You were right, we did need a `try`/`catch` in the polyfill, but in a different place. The reason your `importScripts` hack works is because it was short-circuiting the branch that would otherwise throw the exception. AFAICT this only affects the JSDOMNodeJS environment.

core/js/src/main/scala/cats/effect/unsafe/PolyfillExecutionContext.scala
armanbilge: I think I tried this when I implemented the web worker polyfill and ran into problems, but maybe I did it wrong.

japgolly: No `js.isUndefined` wouldn't work because it only tests whether a strict/eager value is undefined or not, but the problem is an exception is being thrown when attempting to resolve that value. Personally I think it's silly behaviour but sjrd is very, very adamant in these things. It is what it is now so when testing for optional JS features we need to always be ready to catch a (Fatal) exception.

japgolly:
> Ok, I fixed this in #2191, @japgolly would you mind giving it a try?

@armanbilge Yeah no worries, I'll give it a try later today and get back to you

core/js/src/main/scala/cats/effect/unsafe/PolyfillExecutionContext.scala
vasilmkd: Awesome, good to know. Thanks.

vasilmkd:
Is this PR superseded by #2191?

vasilmkd:
Closing in favor of #2191.

PR #2191, opened 2021-08-02
Add JSDOMNodeJSEnv to ci matrix and fix polyfill

armanbilge:
My own attempt to fix the problem reported in #2188.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,7 @@ tags
 .bloop/
 metals.sbt
 .vscode
+
+# npm
+node_modules/
+package-lock.json
--- a/.gitignore
+++ b/.gitignore
@@ -20,4 +20,3 @@ metals.sbt
 
 # npm
 node_modules/
-package-lock.json
--- a/build.sbt
+++ b/build.sbt
@@ -75,7 +75,7 @@ ThisBuild / githubWorkflowBuildPreamble ++= Seq(
     cond = Some("matrix.ci == 'ciJS' || matrix.ci == 'ciJSDOMNodeJS'")
   ),
   WorkflowStep.Run(
-    List("npm install jsdom"),
+    List("npm install"),
     name = Some("Install jsdom"),
     cond = Some("matrix.ci == 'ciJSDOMNodeJS'")
   )
--- a/build.sbt
+++ b/build.sbt
@@ -186,7 +186,7 @@ val CoopVersion = "1.1.1"
 
 replaceCommandAlias(
   "ci",
-  "; project /; headerCheck; scalafmtSbtCheck; scalafmtCheck; clean; test; coreJVM/mimaReportBinaryIssues; root/unidoc213; set Global / useJSEnv := JSEnv.Firefox; testsJS/test; webWorkerTests/test; set Global / useJSEnv := JSEnv.Chrome; testsJS/test; webWorkerTests/test; set Global / useJSEnv := JSEnv.NodeJS"
+  "; project /; headerCheck; scalafmtSbtCheck; scalafmtCheck; clean; test; coreJVM/mimaReportBinaryIssues; root/unidoc213; set Global / useJSEnv := JSEnv.Firefox; testsJS/test; webWorkerTests/test; set Global / useJSEnv := JSEnv.Chrome; testsJS/test; webWorkerTests/test; set Global / useJSEnv := JSEnv.JSDOMNodeJS; testsJS/test; set Global / useJSEnv := JSEnv.NodeJS"
 )
 
 addCommandAlias(
--- a/build.sbt
+++ b/build.sbt
@@ -22,6 +22,7 @@ import org.openqa.selenium.WebDriver
 import org.openqa.selenium.chrome.{ChromeDriver, ChromeOptions}
 import org.openqa.selenium.firefox.{FirefoxOptions, FirefoxProfile}
 import org.openqa.selenium.remote.server.{DriverFactory, DriverProvider}
+import org.scalajs.jsenv.jsdomnodejs.JSDOMNodeJSEnv
 import org.scalajs.jsenv.selenium.SeleniumJSEnv
 
 import JSEnv._
@@ -66,12 +67,19 @@ val GraalVM8 = "graalvm-ce-java8@21.1"
 ThisBuild / githubWorkflowJavaVersions := Seq(ScalaJSJava, LTSJava, LatestJava, GraalVM8)
 ThisBuild / githubWorkflowOSes := Seq(PrimaryOS, Windows)
 
-ThisBuild / githubWorkflowBuildPreamble +=
+ThisBuild / githubWorkflowBuildPreamble ++= Seq(
   WorkflowStep.Use(
     UseRef.Public("actions", "setup-node", "v2.1.2"),
     name = Some("Setup NodeJS v14 LTS"),
     params = Map("node-version" -> "14"),
-    cond = Some("matrix.ci == 'ciJS'"))
+    cond = Some("matrix.ci == 'ciJS' || matrix.ci == 'ciJSDOMNodeJS'")
+  ),
+  WorkflowStep.Run(
+    List("npm install jsdom"),
+    name = Some("Install jsdom"),
+    cond = Some("matrix.ci == 'ciJSDOMNodeJS'")
+  )
+)
 
 ThisBuild / githubWorkflowBuild := Seq(
   WorkflowStep.Sbt(List("${{ matrix.ci }}")),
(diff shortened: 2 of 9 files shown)


build.sbt
```diff
@@ -192,6 +201,7 @@ def browserCiCommand(browser: JSEnv) =
   s"; set Global / useJSEnv := JSEnv.$browser; project rootJS; headerCheck; scalafmtCheck; clean; testsJS/test; webWorkerTests/test; set Global / useJSEnv := JSEnv.NodeJS"
 addCommandAlias("ciFirefox", browserCiCommand(Firefox))
 addCommandAlias("ciChrome", browserCiCommand(Chrome))
+addCommandAlias("ciJSDOMNodeJS", browserCiCommand(JSDOMNodeJS))
```
djspiewak, line 204: Can you also update the `ci` alias? We should probably generate that alias in a way which requires less redundancy.

djspiewak (changes requested):

build.sbt
armanbilge: Fixed, I'll leave a generation scheme for a subsequent bikeshed sorry :)

core/js/src/main/scala/cats/effect/unsafe/PolyfillExecutionContext.scala
```diff
@@ -47,11 +48,16 @@ private[unsafe] object PolyfillExecutionContext extends ExecutionContext {
...
           val oldOnMessage = js.Dynamic.global.onmessage
 
-          js.Dynamic.global.onmessage = { () => postMessageIsAsynchronous = false }
-
-          js.Dynamic.global.postMessage("", "*")
-          js.Dynamic.global.onmessage = oldOnMessage
-          postMessageIsAsynchronous
+          try {
+            // This line throws `ReferenceError: onmessage is not defined` in JSDOMNodeJS environment
+            js.Dynamic.global.onmessage = { () => postMessageIsAsynchronous = false }
```
vasilmkd, line 53: Why not also check if `onmessage` is undefined in the `if` statement?

armanbilge: I considered this too, but `onmessage` in my browser console is `null`.

armanbilge: Well, maybe that makes sense for my browser ... anyway I figured better not to mess with the if conditions since those are important.

vasilmkd: `null` is not the same as `undefined`, so it would pass the if check. Just saying.

vasilmkd: I'm not too fussed about it, but we should maybe try it.

armanbilge: You are right, but IIRC correctly it is also `null` in JSDOMNodeJS env.

armanbilge: Sorry, to clarify I think what I considered was adding a `null` check to the if statment.

core/js/src/main/scala/cats/effect/unsafe/PolyfillExecutionContext.scala
```diff
@@ -47,11 +48,16 @@ private[unsafe] object PolyfillExecutionContext extends ExecutionContext {
...
+            js.Dynamic.global.postMessage("", "*")
+            js.Dynamic.global.onmessage = oldOnMessage
+            postMessageIsAsynchronous
+          } catch {
+            case NonFatal(_) =>
```
japgolly, line 58: This isn't going to work. Scala.JS (unfortunately) throws a fatal exception when referencing something in JS-land that isn't there at runtime. You've got to catch all exceptions.

armanbilge: Thanks again for your eyes on this ... currently this is passing in all test environments, is there something I'm missing??

japgolly:
I just tested this and it does indeed work! Thanks for the revision! :+1:

core/js/src/main/scala/cats/effect/unsafe/PolyfillExecutionContext.scala
japgolly: No it turns out I'm wrong, it seems I'm the one that's missing something :smile: I tested this out and it works just as you're expecting it to.... which is so strange to me because my beliefs were built on experiences. Maybe Scala.JS has received some changes as to how this bit of surface area works and I must've missed it. I'll invalidate my mental cache. Thanks for the collaboration :)

armanbilge:
Hooray!! Thanks for giving it a spin :)

core/js/src/main/scala/cats/effect/unsafe/PolyfillExecutionContext.scala
armanbilge: No worries! It sounded plausible to me too, I could have sworn I had similar experiences 🙃 But on the other hand I was able to replicate an issue with CI here, and make it go away, and it was consistent with your import scripts trick.

vasilmkd (approved):

.gitignore
```diff
@@ -17,3 +17,7 @@ tags
 .bloop/
 metals.sbt
 .vscode
+
+# npm
+node_modules/
+package-lock.json
```
armanbilge: @vasilmkd @djspiewak Just want to make sure you're aware that I decided to ignore the lockfile (created by the `JSDOM` dependency), but not sure about this decision at all. @mpilquist, @ChristopherDavenport, @rossabaker, and I have been struggling with this same issue for fs2 and http4s.

In this case its just a test dependency, but @rossabaker makes a good point about repeatable builds. https://github.com/http4s/http4s/pull/4938#discussion_r683397866

vasilmkd (changes requested):
Holding off until a decision about the `package-lock` file has been made.

vasilmkd:
Wait a second, what @djspiewak asked in [Discord](https://discord.com/channels/632277896739946517/839254871646404638/872835073932660786) is actually a very valid question. If `JSDOMNodeJSEnv` is indeed running in `node`, which we know it is, then this code path should not be exercised at all, because `node` supports `setImmediate` natively. I'm confused here.

armanbilge:
It's possible they are somehow disabling it, to better emulate the browser?

vasilmkd:
I confirm that the native `setImmediate` works on vanilla `node`. I'm investigating further.

armanbilge:
> Holding off until a decision about the package-lock file has been made.

We decided to commit this.

vasilmkd:
I set up a small node project locally.

```
npm init
npm install --save jsdom
```

I copied code from their readme to get a basic set up going:

```js
// From the jsdom readme
const jsdom = require("jsdom");
const { JSDOM } = jsdom;

// these 4 lines are mine
console.log(global.setImmediate);
global.setImmediate(() => console.log('hello setImmediate'));
console.log(setImmediate);
setImmediate(() => console.log('hello again setImmediate'));

// from the jsdom readme
const dom = new JSDOM(`<!DOCTYPE html><p>Hello world</p>`);
console.log(dom.window.document.querySelector("p").textContent); // "Hello world"

```

And I'm seeing this output:
```
[Function: setImmediate] {
  [Symbol(nodejs.util.promisify.custom)]: [Getter]
}
[Function: setImmediate] {
  [Symbol(nodejs.util.promisify.custom)]: [Getter]
}
Hello world
hello setImmediate
hello again setImmediate
```

This IMO means that `setImmediate` is not being disabled by `jsdom`.

armanbilge:
I mean, that all seems reasonable ... now try running it again through the ScalaJS test bridge? https://github.com/scala-js/scala-js-env-jsdom-nodejs

vasilmkd:
It fails there, or rather reports `undefined`. That doesn't mean that `jsdom` is doing something to disable it. It could just mean that our discovery process is wrong.

armanbilge:
> It could just mean that our discovery process is wrong.

That may very well be true, but I feel very inclined not to touch it. JSDOM is a fake environment, only used for testing. Whatever works should suffice.

vasilmkd:
I'm just using the PR to document all discoveries, I think this is more or less good to go.

armanbilge:
> I'm just using the PR to document all discoveries, I think this is more or less good to go.

Oh, absolutely, I appreciate your research efforts! Apologies if came across curt.

vasilmkd:
Found it, `scalajs-js-env-jsdom-nodejs` seems to hide a lot of `global` things https://github.com/scala-js/scala-js-env-jsdom-nodejs/pull/19.

vasilmkd (approved):

armanbilge:
Cool, that's what I thought!

djspiewak (approved):

PR #2193, opened 2021-08-03
Split laws tests into separate specs

vasilmkd:


djspiewak:
Hmm, what's the general motivation here? I usually really like having the law tests together with the other tests unless we're forced to pull them apart, and some of these types don't *have* non-law tests.

vasilmkd:
I was trying to see if it helps the Firefox tests. I didn't manage to accomplish anything really. The changes with inheritance and the cleanups I discovered were merged in #2199, so this PR can fall into oblivion.

PR #2202, opened 2021-08-05
Converted startup runtime collision into a warning

### LINKED ISSUE #1918: Initialization error when using unsafe.implicits.global within IOApp
While updating some examples that demonstrate what *not to do*, I bumped into this issue. Here's my example.

```scala

import cats.effect._
import cats.effect.std.{ Semaphore, Supervisor }

object LeakyState extends IOApp.Simple {

  // global access
  val sem: Semaphore[IO] =
    Semaphore[IO](1).unsafeRunSync()

  def launchMissiles: IO[Unit] =
    sem.permit
      .use { _ =>
        IO.println("Launching missiles") >>
          IO.sleep(3.seconds) >>
          IO.println("Missiles launched, releasing lock")
      }

  def randomSleep: IO[Unit] =
    IO(scala.util.Random.nextInt(100)).flatMap { ms =>
      IO.sleep((ms + 700).millis)
    }.void

  def p1: IO[Unit] =
    sem.permit.use(_ => IO.println("Running P1")) >> randomSleep

  def p2: IO[Unit] =
    sem.permit.use(_ => IO.println("Running P2")) >> randomSleep

  def run: IO[Unit] =
    Supervisor[IO].use { s =>
      s.supervise(launchMissiles) *>
        s.supervise(p1.foreverM).void *>
        s.supervise(p2.foreverM).void *>
        IO.sleep(5.seconds).void
    }

}
```

Initializing (unsafely) the `Semaphore` fails with this error.

```scala
[info] running (fork) examples.state.LeakyState
[error] Exception in thread "main" java.lang.IllegalArgumentException: requirement failed
[error] 	at scala.Predef$.require(Predef.scala:325)
[error] 	at cats.effect.unsafe.IORuntimeCompanionPlatform.installGlobal(IORuntimeCompanionPlatform.scala:67)
[error] 	at cats.effect.IOApp.main(IOApp.scala:58)
[error] 	at cats.effect.IOApp.main$(IOApp.scala:35)
[error] 	at examples.state.LeakyState$.main(LeakyState.scala:9)
[error] 	at examples.state.LeakyState.main(LeakyState.scala)
```

Even when setting `Compile / run / fork := true` in `build.sbt`.

The workaround is to make it a `lazy val` together with forking enabled in `sbt`. So this does not work in Scastie: https://scastie.scala-lang.org/v7S7tnCsQ9yAL9gMSnxyhw

Since this example showcases exactly what users should not do, I don't expect a fix but raising the issue in case this could be a legit bug when interoping with side-effectful frameworks, e.g. initializing an `ActorSystem` within `IOApp`.

djspiewak:
Fixes #1918 

When the global runtime is *already* initialized by the time `main` runs, we now treat this as a warning (which prints to stderr) rather than a fatal error (which is what we did previously). The warning is still important because the user really needs to know that we're ignoring their custom configurations.

We can probably do better than the text on this warning and give the user some actionable steps to track down *why* the warning is appearing.

### DIFF
--- a/core/js/src/main/scala/cats/effect/IOApp.scala
+++ b/core/js/src/main/scala/cats/effect/IOApp.scala
@@ -189,13 +189,20 @@ trait IOApp {
     if (runtime == null) {
       import unsafe.IORuntime
 
-      IORuntime installGlobal {
+      val installed = IORuntime installGlobal {
         IORuntime(
           IORuntime.defaultComputeExecutionContext,
           IORuntime.defaultComputeExecutionContext,
           IORuntime.defaultScheduler,
           () => (),
           runtimeConfig)
+      }
+
+      if (!installed) {
+        System
+          .err
+          .println(
+            "WARNING: Cats Effect global runtime already initialized; custom configurations will be ignored")
       }
 
       _runtime = IORuntime.global
--- a/build.sbt
+++ b/build.sbt
@@ -361,7 +361,8 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
       ProblemFilters.exclude[MissingClassProblem]("cats.effect.SyncIO$Delay"),
       ProblemFilters.exclude[DirectMissingMethodProblem]("cats.effect.IO#IOCont.apply"),
       ProblemFilters.exclude[DirectMissingMethodProblem]("cats.effect.IO#IOCont.copy"),
-      ProblemFilters.exclude[DirectMissingMethodProblem]("cats.effect.IO#IOCont.this")
+      ProblemFilters.exclude[DirectMissingMethodProblem]("cats.effect.IO#IOCont.this"),
+      ProblemFilters.exclude[IncompatibleMethTypeProblem]("cats.effect.unsafe.IORuntimeCompanionPlatform.installGlobal")
     )
   )
   .jvmSettings(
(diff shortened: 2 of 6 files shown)


core/js/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -198,6 +198,13 @@ trait IOApp {
           runtimeConfig)
       }
 
+      if (!installed) {
+        System
+          .err
+          .println(
+            "WARNING: Cats Effect global runtime already initialized; custom configurations will be ignored")
```
vasilmkd, line 205: Should we do the fancy yellow colored [text](https://www.scala-lang.org/api/2.13.6/scala/Console$.html#YELLOW:String(%22\u001B[33m%22))?

vasilmkd (commented):
Bummer, it looks like you forgot the mima filters.

core/js/src/main/scala/cats/effect/IOApp.scala
djspiewak: We could! It generally looks terrible in log output though. Am open to bikeshedding on this one.

vasilmkd: I agree with function over form in this case.

vasilmkd (approved):

PR #2203, opened 2021-08-05
Always enhance caught `Throwable`s

vasilmkd:
This should match the behavior of Cats Effect 2.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -418,7 +418,11 @@ private final class IOFiber[A](
 
             case 1 =>
               val error = ioa.asInstanceOf[Error]
-              runLoop(succeeded(Left(error.t), 0), nextCancelation - 1, nextAutoCede)
+              val t = error.t
+              // We need to augment the exception here because it doesn't get
+              // forwarded to the `failed` path.
+              Tracing.augmentThrowable(runtime.config.enhancedExceptions, t, tracingEvents)
+              runLoop(succeeded(Left(t), 0), nextCancelation - 1, nextAutoCede)
 
             case 2 =>
               val delay = ioa.asInstanceOf[Delay[Any]]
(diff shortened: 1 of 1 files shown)


vasilmkd:
I specifically pushed 2 commits, one augments `NonFatal` `Throwable`s and the other fatal errors. Let me know if I should revert the fatal error case. I know we're not supposed to allocate objects in that case, but I await further instructions.

vasilmkd:
This is not as straightforward as it appears.

vasilmkd:
Or maybe it is.

vasilmkd:
Code:
```scala
import cats.data._
import cats.effect._
import cats.syntax.all._

  def run: IO[Unit] =
    IO.raiseError(new Exception("blah")).attempt.flatTap {
      case Left(t) => IO.delay(t.printStackTrace())
      case Right(_) => ???
    }.void
}
```

Cats Effect 2 output:
```
java.lang.Exception: blah
	at Main$.run(Main.scala:7)
	at flatTap @ Main$.run(Main.scala:7)
	at void @ Main$.run(Main.scala:7)
	at run$ @ Main$.run(Main.scala:5)
	at main$ @ Main$.main(Main.scala:5)
```

Cats Effect 3 new output (This PR):
```
java.lang.Exception: blah
	at Main$.run(Main.scala:7)
	at flatTap @ Main$.run(Main.scala:7)
	at void @ Main$.run(Main.scala:7)
	at run$ @ Main$.run(Main.scala:5)
```

cc @Daenyth

vasilmkd:
The idea is the following:

Each time an exception is caught, we push it through the `failed` path of the runloop. It makes sense to augment the exception here.

Funnily enough, this works for everything _except_ the immediate interpreter for `Attempt`, which directly sends caught exceptions to the `succeeded` path by wrapping them in a `Left(...)`. That's why, the two cases where this happens have special cased augmentation and are marked with comments.

djspiewak (approved):
Awesome!

Issue #2146, opened 2021-07-24
Implement tracing on ScalaJS


djspiewak:
It should be possible given that exceptions work roughly the same.

PR #2207, opened 2021-08-06
Implement tracing for JS

armanbilge:
Towards #2146. This turned out primarily to be a config problem, full tracing seems to work well out-of-the-box on JS, all credit to @vasilmkd.

For configuration, I've created a global `TracingConfig` with a mutable "var" `mode`. Note this means the tracing configuration could be changed while running, not sure if that's reasonable or not.

Regarding cached tracing: the current approach of using the class of the lambda to identify it doesn't seem to work on JS, because all lambdas are the class [`AnonFunction1`](https://github.com/scala-js/scala-js/blob/master/library/src/main/scala/scala/scalajs/runtime/AnonFunctions.scala#L26). Follow-up PRs could possibly figure out a workaround for this.

Thanks in advance!


armanbilge:
Ok, browsers get different stack traces...

vasilmkd:
Can you please post what traces look like on `fastOptJS` and `fullOptJS` when you have the time? Thanks. I want to visualize and document what they look like.

armanbilge:
fastOptJS on Node.js
```
java.lang.Exception
  at {anonymous}()(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-fastopt/main.js:209543:34)
  at scala.scalajs.runtime.AnonFunction1.apply(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-fastopt/main.js:84110:43)
  at flatMap @ tracing.TracingSpec.loop$1(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-fastopt/main.js:209539)
  at flatMap @ tracing.TracingSpec.loop$1(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-fastopt/main.js:209539)
  ...
```

fastOptJS on Firefox
```
java.lang.Exception
  at $c_jl_Throwable(file:///workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-fastopt/main.js:46101:5)
  at $c_jl_Exception(file:///workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-fastopt/main.js:71021:1)
  at tracing.TracingSpec.loop$1(file:///workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-fastopt/main.js:209543:34)
  at scala.scalajs.runtime.AnonFunction1.apply(file:///workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-fastopt/main.js:84110:43)
  at flatMap @ tracing.TracingSpec.loop$1(file:///workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-fastopt/main.js:209539)
  at flatMap @ tracing.TracingSpec.loop$1(file:///workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-fastopt/main.js:209539)
  ...
```

fullOptJS on Node.js
```
java.lang.Exception
  at Vr(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:1633:408)
  at {anonymous}()(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:4478:374)
  at x.c(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:1931:378)
  at e4a(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:2945:270)
  at T4(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:2926:158)
  at KI.Mj(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:2949:222)
  at {anonymous}()(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:953:61)
  at L.U(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:1931:158)
  at Immediate._onImmediate(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:1930:356)
  at processImmediate(internal/timers.js:464:21)
  at Yn.e.Fl @ <jscode>.Yn.e.Fb(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:2478)
  at Yn.e.Fl @ <jscode>.Yn.e.Fb(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:2478)
  ...
```

fullOptJS on Firefox
```
java.lang.Exception
  at Wh(file:///workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:1098:38)
  at Kv(file:///workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:1633:446)
  at Vr(file:///workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:1633:408)
  at Wib/</<(file:///workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:4478:374)
  at x.prototype.c(file:///workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:1931:378)
  at e4a(file:///workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:2945:270)
  at T4(file:///workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:2926:161)
  at KI.prototype.Mj(file:///workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:2949:222)
  at dL.prototype.vt/</<(file:///workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:953:61)
  at L.prototype.U(file:///workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:1931:158)
  at jDa(file:///workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:950:163)
  at lDa(file:///workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:950:342)
  at dL/</<(file:///workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:951:294)
  at x.prototype.c(file:///workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:1931:378)
  at nDa/</<(file:///workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:1930:397)
  at Wh @ <jscode>.Zn(file:///workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:2069)
  at Wh @ <jscode>.Zn(file:///workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/main.js:2069)
  ...
```

core/js/src/main/scala/cats/effect/tracing/TracingPlatform.scala
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  private def calculateTracingEvent(key: String): TracingEvent = {
+    if (isCachedStackTracing) {
+      Try(cache(key)).recover {
+        case _ =>
+          val event = buildEvent()
+          cache(key) = event
+          event
+      }.get
```
vasilmkd: Can we change this to `cache.get(key) match { case Some | case None }`?

vasilmkd (dismissed):

core/js/src/main/scala/cats/effect/tracing/TracingPlatform.scala
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  // We could have a catch-all for non-functions, but explicitly enumerating makes sure we handle each case correctly
+  def calculateTracingEvent[F[_], A, B](cont: Cont[F, A, B]): TracingEvent = {
+    calculateTracingEvent(cont.getClass().getName())
+  }
+
+  private def calculateTracingEvent(key: String): TracingEvent = {
+    if (isCachedStackTracing)
+      cache.getOrElseUpdate(key, buildEvent())
```
vasilmkd, line 52: Even better. I wasn't sure if this would work on `2.12`, but I guess Scala.js has a unified API.

armanbilge: Huh! I'm surprised about that.

core/shared/src/main/scala/cats/effect/tracing/Tracing.scala
```diff
@@ -52,15 +38,34 @@ private[effect] object Tracing extends ClassValue[TracingEvent] {
...
-        return true
+  private[this] final val stackTraceMethodNameFilter: Array[String] = Array(
+    "$c_jl_",
+    "$c_Lcats_effect_"
+  )
+
+  private[this] def applyStackTraceFilter(
+      callSiteClassName: String,
+      callSiteMethodName: String): Boolean = {
+    if (callSiteClassName == "<jscode>") {
```
djspiewak: IMO it would be better to make this platform-specific. It's not a *huge* deal, but it is user-visible performance, and it follows the idiom of the rest of the project sources.

tests/shared/src/test/scala/cats/effect/tracing/TracingSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+ * limitations under the License.
+ */
+
+package tracing // Get out of the CE package so our traces don't get filtered
+
+import cats.effect.testkit.TestInstances
+import cats.effect.BaseSpec
+import cats.effect.IO
+
+class TracingSpec extends BaseSpec with TestInstances {
```
djspiewak, line 24: :+1:

core/js/src/main/scala/cats/effect/tracing/TracingConfig.scala
```diff
@@ -0,0 +1,49 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.tracing
+
+object TracingConfig {
```
djspiewak: So the only way to set the tracing mode is from within scalajs sources? There's no more "javascript native" way of doing this?

djspiewak (dismissed):
Aside from the minor comments, I think it would be worth benchmarking performance on JVM and JS just to make sure we haven't regressed the former and also get a feel for the performance impact on the latter. We're just guessing right now.

core/js/src/main/scala/cats/effect/tracing/TracingConfig.scala
armanbilge: There _are_ more idiomatic ways, but only if you assume Node.js or Webpack etc. The nice thing about those ways, is that during bundling the boolean values could probably get hard-coded and optimized.

vasilmkd:
We've already discussed running JVM benchmarks on Discord. I'll do a round too. Not sure how to rigorously benchmark Scala.js.

core/js/src/main/scala/cats/effect/tracing/TracingConfig.scala
vasilmkd: That's what we're relying on for JVM so that would be nice, however I'm not sure assuming `node` or `webpack` is the way to go.

core/shared/src/main/scala/cats/effect/tracing/Tracing.scala
armanbilge: I can do this, but I'm not crazy about the additional indirection it introduces. If it's a performance concern, what if we introduce a static final boolean flag on JVM side so this gets optimized away?

armanbilge:
Yep I'll post JVM benchmark soon. For JS benchmark there's this? https://github.com/japgolly/scalajs-benchmark

core/shared/src/main/scala/cats/effect/tracing/Tracing.scala
vasilmkd: I don't consider the tracing code as a prime example of teachable or approachable code. I'd like to see this change.

armanbilge: Ok, consider it done!

core/js/src/main/scala/cats/effect/tracing/TracingConfig.scala
armanbilge: Even _if_ we did make that assumption, their idioms are different. So I'm still not sure how could do this without it ultimately being dynamic in some form.

armanbilge:
JVM benchmark:

js-tracing:
```
sbt:root> benchmarks/Jmh/run -wi 10 -i 10 -f 2 -t 1 DeepBindBenchmark
[info] DeepBindBenchmark.async   10000  thrpt   20   955.690 ± 105.005  ops/s
[info] DeepBindBenchmark.delay   10000  thrpt   20  1535.141 ±  94.538  ops/s
[info] DeepBindBenchmark.pure    10000  thrpt   20  2916.520 ± 273.923  ops/s
```

main:
```
[info] DeepBindBenchmark.async   10000  thrpt   20   934.375 ±  28.209  ops/s
[info] DeepBindBenchmark.delay   10000  thrpt   20  1888.475 ± 118.869  ops/s
[info] DeepBindBenchmark.pure    10000  thrpt   20  2894.295 ±  92.649  ops/s
```

js-tracing:
```
sbt:root> benchmarks/Jmh/run -wi 10 -i 10 -f 2 -t 1 DeepBindBenchmark --jvmArgs -Dcats.effect.tracing.mode=full
[info] Benchmark                (size)   Mode  Cnt   Score   Error  Units
[info] DeepBindBenchmark.async   10000  thrpt   20  34.582 ± 1.666  ops/s
[info] DeepBindBenchmark.delay   10000  thrpt   20  49.021 ± 1.346  ops/s
[info] DeepBindBenchmark.pure    10000  thrpt   20  80.893 ± 6.488  ops/s
```

main:
```
sbt:root> benchmarks/Jmh/run -wi 10 -i 10 -f 2 -t 1 DeepBindBenchmark --jvmArgs -Dcats.effect.tracing.mode=full
[info] Benchmark                (size)   Mode  Cnt   Score   Error  Units
[info] DeepBindBenchmark.async   10000  thrpt   20  32.616 ± 1.145  ops/s
[info] DeepBindBenchmark.delay   10000  thrpt   20  45.216 ± 1.346  ops/s
[info] DeepBindBenchmark.pure    10000  thrpt   20  89.558 ± 2.072  ops/s
```

vasilmkd:
I'm also fine with keeping the mutable tracing mode for JS as is for now. JS tracing would be a really new thing in Scala (I think), so let's let people discover it, take it for a spin, maybe report some issues and then we can iterate on the configuration. Let's just make sure we don't paint ourselves in a corner where we have to introduce config source incompatible changes.

armanbilge:
Ok, here's an interesting one! I added the `--enable-source-maps` flag for the NodeJS environment (also requires installing an npm package `source-map-support`) and this is what pops out:

fastOptJS
```
java.lang.Exception
  at {anonymous}()(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-fastopt/https:/raw.githubusercontent.com/typelevel/cats-effect/c7b34177d09c510123348ba37d90ebcb699868cc/tests/shared/src/test/scala/cats/effect/tracing/TracingSpec.scala:30:27)
  at scala.scalajs.runtime.AnonFunction1.apply(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-fastopt/https:/raw.githubusercontent.com/scala-js/scala-js/v1.7.0/library/src/main/scala/scala/scalajs/runtime/AnonFunctions.scala:27:38)
  at flatMap @ tracing.TracingSpec.loop$1(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-fastopt/https:/raw.githubusercontent.com/typelevel/cats-effect/c7b34177d09c510123348ba37d90ebcb699868cc/tests/shared/src/test/scala/cats/effect/tracing/TracingSpec.scala:28)
  at flatMap @ tracing.TracingSpec.loop$1(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-fastopt/https:/raw.githubusercontent.com/typelevel/cats-effect/c7b34177d09c510123348ba37d90ebcb699868cc/tests/shared/src/test/scala/cats/effect/tracing/TracingSpec.scala:28)
  ...
```

fullOptJS:
```
java.lang.Exception
  at java.lang.Exception.<init>(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/file:/home/lars/proj/typelevel/scalacheck/src/main/scala/org/scalacheck/Arbitrary.scala:165:27)
  at {anonymous}()(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/https:/raw.githubusercontent.com/typelevel/cats-effect/c7b34177d09c510123348ba37d90ebcb699868cc/tests/shared/src/test/scala/cats/effect/tracing/TracingSpec.scala:30:27)
  at x.apply__O__O [as c](/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/https:/raw.githubusercontent.com/scala-js/scala-js/v1.7.0/library/src/main/scala/scala/scalajs/runtime/AnonFunctions.scala:27:38)
  at calculateTracingEvent @ <jscode>.calculateTracingEvent__F1__Lcats_effect_tracing_TracingEvent(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/https:/raw.githubusercontent.com/typelevel/cats-effect/c7b34177d09c510123348ba37d90ebcb699868cc/core/js/src/main/scala/cats/effect/tracing/TracingPlatform.scala:34)
  at calculateTracingEvent @ <jscode>.calculateTracingEvent__F1__Lcats_effect_tracing_TracingEvent(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/https:/raw.githubusercontent.com/typelevel/cats-effect/c7b34177d09c510123348ba37d90ebcb699868cc/core/js/src/main/scala/cats/effect/tracing/TracingPlatform.scala:34)
  ...
```

I think with some additional tweaking even `fullOptJS` can give useful traces with this enabled :)

vasilmkd:
Here are some results (cached):

`series/3.x`:
```
DeepBindBenchmark.async   10000  thrpt   20  1502.144 ±  6.247  ops/s
DeepBindBenchmark.delay   10000  thrpt   20  3428.198 ± 53.450  ops/s
DeepBindBenchmark.pure    10000  thrpt   20  4998.132 ± 16.141  ops/s
```

`This PR`:
```
DeepBindBenchmark.async   10000  thrpt   20  1477.639 ± 24.307  ops/s
DeepBindBenchmark.delay   10000  thrpt   20  3360.015 ± 15.644  ops/s
DeepBindBenchmark.pure    10000  thrpt   20  5008.959 ±  8.807  ops/s
```

core/js/src/main/scala/cats/effect/tracing/TracingPlatform.scala
```diff
@@ -0,0 +1,91 @@
...
+  private val cache = js.Dictionary.empty[TracingEvent]
+
+  import TracingConstants._
+
+  def calculateTracingEvent[A](f: Function0[A]): TracingEvent = {
+    calculateTracingEvent(f.asInstanceOf[scalajs.js.Dynamic].sjsr_AnonFunction0__f_f.toString())
+  }
+
+  def calculateTracingEvent[A, B](f: Function1[A, B]): TracingEvent = {
+    calculateTracingEvent(f.asInstanceOf[scalajs.js.Dynamic].sjsr_AnonFunction1__f_f.toString())
```
armanbilge: @djspiewak this magic breaks after `fullOptJS` :(

armanbilge:
Ok, tracing for `fullOptJS` Node.js w/ sourcemaps now works. It looks like this:
```
java.lang.Exception
  at java.lang.Exception.<init>(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/file:/home/lars/proj/typelevel/scalacheck/src/main/scala/org/scalacheck/Arbitrary.scala:165:27)
  at {anonymous}()(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/https:/raw.githubusercontent.com/typelevel/cats-effect/c7b34177d09c510123348ba37d90ebcb699868cc/tests/shared/src/test/scala/cats/effect/tracing/TracingSpec.scala:30:27)
  at x.apply__O__O [as c](/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/https:/raw.githubusercontent.com/scala-js/scala-js/v1.7.0/library/src/main/scala/scala/scalajs/runtime/AnonFunctions.scala:27:38)
  at flatMap @ tracing.TracingSpec.loop$1(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/https:/raw.githubusercontent.com/typelevel/cats-effect/c7b34177d09c510123348ba37d90ebcb699868cc/tests/shared/src/test/scala/cats/effect/tracing/TracingSpec.scala:28)
  at flatMap @ tracing.TracingSpec.loop$1(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-opt/https:/raw.githubusercontent.com/typelevel/cats-effect/c7b34177d09c510123348ba37d90ebcb699868cc/tests/shared/src/test/scala/cats/effect/tracing/TracingSpec.scala:28)
  ...
```

In case you missed it, I have no idea why @larsrh's workspace is leaking into the stack trace!

Some notes/questions:
1. I've only tested this with source maps enabled for Node.js. It's should be possible for the browser too, just not sure how to configure it.
2. Should we run tests with the `--enable-source-maps` flag enabled for Node.js?
3. I know we are already far over-extended, but should we also add a `fullOptJS` run to the test matrix at least for Node.js? Otherwise this stuff isn't under test.

vasilmkd:
Here are my thoughts on everything up to now:

Regarding `node`:
1. It looks like `--enable-source-maps` is the way to go. I think it's a no brainer and should be our official recommendation for people who want usable stack traces on `node`.
2. I would _really_ love to explore a `-Darg` or environment variable approach to setting the tracing mode on `node`, similarly to what we have on the JVM.

This however leaves me a bit unsure what to do for browsers.
I may be wrong, but my gut feeling says to choose a sane default, let's say `cached` and make it unchangeable. Having a mutable tracing mode is bugging me for some reason that I cannot explain. _However_, I wouldn't be opposed to having a hidden method that one has to access through duck typing `.asInstanceOf[{ def setTracingMode(mode: Mode): Unit }]` like we have for the `resetGlobal` method.

I'm just rambling, I'm not sure what to do exactly. I just want to get this right.

armanbilge:
> I would really love to explore a -Darg or environment variable approach to setting the tracing mode on node, similarly to what we have on the JVM.

So I looked into this maybe last week, and I found three different idioms for these sorts of runtime global variables:

1. On Node.js, `process.env` to access environment variables: https://nodejs.org/dist/latest/docs/api/process.html#process_process_env
2. On react-scripts, _also_ `process.env`: https://create-react-app.dev/docs/adding-custom-environment-variables/
3. For webpack (this is what Scala.js bundler uses), an `env` variable available only in `webpack.config.js`: https://webpack.js.org/guides/environment-variables/

armanbilge:
Ok, webpack also has support for `process.env` via a plugin: https://webpack.js.org/plugins/environment-plugin/

So here's a proposal: what if we are opinionated (!!), and implement this via `process.env` with the fallback to `cached` mode if it doesn't exist/isn't set and also possibly that duck-typed backdoor?

armanbilge:
Some benchmark results using #2229. Take with a grain of salt!

main fastOptJS
```
Starting suite: deep bind
[0/3] async 10000 : 34.643 ± 1.435 ops/s
[1/3] delay 10000 : 513.194 ± 21.898 ops/s
[2/3] pure  10000 : 809.888 ± 32.886 ops/s
Suite completed: deep bind
Starting suite: shallow bind
[0/3] async 10000 : 34.668 ± 1.325 ops/s
[1/3] delay 10000 : 531.062 ± 20.492 ops/s
[2/3] pure  10000 : 809.632 ± 21.992 ops/s
Suite completed: shallow bind
```

js-tracing fastOptJS cached
```
Starting suite: deep bind
[0/3] async 10000 : 20.178 ± 0.730 ops/s
[1/3] delay 10000 : 101.374 ± 3.814 ops/s
[2/3] pure  10000 : 153.217 ± 4.441 ops/s
Suite completed: deep bind
Starting suite: shallow bind
[0/3] async 10000 : 25.509 ± 1.181 ops/s
[1/3] delay 10000 : 120.882 ± 2.255 ops/s
[2/3] pure  10000 : 176.194 ± 4.611 ops/s
Suite completed: shallow bind
```

js-tracing fastOptJS full
```
Starting suite: deep bind
[0/3] async 10000 : 2.370 ± 0.074 ops/s
[1/3] delay 10000 : 3.160 ± 0.159 ops/s
[2/3] pure  10000 : 6.074 ± 0.120 ops/s
Suite completed: deep bind
Starting suite: shallow bind
[0/3] async 10000 : 2.585 ± 0.063 ops/s
[1/3] delay 10000 : 3.010 ± 0.073 ops/s
[2/3] pure  10000 : 5.609 ± 0.291 ops/s
Suite completed: shallow bind
```

js-tracing fastOptJS off
```
Starting suite: deep bind
[0/3] async 10000 : 31.602 ± 1.645 ops/s
[1/3] delay 10000 : 250.622 ± 7.983 ops/s
[2/3] pure  10000 : 478.220 ± 16.908 ops/s
Suite completed: deep bind
Starting suite: shallow bind
[0/3] async 10000 : 32.814 ± 1.351 ops/s
[1/3] delay 10000 : 263.651 ± 10.545 ops/s
[2/3] pure  10000 : 479.176 ± 21.826 ops/s
Suite completed: shallow bind
```

main fullOptJS
```
Starting suite: deep bind
[0/3] async 10000 : 35.911 ± 1.151 ops/s
[1/3] delay 10000 : 577.792 ± 36.765 ops/s
[2/3] pure  10000 : 927.688 ± 54.904 ops/s
Suite completed: deep bind
Starting suite: shallow bind
[0/3] async 10000 : 36.322 ± 1.990 ops/s
[1/3] delay 10000 : 591.551 ± 17.359 ops/s
[2/3] pure  10000 : 872.215 ± 30.065 ops/s
Suite completed: shallow bind
```

js-tracing fullOptJS cached
```
Starting suite: deep bind
[0/3] async 10000 : 23.979 ± 1.021 ops/s
[1/3] delay 10000 : 158.711 ± 4.681 ops/s
[2/3] pure  10000 : 279.770 ± 13.381 ops/s
Suite completed: deep bind
Starting suite: shallow bind
[0/3] async 10000 : 27.788 ± 0.784 ops/s
[1/3] delay 10000 : 157.197 ± 4.846 ops/s
[2/3] pure  10000 : 332.340 ± 12.876 ops/s
Suite completed: shallow bind
```

js-tracing fullOptJS full
```
Starting suite: deep bind
[0/3] async 10000 : 2.379 ± 0.056 ops/s
[1/3] delay 10000 : 3.013 ± 0.077 ops/s
[2/3] pure  10000 : 6.413 ± 0.179 ops/s
Suite completed: deep bind
Starting suite: shallow bind
[0/3] async 10000 : 2.704 ± 0.076 ops/s
[1/3] delay 10000 : 2.925 ± 0.075 ops/s
[2/3] pure  10000 : 6.153 ± 0.276 ops/s
Suite completed: shallow bind
```

js-tracing fullOptJS off
```
Starting suite: deep bind
[0/3] async 10000 : 31.740 ± 1.101 ops/s
[1/3] delay 10000 : 265.321 ± 8.767 ops/s
[2/3] pure  10000 : 501.158 ± 14.615 ops/s
Suite completed: deep bind
Starting suite: shallow bind
[0/3] async 10000 : 32.874 ± 1.674 ops/s
[1/3] delay 10000 : 263.035 ± 9.866 ops/s
[2/3] pure  10000 : 494.882 ± 35.125 ops/s
Suite completed: shallow bind
```

djspiewak:
80% slower :-( That's a little steep. I wonder where that's coming from? Maybe using the lambda source is more expensive than we thought?

armanbilge:
Yeah, could be, but then idk what to do :( I'm also going to experiment with alternative cache implementations i.e. `js.Dictionary` vs JS-native map vs scala `mutable.Map` to see if that matters.

djspiewak:
Actually Vasil pointed out that the async benchmarks are *way* closer. Not sure which number to believe.

armanbilge:
Also, like, no guarantees that I didn't completely mess up the benchmarking code 😅

armanbilge:
Consistent with #2231, the choice of map implementation for the cache doesn't appear to make a difference.

Update: benchmarks on 2.13 / Node.js

`fastOptJS` / `js.Dictionary`
```
Starting suite: deep bind
[0/3] async 10000 : 21.425 ± 0.425 ops/s
[1/3] delay 10000 : 104.700 ± 1.772 ops/s
[2/3] pure  10000 : 152.867 ± 1.734 ops/s
Suite completed: deep bind
Starting suite: shallow bind
[0/3] async 10000 : 26.335 ± 0.469 ops/s
[1/3] delay 10000 : 122.488 ± 1.983 ops/s
[2/3] pure  10000 : 179.601 ± 8.012 ops/s
Suite completed: shallow bind
```

`fastOptJS` / `mutable.Map`
```
Starting suite: deep bind
[0/3] async 10000 : 13.560 ± 0.406 ops/s
[1/3] delay 10000 : 53.623 ± 1.591 ops/s
[2/3] pure  10000 : 66.011 ± 0.538 ops/s
Suite completed: deep bind
Starting suite: shallow bind
[0/3] async 10000 : 19.907 ± 0.611 ops/s
[1/3] delay 10000 : 63.358 ± 1.549 ops/s
[2/3] pure  10000 : 86.019 ± 2.392 ops/s
Suite completed: shallow bind
```

`fullOptJS` / `js.Dictionary`
```
Starting suite: deep bind
[0/3] async 10000 : 19.579 ± 1.591 ops/s
[1/3] delay 10000 : 107.697 ± 6.111 ops/s
[2/3] pure  10000 : 192.376 ± 9.413 ops/s
Suite completed: deep bind
Starting suite: shallow bind
[0/3] async 10000 : 22.223 ± 1.235 ops/s
[1/3] delay 10000 : 118.007 ± 7.177 ops/s
[2/3] pure  10000 : 239.164 ± 17.869 ops/s
Suite completed: shallow bind
```

`fullOptJS` / `mutable.Map`
```
Starting suite: deep bind
[0/3] async 10000 : 23.644 ± 0.557 ops/s
[1/3] delay 10000 : 114.980 ± 3.476 ops/s
[2/3] pure  10000 : 173.507 ± 9.118 ops/s
Suite completed: deep bind
Starting suite: shallow bind
[0/3] async 10000 : 26.318 ± 1.568 ops/s
[1/3] delay 10000 : 136.355 ± 6.070 ops/s
[2/3] pure  10000 : 238.910 ± 7.585 ops/s
Suite completed: shallow bind
```

core/js/src/main/scala/cats/effect/tracing/TracingConstants.scala
```diff
@@ -16,12 +16,27 @@
...
 
-  var isCachedStackTracing: Boolean = true
+  @inline def stackTracingMode: String =
+    Option(js.Dynamic.global.process)
+      .filterNot(js.isUndefined)
+      .flatMap(p => Option(p.env).filterNot(js.isUndefined))
+      .flatMap { env =>
+        Option(env.CATS_EFFECT_TRACING_MODE)
+          .filterNot(js.isUndefined)
+          .orElse(Option(env.REACT_APP_CATS_EFFECT_TRACING_MODE).filterNot(js.isUndefined))
```
armanbilge: I think this is an important enough case to hard-code in, although I have no idea if anyone actually uses create-react-app with Scala.js. When using create-react-app all custom environment variables must be prefixed with `REACT_APP_`. See https://create-react-app.dev/docs/adding-custom-environment-variables/

armanbilge:
I think this is almost ready, there is one last tricky issue: in #2221 IIUC the consensus is to use `process.env` to define the tracing mode via an environment variable, e.g. `process.env.CATS_EFFECT_TRACING_MODE`.

Furthermore, my goal was for browser-based users to be able to configure this via a [webpack plugin](https://webpack.js.org/plugins/environment-plugin/) such that the branches for tracing can be eliminated at bundling time.

Unfortunately, this is turning out to be trickier than I thought due to an ugly combination of factors:
1. Said webpack plugin is essentially doing a text replacement of `process.env.CATS_EFFECT_TRACING_MODE` with the desired value.  Note that defining `process.env.CATS_EFFECT_TRACING_MODE` in webpack will _not_ define `process` nor `process.env`.
2. [Scala.js semantics](https://www.scala-js.org/doc/interoperability/global-scope.html#practical-tips) require us to gradually check if `process` is defined, then `process.env`, before accessing `process.env.CATS_EFFECT_TRACING_MODE` to avoid throwing an error.
3. If we bite the bullet and do something like `Try(process.env.CATS_EFFECT_TRACING_MODE)`, then even after webpack simplifies this to `Try("full")` or similar, none of the JS optimizers/bundlers I played with will remove the surrounding try-catch. This obstructs branch elimination.

Do we care enough about this that we should explore other options? Or is it reasonable that browser users will be hit by (whatever) performance penalty of these branches regardless of their selected tracing mode? Thanks.

armanbilge:
[Some reading on stack traces in V8.](https://v8.dev/blog/v8-lite#lazy-source-positions) Unfortunately I think they de-optimized our specific use-case.

> When compiling bytecode from JavaScript, source position tables are generated that tie bytecode sequences to character positions within the JavaScript source code. However, this information is only needed when symbolizing exceptions or performing developer tasks such as debugging, and so is rarely used.
> 
> To avoid this waste, we now compile bytecode without collecting source positions (assuming no debugger or profiler is attached). The source positions are only collected when a stack trace is actually generated, for instance when calling `Error.stack` or printing an exception’s stack trace to the console. This does have some cost, as generating source positions requires the function to be reparsed and compiled, however most websites don’t symbolize stack traces in production and therefore don’t see any observable performance impact.

djspiewak:
The timeouts in CI are concerning.

Some other thoughts…

- Are we now requiring source maps? What happens if someone downstream doesn't *want* to use source maps?
- Is the consensus that the performance impact is manageable? It really sucks that V8 specifically deoptimized this case, though if I'm reading it correctly, it sounds like this is only a one-time penalty.
- Looking through the diff, it *seems* like the `process.env` approach is working? Am I missing anything?

build.sbt
```diff
@@ -78,7 +79,7 @@ ThisBuild / githubWorkflowBuildPreamble ++= Seq(
   WorkflowStep.Run(
     List("npm install"),
     name = Some("Install jsdom"),
```
djspiewak: This is installing more than just jsdom now.

core/js/src/main/scala/cats/effect/tracing/TracingConstants.scala
```diff
@@ -16,15 +16,24 @@
...
+    Try(js.Dynamic.global.process.env.CATS_EFFECT_TRACING_MODE)
+      .toOption
+      .orElse(Try(js.Dynamic.global.process.env.REACT_APP_CATS_EFFECT_TRACING_MODE).toOption)
+      .filterNot(js.isUndefined)
+      .map(_.asInstanceOf[String])
+      .filterNot(_.isEmpty)
+      .getOrElse("cached")
 
-  final val isFullStackTracing: Boolean = false
+  @inline def isCachedStackTracing: Boolean = stackTracingMode.equalsIgnoreCase("cached")
```
djspiewak: Why `def`?

armanbilge:
> The timeouts in CI are concerning.

Darn, I thought I fixed this in 3a0820f. Ok to increase again?

> Are we now requiring source maps? What happens if someone downstream doesn't want to use source maps?

Nope, nothing requires source maps. If they don't use source maps then they get traces looking like https://github.com/typelevel/cats-effect/pull/2207#issuecomment-897103723.

> Is the consensus that the performance impact is manageable? 

Define manageable? :P The async benchmarks seemed better, is this a more realistic use-case for JS?

> It really sucks that V8 specifically deoptimized this case, though if I'm reading it correctly, it sounds like this is only a one-time penalty.

Yes, one-time penalty ... but since we are caching traces anyway, I think we are only generating the stack trace once anyway.

> Looking through the diff, it seems like the `process.env` approach is working?

Yup, this definitely works. I'm just grumpy that we can't get branch-elimination (at JS-source-generation time) when `process.env.CATS_EFFECT_TRACING_MODE` is hard-coded. Hopefully JS runtimes will do this anyway.

build.sbt
armanbilge: 👍 will fix

core/js/src/main/scala/cats/effect/tracing/TracingConstants.scala
armanbilge: Is inlining a `val` the same as inlining a `def`?

armanbilge:
Also want to ask again: do we want to do any CI with fullOptJS enabled?

core/js/src/main/scala/cats/effect/tracing/TracingConstants.scala
djspiewak: Not sure, but the def (inlined or otherwise) is going to be computed repeatedly at all call sites, magnifying the performance cost.

djspiewak:
Do we benefit from CI with fullOpt? I guess I haven't seen the need for it.

Regarding the timeouts, I think my worry is less that we're hitting them and more that they're indicating that the performance regression is significant enough to push us over that threshold.

core/js/src/main/scala/cats/effect/tracing/TracingConstants.scala
armanbilge: Oops, you are right. I think this was back when I was hoping `process.env.CATS_EFFECT_TRACING_MODE` could be inlined itself.

armanbilge:
> Do we benefit from CI with fullOpt? I guess I haven't seen the need for it.

It's so we could check that tracing works in `fullOptJS` as well, I had to adjust the logic to accomodate.

> Regarding the timeouts, I think my worry is less that we're hitting them and more that they're indicating that the performance regression is significant enough to push us over that threshold.

Gotcha. I'm not sure then, should we try harder to improve performance somehow?

vasilmkd:
If it's only about running the tracing spec, let's just add it to the end of the `node` runs.

tests/shared/src/test/scala/cats/effect/tracing/TracingSpec.scala
```diff
@@ -42,16 +42,6 @@ class TracingSpec extends BaseSpec with TestInstances {
...
-
-    // This test is targeted to Scala.js, where we distinguish thunks by their source code
-    "generate unique traces even if thunk source is identical" in {
-      val a = IO(println("foo"))
-      val b = IO(println("foo"))
-      (a, b) match {
-        case (IO.Delay(_, eventA), IO.Delay(_, eventB)) => eventA ne eventB
-        case _ => false
-      }
-    }
```
armanbilge: I added this as an experiment, to see if we'd get lucky. It is the JS equivalent of the cached tracing gotcha for JVM:
> _Caution_: The stack trace cache is indexed by the lambda class reference

On JS, two (anonymous) lambdas with identical source code are not always distinguishable, depending on optimization that occurs.

core/js/src/main/scala/cats/effect/tracing/TracingPlatform.scala
```diff
@@ -64,23 +64,46 @@ private[tracing] abstract class TracingPlatform { self: Tracing.type =>
...
+    "_Lcats_effect_",
+    "_jl_"
+  )
+
+  private[this] final val stackTraceFileNameFilter: Array[String] = Array(
+    "githubusercontent.com/typelevel/cats-effect/",
+    "githubusercontent.com/typelevel/cats/",
+    "githubusercontent.com/scala-js/",
+    "githubusercontent.com/scala/"
   )
```
armanbilge: I regret having to add yet another set of filters. The full set is necessary to fully support tracing when:
1. `fastOptJS` without source maps
2. `fastOptJS` with source maps
3. `fullOptJS` with source maps

Technically, we still generate traces for `fullOptJS` without source maps, but as all names are mangled beyond recognition none of the filtering works.

armanbilge: Btw, if we knew exactly which case we are in we could possibly optimize this a bit. Scala.js offers a `LinkingInfo` object which indicates whether it is `fastOptJS` or `fullOptJS`, but without knowing whether source maps are enabled it's still not clear what can be skipped. Not sure if there is a good way to detect this at runtime.

djspiewak:
Looping back to this…

I think the current status on this PR is that we're concerned the performance hit is too high. We need more measurement to try to get a better intuition for this. Measuring things like `ParallelBenchmark` might be a decent way to get a more realistic feel for it. Additionally, it would be good to figure out *where* the performance hit is coming from. If it's just the initial exception being really expensive, then I'm less concerned about it since the performance normalizes out over time (though the hit is proportionally worse in serverless environments). If it's the caching though which is particularly expensive, then we probably have a more fundamental issue.

armanbilge:
Ran the benchmarks again, including the parallel benchmark. For `traverse` performance looks much closer.

`series/3.x`
```
Starting suite: deep bind
[0/3] async 10000 : 29.974 ± 2.511 ops/s
[1/3] delay 10000 : 404.868 ± 36.810 ops/s
[2/3] pure  10000 : 674.264 ± 66.975 ops/s
Suite completed: deep bind
Starting suite: parallel
[0/8] parTraverse Params(100,100)   : 95.231 ± 6.499 ops/s
[1/8] traverse    Params(100,100)   : 137.093 ± 11.723 ops/s
[2/8] parTraverse Params(100,1000)  : 12.861 ± 0.907 ops/s
[3/8] traverse    Params(100,1000)  : 14.028 ± 1.939 ops/s
[4/8] parTraverse Params(1000,100)  : 8.546 ± 0.765 ops/s
[5/8] traverse    Params(1000,100)  : 13.182 ± 2.201 ops/s
[6/8] parTraverse Params(1000,1000) : 1.298 ± 0.138 ops/s
[7/8] traverse    Params(1000,1000) : 1.471 ± 0.097 ops/s
Suite completed: parallel
Starting suite: shallow bind
[0/3] async 10000 : 31.670 ± 2.512 ops/s
[1/3] delay 10000 : 459.472 ± 50.069 ops/s
[2/3] pure  10000 : 674.040 ± 77.131 ops/s
Suite completed: shallow bind
```

`js-tracing`:
```
Starting suite: deep bind
[0/3] async 10000 : 17.975 ± 2.272 ops/s
[1/3] delay 10000 : 75.496 ± 16.033 ops/s
[2/3] pure  10000 : 123.443 ± 11.648 ops/s
Suite completed: deep bind
Starting suite: parallel
[0/8] parTraverse Params(100,100)   : 42.541 ± 4.642 ops/s
[1/8] traverse    Params(100,100)   : 113.495 ± 10.739 ops/s
[2/8] parTraverse Params(100,1000)  : 10.911 ± 1.462 ops/s
[3/8] traverse    Params(100,1000)  : 12.750 ± 1.616 ops/s
[4/8] parTraverse Params(1000,100)  : 4.899 ± 0.636 ops/s
[5/8] traverse    Params(1000,100)  : 12.690 ± 2.264 ops/s
[6/8] parTraverse Params(1000,1000) : 1.154 ± 0.073 ops/s
[7/8] traverse    Params(1000,1000) : 1.434 ± 0.135 ops/s
Suite completed: parallel
Starting suite: shallow bind
[0/3] async 10000 : 21.072 ± 2.724 ops/s
[1/3] delay 10000 : 114.915 ± 8.198 ops/s
[2/3] pure  10000 : 193.493 ± 17.791 ops/s
Suite completed: shallow bind
```

djspiewak:
Okay I'm :+1: on this pending setting `baseVersion` to `"3.3"`. Additionally, I'm going to rerun the CI a few times. The timeouts are still making me very nervous, because they feel like a sign that we've made something much worse. Follow ups:

- Do more benchmarking to try to see if we can convince ourselves that we haven't destroyed performance somehow. No immediate ideas on what to measure, but we should think more about it. Maybe look at http4s-server benchmarks on Node? Maybe Fs2's benchmarks? (#2284)
- Detect the combination of fullOptJS and *no* source maps. Anyone in this case is unlikely to care about program introspection (by definition) so we should just disable tracing. (#2285)
- Add to the **tracing** documentation with details on all of this stuff (#2286)

djspiewak:
- [x] First CI run
- [x] Second CI run
- [x] Third CI run
- [x] Fourth CI run
- [x] Fifth CI run

djspiewak (approved):

PR #2208, opened 2021-08-07
Clarify scope of guarantee provided by `Unique`

bplommer:


docs/typeclasses/unique.md
```diff
@@ -11,13 +11,20 @@ trait Unique[F[_]] {
...
 ```scala
 val token: F[Unique.Token] = Unique[F].unique
 (token, token).mapN { (x, y) => x === y } <-> Monad[F].pure(false)
 ```
 
+After a token becomes eligible for garbage collection, a subsequent evaluation 
+of `unique` may produce a new token with the same hash. Similarly, the guarantee 
+of uniqueness only applies within a single JVM or JavaScript runtime. If you 
+need a token that is unique across all space and time, use a 
+[UUID](https://docs.oracle.com/javase/7/docs/api/java/util/UUID.html) instead.
```
vasilmkd: Can you please replace this link with the JDK 8 version? That's the oldest supported version that Cats Effect runs on.

vasilmkd (changes requested):

docs/typeclasses/unique.md
bplommer: Done.

vasilmkd (approved):

djspiewak (approved):

Issue #2066, opened 2021-06-18
Add missing tracing events to `IO` constructors and combinators


vasilmkd:
Which constructors and combinators are already covered?

- [x] `IO.delay/IO.apply`
- [x] `IO.uncancelable`
- [ ] `IO.async`
- [x] `IO#flatMap`
- [x] `IO#map`
- [x] `IO#handleErrorWith`
- [x] `IO.blocking`

Which constructors/combinators would we like to be covered, but are not at the moment?

- [ ] `IO.pure` (personally not sure about this one)
- [ ] `IO.raiseError` (ditto)
- [ ] `IO.realTime/IO.monotonic`
- [ ] `IO.executionContext`
- [ ] `IO#attempt` (probably not worth it)
- [ ] `IO.canceled`
- [ ] `IO.onCancel`
- [ ] `IO.cede` (could be worthwhile)
- [ ] `IO#start` (this one could potentially be weird, because it represents the start of another fiber in the current one, not sure if worth it)
- [ ] `IO.racePair` (this one is implemented in terms of `async` already but would most likely be filtered out because it lives in `cats.effect` package, similar concerns to `start`)
- [ ] `IO.sleep`
- [ ] `IO#evalOn` (might show up for every finalizer registered with `onCancel`, but most likely filtered out)
- [ ] `IO.Local` case

PR #2217, opened 2021-08-10
Add tracing events to some of the `IO` combinators

alexandrustana:
This PR should help with the implementation for #2066.
Adds tracing for the following combinators:
- `monotonic`
- `realTime`
- `executionContext`
- `onCancel`
- `cancelled`
- `cede`
- `sleep`
- `evalOn`

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -488,7 +488,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   def map[B](f: A => B): IO[B] = IO.Map(this, f, Tracing.calculateTracingEvent(f.getClass))
 
   def onCancel(fin: IO[Unit]): IO[A] =
-    IO.OnCancel(this, fin)
+    IO.OnCancel(this, fin, Tracing.calculateTracingEvent(fin.getClass))
 
   def onError(f: Throwable => IO[Unit]): IO[A] =
     handleErrorWith(t => f(t).attempt *> IO.raiseError(t))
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1558,15 +1558,15 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
     def tag = 2
   }
 
-  private[effect] case object RealTime extends IO[FiniteDuration] {
+  private[effect] final case class RealTime(event: TracingEvent) extends IO[FiniteDuration] {
```
alexandrustana, line 1561: This type of change (from `case object` to `case class`) was necessary in order to get an accurate localization of the trace.

armanbilge:
Woops, this is going to conflict with #2207 where I adjusted the tracing API to accommodate Scala.js.

vasilmkd:
Yeah, I'm not convinced that all combinators will need tracing info attached to them. Let's review #2207 first, then this PR can follow it.

vasilmkd:
@alexandrustana Thank you for your contribution, I really appreciate it. Sorry this PR fell off the radar. I've had some time to think about adding tracing to these combinators, and I've convinced myself that what we have already is more than enough. Tracing is not free in terms of performance, and most of these combinators and constructors would not add too much to the developer experience. In some future release, we can revisit this, but for now I will close this PR and the original issue. Thank you again.

alexandrustana:
@vasilmkd no problem, thanks for the heads up :)
I'll pick up another issue 😄

PR #2218, opened 2021-08-11
Use a custom jabba index

vasilmkd:
When in doubt, roll your own JDK index.

As a bonus, you get to use the proper names of the JDK versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,7 @@ on:
 
 env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  JABBA_INDEX: 'https://github.com/vasilmkd/jdk-index/raw/main/index.json'
 
 jobs:
   build:
@@ -24,18 +25,18 @@ jobs:
         os: [ubuntu-latest, windows-latest]
         scala: [3.0.1, 2.12.14, 2.13.6]
         java:
-          - adopt@1.8
-          - adopt@1.11
-          - adopt@1.16
-          - graalvm-ce-java8@21.1
+          - adoptium@8
+          - adoptium@11
+          - adoptium@16
+          - graalvm-ce-java8@21.2
         ci: [ciJVM, ciJS, ciFirefox, ciChrome, ciJSDOMNodeJS]
         exclude:
           - ci: ciJS
-            java: adopt@1.11
+            java: adoptium@11
           - ci: ciJS
-            java: adopt@1.16
+            java: adoptium@16
           - ci: ciJS
-            java: graalvm-ce-java8@21.1
+            java: graalvm-ce-java8@21.2
           - os: windows-latest
             scala: 3.0.1
           - os: windows-latest
@@ -43,11 +44,11 @@ jobs:
           - os: windows-latest
             ci: ciJS
           - ci: ciFirefox
-            java: adopt@1.11
+            java: adoptium@11
           - ci: ciFirefox
-            java: adopt@1.16
+            java: adoptium@16
           - ci: ciFirefox
-            java: graalvm-ce-java8@21.1
+            java: graalvm-ce-java8@21.2
           - os: windows-latest
             scala: 3.0.1
           - os: windows-latest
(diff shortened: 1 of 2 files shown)


vasilmkd:
For anyone interested, here's the index I created.

https://github.com/vasilmkd/jdk-index/blob/9a261f33764d7a9f4dd0634495209ebe63836f56/index.json

vasilmkd:
For better integration, we could move the index file to the Cats Effect repository, however there's a slight chicken and egg problem. We need to merge the file in the repository first and then make the build configuration switch in a subsequent PR.

vasilmkd:
I know this is another thing to manage, but I plan to take that responsibility on myself. I've already been doing this for the past year. Unfortunately, the upstream jabba project seems inactive and they haven't been merging PRs or updating the index for months now.

IMO, Cats Effect needs to test on the latest Java versions as soon as they come out, and this is how I intend to do that.

armanbilge:
> Cats Effect needs to test on the latest Java versions as soon as they come out

Pedagogical question: why? 😅

vasilmkd:
It's a foundational library, and I like to test that the runtime does not make some assumptions that future versions of the JDK won't like. I said IMO at the beginning of the sentence. 😄

armanbilge:
> I said IMO at the beginning of the sentence

I know, I reckoned it's a good one so wanted to learn more :)

> the runtime does not make some assumptions that future versions of the JDK won't like

Are there places where it's pushing the boundary of the spec?

vasilmkd:
We definitely don't violate it, but the async node of the `IOFiber` run loop and `cats.effect.unsafe.LocalQueue` do exercise the Java Memory Model quite a bit.

armanbilge:
Presumably, a regression there would be a major regression in Java itself?

armanbilge:
> Cats Effect needs to test on the latest Java versions as soon as they come out

It sounds like to me, we should be testing _before_ they come out, and alerting them to regressions.

vasilmkd:
I'm sure they have a much more strict process of making sure something like that doesn't happen. But there was also an open call to test against JDK 17 [recently](https://contributors.scala-lang.org/t/scala-community-build-on-jdk-17/5153) (we did and it works [fine](https://github.com/typelevel/cats-effect/pull/2135)). It's not just testing against the JMM, but also making sure the Scala versions work properly and that the build passes reliably. It's a sort of an integration test.

I do agree that we should test against early access builds, but I'm not convinced that that's something we need to do all the time. In any case, the official jabba JDK index is too slow to include current or future versions, which IMO is another argument in favor of a custom, lean index, updateable at our own leisure.

armanbilge:
> making sure the Scala versions work properly and that the build passes reliably

Just my opinion, but this sounds like concerns for upstream Scala itself :)

> the official jabba JDK index is too slow to include current or future versions, which IMO is another argument in favor of a custom, lean index, updateable at our own leisure.

This makes sense.

vasilmkd:
Furthermore, the JDK release process has changed since Java 9, with new patch versions dropping regularly every 3 months and each patch release comes with an advice to upgrade, de-facto making it the only supported version of that major JDK version. (i.e. 16.0.2 dropped two weeks ago making 16.0.1 obsolete).

vasilmkd:
> Just my opinion, but this sounds like concerns for upstream Scala itself :)

This is my personal opinion, I love the Scala language, I love the Scala open-source community, and I want everyone to be their best, do the absolute best work they can spare time for, and this includes being a good community citizen. I personally think Cats Effect should lead by example.

That being said, all of what I mentioned in this comment has nothing to do with this PR. In the end we might decide against this change and that's fine. I just thought of an alternative and I proposed it. (I'm also not being defensive, I just maybe failed to convey my initial intentions).

armanbilge:
In that spirit, maybe the JDK index should be a Typelevel repo that can be used by all typelevel projects? Also reduces the bus factor.

vasilmkd:
Or `djspiewak/sbt-github-actions`. I think that would be the easiest integration point.

rossabaker:
One day, the world will embrace the good news that is Nix.  Until then, if it's just a URI, the Typelevel repo sounds smart to me, unless sbt-github-actions has more than a couple people with merge rights.

djspiewak (approved):

PR #2222, opened 2021-08-12
Remove the tracing lambda from IO.apply

vasilmkd:
Curious to see if the deadlock that arose during the original development can still be seen.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -929,10 +929,8 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
    *
    * Alias for `IO.delay(body)`.
    */
-  def apply[A](thunk: => A): IO[A] = {
-    val fn = () => thunk
-    Delay(fn, Tracing.calculateTracingEvent(fn.getClass))
-  }
+  def apply[A](thunk: => A): IO[A] =
+    Delay(() => thunk, Tracing.calculateTracingEvent(thunk.getClass))
 
   /**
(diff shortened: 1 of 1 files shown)


vasilmkd:
https://discord.com/channels/632277896739946517/839263556754472990/875499045085384755

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -929,10 +929,8 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
    *
    * Alias for `IO.delay(body)`.
    */
-  def apply[A](thunk: => A): IO[A] = {
-    val fn = () => thunk
-    Delay(fn, Tracing.calculateTracingEvent(fn.getClass))
-  }
+  def apply[A](thunk: => A): IO[A] =
+    Delay(() => thunk, Tracing.calculateTracingEvent(thunk.getClass))
```
armanbilge, line 933: I've been surprised before, but this isn't really evaluating the `thunk` first, then calling `getClass` ?

armanbilge: And if not, then _how_ do you call `getClass` on the outcome of the `thunk`?

vasilmkd:
The CI is hopelessly stuck.

core/shared/src/main/scala/cats/effect/IO.scala
armanbilge: e.g. see https://scastie.scala-lang.org/Gq9MqhQoS9KPSGZQjV2kNA

vasilmkd:
Fixed in #2226.

Issue #2225, opened 2021-08-13
Cached tracing for `IO.delay` is broken in Scala 3


armanbilge:
As uncovered by a [discord discussion](https://discord.com/channels/632277896739946517/839263556754472990/875532566508212294) with @vasilmkd.

[Code in question](https://github.com/typelevel/cats-effect/blob/series/3.x/core/shared/src/main/scala/cats/effect/IO.scala#L932-L935): 
```scala
def apply[A](thunk: => A): IO[A] = {
  val fn = () => thunk
  Delay(fn, Tracing.calculateTracingEvent(fn.getClass))
}
```

On Scala 3, `fn.getClass` is a constant regardless of the value of `thunk`. Thus, all uses of `delay` end up with the same cached trace.

On Scala 2, `fn.getClass` returns a different value for different `thunk`s (AFAICT, it returns the class of the `Function0` that the `thunk` is syntax sugar for) and thus cached tracing works as expected.

This issue can be observed in [this scastie](https://scastie.scala-lang.org/CyYasg6cR9eSfFCLIrsI9Q) particularly by toggling between Scala 2 and Scala 3. Update: a more precise test demonstrating this problem is now in #2226.

Not sure how to fix this, other than perhaps disabling caching/doing full tracing for every `delay`?

vasilmkd:
Fixed in #2226.

PR #2226, opened 2021-08-13
Fix cached tracing for `IO.delay`

armanbilge:
The first commit demonstrates the issue described in #2225, note the test fails only for Scala 3.


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -931,7 +931,7 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
    */
   def apply[A](thunk: => A): IO[A] = {
     val fn = () => thunk
-    Delay(fn, Tracing.calculateTracingEvent(fn.getClass))
+    Delay(fn, Tracing.calculateTracingEvent(tracing.ThunkClassRetriever(thunk)))
```
vasilmkd: Inline the lambda in `Delay` and remove the braces around the method.

core/jvm/src/main/scala/cats/effect/tracing/ThunkClassRetriever.scala
```diff
@@ -0,0 +1,27 @@
...
+
+private abstract class ThunkClassRetriever {
+  def getClassOf[A](thunk: => A): Class[_]
+}
+
+private[effect] object ThunkClassRetriever {
+  private val impl: ThunkClassRetriever = new ThunkClassRetrieverImpl
+
+  def apply[A](thunk: => A): Class[_] = impl.getClassOf(thunk)
+}
```
vasilmkd: Can you move this into `Tracing.scala`?

vasilmkd: And rename the `apply` method to getThunkClass or something.

armanbilge: Err, wait, not sure. `Tracing.scala` is cross-platform but we can't implement this on JS.

vasilmkd: `Platform`.

core/js/src/main/scala/cats/effect/tracing/ThunkClassRetriever.scala
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.tracing
+
+private[tracing] class ThunkClassRetrieverImpl extends ThunkClassRetriever {
```
vasilmkd: No need for `private[tracing]`, just `private`.

armanbilge: If it's just `private`, then we can't do `new` in `Tracing` I think.

vasilmkd: Also please make the filename `ThunkClassRetrieverImpl`.

armanbilge: Oh I guess we can :) cool

djspiewak (commented):
So this makes a lot of sense, and is also disheartening because it means that Scala 3 isn't optimizing thunk-to-function conversion, and thus Delay nodes are more expensive on that platform. I actually wonder if it would be even better to generalize this machinery slightly, and rather than returning the Class, return a Function0. We can still get the class from it, but effectively we would be cheating the compiler into implementing the optimization it previously did naturally.

armanbilge:
> I actually wonder if it would be even better to generalize this machinery slightly, and rather than returning the Class, return a Function0

Clever! Why not?

vasilmkd:
Please make it a part of tracing...

armanbilge:
Lol!

vasilmkd:
Or keep it in `cats.effect`. Just not in a subpackage.

core/jvm/src/main/java/cats/effect/util/ThunkImpl.java
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+package cats.effect.util;
+
+class ThunkImpl extends Thunk {
+  public <A> scala.Function0<A> asFunction0(scala.Function0<A> thunk) {
+    return thunk;
+  }
+}
```
vasilmkd: Let's add a comment/javadoc on this class explaining the need for its existence. Mention the differences between Scala 2 and Scala 3.

core/jvm/src/main/java/cats/effect/ThunkImpl.java
```diff
@@ -0,0 +1,28 @@
...
+
+package cats.effect;
+
+/**
+ * A utility to convert a by-name `thunk: => A` to a `Function0[A]` (its binary representation).
+ * Scala 2 performs this optimization automatically but on Scala 3 the thunk is wrapped inside of a new `Function0`.
+ * See https://github.com/typelevel/cats-effect/pull/2226
+ */
+class ThunkImpl extends Thunk {
+  public <A> scala.Function0<A> asFunction0(scala.Function0<A> thunk) {
```
vasilmkd: Add an `@Override` here.

core/shared/src/main/scala/cats/effect/tracing/ThunkClassRetriever.scala
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.tracing
+
+private abstract class ThunkClassRetriever {
+  def getClassOf[A](thunk: => A): Class[_]
+}
```
vasilmkd: This is a leftover from the previous encoding. I think this is the final change left to do.

vasilmkd (changes requested):

vasilmkd (commented):
Can we run a quick `DeepBindBenchmark.delay` _without_ tracing on Scala 2.13 and Scala 3? Before and after.

armanbilge:
Command: `benchmarks/Jmh/run -wi 10 -i 10 -f 2 -t 1 DeepBindBenchmark.delay --jvmArgs -Dcats.effect.tracing.mode=none`

2.13 before:
```
[info] Benchmark                (size)   Mode  Cnt     Score    Error  Units
[info] DeepBindBenchmark.delay   10000  thrpt   20  3588.051 ± 75.498  ops/s
```

3 before:
```
[info] Benchmark                (size)   Mode  Cnt     Score    Error  Units
[info] DeepBindBenchmark.delay   10000  thrpt   20  2841.203 ± 45.684  ops/s
```

2.13 after:
```
[info] Benchmark                (size)   Mode  Cnt     Score    Error  Units
[info] DeepBindBenchmark.delay   10000  thrpt   20  3487.627 ± 87.979  ops/s
```

3 after:
```
[info] Benchmark                (size)   Mode  Cnt     Score    Error  Units
[info] DeepBindBenchmark.delay   10000  thrpt   20  3897.315 ± 39.551  ops/s
```

armanbilge:
2.13 @ 7732752:
```
[info] DeepBindBenchmark.delay   10000  thrpt   20  3453.852 ± 110.719  ops/s
```

3 @ 7732752:
```
[info] DeepBindBenchmark.delay   10000  thrpt   20  3431.805 ± 216.061  ops/s
```

armanbilge:
3 @ 7732752 (take 2):
```
[info] DeepBindBenchmark.delay   10000  thrpt   20  3786.509 ± 129.755  ops/s
```

vasilmkd:
Here are my results:

```
Baseline 2.13
DeepBindBenchmark.delay   10000  thrpt   20  3028.428 ± 84.643  ops/s

Baseline 3.0
Benchmark                (size)   Mode  Cnt     Score     Error  Units
DeepBindBenchmark.delay   10000  thrpt   20  2848.246 ± 204.663  ops/s

Java code 2.13
DeepBindBenchmark.delay   10000  thrpt   20  3116.695 ± 84.533  ops/s

Java code 3.0
Benchmark                (size)   Mode  Cnt     Score     Error  Units
DeepBindBenchmark.delay   10000  thrpt   20  3706.874 ± 365.418  ops/s

Cross-platform 2.13
Benchmark                (size)   Mode  Cnt     Score     Error  Units
DeepBindBenchmark.delay   10000  thrpt   20  3328.456 ± 338.778  ops/s

Cross-platform 3.0
Benchmark                (size)   Mode  Cnt     Score     Error  Units
DeepBindBenchmark.delay   10000  thrpt   20  3414.873 ± 111.307  ops/s
```

core/shared/src/main/scala/cats/effect/Thunk.scala
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+
+private object Thunk {
+  private[this] val impl =
```
vasilmkd, line 25: Can you please restore the comment that you put on the Java code.

vasilmkd (changes requested):

vasilmkd (approved):
Great work. I can't tell you how happy I am that we get to keep cached tracing for `apply/delay`.

PR #2228, opened 2021-08-13
Backport thunk

armanbilge:
Via #2226.


vasilmkd (changes requested):
Can you please rebase and drop the first commit. We will have the history in CE3, there's no need to repeat it here.

vasilmkd (approved):

PR #2229, opened 2021-08-14
POC: Scala.js benchmarks

armanbilge:
**tl;dr**: `sbt benchmarksJS/run`
```
[info] Running cats.effect.benchmarks.Main.
Starting suite: deep bind
[0/3] async 10000 : 31.576 ± 4.789 ops/s
[1/3] delay 10000 : 468.774 ± 43.810 ops/s
[2/3] pure  10000 : 698.908 ± 50.539 ops/s
Suite completed: deep bind
Starting suite: shallow bind
[0/3] async 10000 : 33.427 ± 1.590 ops/s
[1/3] delay 10000 : 483.560 ± 32.158 ops/s
[2/3] pure  10000 : 750.375 ± 30.180 ops/s
Suite completed: shallow bind
```

Notes/thoughts:
1. This is using [my fork of scalajs-benchmark](https://github.com/armanbilge/scalajs-benchmark/tree/ce3). Yes, I know :( but, the original code had its own ["Resource Kleisli"](https://github.com/japgolly/scalajs-benchmark/blob/master/benchmark/src/main/scala/japgolly/scalajs/benchmark/Setup.scala#L14-L17) and ["Fiber"](https://github.com/japgolly/scalajs-benchmark/blob/master/benchmark/src/main/scala/japgolly/scalajs/benchmark/engine/Engine.scala#L52-L66)-like constructs and I'm just more comfortable working with the CE3 implementations of these (technically I created a circular dependency between it and CE3). Also it needed a cross to 2.12 which involved dropping the browser interface (cool but unnecessary for us). And it only ran in the JSDOM environment.
2. Scala.js has got some tricky limitations, in that there can only be one main method per project, and you cannot pass command line arguments to the runner. So not sure yet how to enable selecting the suite you want run via sbt.
3. We can probably share the benchmark code between the JVM and JS (i.e., the `IO` to be run).


vasilmkd:
I just remembered, we got a pretty decent benchmark recently, `ParallelBenchmark`. I think it would be nice to see it in action, it should give us another good data point when it comes to `IO` interactions which are somewhat representative, namely `traverse` and `parTraverse`.

A bit of a warning, that benchmark does make use of `Blackhole`, I'm not sure if that's going to be a problem.

Another thing, it has parameters for both the weight of the CPU tasks as well as the size of the lists. Running everything would probably be too long, but it would be nice to see the middle values for both, for example. Again, I'm not sure how long each one takes.

armanbilge:
Cool, I'll try porting that one over too, thanks for the suggestion! The scalajs-benchmark library does have blackholes in it so hopefully those should work fine.

Right now the JS benchmarks run much quicker than their JMH counterparts, we might want to change the [default benchmark settings](https://github.com/armanbilge/scalajs-benchmark/blob/ce3/benchmark/src/main/scala/japgolly/scalajs/benchmark/engine/EngineOptions.scala#L32-L40) although JS runtime is pretty different from JVM I guess?

armanbilge:
@vasilmkd `Params(size, cpuTokens)`
```
Starting suite: parallel
[ 0/10] parTraverse Params(100,100)   : 120.885 ± 10.280 ops/s
[ 1/10] traverse    Params(100,100)   : 144.361 ± 19.643 ops/s
[ 2/10] parTraverse Params(100,1000)  : 15.414 ± 1.052 ops/s
[ 3/10] traverse    Params(100,1000)  : 15.670 ± 2.085 ops/s
[ 4/10] parTraverse Params(100,10000) : 1.670 ± 0.105 ops/s
[ 5/10] traverse    Params(100,10000) : 1.525 ± 0.140 ops/s
[ 6/10] parTraverse Params(1000,100)  : 10.917 ± 1.170 ops/s
[ 7/10] traverse    Params(1000,100)  : 13.364 ± 1.558 ops/s
[ 8/10] parTraverse Params(1000,1000) : 1.398 ± 0.212 ops/s
[ 9/10] traverse    Params(1000,1000) : 1.334 ± 0.288 ops/s
Suite completed: parallel
```

PR #2132, opened 2021-07-16
More scaladoc

djspiewak:
Still lots of stuff to do.

PR #2232, opened 2021-08-15
Fix unlinkable documentation reference in JS platform code

vasilmkd:


core/js/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -23,7 +23,7 @@ import scala.scalajs.js
 /**
  * The primary entry point to a Cats Effect application. Extend this
  * trait rather than defining your own `main` method. This avoids the
- * need to run [[IO.unsafeRunSync]] (or similar) on your own.
+ * need to run `IO.unsafeRunSync` (or similar) on your own.
```
armanbilge: Just a nit, but there is no `IO.unsafeRunSync` in JS land anyway so there's nowhere for this to link :) Perhaps replace with `IO.unsafeRunAsync` or something?

armanbilge:
Just a random bikeshed idea, it seems like there's a lot of duplicated docs between the JVM and JS `IOApp`s. I'm not an expert on scaladoc, but would it help if we made a shared private `trait` for the two to put the documentation so that it could be inherited?

core/js/src/main/scala/cats/effect/IOApp.scala
vasilmkd: I changed it. Thanks.

vasilmkd:
There was a common ancestor before, but Daniel duplicated the code in #2132 at the expense of some documentation duplication. I would leave it as is.

cc @djspiewak

djspiewak (approved):

djspiewak:
I tried deduplicating but surprisingly it isn't really possible while maintaining binary compatibility.

armanbilge:
Binary compatibility confounds me the second time in 5 minutes 🤔

djspiewak:
It surprised me too :-) The problem comes down to the forwarders required for trait constructor encoding.

armanbilge:
IIUC sometimes those are fixable by just adding an override that calls super in the child trait?

djspiewak:
On this one it's the *constructor* which is the problem, not any particular method, so that's a tricky one.

PR #2233, opened 2021-08-15
`LocalQueue` metrics

vasilmkd:
A part of #1568.

Split out from #1789. That change is a bit too big to tackle at once. I've split out the `LocalQueue` metrics. This way I feel like there's some progress being made.

This is just an implementation of the metrics interface, but the code that hooks it up to the `MBeans` server is not in yet. I will do that after I finalize the metrics interfaces.


vasilmkd:
I need to run some quick benchmarks to make sure everything is fine.

vasilmkd:
The scheduling benchmark shows no changes.

djspiewak (approved):

PR #2237, opened 2021-08-16
Swap `Thenable` for `Promise` in `Async.fromPromise`

### LINKED ISSUE #2236: `IO.fromPromise` should accept `js.Thenable`, instead of `js.Promise`
`js.Thenable` is a super-type of `js.Promise`

armanbilge:
Fixes #2236.

### DIFF
--- a/core/js/src/main/scala/cats/effect/IOCompanionPlatform.scala
+++ b/core/js/src/main/scala/cats/effect/IOCompanionPlatform.scala
@@ -18,7 +18,7 @@ package cats.effect
 
 import scalajs.js
 
-import scala.scalajs.js.Promise
+import scala.scalajs.js.{Promise, Thenable}
 
 private[effect] abstract class IOCompanionPlatform { this: IO.type =>
 
@@ -34,6 +34,9 @@ private[effect] abstract class IOCompanionPlatform { this: IO.type =>
     val _ = hint
     apply(thunk)
   }
+
+  def fromThenable[A](iot: IO[Thenable[A]]): IO[A] =
+    asyncForIO.fromThenable(iot)
 
   def fromPromise[A](iop: IO[Promise[A]]): IO[A] =
     asyncForIO.fromPromise(iop)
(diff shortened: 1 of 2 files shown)


vasilmkd (dismissed):

core/js/src/main/scala/cats/effect/IOCompanionPlatform.scala
```diff
@@ -35,6 +35,9 @@ private[effect] abstract class IOCompanionPlatform { this: IO.type =>
     apply(thunk)
   }
 
+  def fromThenable[A](iot: IO[Thenable[A]]): IO[A] =
+    asyncForIO.fromThenable(iot)
```
djspiewak: Can we just make `fromPromise` take a `Thenable` instead? That would mirror the JavaScript convention of treating "everything that has a `then`" as being secretly a `Promise`. It's even fully source and binary compatible in `IO`, and fully binary compatible in `Async`.

djspiewak (changes requested):

core/js/src/main/scala/cats/effect/IOCompanionPlatform.scala
armanbilge: Sure, makes sense!

> It's even fully source and binary compatible in IO, and fully binary compatible in Async.

I feel I am never going to understand binary compatibility 🤔 I didn't think this would be binary compatible because when relaxing the queue signatures to the source/sink supertypes in fs2 in https://github.com/typelevel/fs2/pull/2466, it _wasn't_ binary compatible. Why is it different here?

armanbilge: Ahh, because type erasure.

djspiewak: The binary type is erased to `IO` in the former case, and erased to `Object` in the latter case, so anything goes. :-P

djspiewak (approved):

PR #2239, opened 2021-08-16
Fixed nontermination issue in `par` operators

### LINKED ISSUE #2238: Parallel map2/map2Eval fails to short-circuit on error
```scala
(F.never, F.raiseError(e)).parTupled
```

I believe the above hangs forever. It (correctly) aborts early if we invert the order of the operands.

The problem here is the implementation is done in terms of `Spawn[F]` rather than `Concurrent[F]`, meaning that it boils down to:

1. Start first fiber
2. Start second fiber
3. Join first fiber
4. Join second fiber

If the first fiber never returns, then we can't observe the completion of the second fiber. The only way to resolve this is to use a `Deferred` for each of the two fibers so that their completion handlers can be *within* the body of the fibers themselves, as opposed to happening post-`join`. There might be a fancy way of doing this with extra fibers but I'm not totally sure, and it would probably be less efficient than doing a pair of `Deferred`s.

djspiewak:
Fixes #2238 

This is a relatively high-priority fix, tbh, since `par` operations could be non-terminating in 3.2.2 with the current implementation. The performance impact is pretty real though:

Before:

```
[info] Benchmark                      (cpuTokens)  (size)   Mode  Cnt    Score    Error  Units
[info] ParallelBenchmark.parTraverse        10000    1000  thrpt    5  644.762 ± 49.298  ops/s
[info] ParallelBenchmark.traverse           10000    1000  thrpt    5  236.523 ±  9.855  ops/s
```

After

```
[info] Benchmark                      (cpuTokens)  (size)   Mode  Cnt    Score    Error  Units
[info] ParallelBenchmark.parTraverse        10000    1000  thrpt    5  596.967 ± 20.901  ops/s
[info] ParallelBenchmark.traverse           10000    1000  thrpt    5  237.569 ± 10.316  ops/s
```

So about 7.5% slower in a naively relative comparison. A better comparison is to look at the overhead costs by measuring relative to an idealized `traverse`. I have 16 physical threads, meaning that an (impossible) idealized overhead-free `parTraverse` implementation would get about 3784.368 ops/sec on the first run and 3801.104 ops/sec on the second. The actual throughput was 644.762 and 596.967, respectively, meaning the overhead accounts for 83.0% of the runtime before this change and 84.3% after, which is an increase of about 1.6%. That's unfortunate (also it sucks that the overhead is so high), but it's generally within the bounds of sanity.

In theory, an alternative implementation using two `Deferred`s (one for each fiber) and `guaranteeCase` *within* the body of each fiber would probably be slightly faster, but it would also require some more complex dynamic dispatch machinery similar to the original proposal in #2155. It's probably worth measuring that alternative implementation to see if it's a meaningful improvement.

### DIFF
--- a/tests/shared/src/test/scala/cats/effect/IOSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/IOSpec.scala
@@ -1065,6 +1065,26 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
 
     }
 
+    "parallel" should {
+      "run parallel actually in parallel" in real {
+        val x = IO.sleep(2.seconds) >> IO.pure(1)
+        val y = IO.sleep(2.seconds) >> IO.pure(2)
+
+        List(x, y).parSequence.timeout(3.seconds).flatMap { res =>
+          IO {
+            res mustEqual List(1, 2)
+          }
+        }
+      }
+
+      "short-circuit on error" in ticked { implicit ticker =>
+        case object TestException extends RuntimeException
+
+        (IO.never[Unit], IO.raiseError[Unit](TestException)).parTupled.void must failAs(TestException)
+        (IO.raiseError[Unit](TestException), IO.never[Unit]).parTupled.void must failAs(TestException)
+      }
+    }
+
     "miscellaneous" should {
 
       "round trip non-canceled through s.c.Future" in ticked { implicit ticker =>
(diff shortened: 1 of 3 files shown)


tests/shared/src/test/scala/cats/effect/IOSpec.scala
```diff
@@ -1065,6 +1065,26 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
...
+      "run parallel actually in parallel" in real {
+        val x = IO.sleep(2.seconds) >> IO.pure(1)
+        val y = IO.sleep(2.seconds) >> IO.pure(2)
+
+        List(x, y).parSequence.timeout(3.seconds).flatMap { res =>
+          IO {
+            res mustEqual List(1, 2)
+          }
+        }
+      }
```
vasilmkd, line 1078: Hopefully this won't be flaky in the CI.

vasilmkd:
Another annoyance I just thought of is the fact that `IO` won't side-channel the second exception in the case that both spawned fibers fail.

tests/shared/src/test/scala/cats/effect/IOSpec.scala
djspiewak: It's an old test so it should be fine

djspiewak:
Oh that's a good point. I'll think about how to resolve that.

djspiewak:
Btw given the priority of this fix, I think we shouldn't hold it for the side channel fix. Let's file that and fix in a follow up.

vasilmkd:
Yup, I'm just waiting for the CI.

vasilmkd (approved):

PR #2246, opened 2021-08-19
Implemented `timeoutAndAbandon` for leaky timeout scenarios

djspiewak:
Let's discuss this before merge. I'm still not convinced it's a good idea.

This is very related to @alexandru's discussion here: https://github.com/typelevel/cats-effect/discussions/1979


alexandru:
Just saw the issue, this is good. I would suggest:

1. rename `timeoutAndAbandon` to `timeout`, and rename `timeout` to something else (e.g. `tryTimeoutButWaitOnCancellation`);
2. also provide a sane `race` and `tryRaceButWaitCancellation` pair;

> * This may more closely match intuitions
> * about timeouts, but it also violates backpressure guarantees
> * and intentionally leaks fibers.
> *
> * This combinator should be applied very carefully.

I disagree with that warning BTW 🙂 I think the current `timeout` is the one that's actually unsafe and that should be carefully applied. It's an implementation that I would probably never use, I believe it is that bad, because you have to trust the source, and rely on developer discipline for properly defining tasks, which defeats the purpose of having `timeout` or `race`.  A leaked fiber isn't so awful as a zombie process, as we have tools to detect and act on leaks, whereas a back-pressured zombie is more subtle.

I would also say that the behaviour of `bracket` and related (e.g. `Resource`), with no way to timeout finalisers, is also problematic and a problem in need of a fix. Basically wherever Cats-Effect back-pressures on finalisers there needs to be an obvious alternative that doesn't do that. But that's for another day 🙂

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -598,6 +602,27 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
...
+   * This combinator should be applied very carefully.
+   *
+   * @param duration
+   *   The time span for which we wait for the source to complete; in the event that the
+   *   specified time has passed without the source completing, a `TimeoutException` is raised
+   * @see
+   *   [[timeout]] for a variant which respects backpressure and does not leak fibers
+   */
+  def timeoutAndForget(duration: FiniteDuration): IO[A] =
+    Temporal[IO].timeoutAndForget(this, duration)
```
vasilmkd, line 624: We should maybe rewrite these to `IO.asyncForIO.someMethod`, here and in other places.

vasilmkd (approved):

core/shared/src/main/scala/cats/effect/IO.scala
armanbilge: Out of curiosity, why doesn't the `IO` companion object itself just implement `Async[IO]`?

djspiewak: That's actually a really interesting concept…

djspiewak:
I'll follow up with more cleanup on things, including potentially `IO.type <: Async[IO]`

PR #2249, opened 2021-08-20
Binpack continuations stack to improve memory density

djspiewak:
This removes the old implementation of `ByteStack` and replaces it with a branchless version which binpacks the four bits required for our continuation flags into an `Array[Int]`, as well as pulling the max offset into that same array. This allows us to allocate much smaller continuation stacks which also grow more slowly, as well as slightly lightening up the allocation costs for `IOFiber` itself. The downside is we had to throw away the `failure` fast-path scan (more notes on that later). The main win here is that the memory pressure is dramatically lower. Lots of numbers below…

---

This is a bit of a complex one to analyze. First off, the continuations stack is not at all the bottleneck on `IO`, so these kinds of optimizations are mostly lost in the noise. Despite that, there is a *slight* difference before and after in terms of total run time:

Before:

```
[info] Benchmark                         (cpuTokens)  (size)   Mode  Cnt     Score     Error  Units
[info] DeepBindBenchmark.async                   N/A   10000  thrpt    5  3230.317 ±  50.599  ops/s
[info] DeepBindBenchmark.delay                   N/A   10000  thrpt    5  9334.381 ± 255.966  ops/s
[info] DeepBindBenchmark.pure                    N/A   10000  thrpt    5  9963.281 ± 381.715  ops/s
[info] HandleErrorBenchmark.errorRaised          N/A   10000  thrpt    5  3135.114 ±  35.408  ops/s
[info] HandleErrorBenchmark.happyPath            N/A   10000  thrpt    5  3754.813 ±  50.927  ops/s
[info] ParallelBenchmark.parTraverse           10000    1000  thrpt    5   509.352 ±  55.933  ops/s
[info] ParallelBenchmark.traverse              10000    1000  thrpt    5    63.182 ±   2.599  ops/s
[info] ShallowBindBenchmark.async                N/A   10000  thrpt    5  2490.175 ±  50.183  ops/s
[info] ShallowBindBenchmark.delay                N/A   10000  thrpt    5  9316.155 ± 198.293  ops/s
[info] ShallowBindBenchmark.pure                 N/A   10000  thrpt    5  9532.402 ± 419.648  ops/s

[info] Benchmark                          (size)   Mode  Cnt   Score   Error    Units
[info] WorkStealingBenchmark.scheduling  1000000  thrpt    5  23.424 ± 4.094  ops/min
```

After:

```
[info] Benchmark                         (cpuTokens)  (size)   Mode  Cnt     Score     Error  Units
[info] DeepBindBenchmark.async                   N/A   10000  thrpt    5  2886.952 ± 209.338  ops/s
[info] DeepBindBenchmark.delay                   N/A   10000  thrpt    5  9159.002 ± 671.254  ops/s
[info] DeepBindBenchmark.pure                    N/A   10000  thrpt    5  9934.367 ± 344.390  ops/s
[info] HandleErrorBenchmark.errorRaised          N/A   10000  thrpt    5  3007.812 ±  51.022  ops/s
[info] HandleErrorBenchmark.happyPath            N/A   10000  thrpt    5  3302.089 ±  98.346  ops/s
[info] ParallelBenchmark.parTraverse           10000    1000  thrpt    5   498.043 ±  51.942  ops/s
[info] ParallelBenchmark.traverse              10000    1000  thrpt    5    62.830 ±   3.086  ops/s
[info] ShallowBindBenchmark.async                N/A   10000  thrpt    5  2262.803 ±  30.562  ops/s
[info] ShallowBindBenchmark.delay                N/A   10000  thrpt    5  9343.042 ± 197.581  ops/s
[info] ShallowBindBenchmark.pure                 N/A   10000  thrpt    5  9338.564 ± 490.507  ops/s

[info] Benchmark                          (size)   Mode  Cnt   Score   Error    Units
[info] WorkStealingBenchmark.scheduling  1000000  thrpt    5  25.186 ± 1.621  ops/min
```

This also appears to suggest that the external scanning optimization in `failed` isn't really helping anything.

Notice that almost everything is within the margin for error. The most statistically significant changes are in `errorRaised` and `scheduling`, with the former degrading slightly while the latter improves significantly. All in all, this change is *nearly* a wash outside of memory pressure:

Before:

```
[info] Benchmark                                                           (size)   Mode  Cnt            Score            Error    Units
[info] WorkStealingBenchmark.scheduling                                   1000000  thrpt    5           18.834 ±          6.682  ops/min
[info] WorkStealingBenchmark.scheduling:·gc.alloc.rate                    1000000  thrpt    5         3090.101 ±       1117.344   MB/sec
[info] WorkStealingBenchmark.scheduling:·gc.alloc.rate.norm               1000000  thrpt    5  10775591876.933 ±     357261.592     B/op
[info] WorkStealingBenchmark.scheduling:·gc.churn.PS_Eden_Space           1000000  thrpt    5         3121.574 ±       1264.144   MB/sec
[info] WorkStealingBenchmark.scheduling:·gc.churn.PS_Eden_Space.norm      1000000  thrpt    5  10878006067.200 ± 1003154019.356     B/op
[info] WorkStealingBenchmark.scheduling:·gc.churn.PS_Old_Gen              1000000  thrpt    5           52.054 ±        156.868   MB/sec
[info] WorkStealingBenchmark.scheduling:·gc.churn.PS_Old_Gen.norm         1000000  thrpt    5    180311766.400 ±  551143294.535     B/op
[info] WorkStealingBenchmark.scheduling:·gc.churn.PS_Survivor_Space       1000000  thrpt    5           50.178 ±         28.841   MB/sec
[info] WorkStealingBenchmark.scheduling:·gc.churn.PS_Survivor_Space.norm  1000000  thrpt    5    177866898.933 ±  165077416.099     B/op
[info] WorkStealingBenchmark.scheduling:·gc.count                         1000000  thrpt    5           93.000                    counts
[info] WorkStealingBenchmark.scheduling:·gc.time                          1000000  thrpt    5         6307.000                        ms
```

After:

```
[info] Benchmark                                                           (size)   Mode  Cnt            Score            Error    Units
[info] WorkStealingBenchmark.scheduling                                   1000000  thrpt    5           14.375 ±         11.017  ops/min
[info] WorkStealingBenchmark.scheduling:·gc.alloc.rate                    1000000  thrpt    5         2352.774 ±       1799.100   MB/sec
[info] WorkStealingBenchmark.scheduling:·gc.alloc.rate.norm               1000000  thrpt    5  10775798612.800 ±     931296.668     B/op
[info] WorkStealingBenchmark.scheduling:·gc.churn.PS_Eden_Space           1000000  thrpt    5         2378.270 ±       1751.656   MB/sec
[info] WorkStealingBenchmark.scheduling:·gc.churn.PS_Eden_Space.norm      1000000  thrpt    5  10906725502.667 ± 1287186881.000     B/op
[info] WorkStealingBenchmark.scheduling:·gc.churn.PS_Old_Gen              1000000  thrpt    5           24.499 ±        100.341   MB/sec
[info] WorkStealingBenchmark.scheduling:·gc.churn.PS_Old_Gen.norm         1000000  thrpt    5    113354916.000 ±  458602258.928     B/op
[info] WorkStealingBenchmark.scheduling:·gc.churn.PS_Survivor_Space       1000000  thrpt    5           42.071 ±         50.457   MB/sec
[info] WorkStealingBenchmark.scheduling:·gc.churn.PS_Survivor_Space.norm  1000000  thrpt    5    189144549.333 ±   99755404.342     B/op
[info] WorkStealingBenchmark.scheduling:·gc.count                         1000000  thrpt    5           71.000                    counts
[info] WorkStealingBenchmark.scheduling:·gc.time                          1000000  thrpt    5         6611.000                        ms
```

Notice how the allocation rate is dramatically better. This change lowers the footprint on fibers quite noticeably and improves the probability that other (more meaningful) data is included within the cache line. Thus, in practice, this change should result in gains that are hard to measure in microbenchmarks.

All credit for everything clever here goes to @viktorklang. :-)

### DIFF
--- a/benchmarks/src/main/scala/cats/effect/benchmarks/ParallelBenchmark.scala
+++ b/benchmarks/src/main/scala/cats/effect/benchmarks/ParallelBenchmark.scala
@@ -45,10 +45,10 @@ import java.util.concurrent.TimeUnit
 @OutputTimeUnit(TimeUnit.SECONDS)
 class ParallelBenchmark {
 
-  @Param(Array("100", "1000", "10000"))
+  @Param(Array(/*"100", */"1000"/*, "10000"*/))
   var size: Int = _
 
-  @Param(Array("100", "1000", "10000", "100000", "1000000"))
+  @Param(Array(/*"100", "1000", */"10000"/*, "100000", "1000000"*/))
   var cpuTokens: Long = _
 
   @Benchmark
--- a/benchmarks/src/main/scala/cats/effect/benchmarks/ParallelBenchmark.scala
+++ b/benchmarks/src/main/scala/cats/effect/benchmarks/ParallelBenchmark.scala
@@ -45,10 +45,10 @@ import java.util.concurrent.TimeUnit
 @OutputTimeUnit(TimeUnit.SECONDS)
 class ParallelBenchmark {
 
-  @Param(Array(/*"100", */"1000"/*, "10000"*/))
+  @Param(Array( /*"100", */ "1000" /*, "10000"*/ ))
   var size: Int = _
 
-  @Param(Array(/*"100", "1000", */"10000"/*, "100000", "1000000"*/))
+  @Param(Array( /*"100", "1000", */ "10000" /*, "100000", "1000000"*/ ))
   var cpuTokens: Long = _
 
   @Benchmark
--- a/build.sbt
+++ b/build.sbt
@@ -352,7 +352,8 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
       ProblemFilters.exclude[DirectMissingMethodProblem]("cats.effect.IO#IOCont.copy"),
       ProblemFilters.exclude[DirectMissingMethodProblem]("cats.effect.IO#IOCont.this"),
       ProblemFilters.exclude[IncompatibleMethTypeProblem](
-        "cats.effect.unsafe.IORuntimeCompanionPlatform.installGlobal")
+        "cats.effect.unsafe.IORuntimeCompanionPlatform.installGlobal"),
+      ProblemFilters.exclude[Problem]("cats.effect.ByteStack.*")
     )
   )
   .jvmSettings(
--- a/build.sbt
+++ b/build.sbt
@@ -441,7 +441,9 @@ lazy val example = crossProject(JSPlatform, JVMPlatform)
 lazy val benchmarks = project
   .in(file("benchmarks"))
   .dependsOn(core.jvm)
-  .settings(name := "cats-effect-benchmarks")
+  .settings(
+    name := "cats-effect-benchmarks",
+    javaOptions ++= Seq("-Dcats.effect.tracing.mode=none", "-Dcats.effect.tracing.exceptions.enhanced=false"))
   .enablePlugins(NoPublishPlugin, JmhPlugin)
 
 lazy val docs = project.in(file("site-docs")).dependsOn(core.jvm).enablePlugins(MdocPlugin)
--- a/core/shared/src/main/scala/cats/effect/ByteStack.scala
+++ b/core/shared/src/main/scala/cats/effect/ByteStack.scala
@@ -16,65 +16,64 @@
 
 package cats.effect
 
-private[effect] final class ByteStack(
-    private[this] var buffer: Array[Byte],
-    private[this] var index: Int) {
+private[effect] object ByteStack {
 
-  def this(initBound: Int) =
-    this(new Array[Byte](initBound), 0)
-
-  def this() = this(null, 0)
-
-  def init(bound: Int): Unit = {
-    buffer = new Array(bound)
-    index = 0
+  final def toDebugString(stack: Array[Int], translate: Byte => String = _.toString): String = {
+    val count = size(stack)
+    ((count - 1) to 0 by -1).foldLeft(
+      new StringBuilder()
+      .append("Stack:")
+      .append(" capacity = ").append((stack.length - 1) * 8).append(',')
+      .append(" count = ").append(count).append(',')
+      .append(" content (top-first) = [ ")
+    ){
+      (b, i) =>  b.append(translate(ByteStack.read(stack, i))).append(' ')
+    }.append(']').toString
   }
 
-  def push(b: Byte): Unit = {
-    checkAndGrow()
-    buffer(index) = b
-    index += 1
-  }
+  final def create(initialMaxOps: Int): Array[Int] =
+    new Array[Int](1 + 1 + ((initialMaxOps - 1) >> 3)) // count-slot + 1 for each set of 8 ops
 
-  // TODO remove bounds check
-  def pop(): Byte = {
-    index -= 1
-    buffer(index)
-  }
-
-  def peek(): Byte = buffer(index - 1)
-
-  def isEmpty(): Boolean = index <= 0
-
-  // to allow for external iteration
-  def unsafeBuffer(): Array[Byte] = buffer
-  def unsafeIndex(): Int = index
-
-  def unsafeSet(newI: Int): Unit =
-    index = newI
-
-  def invalidate(): Unit = {
-    index = 0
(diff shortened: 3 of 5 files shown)


vasilmkd:
I'm curious, why not an `Array[Long]`?

vasilmkd:
I've always wanted this change myself, so big :+1: from me.

vasilmkd:
Should we try inlining the methods in `object ByteStack` into `IOFiber`? 😆

viktorklang:
@vasilmkd

>I'm curious, why not an Array[Long]?

My thinking was that it would be unnecessary to store the current stack count in a Long, and if stored as an Int shifted into the Long, it'd be more operations to read and update it, without providing any better binpacking or array access checks.

vasilmkd:
> @vasilmkd
> 
> 
> 
> >I'm curious, why not an Array[Long]?
> 
> 
> 
> My thinking was that it would be unnecessary to store the current stack count in a Long, and if stored as an Int shifted into the Long, it'd be more operations to read and update it, without providing any better binpacking or array access checks.

That's a perfectly valid reason! Thanks for the explanation. I'll take a look at the code more closely.

viktorklang:
@vasilmkd Np—it's a good question! :)

djspiewak:
@vasilmkd can we finalize this before #2258? I want to get this in any hypothetical backport branch.

vasilmkd:
I will review it in a bit.

benchmarks/src/main/scala/cats/effect/benchmarks/ParallelBenchmark.scala
```diff
@@ -45,10 +45,10 @@ import java.util.concurrent.TimeUnit
 @OutputTimeUnit(TimeUnit.SECONDS)
 class ParallelBenchmark {
 
-  @Param(Array("100", "1000", "10000"))
+  @Param(Array( /*"100", */ "1000" /*, "10000"*/ ))
   var size: Int = _
 
-  @Param(Array("100", "1000", "10000", "100000", "1000000"))
+  @Param(Array( /*"100", "1000", */ "10000" /*, "100000", "1000000"*/ ))
```
vasilmkd, line 51: Should we revert this change? I'm pretty sure there exists an override that you can provide as part of the `Jmh/run` command.

build.sbt
```diff
@@ -450,7 +451,9 @@ lazy val example = crossProject(JSPlatform, JVMPlatform)
 lazy val benchmarks = project
   .in(file("benchmarks"))
   .dependsOn(core.jvm)
-  .settings(name := "cats-effect-benchmarks")
+  .settings(
+    name := "cats-effect-benchmarks",
+    javaOptions ++= Seq("-Dcats.effect.tracing.mode=none", "-Dcats.effect.tracing.exceptions.enhanced=false"))
```
vasilmkd, line 456: Big :+1:

vasilmkd (commented):
Again the question about inlining this code in `IOFiber`. It would change all callsites from `invokestatic` to `invokespecial`.

I guess it would need to be duplicated to support `SyncIO`, which is not ideal.

benchmarks/src/main/scala/cats/effect/benchmarks/ParallelBenchmark.scala
djspiewak: It's up to you. There's definitely an override. If we revert this change, then we should toss the override into the class header so we can copy/paste easily. The problem with the current matrix of parameters is it literally takes like 12 hours to run.

vasilmkd: Ok, keep it as is.

djspiewak:
> Again the question about inlining this code in IOFiber. It would change all callsites from invokestatic to invokespecial.

We could check the assembly, but I'd be willing to bet that `invokestatic` is basically as good as an `invokespecial` on a `private[this] final` method.

vasilmkd (approved):

PR #2252, opened 2021-08-21
Fibers dozing

vasilmkd:
Highly experimental and potentially unstable. Do not rebase on this branch (or do so at your own risk). I will force push and add/delete commits with prejudice.

### DIFF
--- a/benchmarks/src/main/scala/cats/effect/benchmarks/SleepBenchmark.scala
+++ b/benchmarks/src/main/scala/cats/effect/benchmarks/SleepBenchmark.scala
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2020-2021 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.benchmarks
+
+import cats.effect.IO
+import cats.effect.unsafe._
+import cats.syntax.all._
+
+import scala.concurrent.duration._
+
+import java.util.concurrent.TimeUnit
+import org.openjdk.jmh.annotations._
+
+/**
+ * To do comparative benchmarks between versions:
+ *
+ *     benchmarks/run-benchmark SleepBenchmark
+ *
+ * This will generate results in `benchmarks/results`.
+ *
+ * Or to run the benchmark from within sbt:
+ *
+ *     jmh:run -i 10 -wi 10 -f 2 -t 1 cats.effect.benchmarks.SleepBenchmark
+ *
+ * Which means "10 iterations", "10 warm-up iterations", "2 forks", "1 thread".
+ * Please note that benchmarks should be usually executed at least in
+ * 10 iterations (as a rule of thumb), but more is better.
+ */
+@State(Scope.Thread)
+@BenchmarkMode(Array(Mode.Throughput))
+@OutputTimeUnit(TimeUnit.MINUTES)
+class SleepBenchmark {
+
+  @Param(Array("10000"))
+  var size: Int = _
+
+  def sleepBenchmark(implicit runtime: IORuntime): Int = {
+    def fiber(i: Int): IO[Int] =
+      IO.sleep(1.nanosecond).flatMap { _ =>
+        IO(i).flatMap { j =>
+          IO.sleep(1.nanosecond).flatMap { _ =>
+            if (j > 1000)
+              IO.sleep(1.nanosecond).flatMap(_ => IO.pure(j))
+            else
+              IO.sleep(1.nanosecond).flatMap(_ => fiber(j + 1))
+          }
+        }
+      }
+
+    List
+      .range(0, size)
+      .traverse(fiber(_).start)
+      .flatMap(_.traverse(_.joinWithNever))
+      .map(_.sum)
+      .unsafeRunSync()
+  }
+
+  @Benchmark
+  def sleep(): Int = {
+    import cats.effect.unsafe.implicits.global
+    sleepBenchmark
+  }
+}
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
@@ -19,6 +19,7 @@ package unsafe
 
 import scala.annotation.switch
 import scala.concurrent.{BlockContext, CanAwait}
+import scala.concurrent.duration.FiniteDuration
 
 import java.util.concurrent.ThreadLocalRandom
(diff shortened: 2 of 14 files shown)


vasilmkd:
Some benchmark results (`benchmarks/Jmh/run -wi 10 -i 10 -f 2 -t 1 SleepBenchmark --jvmArgs -Dcats.effect.tracing.mode=none`):

`series/3.x`:
```
Benchmark             (size)   Mode  Cnt   Score   Error    Units
SleepBenchmark.sleep   10000  thrpt   20  24.447 ± 0.275  ops/min
```

`This PR`:
```
Benchmark             (size)   Mode  Cnt    Score   Error    Units
SleepBenchmark.sleep   10000  thrpt   20  311.734 ± 6.158  ops/min
```

vasilmkd:
Some interesting CPU utilization screenshots taken during the execution of the benchmarks:

`series/3.x`:
Screen Shot 2021-08-21 at 23 24 44

Take a note of the CPU hovering at around 250% (it's how Mac reports CPU usage, with the maximum being 800% for my system) meaning that it uses between 2-3 threads actively. Also take a note at the red bar underneath the blue one, which is the kernel scheduler CPU utilization. This bar is usually flush with the bottom of the graph, and it only shows up when a lot of context switching is going on.

`This PR`:
Screen Shot 2021-08-21 at 23 28 29

Full CPU utilization and no kernel scheduler utilization.

If anyone knows the implementation details of the current Cats Effect timer, this is not surprising, but there are better days coming. 😄 

I will go into detail more soon.

(Please don't get confused by the numbers reported on the left, they are sensitive to when I was taking the screenshot, and the system obviously had to do some context switching, the important part is the flat middle and to the right section in both graphs, while the benchmark was executing).

Edit:
I realized I quoted numbers which are not shown in the original screenshots, here they are:

`series/3.x`:
Screen Shot 2021-08-21 at 23 43 23

`This PR`:
Screen Shot 2021-08-21 at 23 42 47

djspiewak:
Drift measurements:

```
Warming up...
Measuring unloaded...
Unloaded overhead: 0.233673337775
Measuring heavily loaded...
Loaded overhead: 0.41391992757500007
Measuring unloaded 100x...
Unloaded overhead 100x: 0.26452540499166677
Loaded overhead: 0.324064256175
```

The largest improvement by far is the Heavily Loaded 100x Overhead, which is fairly realistic for most network apps.

**Edit** For reference, here's series/3.x:

```
Warming up...
Measuring unloaded...
Unloaded overhead: 0.2656264197166667
Measuring heavily loaded...
Loaded overhead: 0.6483898653416666
Measuring unloaded 100x...
Unloaded overhead 100x: 0.5070073040333334
Loaded overhead: 1.2333863049916665
```

djspiewak:
Published as `3.2-24-5b17308` in case anyone wants to do downstream testing/benchmarking.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -243,7 +254,40 @@ private final class WorkerThread(
...
+        val now = System.nanoTime()
+
+        var cont = true
+        while (cont) {
+          val head = sleepers.head()
+
+          val delay = head.triggerTime - now
+          if (delay <= 0) {
+            if (head.get()) {
+              head.callback.run()
```
viktorklang: Can new sleepers be added as a result of calling `head.callback.run()`?

vasilmkd: No, this `.run()` completes the `IO.async` callback with which sleeping is implemented. This _only_ puts the fiber in the `LocalQueue` and it will be resumed when its turn comes.

viktorklang: Would it not be faster to do that directly here instead? (instead of going via a Runnable (virtual dispatch))

vasilmkd: I see what you mean. The `cats.effect.unsafe.Scheduler` interface essentially prevents that, unfortunately.

vasilmkd: I'll see if I can add a fast-path that circumvents that.

vasilmkd:
New benchmark results (`SleepBenchmark`):

```
Benchmark             (size)   Mode  Cnt    Score    Error    Units
SleepBenchmark.sleep   10000  thrpt   20  341.512 ± 15.411  ops/min
```

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -243,7 +255,40 @@ private final class WorkerThread(
...
+      if (sleepers.nonEmpty) {
+        val now = System.nanoTime()
+
+        var cont = true
+        while (cont) {
+          val head = sleepers.head()
+
+          val delay = head.triggerTime - now
+          if (delay <= 0) {
+            if (head.get()) {
```
viktorklang, line 283: @vasilmkd Don't we need getAndSet(false) here to avoid someone looking at this at the same time?

vasilmkd: Our `cats.effect.unsafe.Scheduler` interface which we're tied to, for better or for worse exposes the following signature: `def sleep(delay: FiniteDuration, task: Runnable): Runnable`. The returned `Runnable` needs to be invoked when we want to cancel the `sleep`. I tried to save on a few allocations by cheating with `SleepCallback extends Runnable`. The run method of this class just sets the `AtomicBoolean` that this class extends (again, cheating allocations) to `false`, indicating that this sleep has been canceled and should not be invoked in the future, when it's time would have come to resume (this is exactly how `ScheduledThreadPoolExecutor` from the JDK works as well).

I don't think there's a need to use anything other than `get` and `set` in this case, we only care to read a volatile boolean value which might have been set by another thread. I believe that's all that's needed.

vasilmkd: In the future I'd like to have a smarter mechanism for removing the reference to the canceled sleep sooner, but this is no different to what we already have in Cats Effect, so I think it's a good enough first step.

benchmarks/src/main/scala/cats/effect/benchmarks/SleepBenchmark.scala
```diff
@@ -0,0 +1,77 @@
...
+ */
+@State(Scope.Thread)
+@BenchmarkMode(Array(Mode.Throughput))
+@OutputTimeUnit(TimeUnit.MINUTES)
+class SleepBenchmark {
+
+  @Param(Array("10000"))
+  var size: Int = _
+
+  def sleepBenchmark(implicit runtime: IORuntime): Int = {
```
viktorklang, line 51: @vasilmkd Might I suggest that you also create an identical method (sans the sleeps) so that in the same benchmark we can see the overhead added by the sleeps side-by-side?

Also, cache the `1.nanosecond` Duration so that the allocation overhead is minimized and the GC time is reduced.

djspiewak: I think this is equivalent to the following benchmark, which replaces the sleeps with `cede`: https://github.com/typelevel/cats-effect/blob/series/3.x/benchmarks/src/main/scala/cats/effect/benchmarks/WorkStealingBenchmark.scala#L52

PR #2254, opened 2021-08-22
Check `WorkerThread` ownership before scheduling

vasilmkd:
This covers the interleaving between two instances of `WorkStealingThreadPool`. I did not have a chance to test the performance impact thoroughly, but in the quick benchmark run I did, the new code was faster, which, I'm not sure whether to trust. I will repeat the run when I have the time.

`series/3.x`:
```
WorkStealingBenchmark.scheduling  1000000  thrpt   10  13.270 ± 1.109  ops/min
```

`This PR`:
```
WorkStealingBenchmark.scheduling  1000000  thrpt   10  13.981 ± 1.537  ops/min
```

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
@@ -355,17 +355,22 @@ private[effect] final class WorkStealingThreadPool(
     val pool = this
     val thread = Thread.currentThread()
 
-    if (thread.isInstanceOf[WorkerThread] &&
-      thread.asInstanceOf[WorkerThread].isOwnedBy(pool)) {
-      thread.asInstanceOf[WorkerThread].reschedule(fiber)
-    } else if (thread.isInstanceOf[HelperThread] &&
-      thread.asInstanceOf[HelperThread].isOwnedBy(pool)) {
-      thread.asInstanceOf[HelperThread].schedule(fiber)
+    if (thread.isInstanceOf[WorkerThread]) {
+      val worker = thread.asInstanceOf[WorkerThread]
+      if (worker.isOwnedBy(pool)) {
+        worker.reschedule(fiber)
+      } else {
+        scheduleExternal(fiber)
+      }
+    } else if (thread.isInstanceOf[HelperThread]) {
+      val helper = thread.asInstanceOf[HelperThread]
+      if (helper.isOwnedBy(pool)) {
+        helper.schedule(fiber)
+      } else {
+        scheduleExternal(fiber)
+      }
     } else {
-      val random = ThreadLocalRandom.current()
-      overflowQueue.offer(fiber, random)
-      notifyParked(random)
-      ()
+      scheduleExternal(fiber)
     }
   }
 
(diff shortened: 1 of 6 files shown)


vasilmkd:
Here's another run:

`series/3.x`:
```
WorkStealingBenchmark.scheduling  1000000  thrpt   20  15.228 ± 0.285  ops/min
```

`This PR`:
```
WorkStealingBenchmark.scheduling  1000000  thrpt   20  14.966 ± 0.491  ops/min
```

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -351,47 +351,51 @@ private[effect] final class WorkStealingThreadPool(
...
+        worker.reschedule(fiber)
+
+      case helper: HelperThread if helper.isOwnedBy(pool) =>
+        helper.schedule(fiber)
+
+      case _ =>
+        val random = ThreadLocalRandom.current()
+        overflowQueue.offer(fiber, random)
+        notifyParked(random)
+        ()
```
djspiewak: I've seen some really weird things with the bytecode emitted by guarded cases. Can we formulate this to be more explicit?

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -351,47 +351,51 @@ private[effect] final class WorkStealingThreadPool(
...
+        worker.schedule(fiber)
+
+      case helper: HelperThread if helper.isOwnedBy(pool) =>
+        helper.schedule(fiber)
+
+      case _ =>
+        val random = ThreadLocalRandom.current()
+        overflowQueue.offer(fiber, random)
+        notifyParked(random)
+        ()
```
djspiewak: Also here

djspiewak (changes requested):

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
vasilmkd: Yeah sure. I agree completely.

vasilmkd:
A new run of the scheduling benchmark:
```
WorkStealingBenchmark.scheduling  1000000  thrpt   20  14.884 ± 0.635  ops/min
```

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -351,11 +351,15 @@ private[effect] final class WorkStealingThreadPool(
...
     val thread = Thread.currentThread()
-    if (thread.isInstanceOf[WorkerThread]) {
-      thread.asInstanceOf[WorkerThread].schedule(fiber)
-    } else if (thread.isInstanceOf[HelperThread]) {
+
+    if (thread.isInstanceOf[WorkerThread] &&
+      thread.asInstanceOf[WorkerThread].isOwnedBy(pool)) {
+      thread.asInstanceOf[WorkerThread].reschedule(fiber)
+    } else if (thread.isInstanceOf[HelperThread] &&
+      thread.asInstanceOf[HelperThread].isOwnedBy(pool)) {
```
djspiewak: This may be slightly faster:

```scala
  val wt = thread.asInstanceOf[WorkerThread]
  if (wt.isOwnedBy(pool)) {
    wt.reschedule(fiber)
  } else {
    val random = ThreadLocalRandom.current()
    overflowQueue.offer(fiber, random)
    notifyParked(random)
    ()
  }
} else if (thread.isInstanceOf[HelperThread]) {
  val ht = thread.asInstanceOf[HelperThread]
  if (ht.isOwnedBy(pool)) {
    ht.schedule(fiber)
  } else {
    val random = ThreadLocalRandom.current()
    overflowQueue.offer(fiber, random)
    notifyParked(random)
    ()
  }
} else {
  val random = ThreadLocalRandom.current()
  overflowQueue.offer(fiber, random)
  notifyParked(random)
  ()
}
```

djspiewak (changes requested):

vasilmkd:
Another run, but thermal throttling might have kicked in. The second fork was down in terms of the score.

```
WorkStealingBenchmark.scheduling  1000000  thrpt   20  14.218 ± 0.894  ops/min
```

djspiewak (approved):

Issue #1029, opened 2020-07-31
Live fiber dumps


RaasAhsan:
Going back and copying some of the ideas I had in https://typelevel.org/cats-effect/tracing/ to concrete issues. Since Cats Effect now supports fiber tracing, it goes without saying that we should support live fiber dumps as well. We should probably build out some of the other tracing features first before tackling this. 

Ideally the dump includes the following information:
1. The up-to-date trace for each fiber
2. Current carrier thread. This is related to #936.
3. Current fiber status. Some ideas are RUNNING, CANCELED, TERMINATED, ASYNC, WAITING
4. If we're in an asynchronous boundary, some auxiliary information reflecting what the asynchronous boundary is. e.g. if we're waiting on a `Deferred`, or if we're sleeping, etc.
5. Fiber identity and possibly ancestry

Some problems we should start thinking about:
1. Fiber runtimes may be managing thousands to hundreds of thousands of fibers at any given moment. Dumping all of them might be infeasible.
2. The act of producing a dump requires that we track all active running fibers. We should garbage collect those handles if the application no longer requires them, otherwise we are effectively creating a memory leak. `WeakReference` and `SoftReference` are of interest here.
3. How often should the manifest of running fibers be updated? I think the only reasonable answer here is at asynchronous boundaries.
4. How do we request a dump? Some ideas are registering a process signal handler, exposing a JMX MBean, or just a regular library API call.
5. Like other tracing features, the implementation of fiber dumps should have negligible performance impact when it isn't being used.

We should aim to get this in for Cats Effect 2, but think about how it would be designed for Cats Effect 3 as well.

Daenyth:
> exposing a JMX MBean

I'd really like if it we had *at least* this, because when I get async locks, I often end up in visualvm trying to analyze a heap dump from the live system.

djspiewak:
Okay I'm pretty sure this is now tractable. I have a PoC for very low overhead tracking of suspended fibers. Task list to get this finished up:

- [ ] Merge @armanbilge's stuff so we don’t create conflicts (#2207)
- [x] Clean up my PoC and PR
- [ ] Merge @vasilmkd's timer stuff so we don’t create conflicts (#2252)
- [ ] Get Vasil to PR his magic on live fiber enumeration
- [ ] Write a thingy which uses both functions (live and suspended) and gets all active fibers
- [ ] Use above thing to print all active traces (note: IOFiber needs an identifying toString)
[...]

PR #2258, opened 2021-08-25
Fast weak bag for tracking suspended fibers

djspiewak:
This is a fun little data structure. It provides the following guarantees:

- O(1) insert
- O(1) remove
- O(n) iteration
- Invalid data is never returned
- The invariants are never corrupted

All operations are "best effort" and are allowed to lose data in rare situations.

The way this works is by appending to a directly-encoded dynamic array and keeping track of the insert offset within the data (an `IOFiber` in this case). Whenever we need to remove a fiber, we just `null` out its index and increment the `fragments` count. Once fragmentation exceeds 50% of the total bag capacity, we defragment by iterating the entire bag and checking `null` status on both the array slot and the `WeakReference` (this is also how we clean up GC'd fibers).

The memory barriers are externalized, and it is expected that data will be lost in contention scenarios. Most often, the changes to this data structure will be published by the volatile writes within the thread pool. This will be much more lossy on ARM than on x86.

More comprehensive benchmarks are coming, but initial results suggest a roughly 6% increase in overhead on the `parTraverse` benchmark, and zero added overhead in `scheduler`.

This forms one required piece of #1029


core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -998,21 +1002,33 @@ private final class IOFiber[A](
...
+     * it should actually be *faster* because CAS isn't primitive but get-and-set is.
+     */
+    val acquired = suspended.getAndSet(false)
+
+    if (acquired) {
+      runtime.unmonitor(monitorIndex)
+    }
+
+    acquired
+  }
```
vasilmkd, line 1024: Since we always branch on `resume()` I wonder if it is better to inline the monitor registration in the code paths where this is taken (in cancel, async, and wherever else) instead of introducing another branch here.

djspiewak: Good point! Let's do that

djspiewak: Actually, on second thought, let's do that in a second pass. There's some subtle complexity there and I want us to be able to iterate on this mechanism before we make it all ugly-ified.

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimePlatform.scala
```diff
@@ -0,0 +1,103 @@
...
+  // used to lazily defragment the buffer (in particular, when (fragments / index) > 0.5)
+  private[this] var fragments: Int = 0
+
+  val shutdown: () => Unit = { () =>
+    this._shutdown()
+    buffer = null
+  }
+
+  private[effect] def suspended(): Set[IOFiber[_]] = {
+    var back = Set[IOFiber[_]]()
```
vasilmkd: How about using a `mutable.Set` and calling [`toSet`](https://www.scala-lang.org/api/2.13.6/scala/collection/mutable/Set.html#toSet[B%3E:A]:scala.collection.immutable.Set[B]) at the end?

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimePlatform.scala
```diff
@@ -0,0 +1,103 @@
...
+    val buf = buffer
+    val max = index
+
+    var i = 0
+    while (i < max) {
+      val ref = buf(i)
+      if (ref != null) {
```
vasilmkd: I know this is probably stupid, but is there a need to be extra explicit and call `ne` instead of `!=`? I've used it in other parts of the codebase.

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimePlatform.scala
```diff
@@ -0,0 +1,103 @@
...
+      i += 1
+    }
+
+    back
+  }
+
+  private[effect] def monitor(self: IOFiber[_]): Int = {
+    checkAndGrow()
+    val idx = index
+    buffer(idx) = new WeakReference(self)
```
vasilmkd, line 63: Too bad `WeakReferences` can be re-set so the objects could be cached.

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimePlatform.scala
```diff
@@ -0,0 +1,103 @@
...
+    index += 1
+    idx
+  }
+
+  private[effect] def unmonitor(idx: Int): Unit = {
+    buffer(idx) = null
+    fragments += 1
+  }
+
+  private[this] def checkAndGrow(): Unit = {
```
vasilmkd, line 74: Can we slap a `@tailrec` on this method? 😆

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -732,6 +735,7 @@ private final class IOFiber[A](
              * for `canceled` in `shouldFinalize`, ensuring no finalisation leaks
              */
             suspended.getAndSet(true)
+            monitorIndex = runtime.monitor(this)
```
vasilmkd, line 738: It might be smart to factor this out in a method `suspendWithFullMemoryFence`. It doesn't have to be a part of this PR.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -998,21 +1002,33 @@ private final class IOFiber[A](
...
+    acquired
+  }
 
-  private[this] def suspend(): Unit =
+  private[this] def suspend(): Unit = {
+    monitorIndex = runtime.monitor(this)
     suspended.set(true)
+  }
+
+  private[effect] def updateMonitorIndex(idx: Int): Unit =
```
vasilmkd, line 1031: `private[effect]` is redundant here.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -1407,6 +1423,12 @@ private final class IOFiber[A](
     Thread.currentThread().interrupt()
     null
   }
+
+  // overrides the AtomicReference#toString
+  override def toString: String = {
+    val state = if (suspended.get()) "SUSPENDED" else "RUNNING"
+    s"cats.effect.IOFiber@${System.identityHashCode(this).toHexString} $state"
+  }
```
vasilmkd, line 1425: FINALLY!

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimePlatform.scala
```diff
@@ -0,0 +1,103 @@
...
+
+        var i = 0
+        var index2 = 0
+        while (i < len) {
+          val ref = buffer(i)
+          if (ref != null && ref.get() != null) {
+            val fiber = ref.get()
+            if (fiber != null) {
+              buffer2(index2) = ref
+              fiber.updateMonitorIndex(index2)
```
vasilmkd, line 88: What happens when this defragmentation process coincides with a fiber waking up?

Can the following scenario happen:
1. This line gets executed, updating the monitor index for the fiber.
2. The fiber wakes up, it removes its reference from the old `buffer` (which is a reference to a different fiber in reality) (`buffer2` isn't published yet)
3. Is this even bad?

vasilmkd (commented):
Very cool data structure and very exciting changes. I'm pumped!

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimePlatform.scala
djspiewak: What could end up happening in that case is the fiber would live forever in the new table. There's a similar scenario which can happen if a fiber `monitor`s itself while table rescaling is happening. In either case, something gets lost.

A fiber that lives forever in the table is still only weakly referenced, so there's no memory leak. If the fiber is still alive, it will be returned from the `suspended()` method despite the fact that it's not suspended. When we implement fiber dumps, that will result in a duplication between the live and suspended sets, which will be unioned away.

The fibers that are *missed* due to monitoring during rescaling or defragmentation are more unfortunate, but they aren't likely to be common cases. There are several race conditions like this throughout the fiber dump process which can miss active or suspended fibers, depending on timing.

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimePlatform.scala
djspiewak: Shouldn't make a difference but I made the change anyway. :-)

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimePlatform.scala
djspiewak: Yeah I briefly thought about doing this. The main reason I didn't is the general bias away from object interning on the JVM.

core/shared/src/main/scala/cats/effect/IOFiber.scala
djspiewak: It is, just trying to stay relatively explicit.

djspiewak:
This CI failure is absolutely fascinating:

```
[error] x Deferred — issue #380: with cooperative full async boundaries
[error]  the value is false (DeferredJVMSpec.scala:174)
```

I don't understand why.

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimePlatform.scala
armanbilge: What's this?

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimePlatform.scala
```diff
@@ -0,0 +1,107 @@
...
+          back += fiber
+        }
+      }
+      i += 1
+    }
+
+    back.toSet
+  }
+
+  private[effect] def monitor(self: IOFiber[_]): Int = {
```
armanbilge, line 60: Why `self`? I found it confusing to read.

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimePlatform.scala
```diff
@@ -0,0 +1,107 @@
...
+        val len = buffer.length
+        val buffer2 = new Array[WeakReference[IOFiber[_]]](len)
+
+        var i = 0
+        var index2 = 0
+        while (i < len) {
+          val ref = buffer(i)
+          if ((ref ne null) && (ref.get() ne null)) {
+            val fiber = ref.get()
+            if (fiber ne null) {
```
armanbilge, line 86: Aren't the conditions slightly redundant here?

armanbilge (commented):
How is this enabled/disabled btw?

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -999,21 +1003,33 @@ private final class IOFiber[A](
...
+
+    if (acquired) {
+      runtime.unmonitor(monitorIndex)
+    }
+
+    acquired
+  }
 
-  private[this] def suspend(): Unit =
+  private[this] def suspend(): Unit = {
```
vasilmkd, line 1026: Thinking about this a little bit more, not every `suspend()` is born equal, i.e. not all suspends need to monitor the fiber, only the one `suspend()` that creates the async callback. All other `suspend()` calls actually just give up on resuming the fiber. That being said, for this to be supported, we need a symmetrical `resume()` which would only unmonitor the fiber on completion of the async callback.

vasilmkd: I'm aware there are some flaws in this argument, but I'm thinking of an alternative which would exploit something along those lines.

vasilmkd: Cancelation of a suspended fiber should unmonitor it.

vasilmkd:
I ran the `ParallelBenchmark.parTraverse` benchmark on a 16 core cloud instance and got the following exception:
```
[error] Exception in thread "io-compute-5" java.lang.ArrayIndexOutOfBoundsException: 4096
[error] 	at cats.effect.unsafe.IORuntimePlatform.monitor(IORuntimePlatform.scala:63)
[error] 	at cats.effect.IOFiber.runLoop(IOFiber.scala:738)
[error] 	at cats.effect.IOFiber.execR(IOFiber.scala:1176)
[error] 	at cats.effect.IOFiber.run(IOFiber.scala:131)
[error] 	at cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:389)
```

core/shared/src/main/scala/cats/effect/IOFiber.scala
armanbilge: Is it possible to include (at least some of) the current IO trace, if available? So maybe we can know what this fiber is doing :)

djspiewak:
IMO the GC approach is *much* more elegant than this PR. I'm in favor of closing this one and concentrating on the approach demonstrated in #2368

PR #2264, opened 2021-08-26
Fix batched queue starvation

### LINKED ISSUE #2263: The `LocalQueue` can starve the `batchedQueue`
Thanks @adamgfraser for the reproduction.

```scala
import cats.effect._
import cats.implicits._

object CatsScheduling extends IOApp {

    io.as(ExitCode.Success)

  def yieldUntil(ref: Ref[IO, Boolean]): IO[Unit] =
    ref.get.flatMap(b => if (b) IO.unit else IO.cede *> yieldUntil(ref))

  val io: IO[Unit] =
    for {
      n    <- IO(java.lang.Runtime.getRuntime.availableProcessors)
      done <- Ref.of[IO, Boolean](false)
      fibers <- List.range(0, n - 1).traverse(_ => yieldUntil(done).start)
      _ <- IO.unit.start.replicateA(200)
      _ <- done.set(true).start
      _ <- IO.unit.start.replicateA(1000)
      _ <- yieldUntil(done)
      _ <- fibers.traverse(_.join)
    } yield ()
}
```

vasilmkd:
Fixes #2263.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/LocalQueue.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/LocalQueue.scala
@@ -644,17 +644,59 @@ private final class LocalQueue {
     null
   }
 
+  /**
+   * Steals a batch of enqueued fibers and transfers the whole batch to the
+   * batched queue.
+   *
+   * This method is called by the runtime to restore fairness guarantees between
+   * fibers in the local queue compared to fibers on the overflow and batched
+   * queues. Every few iterations, the overflow and batched queues are checked
+   * for fibers and those fibers are executed. In the case of the batched queue,
+   * a batch of fibers might be obtained, which cannot fully fit into the local
+   * queue due to insufficient capacity. In that case, this method is called to
+   * drain one full batch of fibers, which in turn creates space for the fibers
+   * arriving from the batched queue.
+   *
+   * Conceptually, this method is identical to [[LocalQueue#dequeue]], with the
+   * main difference being that the `head` of the queue is moved forward by as
+   * many places as there are in a batch, thus securing all those fibers and
+   * transferring them to the batched queue.
+   *
+   * @note Can '''only''' be correctly called by the owner [[WorkerThread]].
+   *
+   * @param batched the batched queue to transfer a batch of fibers into
+   * @param random a reference to an uncontended source of randomness, to be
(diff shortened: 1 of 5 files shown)


djspiewak:
The second (**edit** now removed) commit trades away some fairness for some throughput (about 12% more). With that commit, the bug fix results in an 8% reduction in the `scheduler` benchmark; without the fairness tradeoff, it's about 15%. Not too bad, honestly. I could go either way on the overflow ticks. It's always going to be possible to improve throughput by bumping up that coefficient, but it's probably safer to keep it lower.

vasilmkd:
I pushed a bogus timeout for the test just to see if it passes at all.

Edit: reverted.

djspiewak:
This represents a roughly 21% reduction in performance from series/3.x on the `scheduler` benchmark (there was more thermal throttling when I ran against this PR than there was when I ran against series/3.x, so take those numbers as more of an upper bound; Vasil got a 17.5% reduction in his testing). Lots of ways it can be improved though. Let's get the fix in first and then work on the refinements.

vasilmkd:
Before:
```
WorkStealingBenchmark.scheduling  1000000  thrpt   20  13.567 ± 0.360  ops/min
```

After:
```
WorkStealingBenchmark.scheduling  1000000  thrpt   20  11.165 ± 0.293  ops/min
```

vasilmkd:
I willingly sacrificed some performance here for a quicker turnaround of the fix. We're working on tuning certain parameters of the runtime (necessary tuning which we never really found the time for, but now we'll do it properly) and we should regain some of the performance loss of this PR.

djspiewak (approved):
Random note: batch size is 128, and we check the batched queue once every 128 iterations. However, we may progress as few as 57 fibers every 64 iterations, and we *may* have also taken from the external queue, meaning that we could have only evaluated 57 + 57 - 1 = 113 fibers by the time we go back to the batched queue, which may in turn result in 128 of our local fibers getting dumped to the tail of the tail of the batched queue. The overlap here is 128 - 113 = 15 fibers. These fibers experience unfairness (though not indefinite unfairness, since they'll be at the front of the batch the *next* time through, meaning that they are guaranteed to run).

This is a pathological edge case and will be improved by subsequent iterations. It's fine for now.

senia-psm:
There might be an issue with this fix. Please see #2269

PR #2272, opened 2021-08-28
Revert Async#fromPromise generalization

### LINKED ISSUE #2271: FS2 compilation failures due to change in signature of fromPromise
#2237 changes `fromPromise` to take an `IO[Thenable[A]]` instead of an `IO[Promise[A]]`. As a result, fs2-io has 61 compilation errors when using expressions of the form `F.fromPromise(F.delay(somethingThatReturnsPromise))`. The compiler doesn't know the infer `somethingThatReturnsPromise` as a `Thenable[X]`.

djspiewak:
The source compatibility breakage was somewhat extreme.

Fixes #2271

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -453,7 +453,9 @@ lazy val benchmarks = project
   .dependsOn(core.jvm)
   .settings(
     name := "cats-effect-benchmarks",
-    javaOptions ++= Seq("-Dcats.effect.tracing.mode=none", "-Dcats.effect.tracing.exceptions.enhanced=false"))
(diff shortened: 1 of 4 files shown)


armanbilge:
Can we at least cherry-pick the original `fromThenable` into this? 8e0cc4efb317fca700ce189e3c32f38e8c429693

djspiewak:
I think that's probably what we will do, but I would rather bias back to a safe state and then move forward with additions.

vasilmkd (approved):

PR #2273, opened 2021-08-28
Fix #2236 implement Async.fromThenable (take 2)

### LINKED ISSUE #2236: `IO.fromPromise` should accept `js.Thenable`, instead of `js.Promise`

armanbilge:
Cherry-picks 8e0cc4efb317fca700ce189e3c32f38e8c429693 which implements `fromThenable` as a separate method to avoid problems like #2271.

### DIFF
--- a/core/js/src/main/scala/cats/effect/IOCompanionPlatform.scala
+++ b/core/js/src/main/scala/cats/effect/IOCompanionPlatform.scala
@@ -18,7 +18,7 @@ package cats.effect
 
 import scalajs.js
 
-import scala.scalajs.js.Promise
+import scala.scalajs.js.{Promise, Thenable}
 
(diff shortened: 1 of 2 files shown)


vasilmkd:
@djspiewak I think we should target this at `series/3.x` and use the test that you added in #2272 as a sort of proof that we're not regressing in terms of source compatibility. Thoughts?

armanbilge:
> we're not regressing in terms of source compatibility.

This only adds a new method (with a new name), and doesn't touch the existing one. So I think we're pretty safe with this.

djspiewak:
I agree we're pretty safe, but I think that philosophically we should still target `series/3.x` unless it's an urgently-needed feature.

djspiewak (approved):

Issue #1210, opened 2020-09-19
Testkit documentation


RaasAhsan:
I don't think we have a documentation issue for this one. We should be able to backport anything we write about `TestContext` to CE2 as well.

Kazark:
There are docs for [TestContext](https://typelevel.org/cats-effect/api/3.x/cats/effect/kernel/testkit/TestContext.html), but they are out of date as they reference `Timer`, which does not exist in CE3.

Or if you mean for the migration guide, I am slowly working through how to migrate them [here](https://stackoverflow.com/a/67592671/834176).

djspiewak:
We have `TestControl` pretty well documented at this point.

PR #2276, opened 2021-08-30
Added `TestControl`, with first-class support for time mocking

### LINKED ISSUE #1678: Think more seriously about what we're going to do with time mocking
`TimeT`? Orphan `Clock`? `TestContext`? The first of these is my preferred option but we need to play with it more to make sure it actually works.

### LINKED ISSUE #2220: Loss of the ability to mock Clock in version 3
I am currently in a process of upgrading a relatively large code base that uses Cats Effect and FS2 as building blocks for implementing many micro-service. One thing that is causing too much trouble is the changes made to Clock in the version 3. We have a large number of unit tests that mock the Clock in version 2 to control the timing and test various aspects of the code.

Now it might be my lack of knowledge but so far I have not found a way to mock Type class Clock provided by version 3. This is a complete deal breaker as we are absolutely not going to give up the tests in question. So far, it seams I have only 2 options:

1 - Stay on v2 for foreseeable future and abandon the upgrade effort.
2 - Create a local copy of the Clock trait from version 2 and replace all use of it in the code base.

The first option seams easy at first but that deprives us from taking advantage of any new improvements that come with version 3 especially the IO as I was hoping to replace the Monix Task with it. Also there is a fear of being left behind as the entire eco-system moves on.

The second one seams more logical at this point to me as it might just be very easy replacement. However, it leave a bad taste as it is constant reminder of something that used to work and now doesn't!

I hope I am not overly critical and I want to be absolutely clear that I love working on the code base using these 2 great libraries that make defining very complicated systems a breeze. But I am seriously stuck on this issue for some time now. I see an issue mentioning something similar to this problem (https://github.com/typelevel/cats-effect/issues/1678). But there is no serious activity on that for some months and it is not clear what the scope of that issue is. That is why I decided to open a new issue here.

Thanks in advance for any guidance that can help me solve this issue.

djspiewak:
Fixes #2220
Fixes #1678

Time mocking is back and better than ever! Yay!!! Thoughts very much welcome. Folks who have expressed an interest in this include @SystemFw, @ChristopherDavenport, @benhutchison, and many others. Now is the time to give feedback! I tried to be comprehensive in the `TestControl` scaladoc.

Note that this also includes a binary-compatible (but source- and semantically-incompatible) reworking of `TestContext`. The previous API conflated too many concerns and was very error-prone. Time advancement and schedule ticking is now fully separated. Also, as a bonus, I found (and fixed!) a bug in `tickAll` which was causing all of our test suites to take *dramatically* longer than necessary, and now they run much faster both in CI and on my laptop. Drinks all around.

The primary concept here is that `TestControl` represents a relatively low level mechanism for controlling the runtime. Higher level mechanisms should be built on top of this and integrated within specific downstream test frameworks (such as munit). Note that, just as before, there's still no substitute for running your program on the *real* runtime, and there are several programs which will deadlock on `TestContext` but not on the real thing. However, for anything involving `sleep`, this kind of thing is basically indispensable.

In theory this should be a lot more convenient than what we had to do in CE2 (creating artificial `Clock` instances), since the runtime is automatically threaded through at all points, meaning that you can just use the normal `Clock` and `Temporal` instances and everything will just work out of the box. In fact, time mocking even works on the bare `IO` functions (e.g. `IO.sleep`, `IO.monotonic`, and `IO.realTime`).

As a final goodie, I added support for configuring the source of randomness used in `TestContext`, similar to how ScalaCheck supports optionally setting the seed as a base64-encoded value. I would imagine that downstream test framework support will print out the `seed` in any error messages, as well as provide an API for optionally configuring that seed and passing it to the `TestControl` they use internally to run your examples.

As I said, **now is the time for feedback!** This is a pretty meaningful API addition and I'd like to make sure we get it right.


djspiewak:
Published as `3.3-7-579f468` so that we can experiment in downstream libraries like https://github.com/typelevel/munit-cats-effect and https://github.com/typelevel/cats-effect-testing.

kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/TestContext.scala
```diff
@@ -157,6 +164,28 @@ final class TestContext private () extends ExecutionContext { self =>
...
+   * minimum time-oriented progress on the task queue (e.g. `tick(nextInterval())`).
+   */
+  def nextInterval(): FiniteDuration = {
+    val s = state
+    (s.tasks.min.runsAt - s.clock).max(Duration.Zero)
+  }
+
+  def advance(time: FiniteDuration): Unit =
+    synchronized {
+      require(time > Duration.Zero)
```
vasilmkd: I know this is very minor but this line doesn't need to be inside the critical region.

kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/TestContext.scala
```diff
@@ -157,6 +164,28 @@ final class TestContext private () extends ExecutionContext { self =>
...
+   * `Duration.Zero`. Passing this value to [[tick]] will guarantee
+   * minimum time-oriented progress on the task queue (e.g. `tick(nextInterval())`).
+   */
+  def nextInterval(): FiniteDuration = {
+    val s = state
+    (s.tasks.min.runsAt - s.clock).max(Duration.Zero)
+  }
+
+  def advance(time: FiniteDuration): Unit =
+    synchronized {
```
vasilmkd, line 104: We've had combinations of `ticked` and `real` before, and I can see this line called from inside a `real` `IO`. Should we mark this `scala.util.blocking`?

Ditto for the other `synchronized` blocks later in the code.

kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/TestContext.scala
```diff
@@ -183,73 +212,58 @@ final class TestContext private () extends ExecutionContext { self =>
...
-   * @param time is an optional parameter for simulating time passing;
-   */
-  def tick(time: FiniteDuration = Duration.Zero): Unit = {
-    val targetTime = this.stateRef.clock + time
-    var hasTasks = true
+  @tailrec
+  def tick(): Unit =
+    if (tickOne()) {
+      tick()
+    }
```
vasilmkd, line 157: 🤯

kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/TestContext.scala
```diff
@@ -183,73 +212,58 @@ final class TestContext private () extends ExecutionContext { self =>
...
 
-    while (hasTasks) synchronized {
-      val current = this.stateRef
+  private[testkit] def tick(time: FiniteDuration): Unit = {
+    advance(time)
+    tick()
+  }
```
vasilmkd: Are `tick(time: FiniteDuration)` and `tick$default$1` here for maintaining binary compatibility?

tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
```diff
@@ -880,14 +887,15 @@ class ResourceSpec extends BaseSpec with ScalaCheck with Discipline {
 
         target.use_.unsafeToFuture()
 
-        ticker.ctx.tick(100.millis)
+        ticker.ctx.tick()
+        ticker.ctx.advanceAndTick(100.millis)
```
vasilmkd, line 891: Why do we need a `tick()` followed by an `advanceAndTick` here? Trying to understand.

vasilmkd: Is it the fact that it's right after an `unsafeToFuture()`?

testkit/shared/src/main/scala/cats/effect/testkit/TestControl.scala
```diff
@@ -0,0 +1,260 @@
...
+   * This means that when `sleep` is finally submitted to the runtime,
+   * it is scheduled for the time offset equal to `1.second + 100.millis`,
+   * since time was already advanced `1.second` before it had a chance
+   * to submit. Of course, time has only been advanced by `1.second`,
+   * thus the `sleep` never completes and the `println` cannot ever run.
+   *
+   * There are two possible solutions to this problem: either call [[tick]]
+   * ''first'' (before calling `advanceAndTick`) to ensure that the `sleep`
+   * has a chance to schedule itself, or simply use [[tickAll]] if you
+   * do not need to run assertions between time windows.
```
vasilmkd: Found the explanation for `tick()` followed by an `advanceAndTick()` thanks. Ignore the other question.

testkit/shared/src/main/scala/cats/effect/testkit/TestControl.scala
```diff
@@ -0,0 +1,260 @@
+/*
```
vasilmkd, line 1: Due to lack of comments on whole files, this is fantastic, I'm impressed by the integration. I cannot comment much on the API because I'm too used to it, but overall it makes sense to me.

vasilmkd (dismissed):
Very excited. Marking this as request changes but the changes are pretty minor.

Regarding the binary compatibility, because of the change in behavior, at least to me, the whole situation seems to give a false sense of security that this could be a drop in replacement.

Is there a way to make the `private[testkit]` implementations delegate to the new implementations in a way that is semantically correct in the new model? For example, do the `tick()` then `advanceAndTick()` trick or something.

If the answer is no, maybe it would indeed be better to not guarantee binary compatibility at all and just advise people to upgrade, recompile, fix source changes and change to the new semantics at the same time. Of course, you're much more experienced than me in this regard, so I won't fight any final decision.

kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/TestContext.scala
djspiewak: Yep

tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
djspiewak: This is the case that is documented in the `tick` scaladoc. :-)

kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/TestContext.scala
djspiewak: IMO if this ends up inside a real executor, we're in some serious trouble. :-P It doesn't hurt to add it though

djspiewak:
Noodled a bit on some Specs2 support built on top of this API: typelevel/cats-effect-testing#193

```scala
"execute a simple test with mock time" in {
  var first = false
  var second = false
  var third = false
  val wait = IO.sleep(1.hour)

  val program = for {
    _ <- IO { first = true }
    _ <- wait
    _ <- IO { second = true }
    _ <- wait
    _ <- IO { third = true }
  } yield 42

  program must execute { (control, result) =>
    result() must beNone

    first must beFalse
    second must beFalse
    third must beFalse

    control.tick()

    first must beTrue
    second must beFalse
    third must beFalse

    control.advanceAndTick(1.hour)

    first must beTrue
    second must beTrue
    third must beFalse

    control.advanceAndTick(1.hour)

    first must beTrue
    second must beTrue
    third must beTrue

    result() must beSome(beRight(42))
  }
}
```

kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/TestContext.scala
```diff
@@ -285,18 +314,42 @@ final class TestContext private () extends ExecutionContext { self =>
...
+   * tasks are interleaved. Note that subtle differences between
+   * different runs of identical programs are possible,
+   * particularly if one program auto-`cede`s in a different place
+   * than the other one. This is an excellent and reliable mechanism
+   * for small, tightly-controlled programs with entirely deterministic
+   * side-effects, and a completely useless mechanism for anything
+   * where the scheduler ticks see different task lists despite
+   * identical configuration.
+   */
+  def apply(seed: String): TestContext =
```
alexandru, line 270: Inclusion of a seed is a very good idea 👍

testkit/shared/src/main/scala/cats/effect/testkit/TestControl.scala
```diff
@@ -0,0 +1,260 @@
...
+ *
+ * Note that the above means that it is relatively easy to create a deadlock
+ * on this runtime with a program which would not deadlock on either the JVM
+ * or JavaScript:
+ *
+ * {{{
+ *   // do not do this!
+ *   IO.cede.foreverM.timeout(10.millis)
+ * }}}
+ *
```
SystemFw, line 76: What happens with a program that waits on a deferred with a timeout? Should work right? as that won't end on the task queue until the deferred is complete

testkit/shared/src/main/scala/cats/effect/testkit/TestControl.scala
```diff
@@ -0,0 +1,260 @@
...
+ * use the `tickAll` method, the clock will be automatically advanced by the
+ * minimum amount necessary to reach the next pending task. For example, if
+ * the program contains an [[IO.sleep]] for `500.millis`, and there are no
+ * shorter sleeps, then time will need to be advanced by 500 milliseconds in
+ * order to make that fiber eligible for execution.
+ *
+ * At this point, the process repeats until all tasks are exhausted. If the
+ * program has reached a concluding value or exception, then it will be
+ * produced from the `unsafeRun` method which scheduled the `IO` on the
+ * runtime (pro tip: do ''not'' use `unsafeRunSync` with this runtime, since
```
SystemFw: we need to make sure that the testing integrations do the right thing here

testkit/shared/src/main/scala/cats/effect/testkit/TestControl.scala
```diff
@@ -0,0 +1,260 @@
...
+ * top of these). From the perspective of these functions, all computation is
+ * infinitely fast, and the only effect which advances time is [[IO.sleep]]
+ * (or if something external, such as the test harness, calls the [[advance]]
+ * method). However, an effect such as `IO(System.currentTimeMillis())` will
+ * "see through" the illusion, since the system clock is unaffected by this
+ * runtime. This is one reason why it is important to always and exclusively
+ * rely on `realTime` and `monotonic`, either directly on `IO` or via the
+ * typeclass abstractions.
+ *
+ * WARNING: ''Never'' use this runtime on programs which use the [[IO#evalOn]]
```
SystemFw: I think it's worth adding a sentence about behaviour of `blocking`

testkit/shared/src/main/scala/cats/effect/testkit/TestControl.scala
```diff
@@ -0,0 +1,260 @@
...
+   * ''before'' the [[IO.sleep]] even has a chance to get scheduled!
+   * This means that when `sleep` is finally submitted to the runtime,
+   * it is scheduled for the time offset equal to `1.second + 100.millis`,
+   * since time was already advanced `1.second` before it had a chance
+   * to submit. Of course, time has only been advanced by `1.second`,
+   * thus the `sleep` never completes and the `println` cannot ever run.
+   *
+   * There are two possible solutions to this problem: either call [[tick]]
+   * ''first'' (before calling `advanceAndTick`) to ensure that the `sleep`
+   * has a chance to schedule itself, or simply use [[tickAll]] if you
```
SystemFw: Mm this is tricky. I wonder if this doesn't point out to a best practice of writing the whole test without asserting, i.e. "recording" assertions in the main fiber with `sleep`  while the test program runs in another fiber, and finally using `tickAll`, as opposed to using `advanceAndTick` and asserting

kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/TestContext.scala
```diff
@@ -122,13 +127,14 @@ import scala.util.control.NonFatal
  *   assert(f.value == Some(Success(2))
  * }}}
  */
-final class TestContext private () extends ExecutionContext { self =>
-  import TestContext.{State, Task}
+final class TestContext private (_seed: Long) extends ExecutionContext { self =>
```
SystemFw, line 62: is this intended for end users now, or only `TestControl`?

testkit/shared/src/main/scala/cats/effect/testkit/TestControl.scala
SystemFw: Or more simply that you should always `tick` first

testkit/shared/src/main/scala/cats/effect/testkit/TestControl.scala
djspiewak: Correct!

testkit/shared/src/main/scala/cats/effect/testkit/TestControl.scala
djspiewak: Yep. Special and explicit integrations are needed in munit-cats-effect cats-effect-testing, and Weaver.

testkit/shared/src/main/scala/cats/effect/testkit/TestControl.scala
djspiewak: Good call. To more directly answer, it behaves like JavaScript: the blocking pool and the compute pool are both TestContext.

testkit/shared/src/main/scala/cats/effect/testkit/TestControl.scala
djspiewak: Either one is definitely fine. Honestly I think the answer probably depends on what you're testing. For some things it may be difficult to get "in the middle" of some time-delayed flow to inject assertions.

kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/TestContext.scala
djspiewak: Really both. It's still exposed to downstream users who need it. For example, it's very useful for anyone who wants to evaluate a data type against the CE laws. The main end user cases though are wrapped up in TestControl.

SystemFw: I think some more docs are needed then, I fear a lot of users will stumble upon it but only few will need it

djspiewak: Agreed #1210

djspiewak:
Bumping this a bit, @SystemFw. Any follow-ups needed here? Should we look at the munit integration first?

SystemFw:
Yeah munit integration would be good, and tbh I think it'd be great to have a snapshot to try it out. I'm happy with it though, I hadn't realised that wasn't clear

djspiewak:
Actually published a snapshot quite a while ago. 😃 That's what the Specs2 integration is based on. I'll dig into munit.

djspiewak:
New snapshot published as `3.3-115-fab668b`

djspiewak:
Odd failure:

```
[info] IO monad should
2869
[error]   ! evaluate fibers correctly in presence of a parasitic execution context
2870
[error]    java.util.concurrent.TimeoutException: null (Runners.scala:109)
2871
2872
[info] Total for specification ParasiticECSpec
```

vasilmkd:
Might not be an odd failure. The test uses `ticker` to evaluate equality between generated `IO` values.

djspiewak:
Yeah I'm definitely starting to suspect this is a legitimate failure. It's weird though. I can't reproduce it locally.

SystemFw:
> I'll dig into munit.

btw not necessarily how I want it to be, but for one of the use cases where you don't need to step, this is what I have in fs2:

Setup:
https://github.com/typelevel/fs2/blob/main/core/shared/src/test/scala/fs2/Fs2Suite.scala#L49-L93

Example usage:

https://github.com/typelevel/fs2/blob/2af0571de34cc11b6b2bafeef0baf487f0760ad8/core/shared/src/test/scala/fs2/TimedPullsSuite.scala#L56-L80

It doesn't work with scalacheck-effect atm

djspiewak:
@SystemFw What about this?

```scala
test("execute a simple test with mock time") {
  var first = false
  var second = false
  var third = false
  val wait = IO.sleep(1.hour)

  val program = for {
    _ <- IO { first = true }
    _ <- wait
    _ <- IO { second = true }
    _ <- wait
    _ <- IO { third = true }
  } yield 42

  program execute { (control, result) =>
    assert(result() == None)

    assert(!first)
    assert(!second)
    assert(!third)

    control.tick()

    assert(first)
    assert(!second)
    assert(!third)

    control.advanceAndTick(1.hour)

    assert(first)
    assert(second)
    assert(!third)

    control.advanceAndTick(1.hour)

    assert(first)
    assert(second)
    assert(third)

    assert(result() == Some(Right(42)))
  }
}
```

Btw, I think it might be also useful to define `executeFully` or something like that which just calls `control.tickAll()` and returns `result()`.

SystemFw:
Yeah,`executeFully` is what I have in the links I sent you. Your snippet for the stepthrough case looks great to me :)

djspiewak:
@SystemFw I'm writing some stuff to add to this PR which actually should cover MUnit nicely. I think the Specs2 specialness would remain, and maybe other test frameworks with a similar philosophy.

djspiewak:
@SystemFw Take a look and lmk what you think. You can get infix syntax by `import cats.effect.testkit.syntax.all._`. This import is unfortunate, but without doing some sort of orphan type trick, we don't have a good way of tucking it into the `IO` companion. I'm not *entirely* averse to doing the trick, so long as we're sure it won't cause any problems downstream.

testkit/shared/src/main/scala/cats/effect/testkit/syntax/TestControlSyntax.scala
```diff
@@ -0,0 +1,38 @@
...
+final class TestControlOps[A] private[syntax] (private val wrapped: IO[A]) extends AnyVal {
+
+  def execute[B](config: IORuntimeConfig = IORuntimeConfig(), seed: Option[String] = None)(
+      body: (TestControl, () => Option[Either[Throwable, A]]) => B): B =
+    TestControl.execute(wrapped, config = config, seed = seed)(body)
+
+  def executeFully(
+      config: IORuntimeConfig = IORuntimeConfig(),
+      seed: Option[String] = None): Option[Either[Throwable, A]] =
+    TestControl.executeFully(wrapped, config = config, seed = seed)
```
djspiewak: Bikeshed this API!

SystemFw:
> This import is unfortunate, but without doing some sort of orphan type trick, we don't have a good way of tucking it into the IO companion

We could add the relevant implicits to one of the traits in munit-cats-effect, although strictly speaking the result of `execute` and `executeFully` is not `IO`

danicheg:
> > This import is unfortunate, but without doing some sort of orphan type trick, we don't have a good way of tucking it into the IO companion
> 
> We could add the relevant implicits to one of the traits in munit-cats-effect, although strictly speaking the result of `execute` and `executeFully` is not `IO`

Is it make sense to create another `Suite` that would use the `TestControl` semantic for `IO` execution via `ValueTransform`?

djspiewak:
> Is it make sense to create another Suite that would use the TestControl semantic for IO execution via ValueTransform?

I don't know what those words mean. Hopefully someone who knows munit will answer. :-P

If it's helpful, I've been thinking of `TestControl` as more around *assertions* than around *suites*. So for example, it might make sense to write a test which calls `execute` more than once on different `IO`s (though, stylistically, it's probably better to just have separate tests). This is very different than the rest of the Cats Effect support, which is more about making it possible to write tests *in* `IO`.

djspiewak:
Published a new snapshot as `3.3-138-c458766`

testkit/shared/src/main/scala/cats/effect/testkit/syntax/TestControlSyntax.scala
```diff
@@ -0,0 +1,38 @@
...
+import cats.effect.unsafe.IORuntimeConfig
+
+trait TestControlSyntax {
+  implicit def testControlOps[A](wrapped: IO[A]): TestControlOps[A] =
+    new TestControlOps(wrapped)
+}
+
+final class TestControlOps[A] private[syntax] (private val wrapped: IO[A]) extends AnyVal {
+
+  def execute[B](config: IORuntimeConfig = IORuntimeConfig(), seed: Option[String] = None)(
```
Baccata: `<bikeshedding>`

I don't like this particular extension method. The lexical-scope, and unsafe nature of what happens in it, is communicated in a much clearer  fashion when using `TestControl.execute(io){ ... }`, whereas making it look like a method on "IO" makes it look like some sort of combinator. 

The `unsafe` methods on `IO` are okay, because the prefix makes it really clear that you're leaving the safe zone, it is obvious that they are not combinators. The lack of `unsafe` prefix on the `execute` extension methods breaks this intuition. 

If I was benevolent dictator, I'd remove the `execute` extension method to push for a clearer, resource-like use of `TestControl` and would rename the `executeFully` to `unsafeExecuteFully`. 

`</bikeshedding>`

testkit/shared/src/main/scala/cats/effect/testkit/TestControl.scala
```diff
@@ -0,0 +1,292 @@
...
+ * runtime will detect this situation as an asynchronous deadlock.
+ *
+ * @see
+ *   [[cats.effect.unsafe.IORuntime]]
+ * @see
+ *   [[cats.effect.kernel.Clock]]
+ * @see
+ *   [[tickAll]]
+ */
+final class TestControl private (config: IORuntimeConfig, _seed: Option[String]) {
```
Baccata: I'm thinking about how I'd support it in weaver ... not that it really "needs" supporting, because I'm pretty sure it would just work as is, but I'm thinking of how I'd like people to use it in `weaver`. 

In particular, the paradigm that Anton and I are going for is *USE IO GODDAMNIT !!!*, and enabling people to use a `IO => A` methods breaks that paradigm. 

So my question is : would it be it safe for weaver to encapsulate this (or recreate it) in a construct that'd expose advance/tick methods wrapped in`IO` ? I'm pretty sure it would, but  need to be sure.

SystemFw:
The conversation is scattered all over the place 😄 
Anyway:

>  that would use the TestControl semantic for IO execution via ValueTransform?

This is what the fs2 snippet does. I would prefer it, and it's certainly doable for `executeFully`, but a bit less clear for `execute`

> my question is : would it be it safe for weaver to encapsulate this (or recreate it) in a construct that'd expose advance/tick methods wrapped inIO ? I'm pretty sure it would, but need to be sure.

second this, I think munit-cats-effect would do the same thing, and there's even a question about whether CE itself should do it this way.

It's true that in munit the methods would work as they are, but it makes things less clear, given that other methods in CatsEffectSuite return IOs, and that you would have to use `forall` instead of `forallF` etc

testkit/shared/src/main/scala/cats/effect/testkit/syntax/TestControlSyntax.scala
djspiewak: > If I was benevolent dictator, I'd remove the execute extension method to push for a clearer, resource-like use of TestControl and would rename the executeFully to unsafeExecuteFully.

Fortunately the BDFL owes you a favor or two…

djspiewak:
@Baccata and @SystemFw raise an interesting point, which broadly boils down to "`TestControl` is an effectful thing and should be in `IO`". I thought about this! Honestly I could go either way. My *concern* is it makes the model, which is already complex, even harder to grok. Actually I have two concerns, but first thing's first.

The problem here is the fact that there would end up being an "inner" `IO` and an "outer" `IO`, each running in different runtimes with very different semantics. Values destined for one runtime ending up in the other runtime would be pretty bad. As an example, I give a gold star to anyone who can correctly describe the semantics of the following (assuming an `IO`-wrapped version of the `TestControl` API as it stands):

```scala
for {
  latch <- IO.deferred[Unit]
  _ <- TestControl.execute(latch.get) { (control, results) =>
    control.tick *> latch.set(())
  }
} yield ()
```

Now imagine this is more complex, and instead of just testing the `latch.get` program, we're actually testing something much more complicated. Are we really going to be able to explain those semantics to users?

The counterargument here is that we may, *within* the bounds of the `execute` scope, want to call some application functions that manipulate internal state which affects the program. Those functions will *likely* be in `IO` (if the user is doing what we want them to do 🙂), and if `execute` is pretending to be pure there will be no way to sequence those methods other than doing `executeFully` or something, which sounds tricky.

My second concern about all of this, which I think I have a solution to, is that `tick()` may block, and so running it within an outer `IO` is a bit scary. I'm pretty sure I can fix that though…

As an aside, I'm open to a `Resource`-like API here. The main reason I *didn't* go that route is I really like the idea of marrying the `control` and the `results` together. At least to me, it feels much more intuitive to think of `TestControl` as a sort of handle into the runtime of a specific actively running `IO`. The API sort of encodes that metaphor. It is doable in a resourcy way as well.

Also if we really want to lean into that metaphor, maybe `TestControl` shouldn't expose `runtime` *at all* and just, *itself*, carry the `results`?

djspiewak:
Oh forgot to reply to one point…

> This is what the fs2 snippet does. I would prefer it, and it's certainly doable for executeFully, but a bit less clear for execute

The fact that it's not doable for `execute` is why I shied away from this concept. :-) I think it's important that we make very front-and-center the fact that this is a different sort of runtime, with different rules, operating in a different way, and there are benefits to that which you may or may not leverage. Making it automagically behave exactly like a normal `IO`, but only in one of the two major scenarios, runs a little bit counter to this.

Again, my main concern here is comprehensibility, particularly since this is to be used for testing. This is a *very* subtle bit of machinery, and I want people to understand it enough to use it with confidence and correctness in their tests.

djspiewak:
@SystemFw @Baccata @ChristopherDavenport Do you like this API better? https://gist.github.com/bfed5b1a99f05200314643f466830382 Ignore the scaladoc, it's all outdated now.

danicheg:
> @SystemFw @Baccata @ChristopherDavenport Do you like this API better? https://gist.github.com/bfed5b1a99f05200314643f466830382 Ignore the scaladoc, it's all outdated now.

`IO[Option[Either[Throwable, A]]]` 🤯

vasilmkd:
Can the resulting type `Option[Either[Throwable, A]]` be an `Outcome[Id, Throwable, A]`?

djspiewak:
> Can the resulting type Option[Either[Throwable, A]] be an Outcome[Id, Throwable, A]?

It can be, and that's obviously valid, but I thought that `Option[Either[` might be a little more ergonomic to work with. I'm up for changing it if you or others disagree.

djspiewak:
Okay I think I kinda like the pure API better. Pushed a more complete implementation. Lmk what you think. I didn't change `Option[Either[` to `Outcome` but I'm open to that as well.

djspiewak:
Okay I think I kinda like the pure API better. Pushed a more complete implementation. Lmk what you think. I didn't change `Option[Either[` to `Outcome` but I'm open to that as well.

Baccata:
I do like the pure API, and think I would have provided it in weaver myself, I appreciate not having to do it 😄 

I think an Outcome version would indeed communicate what the API does better. Not too bothered though.

djspiewak:
One quick realization: the `None` case here corresponds to both `Outcome.Canceled` and "did not complete", which is not something `Outcome` represents. I actually think that the `Outcome`-based API would be *less* explicit, not more, particularly since I think cancelation is actually going to show up as a `Left(_: CancellationException)`.

vasilmkd:
Is `None` mapping to both `Canceled` and "did not complete" also the case for the matchers that we currently have for `ticked`?

djspiewak:
> Is None mapping to both Canceled and "did not complete" also the case for the matchers that we currently have for ticked?

It was. I think @wemrysi split them apart, which is why we now have four matchers in `ticked`. The representation here is just a little different and more in line with how end-users usually see the same effects coming out of the `unsafeRun` family.

vasilmkd:
Ok, this does tell me that shoehorning this API into `Outcome` is not the way to go. However, I'm not exactly convinced that canceled and not completed should map to the same return value. On the other hand, I also don't have a good alternative solution at the moment, other than a bespoke 4 case `TestControlOutcome` enum or something.

djspiewak:
> I'm not exactly convinced that canceled and not completed should map to the same return value

Thinking about it (and not, you know, trying it), I'm pretty sure they don't. In fact, I think that cancelation maps to `Some(Left(_: CancellationException))`, which mirrors what users would see if they ran `unsafeRunSync()`.

djspiewak:
New snapshot (including the pure API) published as `3.3-141-b15d57b`

vasilmkd:
> Thinking about it (and not, you know, trying it), I'm pretty sure they don't. In fact, I think that cancelation maps to Some(Left(_: CancellationException)), which mirrors what users would see if they ran unsafeRunSync().

And this is exactly the situation that you described as suboptimal here:
> One quick realization: the None case here corresponds to both Outcome.Canceled and "did not complete", which is not something Outcome represents. I actually think that the Outcome-based API would be less explicit, not more, particularly since I think cancelation is actually going to show up as a Left(_: CancellationException).

I'm playing devil's advocate here, I do not have a preferred API, I want to help you arrive at one.

djspiewak:
Yeah I think I was just wrong in the first half of that comment. 😛 I'll try it though!

djspiewak:
```
scala> TestControl.executeFully(IO.canceled)
val res0: cats.effect.IO[Option[Either[Throwable,Unit]]] = IO(...)

scala> res0.unsafeRunSync()
val res1: Option[Either[Throwable,Unit]] = Some(Left(java.util.concurrent.CancellationException: The fiber was canceled))
```

I think we're good!

tests/jvm/src/test/scala/cats/effect/ParasiticECSpec.scala
```diff
@@ -28,10 +28,12 @@ class ParasiticECSpec extends BaseSpec with TestInstances {
...
     "evaluate fibers correctly in presence of a parasitic execution context" in real {
-      implicit val ticker = Ticker()
+      val test = {
+        implicit val ticker = Ticker()
 
-      val test = IO(implicitly[Arbitrary[IO[Int]]].arbitrary.sample.get).flatMap { io =>
-        IO.delay(io.eqv(io))
+        IO(implicitly[Arbitrary[IO[Int]]].arbitrary.sample.get).flatMap { io =>
+          IO.delay(io.eqv(io))
+        }
```
vasilmkd, line 36: I'm pretty sure this change can be reverted.

vasilmkd (commented):
I am amazed by this PR and the scaladoc is very informative. Thank you for your effort. 🙏🏻 

One question still remains, do people think we need to build a higher level ADT for making the different outcomes more explicit. I'm personally worried about `Left(CancelationException)` not being great from an ergonomics point of view.

Baccata:
> One question still remains, do people think we need to build a higher level ADT for making the different outcomes more explicit. I'm personally worried about `Left(CancelationException)` not being great from an ergonomics point of view.

I agree. Even `Option[Outcome[Id, Throwable, A]]` would be more consistent with the rest of CE

vasilmkd:
I don't hate that. I'm not sure if Daniel considers it a solution if we only do a last-mile wrapping of the `Option[Either[Throwable, A]]` into a `Option[Outcome[Id, Throwable, A]]`.

Baccata:
As long as cancellation is not surfaced as a cancellation exception, I'm happy 😉

vasilmkd:
Most definitely, I'm really not comfortable with that.

djspiewak:
Okay okay. 😛 I'll try the Option if Outcome approach…

SystemFw:
Well, remember that the UX here is writing assertions, and I think Outcome might actually be the worst UX, given that writing assertions that involve pattern matching is generally a bit annoying. In some ways the most immediate return type is actually `IO[Option[A]]` (and `CancelationException` plus other exceptions getting thrown), which integrates with every test framework and favours the most common case


P.S. I'm integrating this into `upperbound` and the experience is otherwise great, especially the fact that you can just call `TestControl.executeFully` with no other changes to munit-cats-effect nor a specific test setup beyond basic IO suites

djspiewak:
See [Discord](https://discord.com/channels/632277896739946517/839263556754472990/886672279327617134) for a long diatribe on why we replaced `executeFully` with `executeEmbed`. :-)

djspiewak:
Published as `3.3-161-09d830c`

SystemFw:
Ok, so I've tried `executeEmbed` 

TL;DR: it's much better, but we should remove the Option from there as well.

Note that I'm only talking `executeEmbed` here, not stepping through. Here are my arguments (some of them might be repeated from Discord):

- Most tests will only care about the actual results, which makes dealing with an Option pointless boilerplate.
- If you are calling `executeEmbed`, you are running the IO to completion, and nontermination is an exceptional outcome, and should therefore be represented as an exception. Note that this isn't the case with `results` inside `execute`, since you are legitimately sampling there.
- It's ok for the resulting Exception to be defined inside TestControl, e.g. as `TestControl.NonTerminationException`. This may appear bespoke, but being able to materialise nontermination (currently as an `Option`) _is_ bespoke to TestControl, so a bespoke exception actually seems very appropriate.
- **`Option` is bad UX on failed tests** . This might actually be the strongest argument. With `Option`, a non terminating program will fail as either `The result was None` or `NoSuchElementException`. Compare to a proper exception: `TestControl.NonTerminationException: the program you have run through TestControl never completed, which might indicate deadlock etc, see these docs etc.`

Hence, having `executeEmbed: F[A]` makes successful tests easier to write, and failed tests easier to understand, and should be preferred to the current API.

djspiewak:
> If you are calling executeEmbed, you are running the IO to completion, and nontermination is an exceptional outcome, and should therefore be represented as an exception. Note that this isn't the case with results inside execute, since you are legitimately sampling there.

I was thinking about it over the past couple days and came round to a similar opinion. I think this is probably the most compelling argument. Also I agree with the counterpoint on "bespokeness".

> Option is bad UX on failed tests . This might actually be the strongest argument. With Option, a non terminating program will fail as either The result was None or NoSuchElementException. Compare to a proper exception: TestControl.NonTerminationException: the program you have run through TestControl never completed, which might indicate deadlock etc, see these docs etc.

I think it gets even worse than that if someone writes their assertions *within* the `IO`, but then that `IO` deadlocks before a failed assertion gets hit, and then they discard the `Option` (e.g. by just returning the results of `executeEmbed` from `test`). In that scenario, there's no failure, but the test is definitely failing. Arguably that's a test where someone failed to account for every possible failure scenario (in this case, nontermination), but the UX should guide them towards checking for such.

So overall I think I'm in alignment. I'll make the change.

djspiewak:
Changes released in `3.3-162-2022ef9`

djspiewak:
Final thoughts @SystemFw or others?

SystemFw:
Really excited this is pretty much done :)
I'm happy with the api, are we set on names? I was thinking about `run` and `stepThrough`, but I wouldn't be mad if we kept the current names at all

djspiewak:
I actually really like the names. Honestly, `run` strikes me as a bit too innocuous. 🙂 Given the weirdness it's performing, I want people to be aware that they're flirting with dragons. Also the documentation pretty consistently refers to "execute" as the verb when running an IO (despite the fact that the production prefix is `unsafeRun`), so it fits pretty well.

The "Embed" suffix also works really well when comparing with the `Fiber` api, which is not an incorrect comparison. It also nudges people a bit to consider the stepping API first, which is good IMO even if it isn't the most commonly optimal function, simply because it invites people to understand a bit better what's going on under the surface.

I'm bikeshedding. 😃 Names matter but probably not as much as I'm implicitly asserting here. I just rather like the current ones.

SystemFw:
>  I just rather like the current ones.

That's ok, let's go ahead with the current ones :)

SystemFw (approved):

bastewart:
Useless comment: just wanted to say I really like this change (i've "borrowed" it temporarily, pending a release of 3.3)! It's a huge improvement over the hand-rolling required in the past to use `TestContext`. Very unrelated as well but I think it's a big validation of the design of CE3 that `TestControl` just exposes a function `IO ~> IO` and end-users don't have to do anything like overwriting the `ContextShift` and `Timer`.

yanns:
This is a really great addition!!! 🥇

PR #2283, opened 2021-08-31
Work stealing runtime queues tuning

vasilmkd:
The initial commits in this PR are a preparation for the main intended change, which is to halve the local queue on overflow, but instead of 1 big batch of 128 fibers, create 4 batches of 32 fibers. At another time, when the batched queue is checked for fibers, it won't cause spillage of the local queue because only a smaller batch of fibers will be enqueued, instead of half of the queue being replaced, as was the case previously.

### DIFF
--- a/core/jvm/src/main/java/cats/effect/unsafe/LocalQueueConstants.java
+++ b/core/jvm/src/main/java/cats/effect/unsafe/LocalQueueConstants.java
@@ -43,7 +43,8 @@ class LocalQueueConstants {
   static final int HalfLocalQueueCapacity = LocalQueueCapacity / 2;
 
   /**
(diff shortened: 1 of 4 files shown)


vasilmkd:
@djspiewak I finalized the changes as a standalone package here. The PR can be merged as is. The fairness towards the batched queue is already improved, we alternate between the external and the batched queue. We did not lose performance and I think that we actually gained something, because adding batches to the local queue is now less disruptive.

I will continue the exploration for unifying the external and batched queues and allowing the application load to dynamically dictate the fairness towards the overflow and batched queue in a separate PR.

djspiewak (approved):

Issue #2244, opened 2021-08-18
`timeout` eagerly creates `TimeoutException`


djspiewak:
This imposes a relatively significant cost (due to the stack trace construction) despite the common case where the timeout is *not* hit and thus the exception is not needed.

ixthe0:
hey @djspiewak, is it OK if I work on this issue?

vasilmkd:
Go for it. Feel free to ask for pointers here or in the Discord.

PR #2294, opened 2021-09-02
Prevent eager creation of TimeoutException in IO.timeout

ixthe0:
This PR addresses #2244 by delaying the construction of a `TimeoutException` until it is really needed (or never)

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -585,7 +585,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
    *        the source completing, a `TimeoutException` is raised
    */
   def timeout[A2 >: A](duration: FiniteDuration): IO[A2] =
-    timeoutTo(duration, IO.pure(duration).flatMap(d => IO.raiseError(new TimeoutException(d.toString))))
+    timeoutTo(duration, IO.defer(IO.raiseError(new TimeoutException(duration.toString))))
 
   /**
    * Returns an IO that either completes with the result of the source within
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -585,7 +585,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
    *        the source completing, a `TimeoutException` is raised
    */
   def timeout[A2 >: A](duration: FiniteDuration): IO[A2] =
-    timeoutTo(duration, IO.raiseError(new TimeoutException(duration.toString)))
+    timeoutTo(duration, IO.pure(duration).flatMap(d => IO.raiseError(new TimeoutException(d.toString))))
 
   /**
    * Returns an IO that either completes with the result of the source within
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -1441,6 +1441,11 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
     override def handleError[A](fa: IO[A])(f: Throwable => A): IO[A] =
       fa.handleError(f)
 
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -585,7 +585,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
    *        the source completing, a `TimeoutException` is raised
    */
   def timeout[A2 >: A](duration: FiniteDuration): IO[A2] =
-    timeoutTo(duration, IO.raiseError(new TimeoutException(duration.toString)))
+    timeoutTo(duration, IO.pure(duration).flatMap(d => IO.raiseError(new TimeoutException(d.toString))))
```
vasilmkd: I think we can save the `IO.pure` allocation here by doing the following:
```scala
IO.defer(IO.raiseError(new TimeoutException(duration.toString)))
```

vasilmkd (changes requested):

core/shared/src/main/scala/cats/effect/IO.scala
ixthe0: @vasilmkd thank you!

Is there a reason to prefer `IO.defer` over `IO.pure`? They seem to introduce a similar level of memory overhead (allocation of `Delay` vs allocation of `Pure`), but `Delay` also includes `TracingEvent`

vasilmkd: Meh, I forgot about the trace. We should still expect the trace to be cached. Otherwise not a huge reason to prefer `defer` other than the fact that it conveys the intent better than `pure` and `flatMap`.

djspiewak: `IO.defer` is a more declaratively correct thing to do. :-)

djspiewak (changes requested):
This looks great! Can you also override `timeout` in the `Async` instance below, delegating to this implementation? That should ensure the semantics are consistent when people use `IO` via the typeclasses.

vasilmkd:
While we're at it, should we attempt to change `GenTemporal#timeout`? That one would need the `IO.unit.flatMap` trick because there's no `defer`.

djspiewak:
It would definitely be ideal yes. Fixing it in IO doesn't fix it for derived transformers.

Also it's actually fixable without the trick, it just requires a redundant implementation of the timeout race.

core/shared/src/main/scala/cats/effect/IO.scala
ixthe0: @vasilmkd @djspiewak makes sense, guys, thank you!

ixthe0:
@vasilmkd correct me if I'm wrong, but the implementation you are referring to would look like so:
```
def timeout[A](fa: F[A], duration: FiniteDuration)(implicit ev: TimeoutException <:< E): F[A] = {
    timeoutTo(fa, duration, unit.flatMap(_ => raiseError[A](ev(new TimeoutException(duration.toString))))
}
```

@djspiewak I am not sure that I caught your idea correctly, but here's how I got it:
```
def timeout[A](fa: F[A], duration: FiniteDuration)(implicit ev: TimeoutException <:< E): F[A] = {
    flatMap(race(fa, sleep(duration))) {
        case Left(a)  => pure(a)
        case Right(_) => raiseError[A](ev(new TimeoutException(duration.toString)))
    }
}
```

I apologize in advance for beginner questions, I still need to get a lot of development context of CE3

djspiewak:
Looks right to me!

vasilmkd:
@ixthe0 I apologize for not giving this PR my full attention, I was really busy with more pressing stuff these days. Running the CI now and will review the code in a bit.

vasilmkd (approved):
Thank you @ixthe0! Great work.

ixthe0:
@vasilmkd @djspiewak couldn't have done it without your help guys

djspiewak (approved):

PR #2298, opened 2021-09-03
PQueue respects FIFO for elements at the same priority

SystemFw:
Currently elements at the same priority get dequeued in LIFO order, which is very weird for a queue. This PR fixes that.

I haven't delved too deeply into BinomialHeap so please double check the logic, worst case scenario there is an implementation strategy that only touches PQueue: keep track of the latest insertion in a Long and use a `Box[A: Order](a: A, insertedAt: Long)`, using `insertedAt` in `Order[Box[A]]` to break ties.

Targeting 3.2.x because I need this.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/internal/BinomialHeap.scala
+++ b/std/shared/src/main/scala/cats/effect/std/internal/BinomialHeap.scala
@@ -139,7 +139,7 @@ private[std] final case class BinomialTree[A](
    * Link two trees of rank r to produce a tree of rank r + 1
    */
   def link(other: BinomialTree[A])(implicit Ord: Order[A]): BinomialTree[A] = {
-    if (Ord.lteqv(value, other.value))
+    if (Ord.lt(value, other.value))
       BinomialTree(rank + 1, value, other :: children)
     else BinomialTree(rank + 1, other.value, this :: other.children)
   }
--- a/tests/shared/src/test/scala/cats/effect/std/PQueueSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/std/PQueueSpec.scala
@@ -120,6 +120,22 @@ class UnboundedPQueueSpec extends BaseSpec with PQueueTests {
     unboundedPQueueTests(
       "UnboundedPQueue mapK",
       PQueue.unbounded[IO, Int].map(_.mapK(FunctionK.id)))
+
+    "degenerate to a normal queue if all priorities are the same" in real {
+      case class E(name: String, index: Int)
+      object E {
+        implicit val orderForE: Order[E] = Order.by(_.index)
+      }
+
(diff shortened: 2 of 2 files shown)


djspiewak (commented):
This _seems_ very reasonable to me but I would like @TimWSpence to weigh in if possible.

SystemFw:
> but I would like @TimWSpence to weigh in if possible.

Yeah, GitHub didn't let me add him to the reviewers

TimWSpence:
Yeah, this looks good to me 👍 The rank and heap properties of `link` are still preserved and then we should pick the order-equal elements up in the reverse order in `take`.

I'm glad I documented all the invariants in the comments 😅

SystemFw:
@djspiewak I need an approval, then I think we can release?

djspiewak (approved):
Looks good to me! Will merge. Holding the release pending the other two issues raised in discord.

SystemFw:
@djspiewak  @TimWSpence 

Unfortunately I've found a case where the PQueue doesn't work 😦 .
Thankfully I had property tests in `upperbound` (really should have done the same in the PR, apologies).
Basically when you have two equal elements and a then a different one, the whole thing gets reversed:

```scala
def p2 = {
    case class E(name: String, index: Int)
    object E {
      implicit val orderForE: Order[E] = Order.by(_.index)
    }

    val results = List(E("a", 0), E("a", 0), E("b", 0))

    PQueue
      .unbounded[IO, E]
      .flatMap { q =>
        results.traverse(q.offer) >>
          q.take.replicateA(results.length).map { out =>
            println(out)
            out == results
          }
      }
      .unsafeRunSync()

    scala> Q.p2
    List(E(b,0), E(a,0), E(a,0))
    val res0: Boolean = false
  }
```

interestingly, StackOverflow seems to indicate that Java/Python/C++ all make priority queue behave as the old PQueue in CE, i.e. don't respect FIFO unless you add an additional shared counter and embed that in the ordering.

So, alternatives:
- Keep fiddling with BinomialHeap. Given the weird nature of the case that broke it, I'm a bit reticent tbh
- Fix the issue in PQueue, by adding a counter to the state and wrapping each element in Box[A: Order](a:  A, insertedAt: Long) to break ties with FIFO
- Revert PQueue to what it was before (like other langs, but it's a gotcha there too), and if you want fifo you need an additional Ref + strategy 2

Thoughts?

djspiewak:
IMO option three is my preferred. Otherwise everyone pays the penalty of the tie breaker overhead, even when people don't need it. The reversion scenario honestly makes sense and I should have thought of it sooner. I'm not sure there's a way to encode a heap which is guaranteed to be stable with respect to prior insertions without significant overhead.

SystemFw:
Let's go with three then. I've already coded it up in `upperbound` using the old PQueue anyway

PR #2305, opened 2021-09-04
Fix helper threads giving up before more work is available

### LINKED ISSUE #2304: `HelperThread` gives up if there's no immediate work available on the pool
Honestly, I was lazy. It's a legitimate bug that I caused because I cut some corners. I need to beef up the HelperThread mechanism to sleep while waiting for new work to arrive, work which might be late, but will eventually unblock the worker threads of the pool.

Identified in https://github.com/typelevel/munit-cats-effect/pull/124. There's a quick and dirty way to avoid the deadlock, which is to let the helper threads busy wait until the fiber that unblocks the worker threads arrives. Obviously I don't need to overstate how terrible that would be.

vasilmkd:
This will eventually fix #2304. For now I have a reproduction as a unit test.

### DIFF
--- a/tests/jvm/src/test/scala/cats/effect/unsafe/HelperThreadParkSpec.scala
+++ b/tests/jvm/src/test/scala/cats/effect/unsafe/HelperThreadParkSpec.scala
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2020-2021 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+package unsafe
+
+import cats.syntax.all._
+
+import scala.concurrent.{Await, Promise}
+import scala.concurrent.duration._
+
+class HelperThreadParkSpec extends BaseSpec {
+
+  "HelperThread" should {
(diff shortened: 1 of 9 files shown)


djspiewak:
Should we target this at series/3.2.x?

tests/jvm/src/test/scala/cats/effect/unsafe/HelperThreadParkSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+      Resource.make(IO(smallRuntime()))(rt => IO(rt.shutdown())).use { rt =>
+        val io = for {
+          p <- IO(Promise[Unit]())
+          _ <- (IO.sleep(50.millis) *> IO(
+            rt.scheduler.sleep(100.millis, () => p.success(())))).start
+          _ <- IO(Await.result(p.future, Duration.Inf))
+        } yield ()
+
+        List
+          .fill(10)(io.start)
```
djspiewak, line 67: Do we need to increase this on machines with higher physical thread counts?

vasilmkd: The test runs on an artificial `IORuntime` with just 2 threads.

vasilmkd:
I released the following snapshot based off of https://github.com/typelevel/cats-effect/pull/2305/commits/614e90f26f7046793fd2c99a9bf5c8131e5a5f81.

```scala
"io.vasilev" %%% "cats-effect" % "3.2-13-4551ba8"
```

vasilmkd:
I did an extensive round of testing using `fs2`, which thankfully uncovered some issues while I was developing the fix. I have managed to fix those issues and the `fs2` test suite was passing reliably in the end.

vasilmkd:
The snapshot release now makes the CI pass in https://github.com/typelevel/munit-cats-effect/pull/124.

Also, my personal fork passed each time I ran it.
https://github.com/vasilmkd/munit-cats-effect/actions/runs/1203576033

vasilmkd:
I found another issue, working on a fix.

vasilmkd:
Superseded by #2312.

PR #2308, opened 2021-09-05
Run CI on `series/3.2.x`

vasilmkd:
We would have to be careful when eventually merging this branch back into `series/3.x`.


djspiewak:
Ah! I thought I had set this to a wildcard. We should do that instead. series/3.* should do the trick.

djspiewak (approved):

PR #2312, opened 2021-09-06
Address issues with the blocking mechanism of the thread pool

vasilmkd:
Addresses issues identified in #2304.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -362,7 +362,10 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
       // changes to `cats.effect.unsafe` package private code
       ProblemFilters.exclude[DirectMissingMethodProblem]("cats.effect.unsafe.IORuntime.this"),
       ProblemFilters.exclude[DirectMissingMethodProblem](
-        "cats.effect.unsafe.IORuntime.<init>$default$6")
+        "cats.effect.unsafe.IORuntime.<init>$default$6"),
+      // introduced by #3182, Address issues with the blocking mechanism of the thread pool
+      // changes to `cats.effect.unsafe` package private code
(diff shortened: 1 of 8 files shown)


vasilmkd:
Next step, publishing locally and testing against the `fs2` test suite.

vasilmkd:
After running the `fs2` test suite for more than 10 times in a row, everything is fine. Continuing with publishing a snapshot release.

vasilmkd:
```scala
"io.vasilev" %%% "cats-effect" % "3.2-23-cc99a6c"
```
is currently propagating.

vasilmkd:
Running the CI of my copy of the original `munit-cats-effect` issue repeatedly seems to be working fine.
https://github.com/vasilmkd/munit-cats-effect/actions/runs/1209650570

vasilmkd:
The snapshot release is also used in the original issue: https://github.com/typelevel/munit-cats-effect/pull/124. Working fine.

vasilmkd:
Running the `BlockingStressSpec` with 10000 fibers which all execute blocking actions (which depending on interleaving can result in up to 10000 helper threads being spawned) works on the Cats Effect thread pool without issues.

Changing the test runtime from the CE thread pool to `scala.concurrent.ExecutionContext.global` and running the same experiment bogs down and potentially ends in a deadlock.

Feel free to try it yourself.

vasilmkd:
A thread snapshot after running the 10000 helper thread test:
Screen Shot 2021-09-07 at 16 48 02

There are no `-blocking-helper-$index` threads visible, which means that they all finished running and died out as soon as the blocking actions were completed.

djspiewak (approved):

Issue #2314, opened 2021-09-06
Plan for `PolyfillExecutor`


djspiewak:
- [x] Get https://github.com/scala-js/scala-js-macrotask-executor publishing 0.1.0
- [x] Swap series/3.x to use it and remove its matrix-o-doom
- [x] Write some more documentation in the former project
- [x] Coordinate with @sjrd and @gzm0 final approvals
- [x] Release 1.0.0
- [x] Update cats-effect to use 1.0.0

PR #2316, opened 2021-09-06
Replaced `PolyfillExecutionContext` with macrotask-executor

### LINKED ISSUE #1664: Extract polyfill executor into a separate project
This is the Cats Effect side of scala-js/scala-js#4129. We need to pull it out into a separate (typelevel?) project. Honestly this is mostly just infrastructure work, untangling the Firefox build machinery and stuff.

djspiewak:
Always outsource your yaks as aggressively as possible. In this case: https://github.com/scala-js/scala-js-macrotask-executor

Fixes #1664
Relates to #2314


armanbilge (approved):
I think we've cleaned up everything we can! Shame about the browsers, but not too shabby :)

vasilmkd:
Are the browsers not going away?

armanbilge:
I'm afraid not. The problem is stack traces are different between Firefox and Chrome (without sourcemaps) and Node (with sourcemaps), so we still need to test the tracing code across them.

PR #2330, opened 2021-09-08
Update the scastie link for CE 3.2.8

vasilmkd:


website/pages/en/index.js
```diff
@@ -60,7 +60,7 @@ class HomeSplash extends React.Component {
                         <span className="hljs-string">"org.typelevel"</span> %% <span className="hljs-string">"cats-effect"</span> % <span className="hljs-string">"3.2.5"</span>
```
djspiewak: I think this also needs updating :-)

djspiewak (changes requested):

website/pages/en/index.js
vasilmkd: Good catch, thank you.

djspiewak (approved):

PR #2338, opened 2021-09-10
Replacing IOFiber#execute with IOFiber#scheduleFiber

### LINKED ISSUE #2318: Remove `IOFiber#execute` because it is identical to `IOFiber#scheduleFiber`
This should be very easy to do.

ixthe0:
A tiny refactoring of IOFiber - replace *.execute method with *.scheduleFiber according to [this issue](https://github.com/typelevel/cats-effect/issues/2318)


vasilmkd (approved):
Thanks for doing this! Great work.

PR #2343, opened 2021-09-10
Split `interruptible`

### LINKED ISSUE #2268: Deprecate `interruptible(boolean)` and replace with `interruptible`/`interruptibleMany`
By "deprecate" I mean "set to `private[effect]` and undo the currying". We need to maintain binary compatibility but we can break source compatibility with this change. The goal should be for `IO.interruptible(...)` to be equivalent to what `IO.interruptible(false)(...)` is *today*, while `IO.interruptibleMany(...)` should be equivalent to `IO.interruptible(true)(...)`. This change should also be applied to `Sync`.

### Discussed in https://github.com/typelevel/cats-effect/discussions/2248

<div type='discussions-op-text'>

<sup>Originally posted by **etspaceman** August 19, 2021</sup>
I'm wondering if we should create an alias around this. For example,  `interruptible(false)` reads like it isn't interruptible. Perhaps we should consider `interruptibleOnce` and `interruptibleMany`. </div>

alexandrustana:
This PR tries to fix #2268.
Running `mimaReportBinaryIssues` doesn't seem to report anything, so I hope the changes are binary compatible 😄 

TODO:
- [x] add scalafix migration

### DIFF
--- a/core/js/src/main/scala/cats/effect/IOCompanionPlatform.scala
+++ b/core/js/src/main/scala/cats/effect/IOCompanionPlatform.scala
@@ -25,7 +25,7 @@ private[effect] abstract class IOCompanionPlatform { this: IO.type =>
   def blocking[A](thunk: => A): IO[A] =
     apply(thunk)
 
-  def interruptible[A](many: Boolean)(thunk: => A): IO[A] = {
+  private[effect] def interruptible[A](many: Boolean, thunk: => A): IO[A] = {
     val _ = many
     apply(thunk)
   }
--- a/core/js/src/main/scala/cats/effect/IOCompanionPlatform.scala
+++ b/core/js/src/main/scala/cats/effect/IOCompanionPlatform.scala
@@ -25,14 +25,14 @@ private[effect] abstract class IOCompanionPlatform { this: IO.type =>
   def blocking[A](thunk: => A): IO[A] =
     apply(thunk)
 
-  def interruptible[A](many: Boolean)(thunk: => A): IO[A] = {
+  private[effect] def interruptible[A](many: Boolean, thunk: => A): IO[A] = {
     val _ = many
     apply(thunk)
   }
 
-  def interruptible[A](thunk: => A): IO[A] = interruptible(false)(thunk)
+  def interruptible[A](thunk: => A): IO[A] = interruptible(false, thunk)
 
-  def interruptibleMany[A](thunk: => A): IO[A] = interruptible(true)(thunk)
+  def interruptibleMany[A](thunk: => A): IO[A] = interruptible(true, thunk)
 
   def suspend[A](hint: Sync.Type)(thunk: => A): IO[A] = {
     val _ = hint
--- a/core/js/src/main/scala/cats/effect/IOCompanionPlatform.scala
+++ b/core/js/src/main/scala/cats/effect/IOCompanionPlatform.scala
@@ -30,9 +30,9 @@ private[effect] abstract class IOCompanionPlatform { this: IO.type =>
     apply(thunk)
   }
 
-  def interruptible[A](thunk: => A): IO[A] = interruptible(false)(thunk)
+  def interruptible[A](thunk: => A): IO[A] = interruptible(false, thunk)
 
-  def interruptibleMany[A](thunk: => A): IO[A] = interruptible(true)(thunk)
+  def interruptibleMany[A](thunk: => A): IO[A] = interruptible(true, thunk)
 
   def suspend[A](hint: Sync.Type)(thunk: => A): IO[A] = {
     val _ = hint
--- a/core/js/src/main/scala/cats/effect/IOCompanionPlatform.scala
+++ b/core/js/src/main/scala/cats/effect/IOCompanionPlatform.scala
@@ -30,6 +30,10 @@ private[effect] abstract class IOCompanionPlatform { this: IO.type =>
     apply(thunk)
   }
 
+  def interruptible[A](thunk: => A): IO[A] = interruptible(false)(thunk)
+
+  def interruptibleMany[A](thunk: => A): IO[A] = interruptible(true)(thunk)
+
   def suspend[A](hint: Sync.Type)(thunk: => A): IO[A] = {
     val _ = hint
     apply(thunk)
--- a/core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
@@ -33,7 +33,9 @@ private[effect] abstract class IOCompanionPlatform { this: IO.type =>
     Blocking(TypeBlocking, fn, Tracing.calculateTracingEvent(fn.getClass))
   }
 
-  private[effect] def interruptible[A](many: Boolean, thunk: => A): IO[A] = {
+  // this cannot be marked private[effect] because of static forwarders in Java
+  @deprecated("use interruptible / interruptibleMany instead", "3.3.0")
+  def interruptible[A](many: Boolean, thunk: => A): IO[A] = {
     val fn = Thunk.asFunction0(thunk)
     Blocking(
       if (many) TypeInterruptibleMany else TypeInterruptibleOnce,
@@ -41,9 +43,15 @@ private[effect] abstract class IOCompanionPlatform { this: IO.type =>
       Tracing.calculateTracingEvent(fn.getClass))
   }
 
-  def interruptible[A](thunk: => A): IO[A] = interruptible(false, thunk)
+  def interruptible[A](thunk: => A): IO[A] = {
+    val fn = Thunk.asFunction0(thunk)
+    Blocking(TypeInterruptibleOnce, fn, Tracing.calculateTracingEvent(fn.getClass))
+  }
 
-  def interruptibleMany[A](thunk: => A): IO[A] = interruptible(true, thunk)
+  def interruptibleMany[A](thunk: => A): IO[A] = {
+    val fn = Thunk.asFunction0(thunk)
+    Blocking(TypeInterruptibleMany, fn, Tracing.calculateTracingEvent(fn.getClass))
+  }
 
   def suspend[A](hint: Sync.Type)(thunk: => A): IO[A] =
     if (hint eq TypeDelay)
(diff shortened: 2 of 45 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1466,8 +1466,9 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
     override def blocking[A](thunk: => A): IO[A] = IO.blocking(thunk)
 
-    override def interruptible[A](many: Boolean)(thunk: => A): IO[A] =
-      IO.interruptible(many)(thunk)
+    override def interruptible[A](thunk: => A): IO[A] = IO.interruptible(false)(thunk)
```
djspiewak: I think we also need to change the definition of `interruptible` on the `IO` companion itself.

djspiewak (changes requested):
This looks really good apart from my one note! And I would generally trust mima: the approached you used is indeed binary compatible.

kernel/shared/src/main/scala/cats/effect/kernel/Sync.scala
```diff
@@ -88,9 +88,13 @@ trait Sync[F[_]] extends MonadCancel[F, Throwable] with Clock[F] with Unique[F]
    *   The side effect which is to be suspended in `F[_]` and evaluated on a blocking execution
    *   context
    */
-  def interruptible[A](many: Boolean)(thunk: => A): F[A] =
+  private[effect] def interruptible[A](many: Boolean)(thunk: => A): F[A] =
     suspend(if (many) InterruptibleMany else InterruptibleOnce)(thunk)
```
djspiewak: Oh, one more thing: I would delegate this implementation to the new public ones, rather than vice versa, since this is the shim.

vasilmkd:
Thanks for taking this on.

core/shared/src/main/scala/cats/effect/IO.scala
armanbilge: Oh, while you do that, you should also fix the `thunk`s like in [#2226](https://github.com/typelevel/cats-effect/pull/2226/files#diff-2dec4bb5f2b890f966e62cfdfcc6d32746235e45b4a14113f56e38b49923132a), looks like I missed them.

core/js/src/main/scala/cats/effect/IOCompanionPlatform.scala
```diff
@@ -30,6 +30,10 @@ private[effect] abstract class IOCompanionPlatform { this: IO.type =>
     apply(thunk)
   }
 
+  def interruptible[A](thunk: => A): IO[A] = interruptible(false)(thunk)
```
djspiewak: Also needs to be accompanied by marking the original `interruptible` as `private[effect]`

core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
```diff
@@ -29,18 +29,22 @@ private[effect] abstract class IOCompanionPlatform { this: IO.type =>
   private[this] val TypeInterruptibleMany = Sync.Type.InterruptibleMany
 
   def blocking[A](thunk: => A): IO[A] = {
-    val fn = () => thunk
+    val fn = Thunk.asFunction0(thunk)
     Blocking(TypeBlocking, fn, Tracing.calculateTracingEvent(fn.getClass))
   }
 
   def interruptible[A](many: Boolean)(thunk: => A): IO[A] = {
-    val fn = () => thunk
```
djspiewak, line 37: The original `interruptible` needs to be marked as `private[effect]

djspiewak (dismissed):

core/js/src/main/scala/cats/effect/IOCompanionPlatform.scala
alexandrustana: Is it necessary though? The class already has the `private[effect]` modifier.

djspiewak: It is! This class is inherited by a subclass which is *not* `private[effect]`.

djspiewak:
> for the output test folder a new version which is not yet released and will contain these changes)

Can we solve this by publishing a snapshot?

core/js/src/main/scala/cats/effect/IOCompanionPlatform.scala
alexandrustana: I've added the modifier. But the only way to still be _binary compatible_ was to keep the methods as is in the `IOCompanionPlatforms`, override the method in `IO.scala` and add the modifier there.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1466,9 +1466,13 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
     override def blocking[A](thunk: => A): IO[A] = IO.blocking(thunk)
 
-    override def interruptible[A](many: Boolean)(thunk: => A): IO[A] =
+    private[effect] override def interruptible[A](many: Boolean)(thunk: => A): IO[A] =
```
djspiewak: Wait you can do that?

vasilmkd: It technically doesn't change the visibility in the bytecode, so yes.

djspiewak: Are we sure it does what we want downstream?

vasilmkd: Yeah, let's not take that risk.

djspiewak:
```
scala> import cats.effect.IO
import cats.effect.IO

scala> IO.interruptible(false) { println("hi") }
val res0: cats.effect.IO[Unit] = IO(...)
```

Looks like it doesn't mask it.

vasilmkd: Ok, thanks for confirming. Good to know.

djspiewak:
For posterity, the problem here is the following:

```
[error] cats-effect: Failed binary compatibility check against org.typelevel:cats-effect_2.13:3.1.0! Found 1 potential problems (filtered 7)
[error]  * static method interruptible(Boolean,scala.Function0)cats.effect.IO in class cats.effect.IO does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("cats.effect.IO.interruptible")
```

Looks like the `private[effect]` overload isn't getting peered into a static forwarder. I'm not totally certain how to trick the compiler into doing that…

djspiewak:
IMO, it's worth breaking binary compatibility with Java interop on this one function in order to get the better API for everyone. Anyone have any objections? Paging @rossabaker and @johnynek for opposing opinions.

johnynek:
just to chime in: 

My feeling is that runtime errors really suck, and if you want to be a super core library, you will be implicated in many diamonds, so try to never break binary compatibility.

But this is your project, of course, so make your own calls.

djspiewak:
I do very much agree with this. The only reason I'm even pondering the breakage is the fact that it's confined to the Java interop exclusively. Scala code paths never hit this forwarder and remain binary compatible.

armanbilge:
FWIW http4s recently made a similar decision to eschew Java forwarders in https://github.com/http4s/http4s/pull/5071#issuecomment-898705809.

alexandrustana:
@djspiewak @vasilmkd, can someone publish a snapshot, please, so I can try and write the scalafix migration? (if it is still needed I mean 😄 )

vasilmkd:
I will right now.

vasilmkd:
```scala
"io.vasilev" %%% "cats-effect" % "3.3-208-b352672"
```
@alexandrustana

vasilmkd (commented):
How did the `build.sbt` changes get pulled in? That PR is not merged yet.

rossabaker:
This one is tough: a marginally better API for what is probably 100% of users, or a landmine for an unanticipated use case.  We chose the Scala users in http4s, but it makes me queasy, and the risk is higher on this project.

A third option is bytecode postprocessing.  I haven't done it in the Scala Era, but I'm pretty sure it would be possible.

vasilmkd:
With a repeated word of caution and previous experience, maybe it is worth reconsidering the extent of this change.

alexandrustana:
Having these in mind, should I close this PR?

vasilmkd:
No, not yet, let's have a discussion with Daniel, whenever he's back.

We'll definitely add `interruptibleOnce` and `interruptibleMany` if this doesn't go through, and the scalafix rules will still be good (with minor changes).

djspiewak:
I'm back ish. :-) To @rossabaker's point… There's another option here, which is that we can keep the current `interruptible` overload public and mark it as `@deprecated`. That's what's visible currently in the PR, and thus there is no binary breakage at all. It's kind of ugly, but it's not really the end of the world, and it's probably less fraught than bytecode shenanigans in the build.

Either way, I think that we can shift the discussion a bit. We agree the changes here are good (I think), and they're binary-compatible as currently written. I'm in favor of merging this PR and having a follow-up if we want to get fancy with hiding the old `interruptible` overload.

scalafix/build.sbt
```diff
@@ -43,3 +43,37 @@ lazy val v3_0_0_tests = project
...
+      "org.typelevel" %% "cats-effect" % "3.0.0"
+    ),
+    scalacOptions += "-P:semanticdb:synthetics:on"
+  )
+
+lazy val v3_3_0_output = project
+  .in(file("v3_3_0/output"))
+  .settings(
+    libraryDependencies ++= Seq(
+      "io.vasilev" %% "cats-effect" % "3.3-208-b352672"
```
djspiewak, line 61: Let's remember to update this to the mainline 3.3.0 once released

djspiewak (approved):
Awesome work, @alexandrustana! Thanks for shepherding this through the process, and sorry for handing you such a thorny change.

fthomas:
FYI: https://github.com/scala-steward-org/scala-steward/pull/2270 adds the v3_3_0 Scalafix to Scala Steward.

alexandrustana:
@djspiewak, @vasilmkd thank you for all the help and support 🙂

vasilmkd:
Thank you @fthomas!

@alexandrustana Thank you for the outstanding work and your patience.

PR #2348, opened 2021-09-13
Clear the `tracingEvents` ring buffer when a fiber ends

### LINKED ISSUE #2317: Clear the `tracingEvents` ring buffer when a fiber ends
Feel free to ask me for pointers.

milanvdm:
Closes https://github.com/typelevel/cats-effect/issues/2317


djspiewak (approved):
Originally I was thinking we would just clear the data structure itself, but this is better since the internal array is fixed length.

vasilmkd:
@djspiewak Check the issue for this change, I left a similar comment there for you. 😄

djspiewak:
LOL. I think this is is better. It means that we're able to avoid the overhead of all of the object headers.

PR #2350, opened 2021-09-14
Fix Cancellation Point in Semaphore

ChristopherDavenport:
We don't want to possible cancel those wakeups at the wrong point in this flatMap chain


vasilmkd (dismissed):

vasilmkd:
Good catch!

diesalbla (dismissed):

djspiewak:
@ChristopherDavenport Can we do this against `series/3.2.x` instead please?

ChristopherDavenport:
Sorry, coded it against it but forgot to target it when I PR'd

vasilmkd (approved):

djspiewak (approved):

PR #2359, opened 2021-09-18
Add embedError convenience function

ChristopherDavenport:
Not much more to it. Just a convenience function. 

Especially useful when you are trying to save the outcome of some computation. But then you need to resume flow.


kernel/shared/src/main/scala/cats/effect/kernel/Outcome.scala
```diff
@@ -162,6 +162,15 @@ object Outcome extends LowPriorityImplicits {
...
     either.fold(Errored(_), a => Succeeded(a.pure[F]))
 
+  def embedError[F[_], A](outcome: Outcome[F, Throwable, A])(
+      implicit F: ApplicativeError[F, Throwable]): F[A] =
+    outcome match {
+      case Succeeded(fa) => fa
+      case Errored(e) => F.raiseError(e)
+      case Canceled() =>
+        F.raiseError(new java.util.concurrent.CancellationException("Outcome was Canceled"))
+    }
```
vasilmkd: `Outcome` has a `fold` method for this pattern match.

vasilmkd (dismissed):

vasilmkd (commented):
I think it's worth bikeshedding on the name and adding an accompanying scaladoc.

kernel/shared/src/main/scala/cats/effect/kernel/Outcome.scala
ChristopherDavenport: I switched to embed which seems more direct. Can use fold if you'd prefer.

vasilmkd: Even better.

vasilmkd:
I think the scaladoc should mention the cancelation case, saying that it embeds cancelation as an error. Otherwise 👍🏻

kernel/shared/src/main/scala/cats/effect/kernel/Outcome.scala
```diff
@@ -166,7 +166,8 @@ object Outcome extends LowPriorityImplicits {
    * embedError allows the restoration to a normal development flow from an Outcome.
    *
    * This can be useful for storing the state of a running computation and then waiters for that
-   * data can act and continue forward on that shared outcome.
+   * data can act and continue forward on that shared outcome. Cancelation is encoded
+   * as an exception.
```
armanbilge:
```suggestion
   * as a `CancellationException`.
```

kernel/shared/src/main/scala/cats/effect/kernel/Outcome.scala
```diff
@@ -162,6 +162,19 @@ object Outcome extends LowPriorityImplicits {
...
     either.fold(Errored(_), a => Succeeded(a.pure[F]))
 
+  /**
+   * embedError allows the restoration to a normal development flow from an Outcome.
+   *
+   * This can be useful for storing the state of a running computation and then waiters for that
+   * data can act and continue forward on that shared outcome. Cancelation is encoded as a
+   * `CancellationException`.
+   */
+  def embedError[F[_], A](outcome: Outcome[F, Throwable, A])(
```
djspiewak: I think this should be a member of `Outcome` like the other `embed`s.

djspiewak (changes requested):

djspiewak (approved):

PR #2360, opened 2021-09-19
Add serialization support 

### LINKED ISSUE #587: Make IO serializable
Sometimes it's useful to serialize IO action and pass it over wire. Monix and ZIO support that by extending `Serializable` trait, cats-effect - not.

gagandeepkalra:
resolves https://github.com/typelevel/cats-effect/issues/587


tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
```diff
@@ -255,6 +257,44 @@ trait IOPlatformSpecification { self: BaseSpec with ScalaCheck =>
         } must completeAs(true)
       }
 
+      "serialise IO when not including evalOn" in ticked { implicit ticker =>
+        import SerializationUtil._
+
```
armanbilge: This is cool! But maybe we can re-use the `SerializableTests` in `cats-kernel-laws`? Example: https://github.com/typelevel/cats/blob/e0784fc0d95b9c4eed0bb1ddce6c69e3fd9a4a4f/tests/src/test/scala/cats/tests/FunctionSuite.scala#L46

tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
```diff
@@ -255,6 +257,44 @@ trait IOPlatformSpecification { self: BaseSpec with ScalaCheck =>
...
+          IO.pure(42).attempt,
+          IO.raiseError(new RuntimeException("Err")).handleErrorWith(_ => IO.pure(42)),
+          IO.canceled,
+          IO.pure(42).onCancel(IO.pure(42)),
+          IO.pure(42).uncancelable,
+          IO.pure(42).start,
+          IO.racePair(IO.pure(42), IO.pure(42)),
+          IO.sleep(0.second),
+          IO.trace
+        ).map(_.as(42)).zipWithIndex
```
armanbilge: Just an idea—I wonder if with some strategic overrides in `AsyncGenerators` you can use it to generate `IO` _without_ `evalOn`.

 https://github.com/typelevel/cats-effect/blob/series/3.x/kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/Generators.scala#L282

vasilmkd: Here's how this is done.
https://github.com/typelevel/cats-effect/blob/ca1da1f333af06aaed3d03b6082092763afa679b/testkit/shared/src/main/scala/cats/effect/testkit/TestInstances.scala#L68-L74

kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/Generators.scala
```diff
@@ -26,12 +26,12 @@ import scala.collection.immutable.SortedMap
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration.FiniteDuration
 
-trait GenK[F[_]] {
+trait GenK[F[_]] extends Serializable {
```
vasilmkd, line 30: For my understanding, why do `GenK[F[_]]` and `Generators1[F[_]` need to be serializable?

tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
armanbilge: Nice!

tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
vasilmkd: @armanbilge Can you please point me to the source code of `SerializableTests`? I was looking for it but I couldn't find it for some reason.

tests/jvm/src/test/scala/cats/effect/SerializationUtil.scala
```diff
@@ -0,0 +1,50 @@
...
+
+package cats.effect
+
+import scala.util.Try
+
+object SerializationUtil {
+  import java.io._
+  import java.util.Base64
+
+  def serialize(`object`: Serializable): Try[String] = {
```
vasilmkd: Use `obj` instead of object with backticks please.

tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
armanbilge: The meat of it is in the law actually.

https://github.com/typelevel/cats/blob/e0784fc0d95b9c4eed0bb1ddce6c69e3fd9a4a4f/kernel-laws/shared/src/main/scala/cats/kernel/laws/SerializableLaws.scala#L26

The test is here:
https://github.com/typelevel/cats/blob/e0784fc0d95b9c4eed0bb1ddce6c69e3fd9a4a4f/kernel-laws/shared/src/main/scala/cats/kernel/laws/discipline/SerializableTests.scala

vasilmkd: That's perfect, let's use that. Thank you.

tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
djspiewak: You could also make `TestContext` extend `Serializable`

kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/Generators.scala
gagandeepkalra: Recursive generator lambda captures both of these references

tests/jvm/src/test/scala/cats/effect/SerializationUtil.scala
gagandeepkalra: deleted this file now

tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
gagandeepkalra: yes that was very nice observation, but I couldn’t see it through. 

Stuck because somehow `ExecutionContextImpl` instance is seeping in

tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
gagandeepkalra: SerializableTests doesn't require Arbitrary, I have used law directly

kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/pure.scala
```diff
@@ -361,8 +361,9 @@ object pure {
       ft.mapK(fk)
     }
 
+  // todo: MVar is not Serializable, release then update here
```
gagandeepkalra, line 365: inclusion of `racePair`test leads us here, we'd need to update Mvar with serialization

gagandeepkalra:
Thanks for the detailed review, 

after adding the property test for IO, I noticed a lot of trait/abstract classes missing serialization, have enhanced those in this very PR

testkit/shared/src/main/scala/cats/effect/testkit/TestInstances.scala
```diff
@@ -76,6 +77,22 @@ trait TestInstances extends ParallelFGenerators with OutcomeGenerators with Sync
     Arbitrary(generators.generators[A])
   }
 
+  def arbitraryIOWithoutContextShift[A: Arbitrary: Cogen]: Arbitrary[IO[A]] = {
```
armanbilge, line 80:
```suggestion
  def arbitraryIOWithoutEvalOn[A: Arbitrary: Cogen]: Arbitrary[IO[A]] = {
```

tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
armanbilge: Good catch!

tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
armanbilge: I think it's fine, `evalOn` was never supposed to be serializable anyway :)

djspiewak:
@gagandeepkalra This is awesome work, but would you mind running `sbt prePR` and committing the results? Scalafmt is unhappy with you. I tried doing it myself, but since you used the `series/3.x` branch on your local fork, I wasn't able to push

gagandeepkalra:
> @gagandeepkalra This is awesome work, but would you mind running `sbt prePR` and committing the results? Scalafmt is unhappy with you. I tried doing it myself, but since you used the `series/3.x` branch on your local fork, I wasn't able to push

Hi @djspiewak, sure thing, let me get to this over the weekend.

Also, what do you think about this- https://github.com/typelevel/cats-effect/pull/2360#discussion_r713246245/ Should I go there and make this change, then the support would be complete, of course this PR won't have to wait.

djspiewak:
@gagandeepkalra I guess I'm a bit confused as to why `PureConc` is affected at all? Shouldn't we exclusively be testing serializability of `IO`?

gagandeepkalra:
I need some help, for `SyncIO` the test fails at this seed: "mWUDuysFx0P-r_sHw_GGVwvm-TcqPoLNs8aOV2g8FuD=" 

yet I'm not able to reproduce the failure locally, would it help to rerun the case (I don't have the permission to re-run)?

armanbilge:
@gagandeepkalra try setting the seed like this:
```scala
implicit def parameters = Parameters(seed =
    Seed.fromBase64("mWUDuysFx0P-r_sHw_GGVwvm-TcqPoLNs8aOV2g8FuD=").toOption
  )
```

gagandeepkalra:
Hi @armanbilge, think I'm making a real silly error somewhere

this is my testcase with your suggestion, the test would not fail wheresoever I put this seed. Please help.
```
package cats.effect

import cats.kernel.laws.discipline.MonoidTests
import cats.kernel.laws.SerializableLaws.serializable
import cats.laws.discipline.AlignTests
import cats.laws.discipline.arbitrary._
import cats.effect.laws.SyncTests
import cats.syntax.all._
import org.scalacheck.{Arbitrary, Gen, Prop}
import Prop.{Exception, Proof, Result, forAll}
import cats.platform.Platform
import org.scalacheck.rng.Seed
import org.specs2.scalacheck.Parameters
import org.typelevel.discipline.specs2.mutable.Discipline

import scala.util.control.NonFatal

class SyncIOSpec extends BaseSpec with Discipline with SyncIOPlatformSpecification {

  implicit override val defaultParameters: Parameters = Parameters(
    seed = Seed.fromBase64("mWUDuysFx0P-r_sHw_GGVwvm-TcqPoLNs8aOV2g8FuD=").toOption
  )

  "sync io monad" should {
    "serialize" in {
      forAll { (io: SyncIO[Int]) => serializable(io)
      }
    }
  }
}
```

armanbilge:
I'm sorry! I think you ran into this bug maybe?
- https://github.com/etorreborre/specs2/issues/1025

That's going to make it pretty hard to replicate 😕

djspiewak:
Here's an example of how you can make it work:

```scala
  checkAll(
    "Kleisli[PureConc]",
    GenTemporalTests[Kleisli[TimeT[PureConc[Int, *], *], MiniInt, *], Int]
      .temporal[Int, Int, Int](10.millis)
    // we need to bound this a little tighter because these tests take FOREVER
  )(Parameters(minTestsOk = 25))
```

Add `seed = ...` to the `Parameters` (obviously also replacing the `checkAll` with the appropriate one) and you're good to go!

armanbilge:
@gagandeepkalra we just updated specs2 in https://github.com/typelevel/cats-effect/pull/2583 which should have a fix for the seed issue. You can merge the latest series/3.x and try again ... thanks!

vasilmkd:
Actually, I probably need to release `discipline-specs2`?

armanbilge:
Oh, are we still only getting specs2 transitively in some subprojects? We should really just fix that :)

gagandeepkalra:
okay so I tried skipping specs2 and working directly with Gen to produce a value then call serialize. The test passed, I'm starting to think the seed we get here is not correct.

I added another case for @djspiewak's suggestion, they both generated similar looking value given the seed.

code:

```scala
package cats.effect

import cats.kernel.laws.SerializableLaws.serializable
import org.scalacheck.rng.Seed
import org.scalacheck.{Arbitrary, Gen, Prop}
import org.specs2.scalacheck.Parameters
import org.typelevel.discipline.specs2.mutable.Discipline

import scala.util.Try
import scala.util.control.NonFatal

class SyncIOSpec extends BaseSpec with Discipline with SyncIOPlatformSpecification {

  implicit override val defaultParameters: Parameters = Parameters(
    minSize = 10000,
    maxSize = 10000,
    seed = Seed.fromBase64("mWUDuysFx0P-r_sHw_GGVwvm-TcqPoLNs8aOV2g8FuD=").toOption
  )

  "sync io monad" should {

    val arb: Arbitrary[SyncIO[Int]] = implicitly[Arbitrary[SyncIO[Int]]]

    val p: Prop = Prop.forAll(arb.arbitrary)(serializable)(identity, implicitly, implicitly)

    check(p, defaultParameters, defaultFreqMapPretty)

    "test Specific Serialize" in {

      val seed: Try[Seed] = Seed.fromBase64("mWUDuysFx0P-r_sHw_GGVwvm-TcqPoLNs8aOV2g8FuD=")
      val value: SyncIO[Int] = arb.arbitrary.apply(Gen.Parameters.default.withInitialSeed(seed.get), seed.get).get

      verifySerializableOrThrow(value)
      true
    }
  }

  def verifySerializableOrThrow[A](a: A): Unit = {
    import java.io.{ByteArrayInputStream, ByteArrayOutputStream, ObjectInputStream, ObjectOutputStream}

    val baos = new ByteArrayOutputStream()
    val oos = new ObjectOutputStream(baos)
    var ois: ObjectInputStream = null // scalastyle:ignore null
    try {
      oos.writeObject(a)
      oos.close()
      val bais = new ByteArrayInputStream(baos.toByteArray)
      ois = new ObjectInputStream(bais)
      val a2: A = ois.readObject().asInstanceOf[A]
      println("BEFORE:" + a)
      println("AFTER:" + a2)
      ois.close()
    } catch {
      case NonFatal(t) => println(t)
    } finally {
      oos.close()
      if (ois != null) ois.close() // scalastyle:ignore null
    }
  }
}

```

djspiewak (dismissed):
So I had a go at this locally and repeatedly ran and re-ran the tests for a few hours. Never had a single failure. I think any issues here were resolved incidentally by something else in `series/3.x` along the way.

vasilmkd:
Love to see accidentally unblocking a big feature.

djspiewak:
Hey! Fun fact: the failures are Scala 3 specific.

djspiewak:
Even more fun fact, it serializes correctly but doesn't *deserialize*. In other words, this *really* smells like an `Externalizable` that wasn't correctly implemented. *Casts summoning spell on @smarter*

What we know is that hte `SyncIO` under test is comprised of the following three stages:

```
SyncIO.FlatMap[Any, Any]
SyncIO.Pure[Any]
SyncIO.Success[A]
```

The only thing non-trivial in there is really `FlatMap`, which contains a `Function1`.

djspiewak:
```
java.lang.RuntimeException
  | => tat scala.runtime.ModuleSerializationProxy$.scala$runtime$ModuleSerializationProxy$$rethrowRuntime(ModuleSerializationProxy.scala:38)
	at scala.runtime.ModuleSerializationProxy$$anon$1.computeValue(ModuleSerializationProxy.scala:30)
	at scala.runtime.ClassValueCompat$JavaClassValue.computeValue(ClassValueCompat.scala:24)
	at java.base/java.lang.ClassValue.getFromHashMap(ClassValue.java:226)
	at java.base/java.lang.ClassValue.getFromBackup(ClassValue.java:208)
	at java.base/java.lang.ClassValue.get(ClassValue.java:114)
	at scala.runtime.ClassValueCompat.get(ClassValueCompat.scala:33)
	at scala.runtime.ModuleSerializationProxy.readResolve(ModuleSerializationProxy.scala:45)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at java.base/java.io.ObjectStreamClass.invokeReadResolve(ObjectStreamClass.java:1250)
	at java.base/java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:2096)
	at java.base/java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1594)
	at java.base/java.io.ObjectInputStream.defaultReadFields(ObjectInputStream.java:2355)
	at java.base/java.io.ObjectInputStream.readSerialData(ObjectInputStream.java:2249)
	at java.base/java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:2087)
	at java.base/java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1594)
	at java.base/java.io.ObjectInputStream.readObject(ObjectInputStream.java:430)
	at cats.effect.SyncIOSpec.liftedTree1$1(SyncIOSpec.scala:247)
	at cats.effect.SyncIOSpec.$init$$$anonfun$159$$anonfun$157$$anonfun$3(SyncIOSpec.scala:252)
```

djspiewak:
Okay, couple problems here. First, accessing `MODULE$` is throwing an exception. This is *probably* why there's a difference between Scala 3 and Scala 2, since this is where you might get some weird initialization order BS going on. Unfortunately, in order to diagnose this further, we really need to know what class we're hitting. The relevant code:

```scala
private[runtime] object ModuleSerializationProxy {
  private val instances = new ClassValueCompat[Object] {
    @nowarn("cat=deprecation") // AccessController is deprecated on JDK 17
    def getModule(cls: Class[_]): Object =
      java.security.AccessController.doPrivileged(
        (() => cls.getField("MODULE$").get(null)): PrivilegedExceptionAction[Object])
    override protected def computeValue(cls: Class[_]): Object =
      try getModule(cls)
      catch {
        case e: PrivilegedActionException =>
          rethrowRuntime(e.getCause)
      }
  }
```

Speaking of which, there's a bug here! (note the above is in the Scala stdlib, shared between 2 and 3). The bug becomes more apparent when I show you `rethrowRuntime`:

```scala
  private def rethrowRuntime(e: Throwable): Object = {
    val cause = e.getCause
    cause match {
      case exception: RuntimeException => throw exception
      case _ => throw new RuntimeException(cause)
    }
  }
```

So… why are we calling `getCause` on the results of `getCause`? I'm willing to bet that we should not be.

At any rate, I'm relatively certain this is definitively problems upstream, so I think the right approach here is to make the serialization tests Scala 2 exclusive.

smarter:
Scala 3 should behave like Scala 2 for object serialization (https://github.com/scala/scala/pull/7297 / https://github.com/lampepfl/dotty/pull/5775), but we emit all objects as `extends Serializable` for convenience and lambdas in objects are emitted as instance methods of the module class, so serializing/deserializing them will trigger the ModuleSerializationProxy codepath unlike Scala 2: https://github.com/lampepfl/dotty/blob/5626f2512d1451bd7d933c26099f82b993e68f04/compiler/src/dotty/tools/backend/jvm/BCodeSkelBuilder.scala#L250-L268

smarter:
> So… why are we calling getCause on the results of getCause? I'm willing to bet that we should not be.

Yeah that seems weird, could you open an issue at scala/scala and ping Jason who wrote that code originally? Meanwhile, I think you could redefine ModuleSerializationProxy in your own project to shadow the one coming from scala-library and add a println to figure out what the name of the deserialized class is.

djspiewak:
scala/bug#12553

I'll see about shadowing to see if we can get the actual class.

djspiewak:
Fun fact: you can't really shadow it because it's not on the classpath. So you can't get it to compile.

smarter:
Not sure I get it, I think you can copy-paste ModuleSerializationProxy.scala and ClassValueCompat.scala in your project and it should compile. But it might be easier to just fire up a debugger and break on rethrowRuntime.

djspiewak:
@smarter Here's something cool: if I shadow everything needed to make the shadowing compile (https://gist.github.com/423c26f4ed8147d0e97b03ee3c406d03), the exception disappears and both serialization and deserialization succeed.

smarter:
Huh. Before going deeper into the rabbit hole, can you try using a more recent version of Scala 3? (latest nightly is `scalaVersion := "3.2.0-RC1-bin-20220308-29073f1-NIGHTLY"`)

smarter:
... actually latest nightly is `scalaVersion := "3.1.3-RC1-bin-20220311-895598f-NIGHTLY"`, I forgot we reverted the version bump.

djspiewak:
@smarter Same issue :-(

djspiewak:
I think we'll probably just disable the tests on Scala 3 for the time being. We'll make a ticket to come back and try again once scala/scala#9970 is released as part of a Scala 3 build. I just don't want to keep this PR languishing forever. I really do wish we had a more minimal reproduction though. Any thoughts on whether this would also be worth filing upstream on Dotty?

smarter:
So I tried the copy-paste ModuleSerializationProxy.scala trick myself and was able to get a better exception after removing the extra getCause:

> Class scala.runtime.ModuleSerializationProxy$ can not access a member of class cats.effect.SyncIO$RealTime$ with modifiers "public static final" (Reflection.java:102)

And the reason the JVM is unhappy is that `SyncIO$RealTime$` is emitted as a private class because it's private in the source unlike scala 2 which iirc always emits all classes (even private inner classes) as public (not even package-private!). So either we need to emit all classes as public too or we need to add a setAccessible in ModuleSerializationProxy.

smarter:
The trouble with using setAccessible is that it will likely stop working when attempting to use the java module system, so we might have no choice but to make every class public.

smarter:
https://github.com/lampepfl/dotty/pull/14686

djspiewak:
> So I tried the copy-paste ModuleSerializationProxy.scala trick myself and was able to get a better exception after removing the extra getCause:

Thank you for doing that! It's absolutely wild that it didn't fail for me when I did that exact series of steps.

> And the reason the JVM is unhappy is that SyncIO$RealTime$ is emitted as a private class because it's private in the source unlike scala 2 which iirc always emits all classes (even private inner classes) as public (not even package-private!). So either we need to emit all classes as public too or we need to add a setAccessible in ModuleSerializationProxy.

Ahhhhhhhhh. Okay that makes sense. That definitely seems like something that will end up being a blocker for things involving serialization (*coughs* Spark), particularly given the fact that lambdas make it very easy to accidentally close over state which would trigger this path.

djspiewak (approved):

PR #2365, opened 2021-09-21
Add documentation to IO.async

kamilkloch:
`k: (Either[Throwable, A] => Unit) => IO[Option[IO[Unit]]]` cries for some explanation :)

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -936,14 +936,14 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
    * Suspends an asynchronous side effect in `IO`.
    *
    * The given function will be invoked during evaluation of the `IO` to "schedule" the
-   * asynchronous callback, where the callback of type `Either[Throwable, A] => Unit` 
-   * is the parameter passed to that function. Only the ''first'' invocation of the callback
-   * will be effective! All subsequent invocations will be silently dropped.
+   * asynchronous callback, where the callback of type `Either[Throwable, A] => Unit` is the
+   * parameter passed to that function. Only the ''first'' invocation of the callback will be
+   * effective! All subsequent invocations will be silently dropped.
    *
-   * The process of registering the callback itself is suspended in `IO` 
-   * (the outer `IO` of `IO[Option[IO[Unit]]]`).
+   * The process of registering the callback itself is suspended in `IO` (the outer `IO` of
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -932,6 +932,44 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
   def defer[A](thunk: => IO[A]): IO[A] =
     delay(thunk).flatten
 
+  /**
+   * Suspends an asynchronous side effect in `IO`.
```
armanbilge, line 936: Swap out `F` for `IO`? :)

armanbilge (dismissed):

core/shared/src/main/scala/cats/effect/IO.scala
kamilkloch: The context is IO, not F. See the existing scaladoc for `IO.async_`.

armanbilge: Oh, I'm so sorry!! Last time I try to review on my phone.

kamilkloch: Sure, no problem, thanks for looking into it!

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -932,6 +932,44 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
...
+   *
+   * The given function will be invoked during evaluation of the `IO` to "schedule" the
+   * asynchronous callback, where the callback of type `Either[Throwable, A] => Unit` 
+   * is the parameter passed to that function. Only the ''first'' invocation of the callback
+   * will be effective! All subsequent invocations will be silently dropped.
+   *
+   * The process of registering the callback itself is suspended in `IO` 
+   * (the outer `IO` of `IO[Option[IO[Unit]]]`).
+   *
+   * The effect returns `Option[IO[Unit]]` which is an optional finalizer to be run in the event
```
```suggestion
   * (the outer `IO` of `IO[Option[IO[Unit]]]`). This effect returns `Option[IO[Unit]]` which is an optional finalizer to be run in the event 
```

armanbilge: ~Nope, ignore. I derped again I think.~ Actually, I think it makes sense.

djspiewak (dismissed):
This is awesome! Sorry I didn't see this sooner!

djspiewak (approved):

Issue #2285, opened 2021-08-31
Disable tracing on fullOptJS without source maps


djspiewak:
This exact combination of configurations almost certainly represents a user who wants maximum performance and doesn't care at all about introspecting their code.

PR #2366, opened 2021-09-21
Disable JS tracing in production

armanbilge:
Implicitly addresses #2285. See also https://github.com/typelevel/cats-effect/discussions/2221#discussioncomment-1260630.

### DIFF
--- a/core/js/src/main/scala/cats/effect/tracing/TracingPlatform.scala
+++ b/core/js/src/main/scala/cats/effect/tracing/TracingPlatform.scala
@@ -20,6 +20,7 @@ import cats.effect.kernel.Cont
 
 import scala.collection.mutable
 import scala.reflect.NameTransformer
+import scala.scalajs.LinkingInfo
 import scala.scalajs.js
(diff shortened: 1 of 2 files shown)


vasilmkd:
This looks good to me. Before I press approve though, didn't you discuss that `fullOptJS` and source maps should be traced? This change seems to not support that? Sorry if this is an obtuse question.

armanbilge:
You're right, we're going back on our previous decision. See https://discord.com/channels/632277896739946517/839263556754472990/889877834150068244

In an ideal world, we'd find a way to support fullOpt/sourcemaps. But the advantage of totally disabling tracing for fullOpt is that it can be dead-code-eliminated.

vasilmkd:
Yes, good. I'm all for it.

vasilmkd (approved):

djspiewak:
Let's do it and wait for the complaints to roll in. :-D It's easier to add functionality later than it is to go back on functionality we've already shipped.

PR #2367, opened 2021-09-21
Make JS IORuntime externally configurable

armanbilge:
Should put JS in lock-step with JVM. Relies on `process.env` instead of system properties.

### DIFF
--- a/core/js/src/main/scala/cats/effect/process.scala
+++ b/core/js/src/main/scala/cats/effect/process.scala
@@ -16,6 +16,8 @@
 
 package cats.effect
 
+import cats.syntax.all._
+
 import scala.scalajs.js
 import scala.util.Try
 
@@ -27,7 +29,8 @@ private[effect] object process {
   def env(key: String): Option[String] =
     Try(js.Dynamic.global.process.env.selectDynamic(key))
       .orElse(Try(js.Dynamic.global.process.env.selectDynamic(s"REACT_APP_$key")))
+      .widen[Any]
+      .collect { case v: String if !js.isUndefined(v) => v }
       .toOption
-      .filterNot(js.isUndefined)
-      .flatMap(x => Try(x.asInstanceOf[String]).toOption)
+
 }
--- a/core/js/src/main/scala/cats/effect/process.scala
+++ b/core/js/src/main/scala/cats/effect/process.scala
@@ -27,7 +27,7 @@ private[effect] object process {
   def env(key: String): Option[String] =
     Try(js.Dynamic.global.process.env.selectDynamic(key))
       .orElse(Try(js.Dynamic.global.process.env.selectDynamic(s"REACT_APP_$key")))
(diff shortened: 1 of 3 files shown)


vasilmkd:
Are we dropping the support for the react app magic environment variables? Or is that no longer an issue?

core/js/src/main/scala/cats/effect/process.scala
```diff
@@ -0,0 +1,33 @@
...
+
+private[effect] object process {
+
+  def argv: Option[List[String]] = Try(
+    js.Dynamic.global.process.argv.asInstanceOf[js.Array[String]].toList.drop(2)).toOption
+
+  def env(key: String): Option[String] =
+    Try(js.Dynamic.global.process.env.selectDynamic(key))
+      .toOption
+      .orElse(Try(js.Dynamic.global.process.env.selectDynamic(s"REACT_APP_$key")).toOption)
```
armanbilge: @vasilmkd moved here?

vasilmkd: Much better indeed, sorry for missing it.

vasilmkd: You can use `.orElse` and the other chained methods directly on `Try` instead of going through `Option`. Obviously convert to `Option` at the very end.

vasilmkd (dismissed):
Some Scala stdlib nits. Otherwise looks great.

core/js/src/main/scala/cats/effect/process.scala
armanbilge: Hmm, seems `Try` doesn't support filtering and rearranging that was a bad idea. So only a modest improvement in the end :)

vasilmkd: Lol, it doesn't support `filterNot`. `filter` works.

vasilmkd: The whole block would have to look like this I guess:
```scala
def env(key: String): Option[String] =
  Try(js.Dynamic.global.process.env.selectDynamic(key))
    .orElse(Try(js.Dynamic.global.process.env.selectDynamic(s"REACT_APP_$key")))
    .collect { case v if !js.isUndefined(v) => v.asInstanceOf[String] }
    .toOption
```

armanbilge: Wow I suck at this 😆 that looks great. Maybe we can make that `case v: String if !js.isUndefined(v)`?

vasilmkd: Even better.

armanbilge: Took one more trick 😆 Nice golfing, thanks!

vasilmkd (commented):
I'll do another pass a bit later.

djspiewak (approved):
Needs documentation, but otherwise looks great!

PR #2369, opened 2021-09-22
Switch to the typelevel jdk index

vasilmkd:
I donated my JDK index to typelevel.


armanbilge (approved):
Thanks for driving us forward! 😀

PR #2371, opened 2021-09-23
Fix error and clarify wording in `MonadCancel` documentation

bastewart:
This was prompted by a [discussion on discord](https://discord.com/channels/632277896739946517/632278585700384799/890606516380512296).

The `MonadCancel` documentation applied nested
`uncancelable` `Poll`s in the incorrect order.

In addition the `MonadCancel.uncancelable` docstring
is slightly ambiguous about what order nested masks should
be applied. I've tried to clarify this...

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
@@ -257,8 +257,8 @@ trait MonadCancel[F[_], E] extends MonadError[F, E] {
    *
    * Masks can also be stacked or nested within each other. If multiple masks are active, all
    * masks must be undone so that cancelation can be observed. In order to completely unmask
-   * within a multi-masked region, the poll corresponding to each mask must be applied,
-   * innermost-first.
+   * within a multi-masked region the poll corresponding to each mask must be applied to the
+   * effect, outermost-first.
    *
    * {{{
    *
@@ -273,7 +273,8 @@ trait MonadCancel[F[_], E] extends MonadError[F, E] {
    * The following operations are no-ops:
    *
    *   1. Polling in the wrong order
-   *   1. Applying the same poll more than once: `poll(poll(fa))`
+   *   1. Subsequent polls when applying the same poll more than once: `poll(poll(fa))` is
+   *      equivalent to `poll(fa)`
    *   1. Applying a poll bound to one fiber within another fiber
    *
    * @param body
--- a/docs/typeclasses/monadcancel.md
+++ b/docs/typeclasses/monadcancel.md
@@ -135,12 +135,14 @@ The use of the `outer` poll within the inner `uncancelable` is simply ignored un
 ```scala
 MonadCancel[F] uncancelable { outer =>
   MonadCancel[F] uncancelable { inner =>
-    outer(inner(fa))
+    inner(outer(fa))
   }
 }
 ```
 
 This is simply equivalent to writing `fa`. Which is to say, `poll` composes in the way you would expect.
+
+They are applied to `fa` outermost-first as when the program is interpreted this order is reversed: the `inner` `poll` will be evaluated first, which removes the inner `uncancelable`, followed by the `outer` `poll`, which removes the outer `uncancelable` and allows cancelation. 
 
 ### Self-Cancelation
 


kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
```diff
@@ -257,8 +257,8 @@ trait MonadCancel[F[_], E] extends MonadError[F, E] {
    *
    * Masks can also be stacked or nested within each other. If multiple masks are active, all
    * masks must be undone so that cancelation can be observed. In order to completely unmask
-   * within a multi-masked region, the poll corresponding to each mask must be applied,
-   * innermost-first.
+   * within a multi-masked region the poll corresponding to each mask must be applied to the
+   * effect, outermost-first.
```
bastewart, line 261: I realise I am literally just inverting the previous docstring but IMO as far as end-users are concerned the correct order to apply the masks is "outermost-first". In the example in this docstring a few lines below (which is copied in this message), `p1` (the outer mask) is being applied first fo `fb`. I think most end-users would interpret the order of application along the lines of program construction (what they write), rather than interpretation.

Example:

```scala
F.uncancelable { p1 =>
  F.uncancelable { p2 =>
    fa *> p2(p1(fb)) *> fc
  }
}
```

I've tried to clarify the inherent ambiguity here (between program construction and interpretation) by adding "**to the effect**" to the wording.

bastewart:
@TimWSpence btw. I really struggled with how to word this! It's hard to be unambiguous in the face of program construction and interpretation being opposites (so people reading the docstring may be biased to reading it as one or the other).

bastewart:
I also believe the `canceled` docstring may be incorrect, but can't comment inline in this PR with that as I didn't change it!

```scala
  /**
   * An effect that requests self-cancelation on the current fiber.
   *
   * In the following example, the fiber requests self-cancelation in a masked region, so
   * cancelation is suppressed until the fiber is completely unmasked. `fa` will run but `fb`
   * will not.
   *
   * {{{
   *
   *   F.uncancelable { _ =>
   *     F.canceled *> fa
   *   } *> fb
   *
   * }}}
   */
  def canceled: F[Unit]
```

This indicates `fb` will not run.

The overall docstring to `MonadCancel` explains that there is no cancellation boundary immediately after `uncancelable` though:

```
 * As noted above, there are some boundaries which are not cancelable boundaries:
[snip]
 *
 * 2. The boundary after `uncancelable`
 *
 * {{{
 *   F.uncancelable(poll => foo(poll)).flatMap(f)
 * }}}
 *
 * It is guaranteed that we will not observe cancelation after `uncancelable` and hence
 * `flatMap(f)` will be invoked. This is necessary for `uncancelable` to compose. Consider for
 * example `Resource#allocated`
```

I think the example in the docstring for `canceled` may be true for a generic `MonadCancel` which implements `*>`/`productR` in a way which means the example desugars to this:

```scala
F.uncancelable { _ =>
  F.canceled *> fa
}.as { (b: B) => b }.ap(fb)
```

Presumably there is a boundary between `as` and `ap`.

`IO` itself though implements `productR` using flatmap, so at runtime this is actually:

```scala
F.uncancelable { _ =>
  F.canceled *> fa
}.flatMap(_ => fb)
```

Which has no boundary according to `MonadCancel` docstring. But, I may be totally misunderstanding...


The example could be changed to something like this to be 100% accurate (by my understanding), but it's pretty ugly:

```
   *   F.uncancelable { _ =>
   *     F.canceled *> fa
   *   }.map(identity) *> fb
```

kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
```diff
@@ -273,7 +273,8 @@ trait MonadCancel[F[_], E] extends MonadError[F, E] {
    * The following operations are no-ops:
    *
    *   1. Polling in the wrong order
-   *   1. Applying the same poll more than once: `poll(poll(fa))`
+   *   1. Subsequent polls when applying the same poll more than once: `poll(poll(fa))` is
+   *      equivalent to `poll(fa)`
```
bastewart, line 277: This is just to be really clear as well. If you're being an ultra-pedant (guilty) then you could read "applying the same poll more than once is noop" as "you must apply the poll exactly once, otherwise the fiber is still masked", rather than what I think it means which is "the first call to poll unmasks, subsequent calls do nothing"

bastewart:
Oh, I must be misunderstanding the difference between a cancellation boundary and a fiber marking itself as cancelled...

>    F.uncancelable(poll => foo(poll)).flatMap(f)
>
> It is guaranteed that we will not observe cancelation after uncancelable and hence flatMap(f) will be invoked.

Perhaps semantically if you invoke `F.canceled` inside `uncancelable` the observation of cancellation wasn't _after_ uncancelable, but during.

kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
vasilmkd: You're right, but I would actually go as far as to rewrite the whole "paragraph" (the entry sentence and the ordered list of points) to be more clear. Would you like to give that a shot?

kernel/shared/src/main/scala/cats/effect/kernel/MonadCancel.scala
vasilmkd: This is an interesting point. I would like to hear more from the other maintainers.

vasilmkd (commented):
Thanks for opening this PR. I'm sorry for falling behind on the review.

djspiewak (approved):
Sorry this kind of got buried! Merging into `series/3.3.x` so it becomes visible immediately. Thank you!

bastewart:
No worries! I had totslly forgotten about it 😬

PR #2372, opened 2021-09-24
Choose an explicit Scala dialect

vasilmkd:
This resolves the following warning:

```
Default dialect is deprecated; use explicit: [Scala211,scala212,Scala212Source3,scala213,Scala213Source3,Sbt0137,Sbt1,scala3]
Also see https://scalameta.org/scalafmt/docs/configuration.html#scala-dialects"
```

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,5 +1,7 @@
 version = 3.0.4
 
+runner.dialect = Scala213Source3
+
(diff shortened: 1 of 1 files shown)


armanbilge (approved):
Brilliant. All my projects have been complaining and now I know how to fix ;)

PR #2375, opened 2021-09-25
Discipline `ParallelF`

armanbilge:
For CI and discussion for now.

### DIFF
--- a/tests/shared/src/test/scala/cats/effect/IOSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/IOSpec.scala
@@ -16,6 +16,7 @@
 
 package cats.effect
 
+import cats.kernel.Eq
 import cats.kernel.laws.discipline.MonoidTests
 import cats.laws.discipline.{
   AlignTests,
@@ -1374,30 +1375,42 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
   }
 
   {
-    implicit val ticker = Ticker()
+    // an alley-eq
+    implicit def eqIOA[A: Eq](implicit ticker: Ticker): Eq[IO[A]] = { (x, y) =>
+      import Outcome._
+      (unsafeRun(x), unsafeRun(y)) match {
+        case (Succeeded(Some(a)), Succeeded(Some(b))) => a eqv b
+        case (Succeeded(Some(_)), _) | (_, Succeeded(Some(_))) => false
+        case _ => true
+      }
+    }
 
-    checkAll(
-      "IO.Par",
-      ParallelTests[IO, IO.Par].parallel[Int, Int]
-    )
-  }
+    {
+      implicit val ticker = Ticker()
 
-  {
-    implicit val ticker = Ticker()
+      checkAll(
+        "IO.Par",
+        ParallelTests[IO, IO.Par].parallel[Int, Int]
+      )
+    }
 
-    checkAll(
-      "IO.Par",
-      CommutativeApplicativeTests[IO.Par].commutativeApplicative[Int, Int, Int]
-    )
-  }
+    {
+      implicit val ticker = Ticker()
 
-  {
-    implicit val ticker = Ticker()
+      checkAll(
+        "IO.Par",
+        CommutativeApplicativeTests[IO.Par].commutativeApplicative[Int, Int, Int]
+      )
+    }
 
-    checkAll(
-      "IO.Par",
-      AlignTests[IO.Par].align[Int, Int, Int, Int]
-    )
+    {
+      implicit val ticker = Ticker()
+
+      checkAll(
+        "IO.Par",
+        AlignTests[IO.Par].align[Int, Int, Int, Int]
+      )
+    }
   }
 
 }
--- a/tests/shared/src/test/scala/cats/effect/IOSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/IOSpec.scala
@@ -17,7 +17,12 @@
 package cats.effect
 
 import cats.kernel.laws.discipline.MonoidTests
-import cats.laws.discipline.{AlignTests, CommutativeApplicativeTests, SemigroupKTests}
+import cats.laws.discipline.{
+  AlignTests,
+  CommutativeApplicativeTests,
+  ParallelTests,
+  SemigroupKTests
+}
 import cats.laws.discipline.arbitrary._
 
 import cats.effect.implicits._
@@ -1373,7 +1378,25 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
 
     checkAll(
       "IO.Par",
+      ParallelTests[IO, IO.Par].parallel[Int, Int]
+    )
+  }
+
+  {
+    implicit val ticker = Ticker()
+
+    checkAll(
+      "IO.Par",
       CommutativeApplicativeTests[IO.Par].commutativeApplicative[Int, Int, Int]
+    )
+  }
+
+  {
+    implicit val ticker = Ticker()
+
+    checkAll(
+      "IO.Par",
+      AlignTests[IO.Par].align[Int, Int, Int, Int]
     )
   }
 
--- a/tests/shared/src/test/scala/cats/effect/IOSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/IOSpec.scala
@@ -16,14 +16,8 @@
 
 package cats.effect
 
-import cats.kernel.Eq
 import cats.kernel.laws.discipline.MonoidTests
-import cats.laws.discipline.{
-  AlignTests,
-  CommutativeApplicativeTests,
-  ParallelTests,
-  SemigroupKTests
-}
+import cats.laws.discipline.{AlignTests, SemigroupKTests}
 import cats.laws.discipline.arbitrary._
 
 import cats.effect.implicits._
@@ -1372,45 +1366,6 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
       "IO",
       AlignTests[IO].align[Int, Int, Int, Int]
     )
-  }
-
-  {
-    // an alley-eq
-    implicit def eqIOA[A: Eq](implicit ticker: Ticker): Eq[IO[A]] = { (x, y) =>
-      import Outcome._
-      (unsafeRun(x), unsafeRun(y)) match {
-        case (Succeeded(Some(a)), Succeeded(Some(b))) => a eqv b
-        case (Succeeded(Some(_)), _) | (_, Succeeded(Some(_))) => false
-        case _ => true
-      }
-    }
-
-    {
-      implicit val ticker = Ticker()
-
-      checkAll(
-        "IO.Par",
-        ParallelTests[IO, IO.Par].parallel[Int, Int]
-      )
-    }
-
-    {
-      implicit val ticker = Ticker()
-
-      checkAll(
-        "IO.Par",
-        CommutativeApplicativeTests[IO.Par].commutativeApplicative[Int, Int, Int]
-      )
-    }
-
-    {
-      implicit val ticker = Ticker()
-
-      checkAll(
-        "IO.Par",
-        AlignTests[IO.Par].align[Int, Int, Int, Int]
-      )
-    }
   }
 
 }
--- a/tests/shared/src/test/scala/cats/effect/IOSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/IOSpec.scala
@@ -17,7 +17,7 @@
 package cats.effect
 
 import cats.kernel.laws.discipline.MonoidTests
-import cats.laws.discipline.{AlignTests, SemigroupKTests}
+import cats.laws.discipline.{AlignTests, CommutativeApplicativeTests, SemigroupKTests}
 import cats.laws.discipline.arbitrary._
 
 import cats.effect.implicits._
@@ -1365,6 +1365,15 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
     checkAll(
       "IO",
       AlignTests[IO].align[Int, Int, Int, Int]
+    )
+  }
+
+  {
+    implicit val ticker = Ticker()
+
+    checkAll(
+      "IO.Par",
+      CommutativeApplicativeTests[IO.Par].commutativeApplicative[Int, Int, Int]
     )
   }
 
--- a/kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/pure.scala
+++ b/kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/pure.scala
@@ -376,7 +376,7 @@ object pure {
         checkM.ifM(
           canceled.ifM(
             // if unmasked and canceled, finalize
-            ctx.finalizers.sequence_.as(true),
+            allocateForPureConc[E].uncancelable(_ => ctx.finalizers.sequence_.as(true)),
             // if unmasked but not canceled, ignore
             false.pure[PureConc[E, *]]
           ),
--- a/kernel/shared/src/main/scala/cats/effect/kernel/instances/GenSpawnInstances.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/instances/GenSpawnInstances.scala
@@ -74,7 +74,7 @@ trait GenSpawnInstances {
               }
 
               a <- F
-                .onCancel(poll(fiberA.join), F.both(fiberA.cancel, fiberB.cancel).void)
+                .onCancel(poll(fiberA.join), bothUnit(fiberA.cancel, fiberB.cancel))
                 .flatMap[A] {
                   case Outcome.Succeeded(fa) =>
                     fa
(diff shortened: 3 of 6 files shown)


armanbilge:
On my local I'm getting
```
[error] x commutative applicative.apply commutativity
[error]  Falsified after 2 passed tests.
[error]  > Labels of failing property:Expected: Kleisli(coop.ApplicativeThread$$anon$2$$Lambda$225/0x0000000800ddf6d8@5f6ddbaa)
[error]  Received: Kleisli(coop.ApplicativeThread$$anon$2$$Lambda$225/0x0000000800ddf6d8@65612e67)> ARG_0: Kleisli(coop.ApplicativeThread$$anon$2$$Lambda$225/0x0000000800ddf6d8@41319c86)
[error]  > ARG_1: Kleisli(cats.data.Kleisli$$$Lambda$767/0x0000000800e9ac00@7ed98197)
[error]  > ARG_2: org.scalacheck.GenArities$$Lambda$1140/0x0000000800f4eb78@3fbdb02f
[error]  The seed is K1xlqS0mCq3mpFbr5nDGHfc1Wd7tsXkF1tdwncpI-BI=
```

armanbilge:
@vasilmkd suggested I add a test for `IO.Par` as well. This also fails with:
```
[info] commutative applicative laws must hold for IO.Par
failing seed for commutative applicative.ap consistent with product + map is OFvQXMxnwvTNvIzVkH0QE8QCT6_A0x6yJB9TLtD4_LB=
[error] x commutative applicative.ap consistent with product + map
[error]  Falsified after 30 passed tests.
[error]  > ARG_1: IO(...)
[error]  The seed is OFvQXMxnwvTNvIzVkH0QE8QCT6_A0x6yJB9TLtD4_LB=
[info] + commutative applicative.applicative homomorphism
[info] + commutative applicative.applicative identity
[info] + commutative applicative.applicative interchange
[info] + commutative applicative.applicative map
[info] + commutative applicative.applicative unit
failing seed for commutative applicative.apply commutativity is zOLgZmdaocdJq0DVDT-OrY9MJBaH2EO69eWWPZKm-0D=
[error] x commutative applicative.apply commutativity
[error]  Falsified after 4 passed tests.
[error]  > ARG_1: IO(...)
[error]  > ARG_2: org.scalacheck.GenArities$$Lambda$2753/0x000000080110ca90@67c544f4
[error]  The seed is zOLgZmdaocdJq0DVDT-OrY9MJBaH2EO69eWWPZKm-0D=
failing seed for commutative applicative.apply composition is ZDFAgHiDvLP6GNdmZiIw71m9Jg02OrHrX47VegRJK8M=
[error] x commutative applicative.apply composition
[error]  Falsified after 30 passed tests.
[error]  > ARG_1: IO(...)
[error]  > ARG_2: IO(...)
[error]  The seed is ZDFAgHiDvLP6GNdmZiIw71m9Jg02OrHrX47VegRJK8M=
[info] + commutative applicative.covariant composition
[info] + commutative applicative.covariant identity
[info] + commutative applicative.invariant composition
[info] + commutative applicative.invariant identity
failing seed for commutative applicative.map2/map2Eval consistency is scYANUhQkt4GtsBNNxGzLQ3UNyDVtLos70HBG8n3sQM=
[error] x commutative applicative.map2/map2Eval consistency
[error]  Falsified after 1 passed tests.
[error]  > ARG_1: IO(...)
[error]  > ARG_2: org.scalacheck.GenArities$$Lambda$2753/0x000000080110ca90@4d3f7e99
[error]  The seed is scYANUhQkt4GtsBNNxGzLQ3UNyDVtLos70HBG8n3sQM=
failing seed for commutative applicative.map2/product-map consistency is jUQjz2NFt2r391jCyymnf5L8IHGrOGZk_1mkyHexQYP=
[error] x commutative applicative.map2/product-map consistency
[error]  Falsified after 22 passed tests.
[error]  > ARG_1: IO(...)
[error]  > ARG_2: org.scalacheck.GenArities$$Lambda$2753/0x000000080110ca90@58bef50e
[error]  The seed is jUQjz2NFt2r391jCyymnf5L8IHGrOGZk_1mkyHexQYP=
[info] + commutative applicative.monoidal left identity
[info] + commutative applicative.monoidal right identity
failing seed for commutative applicative.productL consistent map2 is dCpn2Lwfr7RIGeIyMjn2sDJyifnwwdIQ3AGWiXTYGkN=
[error] x commutative applicative.productL consistent map2
[error]  Falsified after 49 passed tests.
[error]  > ARG_1: IO(...)
[error]  The seed is dCpn2Lwfr7RIGeIyMjn2sDJyifnwwdIQ3AGWiXTYGkN=
failing seed for commutative applicative.productR consistent map2 is 99NbRWtqmqhYc4w6PFH9BBwS25eHjH0clTeXT4aegMF=
[error] x commutative applicative.productR consistent map2
[error]  Falsified after 18 passed tests.
[error]  > ARG_1: IO(...)
[error]  The seed is 99NbRWtqmqhYc4w6PFH9BBwS25eHjH0clTeXT4aegMF=
failing seed for commutative applicative.semigroupal associativity is DrNpl0MlZ9cWFZSX-_KYFD7hdXeZcoYFjWEs59YR4hG=
[error] x commutative applicative.semigroupal associativity
[error]  Falsified after 47 passed tests.
[error]  > ARG_1: IO(...)
[error]  > ARG_2: IO(...)
[error]  The seed is DrNpl0MlZ9cWFZSX-_KYFD7hdXeZcoYFjWEs59YR4hG=
```

vasilmkd:
Not only that, the CI seems to be non-terminating now.

armanbilge:
Just pushed additional tests for `Parallel` and `Align` laws at @vasilmkd's suggestion. They seem to pass for the pure implementation but `IO.Par` fails `Align` with:
```
[info] align laws must hold for IO.Par
failing seed for align.align associativity is BDXwrD6wY1aVopzMx0DIekKqYvehKmI0tr2zS7Nma5L=
[error] x align.align associativity
[error]  Falsified after 45 passed tests.
[error]  > ARG_1: IO(...)
[error]  > ARG_2: IO(...)
[error]  The seed is BDXwrD6wY1aVopzMx0DIekKqYvehKmI0tr2zS7Nma5L=
[info] + align.align homomorphism
[info] + align.alignWith consistent
[info] Total for specification IOSpec
```

armanbilge:
~Switching from `Throwable` to `Int` gets this law passing for the pure spec:~
```
+ commutative applicative.apply commutativity
```

Then it hangs on my local.

EDIT: take it back, that was just a fluke. Still fails with:
```
failing seed for commutative applicative.apply commutativity is Ob_ZmPiZEhR7tDkTSM8oniYk5MZREcyYMhTz1PdbbpP=
[error] x commutative applicative.apply commutativity
[error]  Falsified after 50 passed tests.
[error]  > Labels of failing property:Expected: Kleisli(coop.ApplicativeThread$$anon$2$$Lambda$211/0x0000000800dccbd0@b64763a)
[error]  Received: Kleisli(coop.ApplicativeThread$$anon$2$$Lambda$211/0x0000000800dccbd0@2a17c853)> ARG_0: Kleisli(cats.data.Kleisli$$$Lambda$785/0x0000000800eb4d48@33264062)
[error]  > ARG_1: Kleisli(coop.ApplicativeThread$$anon$2$$Lambda$211/0x0000000800dccbd0@4aaa5800)
[error]  > ARG_2: org.scalacheck.GenArities$$Lambda$1147/0x0000000800f75d78@62a6251a
[error]  The seed is Ob_ZmPiZEhR7tDkTSM8oniYk5MZREcyYMhTz1PdbbpP=
```

djspiewak:
Not sure why this would be hanging, though if we think carefully I'm sure we could come up with an explanation. Probably has something to do with fairness in `TestContext`. But more importantly, this type of testing is going to be fundamentally non-deterministic, I'm afraid, and for the same reason that we remove `IO.race` from the generators. The argument goes something like this…

`race` (and by extension, `par`) *fundamentally* reifies execution time as a value. In particular, if the left takes longer to run than the right, then you get `Right`. This in turn breaks *every* substitutability assumption we have which underlies functional programming and equational reasoning. Consider the functor identity law:

```scala
def identity[A](fa: F[A]) =
  fa.map(identity) <-> fa
```

This is extremely intuitive, and if you think about it, nearly all programs implicitly rely on this identity in some form. The only problem is that it's a lie:

```scala
val left = 0.until(10000).foldLeft(IO.unit)((seed, _) => seed.map(identity))
val right = IO.unit

IO.race(left, right)      // will effectively always yield Right
```

Even more concerningly, imagine if `right` were something more complex, where `IO.unit` *would* normally run faster, but now that we're layering 10,000 `map`s onto it, suddenly `right` pulls ahead. You see the problem. We've broken our equational reasoning within the bounds of this program because we have a combinator (`race`) which can observe the fact that our assumptions are, in fact, lies. The assumption being that we execute our programs on an infinitely fast CPU with infinitely sized memory and stack.

**This limitation is both fundamental and very much expected.**

If you think about it, "fixing" `race` (and `par`) so that they don't expose this issue would result in neither of them actually being useful! We *want* to get some visibility into the real execution time of our programs, and we *want* them to run in parallel, because that's literally what is needed to be practically useful. Thus, despite the fact that `IO.race` literally destroys equational reasoning, we give it a pass because it's completely, unavoidably, useful.

Note that `IO.monotonic` actually has the same issue, we just don't see it in the laws because of how they're tested:

```scala
val left = 0.until(10000).foldLeft(IO.unit)((seed, _) => seed.map(identity))
val right = IO.unit

(IO.monotonic, left, IO.montonic).mapN((s, _, e) => e - s)
(IO.monotonic, right, IO.montonic).mapN((s, _, e) => e - s)
```

The laws say these two programs should return exactly the same value. Of course, the actual results will be dramatically different. No one is claiming that we should fake equational reasoning back into `monotonic` or `realTime`, and so here we are.

There are ways to adjust the definitions of substitutability such that lawfulness is restored. It starts to get really tricky to do without generating scenarios in which the definitions are simply meaningless, and in *most* attempts to do this, the resulting equivalence relation is at minimum undecidable. Since Discipline attempts to actually *run* the equivalence relation, we clearly can't benefit from any of this ivory tower theory, and I see no way to bring things back down to executable earth.

Which is why we don't law test `ParallelF` or *any* program containing `IO.racePair`.

armanbilge:
Thanks for the detailed write-up! I have yet to read it, but I wanted to quicky mention—

> Not sure why this would be hanging, though if we think carefully I'm sure we could come up with an explanation.

I was able to minimize what causes `PureConc` to hang to:
```scala
  type F[A] = PureConc[Unit, A]
  val F = GenConcurrent[F]
  val fa: F[String] = F.pure("a") 
  val fb: F[String] = F.pure("b")
  val fc: F[Unit] = F.raiseError[Unit](())
  run(ParallelF.value(ParallelF(fa).product(ParallelF(fb)).product(ParallelF(fc))))
```

It somehow gets into a situation where the round robin queue never empties, due to `racePair` being called in an infinite loop.

vasilmkd:
Yeah, I'm pretty sure the other PR (simpler map2 for parallel) also shows a bug in the pureconc interpreter. We should dig into it.

armanbilge:
Ok, thanks, I read through the writeup and I'll have to read it at least a few more times, before it completely sinks in. Thanks again for writing it all out nicely.

> But more importantly, this type of testing is going to be fundamentally non-deterministic, I'm afraid

I understand that this is true for `IO`, but for `PureConc` ... really??

> There are ways to adjust the definitions of substitutability such that lawfulness is restored. It starts to get really tricky to do without generating scenarios in which the definitions are simply meaningless, and in most attempts to do this, the resulting equivalence relation is at minimum undecidable.

_Intuitively_, it still seems to me that we should _at least_ be able to pass the laws in the happy path (no cancellation or errors). So either I need to convince myself that we _are_, or realize why we can't be strictly lawful, even in the happy path.

djspiewak:
> I understand that this is true for IO, but for PureConc ... really??

Ah! This is an exciting addendum, but yes even for `PureConc`. The reason is subtle and has to do with other yielding mechanisms that *should* be identity suspensions. But basically, take everything I said above about `IO.race`, and instead of "CPU time" substitute "relative number of yield points". If you can contrive a scenario in which the `left` has more `cede`s than the right, then the right will finish first.

armanbilge:
Bringing in #2403 gets us this result for `ParallelF[PureConc]`:

```
[info] ParallelFSpec
[info] parallel laws must hold for Parallel[F, ParallelF]
[info] + parallel.isomorphic functor
[info] + parallel.isomorphic pure
[info] + parallel.parallel round trip
[info] + parallel.sequential round trip
[info] commutative applicative laws must hold for CommutativeApplicative[ParallelF]
[info] + commutative applicative.ap consistent with product + map
[info] + commutative applicative.applicative homomorphism
failing seed for commutative applicative.apply composition is t3EvoGz8tjGrbg_IiQSN1w9Pt0091Y125qKP9UpCbCA=
[info] + commutative applicative.applicative identity
[info] + commutative applicative.applicative interchange
[info] + commutative applicative.applicative map
[info] + commutative applicative.applicative unit
[info] + commutative applicative.apply commutativity
[error] x commutative applicative.apply composition
[error]  Falsified after 1 passed tests.
[error]  > Labels of failing property:Expected: Kleisli(coop.ApplicativeThread$$anon$2$$Lambda$207/0x0000000800dd8a70@6053ede1)
[error]  Received: Kleisli(coop.ApplicativeThread$$anon$2$$Lambda$207/0x0000000800dd8a70@2e22da25)> ARG_0: Kleisli(coop.ApplicativeThread$$anon$2$$Lambda$207/0x0000000800dd8a70@36e8bc83)
[error]  > ARG_1: Kleisli(cats.data.KleisliFunctions$$Lambda$158/0x0000000800d7b558@f6d0379)
[error]  > ARG_2: Kleisli(cats.data.Kleisli$$$Lambda$814/0x0000000800ebea48@ca21ab2)
[error]  The seed is t3EvoGz8tjGrbg_IiQSN1w9Pt0091Y125qKP9UpCbCA=
[info] + commutative applicative.covariant composition
[info] + commutative applicative.covariant identity
[info] + commutative applicative.invariant composition
[info] + commutative applicative.invariant identity
[info] + commutative applicative.map2/map2Eval consistency
failing seed for commutative applicative.semigroupal associativity is bB9LO0Nqy8NK3aPRV3UGSh17DxXcRgqkt6bfCfHepJB=
[info] + commutative applicative.map2/product-map consistency
[info] + commutative applicative.monoidal left identity
[info] + commutative applicative.monoidal right identity
[info] + commutative applicative.productL consistent map2
[info] + commutative applicative.productR consistent map2
[error] x commutative applicative.semigroupal associativity
[error]  Falsified after 16 passed tests.
[error]  > Labels of failing property:Expected: Kleisli(cats.data.Kleisli$$$Lambda$814/0x0000000800ebea48@8470373)
[error]  Received: Kleisli(cats.data.Kleisli$$$Lambda$814/0x0000000800ebea48@400f2184)> ARG_0: Kleisli(coop.ApplicativeThread$$anon$2$$Lambda$207/0x0000000800dd8a70@4d0afde9)
[error]  > ARG_1: Kleisli(coop.ApplicativeThread$$anon$2$$Lambda$207/0x0000000800dd8a70@5075bcf3)
[error]  > ARG_2: Kleisli(cats.data.Kleisli$$$Lambda$814/0x0000000800ebea48@60197115)
[error]  The seed is bB9LO0Nqy8NK3aPRV3UGSh17DxXcRgqkt6bfCfHepJB=
[info] align laws must hold for Align[ParallelF]
[info] + align.align associativity
[info] + align.align homomorphism
[info] + align.alignWith consistent
[info] Total for specification ParallelFSpec
[info] Finished in 5 seconds, 513 ms
[info] 26 examples, 2404 expectations, 2 failures, 0 error
```

armanbilge:
Still trying to understand the failed laws. Focusing on `commutative applicative.apply composition` to start. Here's a couple instances of failure:
1. Non-equal error outcomes. This is not unreasonable I think, so from this point I'm using `Unit` and the error type (i.e., all errors are equivalent).
2. `Succeeded(None)` and `Errored(())`. This is more troubling. Shouldn't errors always short-circuit? Edit: I think I was confused. Does `Succeeded(None)` represent both non-termination and cancellation??

tests/shared/src/test/scala/cats/effect/IOSpec.scala
```diff
@@ -1368,4 +1374,43 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
...
+  {
+    // an alley-eq
+    implicit def eqIOA[A: Eq](implicit ticker: Ticker): Eq[IO[A]] = { (x, y) =>
+      import Outcome._
+      (unsafeRun(x), unsafeRun(y)) match {
+        case (Succeeded(Some(a)), Succeeded(Some(b))) => a eqv b
+        case (Succeeded(Some(_)), _) | (_, Succeeded(Some(_))) => false
+        case _ => true
+      }
+    }
```
vasilmkd: Can you please define and encapsulate this code in a separate `IOParSpec` file?

armanbilge:
> There are ways to adjust the definitions of substitutability such that lawfulness is restored. It starts to get really tricky to do without generating scenarios in which the definitions are simply meaningless, and in most attempts to do this, the resulting equivalence relation is at minimum undecidable. Since Discipline attempts to actually run the equivalence relation, we clearly can't benefit from any of this ivory tower theory, and I see no way to bring things back down to executable earth.

@djspiewak do you feel this way about the relaxed equivalence I've introduced here?

djspiewak (approved):
I think I'm okay with this given that it's specifically `IOPar` that we're looking at, and we haven't allowed the `Eq` to leak out. I'll write a follow-up PR explaining why exactly the `Eq` needs to be error-oblivious.

PR #2376, opened 2021-09-27
Radically slimmer parallel map2

vasilmkd:
@armanbilge Please read the comments carefully and comment whether you agree with them. I may have overlooked something, it's very late and I apologize in advance.

The tests seem to be passing (the ones that we have in `series/3.x` in any case) with these changes, so this should be "compatible" with the previous implementation.

However, if you rebase the laws checks that you added about ParallelF on this branch, you will see the `???` paths being taken, and I believe this is a bug in the pure interpreter. I haven't tried the IO.Par laws.

PR #2379, opened 2021-09-27
Fix `ApplicativeForParallelF` and friends

armanbilge:
The first goal is getting the `ParallelFSpec` introduced in #2375 to pass, which is accomplished in 6032cf2. 

Note I relaxed the tests from `CommutativeApplicative` to just `Applicative` since atm I do not believe commutativity is feasible without an additional constraint on `E`, h/t to @Baccata for pointing this out.

Currently laws are still not passing for `IO.Par` and in addition these changes also fail the `parallel should short-circuit on error` in `IOSpec`. I don't think it's possible to pass the laws for `Applicative` _and_ short-circuit both the left and right. See also https://github.com/typelevel/cats-effect/pull/2376#discussion_r716312564. Possible compromises:
1. Only short-circuit the right, if an error occurs on the left. Don't short-circuit the left even if the right errors. This effectively left-biases the error channel.
2. Relax the equivalence of errors to something like "all errors are equivalent".

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/instances/GenSpawnInstances.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/instances/GenSpawnInstances.scala
@@ -58,39 +58,14 @@ trait GenSpawnInstances {
               fiberA <- F.start(ParallelF.value(fa).attempt)
               fiberB <- F.start(ParallelF.value(fb).attempt)
 
-              // start a pair of supervisors to ensure that the opposite is canceled on error
-              // _ <- F start {
-              //   fiberB.join flatMap {
-              //     case Outcome.Succeeded(_) => F.unit
-              //     case _ => fiberA.cancel
-              //   }
-              // }
-
-              // _ <- F start {
-              //   fiberA.join flatMap {
-              //     case Outcome.Succeeded(_) => F.unit
-              //     case _ => fiberB.cancel
-              //   }
-              // }
-
               a <- F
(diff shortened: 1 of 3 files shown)


armanbilge:
So far, whenever I attempt to re-introduce (left-biased) short-circuiting I end up with a `ParallelFSpec` that hangs, just like it currently does in #2375. In #2376 @vasilmkd suggested there could be a bug in `PureConc`, although in a different context.

vasilmkd:
Unfortunately, left-biasing the error (i.e. allowing the left to always continue until it succeeds or errors is not feasible. The test that fails in `PureConcSpec` is necessary from a resource safety perspective. I don't know what that means about the laws, unfortunately.

armanbilge:
> The test that fails in PureConcSpec

Darn, I didn't realize there was another test these changes would affect. Thanks for pointing that out.

> is necessary from a resource safety perspective.

Hmm, I can imagine why but I'll have to think more about this to properly understand it.

> Unfortunately, left-biasing the error (i.e. allowing the left to always continue until it succeeds or errors is not feasible.

Even allowing for cancellation?

armanbilge:
> I don't know what that means about the laws, unfortunately.

FWIW I think it may totally make sense to have this as a so-called alley cat. Especially if all it takes to make it lawful is relaxing the equivalence of errors or something along those lines. I've played with this but still haven't gotten it to work though so there may still be some other nuance I'm not getting.

armanbilge:
Superseded by #2403 and #2375.

PR #2382, opened 2021-10-01
Adds weaver-test paragraph to getting-started

Baccata:
As per @djspiewak's request, this adds weaver to the "testing" section of the getting-started documentation. 

It also replaces hardcoded versions of third-parties with scaladex badges (maybe I should keep the versions to facilitate copy paste though, now that I think about it...).

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -39,7 +39,7 @@ import scala.concurrent.duration._
 
 // obviously this isn't actually the problem definition, but it's kinda fun
 object StupidFizzBuzz extends IOApp.Simple {
-  val run = 
+  val run =
     for {
       ctr <- IO.ref(0)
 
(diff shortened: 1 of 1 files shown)


vasilmkd (dismissed):

vasilmkd:
Regarding the versions, I kind of prefer having a specific version that is easily copy paste-able, but I understand that's difficult to maintain. I will leave the decision to others. (I also wouldn't mind if this PR just bumps the versions to the latest ones).

Baccata:
I mean, that's a good point. Bumped versions in 9acec2c . I left the badges because the information is valuable imho.

vasilmkd (dismissed):

docs/getting-started.md
```diff
@@ -96,9 +100,40 @@ class ExampleSuite extends CatsEffectSuite {
 }
 ```
 
+### Weaver-test
+
+[![weaver-cats Scala version support](https://index.scala-lang.org/disneystreaming/weaver-test/weaver-cats/latest-by-scala-version.svg)](https://index.scala-lang.org/disneystreaming/weaver-test/weaver-cats)
+
+[Weaver](https://github.com/disneystreaming/weaver-test) is a test-framework build directly on top of cats-effect. It is designed specifically to handle thousands of tests exercising I/O layers (http, database calls) concurrently. Weaver makes heavy use of the concurrency constructs and abstractions provided by cats-effect to safely share resources (clients) across tests and suite, and runs all tests in parallel by default.
```
armanbilge:
```suggestion
[Weaver](https://github.com/disneystreaming/weaver-test) is a test-framework built directly on top of Cats Effect. It is designed specifically to handle thousands of tests exercising I/O layers (http, database calls) concurrently. Weaver makes heavy use of the concurrency constructs and abstractions provided by cats-effect to safely share resources (clients) across tests and suite, and runs all tests in parallel by default.
```

armanbilge (dismissed):
Why aren't I using weaver? 😆

docs/getting-started.md
vasilmkd: Also, cats-effect in text should be Cats Effect.

Baccata: 7b93721

Baccata:
> Why aren't I using weaver? 

you tell me 😄

vasilmkd (approved):

PR #2390, opened 2021-10-05
Add support for long sleeps in JS scheduler

### LINKED ISSUE #2388: JS `TimeoutOverflowWarning`
Reported by @tpolecat when trying to set a timeout to `Long.MaxValue.ns`

```
(node:7170) TimeoutOverflowWarning: 9223372036854 does not fit into a 32-bit signed integer.
Timeout duration was set to 1.
```

Apparently the maximum timeout on JS is 2<sup>31</sup>-1 ms. See https://stackoverflow.com/questions/12633405/what-is-the-maximum-delay-for-setinterval

One workaround from here: https://stackoverflow.com/a/62002839/1642693

Essentially, to schedule a task that schedules another task and so on, such that the delays add up to the desired duration.

At the very least, we should round down to the maximum possible value instead of to 1 :)

armanbilge:
Fixes #2388. Also removes a duplicate (?) implementation of the default scheduler.

### DIFF
--- a/core/js/src/main/scala/cats/effect/unsafe/SchedulerCompanionPlatform.scala
+++ b/core/js/src/main/scala/cats/effect/unsafe/SchedulerCompanionPlatform.scala
@@ -16,17 +16,24 @@
 
 package cats.effect.unsafe
 
-import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.duration._
 import scala.scalajs.js.timers
 
 private[unsafe] abstract class SchedulerCompanionPlatform { this: Scheduler.type =>
   def createDefaultScheduler(): (Scheduler, () => Unit) =
     (
       new Scheduler {
-        def sleep(delay: FiniteDuration, task: Runnable): Runnable = {
-          val handle = timers.setTimeout(delay)(task.run())
-          () => timers.clearTimeout(handle)
-        }
+        private[this] val maxTimeout = Int.MaxValue.millis
+
+        def sleep(delay: FiniteDuration, task: Runnable): Runnable =
+          if (delay <= maxTimeout) {
+            val handle = timers.setTimeout(delay)(task.run())
+            () => timers.clearTimeout(handle)
+          } else {
+            var cancel: Runnable = () => ()
+            cancel = sleep(maxTimeout, () => cancel = sleep(delay - maxTimeout, task))
+            () => cancel.run()
+          }
 
         def nowMillis() = System.currentTimeMillis()
         def monotonicNanos() = System.nanoTime()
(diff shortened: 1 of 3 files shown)


core/js/src/main/scala/cats/effect/unsafe/SchedulerCompanionPlatform.scala
```diff
@@ -16,17 +16,24 @@
...
 
 private[unsafe] abstract class SchedulerCompanionPlatform { this: Scheduler.type =>
   def createDefaultScheduler(): (Scheduler, () => Unit) =
     (
       new Scheduler {
-        def sleep(delay: FiniteDuration, task: Runnable): Runnable = {
-          val handle = timers.setTimeout(delay)(task.run())
-          () => timers.clearTimeout(handle)
-        }
+        private[this] val maxTimeout = Int.MaxValue.millis
```
vasilmkd: Should we move this on the companion object itself? Above in the class body.

core/js/src/main/scala/cats/effect/unsafe/SchedulerCompanionPlatform.scala
```diff
@@ -16,17 +16,24 @@
...
+
+        def sleep(delay: FiniteDuration, task: Runnable): Runnable =
+          if (delay <= maxTimeout) {
+            val handle = timers.setTimeout(delay)(task.run())
+            () => timers.clearTimeout(handle)
+          } else {
+            var cancel: Runnable = () => ()
+            cancel = sleep(maxTimeout, () => cancel = sleep(delay - maxTimeout, task))
```
vasilmkd, line 35: Sorry, can you explain exactly what this line does?

armanbilge: We create a mutable reference to store the cancellation `Runnable` (initialized to a dummy). Then we schedule a task to occur after the max natively supported timeout and store its cancellation `Runnable`. The scheduled task itself schedules the target task with the remaining sleep time, and updates the reference with the new cancellation `Runnable`.

If that makes sense 😅

vasilmkd: That's just awesome. 😆

vasilmkd (approved):

core/js/src/main/scala/cats/effect/unsafe/SchedulerCompanionPlatform.scala
djspiewak: …holy shit

djspiewak (approved):

Issue #1992, opened 2021-05-24
Kill IOApp when SIGINT repeatedly


djspiewak:
At present, `IOApp` waits for finalizers to complete before allowing exit on <kbd>Ctrl</kbd>-<kbd>C</kbd>. This is generally a good default, but it should be overridable without having to resort to `SIGKILL` if the application is hung. Something like three INT signals within a one second time window should trigger an immediate shutdown, regardless of finalizer status.

rossabaker:
Wait, what?  <kbd>Ctrl</kbd>-<kbd>C</kbd>shouldn't change behavior just because we're angry.  This is why we have different signals that yield different exit codes.

djspiewak:
Interested as to thoughts on this. The scenario here is that the user wants to kill the application, but a finalizer (or uncancelable block) is stuck and thus the application cannot be terminated. The point is that rapidly repeated TERM is likely to indicate direct user input and is almost certainly a signal that they don't care about graceful release, they just want the process to die.

rossabaker:
Well, I guess that's fair for `SIGINT`, which is what <kbd>Ctrl</kbd>-<kbd>C</kbd> sends.  You're only communicating an intent to interrupt the process, and I suppose there's room to detect user rage.  But `SIGTERM`, the default signal of `kill`, ought to remain graceful.

Objection withdrawn.

aeons:
Terraform attempts graceful shutdown on the first ctrl-c and forcefully aborts on the second, with a warning about what might happen. I think that's a pretty good way of doing it.

djspiewak:
> Well, I guess that's fair for SIGINT, which is what Ctrl-C sends. You're only communicating an intent to interrupt the process, and I suppose there's room to detect user rage. But SIGTERM, the default signal of kill, ought to remain graceful.

In which I betray the fact that I always forget which signal is which.

ChristopherDavenport:
In case others needed reference like I just needed to follow this conversation: 

Gnu Termination Signals - https://www.gnu.org/software/libc/manual/html_node/Termination-Signals.html

rossabaker:
My CE2 `IOApp` experience is coming back to me.  Even though the JVM exits with the expected code (signal value + 128), there's no way to distinguish `SIGTERM` (which should be graceful) from `SIGINT` within the application.  The only way (at least on Java 8) was JNI quagmire.

[...]

PR #2394, opened 2021-10-06
Add a configurable timeout for the JVM shutdown hook

### LINKED ISSUE #746: Deadlock in IOApp when calling System.exit
```
object ApiStarter extends IOApp {
  def run(args: List[String]): IO[ExitCode] = Resource.make(IO(System.exit(0)))(_ => IO(())).use { _ =>
    ???
  }
}
```

Hangs forever, clean exit expected.
It seems that it deadlocks on finalizer in IOAppPlatform.

cats-effect 2.0.0
cats 2.0.0

vasilmkd:
Fixes #746 for Cats Effect 3.
Adds an IMO safer alternative to #1992 (by not having to depend on `sun.misc` APIs _and_ providing configuration).

A little bit of background: `Ctrl+C` or `System.exit` on an `IOApp` with an uncancelable main fiber results in an apparent deadlock due to cancelation backpressure. If you want to try it, run the following:

```scala
  val run: IO[Unit] =
    IO(System.exit(0)).uncancelable
}
```

This PR does a few things:
1. It introduces `shutdownHookTimeout` as a configurable option through `-Dcats.effect.shutdown.hook.timeout` and through `IORuntimeConfig`. The diff also includes the usual binary compatibility dance.
2. `IOApp` uses this configuration option to execute a `CountDownLatch.await()` with a timeout in the `io-cancel-hook` thread. IMO, this behavior restores user expectation when they hit `Ctrl+C` or execute `System.exit`, and the application waits forever for an uncancelable fiber to be canceled, and they are free to configure it to match their application needs.
3. I removed the `join`ing of `WorkerThread` instances in the `WorkStealingThreadPool#shutdown` method. We still interrupt the threads. The reason why `join` doesn't "work" in this case is because `System.exit` is an uninterruptible method, so `join`ing a `WorkerThread` that executed `System.exit` looked like the application was hanging, which is bad user experience.
4. The PR as I'm opening has a **BAD** (again, IMO) default for `shutdownHookTimeout`. I set it to `Duration.Inf`, because I wanted to match the previous behavior, but I think we should put it to something like 10 seconds (I swear I have seen other software do this, allow the application to gracefully shutdown within 10 seconds and then forcefully terminate it, but I cannot seem to find a source for this, I'm sorry).
5. I marked a couple of `IORuntimeConfig` _constructors_ (not `apply` methods) as deprecated, because I intend to _remove_ them in a future release, since they were never meant to be public anyway (the default constructor is private).
6. Added a unit test that exercises the new functionality with a `Duration.Zero` timeout.

Interested to hear thoughts on this.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/unsafe/IORuntimeConfig.scala
+++ b/core/shared/src/main/scala/cats/effect/unsafe/IORuntimeConfig.scala
@@ -17,18 +17,22 @@
 package cats.effect
 package unsafe
 
+import scala.concurrent.duration._
+
 final case class IORuntimeConfig private (
     val cancelationCheckThreshold: Int,
     val autoYieldThreshold: Int,
     val enhancedExceptions: Boolean,
-    val traceBufferSize: Int) {
+    val traceBufferSize: Int,
+    val shutdownHookTimeout: Duration) {
 
   def this(cancelationCheckThreshold: Int, autoYieldThreshold: Int) =
     this(
       cancelationCheckThreshold,
       autoYieldThreshold,
       IORuntimeConfig.DefaultEnhancedExceptions,
-      IORuntimeConfig.DefaultTraceBufferSize)
+      IORuntimeConfig.DefaultTraceBufferSize,
+      IORuntimeConfig.DefaultShutdownHookTimeout)
 
   def copy(
       cancelationCheckThreshold: Int = this.cancelationCheckThreshold,
@@ -39,7 +43,8 @@ final case class IORuntimeConfig private (
       cancelationCheckThreshold,
       autoYieldThreshold,
       enhancedExceptions,
-      traceBufferSize)
+      traceBufferSize,
+      shutdownHookTimeout)
 
   // shim for binary compat
   private[unsafe] def copy(
@@ -49,7 +54,8 @@ final case class IORuntimeConfig private (
       cancelationCheckThreshold,
       autoYieldThreshold,
       enhancedExceptions,
-      traceBufferSize)
+      traceBufferSize,
+      shutdownHookTimeout)
 
   private[effect] val traceBufferLogSize: Int =
     Math.round(Math.log(traceBufferSize.toDouble) / Math.log(2)).toInt
@@ -60,6 +66,7 @@ object IORuntimeConfig extends IORuntimeConfigCompanionPlatform {
   // these have to be defs because we forward-reference them from the companion platform
   private[unsafe] def DefaultEnhancedExceptions = true
   private[unsafe] def DefaultTraceBufferSize = 16
+  private[unsafe] def DefaultShutdownHookTimeout = Duration.Inf
 
   def apply(): IORuntimeConfig = Default
 
@@ -68,19 +75,22 @@ object IORuntimeConfig extends IORuntimeConfigCompanionPlatform {
       cancelationCheckThreshold,
       autoYieldThreshold,
       DefaultEnhancedExceptions,
-      DefaultTraceBufferSize)
+      DefaultTraceBufferSize,
+      DefaultShutdownHookTimeout)
 
   def apply(
       cancelationCheckThreshold: Int,
       autoYieldThreshold: Int,
       enhancedExceptions: Boolean,
-      traceBufferSize: Int): IORuntimeConfig = {
+      traceBufferSize: Int,
+      shutdownHookTimeout: Duration): IORuntimeConfig = {
     if (autoYieldThreshold % cancelationCheckThreshold == 0)
       new IORuntimeConfig(
         cancelationCheckThreshold,
         autoYieldThreshold,
         enhancedExceptions,
-        1 << Math.round(Math.log(traceBufferSize.toDouble) / Math.log(2)).toInt)
+        1 << Math.round(Math.log(traceBufferSize.toDouble) / Math.log(2)).toInt,
+        shutdownHookTimeout)
     else
       throw new AssertionError(
         s"Auto yield threshold $autoYieldThreshold must be a multiple of cancelation check threshold $cancelationCheckThreshold")
--- a/core/shared/src/main/scala/cats/effect/unsafe/IORuntimeConfig.scala
+++ b/core/shared/src/main/scala/cats/effect/unsafe/IORuntimeConfig.scala
@@ -26,10 +26,7 @@ final case class IORuntimeConfig private (
     val traceBufferSize: Int,
     val shutdownHookTimeout: Duration) {
 
-  @deprecated(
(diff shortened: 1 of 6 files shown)


vasilmkd:
I have no idea why the `IORuntimeConfig` companion object constructor started showing a stack overflow error on Scala.js, nor how to solve it.

armanbilge:
Ah, it's because there's a circular dependency 😆 

```scala
  def apply(): IORuntimeConfig = Default
```

```scala
  protected final val Default: IORuntimeConfig =
    IORuntimeConfig().copy(cancelationCheckThreshold = 512, autoYieldThreshold = 1024)
```

vasilmkd:
Still happens if I revert the code. That's what I don't understand. Also happens if I completely switch to using a constructor with `new`.

vasilmkd:
In either case, we're going to have a merge conflict with JS tracing configuration, so we'll solve it then.

armanbilge:
Huh, really? I was able to fix it when I messed with it earlier. Anyway yes we'll see about it then :)

armanbilge:
I haven't looked in depth at your implementation here yet, but I wanted to mention another strategy: I think some apps, when they receive the first `Ctrl+C` start the cancellation process and backpressure (often with a friendly output, "we got the message just cleaning up"). But if you're impatient and slam `Ctrl+C` one or two more times then it forcefully exits. Worth considering here?

vasilmkd:
That's what was discussed in #1992, requires relying on `sun.misc.Signal`, which is internal API, and on top of that, there's a question to be answered how that works with the cancelation hooks that are currently handled by the JVM.

armanbilge:
I see, thanks, apologies I missed that. I'd be all for avoiding `sun.misc.Signal` except aren't we going to need it anyway for fiber dumps?

vasilmkd:
As far as fiber dumps go, that's a signal different than the shutdown signals and I think we should depend on a real library like https://github.com/jnr/jnr-posix to guarantee OS compatibility.

As far as shutdown signals go, I personally think messing with them will bring nothing but headaches, but we'll see what the popular vote says.

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeConfig.scala
```diff
@@ -17,39 +17,73 @@
...
     val autoYieldThreshold: Int,
     val enhancedExceptions: Boolean,
-    val traceBufferSize: Int) {
+    val traceBufferSize: Int,
+    val shutdownHookTimeout: Duration) {
 
+  @deprecated(
+    "Use IORuntimeConfig.apply(cancelationCheckThreshold, autoYieldThreshold, enhancedExceptions, traceBufferSize, shutdownHookTimeout)",
+    "3.3.0")
   def this(cancelationCheckThreshold: Int, autoYieldThreshold: Int) =
```
djspiewak: Let's just make it `private[unsafe]` like the others. We're (lightly) breaking source compat in this release anyway

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeConfig.scala
```diff
@@ -60,27 +94,49 @@ object IORuntimeConfig extends IORuntimeConfigCompanionPlatform {
...
   private[unsafe] def DefaultEnhancedExceptions = true
   private[unsafe] def DefaultTraceBufferSize = 16
+  private[unsafe] def DefaultShutdownHookTimeout = Duration.Inf
 
   def apply(): IORuntimeConfig = Default
 
+  @deprecated(
+    "Use IORuntimeConfig.apply(cancelationCheckThreshold, autoYieldThreshold, enhancedExceptions, traceBufferSize, shutdownHookTimeout)",
+    "3.3.0")
   def apply(cancelationCheckThreshold: Int, autoYieldThreshold: Int): IORuntimeConfig =
```
djspiewak: Same note as above

tests/jvm/src/test/scala/cats/effect/IOAppSpec.scala
```diff
@@ -214,4 +220,13 @@ package examples {
     def run(args: List[String]): IO[ExitCode] =
       IO.pure(ExitCode.Success)
   }
+
+  object ShutdownHookImmediateTimeout extends IOApp.Simple {
```
djspiewak, line 224: I really love this test harness…

djspiewak (commented):
So this is a really interesting change. I like the idea of making it configurable, but I don't like the idea of making it default to anything other than `Duration.Inf`. Arbitrary thresholds are really dangerous. Who's to say that 11 seconds isn't ideal? Or 9?

The *concept* of doing this exists in a bunch of places btw, most notably in Kubernetes. When a pod is being shutdown, it will be sent a `SIGTERM`, after which the scheduler will wait for a configurable grace period (default: 30 seconds) before sending `SIGKILL` and discarding the instance. Note that Cats Effect applications already behave correctly under these circumstances, cleaning up their resources as quickly as possible, but if they *can't* clean things up, correctly signaling to the scheduler that they're stuck and need to be killed. If we were to change the default shutdown timeout to something finite, we would effectively be lying to outside schedulers (like k8s) by claiming we had done a graceful shutdown when actually we had just yolo'd off the cliff.

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeConfig.scala
vasilmkd: Done.

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeConfig.scala
vasilmkd: Done.

vasilmkd:
Ok, I'm glad you like it.

vasilmkd:
Rebased and added JS configuration and as I predicted, the stack overflow issue went away without me having to do anything, so thanks @armanbilge.

vasilmkd:
@djspiewak Hiding the old constructor breaks Java binary compatibility, it is no longer available as a static method.

```
[error] cats-effect: Failed binary compatibility check against org.typelevel:cats-effect_2.13:3.1.0! Found 1 potential problems (filtered 11)
[error]  * static method apply(Int,Int)cats.effect.unsafe.IORuntimeConfig in class cats.effect.unsafe.IORuntimeConfig does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("cats.effect.unsafe.IORuntimeConfig.apply")
```

djspiewak:
Ah good point. I think that only applies to `apply` though, and we can still hide the constructor itself.

vasilmkd:
Should I just add the filter then?

djspiewak (approved):

PR #2399, opened 2021-10-09
update-doc

### LINKED ISSUE #1127: Documentation page for system properties
Since we're relying more and more on system properties to customize the behavior of `IO`, we should document them all in a dedicated doc page for discoverability.

### LINKED ISSUE #1486: Document 'non-daemon thread' debug print features
- sys property to enable/disable
- sys property to control the check interval

All configs defined in here: https://github.com/typelevel/cats-effect/blob/series/2.x/core/jvm/src/main/scala/cats/effect/internals/NonDaemonThreadLogger.scala

This can probably be a section on the `IOApp` doc page

i10416:
close #1127 
close https://github.com/typelevel/cats-effect/issues/1486 too.

### DIFF
--- a/docs/scaling-and-tuning/system-properties.md
+++ b/docs/scaling-and-tuning/system-properties.md
@@ -1,0 +1,30 @@
+---
+id: system-properties
+title: System Properties
+---
+# System properties
+
+There are some properties configurable from system properties.
+
+## JVM
+You can set these properties via JVM arguments.
+```
+-Dcats.effect.propName=propValue
+```
+
+|Property|Value(default)|Description|CE version |
+|---|---|---|---|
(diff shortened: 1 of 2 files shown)


docs/scaling-and-tuning/system-properties.md
```diff
@@ -0,0 +1,23 @@
+---
+id: system-properties
+title: System Properties
+---
+# System properties
```
djspiewak, line 5: This header is actually inserted by docusaurus, so we don't need to add it here.

djspiewak (approved):
This is great, thank you! I really want to refactor some of the documentation organization a bit so we can have a **Core Runtime** section, and I think this would fit nicely there. I'll work on that in a followup

PR #2408, opened 2021-10-13
Global GC fiber bag implementation

vasilmkd:
The global fallback for #2368. This PR does not yet contain the worker-thread-local suspended bags, that will arrive in a follow up PR.

One detail to discuss here is the fact that we're tying this bag to an `IORuntime` instance, which doesn't necessarily make it global the way the tracing cache is for example. I really do not know how we should treat `IORuntime`, but I am going to say that there might be possibilities for further optimizing cache lookups for tracing if we somehow decide that there can only ever be one `IORuntime` per JVM. Probably should be a separate discussion, but still something to think about.

### DIFF
--- a/core/js/src/main/scala/cats/effect/unsafe/SuspendedFiberBag.scala
+++ b/core/js/src/main/scala/cats/effect/unsafe/SuspendedFiberBag.scala
@@ -27,7 +27,7 @@ import scala.collection.mutable
  * a `java.util.WeakHashMap` so that the removal of resumed fibers is handled automatically, but
  * weak references are still not available in Scala.js.
  */
-private final class SuspendedFiberBag {
+private[effect] final class SuspendedFiberBag {
   private[this] val bag: mutable.Map[AnyRef, IOFiber[_]] =
     mutable.Map.empty
 
--- a/core/js/src/main/scala/cats/effect/unsafe/SuspendedFiberBag.scala
+++ b/core/js/src/main/scala/cats/effect/unsafe/SuspendedFiberBag.scala
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2020-2021 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+package unsafe
+
+import scala.collection.mutable
+
+/**
+ * A simple implementation of an unordered bag used for tracking asynchronously suspended fiber
+ * instances on Scala.js. This bag is backed by a mutable hash map and delegates all
+ * functionality to it. Because Scala.js runs in a single-threaded environment, there's no need
+ * for any synchronization. Ideally, we would have liked the Scala.js implementation to also use
+ * a `java.util.WeakHashMap` so that the removal of resumed fibers is handled automatically, but
+ * weak references are still not available in Scala.js.
+ */
+private final class SuspendedFiberBag {
+  private[this] val bag: mutable.Map[AnyRef, IOFiber[_]] =
+    mutable.Map.empty
+
+  /**
+   * Registers a suspended fiber, tracked by the provided key which is an opaque object which
+   * uses reference equality for comparison.
+   *
+   * @param key
+   *   an opaque identifier for the suspended fiber
+   * @param fiber
+   *   the suspended fiber to be registered
+   */
+  def monitor(key: AnyRef, fiber: IOFiber[_]): Unit = {
+    bag(key) = fiber
+  }
+
+  /**
+   * Deregisters a resumed fiber, tracked by the provided key which is an opaque object which
+   * uses reference equality for comparison.
+   *
+   * @param key
+   *   an opaque identifier for the resumed fiber
+   */
+  def unmonitor(key: AnyRef): Unit = {
+    bag -= key
+  }
+}
--- a/core/js/src/main/scala/cats/effect/unsafe/SuspendedFiberBag.scala
+++ b/core/js/src/main/scala/cats/effect/unsafe/SuspendedFiberBag.scala
@@ -17,19 +17,13 @@
 package cats.effect
 package unsafe
 
-import scala.collection.mutable
+import scala.annotation.nowarn
 
 /**
- * A simple implementation of an unordered bag used for tracking asynchronously suspended fiber
- * instances on Scala.js. This bag is backed by a mutable hash map and delegates all
- * functionality to it. Because Scala.js runs in a single-threaded environment, there's no need
- * for any synchronization. Ideally, we would have liked the Scala.js implementation to also use
- * a `java.util.WeakHashMap` so that the removal of resumed fibers is handled automatically, but
- * weak references are still not available in Scala.js.
+ * A no-op implementation of an unordered bag used for tracking asynchronously suspended fiber
+ * instances on Scala.js. We will iterate on this in the future.
  */
 private[effect] final class SuspendedFiberBag {
-  private[this] val bag: mutable.Map[AnyRef, IOFiber[_]] =
-    mutable.Map.empty
 
   /**
    * Registers a suspended fiber, tracked by the provided key which is an opaque object which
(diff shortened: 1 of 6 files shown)


core/js/src/main/scala/cats/effect/unsafe/SuspendedFiberBag.scala
```diff
@@ -0,0 +1,57 @@
...
+
+import scala.collection.mutable
+
+/**
+ * A simple implementation of an unordered bag used for tracking asynchronously suspended fiber
+ * instances on Scala.js. This bag is backed by a mutable hash map and delegates all
+ * functionality to it. Because Scala.js runs in a single-threaded environment, there's no need
+ * for any synchronization. Ideally, we would have liked the Scala.js implementation to also use
+ * a `java.util.WeakHashMap` so that the removal of resumed fibers is handled automatically, but
+ * weak references are still not available in Scala.js.
```
armanbilge: Would this work?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap

djspiewak: NICE FIND

vasilmkd: One challenge with this approach is the fact that the value (`IOFiber`) holds a strong reference to the key, so that kind of prevents collection of the key. We circumvent this by using a `WeakReference` for the value (because `WeakHashMap` holds a strong reference to the value for whatever reason). `WeakReference` is not available on Scala.js. Unless this map holds weak references to the values, we cannot really use it.

vasilmkd: This is also a limitation it seems:
> Because the references are weak, WeakMap keys are not enumerable. There is no method to obtain a list of the keys. If there was, the list would depend on the state of garbage collection, introducing non-determinism. If you want to have a list of keys, you should use a Map.

armanbilge: https://www.scala-js.org/api/scalajs-library/1.7.1/scala/scalajs/js/WeakRef.html

armanbilge: As for the second limitation, I'd have to think about it. Can we keep a list of weak references to keys ourselves? It'll probably need to be cleaned up now and again.

vasilmkd: You're the JS master, but is targeting just ECMA Script 2021 feasible?

armanbilge: Anyway, after this PR lands we can have a play with the JS side. My real objection here was, don't be so presumptuous in the JS comments 😝

armanbilge: > You're the JS master, but is targeting just ECMA Script 2021 feasible?

Yes, we can do a feature test and gracefully fallback. It is already fairly well supported, just not standardized until ES2021.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef#browser_compatibility

vasilmkd: Sounds good, I mainly based my assumptions around this discussion https://github.com/scala-js/scala-js/issues/4477.

core/js/src/main/scala/cats/effect/unsafe/SuspendedFiberBag.scala
```diff
@@ -0,0 +1,57 @@
...
+ */
+private[effect] final class SuspendedFiberBag {
+  private[this] val bag: mutable.Map[AnyRef, IOFiber[_]] =
```
djspiewak: So the problem here is we're retaining a hard reference to these fibers, thus preventing GC. We really need weak references to do this without causing memory leaks, so I think it would be better for now to make the JS fiber bag a no-op, otherwise we regress correctness quite a lot.

core/jvm/src/main/scala/cats/effect/unsafe/SuspendedFiberBag.scala
```diff
@@ -0,0 +1,90 @@
...
+ *      contention between threads. A particular instance is selected using a thread local
+ *      source of randomness using an instance of `java.util.concurrent.ThreadLocalRandom`.
+ *
+ * @note
+ *   The `unmonitor` method is a no-op, but it needs to exist to keep source compatibility with
+ *   Scala.js. The removal of a resumed fiber is done automatically by the GC.
+ */
+private[effect] final class SuspendedFiberBag {
+
+  private[this] val size: Int = Runtime.getRuntime().availableProcessors() << 2
```
djspiewak, line 51: Why `<< 2`?

core/jvm/src/main/scala/cats/effect/unsafe/SuspendedFiberBag.scala
```diff
@@ -0,0 +1,90 @@
...
+   * Registers a suspended fiber, tracked by the provided key which is an opaque object which
+   * uses reference equality for comparison.
+   *
+   * @param key
+   *   an opaque identifier for the suspended fiber
+   * @param fiber
+   *   the suspended fiber to be registered
+   */
+  def monitor(key: AnyRef, fiber: IOFiber[_]): Unit = {
+    val rnd = ThreadLocalRandom.current()
```
djspiewak, line 73: Doesn't this have to be implemented using a `ThreadLocal` behind the scenes? I would assume that's quite slow. We could probably improve it at *least* in the happy path when code is running on our worker threads by exposing the worker-local source of randomness internally.

djspiewak (changes requested):

core/jvm/src/main/scala/cats/effect/unsafe/SuspendedFiberBag.scala
vasilmkd: Our worker threads will each have their own uncontended bag.

core/js/src/main/scala/cats/effect/unsafe/SuspendedFiberBag.scala
vasilmkd: JS has an `unmonitor` method which actually removes mappings from this map when they resume.

core/jvm/src/main/scala/cats/effect/unsafe/SuspendedFiberBag.scala
vasilmkd: I believe @RafalSumislawski measured different configurations for the ThreadSafeHashtable and came up with this.

RafalSumislawski: Indeed https://github.com/typelevel/cats-effect/pull/1937#issuecomment-835304694

core/jvm/src/main/scala/cats/effect/unsafe/SuspendedFiberBag.scala
vasilmkd: `ThreadLocalRandom` is indeed implemented using a `ThreadLocal`, but it's better than having a contended `java.util.Random` instance. We use something similar when external thread enqueue fibers on our work stealing pool.

As far as this PR goes, this only contains the global fallback. I still haven't implemented the bag for the work stealing pool, where each worker thread will have its own uncontended `WeakHashMap`. I will add that in another PR, I didn't want to make this one too big.

armanbilge: Also, landing this one lets me work in parallel on a JS implementation while Vasil tackles the work stealing pool version 😁

core/js/src/main/scala/cats/effect/unsafe/SuspendedFiberBag.scala
djspiewak: Ah but what if they don't resume! Right now, if you suspend a fiber and then let it go out of scope, it will just be GC'd. This change breaks that, but only on JS.

core/jvm/src/main/scala/cats/effect/unsafe/SuspendedFiberBag.scala
djspiewak: > Our worker threads will each have their own uncontended bag.

I was going to ask whether or not they really need it, but thinking more, it makes sense that it would be worth avoiding the synchronization penalty.

core/js/src/main/scala/cats/effect/unsafe/SuspendedFiberBag.scala
vasilmkd: Sorry, I realize now what you mean. Yeah, that's really what I wanted initially, but got confused with the availability of weak references on JS and ended up making a mistake. I will make the JS bag a no-op for now and @armanbilge will iterate on it.

djspiewak (approved):

PR #2409, opened 2021-10-14
Use explicit `isInstanceOf` followed by `asInstanceOf`

vasilmkd:
Found a couple of places where we could have been extra careful.


djspiewak (approved):
I think it's fair to be extra explicit, though it's worth noting that the bytecode is unchanged in each of these cases.

vasilmkd:
You made me change it once in a different PR. 😅

PR #2410, opened 2021-10-14
Override `IOFiber#toString`

djspiewak:


vasilmkd (approved):

djspiewak:
I think that was a sub-60s approval…

armanbilge (commented):
Wait, I made a comment on this in the other PR that got forgotten 😛 

I was wondering, can we use any collected trace information to give some indication what this fiber is/was doing?

djspiewak:
> I was wondering, can we use any collected trace information to give some indication what this fiber is/was doing?

I'm game for something like that. The only tricky bit is the ensuring that `toString` is still thread-safe.

armanbilge:
Absolutely. I can take a stab at it, maybe in a follow-up PR? :)

djspiewak:
Go for it! Maybe something like, instead of `RUNNING` do `RUNNING @ Foo.bar:123`? So just the top-most frame, basically.

armanbilge:
Oh, ok. So not like a Java thread dump which would do the whole stack trace right? I guess since we make fibers willy-nilly it could be overwhelming 😅 

It seems like [`resumeIO`](https://github.com/typelevel/cats-effect/blob/b6f9530816230bb90dc256d0abb971497ccd71f1/core/shared/src/main/scala/cats/effect/IOFiber.scala#L110) is the best source of info on where the fiber is?

djspiewak:
Oh actually you're right, maybe the `@` is useful for both suspension and running. Honestly maybe even more useful for suspension, so you can get an idea of why a fiber is hung. When the fiber is suspended, `resumeIO` will tell you where it is. If it's *not* suspended and it's just running, then you would need to look at the trace ring buffer.

vasilmkd:
`resumeIO` could be useful for suspended fibers, but it's generally `null` for running ones.

On the other hand, is it really that bad to inspect the tracing events circular buffer? It could just happen that it might not be completely up to date, but then the same can be said for any field of a fiber.

PR #2411, opened 2021-10-14
Async cancelation should first re-enqueue the fiber

vasilmkd:
This matches the resumption mechanism in the fact that it simply enqueues the canceled fiber for cancelation, instead of executing the canceled fiber's run loop in place, which can already be running as part of another fiber's run loop (e.g. completing an async callback on a worker thread as part of an `IO.delay`).

It also seems to be slightly faster. I chose the `AsyncBenchmark.race` benchmark because it exercises cancelation.

`series/3.x`:
```
Benchmark            (size)   Mode  Cnt      Score      Error  Units
AsyncBenchmark.race     100  thrpt   20  14542.116 ± 1145.839  ops/s
```

`This PR`:
```
Benchmark            (size)   Mode  Cnt      Score     Error  Units
AsyncBenchmark.race     100  thrpt   20  18157.855 ± 430.937  ops/s
```

### DIFF
--- a/core/js/src/main/scala/cats/effect/IOFiberConstants.scala
+++ b/core/js/src/main/scala/cats/effect/IOFiberConstants.scala
@@ -44,11 +44,13 @@ private object IOFiberConstants {
   final val ExecR: Byte = 0
   final val AsyncContinueSuccessfulR: Byte = 1
   final val AsyncContinueFailedR: Byte = 2
-  final val BlockingR: Byte = 3
-  final val EvalOnR: Byte = 4
-  final val CedeR: Byte = 5
-  final val AutoCedeR: Byte = 6
-  final val DoneR: Byte = 7
+  final val AsyncContinueCanceledR: Byte = 3
+  final val AsyncContinueCanceledWithFinalizerR = 4
+  final val BlockingR: Byte = 5
+  final val EvalOnR: Byte = 6
+  final val CedeR: Byte = 7
+  final val AutoCedeR: Byte = 8
+  final val DoneR: Byte = 9
 
   // ContState tags
   final val ContStateInitial: Int = 0
--- a/core/jvm/src/main/java/cats/effect/IOFiberConstants.java
+++ b/core/jvm/src/main/java/cats/effect/IOFiberConstants.java
@@ -45,11 +45,13 @@ final class IOFiberConstants {
   static final byte ExecR = 0;
   static final byte AsyncContinueSuccessfulR = 1;
   static final byte AsyncContinueFailedR = 2;
-  static final byte BlockingR = 3;
-  static final byte EvalOnR = 4;
-  static final byte CedeR = 5;
-  static final byte AutoCedeR = 6;
-  static final byte DoneR = 7;
+  static final byte AsyncContinueCanceledR = 3;
+  static final byte AsyncContinueCanceledWithFinalizerR = 4;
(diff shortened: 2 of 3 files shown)


djspiewak:
So I'm struggling to see what this does *aside* from effectively injecting an `IO.cede` at the head of any list of finalizers (including if the finalizers are empty). It's not entirely clear to me how that interacts with `race`, but my guess is that the added dose of fairness probably causes the benchmark to wind down just a little faster. Ultimately, I think this is a fairness question more than a throughput one, and I'm trying to reason through whether it matters in practice.

vasilmkd:
It's a fairness issue, sure, but there's also a stack safety issue (which admittedly may not be too realistic in practice). Let's say there's an async callback for a given fiber A that needs to be completed in an `IO.delay` on another fiber B. And let's say that as soon as this callback is completed, cancelation from a third fiber is propagated. Then, fiber B would run `asyncCancel(null)` for fiber A as part of its run loop (run loop in a run loop situation). It's the same reason we don't resume fibers immediately on the happy path (or exception for that matter), but we enqueue them again (at least IMO).

vasilmkd:
It also doesn't inject `IO.cede` for all finalizers, only ones that would be run as part of `IO.async` callback completion when cancelation is propagated by a third party which couldn't win the race.

vasilmkd:
Honestly, same for calling `IOFiber#cancel`, more specifically `asyncCancel(fin)` inside a sequenced `IO.async_`, this will start the canceled fiber's run loop inside the current one. Can a huge chain of fibers be built that causes a stack overflow?

djspiewak (changes requested):
Staring at this some more, I realized that we actually have a bigger fairness issue here. Imagine a very long chain of synchronous finalizers on a fiber, where each finalizer is itself less than the auto-cede threshold. We will hog the carrier thread the whole way through without ever once yielding.

Due to the way the whole machine works, there isn't a good way to pass the current autocede threshold through from one finalizer to the next, so the most direct solution might be to just reschedule the fiber on each iteration of the loop (rather than calling `runLoop` directly). WDYT?

Interestingly, this is probably exactly what's going on in the `race` benchmark, since this situation indirectly arises when you have a long chain of fibers, each of which cancels the next one in its finalizer.

**Edit**: 

> Can a huge chain of fibers be built that causes a stack overflow?

Jinx :-) Yes

vasilmkd:
So what's the game plan here then?

djspiewak:
I have an idea. I'll have a PR up in a minute.

vasilmkd:
Ah, you'll add a cede in `asyncCancel`?

vasilmkd:
@djspiewak I removed your commit because I want to keep this PR focused on the title. Your commit shows a slightly bigger problem, and that is the fact that we call `asyncCancel` from inside our run loop, which then starts a new instance of the run loop with a fresh stack and fresh iteration counts, while also adding a new stack frame. We need to rework `asyncCancel` to fix this. I will do that in a follow-up.

vasilmkd:
`runLoop` should only be called from methods ending with `R`, which are resumption methods called from `run()` whenever a fiber is picked up from a worker queue and executed, not from `K` methods which are continuations popped from the stack. The former creates 1 single run loop stack entry, the latter adds one each time `runLoop` is called. I will do a careful change in a follow-up, as I said.

djspiewak (approved):

PR #2413, opened 2021-10-15
Augment `IOFiber#toString` w/ trace from `IO`

armanbilge:
Following from https://github.com/typelevel/cats-effect/pull/2410#pullrequestreview-779904571.


tests/shared/src/test/scala/cats/effect/IOFiberSpec.scala
```diff
@@ -0,0 +1,53 @@
...
+        s <- IO(f.toString)
+        _ <- f.cancel
+      } yield s.matches(pattern)
+    }
+
+    "toString a suspended fiber" in real {
+      val pattern =
+        raw"cats.effect.IOFiber@[0-9a-f]+ SUSPENDED"
+      for {
+        f <- IO.sleep(1.hour).start
```
armanbilge: I couldn't figure out how to get a suspended fiber with a trace event :( Help please? :)

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -1403,6 +1403,22 @@ private final class IOFiber[A](
...
+    val resumeIO = this.resumeIO
+    val tracingEvents = this.tracingEvents
+    val event =
+      if ((resumeIO ne null))
+        resumeIO.event
+      else if (tracingEvents ne null)
+        tracingEvents.peek
+      else null
+    val frame = if (event ne null) Tracing.getFrame(event) else null
+    val opAndCallSite = if (frame ne null) s" $frame" else ""
```
vasilmkd: Regarding how to construct a suspended fiber with a frame, try adding printlns to this code, to see which of these fields are relevant. It might be happening that `resumeIO` is not null, but its event is, or that `Tracing.getFrame(event)` returns a null, which means that the whole stack trace is filtered out.

armanbilge: Ah, you are right, I forgot about the filtering 😵‍💫

armanbilge: Do you think we should disable filtering? So, we can get at least some indication of the fiber is doing, even it's not sourced from userland code.

armanbilge:
This is what I'm getting locally. I think CI might be different for some reason.

JVM:
```
cats.effect.IOFiber@76193877 RUNNING >> @ fiber.IOFiberSpec.loop$1(IOFiberSpec.scala:31)
cats.effect.IOFiber@25b3fe5a SUSPENDED delay @ cats.effect.IO$$anon$2.$anonfun$apply$5(IO.scala:988)
```

JS:
```
cats.effect.IOFiber@18 RUNNING >> @ fiber.IOFiberSpec.loop$1(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-fastopt/https:/raw.githubusercontent.com/typelevel/cats-effect/00694a4701e9a228bed7d162418851bcdb5c09b5/tests/shared/src/test/scala/cats/effect/IOFiberSpec.scala:31)
cats.effect.IOFiber@1a SUSPENDED delay @ <jscode>.{anonymous}()(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-fastopt/https:/raw.githubusercontent.com/typelevel/cats-effect/00694a4701e9a228bed7d162418851bcdb5c09b5/core/shared/src/main/scala/cats/effect/IO.scala:988)
```

armanbilge:
Yup, this is failing b/c JVM CI is doing this instead:
```
cats.effect.IOFiber@555fc56b RUNNING flatMap @ cats.effect.IO.$greater$greater(IO.scala:135)
```

Edit: working hypothesis: the trace event is sourced from `resumeTag` if not null, otherwise the ring buffer (which aren't necessarily the same). So maybe `resumeTag` becomes `null` in CI but not on my local. Maybe the test needs to be relaxed?

Edit 2: I think I had it backwards, I think the `resumeTag = null` was not being published in CI but it was on my local. Because it is always "published" on JS, and the outputs were matching.

armanbilge:
Ok, relaxed the test to pass in either case now.

I think the bigger issue here is that to get truly meaningful output for the user, we should build a proper trace via the ring buffer. But this is difficult to do in a threadsafe way.

armanbilge:
Just pushed a commit that goes in a different direction:

When the fiber is created, it stores the trace event associated with the `startIO`. Then, it uses this event to implement `toString`.

It gives output like this:
```
cats.effect.IOFiber@3807fb51 RUNNING >> @ fiber.IOFiberSpec.loop$1(IOFiberSpec.scala:31)
cats.effect.IOFiber@73fcfd34 SUSPENDED async_ @ fiber.IOFiberSpec.foreverNever$1(IOFiberSpec.scala:43)

cats.effect.IOFiber@18 RUNNING >> @ fiber.IOFiberSpec.loop$1(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-fastopt/https:/raw.githubusercontent.com/typelevel/cats-effect/c3c134ae99e45dfb5a8af12b82439a070d8abea8/tests/shared/src/test/scala/cats/effect/IOFiberSpec.scala:31)
cats.effect.IOFiber@1a SUSPENDED async @ fiber.IOFiberSpec.foreverNever$1(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-fastopt/https:/raw.githubusercontent.com/typelevel/cats-effect/c3c134ae99e45dfb5a8af12b82439a070d8abea8/tests/shared/src/test/scala/cats/effect/IOFiberSpec.scala:43)
```

Pros:
* (slightly?) more likely to reference user-land code
* published on fiber creation, no weird non-determinism

Cons:
* static, doesn't tell us what the fiber's most recently published activity is
* confusing to users, due to above point

Edit: weirdly, this is still failing CI due to the same issue. So it wasn't race conditions that cause traces to be different, it is something else...

armanbilge:
Changed my mind 🙃 

Reverted the last commit that stores the event for the `IO` that started a fiber. The current `IOFiber#toString` implementation works like this:
1. Create a `List[StackTraceElement]` via the existing `Tracing#getFrames` method that uses the event ring buffer and filters out non-userland frames. _Technically_, there are race conditions/publication issues here, which could cause this `List` to be out-of-order or otherwise incoherent, but we don't care since we will only use the head which, if exists, is guaranteed to be a frame referencing userland code.
2. As a fallback, if the head of said list is empty, we then attempt to get the latest unfiltered frame, which is not guaranteed to point to userland code, but at least will give some indication of what the fiber is doing.

Output like this:
```
cats.effect.IOFiber@3807fb51 RUNNING >> @ fiber.IOFiberSpec.loop$1(IOFiberSpec.scala:31)
cats.effect.IOFiber@73fcfd34 SUSPENDED async_ @ fiber.IOFiberSpec.foreverNever$1(IOFiberSpec.scala:43)

cats.effect.IOFiber@18 RUNNING >> @ fiber.IOFiberSpec.loop$1(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-fastopt/https:/raw.githubusercontent.com/typelevel/cats-effect/c3c134ae99e45dfb5a8af12b82439a070d8abea8/tests/shared/src/test/scala/cats/effect/IOFiberSpec.scala:31)
cats.effect.IOFiber@1a SUSPENDED async @ fiber.IOFiberSpec.foreverNever$1(/workspace/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-fastopt/https:/raw.githubusercontent.com/typelevel/cats-effect/c3c134ae99e45dfb5a8af12b82439a070d8abea8/tests/shared/src/test/scala/cats/effect/IOFiberSpec.scala:43)
```
CI is still failing due to the aforementioned tracing bug (?) that I cannot replicate locally. See https://github.com/typelevel/cats-effect/pull/2413#issuecomment-944489932.

armanbilge:
So, I'm a bit confused how this interacts with https://github.com/typelevel/cats-effect/pull/2427 with also uses `IOFiber#toString` since there would be redundant information. Assuming that `toString` should return a one-liner, I think what I have here makes sense for when `IOFiber#toString` is called from outside the context of a fiber dump. The fiber dump should do something else instead of relying on `IOFiber#toString` to avoid the confusing redundancy.

djspiewak (approved):
image

Jk @armanbilge gets all the credit!

PR #2417, opened 2021-10-16
Update npm lockfile to v2

armanbilge:
```
npm WARN old lockfile 
npm WARN old lockfile The package-lock.json file was created with an old version of npm,
npm WARN old lockfile so supplemental metadata must be fetched from the registry.
npm WARN old lockfile 
npm WARN old lockfile This is a one-time fix-up, please be patient...
npm WARN old lockfile 
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,39 @@
 {
+  "name": "cats-effect",
+  "lockfileVersion": 2,
   "requires": true,
-  "lockfileVersion": 1,
+  "packages": {
+    "": {
+      "devDependencies": {
+        "source-map-support": "^0.5.19"
+      }
+    },
+    "node_modules/buffer-from": {
+      "version": "1.1.2",
+      "resolved": "https://registry.npmjs.org/buffer-from/-/buffer-from-1.1.2.tgz",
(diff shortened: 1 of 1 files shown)


armanbilge:
Woops:
```
npm WARN read-shrinkwrap This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2. I'll try to do my best with it!
```
https://github.com/typelevel/cats-effect/pull/2417/checks?check_run_id=3915688304#step:7:9

vasilmkd:
Do we need a newer npm version for this? Maybe you can tweak the GH action yaml.

armanbilge:
Yes, which I think is related to/depends on the Node.js version. We're currently testing against Node.js 14 which is the current LTS. Node 16 becomes the LTS in ten days, maybe we can update then?



https://nodejs.org/en/about/releases/

vasilmkd:
If it's really 10 days, I'm fine with doing it now. 😄

armanbilge:
Well, there's an argument we should stay on Node.js 14 for a while instead of jumping to the next LTS? For example wait until e.g. AWS offers Node 16 lambdas.

PR #2418, opened 2021-10-17
compile scala code block with mdoc

i10416:
Add mdoc modifiers(`nest`, `reset` , `fail`, `compile-only`) and modify code blocks a bit so that they pass compile.

except
1. `build.sbt` examples like `libraryDependencies += ...`  (at getting-started.md, tutorial.md, async-await.md )
2. ammonite scripts(at getting-started.md)
3. test examples
4. REPL examples like `sbt:demo>` (at migration-guide.md, tutorial.md)
5. CE2 examples at migration-guide.md 
6. some codes left to be modified

I tried to keep original codes as possible, thus there still remains around 35 code blocks to be compiled by mdoc.

note: Dummy implementations and imports I introduced for compilation raise warning like `unused import`, `dead code`.  Should I suppress them?


vasilmkd:
> note: Dummy implementations and imports I introduced for compilation raise warning like unused import, dead code. Should I suppress them?

Can they be suppressed in `mdoc` or do you need to add `@nowarn`? I'm hesitant about the latter, it doesn't look good in documentation.

i10416:
@vasilmkd  I agree with you. `@nowarn` in examples is confusing. 

`mdoc` 's `compile-only` modifier may work. I will try it. 

How about using `@nowarn` annotation only in `mdoc:invisible` parts in order to hide some irrelevant stuffs like temporary imports and mock implementations?

vasilmkd:
Sorry about falling behind on this. We're trying to wrap up a big release. We will work on this together soon. Unless you have a new PR. 😄

vasilmkd:
@i10416 Sorry for being inactive for a bit, I should be more available now. It seems that some of the underlying files have moved on a bit, would you mind rebasing this PR?

Also, to finally answer your question, feel free to use any `mdoc` mode that you feel should be used. I still think that adding `@nowarn` is not the way to go, so we would like to avoid that.

PR #2420, opened 2021-10-17
Intoduce parReplicateAN

RafalSumislawski:
This is a CE counterpart of https://github.com/typelevel/cats/pull/4007 .

In the spots where I would normally use the `parReplicateA` introduced in https://github.com/typelevel/cats/pull/4007, I used an alternative implementation and left TODO's.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/Async.scala
+++ b/core/shared/src/main/scala/cats/effect/Async.scala
@@ -361,6 +361,16 @@ object Async {
     } yield mta
 
   /**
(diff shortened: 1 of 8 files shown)


djspiewak (approved):
This seems reasonable, albeit with a very annoying name. Would you be interested in PRing the equivalent to CE3 as well?

RafalSumislawski:
I'm not happy with the name, but that's what evolved naturally from what we had before it. Yes I will open an equivalent PR to CE3.

PR #2427, opened 2021-10-18
Starting point for working fiber dumps

### LINKED ISSUE #1029: Live fiber dumps

djspiewak:
Where by "working" I mean "actually not quite working yet". Also we should come up with a way to test this. Needs to build on #2408 and related work to get the suspended fibers. Major caveat on this at the end…

This works by registering a signal handler for `SIGINFO` on macOS and BSD, and for `SIGUSR1` on Linux. Confusingly, `SIGINFO` doesn't exist on Linux, and `SIGUSR1` serves a VM-level function on macOS and BSD. Nothing works on Windows, unsurprisingly. This is exactly what the `dd` utility does, btw, so it's not an unfamiliar paradigm. Pleasantly enough, macOS TTYs bind <kbd>Ctrl</kbd>-<kbd>T</kbd> to `SIGINFO`, which means we get a single hotkey for dumping fiber traces. Cool, right?

Anyway, this is obviously very lossy because I'm peeking inside concurrent data structures. Unfortunately, it's *especially* lossy in some very important ways. In particular, we will never see the *currently executing* fiber on a worker thread in the current encoding. Honestly, I'm not 100% sure if it's even possible to circumvent this limitation without adding a lot of overhead. Maybe this is less of a concern though given that, 99% of the time, the most useful traces are going to be the suspended fibers? Worth some discussion, anyway.

When all is said and done, I'd also like to expose this via an MBean, so that users aren't forced to exclusively access this information via POSIX signaling. That also gives an alternative for those poor souls on Windows.

At any rate, this seems pretty cool. It makes no attempt to do anything on ScalaJS.

Fixes #1029

### DIFF
--- a/core/js/src/main/scala/cats/effect/unsafe/SuspendedFiberBag.scala
+++ b/core/js/src/main/scala/cats/effect/unsafe/SuspendedFiberBag.scala
@@ -46,4 +46,6 @@ private[effect] final class SuspendedFiberBag {
    */
   @nowarn("cat=unused-params")
   def unmonitor(key: AnyRef): Unit = {}
+
+  def contents(): Set[IOFiber[_]] = Set()
 }
--- a/core/js/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
+++ b/core/js/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
@@ -28,4 +28,5 @@ private[effect] sealed abstract class WorkStealingThreadPool private ()
   def reportFailure(cause: Throwable): Unit
   private[effect] def rescheduleFiber(fiber: IOFiber[_]): Unit
   private[effect] def scheduleFiber(fiber: IOFiber[_]): Unit
+  private[unsafe] def fiberDump(): String
 }
--- a/core/js/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
+++ b/core/js/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
@@ -28,5 +28,5 @@ private[effect] sealed abstract class WorkStealingThreadPool private ()
   def reportFailure(cause: Throwable): Unit
   private[effect] def rescheduleFiber(fiber: IOFiber[_]): Unit
   private[effect] def scheduleFiber(fiber: IOFiber[_]): Unit
-  private[unsafe] def contents(): (Set[IOFiber[_]], Set[IOFiber[_]])
+  private[unsafe] def contents(): (Set[IOFiber[_]], Map[IOFiber[_], Thread])
 }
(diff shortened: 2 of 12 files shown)


djspiewak:
Also, if you're looking to test this out locally, the following app shows some things (at least on my machine; you need to make sure the replication factor is higher than the number of worker threads):

```scala
import cats.effect.{IO, IOApp}
import cats.syntax.all._

object TestApp extends IOApp.Simple {
  lazy val loop: IO[Unit] = IO.unit >> loop
  val run = loop.start.replicateA(32).flatMap(_.traverse(_.join)).void
}
```

djspiewak:
New example app:

```scala
import cats.effect.{IO, IOApp}
import cats.syntax.all._

object TestApp extends IOApp.Simple {

  lazy val loop: IO[Unit] =
    IO.unit.map(_ => ()) >>
      IO.unit.flatMap(_ =>
        loop)

  val run = for {
    fibers <- loop.start.replicateA(32)

    sleeper = for {
      _ <- IO.unit
      _ <- IO.unit
      _ <- IO.sleep(1.day)
    } yield ()

    _ <- sleeper.start.replicateA(8)
    _ <- fibers.traverse(_.join)
  } yield ()
}
```

And new dump: https://gist.github.com/2053b456076ea614ccf6f0f01eefbbc1

Seems cool

djspiewak:
Better. Now we can detect hard-blocked fibers: https://gist.github.com/5ce47b6af7575c5e117e6fca96c92eb9

```scala
import cats.effect.{IO, IOApp}

import java.util.concurrent.CountDownLatch

object TestApp extends IOApp.Simple {

  lazy val loop: IO[Unit] =
    IO.unit.map(_ => ()) >>
      IO.unit.flatMap(_ => loop)

  val run = for {
    _ <- loop.start.replicateA(20)

    sleeper = for {
      _ <- IO.unit
      _ <- IO.unit
      _ <- IO.sleep(1.day)
    } yield ()

    _ <- sleeper.start.replicateA(8)

    _ <- IO {
      val latch = new CountDownLatch(1)
      latch.await()
    }
  } yield ()
}
```

PR #2429, opened 2021-10-18
Introduce parReplicateAN

RafalSumislawski:
CE3 counterpart of #2420

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
@@ -79,6 +79,12 @@ trait GenConcurrent[F[_], E] extends GenSpawn[F, E] {
   }
 
   /**
+   * Like `Parallel.parReplicateA`, but limits the degree of parallelism.
+   */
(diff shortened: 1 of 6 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
```diff
@@ -78,6 +78,12 @@ trait GenConcurrent[F[_], E] extends GenSpawn[F, E] {
     }
   }
 
+  /**
+   * Like `Parallel.parReplicateA`, but limits the degree of parallelism.
+   */
+  def parReplicateAN[A](n: Int)(replicas: Int, ma: F[A]): F[List[A]] =
```
djspiewak, line 84: Now that I'm staring at the API a bit more… why the separate parameter block?

RafalSumislawski: I'm trying to stay consistent with how the `n: Int` has it's own parameter block in `parSequenceN` and `parTraverseN`

djspiewak: Ah, so it does!

djspiewak (approved):

PR #2433, opened 2021-10-19
Implement `SuspendedFiberBag` for JS

armanbilge:
This was a fairly straightforward port of the JVM version. It feature tests for [`WeakMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap), [`WeakRef`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef), and [`FinalizationRegistry`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry) and gracefully falls back to the no-op implementation.

### DIFF
--- a/core/js/src/main/scala/cats/effect/unsafe/IterableWeakMap.scala
+++ b/core/js/src/main/scala/cats/effect/unsafe/IterableWeakMap.scala
@@ -57,6 +57,12 @@ private[unsafe] class IterableWeakMap[K, V] {
 }
 
 private[unsafe] object IterableWeakMap {
+  private[this] final val Undefined = "undefined"
+  def isAvailable: Boolean =
+    js.typeOf(js.Dynamic.global.WeakMap) != "undefined" &&
+      js.typeOf(js.Dynamic.global.WeakRef) != "undefined" &&
+      js.typeOf(js.Dynamic.global.FinalizationRegistry) != "undefined"
+
   private final case class Entry[K, V](value: V, ref: js.WeakRef[K])
 
   private final case class Finalizer[K](set: mutable.Set[js.WeakRef[K]], ref: js.WeakRef[K]) {
--- a/core/js/src/main/scala/cats/effect/unsafe/IterableWeakMap.scala
+++ b/core/js/src/main/scala/cats/effect/unsafe/IterableWeakMap.scala
@@ -23,7 +23,7 @@ import IterableWeakMap._
 
 // https://github.com/tc39/proposal-weakrefs#iterable-weakmaps
 private[unsafe] class IterableWeakMap[K, V] {
-  private[this] val weakMap = new WeakMap[K, Entry[K, V]]
+  private[this] val weakMap = new WeakMap[K, V]
   private[this] val refSet = mutable.Set[js.WeakRef[K]]()
   private[this] val finalizationGroup =
     new js.FinalizationRegistry[K, Finalizer[K], js.WeakRef[K]](_.cleanup())
@@ -31,27 +31,19 @@ private[unsafe] class IterableWeakMap[K, V] {
   def set(key: K, value: V): Unit = {
     val ref = new js.WeakRef(key)
 
-    weakMap.set(key, Entry(value, ref))
+    weakMap.set(key, value)
     refSet.add(ref)
     finalizationGroup.register(key, Finalizer(refSet, ref), ref)
   }
 
-  def get(key: K): Option[V] = weakMap.get(key).toOption.map(_.value)
-
-  def delete(key: K): Boolean =
-    weakMap.get(key).fold(false) { entry =>
-      weakMap.delete(key)
-      refSet.remove(entry.ref)
-      finalizationGroup.unregister(entry.ref)
-      true
-    }
+  def get(key: K): Option[V] = weakMap.get(key).toOption
 
   def entries(): Iterator[(K, V)] =
     refSet.iterator.flatMap { ref =>
       (for {
         key <- ref.deref()
-        entry <- weakMap.get(key)
-      } yield (key, entry.value)).toOption
+        value <- weakMap.get(key)
+      } yield (key, value)).toOption
     }
 
 }
@@ -62,8 +54,6 @@ private[unsafe] object IterableWeakMap {
     js.typeOf(js.Dynamic.global.WeakMap) != Undefined &&
       js.typeOf(js.Dynamic.global.WeakRef) != Undefined &&
       js.typeOf(js.Dynamic.global.FinalizationRegistry) != Undefined
-
-  private final case class Entry[K, V](value: V, ref: js.WeakRef[K])
 
   private final case class Finalizer[K](set: mutable.Set[js.WeakRef[K]], ref: js.WeakRef[K]) {
     def cleanup(): Unit = {
--- a/core/js/src/main/scala/cats/effect/unsafe/IterableWeakMap.scala
+++ b/core/js/src/main/scala/cats/effect/unsafe/IterableWeakMap.scala
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2020-2021 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.unsafe
+
+import scala.collection.mutable
+import scala.scalajs.js
+
+import IterableWeakMap._
+
+// https://github.com/tc39/proposal-weakrefs#iterable-weakmaps
+private[unsafe] class IterableWeakMap[K, V] {
+  private[this] val weakMap = new WeakMap[K, Entry[K, V]]
+  private[this] val refSet = mutable.Set[js.WeakRef[K]]()
+  private[this] val finalizationGroup =
+    new js.FinalizationRegistry[K, Finalizer[K], js.WeakRef[K]](_.cleanup())
+
+  def set(key: K, value: V): Unit = {
+    val ref = new js.WeakRef(key)
+
+    weakMap.set(key, Entry(value, ref))
+    refSet.add(ref)
+    finalizationGroup.register(key, Finalizer(refSet, ref), ref)
+  }
+
+  def get(key: K): Option[V] = weakMap.get(key).toOption.map(_.value)
+
+  def delete(key: K): Boolean =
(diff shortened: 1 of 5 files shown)


core/js/src/main/scala/cats/effect/unsafe/IterableWeakMap.scala
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+}
+
+private[unsafe] object IterableWeakMap {
+  private[this] final val Undefined = "undefined"
+  def isAvailable: Boolean =
+    js.typeOf(js.Dynamic.global.WeakMap) != "undefined" &&
+      js.typeOf(js.Dynamic.global.WeakRef) != "undefined" &&
+      js.typeOf(js.Dynamic.global.FinalizationRegistry) != "undefined"
```
vasilmkd: You forgot to use `Undefined`? 😄

core/js/src/main/scala/cats/effect/unsafe/IterableWeakMap.scala
```diff
@@ -0,0 +1,74 @@
...
+import scala.collection.mutable
+import scala.scalajs.js
+
+import IterableWeakMap._
+
+// https://github.com/tc39/proposal-weakrefs#iterable-weakmaps
+private[unsafe] class IterableWeakMap[K, V] {
+  private[this] val weakMap = new WeakMap[K, Entry[K, V]]
+  private[this] val refSet = mutable.Set[js.WeakRef[K]]()
+  private[this] val finalizationGroup =
```
vasilmkd: Can you please comment on what each of these fields are used for and what type of references they are holding to the keys and values. It's for my understanding. 😄

core/js/src/main/scala/cats/effect/unsafe/IterableWeakMap.scala
armanbilge: lol!

core/js/src/main/scala/cats/effect/unsafe/IterableWeakMap.scala
armanbilge: Yeah would be good for my understanding too 😆  I just ported from here.
https://github.com/tc39/proposal-weakrefs#iterable-weakmaps

core/js/src/main/scala/cats/effect/unsafe/IterableWeakMap.scala
armanbilge:
```suggestion
  private[this] final val Undefined = "undefined"
  def isAvailable: Boolean =
    js.typeOf(js.Dynamic.global.WeakMap) != Undefined &&
      js.typeOf(js.Dynamic.global.WeakRef) != Undefined &&
      js.typeOf(js.Dynamic.global.FinalizationRegistry) != Undefined
```

core/js/src/main/scala/cats/effect/unsafe/IterableWeakMap.scala
armanbilge: IIUC the `WeakMap` here is the same as its Java equivalent, but its not iterable. So instead, we keep a set of weak references to the keys that we can use for iteration. They are weak so that the keys can still be GCed. The trouble is even though the key is GCed, the set will still contain the `WeakRef`. So lastly the `FinalizationRegistry` registers a finalizer that removes the corresponding `WeakRef` from the set when the key is GCed.

vasilmkd: Cool. Can you please verify if the `WeakMap` holds strong references to the _values_ like `WeakHashMap` does on the JDK. There we use a weak hash map with weak ref values.

vasilmkd:
Is there some special feature that requires the `Entry` class to exist?

armanbilge:
You mean, instead of just using a tuple?

core/js/src/main/scala/cats/effect/unsafe/IterableWeakMap.scala
armanbilge: Hmm, good catch, `WeakMap` might be different here.

> As such native WeakMaps do not prevent garbage collection, which eventually removes references to the key object. "Weak" references also avoid preventing garbage collection of values in the map.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap#why_weakmap

vasilmkd:
Instead of just using a value, not a pair of the key and the value.

armanbilge:
It's needed for the `delete` functionality, which technically we don't need here.

core/js/src/main/scala/cats/effect/unsafe/IterableWeakMap.scala
armanbilge: Ok, I've been staring at the spec and I see nothing there about the values. So I think the safe thing to do is assume they are strong references.

https://262.ecma-international.org/6.0/#sec-weakmap-objects

vasilmkd:
If it's really not needed, let's delete it because we don't need a general solution.

If we do end up needing it, note that we always have a reference to both the key and the value (the opaque object and the fiber instance respectively) when calling the `unmonitor` method, so you can hack it together without allocating an `Entry` object instance.

armanbilge:
> If we do end up needing it, note that we always have a reference to both the key and the value (the opaque object and the fiber instance respectively) when calling the unmonitor method, so you can hack it together without allocating an Entry object instance.

I'm not sure that's true. We also need the `WeakRef` used to store the key in the set.

vasilmkd:
Wait a second, if we assume that the value is a strong reference, our fibers actually hold a strong reference to the key object, thus preventing GC of the keys, regardless of how they are held. We need a weak ref for the value of type `V`.

core/js/src/main/scala/cats/effect/unsafe/SuspendedFiberBag.scala
```diff
@@ -44,6 +43,29 @@ private[effect] final class SuspendedFiberBag {
...
+// Relies on features *standardized* in ES2021, although already offered in many environments
+private final class ES2021SuspendedFiberBag extends SuspendedFiberBag {
+  private[this] val bag = new IterableWeakMap[AnyRef, js.WeakRef[IOFiber[_]]]
+
+  override def monitor(key: AnyRef, fiber: IOFiber[_]): Unit = {
+    bag.set(key, new js.WeakRef(fiber))
+  }
+
+  override def unmonitor(key: AnyRef): Unit = ()
+}
```
armanbilge: @vasilmkd you mean like the `WeakRef` used here?

vasilmkd: Yeah, got confused. All good. I guess this is better to preserve "semantic syntax compatibility" with the JVM code. 😆

vasilmkd:
I'll do another pass tonight.

armanbilge:
So, point-of-interest here. With all this talk about staying on Scala 3.0.x, there's the similar question of whether we should try and stay on an old SJS version. The feeling I got is people think this is less of a big deal, but in theory cats and cats-effect could probably dial back to SJS 1.5.x without consequences. The reason its relevant here is because I'm making use of APIs introduced in 1.7.x :)

core/js/src/main/scala/cats/effect/unsafe/IterableWeakMap.scala
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+package cats.effect.unsafe
+
+import scala.collection.mutable
+import scala.scalajs.js
+
+import IterableWeakMap._
+
+// https://github.com/tc39/proposal-weakrefs#iterable-weakmaps
```
vasilmkd: Let's write a paragraph here, mentioning the original project, what it offers and what we have taken/translated here. There's precedent for giving original projects a more prominent header, let's try to be consistent.

core/js/src/main/scala/cats/effect/unsafe/IterableWeakMap.scala
```diff
@@ -0,0 +1,64 @@
...
+private[unsafe] class IterableWeakMap[K, V] {
```
vasilmkd, line 37: For easier future maintenance, let's add a scaladoc to the class, explaining its purpose (weak maps in JS are not iterable by nature, so we're augmenting them here by relying on a finalization registry which is a construct (that the JVM lacks) for registering cleanup callback functions to be executed when the GC reclaims a registered object. It would be cool if you added scaladoc/comments in the implementation methods, about what happens when, when something is registered, deregistered, cleaned up, etc.

core/js/src/main/scala/cats/effect/unsafe/SuspendedFiberBag.scala
```diff
@@ -17,13 +17,13 @@
 package cats.effect
 package unsafe
 
-import scala.annotation.nowarn
+import scala.scalajs.js
 
 /**
  * A no-op implementation of an unordered bag used for tracking asynchronously suspended fiber
  * instances on Scala.js. We will iterate on this in the future.
  */
```
vasilmkd: I think this can go on the new `NoOp` implementation. We should also add a small scaladoc on the iterable weak map backed one.

core/js/src/main/scala/cats/effect/unsafe/IterableWeakMap.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  def set(key: K, value: V): Unit = {
+    val ref = new js.WeakRef(key)
+
+    weakMap.set(key, value)
+    refSet.add(ref)
+    finalizationGroup.register(key, Finalizer(refSet, ref), ref)
+  }
+
+  def get(key: K): Option[V] = weakMap.get(key).toOption
```
vasilmkd, line 63: Out of curiosity, is there a way to translate an `UndefOr[A]` into an `A` where the undefined case is actually a `null`?

vasilmkd (changes requested):
In the home stretch now, just some scaladoc for easier future maintenance.

core/js/src/main/scala/cats/effect/unsafe/IterableWeakMap.scala
armanbilge: Yes, I think so. `UndefOr` has basically the same interface as `Option`.

vasilmkd: I'm guessing Scala.js isn't actually allocating objects for this construct, right?

armanbilge: No, `UndefOr` is a compile-time construct only.

core/js/src/main/scala/cats/effect/unsafe/IterableWeakMap.scala
armanbilge: @vasilmkd I thought it's kind of like [`Cleaner`](https://docs.oracle.com/javase/9/docs/api/java/lang/ref/Cleaner.html)?

armanbilge:
@vasilmkd added lots of comments, let me know what you think.

vasilmkd (approved):

vasilmkd (approved):

PR #2434, opened 2021-10-19
Initialize tracing buffer if needed

vasilmkd:
Changes in no particular order:
- The `conts` stack was being initialized twice.
- I changed `ArrayStack` so that fewer objects are allocated when an `IOFiber` instance is created (should speed up `start` for immediately canceled fibers now that the "wrapper" `ArrayStack` object is no longer allocated on creation).
- The `tracingEvents` ring buffer is now created only if needed (not at all allocated if tracing is disabled, which wasn't the case before).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -411,7 +411,9 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
       // introduced by #2361, Bye bye helper thread
       // changes to `cats.effect.unsafe` package private code
       ProblemFilters.exclude[MissingClassProblem]("cats.effect.unsafe.HelperThread"),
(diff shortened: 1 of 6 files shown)


vasilmkd:
There's no way to make `CallbackStack` lazily initialized unfortunately because there is no place to store the reference.

djspiewak (dismissed):
Might be worth glancing at the benchmarks. I don't think it'll make a difference, but iirc `final` fields on the JVM are slightly faster than non-`final` in some cases.

vasilmkd:
Ok, I can also go the other way around? Let's pick one and also use it for the `tracingBuffer`.

It also saves on mima exclusions.

vasilmkd:
I know I'm flip flopping here, but I like this approach better.

djspiewak (approved):

PR #2452, opened 2021-10-23
Remove trace case and restore tableswitch in immediate interpreters

vasilmkd:
Before:
```
602: lookupswitch  { // 7
    0: 668
    1: 698
    2: 727
    3: 896
    4: 946
    5: 996
    23: 1022
    default: 1054
}
```

After:
```
602: tableswitch   { // 0 to 5
    0: 640
    1: 670
    2: 699
    3: 868
    4: 918
    5: 968
   default: 994
}
```

IMO, it's not worth reordering `Trace` to fit in the tableswitch.


djspiewak (approved):
Outstanding catch! It's a little hard to tell from the diff. I take it this is just for the eager fast paths?

vasilmkd:
Yes. The immediate interpreters for `map`, `flatMap` and `attempt`. The main one is unchanged.

PR #2456, opened 2021-10-24
Update `sbt-unidoc` to 0.5.0 and enable it on Scala 3

vasilmkd:


project/CI.scala
```diff
@@ -48,7 +48,7 @@ object CI {
         jsEnv = None,
         testCommands = List("test"),
         mimaReport = true,
-        suffixCommands = List("root/unidoc213", "exampleJVM/compile"))
+        suffixCommands = List("root/unidoc", "exampleJVM/compile"))
```
djspiewak, line 51: How were we able to change this and not forced to re-run `githubWorkflowGenerate`? Something's a bit fishy here.

djspiewak (changes requested):

project/CI.scala
djspiewak: Oh, I'm dumb. It's because it's rolled up in `ciJVM`.

djspiewak (approved):

PR #2462, opened 2021-10-25
Remove the bespoke execution contexts stack

vasilmkd:
IMO, `evalOn` is not used frequently enough to warrant the increased footprint of every fiber.


djspiewak (approved):
Ooooh this is nice! I experimented with putting the contexts on the object stack when I first wrote that bit, but it didn't work for some reason. I'll try to remember why but I'm guessing it was an unrelated bug.

PR #2463, opened 2021-10-25
Add configuration forwarders to `IORuntime` and remove prefetched values in `IOFiber`

vasilmkd:
Since I refactored the run loop to count down from `cancelationCheckThreshold` and `autoYieldThreshold` towards 0 during an execution of a fiber, the prefetched values were only relegated to be used in `R` methods, which are executed in `run()` and before the run loop starts. These `R` methods run after async boundaries which are already expensive in terms of memory access. This PR is a compromise between heavily reduced fiber footprint (8 bytes per fiber) and slightly better access to the configuration values through 1 fewer pointer dereference.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -113,9 +113,6 @@ private final class IOFiber[A](
   /* similar prefetch for EndFiber */
   private[this] val IOEndFiber: IO.EndFiber.type = IO.EndFiber
 
-  private[this] val cancelationCheckThreshold: Int = runtime.config.cancelationCheckThreshold
-  private[this] val autoYieldThreshold: Int = runtime.config.autoYieldThreshold
-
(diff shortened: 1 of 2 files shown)


djspiewak (commented):
I would guess that the "tons of `cede`s" WSTP benchmark would be hit the hardest by this change. If there are marginal differences there, then we can chalk it up as a win.

vasilmkd:
Yeah, incoming.

vasilmkd:
I got an idea how to make this free.

vasilmkd:
Ok, I did the benchmark runs.

`series/3.x`:
```
WorkStealingBenchmark.scheduling  1000000  thrpt   20  12.062 ± 0.280  ops/min
```

`This PR`:
```
WorkStealingBenchmark.scheduling  1000000  thrpt   20  12.377 ± 0.211  ops/min
```

Absolutely no difference, within margin of error.

I might revisit this topic in the future, but I think it's worth merging for now.

djspiewak (approved):

PR #2464, opened 2021-10-25
Incorporate MapRef into Std

### LINKED ISSUE #2444: Integration of MapRef into cats-effect
`MapRef` is a fairly simple abstraction that is one of the most critical tools for more advanced concurrent state machines. I propose it move into cats-effect in std that way everyone can leverage it. The name is just what has been used, but its by no means the only name that could be used for it. 

https://github.com/davenverse/mapref

```scala
trait MapRef[F[_], K, V] extends Function1[K, Ref[F, V]]{
  def apply(k: K): Ref[F, V]
}
```

This is then implemented for a variety of implementations ConcurrentHashMap, Ref[F, Map[K, V]], Array[Ref[Map[K, V]], etc.

These all expose `MapRef[F, K, Option[V]]` which when none are not present in the underlying storage and hence does not space leak per key on empty. This is then enhanced via defaulted which lifts `MapRef[F, K, Option[V]] =>  MapRef[F, K, V]` where some default value is instead removed from the map. Giving you a fully keyed state machine with no changes to the underlying state machinery. 

This is already the baseline for mules, single-fibered, ratelimit, lock, as well as its in process of use for keysemaphore, and used as a keyedcircuit for rediculous concurrent. As well as keypool advanced. The use cases continue to expand as a dominating number of state machines are actually keyed.

This is a fairly fundamental abstraction that enables a whole class of state machines that the current ref/deferred alone do not permit without space leak concerns or building something equivalent to this.

Your thoughts are appreciated.

ChristopherDavenport:
Fixes #2444 

Integrates MapRef.

Features: ShardedMapRef, ConcurrentHashMap, and Scala Concurrent Map Implementations, and Defaulted Ref/MapRef.

Why Care? This enables keyed concurrent state machines that don't space leak. It basically allows almost any state machine with a default state to be lifted to a keyed version of the same state machine that is removed entirely when it enters some zero state.


tests/jvm/src/test/scala/cats/effect/std/MapRefTrieSpec.scala
```diff
@@ -0,0 +1,172 @@
...
+        r <- MapRef.ofScalaConcurrentTrieMap[IO, Unit, Int]
+        _ <- r(()).set(Some(0))
+        result <- r(()).tryUpdate(_.map(_+ 1))
+        value <- r(()).get
+      } yield result && value == Some(1)
+
+      op.map(a => a must_=== true)
+    }
+
+    // test("MapRef.ofScalaConcurrentTrieMap - tryUpdate - should fail to update if modification has occurred" ) {
```
ChristopherDavenport: Never got these operational again when upgraded to ce3. Just never got them hammered out.

std/shared/src/main/scala/cats/effect/std/MapRef.scala
```diff
@@ -0,0 +1,594 @@
...
+
+
+  /**
+   * Operates with default and anytime default is present instead information is removed from underlying ref.
+   * This is very useful as a default state can be used to prevent space leaks over high arity maprefs.
+   * 
+   * Also useful for anytime a shared storage location is used for a ref, i.e. DB or Redis to not waste
+   * space.
+   **/
+  def defaultedRef[F[_]: Functor, A: Eq](ref: Ref[F, Option[A]], default: A): Ref[F, A] =
```
ChristopherDavenport: This makes sense here as this is where it necessary. But might be useful elsewhere. We could also rely on global equality if we like being risky or offer a global equality version.

tests/jvm/src/test/scala/cats/effect/std/MapRefTrieSpec.scala
ChristopherDavenport: Fixed

tests/jvm/src/test/scala/cats/effect/std/MapRefJVMSpec.scala
```diff
@@ -169,4 +169,40 @@ class MapRefTrieSpec extends BaseSpec {
 
   }
 
+  "MapRef" should { // Requires unsafeRunSync so doesn't work with JS
```
armanbilge: Interesting note, it's not just that JS doesn't have `unsafeRunSync` so this test doesn't work, it's actually that this test is meaningless on JS anyway because concurrent modifications are impossible :)

armanbilge:
Just a thought, could we abstract the current `RefSpec` in tests and reuse it for the various `MapRef` implementations?

https://github.com/typelevel/cats-effect/blob/series/3.x/tests/shared/src/test/scala/cats/effect/kernel/RefSpec.scala

std/jvm/src/main/scala/cats/effect/std/MapRefCompanionPlatform.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  def inScalaConcurrentTrieMap[G[_]: Sync, F[_]: Concurrent, K, V]: G[MapRef[F, K, Option[V]]] =
+    Sync[G]
+      .delay(scala.collection.concurrent.TrieMap.empty[K, V])
+      .map(MapRef.fromScalaConcurrentMap[F, K, V](_))
+
+  def ofScalaConcurrentTrieMap[F[_]: Concurrent, K, V]: F[MapRef[F, K, Option[V]]] =
+    Concurrent[F]
+      .unit
+      .map(_ => scala.collection.concurrent.TrieMap.empty[K, V])
```
djspiewak: Very leery of something like this. :-)

std/shared/src/main/scala/cats/effect/std/MapRef.scala
```diff
@@ -0,0 +1,607 @@
...
+object MapRef extends MapRefCompanionPlatform {
+
+  private class ShardedImmutableMapImpl[F[_]: Concurrent, K, V](
+      ref: K => Ref[F, Map[K, V]],
+      val keys: F[List[K]]
+  ) extends MapRef[F, K, Option[V]] {
+    class HandleRef(k: K) extends Ref[F, Option[V]] {
+
+      def access: F[(Option[V], Option[V] => F[Boolean])] = for {
+        hasBeenCalled <- Applicative[F].unit.map(_ => new AtomicBoolean(false))
```
djspiewak: …and here too. Why not simply embrace the fact that this is really forcing `Sync[F]`?

std/shared/src/main/scala/cats/effect/std/MapRef.scala
```diff
@@ -0,0 +1,607 @@
...
+      .traverse(_ => Concurrent[F].ref[Map[K, V]](Map.empty))
+      .map(fromSeqRefs(_))
+  }
+
+  /**
+   * Creates a sharded map ref to reduce atomic contention on the Map, given an efficient and
+   * equally distributed Hash, the contention should allow for interaction like a general
+   * datastructure. Created in G, operates in F.
+   */
+  def inShardedImmutableMap[G[_]: Sync, F[_]: Async, K, V](
```
djspiewak, line 171: If we more fully embrace `Sync`, then we don't need to have the `of`/`in` split in the constructors.

std/shared/src/main/scala/cats/effect/std/syntax/MapRefSyntax.scala
```diff
@@ -0,0 +1,50 @@
...
+ * limitations under the License.
+ */
+
+package cats.effect.std.syntax
+
+import cats.effect.std.MapRef
+import cats.syntax.all._
+
+trait MapRefSyntax {
+  implicit def mapRefOptionSyntax[F[_], K, V](
```
djspiewak: You can actually stick this into the `MapRef` companion and it won't require the import.

tests/shared/src/test/scala/cats/effect/std/MapRefSpec.scala
```diff
@@ -0,0 +1,299 @@
...
+import scala.concurrent.duration._
+
+class MapRefSpec extends BaseSpec {
+  private val smallDelay: IO[Unit] = IO.sleep(20.millis)
+  private def awaitEqual[A: Eq](t: IO[A], success: A): IO[Unit] =
+    t.flatMap(a => if (Eq[A].eqv(a, success)) IO.unit else smallDelay *> awaitEqual(t, success))
+
+  "MapRef" should {
+
+    "MapRef.ofSingleImmutableMapRef - concurrent modifications" in real {
```
djspiewak: Uh… we're in Specs2, so this kind of test name is pretty confusing. :-) I would prefer if we did `"ofSingleImmutableMapRef" >> { ... }` and then inside that block have `"concurrent modifications" >> real {` etc etc.

djspiewak (changes requested):
I really like this as an addition to the stdlib. My concerns basically boil down to the following:

- Confusing and IMO unnecessary `of`/`in` constructor duplication
- Inaccurate typeclass constraints faked by using `unit.map(_ => ...)` rather than `delay(...)`
- Syntax requires unnecessary imports
- Minor bikeshedding on test structure
- Lots of unnecessary and likely-inefficient use of `lazy val` rather than `def`

std/shared/src/main/scala/cats/effect/std/MapRef.scala
SystemFw: I'm going to open an issue for what I'm about to say, and I also want to give a general review of MapRef (tl;dr mostly amazing, some api bikeshedding), but I'm going to drop it here in case I forget: we should relax the semantics on `access` to not need `hasBeenCalled`. It was in "broken" form in fs2 for ages, the only reason this was introduced was to conform to the docs, and it's a pain in the arse for things that want to implement Ref, like SignallingRef or MapRef.

durban:
This seems pretty useful. What I find very strange is that it uses universal equality/hash for the keys, while there is a `Hash` typeclass in Cats.

ChristopherDavenport:
Proposal:

1. Remove the `access` requirement on replay so concurrent constraint can get met by Ref implementation. (or allow the concurrent.map unit case for this single location)
2. Make a single default implementation that matches on the effect type like the Queue implementation does that delegates to CHM on Async and Ref on Concurrent
3. All the current constructors live on in MapRef library but the trait and default lives here. That way people can still leverage the interface, but the advanced uses still have the library for those more advanced use cases that would bulk up the binary interface in cats-effect too much.

Thoughts?

Questions with this: Def we want configurable CHM initialCapacity, loadFactor and concurrencyLevel? For The Ref of Map Implementation should there be a default striping we leverage?

SystemFw:
I'm in favour of the proposal above, and I think it would be good to get this into CE :)

durban:
I would recommend basing the default implementation on `TrieMap` instead, since that can support using the `Hash` typeclass for the keys.

djspiewak (approved):

PR #2465, opened 2021-10-25
Reorder branches in run loop

vasilmkd:
`series/3.x`:
```
DeepBindBenchmark.async   10000  thrpt   20  1135.112 ± 19.861  ops/s
```

`This PR`:
```
DeepBindBenchmark.async   10000  thrpt   20  1210.784 ± 16.979  ops/s
```

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -201,31 +201,35 @@ private final class IOFiber[A](
       return
     }
 
-    /* Null IO, blow up but keep the failure within IO */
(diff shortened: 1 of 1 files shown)


vasilmkd:
There are a couple of changes here.

1. We had an off by one error, we were checking `cancelationIterations` and `autoCedeIterations` against 0 instead of their `next` variants, so we were doing one extra iteration in each case.
2. I realized that because `autoCede` is a multiple of `cancelationCheckThreshold`, we only need to check that branch when we force the memory fence. This technically removes a branch from the hot path, hence the performance boost.

djspiewak (approved):

PR #2468, opened 2021-10-26
Status int

vasilmkd:


vasilmkd:
This doesn't actually release any space, it is filled by field alignment.

PR #2475, opened 2021-10-28
New node LTS

vasilmkd:
Node v16 became the new active LTS 2 days ago


armanbilge:
I'm not opposed to adding Node 16, but I'm also not sure how quick we should dump Node 14 until 16 is more widely adopted. See also https://github.com/typelevel/cats-effect/pull/2417#issuecomment-945072814.

vasilmkd:
Ok, it passes. Forgot about that comment, sorry.

armanbilge:
We could of course _add_ to the build matrix 😉

PR #2482, opened 2021-10-30
Enable creating `IOLocal` in `SyncIO`

armanbilge:
Not gonna lie, I want this so I can use `SyncIO` to do unsafe things 😆

But, having these sort of `in` constructors seems standard in CE3.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOLocal.scala
+++ b/core/shared/src/main/scala/cats/effect/IOLocal.scala
@@ -40,7 +40,7 @@ object IOLocal {
 
   def inSyncIO[A](default: A): SyncIO[IOLocal[A]] = in[SyncIO, A](default)
 
-  def in[F[_], A](default: A)(implicit F: Sync[F]): F[IOLocal[A]] =
+  private def in[F[_], A](default: A)(implicit F: Sync[F]): F[IOLocal[A]] =
     F delay {
       new IOLocal[A] { self =>
         override def get: IO[A] =
--- a/core/shared/src/main/scala/cats/effect/IOLocal.scala
+++ b/core/shared/src/main/scala/cats/effect/IOLocal.scala
@@ -36,8 +36,12 @@ sealed trait IOLocal[A] {
 
 object IOLocal {
 
-  def apply[A](default: A): IO[IOLocal[A]] =
-    IO {
+  def apply[A](default: A): IO[IOLocal[A]] = in[IO, A](default)
+
+  def inSyncIO[A](default: A): SyncIO[IOLocal[A]] = in[SyncIO, A](default)
+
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/cats/effect/IOLocal.scala
```diff
@@ -36,8 +36,12 @@ sealed trait IOLocal[A] {
 
 object IOLocal {
 
-  def apply[A](default: A): IO[IOLocal[A]] =
-    IO {
+  def apply[A](default: A): IO[IOLocal[A]] = in[IO, A](default)
+
+  def inSyncIO[A](default: A): SyncIO[IOLocal[A]] = in[SyncIO, A](default)
+
+  def in[F[_], A](default: A)(implicit F: Sync[F]): F[IOLocal[A]] =
```
armanbilge: Maybe make this one `private`? Not sure what place `F` has in a world of `IO` and `SyncIO`.

vasilmkd: Yeah, I'm not entirely sold on `in[F[_], A]`. I don't mind `SyncIO`.

armanbilge: It was mostly a convenient abstraction to share the constructor.

armanbilge:
```suggestion
  private def in[F[_], A](default: A)(implicit F: Sync[F]): F[IOLocal[A]] =
```

armanbilge:
On my 3.3 wishlist, but not a blocker.

ChristopherDavenport:
I'm just going to add, that what you clearly want to do with this is a path to a very dark place.

armanbilge:
Basically, I'm looking for a compromise for https://github.com/typelevel/cats-effect/pull/2067 ...

I want configurable inheritance of my `IOLocals`. Since we don't get this out-of-the-box, my idea is to make an `IO` newtype with its own type-classes that implement locals the way I want. But for these typeclasses to be able to exist outside of an `IO` context (as normally typeclasses do) I need a way to get define an `IOLocal` _outside_ of an `IO`.

ChristopherDavenport:
Like... this? But without the LiftIO constraint?

https://github.com/davenverse/fiberlocal/

Or as in a non-RT FiberLocal. Meaning global state?

armanbilge:
I want an `IOLocal` that does not inherit. Does yours do that?

djspiewak:
I'm definitely skeptical. :-) If we were to do something like this, I'd say let's just have `in` and not specialize it to `inSyncIO`, but… still skeptical.

armanbilge:
I realized I can backdoor this courtesy of `syncStep` so not gonna fight this one :)

PR #2487, opened 2021-11-01
Monitor suspended fibers on all platforms

vasilmkd:
All our supported platforms have reached feature parity, thanks to the awesome work of @armanbilge. It's time to monitor the suspended fibers on JS too, as well as get rid of some unnecessary methods. Furthermore, move everything under `IOFiber`, which we prefer because we like `invokespecial` `private[this] def monitor()` instead of `invokevirtual` `protected def monitor()`.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -1022,8 +1022,7 @@ private final class IOFiber[A](
    * Registers the suspended fiber in the global suspended fiber bag.
    */
   private[this] def monitor(key: AnyRef): Unit = {
-    val fiber = this
-    runtime.suspendedFiberBag.monitor(key, fiber)
+    runtime.suspendedFiberBag.monitor(key, this)
   }
 
   /* can return null, meaning that no CallbackStack needs to be later invalidated */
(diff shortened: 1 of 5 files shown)


core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -1021,7 +1018,13 @@ private final class IOFiber[A](
   private[this] def suspend(): Unit =
     suspended.set(true)
 
-  private[effect] def runtimeForwarder: IORuntime = runtime
+  /**
+   * Registers the suspended fiber in the global suspended fiber bag.
+   */
+  private[this] def monitor(key: AnyRef): Unit = {
+    val fiber = this
+    runtime.suspendedFiberBag.monitor(key, fiber)
```
armanbilge: for my education, why `val fiber = this`? :)

armanbilge (dismissed):
Very nice cleanup 😁

core/shared/src/main/scala/cats/effect/IOFiber.scala
vasilmkd: No real reason, I copied over the code from the previous implementation of the method, which is probably a remnant of something I was trying to micro-optimize, but I don't remember what.

armanbilge (approved):

PR #2492, opened 2021-11-03
Worker-thread-local weak bag for tracking suspended fibers

vasilmkd:
Finalizes the proof of concept started in #2368.


armanbilge (approved):
Much more straightforward and elegant than I imagined!

PR #2504, opened 2021-11-05
Fiber dump docs

armanbilge:
Exciting, but still lots to do :)


docs/fiber-dumps.md
```diff
@@ -0,0 +1,56 @@
...
+## Introduction
+
+In the previous section, we learned how `IO` uses [Tracing](tracing.md) to generate enhanced exceptions that enable us to introspect the execution graph of a fiber, even across asynchronous boundaries.
+As powerful as this is, one shortcoming is that this trace can only be observed when raising an exception or via an explicit `IO.trace` operation, although there are many other situations in which such introspection is extremely valuable, for example to diagnose a deadlocked program.
+
+Fiber dumps solve _exactly_ this problem, by providing a mechanism to view the traces for (nearly) _all_ the fibers in your program at _any_ given time. Essentially, it is `IO`'s answer to thread dumps. This is a sophisticated feature that has been carefully implemented to minimize overhead without memory leaks of terminated fibers.
+
+## Getting a fiber dump
+
+For a running `IOApp` with tracing enabled, it's as simple as <kbd>Ctrl</kbd>-<kbd>T</kbd>. This is supported for _both_ the JVM and Node.js runtimes on macOS/BSD and Linux.
```
djspiewak: I think <kbd>Ctrl</kbd>-<kbd>T</kbd> only works on macOS/BSD, unfortunately, but the signals work fine on Linux.

docs/fiber-dumps.md
```diff
@@ -0,0 +1,56 @@
...
+
+More specifically, sending the `SIGINFO` (macOS/BSD) or `SIGUSR1` (Linux) signals to your app will trigger a fiber dump to `STDERR`.
+
+_TODO, how to do this programatically for Windows/browsers, also mbeans stuff?_
+
+## Interpreting a fiber dump
```
djspiewak: I think we should add a concrete example of a deadlock and what it looks like.

djspiewak (changes requested):
In addition to the comments I noted, I think we should move this under the **core/** subdirectory (which should be created). I want to split out introductory docs from "Core Runtime" documentation. This is also where `TestControl` and tracing docs will end up, as well as the still-to-be-written `IO` and `Resource` docs.

vasilmkd:
The documentation I added about `jlink` and GraalVM `native-image` refer to this document as `core/live-fiber-snapshot.md`. Just FIY. Feel free to decide on the name, but note those two documents might need changing. 😄

djspiewak (approved):

PR #2505, opened 2021-11-05
Rename `SuspendedFiberBag` to `FiberMonitor`

vasilmkd:
This class will not only return the suspended fibers, but also all live fibers on the compute pool, so it makes sense to do the renaming now.

Both the JVM code and the JS code will have a reference to the compute pool, discriminate on its type and the platform specific code will be completely internalized to this class.

I will follow up with an augmented type of `ExecutionContext` on JS which will keep a set of live fibers.

### DIFF
--- a/core/js/src/main/scala/cats/effect/unsafe/FiberMonitor.scala
+++ b/core/js/src/main/scala/cats/effect/unsafe/FiberMonitor.scala
@@ -33,7 +33,7 @@ private[effect] sealed abstract class FiberMonitor {
    * @param fiber
    *   the suspended fiber to be registered
    */
-  def monitor(key: AnyRef, fiber: IOFiber[_]): Unit
+  def monitorSuspended(key: AnyRef, fiber: IOFiber[_]): Unit
 }
 
 /**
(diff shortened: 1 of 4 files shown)


armanbilge:
> This class will not only return the suspended fibers, but also all live fibers on the compute pool, so it makes sense to do the renaming now.

What will be the keys for the live fibers? Themselves?

vasilmkd:
In this PR, `SuspendedFiberBag` becomes `FiberMonitor`, because the class will not only monitor asynchronously suspended fibers, but it will also encapsulate the logic for querying the monitored `compute` pool for all live fibers and creating the live fiber dump.

The name of the class and its methods should match its intended use more closely now.

core/js/src/main/scala/cats/effect/unsafe/FiberMonitor.scala
```diff
@@ -50,20 +56,22 @@ private final class ES2021SuspendedFiberBag extends SuspendedFiberBag {
...
-      new ES2021SuspendedFiberBag
-    else
-      new NoOpSuspendedFiberBag
+  def apply(compute: ExecutionContext): FiberMonitor = {
+    if (LinkingInfo.developmentMode && IterableWeakMap.isAvailable) {
+      if (compute.isInstanceOf[FiberAwareExecutionContext]) {
+        val faec = compute.asInstanceOf[FiberAwareExecutionContext]
+        new ES2021FiberMonitor(faec)
+      } else {
+        new NoOpFiberMonitor()
```
armanbilge: Going on your comment above, can this be `null`? I don't have any strong feeling about this either way. Although it's worth pointing out, tracking suspended fibers is valuable even without being able to track live fibers, but probably not the other way around. I think the question is, if someone else plugs their custom EC into the `IORuntime`, do we make it fiber-aware or not? Not worth dwelling on too much since the EC options are pretty dry anyway 😆 

```suggestion
        new ES2021FiberMonitor(null)
```

armanbilge (approved):
I can kind of see where this is going, and I like it a lot.

PR #2506, opened 2021-11-05
Create fiber-aware EC for JS

armanbilge:
This is a slight adaptation of Daniel's code. The difference is, here we add the fiber to the set _as soon as it's queued_ (rather than when the underlying EC decides to run it). Then, just before the underlying EC starts running it, it removes it. In the process of running, it may re-enqueue itself.

### DIFF
--- a/core/js/src/main/scala/cats/effect/unsafe/FiberAwareExecutionContext.scala
+++ b/core/js/src/main/scala/cats/effect/unsafe/FiberAwareExecutionContext.scala
@@ -23,16 +23,20 @@ import scala.concurrent.ExecutionContext
 private[effect] class FiberAwareExecutionContext(ec: ExecutionContext)
     extends ExecutionContext {
 
-  def fibers: collection.Set[IOFiber[_]] = fiberBag
+  def fibers: Set[IOFiber[_]] = fiberBag.toSet ++ Option(running).toSet
 
+  private[this] var running: IOFiber[_] = null
   private[this] val fiberBag = mutable.Set[IOFiber[_]]()
 
   def execute(runnable: Runnable): Unit = runnable match {
     case r: IOFiber[_] =>
       fiberBag += r
       ec execute { () =>
-        r.run()
+        // We have to remove r _before_ running it, b/c it may be re-enqueued while running
         fiberBag -= r
+        running = r
+        r.run()
+        running = null
       }
 
     case r => r.run()
--- a/core/js/src/main/scala/cats/effect/unsafe/FiberAwareExecutionContext.scala
+++ b/core/js/src/main/scala/cats/effect/unsafe/FiberAwareExecutionContext.scala
@@ -32,6 +32,7 @@ private[effect] class FiberAwareExecutionContext(ec: ExecutionContext)
       fiberBag += r
       ec execute { () =>
         // We have to remove r _before_ running it, b/c it may be re-enqueued while running
+        // B/c JS is single-threaded, nobody can observe the bag while it is running anyway
         fiberBag -= r
         r.run()
       }
--- a/core/js/src/main/scala/cats/effect/unsafe/FiberAwareExecutionContext.scala
+++ b/core/js/src/main/scala/cats/effect/unsafe/FiberAwareExecutionContext.scala
@@ -23,9 +23,8 @@ import scala.concurrent.ExecutionContext
 private[effect] class FiberAwareExecutionContext(ec: ExecutionContext)
     extends ExecutionContext {
 
-  def fibers: Set[IOFiber[_]] = fiberBag.toSet ++ Option(running).toSet
+  def fibers: Set[IOFiber[_]] = fiberBag.toSet
 
-  private[this] var running: IOFiber[_] = null
   private[this] val fiberBag = mutable.Set[IOFiber[_]]()
 
   def execute(runnable: Runnable): Unit = runnable match {
@@ -34,9 +33,7 @@ private[effect] class FiberAwareExecutionContext(ec: ExecutionContext)
       ec execute { () =>
         // We have to remove r _before_ running it, b/c it may be re-enqueued while running
         fiberBag -= r
-        running = r
         r.run()
-        running = null
       }
 
     case r => r.run()
--- a/core/js/src/main/scala/cats/effect/unsafe/FiberAwareExecutionContext.scala
+++ b/core/js/src/main/scala/cats/effect/unsafe/FiberAwareExecutionContext.scala
@@ -20,7 +20,7 @@ package unsafe
 import scala.collection.mutable
 import scala.concurrent.ExecutionContext
 
-private[effect] class FiberAwareExecutionContext(ec: ExecutionContext)
+private[effect] final class FiberAwareExecutionContext(ec: ExecutionContext)
     extends ExecutionContext {
 
   def fibers: Set[IOFiber[_]] = fiberBag.toSet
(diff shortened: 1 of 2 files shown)


armanbilge:
Btw, I think we can use a concurrent-version of this construct to wrap the blocking threadpool on JVM as well.

armanbilge:
One interesting takeaway, is that fiber dumps on JS will _never_ indicate a running fiber due to its single-threaded nature! You can't be observing, if something is running :)

vasilmkd:
@armanbilge Do you have an example of a fiber re-enqueueing itself from inside `.run()`?

armanbilge:
Err, isn't that what `IOFiber` does e.g. for ceding? Or maybe I don't understand how any of this works 😳

vasilmkd:
Actually, you're spot on. That's exactly what happens. On `cede`, the fiber calls `currentCtx.execute(this)`, which will immediately put the fiber in the bag, and then schedule it on the underlying executor. I think this works nicely.

core/js/src/main/scala/cats/effect/unsafe/FiberAwareExecutionContext.scala
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+
+package cats.effect
+package unsafe
+
+import scala.collection.mutable
+import scala.concurrent.ExecutionContext
+
+private[effect] final class FiberAwareExecutionContext(ec: ExecutionContext)
```
vasilmkd: Let's keep this just `private` for now, not `private[effect]`.

vasilmkd:
```suggestion
private final class FiberAwareExecutionContext(ec: ExecutionContext)
```

core/js/src/main/scala/cats/effect/unsafe/FiberAwareExecutionContext.scala
```diff
@@ -0,0 +1,45 @@
...
+  def fibers: Set[IOFiber[_]] = fiberBag.toSet
+
+  private[this] val fiberBag = mutable.Set[IOFiber[_]]()
+
+  def execute(runnable: Runnable): Unit = runnable match {
+    case r: IOFiber[_] =>
+      fiberBag += r
+      ec execute { () =>
+        // We have to remove r _before_ running it, b/c it may be re-enqueued while running
+        // B/c JS is single-threaded, nobody can observe the bag while it is running anyway
```
vasilmkd, line 35:
```suggestion
        // We have to remove the fiber from the bag _before_ running it, because operations like
        // `IO.cede` will add it again. This, while convenient, is not really an issue. The real
        // reason we need to remove the fiber from the bag before running it is to remove the
        // strong reference to it, otherwise, the strong reference would break GC guarantees for
        // suspended operations like `IO.never`. Suspended fibers are still tracked through the
        // iterable weak map mechanism which respects the GC.
```

core/js/src/main/scala/cats/effect/unsafe/FiberAwareExecutionContext.scala
```diff
@@ -0,0 +1,45 @@
...
+
+import scala.collection.mutable
+import scala.concurrent.ExecutionContext
+
+private[effect] final class FiberAwareExecutionContext(ec: ExecutionContext)
+    extends ExecutionContext {
+
+  def fibers: Set[IOFiber[_]] = fiberBag.toSet
+
+  private[this] val fiberBag = mutable.Set[IOFiber[_]]()
```
vasilmkd:
```suggestion
  private[this] val fiberBag = mutable.Set.empty[IOFiber[_]]
```

core/js/src/main/scala/cats/effect/unsafe/FiberAwareExecutionContext.scala
```diff
@@ -0,0 +1,45 @@
...
+package cats.effect
+package unsafe
+
+import scala.collection.mutable
+import scala.concurrent.ExecutionContext
+
+private[effect] final class FiberAwareExecutionContext(ec: ExecutionContext)
+    extends ExecutionContext {
+
+  def fibers: Set[IOFiber[_]] = fiberBag.toSet
```
vasilmkd:
```suggestion
  def liveFibers(): Set[IOFiber[_]] = fiberBag.toSet
```

The `()` will match the JVM code, which will be unsafe and contain side effects.

core/js/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
```diff
@@ -19,10 +19,14 @@ package cats.effect.unsafe
 import org.scalajs.macrotaskexecutor.MacrotaskExecutor
 
 import scala.concurrent.ExecutionContext
+import scala.scalajs.LinkingInfo
 
 private[unsafe] abstract class IORuntimeCompanionPlatform { this: IORuntime.type =>
 
-  def defaultComputeExecutionContext: ExecutionContext = MacrotaskExecutor
+  def defaultComputeExecutionContext: ExecutionContext = if (LinkingInfo.developmentMode)
```
vasilmkd: Let's tie this to `SuspendedFiberBag/FiberMonitor`. If `WeakBag` is not available, IMO, there's no reason to track fibers, if we're not able to offer suspended ones. What do you think?

vasilmkd (changes requested):
This is phenomenal.

Please do not commit my suggestions one by one, apart from the large comment. Add a commit for the rest.

core/js/src/main/scala/cats/effect/unsafe/FiberAwareExecutionContext.scala
vasilmkd: No, wait.

core/js/src/main/scala/cats/effect/unsafe/FiberAwareExecutionContext.scala
armanbilge: 👍 I think we may eventually need this behind a shared trait, so that code can do `isInstanceOf[FiberAwareEC]` or something to check if this method is available.

core/js/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
armanbilge: Oops, good point. Agreed!

core/js/src/main/scala/cats/effect/unsafe/FiberAwareExecutionContext.scala
vasilmkd: I have a plan, don't you worry ;)

armanbilge:
> Please do not commit my suggestions one by one, apart from the large comment. Add a commit for the rest.

Sorry, I'm a bit confused what gets committed when 😅

vasilmkd:
Make 1 commit with me as the co author (or not), do not commit my suggestions one by one. That's what I meant.

core/js/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
armanbilge: I just reverted this for now. When we can see the big picture more clearly we're going to have to figure out the config, which needs to know if (1) in dev mode (2) tracing is enabled and (3) the necessary features are in place, and then do these together (or not). Tracing of course stands alone regardless of fiber dump support.

vasilmkd: Yeah, no problem.

vasilmkd:
This is slightly annoying, but can you please edit the first commit and put Daniel as a co-author? I would really appreciate it. You can force push after that.

vasilmkd (approved):
Nice.

PR #2507, opened 2021-11-06
Track fibers across `IO.blocking` and `IO#evalOn`

vasilmkd:


core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -1257,6 +1263,9 @@ private final class IOFiber[A](
           onFatalFailure(t)
       }
 
+    // Remove the reference to the fiber monitor key
+    objectState.pop()
```
armanbilge, line 1267: 💯

armanbilge (approved):
LGTM but I'm not familiar enough with the runloop to know if you are popping everywhere you need to be :)

vasilmkd:
No worries.

PR #2508, opened 2021-11-06
Create fiber-aware EC for JVM

armanbilge:
This combines the ideas in #2506 and #2507 to create an (effectively) contention-free fiber-aware EC for the JVM. It uses a thread local to keep a weak fiber bag on each submitting thread. There is only a one-time contention when this bag is initialized, so as to register it with the `FiberMonitor`.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/FiberAwareExecutionContext.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/FiberAwareExecutionContext.scala
@@ -17,7 +17,6 @@
 package cats.effect
 package unsafe
 
-import scala.collection.mutable
 import scala.concurrent.ExecutionContext
 
 import java.lang.ref.WeakReference
@@ -33,10 +32,6 @@ private final class FiberAwareExecutionContext(ec: ExecutionContext, monitor: Fi
       bag
     }
   }
-
-  def liveFibers(): Set[IOFiber[_]] = fiberBag.toSet
-
-  private[this] val fiberBag = mutable.Set.empty[IOFiber[_]]
 
   def execute(runnable: Runnable): Unit = runnable match {
     case f: IOFiber[_] =>
--- a/core/jvm/src/main/scala/cats/effect/unsafe/FiberAwareExecutionContext.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/FiberAwareExecutionContext.scala
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2020-2021 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+package unsafe
+
+import scala.collection.mutable
+import scala.concurrent.ExecutionContext
+
+import java.lang.ref.WeakReference
+import java.util.{Map, WeakHashMap}
+
+private final class FiberAwareExecutionContext(ec: ExecutionContext, monitor: FiberMonitor)
+    extends ExecutionContext {
+
+  private[this] val localBag = new ThreadLocal[Map[AnyRef, WeakReference[IOFiber[_]]]] {
+    override def initialValue(): Map[AnyRef, WeakReference[IOFiber[_]]] = {
+      val bag = new WeakHashMap[AnyRef, WeakReference[IOFiber[_]]]
+      monitor.registerExtraBag(bag)
+      bag
+    }
+  }
+
+  def liveFibers(): Set[IOFiber[_]] = fiberBag.toSet
+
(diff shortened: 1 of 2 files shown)


core/jvm/src/main/scala/cats/effect/unsafe/FiberMonitor.scala
```diff
@@ -51,6 +51,9 @@ private[effect] final class FiberMonitor(
   private[this] val size: Int = Runtime.getRuntime().availableProcessors() << 2
   private[this] val bags: Array[Map[AnyRef, WeakReference[IOFiber[_]]]] =
     new Array(size)
+  private[this] val extraBags: Set[Map[AnyRef, WeakReference[IOFiber[_]]]] =
+    // A synchronized weak set
+    Collections.newSetFromMap(Collections.synchronizedMap(new WeakHashMap()))
```
armanbilge: Actually the order should probably be reversed.

```suggestion
    Collections.synchronizedSet(Collections.newSetFromMap(new WeakHashMap()))
```

core/jvm/src/main/scala/cats/effect/unsafe/FiberAwareExecutionContext.scala
```diff
@@ -0,0 +1,47 @@
...
+import java.lang.ref.WeakReference
+import java.util.{Map, WeakHashMap}
+
+private final class FiberAwareExecutionContext(ec: ExecutionContext, monitor: FiberMonitor)
+    extends ExecutionContext {
+
+  private[this] val localBag = new ThreadLocal[Map[AnyRef, WeakReference[IOFiber[_]]]] {
+    override def initialValue(): Map[AnyRef, WeakReference[IOFiber[_]]] = {
+      val bag = new WeakHashMap[AnyRef, WeakReference[IOFiber[_]]]
+      monitor.registerExtraBag(bag)
```
armanbilge, line 31: So, this is not perfect. Since these extra bags are only weakly-registered with the monitor, if the submitting thread is GCed we do "lose" whatever fibers were in its bag, even if they are queued/running on the EC.

armanbilge: One mitigation would be to register the fibers a _second_ time when they run, on the running threads local bag. This would guarantee completeness at the cost of duplication.

Edit: nevermind, a fiber could still be lost between submitting thread being GCed and when it starts running.

vasilmkd:
This looks very interesting. However, before we can consider including it, we need to run the `WorkStealingBenchmark.alloc`, `WorkStealingBenchmark.scheduling` and `DeepBindBenhchmark` on it. We need to compare this PR against a `FixedThreadPoolExecutor`, and then the same executor wrapped in a `FiberAwareExecutionContext`.

Unfortunately however, I will be a pessimist here, allocations before each and every `.run()` of a fiber, memory barriers inherent to thread locals and modifying a large hash table before every `run()`, I think it will be noticeable. That being said, I might be willing to still include it if `WorkStealingBenchmark.alloc` doesn't take a big hit.

I'll ramble here a bit.

There's another experiment that is worth repeating too. When I originally developed the WSTP, I wanted to have a wrapped type `FiberAwareExecutionContext` which our fibers would accept instead of plain `ExecutionContext`. The WSTP would extend that type and foreign ECs would be wrapped. That way, we could have called abstract methods on that class, instead of doing `instanceof` checks and casting. Unfortunately, the `instanceof` checks turned out to be noticeably faster in benchmarks, and we went for that. We've had tons of other optimizations since, so it might be worth exploring this.

With all of this being said, I don't think this should be included in the 3.3.0 release. It can come later when we've had time to experiment a bit more.

armanbilge:
> Unfortunately however, I will be a pessimist here, allocations before each and every `.run()` of a fiber, memory barriers inherent to thread locals and modifying a large hash table before every `run()`, I think it will be noticeable.

Yeah, that all seems fair :) I think the question is, what are those hits relative to the wrapped EC's own performance, which obviously is a question impossible to answer in general. But IIUC a `FixedThreadPoolExecutor` has got stuff worse than this going on anyway?

vasilmkd:
`FixedThreadPoolExecutor` is the most vanilla of thread pools, `n` number of threads all polling tasks from a shared queue.

armanbilge:
Right, I was under the impression it was full of blocking/locks to coordinate that. But I wouldn't know about such things :)

vasilmkd:
It uses a `ConcurrentLinkedQueue` to synchronize the task polling, which is lock-free. It doesn't really matter. The thing is, it used to be the default compute pool in CE2, and is the go-to pool for people that want to keep some old behavior similar to CE2 while migrating, or have problems with the work stealing pool in general.

djspiewak:
Honestly pretty skeptical about this, conceptually. `ThreadLocal` is very expensive, and often exceeds the cost of contention, which is why you don't see it used too often. It also can have some complex GC implications which have their own performance costs. Additionally, we're creating extra allocations on each `execute`, which penalizes third-party ECs even more heavily than they already are.

Ultimately, I really feel like it's okay for live fibers to disappear if you use `evalOn`. This is just one of those things where we're able to offer much stronger vertical integration if people buy into our full runtime. Underscoring this point is the fact that the most useful aspect of fiber dumps is the *suspended* fibers, which still work even on a third party EC; live fibers are just the cherry on top.

vasilmkd:
> Ultimately, I really feel like it's okay for live fibers to disappear if you use evalOn.

We already solved that. 😉 https://github.com/typelevel/cats-effect/pull/2507

This PR is to support live fiber dumps at all on foreign ECs used as compute pools.

armanbilge:
Thanks all! I'm gonna close this then, since its only use would be to enable live dumping for foreign ECs when it's a perfectly good incentive to use the WSTP :)

PR #2509, opened 2021-11-06
Only monitor suspended fibers if tracing is enabled

vasilmkd:


core/jvm/src/main/java/cats/effect/tracing/TracingConstants.java
```diff
@@ -18,7 +18,7 @@
 
 import java.util.Optional;
 
-class TracingConstants {
+public final class TracingConstants {
```
armanbilge: Unfortunate we have to make this fully public. At least we should ~scaladoc~ Javadoc "not public API! stay away!" so that we don't have to worry about bincompat.

armanbilge: Also, if it doesn't have one already, add a private constructor.

vasilmkd: Good call. There's precedent in CE2, but yeah, I'll add that.

core/jvm/src/main/java/cats/effect/tracing/TracingConstants.java
```diff
@@ -18,7 +18,20 @@
...
-class TracingConstants {
+/**
+ * Holds platform-specific flags that control tracing behavior.
+ *
+ * The Scala compiler inserts a volatile bitmap access for module field
+ * accesses. Because the `tracingMode` flag is read in various IO constructors,
+ * we are opting to define it in a Java source file to avoid the volatile
+ * access.
+ *
+ * INTERNAL API with no source or binary compatibility guarantees.
```
armanbilge, line 29: I'm missing the [flourish](https://github.com/akka/akka/blob/d703a2afe0e0c6dc8f1c0321b0f4652600e2ae99/akka-actor/src/main/scala/akka/actor/ActorCell.scala#L363) from my Akka days 😉  but LGTM, thanks!

armanbilge (approved):
Just tangentially related, do we care enough to add a Java formatter? 
https://github.com/sbt/sbt-java-formatter

core/jvm/src/main/java/cats/effect/tracing/TracingConstants.java
vasilmkd: That's awesome. 😆

vasilmkd:
I don't mind a java formatter.

PR #2510, opened 2021-11-06
Fix weak bag for the blocking mechanism

vasilmkd:
Copied commit messages for visibility:

Make the worker thread local fiber bag a constructor argument
    - This works across the blocking mechanism when new worker threads are
    spawned to replace existing blocked threads.

Remove the last bitmap$init creator
    - Uninitialized `var`s do not cause synchronization in the constructor


armanbilge (dismissed):
I'm still weirded out by this volatile thing.

armanbilge (approved):

PR #2513, opened 2021-11-06
Implement the active fiber tracking mechanism

vasilmkd:
Using various atomic variables to publish the active fiber. Waiting for benchmark results.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/LocalQueue.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/LocalQueue.scala
@@ -631,6 +631,12 @@ private final class LocalQueue {
             if (n == 1) {
               // Only 1 fiber has been stolen. No need for any memory
               // synchronization operations.
+
+              if (isStackTracing) {
+                dst.tailPublisherForwarder.lazySet(dstTl)
+                dst.plainStoreTail(dstTl)
+              }
+
               return headFiber
             }
 
--- a/core/jvm/src/main/scala/cats/effect/unsafe/LocalQueue.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/LocalQueue.scala
@@ -28,6 +28,8 @@
 package cats.effect
 package unsafe
 
+import cats.effect.tracing.TracingConstants
+
 import java.util.concurrent.ThreadLocalRandom
 import java.util.concurrent.atomic.AtomicInteger
 
@@ -121,6 +123,7 @@ import java.util.concurrent.atomic.AtomicInteger
 private final class LocalQueue {
 
   import LocalQueueConstants._
+  import TracingConstants._
 
   /**
    * The array of [[cats.effect.IOFiber]] object references physically backing the circular
@@ -394,7 +397,11 @@ private final class LocalQueue {
         }
 
         val fiber = batch(0)
-        worker.setActive(fiber)
+
+        if (isStackTracing) {
+          worker.setActive(fiber)
+        }
+
         // Publish the new tail.
         val newTl = unsignedShortAddition(tl, SpilloverBatchSize - 1)
         tailPublisher.lazySet(newTl)
@@ -467,7 +474,10 @@ private final class LocalQueue {
 
       val idx = index(real)
       val fiber = buffer(idx)
-      worker.setActive(fiber)
+
+      if (isStackTracing) {
+        worker.setActive(fiber)
+      }
 
       if (head.compareAndSet(hd, newHd)) {
         // The head has been successfully moved forward and the fiber secured.
@@ -584,7 +594,10 @@ private final class LocalQueue {
         val headFiberIdx = index(steal)
         val headFiber = buffer(headFiberIdx)
         buffer(headFiberIdx) = null
-        dstWorker.setActive(headFiber)
+
+        if (isStackTracing) {
+          dstWorker.setActive(headFiber)
+        }
 
         // All other fibers need to be transferred to the destination queue.
         val sourcePos = steal + 1
--- a/core/jvm/src/main/scala/cats/effect/unsafe/LocalQueue.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/LocalQueue.scala
@@ -399,7 +399,7 @@ private final class LocalQueue {
         val fiber = batch(0)
 
         if (isStackTracing) {
-          worker.setActive(fiber)
+          worker.active = fiber
         }
 
         // Publish the new tail.
@@ -476,7 +476,7 @@ private final class LocalQueue {
       val fiber = buffer(idx)
 
       if (isStackTracing) {
-        worker.setActive(fiber)
+        worker.active = fiber
       }
 
       if (head.compareAndSet(hd, newHd)) {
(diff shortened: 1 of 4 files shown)


vasilmkd:
`series/3.x`:
```
DeepBindBenchmark.async   10000  thrpt   20  1179.681 ± 16.579  ops/s
```

`This PR`:
```
DeepBindBenchmark.async   10000  thrpt   20  1154.277 ± 29.712  ops/s
```

core/jvm/src/main/scala/cats/effect/unsafe/LocalQueue.scala
```diff
@@ -391,6 +393,8 @@ private final class LocalQueue {
           i += 1
         }
 
+        val fiber = batch(0)
+        worker.setActive(fiber)
```
armanbilge: So if IIUC, we're not actually running the fiber _here_ (this is only the local queue, after all). It will be run by the worker thread after being returned below.

However, we are setting it as _active_ here so that it tags-along in the publications happening in the next lines. This is why the worker thread has to pass down a reference to itself when calling this method.

You're using this pattern for enqueue/dequeuing from the local queue as well as stealing from other worker threads. But, when receiving only a single fiber from the external queue ...

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -257,13 +276,15 @@ private final class WorkerThread(
...
-            val fiber = queue.enqueueBatch(batch)
+            val fiber = queue.enqueueBatch(batch, self)
             // Many fibers have been exchanged between the external and the
             // local queue. Notify other worker threads.
             pool.notifyParked(rnd)
             fiber.run()
           } else if (element.isInstanceOf[IOFiber[_]]) {
             val fiber = element.asInstanceOf[IOFiber[_]]
+            active = fiber
+            parked.lazySet(false)
```
armanbilge: ... instead, you use this lazySet for publication.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -319,7 +343,7 @@ private final class WorkerThread(
 
         case 2 =>
           // Try stealing fibers from other worker threads.
-          val fiber = pool.stealFromOtherWorkerThread(index, rnd)
+          val fiber = pool.stealFromOtherWorkerThread(index, rnd, self)
```
armanbilge, line 366: Do we need to set the active fiber just below here?

core/jvm/src/main/scala/cats/effect/unsafe/LocalQueue.scala
vasilmkd: You're exactly right. We're only setting a reference on the `WorkerThread` so that we can observe it later when creating the fiber dumps.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
vasilmkd: Again right, because in this case, we have no other publisher to queue behind. The `external` queue is a `ConcurrentLinkedQueue` and we do not control its source, so we have to make do with this. Also, this case is not the absolute hottest path. We deliberately branch here to maintain fairness in the pool, so it's not a huge deal to make this case ever so slightly slower.

Anyone who wants to observe latest, will have to read both the `size()` of the `LocalQueue`, which will touch the head and the tail, and the parked signal of the thread, and one of these will give the latest value.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
vasilmkd: `stealFromOtherWorkerThread` calls into `LocalQueue#stealInto` which sets the fiber.

armanbilge: I'm still confused about this one, if you don't mind stepping me through ...

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -176,7 +179,7 @@ private[effect] final class WorkStealingThreadPool(
     val element = externalQueue.poll(random)
     if (element.isInstanceOf[Array[IOFiber[_]]]) {
       val batch = element.asInstanceOf[Array[IOFiber[_]]]
-      destQueue.enqueueBatch(batch)
+      destQueue.enqueueBatch(batch, destWorker)
     } else if (element.isInstanceOf[IOFiber[_]]) {
       val fiber = element.asInstanceOf[IOFiber[_]]
       fiber
```
armanbilge: ... specifically, how do we get here by going through `stealInto`?

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
vasilmkd: I had left out a corner case which I just pushed a fix for.

`stealFromOtherWorkerThread` will depending on the outcome:
1. Call `stealInto`, which will do the publish
2. Call `enqueueBatch` which will do the publish
3. Do it itself if a single fiber comes from the external queue

armanbilge: Thanks, it was case (3) I was looking for.

core/jvm/src/main/scala/cats/effect/unsafe/LocalQueue.scala
```diff
@@ -459,11 +472,16 @@ private final class LocalQueue {
...
+
+      if (isStackTracing) {
+        worker.active = fiber
+      }
+
       if (head.compareAndSet(hd, newHd)) {
         // The head has been successfully moved forward and the fiber secured.
         // Proceed to null out the reference to the fiber and return it.
-        val idx = index(real)
-        val fiber = buffer(idx)
```
armanbilge, line 466: Why is it safe to move these outside of the conditional?

vasilmkd: We're only reading, not modifying anything, and the `compareAndSet`, if successful is what will publish the store. If it fails, the whole outer loop repeats and a new `fiber` will be read and set as the active.

Technically, it might happen that `worker.active` becomes published even if the CAS fails, and observed, but live fiber dumps without wasting performance are an imperfect science, so 🤷🏻.

armanbilge: That all makes sense 👍 ahh, and the reason we need it _outside_ of the conditional is so it's _before_ the CAS so that it can get published.

vasilmkd: By "we're only reading, not modifying" I meant we're not messing with the state of the buffer itself which may throw off other threads that may be stealing from it at the same time. If a wrong read indeed happens, it will be rectified immediately in the next loop.

armanbilge (approved):
With the caveat this part of the source is out-of-my-league, LGTM!

vasilmkd:
It was out of everyone's league and then we learned to bend it to our will. You will too. Also, these are some trying times. 😆

vasilmkd:
(And I think I have a proven track record of fixing the mess I create).

PR #2514, opened 2021-11-07
Schedule fibers on external queue if all worker threads busy

### LINKED ISSUE #2499: Batches of fibers may be scheduled unfairly in compute-heavy applications
This is the result of the following experiment:
- start `n` fibers
- each fiber has an `id`
- the fiber loops, infinitely incrementing an ID-bound value in a `Ref[IO, Map[String, Int]]`, where the `String` is the fiber's `id`, then yields with `IO.cede`.

The result is that the fibers that get started first get to do most of the incrementing (`n = 20`):

image

```scala
import cats.effect.{IO, IOApp, Ref}
import cats.implicits._

object FairnessExperiment extends IOApp.Simple {
  override def run: IO[Unit] = for {
    ref <- Ref[IO].of(Map.empty[String, Int])
    _ <- (1 to 20).map(i => loop(s"Fiber $i", ref)(0).start).toList.parSequence
    _ <- (IO.sleep(1.second) >> ref.get >>= printCountsAsCsv).foreverM
  } yield ()

  def loop(id: String, ref: Ref[IO, Map[String, Int]])(i: Int): IO[Unit] = for {
    _ <- ref.getAndUpdate(s => s.updatedWith(id)(v => Some(v.getOrElse(0) + 1)))
    _ <- IO.cede
    result <- loop(id, ref)(i + 1)
  } yield result

  private def printCountsAsCsv(counts: Map[String, Int]) = IO.delay {
    val columns = counts.toList.sortBy(-_._2)
    println(columns.map(_._1).mkString(","))
    println(columns.map(_._2).mkString(","))
  }
}
```

otto-dev:
Closes #2499 

If all fibers are parked, this may indicate synchronous, compute-heavy fiber usage. In that case the fiber is first placed on the external queue to increase fairness.

While atypical for most CE use-cases, this type of unfair fiber scheduling can happen on a macroscopic and microscopic level for computation-bound applications.

**Before (worst case)**

image

More even work allocation is better.

**After (worst case)**
image

## Benchmarks

Relative performance After/Before (higher is better). This is to ensure that the added fairness does not come at great expense to performance. Please consider that most of these differences fall within the margin of error! The average difference to performance across all benchmarks is 101.5% - almost identical (1.5% improvement).

image

**Before**
```
[info] Benchmark                         (cpuTokens)   (size)   Mode  Cnt      Score      Error    Units
[info] AsyncBenchmark.async                      N/A      100  thrpt   20   9389.945 ± 1208.588    ops/s
[info] AsyncBenchmark.race                       N/A      100  thrpt   20   3760.955 ±  177.710    ops/s
[info] AsyncBenchmark.racePair                   N/A      100  thrpt   20  27440.402 ±  179.971    ops/s
[info] AsyncBenchmark.start                      N/A      100  thrpt   20   4017.429 ±  519.134    ops/s
[info] DeepBindBenchmark.async                   N/A    10000  thrpt   20    697.365 ±  126.651    ops/s
[info] DeepBindBenchmark.delay                   N/A    10000  thrpt   20   1848.787 ±  140.935    ops/s
[info] DeepBindBenchmark.pure                    N/A    10000  thrpt   20   2040.696 ±  190.317    ops/s
[info] ParallelBenchmark.parTraverse           10000     1000  thrpt   20    221.625 ±    2.225    ops/s
[info] ShallowBindBenchmark.async                N/A    10000  thrpt   20    480.995 ±   66.467    ops/s
[info] ShallowBindBenchmark.delay                N/A    10000  thrpt   20   2000.198 ±  202.566    ops/s
[info] ShallowBindBenchmark.pure                 N/A    10000  thrpt   20   1943.593 ±  222.870    ops/s
[info] WorkStealingBenchmark.alloc               N/A  1000000  thrpt   20      2.641 ±    0.052  ops/min
[info] WorkStealingBenchmark.scheduling          N/A  1000000  thrpt   20      9.247 ±    0.269  ops/min
[success] Total time: 5854 s (01:37:34), completed 08/11/2021 2:11:52 PM
```

**After**
```
[info] Benchmark                         (cpuTokens)   (size)   Mode  Cnt      Score      Error    Units
[info] AsyncBenchmark.async                      N/A      100  thrpt   20   8803.194 ± 1408.049    ops/s
[info] AsyncBenchmark.race                       N/A      100  thrpt   20   3799.585 ±  112.200    ops/s
[info] AsyncBenchmark.racePair                   N/A      100  thrpt   20  26703.708 ±  307.931    ops/s
[info] AsyncBenchmark.start                      N/A      100  thrpt   20   3705.231 ±  471.800    ops/s
[info] DeepBindBenchmark.async                   N/A    10000  thrpt   20    846.345 ±  158.343    ops/s
[info] DeepBindBenchmark.delay                   N/A    10000  thrpt   20   1942.372 ±  258.635    ops/s
[info] DeepBindBenchmark.pure                    N/A    10000  thrpt   20   1937.306 ±  223.508    ops/s
[info] ParallelBenchmark.parTraverse           10000     1000  thrpt   20    210.981 ±    3.575    ops/s
[info] ShallowBindBenchmark.async                N/A    10000  thrpt   20    482.870 ±   78.723    ops/s
[info] ShallowBindBenchmark.delay                N/A    10000  thrpt   20   2005.160 ±  321.750    ops/s
[info] ShallowBindBenchmark.pure                 N/A    10000  thrpt   20   2089.791 ±  237.237    ops/s
[info] WorkStealingBenchmark.alloc               N/A  1000000  thrpt   20      2.736 ±    0.054  ops/min
[info] WorkStealingBenchmark.scheduling          N/A  1000000  thrpt   20      9.891 ±    0.353  ops/min
```

Intel(R) Core(TM) i7-4810MQ CPU @ 2.80GHz

`benchmarks/Jmh/run -wi 10 -i 10 -f 2 -t 1 WorkStealingBenchmark.scheduling WorkStealingBenchmark.alloc DeepBindBenchmark.async DeepBindBenchmark.delay DeepBindBenchmark.pure ShallowBindBenchmark.async ShallowBindBenchmark.delay ShallowBindBenchmark.pure AsyncBenchmark.async AsyncBenchmark.race AsyncBenchmark.racePair AsyncBenchmark.start ParallelBenchmark.parTraverse`


core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -408,6 +410,15 @@ private[effect] final class WorkStealingThreadPool(
...
   }
 
+  /**
+   * Determines whether all worker threads are currently parked, which may indicate synchronous,
+   * compute-heavy fiber usage.
+   * @return
+   *   true if all worker threads are currently parked
+   */
+  private[this] def allWorkersParked: Boolean =
+    ((state.get() & UnparkMask) >>> UnparkShift) == threadCount
```
vasilmkd: What this expression does, and what the method says it does is not the same. The `& UnparkMask` and `>>> UnparkShift` combo returns the number of _active_ (or _unparked_) threads.

We have the `notifyShouldWakeup()` method that does mostly the same, but also takes into considerations searching threads. Those are threads which are looking or in the process of stealing fibers from other worker threads. At those points in time, the contention is actually increased in the pool and we use that method to be careful not to increase it further.

I suggest that we delete this method.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -398,7 +400,7 @@ private[effect] final class WorkStealingThreadPool(
 
     if (thread.isInstanceOf[WorkerThread]) {
       val worker = thread.asInstanceOf[WorkerThread]
-      if (worker.isOwnedBy(pool)) {
+      if (worker.isOwnedBy(pool) && !allWorkersParked) {
```
vasilmkd:
```suggestion
      if (worker.isOwnedBy(pool) && !notifyShouldWakeup()) {
```

The idea here is sound. Let's hope the benchmarks agree with us.

vasilmkd (changes requested):
This is very good. We will need to consult with the following benchmarks:

```
WorkStealingBenchmark.scheduling
WorkStealingBenchmark.alloc
DeepBindBenchmark.async
DeepBindBenchmark.delay
DeepBindBenchmark.pure
ShallowBindBenchmark.async
ShallowBindBenchmark.delay
ShallowBindBenchmark.pure
AsyncBenchmark.async
AsyncBenchmark.race
AsyncBenchmark.racePair
AsyncBenchmark.start
ParallelBenchmark.parTraverse
```

You can run the benchmarks from `sbt` using:
```
sbt:root> benchmarks/Jmh/run -wi 10 -i 10 -f 2 -t 1 <benchmark name from above>
```

vasilmkd:
As an added optimization, you should be able to remove this line, because it makes the same check (which is not completely free in terms of performance as it does a read of a volatile variable), and we're making the bet that we're only scheduling on the same worker thread if every other thread is busy.

https://github.com/typelevel/cats-effect/blob/b2276e5a96766fa9df9a86999cc1be246a43bd06/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala#L102

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
otto-dev: Thanks, I have been looking into it now. I don't think we want to use the external queue only if `((st & UnparkMask) >>> UnparkShift) < threadCount`. On the contrary, I'm getting the best results when using the condition `((st & UnparkMask) >>> UnparkShift) == treadCount` instead, which also makes the most sense to me (I.e. no threads are searching, all workers are unparked, means computation heavy fibers). Doing so does not only increase fairness, but preliminary results show that it increases throughput in existing benchmarks.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
otto-dev: Right, I totally misread the domain language. Anyway, we might still need the corrected version of this method (comment below)

otto-dev:
> you should be able to remove this line `pool.notifyParked(rnd)`

Thanks. Can I remove this even if I leave `rescheduleFiber` untouched, and only modify `rescheduleFiber`?

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
otto-dev: For reference, this is the fairness with `!notifyShouldWakeup()`

image

This is the fairness with (now) `!allWorkersBusy()`

image

Preliminary results show an increase in performance, in addition to the increase in fairness.

vasilmkd:
Thanks for taking this on! The graphs looks very encouraging.

> Thanks. Can I remove this even if I leave rescheduleFiber untouched, and only modify rescheduleFiber?

To be completely safe, remove the line from `WorkerThread#schedule` and add it after the call to `schedule` in `WorkerThread#reschedule`. It would also be fun to try the benchmarks without adding the line back in `reschedule`.

otto-dev:
First benchmark result is in, updated the description. (took 30 minutes for both runs, so I'm setting up the other benchmarks separately) (Edit: on hold)

otto-dev:
> we're making the bet that we're only scheduling on the same worker thread if every other thread is busy.

Just trying to work this out: I would have said "we're making the bet that we're **_not_** scheduling on the same worker thread if **_all_** threads are busy". That's at least my intent. Because if all threads are busy, you can have a situation where none of them steal work. So scheduling on the same worker thread (rather than through the external queue) can lead to an unfair distribution of fibers in such a situation. So in that situation we schedule through the external queue.

So in that case, if I read the code correctly, we might have to leave that line in. Please do correct me if I'm missing something!

Running the second set of benchmarks now. Looking promising so far.

vasilmkd:
I'm just going to comment, this change removes the thread affinity of the pool (at full loads, fibers will be jumping around the pool all the time). Which benchmarks exactly are showing peformance improvements?

djspiewak:
I need to think about this more deeply, but at first glance it seems a little concerning. It seems like this would result in child fibers losing all affinity to worker threads, which is an optimization that is not well captured by the current benchmarks. Thread affinity is impactful because the OS kernel will *also* try to do its own version of the same optimization, which is to say, keeping threads affine to host CPUs as much as possible. These two optimizations together improve cache utilization considerably, since child fibers are very likely to access state from their parent(s), which in turn will likely be in cache on the parent's CPU but not on other CPUs.

On the other side of the spectrum, I'm not entirely convinced that this covers a real use-case. We have a nice artificial test which can reproduce this scenario, but even in that scenario, throughput is still optimal, and presumably any truly compute-bound situation is going to care more about throughput than fairness. In any fairness-relevant scenario, other actions (such as `async` completion and internal forking) would take place which trigger the pool to balance on its own, meaning that proactively `start`ing on the external queue is unnecessary contention, since stealing will ultimately happen on its own.

Those are just initial thoughts though. Need to ponder some more.

otto-dev:
> Which benchmarks exactly are showing peformance improvements?

I think this is irrelevant, but here is a list of benchmarks that performed better by score:
- `AsyncBenchmark.race`
- `DeepBindBenchmark.async`
- `DeepBindBenchmark.delay`
- `ShallowBindBenchmark.async`
- `ShallowBindBenchmark.delay`
- `ShallowBindBenchmark.pure`
- `WorkStealingBenchmark.alloc`
- `WorkStealingBenchmark.scheduling`

All differences (better or worse) are within the margin of `error`, and the average score is 101.5% of the previous version. This PR is not meant to improve performance, and I'd argue that it neither improves it, nor makes it worse.

> at full loads, fibers will be jumping around the pool all the time

If that is the case, it doesn't show in the benchmarks, which are supposed to create the highest load possible. This PR only has an effect on computation-bound applications that lack asynchronous components. I don't share this concern.

If there is a particular benchmark you are concerned about, I'm happy to run it with more iterations.

vasilmkd:
The `WorkStealingBenchmark.alloc` and `WorkStealingBenchmark.scheduling` call `cede` every other operation. This _only_ exercises `reschedule`, not `schedule`. The new code isn't being run on them, that's why the results are not changing. I doubt that they are improving because either, again, because no new code is being run.

otto-dev:
Is there any benchmark that would highlight your concern, or should we add a new one?

vasilmkd:
Do you have the scores also? Thanks.

otto-dev:
It's in the `after` section: https://github.com/typelevel/cats-effect/pull/2514#issue-774926194

vasilmkd:
Ok, I've given this some thought and I think I can explain what's going on.

Let's first start with the issue as was originally reported. The test code starts running on a single fiber, which is enqueued by the JVM `main` fiber on the `external` thread of the work stealing thread pool and a single worker thread is woken up to notify it that new work is available. This thread polls the external queue and finds the fiber. It executes it. The fiber starts the fibers that will be running forever.

Here, we arrive to the sequence of operations that creates the situation described in the issue. Each time a `start` is executed by the original worker thread that started the original fiber, the newly created fiber is enqueued on the _same_ worker thread and each time this happens, another worker thread is awoken. Since _all_ of the work available on the pool is on the original worker thread, in its local queue, the thread that is awoken starts the stealing process. This is non-deterministic of course, but because there are a few active threads, each new thread will actually only be able to steal 1-2 fibers from the original worker thread. And then, because all threads have fibers to execute (and the fibers never end, only `cede`), everyone has work to do, so they never attempt to steal again. Again, because the fibers run forever _and_ there are relatively few fibers in the pool (20, 40, even 100 is actually a very low number of fibers), the fibers are never rebalanced, seemingly because there is no need for it. In this situation, throughput is still maximized, because you have all worker threads working at 100% always having something more to do, because the fibers are ceding forever. However, indeed, we are suffering from a fairness issue, because 1 thread is executing 3x the fibers of other threads (for example, it is non-deterministic in nature).

Now, what the change does is the following. It changes how fibers are _started_ on the pool. As the change stands at the moment that this comment is written, if at the point of _starting_ a new fiber, there are threads that are sleeping, it will use the same thread to schedule it and awake other threads to steal the fiber. In other words, on low loads of the pool, the change behaves _exactly_ like the situation that created the fairness issue.

However, what the change actually does is change how fibers are started when all threads have started and instead of the current worker thread, it sends them to the external queue. In a concrete example, if you have 8 threads and 20 fibers, the first 8 fibers will be started on the current thread and all others on the external queue. Furthermore, worker threads are checking the external queue every 61 iterations. They will find the fibers on the external queue, and because each time a single fiber is polled from the external queue, the fibers will more or less balance properly on the worker threads.

One further thing this change does is how it resumes fibers that were asynchronously suspended. If the async callback is executed by a foreign thread, the fiber will be put on the external thread and if the async callback was executed internally, the fiber would be put on the same worker thread. If I remember correctly, I added this behavior _before_ I developed the `ScalQueues` and before the `batched` queue even existed, which was later merged with the external queue. At that time, it made sense, because _stealing_ was less contended than the external queue which wasn't striped like it is now. So at that point, we measured that the current behavior on `series/3.x` was better. However, since the `ScalQueue` striped queue was developed, I guess this experiment wasn't revised. It might just be the case that the external queue is now less contended and can be done faster in practice than stealing.

I was wrong to suggest that this change removes the thread-affinity of fibers, because it doesn't touch `reschedule`. I.e., when a fiber is picked up by a thread, it will stay on it until the end, provided it only does synchronous work and cedes. The difference is of course that the thread this happens on may not be the thread where the fiber that started it ran. I apologize for the confusion I caused there.

The reason why the current behavior on `series/3.x` intuitively makes sense is that fibers started on the same worker thread may get more cache hits because they are likely to have some data in common with the fiber that started them. How realistic this is in practice, I do not know, and frankly I'm not sure we're operating at a level low enough to show these minute differences.

If it were up to me, I would rather simplify this whole code and just start fibers on the external queue, if we can prove that it is indeed faster. I haven't made up my mind yet. I will think about the problem a bit more.

otto-dev:
That's a reasonable description of the intent of this PR @vasilmkd. This is how I summed it up above:
>  if all threads are busy, you can have a situation where none of them steal work

Re:
> just start fibers on the external queue

I think that's that's a significant difference to what this PR does, I could imagine a degradation in performance (needs to be tested). Currently it only uses the external queue if if all threads are unparked an none of them are searching, indicating that we possibly have the non-work-stealing scenario.

Anyway, I might write up a little benchmark later, that's more realistic than the code in the original issue, involving data being pulled from a queue and many thousands of fibers.

vasilmkd:
That would be great. Looking forward to the benchmarks. I'm not opposed to any changes here, I'm just slightly overwhelmed at the moment with other things and I'm also trying to be of help as much as I can. Thank you for your patience.

otto-dev:
In all benchmarks that I have seen, throughput, in the sense of productive CPU utilization, remains unaffected.

But there is at least one benchmark that's interesting, which I have developed, and it measures _fairness_, not _throughput_. This is important to keep in mind, because it requires a significant mind-shift.

Here is how it works:
- The benchmark starts up `2 * cores` CPU intensive fibers, like they could spawn temporarily at times in a real-life application.
- There is an asynchronous `Queue` of "Requests" and a `Queue` of "Responses". Each request is pulled off the queue in sequence, a bunch of temporary fibers are spawned to compute the response, which is then pushed onto the other queue. This is analogous to answering expensive network requests.
- Fairness is measured by measuring the throughput of request-responses, which excludes the fibers that have been initially spawned in step 1). So in other words, it measures how much the first fibers are able to clog up the scheduler unfairly (or rather, how much they don't).

Unfortunately, as you can see below, there is no obviously right answer to which version it better. Based on other benchmarks, I expect that effective CPU utilization remains constant before and after this PR. However, the answer to the question which version is fairer depends on how CPU intensive each fiber is!

## Benchmark

[Link to code](https://gist.github.com/otto-dev/6f00699fadbc2dd90761a174790ee493)

"Computation intensity" means how many Murmur3 integers each fiber calculates.

### Computation intensity per fiber: 10000
#### Before
```
[info] Benchmark                     (size)   Mode  Cnt  Score   Error    Units
[info] FairnessBenchmark.scheduling     100  thrpt   10  7.054 ± 0.867  ops/min
```

#### After
```
[info] Benchmark                     (size)   Mode  Cnt   Score   Error    Units
[info] FairnessBenchmark.scheduling     100  thrpt   10  13.221 ± 1.616  ops/min
```

### Computation intensity per fiber: 10
#### Before
```
[info] Benchmark                     (size)   Mode  Cnt     Score    Error    Units
[info] FairnessBenchmark.scheduling     100  thrpt   10  2588.254 ± 58.580  ops/min
```

#### After
Computation intensity per fiber: 10
```
[info] Benchmark                     (size)   Mode  Cnt     Score    Error    Units
[info] FairnessBenchmark.scheduling     100  thrpt   10  1419.092 ± 43.746  ops/min
```

As you can see, this PR is 87% fairer when each fiber calculates 10000 Murmur3 hashcodes, but the current `3.x` version is 82% fairer when each fiber only calculates 10 hashcodes, and there is a spectrum inbetween.


---

### Appendix

Approximate fairness improvement of this PR expressed as percentage compared to the "before" value, with Computation intensity as `x`.

image


PS: The outcome of this is far removed from my first "fears" when I did the `cede` experiment and saw many fibers starved. This shows that in practice, this does not happen. The only question that's being adressed now is whether/how fairness can be improved as an optimization, not as a repair.

otto-dev:
To conclude this PR, I ran the same benchmark, measuring throughput, not fairness, to confirm that the change in fairness does not come at the expense of throughput. To do so I removed the `createUnfairness` part. [Link to benchmark.](https://gist.github.com/otto-dev/c60383d4653436fe748164a7cf782fca). Also, `size` here is now referring to "Computation intensity" above.

Results (merged code with `series/3.x` as of today):

**Before**
```
[info] Benchmark                       (size)   Mode  Cnt     Score      Error    Units
[info] ThroughputBenchmark.scheduling       1  thrpt   20  4503.731 ±  221.476  ops/min
[info] ThroughputBenchmark.scheduling      10  thrpt   20  5410.462 ± 1208.783  ops/min
[info] ThroughputBenchmark.scheduling     100  thrpt   20  5403.228 ±  160.312  ops/min
[info] ThroughputBenchmark.scheduling    1000  thrpt   20   881.516 ±    6.627  ops/min
[info] ThroughputBenchmark.scheduling   10000  thrpt   20    95.048 ±    0.372  ops/min
[info] ThroughputBenchmark.scheduling  100000  thrpt   20     9.345 ±    0.080  ops/min
```

**After**
```
[info] Benchmark                       (size)   Mode  Cnt     Score      Error    Units
[info] ThroughputBenchmark.scheduling       1  thrpt   20  4225.515 ±  185.137  ops/min
[info] ThroughputBenchmark.scheduling      10  thrpt   20  5385.358 ± 1125.201  ops/min
[info] ThroughputBenchmark.scheduling     100  thrpt   20  5776.103 ±  104.549  ops/min
[info] ThroughputBenchmark.scheduling    1000  thrpt   20   907.226 ±   15.629  ops/min
[info] ThroughputBenchmark.scheduling   10000  thrpt   20    98.211 ±    1.207  ops/min
[info] ThroughputBenchmark.scheduling  100000  thrpt   20     9.335 ±    0.158  ops/min
```

This was the last thing I was going to do for this PR. My conclusion:

The PR noticeably improves fairness for computation intensive applications while having no significant impact on throughput.

There is a slight fairness reduction for non-computation intensive applications. But that doesn't matter. If you have millions of "light" short-lived fibers, you are not going to see that, because throughput stays the same. But if you have thousands of long-lived computation intensive fibers, you may (e.g. an AI driven trading bot).

That's it from me. I leave it to the maintainers to make a decision.

PS: I don't know why very short-lived fibers have less throughput than fibers that do a bit more. I re-ran this to confirm, and this really is the case in both situations.

djspiewak:
This is awesome information gathering. Need to think about this a bit…

djspiewak:
I had to think about this for quite a while, but I think I finally know what's going on and why it defies intuitions. It comes down to two things.

First, `Queue` is built on `Deferred`, which in turn uses `async` under the surface to fulfill continuations on `poll`. `async` automatically shifts on completion, which means it is *tempting* to view this as a reasonable way of approximating external NIO calls. But it isn't. The problem is that the automatic shifting inside of `async` has a check to see if the callback is being fulfilled *from* a worker thread, which in this case, it is! This means that `poll` is never going to shift, which in turn means that the "external source of work" is actually an *internal* source of work.

Which brings us back to the unfairness. By pushing `start` onto the external queue, what you're actually doing here is sending all the murmur load to the back of the line, and so *naturally* the measured work is completed sooner. That's not a sign that fairness is improved, though, it's just a sign that the runtime has been tweaked in such a way that it encodes a priority queue with respect to this exact test.

I'm pretty sure that's the *main* thing that's going on here. To fix this, the `Queue` would need to be replaced with a `Dispatcher`, which in turn would need to be fulfilled from another pool (`blocking` is the easiest thing). Fixing this and then measuring before/after may still show some advantage to pushing `start` to an external queue, but it's going to be WAY less significant, particularly since `Dispatcher` itself needs `start` to function.

The secondary thing that's happening here is the use of `blocking` around `consumeCPU`, which I think is probably wrong. `BlackHole` isn't a blocking action, it's just a CPU sink, similar to the murmur hash, so it should be on the compute pool. Pushing it to the blocking pool creates some other effects which are hard to model and are almost certainly conflated with the measurement, so it would probably be better to just not do that.

Anyway, I still think this is a *really* interesting line of investigation, but I'm pretty sure that it doesn't yet prove that `start` should be on the external queue.

djspiewak:
I'm going to close this. While I think this was an absolutely fantastic line of investigation, we weren't able to really come up with a test which showed a meaningful improvement to come from the change. In the meantime, several other scenarios have come up which demonstrate some of the value of treating `async` continuations as internal when they come from compute-bound threads, particularly related to `Queue`s and `Dispatcher`.

As for `start`, its thread affinity is harder to reason about, because it comes down to what people do with fibers *in practice*, and in practice, most people aren't using `start` directly but rather are using some of the `par` operators, or higher-level variants thereof in Fs2. In these cases, fibers are very much an implementation detail around something which is semantically a continuation of what came before, and thus all the same arguments about affinity apply. If we had a long series of sequential `parMapN`s with meaningful working sets, I suspect we would be able to reproduce this intuition in benchmark form.

Which generally brings me to a larger conclusion: we should try to capture some of this tuning around affinity in objective form. The work I've been doing on #2771 certainly provides some of that, but I think we can probably go further. It would also be really good to capture some of what was learned here in documentation form, since the insights in this PR are extremely profound.

Issue #1838, opened 2021-03-29
Dispatcher docs don't say how to get one


kubukoz:
(on the `docs` branch)

LLCampos:
Some feedback from someone who has been struggling to understand `Dispatcher`.

For example, the **Motivation section** basically says "this does what this other thing did in cats-effect 2" which is is not really useful for people who are not experienced in CE2. :crying_cat_face: 

I think that having an example of a practical problem that can be solved with Dispatcher would be excellent!

And also a little explanation of exactly **how** it works in the inside - right now, it fells like magic to me. :)

kubukoz:
Personally I think none of the docs should refer to CE2 except the migration guide 😅

LLCampos:
We can close this one? https://github.com/typelevel/cats-effect/pull/2515 already merged.

armanbilge:
And live! :)
https://typelevel.org/cats-effect/docs/std/dispatcher

Issue #2448, opened 2021-10-21
Make dispatcher safer 


Baccata:
The current Dispatcher suffers from the following problems : 

* algebraic product components of the effect are lost at the point of dispatch (WriterT's log) 
* algebraic coproduct components (EitherT, OptionT) of the effect put the dispatch call at risk of never returning. 
* self cancellation of dispatched computations prevent the dispatch call from ever returning

This last two points are pretty bad : even if Dispatcher can in theory work against any effect type, in practice it makes it really easy for the user to shoot themselves in the foot. 

The problem could be mitigated by requiring the dispatch methods (`unsafeRunSomething`) to take a callback construct, that'd allow to witness these three above points : 

```scala 
trait DispatchObserver[F[_]] {

  def onCancel: F[Unit]

  // When the computation returns an algebraic failure and a value
  // cannot be produced (for instance OptionT.none)
  def onAlgebraicFailure(failed: F[Unit]): F[Unit]

  // When the computation returns an algebraic success.
  def onAlgebraicSuccess(success: F[Unit]): F[Unit]

}
```

See a POC [here](https://github.com/Baccata/dispatch-bridge-experiment/blob/main/src/main/scala/SaferDispatcher.scala#L12-L22) 

The incorporation of such a callback would facilitate writing constructs like [this one](https://github.com/Baccata/dispatch-bridge-experiment/blob/main/src/main/scala/DispatchBridge.scala#L50-L55), allowing for communicating cancellation and algebraic information through unsafe regions, in a way that is similar to what [cats-effect-cps](https://github.com/typelevel/cats-effect-cps) does today. 

A concrete benefit would be a better integration with libraries like https://github.com/sangria-graphql/sangria, which force an unsafe region onto the user.

djspiewak:
Isn't this isomorphic to `unsafeRunOutcome[A](fa: F[A]): Outcome[F, E, A]`?

Baccata:
@djspiewak it is not, because what do you do with the effect once when you get a successful outcome ? You still have to run it through the dispatcher to get the value, and if the F[A] is `OptionT.none`, you're still f***d.

The outcome does let you detect cancellation though, but then what do you do with it ? You're in unsafe land with that information, so you need to make another dispatcher call to send it over to safe land, or pass some exception to signal cancellation. The latter is fair enough I guess, but the former is not great due to the having to repeat the dispatch.

[...]

PR #2515, opened 2021-11-07
Some tentative improvements to `Dispatcher` docs

LLCampos:
Inspired by https://github.com/typelevel/cats-effect/issues/1838.

### DIFF
--- a/docs/std/dispatcher.md
+++ b/docs/std/dispatcher.md
@@ -5,7 +5,7 @@ title: Dispatcher
 
 ![](assets/dispatcher.jpeg)
 
-`Dispatcher` is a fiber-based [Supervisor](https://typelevel.org/cats-effect/docs/std/supervisor) utility for evaluating effects across an impure boundary. 
+`Dispatcher` is a fiber-based utility for evaluating effects across an impure boundary. An instance of `Dispatcher` can be derived for any effect type conforming to the `Async` typeclass.
 This is useful when working with reactive interfaces that produce potentially many values (as opposed to one), and for each value, some effect in `F` must be performed 
 (like inserting each value into a queue).
 
--- a/docs/std/dispatcher.md
+++ b/docs/std/dispatcher.md
@@ -6,7 +6,7 @@ title: Dispatcher
 ![](assets/dispatcher.jpeg)
 
 `Dispatcher` is a fiber-based [Supervisor](https://typelevel.org/cats-effect/docs/std/supervisor) utility for evaluating effects across an impure boundary. 
-This is useful when working with reactive interfaces that produce potentially many values (as opposed to one), and for each value, some effect in F must be performed 
+This is useful when working with reactive interfaces that produce potentially many values (as opposed to one), and for each value, some effect in `F` must be performed 
 (like inserting each value into a queue).
 
 Let's say we are integrating with an interface that looks like this:
--- a/docs/std/dispatcher.md
+++ b/docs/std/dispatcher.md
@@ -23,7 +23,7 @@ abstract class ImpureInterface {
 }
 ```
 
-As an user we're supposed to create an implementation of this class where we implement `onMessage`.
+As a user we're supposed to create an implementation of this class where we implement `onMessage`.
 
 Let's say that we have a [`queue`](https://typelevel.org/cats-effect/docs/std/queue) and that we want to add each message received to the queue. We might,
 naively, try to do the following
@@ -44,27 +44,28 @@ for {
 
 This will print "Value not found in queue :(", since our implementation of `onMessage` 
 doesn't really *do* anything. It just creates an `IO` value which is just a description of an effect,
-without actually executing it. (Here the `scalac` option `-Ywarn-value-discard` might help hint this problem.)
+without actually executing it. (Here the `scalac` option `-Ywarn-value-discard` might help hint at this problem.)
 
 It is in these cases that `Dispatcher` comes in handy. Here's how it could be used:
 
 ```scala
 Dispatcher[IO].use { dispatcher =>
-    for {
-      queue <- Queue.unbounded[IO, String]
-      impureInterface <-
-        IO.delay {
-          new ImpureInterface {
-            override def onMessage(msg: String): Unit =
-              dispatcher.unsafeRunSync(queue.offer(msg))
-          }
+  for {
+    queue <- Queue.unbounded[IO, String]
+    impureInterface <-
+      IO.delay {
+        new ImpureInterface {
+          override def onMessage(msg: String): Unit =
+            dispatcher.unsafeRunSync(queue.offer(msg))
         }
-      _ <- IO.delay(impureInterface.init())
-      value <- queue.tryTake
-    } yield value match {
-      case Some(v) => println(s"Value found in queue! $v")
-      case None    => println("Value not found in queue :(")
-    }
+      }
+    _ <- IO.delay(impureInterface.init())
+    value <- queue.tryTake
+  } yield value match {
+    case Some(v) => println(s"Value found in queue! $v")
+    case None    => println("Value not found in queue :(")
+  }
+}
 ```
 
 It prints "Value found in queue! init"!
@@ -93,14 +94,14 @@ trait Dispatcher[F[_]] extends DispatcherPlatform[F] {
 }
 ```
 
-An instance of `Dispatcher` is very cheap - it is encouraged to instantiate it 
+Creating an instance of `Dispatcher` is very cheap - you are encouraged to instantiate it 
 where necessary rather than wiring a single instance throughout an application.
 
 
-## Cats-Effect 2
+## Cats Effect 2
 
-Users of cats effect 2 may be familiar with the `Effect` and `ConcurrentEffect`
-typeclasses. These have been removed as they contrained implementations of the
+Users of Cats Effect 2 may be familiar with the `Effect` and `ConcurrentEffect`
+typeclasses. These have been removed as they constrained implementations of the
 typeclasses too much by forcing them to be embeddable in `IO` via `def
 toIO[A](fa: F[A]): IO[A]`. However, these typeclasses also had a valid use-case
 for unsafe running of effects to interface with impure APIs (`Future`, `NIO`,
--- a/docs/std/dispatcher.md
+++ b/docs/std/dispatcher.md
@@ -5,20 +5,70 @@ title: Dispatcher
 
 ![](assets/dispatcher.jpeg)
 
-## Motivation
+`Resource` is a fiber-based [Supervisor](https://typelevel.org/cats-effect/docs/std/supervisor) utility for evaluating effects across an impure boundary. 
+This is useful when working with reactive interfaces that produce potentially many values (as opposed to one), and for each value, some effect 
+in F must be performed (like inserting it into a queue).
 
-Users of cats effect 2 may be familiar with the `Effect` and `ConcurrentEffect`
-typeclasses. These have been removed as they constrained implementations of the
-typeclasses too much by forcing them to be embeddable in `IO` via `def
-toIO[A](fa: F[A]): IO[A]`. However, these typeclasses also had a valid use-case
-for unsafe running of effects to interface with impure APIs (`Future`, `NIO`,
-etc).
+Let's say we are integrating with an interface that looks like this:
 
-## Dispatcher
+```scala
+abstract class ImpureInterface {
+  def onMessage(msg: String): Unit
 
-`Dispatcher` addresses this use-case but can be constructed for any effect
-type with an `Async` instance, rather than requiring a primitive typeclass
-implementation.
+  def init(): Unit = {
+    onMessage("init")
+  }
+
+  // ...
+}
+```
+
+As an user we're supposed to create an implementation of this class where we implement `onMessage`.
+
+Let's say that we have a [`queue`](https://typelevel.org/cats-effect/docs/std/queue) and that we want to add each message received to the queue. We might,
+naively, try to do the following
+
+```scala
+for {
+    queue <- Queue.unbounded[IO, String]
+    impureInterface = new ImpureInterface {
+      override def onMessage(msg: String): Unit = queue.offer(msg) // This returns an IO, so nothing really happens!
+    }
+    _ <- IO.delay(impureInterface.init())
+    value <- queue.tryTake
+  } yield value match {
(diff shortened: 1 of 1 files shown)


docs/std/dispatcher.md
```diff
@@ -5,20 +5,70 @@ title: Dispatcher
 
 ![](assets/dispatcher.jpeg)
 
-## Motivation
+`Resource` is a fiber-based [Supervisor](https://typelevel.org/cats-effect/docs/std/supervisor) utility for evaluating effects across an impure boundary.
```
vasilmkd: You probably meant `Dispatcher` instead of `Resource`?

docs/std/dispatcher.md
```diff
@@ -5,20 +5,70 @@ title: Dispatcher
 
 ![](assets/dispatcher.jpeg)
 
-## Motivation
+`Resource` is a fiber-based [Supervisor](https://typelevel.org/cats-effect/docs/std/supervisor) utility for evaluating effects across an impure boundary. 
+This is useful when working with reactive interfaces that produce potentially many values (as opposed to one), and for each value, some effect 
+in F must be performed (like inserting it into a queue).
```
vasilmkd: like inserting each value into a queue

docs/std/dispatcher.md
```diff
@@ -5,20 +5,70 @@ title: Dispatcher
...
+    _ <- IO.delay(impureInterface.init())
+    value <- queue.tryTake
+  } yield value match {
+    case Some(v) => println(s"Value found in queue! $v")
+    case None    => println("Value not found in queue :(")
+  }
+```
+
+This will print "Value not found in queue :(", since our implementation of `onMessage` 
+doesn't really *do* anything. It just creates an `IO`.
```
vasilmkd: It just creates an `IO` value, which is just a description of an effect, without actually executing it.

vasilmkd: Here the `scalac` option `-Ywarn-value-discard` might help hint this problem.

docs/std/dispatcher.md
```diff
@@ -43,6 +93,15 @@ trait Dispatcher[F[_]] extends DispatcherPlatform[F] {
 ```
 
 An instance of `Dispatcher` is very cheap - it allocates a single fiber so it
-is encouraged to instantiate it where necessary,
-as close to the unsafe region as possible,
-rather than wiring a single instance throughout an application.
+is encouraged to instantiate it where necessary rather than wiring
```
vasilmkd: Let's remove the `it allocates a single fiber` part of the sentence, since this is no longer true. Let's go with it's encouraged to instantiate it...

docs/std/dispatcher.md
```diff
@@ -43,6 +93,15 @@ trait Dispatcher[F[_]] extends DispatcherPlatform[F] {
...
+
+
+## Cats-Effect 2
+
+Users of cats effect 2 may be familiar with the `Effect` and `ConcurrentEffect`
+typeclasses. These have been removed as they contrained implementations of the
+typeclasses too much by forcing them to be embeddable in `IO` via `def
+toIO[A](fa: F[A]): IO[A]`. However, these typeclasses also had a valid use-case
+for unsafe running of effects to interface with impure APIs (`Future`, `NIO`,
+etc).
```
vasilmkd: The last sentence sounds like this is no longer true. Does `Dispatcher` not cover these cases?

vasilmkd (changes requested):
This is amazing, thank you so much!

I just added some small improvements, hopefully you will like them.

Please also rebase this PR on `series/3.2.x` and change the base of this PR to that branch since that's where we're publishing stable documentation from.

Thank you.

docs/std/dispatcher.md
LLCampos: This part was copied from the current docs. It made more sense in that context since it was followed by "`Dispatcher` now addresses this use-case.". I'm going to add that sentence.

LLCampos:
@vasilmkd, thanks! Applied the recommended changes. :)

docs/std/dispatcher.md
```diff
@@ -5,20 +5,71 @@ title: Dispatcher
 
 ![](assets/dispatcher.jpeg)
 
-## Motivation
+`Dispatcher` is a fiber-based [Supervisor](https://typelevel.org/cats-effect/docs/std/supervisor) utility for evaluating effects across an impure boundary.
```
vasilmkd: I'm not completely sure we should link `Dispatcher` to `Supervisor` just yet. Unfortunately, `Dispatcher` does not offer the backpressure that `Supervisor` does on scope exit, yet. We have planned improvements to `Dispatcher` coming most likely in the Cats Effect `3.4.x` lineage of releases.

For now, I would prefer if we used the old wording that `Dispatcher` is a functional interface with derivable instances for any effect type conforming to the `Async` typeclass.

armanbilge: Huh, I didn't realize that ... what guarantees does `Dispatcher` offer on scope exit then?

Btw this paragraph is taken straight from the current scaladocs actually 😆

armanbilge: Yeah, seems like we should fix the scaladocs ...
https://github.com/typelevel/cats-effect/blob/b2276e5a96766fa9df9a86999cc1be246a43bd06/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala#L37-L38

docs/std/dispatcher.md
armanbilge: I think this also needs to be updated in the scaladocs?

docs/std/dispatcher.md
```diff
@@ -5,20 +5,71 @@ title: Dispatcher
 
 ![](assets/dispatcher.jpeg)
 
-## Motivation
+`Dispatcher` is a fiber-based [Supervisor](https://typelevel.org/cats-effect/docs/std/supervisor) utility for evaluating effects across an impure boundary. 
+This is useful when working with reactive interfaces that produce potentially many values (as opposed to one), and for each value, some effect in F must be performed
```
armanbilge:
```suggestion
This is useful when working with reactive interfaces that produce potentially many values (as opposed to one), and for each value, some effect in `F` must be performed 
```

docs/std/dispatcher.md
vasilmkd: Nothing, there are reported gotchas. The worst thing that happens I guess is that a submitted effect doesn't get a chance to start at all.

armanbilge: > The worst thing that happens I guess is that a submitted effect doesn't get a chance to start at all.

I see. So but does it backpressure on _running_ effects? It's just it may close before the queue is empty, hence a submitted effect doesn't start? And does it not start silently, or with a "dispatcher closed" error (I've gotten a few of those in my day :).

vasilmkd: I honestly cannot answer this question correctly. 😕

vasilmkd:
Yeah, the last thing here to do would be to amend the `Supervisor` mention, because `Dispatcher` really isn't a valid instance of a `Supervisor`. When we change that, we'll update the docs along with it.

armanbilge:
I'm looking at the scaladocs for Supervisor and Dispatcher. It seems, maybe the difference is that Supervisor backpressures for _completion_ of submitted effects, whereas Dispatcher backpressures for _cancellation_ of submitted effects?

```scala
  /**
   * Create a [[Dispatcher]] that can be used within a resource scope. Once the resource scope
   * exits, all active effects will be canceled, and attempts to submit new effects will throw
   * an exception.
   */
```

```scala
  /**
   * Creates a [[cats.effect.kernel.Resource]] scope within which fibers can be monitored. When
   */
```

docs/std/dispatcher.md
```diff
@@ -5,20 +5,71 @@ title: Dispatcher
...
-implementation.
+  def init(): Unit = {
+    onMessage("init")
+  }
+
+  // ...
+}
+```
+
+As an user we're supposed to create an implementation of this class where we implement `onMessage`.
```
armanbilge:
```suggestion
As a user we're supposed to create an implementation of this class where we implement `onMessage`.
```

docs/std/dispatcher.md
```diff
@@ -5,20 +5,71 @@ title: Dispatcher
...
+    value <- queue.tryTake
+  } yield value match {
+    case Some(v) => println(s"Value found in queue! $v")
+    case None    => println("Value not found in queue :(")
+  }
+```
+
+This will print "Value not found in queue :(", since our implementation of `onMessage` 
+doesn't really *do* anything. It just creates an `IO` value which is just a description of an effect,
+without actually executing it. (Here the `scalac` option `-Ywarn-value-discard` might help hint this problem.)
```
armanbilge:
```suggestion
without actually executing it. (Here the `scalac` option `-Ywarn-value-discard` might help hint at this problem.)
```

docs/std/dispatcher.md
```diff
@@ -5,20 +5,71 @@ title: Dispatcher
...
+              dispatcher.unsafeRunSync(queue.offer(msg))
+          }
+        }
+      _ <- IO.delay(impureInterface.init())
+      value <- queue.tryTake
+    } yield value match {
+      case Some(v) => println(s"Value found in queue! $v")
+      case None    => println("Value not found in queue :(")
+    }
+```
```
armanbilge, line 68:
```suggestion
```scala
Dispatcher[IO].use { dispatcher =>
  for {
    queue <- Queue.unbounded[IO, String]
    impureInterface <-
      IO.delay {
        new ImpureInterface {
          override def onMessage(msg: String): Unit =
            dispatcher.unsafeRunSync(queue.offer(msg))
        }
      }
    _ <- IO.delay(impureInterface.init())
    value <- queue.tryTake
  } yield value match {
    case Some(v) => println(s"Value found in queue! $v")
    case None    => println("Value not found in queue :(")
  }
}
```
```

**Comment by vasilmkd** [comment]
Not really, cancelation is implemented as enqueueing another effect that cancels the first. And there's no guarantee that that will get executed, like any other effect.

The funny thing is, even with all of these limitations, Dispatcher is still very useful. However, we will need to make it much better and soon.

File: docs/std/dispatcher.md (hunk 10 above)
Comment (armanbilge) [reply docs/std/dispatcher.md:68]: Woops, it seems it loses the ` ``` ` in a GH comments suggestion.

File: docs/std/dispatcher.md [hunk 11]
```diff
@@ -5,20 +5,71 @@ title: Dispatcher
...
+            override def onMessage(msg: String): Unit =
+              dispatcher.unsafeRunSync(queue.offer(msg))
+          }
+        }
+      _ <- IO.delay(impureInterface.init())
+      value <- queue.tryTake
+    } yield value match {
+      case Some(v) => println(s"Value found in queue! $v")
+      case None    => println("Value not found in queue :(")
+    }
```
armanbilge:
```suggestion
Dispatcher[IO].use { dispatcher =>
  for {
    queue <- Queue.unbounded[IO, String]
    impureInterface <-
      IO.delay {
        new ImpureInterface {
          override def onMessage(msg: String): Unit =
            dispatcher.unsafeRunSync(queue.offer(msg))
        }
      }
    _ <- IO.delay(impureInterface.init())
    value <- queue.tryTake
  } yield value match {
    case Some(v) => println(s"Value found in queue! $v")
    case None    => println("Value not found in queue :(")
  }
}
```

vasilmkd: Is this safe to commit?

armanbilge: Should be I think, adds a closing brace and fixes the indent.

docs/std/dispatcher.md
```diff
@@ -42,7 +93,15 @@ trait Dispatcher[F[_]] extends DispatcherPlatform[F] {
 }
 ```
 
-An instance of `Dispatcher` is very cheap - it allocates a single fiber so it
-is encouraged to instantiate it where necessary,
-as close to the unsafe region as possible,
-rather than wiring a single instance throughout an application.
+An instance of `Dispatcher` is very cheap - it is encouraged to instantiate it 
+where necessary rather than wiring a single instance throughout an application.
```
armanbilge:
```suggestion
Creating an instance of `Dispatcher` is very cheap - you are encouraged to instantiate it 
where necessary rather than wiring a single instance throughout an application.
```

docs/std/dispatcher.md
```diff
@@ -42,7 +93,15 @@ trait Dispatcher[F[_]] extends DispatcherPlatform[F] {
...
+
+
+## Cats-Effect 2
+
+Users of cats effect 2 may be familiar with the `Effect` and `ConcurrentEffect`
```
armanbilge: Sorry for all the tiny nits!

```suggestion
## Cats Effect 2

Users of Cats Effect 2 may be familiar with the `Effect` and `ConcurrentEffect`
```

docs/std/dispatcher.md
```diff
@@ -42,7 +93,15 @@ trait Dispatcher[F[_]] extends DispatcherPlatform[F] {
...
+typeclasses. These have been removed as they contrained implementations of the
```
armanbilge:
```suggestion
typeclasses. These have been removed as they constrained implementations of the
```

docs/std/dispatcher.md
vasilmkd:
```suggestion
`Dispatcher` is a fiber-based utility for evaluating effects across an impure boundary. An instance of `Dispatcher` can be derived for any effect type conforming to the `Async` typeclass.
```

armanbilge: FWIW, using it for _any_ Async can land you into all sorts of trouble 😛 e.g. see https://github.com/typelevel/cats-effect/issues/2448#issuecomment-959181978.

armanbilge: Maybe we should just fess up a bit in the docs and add a caveats section 😅

vasilmkd: Sure, I don't mind.

LLCampos:
> Yeah, the last thing here to do would be to amend the `Supervisor` mention, because `Dispatcher` really isn't a valid instance of a `Supervisor`. When we change that, we'll update the docs along with it.

`Supervisor` mention amended ;)

vasilmkd (approved):

armanbilge (approved):
This was long overdue, thanks! Tracking corresponding changes to scaladocs in https://github.com/typelevel/cats-effect/issues/2519.

PR #2516, opened 2021-11-07
Snapshots of queues

vasilmkd:


core/jvm/src/main/scala/cats/effect/unsafe/LocalQueue.scala
```diff
@@ -858,6 +858,18 @@ private final class LocalQueue {
...
 
+  /**
+   * Returns a snapshot of the fibers currently enqueued on this local queue.
+   *
+   * @return
+   *   a set of the currently enqueued fibers
+   */
+  def snapshot(): Set[IOFiber[_]] = {
+    // load fence to get a more recent snapshot of the enqueued fibers
+    val _ = size()
```
armanbilge, line 869: Hotspot trivia: do we have any guarantee that it won't happily optimize this away? 😀

vasilmkd: https://github.com/typelevel/cats-effect/blob/b2276e5a96766fa9df9a86999cc1be246a43bd06/core/shared/src/main/scala/cats/effect/IOFiber.scala#L1217-L1221

We're relying on this and it works on ARM which is the much looser (w.r.t. memory ordering) architecture. If it doesn't work, we probably have bigger problems than printing some stale fibers out.

vasilmkd: I'm not an expert on this, but I think it would be a violation of the Java Memory Model, since these are atomic variables and can and should be used for synchronization. If it ever doesn't work, we'll switch to using `Unsafe`.

armanbilge: Interesting stuff 🤓

armanbilge (approved):

PR #2517, opened 2021-11-07
A snapshot of live fibers on the work stealing thread pool

vasilmkd:


core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -445,6 +445,28 @@ private[effect] final class WorkStealingThreadPool(
     ()
   }
 
+  /**
+   * Returns a snapshot of the fibers currently live on this thread pool.
+   *
+   * @return
+   *   a set of currently live fibers and a mapping of active fibers associated to their carrier
+   *   threads
```
armanbilge: Checking my intuition: if a fiber is in the external queue, it appears in the live set but not in the map keys?

vasilmkd: Yeah. Is the wording weird? Feel free to suggest something else.

armanbilge: Just to throw another idea out there: what if we returned a `Set` representing the external queue, and a `Thread -> Set` map for the rest? At least for me it better matches conceptually the underlying architecture of the pool

vasilmkd: That's a great idea. Let's do it.

armanbilge: When we print it out, we could even organize by thread: this thread is running this fiber, and has queued these, etc.

vasilmkd: Yeah, totally.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -445,6 +446,37 @@ private[effect] final class WorkStealingThreadPool(
     ()
   }
 
+  /**
+   * Returns a snapshot of the fibers currently live on this thread pool.
+   *
+   * @return
+   *   a pair consisting of the set of fibers on the external thread and a map associating
```
armanbilge:
```suggestion
   *   a pair consisting of the set of fibers on the external queue and a map associating
```

armanbilge (dismissed):
👍 assuming it compiles this time 😉

armanbilge (approved):

PR #2518, opened 2021-11-07
Suspended fibers snapshot

vasilmkd:


core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -450,19 +450,20 @@ private[effect] final class WorkStealingThreadPool(
    * Returns a snapshot of the fibers currently live on this thread pool.
    *
    * @return
-   *   a pair consisting of the set of fibers on the external queue and a map associating worker
+   *   a 3-tuple consisting of the set of fibers on the external queue, a map associating worker
    *   threads to the currently active fiber and fibers enqueued on the local queue of that
-   *   worker thread
+   *   worker thread and a set of suspended fibers tracked by this thread pool
```
armanbilge, line 455: maybe bullet-point/enumerate this? 😆 getting a bit unwieldy

core/jvm/src/main/scala/cats/effect/unsafe/FiberMonitor.scala
```diff
@@ -101,4 +103,29 @@ private[effect] object FiberMonitor {
...
+          if (v ne null) {
+            buffer += v
+          }
+        }
+
+        buffer.toSet
+      } catch {
+        case _: ConcurrentModificationException =>
+          buffer.clear()
+          contents()
```
armanbilge: I've been wondering about this. How do we know this doesn't become an infinite loop?

vasilmkd: If it does for anyone, we'll fix it then.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
vasilmkd: It's an internal method, this is too much as it is.

core/jvm/src/main/scala/cats/effect/unsafe/FiberMonitor.scala
armanbilge: Assuming that people only use this on hung programs, it's almost definitely okay. If someone tries to fiber dump a live program, this could end up indefinitely blocking a thread. Yolo?

armanbilge (approved):
Thanks!

PR #2525, opened 2021-11-09
Order snapshot data by trustworthiness

vasilmkd:


core/jvm/src/main/scala/cats/effect/unsafe/FiberMonitor.scala
```diff
@@ -97,8 +97,30 @@ private[effect] final class FiberMonitor(
...
-      val foreign = foreignFibers()
+      val (rawExternal, workersMap, rawSuspended) = compute.liveFibers()
+      val rawForeign = foreignFibers()
+
+      // We trust the sources of data in the following order, ordered from
+      // most trustworthy to least trustworthy.
+      // 1. Fibers from the worker threads
+      // 2. Fibers from the external queue
+      // 3. Fibers from the foreign synchronized fallback weak GC maps
+      // 4. Fibers from the suspended thread local GC maps
```
armanbilge, line 108: Fascinating! How did you determine this? 😁

core/jvm/src/main/scala/cats/effect/unsafe/FiberMonitor.scala
```diff
@@ -97,8 +97,30 @@ private[effect] final class FiberMonitor(
...
+
+      val external = workersMap.foldLeft(rawExternal) {
+        case (acc, (_, (active, local))) =>
+          (acc -- local) - active
```
armanbilge: Instead of folding, could we just toss all of these into a single set to use in a single set difference? It took me a little while to understand what was going on here.

core/jvm/src/main/scala/cats/effect/unsafe/FiberMonitor.scala
vasilmkd: It's more or less ordered by how much synchronization each data source entails.
1. Synchronized and "small", the snapshot should be consistent
2. The external queue is a bunch of `ConcurrentLinkedQueues` which are all synchronized on each linked element
3. The GC maps that are synchronized (may be out of date)
4. The GC maps that are thread local and not synchronized (may be out of date)

It's far from scientific. I'm not claiming this is great. But I do personally like the by-worker-thread approach to the output. I personally don't see a problem with tweaking the output in subsequent releases if something doesn't pan out.

armanbilge: Thanks, that seems plenty decent reasoning for me :) I like it!

core/jvm/src/main/scala/cats/effect/unsafe/FiberMonitor.scala
vasilmkd: Can you please post a short code snippet? I don't quite see how to do that.

armanbilge: I was thinking something like:
```scala
val toRemove = workersMap.map {
  case (_, (active, local)) => local + active
}.reduce(_ ++ _)
val external = rawExternal -- toRemove
```

Apologies if I'm being stupid/dense and not getting this 😆

vasilmkd: It's actually a great simplification. Thanks.

armanbilge (approved):

vasilmkd:
Let's hear from @djspiewak on this one.

PR #2528, opened 2021-11-10
Use the tracing constants for speed

vasilmkd:
Every little bit helps.

`series/3.x`:
```
DeepBindBenchmark.async   10000  thrpt   20  1160.801 ± 18.224  ops/s
```

`This PR`:
```
DeepBindBenchmark.async   10000  thrpt   20  1216.760 ± 17.077  ops/s
```


armanbilge (approved):
Nice! When can I unleash you onto some of my projects? 😉

PR #2534, opened 2021-11-12
Compute pool metrics

vasilmkd:
I've been putting off doing this for so long.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
@@ -566,7 +566,7 @@ private[effect] final class WorkStealingThreadPool(
     }
   }
 
-  def blockedWorkerThreadCounterForwarder: AtomicInteger =
+  private[unsafe] def blockedWorkerThreadCounterForwarder: AtomicInteger =
     blockedWorkerThreadCounter
 
   /*
@@ -584,7 +584,7 @@ private[effect] final class WorkStealingThreadPool(
    * @return
    *   the number of worker threads backing the compute pool
    */
-  def getWorkerThreadCount(): Int =
+  private[unsafe] def getWorkerThreadCount(): Int =
     threadCount
 
   /**
@@ -594,7 +594,7 @@ private[effect] final class WorkStealingThreadPool(
    * @return
    *   the number of active worker threads
    */
-  def getActiveThreadCount(): Int = {
+  private[unsafe] def getActiveThreadCount(): Int = {
     val st = state.get()
     (st & UnparkMask) >>> UnparkShift
   }
@@ -606,7 +606,7 @@ private[effect] final class WorkStealingThreadPool(
    * @return
    *   the number of worker threads searching for work
    */
-  def getSearchingThreadCount(): Int = {
+  private[unsafe] def getSearchingThreadCount(): Int = {
     val st = state.get()
     st & SearchMask
   }
@@ -618,7 +618,7 @@ private[effect] final class WorkStealingThreadPool(
    * @return
    *   the number of blocked worker threads
    */
-  def getBlockedWorkerThreadCount(): Int =
+  private[unsafe] def getBlockedWorkerThreadCount(): Int =
     blockedWorkerThreadCounter.get()
 
   /**
@@ -627,7 +627,7 @@ private[effect] final class WorkStealingThreadPool(
    * @return
    *   the total number of fibers enqueued on all local queues
    */
-  def getLocalQueueFiberCount(): Long =
+  private[unsafe] def getLocalQueueFiberCount(): Long =
     localQueues.map(_.size().toLong).sum
 
   /**
@@ -640,6 +640,6 @@ private[effect] final class WorkStealingThreadPool(
    * @return
    *   the number of asynchronously suspended fibers
    */
-  def getSuspendedFiberCount(): Long =
+  private[unsafe] def getSuspendedFiberCount(): Long =
     workerThreads.map(_.getSuspendedFiberCount().toLong).sum
 }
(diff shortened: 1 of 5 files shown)


core/jvm/src/main/scala/cats/effect/unsafe/metrics/ComputePoolSampler.scala
```diff
@@ -0,0 +1,35 @@
...
+package metrics
+
+/**
+ * An implementation of the [[ComputePoolSamplerMBean]] interface which simply delegates to the
+ * corresponding methods of the [[cats.effect.unsafe.WorkStealingThreadPool]] being monitored.
+ *
+ * @param queue
+ *   the monitored local queue
+ */
+private[unsafe] class ComputePoolSampler(compute: WorkStealingThreadPool)
```
armanbilge: Can this be final?

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -563,4 +565,81 @@ private[effect] final class WorkStealingThreadPool(
       Thread.currentThread().interrupt()
     }
   }
+
+  private[unsafe] def blockedWorkerThreadCounterForwarder: AtomicInteger =
+    blockedWorkerThreadCounter
```
armanbilge, line 570: Pedagogical: why keep this behind a forwarder?

armanbilge (dismissed):
Don't have experience with MBeans but this seems straightforward enough 👍

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
vasilmkd: Because of this BS:

```scala
class Stupid {
  val blah = 5
}
```

compiles to the following bytecode:

```
public class cats.effect.Stupid {
  private final int blah;

  private volatile boolean bitmap$init$0;

  public int blah();
    Code:
       0: aload_0
       1: getfield      #15                 // Field bitmap$init$0:Z
       4: ifeq          14
       7: aload_0
       8: getfield      #17                 // Field blah:I
      11: goto          24
      14: new           #19                 // class scala/UninitializedFieldError
      17: dup
      18: ldc           #21                 // String Uninitialized field: /Users/vasil/Code/cats-effect/core/shared/src/main/scala/cats/effect/Stupid.scala: 20
      20: invokespecial #25                 // Method scala/UninitializedFieldError."<init>":(Ljava/lang/String;)V
      23: athrow
      24: pop
      25: aload_0
      26: getfield      #17                 // Field blah:I
      29: ireturn

  public cats.effect.Stupid();
    Code:
       0: aload_0
       1: invokespecial #30                 // Method java/lang/Object."<init>":()V
       4: aload_0
       5: iconst_5
       6: putfield      #17                 // Field blah:I
       9: aload_0
      10: iconst_1
      11: putfield      #15                 // Field bitmap$init$0:Z
      14: return
}
```

vasilmkd: The `blah()` getter is synchronized on the `bitmap` field, which is volatile, incurring unnecessary synchronization points.

core/jvm/src/main/scala/cats/effect/unsafe/metrics/ComputePoolSampler.scala
vasilmkd: Good catch. It was an oversight. Thanks.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
armanbilge: Huh, I thought this happened even for `private[this]` fields which is why you moved everything to the constructor?

armanbilge (approved):

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
vasilmkd: It does in the constructor. But for `private[this]` fields, the getter is not automatically generated by the compiler. A "forwarder" method accesses the field raw.

vasilmkd:
```
class Forwarder {
  private[this] val blah: Int = 5

  def blahForwarder: Int = blah
}
```

compiles to

```
public class cats.effect.Forwarder {
  private final int blah;

  private volatile boolean bitmap$init$0;

  public int blahForwarder();
    Code:
       0: aload_0
       1: getfield      #16                 // Field blah:I
       4: ireturn

  public cats.effect.Forwarder();
    Code:
       0: aload_0
       1: invokespecial #22                 // Method java/lang/Object."<init>":()V
       4: aload_0
       5: iconst_5
       6: putfield      #16                 // Field blah:I
       9: aload_0
      10: iconst_1
      11: putfield      #24                 // Field bitmap$init$0:Z
      14: return
}
```

vasilmkd: Forwarder is a Cats Effect nomenclature. I think I heard Daniel say it once, or use it, and I continued doing so. It doesn't have any meaning. It's the same as getter, the point is to write it ourselves and circumvent the synchronization that the Scala compiler would do otherwise. 😄

armanbilge: Thanks 🤓 I'm very glad there's a workaround! But I wish I understood _why_ the Scala compiler does this.

PR #2473, opened 2021-10-27
Fit a fiber on a single cache line

vasilmkd:
This is not production ready. And is also probably not the best combination of fields to achieve 64 bytes of `IOFiber` object footprint.

1. I was examining bytecode and came across the following:
`private[this] val` fields which are _inside_ the class body generate a `bitmap$init` volatile field. I don't exactly know why. However, it can be circumvented (on all Scala versions) by putting the same `private[this] val` fields inside the primary constructor for the class. Thus, the `series/3.x` bytecode looks like the following:
```
private volatile int bitmap$init$0;

public cats.effect.IOFiber(scala.collection.immutable.Map<cats.effect.IOLocal<?>, java.lang.Object>, scala.Function1<cats.effect.kernel.Outcome<cats.effect.IO, java.lang.Throwable, A>, scala.runtime.BoxedUnit>, cats.effect.IO<A>, scala.concurrent.ExecutionContext, cats.effect.unsafe.IORuntime);
Code:
    0: aload_0
    1: aload         5
    3: putfield      #255                // Field runtime:Lcats/effect/unsafe/IORuntime;
    6: aload_0
    7: invokespecial #1193               // Method cats/effect/IOFiberPlatform."<init>":()V
    10: aload_0
    11: invokestatic  #1197               // InterfaceMethod cats/effect/kernel/Fiber.$init$:(Lcats/effect/kernel/Fiber;)V
    14: aload_0
    15: new           #401                // class cats/effect/ArrayStack
    18: dup
    19: invokespecial #1198               // Method cats/effect/ArrayStack."<init>":()V
    22: putfield      #399                // Field objectState:Lcats/effect/ArrayStack;
    25: aload_0
    26: aload_0
    27: getfield      #223                // Field bitmap$init$0:I
    30: iconst_1
    31: ior
    32: putfield      #223                // Field bitmap$init$0:I
    35: aload_0
    36: aload         4
    38: putfield      #265                // Field currentCtx:Lscala/concurrent/ExecutionContext;
    41: aload_0
    42: aload_0
    43: getfield      #223                // Field bitmap$init$0:I
    46: iconst_2
    47: ior
    48: putfield      #223                // Field bitmap$init$0:I
    51: aload_0
    52: iconst_0
    53: putfield      #465                // Field canceled:Z
    56: aload_0
    57: aload_0
    58: getfield      #223                // Field bitmap$init$0:I
    61: iconst_4
    62: ior
    63: putfield      #223                // Field bitmap$init$0:I
[...]

PR #2540, opened 2021-11-13
Remove `-Xcheckinit` compiler flag

armanbilge:
I'm pretty sure this is responsible for those auto-generated volatile fields that have been taunting @vasilmkd 😆

### DIFF
--- a/project/Common.scala
+++ b/project/Common.scala
@@ -30,12 +30,6 @@ object Common extends AutoPlugin {
           Nil
         else
           Seq(compilerPlugin("com.olegpy" %% "better-monadic-for" % "0.3.1"))
-      },
-      scalacOptions ++= {
-        if (isDotty.value)
-          Nil
-        else
-          Seq("-Xcheckinit")
       }
     )
 }


vasilmkd:
I'm in favor of removing this, because it causes a volatile field `bitmap$init` to be created on a lot of classes. This slows down constructors and field accesses (which we work hard to avoid), but the constructors remain. Examples:
https://github.com/typelevel/cats-effect/pull/2534#discussion_r748775008 and https://github.com/typelevel/cats-effect/pull/2473.

armanbilge:
I don't know how hard it would be to setup, but if this check is important, maybe we can enable this flag only for testing/CI but remove for release?

vasilmkd:
No, I don't agree with that. That could hide bugs. Either there is a good reason that this flag was included and @djspiewak can shed some light on it, or I really don't know why this is here and we can remove it everywhere.

armanbilge:
Btw it was introduced here:
https://github.com/typelevel/cats-effect/pull/874/commits/de1f68bb4fff5f4fbbf3432069258e96609d02b5#diff-928a6eabff1094f1b64a2d1b6bc7574c5a483476ec3fee98c16d1fe46aae9d0d

vasilmkd:
Before:
```
[info] Benchmark             (size)   Mode  Cnt      Score     Error  Units
[info] AsyncBenchmark.start     100  thrpt    5  12890.043 ± 107.889  ops/s
```

After:
```
[info] Benchmark             (size)   Mode  Cnt      Score     Error  Units
[info] AsyncBenchmark.start     100  thrpt    5  15683.946 ± 296.763  ops/s
```

djspiewak (approved):
Sincerely no idea why I explicitly added this. So odd.

armanbilge:
@vasilmkd should we consider reverting some of the changes you made to workaround this? E.g. moving things into constructors. Or maybe those were good ideas anyway 👍

PR #2541, opened 2021-11-14
Undo workarounds for `bitmap$init`

armanbilge:
Don't feel strongly about this, but I think the old code was just a little cleaner and better scoped. I think these were the only two places?

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
@@ -102,7 +102,7 @@ private[effect] final class WorkStealingThreadPool(
    */
   private[this] val done: AtomicBoolean = new AtomicBoolean(false)
 
-  private[this] val blockedWorkerThreadCounter: AtomicInteger = new AtomicInteger(0)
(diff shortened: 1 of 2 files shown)


vasilmkd:
This is not a workaround, this is actual functionality. When blocking occurs, a new worker thread is spawned that replaces the blocked one. When the blocked one finishes, it dies. The fiber bag needs to be transferred from the old thread to the replacement thread, otherwise we lose references to fibers. Please remove that commit. You can restore the `blocking = false`.

armanbilge:
Woops! That makes more sense. Your commit messages were plenty clear enough on that fact to make me stare hard at it, but that doesn't get the brain working 😉

vasilmkd:
The `blockedWorkerThreadCounterForwarder` is ok to remove.

vasilmkd:
Can you please remove the commits and force push?

armanbilge:
This one strikes again for windows/graalvm:
```
[info] IOAppSpec
[info] IOApp (jvm) should
[error]   x evaluate and print hello world
[error]    1 != 0 (IOAppSpec.scala:39)
[error]   x pass all arguments to child
[error]    1 != 0 (IOAppSpec.scala:46)
```

vasilmkd (approved):
Nice cleanup.

PR #2543, opened 2021-11-14
Inline the head and the tail of the local queue while avoiding false sharing

vasilmkd:
False sharing is not a joke Jim, millions of fibers suffer every day. https://www.youtube.com/watch?v=WaaANll8h18

I ran the benchmarks twice each, and let the computer cool down between runs.

`series/3.x`:
```
WorkStealingBenchmark.scheduling  1000000  thrpt   20  11.439 ± 0.317  ops/min

WorkStealingBenchmark.scheduling  1000000  thrpt   20  11.475 ± 0.338  ops/min
```

`This PR`:
```
WorkStealingBenchmark.scheduling  1000000  thrpt   20  12.969 ± 0.350  ops/min

WorkStealingBenchmark.scheduling  1000000  thrpt   20  13.354 ± 0.277  ops/min
```

I still need to do some cleanups.

### DIFF
--- a/core/jvm/src/main/java/cats/effect/unsafe/LocalQueue.java
+++ b/core/jvm/src/main/java/cats/effect/unsafe/LocalQueue.java
@@ -19,30 +19,36 @@ package cats.effect.unsafe;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 
 @SuppressWarnings("unused")
-class HeadPadding {
-  private final int hp00 = 0;
-  private final long hp01 = 0;
-  private final long hp02 = 0;
-  private final long hp03 = 0;
-  private final long hp04 = 0;
-  private final long hp05 = 0;
-  private final long hp06 = 0;
-  private final long hp07 = 0;
-  private final long hp08 = 0;
-  private final long hp09 = 0;
-  private final long hp10 = 0;
-  private final long hp11 = 0;
-  private final long hp12 = 0;
-  private final long hp13 = 0;
-  private final long hp14 = 0;
-  private final long hp15 = 0;
-  private final long hp16 = 0;
+class ClassPadding {
+  private final int p00 = 0;
+}
+
+@SuppressWarnings("unused")
+class HeadPadding extends ClassPadding {
+  private final long p01 = 0;
+  private final long p02 = 0;
+  private final long p03 = 0;
+  private final long p04 = 0;
+  private final long p05 = 0;
+  private final long p06 = 0;
+  private final long p07 = 0;
+  private final long p08 = 0;
+  private final long p09 = 0;
+  private final long p10 = 0;
+  private final long p11 = 0;
+  private final long p12 = 0;
+  private final long p13 = 0;
+  private final long p14 = 0;
+  private final long p15 = 0;
+  private final long p16 = 0;
 }
 
 @SuppressWarnings("unused")
 class Head extends HeadPadding {
   protected static final AtomicIntegerFieldUpdater<Head> updater =
       AtomicIntegerFieldUpdater.newUpdater(Head.class, "head");
+
+  private final int p00 = 0;
 
   /**
    * The head of the queue.
@@ -73,23 +79,22 @@ class Head extends HeadPadding {
 
 @SuppressWarnings("unused")
 class TailPadding extends Head {
-  private final int tp00 = 0;
-  private final long tp01 = 0;
-  private final long tp02 = 0;
-  private final long tp03 = 0;
-  private final long tp04 = 0;
-  private final long tp05 = 0;
-  private final long tp06 = 0;
-  private final long tp07 = 0;
-  private final long tp08 = 0;
-  private final long tp09 = 0;
-  private final long tp10 = 0;
-  private final long tp11 = 0;
-  private final long tp12 = 0;
-  private final long tp13 = 0;
-  private final long tp14 = 0;
-  private final long tp15 = 0;
-  private final long tp16 = 0;
+  private final long p01 = 0;
+  private final long p02 = 0;
+  private final long p03 = 0;
+  private final long p04 = 0;
+  private final long p05 = 0;
+  private final long p06 = 0;
+  private final long p07 = 0;
+  private final long p08 = 0;
+  private final long p09 = 0;
+  private final long p10 = 0;
+  private final long p11 = 0;
+  private final long p12 = 0;
+  private final long p13 = 0;
+  private final long p14 = 0;
+  private final long p15 = 0;
+  private final long p16 = 0;
 }
 
 @SuppressWarnings("unused")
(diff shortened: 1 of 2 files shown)


vasilmkd:
Big boy `LocalQueue`:

```
cats.effect.unsafe.LocalQueue object internals:
OFF  SZ                    TYPE DESCRIPTION                              VALUE
  0   8                         (object header: mark)                    N/A
  8   4                         (object header: class)                   N/A
 12   4                     int HeadPadding.hp00                         N/A
 16   8                    long HeadPadding.hp16                         N/A
 24   8                    long HeadPadding.hp01                         N/A
 32   8                    long HeadPadding.hp02                         N/A
 40   8                    long HeadPadding.hp03                         N/A
 48   8                    long HeadPadding.hp04                         N/A
 56   8                    long HeadPadding.hp05                         N/A
 64   8                    long HeadPadding.hp06                         N/A
 72   8                    long HeadPadding.hp07                         N/A
 80   8                    long HeadPadding.hp08                         N/A
 88   8                    long HeadPadding.hp09                         N/A
 96   8                    long HeadPadding.hp10                         N/A
104   8                    long HeadPadding.hp11                         N/A
112   8                    long HeadPadding.hp12                         N/A
120   8                    long HeadPadding.hp13                         N/A
128   8                    long HeadPadding.hp14                         N/A
136   8                    long HeadPadding.hp15                         N/A
144   4                     int Head.head                                N/A
148   4                     int TailPadding.tp00                         N/A
152   8                    long TailPadding.tp16                         N/A
160   8                    long TailPadding.tp01                         N/A
168   8                    long TailPadding.tp02                         N/A
176   8                    long TailPadding.tp03                         N/A
184   8                    long TailPadding.tp04                         N/A
192   8                    long TailPadding.tp05                         N/A
200   8                    long TailPadding.tp06                         N/A
208   8                    long TailPadding.tp07                         N/A
216   8                    long TailPadding.tp08                         N/A
224   8                    long TailPadding.tp09                         N/A
232   8                    long TailPadding.tp10                         N/A
240   8                    long TailPadding.tp11                         N/A
248   8                    long TailPadding.tp12                         N/A
256   8                    long TailPadding.tp13                         N/A
264   8                    long TailPadding.tp14                         N/A
272   8                    long TailPadding.tp15                         N/A
280   4                     int Tail.tail                                N/A
284   4                     int Tail.tailPublisher                       N/A
288   8                    long LocalQueuePadding.qp16                   N/A
296   8                    long LocalQueuePadding.qp01                   N/A
304   8                    long LocalQueuePadding.qp02                   N/A
312   8                    long LocalQueuePadding.qp03                   N/A
320   8                    long LocalQueuePadding.qp04                   N/A
328   8                    long LocalQueuePadding.qp05                   N/A
336   8                    long LocalQueuePadding.qp06                   N/A
344   8                    long LocalQueuePadding.qp07                   N/A
352   8                    long LocalQueuePadding.qp08                   N/A
360   8                    long LocalQueuePadding.qp09                   N/A
368   8                    long LocalQueuePadding.qp10                   N/A
376   8                    long LocalQueuePadding.qp11                   N/A
384   8                    long LocalQueuePadding.qp12                   N/A
392   8                    long LocalQueuePadding.qp13                   N/A
400   8                    long LocalQueuePadding.qp14                   N/A
408   8                    long LocalQueuePadding.qp15                   N/A
416   4                     int LocalQueuePadding.qp00                   N/A
420   4   cats.effect.IOFiber[] LocalQueue.buffer                        N/A
424   8                    long LocalQueue.totalFiberCount               N/A
432   8                    long LocalQueue.totalSpilloverCount           N/A
440   8                    long LocalQueue.successfulStealAttemptCount   N/A
448   8                    long LocalQueue.stolenFiberCount              N/A
Instance size: 456 bytes
Space losses: 0 bytes internal + 0 bytes external = 0 bytes total
```

djspiewak:
Daaaaaamn. Okay I'm going to need a primer on what's going on here. :-P

vasilmkd:
@djspiewak

Here's an X-ray scan from before:

```
cats.effect.unsafe.LocalQueue object internals:
OFF  SZ                                        TYPE DESCRIPTION                              VALUE
  0   8                                             (object header: mark)                    N/A
  8   4                                             (object header: class)                   N/A
 12   4                                         int LocalQueue.tail                          N/A
 16   8                                        long LocalQueue.totalFiberCount               N/A
 24   8                                        long LocalQueue.totalSpilloverCount           N/A
 32   8                                        long LocalQueue.successfulStealAttemptCount   N/A
 40   8                                        long LocalQueue.stolenFiberCount              N/A
 48   1                                        byte LocalQueue.bitmap$init$0                 N/A
 49   3                                             (alignment/padding gap)                  
 52   4                       cats.effect.IOFiber[] LocalQueue.buffer                        N/A
 56   4   java.util.concurrent.atomic.AtomicInteger LocalQueue.head                          N/A
 60   4   java.util.concurrent.atomic.AtomicInteger LocalQueue.tailPublisher                 N/A
Instance size: 64 bytes
Space losses: 3 bytes internal + 0 bytes external = 3 bytes total
```

The `AtomicInteger head` and `AtomicInteger tailPublisher` are just references, the field that the actual `CAS` contention is happening is somewhere at a different location in memory. A pointer has to be followed every time these are used.

This PR inlines the volatile fields `head` and `tailPublisher` inside the `LocalQueue` itself. It uses the `AtomicIntegerFieldUpdater` mechanism to allow CAS on normal volatile fields (instead of going through `AtomicInteger`).

However, we cannot just take these fields and plop them into the class. That would create false sharing, i.e. two independent fields residing on the same cache line while different threads independently write to each variable. Each write to either field in this case invalidates the cache of the other thread, slowing everyone down.

The way to combat this is to isolate both fields in separate cache lines. It is recommended to insert padding of 128 bytes between the fields to make sure that they end up on different cache lines. (Most machines will be fine with 64 bytes padding, but our software might be running on big servers and there are only a few local queues per JVM, so it's not like we're in a situation where we have to be thrifty about this stuff and so, why take the chance).

However, we cannot just insert fields in a class from top to bottom and expect the JVM to order the fields in that order in memory. The JVM is mostly free to do whatever it wants in the name of optimization. However, it is obliged to follow one rule. Namely, the fields of a superclass always come before the fields of a subclass. This PR uses this rule to create a hierarchy of superclasses for `LocalQueue` which define:
1. 128 bytes of padding before the head field
2. The head field
3. 128 bytes of padding before the tail field
4. The tail field
5. 128 bytes of padding before the rest of the `LocalQueue` class
6. And finally the `LocalQueue` class completes this chain

Now, in whatever way the JVM orders these fields, there will always be 128 bytes of padding both before and after the head and tail fields respectively, and thus they are guaranteed to fall on a cache line which will not be shared with the other variable.

Overall, this change improves the spatial locality of `LocalQueue` (data necessary for `enqueue/dequeue` is right next to the `head/tail` in memory, instead of dereferencing an `AtomicInteger` pointer and might lead to more optimized fetching and cache transfer).

And there's no denying that we're seeing improvements.

djspiewak:
This is unbelievably cool

armanbilge:
Brilliant 😮

vasilmkd:
This PR is ready.

djspiewak (dismissed):

vasilmkd:
Rebased and fixed merge conflicts.

djspiewak (approved):

PR #2548, opened 2021-11-15
Reflectively register signal handling if available

vasilmkd:
Danger. Absolutely abhorrent code ahead. But at least we don't link against `sun.misc` code.


vasilmkd:
A good test for this is to try and build a native image. I will do it later tonight.

vasilmkd:
Without any configuration, building a GraalVM native image results in the following:
```
Exception in thread "main" com.oracle.svm.core.jdk.UnsupportedFeatureError: Proxy class defined by interfaces [interface sun.misc.SignalHandler] not found. Generating proxy classes at runtime is not supported. Proxy classes need to be defined at image build time by specifying the list of interfaces that they implement. To define proxy classes use -H:DynamicProxyConfigurationFiles=<comma-separated-config-files> and -H:DynamicProxyConfigurationResources=<comma-separated-config-resources> options.
        at com.oracle.svm.core.util.VMError.unsupportedFeature(VMError.java:87)
        at com.oracle.svm.reflect.proxy.DynamicProxySupport.getProxyClass(DynamicProxySupport.java:146)
        at java.lang.reflect.Proxy.getProxyConstructor(Proxy.java:66)
        at java.lang.reflect.Proxy.newProxyInstance(Proxy.java:1037)
        at cats.effect.Signal$2.handleSignal(Signal.java:115)
        at cats.effect.Signal.handleSignal(Signal.java:64)
        at cats.effect.IOApp.$anonfun$main$7(IOApp.scala:251)
        at cats.effect.IOApp.$anonfun$main$7$adapted(IOApp.scala:247)
        at scala.collection.immutable.List.foreach(List.scala:333)
        at cats.effect.IOApp.main(IOApp.scala:247)
        at cats.effect.IOApp.main$(IOApp.scala:200)
        at io.vasilev.Main$.main(Main.scala:7)
        at io.vasilev.Main.main(Main.scala)
```

After consulting https://www.graalvm.org/reference-manual/native-image/DynamicProxy/, the following configuration is necessary:

```json
// in file dynamic-proxy-config.json
[
    ["sun.misc.SignalHandler"]
]
```

and 

```
-H:DynamicProxyConfigurationFiles=dynamic-proxy-config.json
```

passed as parameter to the `native-image` generation process.

After using this config, native images also support signals. Unfortunately, there is no way to not provide the configuration and degrade to not using signals.

I don't mind adding a documentation section for this, as a solution. In any case, we should expect bug reports for this. But then again, if we link to `sun.misc` API directly, that's a whole other can of worms, so...

Edit: This is fixed now by adding this configuration to the jar.

vasilmkd:
Added the documentation page.

vasilmkd:
I should also try a JVM built using `jlink`.

vasilmkd:
`jlink` works and works great. No further configuration is necessary when building a JDK image. We only depend on `java.base` by default (with this PR). Running a Cats Effect on such an image degrades to not registering signals, which is what we wanted.

If users want signal support, they have to add the `jdk.unsupported` module to their `jlink` JDK. Then signals work as normal.

vasilmkd:
Ok, I'm actually done here now.

vasilmkd:
https://www.graalvm.org/reference-manual/native-image/BuildConfiguration/#runtime-vs-build-time-initialization

djspiewak (dismissed):
Very, very nice.

djspiewak (approved):

PR #2550, opened 2021-11-16
Print fiber dump incrementally rather than in one large string

djspiewak:


core/jvm/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -249,7 +249,7 @@ trait IOApp {
         try {
           Signal.handle(
             new Signal(name),
-            _ => runtime.fiberMonitor.liveFiberSnapshot().foreach(System.err.println(_)))
+            _ => runtime.fiberMonitor.liveFiberSnapshot(System.err.print(_)))
```
vasilmkd, line 252:
```suggestion
            _ => runtime.fiberMonitor.liveFiberSnapshot(System.err.print))
```

Nitpicking, feel free to ignore.

vasilmkd (approved):

core/jvm/src/main/scala/cats/effect/IOApp.scala
djspiewak: I think Scala 3 is going to eventually remove the eta-lifting magic in this expression, so that's part of why I've been trying to get in the habit of always adding the `_`.

Issue #2552, opened 2021-11-16
`raceOutcome` is not tested (directly)


armanbilge:
... and laws/tests :)

Slipped a fix in here, but probably deserves its own PR:
https://github.com/typelevel/cats-effect/pull/2545/commits/bdc7429e6b333500a07d357b7dbe783f34aeee26#diff-5acd9e5b3bc8897d205aa4590f18c307679d41279ca72a125764900fffe9d568

Discord discussion:
https://discord.com/channels/632277896739946517/839263556754472990/910242079341412433

Related:
https://github.com/typelevel/cats-effect/pull/1392

vasilmkd:
Removing this from the milestone because it is not a blocker and the fix is part of another PR.

armanbilge:
The fix went through in https://github.com/typelevel/cats-effect/pull/2545, but we are still lacking laws/tests.

djspiewak:
Let's leave this issue open to track the lack of tests

PR #2553, opened 2021-11-16
Fix raceOutcome, add laws

armanbilge:
Towards #2552, but currently in a bincompat mess.


djspiewak:
Yeah this is rough. :-) I think though that there isn't that much value in a law which is literally the definition. I mean, we probably have a couple of those, but I don't think it's all that meaningful. So if we skip the laws, the bincompat problems go out the window.

armanbilge:
Yeah, I was afraid of that 🙃 so there really are no better laws for this then? If laws can't help us, maybe there's a test or two we can add for this (to our internal tests module)?

PR #2554, opened 2021-11-17
typo fix in io.scala notes

amitbaer:


vasilmkd:
Thanks for opening this PR!

Unfortunately, I'm going to have to close this PR as "memoization" is correct in this context and refers to the following [concept](https://en.wikipedia.org/wiki/Memoization).

PR #2555, opened 2021-11-17
Remove `evalOnR` as it is a duplicate of `autoCedeR`

vasilmkd:


armanbilge (approved):
IIUC, doing an `evalOn` is the same as resuming after a `cede`?

vasilmkd:
This is the resumption method, which runs when `IOFiber#run` is called by a worker thread. This part is the same.

`evalOn` is preceded with a preparation step which changes the `executionCtx` variable to point to the new `EC` that we're switching to.

PR #2556, opened 2021-11-17
MBeans

vasilmkd:
Finally. Only 8 months late.


vasilmkd:
Screen Shot 2021-11-17 at 18 02 21

vasilmkd:
You have Cats Effect 3.3.0.

djspiewak (dismissed):
LOL. Nice work. :-)

core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
```diff
@@ -36,13 +36,20 @@ final class IORuntime private (
...
 
   private[effect] val fiberMonitor: FiberMonitor = FiberMonitor(compute)
 
+  private[this] val unregisterMBean = registerFiberMonitorMBean(fiberMonitor)
+
+  val shutdown: () => Unit = { () =>
+    unregisterMBean()
+    _shutdown()
+  }
+
```
vasilmkd: It's horrible that this code is in the constructor, let me see if I can do something about it.

vasilmkd:
Screen Shot 2021-11-17 at 18 45 03

Still here.

core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala

vasilmkd:
Ok, this is now not the same as before. If the registration code is in the constructor, every `IORuntime` constructed would register the MBean. Otherwise, only the `IOApp` and `global` ones do.

@djspiewak Thoughts?

I can easily revert the last commit.

djspiewak:
@vasilmkd IMO it's better to just do it in `IOApp` and `global`. Better not to mess with random third-party runtimes.

vasilmkd:
Ok, ready to go then.

Edit: See questions below.

vasilmkd:
Actually, another question. Should we do that for `WSTP`s as well?

vasilmkd:
At the moment, every `WSTP` constructed registers metrics. Should we also tie it to `IOApp` and `global` only?

djspiewak:
> At the moment, every WSTP constructed registers metrics. Should we also tie it to IOApp and global only?

I think tying it to `IOApp` and `global` is better in general, though maybe this is academic since you can't actually create a WSTP from third-party code.

armanbilge:
> though maybe this is academic since you can't actually create a WSTP from third-party code.

Ah, you can't. I was mildly surprised since for a moment there it seemed like you could :) I think we can merge this PR as is.

djspiewak (approved):

PR #2560, opened 2021-11-18
Thin ci a bit

armanbilge:


armanbilge:
Cool, I think this is down to 20 checks from 30 on main.

vasilmkd:
Is it better to state the sets of versions explicitly? Currently, the outcome is dependent on how they're ordered in the lists above. I know we control both, I'm just saying changing the former may have unintended consequences later in the file.

armanbilge:
Also is there any hope for this windows ci?

vasilmkd:
Most of the `IOApp` tests are disabled on Windows. I don't think it would be the end of the world if we disabled them all.

armanbilge:
Down to 19!

djspiewak:
We can probably cut this down even further by removing the JavaScript tests on Windows and Mac.

**Edit:** Nvm they already are derp

djspiewak (approved):

PR #2557, opened 2021-11-17
JS trace filtering fixes

armanbilge:
Some progress/workarounds for #2551. Still not quite right though I think.

PR #2575, opened 2021-11-23
Remove `-P:scalajs:mapSourceURI` for JS tests

armanbilge:
This is what tripped me up in https://github.com/typelevel/cats-effect/pull/2557#issuecomment-971989012.

On JVM, traces are filtered by package name, so we test outside the usual package.

On JS, traces are _also_ filtered by filename (e.g. starting with `https://raw.githubusercontent.com/typelevel/cats-effect/`). Removing this setting is equivalent to getting out of the package.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -479,6 +479,7 @@ lazy val tests = crossProject(JSPlatform, JVMPlatform)
     Test / fork := true,
     Test / javaOptions += s"-Dsbt.classpath=${(Test / fullClasspath).value.map(_.data.getAbsolutePath).mkString(File.pathSeparator)}")
   .jsSettings(
+    // The default configured mapSourceURI is used for trace filtering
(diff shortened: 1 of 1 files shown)


armanbilge:
Previous failure is a JVM flake, publication issue I assume or some other race condition.
```
[info] IOFiberSpec
[info] IOFiber should
[info]   + toString a running fiber
[error]   x toString a suspended fiber
[error]    'cats.effect.IOFiber@69fe2de8 RUNNING' doesn't match 'cats.effect.IOFiber@[0-9a-f][0-9a-f]+ SUSPENDED(: async_? @ fiber.IOFiberSpec.foreverNever\$[0-9]\(.*IOFiberSpec.scala:[0-9]{2}\))?' (IOFiberSpec.scala:51)
```

djspiewak (approved):

PR #2576, opened 2021-11-23
Add instructions about source maps

armanbilge:


docs/tracing.md
```diff
@@ -102,6 +102,15 @@ specify the following JVM system properties:
 ```
 
 For instructions how to configure these settings on JS see the [`IORuntime` configuration page](scaling-and-tuning/io-runtime-config.md).
+Additionally, we recommend enabling [source maps](https://nodejs.medium.com/source-maps-in-node-js-482872b56116) so that traces include line numbers to Scala code (and not line numbers to generated JS code, which are not so helpful!).
```
armanbilge: Any concerns this is not separated enough from JVM section and will be confusing?

djspiewak (approved):
Seems reasonable to me!

PR #2578, opened 2021-11-24
Do not update LocalQueue counters if tracing is disabled

vasilmkd:


armanbilge (commented):
Apologies, I don't think I've paid enough attention: how does stack tracking enabled/disabled relate to the JMX metrics? (Are these counters JMX metrics?)

vasilmkd:
These counters are the implementation details of the jmx metrics. Those interfaces show these numbers.

The MBeans are not registered if tracing is disabled, so there's no need to calculate them. I disabled them because there are some places where we have to publish them more strictly, so this saves a memory barrier here and there.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -537,9 +540,14 @@ private final class WorkerThread(
 
       // With another `WorkerThread` started, it is time to execute the blocking
       // action.
-      val result = thunk
-
-      pool.blockedWorkerThreadCounter.decrementAndGet()
+      val result =
+        try thunk // Exceptions are caught in IOFiber.
+        finally {
```
armanbilge, line 545: Why do we need finally if exceptions are caught?

armanbilge (approved):
Thanks for clarifying about that! :)

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
vasilmkd: If an exception occurs in the `thunk`, the counter would not be decremented. This decrements it while allowing the exception to bubble up to the proper handler.

armanbilge: Ah, the catch happens in the caller, not inside of the the `thunk`. Thanks.

PR #2581, opened 2021-11-24
Re-work JS trace filtering

armanbilge:
If this works, hopefully we don't ever have to think about this again.

PR #2582, opened 2021-11-24
Begone, source-map-support

armanbilge:
So in a surprising twist from https://github.com/scala-js/scala-js/issues/4599, it turns out we _don't_ need the npm package source-map-support (and all the `package-lock.json` trouble it brings!) if we disable the default source map configuration in Scala.js and replace it with the Node.js native source map feature.

Hopefully this will eventually be built-in to Scala.js, but for now I've also updated the tracing docs.


armanbilge:
Hmm, this PR might not stand alone. I'll merge it into https://github.com/typelevel/cats-effect/pull/2581.

armanbilge:
I don't think this works.

PR #2584, opened 2021-11-25
Tighten up specs2 dependencies

armanbilge:
Removes unnecessary ceremony around Scala 2/3 and explicitly brings in specs2-scalacheck so we don't rely on discipline-specs2 for transitive updates.


vasilmkd:
I'm not against this per se, but what this PR does is duplicate the whole build config of `discipline-specs2`, line by line, which seems wasteful to me. That repo is already set up for a quick release.

armanbilge:
That's fair, but I don't think I added any config that wasn't already here. Just reorganized to get a +22/-43 diff :) we could reduce even more if we moved the common settings to the auto-plugin.

IMHO we shouldn't need a release of discipline-specs2 to get the latest specs2 bug fixes.

djspiewak (approved):

PR #2587, opened 2021-11-27
Support (suspended) fiber dumps on foreign runtimes

armanbilge:
This enables fiber dumps even if the runtime isn't the WSTP. Since we are not tracking live/active fibers on this foreign runtime, we can only capture fibers that are "foreign" to the runtime i.e., suspended and `evalOn`.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/IOApp.scala
+++ b/core/jvm/src/main/scala/cats/effect/IOApp.scala
@@ -279,7 +279,8 @@ trait IOApp {
           latch.countDown()
         })(runtime)
 
-    runtime.fiberMonitor.monitorSuspended(fiber, fiber)
+    if (isStackTracing)
+      runtime.fiberMonitor.monitorSuspended(fiber, fiber)
 
     def handleShutdown(): Unit = {
       if (latch.getCount() > 0) {
--- a/core/jvm/src/main/scala/cats/effect/IOApp.scala
+++ b/core/jvm/src/main/scala/cats/effect/IOApp.scala
@@ -279,6 +279,8 @@ trait IOApp {
           latch.countDown()
         })(runtime)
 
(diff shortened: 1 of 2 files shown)


core/jvm/src/main/scala/cats/effect/unsafe/FiberMonitor.scala
```diff
@@ -96,53 +96,58 @@ private[effect] final class FiberMonitor(
...
-      val localAndActive = workersMap.foldLeft(Set.empty[IOFiber[_]]) {
-        case (acc, (_, (active, local))) =>
-          (acc ++ local) ++ active.toSet
-      }
-      val external = rawExternal -- localAndActive
-      val foreign = rawForeign -- localAndActive -- external
-      val suspended = rawSuspended -- localAndActive -- external
+    if (TracingConstants.isStackTracing)
+      Option(compute).fold {
+        printFibers(foreignFibers(), "ACTIVE")(print)
```
armanbilge, line 101: I'm not sure `ACTIVE` is what we want here. This is how foreign fibers are currently described in the WSTP's fiber dumps.

core/jvm/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -279,6 +279,8 @@ trait IOApp {
           latch.countDown()
         })(runtime)
 
+    runtime.fiberMonitor.monitorSuspended(fiber, fiber)
```
armanbilge: By explicitly monitoring the main fiber, we are guaranteed that it appears in the fiber dump even if a foreign runtime is being used. I think this makes good sense, WDYT? So the user isn't left wondering if their app is running 😉

djspiewak (approved):

PR #2593, opened 2021-11-28
Swap `System.exit` with `Runtime#halt`

### LINKED ISSUE #2586: `System.exit(1)` vs `Runtime#halt(1)` for fatal exceptions in `IOApp`
Continuing from:
- https://github.com/http4s/http4s/issues/5582#issuecomment-980693520.

alexandrustana:
This PR aims to fix #2586.

Even though these explanations
- https://github.com/http4s/http4s/issues/5582#issuecomment-978118101
- https://discordapp.com/channels/632277896739946517/632278585700384799/913131970702237696

mention that when encountering SOE it is safe to recover from it, they also state that it's not necessarily ideal to create a special case for handling it.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/internals/PoolUtils.scala
+++ b/core/jvm/src/main/scala/cats/effect/internals/PoolUtils.scala
@@ -57,7 +57,7 @@ private[internals] object PoolUtils {
             case t: Throwable =>
               // under most circumstances, this will work even with fatal errors
               t.printStackTrace()
-              System.exit(1)
+              Runtime.getRuntime().halt(1)
           }
       })
 


core/jvm/src/main/scala/cats/effect/internals/PoolUtils.scala
```diff
@@ -57,7 +57,7 @@ private[internals] object PoolUtils {
             case t: Throwable =>
               // under most circumstances, this will work even with fatal errors
               t.printStackTrace()
-              System.exit(1)
+              Runtime.getRuntime().halt(1)
```
armanbilge, line 60: Should we check if `t` is a fatal exception before deciding what to do?

alexandrustana: It is checked above that it is `NonFatal`, so I'm guessing we will only get fatal exceptions at this point.

armanbilge: Aha, so it is! Apologies for the tunnel vision 😅

alexandrustana: No worries, thank you for checking the PR 🙂

armanbilge (approved):

djspiewak (approved):

PR #2598, opened 2021-11-28
Updated scastie to use 3.3.0

djspiewak:


vasilmkd:
Should we use Scala `2.13.7` in Scastie?

vasilmkd (approved):

PR #2604, opened 2021-11-29
Fix `Console` on JS 

### LINKED ISSUE #2588: `Console` on Node.js is underpowered and probably blocks
This was first reported by @ChristopherDavenport in [Discord](https://discord.com/channels/632277896739946517/632278585700384799/900962949374550027).

1. `System.out.print` delegates to `console.log` which is standard JS but has various limitations. Node.js has a more featured `process.stdout`.
2. Furthermore, `process.stdout` is async so `console.log` on Node.js is almost definitely a blocking op.
3. I'm actually not sure what `readLine` does right now, but it definitely doesn't do what we want :)

This requires a source-breaking change to obtain the additional `Async` constraint. For browsers and bincompat we will also keep the old `Sync` based implementation.

I actually started a branch to fix this at the time but didn't follow-through 😅
https://github.com/armanbilge/cats-effect/tree/feature/nodejs-console

armanbilge:
Fixes #2588.
1. Deprecates `readLine` on JS. [It really never should have been there to begin with.](https://github.com/scala-js/scala-js/blob/97187da7cf83f98ee3389fd617f312c84944a6cf/javalanglib/src/main/scala/java/lang/System.scala#L42). Users on Node.js are encouraged to use `fs2.io.stdin` instead (see https://github.com/typelevel/fs2/pull/2728).
2. Implements an async `Console` implementation specifically for Node.js. I didn't bother with `readLine` because it's annoying here and literally a one line implementation in fs2.io (via lots of shared machinery).
3. Separates `Console#make` into a `Sync`-based JVM version and an `Async`-based JS version. This is source-breaking, hence I've bumped the base version to 3.4.


armanbilge:
Darn, MiMa is unhappy because I moved `readLine` into a platform trait.
```
[error] cats-effect-std: Failed binary compatibility check against org.typelevel:cats-effect-std_2.13:3.2.8! Found 1 potential problems
[error]  * in current version, classes mixing cats.effect.std.Console need be recompiled to wire to the new static mixin forwarder method all super calls to method readLine()java.lang.Object
[error]    filter with: ProblemFilters.exclude[NewMixinForwarderProblem]("cats.effect.std.Console.readLine")
```

Options:
1. Don't deprecate `readLine` on JS.
2. Duplicate the entire `Console` trait in each platform.
3. Declare `Console` sealed and break bin-compat. Probably not, but is there a real motivation for someone to implement their own `Console`?

djspiewak:
How much work is it to get a functioning `readLine` on JS, now that we have the `Async` constraint? I get that it's duplicating stuff with fs2, but at the same time… we kind of promise this functionality in the API, and even on the JVM we go a really long way to try to make it functional.

Regarding people implementing their own `Console`… I'm sure it isn't common, but two cases I can think of would be tests and redirection (like if you want to redirect to a logger, or add some richer semantic to printing).

armanbilge:
> How much work is it to get a functioning readLine on JS, now that we have the Async constraint?

Well, we'd have to start by writing up a proposal to the ECMAScript folks :wink: but yeah, snark aside, we can't really keep this promise across all runtimes, which is unideal but not the end of the world. So we're really only talking about Node.js. By using fs2.io you are explicitly buying into the Node.js runtime.

So I did start work towards this in 6fc981b. The idea is to use the same approach as JVM, reading a byte at a time and attempting to decode. This strategy is _possible_, but is annoying and unperformant for a couple reasons:
1. Node.js `process.stdin` is not at all optimized for reading a byte at a time. The best you can do is read an entire byte buffer of length 1. If you read anymore than that, and then discard the unused bytes, then you mess up the next `readLine`. Unless we establish some sort of global state.
2. On JVM, we can just keep reading a byte at a time until there is no more. On JS, when there are no more bytes, it doesn't necessarily mean we are done, it means we need to asynchronously yield until more data can be read into the buffer. So we have the fun choice of whether we should duplicate the JVM code and adjust or find a way to tweak it so it can be shared.
3. I don't even want to think about how to test this 🙃 

So yes yes this is all possible, just a yak. 

In summary, I feel like a (somewhat) undeliverable promise was made, and unsurprisingly never delivered on. Seeing as we can never fully deliver on it, and even when we can, it's an unperformant maintenance burden, I just want to cut our losses 😆 

Edit: also worth noting that nobody complained about this.

armanbilge:
@djspiewak any thoughts about this one?

std/js/src/main/scala/cats/effect/std/Console.scala
```diff
@@ -86,14 +86,15 @@ object Console extends ConsoleCompanionCrossPlatform {
...
 
     private def write(writable: js.Dynamic, s: String): F[Unit] =
       F.async_[Unit] { cb =>
-        writable.write(
-          s,
-          (e: js.UndefOr[js.Error]) => cb(e.map(js.JavaScriptException(_)).toLeft(())))
+        if (writable.write(s).asInstanceOf[Boolean]) // no backpressure
+          cb(Right(()))
+        else // wait for drain event
+          writable.once("drain", () => cb(Right(())))
```
armanbilge, line 97: I'm very suspicious about this 🤔

djspiewak (approved):

PR #2613, opened 2021-11-30
Sync run convenience function

### LINKED ISSUE #2530: Bring back IO#runAsync and friends?
Nice to meet you!

I'm having a difficulty using Cats Effect 3 on Scala.js due to the change of the thread model.

Some JS functions such as [`Event.preventDefault()`](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault) have to be run synchronously or won't have any effect. In CE2, we can just use `IO` because it is run synchronously until a first async boundary. For example, if we wanted to implement some `click` event handler:

```scala
val more: IO[Unit] = ???
// OK in CE2, bad in CE3
def onClick(e: org.scalajs.dom.Event): IO[Unit] =
  IO{ e.preventDefault() } *> more
```

In CE3, we must use `SyncIO` without converting it into an `IO`. Unless I'm missing something, we have to resort to `IO#unsafeRunAndForget`:

```scala
val more: IO[Unit] = ???
// bad
def onClick(e: org.scalajs.dom.Event): SyncIO[Unit] =
  SyncIO{ e.preventDefault() }.to[IO] *> more
// OK
def onClick(e: org.scalajs.dom.Event): SyncIO[Unit] =
  SyncIO{ e.preventDefault(); more.unsafeRunAndForget() }
// OK
def onClick(e: org.scalajs.dom.Event): SyncIO[Unit] =
  SyncIO{ e.preventDefault() } *> SyncIO{ more.unsafeRunAndForget() }
```

The story is complicated because often `more` in the above example must be asynchronous. For example, it can be an AJAX call triggered by clicking a link. I guess this is going to be a common pattern in my project, and I definitely wouldn't like to sprinkle `unsafeRunAndForget` all over it.

If there were `IO#runAsync` like in CE2, the last code could be rewritten like this:

```scala
def onClick(e: org.scalajs.dom.Event): SyncIO[Unit] =
  SyncIO{ e.preventDefault() } *> more.runAsync(_ => IO.unit)
```

(I'd like to point out `IO#runAsync` would be equivalent to [`runAff_`](https://pursuit.purescript.org/packages/purescript-aff/6.0.0/docs/Effect.Aff#v:runAff_) in `purescript-aff`, and `IO#runAndForget` would correspond to `launchAff_`.)

The migration guide refers to `Dispatcher` as the replacement of `IO#runAsync` but it doesn't have any methods that return `SyncIO[_]`. I think it is reasonable because `Dispatcher` shouldn't be coupled to `SyncIO`, but it may be helpful to have `IO`, as a concrete effect type, have a conversion from `IO` to `SyncIO`.

djspiewak:
Fixes #2530 
Supersedes #2478

Adds `unsafeRunSyncToFuture` to `IO` on Scala.js only. This uses a new overload of `syncStep` which takes an integer number of steps to allow prior to forcing a yield, defaulting that value to the `autoYieldThreshold` from the runtime itself. We could also mark the old `syncStep` as `private[effect]` if we want to clean things up a bit.

This is a good example of a change which is not source-breaking but which *is* forward-incompatible. I bumped the version to 3.4 to account for this. Please discuss and bikeshed.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -26,7 +26,7 @@ import org.scalajs.jsenv.selenium.SeleniumJSEnv
 
 import JSEnv._
 
-ThisBuild / baseVersion := "3.3"
+ThisBuild / baseVersion := "3.4"
 
 ThisBuild / organization := "org.typelevel"
 ThisBuild / organizationName := "Typelevel"
--- a/core/js/src/main/scala/cats/effect/IOPlatform.scala
+++ b/core/js/src/main/scala/cats/effect/IOPlatform.scala
@@ -16,6 +16,7 @@
 
 package cats.effect
 
+import scala.concurrent.Future
 import scala.scalajs.js.{|, Function1, JavaScriptException, Promise, Thenable}
 
 abstract private[effect] class IOPlatform[+A] { self: IO[A] =>
@@ -35,4 +36,11 @@ abstract private[effect] class IOPlatform[+A] { self: IO[A] =>
           resolve(value)
           ()
       })
+
+  def unsafeRunSyncToFuture()(implicit runtime: unsafe.IORuntime): Future[A] =
+    self.syncStep(runtime.config.autoYieldThreshold).attempt.unsafeRunSync() match {
+      case Left(t) => Future.failed(t)
+      case Right(Left(ioa)) => ioa.unsafeToFuture()
+      case Right(Right(a)) => Future.successful(a)
+    }
 }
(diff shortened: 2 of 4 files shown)


armanbilge (dismissed):
I mean, this looks really great to me! 😁 

I think it's worth deprecating the old `syncStep`, totally removing it without a deprecation phase is a bit rough :P but it would be good to force users to reckon with the ceding/limit issue.

armanbilge:
It occurred to me, we should also have `unsafeRunSyncToPromise` since this is for JS anyway :)

armanbilge (approved):
@djspiewak I added an `unsafeRunSyncToPromise` method as well, since this is JS land.

PR #2617, opened 2021-12-01
Add Resource#allocatedCase

### LINKED ISSUE #2472: Unexpected finalization of uncancelable resource
**Scala version:** 2.13.6
**Cats-effect version:** 3.2.9

**Minimal example:**
```scala
import cats.effect.{IO, IOApp, Resource}
import cats.effect.syntax.monadCancel._

  def resource(prefix: String): Resource[IO, Unit] =
    Resource.makeCase(IO.unit) {
      case (_, Resource.ExitCase.Succeeded) => IO.println(s"[$prefix]. Release. Success")
      case (_, Resource.ExitCase.Errored(e)) => IO.println(s"[$prefix]. Release. Error: $e")
      case (_, Resource.ExitCase.Canceled) => IO.println(s"[$prefix]. Release. Canceled")
    }

  def run: IO[Unit] =
    for {
      r1 <- resource("general").use(_ => IO.raiseError(new RuntimeException("oops"))).attempt

      // emulate error
      r2 <- resource("uncancelable").uncancelable.use(_ => IO.raiseError(new RuntimeException("oops"))).attempt

      // emulate cancelation
      d <- IO.deferred[Unit]
      fiber <- resource("uncancelable cancel").uncancelable.use(_ => d.complete(()) >> IO.never).start
      _ <- d.get // wait till `use` hit `never`
      _ <- fiber.cancel
      
      _ <- IO.println("r1: " + r1)
      _ <- IO.println("r2: " + r2)
    } yield ()

}
```

The output:
```
[general]. Release. Error: java.lang.RuntimeException: oops
[uncancelable]. Release. Success
[uncancelable cancel]. Release. Success

r1: Left(java.lang.RuntimeException: oops)
r2: Left(java.lang.RuntimeException: oops)
```

For some reason, a resource that has been marked as `uncancelable` always has `ExitCase.Succeeded` during the release stage. Even though it has been canceled or completed with an error.

`Resource.uncancelable[IO, Unit](_ => resource("uncancelable"))` has the same behavior.

TimWSpence:
Add `Resource#allocatedCase` as a dual of `applyCase`.

Addresses #2472 by tracking ExitCases. Alters existing combinators to take advantage of `allocatedCase` instead of `allocated`

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -385,7 +385,7 @@ sealed abstract class Resource[F[_], +A] {
     def continue[C](
         current: Resource[F, C],
         stack: Stack[C],
-        release: ExitCase => F[Unit]): F[(B, ExitCase => F[Unit])] =
+        release: List[ExitCase => F[Unit]]): F[(B, List[ExitCase => F[Unit]])] =
       loop(current, stack, release)
 
     // Interpreter that knows how to evaluate a Resource data structure;
@@ -393,13 +393,13 @@ sealed abstract class Resource[F[_], +A] {
     @tailrec def loop[C](
         current: Resource[F, C],
         stack: Stack[C],
-        release: ExitCase => F[Unit]): F[(B, ExitCase => F[Unit])] =
+        release: List[ExitCase => F[Unit]]): F[(B, List[ExitCase => F[Unit]])] =
       current match {
         case Allocate(resource) =>
           F uncancelable { poll =>
             resource(poll) flatMap {
               case (b, rel) =>
-                val rel2 = (ec: ExitCase) => rel(ec).guarantee(release(ec))
+                val rel2 = rel :: release
 
                 stack match {
                   case Nil =>
@@ -439,7 +439,13 @@ sealed abstract class Resource[F[_], +A] {
           fa.flatMap(a => continue(Resource.pure(a), stack, release))
       }
 
-    loop(this, Nil, _ => F.unit)
+    loop(this, Nil, List.empty).map {
+      case (b, rels) =>
+        (
+          b,
+          (e: ExitCase) =>
+            rels.foldLeft[F[Unit]](F.unit) { case (rel, f) => rel.guarantee(f(e)) })
+    }
   }
 
   /**
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -374,7 +374,7 @@ sealed abstract class Resource[F[_], +A] {
    * release actions (like the `before` and `after` methods of many test frameworks), or complex
    * library code that needs to modify or move the finalizer for an existing resource.
    */
-  def allocatedFull[B >: A](
+  def allocatedCase[B >: A](
       implicit F: MonadCancel[F, Throwable]): F[(B, ExitCase => F[Unit])] = {
     sealed trait Stack[AA]
     case object Nil extends Stack[B]
@@ -462,7 +462,7 @@ sealed abstract class Resource[F[_], +A] {
    */
   def allocated[B >: A](implicit F: MonadCancel[F, Throwable]): F[(B, F[Unit])] =
     F.uncancelable(poll =>
-      poll(allocatedFull).map { case (b, fin) => (b, fin(ExitCase.Succeeded)) })
+      poll(allocatedCase).map { case (b, fin) => (b, fin(ExitCase.Succeeded)) })
 
   /**
    * Applies an effectful transformation to the allocated resource. Like a `flatMap` on `F[A]`
@@ -495,19 +495,19 @@ sealed abstract class Resource[F[_], +A] {
     }
 
   def forceR[B](that: Resource[F, B])(implicit F: MonadCancel[F, Throwable]): Resource[F, B] =
-    Resource.applyFull { poll => poll(this.use_ !> that.allocatedFull) }
+    Resource.applyFull { poll => poll(this.use_ !> that.allocatedCase) }
 
   def !>[B](that: Resource[F, B])(implicit F: MonadCancel[F, Throwable]): Resource[F, B] =
     forceR(that)
 
   def onCancel(fin: Resource[F, Unit])(implicit F: MonadCancel[F, Throwable]): Resource[F, A] =
-    Resource.applyFull { poll => poll(this.allocatedFull).onCancel(fin.use_) }
+    Resource.applyFull { poll => poll(this.allocatedCase).onCancel(fin.use_) }
 
   def guaranteeCase(
       fin: Outcome[Resource[F, *], Throwable, A @uncheckedVariance] => Resource[F, Unit])(
       implicit F: MonadCancel[F, Throwable]): Resource[F, A] =
     Resource.applyFull { poll =>
-      poll(this.allocatedFull).guaranteeCase {
+      poll(this.allocatedCase).guaranteeCase {
         case Outcome.Succeeded(ft) =>
           fin(Outcome.Succeeded(Resource.eval(ft.map(_._1)))).use_.handleErrorWith { e =>
             ft.flatMap(_._2(ExitCase.Errored(e))).handleError(_ => ()) >> F.raiseError(e)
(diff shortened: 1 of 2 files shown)


TimWSpence:
I haven't ported `start` to use `allocatedFull` yet. It wasn't immediately obvious to me how to do it. I'm assuming it's possible and that I should do it?

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -383,21 +385,21 @@ sealed abstract class Resource[F[_], +A] {
...
         stack: Stack[C],
-        release: F[Unit]): F[(B, F[Unit])] =
+        release: ExitCase => F[Unit]): F[(B, ExitCase => F[Unit])] =
       current match {
         case Allocate(resource) =>
           F uncancelable { poll =>
             resource(poll) flatMap {
               case (b, rel) =>
-                val rel2 = rel(ExitCase.Succeeded).guarantee(release)
+                val rel2 = (ec: ExitCase) => rel(ec).guarantee(release(ec))
```
TimWSpence: This is not stack-safe 😭 It would be trivial if we had `Defer` here but I can't force that on things like `uncancelable` that call this now. If anyone has any thoughts then I would love to hear them!

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -954,30 +960,32 @@ object Resource extends ResourceFOInstances0 with ResourceHOInstances0 with Reso
...
             def apply[G[_]](implicit G: MonadCancel[G, Throwable]) = { (cb, ga, nt) =>
-              type D[A] = Kleisli[G, Ref[G, F[Unit]], A]
+              type D[A] = Kleisli[G, Ref[G, ExitCase => F[Unit]], A]
 
               val nt2 = new (Resource[F, *] ~> D) {
                 def apply[A](rfa: Resource[F, A]) =
                   Kleisli { r =>
-                    nt(rfa.allocated) flatMap { case (a, fin) => r.update(_ !> fin).as(a) }
+                    nt(rfa.allocatedFull) flatMap {
+                      case (a, fin) => r.update(f => (ec: ExitCase) => f(ec) !> fin(ec)).as(a)
```
TimWSpence: TODO: This looks stack-unsafe as well. Need to check!

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -1078,13 +1086,14 @@ object Resource extends ResourceFOInstances0 with ResourceHOInstances0 with Reso
...
-        val alloc = self.allocated.allocated
+      Resource.applyFull { poll =>
+        val alloc = self.allocatedFull.allocatedFull
 
         poll(alloc) map {
           case ((a, rfin), fin) =>
-            val composedFinalizers = fin !> rfin.allocated.flatMap(_._2)
-            (a, (_: Resource.ExitCase) => composedFinalizers)
+            val composedFinalizers =
+              (ec: ExitCase) => fin(ec) !> rfin(ec).allocatedFull.flatMap(_._2(ec))
```
TimWSpence: As above - stack-unsafe?

SystemFw:
See if you can add tests for the stack safety issues please. This trick should solve it:
https://github.com/typelevel/cats-effect/blob/series/3.x/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala#L333

TimWSpence:
> See if you can add tests for the stack safety issues please. This trick should solve it: https://github.com/typelevel/cats-effect/blob/series/3.x/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala#L333

Yeah absolutely! I already added one for `allocatedFull` and thought I'd get the implementation of that sorted first before I looked at the others

And thanks! 👍

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -437,9 +440,29 @@ sealed abstract class Resource[F[_], +A] {
...
+   * For this reason, this is an advanced and potentially unsafe api which can cause a resource
+   * leak if not used correctly, please prefer [[use]] as the standard way of running a
+   * `Resource` program.
+   *
+   * Use cases include interacting with side-effectful apis that expect separate acquire and
+   * release actions (like the `before` and `after` methods of many test frameworks), or complex
+   * library code that needs to modify or move the finalizer for an existing resource.
+   */
+  def allocated[B >: A](implicit F: MonadCancel[F, Throwable]): F[(B, F[Unit])] =
+    allocatedFull.map { case (b, fin) => (b, fin(ExitCase.Succeeded)) }
```
SystemFw: I think this requires `uncancelable`

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -383,21 +385,22 @@ sealed abstract class Resource[F[_], +A] {
...
-        release: F[Unit]): F[(B, F[Unit])] =
+        release: ExitCase => F[Unit]): F[(B, ExitCase => F[Unit])] =
       current match {
         case Allocate(resource) =>
           F uncancelable { poll =>
             resource(poll) flatMap {
               case (b, rel) =>
-                val rel2 = rel(ExitCase.Succeeded).guarantee(release)
+                // Insert F.unit to emulate defer for stack-safety
+                val rel2 = (ec: ExitCase) => rel(ec).guarantee(F.unit >> release(ec))
```
SystemFw, line 403: I suggested this because it solved a very similar problem in Resource, but just double checking: you've fixed an actual failed test with this right?

SystemFw (dismissed):

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
SystemFw: Ah, I had forgotten the details of our previous convo, you did :)

SystemFw:
This is a great addition Tim ❤️

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
TimWSpence: 🤦 I in fact used this precise example in the `MonadCancel` docs 😂

vasilmkd:
Bug fixes for days! Sorry I wasn't around for the review.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -373,7 +374,8 @@ sealed abstract class Resource[F[_], +A] {
    * release actions (like the `before` and `after` methods of many test frameworks), or complex
    * library code that needs to modify or move the finalizer for an existing resource.
    */
-  def allocated[B >: A](implicit F: MonadCancel[F, Throwable]): F[(B, F[Unit])] = {
+  def allocatedFull[B >: A](
```
armanbilge: Bikeshedding question: would `allocatedCase` be a more appropriate name for this method? At least based on the `make` functions, `makeFull` has to do with `Poll`, `makeCase` has to do with tracking the `ExitCase` (as is done here), and `makeCaseFull` has to do with both.

vasilmkd: Yes, I agree with this.

TimWSpence: Thanks! Very good point 👍

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -373,7 +374,8 @@ sealed abstract class Resource[F[_], +A] {
    * release actions (like the `before` and `after` methods of many test frameworks), or complex
    * library code that needs to modify or move the finalizer for an existing resource.
    */
-  def allocated[B >: A](implicit F: MonadCancel[F, Throwable]): F[(B, F[Unit])] = {
+  def allocatedCase[B >: A](
```
armanbilge:
```suggestion
  private[effect] def allocatedCase[B >: A](
```

@vasilmkd just want to point out that regardless of our release strategy for patches this time around, we can get this PR in like this by keeping the new API internal since it still fixes a lot of bugs in existing public APIs. Then when we are ready we can un-privatize it :)

vasilmkd: I think we should add it in 3.3.1, but let's wait for Daniel.

armanbilge: @djspiewak 👍 in discord
https://discord.com/channels/632277896739946517/839263556754472990/922206112730722325

djspiewak (approved):
This is awesome thank you!

armanbilge:
Failure is b/c of https://github.com/typelevel/cats-effect/pull/2665
```
[info] AsyncPlatform CompletableFuture conversion should
[error]   x cancel CompletableFuture on fiber cancellation
[error]    Expected: java.util.concurrent.CancellationException. Got nothing (AsyncPlatformSpec.scala:40)
```

I think it's just a flaky test, and I think I know how to fix it.

PR #2635, opened 2021-12-09
Adding missing parameter to `retry` in test-runtime.md docs.

zarthross:


docs/core/test-runtime.md
```diff
@@ -46,7 +46,7 @@ def retry[A](ioa: IO[A], delay: FiniteDuration, max: Int, random: Random[IO]): I
   else
     ioa handleErrorWith { _ =>
       random.betweenLong(0L, delay.toNanos) flatMap { ns =>
-        IO.sleep(ns.nanos) *> retry(ioa, delay * 2, max - 1)
+        IO.sleep(ns.nanos) *> retry(ioa, delay * 2, max - 1, random)
```
armanbilge, line 49: Thanks for the PR! If you don't mind fixing it, I think the `mdoc scala` for this block needs to become a `scala mdoc` so this mistake won't slip through next time :) might need some additional modifiers ...

zarthross: Done.  There are others in this file that could probably use the same change in order to compile, but since `testkit` isn't a dep of docs I'd have to add it.  Happy to do a follow up PR to do that if you think its fine to add that dep.

armanbilge: Thanks! `testkit` as a dep of docs seems fine to me, a followup PR would be much appreciated!

armanbilge (approved):

Issue #2634, opened 2021-12-07
[3.3.0] Performance regression


iRevive:
_This issue was originally created in fs2, so version numbers below refer to fs2_.

I've observed a performance degradation in some scenarios on **3.2.3** version. 

The throughput on small byte streams remained the same. 
The throughput on bigger byte streams is decreased roughly by ~20%.  
The memory allocation decreased by 10-15% on bigger byte streams.  

See the benchmarks below for more details.

#### Stream usage

The project utilizes a TCP socket from the fs2-io module. I cannot share many details due to NDA, but the generalized usage of the socket is the following one:
```scala
val streamDecoder: StreamDecoder[Structure] =
  StreamDecoder.many(StructureDecoder)

socket.reads
  .through(streamDecoder.toPipeByte)
  .evalMap(structure => queue.offer(Right(structure)))
  .compile
  .background
```

#### Consumed bytes per single invocation:
- createOne: 167 bytes
- returnRandomUUID: 135 bytes
- return100Record: 2683 bytes

## 3.2.2 

Operation average time:
```
Benchmark                           Mode  Cnt   Score        Error   Units
DriverBenchmark.createOne           avgt   25   4.725 ±      0.396   ms/op
DriverBenchmark.return100Records    avgt   25   8.613 ±      0.488   ms/op
DriverBenchmark.returnRandomUUID    avgt   25   3.542 ±      1.156   ms/op
```

Memory allocation:
```
DriverBenchmark.return100Records:·gc.alloc.rate                        201.323 ±     11.576  MB/sec
DriverBenchmark.return100Records:·gc.alloc.rate.norm               3117144.635 ±  13036.746    B/op
DriverBenchmark.return100Records:·gc.churn.G1_Eden_Space               199.083 ±     23.093  MB/sec
DriverBenchmark.return100Records:·gc.churn.G1_Eden_Space.norm      3076695.652 ± 283692.848    B/op
DriverBenchmark.return100Records:·gc.churn.G1_Old_Gen                    0.074 ±      0.113  MB/sec
DriverBenchmark.return100Records:·gc.churn.G1_Old_Gen.norm            1266.657 ±   1981.461    B/op
DriverBenchmark.return100Records:·gc.churn.G1_Survivor_Space             0.484 ±      0.444  MB/sec
DriverBenchmark.return100Records:·gc.churn.G1_Survivor_Space.norm     7614.493 ±   6998.312    B/op
[...]

PR #2642, opened 2021-12-10
Only use `FiberAwareExecutionContext` if tracing is enabled

armanbilge:
Addresses the JS performance regression reported by @rpiaggio in https://github.com/typelevel/cats-effect/issues/2634#issuecomment-991202431. Credit to @vasilmkd for identifying the problem.


vasilmkd:
Sorry for being picky, but for anyone reading this, this PR does not actually improve performance, it just makes sure that when tracing is disabled, there are no debugging features enabled in the runtime, in this case, tracking fibers for generating fiber dumps.

vasilmkd (approved):

armanbilge:
Thanks for the clarification 👍

Issue #2636, opened 2021-12-09
joinWithUnit for Fibers


etspaceman:
We have `joinWithNever`, in which a cancelled fiber is suspended and never finished. This mimics the behavior of CE2 from what I understand.

I would like to see a `joinWithUnit`, in which a cancelled fiber's result is ignored. Something like this:

```scala
  def joinWithUnit(implicit F: GenSpawn[F, E]): F[A] =
    joinWith(F.unit)
```

armanbilge:
I think for this to work you'd need need an implicit `ev: A <:< Unit` (i.e., the `Fiber` itself has to be of `Unit` return type. Edit: `Unit <:< A` actually 🤔

PR #2643, opened 2021-12-11
Adds joinWithUnit for Fiber

etspaceman:
Solves #2636

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Fiber.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Fiber.scala
@@ -80,6 +80,8 @@ trait Fiber[F[_], E, A] {
    * fiber completes with [[Outcome.Errored]], the error is raised. If the fiber completes with
    * [[Outcome.Canceled]], the result is ignored.
    */
-  def joinWithUnit(implicit F: GenSpawn[F, E], ev: Unit <:< A): F[A] =
-    joinWith(F.unit.map(ev(_)))
+  def joinWithUnit(implicit F: MonadCancel[F, E], ev: Unit <:< A): F[A] = {
+    val _ = ev
+    joinWith(F.unit.asInstanceOf[F[A]])
+  }
 }
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Fiber.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Fiber.scala
@@ -72,4 +72,14 @@ trait Fiber[F[_], E, A] {
    */
   def joinWithNever(implicit F: GenSpawn[F, E]): F[A] =
     joinWith(F.never)
+
+  /**
+   * Awaits the completion of the bound fiber and returns its result once it completes.
+   *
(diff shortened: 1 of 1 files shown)


vasilmkd:
I'm against this method because it is no different from `joinWith(F.unit)` if your effect is already `F[Unit]`.

etspaceman:
Its really just a convenience method for a common pattern. What's the issue with adding it?

vasilmkd:
No issue. I was just voicing my opinion. My opinion may be wrong. 😄

armanbilge:
> I'm against this method because it is no different from `joinWith(F.unit)` if your effect is already `F[Unit]`.

Exactly, saves some typing and gets you autocomplete :) unless there's some reason this is an anti-pattern we don't want to encourage I'm 👍

etspaceman:
@djspiewak any thoughts on this one?

kernel/shared/src/main/scala/cats/effect/kernel/Fiber.scala
```diff
@@ -72,4 +72,14 @@ trait Fiber[F[_], E, A] {
...
     joinWith(F.never)
+
+  /**
+   * Awaits the completion of the bound fiber and returns its result once it completes.
+   *
+   * If the fiber completes with [[Outcome.Succeeded]], the successful value is returned. If the
+   * fiber completes with [[Outcome.Errored]], the error is raised. If the fiber completes with
+   * [[Outcome.Canceled]], the result is ignored.
+   */
+  def joinWithUnit(implicit F: GenSpawn[F, E], ev: Unit <:< A): F[A] =
```
djspiewak:
```suggestion
  def joinWithUnit(implicit F: MonadCancel[F, E], ev: Unit <:< A): F[A] =
```

kernel/shared/src/main/scala/cats/effect/kernel/Fiber.scala
```diff
@@ -72,4 +72,14 @@ trait Fiber[F[_], E, A] {
...
+    joinWith(F.unit.map(ev(_)))
```
djspiewak: I'm actually in favor of using a cast here since it avoids an allocation.

```suggestion
    val _ = ev
    joinWith(F.unit.asInstanceOf[F[A]])
```

(also assume the addition of some wrapping braces)

djspiewak (changes requested):
I think this is honestly quite reasonable. It's a fairly common pattern and the auto-complete alone is worth it. Left some minor comments.

kernel/shared/src/main/scala/cats/effect/kernel/Fiber.scala
etspaceman: This is done

kernel/shared/src/main/scala/cats/effect/kernel/Fiber.scala
etspaceman: This is done

djspiewak (approved):

PR #2646, opened 2021-12-11
Relax `Resource#mapK` constraints

armanbilge:
Branch based on https://github.com/typelevel/cats-effect/pull/2617.

Earlier on [Discord](https://discord.com/channels/632277896739946517/632278585700384799/919255475772604446) a user lamented the need for a `MonadCancel` constraint on _both_ the `F` and the `G` of `Resource#mapK`. This prompted a lengthy discussion about the necessity for the constraint on `G`.

Here, I've re-implemented `mapK` in terms of the new `allocatedFull` and accordingly the constraint on `G` is reduced to `Functor`.

However, the constraint on `F` did have to be strengthened from `MonadCancel[F, _]` to `MonadCancel[F, Throwable]` because `allocatedFull` expects the latter constraint. This manifests as a source-breaking change that will ripple through e.g. semaphore.

I also made an attempt to completely remove `Functor` constraint on `G` in https://github.com/armanbilge/cats-effect/commit/89f4f41507084e9ef0ef201076a69b7d5e0f72c5. This requires creating a `MapK` node. This is difficult for two reasons:
1. `Resource` is a sealed public ADT used in e.g. fs2, so adding new nodes breaks compatibility.
2. `Resource#attempt[E]` is difficult (impossible?) to implement for a generic `E`, because of the hard-coding for `Throwable` mentioned above.

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -385,7 +385,7 @@ sealed abstract class Resource[F[_], +A] {
     def continue[C](
         current: Resource[F, C],
         stack: Stack[C],
-        release: ExitCase => F[Unit]): F[(B, ExitCase => F[Unit])] =
+        release: List[ExitCase => F[Unit]]): F[(B, List[ExitCase => F[Unit]])] =
       loop(current, stack, release)
 
     // Interpreter that knows how to evaluate a Resource data structure;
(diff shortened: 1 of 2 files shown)


armanbilge:
If there's interest in this, I'd like to take one more stab at it. I still have a small optimism it may be possible to:
1. completely remove the constraint on `G`, without adding another node to the ADT
2. not force the error to be `Throwable`

Not really sure, but I have some ideas 🙃

PR #2647, opened 2021-12-12
Port the `Sync.blocking` scaladoc to `IO.blocking`

danicheg:
Also, this adds a note about the cancelation of effects created via `blocking`.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -1549,7 +1549,9 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
      * Like [[IO.delay]] but intended for thread blocking operations. `blocking` will shift the
      * execution of the blocking operation to a separate threadpool to avoid blocking on the main
      * execution context. See the thread-model documentation for more information on why this is
-     * necessary.
+     * necessary. Note that created effect will be uncancelable. If you're wondering about
+     * an interruption of evaluation, for instance via [[IO.timeout]], then you should use
+     * [[IO.interruptible]] or [[IO.interruptibleMany]].
      *
      * {{{
      * IO.blocking(scala.io.Source.fromFile("path").mkString)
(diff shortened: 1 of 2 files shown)


vasilmkd (dismissed):

kernel/shared/src/main/scala/cats/effect/kernel/Sync.scala
```diff
@@ -63,7 +63,9 @@ trait Sync[F[_]] extends MonadCancel[F, Throwable] with Clock[F] with Unique[F]
    * Like [[Sync.delay]] but intended for thread blocking operations. `blocking` will shift the
    * execution of the blocking operation to a separate threadpool to avoid blocking on the main
    * execution context. See the thread-model documentation for more information on why this is
-   * necesary.
+   * necessary. Note that created effect will be uncancelable. If you're wondering about
+   * an interruption of evaluation then you should use [[Sync.interruptible]]
+   * or [[Sync.interruptibleMany]].
```
armanbilge:
```suggestion
   * necessary. Note that the created effect will be uncancelable; if you need cancellation
   * then you should use [[Sync.interruptible]] or [[Sync.interruptibleMany]].
```

armanbilge: This is a good note! Apologies for the nitpicking.

danicheg: Thanks! Only I changed 'cancellation' -> 'cancelation'

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1545,6 +1545,22 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
     override def delay[A](thunk: => A): IO[A] = IO(thunk)
 
+    /**
+     * Like [[IO.delay]] but intended for thread blocking operations. `blocking` will shift the
+     * execution of the blocking operation to a separate threadpool to avoid blocking on the main
+     * execution context. See the thread-model documentation for more information on why this is
+     * necessary. Note that created effect will be uncancelable. If you're wondering about
+     * an interruption of evaluation, for instance via [[IO.timeout]], then you should use
+     * [[IO.interruptible]] or [[IO.interruptibleMany]].
```
armanbilge: Sorry, can we update this one too 😅

armanbilge (approved):
LGTM thank you very much!

vasilmkd:
@danicheg Thanks for this PR. Very much appreciated. @armanbilge Thanks for nailing the final details.

PR #2650, opened 2021-12-13
Implement `std.Env[F]` for cross-platform environment variables

### LINKED ISSUE #2532: `std.Env[F]` for cross-platform environment variables?
Currently there is no JVM/JS cross-platform API for accessing environment variables. @bpholt and I wonder whether this might makes sense to include in CE3 `std`. 

Why:
* `IOApp` already provides cross-platform command line args via a shim for JS. It doesn't seem too much of a stretch to also provide cross-platform env.
* We've actually already internally shimmed env for JS, to configure tracing etc. at runtime. So it would just be a matter of exposing that code in a public API.
* `Console[F]` might also be a precedent for this kind of thing?

Why not:
* Dubious whether reading environment variables is an effect. @ChristopherDavenport had some thoughts about this.
* Just generally out-of-scope for CE3.

armanbilge:
Fixes https://github.com/typelevel/cats-effect/issues/2532. As @SystemFw [pointed out](https://discord.com/channels/632277896739946517/839263556754472990/919679231993184406):
> I think some things should be quite uncontroversial, like `Env`
> it's used all over the place and the design space is tiny

So, here we are 😁


core/js/src/main/scala/cats/effect/process.scala
```diff
@@ -26,12 +28,10 @@ private[effect] object process {
...
 
-  def env(key: String): Option[String] =
-    Try(js.Dynamic.global.process.env.selectDynamic(key))
-      .orElse(Try(js.Dynamic.global.process.env.selectDynamic(s"REACT_APP_$key")))
-      .widen[Any]
-      .collect { case v: String if !js.isUndefined(v) => v }
-      .toOption
+  def env(key: String): Option[String] = {
+    val env = Env.make[SyncIO]
+    OptionT(env.get(key)).orElseF(env.get(s"REACT_APP_$key")).value.unsafeRunSync()
```
kubukoz, line 33: I don't really like this, feels like React users could add this fallback on their own (or just use the key with this prefixed from the beginning).

kubukoz (dismissed):

std/jvm/src/main/scala/cats/effect/std/EnvCompanionPlatform.scala
```diff
@@ -0,0 +1,29 @@
...
+
+package cats.effect.std
+
+import cats.effect.kernel.Sync
+
+import scala.collection.immutable.Map
+
+private[std] class EnvCompanionPlatform {
+  private[std] final class SyncEnv[F[_]](implicit F: Sync[F]) extends Env[F] {
+    def get(name: String): F[Option[String]] = F.delay(sys.env.get(name))
```
kubukoz: `System.getenv` should be better as it doesn't force the entire environment. I think there's a difference in the security manager...

core/js/src/main/scala/cats/effect/process.scala
armanbilge: FTR this is only used for detecting CE config (e.g., tracing, etc.).

tests/shared/src/test/scala/cats/effect/std/EnvSpec.scala
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package cats.effect
+package std
+
+class EnvSpec extends BaseSpec {
+
+  "Env" should {
+    "retrieve a variable from the environment" in real {
+      Env[IO].get("HOME").flatMap(x => IO(x must beSome))
```
kubukoz, line 24: You could set a concrete value with `envVars` in sbt (assuming we have `fork` in the JVM tests), should make for more accurate testing.

armanbilge: Yeah, tried that first. Doesn't work for Node.js and was lazy to split the tests (:

core/js/src/main/scala/cats/effect/process.scala
kubukoz: ah wait I thought this was used _in_ Env. LGTM

tests/shared/src/test/scala/cats/effect/std/EnvSpec.scala
kubukoz: nope, it was actually skipped lmao

std/jvm/src/main/scala/cats/effect/std/EnvCompanionPlatform.scala
armanbilge: Oh, good point. Didn't know that, here's the source for anyone curious.
https://github.com/scala/scala/blob/4b2151cbe4c857d49556b70c3fa0f5e236fd7754/src/library/scala/sys/package.scala#L67

std/jvm/src/main/scala/cats/effect/std/EnvCompanionPlatform.scala
```diff
@@ -0,0 +1,32 @@
...
+import scala.collection.immutable.Map
+
+private[std] class EnvCompanionPlatform {
+  private[std] final class SyncEnv[F[_]](implicit F: Sync[F]) extends Env[F] {
+
+    def get(name: String): F[Option[String]] =
+      F.delay(Option(System.getenv(name))) // sys.env copies the entire env into a Map
+
+    def toMap: F[Map[String, String]] =
+      F.delay(Map.from(sys.env)) // a somewhat redundant copy, to shake the unsafe withDefault
```
armanbilge: Alternatively, we could build the `Map` ourselves directly from `System.getenv()`, but I didn't want to fight deprecation warnings for the Java collections converters across 2.12/2.13.

rossabaker (changes requested):
:+1: on adding it to `std`.  :-1: on `toMap`.  :-1: to me appealing to Haskell, but I think they got this right.

armanbilge:
Thanks. I don't have any strong feeling about `toMap` (it's honestly annoying to implement) but I am curious why it's not a good idea generally :)

rossabaker:
Hmm.  My inline comment did not arrive with my ❌.  Sorry.

On some platforms, notably Windows, environments are case insensitive.  That's why the Scala map has a default, and why even `Map.empty ++ sys.env` breaks the property that `get(x) <-> toMap(_.get(x))`.  You can hack around that, but a `Map` whose lookup is inconsistent with the equality of its keys will always do weird shit.

In [Haskell](https://hackage.haskell.org/package/base-4.16.0.0/docs/System-Environment.html), they have (translating to Scala and tagless):
* `String => F[Option[String]]` (like our `get`)
* `F[List[(String, String)]]` (instead of our `toMap`).

We might quibble about `List` and the ordering it might imply, but separating the lookup from the iteration insulates the base library from weird shit.

armanbilge:
Aha, thanks. That was bothering me too. Returning a `List` of key-value pairs seems good to me. B/c of the unordered thing, could we return a `Set` here?

rossabaker:
`Set`'s featured operation is `contains`, which has the same problems as the `Map`.  `scala.collection.immutable.Iterable` says just the right amount here.

std/shared/src/main/scala/cats/effect/std/Env.scala
```diff
@@ -0,0 +1,109 @@
...
+import scala.collection.immutable.Iterable
+
+/**
+ * Effect type agnostic `Env` with common methods to read environment variables
+ */
+trait Env[F[_]] { self =>
+
+  def get(name: String): F[Option[String]]
+
+  def entries: F[Iterable[(String, String)]]
```
armanbilge, line 33: 🚲 🏠 ?

kubukoz: lgtm

rossabaker (approved):
Now people can pick their own tradeoffs with `entries.map(SomeOtherCollectionFactory)`... or probably just iterate over it.

djspiewak (approved):
Tentatively approved, though I'm not thrilled about `Iterable`. Does Cats have a `Foldable[Iterable]`?

rossabaker:
There is not.  There is a `Foldable[collection.Iterable]` in alleycats.  Some of the `Foldable` operations suggest an ordering, which is the reason I objected to `List`.

A `List` isn't the end of the world: the ordering is arbitrary, but harmless.  I still think `Map` and `Set` lead to platform-specific dragons.

PR #2654, opened 2021-12-15
Enrich the `delay` scaladoc for `IO` and `Sync`

danicheg:
This is a minor thing but should help in picking a suitable constructor.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -932,8 +932,7 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
   /**
    * Suspends a synchronous side effect in `IO`.
    * Use [[IO.apply]] if your side effect is not thread-blocking;
-   * otherwise you should use [[IO.blocking]] or `IO.interruptible`
-   * depending on the need of the created effect cancelation.
+   * otherwise you should use [[IO.blocking]] (uncancelable) or `IO.interruptible` (cancelable).
    *
    * Alias for [[IO.delay]].
    */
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -931,13 +931,19 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
   /**
    * Suspends a synchronous side effect in `IO`.
+   * Use [[IO.apply]] if your side effect is not thread-blocking;
+   * otherwise you should use [[IO.blocking]] or `IO.interruptible`
```
danicheg: Unfortunately, the direct link to the `IO.interruptible` is ambiguous. And it could be fixed only by a fully qualified name. That's too extra, I think.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -931,13 +931,19 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
...
+   * depending on the need of the created effect cancelation.
```
```suggestion
   * otherwise you should use [[IO.blocking]] (uncancelable) or `IO.interruptible` (cancelable).
```
WDYT?

core/shared/src/main/scala/cats/effect/IO.scala
armanbilge: Hmm, that's annoying. I actually don't know how to do that anyway :)

core/shared/src/main/scala/cats/effect/IO.scala
danicheg: It looks better, thanks.

armanbilge (approved):
Thanks, this will help users a lot I think!

vasilmkd (approved):
Thank you so much!

PR #2657, opened 2021-12-16
Introduce `IORuntimeBuilder`

majk-p:
I'm working on migrating applications from CE2 to CE3. In CE2 I used `IOApp` and prior to running any `IO`, I'd override the default execution context with the one wrapped with Kanela.

With CE3 we have this `IORuntime` that hides the underlying execution context. If I wanted to reproduce that approach with CE3, I'd basically have to create the whole `IORuntime` on my own. Instead of doing that, doing `runtime.copy(context = wrap(runtime.context))` would be enough.

This PR introduces a method that exposes the necessary API for my use case. What do you think about it? Is it the right approach or is there anything else I should try?

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeBuilderPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeBuilderPlatform.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020-2021 Typelevel
+ * Copyright 2020-2022 Typelevel
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeBuilderPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeBuilderPlatform.scala
@@ -35,8 +35,8 @@ private[unsafe] abstract class IORuntimeBuilderPlatform { self: IORuntimeBuilder
     val runtimeConfig = customConfig.getOrElse(IORuntimeConfig())
 
     runtime = IORuntime.apply(
-      computeWrapper(compute),
-      blockingWrapper(blocking),
+      computeTransform(compute),
+      blockingTransform(blocking),
       scheduler,
       shutdown,
       runtimeConfig
--- a/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeBuilderPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeBuilderPlatform.scala
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2020-2021 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.unsafe
+import scala.concurrent.ExecutionContext
+
+private[unsafe] abstract class IORuntimeBuilderPlatform {
+
+  def build(
+      computeWrapper: ExecutionContext => ExecutionContext = identity,
+      blockingWrapper: ExecutionContext => ExecutionContext = identity,
+      customConfig: Option[IORuntimeConfig] = None,
+      customScheduler: Option[Scheduler] = None,
+      customShutdown: Option[() => Unit] = None
+  ) = {
+    var runtime: IORuntime = null
+    val compute = IORuntime.createDefaultComputeThreadPool(runtime)._1
+    val blocking = IORuntime.createDefaultBlockingExecutionContext()._1
+    val scheduler = customScheduler.getOrElse(IORuntime.createDefaultScheduler()._1)
+    val shutdown = customShutdown.getOrElse(() => ())
+    val runtimeConfig = customConfig.getOrElse(IORuntimeConfig())
+
+    runtime = IORuntime.apply(
+      computeWrapper(compute),
+      blockingWrapper(blocking),
+      scheduler,
+      shutdown,
+      runtimeConfig
+    )
+    runtime
+  }
+
+}
--- a/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeBuilderPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeBuilderPlatform.scala
@@ -30,7 +30,7 @@ private[unsafe] abstract class IORuntimeBuilderPlatform { self: IORuntimeBuilder
       computeShutdown()
       blockingShutdown()
       schedulerShutdown()
-      extraShutdownHooks.foreach(_())
+      extraShutdownHooks.reverse.foreach(_())
     }
     val runtimeConfig = customConfig.getOrElse(IORuntimeConfig())
 
--- a/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeBuilderPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeBuilderPlatform.scala
@@ -15,22 +15,23 @@
  */
 
 package cats.effect.unsafe
-import scala.concurrent.ExecutionContext
 
-private[unsafe] abstract class IORuntimeBuilderPlatform {
+private[unsafe] abstract class IORuntimeBuilderPlatform { self: IORuntimeBuilder =>
 
-  def build(
-      computeWrapper: ExecutionContext => ExecutionContext = identity,
-      blockingWrapper: ExecutionContext => ExecutionContext = identity,
-      customConfig: Option[IORuntimeConfig] = None,
-      customScheduler: Option[Scheduler] = None,
-      customShutdown: Option[() => Unit] = None
-  ) = {
+  protected def platformSpecificBuild = {
     var runtime: IORuntime = null
-    val compute = IORuntime.createDefaultComputeThreadPool(runtime)._1
-    val blocking = IORuntime.createDefaultBlockingExecutionContext()._1
-    val scheduler = customScheduler.getOrElse(IORuntime.createDefaultScheduler()._1)
-    val shutdown = customShutdown.getOrElse(() => ())
+    val (compute, computeShutdown) =
+      customCompute.getOrElse(IORuntime.createDefaultComputeThreadPool(runtime))
+    val (blocking, blockingShutdown) =
+      customBlocking.getOrElse(IORuntime.createDefaultBlockingExecutionContext())
+    val (scheduler, schedulerShutdown) =
+      customScheduler.getOrElse(IORuntime.createDefaultScheduler())
+    val shutdown = () => {
+      computeShutdown()
+      blockingShutdown()
+      schedulerShutdown()
+      extraShutdownHooks.foreach(_())
+    }
     val runtimeConfig = customConfig.getOrElse(IORuntimeConfig())
 
     runtime = IORuntime.apply(
--- a/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
+++ b/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
@@ -52,7 +52,6 @@ final class IORuntime private[unsafe] (
   private[effect] val traceBufferLogSize: Int = config.traceBufferLogSize
 
   override def toString: String = s"IORuntime($compute, $scheduler, $config)"
-
 }
 
 object IORuntime extends IORuntimeCompanionPlatform {
@@ -63,5 +62,4 @@ object IORuntime extends IORuntimeCompanionPlatform {
       shutdown: () => Unit,
       config: IORuntimeConfig): IORuntime =
     new IORuntime(compute, blocking, scheduler, FiberMonitor(compute), shutdown, config)
-
 }
--- a/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
+++ b/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
@@ -52,6 +52,15 @@ final class IORuntime private[unsafe] (
   private[effect] val traceBufferLogSize: Int = config.traceBufferLogSize
 
   override def toString: String = s"IORuntime($compute, $scheduler, $config)"
+
+  def wrappingCompute(wrapCompute: ExecutionContext => ExecutionContext): IORuntime = 
+    IORuntime(
+      wrapCompute(compute),
+      blocking,
+      scheduler,
+      shutdown,
+      config
+    )
 }
 
 object IORuntime extends IORuntimeCompanionPlatform {
--- a/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
+++ b/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
@@ -63,6 +63,6 @@ object IORuntime extends IORuntimeCompanionPlatform {
       config: IORuntimeConfig): IORuntime =
     new IORuntime(compute, blocking, scheduler, FiberMonitor(compute), shutdown, config)
 
-  def builder = 
+  def builder: IORuntimeBuilder =
     IORuntimeBuilder()
 }
--- a/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
+++ b/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
@@ -62,4 +62,7 @@ object IORuntime extends IORuntimeCompanionPlatform {
       shutdown: () => Unit,
       config: IORuntimeConfig): IORuntime =
     new IORuntime(compute, blocking, scheduler, FiberMonitor(compute), shutdown, config)
+
+  def builder = 
+    IORuntimeBuilder()
 }
--- a/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
+++ b/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
@@ -63,6 +63,6 @@ object IORuntime extends IORuntimeCompanionPlatform {
       config: IORuntimeConfig): IORuntime =
     new IORuntime(compute, blocking, scheduler, FiberMonitor(compute), shutdown, config)
 
-  def builder: IORuntimeBuilder =
+  def builder(): IORuntimeBuilder =
     IORuntimeBuilder()
 }
--- a/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
+++ b/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
@@ -53,14 +53,6 @@ final class IORuntime private[unsafe] (
 
   override def toString: String = s"IORuntime($compute, $scheduler, $config)"
 
-  def wrappingCompute(wrapCompute: ExecutionContext => ExecutionContext): IORuntime = 
-    IORuntime(
-      wrapCompute(compute),
-      blocking,
-      scheduler,
-      shutdown,
-      config
-    )
 }
 
 object IORuntime extends IORuntimeCompanionPlatform {
@@ -71,4 +63,6 @@ object IORuntime extends IORuntimeCompanionPlatform {
       shutdown: () => Unit,
       config: IORuntimeConfig): IORuntime =
     new IORuntime(compute, blocking, scheduler, FiberMonitor(compute), shutdown, config)
+
+  def builder: IORuntimeBuilder = IORuntimeBuilder()
 }
--- a/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
+++ b/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
@@ -64,5 +64,4 @@ object IORuntime extends IORuntimeCompanionPlatform {
       config: IORuntimeConfig): IORuntime =
     new IORuntime(compute, blocking, scheduler, FiberMonitor(compute), shutdown, config)
 
-  def builder: IORuntimeBuilder = IORuntimeBuilder()
 }
--- a/core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
+++ b/core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020-2021 Typelevel
+ * Copyright 2020-2022 Typelevel
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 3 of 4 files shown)


core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
```diff
@@ -52,6 +52,15 @@ final class IORuntime private[unsafe] (
   private[effect] val traceBufferLogSize: Int = config.traceBufferLogSize
 
   override def toString: String = s"IORuntime($compute, $scheduler, $config)"
+
+  def wrappingCompute(wrapCompute: ExecutionContext => ExecutionContext): IORuntime =
```
majk-p: Suggestions of the better name are obviously welcome :wink:

armanbilge:
@majk-p and I had a very nice chat in [Discord](https://discord.com/channels/632277896739946517/839263556754472990/921005196274958357) :) Some notes:
1. The current API is flexible enough to do what is needed, but too difficult to use. Particularly, the fact that `createDefaultComputeThreadPool` needs a reference to the runtime being created is confusing.
2. I like how this PR provides a simple interface to the user, but I'm worried if being able to access and share a compute pool between multiple `IORuntime`s could violate some of our internal assumptions e.g. when registering mbeans, fiber bags, etc. 
3. I propose we introduce an API `IORuntimeBuilder` that has the defaults baked in, but also exposes methods such as the one proposed in this PR. No resources should get allocated until the user `.build`s it. This should help solve the complexity in (1).

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  def withScheduler(scheduler: Scheduler) = {
+    customScheduler = Some(scheduler)
+    this
+  }
+
+  def withShutdown(shutdown: () => Unit) = {
+    customShutdown = Some(shutdown)
+    this
+  }
```
armanbilge: I'm not sure if we should make the shutdown overridable, since it's supposed to shutdown the threadpools that were created for this runtime. Instead, I think we should collect a shutdown for each option, and automatically combine the shutdowns in the builder.

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
```diff
@@ -0,0 +1,61 @@
...
+import scala.concurrent.ExecutionContext
+
+final class IORuntimeBuilder protected (
+    protected var computeWrapper: ExecutionContext => ExecutionContext = identity,
+    protected var blockingWrapper: ExecutionContext => ExecutionContext = identity,
+    protected var customConfig: Option[IORuntimeConfig] = None,
+    protected var customScheduler: Option[Scheduler] = None,
+    protected var customShutdown: Option[() => Unit] = None
+) extends IORuntimeBuilderPlatform {
+  def withComputeWrapper(wrapper: ExecutionContext => ExecutionContext) = {
```
armanbilge: This is good for your usecase :) can we also add options for someone to completely replace the compute and blocking pools?

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
```diff
@@ -0,0 +1,61 @@
...
+    blockingWrapper = wrapper.andThen(blockingWrapper)
+    this
+  }
+
+  def withConfig(config: IORuntimeConfig) = {
+    customConfig = Some(config)
+    this
+  }
+
+  def withScheduler(scheduler: Scheduler) = {
```
armanbilge: We should also ask for an operation to shutdown the `Scheduler`.

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeBuilderPlatform.scala
```diff
@@ -0,0 +1,46 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.unsafe
+import scala.concurrent.ExecutionContext
+
+private[unsafe] abstract class IORuntimeBuilderPlatform {
```
armanbilge:
```suggestion
private[unsafe] abstract class IORuntimeBuilderPlatform { self: IORuntimeBuilder =>
```

Then, you will be able to access all of its member variables, without need to request them as parameters in the `build` method below.

armanbilge (commented):
Nice job navigating the cross-platform stuff! I really like how this is looking :)

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
majk-p: Thank you, I wasn't sure how the shutdown logic works. I've done as asked, also I've added `def addShutdownHook(shutdown: () => Unit)` for introducing custom shutdown logic.

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
majk-p: Sure thing, done :wink:

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
majk-p: Done :+1:

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeBuilderPlatform.scala
majk-p: For some reason, I thought it wouldn't work when I first tried it. Seems like my IDE was playing tricks on me. Done now :+1:

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
```diff
@@ -0,0 +1,128 @@
...
+package cats.effect
+package unsafe
+
+import scala.concurrent.ExecutionContext
+
+/**
+ * Builder object for creating custom `IORuntime`s. Useful for creating `IORuntime` based on the
+ * default one but with some wrappers around execution contexts or custom shutdown hooks.
+ */
+final class IORuntimeBuilder protected (
```
vasilmkd, line 26: Default parameters are a bit of a pain to deal with in library APIs when it comes to binary compatibility. Let's wait for some of the more experienced folks to weigh in.

armanbilge: True. Fortunately this is a `protected` constructor (it can probably be made `private` btw :) so no concerns with bincompat.

vasilmkd: The public methods also contain default parameters. GitHub should allow file level comments.

armanbilge: Ah, haven't looked at the latest updates. Those defaults should probably just be removed IMO.

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
```diff
@@ -0,0 +1,128 @@
...
+
+  /**
+   * Override the defaul computing execution context
+   *
+   * @param compute
+   *   execution context for compute
+   * @param shutdown
+   *   method called upon compute context shutdown
+   */
+  def overrideCompute(compute: ExecutionContext, shutdown: () => Unit = () => ()) = {
```
armanbilge:
```suggestion
  def setCompute(compute: ExecutionContext, shutdown: () => Unit) = {
```

Or, `withCompute`? Not sure what's more idiomatic, but I think `set` is for mutable and `with` is for immutable.

Vasil makes a good point about default parameters. If we really want to make `shutdown` optional we can add an overload. I think we should make it required though, to force users to think about shutdown. It's easy to type `() => ()` anyway.

armanbilge: Oh, another thing to consider: what if they `setCompute` twice, should we `shutdown` the previous one? We can either not care (it's a strange thing to do) or we can only allow them to call `setCompute` at most once (if there's already a custom compute set, raise an error).

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
```diff
@@ -0,0 +1,128 @@
...
+
+  /**
+   * Override the defaul blocking execution context
+   *
+   * @param compute
+   *   execution context for blocking
+   * @param shutdown
+   *   method called upon blocking context shutdown
+   */
+  def overrideBlocking(blocking: ExecutionContext, shutdown: () => Unit = () => ()) = {
```
armanbilge: Same comments as above :)

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
```diff
@@ -0,0 +1,128 @@
...
+
+  /**
+   * Override the defaul scheduler
+   *
+   * @param compute
+   *   execution context for compute
+   * @param shutdown
+   *   method called upon compute context shutdown
+   */
+  def withScheduler(scheduler: Scheduler, shutdown: () => Unit = () => ()) = {
```
armanbilge: same comments here 😉

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
```diff
@@ -0,0 +1,128 @@
...
+   *
+   * @param shutdown
+   * @return
+   */
+  def addShutdownHook(shutdown: () => Unit) = {
+    extraShutdownHooks = shutdown :: extraShutdownHooks
+    this
+  }
+
+  def build: IORuntime =
```
armanbilge: I wonder if we should only allow to call this method once, so a builder can't be re-used.

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
majk-p: WDYT about doubling the methods for convenience like below?
```scala
  def overrideCompute(compute: ExecutionContext, shutdown: () => Unit = () => ()) = {
    customCompute = Some((compute, shutdown))
    this
  }
  def overrideCompute(compute: ExecutionContext ) = {
    this.overrideCompute(compute, () => ())
  }
```

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeBuilderPlatform.scala
```diff
@@ -0,0 +1,47 @@
...
+      customCompute.getOrElse(IORuntime.createDefaultComputeThreadPool(runtime))
+    val (blocking, blockingShutdown) =
+      customBlocking.getOrElse(IORuntime.createDefaultBlockingExecutionContext())
+    val (scheduler, schedulerShutdown) =
+      customScheduler.getOrElse(IORuntime.createDefaultScheduler())
+    val shutdown = () => {
+      computeShutdown()
+      blockingShutdown()
+      schedulerShutdown()
+      extraShutdownHooks.foreach(_())
```
armanbilge: I think we want to run `extraShutdownHooks` in reverse, because the list is built in reverse order that they are added.

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  /**
+   * Introduce additional shutdown hook to be executed after compute, blocking and scheduler
+   * shutdown logic is invoked
+   *
+   * @param shutdown
+   * @return
+   */
+  def addShutdownHook(shutdown: () => Unit) = {
```
armanbilge, line 126: Actually, I'm not sure what a realistic use-case for this method is, do you have one in mind? Seems strange for shutting down an `IORuntime` to also shutdown resources not related to it :)

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
armanbilge:
```suggestion
  def transformCompute(wrapper: ExecutionContext => ExecutionContext) = {
```

Maybe? Naming things is hard 😆

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  def build: IORuntime =
+    platformSpecificBuild
+}
+
+object IORuntimeBuilder {
+  def apply(): IORuntimeBuilder =
+    new IORuntimeBuilder()
+}
```
armanbilge, line 143: One more idea: maybe we should have a `Builder` method inside of the `IORuntime` object. Or even move all of this into a nested class.

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
majk-p: I can think of a such situation. Say I'm creating `InstrumentedIOApp` that uses custom `IORuntime`. I would use `IORuntimeBuilder().withComputeWrapper(addInstrumentation)` and then I might want to go with `addShutodwnHook` to make sure that all instrumentation modules are stopped. I guess this might also be helpful when debugging.
Do you feel like we should drop it?

armanbilge: Aha, thanks for explaining. That makes sense to me 👍 let's keep it.

(One alternative would be for to collect the shutdown hook in `withComputeWrapper`, like we do for `setCompute` etc., but that seems more confusing to me.)

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
majk-p: I've renamed all the methods that do override to use the `set` prefix. Allowing to set compute/blocking/shutdown and build just once sounds great. Just to make sure I got you right, do you think we should throw some runtime exception?
Sounds most straightforward, but I wonder if it's worth it to prevent on type-level somehow? or perhaps it would just be an overengineering?

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
majk-p: Done ;)

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
majk-p: Done ;)

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeBuilderPlatform.scala
majk-p: Good catch, thank you! done :+1:

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
majk-p: Sounds reasonable. Do you think this should be just based on some mutable bool state or do you have something more sophisticated in mind?

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
majk-p: I've added the `builder` to `IORuntime`, although I'd prefer to keep `IORuntimeBuiler` separate for clarity, it's kinda large due to the number of mutation methods. I find the relation with `IORuntimeBuilderPlatform` more straightforward this way.

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
armanbilge: Nah, a mutable boolean is fine :)

```suggestion
  def build(): IORuntime =
```

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
armanbilge: Typelevel stuff is cool but probably too over-engineered for this 😆 I was just thinking a runtime exception.

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
armanbilge: Thanks! Agree about organization; anyway we can decide this in a final bikeshedding :)

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
majk-p: Done

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
majk-p: Much better :wink:

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
majk-p: Done :wink:

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
```diff
@@ -0,0 +1,143 @@
...
+ */
+final class IORuntimeBuilder protected (
+    protected var customCompute: Option[(ExecutionContext, () => Unit)] = None,
+    protected var computeTransform: ExecutionContext => ExecutionContext = identity,
+    protected var customBlocking: Option[(ExecutionContext, () => Unit)] = None,
+    protected var blockingTransform: ExecutionContext => ExecutionContext = identity,
+    protected var customConfig: Option[IORuntimeConfig] = None,
+    protected var customScheduler: Option[(Scheduler, () => Unit)] = None,
+    protected var extraShutdownHooks: List[() => Unit] = Nil,
+    protected var builderExecuted: Boolean = false
```
armanbilge, line 34: I wonder if these should be volatiles or atomics or something, oh the joys of mutable code 😆

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
```diff
@@ -0,0 +1,143 @@
...
+) extends IORuntimeBuilderPlatform {
+
+  /**
+   * Override the defaul computing execution context
```
armanbilge:
```suggestion
   * Set a custom compute pool
```

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
```diff
@@ -0,0 +1,143 @@
...
+   *
+   * @param compute
+   *   execution context for compute
```
armanbilge:
```suggestion
   *   [[ExecutionContext]] for compute
```

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
```diff
@@ -0,0 +1,143 @@
...
+   * @param shutdown
+   *   method called upon compute context shutdown
```
armanbilge:
```suggestion
   *   shutdown hook upon [[IORuntime]] shutdown
```

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
```diff
@@ -0,0 +1,143 @@
...
+   */
+  def setCompute(compute: ExecutionContext, shutdown: () => Unit) = {
+    if (customCompute.isDefined) {
+      throw new RuntimeException("Compute can only be set once")
```
armanbilge:
```suggestion
      throw new IllegalStateException("Compute can be set only once")
```

armanbilge (commented):
Just some nitpicks, thanks for adding scaladocs for everything :)

Let's give the other maintainers an opportunity to take a look as well.

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
majk-p: Seems this has actually broken the build, it fails with
>Could not find any member to link

Looks like [fixing this would require a workaround](https://stackoverflow.com/a/23177275/2040487) in build.sbt. Should we go this way or just unlink it?

armanbilge: Woops, sorry. `` `ExecutionContext` `` is fine 👍

majk-p: Fixed

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
majk-p: Do you think it's a real case to use builder across multiple threads or something? I'd say it's overengineering at this point :wink:

armanbilge: I have no idea what people will do :P mostly it feels strange to write public APIs that are not threadsafe (usually we get this for free with immutability). We can leave this for now, it's just a minor thing.

core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
```diff
@@ -62,4 +62,7 @@ object IORuntime extends IORuntimeCompanionPlatform {
       shutdown: () => Unit,
       config: IORuntimeConfig): IORuntime =
     new IORuntime(compute, blocking, scheduler, FiberMonitor(compute), shutdown, config)
+
+  def builder: IORuntimeBuilder =
```
djspiewak:
```suggestion
  def builder(): IORuntimeBuilder =
```

Yay, mutable code!

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
```diff
@@ -0,0 +1,143 @@
...
+   *
+   * @param transform
+   */
+  def transformCompute(transform: ExecutionContext => ExecutionContext) = {
+    computeTransform = transform.andThen(computeTransform)
+    this
+  }
+
+  /**
+   * Override the defaul blocking execution context
```
djspiewak:
```suggestion
   * Override the default blocking execution context
```

djspiewak (changes requested):
I finally got to this! Sorry for the delay.

Couple notes, and one bit of bikeshedding: I think we should use `with` rather than `set` for the main build functions. Beyond that, I think this is good. In general, I'd rather see people *not* building custom `IORuntime`s on the regular, but when you need to do it, this is much much nicer than what was previously necessary.

majk-p:
Thanks for the feedback @djspiewak! I've made the changes, just not sure why the `headerCheck` would fail, I've launched `prePR` prior to this last commit :thinking:

kubukoz:
Needs a merge from series/3.x, then headers can be regenerated :)

djspiewak (approved):

PR #2659, opened 2021-12-17
Add fromCompletionStage to Async / IO

etspaceman:
CompletionStage is a more generic instance of CompletableFuture, and it can be applied to a number of things. An example can be the [RFuture](https://www.javadoc.io/doc/org.redisson/redisson/3.2.0/org/redisson/api/RFuture.html) from the Redisson library. 

This PR implements a `fromCompletionStage` function on Async and IO, which simply converts the CompletionStage to a completable future and runs `fromCompletableFuture`.

I tried implementing this the other way (move the core logic of `fromCompletableFuture` to `fromCompletionStage`, and simply have `fromCompletableFuture` run `fromCompletionStage` with a CompletableFuture value), but `CompletionStage` does not have a `cancel` method, so I was unable to replicate what was there. 

LMK any thoughts here. Happy to adjust as needed

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
@@ -19,8 +19,7 @@ package cats.effect
 import cats.effect.tracing.Tracing
 
 import java.time.Instant
-import java.util.concurrent.CompletableFuture
-import java.util.concurrent.CompletionStage
+import java.util.concurrent.{CompletableFuture, CompletionStage}
 
 private[effect] abstract class IOCompanionPlatform { this: IO.type =>
 
--- a/core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
@@ -20,6 +20,7 @@ import cats.effect.tracing.Tracing
 
 import java.time.Instant
 import java.util.concurrent.CompletableFuture
+import java.util.concurrent.CompletionStage
 
 private[effect] abstract class IOCompanionPlatform { this: IO.type =>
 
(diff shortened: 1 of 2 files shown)


armanbilge (dismissed):
Seems straightforward to me :)

core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
```diff
@@ -20,6 +20,7 @@ import cats.effect.tracing.Tracing
 
 import java.time.Instant
 import java.util.concurrent.CompletableFuture
+import java.util.concurrent.CompletionStage
```
vasilmkd: Nitpicking, can you please merge these imports into one?

`import java.util.concurrent.{CompletableFuture, CompletionStage}`.

Same in `AsyncPlatform.scala`.

vasilmkd:
Are there any tests for `fromCompletableFuture` btw? I think it would be nice to add some.

etspaceman:
We do already have tests for `fromCompletableFuture` in `IOPlatformSpecification`

core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
armanbilge: If we care about such things, I highly recommend we add this scalafix:
https://github.com/liancheng/scalafix-organize-imports

etspaceman: I use this myself and very much enjoy it :D

armanbilge: @etspaceman would you be able to set that up for us? :)

armanbilge (approved):

core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
etspaceman: Sure I could do that in a separate PR if you'd like

vasilmkd:
Approved but let's wait for @djspiewak to merge it. Releases and stuff...

vasilmkd:
Very nice addition. Thank you @etspaceman.

djspiewak (approved):

PR #2660, opened 2021-12-17
Scalafix Organize Imports

etspaceman:
Naturally a ton of file changes. The main ones that changes were made to manually were:

- .scalafix.conf
- build.sbt
- project/Common.scala
- project/CI.scala
- project/plugins.scala
- .github/workflows/ci.yml

I also had to make a few updates manually to imports to make things work properly. E.g.:

```scala
import org.scalatest.Prop
import Prop.forAll
```

This was not working properly because `import Prop._` would get sent to the top, which would then cause it to no longer be in scope. I updated these types of imports to use fully qualified imports, e.g. `import org.scalatest.Prop.forAll`.

The ordering of the imports is just copied from what I've done in my other projects. The ordering is fully debatable and can be changed to whatever the team here sees fit.


.scalafix.conf
```diff
@@ -0,0 +1,11 @@
+rules = [OrganizeImports]
+
+OrganizeImports {
+  groupedImports = Merge
+  groups = [
```
etspaceman, line 7: This organizes imports separated by new lines. E.g.:

```scala
import java.util.CompletableFuture

import org.openjdk.jmh.annotations._

import cats.Eq
```

vasilmkd:
Is it possible to configure this? I believe a lot of files in the codebase are doing local imports and dependency imports first, then imports from the `scala.*` standard library and then imports from the `java.*` standard library.

.scalafix.conf
etspaceman: @vasilmkd this is where the configuration is. We can change it up to whatever you see fit 😄

vasilmkd: Nice, I think going for `["*", "scala.", "java."]` should give the least diff.

etspaceman: Do you want the cats dependencies in their own grouping?

.github/workflows/ci.yml
```diff
@@ -138,6 +138,10 @@ jobs:
       - shell: bash
         run: sbt ++${{ matrix.scala }} '${{ matrix.ci }}'
 
+      - if: matrix.scala == '2.13.7'
+        shell: bash
+        run: sbt ++${{ matrix.scala }} scalafixAll --check
```
etspaceman: Scalafix semanticDb doesn't have support for Scala 3 and thus OrganizeImports wont work. So this will only run the check on 2.13.x

.scalafix.conf
vasilmkd: On the surface, no, but if it looks particularly nice, why not.

etspaceman: Where should those be? At the top or bottom?

etspaceman: Updated to include the sorting you provided. Let me know if you want a different sort

.scalafmt.conf
```diff
@@ -53,7 +53,7 @@ trailingCommas = never
 
 rewrite {
   // RedundantBraces honestly just doesn't work, otherwise I'd love to use it
-  rules = [PreferCurlyFors, RedundantParens, SortImports]
```
etspaceman, line 56: SortImports conflicts with the OrganizeImports so `scalafixAll --check` doesn't work properly in the CI. Removed it here.

.scalafix.conf
vasilmkd: "scala." in the middle of "*" and "java." will be a smaller diff.

vasilmkd: Please add a newline at the end of this file as well.

build.sbt
```diff
@@ -129,6 +129,11 @@ ThisBuild / githubWorkflowBuildPreamble ++= Seq(
 )
 
 ThisBuild / githubWorkflowBuild := Seq(
+  WorkflowStep.Sbt(
+    List("root/scalafixAll --check"),
+    name = Some("Check that scalafix has been run"),
+    cond = Some(s"matrix.scala != '$Scala3' && matrix.os != '$Windows'")
```
etspaceman, line 134: Seem to have a weird exception in the CI on windows machines here. See https://github.com/typelevel/cats-effect/runs/4564105278?check_suite_focus=true

Might be worth opening a bug w/ Scalafix

etspaceman: Also Scala3 would require us to disable the `removeUnused` imports that this plugin offers. After talking w/ @vasilmkd and @armanbilge we decided to disable this check in the Scala 3 CIs.

etspaceman: Created an issue: https://github.com/scalacenter/scalafix/issues/1510

.scalafix.conf
```diff
@@ -0,0 +1,6 @@
+rules = [OrganizeImports]
+
+OrganizeImports {
+  groupedImports = Merge
+  groups = ["cats.", "*", "scala.", "java."]
```
djspiewak: Quick note: I would generally remove the leading `"cats."` here. It's something of a coincidence that those imports tend to come first, stemming from the fact that `ca` is alphabetically ordered before almost any other plausible top level package names. Beyond that, this looks like what I would (subjectively!) want!

djspiewak (changes requested):
This is great and thank you for taking this on! One quick bikeshed nit and it's ready for merge.

djspiewak:
Oh actually, additional idea: can you add this to the `prePR` alias? That alias already runs scalafmt, so it would be helpful if it could also run the import organization.

.scalafix.conf
etspaceman: Yep, can do. I'll do that when I get a chance.

build.sbt
```diff
@@ -241,7 +246,9 @@ addCommandAlias(CI.JS.command, CI.JS.toString)
 addCommandAlias(CI.Firefox.command, CI.Firefox.toString)
 addCommandAlias(CI.Chrome.command, CI.Chrome.toString)
 
-addCommandAlias("prePR", "; root/clean; scalafmtSbt; +root/scalafmtAll; +root/headerCreate")
+addCommandAlias(
+  "prePR",
+  "; root/clean; root/scalafixAll; scalafmtSbt; +root/scalafmtAll; +root/headerCreate")
```
etspaceman, line 250: @djspiewak The scalafix command was added to prePR.

.scalafix.conf
etspaceman: This is done

djspiewak (changes requested):
Ah! I misunderstood how this config works. Also scanned through and I'm going to bikeshed some more…

- Is there a way to make it group by top-level package? It seems that `*` just lumps everything together, as opposed to splitting apart (e.g.) `cats` from `org`
- We should add `javax.` to the end of the list. I forgot about that one!

The ideal would be something like this:

```scala
import cats.{Functor, Monad}
import cats.effect.IO
import cats.effect.kernel.Resource

import org.thingy.Wut

import sun.misc.Unsafe

import scala.collection.immutable

import java.util.Collection

import javax.swing.JFrame
```

etspaceman:
We could do that as long as we are explicit about it. E.g. "org.", "com.", etc. The rule also allows for regex groups. fwiw. 

We can also control the newlines, so we can have pretty fine tuned control. 

https://github.com/liancheng/scalafix-organize-imports#221-value-type

Feel free to play around with the PR to test out different configurations. Fair warning, when changing the config locally, I've often had to run it 2x for the compile to work properly. Not sure why. But once it's run successfully on all files, the rule works seamlessly going forward.

etspaceman:
@djspiewak I tried to capture all of the top-level packages explicitly in the config. LMK what you think.

.scalafix.conf
```diff
@@ -0,0 +1,15 @@
+rules = [OrganizeImports]
+
+OrganizeImports {
+  groupedImports = Merge
+  importsOrder = SymbolsFirst
```
etspaceman, line 5: SymbolsFirst matches what scalafmt was doing. The default for this plugin is Ascii, which is why some of the sorting changed on the initial runs.

.scalafmt.conf
etspaceman: Technically we should be able to add this back but I'm going to keep it out because we are now controlling this through scalafix

.scalafix.conf
djspiewak: Odd. Scanning through the diff, it seems that symbols are still being sorted last, not first. Am I missing something?

etspaceman: I was a bit confused too. There's another option that I think I have to set to be the same: `importSelectorsOrder`. I'll set that here and push a change.

djspiewak: That looks much better!

djspiewak (approved):

PR #2661, opened 2021-12-17
Register the live fiber snapshot trigger mbean on IORuntime construction

vasilmkd:
It should be done for every constructed `IORuntime`, not just for the global one.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
+++ b/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
@@ -60,6 +60,9 @@ object IORuntime extends IORuntimeCompanionPlatform {
       blocking: ExecutionContext,
       scheduler: Scheduler,
       shutdown: () => Unit,
-      config: IORuntimeConfig): IORuntime =
-    new IORuntime(compute, blocking, scheduler, FiberMonitor(compute), shutdown, config)
+      config: IORuntimeConfig): IORuntime = {
+    val fiberMonitor = FiberMonitor(compute)
+    registerFiberMonitorMBean(fiberMonitor)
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
```diff
@@ -60,6 +60,9 @@ object IORuntime extends IORuntimeCompanionPlatform {
       blocking: ExecutionContext,
       scheduler: Scheduler,
       shutdown: () => Unit,
-      config: IORuntimeConfig): IORuntime =
-    new IORuntime(compute, blocking, scheduler, FiberMonitor(compute), shutdown, config)
+      config: IORuntimeConfig): IORuntime = {
+    val fiberMonitor = FiberMonitor(compute)
+    registerFiberMonitorMBean(fiberMonitor)
+    new IORuntime(compute, blocking, scheduler, fiberMonitor, shutdown, config)
```
armanbilge: The `register` function returns an `unregister` hook right, should that be tagged onto the `shutdown`?

vasilmkd: Good catch, thanks, I was in a hurry yesterday.

armanbilge (approved):

vasilmkd:
Wait, please don't merge this, it seems like there's some flakiness.

armanbilge (changes requested):
Until the flakiness is investigated

vasilmkd:
Moving the registration to `IORuntime.apply` clashes with the test runtime, where a new instance is created for every spec. I know this can be worked around but I'd rather not do it at all. It's also becoming clear to me that in light of #2663, I was right to be cautious of giving metrics to other thread pools.

Issue #2637, opened 2021-12-09
Queue: take for more than one record


etspaceman:
FS2 Queue's had methods that would allow you to take a chunk of records, which were really convenient. The behavior would be that we would run the `tryTake` method until N was hit, or None was returned. I would like to see the same in the Cats Effect queue. Something like this:

```scala
def tryTakeChunk(
        maxN: Int
      )(implicit A: Async[F]
      ): F[List[A]] = {
      A.tailRecM[(List[A], Int), List[A]](
        (List.empty, 0)
      ) { case (list, i) =>
        if (list.size === maxN) list.reverse.asRight.pure[F]
        else {
          queue.tryTake.map {
            case None => Right(list.reverse)
            case Some(x) => Left((list.prepend(x), i + 1))
          }
        }
      }
    }
```

armanbilge:
Related?
- https://github.com/typelevel/cats-effect/pull/2164

etspaceman:
It was related, yes, though that PR was messy and not well thought through. 😅

armanbilge:
Seems it might be worth a take 2 then? :)

etspaceman:
Absolutely! I opened this issue after discussing w/ @djspiewak offline in a related matter, just to track the effort.

armanbilge:
Hmm, looking at back that PR it seemed there was some overlapping scope with fs2? I think I proposed an fs2-solution for this in https://github.com/typelevel/cats-effect/pull/2164#issuecomment-903209604.

etspaceman:
Maybe!

etspaceman:
After thinking about this more, I think this should live in cats-effect. A user shouldn't have to make an FS2 Stream to get more than 1 record from a queue.

vasilmkd:
I propose that we break compatibility here and do this properly, if possible.

armanbilge:
I don't have enough experience to know when it's ok to break bincompat, so I err on conservatism especially as CE3 gets more established in the ecosystem. So 👎 but happy to be overriden by someone who knows better :)

OTOH, can we implement these as implicit syntax?

armanbilge:
Also linking to https://github.com/typelevel/cats-effect/issues/2639 which would need some breaking changes I think.

vasilmkd:
Yeah, #2639 cannot be done without breakage, unfortunately.

armanbilge:
Well, you could "cheat" by taking something out and putting it back to implement peek.

[...]

PR #2670, opened 2021-12-19
Add batch offer/take methods to Queue and Dequeue

etspaceman:
Solves #2637

TODO
- [x] Add tests


armanbilge:
👍 to the methods, but my 2 cents is we should add them as syntax extensions so we don't have to break binary compatibility.

etspaceman:
I can certainly do that. @djspiewak WDYT?

djspiewak:
Actually I think we can keep them on the main trait, we just have to define the default (naive) implementations there. We can't add new *abstract* members, but we can add new concrete ones.

armanbilge:
Oh, is the "naive" implementation just `N = 1`? That works.

armanbilge:
Except, I don't think that's possible. Because we can't get to a `List` without `map`ping, but we don't have `Functor` unless I'm missing something.

etspaceman:
The implementation i designed requires a Monad for tailRecM, which i wouldn't have on the trait.

etspaceman:
I have moved the methods to a syntax area per this discussion

std/shared/src/main/scala/cats/effect/std/syntax/package.scala
```diff
@@ -19,7 +19,9 @@ package cats.effect.std
 package object syntax {
 
   object all extends AllSyntax
-
   object supervisor extends SupervisorSyntax
+  object backpressure extends BackpressureSyntax
```
etspaceman: I noticed BackpressureSyntax didn't have an associated object so I created one here.

armanbilge:
Sorry, after looking at your changes I just realized what Daniel may have been getting at. We _can_ actually define the methods on the trait by requiring the additional constraints, exactly as you've done for the syntax extensions:
```scala
  def tryTakeBackN(maxN: Option[Int])(implicit F: Monad[F]): F[Option[List[A]]] = 
    _tryTakeN(maxN)(wrapped.tryTakeBack)
```

This is probably nicer 🙃

etspaceman:
Ok, can do

etspaceman:
Alright I think this is at a pretty good point. LMK what you think @djspiewak @armanbilge

etspaceman:
Per a conversation w/ @djspiewak I added tryOfferN/tryOfferFrontN/tryOfferBackN methods as well.

std/shared/src/main/scala/cats/effect/std/Dequeue.scala
```diff
@@ -373,6 +442,20 @@ trait DequeueSink[F[_], A] extends QueueSink[F, A] {
    */
   def tryOfferFront(a: A): F[Boolean]
 
+  /**
+   * Attempts to enqueue the given elements at the front of the queue without semantically
+   * blocking. If an item in the list cannot be enqueued, the remaining elements will be
+   * returned. This is a convenience method that recursively runs `tryOffer` and does not offer
+   * any additional performatnce benefits.
```
djspiewak:
```suggestion
   * any additional performance benefits.
```

There are a couple of these :-)

djspiewak (changes requested):
Minor issue and then good to go. The inconsistency in the constraints on the methods is unfortunate but unavoidable, and it's better than a `Queue2` or something equally hideous.

std/shared/src/main/scala/cats/effect/std/Dequeue.scala
etspaceman: Fixed

djspiewak (approved):

etspaceman:
CI seems to have failed due to an unrelated test, undeterministically:

```
[error]   x exit on fatal error with other unsafe runs
[error]    java.lang.InterruptedException
[error]    	at java.util.concurrent.locks.AbstractQueuedSynchronizer.tryAcquireSharedNanos(AbstractQueuedSynchronizer.java:1326)
[error]    	at java.util.concurrent.CountDownLatch.await(CountDownLatch.java:277)
[error]    	at cats.effect.IOPlatform.$anonfun$unsafeRunTimed$2(IOPlatform.scala:81)
[error]    	at scala.runtime.java8.JFunction0$mcZ$sp.apply(JFunction0$mcZ$sp.scala:17)
[error]    	at cats.effect.unsafe.WorkerThread.blockOn(WorkerThread.scala:572)
[error]    	at scala.concurrent.package$.blocking(package.scala:124)
[error]    	at cats.effect.IOPlatform.unsafeRunTimed(IOPlatform.scala:81)
[error]    	at cats.effect.IOPlatform.unsafeRunSync(IOPlatform.scala:42)
[error]    	at catseffect.examples.FatalErrorUnsafeRun$.$anonfun$run$3(examplesplatform.scala:42)
[error]    	at cats.effect.unsafe.WorkerThread.blockOn(WorkerThread.scala:607)
[error]    	at scala.concurrent.package$.blocking(package.scala:124)
[error]    	at cats.effect.IOFiber.runLoop(IOFiber.scala:948)
[error]    	at cats.effect.IOFiber.execR(IOFiber.scala:1293)
[error]    	at cats.effect.IOFiber.run(IOFiber.scala:123)
[error]    	at cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:530) doesn't contain 'Boom!' (IOAppSpec.scala:217)
```

djspiewak:
Oh this is fun. I think this is a race condition with the interruption handling in the worker threads.

Edit: #2720

PR #2682, opened 2021-12-23
Use the done signal when parking worker threads

vasilmkd:
A little detail I missed during the implementation of #2600.


vasilmkd:
I will include this in another PR.

PR #2683, opened 2021-12-23
Add blocking benchmarks

vasilmkd:
I will be playing with these in the near future.


vasilmkd:
I will include this in another PR.

PR #2685, opened 2021-12-23
Make `IORuntime#blocking` public

armanbilge:
I understand there is/was some contention around this, so apologies for poking the bear and feel free to close.

My assumptions are:
1. You can achieve the same thing by manually creating a default `IORuntime` using `IORuntime.createDefaultBlockingExecutionContext` and keeping a reference to it.
2. There are APIs which will ask for a blocking compute pool e.g. to run a `Future`.

Assuming the above are true, hiding it just makes things unnecessarily difficult when there is a reasonable use-case for it to be public.

If the concern is about leaking this to JS:
1. The _requirement_ for a blocking EC is already in the JS `IORuntime` constructor.
2. Various blocking-related methods are already leaked to JS one way or another.
3. Of course, this can be worked around, if it is truly a concern. In which case, I'd argue we should fix it everywhere, but that's a separate discussion.

An aside, but it also occurred to me that while there is `Async#executionContext`, AFAICT there is no way to get access to the blocking EC from inside of `F`. Which for situations like the above, where you are interacting with a `Future`-based API that wants a blocking `ExecutionContext` is unfortunate.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
+++ b/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
@@ -34,7 +34,7 @@ avoid unintentionally degrading your application performance.
 """)
 final class IORuntime private[unsafe] (
     val compute: ExecutionContext,
-    private[effect] val blocking: ExecutionContext,
+    val blocking: ExecutionContext,
     val scheduler: Scheduler,
     private[effect] val fiberMonitor: FiberMonitor,
     val shutdown: () => Unit,


djspiewak:
IMO this isn't the direction we should go, particularly in light of #2687. The reason the blocking EC is private is two-fold:

- The goal was always to find a way to *not* have it in the future (the future is now!), and making it public would severely constrain such efforts
- There's almost no penalty to having multiple blocking pools, since each is just a cached unbounded thread pool. So anyone who *needs* a blocking thread pool as an `ExecutionContext` can always just make one of their own

armanbilge:
> The goal was always to find a way to not have it in the future (the future is now!), and making it public would severely constrain such efforts

Even with https://github.com/typelevel/cats-effect/pull/2687 we still have a blocking EC, it just happens to be the same as the compute EC. We can create a wrapper it so that runnables submitted to the wrapper EC are properly submitted as blocking to the underlying EC.

> There's almost no penalty to having multiple blocking pools, since each is just a cached unbounded thread pool. So anyone who needs a blocking thread pool as an `ExecutionContext` can always just make one of their own.

Fair enough. IMHO this is a batteries-included sort of thing—creating your own is less a performance penalty, but more an engineering one (how should I make one? where do I put it? how should I manage its lifecycle?) when there's a perfectly good one inside the runtime system I already bought into. But, just my two cents.

rossabaker:
If it's about engineering cost, I'd consider a convenience def that could be deprecated later with nice names and such, rather than a public member of a fundamental construct.  But even "something new to deprecate later" makes my argument wobbly.  Slight agreement with Daniel.

djspiewak:
> But even "something new to deprecate later" makes my argument wobbly

Also consider the fact that it would be "something new to deprecate immediately", since @vasilmkd's PR is already ready. :-)

armanbilge:
I still don't understand why there is something to deprecate—unless there is literally no more blocking happening.

djspiewak:
Looping back to this…

The problem with exposing a `IORuntime#blocking` which just wraps things in an `IO.blocking` and feeds it to the WSTP is that it's a *lot* less efficient for such purely-external things than a plain `cachedThreadPool` would be. So really, users would be better off making that for themselves.

Really what we're asking about here is whether or not Cats Effect should provide some trivial little utility which constructs a cached thread pool. Is that worthy of **std**? I'm not sure. Discussion welcome. But I think doing it via `IORuntime` would be the wrong way to go about it.

armanbilge:
> Really what we're asking about here is whether or not Cats Effect should provide some trivial little utility which constructs a cached thread pool.

Well, we already have `IORuntime.createDefaultBlockingExecutionContext` which does this. I don't think it's worth including something in **std**.

> The problem with exposing a `IORuntime#blocking` which just wraps things in an `IO.blocking` and feeds it to the WSTP is that it's a lot less efficient for such purely-external things than a plain `cachedThreadPool` would be.

Thanks for the additional explanation. Ok, so IIUC:
1. Wrapping things in an `IO` before submitting them to a threadpool is not efficient. (Or is this specifically bad for `IO.blocking`?)
2. We expose the WSTP as an `ExecutionContext` in this manner, but the trade-off makes sense there b/c having more than one compute pool in the application hurts performance much worse.
3. B/c the same penalty doesn't apply to multiple blocking pools, the trade-off doesn't make sense in that case.

In fact, since the WSTP _is_ exposed and is capable of handling the `scala.concurrent.blocking` thing, users could make this so-called blocking EC wrapper themselves. However, can is not should, and this falls under the category of something we don't want to encourage. Makes sense to me.

PR #2699, opened 2021-12-29
Removed `interruptible` dependency on an explicit `ExecutionContext`

djspiewak:
This does result in *slightly* more overhead, but it's a lot more composable and provides compatibility with #2687 (and all future revisions in this area).


vasilmkd:
When this eventually runs on the WSTP it will have less overhead because `start` should be faster than `EC#execute` (internal fork vs external enqueue).

vasilmkd (approved):

PR #2703, opened 2021-12-30
Update site for 3.3.2

djspiewak:


vasilmkd:
It seems the build still contains the old release. Isn't it early for the new one?

djspiewak:
I try to just update both as soon as I can, rather than waiting to do it in a particular order. In theory, Dependabot should reconcile things within a day or two.

vasilmkd:
What I actually meant is that the scastie build the new link points to still contains `3.3.1` and not `3.3.2`. The URL needs to be regenerated for the new build.

armanbilge:
Updating this link every release seems annoying :) if you're willing to eat some JavaScript there might be a way to dynamically generate the link just from the version number.

Here's a snippet from [curly4s](https://davenverse.github.io/curly4s/) that calls the scastie API.

```scala
def setOutput(code: String): IO[Unit] = IO {
    js.Dynamic.global.scastie.Embedded(
      "#output",
      js.Dynamic.literal(
        code = "\n" + code,
        sbtConfig = """libraryDependencies += "org.http4s" %% "http4s-core" % "0.23.6"""",
        scalaVersion = "2.13.7",
        targetType = "jvm"
      )
    )
  }
```

armanbilge:
So I looked into this a bit. It seems if you make a POST to the endpoint https://scastie.scala-lang.org/api/save with a JSON body like this:
```json
{
  "_isWorksheetMode" : true,
  "code" : "\n{\n  import _root_.org.http4s.{Request, Method, Uri, Headers}\n  import _root_.fs2.{Stream, Pure}\n  Request[Pure](\n    method = Method.fromString(\"GET\").fold(throw _, identity),\n    uri = Uri.unsafeFromString(\"https://www.icanhazdadjoke.com\"),\n    headers = Headers((\"Accept\",\"application/json\")),\n    body = Stream()\n  )\n}",
  "target" : {
    "scalaVersion" : "2.13.7",
    "tpe" : "Jvm"
  },
  "libraries" : [  ],
  "librariesFromList" : [  ],
  "sbtConfigExtra" : "libraryDependencies += \"org.http4s\" %% \"http4s-core\" % \"0.23.6\"",
  "sbtPluginsConfigExtra" : "",
  "isShowingInUserProfile" : false
}
```

Then the response is an ID for a scastie:
```json
{
  "base64UUID" : "gyolUrkORBSRn5DPYVrHNw"
}
```

So, I'd say we can completely automate this :)

armanbilge:
And here's the config for the CE3 scastie:
```json
{
    "_isWorksheetMode": false,
    "code": "import cats.effect.{IO, IOApp}\nimport cats.effect.std.Random\n\nimport scala.concurrent.duration._\n\nobject Hello extends IOApp.Simple {\n  \n  def sleepPrint(word: String, name: String, rand: Random[IO]) =\n    for {\n      delay <- rand.betweenInt(200, 700)\n      _     <- IO.sleep(delay.millis)\n      _     <- IO.println(s\"$word, $name\")\n    } yield ()\n  \n  val run =\n    for {\n      rand <- Random.scalaUtilRandom[IO]\n      \n      // try uncommenting first one locally! Scastie doesn't like System.in\n      // name <- IO.print(\"Enter your name: \") >> IO.readLine\n      name <- IO.pure(\"Daniel\")\n\n      english <- sleepPrint(\"Hello\", name, rand).foreverM.start\n      french  <- sleepPrint(\"Bonjour\", name, rand).foreverM.start\n      spanish <- sleepPrint(\"Hola\", name, rand).foreverM.start\n\n      _ <- IO.sleep(5.seconds)\n      _ <- english.cancel >> french.cancel >> spanish.cancel\n    } yield ()\n}\n",
    "isShowingInUserProfile": false,
    "libraries": [
        {
            "artifact": "cats-effect",
            "groupId": "org.typelevel",
            "target": {
                "scalaVersion": "2.13.6",
                "tpe": "Jvm"
            },
            "version": "3.3.2"
        }
    ],
    "librariesFromList": [],
    "sbtConfigExtra": "scalacOptions ++= Seq(\n  \"-deprecation\",\n  \"-encoding\", \"UTF-8\",\n  \"-feature\",\n  \"-unchecked\"\n)\n\nturbo := false",
    "sbtPluginsConfigExtra": "",
    "target": {
        "scalaVersion": "2.13.7",
        "tpe": "Jvm"
    }
}
```

So, the only remaining question is where to put it 🤔

djspiewak:
> So, the only remaining question is where to put it

What if we have a custom GHA workflow which triggers off of *tag* push? We only ever push the tags when a release is made.

armanbilge:
Eh, I don't think the issue is how frequently it runs, but rather _where_ it runs and how the url gets populated the scastie id. Two ideas:
1. Clicking this button on the site calls to the scastie API, gets the ID, and then opens a new page there. This is pretty easy. We'd still have to update the CE version number each release with PRs like this.
2. Put the CE version and scastie id into the site config and reference them from here. Then, we need a mechanism for updating the config from a workflow with this information.

PR #2705, opened 2021-12-31
Detect sbt thread cleaner and terminate workers

### LINKED ISSUE #2704: version 3.3.1 doesn't properly terminate on error in SBT
build.sbt
```
ThisBuild / scalaVersion := "2.13.7"
ThisBuild / organization := "org.example"

val commonSettings = Seq(
  version := "0.0.1",
  scalacOptions ++= Seq("-deprecation", "-Xfatal-warnings")
)

val root = (project in file("."))
  .settings(commonSettings)
  .settings(
    name := "ds",
    libraryDependencies += "org.typelevel" %% "cats-effect" % "3.3.1"
  )
```

Main.scala
```scala

  def run: IO[Unit] =
    IO.raiseError(new RuntimeException)
}
```

```
% sbt
[info] welcome to sbt 1.5.5 (Oracle Corporation Java 13.0.1)
> run
```

Expected behavior: IOApp completes with error, sbt is usable
Actual behavior: IOApp completes with error, sbt starts consuming all available CPU(up to 800% on my PC)

After I change `ce3` version to `3.3.0` the problem disappears.

djspiewak:
Also added a non-fatal test which *doesn't* reproduce the issue. I wasn't able to come up with a good way of rigorously testing this.

Fixes #2704

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
@@ -287,7 +287,7 @@ private final class WorkerThread(
 
         // the only way we can be interrupted here is if it happened *externally* (probably sbt)
         if (isInterrupted())
-          done.set(true)
+          pool.shutdown()
         else
           // Spurious wakeup check.
           cont = parked.get()
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
@@ -285,8 +285,12 @@ private final class WorkerThread(
         // Park the thread until further notice.
         LockSupport.park(pool)
 
-        // Spurious wakeup check.
-        cont = parked.get()
+        // the only way we can be interrupted here is if it happened *externally* (probably sbt)
+        if (isInterrupted())
+          done.set(true)
+        else
+          // Spurious wakeup check.
+          cont = parked.get()
       }
     }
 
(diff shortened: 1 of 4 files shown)


armanbilge:
It's super annoying, but you have to register the new example app here for JS:
https://github.com/typelevel/cats-effect/blob/63fe08c66546caf0615a30e75a08c37dd8982ac5/tests/js/src/main/scala/catseffect/examplesplatform.scala#L31-L44

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -285,8 +285,12 @@ private final class WorkerThread(
         // Park the thread until further notice.
         LockSupport.park(pool)
 
-        // Spurious wakeup check.
-        cont = parked.get()
+        // the only way we can be interrupted here is if it happened *externally* (probably sbt)
+        if (isInterrupted())
+          done.set(true)
```
vasilmkd: This clashes with the `shutdown()` method on `WorkStealingThreadPool`. I wonder if that method should be called instead, there's already a CAS which will make sure to call it once. Only setting `done` won't clean up all resources.

vasilmkd: The problem with all this is that I don't know if the rest of the thread pools in the same `IORuntime` are shut down. Technically, `IORuntime#shutdown()` can also be called, since the `WSTP` has a reference to its `IORuntime`.

djspiewak: I can swap it to call shutdown(). I'm actually not as worried about the other pools in the runtime because this is such an odd scenario. It'a kind of like being shut down from the inside out, and we can assume it's happening to everyone and not just us.

vasilmkd (approved):
Thanks for taking this on.

Issue #2630, opened 2021-12-05
`IOApp` and leaked fibers


armanbilge:
While testing https://github.com/http4s/http4s/pull/5657 on Node.js we encountered a strange "bug": the `IOApp` would terminate, but the Node.js process _wouldn't_. The same program terminated normally on JVM.

[With the help of fiber dumps](https://gist.github.com/armanbilge/496dbe8e0ce7d77466f5429c500175fd) 😁 we were able to identify some leaked fibers. The app wasn't _blocked_ per-say (`IOApp#run` had already terminated and set the exit code) but Node.js by default won't terminate the process until the event loop queue is empty (including scheduled events). So as long as those fibers were alive so would be the process.

So overall this was a net win, but this behavior was confusing. Also I'm trying really hard to have identical semantics across JVM/Node.js as much as possible (browsers are another story of course).

Some ideas:
1. Leave as-is. These are different platforms after all. Plus I can argue that cross-building for Node.js helps find bugs 😉 
2. Force exit on Node.js after `IOApp#run` terminates, leaks be damned.
3. Try and match the behavior on JVM? This seems hard in general and is not a good UX anyway.
4. Hard-exit on both platforms, but maybe issue a best-effort warning about leaked fibers.

armanbilge:
Some discord discussion: https://discord.com/channels/632277896739946517/839254871646404638/917150285179289650

Here's an idea, albeit opinionated: if `IOApp` gets to normal exit and there are still live fibers in the fiberbag, to do a fiber dump before exiting to complain about these leaks. This would work best with https://github.com/typelevel/cats-effect/issues/2628, and of course should be suppressible via an option for those who don't care :)

This is particularly helpful for Node.js: when the app terminates normally but there are leaked fibers preventing process termination, somehow fiber dumps stopped working. I couldn't quite understand why. So getting the dump for free would be a big help.

One last thought: just realized the connection to dispatcher, which cancels its fibers and backpressures on close.

PR #2707, opened 2021-12-31
Hard exit in Node.js `IOApp`

armanbilge:
This fixes https://github.com/typelevel/cats-effect/issues/2630 by enforcing a hard-exit. A graceful exit in Node.js waits for the event loop to become empty, which may not happen if there are leaked fibers.

I still think it would be good to issue a warning about leaked fibers, but this at least solves the confusing problem of the app completely hanging on termination.


djspiewak (approved):
I think this is the right thing to do. At the very least it matches the JVM semantics.

PR #2708, opened 2021-12-31
IOFiber - Auxiliary methods.

diesalbla:
Introduce a factory `spawnFiber` method, to create new IOFiber objects for the `Start` and `Race` cases. This factory uses the localState, execution context, and runtime, of this fiber.

From the `scheduleFiber` and `rescheduleFiber` methods we remove the "ec" parameter. It seems that the `currentCtx` is always used, and it is not modified from load to call.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -133,6 +133,9 @@ private final class IOFiber[A](
     }
   }
 
+  private[this] def spawnFiber(spawnIO: IO[_]): IOFiber[Any] =
+    new IOFiber[Any](localState, null, spawnIO, currentCtx, runtime)
+
   /* backing fields for `cancel` and `join` */
 
   /* this is swapped for an `IO.unit` when we complete */
@@ -152,10 +155,9 @@ private final class IOFiber[A](
           /* if we have async finalizers, runLoop may return early */
           IO.async_[Unit] { fin =>
             // println(s"${name}: canceller started at ${Thread.currentThread().getName} + ${suspended.get()}")
-            val ec = currentCtx
             resumeTag = AsyncContinueCanceledWithFinalizerR
             objectState.push(fin)
-            scheduleFiber(ec, this)
+            scheduleFiber(this)
           }
         } else {
           /*
@@ -224,8 +226,7 @@ private final class IOFiber[A](
       if (nextAutoCede <= 0) {
         resumeTag = AutoCedeR
         resumeIO = _cur0
-        val ec = currentCtx
-        rescheduleFiber(ec, this)
+        rescheduleFiber(this)
         return
       }
     }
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -132,6 +132,10 @@ private final class IOFiber[A](
     }
   }
 
+  @inline
```
djspiewak: This is ignored btw and doesn't need to be here

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -1257,6 +1232,9 @@ private final class IOFiber[A](
   }
 
   /* Implementations of resume methods */
+  @inline
```
djspiewak: Also unnecessary

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -1257,6 +1232,9 @@ private final class IOFiber[A](
...
+  private[this] def runtimeRun(io: IO[Any]): Unit =
```
djspiewak: Not to bikeshed too much, but I think this might be better-named something like `initRunLoop`

djspiewak (changes requested):

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -1229,12 +1204,12 @@ private final class IOFiber[A](
...
-      wstp.scheduleFiber(fiber)
-    } else {
-      scheduleOnForeignEC(ec, fiber)
+  private[this] def scheduleFiber(fiber: IOFiber[_]): Unit = {
+    val ec: ExecutionContext = currentCtx
+    ec match {
+      case wstp: WorkStealingThreadPool =>
+        wstp.scheduleFiber(fiber)
+      case _ => scheduleOnForeignEC(ec, fiber)
     }
```
vasilmkd: @djspiewak You made me explicitly switch to the isInstanceOf/asInstanceOf style once. Is that still desirable?

djspiewak: Oh, good catch! I missed that. Yes, I would prefer the explicit style here because we're less reliant on how pattern matching is encoded, and the JVM optimization is a bit fragile sometimes.

djspiewak:
```
[error] /home/runner/work/cats-effect/cats-effect/core/shared/src/main/scala/cats/effect/IOFiber.scala:630:23: local val ec in method loop is never used
62
[error]                   val ec = currentCtx
63
[error]                       ^
64
```

Warnings are fatal in CI and non-fatal locally. :-)

vasilmkd (dismissed):
Nice cleanup. Thanks for tackling this.

vasilmkd:
Can we target this at `series/3.3.x` instead? And the follow up PR.

diesalbla:
> Can we target this at series/3.3.x instead? And the follow up PR. 

Done.

PR #2710, opened 2022-01-01
Revert "Update sbt-scalajs, scalajs-compiler, ... to 1.8.0 in series/3.x"

djspiewak:
Reverts typelevel/cats-effect#2645


armanbilge (approved):
We can re-evaluate before CE 3.4.0 but a good rule of thumb might be to tie the SJS version to our current Scala 3 version. E.g. Scala 3.0 requires SJS 1.5+, Scala 3.1 requires SJS 1.7+, and we'll see what is needed in Scala 3.2.

armanbilge:
Looks like CI on series/3.x is broken until we merge in series/3.3.x I think.

PR #2714, opened 2022-01-01
Set the callback unsafely without memory barriers and allocations

vasilmkd:
`series/3.3.x`:
```scala
Benchmark                      (cpuTokens)  (size)   Mode  Cnt    Score   Error  Units
ParallelBenchmark.parTraverse        10000    1000  thrpt   20  313.308 ± 8.235  ops/s
```

`This PR`:
```scala
Benchmark                      (cpuTokens)  (size)   Mode  Cnt    Score   Error  Units
ParallelBenchmark.parTraverse        10000    1000  thrpt   20  320.792 ± 1.811  ops/s
```


djspiewak (approved):
Very clever! Nice catch

PR #2719, opened 2022-01-01
Attempt to detect host environments (like unforked sbt)

### LINKED ISSUE #2706: Errors raised in unforked `IOApp`s terminate sbt 1.6.x
Simple reproducer:

```scala
object Boom extends IOApp.Simple {
  val run = IO.raiseError(new RuntimeException)
}
```

When `run` unforked in sbt 1.6.0, this will print the error and then terminate the sbt runtime. I believe this is happening because sbt 1.6.0 isn't suppressing `System.exit` in the same way that previous versions did, which results in VM termination. We can resolve this by using `throw` instead of `printStackTrace()` etc, but that has other problems for exit codes on worker interruption.

djspiewak:
Rather than printing and calling `System.exit`, we just `throw` whenever we're *not* forked. The detection here actually seems robust enough to lean on, particularly since the degradation path is fine. This also has the pleasing side effect of integrating better with sbt's `last` command.

Fixes #2706

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/IOApp.scala
+++ b/core/jvm/src/main/scala/cats/effect/IOApp.scala
@@ -200,6 +200,9 @@ trait IOApp {
   def run(args: List[String]): IO[ExitCode]
 
   final def main(args: Array[String]): Unit = {
+    // checked in openjdk 8-17; this attempts to detect when we're running under artificial environments, like sbt
+    val isForked = Thread.currentThread().getName() == "main"
+
     if (runtime == null) {
       import unsafe.IORuntime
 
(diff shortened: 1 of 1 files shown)


core/jvm/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -200,6 +200,9 @@ trait IOApp {
   def run(args: List[String]): IO[ExitCode]
 
   final def main(args: Array[String]): Unit = {
+    // checked in openjdk 8-17; this attempts to detect when we're running under artificial environments, like sbt
+    val isForked = Thread.currentThread().getName() == "main"
```
armanbilge: Not sure what alternatives you considered, but I wonder if another candidate is `Thread.getId()` which seems to be `1` for the main thread. Unlike the thread name, at least it's immutable? /shrug

djspiewak: Ah! I'll check on the id. I would prefer that to a string comparison.

vasilmkd (approved):

core/jvm/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -200,6 +200,9 @@ trait IOApp {
   def run(args: List[String]): IO[ExitCode]
 
   final def main(args: Array[String]): Unit = {
+    // checked in openjdk 8-17; this attempts to detect when we're running under artificial environments, like sbt
```
LeifW, line 203: Does this comment mean that this has been tested on those Java versions? I'm on Java 23, and `Thread.currentThread().getId()` returns 1 - though that `.getId()` got deprecated in Java 19, because:
> This method is not final and may be overridden to return a value that is not the thread ID. Use [threadId()](https://download.java.net/java/early_access/valhalla/docs/api/java.base/java/lang/Thread.html#threadId()) instead

PR #2724, opened 2022-01-02
Handle interruption around blocking waits

### LINKED ISSUE #2720: `FatalErrorUnsafeRun` test is now nondeterministic
Rarely failing with this error:

```
[error]   x exit on fatal error with other unsafe runs
[error]    java.lang.InterruptedException
[error]    	at java.util.concurrent.locks.AbstractQueuedSynchronizer.tryAcquireSharedNanos(AbstractQueuedSynchronizer.java:1326)
[error]    	at java.util.concurrent.CountDownLatch.await(CountDownLatch.java:277)
[error]    	at cats.effect.IOPlatform.$anonfun$unsafeRunTimed$2(IOPlatform.scala:81)
[error]    	at scala.runtime.java8.JFunction0$mcZ$sp.apply(JFunction0$mcZ$sp.scala:17)
[error]    	at cats.effect.unsafe.WorkerThread.blockOn(WorkerThread.scala:572)
[error]    	at scala.concurrent.package$.blocking(package.scala:124)
[error]    	at cats.effect.IOPlatform.unsafeRunTimed(IOPlatform.scala:81)
[error]    	at cats.effect.IOPlatform.unsafeRunSync(IOPlatform.scala:42)
[error]    	at catseffect.examples.FatalErrorUnsafeRun$.$anonfun$run$3(examplesplatform.scala:42)
[error]    	at cats.effect.unsafe.WorkerThread.blockOn(WorkerThread.scala:607)
[error]    	at scala.concurrent.package$.blocking(package.scala:124)
[error]    	at cats.effect.IOFiber.runLoop(IOFiber.scala:948)
[error]    	at cats.effect.IOFiber.execR(IOFiber.scala:1293)
[error]    	at cats.effect.IOFiber.run(IOFiber.scala:123)
[error]    	at cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:530) doesn't contain 'Boom!' (IOAppSpec.scala:217)
```

I think this is happening because of the interruption handling on the workers, creating a race condition between the initiating exception and the interruption which results.

vasilmkd:
Resolves #2720.

Removes code that captures mutable variables and volatile at that, for which the Scala runtime has specific classes (`VolatileMutableReference` or something similar, TIL).

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/IOApp.scala
+++ b/core/jvm/src/main/scala/cats/effect/IOApp.scala
@@ -294,16 +294,12 @@ trait IOApp {
         val cancelLatch = new CountDownLatch(1)
         fiber.cancel.unsafeRunAsync(_ => cancelLatch.countDown())(runtime)
 
-        try {
-          val timeout = runtimeConfig.shutdownHookTimeout
-          if (timeout.isFinite) {
-            blocking(cancelLatch.await(timeout.length, timeout.unit))
-            ()
-          } else {
-            blocking(cancelLatch.await())
-          }
-        } catch {
-          case _: InterruptedException =>
+        val timeout = runtimeConfig.shutdownHookTimeout
+        if (timeout.isFinite) {
+          blocking(cancelLatch.await(timeout.length, timeout.unit))
+          ()
+        } else {
+          blocking(cancelLatch.await())
         }
       }
 
--- a/core/jvm/src/main/scala/cats/effect/IOApp.scala
+++ b/core/jvm/src/main/scala/cats/effect/IOApp.scala
@@ -270,18 +270,18 @@ trait IOApp {
     val fiber =
       ioa.unsafeRunFiber(
         {
-          queue.offer(new CancellationException("IOApp main fiber was canceled"))
           counter.decrementAndGet()
+          queue.offer(new CancellationException("IOApp main fiber was canceled"))
           ()
         },
         { t =>
-          queue.offer(t)
           counter.decrementAndGet()
+          queue.offer(t)
           ()
         },
         { a =>
-          queue.offer(a)
           counter.decrementAndGet()
+          queue.offer(a)
           ()
         }
       )(runtime)
--- a/core/jvm/src/main/scala/cats/effect/IOApp.scala
+++ b/core/jvm/src/main/scala/cats/effect/IOApp.scala
@@ -361,9 +361,6 @@ trait IOApp {
         case t: Throwable =>
           t.printStackTrace()
           rt.halt(1)
-        case null =>
-          println(
-            s"result is null but is interrupted? ${Thread.currentThread().isInterrupted()}")
       }
     } catch {
       // this handles sbt when fork := false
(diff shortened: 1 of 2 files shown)


core/jvm/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -261,43 +262,48 @@ trait IOApp {
...
+    val counter = new AtomicInteger(1)
 
     val ioa = run(args.toList)
 
     val fiber =
       ioa.unsafeRunFiber(
         {
-          error = new CancellationException("IOApp main fiber was canceled")
-          latch.countDown()
+          counter.decrementAndGet()
```
djspiewak, line 273: TIL `decrement` isn't a method on `AtomicInteger.

core/jvm/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -261,43 +262,48 @@ trait IOApp {
...
           if (timeout.isFinite) {
-            cancelLatch.await(timeout.length, timeout.unit)
+            blocking(cancelLatch.await(timeout.length, timeout.unit))
+            ()
           } else {
-            cancelLatch.await()
+            blocking(cancelLatch.await())
           }
+        } catch {
+          case _: InterruptedException =>
```
djspiewak: Won't this happen in Sbt if someone hits Ctrl-C repeatedly? I mean, if the interruption were working *at all* in Sbt, which it isn't.

core/jvm/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -355,6 +361,9 @@ trait IOApp {
         case t: Throwable =>
           t.printStackTrace()
           rt.halt(1)
+        case null =>
+          println(
+            s"result is null but is interrupted? ${Thread.currentThread().isInterrupted()}")
```
djspiewak: Debug code? Feels like this case should be handled similarly to the `catch { case _: InterruptedException => ...}` below.

djspiewak (changes requested):
The use of the `Queue` is super nice honestly. I never thought of doing it that way. Some quick comments on the interrupted `poll` cases but otherwise this looks good to me!

core/jvm/src/main/scala/cats/effect/IOApp.scala
vasilmkd: What do you suggest should be done here? No `catch` at all?

core/jvm/src/main/scala/cats/effect/IOApp.scala
vasilmkd: Yes. I had a bug in my original implementation because I accidentally called `queue.poll()` instead of `queue.take()` and I was getting strange `null` results because `poll` is non-blocking and returns `null` if the queue is empty. I forgot to remove this code.

core/jvm/src/main/scala/cats/effect/IOApp.scala
vasilmkd: I removed it. This matches the previous behavior.

djspiewak (approved):

PR #2731, opened 2022-01-06
Invoke `process.exit` within a `Try` in IOApp.js

### LINKED ISSUE #2722: live fiber snapshot test is flaky on JVM
ubuntu-latest, 2.13.7, graalvm-ce-java11@21.3

https://github.com/typelevel/cats-effect/runs/4681567043?check_suite_focus=true#step:9:3480

```
[error]   x live fiber snapshot
[error]     doesn't contain 'cats.effect.IOFiber' (IOAppSpec.scala:271)
```

### LINKED ISSUE #2730: `Uncaught TypeErrror` in browser console after upgrading to 3.3.3
After upgrading from 3.3.2 to 3.3.3 I started seeing error messages in the browser console on startup. The message is of the form `Uncaught TypeError: f$1 is not a function`.

One of the lines of code in the callstack is
```
hardExit$3.apply$mcVI$sp__I__V(c.Lcats_effect_ExitCode__f_code)
```
With the line above it is:
```
var c = $j_cats$002eeffect$002eExitCode.$as_Lcats_effect_ExitCode(c$2);
 hardExit$3.apply$mcVI$sp__I__V(c.Lcats_effect_ExitCode__f_code)
```
This looks similar to this line of code from #2707: https://github.com/typelevel/cats-effect/blob/614e98982602263d1447f3a9975be84bb4c23a67/core/js/src/main/scala/cats/effect/IOApp.scala#L243

I haven't been able to verify it is the problem, but the cats effect upgrade is the only change. If needed, I could try to create a minimal example. The application it is occurring in does not qualify as simple...

armanbilge:
Fixes #2730.

We have a `Try` in case reading the global variable `process.exit` fails, but just because it succeeded, doesn't mean we got a legit `exit` function. In #2730, `process` existed but `process.exit` was `undefined`, and in theory there could be just about anything in there :) in which case attempting some sort of typecheck seems fruitless. So this PR just wraps it all up in one more `Try`.

Since browsers can configure CE via a `process.env` global, this sort of situation can happen quite easily.

I didn't add a test, but I would like to. I need to think about how to set this one up.

### DIFF
--- a/core/js/src/main/scala/cats/effect/IOApp.scala
+++ b/core/js/src/main/scala/cats/effect/IOApp.scala
@@ -219,8 +219,10 @@ trait IOApp {
     val argList = process.argv.getOrElse(args.toList)
 
     // Store the default process.exit function, if it exists
-    val hardExit =
-      Try(js.Dynamic.global.process.exit.asInstanceOf[js.Function1[Int, Unit]]: Int => Unit)
+    val hardExit: Int => Unit =
+      Try(js.Dynamic.global.process.exit.asInstanceOf[js.Function1[Int, Unit]])
+        // we got *something*, but we don't know what it is really. so wrap in a Try
+        .map(f => (i: Int) => { Try(f(i)); () })
         .getOrElse((_: Int) => ())
 
     var cancelCode = 1 // So this can be updated by external cancellation
--- a/tests/js/src/main/scala/catseffect/examplesplatform.scala
+++ b/tests/js/src/main/scala/catseffect/examplesplatform.scala
@@ -43,11 +43,15 @@ package examples {
     register(FatalErrorUnsafeRun)
     register(Finalizers)
     register(LeakedFiber)
+    register(UndefinedProcessExit)
 
     @nowarn("cat=unused")
     def main(paperweight: Array[String]): Unit = {
       val args = js.Dynamic.global.process.argv.asInstanceOf[js.Array[String]]
       val app = args(2)
+      if (app == UndefinedProcessExit.getClass.getName.init)
+        // emulates the situation in browsers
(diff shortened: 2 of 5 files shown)


armanbilge:
Oh woops, I need to re-target this.

armanbilge:
We now have a test, and @toddburnside kindly verified this fix as well.

armanbilge:
This test on JVM has failed three times in a row 😩 

```
[error]   x live fiber snapshot
[error]     doesn't contain 'cats.effect.IOFiber' (IOAppSpec.scala:278)
```

Linking to https://github.com/typelevel/cats-effect/issues/2722.

armanbilge:
Same failure ?? maybe I actually broke it somehow

armanbilge:
It's just CI it seems. Locally:
```
sbt:root> testsJVM/testOnly *.IOAppSpec
[info] IOAppSpec
[info] IOApp (jvm) should
[info]   + evaluate and print hello world
[info]   + pass all arguments to child
[info]   + run finalizers on TERM
[info]   + exit on non-fatal error
[info]   + exit on fatal error
[info]   + exit on fatal error with other unsafe runs
[info]   + exit on canceled
[info]   + exit with leaked fibers
[info]   + warn on global runtime collision
[info]   + abort awaiting shutdown hooks
[info]   + shutdown on worker thread interruption
[info]   + live fiber snapshot
[info] Total for specification IOAppSpec
[info] Finished in 8 seconds, 644 ms
[info] 12 examples, 0 failure, 0 error
[info] Passed: Total 12, Failed 0, Errors 0, Passed 12
[success] Total time: 10 s, completed Jan 6, 2022, 3:26:36 PM
```

djspiewak (approved):

Issue #2333, opened 2021-09-09
Event loop implementation?


armanbilge:
This has come up at least a couple times in different contexts and I think this would be a cool idea with some interesting applications. Plus @keynmol may have volunteered himself to take this on 😉 

Why:
- JVM: `IO#unsafeRunSync` in single-threaded environments (e.g. AWS lambda) without the penalty of creating threads or shifting.
- JS: `unsafeRunSync` (!!) via a new "`TemporalSyncIO`" data type that implements `Temporal` and `Sync`. @djspiewak pointed out this would be similar to the microtask loop (with the addition of timers).
- Native: when Native rolls around, IIUC this _should_ give us a single-threaded runtime essentially for free.

armanbilge:
It occurred to me, another place this could be very useful is in fs2's sync compiler, which gets tripped up when used to run streams that need `Concurrent`. I think Ross suggested something like this in https://github.com/typelevel/fs2/pull/2650#issuecomment-930378990.

See:
https://github.com/typelevel/fs2/issues/2371
https://github.com/typelevel/fs2/pull/2650
https://github.com/typelevel/fs2/pull/2657

bilki:
Not sure whether it could be the solution for this use case, but I'm posting it, as we discussed on discord: while trying to work with some Java frameworks, like QtJambi or LWJGL, sometimes it's required for certain functions to be run on the main thread.

A workaround could be to execute what's needed to be run on the main thread _before_ entering the `IO` context, then proceed with the rest of the program. But as soon as you need to run a function on the main thread again inside the `IO`, you're back to square one.

I don't have a repo to show the use case, but a sample snippet could look like more or less like:

```scala
package com.example

import cats.effect.{IO, IOApp}
import io.qt.widgets.*

object Sample extends IOApp.Simple:
  override val run =
    for
      app <- IO.delay(QApplication.initialize(Array.empty))
      _   <- // IO stuff running outside the main thread
      _   <- IO.delay(QMessageBox.information(null, "QtJambi", "Hello World"))
      _   <- // even more IO stuff running outside the main thread
      _   <- IO.delay(QApplication.shutdown)
    yield ()
```

djspiewak:
Crazy idea, but what if you could do something like `.evalOn(IO.mainThread)`?

bilki:
[...]

PR #2732, opened 2022-01-06
Expose an `IOApp#MainThread` executor

djspiewak:
Solution for part of #2333.

This builds on @vasilmkd's insight that we can safely use an `ArrayBlockingQueue` to ferry results rather than a `var` and a `CountDownLatch`. With this change, the queue is expanded to allow values of type `Runnable`, indicating the enqueue of an action using the `MainThread` `ExecutionContext`.

There are several subtleties with this, one of them being that the main thread is *also* the way in which the program needs to end. Thus, whenever the main fiber completes, we `clear()` the queue to remove any `Runnable`s which have been enqueued by orphaned fibers (which are about to die anyway). Some care must be taken here to ensure we don't accidentally `clear()` the real results. An additional subtlety is that the queue *must* be strictly bounded in size, otherwise we risk forcing an allocation when we propagate fatal errors back to the main thread. In practice, this is unlikely to be a serious issue since it is expected that most use-cases which require this functionality will have relatively limited throughput, but it's still at least possible to overrun the currently hardcoded limit of `32` outstanding tasks. In order to avoid creating a trivial deadlock scenario, we first attempt to `offer` tasks, and if that fails, we use the blocking executor in the runtime itself to shunt off a blocking `put`.

In practice, it is recommended that users gate any serious use of `MainThread` with some higher semantic rate limiting, such as `Semaphore`.


armanbilge:
Regarding a test:
1. Add the `IOApp` here: https://github.com/typelevel/cats-effect/blob/series/3.x/tests/jvm/src/main/scala/catseffect/examplesplatform.scala
2. Make a same-named dummy `IOApp` here (no need to register it): https://github.com/typelevel/cats-effect/blob/series/3.x/tests/js/src/main/scala/catseffect/examplesplatform.scala
3. Run the test conditionally like this: https://github.com/typelevel/cats-effect/blob/614e98982602263d1447f3a9975be84bb4c23a67/tests/jvm/src/test/scala/cats/effect/IOAppSpec.scala#L238

core/jvm/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -183,6 +184,32 @@ trait IOApp {
...
+   */
+  protected lazy val MainThread: ExecutionContext =
+    new ExecutionContext {
+
+      def reportFailure(t: Throwable): Unit =
+        t.printStackTrace()
+
+      def execute(r: Runnable): Unit =
+        if (!queue.offer(r)) {
+          blocking(queue.put(r))
```
armanbilge: Is it possible that the main thread can deadlock? Since it has no capability to handle blocking, if it schedules fibers on itself up to capacity.

djspiewak: Yeah you can definitely deadlock the main thread. Off the top of my head, this would probably do it: `IO.cede.foreverM.start.foreverM.evalOn(MainThread)`

armanbilge: Hmm. Instead of wrapping in `block`, could we actually run it (the enqueue) on the blocking EC? So at least main doesn't become dead-locked.

vasilmkd:
I support exposing the main thread as an execution context to `IOApps`. I dislike reusing the queue I introduced for this purpose. My ideal solution would be to create our own non-daemon `io-main` thread which would do what is currently being done in `public static void main` by the `main` thread, while the `ExecutionContext` backed by the JVM main thread would use its own queue for `Runnables`. There is no need for the mechanism behind `IOApp` to actually run on the JVM `main` thread, and the JVM will wait for all non-daemon threads to exit before exiting, thus achieving both separation of concerns and avoiding weird bugs around deadlocking the thread which needs to deliver the final program result.

djspiewak:
> There is no need for the mechanism behind IOApp to actually run on the JVM main thread, and the JVM will wait for all non-daemon threads to exit before exiting, thus achieving both separation of concerns and avoiding weird bugs around deadlocking the thread which needs to deliver the final program result.

The problem with this is there's a real difference between `System.exit(0)` and allowing the `main` to return (which is part of why we check for `ExitCode.Success` and allow execution to just fall through in that event). I fear that if we start relying on daemon thread behavior and `System.exit`, we're going to be setting ourselves up for a world of unforeseen hurt.

I'd like to better understand why you object to reusing the `main` thread in this fashion. The `BlockingQueue` is there and actually works extremely well. The fact that it has an arbitrary limit of 1024 is simply a random choice I made; there's no particular reason we can't use an unbounded queue here. The only tradeoff on overloading this queue is we delay the results of the program by enqueueing thunks. I'm not sure whether or not this is a real problem, but it seems like a fair objection at any rate.

I'm also curious as to how you would see the `io-main` approach working. We need to use *the main thread* in order to handle cases like OpenGL (see some of the discussion in #2333). We can't just use a secondary thread with a different ID. This in turn would seem to imply that we need to do something at least somewhat like what's in this PR, and we can't shunt to a separate thread.

djspiewak:
> There is no need for the mechanism behind IOApp to actually run on the JVM main thread, and the JVM will wait for all non-daemon threads to exit before exiting, thus achieving both separation of concerns and avoiding weird bugs around deadlocking the thread which needs to deliver the final program result.

The problem with this is there's a real difference between `System.exit(0)` and allowing the `main` to return (which is part of why we check for `ExitCode.Success` and allow execution to just fall through in that event). I fear that if we start relying on daemon thread behavior and `System.exit`, we're going to be setting ourselves up for a world of unforeseen hurt.

I'd like to better understand why you object to reusing the `main` thread in this fashion. The `BlockingQueue` is there and actually works extremely well. The fact that it has an arbitrary limit of 1024 is simply a random choice I made; there's no particular reason we can't use an unbounded queue here. The only tradeoff on overloading this queue is we delay the results of the program by enqueueing thunks. I'm not sure whether or not this is a real problem, but it seems like a fair objection at any rate.

I'm also curious as to how you would see the `io-main` approach working. We need to use *the main thread* in order to handle cases like OpenGL (see some of the discussion in #2333). We can't just use a secondary thread with a different ID. This in turn would seem to imply that we need to do something at least somewhat like what's in this PR, and we can't shunt to a separate thread.

core/jvm/src/main/scala/cats/effect/IOApp.scala
djspiewak: > Hmm. Instead of wrapping in block, could we actually run it (the enqueue) on the blocking EC? So at least main doesn't become dead-locked.

That does seem like a better approach tbh.

vasilmkd:
We were talking past each other, but it doesn't really matter. I'm worried that the main thread can be used with `evalOn` and I don't have anything smart to suggest there.

djspiewak:
>  I'm worried that the main thread can be used with evalOn

I'm moderately worried about it as well, though I think if we make the queue unbounded, it should resolve the issues with deadlocking. Additionally, tossing a `clear()` right before we enqueue the results should avoid the majority of the problems associated with locking out program termination.

We should definitely be a bit more "beware of clown snakes" in the documentation than I am right now, though.

djspiewak:
Alright, I think this is ready for review. @vasilmkd I believe you still have some objections? I'm definitely not married to this idea, but I thought it was interesting enough to push to completion, and I can't see another way to handle thread-id-sensitive scenarios.

armanbilge:
> I can't see another way to handle thread-id-sensitive scenarios

We could still just have an event loop EC running on the main thread. But, I think the concept here is pretty cool :)

djspiewak:
> We could still just have an event loop EC running on the main thread

I thought through that a bit, but honestly I can't see a difference between that and this. At the end of the day, we still need all of the machinery which was previously on the main thread to *remain* on the main thread (e.g. error propagation), so the weird blending of the results callback and the executor seems pretty fundamental.

vasilmkd:
Alright, finally doing the review. Can you please reassure me about `clear()`.

I have some time this evening, can I please quickly prototype my original comment above? It will illustrate most of the points of contention on this PR and we will walk through them comparatively.

djspiewak:
> can I please quickly prototype my original comment above?

Please do! :-)

> Can you please reassure me about clear().

So `clear()` is ultimately encoding a *daemon* semantic on the main thread executor. Let's compare what it does to what implicitly happens to all other fibers.

In the event that you have orphaned fibers running in the background when the main fiber produces a result, those fibers are basically unaffected (because orphan) and will continue executing. However, the *threads* they are executing on will be abruptly terminated when the application itself exits a moment later. Thus, those fibers never release their resources, never finish their processing, and simply cease to exist along with the rest of the universe around them. We get this for free on every thread that *isn't* `main`.

With the main thread, the problem is that we can't exit the application until that thread is done working, but the way we've defined things, that thread will *not* be done working until all of the work on the `MainThread` executor is flushed. So if you have an orphan fiber which is `evalOn`ed to `MainThread`, that orphan will continue pumping work into the executor and thus blocking termination. So what `clear()` is meant to do is emulate daemon thread semantics on `main` itself by basically terminating all work as swiftly as possible just as the process itself is about to terminate.

I'm relatively confident we need to have *something* a bit like this, though perhaps encoded in a different way, because otherwise a spammy orphan fiber could block process exit, but only if `evalOn`ed onto `MainThread`, which seems broken.

vasilmkd:
Ok, that all makes sense, and thanks for documenting this here, for future reference. Basically there is no way to avoid what you described, so even in my prototype, I would still have to signal this interruption to the main thread. Humor me this, would anything bad happen if we call `Thread.interrupt()` on the main thread?

vasilmkd:
Alright, my idea is actually not very different in the end and I've convinced myself it doesn't "solve" any of the issues in a better way. Let's go with this PR. Sorry for dragging this out for so long.

vasilmkd (approved):

vasilmkd:
I've looked at the code thoroughly, I am able to maintain it if any issues arise (I'm mainly worried about unforseen use cases, but let's wait and see and fix later).

armanbilge (approved):
> Alright, my idea is actually not very different in the end and I've convinced myself it doesn't "solve" any of the issues in a better way. Let's go with this PR. Sorry for dragging this out for so long.

Yes, sorry from me too, Daniel also had to convince me the event loop EC would fundamentally have the same issues 🙃

PR #2733, opened 2022-01-07
Correct `IO#racePair` docs

TimWSpence:


armanbilge (dismissed):
Thanks!

armanbilge:
Oh, actually can this be re-targeted at 3.3.x?

TimWSpence:
> Oh, actually can this be re-targeted at 3.3.x?

I tried and failed horribly 😂 What is the preferred approach to this? Change the base in GH and then rebase my branch onto 3.3.x?

armanbilge:
Yeah, unfortunately can't be done just from GH :)

TimWSpence:
> Yeah, unfortunately can't be done just from GH :)

Yeah I suspected not! Done now (I think 😂 )

djspiewak:
This is great. It does need a run of `sbt prePR` though :-)

djspiewak (dismissed):
This is great. It does need a run of sbt prePR though :-)

djspiewak:
So correction: docs stuff should ideally be targeted at `series/3.x`. Sorry to keep bouncing it around a bit :-)

armanbilge:
Wait, why is that? Are we publishing the docs from 3.x?

Edit: If we are, that doesn't make sense because it means are publishing docs for new APIs e.g. `allocatedCase` that have not yet been published. Furthermore, this fix should benefit users who read their scaladocs in their IDE. In both cases, I would think we want it merged into 3.3.x.

armanbilge (approved):

PR #2743, opened 2022-01-10
Workaround for dotty#14240

armanbilge:
Reported by @mn98. This turned out to be a Scala 3 bug reported in https://github.com/lampepfl/dotty/issues/14240, but we can code defensively around it in the mean time.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,6 +1,11 @@
 version = 3.2.1
 
 runner.dialect = Scala213Source3
+fileOverride {
+  "glob:**/scala-3/**" {
+    runner.dialect = scala3
+  }
+}
 
 project.excludeFilters = [
   "scalafix/*"
--- a/core/shared/src/main/scala-3/cats/effect/ExitCode.scala
+++ b/core/shared/src/main/scala-3/cats/effect/ExitCode.scala
@@ -16,18 +16,13 @@
 
 package cats.effect
 
-import scala.annotation.targetName
-
 /**
  * Represents the exit code of an application.
  *
  * `code` is constrained to a range from 0 to 255, inclusive.
  */
 // The strange encoding of this class is due to https://github.com/lampepfl/dotty/issues/14240
-sealed class ExitCode private[effect] (val code: Int)
-    extends Product
-    with Equals
-    with Serializable {
+sealed class ExitCode private[effect] (val code: Int) extends Product, Equals, Serializable {
 
   def _1: Int = code
 
--- a/core/shared/src/main/scala-3/cats/effect/ExitCode.scala
+++ b/core/shared/src/main/scala-3/cats/effect/ExitCode.scala
@@ -28,6 +28,9 @@ sealed class ExitCode private[effect] (val code: Int)
     extends Product
     with Equals
     with Serializable {
+
+  def _1: Int = code
+
   def canEqual(that: Any): Boolean = that.isInstanceOf[ExitCode]
 
   override def equals(that: Any): Boolean = that match {
@@ -55,11 +58,7 @@ object ExitCode {
    */
   def apply(i: Int): ExitCode = new ExitCode(i & 0xff)
 
-  @targetName("unapply")
-  private[effect] def unapplyBinCompat(ec: ExitCode): Option[Int] = Some(ec.code)
-
-  @targetName("unapplyTotal")
-  def unapply(ec: ExitCode): Some[Int] = Some(ec.code)
+  def unapply(ec: ExitCode): ExitCode = ec
 
   val Success: ExitCode = ExitCode(0)
   val Error: ExitCode = ExitCode(1)
--- a/core/shared/src/main/scala-3/cats/effect/ExitCode.scala
+++ b/core/shared/src/main/scala-3/cats/effect/ExitCode.scala
@@ -16,6 +16,8 @@
 
 package cats.effect
 
+import scala.annotation.targetName
+
 /**
  * Represents the exit code of an application.
  *
@@ -53,7 +55,11 @@ object ExitCode {
    */
   def apply(i: Int): ExitCode = new ExitCode(i & 0xff)
 
-  def unapply(ec: ExitCode): Option[Int] = Some(ec.code)
+  @targetName("unapply")
+  private[effect] def unapplyBinCompat(ec: ExitCode): Option[Int] = Some(ec.code)
+
+  @targetName("unapplyTotal")
+  def unapply(ec: ExitCode): Some[Int] = Some(ec.code)
 
   val Success: ExitCode = ExitCode(0)
   val Error: ExitCode = ExitCode(1)
--- a/core/shared/src/main/scala/cats/effect/ExitCode.scala
+++ b/core/shared/src/main/scala/cats/effect/ExitCode.scala
@@ -21,7 +21,26 @@ package cats.effect
  *
  * `code` is constrained to a range from 0 to 255, inclusive.
  */
-sealed abstract case class ExitCode private (code: Int)
+// The strange encoding of this class is due to https://github.com/lampepfl/dotty/issues/14240
+sealed class ExitCode private[effect] (val code: Int)
+    extends Product
+    with Equals
+    with Serializable {
+  def canEqual(that: Any): Boolean = that.isInstanceOf[ExitCode]
+
+  override def equals(that: Any): Boolean = that match {
+    case ExitCode(code) => this.code == code
+    case _ => false
+  }
+
+  def productArity: Int = 1
+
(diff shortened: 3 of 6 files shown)


core/shared/src/main/scala/cats/effect/ExitCode.scala
```diff
@@ -21,7 +21,26 @@ package cats.effect
...
+  override def equals(that: Any): Boolean = that match {
+    case ExitCode(code) => this.code == code
+    case _ => false
+  }
+
+  def productArity: Int = 1
+
+  def productElement(n: Int): Any = if (n == 0) code else throw new NoSuchElementException
+
+  override def hashCode = code
```
armanbilge: I cheated on this one. If we really care about keeping this stable, I can figure out what's supposed to go here.

core/shared/src/main/scala/cats/effect/ExitCode.scala
```diff
@@ -32,7 +51,9 @@ object ExitCode {
    *   the value whose 8 least significant bits are used to construct an exit code within the
    *   valid range.
    */
-  def apply(i: Int): ExitCode = new ExitCode(i & 0xff) {}
+  def apply(i: Int): ExitCode = new ExitCode(i & 0xff)
+
+  def unapply(ec: ExitCode): Option[Int] = Some(ec.code)
```
djspiewak:
```suggestion
  def unapply(ec: ExitCode): Some[Int] = Some(ec.code)
```

This way we keep exhaustiveness checking

djspiewak (dismissed):
This is really annoying :-( As long as mima blesses it though I think we'll be okay.

core/shared/src/main/scala/cats/effect/ExitCode.scala
vasilmkd: As a wrapper of an integer, this is ok.

vasilmkd (changes requested):
Just Daniel's suggestion.

core/shared/src/main/scala/cats/effect/ExitCode.scala
armanbilge: This is how I did it at first, but mima complained actually. So I have no idea if/how we are getting exhaustive checking right now, maybe not at the bytecode level 🤔

armanbilge:
Actually, hang on. The lack of exhaustivity check would be source-breaking, and quite frankly annoying. This is no good. I wonder if we can work around it.

core/shared/src/main/scala-3/cats/effect/ExitCode.scala
```diff
@@ -0,0 +1,66 @@
...
+   *   the value whose 8 least significant bits are used to construct an exit code within the
+   *   valid range.
+   */
+  def apply(i: Int): ExitCode = new ExitCode(i & 0xff)
+
+  @targetName("unapply")
+  private[effect] def unapplyBinCompat(ec: ExitCode): Option[Int] = Some(ec.code)
+
+  @targetName("unapplyTotal")
+  def unapply(ec: ExitCode): Some[Int] = Some(ec.code)
```
armanbilge: Bincompat is so much more fun on Scala 3 :)

armanbilge:
@djspiewak we really need that sbt-spiewak release with mima for Scala 3 enabled ... 

I manually added:
```scala
mimaPreviousArtifacts += "org.typelevel" %%% "cats-effect" % "3.3.4",
```

Not only did I miss the `_1` method, it seems the `unapply` encoding on Scala 3 actually doesn't use `Option` at all?

```
[error] cats-effect: Failed binary compatibility check against org.typelevel:cats-effect_3:3.3.4! Found 3 potential problems
[error]  * static method unapply(cats.effect.ExitCode)cats.effect.ExitCode in class cats.effect.ExitCode has a different result type in current version, where it is scala.Option rather than cats.effect.ExitCode
[error]    filter with: ProblemFilters.exclude[IncompatibleResultTypeProblem]("cats.effect.ExitCode.unapply")
[error]  * method _1()Int in class cats.effect.ExitCode does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("cats.effect.ExitCode._1")
[error]  * method unapply(cats.effect.ExitCode)cats.effect.ExitCode in object cats.effect.ExitCode has a different result type in current version, where it is scala.Option rather than cats.effect.ExitCode
[error]    filter with: ProblemFilters.exclude[IncompatibleResultTypeProblem]("cats.effect.ExitCode.unapply")
[error] stack trace is suppressed; run last coreJVM / mimaReportBinaryIssues for the full output
[error] (coreJVM / mimaReportBinaryIssues) Failed binary compatibility check against org.typelevel:cats-effect_3:3.3.4! Found 3 potential problems
```

armanbilge:
Forgot to indicate ready-for-review :)

core/shared/src/main/scala-3/cats/effect/ExitCode.scala
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Represents the exit code of an application.
+ *
+ * `code` is constrained to a range from 0 to 255, inclusive.
+ */
+// The strange encoding of this class is due to https://github.com/lampepfl/dotty/issues/14240
+sealed class ExitCode private[effect] (val code: Int)
+    extends Product
+    with Equals
+    with Serializable {
```
vasilmkd: I wouldn't object if the new syntax was used in this file.

core/shared/src/main/scala-3/cats/effect/ExitCode.scala
```diff
@@ -0,0 +1,65 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+
+import scala.annotation.targetName
```
vasilmkd: This might be unused. Please double check.

core/shared/src/main/scala-3/cats/effect/ExitCode.scala
```diff
@@ -0,0 +1,65 @@
...
+  def canEqual(that: Any): Boolean = that.isInstanceOf[ExitCode]
+
+  override def equals(that: Any): Boolean = that match {
+    case ExitCode(code) => this.code == code
+    case _ => false
+  }
+
+  def productArity: Int = 1
+
+  def productElement(n: Int): Any = if (n == 0) code else throw new NoSuchElementException
```
vasilmkd, line 38: The quirks of Scala 3. Scala 3 supports creating instances of objects without the `new` keyword, but it can't do it in this case because it cannot see through the `type NoSuchElementException = java.util.NoSuchElementException` alias. It works if you import it, which is whatever.

More on the subject https://docs.scala-lang.org/scala3/reference/other-new-features/creator-applications.html.

vasilmkd (changes requested):
Here's more on how `unapply` and pattern matching has changed in Scala 3. https://docs.scala-lang.org/scala3/reference/changed-features/pattern-matching.html

And they say Scala 3 is not a new language.

core/shared/src/main/scala-3/cats/effect/ExitCode.scala
armanbilge: Our scalac options like the braces I think.

core/shared/src/main/scala-3/cats/effect/ExitCode.scala
armanbilge: Good catch, no unused warnings :(

core/shared/src/main/scala-3/cats/effect/ExitCode.scala
vasilmkd: Does `fs2` have Scala 3 specific sources with the new syntax or did I misremember something. If we're not set up for it in `cats-effect`, this change is not worth that.

armanbilge: I mean, I hope the split sources are only temporary until the bug is fixed. But wishful thinking maybe :)

.scalafmt.conf
```diff
@@ -1,6 +1,11 @@
 version = 3.2.1
 
 runner.dialect = Scala213Source3
+fileOverride {
+  "glob:**/scala-3/**" {
+    runner.dialect = scala3
+  }
+}
```
vasilmkd, line 8: Does this do anything? Genuine question.

armanbilge: `scalafmt` raises a warning without it.

vasilmkd (approved):

djspiewak (approved):

PR #2749, opened 2022-01-12
Adds CommutativeApplicative instance for Resource.Par

skennedy:
It's my (admittedly limited!) understanding that Parallel implies CommutativeApplicative - so this adds a CommutativeApplicative instance to Resource's implicit scope (like there was for Parallel). Happy to be enlightened if this was deliberately missing for some reason...

I came across this converting code from CE2 to CE3 - the following stops compiling without this:

```
import cats.Parallel
import cats.effect.Concurrent
import cats.effect.Resource
import cats.syntax.parallel._

object Example {
  def foo[F[_]: Concurrent: Parallel, K, V1, V2](map: Map[K, V1]): Resource[F, Map[K, V2]] = {
    val op: V1 => Resource[F, V2] = ???

    map.parUnorderedTraverse(op)
  }

}
```

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -1103,6 +1103,10 @@ object Resource extends ResourceFOInstances0 with ResourceHOInstances0 with Reso
 
   implicit def parallelForResource[F[_]: Concurrent]: Parallel.Aux[Resource[F, *], Par[F, *]] =
     spawn.parallelForGenSpawn[Resource[F, *], Throwable]
+
+  implicit def commutativeApplicativeForResource[F[_]: Concurrent]
+      : CommutativeApplicative[Par[F, *]] =
+    spawn.commutativeApplicativeForParallelF[Resource[F, *], Throwable]
 }
 
 private[effect] trait ResourceHOInstances0 extends ResourceHOInstances1 {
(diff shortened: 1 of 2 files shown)


vasilmkd:
Thank you for opening this PR. I will let @djspiewak have the final say on this.

In the meantime, adding the import `cats.effect.implicits._` resolves the compilation issue, as a workaround. This import brings into scope the same instance used in this PR. Hope this is helpful.

armanbilge:
Maybe it would be helpful to add some tests for this? Like we did in https://github.com/typelevel/cats-effect/pull/2375.

djspiewak (dismissed):
Nice catch! I'm in favor of including this. I think I removed it when, for a brief window, `Parallel[Resource[F, *]]` did *not* in fact imply `CommutativeApplicative`. I would like to see at least one test though.

skennedy:
Great, I'd be happy to look into tests - I guess this would be mainly a case of enabling the laws tests from #2375 but for Resource?

armanbilge:
Thanks! Yes, I'd try adding the laws tests for `Resource[PureConc, *]` in [`ParallelFSpec`](https://github.com/armanbilge/cats-effect/blob/614e98982602263d1447f3a9975be84bb4c23a67/tests/shared/src/test/scala/cats/effect/kernel/ParallelFSpec.scala). That should probably be sufficient.

armanbilge (approved):

PR #2751, opened 2022-01-13
Invalid comment in IO.Par

jmgimeno:
The doc stated that Par.apply and Par.unwrap exist as methods to convert between IO and IO.Par but this is no longer the case. 

The change removes this paragraph and replaces it with one suggesting to ways to do the conversion:
- the Parallel[IO] instance
- methods apply / value from the object cats.effect.kernel.Par.ParallelF


armanbilge:
Thanks! If you don't mind targeting your PR at the series/3.3.x branch your changes can go live sooner :)

jmgimeno:
@armanbilge so I create another PR, with the same change, for the series/3.3.x branch?

armanbilge:
Yes, exactly. Or if you are handy with git you can try and rebase this one 😆 thanks!

jmgimeno:
Done

armanbilge:
Superseded by https://github.com/typelevel/cats-effect/pull/2752.

PR #2757, opened 2022-01-14
Add IO.rethrow and IO.debug

BalmungSan:
IMHO they both come in handy, especially when using concrete `IO` which many newcomers do.

Let me know if I could have done better in the **Scaladoc** comments or in the tests :)

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -166,11 +166,11 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
    * `IO.raiseError`. Thus:
    *
    * {{{
-   * IO.raiseError(ex).attempt.unsafeRunAsync === Left(ex)
+   * IO.raiseError(ex).attempt.unsafeRunSync() === Left(ex)
    * }}}
    *
    * @see
-   *   [[IO.raiseError]]
+   *   [[IO.raiseError]] [[rethrow]]
    */
   def attempt: IO[Either[Throwable, A]] =
     IO.Attempt(this)
@@ -490,6 +490,25 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
     IO.racePair(this, that)
 
   /**
+   * Inverse of `attempt`
+   *
+   * This function raises any materialized error.
+   *
+   * {{{
+   * IO(Right(a)).rethrow === IO.pure(a)
+   * IO(Left(ex)).rethrow === IO.raiseError(ex)
+   *
+   * // Or more generally:
+   * io.attempt.rethrow === io // For any io.
+   * }}}
+   *
+   * @see
+   *   [[IO.raiseError]] [[attempt]]
+   */
+  def rethrow[B](implicit ev: A <:< Either[Throwable, B]): IO[B] =
+    flatMap(a => IO.fromEither(ev(a)))
+
+  /**
    * Returns a new value that transforms the result of the source, given the `recover` or `map`
    * functions, which get executed depending on whether the result ends in error or if it is
    * successful.
@@ -558,6 +577,14 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
       IO.unit
     else
       flatMap(_ => replicateA_(n - 1))
+
+  /**
+   * Logs the value of this `IO` _(even if it is an error)_ to the stdout.
+   *
+   * This operation is intended as a quick debug, not as proper logging.
+   */
+  def debug: IO[A] =
+    attempt.flatTap(IO.println).rethrow
 
   /**
    * Returns an IO that will delay the execution of the source by the given duration.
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -489,6 +489,21 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
     (FiberIO[A @uncheckedVariance], OutcomeIO[B])]] =
     IO.racePair(this, that)
 
+  /**
+   * Inverse of `attempt`
+   *
+   * This function raises any materialized error.
+   *
+   * {{{
+   * IO.pure(3).attempt.rethrow.unsafeRunSync() === 3
```
armanbilge: Maybe the example should be more like `IO(Left(ex)).rethrow`? So that you don't need to know the signature of `attempt`, and also the `A <:< Either[Throwable, B]` is difficult to know what that means exactly.

BalmungSan: @armanbilge

> So that you don't need to know the signature of attempt

Yeah, I also thought about it but then what should be after the `===`? A `throw ex`?
_(one may also raise the argument that since this method is just the inverse of `attempt` the it is okay to expect to know it, like the example of `attempt` shows `IO.raiseError`)_

> and also the A <:< Either[Throwable, B] is difficult to know what that means exactly.

Not sure what you mean? Like, should I add a comment note? or what?

armanbilge: Maybe?
```scala
IO(Right(a)).rethrow === IO(a)
IO(Left(ex)).rethrow === IO.raiseError(ex)
```

Edit: actually, we have to be careful, `IO(Left(ex)).rethrow` is technically `IO(IO.raiseError(ex)).flatten` I believe. We could write it in terms of `pure` but that might confuse things.

> Not sure what you mean? Like, should I add a comment note? or what?

Sorry, I meant this is another good reason to have an example that specifically demonstrates the use of `Either`, because it is not straightforward to understand from the method signature that it has to be an `IO[Either[Throwable, A]]`.

BalmungSan: @armanbilge ahhh makes sense, thanks for the snippets; I just pushed an update.

BalmungSan:
Uhm the failing tests doesn't look like my fault, does it?

vasilmkd:
I'm asking from a user experience perspective, the printout of `debug` has a `Left/Right` wrapping around the result, which is not necessarily a part of the type signature. I wonder if a pattern match on the result of `attempt` and then printing out the vanilla result or the throwable is a better experience. Otherwise I quite like this change.

BalmungSan:
@vasilmkd yeah, I was thinking the same when I wrote the function, I was just worried that doing too much for a quick debug method was not worth it and may encourage users to use it for real logging _(because originally I thought of including things like thread name and timestamp)_

However, on second thought, a simple prefix like `Success: ${a}` and `Failure: ${ex.getMessage}` may be good enough.
What do you think?

wemrysi:
> However, on second thought, a simple prefix like Success: ${a} and Failure: ${ex.getMessage} may be good enough.
What do you think?

I think a prefix would be helpful for communicating the outcome at-a-glance (especially for values like `IO[Unit]`) 👍  Speaking of outcome, how about `Succeeded: ${a}` and `Errored: ${ex}` to be consistent with the cases in `Outcome` and include the name of the `Throwable` (as sometimes messages aren't terribly meaningful out of context)?

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -559,6 +578,14 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
     else
       flatMap(_ => replicateA_(n - 1))
 
+  /**
+   * Logs the value of this `IO` _(even if it is an error)_ to the stdout.
+   *
+   * This operation is intended as a quick debug, not as proper logging.
+   */
+  def debug: IO[A] =
```
```suggestion
  def debug(implicit A: Show[A] = Show.fromToString): IO[A] =
```

Just to be consistent with `IO.println` and to give extra flexibility on the `A` rendering.

djspiewak (changes requested):
Honestly both of these are fantastic. Couple thoughts.

1. Echoing @wemrysi's suggestion of following `Outcome`'s vernacular to render results. Though, there shouldn't be a situation where you render `Canceled` (or maybe we should?!). Literally an implementation built on `guaranteeCase` would do nicely, and would avoid the need to explicitly `rethrow`.
2. Might be worth looking at Fs2's `Stream#debug` function, which provides some extra configurability for the prefix (for everyone who likes to `println(">>>> THE THING HERE")`) and such

BalmungSan:
@vasilmkd @wemrysi @djspiewak

Hey all, sorry for the delay.
I just pushed a couple of changes regarding `debug`, I hope you like them :)
Let me know if you think there is something else that may be done.

Also, do you know if it is possible _(and worth it)_ to test `debug`?

djspiewak (approved):
Very nice! Also I don't see a need to test `debug`. It's basically just `println` :-P

PR #2758, opened 2022-01-15
Broke out contributing guidelines into a separate document

djspiewak:


CONTRIBUTING.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Process
+
+Fairly standard pull request workflow. Please do *not* rebase commits that you've already pushed to a non-draft pull request unless explicitly requested. It's also very helpful if you toggle on the "Allow edits and access to secrets by maintainers" check box. There are no secrets, so you aren't leaking anything. What this does is allows maintainers to push last-minute trivial fixes to your PR (e.g. scalafmt) and ensure it can be merged with appropriate attribution.
+
+We maintain multiple primary branches within the Cats Effect repository. The active branches are dependent on the most recent release. The following examples assume the most recent releases were 3.4.1 and 2.5.7 (in the 2.x and 3.x lineages, respective). For those most recent releases, the active branches will be as follows:
+
+- `series/3.x` (default)
+  + Any *binary compatible* changes on top of a 3.x version which are not source- or forward-compatible
+- `series/3.4.x`
```
vasilmkd, line 25: Did you mean `series/3.3.x` or is this meant for the future?

Also, it might be good to mention that changes to docs should go to that branch for quick deployment.

Finally, not sure if `forward compatible` is easily understandable for people, maybe we can simplify both `series/3.x` and this branch as:
- `series/3.x` - new methods and backward compatible API changes
- `series/3.3.x` - bugfixes

vasilmkd (commented):
Just one question/suggestion. Otherwise great addition.

CONTRIBUTING.md
djspiewak: Was more future-focusing.

As for "forward compatibility", I tried to illustrate it a little better (without loss of precision) in the following paragraph:

> When in doubt, just target your work against `series/3.x`. If you're *very* certain that your changes will be source- *and* forward-compatible, it's safe to target the latest major/minor series (e.g. `series/3.4.x`). Note that forward-compatibility means that you cannot *add or remove* an API, while backward-compatibility means you cannot *remove* an API.

Does that help? Or would it be better to refine the bullet list?

vasilmkd (approved):

PR #2773, opened 2022-01-24
Configurable caching of blocking threads

vasilmkd:
I invite you to bikeshed on the JVM property name.

This does slightly leak into JS land, because `IORuntimeConfig` is a shared source. The new parameter could be hidden in JS with a bit of source splitting. Let me know if that's worth the maintenance burden.


rossabaker:
Is it assumed that system properties are all read at startup?  Unlike environment variables, which are theoretically mutable, system properties are certainly mutable.

djspiewak:
They're read when `IORuntimeConfig` is classloaded.

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeConfigCompanionPlatform.scala
```diff
@@ -44,11 +44,17 @@ private[unsafe] abstract class IORuntimeConfigCompanionPlatform { this: IORuntim
         .map(Duration(_))
         .getOrElse(DefaultShutdownHookTimeout)
 
+    val blockingCacheExpiration =
+      Try(System.getProperty("cats.effect.blocking.cache.expiration"))
```
djspiewak: We should probably start thinking of namespaces a bit more hierarchically. Maybe `cats.effect.runtime.blocking.expiration`? Bikeshedding is hard.

djspiewak (changes requested):
I like this quite a bit, but I do think we should split it away from JavaScript. It's kind of a pain to have these split APIs, but we've been trying to follow a paradigm of keeping things fairly independent and not foisting APIs on JS which don't make sense on that platform.

djspiewak (approved):

djspiewak:
Btw we'll need to document this, but it's good enough to just get it in for now. The properties documentation is kind of bad right now.

PR #2775, opened 2022-01-24
Cats effect testing does not support munit

Guisanpea:
This PR fixes the fact that the documentation points to it

### DIFF
--- a/docs/guides/testing.md
+++ b/docs/guides/testing.md
@@ -9,7 +9,7 @@ title: Testing
 
 Relatively few libraries support cats-effect directly at this time. However, most (if not all) popular testing frameworks have libraries offering some level of integration
 
-- [cats-effect-testing](https://github.com/djspiewak/cats-effect-testing): Supports Scalatest, Specs2, munit, minitest, µTest, and scalacheck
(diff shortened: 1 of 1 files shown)


vasilmkd (approved):
Thanks for the contribution!

docs/guides/testing.md
```diff
@@ -9,7 +9,7 @@ title: Testing
 
 Relatively few libraries support cats-effect directly at this time. However, most (if not all) popular testing frameworks have libraries offering some level of integration
 
-- [cats-effect-testing](https://github.com/djspiewak/cats-effect-testing): Supports Scalatest, Specs2, munit, minitest, µTest, and scalacheck
+- [cats-effect-testing](https://github.com/djspiewak/cats-effect-testing): Supports Scalatest, Specs2, minitest, µTest, and scalacheck
```
sideeffffect, line 12: It might as well point to https://github.com/typelevel/cats-effect-testing

vasilmkd: That's a good observation, thanks. I do believe that the URL redirects correctly, but the change would be appreciated.

sideeffffect, line 12: It's https://github.com/typelevel/munit-cats-effect what supports MUnit

sideeffffect (changes requested):

PR #2776, opened 2022-01-25
Faster runnables on the WSTP

vasilmkd:
What it says on the tin.

`series/3.3.x`:
```scala
Benchmark                                  (size)   Mode  Cnt    Score   Error    Units
WorkStealingBenchmark.runnableScheduling  1000000  thrpt   20  222.826 ± 2.604  ops/min
```

`This PR`:
```scala
Benchmark                                  (size)   Mode  Cnt    Score   Error    Units
WorkStealingBenchmark.runnableScheduling  1000000  thrpt   20  537.854 ± 2.387  ops/min
```

What is the big idea?

Instead of allocating a full fiber with all of the necessary data structures like continuation and finalizer stacks which are needed to support the run loop, a special fiber is created that starts from a bespoke state that only runs the provided runnable and exits. I've managed to bring the implementation down to a single allocation of the `IOFiber` wrapper object around the runnable. Unfortunately, I don't think that anything better can be achieved due to the following reasons.

1. The `IOFiber` code nicely interacts with the provided `IORuntime` and shuts it down on a fatal error, this complicated code path is reused.
2. There is an alternative world where the whole WSTP, LocalQueue and WorkerThreads are reworked to hold `Runnable` instances instead of `IOFiber` instances. While this might look nice in theory, it doesn't interact well with the live fiber snapshots (runnables are simply not full `IOFibers`) and this creates additional branches when assigning fibers as active. Furthermore, the WorkerThreads need to add branches before every `run()` (which also becomes an `invokeinterface` instead of the current `invokevirtual`) so that they can wrap it in a `try/catch` to recreate the runtime shutdown code path on fatal errors. All in all, I don't feel comfortable adding all of these branches on the hot path, which I've worked so hard to optimize before. (I tried this whole approach and I didn't like it). The upside to this approach is that there's no wrapping of external `Runnables` executed on our runtime (think using the WSTP for execution of `Future`s without any overhead), but ultimately I think this is a wrong optimization to pursue, as it is inherently slowing down our own fibers.

I think this PR adds enough performance to still make our runtime an extremely safe general purpose EC with very good performance.

FWIW, here's how much one layer of wrapping is worth:
```scala
Benchmark                                             (size)   Mode  Cnt     Score    Error    Units
WorkStealingBenchmark.runnableSchedulingScalaGlobal  1000000  thrpt   20  1363.152 ± 12.125  ops/min
```

### DIFF
--- a/benchmarks/src/main/scala/cats/effect/benchmarks/WorkStealingBenchmark.scala
+++ b/benchmarks/src/main/scala/cats/effect/benchmarks/WorkStealingBenchmark.scala
@@ -132,7 +132,7 @@ class WorkStealingBenchmark {
       }
     }
 
-    (0 to theSize).foreach(_ => run(0))
+    (0 until theSize).foreach(_ => run(0))
 
     countDown.await()
   }
@@ -144,6 +144,11 @@ class WorkStealingBenchmark {
   @Benchmark
   def runnableScheduling(): Unit = {
     runnableSchedulingBenchmark(cats.effect.unsafe.implicits.global.compute)
+  }
+
+  @Benchmark
+  def runnableSchedulingScalaGlobal(): Unit = {
+    runnableSchedulingBenchmark(ExecutionContext.global)
   }
 
   lazy val manyThreadsRuntime: IORuntime = {
--- a/core/js/src/main/scala/cats/effect/IOFiberConstants.scala
+++ b/core/js/src/main/scala/cats/effect/IOFiberConstants.scala
@@ -42,7 +42,8 @@ private object IOFiberConstants {
   final val BlockingR: Byte = 5
   final val CedeR: Byte = 6
   final val AutoCedeR: Byte = 7
-  final val DoneR: Byte = 8
+  final val ExecuteRunnableR: Byte = 8
+  final val DoneR: Byte = 9
 
   // ContState tags
   final val ContStateInitial: Int = 0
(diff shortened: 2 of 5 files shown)


vasilmkd:
Actually, I want to think about this one a bit more.

djspiewak:
@vasilmkd This looks really good. I'm going to do a quick run of the scheduler benchmarks locally to check that we haven't regressed anything but otherwise I think this can make 3.3.5

djspiewak:
Before:

```
[info] Benchmark                                              (size)   Mode  Cnt    Score   Error    Units
[info] WorkStealingBenchmark.alloc                           1000000  thrpt   20   10.782 ± 0.041  ops/min
[info] WorkStealingBenchmark.manyThreadsSchedulingBenchmark  1000000  thrpt   20   42.274 ± 0.679  ops/min
[info] WorkStealingBenchmark.runnableScheduling              1000000  thrpt   20  170.456 ± 0.535  ops/min
[info] WorkStealingBenchmark.scheduling                      1000000  thrpt   20   51.538 ± 4.001  ops/min
```

After:

```
[info] Benchmark                                              (size)   Mode  Cnt     Score   Error    Units
[info] WorkStealingBenchmark.alloc                           1000000  thrpt   20    10.757 ± 0.058  ops/min
[info] WorkStealingBenchmark.manyThreadsSchedulingBenchmark  1000000  thrpt   20    44.989 ± 3.995  ops/min
[info] WorkStealingBenchmark.runnableScheduling              1000000  thrpt   20   467.229 ± 1.090  ops/min
[info] WorkStealingBenchmark.runnableSchedulingScalaGlobal   1000000  thrpt   20  2231.549 ± 7.020  ops/min
[info] WorkStealingBenchmark.scheduling                      1000000  thrpt   20    51.915 ± 1.784  ops/min
```

Looks stellar to me!

djspiewak (approved):

PR #2777, opened 2022-01-25
Add `andWait` method to IO

iRevive:
Quality of life improvement.

No need to import `cats.effect.syntax.temporal._` anymore.


vasilmkd (approved):
Nice. Thank you.

PR #2780, opened 2022-01-28
Add `recover` and `recoverWith` to IO.

kamilkloch:
No need to `import cats.syntax.all._`


djspiewak (approved):
This is great! Thank you!

PR #2781, opened 2022-01-28
Remove repeated CE2 dispatcher qualification

jisantuc:
This small PR removes some repeated text from dispatcher docs. Identical text is included further down under a `Cats Effect 2` header -- I chose to remove the top copy of the text, since analogies to CE2 seemed less urgent to communicate than the description / explanation of what dispatcher is for and how it works.

### DIFF
--- a/docs/std/dispatcher.md
+++ b/docs/std/dispatcher.md
@@ -7,8 +7,6 @@ title: Dispatcher
 
 `Dispatcher` is a fiber-based [`Supervisor`](./supervisor.md) utility for evaluating effects across an impure boundary. This is useful when working with reactive interfaces that produce potentially many values (as opposed to one), and for each value, some effect in `F` must be performed (like inserting each value into a queue).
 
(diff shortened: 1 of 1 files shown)


djspiewak:
My one reservation on this is the loss of SEO. How many people are googling "concurrenteffect cats effect 3" these days?

djspiewak (approved):

jisantuc:
> My one reservation on this is the loss of SEO.

I've been trying for three days to parse whether you're joking or not -- are a lot of people googling concurrenteffect cats effect 3 these days? FWIW the current first result takes you to the page with [this section](https://typelevel.org/cats-effect/docs/migration-guide#effect-concurrenteffect-synceffect)

ETA: I guess we're not even at a year yet of CE3 being available so I think it's probably a serious concern, but landing on the migration guide is still really good I think, and this change didn't remove `ConcurrentEffect` from the dispatcher docs, just the verbatim repeated paragraph

djspiewak:
> I've been trying for three days to parse whether you're joking or not -- are a lot of people googling concurrenteffect cats effect 3 these days? FWIW the current first result takes you to the page with this section

It was an uncaffeinated and rushed comment. :-) I was being serious, and I do think folks are probably still googling that (judging by questions in Discord), but the fact that it's just a verbatim paragraph makes it all moot. Either way, thank you for the fix!

PR #2784, opened 2022-01-29
Create and use java.util.ConcurrentThreadLocalRandom on the same flatMap call

### LINKED ISSUE #2766: Random.javaUtilConcurrentThreadLocalRandom not thread-local
I think that `cats.effect.std.Random.javaUtilConcurrentThreadLocalRandom` might call a `ThreadLocalRandom` from the incorrect thread. Its unlikely this would cause any problems on OpenJDK, since there *every* `ThreadLocalRandom` is the "same", but... it's still not allowed, and who knows what other JVM-s might do.

I think this could happen in the following way. Inlining a few things, what it does is like this:
```scala
def nextBoolean: F[Boolean] =
  for {
    r <- Sync[F].delay(new SRandom(java.util.concurrent.ThreadLocalRandom.current()))
    out <- Sync[F].delay(r.nextBoolean())
  } yield out
```
So there is a `flatMap` between the call to `.current()` and (for example) the call to `.nextBoolean()`. And if there is an "autocede" between them, the fiber could be rescheduled to another JVM thread. And then `.nextBoolean()` would be called on another thread's `ThreadLocalRandom`. At least I think this could happen.

PiotrBosak:
also moved around some methods in `ScalaRandom` to keep the same order as in `Random` trait.
Fixes #2766

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Random.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Random.scala
@@ -329,8 +329,7 @@ object Random {
     }
 
   def javaUtilConcurrentThreadLocalRandom[F[_]: Sync]: Random[F] =
-    new ScalaRandom[F](
-      Sync[F].delay(new SRandom(java.util.concurrent.ThreadLocalRandom.current()))) {}
+    new JavaUtilConcurrentThreadLocalRandom[F] {}
 
   def javaSecuritySecureRandom[F[_]: Sync](n: Int): F[Random[F]] =
     for {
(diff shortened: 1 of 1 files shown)


std/shared/src/main/scala/cats/effect/std/Random.scala
```diff
@@ -516,9 +520,47 @@ object Random {
...
-      if (condition) ().pure[F]
-      else new IllegalArgumentException(errorMessage).raiseError[F, Unit]
+  private abstract class ThreadLocalRandom[F[_]: Sync] extends RandomCommon[F] {
+    def nextBoolean: F[Boolean] =
+      Sync[F].delay(new SRandom(java.util.concurrent.ThreadLocalRandom.current()).nextBoolean())
+
+    def nextBytes(n: Int): F[Array[Byte]] = {
+      val bytes = new Array[Byte](0 max n)
+      Sync[F]
+        .delay(new SRandom(java.util.concurrent.ThreadLocalRandom.current()).nextBytes(bytes))
```
armanbilge: Maybe we should factor this out 😅 
```scala
private[this] def localRandom = new SRandom(java.util.concurrent.ThreadLocalRandom.current())
```

armanbilge (commented):
Thanks for the PR :)

std/shared/src/main/scala/cats/effect/std/Random.scala
PiotrBosak: yes, I was thinking about it, but figured the function is side-effectful so I ended up not doing it.
Makes sense though since it's not part of the interface.

armanbilge: Good point, you may want to slap a `()` on that: `localRandom()`.

PiotrBosak: done

armanbilge:
Closing/re-opening to trigger CI on a fresh merge with series/3.3.x.

djspiewak (approved):

djspiewak:
Looks good, thank you!

PR #2785, opened 2022-01-30
Simplify randomness in `TestContext`

vasilmkd:
The idea of this PR is, instead of traversing an iterator and creating an array of up to 10 tasks and then randomly selecting one, this PR adds a random component to each task (just a `Long` value) which is used when comparing two `Task` instances, such that the random component is used when disambiguating tasks which are supposed to run at the same time, effectively enforcing a random ordering of tasks that can run at the same time, while tasks that run at different times still keep their expected temporal ordering. This further allows to simply pick the head task, because the eligible tasks have already been randomized when inserting into the sorted set by the comparator.

This PR further cuts down on unnecessary allocations.

### DIFF
--- a/kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/TestContext.scala
+++ b/kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/TestContext.scala
@@ -326,10 +326,8 @@ object TestContext {
 
         def compare(x: Task, y: Task): Int =
           x.runsAt.compare(y.runsAt) match {
-            case nonZero if nonZero != 0 =>
-              nonZero
-            case _ =>
-              longOrd.compare(x.id, y.id)
+            case 0 => longOrd.compare(x.id + x.rnd, y.id + y.rnd)
+            case nonZero => nonZero
           }
       }
   }
--- a/kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/TestContext.scala
+++ b/kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/TestContext.scala
@@ -71,7 +71,7 @@ final class TestContext private (_seed: Long) extends ExecutionContext { self =>
 
   def execute(r: Runnable): Unit =
     synchronized {
-      stateRef = stateRef.execute(r)
+      stateRef = stateRef.execute(r, random)
     }
 
   def reportFailure(cause: Throwable): Unit =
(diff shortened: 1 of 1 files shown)


kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/TestContext.scala
```diff
@@ -130,24 +133,32 @@ final class TestContext private (_seed: Long) extends ExecutionContext { self =>
...
           stateRef = current.copy(tasks = rest)
-          // execute task
-          try head.task.run()
-          catch { case NonFatal(ex) => reportFailure(ex) }
-          true
+          head
 
         case None =>
-          false
+          null
```
djspiewak: IMO this allocation reduction just isn't worth it. `TestControl` is, by definition, only used in scenarios where performance really isn't necessary. I would prefer to keep the code clarity.

kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/TestContext.scala
```diff
@@ -130,24 +133,32 @@ final class TestContext private (_seed: Long) extends ExecutionContext { self =>
...
-    synchronized {
+  def tickOne(): Boolean = {
+    val head = synchronized {
       val current = stateRef
 
-      // extracting one task by taking the immediate tasks
-      extractOneTask(current, current.clock, random) match {
-        case Some((head, rest)) =>
+      // extracing one task
+      current.tasks.headOption.filter(_.runsAt <= current.clock) match {
```
djspiewak: Shouldn't we invert this?

```suggestion
      current.tasks.filter(_.runsAt <= current.clock).headOption match {
```

I believe that by doing it in the order you do, you're forcing a lot of churn and and failed takes when most tasks are delayed.

djspiewak (changes requested):
I like this idea a lot! A few minor notes and otherwise looks good.

kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/TestContext.scala
vasilmkd: Sorry, I don't quite understand this suggestion. I didn't change the code, it was already like that. The code as it is doesn't traverse the data structure to filter it, while the suggestion does, I would think the latter does strictly more work.

djspiewak: Oh, I think I missed the fact that `tasks` is sorted, so you're guaranteed to get one that you can run if *any* can be run.

kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/TestContext.scala
vasilmkd: Done.

djspiewak (approved):

PR #2787, opened 2022-01-30
Lock-free test context

vasilmkd:
This one is one of those "because it can be done", not sure if it should be done. Opening for CI.

Based on #2785.


djspiewak:
This is a fun one to think about :-) I'm in favor, given that we usually run `TestContext` via `TestControl`, so these `synchronized` blocks are being hit on worker threads.

djspiewak (approved):

PR #2798, opened 2022-02-01
Be more paranoid when making JS Console

armanbilge:
Followup to https://github.com/typelevel/cats-effect/pull/2604?notification_referrer_id=NT_kwDOAC-ZRLIyNzQ3NzkwNzk2OjMxMTk0Mjg#event-5989296410. I don't want a repeat of https://github.com/typelevel/cats-effect/issues/2730.

### DIFF
--- a/std/js/src/main/scala/cats/effect/std/Console.scala
+++ b/std/js/src/main/scala/cats/effect/std/Console.scala
@@ -82,9 +82,10 @@ object Console extends ConsoleCompanionCrossPlatform {
       .flatMap(Option(_))
       .filterNot(js.isUndefined(_))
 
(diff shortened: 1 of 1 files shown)


djspiewak (dismissed):
It surprises me that the browser tests didn't catch this.

armanbilge:
I'm not sure if there actually was a bug here, assuming that `Try` catches exceptions in `filter` (edit: it does). But we've been burned before.

There's always a special case that I forget, which I should figure out how to add to our tests: some browser bundlers do define a `process` global without defining the usual Node.js things _on_ `process`.

djspiewak (approved):

PR #2805, opened 2022-02-02
Add documentation to IO.blocking, IO.interruptible, IO.interruptibleMany

kamilkloch:
Copy & paste of Sync#blocking docs, with small modfications.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
@@ -77,7 +77,7 @@ private[effect] abstract class IOCompanionPlatform { this: IO.type =>
 
   /**
    * Like [[blocking]] but will attempt to abort the blocking operation using thread interrupts
-   * in the event of cancellation. The interrupt will be attempted repeatedly until the blocking
+   * in the event of cancelation. The interrupt will be attempted repeatedly until the blocking
    * operation completes or exits.
    *
    * @param thunk
--- a/core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
@@ -62,7 +62,7 @@ private[effect] abstract class IOCompanionPlatform { this: IO.type =>
 
   /**
    * Like [[blocking]] but will attempt to abort the blocking operation using thread interrupts
-   * in the event of cancellation. The interrupt will be attempted only once.
+   * in the event of cancelation. The interrupt will be attempted only once.
    *
    * @param thunk
    *   The side effect which is to be suspended in `IO` and evaluated on a blocking execution
--- a/core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
@@ -64,6 +64,12 @@ private[effect] abstract class IOCompanionPlatform { this: IO.type =>
    * Like [[blocking]] but will attempt to abort the blocking operation using thread interrupts
    * in the event of cancelation. The interrupt will be attempted only once.
    *
+   * Note the following tradeoffs:
+   *  - this has slightly more overhead than [[blocking]] due to the machinery necessary for the interrupt coordination,
+   *  - thread interrupts are very often poorly considered by Java (and Scala!) library authors, and it is possible
+   *    for interrupts to result in resource leaks or invalid states. It is important to be certain that this
+   *    will not be the case before using this mechanism.
+   *
    * @param thunk
    *   The side effect which is to be suspended in `IO` and evaluated on a blocking execution
    *   context
@@ -79,6 +85,12 @@ private[effect] abstract class IOCompanionPlatform { this: IO.type =>
    * Like [[blocking]] but will attempt to abort the blocking operation using thread interrupts
    * in the event of cancelation. The interrupt will be attempted repeatedly until the blocking
    * operation completes or exits.
+   *
+   * Note the following tradeoffs:
+   *  - this has slightly more overhead than [[blocking]] due to the machinery necessary for the interrupt coordination,
+   *  - thread interrupts are very often poorly considered by Java (and Scala!) library authors, and it is possible
+   *    for interrupts to result in resource leaks or invalid states. It is important to be certain that this
+   *    will not be the case before using this mechanism.
    *
    * @param thunk
    *   The side effect which is to be suspended in `IO` and evaluated on a blocking execution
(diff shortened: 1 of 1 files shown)


core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
```diff
@@ -28,6 +28,23 @@ private[effect] abstract class IOCompanionPlatform { this: IO.type =>
   private[this] val TypeInterruptibleOnce = Sync.Type.InterruptibleOnce
   private[this] val TypeInterruptibleMany = Sync.Type.InterruptibleMany
 
+  /**
+   * Intended for thread blocking operations. `blocking` will shift the execution of the
+   * blocking operation to a separate threadpool to avoid blocking on the main execution
+   * context. See the thread-model documentation for more information on why this is necessary.
+   * Note that the created effect will be uncancelable; if you need cancellation then you should
+   * use [[interruptible]] or [[interruptibleMany]].
```
armanbilge: Unfortunately it seems scaladoc doesn't like `[[interruptible]]` b/c there's two methods named `interruptible` and it doesn't know which to link.

vasilmkd: There is a way to disambiguate but I'm never sure how to do it exactly. I think this might be helpful and there should also exist a page for Scala 2 but I cannot find it at the moment.

https://docs.scala-lang.org/scala3/guides/scaladoc/linking.html

core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
```diff
@@ -28,6 +28,23 @@ private[effect] abstract class IOCompanionPlatform { this: IO.type =>
   private[this] val TypeInterruptibleOnce = Sync.Type.InterruptibleOnce
   private[this] val TypeInterruptibleMany = Sync.Type.InterruptibleMany
 
+  /**
+   * Intended for thread blocking operations. `blocking` will shift the execution of the
+   * blocking operation to a separate threadpool to avoid blocking on the main execution
+   * context. See the thread-model documentation for more information on why this is necessary.
+   * Note that the created effect will be uncancelable; if you need cancellation then you should
```
djspiewak: Minor nit, but we tend to use the alternative spelling, "cancelation".

```suggestion
   * Note that the created effect will be uncancelable; if you need cancelation then you should
```

core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
```diff
@@ -43,11 +60,32 @@ private[effect] abstract class IOCompanionPlatform { this: IO.type =>
       Tracing.calculateTracingEvent(fn.getClass))
   }
 
+  /**
+   * Like [[blocking]] but will attempt to abort the blocking operation using thread interrupts
+   * in the event of cancellation. The interrupt will be attempted only once.
```
djspiewak:
```suggestion
   * in the event of cancelation. The interrupt will be attempted only once.
```

core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
```diff
@@ -43,11 +60,32 @@ private[effect] abstract class IOCompanionPlatform { this: IO.type =>
       Tracing.calculateTracingEvent(fn.getClass))
   }
 
+  /**
+   * Like [[blocking]] but will attempt to abort the blocking operation using thread interrupts
```
djspiewak, line 64: It might be worth tossing some notes on here about the tradeoffs. Specifically, I would note the following two caveats:

- This has slightly more overhead than `blocking` due to the machinery necessary for the interrupt coordination
- Thread interrupts are very often poorly considered by Java (and Scala!) library authors, and it is possible for interrupts to result in resource leaks or invalid states. It's important to be certain that this will *not* be the case before using this mechanism.

core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
```diff
@@ -43,11 +60,32 @@ private[effect] abstract class IOCompanionPlatform { this: IO.type =>
...
+   */
   def interruptible[A](thunk: => A): IO[A] = {
     val fn = Thunk.asFunction0(thunk)
     Blocking(TypeInterruptibleOnce, fn, Tracing.calculateTracingEvent(fn.getClass))
   }
 
+  /**
+   * Like [[blocking]] but will attempt to abort the blocking operation using thread interrupts
+   * in the event of cancellation. The interrupt will be attempted repeatedly until the blocking
+   * operation completes or exits.
```
djspiewak:
```suggestion
   * in the event of cancelation. The interrupt will be attempted repeatedly until the blocking
   * operation completes or exits.
```

I might also say a few words about *why* this exists. In particular, noting that library authors tend to handle interruption incorrectly, and they may set up a situation where they suppress an interrupt unintentionally (e.g. by catching `Throwable` or even `Exception`) and then resume their work. Repeatedly interrupting is one workaround for this situation, though it does come at a cost, since a *second* blocking thread must be tied down in a hot `while` loop just calling `interrupt()` repeatedly, so the overhead of this effect is relatively high and should only be used in situations where there are no other options.

djspiewak (changes requested):
This looks like a really good start, thank you!

Once the `IO` docs are done, can we copy them over to `Sync` as well? Also IIRC there's a way to inherit scaladoc between members to avoid redundancy, but I can't remember how that works exactly.

armanbilge:
> Also IIRC there's a way to inherit scaladoc between members to avoid redundancy, but I can't remember how that works exactly.

I think we could take advantage of this if the companion objects for `IO` and `SyncIO` inherited directly from the `Async` and `Sync` typeclasses. We'd also get a lot of currently duplicated methods for free.

core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
kamilkloch: I tried `[[interruptible[A](=>A):IO[A]*]]`, but it did not help. I give up.

kamilkloch: @vasilmkd Could you check if this works in VSCode?  Here is the scaladoc cheatsheeet: https://github.com/scala/scala/blob/2.12.x/test/scaladoc/resources/links.scala

core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
kamilkloch: @djspiewak I basically added the above comment, please take a look.

kamilkloch:
> > Also IIRC there's a way to inherit scaladoc between members to avoid redundancy, but I can't remember how that works exactly.
> 
> I think we could take advantage of this if the companion objects for `IO` and `SyncIO` inherited directly from the `Async` and `Sync` typeclasses. We'd also get a lot of currently duplicated methods for free.

A great idea, in general. The burden of syncing documentation across two separate locations would go away.

djspiewak (approved):

PR #2807, opened 2022-02-05
Enable configuration of non-daemon thread logging within `IOApp`

djspiewak:
Brings it more in line with the other configuration options.


build.sbt
```diff
@@ -457,7 +457,16 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
...
+      ProblemFilters.exclude[MissingClassProblem]("cats.effect.unsafe.WorkerThread$Data"),
+      // introduced by #4248
+      // changes to package private code
+      ProblemFilters.exclude[DirectMissingMethodProblem](
+        "cats.effect.NonDaemonThreadLogger.isEnabled"),
+      ProblemFilters.exclude[DirectMissingMethodProblem](
+        "cats.effect.NonDaemonThreadLogger.sleepIntervalMillis"),
+      ProblemFilters.exclude[DirectMissingMethodProblem](
+        "cats.effect.NonDaemonThreadLogger.this"),
+      ProblemFilters.exclude[MissingClassProblem]("cats.effect.NonDaemonThreadLogger$")
```
armanbilge, line 536: `#2807` actually?

Btw are these Scala 3 specific exclusions? MiMa doesn't recognize package-privates yet on Scala 3, but it will in the next release at which point we can remove them.

djspiewak: Oh whoops. Yeah, 2807

They might be scala 3 specific. I'll check that out.

sbly (dismissed):
Just reviewing PRs to try to understand the codebase better. This one looks good to me!

PR #2808, opened 2022-02-05
Skip tracing tests on WSLv2 with JavaScript

djspiewak:


tests/jvm/src/test/scala/cats/effect/DetectPlatform.scala
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+
+trait DetectPlatform {
+  def isWSL: Boolean = System.getProperty("os.version").contains("-WSL")
+  def isJS: Boolean = true
```
armanbilge: 😝 

```suggestion
  def isJS: Boolean = false
```

djspiewak: Oh my…

armanbilge (approved):

Issue #2771, opened 2022-01-22
High-performance `Queue` implementation


djspiewak:
We can up-level the performance on `Queue` by about an order of magnitude by taking an `Async` constraint instead of `Concurrent`. However, the ecosystem relies on being able to use `Concurrent` since it's a much weaker constraint. Thus, we need to do a `match` on the runtime type of the typeclass. If it's `Concurrent`, we use the existing implementation; if it's `Async`, we use the higher performance one.

This level of shenanigans is worthwhile since `Queue` is so widely used within the critical path of most ecosystem libraries.

PR #2809, opened 2022-02-05
Refine derived type classes

### LINKED ISSUE #2772: Special-case more specific runtime types in derived typeclasses
In each of the derived typeclasses (e.g. `Concurrent` for `EitherT`), we should `match` on the runtime type of the `F` constraint and try to provide the more specific runtime type if at all possible (e.g. if the `F` constraint is `Async` and not `Concurrent`, then we should instantiate an `Async` for `EitherT` and return *that*, preserving the runtime specificity). This is needed to ensure that #2771 is fairly robust.

ybasket:
When deriving type class instances for monad transformers, pattern match at runtime to find out the strongest type class instance that can be derived and return this one instead.

Open questions: 
1. Do `Clock` and `Unique` need the same treatment? Currently they don't get it (why would you want an `Async` if you ask for `Clock` other than general consistency?). `Unique` doesn't even define instances for monad transformers directly. But I might overlook valid use cases.
2. When pattern matching (let's say for `OptionT` in `GenSpawn`), a naive version would call `GenTemporal.genTemporalForOptionT` if it detects a runtime type of `GenTemporal`. But as `GenTemporal.genTemporalForOptionT` itself also pattern matches to see whether it's actually an `Async`, there'd be two pattern matches happening before the actual instance is created. To avoid this, this PR introduces methods like `GenTemporal.instantiateGenTemporalForOptionT` that don't pattern match, but directly instantiate. So the derivation in `GenSpawn` that pattern matched already can use this to avoid the second pattern match from happening (this duplication would happen for every transformer in a stack, so it's a bigger problem with huge stacks). So one method call is added – is that a performance problem? One could easily inline all the `instantiateXYZ` methods, but that leads to code duplication and (if I'm not mistaken) to more anonymous classes being created. Opinions on what's the better tradeoff here?

Closes #2772

### DIFF
--- a/tests/shared/src/test/scala/cats/effect/kernel/DerivationRefinementSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/kernel/DerivationRefinementSpec.scala
@@ -1,0 +1,109 @@
+/*
+ * Copyright 2020-2022 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.kernel
+
+import cats.data._
+import cats.effect.{BaseSpec, IO, SyncIO}
+import org.specs2.matcher.Matchers
+
+import scala.reflect.ClassTag
+
+class DerivationRefinementSpec extends BaseSpec with Matchers {
+
+  type AsyncStack[F[_], A] = Kleisli[OptionT[EitherT[IorT[F, Int, *], String, *], *], Unit, A]
+  type SyncStack[F[_], A] = StateT[ReaderWriterStateT[F, String, Int, Unit, *], Boolean, A]
+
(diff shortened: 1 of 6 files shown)


tests/shared/src/test/scala/cats/effect/kernel/DerivationRefinementSpec.scala
```diff
@@ -0,0 +1,109 @@
...
+
+package cats.effect.kernel
+
+import cats.data._
+import cats.effect.{BaseSpec, IO, SyncIO}
+import org.specs2.matcher.Matchers
+
+import scala.reflect.ClassTag
+
+class DerivationRefinementSpec extends BaseSpec with Matchers {
```
ybasket, line 26: This test covers all monad transformers, but only `Sync` and `Async` as "target" type classes of the refinement as it's hard (boilerplate) to create instances of `Temporal` and alike that aren't stronger at runtime. At least on Scala 2, on Scala 3 this can probably be solved using `export` clauses. Would you want a Scala 3-only test that covers more derivation paths than this one or is the added complexity by having a Scala 3 only test not worth it?

ybasket: I tried the `export` trick, it works quite nicely. Scala 3-only source files with imports require some scalafix hacking though, so I didn't add it to this PR. Code can be inspected at https://github.com/ybasket/cats-effect/pull/1/files, happy to take the commit into this PR or create a follow-up if people consider it valuable.

djspiewak:
> Unique doesn't even define instances for monad transformers directly

Ah, this is an interesting oversight!

Thank you for taking this on! Will review shortly. This is really awesome and it's the other important half of #2771

djspiewak (approved):
This is all excellent :-) Thank you!

PR #2810, opened 2022-02-06
Introduce `GenDeferred`

armanbilge:
This PR introduces `GenDeferred` as a generalization of `Deferred`:

```scala
trait GenDeferred[F[_], G[_], A] extends DeferredSource[F, A] with DeferredSink[G, A]
```

Conceptually, this enables two independent effect systems to communicate/synchronize via a shared data structure.

In practice, there are often situations where it is useful to instantiate a `GenDeferred` where `F` is any `Async` type and `G` is `SyncIO` when interopping with unsafe code. This makes it possible to complete the deferred "in-place" via `SyncIO#unsafeRunSync` without requiring a `Dispatcher` and avoiding a thread-shift to the compute pool to run this trivial operation.

I would like to follow-up with a similar `GenQueue[F, G, A]` such that an unbounded queue can be implemented for `Async[F]` and `Sync[G]`.

Example use-cases include fs2-grpc (cc @ahjohannessen) and fs2-io.js which both rely heavily on `Deferred` and `Queue` to interop with unsafe code.

Credit to @ChristopherDavenport for first exploring this idea in https://github.com/davenverse/condemned.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -299,6 +299,16 @@ lazy val kernel = crossProject(JSPlatform, JVMPlatform)
   .jsSettings(
     libraryDependencies += "org.scala-js" %%% "scala-js-macrotask-executor" % MacrotaskExecutorVersion % Test
   )
+  .settings(
+    mimaBinaryIssueFilters ++= Seq(
+      ProblemFilters.exclude[MissingClassProblem]("cats.effect.kernel.Deferred$State"),
+      ProblemFilters.exclude[MissingClassProblem]("cats.effect.kernel.Deferred$State$"),
+      ProblemFilters.exclude[MissingClassProblem]("cats.effect.kernel.Deferred$State$Set"),
+      ProblemFilters.exclude[MissingClassProblem]("cats.effect.kernel.Deferred$State$Set$"),
+      ProblemFilters.exclude[MissingClassProblem]("cats.effect.kernel.Deferred$State$Unset"),
+      ProblemFilters.exclude[MissingClassProblem]("cats.effect.kernel.Deferred$State$Unset$")
+    )
+  )
 
 /**
  * Reference implementations (including a pure ConcurrentBracket), generic ScalaCheck
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Deferred.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Deferred.scala
@@ -46,7 +46,7 @@ import java.util.concurrent.atomic.AtomicReference
  * Finally, the blocking mentioned above is semantic only, no actual threads are blocked by the
  * implementation.
  */
-abstract class Deferred[F[_], A] extends DeferredSource[F, A] with DeferredSink[F, A] {
+abstract class Deferred[F[_], A] extends GenDeferred[F, F, A] {
 
   /**
    * Modify the context `F` using transformation `f`.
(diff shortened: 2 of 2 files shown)


durban:
I didn't look too closely, but I think I've already did this in #1583 (it wasn't merged).

armanbilge:
@durban thanks for chiming in, yes the concept looks the same :) apologies that I wasn't aware of your prior art.

One year on, with at least two applications, I wonder if opinions have changed? 😅

ahjohannessen:
I hope this and the follow up could land in cats-effect.
Library authors apparently have a need for something like this. I mean @armanbilge does this, @ChristopherDavenport makes condemned and fs2-grpc could use this.

durban:
> One year on, with at least two applications, I wonder if opinions have changed?

My opinion haven't changed: I still think it's a good idea :smile: (Although as it turned out, my original use case required even more than this...)

bplommer:
Chiming in with another closed PR in the same vicinity: https://github.com/typelevel/cats-effect/pull/1889 is a bit unfocused, but among other things it takes a slightly different approach of just letting you change the effect type of an existing `SyncDeferred`/`AsyncDeferred` to an arbitrary effect with the relevant instance. It looks like your approach is simpler to implement though 😅

armanbilge:
Wow, another one! Thanks for the pointer. It seems you were actually able to execute the changes for `Queue` as well which is awesome since I shied away from it here. Also Daniel seemed _slightly_ less skeptical in that PR which is a big win in of itself 😉

bplommer:
> It seems you were actually able to execute the changes for `Queue` as well which is awesome since I shied away from it here.

Ah yes - not sure whether the `Deferred` part is any more complex than your version 🤔

armanbilge:
Yes, I held off since we are getting a real `AsyncQueue` as part of https://github.com/typelevel/cats-effect/issues/2771. So it doesn't make sense to rework the current `Concurrent` implementation.

SystemFw:
Quick comment I've already mentioned on Discord: I don't have issues with the idea, but I'd rather we didn't have the subtyping hierarchy

armanbilge:
Big thanks to Daniel for patiently walking me through _exactly_ what happens when you complete a `Deferred` and why doing this outside of the compute pool (i.e., not via `Dispatcher`) can actually make things _worse_.

A key part of completing a `Deferred` is notifying any waiting readers.

https://github.com/typelevel/cats-effect/blob/f423ac754fecdf563cc60b4ed4800374ff80c251/kernel/shared/src/main/scala/cats/effect/kernel/Deferred.scala#L157-L158

To do so, you fulfill a callback `next` with the completed value `a`, _for each reader_.

https://github.com/typelevel/cats-effect/blob/f423ac754fecdf563cc60b4ed4800374ff80c251/kernel/shared/src/main/scala/cats/effect/kernel/Deferred.scala#L167

The callback function itself is ultimately defined here.

https://github.com/typelevel/cats-effect/blob/f423ac754fecdf563cc60b4ed4800374ff80c251/core/shared/src/main/scala/cats/effect/IOFiber.scala#L626

Which involves scheduling the semantically-blocked fiber in question.

https://github.com/typelevel/cats-effect/blob/f423ac754fecdf563cc60b4ed4800374ff80c251/core/shared/src/main/scala/cats/effect/IOFiber.scala#L667

Which takes us to here.

https://github.com/typelevel/cats-effect/blob/f423ac754fecdf563cc60b4ed4800374ff80c251/core/shared/src/main/scala/cats/effect/IOFiber.scala#L1284-L1287

And finally, to here (assuming our fiber is indeed assigned to the WSTP).

https://github.com/typelevel/cats-effect/blob/f423ac754fecdf563cc60b4ed4800374ff80c251/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala#L418-L431

And suddenly, it's obvious why it makes a _huge_ difference whether the current thread we are running on belongs to the compute pool (where `Dispatcher` places us) _or_ we are completing these callbacks from an external thread. It's the difference between scheduling _on the same worker thread_ vs submitting to the WSTP external queue.

In summary: the context shift is _fundamental_ as Daniel has been saying all along. So, better to get on the compute pool ASAP.

Sadly, this still leaves JavaScript in an awkward place since there is no WSTP-fastpath. The initial `yield` is wasteful since it's not taking us to a better place like it does on JVM. But I'm not sure if that's good enough reason to muck up the hiearchy with this stuff.

_Note: edited to clarify the final hop to actually scheduling on the WSTP._

djspiewak:
@armanbilge What if we have a JS specialization of `Dispatcher`?

armanbilge:
Sure, but what would it do / how would it work?

armanbilge:
Trying a different idea in https://github.com/typelevel/cats-effect/pull/2835.

PR #2811, opened 2022-02-07
Support shutting down multiple runtimes

vasilmkd:
This PR does many things.

1. Does #2661 properly. All runtimes can now be tracked through MBeans.
2. Adds support for shutting down _all_ `IORuntime` instances on a fatal error, not just the runtime that owns the fiber that failed with a fatal error.
3. Through the same mechanism gets rid of the `self: => IORuntime` forward reference in the WSTP. For this however, #2773 needs to be reverted first and done again.

### DIFF
--- a/benchmarks/src/main/scala/cats/effect/benchmarks/WorkStealingBenchmark.scala
+++ b/benchmarks/src/main/scala/cats/effect/benchmarks/WorkStealingBenchmark.scala
@@ -175,7 +175,7 @@ class WorkStealingBenchmark {
       (Scheduler.fromScheduledExecutor(executor), () => executor.shutdown())
     }
 
-    val compute = new WorkStealingThreadPool(256, "io-compute", manyThreadsRuntime)
+    val compute = new WorkStealingThreadPool(256, "io-compute")
 
     val cancelationCheckThreshold =
       System.getProperty("cats.effect.cancelation.check.threshold", "512").toInt
--- a/build.sbt
+++ b/build.sbt
@@ -468,7 +468,23 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
           ProblemFilters.exclude[DirectMissingMethodProblem](
             "cats.effect.unsafe.WorkStealingThreadPool.localQueuesForwarder"),
           ProblemFilters.exclude[DirectMissingMethodProblem](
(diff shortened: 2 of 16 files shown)


vasilmkd:
This breaks the static forwarders to the old default methods. I would appreciate any guidance on how to fix this, if it is even possible.

armanbilge:
Is this only on Scala 3, or in general?

vasilmkd:
If we opt for deprecation and a new method with a new name instead, everything is fine of course.

armanbilge:
Oh, you meant methods with `default` in the name, not methods with default parameters 😆

Yeah, there's no way to retain the static forwarders without keeping it public.

vasilmkd:
> Oh, you meant methods with `default` in the name, not methods with default parameters 😆

Actually both. 😆

armanbilge:
Btw, seems to me you could keep the old name and opt for overload?

vasilmkd:
I just tried it, the new method cannot have default parameters if we wish to keep static forwarder binary compatibility.

armanbilge:
Ah yeah, forgot about those. Anyway overloads aren't always a good idea 😬

armanbilge:
Here's how you can do it without compromising bincompat 😉 https://github.com/typelevel/cats-effect/commit/9710e21ce18f16ce0b913e0be0f30c315493c967

vasilmkd:
Can the method you added be `private[unsafe]`. I'm guessing yes? If yes, let's do it.

armanbilge:
Nope 😕 it's static forwarders again. Although, I'd argue removing this static forwarder is less likely to be harmful. It's very plausible that a Java user might call `createDefaultComputeThreadPool` but less plausible that they would call `createDefaultComputeThreadPool$default$3`.

```
[error]  * synthetic static method createDefaultComputeThreadPool$default$3()java.lang.String in class cats.effect.unsafe.IORuntime does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("cats.effect.unsafe.IORuntime.createDefaultComputeThreadPool$default$3")
```

vasilmkd:
Not necessarily all the way there (in terms of API), requesting review for comments. Features are finalized.

djspiewak (approved):

PR #2815, opened 2022-02-08
Remove duplicate documentation page

Daenyth:
Keep the better one

I'd also like to get a section on the IOApp fields that can be overridden, but we can enhance that after


site-docs/sidebars.json
```diff
@@ -12,7 +12,6 @@
       "tracing",
       "thread-model",
       "schedulers",
-      "scaling-and-tuning/io-runtime-config",
```
armanbilge, line 15: Actually I think we want to keep this one...

site-docs/sidebars.json
```diff
@@ -43,8 +42,8 @@
       "std/supervisor",
       "std/async-await"
     ],
-    "Scaling and Tuning":[
-      "scaling-and-tuning/system-properties"
+    "Scaling and Tuning": [
+      "scaling-and-tuning/io-runtime-config"
     ]
```
armanbilge, line 47: ... and remove this one :)

armanbilge (commented):
Thanks for chasing this up! Just a minor thing.

site-docs/sidebars.json
armanbilge: In fact, it can probably be moved to that same directory as the other core runtime stuff.

vasilmkd:
Gentle ping. @Daenyth 😆

PR #2819, opened 2022-02-10
Put instances for `IO.Par` into implicit scope

### LINKED ISSUE #2818: CommutativeApplicative[IO.Par] not available by default
```
import cats.effect.instances.all.commutativeApplicativeForParallelF
```
is required for 
```
List(1,2,3).parUnorderedTraverse(_ => IO.unit)
```
to compile, which is probably fixable as long as `c.e.IO` is concerned and not a generic `F[_]`

Creating issue per to discord request :)

armanbilge:
Fixes https://github.com/typelevel/cats-effect/issues/2818.


armanbilge:
Interesting, seems like this doesn't work in 2.12.

build.sbt
```diff
@@ -515,7 +515,13 @@ lazy val tests: CrossProject = crossProject(JSPlatform, JVMPlatform)
...
       "org.typelevel" %%% "cats-kernel-laws" % CatsVersion % Test
     ),
-    buildInfoPackage := "catseffect"
+    buildInfoPackage := "catseffect",
+    Test / unmanagedSourceDirectories ++= {
+      if (scalaBinaryVersion.value != "2.12")
+        Seq(baseDirectory.value / ".." / "shared" / "src" / "test" / "scala-2.13+")
+      else
+        Seq.empty
+    }
```
armanbilge, line 524: 😩

vasilmkd: I'm personally fine with duplicating code in `scala-2.13` and `scala-3` as separate directories, which doesn't need unmanaged source directories.

vasilmkd: Especially for tests, not worth it.

armanbilge: Yeah, me too. Too easy for a test not to compile/run without realizing.

I was thinking we could also symlink it.

tests/shared/src/test/scala-2.13/cats/effect/IOParImplicitSpec.scala
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+package not.cats.effect // verifies scoping outside of CE
+
+import cats.{Align, CommutativeApplicative}
+import cats.effect.{BaseSpec, IO}
+import cats.syntax.all._
+
+// Note: this file is duplicated across 2.13/3 source directories
```
djspiewak: I'd rather see it symlinked, tbh, or go back to the `unmanagedSourceDirectories` thing from the original changeset. It would break compilation entirely on non-WSL Windows, but I'm cool with that. These kinds of copy/paste duplications inevitably bite in the long run, by which we have long since forgotten why we did it in the first place.

djspiewak (changes requested):

tests/shared/src/test/scala-2.13/cats/effect/IOParImplicitSpec.scala
armanbilge: Reverted since I already did the work.

djspiewak (approved):

PR #2827, opened 2022-02-16
Add `IOLocal.unsafeCreate`

bplommer:
The use case I have in mind is to be able to create a global instance for tracing purposes:

```
type TraceCtx
object TraceCtx {
  def empty: TraceCtx = ???
  val ioLocalInstance: IOLocal[TraceCtx] = IOLocal.unsafeCreate(empty)
}
```

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOLocal.scala
+++ b/core/shared/src/main/scala/cats/effect/IOLocal.scala
@@ -35,35 +35,33 @@ sealed trait IOLocal[A] {
 }
 
 object IOLocal {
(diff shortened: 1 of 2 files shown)


armanbilge:
Heh, my turn to point you to an old closed PR:
- https://github.com/typelevel/cats-effect/pull/2482

Btw, I closed it because I realized you can cheat via `syncStep` and `unsafeRunSync`.

Edit: also, cheating on the RT did totally bite me in the end 🙃

bplommer:
Interesting. I also have another WIP that takes an alternative approach, getting RT by requiring the user to pass a key. I'll push it and see what people think.

PR #2828, opened 2022-02-16
Non-effectful `IOLocal` allocation based on a user-supplied key

bplommer:
This is intended to allow instantiating `IOLocal` in a non-effectful but referentially transparent way.

The diff is ugly 😕

An alternative to #2827

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOLocal.scala
+++ b/core/shared/src/main/scala/cats/effect/IOLocal.scala
@@ -17,6 +17,7 @@
 package cats.effect
 
 import cats.kernel.Hash
+
 import scala.reflect.ClassTag
 
 sealed trait IOLocal[A] {
--- a/core/shared/src/main/scala/cats/effect/IOLocal.scala
+++ b/core/shared/src/main/scala/cats/effect/IOLocal.scala
@@ -16,6 +16,9 @@
 
 package cats.effect
 
+import cats.kernel.Eq
+import scala.reflect.ClassTag
+
 sealed trait IOLocal[A] {
 
   def get: IO[A]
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/effect/IOLocal.scala
```diff
@@ -37,33 +40,97 @@ sealed trait IOLocal[A] {
...
-        override def get: IO[A] =
-          IO.Local(state => (state, state.get(self).map(_.asInstanceOf[A]).getOrElse(default)))
+    IO(new SelfKeyedIOLocal[A](default))
+
+  /**
+   * Creates an [[IOLocal]] instance keyed on `k`. Two instances of `IOLocal` created with equal
+   * keys will operate the same state when their operations are sequenced together.
+   *
+   * Warning: calling this twice with equal keys but different value types will result in
+   * runtime errors.
```
armanbilge, line 51: I think this is the problem that `Vault` solves.
https://github.com/typelevel/vault/blob/34aed00f00394f3c525e57f258a1b5e607f93266/core/src/main/scala/org/typelevel/vault/Key.scala#L112

bplommer: Sure but that's effectful. If you're happy to allocate effectfully you can just do `apply`.

armanbilge: I could be confused, but people `unsafeRunSync` on vault pretty much everywhere.

bplommer: Okay but then we lose RT? 

Are you suggesting we should change `IOLocal` to be backed by `Vault` or something vault-like?

bplommer: There's definitely refinement to be done on the implementation, I guess the main thing at this stage is to get feedback on the concept.

armanbilge: Oh, right. I _am_ confused, sorry :)

IMHO loss of type-safety is much worse than loss of RT.

bplommer: How about requiring a classtag like with `forClass` and keying on the combination of key and class?

... oh, crap, class tags aren't going to be enough for type safety, are they. And typetags don't exist in scala 3. Hmm 😕

bplommer: > IMHO loss of type-safety is much worse than loss of RT.

I’m not sure I agree that one is inherently worse than the other. It seems reasonable to allow users to choose which one is provided and which one they have to take care of themselves.

armanbilge: I'm not sure what this PR wants to do is possible. The other PR is more viable (esp. in the form of a `Sync` constructor which is completely safe). If your goal is the same as stated in the other PR, I really think the answer is `syncStep` to get a `SyncIO` and then `unsafeRunSync` on that. It will work, but is annoying enough to remind you what you are doing :)

armanbilge: > I’m not sure I agree that one is inherently worse than the other.

Well, my reasoning is that we expect type-safety to be enforced by the compiler, but not RT (that's on us to enforce). Which is why lack of type-safety seems so much more dangerous.

bplommer: Yeah, I think you're right. Closing this.

PR #2829, opened 2022-02-16
Indicate a completed state in IOFiber#toString

### LINKED ISSUE #2628: Indicate if fiber is completed in dump/toString
If a fiber hasn't been GCed (either due to non-determinism or some strong reference) or we are directly calling `toString` on it, we should indicate if it is complete.

iRevive:
My naive implementation of https://github.com/typelevel/cats-effect/issues/2628. 

When fiber is completed, the stack trace events are not available anymore. 
Technically, there is room for premature optimization:
```scala
val hash = System.identityHashCode(this).toHexString
if (isDone) {
  s"cats.effect.IOFiber@$hash COMPLETED"
} else {
  val state = if (suspended.get()) "SUSPENDED" else "RUNNING"

  val opAndCallSite =
    Tracing.getFrames(tracingEvents).headOption.map(frame => s": $frame").getOrElse("")

   s"cats.effect.IOFiber@$hash $state$opAndCallSite"
}
```

But I don't think it's worth it.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -1518,7 +1518,7 @@ private final class IOFiber[A] private (
 
   // overrides the AtomicReference#toString
   override def toString: String = {
-    val state = if (suspended.get()) "SUSPENDED" else "RUNNING"
+    val state = if (suspended.get()) "SUSPENDED" else if (isDone) "COMPLETED" else "RUNNING"
     val tracingEvents = this.tracingEvents
 
     // There are race conditions here since a running fiber is writing to `tracingEvents`,
@@ -1529,6 +1529,9 @@ private final class IOFiber[A] private (
 
     s"cats.effect.IOFiber@${System.identityHashCode(this).toHexString} $state$opAndCallSite"
   }
+
+  private[this] def isDone: Boolean =
+    resumeTag == DoneR
 
   private[effect] def prettyPrintTrace(): String =
     if (isStackTracing) {
--- a/tests/shared/src/test/scala/cats/effect/IOFiberSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/IOFiberSpec.scala
@@ -50,5 +50,18 @@ class IOFiberSpec extends BaseSpec {
         _ <- IO(s must beMatching(pattern))
       } yield ok
     }
+
+    "toString a completed fiber" in real {
+      def done = IO.unit.start
+      val pattern = raw"cats.effect.IOFiber@[0-9a-f][0-9a-f]+ COMPLETED"
+      for {
+        f <- done.start
+        _ <- IO.sleep(1.milli)
+        s <- IO(f.toString)
+        // _ <- IO.println(s)
+        _ <- f.cancel
+        _ <- IO(s must beMatching(pattern))
+      } yield ok
+    }
   }
 }


core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -1530,6 +1530,9 @@ private final class IOFiber[A] private (
     s"cats.effect.IOFiber@${System.identityHashCode(this).toHexString} $state$opAndCallSite"
   }
 
+  private def isDone: Boolean =
```
djspiewak: Minor nit: `private[this]` would be preferred for this type of thing, since it will inline better.

djspiewak (changes requested):
This looks super-great! Aside from the minor nit, I think it's ready to go!

armanbilge:
In the original issue I made a note https://github.com/typelevel/cats-effect/issues/2628#issuecomment-986169842 that this needs to be changed in a couple places, besides `toString`.

I think for this change to take effect for fiber dumps we need to change it somewhere in here too:
https://github.com/typelevel/cats-effect/blob/f423ac754fecdf563cc60b4ed4800374ff80c251/core/jvm/src/main/scala/cats/effect/unsafe/FiberMonitor.scala#L93

And also for JS.

armanbilge:
Actually, maybe the change should be here:
https://github.com/typelevel/cats-effect/blob/f423ac754fecdf563cc60b4ed4800374ff80c251/core/shared/src/main/scala/cats/effect/unsafe/FiberMonitorShared.scala#L26-L31

If the fiber is done it can ignore the provided status `String`.

djspiewak:
@armanbilge Don't we already have something somewhere which puts the fiber status onto dumps? I thought we had it elsewhere because the fiber dump capability is able to more accurately determine *what* a fiber is doing than the fiber itself can.

armanbilge:
> I thought we had it elsewhere because the fiber dump capability is able to more accurately determine what a fiber is doing than the fiber itself can.

Yes, that's exactly why we have a separate `fiberString` that takes a `status` as an argument. However, that currently does not take into account whether the fiber is done or not.

djspiewak:
> Yes, that's exactly why we have a separate fiberString that takes a status as an argument. However, that currently does not take into account whether the fiber is done or not.

Ah! Yes, I'd like to see that improved.

iRevive:
Thanks for the details. I will implement mentioned suggestions

armanbilge:
Another possibility I didn't consider this morning is that we should filter out the fibers that are done, and not include them in the dump at all (either they haven't been GCed yet, or there's a strong reference). We should see what @djspiewak thinks.

Edit: since the trace is no longer available, filtering them out is almost definitely the right thing to do actually. Since otherwise they are just taking space without much information :)

djspiewak:
> Another possibility I didn't consider this morning is that we should filter out the fibers that are done, and not include them in the dump at all (either they haven't been GCed yet, or there's a strong reference). We should see what @djspiewak thinks.

This is a good point. IMO yes, a done fiber shouldn't be displayed in the dump.

djspiewak (approved):

armanbilge:
This PR still needed some follow-up work, I'll re-open the issue.

armanbilge:
Also, this could probably go into 3.3.x, no?

djspiewak:
> Also, this could probably go into 3.3.x, no?

Probably yes, now that I think about it. We would need to revert this though, PR the revert, then re-PR the change with a new base. Doesn't seem worth it. We can just let this fall through to 3.4.0.

PR #2834, opened 2022-02-20
Fixed console example

iamonzon:
Added Monad constraint and removed extra implicit

### DIFF
--- a/std/js/src/main/scala/cats/effect/std/Console.scala
+++ b/std/js/src/main/scala/cats/effect/std/Console.scala
@@ -36,7 +36,7 @@ import java.nio.charset.Charset
  *  import cats.effect.std.Console
  *  import cats.syntax.all._
  *
- *  def myProgram[F[_] : Console : Monad]: F[Unit] =
+ *  def myProgram[F[_]: Console: Monad]: F[Unit] =
  *    for {
  *      _ <- Console[F].println("Please enter your name: ")
  *      n <- Console[F].readLine
--- a/std/js/src/main/scala/cats/effect/std/Console.scala
+++ b/std/js/src/main/scala/cats/effect/std/Console.scala
@@ -32,18 +32,16 @@ import java.nio.charset.Charset
  *
  * @example
  *   {{{
- * import cats.effect.std.Console
- * import cats.effect.kernel.Sync
- * import cats.syntax.all._
+ *  import cats.Monad
+ *  import cats.effect.std.Console
+ *  import cats.syntax.all._
  *
- * implicit val console = Console.sync[F]
- *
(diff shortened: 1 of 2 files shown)


std/js/src/main/scala/cats/effect/std/Console.scala
```diff
@@ -32,18 +32,16 @@ import java.nio.charset.Charset
...
  *
- * implicit val console = Console.sync[F]
- *
- * def myProgram[F[_]: Console]: F[Unit] =
- *   for {
- *     _ <- Console[F].println("Please enter your name: ")
- *     n <- Console[F].readLine
- *     _ <- if (n.nonEmpty) Console[F].println("Hello, " + n) else Console[F].errorln("Name is empty!")
- *   } yield ()
+ *  def myProgram[F[_] : Console : Monad]: F[Unit] =
```
armanbilge: I _think_ this is the style enforced by our scalafmt, but I personally don't care :)

```suggestion
 *  def myProgram[F[_]: Console: Monad]: F[Unit] =
```

armanbilge (dismissed):
Thanks for fixing this, LGTM! Much appreciated :)

Ideally we'd set up something like [sbt-doctest](https://github.com/tkawachi/sbt-doctest) to avoid this in the future :)

std/js/src/main/scala/cats/effect/std/Console.scala
iamonzon: Oh, you're right. 
I thought I've ran `sbt prePR` to make sure it was in the right format. My bad, I accepted the commit

PS: I didn't realize it was going to require approbation again :sweat_smile:

armanbilge: I don't think scalafmt can check/format code in scaladocs, so it's not your fault :) thanks!

armanbilge (approved):

std/js/src/main/scala/cats/effect/std/Console.scala
```diff
@@ -32,18 +32,16 @@ import java.nio.charset.Charset
  *
  * @example
  *   {{{
- * import cats.effect.std.Console
- * import cats.effect.kernel.Sync
- * import cats.syntax.all._
+ *  import cats.Monad
+ *  import cats.effect.std.Console
+ *  import cats.syntax.all._
```
samirbajaj, line 37: Nitpick: why the leading space?

djspiewak: I would guess scalafmt did something odd. It seems to really struggle with code snippets in scaladoc.

PR #2835, opened 2022-02-23
Add `Async#syncStep`

armanbilge:
Takes the ideas from https://github.com/typelevel/cats-effect/pull/1945 and https://github.com/typelevel/cats-effect/pull/2613 and replays it for `Async[F]`. I think this is the law we are aiming for:

```scala
F.syncStep[G, A](fa, limit).to[F].flatMap {
  case Left(fa) => fa
  case Right(a) => F.pure(a)
} <-> fa
```

We should also consider adding `Dispatcher#unsafeRunSyncToFuture` and `Dispatcher#unsafeRunSyncToPromise` to match the methods on `IO`. Slightly annoying is the need to explicitly pass a `limit` since it cannot be retrieved from the `IORuntime` in this case.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -1647,8 +1647,7 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
     override def syncStep[G[_], A](fa: IO[A], limit: Int)(
         implicit G: Sync[G]): G[Either[IO[A], A]] = {
-      type G0[+A] = G[A @uncheckedVariance] // helps with type inference
-      def interpret[B](io: IO[B], limit: Int): G0[Either[IO[B], (B, Int)]] = {
+      def interpret[B](io: IO[B], limit: Int): G[Either[IO[B], (B, Int)]] = {
         if (limit <= 0) {
           G.pure(Left(io))
         } else {
@@ -1675,7 +1674,7 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
               interpret(ioe, limit - 1)
                 .map {
                   case Left(_) => Left(io)
-                  case Right((a, limit)) => Right((a.asRight[Throwable], limit))
+                  case Right((a, limit)) => Right((a.asRight[Throwable].asInstanceOf[B], limit))
                 }
                 .handleError(t => Right((t.asLeft[IO[B]], limit - 1)))
 
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -1647,7 +1647,8 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
     override def syncStep[G[_], A](fa: IO[A], limit: Int)(
         implicit G: Sync[G]): G[Either[IO[A], A]] = {
-      def interpret[B](io: IO[B], limit: Int): G[Either[IO[B], (B, Int)]] = {
+      type G0[+A] = G[A @uncheckedVariance] // helps with type inference
+      def interpret[B](io: IO[B], limit: Int): G0[Either[IO[B], (B, Int)]] = {
         if (limit <= 0) {
           G.pure(Left(io))
         } else {
@@ -1674,7 +1675,7 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
               interpret(ioe, limit - 1)
                 .map {
                   case Left(_) => Left(io)
-                  case Right((a, limit)) => Right((a.asRight[Throwable].asInstanceOf[B], limit))
+                  case Right((a, limit)) => Right((a.asRight[Throwable], limit))
                 }
                 .handleError(t => Right((t.asLeft[IO[B]], limit - 1)))
 
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -36,6 +36,7 @@ import cats.{
 }
 import cats.data.Ior
 import cats.effect.instances.spawn
+import cats.effect.kernel.Sync
 import cats.effect.std.{Console, Env}
 import cats.effect.tracing.{Tracing, TracingEvent}
 import cats.syntax.all._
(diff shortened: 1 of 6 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1689,6 +1645,55 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
...
+              interpret(ioe, limit - 1).flatMap {
+                case Left(_) => G.pure(Left(io))
+                case Right((a, limit)) => interpret(f(a), limit - 1)
+              }
+
+            case IO.Attempt(ioe) =>
+              interpret(ioe, limit - 1)
+                .map {
+                  case Left(_) => Left(io)
+                  case Right((a, limit)) => Right((a.asRight[Throwable].asInstanceOf[B], limit))
```
armanbilge, line 1677: I couldn't figure out how to get the compiler to like this without the cast 😕

kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
```diff
@@ -398,6 +431,14 @@ object Async {
     override def unique: IorT[F, L, Unique.Token] =
       delay(new Unique.Token())
 
+    override def syncStep[G[_], A](fa: IorT[F, L, A], limit: Int)(
+        implicit G: Sync[G]): G[Either[IorT[F, L, A], A]] =
+      G.map(F.syncStep[G, Ior[L, A]](fa.value, limit)) {
+        case Left(ior) => Left(IorT(ior))
+        case Right(Ior.Right(a)) => Right(a)
+        case Right(right) => Left(IorT.fromIor(right))
+      }
```
armanbilge: The transformers were quite interesting to implement :) the law is actually quite important here in guiding the right behavior: we can only return `A` if we pattern match the `pure` representation for the underlying datatype.

djspiewak:
Will look at this in more detail in a bit. I want to leave it open for discussion for now though since this would be a pretty significant addition to `Async`. Paging @SystemFw for galaxy brain thoughts, as well as absolutely anyone else who wants to weigh in.

armanbilge:
I just added a slightly tweaked version of my law from above. I didn't realize `SyncIO#to` was not defined on `Sync` itself (makes sense in retrospect) but actually there's nothing to stop us from `syncStep`ing into the same effect 😁 

https://github.com/typelevel/cats-effect/blob/d82ac6ba5b94ddc894e1db15f472384b7d53b05d/laws/shared/src/main/scala/cats/effect/laws/AsyncLaws.scala#L69-L77

kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
```diff
@@ -149,6 +149,23 @@ trait Async[F[_]] extends AsyncPlatform[F] with Sync[F] with Temporal[F] {
...
+   * completion, the `limit` number of stages, or until the first asynchronous boundary,
+   * whichever is encountered first.
+   *
+   * Note that `syncStep` is merely a hint to the runtime system; implementations have the
+   * liberty to interpret this method to their liking as long as it obeys the respective laws.
+   * For example, a lawful implementation of this function is `G.pure(Left(fa))`, in which case
+   * the original `F[A]` value is returned unchanged.
+   *
+   * @param limit
+   *   The number of stages to evaluate prior to forcibly yielding to `F`
```
djspiewak:
```suggestion
   *   The maximum number of stages to evaluate prior to forcibly yielding to `F`
```

kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
```diff
@@ -271,6 +288,14 @@ object Async {
     override def unique: OptionT[F, Unique.Token] =
       delay(new Unique.Token())
 
+    override def syncStep[G[_], A](fa: OptionT[F, A], limit: Int)(
+        implicit G: Sync[G]): G[Either[OptionT[F, A], A]] =
+      G.map(F.syncStep[G, Option[A]](fa.value, limit)) {
+        case Left(foption) => Left(OptionT(foption))
+        case Right(None) => Left(OptionT.none)
+        case Right(Some(a)) => Right(a)
```
djspiewak, line 296: I really hadn't considered this. The bare `A` in the `Right` creates this scenario. I wonder if we should even special case it or simply allow it to fall through as a `Left`, but then it would seem to mean that this function is just dramatically less useful overall in the context of monad transformers.

This reminds me a bit of some of the work @Baccata did on `Dispatcher`.

laws/shared/src/main/scala/cats/effect/laws/AsyncLaws.scala
```diff
@@ -65,6 +65,32 @@ trait AsyncLaws[F[_]] extends GenTemporalLaws[F, Throwable] with SyncLaws[F] {
 
   def evalOnNeverIdentity(ec: ExecutionContext) =
     F.evalOn(F.never[Unit], ec) <-> F.never[Unit]
+
+  def syncStepIdentity[A](fa: F[A], limit: Int) =
+    F.syncStep[F, A](fa, limit)(syncF).flatMap {
```
djspiewak: Can't we just pass `F` here?

armanbilge: We could. See my comment.
https://github.com/typelevel/cats-effect/blob/d82ac6ba5b94ddc894e1db15f472384b7d53b05d/laws/shared/src/main/scala/cats/effect/laws/AsyncLaws.scala#L75-L77

kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
armanbilge: Yeah, neither had I. I think this is okay, it's lawful and there's nothing inherently unsafe about it.

armanbilge: In a sense, this implementation is no different than what we are implementing for `IO`. The current doc says:

> runs the original `F` to its completion, the `limit` number of stages, or until the first asynchronous boundary

But what it really should say is
> or until the first stage that which cannot be expressed with `Sync` (typically an asynchronous boundary)

Even with `IO` there are non-async stages (such as interacting with `IOLocal`s) that cannot be expressed with `Sync` and thus force us to end the `syncStep`. It's no different for a transformer.

laws/shared/src/main/scala/cats/effect/laws/AsyncLaws.scala
djspiewak: IMO it isn't worth the hack to catch cheaters. :-) If they're going to optimize in that fashion (or in any other), they're responsible for making sure things are well tested.

kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
djspiewak: I like this explanation a lot. Would you mind adding that change to the wording, as well?

djspiewak:
Approved once the above two minor nits are addressed. I think this is a good change overall.

djspiewak:
Oh actually, one further request: can you add this to `AsyncSyntax`?

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1647,7 +1647,8 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
     override def syncStep[G[_], A](fa: IO[A], limit: Int)(
         implicit G: Sync[G]): G[Either[IO[A], A]] = {
-      def interpret[B](io: IO[B], limit: Int): G[Either[IO[B], (B, Int)]] = {
+      type G0[+A] = G[A @uncheckedVariance] // helps with type inference
+      def interpret[B](io: IO[B], limit: Int): G0[Either[IO[B], (B, Int)]] = {
```
armanbilge: Sorry, one last hack. I'd rather do this than the `asInstanceOf` below: type-erasure makes this okay, but unsafe casts are a footgun.

Thoughts?

armanbilge: Nevermind, Scala 3 doesn't like this.

armanbilge:
> We should also consider adding `Dispatcher#unsafeRunSyncToFuture` and `Dispatcher#unsafeRunSyncToPromise` to match the methods on `IO`. Slightly annoying is the need to explicitly pass a limit since it cannot be retrieved from the `IORuntime` in this case.

I tried doing this, but realized we basically can't.

1. We can't provide default implementations of these without requiring an `Async[F]` constraint on the methods.
2. Without `SyncIO` we have no way to `unsafeRunSync` anyway 🙃 

In practice this is not a blocker since these can be implemented in user-land, but it is kind of annoying. Maybe could add them as syntax for `Dispatcher` in core?

djspiewak:
> In practice this is not a blocker since these can be implemented in user-land, but it is kind of annoying. Maybe could add them as syntax for Dispatcher in core?

Hmm, let's tackle this one in a follow up. We basically have to weigh whether a syntax enrichment in **core** is better or worse than a significantly-altered signature in **std** (e.g. taking a function of type `G[A] => A` as a parameter).

djspiewak (approved):

armanbilge:
> taking a function of type `G[A] => A` as a parameter

It would have to be `G[A] => Either[Throwable, A]` I think, but I suppose that's an option too 🤔

PR #2838, opened 2022-02-23
Close cancelation leak in fromCompletableFuture

TimWSpence:
The fiber could be canceled between sequencing the completable future and sequencing the `async` node that installs the finalizer.

Moving sequencing the completable future inside the `async` node means that it's run inside an `uncancelable` and hence safe


djspiewak (dismissed):
Nice catch! You're absolutely correct. Would you mind targeting this at series/3.3.x though?

TimWSpence:
> Nice catch! You're absolutely correct. Would you mind targeting this at series/3.3.x though?

🤦 that is what I intended to do

djspiewak (approved):

satorg:
Just curious – is there a possibility to make features like this one testable in principle? Seems that currently it is all based on assumptions and deep knowledge of the library plumbings only.

djspiewak:
There are a couple ways to test it, including a custom monad transformer which injects `cede` after every `flatMap`. Another approach is to scalacheck several different *prefix* effects and then cancel at different times. It's difficult to catch though. These types of things are more easily caught analytically than empirically.

PR #2843, opened 2022-02-25
Add Semaphore#tryPermit

TimWSpence:
A hopefully safer abstraction over `Semaphore#tryAcquire` which ensures that we always release the permit

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Semaphore.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Semaphore.scala
@@ -255,7 +255,7 @@ object Semaphore {
           Resource.makeFull { (poll: Poll[F]) => poll(acquire) } { _ => release }
 
         def tryPermit: Resource[F, Boolean] =
-          Resource.makeFull { (poll: Poll[F]) => poll(tryAcquire) } { _ => release }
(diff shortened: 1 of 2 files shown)


std/shared/src/main/scala/cats/effect/std/Semaphore.scala
```diff
@@ -247,6 +254,9 @@ object Semaphore {
         def permit: Resource[F, Unit] =
           Resource.makeFull { (poll: Poll[F]) => poll(acquire) } { _ => release }
 
+        def tryPermit: Resource[F, Boolean] =
+          Resource.makeFull { (poll: Poll[F]) => poll(tryAcquire) } { _ => release }
```
SystemFw: shouldn't this only release when acquired?

SystemFw (changes requested):

std/shared/src/main/scala/cats/effect/std/Semaphore.scala
TimWSpence: 🤦 yes. Yes it should

SystemFw: Also, you shouldn't `poll`, and make it uncancelable instead

SystemFw (approved):

djspiewak (approved):

Issue #2663, opened 2021-12-18
Contention on Map when using IO.unsafeToFuture()


yanns:
From https://discord.com/channels/632277896739946517/632278585700384799/921743315169325087

In a project based on `Future`, we introduce `IO` step by step. We use `IO.unsafeToFuture()` for interoperability.
I can observe the following locking:

<img width="772" alt="Screenshot 2021-12-18 at 13 35 46" src="https://user-images.githubusercontent.com/51669/146643621-5bc88393-80b5-4c11-8733-e5c75dabb4ca.png">

The whole application is running with one main `ExecutionContext` using a `ForkJoinPool` building very similarly to `scala.concurrent.ExecutionContext.opportunistic`.

We build our own `IORuntime` to re-use the main `ExecutionContext` like this:

```scala
import cats.effect.unsafe.{IORuntimeConfig, Scheduler}

import java.util.concurrent.ScheduledThreadPoolExecutor
import scala.concurrent.{blocking, ExecutionContext}

object IORuntimeFactory {

  /**
   * To be able to run some [[cats.effect.IO]], for example into a [[scala.concurrent.Future]],
   * we need a [[cats.effect.unsafe.IORuntime]]. This one is build from existing execution context.
   * It does not follow the best practices from cats effects as we are using the same
   * execution context for blocking and non blocking operations.
   *
   * This should be ok as long as we don't use it a lot.
   *
   * In general, it's better to avoid using it, and to consider using a [[cats.effect.std.Dispatcher]] instead.
   */
  def from(ec: ExecutionContext): cats.effect.unsafe.IORuntime = {
    val compute = ec
    val blockingEC = new ExecutionContext {
      override def execute(runnable: Runnable): Unit = ec.execute(() => blocking(runnable.run()))
      override def reportFailure(cause: Throwable): Unit = ec.reportFailure(cause)
    }
    cats.effect.unsafe.IORuntime(compute, blockingEC, scheduler, () => (), IORuntimeConfig())
  }

  /** same scheduler as in [[cats.effect.unsafe.IORuntime.global]] */
  val scheduler: Scheduler = {
    val threadPrefix: String = "io-scheduler"
    val scheduler = new ScheduledThreadPoolExecutor(
      1,
      { r =>
        val t = new Thread(r)
        t.setName(threadPrefix)
        t.setDaemon(true)
        t.setPriority(Thread.MAX_PRIORITY)
        t
      })
    scheduler.setRemoveOnCancelPolicy(true)
[...]

PR #2844, opened 2022-02-26
Thread local fallback weak bag

vasilmkd:
Somewhat addresses #2663.


armanbilge:
I know benchmarks are incoming 😉 but does this have the same basic caveat as https://github.com/typelevel/cats-effect/pull/2508#issuecomment-962658483?
> `ThreadLocal` is very expensive, and often exceeds the cost of contention, which is why you don't see it used too often. It also can have some complex GC implications which have their own performance costs.

vasilmkd:
I wanted to update the original comment, but I'll answer here.

Benchmarks are not coming because I'm not sure what to measure exactly.

The code that this replaces is _literally_ a `ThreadLocalRandom`, which is, you guessed it, a `ThreadLocal`. And in #2663, it was the locking contention that showed up, instead of the thread local usage.

I would like to call @yanns into the conversation if they would be willing to test this change out, with their workflow and measurement in Mission Control. Thanks in advance.

vasilmkd:
Managed to come up with a benchmark.

`series/3.3.x`:
```scala
Benchmark                        (size)   Mode  Cnt    Score   Error  Units
ThreadLocalBenchmark.contention    2000  thrpt   20  287.505 ± 3.640  ops/s
```

`This PR`:
```scala
Benchmark                        (size)   Mode  Cnt    Score    Error  Units
ThreadLocalBenchmark.contention    2000  thrpt   20  306.480 ± 22.891  ops/s
```

In reality, the improvement is a bit misleading due to the bigger margin of error, so it might be a toss up, but thread locals are not strictly slower I guess.

vasilmkd:
Another run is more or less the same:

`This PR`:
```scala
Benchmark                        (size)   Mode  Cnt    Score    Error  Units
ThreadLocalBenchmark.contention    2000  thrpt   20  312.366 ± 30.404  ops/s
```

durban:
It's unclear to me, what guarantees that the `bag.toSet` call in `def foreignFibers()` "sees" the fibers inserted by `monitorFallback`, as `monitorFallback` accesses a bag directly, without any synchronization (and `WeakBag` itself does not seem thread-safe).

vasilmkd:
That's great input @durban. Thank you.

Another benchmark run with the latest changes:
`This PR`:
```scala
Benchmark                        (size)   Mode  Cnt    Score    Error  Units
ThreadLocalBenchmark.contention    2000  thrpt   20  327.031 ± 20.432  ops/s
```

djspiewak (approved):

core/jvm/src/main/scala/cats/effect/unsafe/FiberMonitor.scala
```diff
@@ -172,4 +168,15 @@ private[effect] object FiberMonitor {
       new FiberMonitor(null)
     }
   }
+
+  private[FiberMonitor] final val Bags: ThreadLocal[WeakBag[IOFiber[_]]] =
+    ThreadLocal.withInitial { () =>
+      val bag = new WeakBag[IOFiber[_]]()
+      BagReferences.offer(new WeakReference(bag))
```
armanbilge, line 175: @vasilmkd sorry, I had a follow-up question about this change.

Is it possible that we could lose track of suspended fibers, if the threads that they were suspended from no longer exist? Is that even a realistic situation 😆

vasilmkd: That's a possibility, yes. The change was made with the intention that having an already inaccurate reporting mechanism remain that way is better than a memory leak. If people disagree, PRs are welcome.

armanbilge: That's fair, thanks.

Probably over-complicated but I wonder if we could use a `PhantomReference` to "evacuate" the contents of the bag when its owning thread gets GCed.

Btw, since the WSTP also dynamically adds/removes threads, how is this problem handled there?

vasilmkd: The WSTP does not use this code path. I'm open to exploring Phantom References.

armanbilge: > The WSTP does not use this code path.

Right :) but it still uses a thread-local fiber bag right? And the threads may be added/removed as the WSTP resizes itself? So it seems like it's a very similar problem.

vasilmkd: In the WSTP, each worker thread also has a thread local weak bag, in the sense that each thread has a unique reference to a bag. There is a transfer process of all important data structures when a thread is swapped out which solves this, which essentially ties the lifecycle of the bag to the pool, not to the thread, something which is not available on threads we do not control.

vasilmkd: With Phantom References, there needs to be a dedicated thread which serves the reference queue when these bags expire, which polls references from it and does the cleanup. I'm not very enthusiastic about that but I can be persuaded.

armanbilge: Yeah, I wasn't sure if it's worth it :) instead of a dedicated thread, is this something we can schedule on the runtime itself?

vasilmkd: That's what we had before. It requires solving mapping of threads to bags, which was done using locking. If we come up with a concurrent weak bag/hash map, then sure. But not even JCTools has that afaik. It's a big undertaking.

Edit: I misunderstood your comment and answered something completely different.

vasilmkd: Scheduling on the runtime requires answering how often do to run it, which to me doesn't seem like a good strategy for something considered to be memory beneficial/critical. And ReferenceQueue is not too smart of an interface either. You can poll it in a non-blocking way, and when it returns null, when do you try again? The proper way IMO is to block on it and run cleanup on each expiry.

armanbilge: No it doesn't seem very elegant :) I feel like in practice, there must be some reasonable rate at which we can check the `ReferenceQueue` ... if an application is adding/removing threads too fast seems like its performance would be bounded by other factors anyway. But I don't really know about such things :)

After thinking about this more, seems like it could be important. A situation in which there is a deadlock seems like exactly the situation when a dynamically resizing threadpool would start culling threads due to lack of work, which could cause GC of the fiber bag holding the fibers would help diagnose the deadlock.

vasilmkd: @djspiewak 👆🏻

PR #2846, opened 2022-02-27
Add `unsafeRunSync*` syntax for JS `Dispatcher`

armanbilge:
Following up on https://github.com/typelevel/cats-effect/pull/2835#issuecomment-1053265252, replays the convenience methods added to `IO` in https://github.com/typelevel/cats-effect/pull/2613 for `Dispatcher` (JS-only).

It needs to be added as syntax in core due to the dependency on `SyncIO` for the `unsafeRunSync` execution.

Daniel proposed an alternative in https://github.com/typelevel/cats-effect/pull/2835#issuecomment-1053635478:

> We basically have to weigh whether a syntax enrichment in **core** is better or worse than a significantly-altered signature in **std** (e.g. taking a function of type `G[A] => A` as a parameter).

IIUC it's proposing something like this on `Dispatcher`:
```scala
def unsafeRunSyncToFuture[G[_]: Sync, A](
  fa: F[A],
  limit: Int,
  unsafeRunSync: G[A] => Either[Throwable, A]
)(implicit F: Async[F]): Future[A] = {
  unsafeRunSync(F.syncStep[G, A](fa, limit)) match {
    case Left(t) => Future.failed(t)
    case Right(Left(fa)) => unsafeToFuture(fa)
    case Right(Right(a)) => Future.successful(a)
  }
}
```
which I would then call like this:
```scala
dispatcher.unsafeRunSyncToFuture(fa, 32, (_: SyncIO[A]).attempt.unsafeRunSync())
```

But at that point, I think I could just as easily write:
```scala
fa.syncStep[SyncIO](32).unsafeRunSync().fold(
  dispatcher.unsafeRunAndForget(_),
  _ => () // whatever
)
```

This assumes I don't care about the result ... which in many cases I don't, because I'm completing a `Deferred` or `offer`ing to a `Queue`.

The syntax proposed here gets us this:
```scala
dispatcher.unsafeRunSyncToFuture(fa, 32)
```

So, subjective obviously. And not necessarily mutually exclusive.

I also wish there was a better story for the `limit` parameter, which we can't fish out of the `IORuntime` in this case. So it will almost definitely be hardcoded, which requires some iota of thinking or tempts a lazy `Int.MaxValue` which is probably fine for the `complete`/`offer` use-cases.

### DIFF
--- a/core/js/src/main/scala/cats/effect/IOPlatform.scala
+++ b/core/js/src/main/scala/cats/effect/IOPlatform.scala
@@ -37,6 +37,17 @@ abstract private[effect] class IOPlatform[+A] { self: IO[A] =>
           ()
       })
 
+  /**
+   * Evaluates the effect and produces the result in a `Future`.
+   *
+   * This is similar to `unsafeToFuture` in that it evaluates the `IO` as a side effect in a
+   * non-blocking fashion, but begins by taking a `syncStep` limited by the runtime's auto-yield
+   * threshold. This function should really only be used if it is critical to attempt to
+   * evaluate this `IO` without first yielding to the event loop.
+   *
+   * @see
+   *   [[IO.syncStep(limit:Int)*]]
+   */
   def unsafeRunSyncToFuture()(implicit runtime: unsafe.IORuntime): Future[A] =
     self.syncStep(runtime.config.autoYieldThreshold).attempt.unsafeRunSync() match {
       case Left(t) => Future.failed(t)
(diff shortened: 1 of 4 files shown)


djspiewak:
Btw haven't forgotten about this. I'm thinking we might be able to make it a bit more convenient in some cases. Will see if I can find time to experiment a bit.

djspiewak (approved):
Did some experimentation. I think this is about the best we can do for now. We could remove the `syncLimit` parameter for any `F: LiftIO`, but that would actually be really weird because then you would have code that would compile until you *add* a constraint, at which point scalac would start telling you that you're missing a parameter. So that seems worse than just having it behave this way all the time.

PR #2852, opened 2022-03-01
Add IO.randomUUID convenience method

Daenyth:


armanbilge (dismissed):

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -36,7 +36,7 @@ import cats.{
 }
 import cats.data.Ior
 import cats.effect.instances.spawn
-import cats.effect.std.{Console, Env}
+import cats.effect.std.{Console, Env, GenUUID}
```
djspiewak: I believe you mean `UUIDGen` :-)

djspiewak (changes requested):
@Daenyth Looks like this needs to be split

core/shared/src/main/scala/cats/effect/IO.scala
Daenyth: Damnit

wjoel:
Wouldn't `Random` be a better place for this? https://github.com/typelevel/cats-effect/blob/series/3.x/std/shared/src/main/scala/cats/effect/std/Random.scala

djspiewak:
It could be, yes, but this is also a wildly common case, so I think it's fair to lean a bit toward better convenience. `Random` isn't any harder to summon than `UUIDGen`, but IO is just… there.

wjoel:
Fair enough (even if I personally don't generate random UUIDs *that* often ;) ), but maybe the help text could include a reference `Random` so newbies that type `IO.rando...` have a better chance of discovering it. Just a thought.

djspiewak (approved):

Daenyth:
I was looking at that too - `UUIDGen` doesn't actually use `Random`, it just wraps `UUID.randomUUID()` in `Sync[F].delay`

djspiewak:
This actually might be something that should be addressed, since randomUUID can block threads.

rossabaker:
I have seen `UUID.randomUUID` kill more than one service in the years BCE (Before Cats Effect).  The synchronization isn't obvious until you start profiling why you're getting single-threaded performance.

PR #2854, opened 2022-03-01
Sequential `Dispatcher`

### LINKED ISSUE #2768: Allow `Dispatcher` to be configured for single-producer scenarios
Right now, `Dispatcher` somewhat pessimistically assumes that you will be using it from multiple producers simultaneously, and the actions you're enqueuing might be very lengthy. This forces its implementation to stripe internally, as well as `.start` every action it is given, both of which sacrifice ordering properties entirely. Additionally, this adds some extra overhead relative to what we theoretically could do if we had a guarantee that only a single fiber was producer to a given `Dispatcher`.

We should add this as a configurable mode. In particular, what this mode would do is suppress the striping *and* the `start`ing. Users would not be able to share such `Dispatcher`s across fibers (or they could, but it would be very slow to do so), but if they stay affine to a single fiber the results would be *better* performance and guaranteed ordering in the sense that `fa.unsafeRunAndForget(); fb.unsafeRunAndForget()` would produce the same outcome as `fa.guarantee(fb).unsafeRunAndForget()`.

djspiewak:
This is a little less of an SPSC `Dispatcher` and more MPSC, and definitively a lot more about "short actions that you want to be sequential". Basically this special-cases just a bit for the `q.offer` case, which is obviously quite common in practice. Haven't added benchmarks back, but anecdotally from test runs this appears to be around a 1-2 order of magnitude improvement in those cases.

Closes #2768

### DIFF
--- a/benchmarks/src/main/scala/cats/effect/benchmarks/DispatcherBenchmark.scala
+++ b/benchmarks/src/main/scala/cats/effect/benchmarks/DispatcherBenchmark.scala
@@ -66,7 +66,7 @@ class DispatcherBenchmark {
         }
       }))
 
-    Dispatcher[IO]()
+    Dispatcher.parallel[IO]
       .use { disp =>
         List
           .range(0, size)
--- a/benchmarks/src/main/scala/cats/effect/benchmarks/DispatcherBenchmark.scala
+++ b/benchmarks/src/main/scala/cats/effect/benchmarks/DispatcherBenchmark.scala
@@ -66,7 +66,7 @@ class DispatcherBenchmark {
         }
       }))
 
-    Dispatcher[IO]
+    Dispatcher[IO]()
       .use { disp =>
         List
           .range(0, size)
--- a/benchmarks/src/main/scala/cats/effect/benchmarks/DispatcherBenchmark.scala
+++ b/benchmarks/src/main/scala/cats/effect/benchmarks/DispatcherBenchmark.scala
@@ -66,7 +66,8 @@ class DispatcherBenchmark {
         }
       }))
 
-    Dispatcher.parallel[IO]
+    Dispatcher
+      .parallel[IO]
       .use { disp =>
         List
           .range(0, size)
--- a/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
@@ -47,6 +47,11 @@ import java.util.concurrent.atomic.{AtomicBoolean, AtomicReference}
 trait Dispatcher[F[_]] extends DispatcherPlatform[F] {
 
   /**
+   * Indicates the mode in which this `Dispatcher` is operating (set at construction time).
+   */
+  def mode: Dispatcher.Mode = Dispatcher.Mode.Parallel
+
+  /**
    * Submits an effect to be executed, returning a `Future` that holds the result of its
    * evaluation, along with a cancelation token that can be used to cancel the original effect.
    */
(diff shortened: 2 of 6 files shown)


benchmarks/src/main/scala/cats/effect/benchmarks/DispatcherBenchmark.scala
```diff
@@ -66,7 +66,7 @@ class DispatcherBenchmark {
         }
       }))
 
-    Dispatcher[IO]
+    Dispatcher[IO]()
```
vasilmkd: Bikeshedding time, this or a separate named method `Dispatcher.sequential`? I think it's also worth mentioning that default parameters in methods are a pain for library writers.

vasilmkd (commented):
Nice and clean. Only the one question.

benchmarks/src/main/scala/cats/effect/benchmarks/DispatcherBenchmark.scala
armanbilge: Yep, I expressed the same in discord: `Dispatcher.sequential` and `Dispatcher.parallel` IMO. We can consider deprecating `Dispatcher.apply` instead of flat-out breaking source.

Also, not sure if there can be additional modes in the future but using a public ADT here would prevent that.

djspiewak: > Also, not sure if there can be additional modes in the future but using a public ADT here would prevent that.

Refinement: only prevents it from a *source* compatibility standpoint. Binary compatibility is fine.

armanbilge: No, binary compatibility is not fine unless everyone defensively adds a `case _ =>` to their pattern matches.

See also:
- https://github.com/lightbend/mima/issues/200

djspiewak: Btw I'm definitely fine with the bikeshedding. :-) From an API design standpoint, do we *like* `Dispatcher.parallel` and `Dispatcher.sequential` rather than the `Mode` ADT?

djspiewak: > No, binary compatibility is not fine unless everyone defensively adds a case _ => to their pattern matches.

Oh I see what you mean. Hmm.

djspiewak:
So one immediate problem with doing away with the ADT is it means we can't support `Dispatcher#mode`. Maybe that's a non-issue though? How much do we feel like people are going to need to runtime-introspect their dispatchers and case on their modality?

armanbilge:
TBH I hadn't considered allowing introspection at all. When I need a dispatcher, I usually create exactly what I need on-the-spot, instead of asking someone else to give it to me?

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -96,12 +96,51 @@ object Dispatcher {
 
   private[this] val Completed: Either[Throwable, Unit] = Right(())
 
+  @deprecated("3.4.0", "please use parallel or sequential instead")
```
djspiewak: Maybe I shouldn't be so polite?

ahjohannessen: No need for `please` :)

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -96,7 +96,9 @@ object Dispatcher {
 
   private[this] val Completed: Either[Throwable, Unit] = Right(())
 
-  @deprecated("3.4.0", "please use parallel or sequential instead")
+  @deprecated(
+    "3.4.0",
+    "use parallel or sequential instead; the former corresponds to the previous semantics of this method")
```
armanbilge: _Current_ semantics! I would just say this method delegates to `parallel`.

djspiewak:
```suggestion
  @deprecated(
    "3.4.0",
    "use parallel or sequential instead; the former corresponds to the current semantics of this method")
```

PR #2825, opened 2022-02-16
Use Java release option

vasilmkd:
## Example 1

The easy example first, APIs that don't exist on JDK 8 (`VarHandle`):

```java
package io.vasilev;

import java.lang.invoke.MethodHandles;
import java.lang.invoke.VarHandle;

public class VarHandles {
    private volatile int n;

    private static final VarHandle N;

    static {
        try {
            N = MethodHandles.lookup().findVarHandle(VarHandles.class, "n", int.class);
        } catch (Exception e) {
            throw new ExceptionInInitializerError(e);
        }
    }
}
```

Compiling:
```sh
javac io/vasilev/VarHandles.java --release 8
io/vasilev/VarHandles.java:4: error: cannot find symbol
import java.lang.invoke.VarHandle;
                       ^
  symbol:   class VarHandle
  location: package java.lang.invoke
io/vasilev/VarHandles.java:9: error: cannot find symbol
    private static final VarHandle N;
                         ^
  symbol:   class VarHandle
  location: class VarHandles
io/vasilev/VarHandles.java:13: error: cannot find symbol
            N = MethodHandles.lookup().findVarHandle(VarHandles.class, "n", int.class);
                                      ^
  symbol:   method findVarHandle(Class<VarHandles>,String,Class<Integer>)
  location: class Lookup
3 errors
```

---
## Example 2

Our favorite example, `ByteBuffer#flip()` on JDK 9+ overrides `Buffer#flip()` with a more specific type.

- On JDK 8 we have:
    1. `Buffer.flip()` has the JVM signature `java/nio/Buffer.flip:()Ljava/nio/Buffer`
    2. `ByteBuffer.flip()` has the JVM signature `java/nio/ByteBuffer.flip:()L/java/nio/Buffer`

- On JDK 9+ we have:
    1. `Buffer.flip()` has the same JVM signature `java/nio/Buffer.flip:()Ljava/nio/Buffer`
    2. `ByteBuffer.flip()` has the new JVM signature `java/nio/ByteBuffer.flip:()L/java/nio/ByteBuffer` and thus cannot be called on JDK 8.

Let's examine the output of `javac --relase 8` in this case, when we have an explicit return type in the code.

```java
package io.vasilev;

import java.nio.ByteBuffer;

public class Buffers {

    private static final ByteBuffer buffer = ByteBuffer.allocate(16);

    static {
        ByteBuffer flipped = buffer.flip();
    }
}
```

```sh
[...]

PR #2857, opened 2022-03-03
Migrate to sbt-typelevel

armanbilge:
Taking the conservative approach and not mixing up the build too much with sbt-typelevel gospel 😉 

Opening for CI. I anticipate this PR will fail while we get MiMa up-to-speed on Scala 3 and possible Scala.js.


build.sbt
```diff
@@ -101,17 +100,17 @@ val Scala3 = "3.0.2"
 
 ThisBuild / crossScalaVersions := Seq(Scala3, "2.12.15", Scala213)
 
-ThisBuild / githubWorkflowUseSbtThinClient := false
 ThisBuild / githubWorkflowTargetBranches := Seq("series/3.*")
+ThisBuild / tlCiReleaseTags := false
+ThisBuild / tlCiReleaseBranches := Nil
```
armanbilge, line 120: No CI publishing :)

.github/workflows/ci.yml
```diff
@@ -12,10 +12,15 @@ on:
...
   push:
     branches: [series/3.*]
+    tags: [v*]
 
 env:
+  PGP_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
+  SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
+  SONATYPE_CREDENTIAL_HOST: ${{ secrets.SONATYPE_CREDENTIAL_HOST }}
+  SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
+  PGP_SECRET: ${{ secrets.PGP_SECRET }}
```
armanbilge: Hmm, I guess these are baked in a little too hard :) anyway should be harmless.

build.sbt
```diff
@@ -156,7 +155,7 @@ val ciVariants = CI.AllCIs.map(_.command)
 val jsCiVariants = CI.AllJSCIs.map(_.command)
 ThisBuild / githubWorkflowBuildMatrixAdditions += "ci" -> ciVariants
 
-ThisBuild / githubWorkflowBuildMatrixExclusions ++= {
+ThisBuild / githubWorkflowBuildMatrixExclusions := {
```
armanbilge, line 173: sbt-typelevel already does some of these exclusions for us. Rather than tease apart what's been upstreamed we'll just keep using our own.

project/Common.scala
```diff
@@ -16,20 +16,18 @@
...
     Seq(
-      libraryDependencies ++= {
-        if (isDotty.value)
-          Nil
-        else
-          Seq(compilerPlugin("com.olegpy" %% "better-monadic-for" % "0.3.1"))
-      }
+      headerLicense := Some(
+        HeaderLicense.ALv2(s"${startYear.value.get}-2022", organizationName.value)
+      )
```
armanbilge, line 31: Sorry, this is a spiewak-ism 😝 Vasil has a PR to upstream where it belongs.

- https://github.com/sbt/sbt-header/pull/282

armanbilge:
I just added like a 100+ mima filters. At a glance they all seem reasonable but would be good for someone to look with a more careful eye.

build.sbt
```diff
@@ -27,6 +27,7 @@ import sbtcrossproject.CrossProject
 import JSEnv._
 
 ThisBuild / tlBaseVersion := "3.3"
+ThisBuild / tlUntaggedAreSnapshots := false
```
armanbilge: I almost forgot 😆

build.sbt
```diff
@@ -100,18 +100,20 @@ val Scala213 = "2.13.7"
...
 ThisBuild / githubWorkflowTargetBranches := Seq("series/3.*")
+ThisBuild / tlCiReleaseTags := false
+ThisBuild / tlCiReleaseBranches := Nil
 
 val OldGuardJava = JavaSpec.temurin("8")
 val LTSJava = JavaSpec.temurin("11")
 val LatestJava = JavaSpec.temurin("17")
 val ScalaJSJava = OldGuardJava
-val GraalVM = JavaSpec.graalvm("21.3.0", "11")
+val GraalVM = JavaSpec.graalvm("11")
```
djspiewak, line 126: Why is the Graal version not specified. This feels a little weird in general since it suggests that the GraalVM version would be either unstable or unconfigurable.

build.sbt
```diff
@@ -100,18 +100,20 @@ val Scala213 = "2.13.7"
 val Scala3 = "3.0.2"
 
 ThisBuild / crossScalaVersions := Seq(Scala3, "2.12.15", Scala213)
+ThisBuild / tlVersionIntroduced := Map("3" -> "3.1.1")
+ThisBuild / tlJdkRelease := Some(8)
```
djspiewak, line 116: Why is this needed if we aren't doing CI releases?

project/Common.scala
djspiewak: It does indeed belong in upstream

.github/workflows/ci.yml
djspiewak: Bleh. :-) I'd definitely like to see these cleaned up, but it's kind of okay for now.

djspiewak (changes requested):
Btw I do agree with the reformattings, but why are they happening as part of this PR?

.github/workflows/ci.yml
```diff
@@ -12,10 +12,15 @@ on:
     branches: [series/3.*]
   push:
     branches: [series/3.*]
+    tags: [v*]
```
djspiewak, line 15: Oh actually one extra bit: we don't need to build tags because we're not doing CI releases. This appears to be another thing that's a bit too baked-in.

build.sbt
armanbilge: I can revert this. Most people probably want the latest Graal and not a specific version (which is easy to forget to update, as we've forgotten to here :)

build.sbt
armanbilge: Because neither of these has anything to do with CI release :)

`tlVersionIntroduced` tells MiMa not to look for Scala 3 artifacts for CE versions older than 3.1.1.

`tlJdkRelease` is the upstreamed version of https://github.com/typelevel/cats-effect/pull/2825.

.github/workflows/ci.yml
armanbilge: Yep, I noticed this with the other stuff. It's also harmless but will definitely fix upstream.

armanbilge:
Formatting changes are probably because sbt-typelevel brings you latest sbt-scalafmt which this branch missed.

- https://github.com/typelevel/cats-effect/pull/2690

which includes a fix for:

- https://github.com/scalameta/sbt-scalafmt/issues/190

build.sbt
armanbilge: Btw @vasilmkd what are your thoughts on this topic, maybe we need to revise the JDK index to support this?

.github/workflows/ci.yml
djspiewak: Not *quite* harmless since it counts against our running instance quota, but mostly harmless yes. :-)

build.sbt
vasilmkd: We were already on latest, I was doing it manually. Graalvm also follows the JDK release cycle, so as soon as the latest drops, the previous one is obsolete and unsupported. I don't want to sound dismissive, but I'm not interested in changing the index to support old versions.

armanbilge: I see the latest in the index is 22.0.0.2 and CE is on 21.3.0, so we've fallen behind again (at least on this branch).

Ok @djspiewak your call, how do you want to configure this for CE?

> version would be either unstable

It depends on what you are targeting. If you are targeting JDK whatever then shouldn't it be as stable as any of the other JVMs?

OTOH, if you are targeting specific GraalVM APIs then obviously it makes a difference. Actually [I have a project like this](https://github.com/armanbilge/fungio/).

FTR I didn't remove the ability to configure a specific GraalVM (it still shells out to [DeLaGuardo/setup-graalvm](https://github.com/DeLaGuardo/setup-graalvm) like you set it up) although it is now a little more "discouraged" because I expect 99% of projects don't actually want this.

Finally, if you truly want stability, you should really pin the JDK index to a particular hash anyway :)

vasilmkd:
In light of https://github.com/lampepfl/dotty/issues/14624 and https://github.com/lampepfl/dotty/pull/14662, I think we can go ahead and apply the MiMa filters for the Scala 3 issue.

armanbilge:
After this PR merges, I'll do a follow-up which adds JDK17/Scala3 to CI. That will force us to add the mima filters.

.github/workflows/ci.yml
armanbilge: So actually this one is harder to fix, because it controls the _entire_ workflow file. Even though neither CI nor publishing may not need to run on tags, in general there may be other jobs in the workflow (such as deploying a website) that need to be triggered.

djspiewak:
`show version`
`3.3.9-18-b0b4a99-20220402T054540Z-SNAPSHOT`

How do I make it stop? :-( Worktrees do seem to work though.

armanbilge:
Cool! So not a blocker ... right? 😇

djspiewak:
LOL. Well, I really want to make that version stuff at least *configurable*, because the transitive impact it has on the ecosystem is… unpleasant. Not a hard blocker, but definitely a something. I'll see if I can fix it myself in sbt-tl

djspiewak:
Oh, it's more sbt-git issues: 

```
sbt:root> show gitUncommittedChanges
[info] benchmarks / gitUncommittedChanges
[info] 	true
[info] testsJVM / gitUncommittedChanges
[info] 	true
```

There are no uncommitted changes.

djspiewak:
Alright, managed to get it working, sort of. `3.3.9-20-44bf4e1` has been published to sonatype so we can make sure things are in good shape. I published it using Java 11, fwiw, so we can test the release flag while we're at it.

Basically, we needed one last fix for sbt-git which I completely forgot about. Namely, they're determining dirty directories incorrectly (by using `git status` and then grepping for text). This has issues between different versions of `git`, and so the workaround is to use `git status -s` and check to see if it returns *any* results.

djspiewak:
Release looks good. Branch appears to be generally functional. I'm in favor of merging this, though it would be nice to see the sbt-git workaround upstreamed at some point.

build.sbt
djspiewak: We shoudl try to inherit more in the future, but this is a fine start.

djspiewak (approved):

vasilmkd:
Let's go! 🚀

vasilmkd (approved):

armanbilge:
🎉 I wasn't sure if we'd make it 😂 thank you! Obviously still plenty to upstream 😬

PR #2859, opened 2022-03-07
Fix `IO#syncStep` double-execution

### LINKED ISSUE #2858: IO.syncStep leads to double execution
Hey guys, 
I found which seems to me a critical bug during migration of codebase from ce2 to ce3

The code below prints `hello` twice, I would expect `hello` to be printed once.

```scala
    IO
      .apply { println("hello") }
      .flatMap { _ => IO.cede }
      .as(Succeeded)
      .syncStep
      .unsafeRunSync() match {
      case Right(a) => Future.successful(a)
      case Left(a)  => a.unsafeToFuture()
    }
```

armanbilge:
WIP. Starts by adding a reproducer.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -843,7 +843,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
 
         case IO.Map(ioe, f, _) =>
           interpret(ioe).map {
-            case Left(_) => Left(io)
+            case Left(io) => Left(io.map(f))
             case Right(a) => Right(f(a))
           }
 
@@ -919,7 +919,7 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
    * Newtype encoding for an `IO` datatype that has a `cats.Applicative` capable of doing
    * parallel processing in `ap` and `map2`, needed for implementing `cats.Parallel`.
    *
-   * For converting back and forth you can use either the `Parallel[IO]` instance or 
+   * For converting back and forth you can use either the `Parallel[IO]` instance or
    * the methods `cats.effect.kernel.Par.ParallelF.apply` for wrapping any `IO` value and
    * `cats.effect.kernel.Par.ParallelF.value` for unwrapping it.
    *
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -849,7 +849,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
 
         case IO.FlatMap(ioe, f, _) =>
           interpret(ioe).flatMap {
-            case Left(_) => SyncIO.pure(Left(io))
+            case Left(io) => SyncIO.pure(Left(io.flatMap(f)))
             case Right(a) => interpret(f(a))
           }
 
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -864,7 +864,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
         case IO.HandleErrorWith(ioe, f, _) =>
           interpret(ioe)
             .map {
-              case Left(_) => Left(io)
+              case Left(io) => Left(io.handleErrorWith(f))
               case Right(a) => Right(a)
             }
             .handleErrorWith(t => interpret(f(t)))
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -856,7 +856,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
         case IO.Attempt(ioe) =>
           interpret(ioe)
             .map {
-              case Left(_) => Left(io)
+              case Left(io) => Left(io.attempt)
               case Right(a) => Right(a.asRight[Throwable])
             }
             .handleError(t => Right(t.asLeft[IO[B]]))
--- a/tests/shared/src/test/scala/cats/effect/IOSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/IOSpec.scala
@@ -1345,7 +1345,7 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
 
       "should not execute side-effects twice (#2858)" in ticked { implicit ticker =>
         var i = 0
-        val io = (IO(i += 1) *> IO.cede).syncStep.unsafeRunSync() match {
+        val io = (IO(i += 1) *> IO.cede as ()).syncStep.unsafeRunSync() match {
           case Left(io) => io
           case Right(_) => IO.unit
         }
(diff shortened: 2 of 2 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -843,7 +843,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
 
         case IO.Map(ioe, f, _) =>
           interpret(ioe).map {
-            case Left(_) => Left(io)
+            case Left(io) => Left(io.map(f))
```
t3hnar, line 846: > but I suspect more bugs

@armanbilge why do you suspect more bugs?

armanbilge: It's because the same mistake was made in 4 places :) I think I fixed all of them now.

vasilmkd: Is it my code? I remember going back and forth on this one and possibly mucking it up. Sorry.

armanbilge: Eh, blame the reviewer 😉 

Anyway, thanks to @t3hnar for catching this, I'm about the invest the CE.js world pretty deep into `syncStep`.

t3hnar: `syncStep` is critical functional for projects in my company as well. Mostly because it is used for integration with akka/kafka/cassandra/java libraries, and things could not be done without it, as otherwise you would be either loosing order guarantees or blocking threads.

armanbilge: @t3hnar I'd be interested to see a specific example of your use-case. I wonder if the work in https://github.com/typelevel/cats-effect/pull/2854 will be applicable to you? It provides an alternate mode for `Dispatcher` which guarantees ordering of effects.

Actually, `syncStep` was primarily introduced to help with JavaScript interop. In JVM-world, I think use-cases should be much more rare and `Dispatcher` is probably a better tool.

t3hnar: Any code where you have java listeners called in exact order that should start executing `IO` based code.
And ordering guaranteed are modelled inside of `IO` composition for instance via `Ref` (not crossing async boundary)

Also in cases where you cannot use `IO.async` - because you want to have `IO[IO[A]]` instead of classic `IO[A]` - outer IO is about java api being called, inner IO is about execution is completed.


```scala
def javaApi: JavaFuture[A] = ???
def scalaApi: IO[IO[A]] = {
  for {
  future <- IO { javaApi } // because this gives order guarantee during method call and does not cross async boundary
  } yield {
    future.javaFutureToIO
  }
}
```

t3hnar (dismissed):

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -843,28 +843,28 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
...
             case Right(a) => Right(f(a))
           }
 
         case IO.FlatMap(ioe, f, _) =>
           interpret(ioe).flatMap {
-            case Left(_) => SyncIO.pure(Left(io))
+            case Left(io) => SyncIO.pure(Left(io.flatMap(f)))
```
djspiewak, line 852: I have no intuition as to *why* this would cause doubled execution…

djspiewak (dismissed):
This is a really nice catch! Honestly I'm surprised we didn't see it before, but I'm even more surprised that it caused doubled execution. That's the part that really doesn't make sense to me.

core/shared/src/main/scala/cats/effect/IO.scala
t3hnar: because we returned original `io` as `IO ` and synchronous part of it as `SyncIO`
and executing one after another leads to double execution of synchronous part

t3hnar:
@armanbilge @djspiewak any plans on releasing this fix anytime soon? :)

core/shared/src/main/scala/cats/effect/IO.scala
djspiewak: > Any code where you have java listeners called in exact order that should start executing IO based code.

This really sounds like exactly what #2854 was designed to fix. 

> Also in cases where you cannot use IO.async - because you want to have IO[IO[A]] instead of classic IO[A] - outer IO is about java api being called, inner IO is about execution is completed.

Fwiw, you might want to consider modeling this as `Resource[IO, IO[A]]` if `javaApi` involves any scarce resources. I definitely get what you're going at though. This type of pattern is annoying precisely because older-style APIs (such as in Java) often model these types of lifecycles implicitly ("kick off a running future and which you can ask for later") rather than explicitly (`IO[IO[A]]`).

djspiewak:
@t3hnar If this is urgent for you, I can see about getting it out later today

t3hnar:
@djspiewak would be lovely! thanks a lot!

core/shared/src/main/scala/cats/effect/IO.scala
t3hnar: > This really sounds like exactly what https://github.com/typelevel/cats-effect/pull/2854 was designed to fix.

already subscribed and looking forward

tests/shared/src/test/scala/cats/effect/IOSpec.scala
```diff
@@ -1342,6 +1342,50 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
             case Right(_) => SyncIO.raiseError[Unit](new RuntimeException("Boom!"))
           } must completeAsSync(())
       }
+
+      "should not execute side-effects twice for map (#2858)" in ticked { implicit ticker =>
```
Daenyth:
```suggestion
      "should not execute effects twice for map (#2858)" in ticked { implicit ticker =>
```

Reading this the first time, I thought from the description that this was about shapes like `io.map(_ => i += 1)`

core/shared/src/main/scala/cats/effect/IO.scala
armanbilge: Here's an example:
```scala
val io1 = IO(println(...))
val io2 = io1 *> IO.cede
val io3 = io2.void
```

If you run `io3.syncStep.unsafeRunSync`, roughly this happens:
1. We pattern match `io3 @ IO.Map(io2, _ => ())`
2. We interpret `io2`:
    1. We pattern match `io2 @ IO.FlatMap(io1, _ => IO.cede)`
    2. We interpret `io1`:
        1. We pattern match `io1 @ IO.delay(println(...))`
        2. We run it synchronously, and return `Right(())`
    3. We apply the flatMap `f` to `()` and get `IO.cede`.
    4. We interpret `IO.cede`, but can't, so we return `Left(IO.cede)`
3. We get `Left(IO.cede)`, so we have to bail. Now, we have an interesting question: should we return the `Left(io.void)` with the `io` we've just been given, or should we return `Left(io3)` that we started the iteration with? The old code was doing the latter.

Hope that makes sense.

tests/shared/src/test/scala/cats/effect/IOSpec.scala
armanbilge: Oh yeah, that's unclear. It's observed as the side-effects being run twice, but it is in fact the entire effect being run twice. Thanks!

djspiewak (approved):

PR #2860, opened 2022-03-07
Added publish settings autoplugin

djspiewak:
Temporary fix until #2857 is ready

### DIFF
--- a/project/Publish.scala
+++ b/project/Publish.scala
@@ -1,0 +1,12 @@
+import sbt._, Keys._
+
+import sbtspiewak.SpiewakSonatypePlugin
+import xerial.sbt.Sonatype.SonatypeKeys._
+
+object Publish extends AutoPlugin {
+
+  override def requires = SpiewakSonatypePlugin
+  override def trigger = allRequirements
+
+  override def projectSettings = Seq(sonatypeCredentialHost := "s01.oss.sonatype.org")
+}


vasilmkd (approved):

armanbilge (dismissed):
I expect to be overridden, but 👎 

This PR and https://github.com/typelevel/cats-effect/pull/2857 both fix publishing to s01. The difference is that this PR supports git worktrees but no Scala3 or JS MiMa, whereas https://github.com/typelevel/cats-effect/pull/2857 supports MiMa for all artifacts, but no git worktrees.

And we literally just picked up Scala 3 issues, with MiMa, while testing local release. Unless I'm missing something, there are certainly ways to use git without worktrees, but no-one is checking MiMa unless you configure it :)

I'd like to see a PR that fixes MiMa along with publishing, no preference at all as to how.

vasilmkd:
I agree with the sentiment, but for the next major relese 3.4.x. 3.3.7 should take precedence in this case.

armanbilge:
Well, I said my piece, and to his credit @djspiewak make a good effort to try and work around this. But it's a long thread of issues with sbt-git and upstream friends unraveling as we pull. So let's not block.

djspiewak:
To summarize what @armanbilge and I decided:

1. Merging this to unblock publication
2. Releasing the next thing in 3.3.x
3. I'm going to work on fixing sbt-tl to workaround the jgit issues
4. We'll migrate CE to sbt-tl directly

PR #2862, opened 2022-03-08
Updated versions for 3.3.7

djspiewak:

PR #2863, opened 2022-03-08
Released 3.3.7

armanbilge:
Following up on https://github.com/typelevel/cats-effect/pull/2862.

The compiler is giving me a really hard time with the effect-agnostic `syncStep` :( I tried shuffling things again but still ended up with an unsafe cast in there. It works fine with concrete `SyncIO` so there must be some way to make it work ...

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -1657,8 +1657,9 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
     override def syncStep[G[_], A](fa: IO[A], limit: Int)(
         implicit G: Sync[G]): G[Either[IO[A], A]] = {
-      type H[+A] = G[A @uncheckedVariance]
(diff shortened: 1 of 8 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1852,3 +1810,51 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
   }
 
 }
+
+private object SyncStep {
+  def interpret[G[+_], B](io: IO[B], limit: Int)(
```
armanbilge, line 1815: So this is annoying, but it works. By extracting out the sync step interpreter to its own scope avoids the implicits floating around inside the `IO` companion. Then, the compiler inference is able to kick-in and do the right thing by itself without requiring explicit types or unsafe casts.

djspiewak (approved):
I think this is basically happening because GADTs are a wonky path through subtyping, which is partially disabled when we have an invariant constructor.

PR #2864, opened 2022-03-08
Exclude completed fibers from the live runtime snapshot

### LINKED ISSUE #2628: Indicate if fiber is completed in dump/toString

iRevive:
A follow-up to https://github.com/typelevel/cats-effect/pull/2829.

The test I wrote is passing on the 3.x branch as well.
Hence I'm trying to find a way to emulate a required behavior: obtain a completed fiber in a live snapshot. 
I can assume, I need a strong reference on a fiber?


core/js/src/main/scala/cats/effect/unsafe/FiberMonitor.scala
```diff
@@ -57,8 +57,8 @@ private final class ES2021FiberMonitor(
 
   def liveFiberSnapshot(print: String => Unit): Unit =
     Option(compute).foreach { compute =>
-      val queued = compute.liveFibers()
-      val rawForeign = bag.toSet
+      val queued = compute.liveFibers().filterNot(_.isDone)
```
iRevive, line 60: Perhaps it's better to tweak `liveFibers` method to exclude completed fibers? Instead of `liveFibers().filterNot(_.isDone)`

armanbilge:
> I can assume, I need a strong reference on a fiber?

Yes, that would be the way to do it 👍

iRevive:
I've tried the following test:
```scala
val rt: IORuntime = ... // compute: WorkStealingThreadPool

for {
  cdl     <- CountDownLatch[IO](1)
  fiber   <- IO.println("test").flatMap(_ => cdl.await).start
  _       <- IO.println(fiber) // cats.effect.IOFiber@433688b3 RUNNING: $anonfun$real$1 ...
  _       <- IO.delay(rt.fiberMonitor.liveFiberSnapshot(line => println(line))) // shows 1 fiber in RUNNING state
  _       <- cdl.release // allow further execution
  outcome <- fiber.join
  _       <- IO.println(fiber) // cats.effect.IOFiber@433688b3 COMPLETED
  _       <- IO.delay(rt.fiberMonitor.liveFiberSnapshot(line => println(line))) // shows 0 fibers
} yield ()
```


Even though I have a strong reference on the fiber, it's still being removed from the `WeakBag`. 
Could someone suggest a hint, on how I can keep a completed fiber in a runtime snapshot?

testkit/shared/src/main/scala/cats/effect/testkit/TestControl.scala
```diff
@@ -121,6 +121,7 @@ import java.util.concurrent.atomic.AtomicReference
  */
 final class TestControl[A] private (
     ctx: TestContext,
+    runtime: IORuntime,
```
iRevive: MiMa says this change is not binary compatible.
Is it safe to filter it out, since the constructor is private?

```
ProblemFilters.exclude[DirectMissingMethodProblem]("cats.effect.testkit.TestControl.this")
```

iRevive:
I tested this branch locally and the completed fibers are excluded from the snapshot.

<details>
  <summary>CE 3.3.7</summary>
ce-3 3 7
</details>

<details>
  <summary>Current branch</summary>
ce-exclude-completed-fibers
</details>

Only proper unit testing is left.

testkit/shared/src/main/scala/cats/effect/testkit/TestControl.scala
armanbilge: Is the MiMa error only on Scala 3? We have a special section for Scala 3-only MiMa issues because it does not yet automatically recognize `private`. You can add the filter there.

armanbilge:
> Even though I have a strong reference on the fiber, it's still being removed from the WeakBag.
> Could someone suggest a hint, on how I can keep a completed fiber in a runtime snapshot?

Sorry, I got confused what the problem is. Isn't the goal of this PR that the completed fiber isn't in the snapshot?

iRevive:
> Sorry, I got confused what the problem is. Isn't the goal of this PR that the completed fiber isn't in the snapshot?

Yes, that's correct.

Unfortunately, the unit test and the example I mentioned above are passing even without my changes (e.g. CE 3.3.7).

testkit/shared/src/main/scala/cats/effect/testkit/TestControl.scala
iRevive: The error occurred on 2.13 https://github.com/typelevel/cats-effect/runs/5509038075?check_suite_focus=true#step:13:4155

armanbilge: Oh, interesting, I thought MiMa was smarter than that :) yes, you can add a filter that applies to all Scala versions then, thanks!

iRevive:
I managed to implement a valid unit test eventually. The PR is ready for review.

core/js/src/main/scala/cats/effect/unsafe/FiberMonitor.scala
djspiewak: Eh, I think this is fine. We can reconsider if we have other uses for `liveFibers` in the future.

djspiewak (approved):
Nice job on making this unit testable! I do agree that maybe `liveFibers()` should do the filtering, but we can clean that up later.

PR #2868, opened 2022-03-10
Implemented unhandled error reporting for orphan fibers

### LINKED ISSUE #2842: Missing information about failing background fibers
Tested against 3.3.0 and 3.3.5

Following code doesn't print any information about the fact that an exception was thrown in one of the fibers.

```
object Main extends IOApp {
    Thread.setDefaultUncaughtExceptionHandler((t, e) => println("Uncaught exception in thread: " + t, e))
    for {
      _ <- IO
        .sleep(5.seconds)
        .flatMap(_ => IO.delay(println("after sleep")))
        .flatMap(_ => IO.raiseError(new RuntimeException("something went wrong")))
        .start
      _ <- IO.delay(println("started"))
      _ <- IO.never[Unit]
    } yield ExitCode.Success
  }
}
```

I would expect it to print some information to stdout/stderr. It would be even better if there was a way to customize that information. 

For example in zio there is a `Plaform` object that contains an error reporter which can be customized by the user.
> In the `Platform` that is a part of every runtime, there is an error reporter that will be called by ZIO to report every unhandled error

Link to the original discussion on discord: https://discord.com/channels/632277896739946517/632278585700384799/946443755286855702

djspiewak:
This enriches `CallbackStack` with the ability to report when no callbacks were invoked. Note that this isn't entirely perfect, since a fiber which `join`s and is later `cancel`ed within the `join` will empty out its callback, but the `join`ed fiber may finish with an error at the exact same moment. If this happens and the `join`ing fiber is the *only* fiber, the error would have been side-channeled but might end up being missed. I think this is a fairly reasonable tradeoff though, at least for the moment. The alternative is that we force a memory barrier in this case, which I think is a bit too much deoptimization for this edge case.

Closes #2842

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -454,7 +454,10 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
       ProblemFilters.exclude[MissingClassProblem]("cats.effect.unsafe.WorkerThread$Data"),
       // introduced by #2844, Thread local fallback weak bag
       // changes to `cats.effect.unsafe` package private code
-      ProblemFilters.exclude[MissingClassProblem]("cats.effect.unsafe.SynchronizedWeakBag")
+      ProblemFilters.exclude[MissingClassProblem]("cats.effect.unsafe.SynchronizedWeakBag"),
+      // introduced by #2868
+      // added signaling from CallbackStack to indicate successful invocation
+      ProblemFilters.exclude[DirectMissingMethodProblem]("cats.effect.CallbackStack.apply")
     ) ++ {
       if (isDotty.value) {
         // Scala 3 specific exclusions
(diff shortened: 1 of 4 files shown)


djspiewak:
Test failures here are legit. I think it's a race condition in how I setup the test which is being surfaced randomly by the `TestContext` shuffling.

vasilmkd (approved):

RobertoUa:
Is it possible to override `reportFailure` in the default ctx? For instance, I want to report an exception to Sentry.

armanbilge:
@RobertoUa I _think_ that's what https://github.com/typelevel/cats-effect/pull/2869 is doing, which will land in 3.4.0.

PR #2869, opened 2022-03-10
Added support for configuring compute pool error reporter

djspiewak:
I think this is kind of cool, and particularly the ability to configure error reporting to do something non-trivial and even asynchronous. Unfortunately though, this does come with some drawbacks. In particular, the default error reporter now delegates to `unsafeRunAndForget()`, which is not the cheapest thing in the world. We could make it slightly cheaper by having a version which doesn't use `fiberErrorCbs`.

Additionally, we should probably think careful about whether there are any cases where we use `reportFailure` internally assuming that it *isn't* going to leverage the runtime. I did a quick pass and couldn't find any, but it deserves a second pair of eyes.


djspiewak:
I think this is ready for a second look, @vasilmkd

build.sbt
```diff
@@ -461,7 +461,10 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
       ProblemFilters.exclude[MissingClassProblem]("cats.effect.unsafe.WorkerThread$Data"),
       // introduced by #2844, Thread local fallback weak bag
       // changes to `cats.effect.unsafe` package private code
-      ProblemFilters.exclude[MissingClassProblem]("cats.effect.unsafe.SynchronizedWeakBag")
+      ProblemFilters.exclude[MissingClassProblem]("cats.effect.unsafe.SynchronizedWeakBag"),
+      // introduced by #2869
+      // package-private method
+      ProblemFilters.exclude[DirectMissingMethodProblem]("cats.effect.IO.unsafeRunFiber")
```
armanbilge, line 527: Sorry to be annoying: are we sure this is not Scala 3 only?

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -821,8 +821,17 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
    * Note that errors still get logged (via IO's internal logger), because errors being thrown
    * should never be totally silent.
    */
-  def unsafeRunAndForget()(implicit runtime: unsafe.IORuntime): Unit =
-    unsafeRunAsync(_ => ())
+  def unsafeRunAndForget()(implicit runtime: unsafe.IORuntime): Unit = {
+    val _ = unsafeRunFiber((), _ => (), _ => ())
```
sbly, line 835: Since this is no longer calling delegating to `unsafeRunAsync`, that means it won't throw a `new CancellationException("The fiber was canceled")` when cancelled, correct? Just checking that's intended?

sbly (dismissed):
Looks good to me! Just one minor comment that probably doesn't matter :)

core/shared/src/main/scala/cats/effect/IO.scala
djspiewak: Good catch! That regression was *not* intended.

djspiewak: Oh actually, looking at this more, the cancelation exception was eaten *anyway*, so this is still matching the existing behavior.

sbly: Gotcha, thanks for explaining!

PR #2870, opened 2022-03-11
IORuntimeBuilder: fix typos and update documentation

iRevive:


core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
```diff
@@ -38,11 +38,11 @@ final class IORuntimeBuilder protected (
    * Set a custom compute pool
    *
    * @param compute
-   *   `ExecutionContext` for compute
+   *   the [[scala.concurrent.ExecutionContext ExecutionContext]] for a compute pool
    * @param shutdown
-   *   shutdown hook upon [[IORuntime]] shutdown
+   *   the shutdown logic of the given execution context
```
iRevive: Is there is a better word than `logic`? I was thinking about: `strategy`, `method`.
Also, I can keep `shutdown hook`.

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
```diff
@@ -38,11 +38,11 @@ final class IORuntimeBuilder protected (
    * Set a custom compute pool
    *
    * @param compute
-   *   `ExecutionContext` for compute
+   *   the [[scala.concurrent.ExecutionContext ExecutionContext]] for a compute pool
```
armanbilge:
```suggestion
   *   the [[scala.concurrent.ExecutionContext ExecutionContext]] for the compute pool
```

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeBuilder.scala
armanbilge: IMO I think `shutdown hook` is the right term. For example Java has a ["Shutdown Hooks API"](https://docs.oracle.com/javase/8/docs/technotes/guides/lang/hook-design.html). But maybe it's confusing?

```suggestion
   *   [[IORuntime]] shutdown hook
```

djspiewak (approved):

Issue #2727, opened 2022-01-03
Document most common failure modes when using `Dispatcher` 


LLCampos:
For example, I'm constantly hitting "dispatcher already shutdown" issues, and I always get really confused on why. Some troubleshooting docs would be great.

PR #2875, opened 2022-03-14
Give each Dispatcher worker its own Supervisor

wjoel:
In programs that use the same dispatcher a lot, the `stateRef` in `Supervisor` ends up being a hotspot, and a significant amount of time is spent in `spin` attempting to update it.

While the documentation says that dispatchers are cheap and should be constructed often, web servers like blaze-server, jetty-server, http4s-netty, and fs2-netty, use one `Dispatcher` per server.
I don't think that's wrong, and I've also experienced problems with "dispatcher already shutdown" when trying to use shorter-lived ones, as reported in https://github.com/typelevel/cats-effect/issues/2727

Given that, we can fix the hotspot in `Dispatcher` by creating one `Supervisor` for each CPU/worker, which is what this PR does.

Results (from a Ryzen 5900X, I expect the improvement to scale with the number of cores) look good:

Baseline on `series/3.x`:
```
DispatcherBenchmark.scheduling    1000  thrpt   20  167.852 ± 1.801  ops/min
```
This PR:
```
DispatcherBenchmark.scheduling    1000  thrpt   20  600.486 ± 6.002  ops/min
```

I'm also doing some work with Techempower HTTP benchmarks, and this change results in a 5% increase in performance on plaintext using blaze (but interestingly, gets the best result on 16384 concurrent requests while it otherwise peaks at 4096 concurrent requests), and an even greater increase when using a modified http4s-netty backend.

It's possible that there's a better fix for the underlying issue, but this one is simple, and a 3x improvement in (benchmark) performance makes it worth considering, in my opinion. Even if it does making constructing a `Dispatcher` a tiny bit less cheap.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
@@ -111,7 +111,7 @@ object Dispatcher {
     final case class CancelToken(cancelToken: () => Future[Unit]) extends CancelState
 
     for {
-      supervisor <- Supervisor[F]
+      supervisors <- (0 until Cpus).toList.traverse(_ => Supervisor[F])
       latches <- Resource.eval(F delay {
         val latches = new Array[AtomicReference[() => Unit]](Cpus)
         var i = 0
@@ -139,6 +139,7 @@ object Dispatcher {
 
       _ <- {
         def dispatcher(
+            supervisor: Supervisor[F],
             latch: AtomicReference[() => Unit],
             state: Array[AtomicReference[List[Registration]]]): F[Unit] =
           for {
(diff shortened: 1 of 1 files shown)


wjoel:
I think I'll need some help with that test failure.

djspiewak:
Holy crap. Those numbers are legit.

djspiewak:
I'll take a look at the failure in a bit. Would you mind targeting this at series/3.3.x though? It's forward compatible so we can release it much sooner this way.

wjoel:
> I'll take a look at the failure in a bit. Would you mind targeting this at series/3.3.x though? It's forward compatible so we can release it much sooner this way.

Woops, didn't mean to close this one, but ended up with https://github.com/typelevel/cats-effect/pull/2876 instead.

PR #2878, opened 2022-03-15
Extra-concise version of `IOApp.Simple`

bplommer:
I'm really on the fence about whether this is a good idea or just me being phenomenally lazy, so curious what others think 🤷

### DIFF
--- a/core/js/src/main/scala/cats/effect/IOApp.scala
+++ b/core/js/src/main/scala/cats/effect/IOApp.scala
@@ -288,5 +288,5 @@ object IOApp {
    * A further simplified version of IOApp.Simple, allowing the IO program to be passed as a
    * constructor argument.
    */
-  class Run(val run: IO[Unit]) extends IOApp.Simple
+  class Run(override val run: IO[Unit]) extends IOApp.Simple
 }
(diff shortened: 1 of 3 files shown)


djspiewak:
Lmao. Let's let people bikeshed this a bit.

armanbilge:
What I said on Discord is we could really benefit from a way to integrate with Scala 3 `@main` annotation.

djspiewak:
> What I said on Discord is we could really benefit from a way to integrate with Scala 3 @main annotation.

Yeah that would be a huge win, but I think it's not configurable right now.

mpilquist:
`@main` support would be nice. I don't like having too many ways of doing the same thing though and I'm also a bit concerned about moving the creation of the main `IO` in to a constructor call. Reminiscent of `App`.

bplommer:
> What I said on Discord is we could really benefit from a way to integrate with Scala 3 `@main` annotation.

I think it would be reasonably simple to add syntax for something like `@main def run(arg1: Foo, arg2: Bar) = myIO(arg1, arg2).unsafeRunIOApp` that desugars to something like
```scala
@main def run(arg1: Foo, arg2: Bar) = new IOApp.Simple {
  override val run: IO[Unit] = myIO(arg1, arg2)
}.main(Array.empty)
```

armanbilge:
Scala 3.2.0 now offers https://scala-lang.org/api/3.x/scala/annotation/MainAnnotation.html#

PR #2880, opened 2022-03-17
Use Sync[F].blocking instead of Sync[F].delay for randomUUID

### LINKED ISSUE #2865: UUIDGen does not use Random[F]
> I have seen `UUID.randomUUID` kill more than one service in the years BCE (Before Cats Effect).  The synchronization isn't obvious until you start profiling why you're getting single-threaded performance.

_Originally posted by @rossabaker in https://github.com/typelevel/cats-effect/issues/2852#issuecomment-1059801110_

Given this, we probably want to offer a `Random[F] => UUIDGen[F]` implicit that's higher-priority than the `Sync`-based one, and consider deprecating the `Sync` instance, or at least making it *explicitly* summoned rather than implicit so the user has to opt-in to using the less-safe implementation

BalmungSan:
Relates to https://github.com/typelevel/cats-effect/issues/2865#issuecomment-1062523959

I know it could be better, I just think the quick fix is worth it and it can be superseded later with a better version.

_(I hope I didn't break bincompat with the unrelated changes)_

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/UUIDGen.scala
+++ b/std/shared/src/main/scala/cats/effect/std/UUIDGen.scala
@@ -38,9 +38,11 @@ trait UUIDGen[F[_]] {
 object UUIDGen {
   def apply[F[_]](implicit ev: UUIDGen[F]): UUIDGen[F] = ev
 
-  implicit def fromSync[F[_]: Sync]: UUIDGen[F] = new UUIDGen[F] {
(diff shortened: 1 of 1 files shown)


armanbilge:
Oh, this is super annoying, but can you retarget this at `series/3.3.x`? For the release this weekend :)

djspiewak (approved):

PR #2885, opened 2022-03-19
High(er) performance async `Queue`

### LINKED ISSUE #2771: High-performance `Queue` implementation

djspiewak:
First off, one thing I discovered as part of this is that the existing `Queue` is *really* a lot faster than you probably think. But we can do better.

This relies on the runtime typecase trick on the typeclass to see if the `GenConcurrent` is secretly actually an `Async`. When it is, and when the bound is `> 1`, we transparently swap to this more efficient implementation. Benchmarks to follow, but it's somewhere between 2x and 10x faster, depending on the scenario and the number of physical threads. Implementation is based on jctools for the bounded part, and all credit is due to @viktorklang for the unbounded part. Bugs are probably my fault.

Lots of room to improve here! A non-exhaustive list:

- We can support `unbounded` without too much trouble, just by building on a pair of `UnboundedUnsafe`s
- `circular` is also pretty easy with what we have
- `synchronous` is going to require a brand new structure and a lot of thought. That's a very tricky case
- Specialized versions for ScalaJS would technically be optimal
- The jctools bounded queue does have support for a `takeAll`-like operation, meaning we can override `tryTakeN` and it will be *much* better than the naive version, which should in turn make Fs2 `Channel` vastly better
- ~~Right now, it starts failing if you enqueue more than `Long.MaxValue`. This is easily fixable, I was just lazy~~

An even more substantial optimization will be striping consumers. While this is an mpmc queue, the far-and-away most common scenario is mpsc, which would allow us to optimize it a lot. We can take advantage of that by striping sc queues so long as we don't care *as* much about fairness in that case, and since we know our consuming threads are bounded by the physical threads, we also know our stripes will be strictly bounded.

Anyway, lots of room for future excitement.

Closes #2771


djspiewak:
Auspicious…

djspiewak:
Benchmarks:

```
[info] Benchmark                                         (size)   Mode  Cnt      Score     Error    Units
[info] QueueBenchmark.asyncEnqueueDequeueContended       100000  thrpt   10  21800.912 ± 703.188  ops/min
[info] QueueBenchmark.asyncEnqueueDequeueMany            100000  thrpt   10   9667.855 ± 118.857  ops/min
[info] QueueBenchmark.asyncEnqueueDequeueOne             100000  thrpt   10   9448.161 ± 122.768  ops/min
[info] QueueBenchmark.concurrentEnqueueDequeueContended  100000  thrpt   10   9625.092 ±  61.529  ops/min
[info] QueueBenchmark.concurrentEnqueueDequeueMany       100000  thrpt   10   3647.851 ±  65.173  ops/min
[info] QueueBenchmark.concurrentEnqueueDequeueOne        100000  thrpt   10   3769.850 ±  52.817  ops/min
```

std/shared/src/main/scala/cats/effect/std/Queue.scala
```diff
@@ -314,6 +337,379 @@ object Queue {
...
+
+                // it worked! clear ourselves out of the queue
+                clear()
+                // our retry succeeded
+                succeeded = true
+
+                // manually complete our own callback
+                // note that we could have a race condition here where we're already completed
+                // async will deduplicate these calls for us
+                // additionally, the continuation (below) is held until the registration completes
```
durban: This ("the continuation is held") seems important here. Is this guaranteed by any `Async`?

std/shared/src/main/scala/cats/effect/std/Queue.scala
```diff
@@ -67,11 +70,31 @@ object Queue {
    * @return
    *   an empty, bounded queue
    */
-  def bounded[F[_], A](capacity: Int)(implicit F: GenConcurrent[F, _]): F[Queue[F, A]] = {
+  def bounded[F[_], A](capacity: Int)(implicit F: GenConcurrent[F, _]): F[Queue[F, A]] =
+    F match {
+      case f0: Async[F] =>
```
durban, line 75: This is not strictly related to this PR, but what's stopping someone implementing `Cont` from doing the same, and recovering an `Async[G]` from the `MonadCancel[G, Throwable]`?

std/shared/src/main/scala/cats/effect/std/Queue.scala
```diff
@@ -314,6 +337,379 @@ object Queue {
...
+    private[this] val takers = new UnsafeUnbounded[Either[Throwable, Unit] => Unit]()
+    private[this] val offerers = new UnsafeUnbounded[Either[Throwable, Unit] => Unit]()
+
+    // private[this] val takers = new ConcurrentLinkedQueue[AtomicReference[Either[Throwable, Unit] => Unit]]()
+    // private[this] val offerers = new ConcurrentLinkedQueue[AtomicReference[Either[Throwable, Unit] => Unit]]()
+
+    def offer(a: A): F[Unit] = F defer {
+      try {
+        // attempt to put into the buffer; if the buffer is full, it will raise an exception
+        buffer.put(a)
```
durban: Since `offer` and `take` both first try to use `buffer` I think it can happen that a waiter is bypassed. For example, initially the queue is empty, and there is one taker waiting. Then, `offer` and `take` are racing, and `take` immediately removes the item inserted by `offer`, thus bypassing the waiter, which was earlier than `take`. (Although this whole thing might be unobservable. It seems a "fairness" issue, and not necessarily a "semantics" issue.)

durban (commented):
A general comment: none of these queues seem lock-free (which is not necessarily a problem, it's just an observation.)

djspiewak:
They're definitely all lock free. They aren't _contention_ free though.

std/shared/src/main/scala/cats/effect/std/Queue.scala
djspiewak: It is! And I agree it would be worth spelling it out more. We use this trick a lot though.

std/shared/src/main/scala/cats/effect/std/Queue.scala
djspiewak: Doing so and then setting up weird forking scenarios would result in runtime errors. So, caveat emptor.

std/shared/src/main/scala/cats/effect/std/Queue.scala
djspiewak: You're correct! There are actually a couple cases like this. Strict fairness is not guaranteed in highly contended races.

durban:
Regarding lock freedom:

Let's look at `UnsafeUnbounded`: in `take` an interesting case is when `taken` is `null`, but `last.get() ne null`. In this case what it does is: `take() // Waiting for prevLast.set(cell), so recurse`. It waits on another thread (`put`), by spinning. This is... well, it's a _spinlock_. Acquired by `put` with `last.getAndSet(cell)`, and released either with `first.set(cell)` or with `prevLast.set(cell)`. And `take` is waiting (by spinning) on this lock with the recursive call.

There is another similar case (spinwaiting on another thread) in `take` after the long comment.

In `UnsafeBounded` there is something similar going on (if I understand correctly): `put` acquires a spinlock with `tail.compareAndSet(currentTail, currentTail + 1)`, writes the data into `buffer` while holding the lock, then releases it with `sequenceBuffer.incrementAndGet(project(currentTail))`. `take` waits on this lock by spinning until `seq == currentHead + 1`.

viktorklang:
The spinlock in UnsafeUnbounded is not needed for correctness, only
performance.--
Cheers,
√

djspiewak:
Depends on #3000

djspiewak:
```
[info] Benchmark                                                  (size)   Mode  Cnt      Score     Error    Units
[info] QueueBenchmark.boundedAsyncEnqueueDequeueContended         100000  thrpt   10  29510.955 ± 206.937  ops/min
[info] QueueBenchmark.boundedAsyncEnqueueDequeueMany              100000  thrpt   10   6105.196 ±  27.218  ops/min
[info] QueueBenchmark.boundedAsyncEnqueueDequeueOne               100000  thrpt   10   6261.579 ±   5.845  ops/min
[info] QueueBenchmark.boundedConcurrentEnqueueDequeueContended    100000  thrpt   10  10060.914 ±  44.569  ops/min
[info] QueueBenchmark.boundedConcurrentEnqueueDequeueMany         100000  thrpt   10   4157.941 ±   7.719  ops/min
[info] QueueBenchmark.boundedConcurrentEnqueueDequeueOne          100000  thrpt   10   4278.601 ±  16.290  ops/min
[info] QueueBenchmark.unboundedAsyncEnqueueDequeueContended       100000  thrpt   10  30820.903 ±  93.492  ops/min
[info] QueueBenchmark.unboundedAsyncEnqueueDequeueMany            100000  thrpt   10  16245.842 ±  47.942  ops/min
[info] QueueBenchmark.unboundedAsyncEnqueueDequeueOne             100000  thrpt   10  16594.351 ± 126.705  ops/min
[info] QueueBenchmark.unboundedConcurrentEnqueueDequeueContended  100000  thrpt   10  10073.977 ±  76.811  ops/min
[info] QueueBenchmark.unboundedConcurrentEnqueueDequeueMany       100000  thrpt   10   4155.995 ±   5.937  ops/min
[info] QueueBenchmark.unboundedConcurrentEnqueueDequeueOne        100000  thrpt   10   4274.642 ±   6.468  ops/min
```

Plenty more room to improve, but I'll take a 2-4x as a starting point.

PR #2895, opened 2022-03-20
Use Performance API for high-precision time on JS

armanbilge:
Follow-up to https://github.com/typelevel/cats-effect/pull/2416.

Always an adventure on JS :) there's nothing in the ES standard, but amazingly both [browsers](https://developer.mozilla.org/en-US/docs/Web/API/Performance) and [Node.js](https://nodejs.org/api/perf_hooks.html#performance-measurement-apis) have implemented the same APIs for once! With of course the caveat that for whatever reason [`performance.timeOrigin`](https://developer.mozilla.org/en-US/docs/Web/API/Performance/timeOrigin) is "experimental technology" on the web 🙃


djspiewak (dismissed):

armanbilge:
One learning from this PR is that `IO.cede` on browsers very nearly takes a millisecond ...

armanbilge:
Urgh, extremely annoying! On browsers `IO.cede` is too slow; on Node.js, without it is too fast.

armanbilge:
Hmm, even with the `cede` Node.js is having issues :( I can't replicate locally though.
```
[error]   x use high-precision time
[error]    0 microseconds is equal to 0 nanoseconds (/home/runner/work/cats-effect/cats-effect/tests/js/target/scala-2.13/cats-effect-tests-test-fastopt/https:
```

Update: the problem is with Node 14. Hmmm.

armanbilge:
Node 14 is LTS for one more year. I guess we should keep it in CI :(
https://nodejs.org/en/about/releases/

core/js/src/main/scala/cats/effect/unsafe/SchedulerCompanionPlatform.scala
```diff
@@ -38,8 +40,30 @@ private[unsafe] abstract class SchedulerCompanionPlatform { this: Scheduler.type
...
         def monotonicNanos() = System.nanoTime()
-        override def nowMicros(): Long = nowMillis() * 1000
+        override def nowMicros(): Long = nowMicrosImpl()
       },
       () => ())
 
+  private[this] val nowMicrosImpl: () => Long = {
+    def test(performance: Performance): Try[Performance] = Try {
+      // take it for a spin
+      require(!(performance.timeOrigin + performance.now()).isNaN)
```
djspiewak: Exceptions are pretty expensive. Can we turn this into a simple `if`/`else`?

djspiewak (changes requested):

core/js/src/main/scala/cats/effect/unsafe/SchedulerCompanionPlatform.scala
armanbilge: Even for a one-time initialization?

armanbilge: Yeah, you're probably right :( I've used `Try` in many places, maybe I should dial those back.

armanbilge:
This is looking to be flaky on Chrome. Unfortunately the resolution there is 100 microseconds (versus 5 microseconds on Firefox with `privacy.reduceTimerPrecision=false`). I guess I can try `cede` again, but that may be dangerously slow 😅

djspiewak (dismissed):

djspiewak:
@armanbilge This is ready to go once the conflicts are resolved.

djspiewak (approved):

PR #2897, opened 2022-03-21
Docs: Use "fiber blocking" instead of "semantic blocking"

Daenyth:
From the discord discussions, everyone seems to like this phrasing better - it
helps avoid confusion for newcomers, about the difference between thread
blocking and fiber blocking

### DIFF
--- a/docs/thread-model.md
+++ b/docs/thread-model.md
@@ -16,13 +16,13 @@ many of our users are already at least somewhat familiar with it.
 The high-level goals of threading are covered in detail by [Daniel's gist](https://gist.github.com/djspiewak/46b543800958cf61af6efa8e072bfd5c) so I'll
 just give the executive summary. We are aiming for:
 
- * A single thread pool of roughly the number of available processors for compute-based operations
+* A single thread pool of roughly the number of available processors for compute-based operations
    (depending on your application you may get better performance by leaving one or two cores
    free for GC, etc)
- * An unbounded, cached threadpool for blocking operations
- * 1 or 2 high-priority threads for handling asynchronous I/O events, the handling of which should
+* An unbounded, cached threadpool for blocking operations
+* 1 or 2 high-priority threads for handling asynchronous I/O events, the handling of which should
    immediately be shifted to the compute pool
-  
+
 The goal of this is to minimize the number of expensive thread context shifts
 and to maximize the amount of time that our compute pool is doing useful work.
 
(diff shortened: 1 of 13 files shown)


docs/thread-model.md
```diff
@@ -104,40 +105,40 @@ unbounded, cached threadpool and then shift computation back to the compute pool
...
 
-### Semantic blocking
+### Fiber blocking
 
 Of course, we do also need the ability to tell fibers to wait for conditions to
 be fulfilled. If we can't call thread blocking operations (eg Java/Scala builtin
 locks, semaphores, etc) then what can we do? It seems we need a notion of
 _semantic_ blocking, where the execution of a fiber is suspended and control of
-the thread it was running on is yielded.
+the thread it was running on is yielded. This concept is called "fiber blocking" in cats-effect.
```
armanbilge, line 114: We should make the connection to the old name, something like this I reckon.

```suggestion
the thread it was running on is yielded. This concept is called "fiber blocking" in cats-effect (previously described as "semantic blocking").
```

Daenyth: Do you think that's still needed in the context of the previous sentence?

> It seems we need a notion of
_semantic_ blocking, where the execution of a fiber is suspended and control of
the thread it was running on is yielded. This concept is called "fiber blocking" in cats-effect.

armanbilge: Oh, missed that, sorry. Hm. I think it's still good to call this out explicitly, like as if we renamed a method. Maybe even in the section title.

docs/thread-model.md
```diff
@@ -334,16 +336,16 @@ the event that the fiber is canceled.
...
+* This means that we can implement thread affinity, where a fiber that yields is most
    likely to be re-scheduled on the same thread. This makes yielding much cheaper
    as if the fiber is immediately re-scheduled we don't even have to flush CPU caches
- * Consequently we can support auto-yielding where a fiber will insert an `IO.cede`
+* Consequently we can support auto-yielding where a fiber will insert an `IO.cede`
    every fixed number of iterations of the runloop, stopping a rogue cpu-bound fiber
    from inadvertently pinning a CPU core
 
-## And that's it!
+## And that's it
```
armanbilge: Don't like exclamations? 😆

Daenyth: Accidental

Daenyth: Actually I see now it's vscode's markdown formatter removing it automatically - apparently trailing punctuation is a lint

armanbilge (approved):

djspiewak (approved):

PR #2873, opened 2022-03-12
The `WSTP` can run Futures just as fast as `ExecutionContext.global`

vasilmkd:
Not ready for prime time, opening to show some benchmark results. If no one objects, I would still need to fix live fiber snapshots (which I mostly commented out) and handling of fatal exceptions.

If anyone feels like bikeshedding, I would appreciate input on how to (or if at all) present external runnables, since they are not instances of `IOFiber`.

PR #2898, opened 2022-03-22
Use `justFibers` helper in `foreignFibers`

aeons:
The `justFibers` helper introduced in #2873 is used in all relevant places except for one.

This PR swithes the last place to use the helper function.


vasilmkd (approved):
Good catch. Thanks!

Issue #2902, opened 2022-03-24
Random.javaSecuritySecureRandom doesn't link on Scala.js


rossabaker:
```scala
//> using scala "3.1.1"
//> using lib "org.typelevel::cats-effect:3.3.8"

import cats.effect._, std._

@main def oops = Random.javaSecuritySecureRandom[SyncIO].unsafeRunSync()
```

Compiles, but doesn't link.  Probably should be in a platform trait?

```console
$ scala-cli run oops.scala --js
Compiling project (Scala 3.1.1, Scala.JS)
Compiled project (Scala 3.1.1, Scala.JS)
Referring to non-existent class cats.effect.std.Random
  called from static oops.main([java.lang.String)void
  called from core module module initializers
Referring to non-existent class cats.effect.std.Random
  called from oops$package$.oops()cats.effect.std.Random
  called from static oops.main([java.lang.String)void
  called from core module module initializers
involving instantiated classes:
  oops$package$
Referring to non-existent class cats.effect.SyncIO
  called from oops$package$.oops()cats.effect.std.Random
  called from static oops.main([java.lang.String)void
  called from core module module initializers
involving instantiated classes:
  oops$package$
Referring to non-existent class cats.effect.std.Random$
  called from oops$package$.oops()cats.effect.std.Random
  called from static oops.main([java.lang.String)void
  called from core module module initializers
involving instantiated classes:
  oops$package$
Referring to non-existent class cats.effect.std.Random
  called from oops$package$.oops()cats.effect.std.Random
  called from static oops.main([java.lang.String)void
  called from core module module initializers
involving instantiated classes:
  oops$package$
Referring to non-existent class cats.effect.SyncIO$
  called from oops$package$.oops()cats.effect.std.Random
  called from static oops.main([java.lang.String)void
  called from core module module initializers
involving instantiated classes:
  oops$package$
Referring to non-existent class cats.effect.std.Random$
  called from oops$package$.oops()cats.effect.std.Random
  called from static oops.main([java.lang.String)void
  called from core module module initializers
involving instantiated classes:
  oops$package$
Referring to non-existent method cats.effect.std.Random$.javaSecuritySecureRandom(cats.effect.kernel.Sync)java.lang.Object
  called from oops$package$.oops()cats.effect.std.Random
  called from static oops.main([java.lang.String)void
  called from core module module initializers
involving instantiated classes:
  oops$package$
Referring to non-existent class cats.effect.SyncIO
[...]

PR #2904, opened 2022-03-24
Fix CompletableFuture cancelation

### LINKED ISSUE #2903: `Async.fromCompletableFuture` should block if cancelation fails
`Async.fromCompletableFuture` has a [cancel operation](https://github.com/typelevel/cats-effect/blob/a045918379d69c699e8d7b47e6d81dc112aa1eb1/kernel/jvm/src/main/scala/cats/effect/kernel/AsyncPlatform.scala#L45) that completes after calling `cancel` on the underlying `CompletableFuture`, but if the cancelation fails then `CompletableFuture.cancel` will immediately return `false` - fiber cancelation in this case should block until the future completes.

bplommer:
I think this fixes #2903 but I haven't yet written a regression test.

### DIFF
--- a/kernel/jvm/src/main/scala/cats/effect/kernel/AsyncPlatform.scala
+++ b/kernel/jvm/src/main/scala/cats/effect/kernel/AsyncPlatform.scala
@@ -42,7 +42,14 @@ private[kernel] trait AsyncPlatform[F[_]] extends Serializable { this: Async[F] 
               }))
           }
 
-          Some(void(delay(cf.cancel(false))))
+          Some(flatMap(delay(cf.cancel(false))) {
+            case true => unit
+            case false => // failed to cancel - block until completion
+              async_[Unit] { cb =>
+                cf.handle[Unit]((_, _) => cb(Right(())))
+                ()
+              }
+          })
         }
       }
     }


bplommer:
cc @Daenyth

vasilmkd:
LGTM, but a test would be really good for this.

armanbilge:
We discussed how to write a test on Discord.
https://discord.com/channels/632277896739946517/632278585700384799/956566207211905115

The tl;dr is create an uncancelable "never" `CompletableFuture` that will not complete, and always returns `false` when attempts to be canceled. The correct implementation should get backpressured on cancellation, whereas the broken one should terminate.

bplommer:
There was some discussion on Discord about how to test this - I can't promise to do that in the next couple of days, so feel free to fork this PR with a test.

Edit: what @armanbilge said

vasilmkd:
@armanbilge Can you please handle this? Thanks.

armanbilge:
Yup I'm working on https://github.com/typelevel/cats-effect/issues/2902, I can do this next.

vasilmkd:
Closing in favor of #2907.

Issue #2882, opened 2022-03-18
Make `UUIDGen` more performant


armanbilge:
As of https://github.com/typelevel/cats-effect/pull/2880 it delegates to `Sync#blocking`. 

Some ideas in https://github.com/typelevel/cats-effect/issues/2865#issuecomment-1062567227 to improve the status quo:
> While `UUID.randomUUID` is blocking, we can do better:
>
> * Try to create a `SecureRandom.getInstance("NativePRNGNonBlocking")`.  `"NativePRNG"` appears to be equivalent on `nextBytes` on Unix, but that's platform specific. 
> * Until Java 9, there's still a mutex, but [a pool](https://github.com/f4b6a3/uuid-creator/blob/dbc54155c6cb2893eff99876b75ead20382fe331/src/main/java/com/github/f4b6a3/uuid/factory/function/impl/DefaultRandomFunction.java) fixes that.

rossabaker:
http4s/http4s#6138 is another hidden instance of this problem.  Maybe what we really need is a way of safely and efficiently generating random bytes.  UUIDGen, and this http4s use case, could both build on that.

armanbilge:
Yes, I've wondered if we need a `SecureRandom[F]` type class that extends `Random[F]`. But btw, at what point does this start scope-creeping outside of CE3? If we get a `SecureRandom` it will need a JS implementation, which gets into [bobcats territory](https://github.com/typelevel/bobcats/blob/fbaebfe60a9faf3870c51faa8d5d5f224bf7428c/core/js/src/main/scala/bobcats/unsafe/SecureRandom.scala#L27) with different implementations for Node.js and browsers and a CI matrix spanning all of them.

armanbilge:
Ross has demoed his proposal in https://github.com/http4s/http4s/pull/6165.

djspiewak:
I rather like Ross' proposal.

rossabaker:
I reformed it a bit to add the pool.  This is a win when we get caught in a mutex, which we always will in Java 8.  In Java 9, if we have a non-blocking but threadsafe instance, we probably don't want that pool.  But detecting thread safety is [super gross](https://github.com/openjdk/jdk/blob/4df67426ed02f18af0757897acb28b636a317a91/src/java.base/share/classes/java/security/SecureRandom.java#L229-L236).

rossabaker:
Would `SecureRandom` have any additional methods, or just a marker trait that it's, like, secure and stuff?

[...]

PR #2905, opened 2022-03-24
Add a SecureRandom algebra

rossabaker:
Adds a `SecureRandom` extension to the `Random` algebra.  It's a marker trait, for clients that need more assurance of cryptographic strength.

Open questions:
* [ ] Do we want a `nextBits(n: Int)` now, while we can?  It's not specific to secure randoms.
* [ ] Do we want a `nextBigInt(n: Int)` now, while we can?  It's not specific to secure randoms.
* [ ] Do we want to expose `algorithm`?  If so, it will cost us the pooled instance.  I don't think it's all that useful after #2882.


rossabaker:
* As is, this will fail on Scala.js like #2902 did.
* I thought about an `algorithm: String`, but that gets janky with the round-robin implementations.
* This is our opportunity to add `nextBits(n: Int): F[String]` without a breaking change.  That's also janky because we have to request bytes and throw away up to 7 bits when the Java method is _right there_ but protected. 😠 
* This is our opportunity to add `nextBigInt: F[BigInt]` without a breaking change, but I replaced my itch for that with `nextBits`.
* This is where the blocking enhancements of #2882 would go, but I _think_ that could still be a post-release improvement.

rossabaker:
Regarding `nextBits` and `nextBigInt`, there's nothing "Secure" about them.  They weren't thought of in time for `Random`, but are applicable to those, and could be derived methods on the `Random` object.  Putting them here would really only be for syntax.

I also don't see a method that lifts an extant `java.security.SecureRandom` in the original `Random`.  That should also be added here.  I don't see how else we can customize the algorithm today without mucking globally with Java Security Providers.

rossabaker:
I'm unqualified to comment on #2906, but I'll rebase this on the results of that.

armanbilge:
We should also point the browser CI at the new spec.
https://github.com/typelevel/cats-effect/blob/ac120aa7184e11af2dcb8e512598607c53768f2b/project/CI.scala#L73

rossabaker:
I expected that to change some YAML too.  I'm not familiar with that file.

rossabaker:
Is `JavaSecureRandom` in `RandomCompanionPlatform` on the JVM and in the `std` package on JS?  I am confused.

armanbilge:
Yes. On JVM `JavaSecureRandom` is a type alias for `java.security.SecureRandom`. On JS it's a shim that's implemented in `cats.effect.std`. Honestly they can be named/organized however.

rossabaker:
I must be missing something.  I thought the point was that they had the same FQCN.  How are we supposed to point to both of them in the shared code?

rossabaker:
Uh-oh.  I think it works on `Random` because it's in the same package for JS and in the same object for the JVM.  That doesn't work for `SecureRandom`.

armanbilge:
Yeah, sorry, I wasn't thinking ahead to your PR very much 😅 

Can `SecureRandom` get the platform trait instead? And then the deprecated methods in `Random` can delegate to `SecureRandom` instead so they don't need to know about `JavaSecureRandom`?

rossabaker:
That would create a circular reference between `Random` and `SecureRandom` I was trying to avoid.  They're in the same package, so probably not the worst thing.

armanbilge:
If `SecureRandom`'s constructors get smart about selecting a non-blocking implementation, we probably don't want to duplicate that logic in `Random`'s deprecated constructors, so I think it was inevitable 😅

std/js/src/main/scala/cats/effect/std/RandomCompanionPlatform.scala
```diff
@@ -16,4 +16,5 @@
 
 package cats.effect.std
 
+// Vestigial shim
```
rossabaker, line 19: I couldn't get rid of these without angering MiMa, even though I could empty them on both platforms.

std/js/src/main/scala/cats/effect/std/SecureRandomCompanionPlatform.scala
```diff
@@ -0,0 +1,61 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.std
+
+import scala.scalajs.js
+
+private[std] trait SecureRandomCompanionPlatform {
+  private[std] final class JavaSecureRandom extends java.util.Random {
```
rossabaker: This move is so they have the same FQCN.

std/js/src/main/scala/cats/effect/std/RandomCompanionPlatform.scala
armanbilge: That's ok. We'll need them again to deal with all the other stuff that doesn't link in https://github.com/typelevel/cats-effect/issues/2909.

std/shared/src/main/scala/cats/effect/std/Random.scala
```diff
@@ -332,20 +278,15 @@ object Random extends RandomCompanionPlatform {
   def javaUtilConcurrentThreadLocalRandom[F[_]: Sync]: Random[F] =
     new ThreadLocalRandom[F] {}
 
+  @deprecated("Call SecureRandom.javaSecuritySecureRandom", "3.4.0")
   def javaSecuritySecureRandom[F[_]: Sync](n: Int): F[Random[F]] =
```
armanbilge, line 326: So actually thanks to type erasure we could tighten the return type here to `F[SecureRandom[F]]`. But given that this method is deprecated and that these sorts of tricks can cause source-breakage or type-inference issues probably not worth it.

std/shared/src/main/scala/cats/effect/std/SecureRandom.scala
```diff
@@ -0,0 +1,126 @@
...
+    for {
+      ref <- Ref[F].of(0)
+      array <- Sync[F].delay(Array.fill(n)(new SRandom(new JavaSecureRandom())))
+    } yield {
+      def incrGet = ref.modify(i => (if (i < (n - 1)) i + 1 else 0, i))
+      def selectRandom = incrGet.map(array(_))
+      new ScalaRandom[F](selectRandom) with SecureRandom[F] {}
+    }
+
+  def javaSecuritySecureRandom[F[_]: Sync]: F[SecureRandom[F]] =
```
armanbilge, line 122: I assume we will port the ideas from https://github.com/http4s/http4s/pull/6165 in future PRs since that can go into any patch. Seems a good-first-issue for a contributor :)

armanbilge (dismissed):
👍 I think this is a very important addition.

Regarding `nextBits` and `nextBigInt`: I think we should add these on `Random` itself. To do so compatibly we'll have to clumsily ask for constraints on the methods so we can provide default implementations in the parent trait. There's precedent doing this to add methods to `Queue`.

std/shared/src/main/scala/cats/effect/std/Random.scala
rossabaker: Yeah, I think I don't want to think about all that.

rossabaker:
So `getAlgorithm` (for approximate parity with the Java, but no hugely compelling use case) is the big question mark.  We can't add that abstractly after we hit the big green button.

djspiewak (dismissed):
I like this change quite a bit. I defer on whether `getAlgorithm` is meaningful (though as a moment of bikeshedding, I would say `algorithm` might be a more idiomatic name).

armanbilge:
I can fix the conflicts since they are my fault.

djspiewak (dismissed):

djspiewak (dismissed):

armanbilge (dismissed):
Finally!

rossabaker:
> I defer on whether getAlgorithm is meaningful 

It's probably mostly not, but I did use Java's to [determine whether the Java implementation is threadsafe](https://github.com/http4s/http4s/blob/c68e46175f3e7e2d2cf00d5755072e86ed1ecd0c/core/src/main/scala/org/http4s/internal/Random.scala#L77-L80).  Hopefully I suffered so the Scala users don't have to.

djspiewak:
@rossabaker Approved for merge once conflicts are fixed

armanbilge:
I'll fix the conflicts. Again 🙃

PR #2906, opened 2022-03-24
Implement `Random#javaSecuritySecureRandom` on Scala.js

### LINKED ISSUE #2902: Random.javaSecuritySecureRandom doesn't link on Scala.js

armanbilge:
Fixes https://github.com/typelevel/cats-effect/issues/2902 by implementing our own `java.security.SecureRandom` shim that delegates to browser and Node.js crypto APIs.

While working on this I noticed there are other APIs that won't link on Scala.js, like the thread local stuff. Those really should be moved to platform traits in a PR targeting series/3.x.

### DIFF
--- a/std/js/src/main/scala/cats/effect/std/SecureRandom.scala
+++ b/std/js/src/main/scala/cats/effect/std/SecureRandom.scala
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2020-2022 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.std
+
+import java.util.Random
+import scala.scalajs.js
+
+private final class SecureRandom extends Random {
+
+  private val nextBytes: Int => js.typedarray.Int8Array =
+    if (js.typeOf(js.Dynamic.global.crypto) != "undefined") // browsers
+      { numBytes =>
+        val bytes = new js.typedarray.Int8Array(numBytes)
+        js.Dynamic.global.crypto.getRandomValues(bytes)
+        bytes
+      } else {
+      val crypto = js.Dynamic.global.require("crypto")
+
+      // Node.js
+      { numBytes =>
+        val bytes = crypto.randomBytes(numBytes).asInstanceOf[js.typedarray.Uint8Array]
+        new js.typedarray.Int8Array(bytes.buffer, bytes.byteOffset, bytes.byteLength)
+
+      }
+    }
+
+  override def nextBytes(bytes: Array[Byte]): Unit = {
+    nextBytes(bytes.length).copyToArray(bytes)
+    ()
+  }
+
+  override protected final def next(numBits: Int): Int = {
+    val numBytes = (numBits + 7) / 8
+    val b = new js.typedarray.Int8Array(nextBytes(numBytes).buffer)
+    var next = 0
+
+    var i = 0
+    while (i < numBytes) {
+      next = (next << 8) + (b(i) & 0xff)
+      i += 1
+    }
+
+    next >>> (numBytes * 8 - numBits)
+  }
+
+}
--- a/std/js/src/main/scala/cats/effect/std/SecureRandom.scala
+++ b/std/js/src/main/scala/cats/effect/std/SecureRandom.scala
@@ -16,10 +16,9 @@
 
 package cats.effect.std
 
-import java.util.Random
 import scala.scalajs.js
 
-private final class SecureRandom extends Random {
+private final class SecureRandom extends java.util.Random {
 
   private val nextBytes: Int => js.typedarray.Int8Array =
     if (js.typeOf(js.Dynamic.global.crypto) != "undefined") // browsers
--- a/project/CI.scala
+++ b/project/CI.scala
@@ -69,7 +69,7 @@ object CI {
         command = "ciFirefox",
         rootProject = "rootJS",
         jsEnv = Some(JSEnv.Firefox),
-        testCommands = List("testOnly *tracing*"),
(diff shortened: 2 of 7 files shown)


std/js/src/main/scala/cats/effect/std/SecureRandom.scala
```diff
@@ -0,0 +1,59 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.std
+
+import scala.scalajs.js
+
+private final class SecureRandom extends java.util.Random {
```
rossabaker: If this gets called `SecureRandom`, the algebra in #2902 will need another name.

armanbilge: Oh yeah. We can rename it `UnsafeSecureRandom` or `JdkSecureRandom`.

rossabaker: I wouldn't call it `Jdk` when it doesn't come from one, but it's private.  I just mostly want it to be called something else. 😆

vasilmkd: @armanbilge Do you want to do the renaming now, or should we press on? Otherwise, this is good to go.

std/js/src/main/scala/cats/effect/std/SecureRandom.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  private val nextBytes: Int => js.typedarray.Int8Array =
+    if (js.typeOf(js.Dynamic.global.crypto) != "undefined") // browsers
```
armanbilge: There's a trade-off initializing this in the class instead of the companion object. Caching is good, but this is not unreasonable if `SecureRandom` instances are not created very often.

Lazy initialization is important for environments that don't have _any_ secure random capability, since there is no reasonable fallback implementation we can use in those situations.

std/js/src/main/scala/cats/effect/std/SecureRandom.scala
armanbilge: I'll rename it `JavaSecureRandom`.

vasilmkd (approved):

PR #2907, opened 2022-03-24
Fix `Async#fromCompletableFuture` cancelation

armanbilge:
Fixes https://github.com/typelevel/cats-effect/issues/2903. Supersedes https://github.com/typelevel/cats-effect/pull/2904.

The bad news is, my test doesn't pass with @bplommer's fix 🤔 opening in case anyone has some ideas.

/cc @Daenyth for the test


armanbilge:
Update: I was being dumb and not giving a chance for the fiber to start before canceling. Needed an `IO.cede`.

armanbilge:
Wait, nope, still broken 🤔

vasilmkd (approved):

PR #2908, opened 2022-03-24
Use specs2 on Scala 3 without `for3Use2_13`

armanbilge:
Bid farewell to `for3Use2_13`!


djspiewak:
This is very cool. We can't merge it because of the snapshot, but it's a lovely step in the right direction.

@etorreborre Is the plan to eventually release this in a regular version? Like 4.15 maybe?

armanbilge:
The problem is IIUC it's not binary-compatible with the 4.x series, because it's missing things that did not cross-compile.

djspiewak:
Doesn't Specs2 break bincompat between minor versions? I thought it was using pvp rather than semver. Regardless though, I think that's less of a problem overall since it's just a test dependency in most of its downstream consumers.

etorreborre:
@djspiewak @armanbilge not the best situation indeed:

 1. there's the regular 4.x series for Scala 2 with full support for everything
 2. there's a 4.x-cross series for Scala 2/3 with partial support and impacted type inference 
 3. there's the 5.x series for Scala 3 with some breaking changes compared to 4.x and 4.x-cross

I'm already not a fan of having to support 3 different branches but I don't know what else to do to satisfy everyone. 

In terms of versions maybe I can change the artifact names to be: `specs2-cross-core`, `specs2-cross-matcher`, etc... with regular version numbers? Maybe that would make more sense since the feature set is different?

djspiewak:
@etorreborre It kind of seems to me like the cross branch should subsume 4.x entirely, since it gives you mostly source-compatible cross building while transitioning progressively towards 5.x. The downside is that the version number sequencing is kind of messed up now :-( I'm really not sure what to do, honestly.

etorreborre:
> It kind of seems to me like the cross branch should subsume 4.x entirely, since it gives you mostly source-compatible cross building while transitioning progressively towards 5.x.

That would make my life easier but this will force people to start transitioning which possibly some people don't want to do until the last minute.

> The downside is that the version number sequencing is kind of messed up now 

I can start the cross versions at `4.20` and we forget about `4.14.1-cross`. 

I can also raise the level of hackiness and say:

 - from now on `4.20.x` to `4.99.x` will be Scala 2 only with as many fixes as we can cram in those version numbers. After that, it's too late you have to migrate, there will be no more fixes
 - `4.1xx.x` is for the Scala 2 / Scala 3 versions with some degraded functionalities depending on what you use
 - `5.x` is the Scala 3 new thing

That feels quite dirty but maybe pragmatic enough in that situation. You have a better pulse on the community these days than me so I will trust your judgement if you want to recommend something

armanbilge:
Actually, I have an idea: at least from our experiment, it seems that this 4.14.1-cross is source-compatible-enough with 4.14.1 (as in, I didn't change a single test).

So ... what if 4.x continue as usual, publishing only for Scala 2. And the 4.x-cross branch only publishes for Scala 3, with versions exactly tracking the 4.x branch (i.e., no `-cross` suffix). This way we don't have to hack the versioning. WDYT?

Edit: btw, to reduce the 3-branch problem, the 4.x-cross branch can be subsumed by the 4.x branch by either sharing sources, or where that's not possible, splitting sources into scala-2 and scala-3 directories. The scala-2 sources should remain unchanged, but the scala-3 sources may have degraded functionality.

symbiont-eric-torreborre:
> So ... what if 4.x continue as usual, publishing only for Scala 2. And the 4.x-cross branch only publishes for Scala 3, with versions exactly tracking the 4.x branch (i.e., no -cross suffix). This way we don't have to hack the versioning. WDYT?

That's a very good idea!

> Edit: btw, to reduce the 3-branch problem, the 4.x-cross branch can be subsumed by the 4.x branch by either sharing sources, or where that's not possible, splitting sources into scala-2 and scala-3 directories. The scala-2 sources should remain unchanged, but the scala-3 sources may have degraded functionality.

I'm afraid that this will turn into something a bit hard to follow and maintain, I'm going to try your first idea.

etorreborre:
@armanbilge here we go, I published `4.15.0` from both branches

armanbilge:
@etorreborre fantastic! We can cut a proper discipline-specs2 release with that and we should be solid!

- https://github.com/typelevel/discipline-specs2/pull/351

build.sbt
```diff
@@ -256,9 +256,11 @@ ThisBuild / autoAPIMappings := true
 val CatsVersion = "2.7.0"
 val Specs2Version = "4.16.0"
 val ScalaCheckVersion = "1.16.0"
-val DisciplineVersion = "1.3.1"
+val DisciplineVersion = "1.5-a28cf96-SNAPSHOT"
```
armanbilge: @vasilmkd once we get that discipline-specs2 release we can finally land this, hooray!

djspiewak (approved):
This is great! Thank you both so much! @etorreborre @armanbilge

PR #2914, opened 2022-03-25
High(er) performance unbounded queue

### LINKED ISSUE #2886: Optimized async unbounded `Queue`
Using the same trick (and same primitives) that we used in #2885 for `bounded`, implement an `unbounded` queue for `Async[F]`. This should essentially be two `UnsafeUnbounded`s glued together: `takers` and `values`.

djspiewak:
Builds on #2885 to handle `unbounded` cases. The performance gap between this and the `Concurrent` version is even larger than with the bounded queue.

Closes #2886


djspiewak:
Depends on #3000

PR #2917, opened 2022-03-26
High(er) performance `tryTakeN` for `Queue.bounded`

### LINKED ISSUE #2890: Implement optimized `tryTakeN` for async `Queue.bounded`
Following up from #2885, it should be possible to implement a very fast `tryTakeN` by porting `drain` from the jctools `MpmcArrayQueue`. This should result in a pretty dramatic optimization of Fs2's `Channel`, which is at the heart of a lot of programs now.

djspiewak:
Ports the `drain` function from the jctools `MpmcArrayQueue`. This basically just fast-paths the single-consumer bulk consumption scenario (which is to say, Fs2's `Channel`).

Builds on #2885 
Closes #2890


djspiewak:
Depends on #3000

tests/shared/src/test/scala/cats/effect/std/QueueSpec.scala
```diff
@@ -124,7 +124,7 @@ class BoundedQueueSpec extends BaseSpec with QueueTests[Queue] {
     }
 
     "offer/take at high contention" in real {
-      val size = 100000
+      val size = 10000
```
armanbilge: If you want to keep it big for JVM.

```suggestion
      val size = if (cats.platform.Platform.isJVM) 100000 else 10000
```

djspiewak: Wait `cats.platform.Platform` is public?

armanbilge: It comes from `cats-kernel-laws` and is `private[cats]`. Good for us.

PR #2918, opened 2022-03-26
Make Queue#tryTakeN, tryTakeFrontN and tryTakeBackN return F[List[A]]

### LINKED ISSUE #2915: Swap `Queue#tryTakeN`'s signature to `F[List[A]]`
I don't think the `Option` really serves any purpose other than increasing boxing. We can still do this because it hasn't been released yet!

leusgalvan:
Fixes https://github.com/typelevel/cats-effect/issues/2915. 

Decided to tackle all tryTake methods because it felt more consistent.


djspiewak (approved):
This is awesome, thank you!

PR #2919, opened 2022-03-26
Add "with Executor" to ExecutionContexts that are also Executors

wjoel:
I don't know if this a good idea or not, to be honest, but since https://github.com/typelevel/cats-effect/pull/2873 it might be?

### DIFF
--- a/core/js/src/main/scala/cats/effect/unsafe/FiberAwareExecutionContext.scala
+++ b/core/js/src/main/scala/cats/effect/unsafe/FiberAwareExecutionContext.scala
@@ -20,7 +20,11 @@ package unsafe
 import scala.collection.mutable
 import scala.concurrent.ExecutionContext
 
-private final class FiberAwareExecutionContext(ec: ExecutionContext) extends ExecutionContext {
+import java.util.concurrent.Executor
+
+private final class FiberAwareExecutionContext(ec: ExecutionContext with Executor)
+    extends ExecutionContext
+    with Executor {
 
   def liveFibers(): Set[IOFiber[_]] = fiberBag.toSet
 
--- a/core/js/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
+++ b/core/js/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
@@ -23,9 +23,11 @@ import org.scalajs.macrotaskexecutor.MacrotaskExecutor
 import scala.concurrent.ExecutionContext
 import scala.scalajs.LinkingInfo
 
+import java.util.concurrent.Executor
+
 private[unsafe] abstract class IORuntimeCompanionPlatform { this: IORuntime.type =>
 
-  def defaultComputeExecutionContext: ExecutionContext =
+  def defaultComputeExecutionContext: ExecutionContext with Executor =
     if (LinkingInfo.developmentMode && TracingConstants.isStackTracing && FiberMonitor.weakRefsAvailable)
       new FiberAwareExecutionContext(MacrotaskExecutor)
     else
--- a/core/js/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
+++ b/core/js/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
@@ -19,11 +19,14 @@ package unsafe
 
 import scala.concurrent.ExecutionContext
 
+import java.util.concurrent.Executor
+
 // Can you imagine a thread pool on JS? Have fun trying to extend or instantiate
 // this class. Unfortunately, due to the explicit branching, this type leaks
 // into the shared source code of IOFiber.scala.
 private[effect] sealed abstract class WorkStealingThreadPool private ()
-    extends ExecutionContext {
+    extends ExecutionContext
+    with Executor {
   def execute(runnable: Runnable): Unit
   def reportFailure(cause: Throwable): Unit
   private[effect] def reschedule(runnable: Runnable): Unit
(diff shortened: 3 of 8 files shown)


wjoel:
Is there any way to fix the bincompat issue for Scala 3? Is there any interest in this addition, or should I just close the PR?

armanbilge:
Bincompat problems are (often) fixable. I feel like you are reaching for `ExecutionContextExecutor` maybe?
https://www.scala-lang.org/api/2.12.0/scala/concurrent/ExecutionContextExecutor.html

armanbilge:
Bincompat problems are (often) fixable. I feel like you are reaching for `ExecutionContextExecutor` maybe?
https://www.scala-lang.org/api/2.12.0/scala/concurrent/ExecutionContextExecutor.html

wjoel:
@armanbilge I started out with that, but it was not accepted at all by bincompat (even on 2.13, etc.) which is why I changed to `with Executor`. That works for all versions except Scala 3, which I'm guessing is because it ends up being ordered differently in the JVM bytecode produced by the Scala 3 compiler?

I'm pretty clueless about bincompat in general, but it makes sense to me that changing to a different trait like `ExecutionContextExecutor` might not be acceptable, but adding an additional trait should be, only in the case of Scala 3 it looks like it's changing the "base" to `Executor` instead of just appending it at the end.

joroKr21:
Scala 2 erases `with` to the first element but Scala 3 takes both into account.
I'm not sure why it picks `Executor` in this case - maybe because it's a Java interface?
This is the same issue which caused the binary incompatibility in Cats 2.7  🙈

armanbilge:
Yes, that's right, and in fact Cats Effect also was not checking binary compatibility on Scala 3 until recently 😅

In any case, I think we are focused on the wrong problem here. AFAICT these changes are also source-breaking, regardless of whether they are bincompatible or not, which is also not ideal. There are ways we can fix/workaround this.

However, it's still not obvious to me what exactly the goal is with these changes? :)

rossabaker:
> However, it's still not obvious to me what exactly the goal is with these changes? :)

Many Java APIs are expressed in terms of an `Executor`.  It's nice if it can be done without breakage.  If not, there's always the adapter pattern.  [blows dust off Gang of Four book]

wjoel:
Yep, pretty much exactly what Ross said.

I've been experimenting with vert.x pgclient, and with this change it's possible to have it use the same threads as cats-effect does, instead of creating "nCpus" new ones. Similarly, it would be possible to use cats-effect threads to for Netty in http4s-netty.

Not sure if either is really a good idea, but it's interesting to have the option, and it might be a good idea for at least some use cases.

armanbilge:
Right, I see. Just to clarify, do you want to use the CE EC for vert.x/Netty, or use their `Executor` for CE? Do you have an example you can point to?

The tricky thing about this change is even if we can make it source- and binary-compatible, what about all the people currently building `IORuntime`s with `ExecutonContext`s that are not `Executor`s? No matter what changes we make here, we can't change that :) at least not without an adapter pattern.

What I could see working, is exposing a way to create a default `ExecutionContextExecutor` (i.e. the WSTP). Then, you can manually create an `IORuntime` and vert.x/netty whatever that both share that.

armanbilge:
Maybe we can do `isInstanceOf` to check if the EC is already an `Executor`, and otherwise fallback to Ross's adapter idea. That would always work, and wouldn't wrap for the WSTP.

In fact, we should also consider making the change in `Async`.
https://github.com/typelevel/cats-effect/blob/0a6bae21333e96c227eb031e6db777aa1c8d0197/kernel/shared/src/main/scala/cats/effect/kernel/Async.scala#L140

wjoel:
> Right, I see. Just to clarify, do you want to use the CE EC for vert.x/Netty, or use their `Executor` for CE? Do you have an example you can point to?

The first, use CE EC for vert.x/Netty. I don't have an example anymore, I undid it some weeks ago. :/
 
> The tricky thing about this change is even if we can make it source- and binary-compatible, what about all the people currently building `IORuntime`s with `ExecutonContext`s that are not `Executor`s? No matter what changes we make here, we can't change that :) at least not without an adapter pattern.

Oh, right, that's true. I see. That's a bit of a showstopper. I like your `def executionContext` idea for `Async`, that would also work for what I have in mind.

> What I could see working, is exposing a way to create a default `ExecutionContextExecutor` (i.e. the WSTP). Then, you can manually create an `IORuntime` and vert.x/netty whatever that both share that.

Yes, I think changing just `WorkStealingThreadPool` and making it public (so I could construct it manually and then create an `IORuntime` using it) would be enough for the use cases I had in mind. In retrospect, I think I started adding `with Executor` to the CE runtime mostly to get my hands on `WSTP with Executor` from the IOApp's `IORuntime` (since WSTP isn't public).

armanbilge:
> Yes, I think changing just WorkStealingThreadPool and making it public

Actually, I think it _is_ public now (or will be as of 3.4.0 :)

https://github.com/typelevel/cats-effect/blob/0a6bae21333e96c227eb031e6db777aa1c8d0197/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala#L35-L39

armanbilge:
> I like your `def executionContext` idea for `Async`, that would also work for what I have in mind.

Just came across a need for this myself in https://github.com/http4s/http4s-jdk-http-client/pull/641, so I opened https://github.com/typelevel/cats-effect/pull/3001. I'm hopeful that should be sufficient, and we won't need the more complex sort of changes this PR is proposing 😅

wjoel:
I like your PR in https://github.com/typelevel/cats-effect/pull/3001 better as well. Closing this one. :)

PR #2940, opened 2022-04-04
Added orElse combinator to IO

walesho:
Added orElse combinator to IO. 

Example of Usage 
`val firstIO = IO.raiseError[Int]`
 `val secondIO = IO.pure(42)`
 `val program = firstIO orElse secondIO // this would result to secondIO `

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -457,7 +457,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
    *   IO to be executed (if the current IO fails)
    * @return
    */
-  def orElse[B >: A](other: IO[B]): IO[B] =
+  def orElse[B >: A](other: => IO[B]): IO[B] =
     handleErrorWith(_ => other)
 
   /**
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -452,13 +452,13 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
     handleErrorWith[B](t => IO.pure(f(t)))
 
   /**
-   * Runs the current IO, if it fails with an error(exception), the other IO  will be executed.
+   * Runs the current IO, if it fails with an error(exception), the other IO will be executed.
    * @param other
-   * IO to be executed (if the current IO fails)  
+   *   IO to be executed (if the current IO fails)
    * @return
    */
   def orElse[B >: A](other: IO[B]): IO[B] =
-    SemigroupK[IO].combineK(this, other)
+    handleErrorWith(_ => other)
 
   /**
    * Handle any error, potentially recovering from it, by mapping it to another `IO` value.
@@ -1493,7 +1493,7 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
   protected class IOSemigroupK extends SemigroupK[IO] {
     final override def combineK[A](a: IO[A], b: IO[A]): IO[A] =
-      a.handleErrorWith(_ => b)
+      a orElse b
   }
 
   implicit def alignForIO: Align[IO] = _alignForIO
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -458,7 +458,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
    * @return
    */
   def orElse[B >: A](other: IO[B]): IO[B] =
-    handleErrorWith(_ => other)
+    SemigroupK[IO].combineK(this, other)
 
   /**
    * Handle any error, potentially recovering from it, by mapping it to another `IO` value.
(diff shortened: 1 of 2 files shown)


rossabaker:
This is what the `MonoidK` instance does for `combineK`.  I think it's a useful enough operation to give it a name like this.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -451,6 +451,15 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
...
     handleErrorWith[B](t => IO.pure(f(t)))
 
+  /**
+   * Runs the current IO, if it fails with an error(exception), the other IO  will be executed.
+   * @param other
+   * IO to be executed (if the current IO fails)  
+   * @return
+   */
+  def orElse[B >: A](other: IO[B]): IO[B] =
+    handleErrorWith(_ => other)
```
satorg: If it just calls to `handleErrorWith` then (perhaps) it would be better to add such a method directly to Cats' `ApplicativeError` type class instead of just `IO`. Because apparently it would make `orElse` functionality uniformly available for all `ApplicativeError` implementers (which includes IO from Cats Effect 2.x, Monix Task, fs2.Stream, etc). Wdyt?

djspiewak (commented):
This is great! Can you please run `sbt prePR` though? Additionally, as @rossabaker pointed out, these are the same semantics as `MonoidK[IO]`, so it might be worth adjusting that instance to delegate to `orElse` to avoid duplication and improve test coverage.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -458,7 +458,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
    * @return
    */
   def orElse[B >: A](other: IO[B]): IO[B] =
-    handleErrorWith(_ => other)
+    SemigroupK[IO].combineK(this, other)
```
rossabaker: I would go the other way around: leave this like you had it, and implement SemigroupK in terms of it.

walesho: Thanks for the input. I would follow that approach.

core/shared/src/main/scala/cats/effect/IO.scala
rossabaker: It's already [syntax on `ApplicativeError`](https://github.com/typelevel/cats/blob/e0784fc0d95b9c4eed0bb1ddce6c69e3fd9a4a4f/core/src/main/scala/cats/syntax/applicativeError.scala#L76-L77).  I'm not sure why it shouldn't also be on the type class.

Regardless, I think the trend on `IO` has been toward a rich interface without typeclass syntax.  Adding it here is consistent with `Either`, `EitherT`, etc.

satorg: > I'm not sure why it shouldn't also be on the type class.

I'm not sure either. I wonder if there is any convention on that in general. The syntax was introduced in typelevel/cats#2243 but seems the option for adding it to the typeclass had not been discussed nor challenged.

In general, having a method on typeclass makes it possible to override it in a particular implementation (e.g. for the sake of optimization). Not sure if it may make sense to do that for this `orElse` though.

satorg: > It's already [syntax on ApplicativeError](https://github.com/typelevel/cats/blob/e0784fc0d95b9c4eed0bb1ddce6c69e3fd9a4a4f/core/src/main/scala/cats/syntax/applicativeError.scala#L76-L77).

Btw, the syntax for `ApplicativeError` is defined in this way:
```scala
  def orElse(other: => F[A])(implicit F: ApplicativeError[F, E]): F[A] =
    F.handleErrorWith(fa)(_ => other)
```
Notice it takes `other` as a by-name param while the suggested addition to `IO` takes it as just a value.

Wouldn't it be better to keep these two methods in correspondence to each other? I.e. to declare the latter's `other` as a by-name parameter too? @rossabaker @djspiewak , wdyt?

armanbilge: Right, good catch. This should definitely be a by-name.

armanbilge: Btw, it's really OOPy but if we had inheritable syntax traits for the monad classes themselves, couldn't we get all this syntax for free and avoid oopsies? Similar to how I've suggested that the `IO` companion object itself should implement `Async[IO]`.

djspiewak: So the by-name parameter point is a really important one. If this already exists as syntax on `ApplicativeError`, then we need to be consistent with that.

djspiewak (changes requested):

djspiewak (approved):
Looks good to me!

armanbilge:
CI failure is because series/3.x is behind series/3.3.x so MiMa is rightfully upset :) merging https://github.com/typelevel/cats-effect/pull/2947 should fix this, I'll look into that now.

rossabaker (approved):

armanbilge:
Closing / re-opening to trigger CI on a fresh merge commit, apologies.

armanbilge:
Congrats on your first PR! @walesho 🎉

PR #2941, opened 2022-04-04
Add `GenTemporal#cachedRealTime`

brendanmaguire:
This method provides a nested effect which calculates time using a cached offset and `monotonic`. This is much faster than calling `realTime` repeatedly and is useful for some applications where precise time is not required

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
@@ -143,8 +143,8 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
    * and, when the inner effect is run, the offset is used in combination with
    * `Clock[F]#monotonic` to give an approximation of the real time. The practical benefit of
    * this is a reduction in the number of syscalls, since `realTime` will only be sequenced once
-   * per `ttl` window, and it tends to be (on most platforms) multiple orders of
-   * magnitude slower than `monotonic`.
+   * per `ttl` window, and it tends to be (on most platforms) multiple orders of magnitude
+   * slower than `monotonic`.
    *
    * This should generally be used in situations where precise "to the millisecond" alignment to
    * the system real clock is not needed. In particular, if the system clock is updated (e.g.
@@ -159,10 +159,9 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
   def cachedRealTime(ttl: Duration): F[F[FiniteDuration]] = {
     implicit val self = this
 
-    val cacheValuesF = for {
-      realTimeNow <- realTime
-      cacheRefreshTime <- monotonic
-    } yield (cacheRefreshTime, realTimeNow - cacheRefreshTime)
+    val cacheValuesF = (realTime, monotonic) mapN {
+      case (realTimeNow, cacheRefreshTime) => (cacheRefreshTime, realTimeNow - cacheRefreshTime)
+    }
 
     // Take two measurements and keep the one with the minimum offset. This will no longer be
     // required when `IO.unyielding` is merged (see #2633)
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
@@ -143,7 +143,10 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
    * and, when the inner effect is run, the offset is used in combination with
    * `Clock[F]#monotonic` to give an approximation of the real time.
    *
-   * This should only be used in situations where precise time does not need to be guaranteed.
+   * This should generally be used in situations where precise "to the millisecond" alignment to the
+   * system real clock is not needed. In particular, if the system clock is updated (e.g. via an NTP sync),
+   * the inner effect will not observe that update until up to `refreshPeriod`. This is an acceptable
+   * tradeoff in most practical scenarios, particularly with frequent sequencing of the inner effect.
    *
    * @param refreshPeriod
    *   The period of time after which the cached real time will be refreshed. Note that it will
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
@@ -141,45 +141,46 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
    * Returns a nested effect which returns the time in a much faster way than
    * `Clock[F]#realTime`. This is achieved by caching the real time when the outer effect is run
    * and, when the inner effect is run, the offset is used in combination with
-   * `Clock[F]#monotonic` to give an approximation of the real time. The practical benefit of this is a reduction
-   * in the number of syscalls, since `realTime` will only be sequenced once per `refreshTime` window, and it
-   * tends to be (on most platforms) multiple orders of magnitude slower than `monotonic`.
+   * `Clock[F]#monotonic` to give an approximation of the real time. The practical benefit of
+   * this is a reduction in the number of syscalls, since `realTime` will only be sequenced once
+   * per `refreshTime` window, and it tends to be (on most platforms) multiple orders of
+   * magnitude slower than `monotonic`.
    *
-   * This should generally be used in situations where precise "to the millisecond" alignment to the
-   * system real clock is not needed. In particular, if the system clock is updated (e.g. via an NTP sync),
-   * the inner effect will not observe that update until up to `refreshPeriod`. This is an acceptable
-   * tradeoff in most practical scenarios, particularly with frequent sequencing of the inner effect.
+   * This should generally be used in situations where precise "to the millisecond" alignment to
+   * the system real clock is not needed. In particular, if the system clock is updated (e.g.
+   * via an NTP sync), the inner effect will not observe that update until up to `ttl`. This is
+   * an acceptable tradeoff in most practical scenarios, particularly with frequent sequencing
+   * of the inner effect.
    *
-   * @param refreshPeriod
+   * @param ttl
    *   The period of time after which the cached real time will be refreshed. Note that it will
    *   only be refreshed upon execution of the nested effect
    * @see
    *   [[timeout]] for a variant which respects backpressure and does not leak fibers
    */
-  def cachedRealTime(refreshPeriod: Duration): F[F[FiniteDuration]] = {
-    val cacheValuesF = flatMap(realTime)(realTimeNow =>
-      map(monotonic)(cacheRefreshTime => (cacheRefreshTime, realTimeNow - cacheRefreshTime)))
+  def cachedRealTime(ttl: Duration): F[F[FiniteDuration]] = {
+    implicit val self = this
+
+    val cacheValuesF = for {
+      realTimeNow <- realTime
+      cacheRefreshTime <- monotonic
+    } yield (cacheRefreshTime, realTimeNow - cacheRefreshTime)
 
     // Take two measurements and keep the one with the minimum offset. This will no longer be
     // required when `IO.unyielding` is merged (see #2633)
-    val minCacheValuesF = flatMap(cacheValuesF) {
-      case cacheValues1 @ (_, offset1) =>
-        map(cacheValuesF) {
-          case cacheValues2 @ (_, offset2) if offset2 < offset1 => cacheValues2
-          case _ => cacheValues1
-        }
+    val minCacheValuesF = (cacheValuesF, cacheValuesF) mapN {
+      case (cacheValues1 @ (_, offset1), cacheValues2 @ (_, offset2)) =>
+        if (offset1 < offset2) cacheValues1 else cacheValues2
     }
 
-    val cacheValuesRefF = flatMap(minCacheValuesF)(ref)
-
-    map(cacheValuesRefF) { cacheValuesRef =>
-      flatMap(monotonic) { timeNow =>
-        flatMap(cacheValuesRef.access) {
-          case ((cacheRefreshTime, offset), setCacheValues) =>
-            if (timeNow >= cacheRefreshTime + refreshPeriod)
-              flatMap(minCacheValuesF) {
+    minCacheValuesF.flatMap(ref).map { cacheValuesRef =>
+      monotonic.flatMap { timeNow =>
+        cacheValuesRef.get.flatMap {
+          case (cacheRefreshTime, offset) =>
+            if (timeNow >= cacheRefreshTime + ttl)
+              minCacheValuesF.flatMap {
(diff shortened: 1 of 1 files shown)


brendanmaguire:
As described in the discussion in #2416 , and in particular [this comment](https://github.com/typelevel/cats-effect/pull/2416#issuecomment-1073308423).

brendanmaguire:
@djspiewak & @armanbilge , would you mind taking a look. It seems messy to me but I'm not sure how to clean it up. Suggestions welcome.

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
```diff
@@ -136,6 +136,52 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
...
       }
     }
+
+  /**
+   * Returns a nested effect which returns the time in a much faster way than
+   * `Clock[F]#realTime`. This is achieved by caching the real time when the outer effect is run
+   * and, when the inner effect is run, the offset is used in combination with
+   * `Clock[F]#monotonic` to give an approximation of the real time.
+   *
+   * This should only be used in situations where precise time does not need to be guaranteed.
```
djspiewak:
```suggestion
   * This should generally be used in situations where precise "to the millisecond" alignment to the
   * system real clock is not needed. In particular, if the system clock is updated (e.g. via an NTP sync),
   * the inner effect will not observe that update until up to `refreshPeriod`. This is an acceptable
   * tradeoff in most practical scenarios, particularly with frequent sequencing of the inner effect.
```

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
```diff
@@ -136,6 +136,52 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
           start(f.cancel) *> raiseError[A](ev(new TimeoutException(duration.toString)))
       }
     }
+
+  /**
+   * Returns a nested effect which returns the time in a much faster way than
+   * `Clock[F]#realTime`. This is achieved by caching the real time when the outer effect is run
+   * and, when the inner effect is run, the offset is used in combination with
+   * `Clock[F]#monotonic` to give an approximation of the real time.
```
djspiewak:
```suggestion
   * `Clock[F]#monotonic` to give an approximation of the real time. The practical benefit of this is a reduction
   * in the number of syscalls, since `realTime` will only be sequenced once per `refreshTime` window, and it
   * tends to be (on most platforms) multiple orders of magnitude slower than `monotonic`.
```

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
```diff
@@ -136,6 +136,52 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
...
+      case cacheValues1 @ (_, offset1) =>
+        map(cacheValuesF) {
+          case cacheValues2 @ (_, offset2) if offset2 < offset1 => cacheValues2
+          case _ => cacheValues1
+        }
+    }
+
+    val cacheValuesRefF = flatMap(minCacheValuesF)(ref)
+
+    map(cacheValuesRefF) { cacheValuesRef =>
```
djspiewak: Random fyi: you can use cats syntax within this class. :-) There should be a `private implicit val self = this` or something like that within `GenTemporal`, meaning you can safely do things like `cacheValuesRefF.map { ...}`.

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
```diff
@@ -136,6 +136,52 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
...
+
+    // Take two measurements and keep the one with the minimum offset. This will no longer be
+    // required when `IO.unyielding` is merged (see #2633)
+    val minCacheValuesF = flatMap(cacheValuesF) {
+      case cacheValues1 @ (_, offset1) =>
+        map(cacheValuesF) {
+          case cacheValues2 @ (_, offset2) if offset2 < offset1 => cacheValues2
+          case _ => cacheValues1
+        }
+    }
```
djspiewak:
```suggestion
    val minCacheValuesF = (cacheValuesF, cacheValuesF) mapN {
      case ((_, offset1), (_, offset2)) => offset1.min(offset2)
    }
```

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
```diff
@@ -136,6 +136,52 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
...
+
+    val cacheValuesRefF = flatMap(minCacheValuesF)(ref)
+
+    map(cacheValuesRefF) { cacheValuesRef =>
+      flatMap(monotonic) { timeNow =>
+        flatMap(cacheValuesRef.access) {
```
djspiewak: I would probably use `Ref#set` and `Ref#get` directly instead.

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
```diff
@@ -136,6 +136,52 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
...
+   *
+   * This should only be used in situations where precise time does not need to be guaranteed.
+   *
+   * @param refreshPeriod
+   *   The period of time after which the cached real time will be refreshed. Note that it will
+   *   only be refreshed upon execution of the nested effect
+   * @see
+   *   [[timeout]] for a variant which respects backpressure and does not leak fibers
+   */
+  def cachedRealTime(refreshPeriod: Duration): F[F[FiniteDuration]] = {
```
djspiewak: What do you think about `ttl` as a name for this parameter?

djspiewak (commented):
Doesn't seem too messy to me. :-) It's a surprisingly involved effect. The big thing that would clean it up is just leveraging the syntax.

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
brendanmaguire: Sounds good to me :+1: I'll change it.

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
brendanmaguire: Ah! I didn't know this. I will use the more idiomatic approach.

I didn't notice `for` comprehensions in the codebase (apart from tests). Is there a reason for this?

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
brendanmaguire: I get the following compilation error when attempting to use `mapN`:

```
Could not find an instance of Functor for F
```

Am I missing an import? :thinking:

brendanmaguire: Ah, I see this is what you were referring to with the `private implicit val self = this`. Thanks for the tip!

brendanmaguire: I've changed this locally to use `mapN` but your suggestion won't quite work as we need to retain both the `cacheRefreshTime` (to determine when to refresh) and `offset` (to determine the time). I'll push what I have and you can see if it looks right.

brendanmaguire:
Thanks for the feedback @djspiewak :slightly_smiling_face: Changes pushed.

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
```diff
@@ -136,6 +136,58 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
...
+   * the system real clock is not needed. In particular, if the system clock is updated (e.g.
+   * via an NTP sync), the inner effect will not observe that update until up to `ttl`. This is
+   * an acceptable tradeoff in most practical scenarios, particularly with frequent sequencing
+   * of the inner effect.
+   *
+   * @param ttl
+   *   The period of time after which the cached real time will be refreshed. Note that it will
+   *   only be refreshed upon execution of the nested effect
+   * @see
+   *   [[timeout]] for a variant which respects backpressure and does not leak fibers
```
brendanmaguire:
```suggestion
```

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
djspiewak: `for` comprehensions are somewhat tricky for a few reasons. First, the desugaring is unintuitive and can create issues, both semantic and performance-related, if you aren't careful (particularly in recursive cases). Second, the desugaring is actually *different* between Scala 2 and Scala 3, so generally we try not to use them unless it's really convincingly safe to do so.

Stylistically though, I quite like them :-)

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
```diff
@@ -136,6 +136,56 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
...
       }
     }
+
+  /**
+   * Returns a nested effect which returns the time in a much faster way than
+   * `Clock[F]#realTime`. This is achieved by caching the real time when the outer effect is run
+   * and, when the inner effect is run, the offset is used in combination with
+   * `Clock[F]#monotonic` to give an approximation of the real time. The practical benefit of
+   * this is a reduction in the number of syscalls, since `realTime` will only be sequenced once
+   * per `refreshTime` window, and it tends to be (on most platforms) multiple orders of
```
djspiewak:
```suggestion
   * per `ttl` window, and it tends to be (on most platforms) multiple orders of
```

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
```diff
@@ -136,6 +136,56 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
...
+      realTimeNow <- realTime
+      cacheRefreshTime <- monotonic
+    } yield (cacheRefreshTime, realTimeNow - cacheRefreshTime)
+
+    // Take two measurements and keep the one with the minimum offset. This will no longer be
+    // required when `IO.unyielding` is merged (see #2633)
+    val minCacheValuesF = (cacheValuesF, cacheValuesF) mapN {
+      case (cacheValues1 @ (_, offset1), cacheValues2 @ (_, offset2)) =>
+        if (offset1 < offset2) cacheValues1 else cacheValues2
+    }
```
djspiewak, line 171: Much cleaner, IMO!

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
```diff
@@ -136,6 +136,56 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
...
+   *   The period of time after which the cached real time will be refreshed. Note that it will
+   *   only be refreshed upon execution of the nested effect
+   */
+  def cachedRealTime(ttl: Duration): F[F[FiniteDuration]] = {
+    implicit val self = this
+
+    val cacheValuesF = for {
+      realTimeNow <- realTime
+      cacheRefreshTime <- monotonic
+    } yield (cacheRefreshTime, realTimeNow - cacheRefreshTime)
```
djspiewak: Btw this can also be done with `mapN`, and it's probably cleaner to do so.

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
```diff
@@ -136,6 +136,56 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
...
+            if (timeNow >= cacheRefreshTime + ttl)
+              minCacheValuesF.flatMap {
+                case cacheValues @ (cacheRefreshTime, offset) =>
+                  cacheValuesRef.set(cacheValues).map(_ => cacheRefreshTime + offset)
+              }
+            else
+              pure(timeNow + offset)
+        }
+      }
+    }
```
djspiewak, line 186: IMO if we're going to use `for` here, this is the place to do it. :-)

djspiewak (changes requested):
Really minor required change and some stylistic nits, but this is looking very good!

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
brendanmaguire: Ah yes, this makes sense. I will remove them.

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
brendanmaguire: I'm happy to remove the use of `for` given the reasons you've outlined above.

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
brendanmaguire: Good point :slightly_smiling_face: Will change.

brendanmaguire:
Thanks again for the feedback @djspiewak . I will start using `mapN` more often in my own code after this :slightly_smiling_face: Changes made.

djspiewak:
Reopening to get a new merge commit for CI

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
```diff
@@ -136,6 +136,55 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
...
+        cacheValuesRef.get.flatMap {
+          case (cacheRefreshTime, offset) =>
+            if (timeNow >= cacheRefreshTime + ttl)
+              minCacheValuesF.flatMap {
+                case cacheValues @ (cacheRefreshTime, offset) =>
+                  cacheValuesRef.set(cacheValues).map(_ => cacheRefreshTime + offset)
+              }
+            else
+              pure(timeNow + offset)
+        }
```
He-Pin, line 184: can this be rewrite to for expression?

djspiewak: It could be but we tend to avoid them in most places since their desugaring is both unintuitive (creating an extra `map` in most cases) *and* not the same across Scala 2 and 3 (due to our use of a compiler plugin on the former). Perhaps more relevantly though, I don't think a `for`-comprehension would help much here with readability, since you would still have the testing due to the `F[F[FiniteDuration]]` result type.

djspiewak (approved):

PR #2951, opened 2022-04-07
Implemented configurability for `Supervisor` scope termination

### LINKED ISSUE #1881: Consider customizable Supervisor/Dispatcher
Users have expressed interest in being able to change the behavior of these abstractions when they're being shut down (i.e. waiting for tasks to complete before exiting).

It should be investigated how we can best achieve that.

djspiewak:
This implements an `await` flag on the `Supervisor` constructor. This will in turn be added to `Dispatcher`'s parameters as a follow up to #2854. Specifically, this changes the scope termination semantics from `_.parTraverse(_.cancel)` to `_.traverse(_.join).void`, which should resolve some of the most common cases that lead to `dispatcher already shutdown`.

Note that I'm cheating a bit with binary compatibility since this removes the static forwarder (only used from Java) for `Supervisor`'s constructor.

Fixes #1881

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Supervisor.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Supervisor.scala
@@ -107,28 +107,33 @@ object Supervisor {
    * Creates a [[cats.effect.kernel.Resource]] scope within which fibers can be monitored. When
    * this scope exits, all supervised fibers will be finalized.
    */
-  def apply[F[_]](implicit F: Concurrent[F]): Resource[F, Supervisor[F]] = {
+  def apply[F[_]](await: Boolean = false)(
+      implicit F: Concurrent[F]): Resource[F, Supervisor[F]] = {
     F match {
-      case asyncF: Async[F] => applyForAsync(asyncF)
-      case _ => applyForConcurrent
+      case asyncF: Async[F] => applyForAsync(await)(asyncF)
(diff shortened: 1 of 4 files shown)


std/shared/src/main/scala/cats/effect/std/Supervisor.scala
```diff
@@ -107,74 +107,92 @@ object Supervisor {
    * Creates a [[cats.effect.kernel.Resource]] scope within which fibers can be monitored. When
    * this scope exits, all supervised fibers will be finalized.
    */
-  def apply[F[_]](implicit F: Concurrent[F]): Resource[F, Supervisor[F]] = {
+  def apply[F[_]](await: Boolean = false)(
```
armanbilge: Is this a source-breaking change? Since the `()` will be required I think.

djspiewak: It is

armanbilge (approved):
This was quite straightforward actually :)

wjoel:
This is great, but it only fixes the Supervisor part of https://github.com/typelevel/cats-effect/issues/1881. @iRevive has expressed interest in addressing the Dispatcher part in https://github.com/typelevel/cats-effect/issues/1881#issuecomment-1088363444, so maybe that issue should be kept open until those changes arrive?

djspiewak:
Good point!

PR #2954, opened 2022-04-09
Duration

### LINKED ISSUE #2855: Generalize `timeout` and friends to take a `Duration`
Right now, every flavor of this method takes a `FiniteDuration`, which is *correct* but less convenient. We should expand them to `Duration` and silently degrade to identity when it's `Duration.Inf`. This can be done trivially on `IO` because of contravariant expansion, but it might requires some bincompat shenanigans on `Temporal`.

b3nk3i:
Generalize timeout and friends to take a Duration #2855


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -664,7 +664,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
    *   is the time span for which we wait for the source to complete; in the event that the
    *   specified time has passed without the source completing, a `TimeoutException` is raised
    */
-  def timeout[A2 >: A](duration: FiniteDuration): IO[A2] =
+  def timeout[A2 >: A](duration: Duration): IO[A2] =
```
armanbilge, line 670: Thanks for the PR! At a glance your changes are looking good 👍 

It's a little bit annoying, but for binary compatibility, we actually need to keep the old methods in addition to adding the new ones.

b3nk3i: thanks for the feedback :)
I can put back the old method

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
```diff
@@ -97,11 +99,19 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
...
+  def timeout[A](fa: F[A], duration: Duration)(implicit ev: TimeoutException <:< E): F[A] = {
+    handleFinite(fa, duration) { finiteDuration =>
+      flatMap(race(fa, sleep(finiteDuration))) {
+        case Left(a) => pure(a)
+        case Right(_) => raiseError[A](ev(new TimeoutException(duration.toString())))
+      }
+    }
+  }
+
+  def handleFinite[A](fa: F[A], duration: Duration)(f: FiniteDuration => F[A]): F[A] = {
```
djspiewak: This should probably be private. 🙂

core/shared/src/main/scala/cats/effect/IO.scala
djspiewak: More specifically, I believe we need a protected overload which handles the FiniteDuration case, and the public overload then becomes the Duration. This would preserve binary compatibility and semantics with older implementers of the trait.

djspiewak (changes requested):
This is looking like a very nice start! Thank you for taking this on.

core/shared/src/main/scala/cats/effect/IO.scala
walesho: What would be the practical implication of taking a Duration? Since we can already do IO.timeout(5 Seconds) using FiniteDuration as it were.

satorg: @walesho I'd guess it is just for the sake of generalization. If `Duration.Inf` is passed to the `timeout` method, then the operation should never timeout. It may not be that useful with literals and constant values, but imagine if the timeout value is taken from configs. So there will be an easy way to disable timeouts from the configs simply by passing an infinite value out there.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -664,7 +665,10 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
...
+    handleFinite(this, duration)(finiteDuration => timeout(finiteDuration))
+
+  protected def timeout[A2 >: A](duration: FiniteDuration): IO[A2] =
```
djspiewak: This can actually be `private[effect]` instead of `protected`, since `IO` is final.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -683,11 +687,16 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
    * @param fallback
    *   is the task evaluated after the duration has passed and the source canceled
    */
-  def timeoutTo[A2 >: A](duration: FiniteDuration, fallback: IO[A2]): IO[A2] =
+  def timeoutTo[A2 >: A](duration: Duration, fallback: IO[A2]): IO[A2] = {
+    handleFinite[IO, A2](this, duration)(finiteDuration => timeoutTo(finiteDuration, fallback))
+  }
+
+  protected def timeoutTo[A2 >: A](duration: FiniteDuration, fallback: IO[A2]): IO[A2] = {
```
djspiewak: Same here…

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
```diff
@@ -214,6 +228,13 @@ object GenTemporal {
         instantiateGenTemporalForWriterT(temporal)
     }
 
+  def handleFinite[F[_], A](fa: F[A], duration: Duration)(f: FiniteDuration => F[A]): F[A] = {
```
djspiewak: This needs to be `private[kernel]` instead of public, since this is unintentionally adding to the public API surface!

djspiewak (changes requested):
Almost there! Just a quick tweak to avoid bloating the public API. I'll work on fixing the scaladoc links and will push the changes, since that's the most annoying thing in the universe.

core/shared/src/main/scala/cats/effect/IO.scala
armanbilge: Note that the type parameter here is unnecessary since `IO` is covariant but I guess we'd rather not break source-compatibility at this stage /shrug

kernel/shared/src/main/scala-2.12/cats/effect/kernel/syntax/GenTemporalSyntaxCompat.scala
```diff
@@ -0,0 +1,54 @@
...
+
+  @deprecated("Preserved for binary-compatibility", "3.4.0")
+  def timeout$extension[F[_], A, E](
+      wrapped: F[A],
+      duration: FiniteDuration,
+      F: GenTemporal[F, E],
+      timeoutToE: TimeoutException <:< E): F[A] = F.timeout(wrapped, duration)(timeoutToE)
+
+  @deprecated("Preserved for binary-compatibility", "3.4.0")
+  def timeoutAndForget$extension[F[_], A, E](
```
armanbilge, line 47: Yes, they have to be public. Yes, it's really annoying.

djspiewak: Why can't they be `private[kernel]`?

armanbilge: Because static forwarders and binary compatibility. This is only on 2.12 btw.

djspiewak: Grrrrrrrrrrrrr.

Okay :-P

djspiewak (approved):

PR #2959, opened 2022-04-13
Add configurable `Dispatcher`

iRevive:
Follow up to https://github.com/typelevel/cats-effect/issues/1881#issuecomment-1096987607. 

The PR is based on https://github.com/typelevel/cats-effect/pull/2854


tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
```diff
@@ -120,13 +142,139 @@ class DispatcherSpec extends BaseSpec {
...
+        } yield runner
+
+      for {
+        releaseInner <- CountDownLatch[IO](1)
+        fiberLatch <- CountDownLatch[IO](1)
+
+        fiber <- makeRunner(releaseInner).use { runner =>
+          for {
+            cdl <- CountDownLatch[IO](1)
+            _ <- IO(runner.unsafeRunAndForget(cdl.release >> fiberLatch.await))
```
iRevive, line 265: Without explicit delay (via `cdl.release` and `cdl.await`) the test does not pass on the sequential dispatcher.
As far as I debugged, the dispatcher is closed faster than the task is picked up by the queue. 

A test to reproduce:
```scala
"effect is not executed" in real {
  for {
    fiberLatch <- CountDownLatch[IO](1)
    fiber <- Dispatcher
      .sequential[IO](await = true)
      .use(runner => IO(runner.unsafeToFuture(fiberLatch.await)))
      .start
    _ <- fiberLatch.release
    future <- fiber.joinWithNever
  } yield {
    scala.concurrent.Await.result(future, Duration.Inf) // never completes
    ok
  }
}
```

The test fails on the original branch (https://github.com/typelevel/cats-effect/pull/2854) as well.

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -111,22 +192,32 @@ object Dispatcher {
     final case class CancelToken(cancelToken: () => Future[Unit]) extends CancelState
 
     for {
-      supervisor <- Supervisor[F]
+      (workers: Int, fork: (F[Unit] => F[F[Unit]])) <- mode match {
```
iRevive: Unfortunately, JS requires a little help with types there.

tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
iRevive: cc @djspiewak

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -110,23 +191,37 @@ object Dispatcher {
...
+    val (workers, makeFork) =
+      mode match {
+        case Mode.Parallel =>
+          (Cpus, Supervisor[F](await).map(s => s.supervise(_: F[Unit]).map(_.cancel)))
+
+        case Mode.Sequential =>
+          (
+            1,
+            Resource
+              .pure[F, F[Unit] => F[F[Unit]]]((_: F[Unit]).as(F.unit).handleError(_ => F.unit))
```
iRevive, line 203: Is `as(F.unit)` necessary, since we are already operating on `F[Unit]`?

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
armanbilge: Why is this JS specific? This code is in a shared file right? Actually the JVM/JS compilers are essentially the same, only the backend is different, which is why I'm surprised and curious :)

iRevive: CI was upset with these errors: https://github.com/typelevel/cats-effect/runs/6006857516?check_suite_focus=true#step:17:62

```
[error] -- [E007] Type Mismatch Error: /home/runner/work/cats-effect/cats-effect/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala:207:61 
[error] 207 |        val latches = new Array[AtomicReference[() => Unit]](workers)
[error]     |                                                             ^^^^^^^
[error]     |                                             Found:    (workers : Any)
[error]     |                                             Required: Int
[error] -- [E134] Type Error: /home/runner/work/cats-effect/cats-effect/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala:209:17 
[error] 209 |        while (i < workers) {
[error]     |               ^^^
[error]     |None of the overloaded alternatives of method < in class Int with types
[error]     | (x: Double): Boolean
[error]     | (x: Float): Boolean
[error]     | (x: Long): Boolean
[error]     | (x: Int): Boolean
[error]     | (x: Char): Boolean
[error]     | (x: Short): Boolean
[error]     | (x: Byte): Boolean
[error]     |match arguments ((workers : Any))
```

I can assume, something is going wrong with unapply in `withFilter`.

armanbilge: Hmm, I see compile errors in the JVM job too.
https://github.com/typelevel/cats-effect/runs/6006857453?check_suite_focus=true#step:17:52

iRevive: Ah, it's less suspicious now.

djspiewak:
It looks like this has been rebased. Please don't do that. :-) Since this is building on top of another PR, it needs to be based *on that branch*, merged with `series/3.x`.

In general, we don't rebase commits on non-draft PRs (outside of weird circumstances) for exactly this reason.

tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
djspiewak: Ahhhhhh! Interesting. Whoops.

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -96,12 +96,93 @@ object Dispatcher {
...
+   *     }
+   *   }}}
+   *
+   * @param await
+   *   the termination policy of the internal [[Supervisor]].
+   *   - true - wait for the completion of the active fibers
+   *   - false - cancel the active fibers
    */
-  def apply[F[_]](implicit F: Async[F]): Resource[F, Dispatcher[F]] = {
+  def parallel[F[_]: Async](await: Boolean): Resource[F, Dispatcher[F]] =
```
djspiewak, line 155: IMO we should mirror `Supervisor` here and create zero-parameter overloads that default `await` to `false`.

djspiewak (changes requested):
Changes look very good! I would add some overloads and undo the rebasing, but otherwise this is ready to go.

djspiewak (dismissed):
Looks fantastic! Thank you!

djspiewak:
I'm going to launder my changes through this PR, since the original one hasn't really been approved yet. :-P

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
iRevive: Good point. I added overloaded methods.

iRevive:
The documentation can be slightly improved:
```scala
/**
 * @define createDispatcher
 *   Create a [[Dispatcher]] that can be used within a resource scope. Once the resource scope
 *   exits, all active effects will be canceled, and attempts to submit new effects will throw
 *   an exception.
 *
 * @define parallelDetails
 *.  This corresponds to a pattern in which a single `Dispatcher` is being used by multiple ...
 *
 * @define sequentialDetails
 *.  This corresponds to a [[Dispatcher]] mode in which submitted actions are evaluated strictly ...
 */
object Dispatcher {

  /**
   * $createDispatcher
   * $parallelDetails
   */
  def parallel[F[_]: Async]: Resource[F, Dispatcher[F]] =
    parallel[F](await = false)
    
  /**
   * $createDispatcher
   * $sequentialDetails
   */
  def sequential[F[_]: Async]: Resource[F, Dispatcher[F]] =
    sequential[F](await = false)
    
  /**
   * $createDispatcher
   * $parallelDetails
   */
  def parallel[F[_]: Async](await: Boolean): Resource[F, Dispatcher[F]] =
    apply(Mode.Parallel, await)
    
   /**
   * $createDispatcher
   * $sequentialDetails
   */
  def sequential[F[_]: Async](await: Boolean): Resource[F, Dispatcher[F]] =
    apply(Mode.Sequential, await)
}
```

That way every overloaded method will have detailed documentation.

djspiewak:
> The documentation can be slightly improved

You sir are a master of understatement. :-D

I knew there was a way to define this type of expansion but can never remember it! We should definitely do that, though with an added sentence describing the default semantic on the no-args overloads.

armanbilge:
Woah, that's cool.

djspiewak:
Interesting! `DispatcherSpec` failed

iRevive:
Here are examples of the [macro definitions](https://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html#macros):

<details>
  <summary>Generated scaladoc</summary>
  image
</details>

<details>
  <summary>IntelliJ: method popup</summary>
image
</details>

<details>
  <summary>IntelliJ: source code (published jar) with <a href="https://www.jetbrains.com/help/idea/working-with-code-documentation.html#toggle-rendered-view" target="_blank">rendered view</a></summary>
  
image

</details>

<details>
  <summary>IntelliJ: source code (published jar)</summary>
image
</details>

I have mixing feelings tbh. I personally prefer raw scaladoc as a source of documentation during development. And with macro definitions, raw scaladoc is kinda meaningless.

Also, I'm not sure how Metals handle it since I'm not using it.

iRevive:
Weird. The tests are passing locally on `GraalVM 11` and `Azul VM 15.0`.

Could you restart CI jobs, please? Seems I do not have enough permissions for that.

djspiewak (approved):
Thank you for taking this on!

PR #2963, opened 2022-04-15
Override racePair in _asyncForIO

### LINKED ISSUE #2962: using timeout prints stack trace to console if effect throws exception
```scala
IO.delay(throw new Exception("moi 1")).timeout(10 minutes).unsafeRunSync()
```
does seem to work as expected, but does print exception to console 
```shell
java.lang.Exception: moi 1
  | => rat test.TimeoutSpec.$anonfun$new$4(TimeoutSpec.scala:20)
	at delay @ test.TimeoutSpec.$anonfun$new$3(TimeoutSpec.scala:20)
```

Affected cats-effect version are 3.3.8+

example project for bug https://github.com/margussipria/cats-effect-timout-bug

durban:
... so that it uses the specialized `RacePair`. In turn, this causes `race`, `timeout`, etc. to use it too. Incidentally, this fixes #2962 (which is based on `timeout`). However, nothing stops anybody for doing the same with start/join, which will still report the error.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -1536,6 +1536,11 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
     def start[A](fa: IO[A]): IO[FiberIO[A]] =
       fa.start
 
+    override def racePair[A, B](
+        left: IO[A],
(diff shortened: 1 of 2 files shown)


djspiewak (approved):

djspiewak:
This is definitely a fascinating oversight! Thank you for catching it. I'm still a bit worried about the underlying bug, so I'll leave that open to investigate, but it's nice to fix this case.

PR #2965, opened 2022-04-17
Fixed `TestControl#nextInterval` and added `tickFor`

### LINKED ISSUE #2615: Expose a tickFor(time: FiniteDuration) in TestControl
In some cases the logic you are trying to test might not be adequate for manually ticking and advancing directly.

An simplified example:

```scala
def functionToTest[A](sleep: FiniteDuration, calc: => IO[A], emit: A => IO[Unit]): IO[Unit] = {
    calc.flatMap(emit) <* IO.sleep(sleep)
}
```
If I want to test how many times `emit` is called in a given period of time, I would need to understand exactly how and when fibers are scheduled (not only in the `functionToTest` but also if anything is scheduled in `calc` or `emit`.

I would propose to add some utilities to advance the clock and tick, e.g.:

```scala
  def advanceFor(duration: FiniteDuration): IO[Unit] = {
    Monad[IO].iterateWhileM(Option(0.millis)) { cur => 
      control.nextInterval.attempt.map(_.toOption)
        .flatMap {
        case Some(n) if n > 0.millis => (control.advance(n) <* control.tick).as(Some(cur.getOrElse(0.millis) + n))
        case Some(_) => control.tick.as(cur)
        case None => IO.pure(None)
        }
    } { cur => 
      cur.nonEmpty && cur.exists(_ < duration) 
    }.void
  }
```

```scala
//The test now becomes rather simple:

val expected = 60
Ref.of[IO,Vector[Int]](Vector.empty).flatMap { ref =>
    val test = functionToTest(1.second, IO.pure(1), (a: Int) => ref.modify(cur => (cur.appended(a, ()))).timeout(1.minute).attempt
    TestControl.execute(test).flatMap { control =>
      for {
          _ <- control.tick
          _ <- control.advanceFor(1.minute)
          _ <- ref.get.map(acc => assert(acc.forall(_ == 1) && acc.size == expected))
        } yield ()
    }
  }
```

Not sure if this makes sense, or there is a better way to achieve the same with the current API.

djspiewak:
This function implements the semantic that most people probably want when they reach for `advanceAndTick`.

Fixes #2615


tests/shared/src/test/scala/cats/effect/testkit/TestControlSpec.scala
```diff
@@ -112,6 +113,59 @@ class TestControlSpec extends BaseSpec {
...
     }
+
+    "produce Duration.Zero from nextInterval when no tasks" in real {
+      TestControl.execute(deadlock) flatMap { control =>
+        for {
+          _ <- control.tick
+          i <- control.nextInterval
+          _ <- IO(i mustEqual Duration.Zero)
+        } yield ok
+      }
```
armanbilge, line 124: I haven't followed any of this API at all, so dumb question: why `Duration.Zero` instead of `Duration.Infinite`? I see the doc says
> If fibers are currently eligible for execution, the result will be `Duration.Zero`.

But AFAICT there are no fibers eligible in this case.

djspiewak: You're right, this is probably a better semantic. I'll adjust it.

djspiewak: Actually correction: the bigger problem is that `nextInterval` produces a `FiniteDuration` and not a `Duration`, so we can't really make that change without playing the package private game. I'm not sure whether or not it's worth it in this case.

In a sense, `nextInterval`'s contract is meant to be "you can pass this result to `advance` and the most incremental possible result will happen". `advance` only takes a `FiniteDuration`, and that's (IMO) a good thing, which transitively implies that `Duration.Zero` is probably the right answer here.

The prescriptively correct thing to do would have been to produce an `Option[FiniteDuration]` (or isomorphically, just `Duration`) from `nextInterval`, but sicne this is already detectable via `results` (which has cleaner overall semantics), I'm not sure if there's value there.

armanbilge: Yeah, right, if we can't change the API the next best thing is the docs, since that bit I quoted may not be true in general.

armanbilge: Actually, I'm confused. How do you distinguish "more stuff to do" from deadlock? Isn't `results` `None` in both situations?

djspiewak: Oh right yes. I should have referred to `isDeadlocked`, which will produce `true` if you're stuck.

tests/shared/src/test/scala/cats/effect/testkit/TestControlSpec.scala
```diff
@@ -112,6 +113,59 @@ class TestControlSpec extends BaseSpec {
...
+        }
+      }
+
+      "advance incrementally in minimum steps" in real {
+        val step = IO.sleep(1.second) *> IO.realTime
+
+        TestControl.execute((step, step).tupled) flatMap { control =>
+          for {
+            _ <- control.tickFor(1.second + 500.millis)
+            _ <- control.tickAll
```
armanbilge, line 161: Another dumb question 😄 what is this test exercising? That `tickAll` doesn't tick too far forward? I can't see how this tests "advance incrementally in minimum steps".

kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/TestContext.scala
```diff
@@ -109,8 +109,12 @@ final class TestContext private (_seed: Long) extends ExecutionContext { self =>
    */
   def nextInterval(): FiniteDuration = {
     val current = stateRef
-    val diff = current.tasks.first().runsAt - current.currentNanos.get().nanos
-    diff.max(Duration.Zero)
+    if (current.tasks.isEmpty) {
```
armanbilge, line 112: I think scaladoc for this needs to be updated as well.

tests/shared/src/test/scala/cats/effect/testkit/TestControlSpec.scala
djspiewak: The Scaladoc explains a bit better. 🙂 It's the difference between this and `advanceAndTick`.

armanbilge: Ahh, this is literally the snippet in the scaladoc. I blame your PR review technique of starting with the tests which I've been trying :P

Ok, right. So really what this is testing is that it doesn't do weird unintuitive stuff like `advanceAndTick` does.

testkit/shared/src/main/scala/cats/effect/testkit/TestControl.scala
```diff
@@ -209,6 +214,58 @@ final class TestControl[A] private (
...
+   *
+   * @see
+   *   [[tick]]
+   * @see
+   *   [[advance]]
+   */
+  def tickFor(time: FiniteDuration): IO[Unit] =
+    tick *> nextInterval flatMap { next =>
+      val step = time.min(next)
+      val remaining = time - step
```
armanbilge, line 255: We can move this inside the `else`

testkit/shared/src/main/scala/cats/effect/testkit/TestControl.scala
```diff
@@ -209,6 +214,58 @@ final class TestControl[A] private (
...
+   *       _ <- IO(assert(r == Some(Outcome.succeeded(1.second, 2.seconds))))
+   *     } yield ()
+   *   }
+   * }}}
+   *
+   * Notably, the first component of the results tuple here is `1.second`, meaning that the
+   * first `IO.realTime` evaluated after the clock had ''only'' advanced by `1.second`. This is
+   * in contrast to what would have happened with `control.advanceAndTick(1.second +
+   * 500.millis)`, which would have caused the first `realTime` to produce `1500.millis` as a
+   * result, rather than the correct answer of `1.second`. In other words, [[advanceAndTick]] is
```
armanbilge: Sorry, still confused 😕

The `advanceAndTick` example uses `IO.sleep(100.millis) *> IO.println("Hello, World!")`. Then it says
> This means that when `sleep` is finally submitted to the runtime, it is scheduled for the time offset equal to `1.second + 100.millis`

Why wouldn't the same thing happen here? I would have thought the `IO.realTime` would produce `2500.millis` actually.

kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/TestContext.scala
djspiewak: This is what the scaladoc is covering. :-)

testkit/shared/src/main/scala/cats/effect/testkit/TestControl.scala
djspiewak: `tickFor` will advance the clock by the minimum of however much time is needed to get to the next task, or the given bound. That minimum is important and it's what `advanceAndTick` *doesn't* do. What happens with `advanceAndTick` is you drive all the way out to the given bound, and *then* you tick, which has all sorts of bad consequences. `tickFor` probably more closely matches the intuition about "make time pass".

kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/TestContext.scala
armanbilge: AFAICT the scaladoc above this doesn't mention anything about deadlocks.
```
  /**
   * Returns the current interval between "now" and the earliest scheduled task. If there are
   * tasks which will run immediately, this will return `Duration.Zero`. Passing this value to
   * [[tick]] will guarantee minimum time-oriented progress on the task queue (e.g.
   * `tick(nextInterval())`).
   */
```

testkit/shared/src/main/scala/cats/effect/testkit/TestControl.scala
armanbilge: Right, I understand about `tickFor`. What I'm confused about is `advanceAndTick`.

For this code:
```scala
val program = IO.sleep(100.millis) *> IO.println("Hello, World!")
TestControl.execute(program).flatMap(_.advanceAndTick(1.second))
```
The described outcome is
> This means that when `sleep` is finally submitted to the runtime, it is scheduled for the time offset equal to `1.second + 100.millis`

Meanwhile for this code:
```scala
val tick = IO.sleep(1.second) *> IO.realTime
(tick, tick).tupled
```
The described outcome is
> This is  in contrast to what would have happened with `control.advanceAndTick(1.second + 500.millis)`, which would have caused the first `realTime` to produce `1500.millis` as a result

AFAICT, both programs start with a `sleep` for less than the `time` passed to `advanceAndTick`.

So the part I'm confused, is why the `sleep` in the first example is _additive_ (1s+100millis) while the `sleep` in the second example is simply taking the max(1s,1.5s)

djspiewak: Stared at this for about ten minutes before realizing it's simply a typo in the docs. :-P

armanbilge (approved):

PR #2966, opened 2022-04-18
Relax `LensRef` constraints

armanbilge:
Follow-up to https://github.com/typelevel/cats-effect/pull/2901#issuecomment-1100936339.

This drops the semantic from https://github.com/typelevel/cats-effect/pull/827#discussion_r404018298 that:
> writes to A that don't touch B don't break `access`'s setter

As a result, we are able to relax the constraints from `Sync[F]` to `Functor[F]` (!!) and drop the requirement that `B <: AnyRef`, both of which were very annoying in practice.

Thoughts? /cc @kubukoz who reviewed https://github.com/typelevel/cats-effect/pull/827.

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Ref.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Ref.scala
@@ -257,7 +257,7 @@ object Ref {
    * }}}
    */
   def lens[F[_], A, B](ref: Ref[F, A])(get: A => B, set: A => B => A)(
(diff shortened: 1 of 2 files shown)


armanbilge:
Oh, also some convo about that semantic in https://gitter.im/typelevel/cats-effect-dev?at=5e8b0e65d021c45cc7c995a6

"man we should have had this conversation on the ticket :(" - Fabio 😆

kubukoz (approved):
Happy to have it either way, just thought it'd be more efficient with this semantic.

PR #2967, opened 2022-04-20
Add Ref.empty Monoid-based shortcut

ivan-klass:
In practice, it's very common to initialize Ref with some empty values - e.g. with empty Option, Vector, Map, etc.
I find this builder method to be useful

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Ref.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Ref.scala
@@ -178,7 +178,7 @@ object Ref {
   /**
    * Creates a Ref with empty content
    */
-  def empty[F[_]: Make, A: Monoid]: F[Ref[F, A]] = of(Monoid[A].empty)
+  def ofEmpty[F[_]: Make, A: Monoid]: F[Ref[F, A]] = of(Monoid[A].empty)
 
   /**
    * Creates a Ref starting with the value of the one in `source`.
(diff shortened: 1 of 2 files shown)


ChristopherDavenport:
Personally a monoid is not empty in a way ref is. It's a different concept with the same name which causes confusion. Something like `def fromMonoidEmpty` makes it clear the two concept work differently.

ivan-klass:
@ChristopherDavenport sorry, I'm probably missing something. Is there some empty concept of the Ref itself? Like an unresolved Deferred? I can't see it in the sources

ivan-klass:
@ChristopherDavenport what about `ofEmpty` naming?
```scala
Ref[F].ofEmpty[String]
```

djspiewak:
@ChristopherDavenport Just to continue the conversation a bit late… Is a `Ref` ever really empty though? IMO `Ref.empty[A]` isn't a terrible naming, since `Ref` always contains some sort of value, so the only thing it *can* be would be the empty for `A`, which is to say `Monoid[A].empty`. It would definitely be different if we were talking about `Deferred`.

SystemFw:
I think this is a case of convenience over being super principled.

> since Ref always contains some sort of value, so the only thing it can be would be the empty for A, which is to say Monoid[A].empty

in particular there really is no such thing as the empty for `A`, `Monoid[A].empty` is defined in the context of `|+|`, so strictly speaking unless you only update the `Ref` monoidally, the `[A: Monoid]` constraint is not quite right, really what you're looking for is `alleycats` [Empty](https://github.com/typelevel/cats/blob/main/alleycats-core/src/main/scala/alleycats/Empty.scala), or arguably no typeclass at all (for the same reasons why `Empty` is in alleycats)

I have to say that in practice it is common to wanting to create `Ref`s with empty `Map`s, `List`s or `Option`s, and those happen to be `Monoid.empty`, so I see the appeal. What I would say though is to either not add it at all, or to add it as `empty`, the whole point is convenience here so cumbersome naming is not desirable

ChristopherDavenport:
I withdraw, I don't feel strongly enough here.

ivan-klass:
@SystemFw @djspiewak @ChristopherDavenport thank you for your feedback. I've reverted back to `.empty` method name

armanbilge (approved):
I will definitely be using this :)

SystemFw (approved):

PR #2970, opened 2022-04-24
IOAppSpec: running tests on Windows

### LINKED ISSUE #2601: Some IOAppSpec tests can be executed on all platforms
Currently, most of the example `IOApp` tests are not executed on Windows, because it doesn't support proper interruption of processes through the `java.lang.Process` API. These tests are the ones calling the `Handle#term` method. However, some tests do not call this method and instead wait for the monitored app to finish by itself. These should be run on Windows too.

The solution to this issue looks at `IOAppSpec.scala` and tries to run as much of the tests marked with `skipped("cannot observe graceful process termination on Windows")` unconditionally on all platforms. This probably means moving the tests from the `else` block above the `if/else` and removing the `skipped` tests. Feel free to reorganize the file as you see fit.

I'm tagging this issue as a "good first issue" so that new or less experienced people can pick it up, but I must warn them. Dealing with processes is annoying and tedious. Please do not get frustrated if something doesn't work on first try. Feel free to open a PR and the CI will do most of the testing.

That being said, this shouldn't be hard to do, but if you're stuck, we'll help you.

caiormt:
Running tests on Windows that doesn't depend on process interruption.

Closes #2601

### DIFF
--- a/tests/jvm/src/test/scala/cats/effect/IOAppSpec.scala
+++ b/tests/jvm/src/test/scala/cats/effect/IOAppSpec.scala
@@ -249,6 +249,9 @@ class IOAppSpec extends Specification {
         if (isJava8) {
           "live fiber snapshot" in skipped(
             "JDK 8 does not have free signals for live fiber snapshots")
+        } else if (isWindows) {
+          "live fiber snapshot" in skipped(
+            "cannot observe signals sent to process termination on Windows")
         } else {
           "live fiber snapshot" in {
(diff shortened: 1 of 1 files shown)


caiormt:
Broken test on Windows uses the Platform#sendSignal.

I assume that has the same root problem as Handle#term based on process interruption.

djspiewak (approved):
Thanks for taking this on! Sorry it took me so long to get to the review. :-(

PR #2971, opened 2022-04-24
SupervisorSpec: Covering all branches

### LINKED ISSUE #2950: Ensure `SupervisorSpec` is covering all cases
Right now, we're only testing the async branch. We need to ensure that we're explicitly covering both the async and concurrent implementations.

caiormt:
Explicit testing concurrent and async branches.

Closes #2950


djspiewak (approved):
This is great, thank you!

PR #2972, opened 2022-04-26
Fix leaking array ref in `Random`

catostrophe:
There's a bug in `cats.effect.std.Random` impl. In some cases `nextBytes` returns a ref to the same array. The bug is demonstrated in the newly added tests.

N.B. I changed the impl of both `ThreadLocalRandom` and `ScalaRandom`, although the bug appears only in the former. `ScalaRandom` works fine due to a delay in a preliminary `flatMap` before array creation. But I made the code identical to `ThreadLocalRandom` as to prevent the wrong pattern from being blindly copied somewhere else.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Random.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Random.scala
@@ -457,9 +457,12 @@ object Random extends RandomCompanionPlatform {
     def nextBytes(n: Int): F[Array[Byte]] =
       for {
         r <- f
-        bytes = new Array[Byte](0 max n)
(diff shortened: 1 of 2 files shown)


djspiewak:
Excellent catch! Should we be targeting this at `series/3.3.x` so we can release it sooner, though?

catostrophe:
Yes, shall I change the target branch?

armanbilge:
Yep, you'll also probably have to rebase and force-push :)

catostrophe:
Done. The build has failed on some unrelated (flaky?) js test. Could you restart?

armanbilge (approved):
Thank you!

PR #2976, opened 2022-05-03
add `useEval` (`useIt`) method to `Resource`

mberndt123:
Hey there,

I've added a method I call `useIt`, which is equivalent to `use(identity)`. As a motivation, consider this example from the cats-effect documentation:

```scala
  (
    for {
      in1 <- file("file1")
      in2 <- file("file2")
      out <- file("file3")
    } yield (in1, in2, out)
  ).use { case (file1, file2, file3) =>
    for {
      bytes1 <- read(file1)
      bytes2 <- read(file2)
      _ <- write(file3, bytes1 ++ bytes2)
    } yield ()
  }
```
This looks very clumsy to me. Every allocated resource is mentioned at least 3 times here: once to the left of `<-`, once to the right of `yield`, and then again to the right of `case`. The reader also needs to keep track of the fact that `in1` is the same as `file1` etc., creating opportunities for confusion.

I find it more elegant to simply `yield` the action to be performed and then call `useIt`. 

```scala
  (
    for {
      in1 <- file("file1")
      in2 <- file("file2")
      out <- file("file3")
    } yield for {
      bytes1 <- read(in1)
      bytes2 <- read(in2)
      _ <- write(out, bytes1 ++ bytes2)
    } yield ()
  ).useIt
```

There is precedent for this kind of shortcut, e. g. `combineAll` is a shortcut for `foldMap(identity)`, and I hope that having this method, along with an example in the documentation, helps people discover what I consider to be a neat little trick.

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -211,7 +211,7 @@ sealed abstract class Resource[F[_], +A] extends Serializable {
    * release it.
    */
 
-  def useIt[B](implicit ev: A <:< F[B], F: MonadCancel[F, Throwable]): F[B] =
+  def useEval[B](implicit ev: A <:< F[B], F: MonadCancel[F, Throwable]): F[B] =
     use(ev)
 
   /**
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -207,6 +207,14 @@ sealed abstract class Resource[F[_], +A] extends Serializable {
     fold(f, identity)
 
   /**
+   * For a resource that allocates an action (type `F[B]`), allocate that action, run it and
+   * release it.
+   */
+
+  def useIt[B](implicit ev: A <:< F[B], F: MonadCancel[F, Throwable]): F[B] =
+    use(ev)
+
+  /**
    * Allocates a resource with a non-terminating use action. Useful to run programs that are
    * expressed entirely in `Resource`.
    *


satorg:
Just a side note (regardless of the value of this change):
if there're several effects which are independent of each other, then sometimes a parallel composition can come in handy:
```scala
(file("file1"), file("file2"), file("file3"))
  .tupled
  .use { case (file1, file2, file3) =>
    for {
      (bytes1, bytes2) <- (read(file1), read(file2)).tupled
      _ <- write(file3, bytes1 ++ bytes2)
    } yield ()
  }
```
Also this approach (with small changes in the code above) allows running the independent effects concurrently.

mberndt123:
Yes, that's also something that I like to do. But it doesn't always work, sometimes you really do need flatMap.

Daenyth:
`use(identity)` is always a resource leak bug

The value you return from `use` will be a *closed* resource, so there's never a point to this shape and it certainly shouldn't be in the api

Edit:
I misread the shape of the api - `use(F.pure(_))` is the shape to avoid.

Even so, IMO this still isn't likely to be useful enough

mberndt123:
Hi @Daenyth,

Thank you for your feedback. 

I would like to better understand why you think it's not useful. Do you also think that `combineAll` is useless and people should write `foldMap(identity)` instead? Or do you think it's not a common use case? Because I see this pattern all the time. Almost every program works by allocating a bunch of resources at startup (DB connection pool, HTTP client, message broker connections, …) and then performing an IO action. So almost every realistic program has at least one use for this method :-)

SystemFw:
I don't think this method is all that useful, but I also don't think it's harmful, it's just a convenience for people that use `for` heavily (which I don't recommend, but it's of course legitimate). So, I'm not against adding it, although I'd probably name it `useNow`

djspiewak (commented):
Definitely on the fence about this. It's kind of skating pretty close to a shortcut for `use(pure(_))`, which is to say, a resource leak. I can see it being useful beyond just `for` comprehensions though. It's kind of like the inverse of `background`. Maybe something like `useEval` or `useFlatten` would be a more descriptive name?

mberndt123:
fwiw, I'd rather not call it `useNow` as that name is used in another effect system as a shortcut for `use(_.pure)`. That's useful for code such as `for (f <- openFile; content <- Resource.eval(readFile(f))) yield content` and doesn't necessarily imply a use-after-free error. I rather like the name `useIt`.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -206,6 +206,14 @@ sealed abstract class Resource[F[_], +A] extends Serializable {
   def use[B](f: A => F[B])(implicit F: MonadCancel[F, Throwable]): F[B] =
     fold(f, identity)
 
+  /**
+   * For a resource that allocates an action (type `F[B]`), allocate that action, run it and
+   * release it.
+   */
+
+  def useIt[B](implicit ev: A <:< F[B], F: MonadCancel[F, Throwable]): F[B] =
```
djspiewak:
```suggestion
  def useEval[B](implicit ev: A <:< F[B], F: MonadCancel[F, Throwable]): F[B] =
```

djspiewak (changes requested):
After sleeping on this for a while, I think I'm in favor of adding it. I am going to bike-shed the name though. :-) Let's go with `useEval` instead. `useFlatten` is also a reasonable alternative and I wouldn't quibble if you prefer it. `useIt` is nice and short, but it's also not particularly descriptive of what's going on here.

I'm not concerned about the resource leak side of things, since this is basically just a slightly indirect way of constructing an otherwise-safe `use`.

mberndt123:
@djspiewak,

OK, `useEval` seems reasonable. Thank you for your support.

djspiewak (approved):
Thanks for sticking with this! Sorry again for how busy I've been over the past few months.

PR #2978, opened 2022-05-05
Revert "Merge pull request #2868 from djspiewak/bug/report-fiber-failure"

armanbilge:
This reverts commit 1a0893ce0d60c6fe4fa93b789e1ba3f4d4d94863, reversing
changes made to 544b95d585ad9d0592945d66854e70fcee4db223.

See https://github.com/typelevel/cats-effect/issues/2962.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -473,10 +473,7 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
       ProblemFilters.exclude[MissingClassProblem]("cats.effect.unsafe.WorkerThread$Data"),
       // introduced by #2844, Thread local fallback weak bag
       // changes to `cats.effect.unsafe` package private code
-      ProblemFilters.exclude[MissingClassProblem]("cats.effect.unsafe.SynchronizedWeakBag"),
-      // introduced by #2868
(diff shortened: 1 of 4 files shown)


djspiewak:
I'm not sure we really need to go this far. The acute issue has been fixed more directly (though not released). Basically the semantic we have now is that we *over* log rather than *under* logging, and I would definitely prefer over-logging in general, particularly once we have configurability in `IOApp` (in 3.4.0).

armanbilge:
Yep, I missed the fact that it continued to be an issue because the fix was unreleased, see https://github.com/typelevel/cats-effect/issues/2962#issuecomment-1118728110. But I still think this can have all sorts of unintended consequences.

For example, even though it was fixed for `IO`, if a monad transformer (such as `Kleisli`) doesn't override that method then you end up in the same situation (we should check that, and fix it).

djspiewak:
Actually I think we can fix it in general here:

```scala
        fibA <- start(guaranteeCase(fa)(oc => result.complete(Left(oc)).void))
        fibB <- start(guaranteeCase(fb)(oc => result.complete(Right(oc)).void))
```

All we need is a `handleError(_ => ())` around the whole thing.

PR #2979, opened 2022-05-06
docs: update deps at getting started page

char16t:
Update versions of `weaver-cats` and `munit-cats-effect-3` libs to latest.


djspiewak:
Thanks for the catch!

PR #2984, opened 2022-05-10
Add On Air Entertainment to the list of adopters

ProxyLiveSolutions:
Added On Air Entertainment to the list of Cats Effect adopters.


vasilmkd (approved):
Thanks!

PR #2990, opened 2022-05-18
Docs clarification on how expensive interruptibleMany can be

TimWSpence:
Wow. Much expensive. Very exception


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1583,6 +1583,11 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
      * Like [[IO.blocking]] but will attempt to abort the blocking operation using thread
      * interrupts in the event of cancelation. The interrupt will be attempted repeatedly until
      * the blocking operation completes or exits.
+
+     * Note that this _really_ means what it says - it will throw exceptions in a
```
armanbilge: Maybe `@note`? :P

TimWSpence: I did not know that was a thing!

armanbilge (approved):

PR #3000, opened 2022-05-21
Harden queue cancelation semantics

### LINKED ISSUE #2920: Harden `Queue#take`'s cancelation properties
`take` must have the property such that it either does not remove the element, or it removes the element *and returns it*, even when canceled. Thus the whole thing must be uncancelable aside from an atomic `poll` (when semantic blocking is needed). Without this property, it is possible in rare race conditions to cancel a `take` and lose data.

Applies to both the existing `Concurrent` queue and the new `Async` queue in #2885

djspiewak:
This revises `Queue` to ensure that the values are always passed through the backing value queue, rather than being passed directly to the taker. This ensures that values can never be "lost" even if `take` is canceled. Note that I would really like to iterate the unit test a lot more times, but it starts timing out rather quickly.

Closes #2920


djspiewak:
Oh interestingly… I think this test isn't really demonstrating the issue, since `Dequeue` hasn't been fixed yet and still exhibits the issue, but the test passes.

djspiewak:
Confirmed by reverting that this new test does indeed reproduce the non-atomic `take` semantics. As it turns out, the non-`BoundedQueue`s are already hardened.

durban:
I think there is a chance of lost wakeups here: `offer` can wake up a `take` fiber which already have been cancelled; this causes other takers not to be waked. I wrote a test which (I think) demonstrates this:

```scala

import munit.FunSuite

import cats.effect.IO
import cats.effect.std.Queue
import cats.effect.unsafe.IORuntime

final class CatsQueueTest2 extends FunSuite {

  @volatile
  private[this] var taken1: String = null

  test("Cancel") {
    val tsk = for {
      q <- Queue.bounded[IO, String](capacity = 64)
      take1 <- IO.uncancelable { poll =>
        poll(q.take).flatTap { taken =>
          IO { this.taken1 = taken }
        }
      }.start
      take2 <- q.take.start
      _ <- IO.sleep(500.millis)
      _ <- IO.println("Starting race...")
      _ <- IO.both(q.offer("foo"), take1.cancel)
      _ <- IO.println("End race.")
      t1 <- IO { this.taken1 }
      _ <- if (t1 ne null) {
        IO.println("Too late") <* take2.cancel
      } else {
        IO.println("OK, joining...") <* take2.join
      }
    } yield ()
    (tsk.replicateA_(16)).unsafeRunSync()(IORuntime.global)
  }
}
```

This test occasionally deadlocks. I honestly don't know, which is worse: deadlocks like this, or losing elements (which is fixed by this PR).

djspiewak:
> I think there is a chance of lost wakeups here: offer can wake up a take fiber which already have been cancelled; this causes other takers not to be waked

Nice catch. I actually have a solve for this in the `Async` version which I think can be ported to the `Concurrent` one fairly easily.

PR #3001, opened 2022-05-22
Add `Async#executor`

armanbilge:
This is helpful for Java interop. See:
- https://github.com/typelevel/cats-effect/pull/2919#issuecomment-1101390355
- https://github.com/http4s/http4s-jdk-http-client/pull/641

/cc @wjoel

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
@@ -141,7 +141,7 @@ trait Async[F[_]] extends AsyncPlatform[F] with Sync[F] with Temporal[F] {
   def executionContext: F[ExecutionContext]
 
   /**
-   * Obtain a reference to the current execution context as a `java.util.concurrent.Executor]]`.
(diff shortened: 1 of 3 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
```diff
@@ -139,6 +140,14 @@ trait Async[F[_]] extends AsyncPlatform[F] with Sync[F] with Temporal[F] {
    */
   def executionContext: F[ExecutionContext]
 
+  /**
+   * Obtain a reference to the current execution context as a `java.util.concurrent.Executor]]`.
```
armanbilge:
```suggestion
   * Obtain a reference to the current execution context as a `java.util.concurrent.Executor`.
```

djspiewak:
Does this have to be source-split for ScalaJS?

armanbilge:
No, interestingly enough the `MacrotaskExecutor` is already an `ExecutionContextExecutor` :)

wjoel:
Looks great, much cleaner than my PR. :+1:

djspiewak (approved):

PR #3002, opened 2022-05-22
Highlight uncancelable behavior in some scaladocs

danicheg:
This is a follow-up to the Discord [discussion](https://discord.com/channels/632277896739946517/632278585700384799/977633287801557103). All credits to Daniel.


durban:
I don't know, what's in the linked discussion (I can't read it), but I don't think this is correct. Not having a finalizer doesn't make the effect uncancellable: the `case None =>` branch also has a `poll`.

djspiewak (dismissed):

PR #3007, opened 2022-05-24
Add combinators for choosing random elements of collections

cb372:
This PR adds a couple of handy functions to `Random` for choosing a random value from a collection.

I'm open to bikeshedding on the naming.

## oneOf

```scala
def oneOf[A](x: A, xs: A*): F[A]
```

Hopefully fairly self-explanatory.

But I wonder if it would make more sense to require at least 2 arguments like Scalacheck does:

```scala
def oneOf[A](x1: A, x2: A, xs: A*): F[A]
```

## elementOf

```scala
def elementOf[A](xs: Iterable[A]): F[A]
```

This takes an `Iterable` so you can choose a random element directly from e.g. a `Map` or a `Set` without having to first copy the contents to an intermediate `Seq`.

## Benchmarks

Performance summary:

* Choosing a random element from a `Vector` is O(1). This is as expected, since `Vector#size` and `Vector#apply` are both O(1).
* For `List` it's O(n). Also expected, because we need to traverse the list twice: once to calculate its size, and again to select the element at the randomly chosen index
* For `Map` it's slightly worse than O(n). Not quite sure why. I'd expect it to be O(n) and faster than `List` because we only do one traversal, not two (`HashMap#size` is O(1)).

Details:

Throughput, so higher is better.

```
[info] Benchmark                         (size)   Mode  Cnt       Score      Error  Units
[info] RandomBenchmark.elementOfList      10000  thrpt   10   21269.536 ±  848.270  ops/s
[info] RandomBenchmark.elementOfList     100000  thrpt   10    1865.345 ±  495.658  ops/s
[info] RandomBenchmark.elementOfList    1000000  thrpt   10     175.340 ±    3.155  ops/s
[info] RandomBenchmark.elementOfMap       10000  thrpt   10   21222.837 ± 2378.972  ops/s
[info] RandomBenchmark.elementOfMap      100000  thrpt   10    1432.057 ±  796.681  ops/s
[info] RandomBenchmark.elementOfMap     1000000  thrpt   10     114.991 ±    9.787  ops/s
[info] RandomBenchmark.elementOfVector    10000  thrpt   10  103479.262 ± 4450.984  ops/s
[info] RandomBenchmark.elementOfVector   100000  thrpt   10   96840.616 ± 7714.952  ops/s
[info] RandomBenchmark.elementOfVector  1000000  thrpt   10   93826.712 ± 5311.399  ops/s
```

## Alternative implementation: reservoir sampling

In an attempt to avoid traversing a List twice, I experimented with [reservoir sampling](https://en.wikipedia.org/wiki/Reservoir_sampling) (specifically Algorithm R, with k=1). That only does one traversal, as you doesn't need to calculate the list's length. On the other hand, it needs to generate one random integer per element of the list. Benchmarks confirmed it is dog slow, so I removed this implementation. Take a look at b022844fc37f384d6e9b8490b8147ab31584dee8 if you're interested.

Algorithm L ([Wikipedia](https://en.wikipedia.org/wiki/Reservoir_sampling#An_optimal_algorithm), [paper](https://dl.acm.org/doi/10.1145/198429.198435)) might be worth a try, as it doesn't require as many random numbers to be generated.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Random.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Random.scala
@@ -144,6 +144,19 @@ trait Random[F[_]] { self =>
   def shuffleVector[A](v: Vector[A]): F[Vector[A]]
 
   /**
+   * Pseudorandomly chooses one of the given values.
+   */
+  def oneOf[A](x: A, xs: A*): F[A]
+
+  /**
+   * Pseudorandomly chooses an element of the given collection.
+   *
+   * @return
+   *   a failed effect (NoSuchElementException) if the given collection is empty
+   */
+  def elementOf[A](xs: scala.collection.Seq[A], reservoir: Boolean): F[A]
+
+  /**
    * Modifies the context in which this [[Random]] operates using the natural transformation
    * `f`.
    *
@@ -206,6 +219,11 @@ trait Random[F[_]] { self =>
       override def shuffleVector[A](v: Vector[A]): G[Vector[A]] =
         f(self.shuffleVector(v))
 
+      override def oneOf[A](x: A, xs: A*): G[A] =
+        f(self.oneOf(x, xs: _*))
+
+      override def elementOf[A](xs: scala.collection.Seq[A], reservoir: Boolean): G[A] =
+        f(self.elementOf(xs, reservoir))
     }
 }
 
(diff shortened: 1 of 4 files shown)


std/shared/src/main/scala/cats/effect/std/Random.scala
```diff
@@ -143,6 +144,19 @@ trait Random[F[_]] { self =>
    */
   def shuffleVector[A](v: Vector[A]): F[Vector[A]]
 
+  /**
+   * Pseudorandomly chooses one of the given values.
+   */
+  def oneOf[A](x: A, xs: A*): F[A]
```
armanbilge: Unfortunately these changes are not binary-compatible 😕 

What we can do is is take an implicit constraint (probably a `Functor[F]` or `Monad[F]`) and use that to provide a default implementation.

cb372: I see, that's annoying. I didn't even consider bincompat.

I've done as you suggested and MiMa seems happy now.

I had to remove the new methods from `mapK` because I couldn't find a way to supply the implicit constraint for `F` rather than `G`. Is that ok?

djspiewak (approved):

PR #3008, opened 2022-05-25
Print fatal errors

### LINKED ISSUE #2961: `IO#unsafeRunAsync` and `unsafeRunAndForget` eat fatal errors
What happens here is `unsafeRunFiber` sets up the fatal error callback relocation thing, but then `unsafeRunAsync` has to box the error into a `Left`. This *may* work, but even if it works, `unsafeRunAndForget` is just going to drop it on the floor and the `IORuntime` silently dies. That seems like a bad experience.

I think what should probably happen here is we should register the callback as usual, but then in these kinds of cases where can't reliably continue the error relocation, we should just give up and dump the exception to stderr. Note that `unsafeToFuture` doesn't have this problem because the `Promise` is already allocated, so the fatal errors *can* be safely relocated into the resultant `Future` (and processed upon) without running into the cases where the JVM no longer functions.

sbly:
This is meant to address #2961. This is my first PR. I'm not sure I understood the ticket correctly. I'm not sure if this is the right place to address the issue, or if it needs to be done deeper, inside the IO runloop. Could I get some feedback? Thank you

I wrote a test for this locally to observe the behavior, but it seems with or without the change fatal errors get printed, which I think might be because of the test framework though. So I'm not sure how to test this behavior

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -797,9 +797,10 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
    * impure side effects.
    *
    * Any exceptions raised within the effect will be passed to the callback in the `Either`. The
-   * callback will be invoked at most *once*. Note that it is very possible to construct an IO
-   * which never returns while still never blocking a thread, and attempting to evaluate that IO
-   * with this method will result in a situation where the callback is *never* invoked.
+   * callback will be invoked at most *once*. In addition, fatal errors will be printed. Note
+   * that it is very possible to construct an IO which never returns while still never blocking
+   * a thread, and attempting to evaluate that IO with this method will result in a situation
+   * where the callback is *never* invoked.
    *
    * As the name says, this is an UNSAFE function as it is impure and performs side effects. You
    * should ideally only call this function ''once'', at the very end of your program.
@@ -808,8 +809,12 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
       implicit runtime: unsafe.IORuntime): Unit = {
     unsafeRunFiber(
       cb(Left(new CancellationException("The fiber was canceled"))),
-      t => cb(Left(t)),
-      a => cb(Right(a)))
+      t => {
+        if (NonFatal(t)) t.printStackTrace()
+        cb(Left(t))
+      },
+      a => cb(Right(a))
+    )
     ()
   }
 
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -832,7 +832,11 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
    * should never be totally silent.
    */
   def unsafeRunAndForget()(implicit runtime: unsafe.IORuntime): Unit =
-    unsafeRunAsync(_ => ())
+    unsafeRunAsync {
+      case Left(NonFatal(_)) => ()
+      case Left(e) => System.err.println(e)
```
armanbilge: I wonder if we can use `e.printStackTrace()` here instead :)

armanbilge (commented):
Thanks for the PR! To appease CI you can run `scalafixAll` in sbt. Perhaps we forgot to add this to the `prePR` command, sorry about that :)

core/shared/src/main/scala/cats/effect/IO.scala
sbly: Whoops, good point :)

sbly:
@armanbilge Thanks for the review! Running `scalafixAll` worked. I checked the `prePR` command and it should be calling it:

```
addCommandAlias(
  "prePR",
  "; root/clean; root/scalafixAll; scalafmtSbt; +root/scalafmtAll; +root/headerCreate")
```

but when I use `prePR` directly, it doesn't seem to call scalafix. I think maybe this has to do with project aggregation in sbt, but I'm not familiar enough with this project's structure to diagnose the issue

djspiewak (changes requested):
Right idea, but not quite there! :-) So the problem with `unsafeRunAndForget` comes from the fact that it is delegating to `unsafeRunAsync`, and `unsafeRunAsync` boxes fatal errors. Note that `unsafeRunAsyncOutcome` has the same issue.

There are two things we could do here, and I think I've convinced myself that one of them is better. Specifically, in `unsafeRunAsync` and `unsafeRunAsyncOutcome`, we check to see if the error is *not* `NonFatal` and then specifically print it in that case. We can still *try* to box it and send it along, but this is basically the last moment that we can be absolutely certain we have our hands on the error before more bad things become possible due to the type signatures.

In theory, `unsafeRunAndForget` shouldn't need to be changed.

sbly:
@djspiewak Thanks for the feedback! I implemented your suggestion.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -807,16 +809,27 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
       implicit runtime: unsafe.IORuntime): Unit = {
     unsafeRunFiber(
       cb(Left(new CancellationException("The fiber was canceled"))),
-      t => cb(Left(t)),
-      a => cb(Right(a)))
+      t => {
+        if (!NonFatal(t)) {
```
djspiewak, line 813: TIL `NonFatal` has an `apply` method.

djspiewak (approved):
Sorry this took me so long!

Issue #2558, opened 2021-11-17
"are source maps just… really really really slow?"


armanbilge:
Running the tracing example from the docs on Node.js with the exact same tracing parameters.
https://typelevel.org/cats-effect/docs/tracing#complete-code

```bash
CATS_EFFECT_TRACING_BUFFER_SIZE=64 CATS_EFFECT_TRACING_MODE=FULL node --enable-source-maps example/js/target/scala-2.13/cats-effect-example-fastopt/main.js
```

With sourcemaps:
```
real    0m4.981s
user    0m4.218s
sys     0m1.119s
```

Without:
```
real    0m0.169s
user    0m0.152s
sys     0m0.025s
```

One idea is to provide tooling/utility that takes unsourcemapped exceptions + a sourcemap and provides sourcemapped exceptions.

armanbilge:
Relevant to https://github.com/typelevel/cats-effect/issues/2284.

armanbilge:
Another idea to try is loading source-maps into memory (instead of file access) as demonstrated here:
https://www.npmjs.com/package/source-map-support#options

H/t @ChristopherDavenport

armanbilge:
A follow up on this: my measurements used `node --enable-source-maps` which actually doesn't use source-map-support (see https://github.com/scala-js/scala-js-js-envs/issues/20, https://github.com/typelevel/cats-effect/pull/2582) so the comment above might not be relevant actually.

Also, I'm not sure if using the default source map functionality in Scala.js (which uses source-map-support) has the same performance problem. I don't think so, but should check again.

Issue #2284, opened 2021-08-31
Do more benchmarking on ScalaJS tracing


djspiewak:
It's important to convince ourselves that we haven't somehow entirely destroyed performance on scalajs. The benchmark results are *really* inconsistent and it would be immensely helpful if someone who understands V8 and some of ScalaJS were able to help out on this a bit, since it's difficult to explain the numbers we're seeing in #2207.

Running `hey` benchmarks on http4s-server and even perhaps the Fs2 benchmarks on ScalaJS would probably be the best way to start to get some confidence over how this change manifests in realistic workloads.

djspiewak:
Requires #2684

armanbilge:
V8 performance notes, but might be a bit dated: https://github.com/thlorenz/v8-perf

PR #3011, opened 2022-05-30
Disable tracing for browsers by default on JS

armanbilge:
So this is kind of sad, and this PR is definitely intended to be a RFC. Right now the performance impact is just too detrimental, particularly in browsers/frontend where Scala.js is most popular. Many frontend projects are running into this and manually disabling it anyway, if not being totally turned off from using CE.js because it seems super slow. So at least at this point in time, I wonder if tracing enabled by default in dev mode is doing more harm than good, and it should be opt-in for when you're actively trying to debug something 😕

Linking to:
- https://github.com/typelevel/cats-effect/issues/2634#issuecomment-991202431
- https://github.com/typelevel/cats-effect/issues/2284
- https://github.com/typelevel/cats-effect/issues/2558

### DIFF
--- a/core/js/src/main/scala/cats/effect/tracing/TracingConstants.scala
+++ b/core/js/src/main/scala/cats/effect/tracing/TracingConstants.scala
@@ -20,7 +20,7 @@ package tracing
 private[effect] object TracingConstants {
 
   private[this] final val stackTracingMode: String =
-    process.env("CATS_EFFECT_TRACING_MODE").filterNot(_.isEmpty).getOrElse("cached")
+    process.env("CATS_EFFECT_TRACING_MODE").filterNot(_.isEmpty).getOrElse("none")
 
   final val isCachedStackTracing: Boolean = stackTracingMode.equalsIgnoreCase("cached")
 
--- a/core/js/src/main/scala/cats/effect/tracing/TracingConstants.scala
+++ b/core/js/src/main/scala/cats/effect/tracing/TracingConstants.scala
@@ -17,10 +17,19 @@
 package cats.effect
 package tracing
 
+import scala.scalajs.js
+
 private[effect] object TracingConstants {
 
   private[this] final val stackTracingMode: String =
-    process.env("CATS_EFFECT_TRACING_MODE").filterNot(_.isEmpty).getOrElse("none")
+    process.env("CATS_EFFECT_TRACING_MODE").filterNot(_.isEmpty).getOrElse {
+      if (js.typeOf(js.Dynamic.global.process) != "undefined"
+        && js.typeOf(js.Dynamic.global.process.release) != "undefined"
+        && js.Dynamic.global.process.release.name == "node".asInstanceOf[js.Any])
+        "cached"
+      else
+        "none"
+    }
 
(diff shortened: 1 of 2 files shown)


djspiewak:
Definitely super sad. I think the core assumption we made is that performance is less relevant in dev mode. Is that assumption where we erred?

armanbilge:
> I think the core assumption we made is that performance is less relevant in dev mode.

I think this is definitely true, but only to an extent, and currently the performance impact is just too noticeable. To be fair, I haven't put enough effort into trying to better understand why this is and see if we can improve it somehow.

Politely pinging @davesmith00000, @buntec, and @cquiroz for their thoughts.

armanbilge:
Also politely pinging @kubukoz who has used CE in Node.js, for a non-frontend perspective on this, thanks!

davesmith00000:
> I think the core assumption we made is that performance is less relevant in dev mode. Is that assumption where we erred?

I can see how you'd arrive at that assumption, thinking of things like AWS Lambda's and microservices.

My own experience:

I look after two Scala.js libraries. Indigo (game engine) and Tyrian (Elm-like web framework). Indigo is just straight up Scala.js and I recently converted Tyrian to use CE3, replacing a home-brew `Task` effect type that was designed only with Scala.js in mind.

In terms of performance, Indigo runs almost as quickly under `fastOpt` as `fullOpt` - the main win of the latter is a slight performance edge, slightly better memory use, and a massive improvement in file size. Tyrian used to be the same. When I converted it to CE3, the performance drop under `fastOpt` was so noticeable that I assumed I'd messed up the implementation. Literally watching console log lines appear one after another, and I was saved from pulling it out again by @armanbilge saying "Oh, did you try `fullOpt`?". 😄

That was weeks ago, I only found out about the tracing/debugging and that you can disable it today. 🤷 😄 

Unlike the lambda / microservice use-case, in web dev, `fastOpt` as a 'dev mode' is about fast feedback, coding, trying, changing things, trying it again - visual things you can't really unit test because they're about how they feel subjectively rather than provable behaviour. If the performance difference is very high, the app feels different and that makes your job harder.

Personally I'd be in favour of inverting the logic so that the tracing is opt-in, however, if you're expecting most Scala.js folks to be using CE3 for server-side/server-less components, then I'd understand a decision to leave it as-is.

Just my ten cents, and for what it's worth, I'm glad I did the conversion to CE3 in the end. 😄

cquiroz:
The effect is noticeable, we ported to CE3 early on and when tracing was introduced the difference was right there. It took as a while to understand why and we  are explicitly disabling it. This is the context of front end apps in the browser

I adhere to @davesmith00000 comments that on development we want a fast development cycle thus maybe be better to disable it by default or make it opt-in.

armanbilge:
Thank you all for the feedback! I really appreciate your insights. I think it makes sense to go ahead with this change, although perhaps we can still enable it automatically if the runtime is Node.js. Still, it's pretty easy to set an env variable for Node.js.

More pressing, I need to find a way to re-activate tracing for CI across all the environments ...

djspiewak (approved):
Much sadness. We should document this, and hopefully also figure out a way to resolve the performance issues.

PR #3012, opened 2022-06-01
Rename worker threads in blocking regions

### LINKED ISSUE #3010: Change thread name when it becomes a "blocker"
Simple thing, but when a thread transmutes into a blocking agent, we should change its name from `io-compute` to `io-blocker` or something like that. This seems to come up a lot in Discord and it's not a difficult solve. :-)

aeons:
This adds a new parameter to the `WorkStealingThreadPool` constructor, but it looks like the constructor is private to the project.

Closes #3010

### DIFF
--- a/tests/jvm/src/test/scala/cats/effect/unsafe/WorkerThreadNameSpec.scala
+++ b/tests/jvm/src/test/scala/cats/effect/unsafe/WorkerThreadNameSpec.scala
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2020-2022 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.unsafe
+
+import cats.effect.{BaseSpec, IO}
+import cats.effect.testkit.TestInstances
+import cats.syntax.all._
+
+class WorkerThreadNameSpec extends BaseSpec with TestInstances {
+
+  override def runtime(): IORuntime = {
+    lazy val rt: IORuntime = {
+      val (blocking, blockDown) =
+        IORuntime.createDefaultBlockingExecutionContext(threadPrefix =
+          s"io-blocking-${getClass.getName}")
+      val (scheduler, schedDown) =
+        IORuntime.createDefaultScheduler(threadPrefix = s"io-scheduler-${getClass.getName}")
+      val (compute, compDown) =
+        IORuntime.createDefaultComputeThreadPool(
+          rt,
+          threads = 1,
+          threadPrefix = s"io-compute-${getClass.getName}",
+          blockerThreadPrefix = s"io-blocker-${getClass.getName}")
+
+      IORuntime(
+        compute,
+        blocking,
+        scheduler,
+        { () =>
+          compDown()
+          blockDown()
+          schedDown()
+        },
+        IORuntimeConfig()
+      )
+    }
+
+    rt
+  }
+
+  "WorkerThread" should {
+    "rename itself when entering and exiting blocking region" in real {
+      for {
+        computeThread <- threadInfo
+        (computeThreadName, _) = computeThread
+        blockerThread <- IO.blocking(threadInfo).flatten
+        (blockerThreadName, blockerThreadId) = blockerThread
+        _ <- IO.cede
+        // Force the previously blocking thread to become a compute thread by converting
+        // the pool of compute threads (size=1) to blocker threads
+        resetComputeThreads <- List.fill(2)(threadInfo <* IO.blocking(())).parSequence
+      } yield {
+        // Start with the regular prefix
+        computeThreadName must startWith("io-compute")
+        // Check that entering a blocking region changes the name
+        blockerThreadName must startWith("io-blocker")
+        // Check that the same thread is renamed again when it is readded to the compute pool
+        resetComputeThreads.exists {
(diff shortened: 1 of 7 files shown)


tests/jvm/src/test/scala/cats/effect/unsafe/WorkerThreadNameSpec.scala
```diff
@@ -0,0 +1,43 @@
...
+
+import scala.concurrent.blocking
+
+class WorkerThreadNameSpec extends BaseSpec with TestInstances {
+
+  "WorkerThread" should {
+    "rename itself when entering and exiting blocking region" in real {
+      for {
+        computeThreadName <- getThreadName
+        blockerThreadName <- blocking { getThreadName }
```
aeons: Is it correctly understood that this works the same as using `IO.blocking`?

armanbilge: I'm not sure, so Vasil will have to comment. But my understanding was the `IO.blocking` calls can be shifted to an existing blocking thread if available. However because `blocking { }` is synchronous, it forces the _current_ thread to become a blocking thread. So possibly not the same.

Also the semantics here seem a bit strange, maybe you actually want `IO(blocking(Thread.currentThread().getName()))`?

tests/jvm/src/test/scala/cats/effect/unsafe/WorkerThreadNameSpec.scala
```diff
@@ -0,0 +1,43 @@
...
+        _ <- IO.cede
+        resetComputeThreadName <- getThreadName
```
armanbilge: From my understanding, this doesn't check if the thread name is reset. All it's checking is that the fiber is now running on a compute thread, which is not the same thing.

It seems to me all of your assertions should check that the underlying thread is the same (by reference equality) so that we are in fact seeing name changes (and not simply thread jumps).

armanbilge:
Oh also, this PR can probably be targeted at 3.3.x, so we can release it sooner :)

tests/jvm/src/test/scala/cats/effect/unsafe/WorkerThreadNameSpec.scala
vasilmkd: I would not test `scala.concurrent.blocking` and I would advise against using it to wrap effects.

The more correct usage is to wrap effectful code directly, like Arman suggested.

vasilmkd: > I'm not sure, so Vasil will have to comment. But my understanding was the IO.blocking calls can be shifted to an existing blocking thread if available. However because blocking { } is synchronous, it forces the current thread to become a blocking thread. So possibly not the same.

`IO.blocking` calls are either:
1. Rewritten as `IO(scala.concurrent.blocking(thunk))` and then executed in place, by making the current compute thread become a blocking thread and replacing it with a cached/new thread.
2. Shifted completely to the `blocking` pool and then shifted back to the `compute` pool.

Cached threads that blocked at one time can be used as replacements for _compute_ threads again, but not directly as blocking threads.

armanbilge: > Cached threads that blocked at one time can be used as replacements for compute threads again, but not directly as blocking threads.

Oh interesting, at the risk of going on a tangent why is that?

aeons: Changing it to `IO.blocking`

tests/jvm/src/test/scala/cats/effect/unsafe/WorkerThreadNameSpec.scala
aeons: You are right. My first test that didn't work, without the `cede`, had the same thread id for all three fibers. Now the first two are the same and correctly changes name, but the last one is on a different thread.

tests/jvm/src/test/scala/cats/effect/unsafe/WorkerThreadNameSpec.scala
vasilmkd: > Oh interesting, at the risk of going on a tangent why is that?

If you shift a fiber to a dedicated blocking thread (regardless of whether it's cached or spawned just now) you incur 2 context switches (one by switching to the blocking thread and one by switching back to the compute pool). Runnin the blocking code in place and replacing the thread only incurs 1 context switch, when you go back to the compute pool. This is demonstrably quicker on benchmarks, something CE was criticized for (inefficiency of blocking code execution) now it benches the same as the best of them.

vasilmkd: https://github.com/typelevel/cats-effect/pull/2687

tests/jvm/src/test/scala/cats/effect/unsafe/WorkerThreadNameSpec.scala
aeons: I removed my debug println (😳) and updated the test. 

What it does now is 
- create a runtime with a single thread threadpool, 
- check that the compute thread is correctly renamed when entering a blocking region, 
- then force two threads to become blocking, 
- thus forcing the previously blocking thread to be readded to the compute pool
- and verify that the thread has been renamed again

It works, but feels iffy.

aeons:
I guess that involves rebasing on series/3.3.x and changing the target branch here, right?

Just confirming before I force push.

vasilmkd:
I would like to review this PR too. Please give me a few days. Sorry for the delay.

armanbilge:
Sorry, missed your message! Yes, you should rebase and force push.

aeons:
It should be ready now, targeting `series/3.3.x`.

djspiewak:
Holding the merge on this until @vasilmkd approves.

djspiewak:
@vasilmkd Any availability to review within the next day or so? I'd like to get this merged. All good if you're not able to get to it.

vasilmkd:
I will review it tomorrow, I don't have a computer at the moment. I'm sorry for the delay. 😄

vasilmkd (approved):
Great work, thanks for chasing this down, and sorry for the delay.

Issue #2639, opened 2021-12-10
Add peek method to Queue


esgott:
The original `fs2.Queue` had a `peek` method on its interface. The `Queue` in Cats Effect doesn't have this. It would be good to add it, as it makes the migration to Cats Effect 3 easier.

MrKustra94:
I may take it :)

armanbilge:
Thanks, much appreciated! Unfortunately I'm not sure if it's possible to implement this one in a backwards-compatible way, see some commentary in https://github.com/typelevel/cats-effect/issues/2637#issuecomment-991954088.

Issue #512, opened 2019-04-24
Proposal: IO#raceMany that accepts collection of IO[T], returns winner and cancels others


Adam-Kurzawa:
I've been considering a race scenario implementation for a collection of IOs. Concept itself is similar to IO#race but stretched out to many IOs. Input is a collection of IO[T]. The return value is a race winner (first completed) wrapped as IO[Either[Throwable, T]] while other IOs are canceled. 

I've pushed an example to my repo, just have in mind this is quick and dirty effort of one evening ;) Based on original IO#race code.

https://github.com/Adam-Kurzawa/cats-race-many/blob/master/src/main/scala/cats/effect/internals/IORaceCollection.scala

djspiewak:
Hmm, what would this be useful for? The "cancel everything" semantic is pretty blunt.

LukaJCB:
I think this could easily be done using something like `foldMapK` and a `MonoidK` instance that races, something similar to what's discussed in #405

djspiewak:
Exactly. The only advantage to putting it on `IO` would be for optimization reasons.

Adam-Kurzawa:
I will reveal some more context behind this idea. I have seen a code depending on a racing scenario - it used from 2 to 4 (depending on config env) parrallel calls to cache to mitigate network latency spikes. Later it was implemented into HTTP request mechanism to handle one troublesome service. Implementation was done by folding list of IOs into IO.race, like
```
list.foldLeft(io) { racingAcc, nextInList => IO.race(racingAcc, nextInList) }
```
It worked but with some minor issues when used in racing HTTP requests. Also the whole IO-folding code was ugly when combined with some exception handling and logging (excerpt above is just to show a concept). I tried to refactor it but didn't end with anything better than fold. After that I experimented with Cats Effect internals and came with this idea.
1. @LukaJCB I will have a look into this. Maybe that is the solution after all.
2. I think generalisation is a good thing. Why are we limiting racing to just 2 elements? Why not to 3? IMHO racing a collection makes Cats Effects API more elegant and doesn't limit users' creativity.
3. If this low-level race turns out more performant than some folding/flatmaping magic than hell yeah!

djspiewak:
@Adam-Kurzawa Did you have a chance to look at the `SemigroupK` solution? I can definitely see value in what you're proposing here (only two years later!).

Adam-Kurzawa:
[...]

PR #3013, opened 2022-06-02
Add peek method to all queue interfaces and implementations

sbly:
This implements #2639 

This only adds methods, but it seems it may break backwards-compatibility according to that discussion. Is that because it adds methods to an interface?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,6 @@ metals.sbt
 
 # npm
 node_modules/
+
+# mill
+out/
--- a/std/shared/src/main/scala/cats/effect/std/Dequeue.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Dequeue.scala
@@ -49,6 +49,8 @@ trait Dequeue[F[_], A] extends Queue[F, A] with DequeueSource[F, A] with Dequeue
       def tryOfferFront(a: A): G[Boolean] = f(self.tryOfferFront(a))
       def takeFront: G[A] = f(self.takeFront)
       def tryTakeFront: G[Option[A]] = f(self.tryTakeFront)
+      def peekFront: G[Option[A]] = f(self.peekFront)
+      def peekBack: G[Option[A]] = f(self.peekBack)
       def reverse: G[Unit] = f(self.reverse)
       def size: G[Int] = f(self.size)
     }
@@ -97,6 +99,12 @@ object Dequeue {
           override def takeFront: F[B] =
             fa.takeFront.map(f)
 
+          override def peekFront: F[Option[B]] =
+            fa.peekFront.map(_.map(f))
+
+          override def peekBack: F[Option[B]] =
+            fa.peekBack.map(_.map(f))
+
           override def tryTakeFront: F[Option[B]] =
             fa.tryTakeFront.map(_.map(f))
 
@@ -134,6 +142,9 @@ object Dequeue {
     override def tryTakeBack: F[Option[A]] =
       _tryTake(queue => queue.tryPopBack)
 
+    override def peekBack: F[Option[A]] =
+      _peek(queue => queue.peekBack)
+
     override def offerFront(a: A): F[Unit] =
       _offer(a, queue => queue.pushFront(a))
 
@@ -145,6 +156,9 @@ object Dequeue {
 
     override def tryTakeFront: F[Option[A]] =
       _tryTake(queue => queue.tryPopFront)
+
+    override def peekFront: F[Option[A]] =
+      _peek(queue => queue.peekFront)
 
     override def reverse: F[Unit] =
       state.update {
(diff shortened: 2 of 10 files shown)


armanbilge:
Thanks for the PR!

> Is that because it adds methods to an interface?

Right, exactly. We can only add new methods if we also can provide a default implementation. Unfortnately I don't think we can do that here.

Also, I was a bit confused about the value of peek for a concurrent queue, see https://github.com/typelevel/cats-effect/issues/2637#issuecomment-991954827, but that's just my opinion :)

sbly:
@armanbilge No problem! Thanks for the explanation

Yeah you're right that it might not be that useful in concurrent situations. It seems like the original motivation was simply to make the migration to CE3 easier. That said, if it's decided that this isn't worth it i'm OK with that too

armanbilge:
> It seems like the original motivation was simply to make the migration to CE3 easier. That said, if it's decided that this isn't worth it i'm OK with that too

It's certainly worth considering if we can provide a default implementation, so that this change is binary-compatible :)

sbly:
@armanbilge Unfortunately, I don't see a reasonable way to do that. Even `tryTake.flatMap(tryOffer)` wouldn't work. Any suggestions?

LMnet:
I want to add a few thoughts about necessity of the `peek` method. I have a pretty complex concurrent component in my codebase which is heavily rely on a this method. I will not go too much into details about my component, but in general it uses `peek` to wait for the first new element in an empty queue. And after that it use `take` to get an actual element. I can't use `take` without a `peek`, because I have more than 1 queue, and I want to get an element only from the first filled queue. With `take` I will get elements from all queues, but it's not what I want.

So, `peek` is usable sometimes. Although, for my specific use-case method like `waitForNonEmpty` will be enough.

armanbilge:
@LMnet If I understand your problem correctly, I think you are looking for `race` :)

```scala
IO.race(queue1.take, queue2.take)
```

The first queue that gets an element will win the race and `take` successfully, and the other `take` will be cancelled.

If you want to race `take` on multiple queues you can have a look at https://github.com/typelevel/cats-effect/issues/512#issuecomment-1072337740.

djspiewak:
Sorry for the delay in getting to this…

My concern about this method is there really isn't a way to do it atomically in a default implementation. Now, seeing as we control all realistic `Queue` implementations, maybe that's not such a huge concern, but it's definitely something to think about. Are we okay with it being non-atomic (and thus potentially corrupting ordering and/or fiber blocking) on third-party implementations of `Queue`?

armanbilge:
> Now, seeing as we control all realistic `Queue` implementations, maybe that's not such a huge concern

If we really feel that way, a "safer" way forward might be to seal `Queue` in Cats Effect 3.4, but not make any other changes. This will be source-breaking, but not binary-breaking. Assuming that nobody complains about this, I'd say from 3.5 forward we can freely add new abstract methods to the `Queue` interface, including `peek`.

LMnet:
<details>
  <summary>@armanbilge since it's off-topic, I have hided my answer under a spoiler</summary>
  
  > If I understand your problem correctly, I think you are looking for race :)
 
  In theory it should work. But in practice `IO.race(queue1.take, queue2.take)` will return the first element, but will not cancel the other `take`. Because `take` is uncancellable inside. You can check it through the simple test:
  ```scala
  for {
    queue1 <- Queue.unbounded[IO, String]
    queue2 <- Queue.unbounded[IO, String]
    _ <- queue1.offer("1")
    _ <- queue2.offer("2")
    res <- IO.race(queue1.take, queue2.take)
    size1 <- queue1.size
    size2 <- queue2.size
  } yield {
    assert(size1 + size2 == 1)
  }
  ```
  
  This test will always fail, because both `take` will dequeue elements from their queues. But because of `race` only one of the results will be returned. Another one will be dropped.

  Maybe it's somehow possible to support cancellation in the queue, but now it's not working.
</details>

djspiewak:
> If we really feel that way, a "safer" way forward might be to seal Queue in Cats Effect 3.4, but not make any other changes. This will be source-breaking, but not binary-breaking. Assuming that nobody complains about this, I'd say from 3.5 forward we can freely add new abstract methods to the Queue interface, including peek.

I mean, not really. :-) It still breaks bincompat with anything that was compiled before the `sealed` modifier was added. In particular, anyone calling a hypothetical abstract `peek` on `Queue` where the underlying implementation predates the addition of the `sealed` modifier would get a link error.

armanbilge:
> In particular, anyone calling a hypothetical abstract `peek` on `Queue` where the underlying implementation predates the addition of the `sealed` modifier would get a link error.

Is that better or worse than silently corrupting said `Queue`? :)

LMnet:
Is there any chances that this pr will be merged and released with the next patch version?

Jasper-M:
IMO the method being added here should be called `tryPeek`. `peek` should be like `take` without actually removing the element.

djspiewak:
I genuinely don't think that there's a way this can be done safely in the current API. Given the tradeoffs involved in this *type* of function, I would rather it exist in userspace (`q.tryTake.flatMap(_.traverse(q.offer(_))).uncancelable`), where those tradeoffs are more directly visible.

PR #3039, opened 2022-06-19
Add `Async#fromJavaFuture`

armanbilge:
~~Still needs some tests.~~ /cc @Daenyth

### DIFF
--- a/kernel/jvm/src/main/scala/cats/effect/kernel/AsyncPlatform.scala
+++ b/kernel/jvm/src/main/scala/cats/effect/kernel/AsyncPlatform.scala
@@ -16,7 +16,13 @@
 
 package cats.effect.kernel
 
-import java.util.concurrent.{CompletableFuture, CompletionException, CompletionStage}
+import java.util.concurrent.{
+  CancellationException,
+  CompletableFuture,
+  CompletionException,
+  CompletionStage,
+  Future
+}
 
 private[kernel] trait AsyncPlatform[F[_]] extends Serializable { this: Async[F] =>
 
@@ -53,4 +59,16 @@ private[kernel] trait AsyncPlatform[F[_]] extends Serializable { this: Async[F] 
         }
       }
     }
+
+  def fromJavaFuture[A](fut: F[Future[A]], mayInterruptIfRunning: Boolean): F[A] =
+    flatMap(fut) { fut =>
+      map(
+        race(
+          onCancel(never, void(delay(fut.cancel(mayInterruptIfRunning)))),
+          recoverWith(blocking(fut.get())) {
+            case _: CancellationException => productR(canceled)(never)
+          }
+        ))(_.toOption.get)
+    }
+
 }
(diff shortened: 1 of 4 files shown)


tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
```diff
@@ -274,7 +274,9 @@ class DispatcherSpec extends BaseSpec {
             _ <- cdl.await // make sure the execution of fiber has started
           } yield ()
         }.start
-        _ <- releaseInner.await.timeoutTo(2.seconds, IO(false must beTrue)) // release process has started
+        _ <- releaseInner
+          .await
+          .timeoutTo(2.seconds, IO(false must beTrue)) // release process has started
```
armanbilge, line 279: Hmm, not sure why this changed.

djspiewak:
I'm a little on the fence about this one. IMO if we're going to do it, we should at least double-check to see if the `Future` is secretly a `CompletableFuture`, to avoid creating a foot gun.

armanbilge:
Oh, I missed the fact that `CompletableFuture` extends `Future`. Fair point.

armanbilge:
For the record I'm skeptical about this one too now. Somebody asked on the Discord about just `Future` and the cancellation logic doesn't come for free so it seemed worth adding.

kernel/jvm/src/main/scala/cats/effect/kernel/AsyncPlatform.scala
```diff
@@ -53,4 +59,18 @@ private[kernel] trait AsyncPlatform[F[_]] extends Serializable { this: Async[F]
         }
       }
     }
+
+  def fromJavaFuture[A](fut: F[Future[A]], mayInterruptIfRunning: Boolean): F[A] =
+    flatMap(fut) {
+      case fut: CompletableFuture[A] => fromCompletableFuture(pure(fut))
```
armanbilge, line 65: I can't remember why we don't use `mayInterruptIfRunning` for `fromCompletableFuture`.

djspiewak: I think it's because interruption in general is weird and hard to be confident in.

armanbilge: In that case maybe should I remove it for `fromJavaFuture` too?

Daenyth (approved):
lgtm, better than what I wrote before

armanbilge:
More thoughts from Daniel:
> second, it makes it seem like it's fine to convert from Future to IO, in the same way that it is with CF
> when really it's not
> this in turn may lead users to select APIs which return Future rather than CF
> since they seem equivalent
> basically the argument is that putting something in the API elevates the case to something more "supported" and "okay"
> but really it's not okay. it's terrible. sometimes you have to do it, but not often

So that seems like a convincing argument not to add this. The workaround is to do `IO.blocking(future.get())` or inline what I have here if you want cancelation.

PR #3044, opened 2022-06-21
Add Sync helpers for scala.Predef assertions

DavidGregory084:
I'm opening this for discussion after the idea came up while chatting with @armanbilge on typelevel/bobcats#106.

I'm not 100% sure whether this is a useful addition - I have some sympathy for the idea that `assert` and `require` express critical preconditions and that it's okay if the runtime explodes if those preconditions are not satisfied.

I chose `Sync` because I think it's the first place in the hierarchy that knows that the error channel is `Throwable`.

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Sync.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Sync.scala
@@ -229,7 +229,6 @@ trait Sync[F[_]] extends MonadCancel[F, Throwable] with Clock[F] with Unique[F] 
    *
    * @param requirement
    *   the expression to test
-   * @return
    */
   def require(requirement: Boolean): F[Unit] =
     delay(Predef.require(requirement))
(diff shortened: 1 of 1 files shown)


armanbilge:
> I chose `Sync` because I think it's the first place in the hierarchy that knows that the error channel is `Throwable`.

Thanks for the PR! I guess I wasn't very clear in https://github.com/typelevel/bobcats/pull/106#discussion_r901722682 but actually I think these belong in Cats alongside `ApplicativeThrow#catchNonFatal` and friends (note that `ApplicativeThrow` is a type alias for `ApplicativeError[*, Throwable]`).

`Sync` is a very strong constraint and may not always be where this functionality is needed.

DavidGregory084:
Hmm OK, so we would add these as extension methods on `ApplicativeThrow`?

armanbilge:
So Cats add the methods to `ApplicativeError` but requires evidence that `Throwable <:< E`
https://github.com/typelevel/cats/blob/9839b7498e81b1405d3e41c88d24ce73e316e69d/core/src/main/scala/cats/ApplicativeError.scala#L247

rossabaker:
Because I had to look it up, `AssertionError` is `NonFatal`.  I agree that `ApplicativeError` would work well.

DavidGregory084:
Superseded by typelevel/cats#4249

PR #3046, opened 2022-06-23
Add documentation for `IOLocal`

### LINKED ISSUE #2117: Write documentation for `IOLocal`
What it says on the tin. It needs both scaladoc and should also have a page in the site (**docs/** subdirectory).

iRevive:
Closes https://github.com/typelevel/cats-effect/issues/2117.

It is a compilation of comments from the discord mostly. And almost a real-world usage example.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOLocal.scala
+++ b/core/shared/src/main/scala/cats/effect/IOLocal.scala
@@ -25,12 +25,13 @@ package cats.effect
  * That means, two fibers can never access the same [[IOLocal]], they will always be working on
  * their own copies.
  *
- * In some scenarios, [[IOLocal]] can be considered as a replacement for
+ * In some scenarios, [[IOLocal]] can be considered as an alternative to
  * [[https://typelevel.org/cats-mtl/mtl-classes/local.html cats.mtl.Local]]
  *
  * @example
  *   {{{
  *  import cats.Monad
+ *  import cats.effect.{IO, IOLocal, Sync, Resource}
  *  import cats.effect.std.{Console, Random}
  *  import cats.syntax.flatMap._
  *  import cats.syntax.functor._
--- a/core/shared/src/main/scala/cats/effect/IOLocal.scala
+++ b/core/shared/src/main/scala/cats/effect/IOLocal.scala
@@ -19,33 +19,40 @@ package cats.effect
 /**
  * [[IOLocal]] provides a handy way of manipulating a context on different scopes.
  *
- * In some scenarios, [[IOLocal]] can be considered as an alternative to [[cats.data.Reader]] or
- * [[cats.data.Kleisli]].
+ * In some scenarios, [[IOLocal]] can be considered as an alternative to [[cats.data.Kleisli]].
  *
  * [[IOLocal]] should not be treated as [[cats.effect.kernel.Ref Ref]], since the former abides
  * different laws.
  *
- * For example, two forked fibers can never access the same [[IOLocal]], they will always be
- * working on their own copies.
+ * Once a fiber is forked, for example by `Spawn[F].start`, the forked fiber manipulates the
+ * copy of the parent's context. For example, two forked fibers will never see each other's
+ * modifications to the same [[IOLocal]], each fiber will only see its own modifications.
  *
  * ===Operations on [[IOLocal]] are visible to the fiber===
  *
  * {{{
- *  def inc(idx: Int, local: IOLocal[Int]): IO[Unit] =
- *    local.update(_ + 1) >> local.get.flatMap(current => IO.println(s"child $$idx: $$current"))
+ * ┌────────────┐               ┌────────────┐               ┌────────────┐
+ * │  Fiber A   │ update(_ + 1) │  Fiber A   │ update(_ + 1) │  Fiber A   │
+ * │ (local 42) │──────────────►│ (local 43) │──────────────►│ (local 44) │
+ * └────────────┘               └────────────┘               └────────────┘
+ * }}}
+ *
+ * {{{
+ *  def inc(name: String, local: IOLocal[Int]): IO[Unit] =
+ *    local.update(_ + 1) >> local.get.flatMap(current => IO.println(s"fiber $$name: $$current"))
  *
  *  for {
  *    local   <- IOLocal(42)
  *    _       <- inc(1, local)
  *    _       <- inc(2, local)
  *    current <- local.get
- *    _       <- IO.println("parent: $$current")
+ *    _       <- IO.println(s"fiber A: $$current")
  *  } yield ()
  *
- *  // output
- *  // child 1: 43
- *  // child 2: 44
- *  // parent: 44
+ *  // output:
+ *  // update 1: 43
+ *  // update 2: 44
+ *  // fiber A: 44
  * }}}
  *
  * ===A forked fiber operates on a copy of the parent [[IOLocal]]===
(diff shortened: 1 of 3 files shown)


iRevive:
Hm. I assume the failing test is flaky?

docs/core/io-local.md
```diff
@@ -0,0 +1,64 @@
+---
+id: io-local
+title: IOLocal
+---
+
+`IOLocal` provides a handy way of sharing a context within the fiber runtime.
```
armanbilge: Personally, when I think about `IOLocal` I think less about "sharing". Indeed, when new fibers are spawned they inherit the context, but they do not share it (the way a `Ref` lets you share concurrent state, anyway)—any changes they make, are local to that fiber.

docs/core/io-local.md
```diff
@@ -0,0 +1,64 @@
+---
+id: io-local
+title: IOLocal
+---
+
+`IOLocal` provides a handy way of sharing a context within the fiber runtime.  
+`IO` stores local states as `scala.collection.immutable.Map[IOLocal[_], Any]` under the hood.
```
armanbilge: I feel like this is leaking an implementation detail and I'm not entirely sure if it's helpful for understanding.

docs/core/io-local.md
```diff
@@ -0,0 +1,64 @@
+---
+id: io-local
+title: IOLocal
+---
+
+`IOLocal` provides a handy way of sharing a context within the fiber runtime.  
+`IO` stores local states as `scala.collection.immutable.Map[IOLocal[_], Any]` under the hood.  
+That means, two fibers can never access the same `IOLocal`, they will always be working on their own copies.
```
armanbilge: Ha 😂 I actually regretted this a little after I wrote it in Discord, because I think it is a little confusing. After you create an `IOLocal[A]`, indeed two fibers _can_ use it at the same time, but they will be working on their own copies of the underlying value. Sort of an `IOLocal` is a "key" to some state that is unique to each individual fiber.

docs/core/io-local.md
```diff
@@ -0,0 +1,64 @@
+---
+id: io-local
+title: IOLocal
+---
+
+`IOLocal` provides a handy way of sharing a context within the fiber runtime.  
+`IO` stores local states as `scala.collection.immutable.Map[IOLocal[_], Any]` under the hood.  
+That means, two fibers can never access the same `IOLocal`, they will always be working on their own copies.  
+In some scenarios, `IOLocal` can be considered as an alternative to [[https://typelevel.org/cats-mtl/mtl-classes/local.html cats.mtl.Local]].
```
armanbilge: Or more simply, `Kleisli` / `Reader` which folks may be more familiar with :)

docs/core/io-local.md
```diff
@@ -0,0 +1,64 @@
...
+id: io-local
+title: IOLocal
+---
+
+`IOLocal` provides a handy way of sharing a context within the fiber runtime.  
+`IO` stores local states as `scala.collection.immutable.Map[IOLocal[_], Any]` under the hood.  
+That means, two fibers can never access the same `IOLocal`, they will always be working on their own copies.  
+In some scenarios, `IOLocal` can be considered as an alternative to [[https://typelevel.org/cats-mtl/mtl-classes/local.html cats.mtl.Local]].  
+
+See the following example, of how the `IOLocal` can be used for the propagation of a `TraceId`:
```
armanbilge: Tracing is cool of course, but I wonder if a few demos of some of the special `IOLocal` properties could be a better example of how it works.
e.g.
- "copy locals to children fibers"
- "child local manipulation is invisible to parents"
- "parent local manipulation is invisible to children"

https://github.com/typelevel/cats-effect/blob/7cb291426cadb359c786b1d318a46bced55d4ba3/tests/shared/src/test/scala/cats/effect/IOLocalSpec.scala#L20

armanbilge (commented):
Thanks for spear-heading this!!

docs/core/io-local.md
armanbilge: Actually, `IOLocal` works very much like a `Vault` https://typelevel.org/vault/

docs/core/io-local.md
armanbilge: That is to say, many fibers can use the same key at the same time, but the "box" they open with it will belong only to the fiber and not be accessible by any other fiber.

iRevive: The more I experiment with `IOLocal` the more non-trivial it feels:

## Example 1 (flatMap)

```scala
def inc(local: IOLocal[Int]): IO[Unit] =
  local.update(_ + 1) >> local.get.flatMap(current => IO.println(s"Current: $current"))
  
for {
  local   <- IOLocal(42)
  _       <- inc(local)
  _       <- inc(local)
  current <- local.get
  _       <- IO.println(current)
} yield ()
```

The output is:
```
Current: 43
Current: 44
44
```

## Example 2 (fiber)

```scala
def inc(local: IOLocal[Int]): IO[Unit] =
  local.update(_ + 1) >> local.get.flatMap(current => IO.println(s"Current: $current"))
  
for {
  local   <- IOLocal(42)
  fiber1  <- inc(local).start
  fiber2  <- inc(local).start
  _       <- fiber1.joinWithNever
  _       <- fiber2.joinWithNever
  current <- local.get
  _       <- IO.println(current)
} yield ()
```

The output is:
```
Current: 43
Current: 43
42
```

## 

It's reasonable but slightly nonintuitive. I assume someone can treat `IOLocal` as `Ref` by mistake.

At this point, it makes sense to distinguish `IOLocal` from `Ref` in docs as far as possible (as you suggested in Discord). 

I will improve the example and the doc.

## 

If I understand correctly, a forked fiber (`.start`, in the example above) operates on a duplicated state of the parent fiber.

iRevive: Basically, that's what you've mentioned here https://github.com/typelevel/cats-effect/pull/3046/files#r905350433. 

We need various examples to explain the behavior of `IOLocal`.

core/shared/src/main/scala/cats/effect/IOLocal.scala
```diff
@@ -16,6 +16,144 @@
...
+ * In some scenarios, [[IOLocal]] can be considered as an alternative to [[cats.data.Reader]] or
+ * [[cats.data.Kleisli]].
+ *
+ * [[IOLocal]] should not be treated as [[cats.effect.kernel.Ref Ref]], since the former abides
+ * different laws.
+ *
+ * For example, two forked fibers can never access the same [[IOLocal]], they will always be
+ * working on their own copies.
+ *
+ * ===Same-level operations on [[IOLocal]] are visible to the parent===
```
iRevive: I'm not sure about `same-level` tbh. Is there a better phrasing?

iRevive: Maybe `non-forked`?

armanbilge: I _think_ what you mean is that operations that occur within a fiber, are visible to that fiber? In this situation there is no `.start`, so there is only one fiber, so there is no "parent" or "child", just the one fiber :)

iRevive: > I think what you mean is that operations that occur within a fiber, are visible to that fiber

Exactly, thanks

core/shared/src/main/scala/cats/effect/IOLocal.scala
```diff
@@ -17,19 +17,89 @@
...
+ * [[IOLocal]] provides a handy way of manipulating a context on different scopes.
+ *
+ * In some scenarios, [[IOLocal]] can be considered as an alternative to [[cats.data.Reader]] or
+ * [[cats.data.Kleisli]].
+ *
+ * [[IOLocal]] should not be treated as [[cats.effect.kernel.Ref Ref]], since the former abides
+ * different laws.
+ *
+ * For example, two forked fibers can never access the same [[IOLocal]], they will always be
+ * working on their own copies.
```
armanbilge: See https://github.com/typelevel/cats-effect/pull/3046#discussion_r905348981. Maybe we should say that two fibers will never see each other's modifications to the same `IOLocal`, each fiber will only see its own modifications?

core/shared/src/main/scala/cats/effect/IOLocal.scala
```diff
@@ -16,26 +16,193 @@
...
+ *                │      └────────────┘               └────────────┘
+ * ┌────────────┐─┘                                   ┌────────────┐
+ * │  Fiber A   │                                     │  Fiber A   │
+ * │ (local 42) │────────────────────────────────────►│ (local 42) │
+ * └────────────┘─┐                                   └────────────┘
+ *                │      ┌────────────┐               ┌────────────┐
+ *                │ fork │  Fiber C   │ update(_ + 1) │  Fiber C   │
+ *                └─────►│ (local 42) │──────────────►│ (local 43) │
+ *                       └────────────┘               └────────────┘
+ * }}}
```
armanbilge, line 76: Woah!! 🤩

keynmol: ❤️

djspiewak (approved):
This is so great! Thank you!

PR #3048, opened 2022-06-25
Add `Sync.Type.BlockingOrInterruptible`

armanbilge:
For example this might be useful in https://github.com/typelevel/fs2/pull/2932 to allow the user to configure what kind of blocking they want, without being allowed to pass `Delay`.

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Sync.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Sync.scala
@@ -290,8 +290,9 @@ object Sync {
 
   object Type {
     case object Delay extends Type
-    case object Blocking extends Type
-    case object InterruptibleOnce extends Type
-    case object InterruptibleMany extends Type
+    sealed trait BlockingOrInterruptible extends Type
+    case object Blocking extends BlockingOrInterruptible
+    case object InterruptibleOnce extends BlockingOrInterruptible
+    case object InterruptibleMany extends BlockingOrInterruptible
   }
 }


kernel/shared/src/main/scala/cats/effect/kernel/Sync.scala
```diff
@@ -290,8 +290,9 @@ object Sync {
 
   object Type {
     case object Delay extends Type
-    case object Blocking extends Type
-    case object InterruptibleOnce extends Type
-    case object InterruptibleMany extends Type
+    sealed trait BlockingOrInterruptible extends Type
+    case object Blocking extends BlockingOrInterruptible
+    case object InterruptibleOnce extends BlockingOrInterruptible
+    case object InterruptibleMany extends BlockingOrInterruptible
```
armanbilge, line 296: Something like this could work too.
```suggestion
    sealed abstract class Blocking extends Type
    case object Blocking extends Blocking
    case object InterruptibleOnce extends Blocking
    case object InterruptibleMany extends Blocking
```

djspiewak:
Technically this is also doable without the subtyping. :-) A typeclass which characterizes the precise types which are allowed would give you the same effect without changing the upstream.

armanbilge:
Truthfully that sounds pretty inconvenient. That also sounds like a no 🙃

djspiewak:
Part of what I'm concerned about is the fact that it's pretty bespoke to the case in question. Technically you probably wouldn't want to expose *both* `InterruptibleMany` and `InterruptibleOnce`, since only the latter makes sense in this case. So, just `InterruptibleOnce` and `Blocking`, but then is this better than just a plain `Boolean`?

armanbilge:
Ah, perhaps that's the part I don't understand. Why wouldn't `interruptibleMany` make sense here?

djspiewak:
> Why wouldn't interruptibleMany make sense here?

You know that the action doesn't catch and swallow the `InterruptedException`, so if it can be interrupted, `InterruptibleOnce` will successfully interrupt it.

armanbilge:
Ok, thanks :) it's not entirely clear to me how we can know that about an arbitrary `OutputStream` but I don't grok thread interruption anyway 😆 I do see that this is bespoke to the usecase.

PR #3054, opened 2022-06-27
Fix new blocking worker thread naming change for bincompat

djspiewak:


core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
```diff
@@ -29,12 +29,14 @@ import javax.management.ObjectName
 
 private[unsafe] abstract class IORuntimeCompanionPlatform { this: IORuntime.type =>
 
+  private[this] val DefaultBlockerPrefix = "io-compute-blocker"
```
armanbilge:
```suggestion
  private[this] final val DefaultBlockerPrefix = "io-compute-blocker"
```

Then the compiler inlines it. In theory :)

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
```diff
@@ -104,6 +106,13 @@ private[unsafe] abstract class IORuntimeCompanionPlatform { this: IORuntime.type
       })
   }
 
+  @deprecated("bincompat shim for previous default method overload", "3.3.13")
+  def createDefaultComputeThreadPool(
+      self: () => IORuntime,
+      threads: Int,
+      threadPrefix: String): (WorkStealingThreadPool, () => Unit) =
+    createDefaultComputeThreadPool(self(), threads, threadPrefix, DefaultBlockerPrefix)
```
armanbilge: I'm surprised if this doesn't work, but whatever :)
```suggestion
    createDefaultComputeThreadPool(self(), threads, threadPrefix)
```

armanbilge (dismissed):

armanbilge (approved):

PR #3057, opened 2022-06-27
Scala Native

### LINKED ISSUE #1302: Experiment with scala native

armanbilge:
Closes https://github.com/typelevel/cats-effect/issues/1302.


std/native/src/main/scala/cats/effect/std/SecureRandomCompanionPlatform.scala
```diff
@@ -0,0 +1,56 @@
...
+
+  }
+
+}
+
+@extern
+@nowarn
+private[std] object sysrandom {
+  def getentropy(buf: Ptr[Byte], buflen: CSize): Int = extern
+}
```
armanbilge, line 67: This works on Linux and macOS. I reckon that's plenty good for now. Anyone can implement `SecureRandom` for themselves to suit their requirements.

build.sbt
```diff
@@ -322,12 +345,15 @@ lazy val kernel = crossProject(JSPlatform, JVMPlatform)
   .jsSettings(
     libraryDependencies += "org.scala-js" %%% "scala-js-macrotask-executor" % MacrotaskExecutorVersion % Test
   )
+  .nativeSettings(
+    libraryDependencies += "io.github.cquiroz" %%% "scala-java-time" % "2.4.0"
+  )
```
armanbilge, line 368: I don't think Scala Native has the same artifact size concerns as Scala.js, so I reckon we should ship `java.time` support out-of-the-box.

core/native/src/main/scala/cats/effect/unsafe/QueueExecutorScheduler.scala
```diff
@@ -0,0 +1,98 @@
...
+import scala.collection.mutable.{ListBuffer, PriorityQueue}
+import scala.concurrent.ExecutionContextExecutor
+import scala.concurrent.duration._
+import scala.util.control.NonFatal
+
+import java.time.Instant
+import java.time.temporal.ChronoField
+
+// inspired by https://github.com/scala-native/scala-native/blob/7058cea339168f9456b8c42f1dc4909653e465e8/nativelib/src/main/scala/scala/scalanative/runtime/ExecutionContext.scala
+private[effect] object QueueExecutorScheduler extends ExecutionContextExecutor with Scheduler {
```
armanbilge: This uses `Thread.sleep()` instead of busy-waiting on the global EC when there's nothing to do. But, it's impractical to use for tests (although _possible_) and it's not especially useful as a real runtime either.

armanbilge: That is to say, I don't really think it's worth keeping.

armanbilge: Thanks to some brainstorming from Daniel this has evolved into a much more interesting `PollingExecutorScheduler` abstraction :)

rossabaker:
Is there a way to limit the number of concurrent jobs?  Now that the matrix has grown past 20, this is using all of the org's runners.

armanbilge:
Yeah, that is indeed annoying 😕

GHA does have some controls for concurrency. I think to make it work though we'd have to split the matrix roughly in half (two batches of 10) and enforce that only one such batch may be running at a time across the entire repo.
https://docs.github.com/en/actions/using-jobs/using-concurrency

On the other hand, this will suck when its crickets everywhere else but we're trying to do something here.

armanbilge:
Another alternative I've thought about is running a slimmer CI matrix for PRs and using the full matrix only on merges.

djspiewak:
> Is there a way to limit the number of concurrent jobs? Now that the matrix has grown past 20, this is using all of the org's runners.

I'm honestly not sure it can be helped. Cats Effect is often affected by the *other* direction of this problem, btw, where builds on Cats or Fs2 can eat up runners and cause builds here to take a remarkably long time. Gating concurrency doesn't really solve the problem, which is that we can't get more runners without paying a significant premium (and even then we don't get many).

rossabaker:
This is a better discussion for the maintainers forum.  This is strangling us all.  I'll move it there.

armanbilge:
@keynmol kindly reports that all tests are passing on macOS M1 + 2.13 + Native

core/native/src/main/scala/cats/effect/unsafe/PollingExecutorScheduler.scala
```diff
@@ -0,0 +1,181 @@
...
+        val runnable = todo.poll()
+        try {
+          runnable.run()
+        } catch {
+          case NonFatal(t) =>
+            reportFailure(t)
+        }
+      }
+
+      continue = needsPoll || !executeQueue.isEmpty() || !sleepQueue.isEmpty()
```
armanbilge: This is wrong. `needsPoll` may have been false, but then we run a bunch of tasks that may schedule new stuff for `poll`ing. Indeed, we may have drained the queue. So we could end up not continuing, when in fact we should be `poll`ing.

core/native/src/main/scala/cats/effect/IOCompanionPlatform.scala
```diff
@@ -23,11 +23,12 @@ import java.time.Instant
 private[effect] abstract class IOCompanionPlatform { this: IO.type =>
 
   def blocking[A](thunk: => A): IO[A] =
-    apply(thunk)
+    // do our best to mitigate blocking
+    IO.cede *> apply(thunk) <* IO.cede
```
armanbilge: Downstream libraries are riddled with blocking calls (e.g. DNS resolution in ip4s, file I/O in fs2). Actually even `IO.println` is blocking! Since we don't have a blocking threadpool to run them on, this seems like the best we can do to mitigate their effects.

armanbilge: Actually we can do even better here: this should be `apply(thunk).guarantee(IO.cede)`. We don't want to skip that `cede` even if we errored.

djspiewak:
So this is all in `series/3.3.x` now right? How do we untangle the git gordian knot?

armanbilge:
No, this is not really in `series/3.3.x`. Actually pulling that off would have been too painful because Scala Native requires Scala 3.1+ and the JVM/JS `series/3.3.x` is stuck on Scala 3.0. Hence my "backport" PR https://github.com/typelevel/cats-effect/pull/3138 really started a parallel branch `series/3.3.x-native`. That branch can simply be abandoned.

Yes, sorry, not ideal. But I wasn't prepared to let this die-on-the-vine like my `Env` PR from \*checks watch\* ...

djspiewak:
> Yes, sorry, not ideal. But I wasn't prepared to let this die-on-the-vine like my Env PR from *checks watch* ...

*whistles innocently*

Either way, given that this has now been released, I think it's best if we get it into series/3.x to form a part of 3.4. :-)

armanbilge:
Ok, @djspiewak, let's land this one :) tbh I was quite sad to ship this out in 3.3.14 without your review, so I'm looking forward to that 😇 

Question: do we want to add a macOS CI job? There a couple places where we directly call native APIs (high-precision time and cryptographically strong random numbers). They are all supported on Linux/macOS, but in theory we should be careful.

djspiewak:
> Question: do we want to add a macOS CI job? There a couple places where we directly call native APIs (high-precision time and cryptographically strong random numbers). They are all supported on Linux/macOS, but in theory we should be careful.

I hate to say it, but we should be careful. :-( So that means one more matrix entry.

armanbilge:
Something I haven't done in this PR is setup the test harness so we can run the `IOAppSpec` with Native executables. It's possible by the same trick we use to run the `IOAppSpec` for JS. So just a yak I'd be happy to put off 😆

core/native/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    // An infinite heartbeat to keep main alive.  This is similar to
+    // `IO.never`, except `IO.never` doesn't schedule any tasks and is
+    // insufficient to keep main alive.  The tick is fast enough that
+    // it isn't silently discarded, as longer ticks are, but slow
+    // enough that we don't interrupt often.  1 hour was chosen
+    // empirically.
+    lazy val keepAlive: IO[Nothing] =
+      IO.sleep(1.hour) >> keepAlive
```
djspiewak, line 212: Do we actually need this with Scala Native?

core/native/src/main/scala/cats/effect/IOCompanionPlatform.scala
```diff
@@ -0,0 +1,65 @@
...
+
+import java.time.Instant
+
+private[effect] abstract class IOCompanionPlatform { this: IO.type =>
+
+  private[this] val TypeDelay = Sync.Type.Delay
+
+  def blocking[A](thunk: => A): IO[A] =
+    // do our best to mitigate blocking
+    IO.cede *> apply(thunk).guarantee(IO.cede)
```
djspiewak, line 29: 😱

core/native/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -0,0 +1,244 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+
+import scala.concurrent.CancellationException
+import scala.concurrent.duration._
+
+/**
```
djspiewak, line 22: Can we write some words in here which clarifies 1) the experimental nature of things, and 2) the limitations imposed by this particular `IOApp`? It's fair to call out to EpollCat for now since that's the only usable runtime.

core/native/src/main/scala/cats/effect/tracing/TracingPlatform.scala
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package cats.effect.tracing
+
+import scala.annotation.nowarn
+
+private[tracing] abstract class TracingPlatform { self: Tracing.type =>
+
+  @nowarn("msg=never used")
+  def calculateTracingEvent(key: Any): TracingEvent = null
```
djspiewak, line 24: This will be fun™…

std/native/src/main/scala/cats/effect/std/SecureRandomCompanionPlatform.scala
djspiewak: We definitely need to write up some serious words about the limitations of the current implementation. :-)

tests/native/src/test/scala/cats/effect/DetectPlatform.scala
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+
+trait DetectPlatform {
+  def isWSL: Boolean = System.getProperty("os.version").contains("-WSL")
```
djspiewak, line 20: How are system properties handled on Scala Native?

djspiewak (commented):
Mostly just worried about the lack of documentation and caveats!

core/native/src/main/scala/cats/effect/IOApp.scala
armanbilge: I believe so! The whole point of this is to keep an `IO.never` program alive. In the `IO.never` case the EC queue will be empty. In theory we could just have the main thread go into a permanently parked state in this situation.

But in practice the EC actually has to stop running if its queues are empty so that it can yield to the Scala Native global EC, to make test frameworks work correctly. See also the doc comment on `poll(...)`.

core/native/src/main/scala/cats/effect/IOApp.scala
armanbilge: Yep can write some doc. Besides the `blocking` issue I wouldn't really say `IOApp` is limited. Anything you can do in Scala Native today works fine with this `IOApp`. If you want to do async I/O with JDK APIs, tough, because they are not implemented anyway, but that's not this `IOApp`s fault.

> It's fair to call out to EpollCat for now since that's the only usable runtime.

It's not actually. It's only necessary for `fs2.io.net` and thus ember or skunk etc.

I have a `CurlApp` in [http4s-curl](https://github.com/http4s/http4s-curl) that's perfectly good if the only non-blocking I/O you need are HTTP requests (this is sufficient to implement a Lambda runtime btw). Also that runtime is theoretically capable of polling on arbitrary file descriptors, but I have not exposed that.

I also just wrote an `io_uring`-based runtime in https://github.com/armanbilge/fs2-io_uring/pull/2. But that's still a bit premature :)

Finally, see also https://github.com/typelevel/cats-effect/pull/3153#discussion_r974603033 where we discussed some interesting native applications built with vanilla `IOApp`, including CLIs and an LSP.

tests/native/src/test/scala/cats/effect/DetectPlatform.scala
armanbilge: No idea what this would say for WSL. But I believe it does attempt to populate them as the JVM would.

std/native/src/main/scala/cats/effect/std/SecureRandomCompanionPlatform.scala
armanbilge: Sure. But note that anyone who tries to use this on an unsupported platform will get an error at _linking time_, before runtime. So it's not evil, just annoying.

core/native/src/main/scala/cats/effect/IOApp.scala
djspiewak: Fascinating. It's identical code but actually handled in an entirely different way.

PR #3063, opened 2022-06-29
Added fiber restart functionality to `Supervisor`

djspiewak:
Needs documentation and stuff. This adds support for configurable supervised restarts, similar to how supervisor nets work in Erlang. The idea is basically that you can configure a supervision predicate which optionally restarts a supervised fiber when it completes in a certain way. Due to type constraints, it is impossible to case on the *value* being returned when it is successful, but I think this is fine.

Thoughts welcome.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
@@ -204,7 +204,8 @@ object Dispatcher {
     val (workers, makeFork) =
       mode match {
         case Mode.Parallel =>
-          (Cpus, Supervisor[F](await).map(s => s.supervise(_: F[Unit]).map(_.cancel)))
+          // TODO we have to do this for now because Scala 3 doesn't like it (lampepfl/dotty#15546)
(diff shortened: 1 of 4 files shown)


armanbilge:
So question about this: is it possible to implement this as new method on `Supervisor`?

The bincompat constraint makes it slightly ugly, and I guess the default implementation wouldn't even need to delegate to `supervise` 🤔 unless there's a dumbed down default implementation in terms of `Monad` or something.

But on the other hand, I think the API would be more flexible: a single `Supervisor` could be used for restarting and non-restarting fibers and also `checkRestart` would be able to introspect the `A`.

```scala
def supervisedRestart[A](fa: F[A])(
  checkRestart: Outcome[F, Throwable, A] => Boolean
)(implicit F: Concurrent[F]): F[Fiber[F, Throwable, A]]
```

djspiewak:
> So question about this: is it possible to implement this as new method on Supervisor?

Okay this is very smart and I agree it's much better. It allows users to think about `Supervisor` in terms of its lifecycle scope, while restart semantics are on a per-fiber basis. Still doesn't allow people to inspect the `A` because it is wrapped in `F`, but the types are simpler and we avoid the existential.

PR #3064, opened 2022-06-29
Handle `Uncancelable` and `OnCancel` in `syncStep` interpreter

armanbilge:
`SyncIO` is effectively uncancelable, so we don't need to do anything special in this case.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -869,6 +869,11 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
             }
             .handleErrorWith(t => interpret(f(t)))
 
+        case IO.Uncancelable(body, _) =>
+          interpret(body(new Poll[IO] {
+            def apply[A](ioa: IO[A]): IO[A] = ioa
+          }))
+
         case _ => SyncIO.pure(Left(io))
       }
 
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -874,6 +874,8 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
             def apply[C](ioc: IO[C]): IO[C] = ioc
           }))
 
+        case IO.OnCancel(ioa, _) => interpret(ioa)
+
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -869,6 +869,13 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
             }
             .handleErrorWith(t => interpret(f(t)))
 
+        case IO.Uncancelable(body, _) =>
+          interpret(body(new Poll[IO] {
+            def apply[C](ioc: IO[C]): IO[C] = ioc
+          }))
+
+        case IO.OnCancel(ioa, _) => interpret(ioa)
+
```
djspiewak, line 878: `Sync` extends `MonadCancel`, so we should actually delegate here rather than assuming that the underlying type is uncancelable.

djspiewak (changes requested):

core/shared/src/main/scala/cats/effect/IO.scala
armanbilge: Ouch, good catch.

armanbilge: Wait, no, this is interpreting specifically into `SyncIO`, so there's nothing to do here. We'll have to fix this on merge into `series/3.x`

armanbilge: In fact, our tests catch this mistake on `series/3.x` anyway :)

```
failing seed for async.syncStep identity is ncXLceSQmGLs4bYfy4oHs3x-u4VCJHPY3rB6uQ5UJwO=
[error] x async.syncStep identity
[error]  Falsified after 13 passed tests.
[error]  > ARG_1: 3
[error]  > ARG_1_ORIGINAL: 2147483647
[error]  The seed is ncXLceSQmGLs4bYfy4oHs3x-u4VCJHPY3rB6uQ5UJwO=
```

armanbilge: `series/3.x` version in https://github.com/typelevel/cats-effect/pull/3065.

djspiewak (approved):

PR #3065, opened 2022-06-29
Handle `Uncancelable` and `OnCancel` in `syncStep` interpreter in 3.x

armanbilge:
Forward merge of https://github.com/typelevel/cats-effect/pull/3064.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -869,6 +869,11 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
             }
             .handleErrorWith(t => interpret(f(t)))
 
+        case IO.Uncancelable(body, _) =>
+          interpret(body(new Poll[IO] {
+            def apply[A](ioa: IO[A]): IO[A] = ioa
+          }))
+
         case _ => SyncIO.pure(Left(io))
       }
 
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -1961,13 +1961,16 @@ private object SyncStep {
             }
             .handleErrorWith(t => interpret(f(t), limit - 1))
 
-        case IO.Uncancelable(body, _) =>
+        case IO.Uncancelable(body, _)
+            if G.isInstanceOf[MonadCancel.Uncancelable[G, Throwable] @unchecked] =>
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1961,6 +1961,17 @@ private object SyncStep {
...
+        case IO.Uncancelable(body, _)
+            if G.isInstanceOf[MonadCancel.Uncancelable[G, Throwable] @unchecked] =>
+          val ioa = body(new Poll[IO] {
+            def apply[C](ioc: IO[C]): IO[C] = ioc
+          })
+          interpret(ioa, limit)
+
+        case IO.OnCancel(ioa, _)
+            if G.isInstanceOf[MonadCancel.Uncancelable[G, Throwable] @unchecked] =>
+          interpret(ioa, limit)
```
armanbilge, line 1973: So this is kind of cheating, but I'm not sure if we can do better.

For `OnCancel`, what happens if we need to bail out the `finalizer`? Then we would need to return a `Left(...)` but I don't see how that's possible if we're in the process of being canceled.

I haven't thought as deeply about `Uncancelable`, but the mental gymnastics of implementing `Poll[IO]` in terms of `Poll[G]` were enough to put me off 😂

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1961,6 +1961,17 @@ private object SyncStep {
             }
             .handleErrorWith(t => interpret(f(t), limit - 1))
 
+        case IO.Uncancelable(body, _)
+            if G.isInstanceOf[MonadCancel.Uncancelable[G, Throwable] @unchecked] =>
```
djspiewak, line 1965: Okay, so finally looped around to this. The correct way to do something like this is via `rootCancelScope`:

```suggestion
        case IO.Uncancelable(body, _) if G.rootCancelScope == CancelScope.Uncancelable =>
```

djspiewak (changes requested):

core/shared/src/main/scala/cats/effect/IO.scala
armanbilge: Ah, right! Forgot about this.

djspiewak: I already made the change locally :-)

Issue #3051, opened 2022-06-26
Unable to compile an hello world example with GraalVM native-image on any 3.3.x using `sbt-native-image`


TonioGela:
Hello 👋!

As per the title, I've tried compiling a dead stupid Hello World

```scala
object Main extends IOApp.Simple {
  def run: IO[Unit] = IO.println("Hello cats!")
}
```
using [sbt-native-image](https://github.com/scalameta/sbt-native-image) 0.3.1 with these configurations

```sbt
ThisBuild / organization := "com.example"
ThisBuild / scalaVersion := "2.13.8"

lazy val root = (project in file(".")).enablePlugins(NativeImagePlugin).settings(
  name := "cats-effect-3-quick-start",
  libraryDependencies += "org.typelevel" %% "cats-effect" % "<CE_VERSION>",
  Compile / mainClass := Some("com.example.Main"),
  nativeImageOptions ++= List("-H:+ReportExceptionStackTraces", "--no-fallback")
)
// GraalVM version is graalvm-java11@20.2.0 as per sbt-native-image default

```
and while for `CE_VERSION` in `2.5.5` to `3.2.9` I get a native executable that works flawlessly, for `CE_VERSION`s in `3.3.0` to `3.3.12` I consistently get these errors 👇 

<details>
<summary>Stacktraces</summary>

```
[/Users/toniogela/Desktop/cats-effect-3-quick-start/target/native-image/cats-effect-3-quick-start:95767]    classlist:   3,666.84 ms,  0.96 GB
[/Users/toniogela/Desktop/cats-effect-3-quick-start/target/native-image/cats-effect-3-quick-start:95767]        (cap):   2,111.30 ms,  0.96 GB
[/Users/toniogela/Desktop/cats-effect-3-quick-start/target/native-image/cats-effect-3-quick-start:95767]        setup:   3,271.69 ms,  0.95 GB
[/Users/toniogela/Desktop/cats-effect-3-quick-start/target/native-image/cats-effect-3-quick-start:95767]     (clinit):     243.10 ms,  1.46 GB
[/Users/toniogela/Desktop/cats-effect-3-quick-start/target/native-image/cats-effect-3-quick-start:95767]   (typeflow):   6,919.81 ms,  1.46 GB
[/Users/toniogela/Desktop/cats-effect-3-quick-start/target/native-image/cats-effect-3-quick-start:95767]    (objects):   6,171.18 ms,  1.46 GB
[...]

Issue #2989, opened 2022-05-18
Cats Effect + scala-cli example


armanbilge:
It's annoying that we need this, but this trips everyone up, most recently @TimWSpence. We should add something to the docs demonstrating how to use CE with scala-cli by explicitly creating an `IOApp` in a `.scala` file, since stuff like `unsafeRunSync()` at the top-level of a `.sc` file will just hang.

Daenyth:
Somebody wrote this, idk seems pretty good ;)

https://github.com/armanbilge/http4s-cli

TonioGela:
There's the [FAQ](https://typelevel.org/cats-effect/docs/faq#scala-cli) section of the documentation that contains an example. Is it worth extracting in its own subsection? And maybe adding a section that shows the usage of decline-effect + CE to write a command-line tool. What do you think?

PR #3069, opened 2022-07-02
Native Image documentation example

TonioGela:
Hello! 👋
I added an example of how to use `sbt-native-image` to produce a native-image executable, including some brief information about what was discovered/explained in #3051 (and a link).

<details>
<summary>SPOILER: Opinionated addition</summary>
A little addition to the docs may be the usage of `hyperfine` (a tool written in Rust) to measure the startup time of the native executable. I added it since its report is so much more detailed compared to the one of `time`:

```sh
$ time ./HelloWorld
Hello world
./HelloWorld  0.01s user 0.01s system 99% cpu 0.022 total

$ hyperfine --warmup 30 ./HelloWorld
Benchmark 1: ./HelloWorld
  Time (mean ± σ):      20.4 ms ±   0.5 ms    [User: 11.3 ms, System: 10.0 ms]
  Range (min … max):    19.3 ms …  21.9 ms    110 runs
```
I can replace that piece with the `time` output if we don't want to advertise external tools
</details>

The `scala-cli` section in `faq.md` can IMHO be extracted in its file so that #2989 can be closed and the doc tidied up a bit.

### DIFF
--- a/docs/core/native-image.md
+++ b/docs/core/native-image.md
@@ -53,7 +53,8 @@ object Main extends IOApp.Simple {
 
 The code can be compiled using `sbt nativeImage` and a native-image executable can then
 be found under `target/native-image/cats-effect-3-hello-world`, and executed as any native
-executable with the benefit of a really fast startup time
+executable with the benefit of a really fast startup time ([hyperfine](https://github.com/sharkdp/hyperfine) 
+is a command line benchmarking tool written in Rust)
 
 ```sh
 $ ./target/native-image/cats-effect-3-hello-world
--- a/docs/core/native-image.md
+++ b/docs/core/native-image.md
@@ -66,4 +66,4 @@ Benchmark 1: ./target/native-image/cats-effect-3-hello-world
 ```
 
 Another way to get your cats effect app compiled with native-image is to leverage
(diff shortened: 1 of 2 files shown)


docs/core/native-image.md
```diff
@@ -17,3 +17,53 @@ images seamless, without the need of extra configuration. The only caveat
...
+
+The code can be compiled using `sbt nativeImage` and a native-image executable can then
+be found under `target/native-image/cats-effect-3-hello-world`, and executed as any native
+executable with the benefit of a really fast startup time
+
+```sh
+$ ./target/native-image/cats-effect-3-hello-world
+Hello Cats Effect!
+
+$ hyperfine ./target/native-image/cats-effect-3-hello-world
```
djspiewak, line 63: I think this is completely fine actually but I would link the tool so that people know what this is referencing. :-)

djspiewak (changes requested):

docs/core/native-image.md
TonioGela: Cool, I added a reference to it. How about the whole text? I'm not a native speaker, so sometimes, my phrasing could be confusing. It's not a poem of course, it's technical documentation, but I have to ask.

djspiewak (approved):
Looks good, thank you!

PR #3074, opened 2022-07-04
Resolve race condition in `interruptibleMany` after interruption

### LINKED ISSUE #3073: Application locks up and fiber dumps don't work anymore
My application completely locks up after some time of running fine. Symtopms:
* Nothing gets processed anymore
* Fiber dumps (kill -USR1 <pid>) give an error (see below)
* Thread dump shows no io-compute threads anymore, only the io-scheduler is still there. Also no io-compute-blocker
* Nothing shown on the application log - I have a Resource around everything that logs on termination

The application is a http4s webserver that makes some calls to a legacy library (keycloak-admin) using IO.interruptibleMany. The results are wrapped in streams (actually I do a paged load using streams with fs2.Stream.unfoldChunkEval). Running cats-effect 3.3.13 and fs2 3.2.9

Error from the fiber dump
```
Exception in thread "SIGUSR1 handler" java.lang.NullPointerException: Cannot invoke "cats.effect.unsafe.WeakBag.toSet()" because "this.fiberBag" is null
	at cats.effect.unsafe.WorkerThread.suspendedSnapshot(WorkerThread.scala:219)
	at cats.effect.unsafe.WorkStealingThreadPool.liveFibers(WorkStealingThreadPool.scala:491)
	at cats.effect.unsafe.FiberMonitor.$anonfun$liveFiberSnapshot$2(FiberMonitor.scala:92)
	at cats.effect.unsafe.FiberMonitor.$anonfun$liveFiberSnapshot$2$adapted(FiberMonitor.scala:91)
	at scala.Option.fold(Option.scala:263)
	at cats.effect.unsafe.FiberMonitor.liveFiberSnapshot(FiberMonitor.scala:91)
	at cats.effect.IOApp.$anonfun$main$8(IOApp.scala:260)
	at cats.effect.Signal.lambda$invocationHandlerFromConsumer$0(Signal.java:105)
	at com.sun.proxy.$Proxy3.handle(Unknown Source)
	at jdk.unsupported/sun.misc.Signal$InternalMiscHandler.handle(Signal.java:198)
	at java.base/jdk.internal.misc.Signal$1.run(Signal.java:220)
	at java.base/java.lang.Thread.run(Thread.java:832)
```

djspiewak:
The race condition here is that the busy-loop which spams interruption on the worker thread can continue to run *after* the interruption has been caught and handled. That… actually literally the point. This is problematic though because it means the worker thread will be returned to the pool in an interrupted state, which in turn shuts down the pool. The solution is to busy-wait (we can't hard-block because that would, itself, be interrupted) until the interruption loop is finished, and then finally clear the interrupted bit at the very end.

Fixes #3073

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/IOFiberPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/IOFiberPlatform.scala
@@ -59,30 +59,33 @@ private[effect] abstract class IOFiberPlatform[A] extends AtomicBoolean(false) {
               val result =
                 try {
                   canInterrupt.release()
-                  val back = Right(cur.thunk())
+
+                  val back =
+                    try {
+                      Right(cur.thunk())
+                    } catch {
+                      // this won't suppress the interruption
+                      case NonFatal(t) => Left(t)
+                    }
 
                   // this is why it has to be a semaphore rather than an atomic boolean
                   // this needs to hard-block if we're in the process of being interrupted
                   // once we acquire this lock, we cannot be interrupted
                   canInterrupt.acquire()
-                  manyDone.set(true) // in this case, we weren't interrupted
+
+                  if (many) {
+                    manyDone.set(true) // in this case, we weren't interrupted
+                  }
+
                   back
                 } catch {
                   case _: InterruptedException =>
                     null
-
-                  case NonFatal(t) =>
-                    Left(t)
                 } finally {
                   canInterrupt.tryAcquire()
                   done.set(true)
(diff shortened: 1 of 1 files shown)


djspiewak:
Released as `3.3.13-7-ea6f370`.

I've run this on the reproducer several times and it appears to be fixed!

armanbilge (approved):
Sadly I don't understand everything about how this works. But these changes make sense AFAICT.

PR #3080, opened 2022-07-07
Override `memoize` for `Resource`

### LINKED ISSUE #3079: `Resource[F, A]#memoize` releases right after first use
Scastie link: https://scastie.scala-lang.org/wxfOvT64RDSGkjRfeAqWgA

Memoized resource allocates just before the first use (as expected), but releases right after it neither holding resource until 'memo release' nor allocating anew before second usage:

```console
scala> val memo = Resource.make(IO.println("Allocate"))(_ => IO.println("Release")).memoize

val memo:
  cats.effect.kernel.Resource[cats.effect.IO,
    cats.effect.kernel.Resource[cats.effect.IO, Unit]
  ] = Bind(Bind(Eval(IO(...)),cats.effect.kernel.Resource$$Lambda$11226/0x000000080229f008@5e3f1864),cats.Monad$$Lambda$11228/0x000000080229f9c8@1ce88df5)

scala> memo.use { r =>
     |   for {
     |     _ <- IO.println("First:")
     |     _ <- r.use(_ => IO.println("First done"))
     |     _ <- IO.println("Second:")
     |     _ <- r.use(_ => IO.println("Second done"))
     |   } yield ()
     | }.unsafeRunSync()
First:
Allocate
First done
Release
Second:
Second done
```

Cats-Effect 3.3.13, Scala 2.13.8

armanbilge:
Fixes https://github.com/typelevel/cats-effect/issues/3079#issuecomment-1177844860. Or so I thought 🤔

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -1246,6 +1246,13 @@ abstract private[effect] class ResourceConcurrent[F[_]]
 
   override def both[A, B](fa: Resource[F, A], fb: Resource[F, B]): Resource[F, (A, B)] =
     fa.both(fb)
+
+  override def memoize[A](fa: Resource[F, A]): Resource[F, Resource[F, A]] = {
+    Resource
+      .makeCaseFull[F, F[(A, Resource.ExitCase => F[Unit])]](poll =>
+        poll(F.memoize(fa.allocatedCase))) { (memo, exit) => memo.flatMap(_._2.apply(exit)) }
+      .map(memo => Resource.eval(memo.map(_._1)))
+  }
 }
 
 private[effect] trait ResourceClock[F[_]] extends Clock[Resource[F, *]] {
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -1249,7 +1249,7 @@ abstract private[effect] class ResourceConcurrent[F[_]]
 
   override def memoize[A](fa: Resource[F, A]): Resource[F, Resource[F, A]] = {
     Resource.eval(F.ref(false)).flatMap { allocated =>
-      val fa2 = F.uncancelable(poll => allocated.set(true) *> poll(fa.allocatedCase))
+      val fa2 = F.uncancelable(poll => poll(fa.allocatedCase) <* allocated.set(true))
       Resource
         .makeCaseFull[F, F[(A, Resource.ExitCase => F[Unit])]](poll => poll(F.memoize(fa2)))(
           (memo, exit) => allocated.get.ifM(memo.flatMap(_._2.apply(exit)), F.unit))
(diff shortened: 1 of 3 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -1246,6 +1246,13 @@ abstract private[effect] class ResourceConcurrent[F[_]]
 
   override def both[A, B](fa: Resource[F, A], fb: Resource[F, B]): Resource[F, (A, B)] =
     fa.both(fb)
+
+  override def memoize[A](fa: Resource[F, A]): Resource[F, Resource[F, A]] = {
+    Resource
+      .makeCaseFull[F, F[(A, Resource.ExitCase => F[Unit])]](poll =>
+        poll(F.memoize(fa.allocatedCase))) { (memo, exit) => memo.flatMap(_._2.apply(exit)) }
+      .map(memo => Resource.eval(memo.map(_._1)))
+  }
```
armanbilge: So one issue with this implementation is that it always forces allocation in the `release` step. Actually this is giving me deja vu.

Still, it seems like with more work even that could be solved.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -1246,6 +1246,16 @@ abstract private[effect] class ResourceConcurrent[F[_]]
...
+
+  override def memoize[A](fa: Resource[F, A]): Resource[F, Resource[F, A]] = {
+    Resource.eval(F.ref(false)).flatMap { allocated =>
+      val fa2 = F.uncancelable(poll => poll(fa.allocatedCase) <* allocated.set(true))
+      Resource
+        .makeCaseFull[F, F[(A, Resource.ExitCase => F[Unit])]](poll => poll(F.memoize(fa2)))(
+          (memo, exit) => allocated.get.ifM(memo.flatMap(_._2.apply(exit)), F.unit))
+        .map(memo => Resource.eval(memo.map(_._1)))
+    }
+  }
```
djspiewak, line 1258: I'm relatively certain this isn't enough. :-) We should make sure that all of the `GenConcurrent` tests for `memoize` are applied to this implementation.

djspiewak (commented):
So my concern with this is it makes the semantics moderately incoherent, which in turn may have an impact on abstract code. What I'm not convinced of is that this is necessarily *wrong*, since the `memoize` semantics we inherit are consistent with the way that fiber lifecycles work, but inconsistent with the way that `Resource` lifecycles work. Does that make it wrong? Is the `memoize` implementation upstream assuming something that it shouldn't?

Basically, I feel like we need a bit more of a first-principles argument here and it's eluding me.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
armanbilge: Hmm, why is that? All the existing tests in `MemoizeSpec` appear to work, unless I did something dumb.

armanbilge:
> So my concern with this is it makes the semantics moderately incoherent, which in turn may have an impact on abstract code.

In what sense? Is there an example of a program you can write with only `Concurrent` combinators that behaves differently after this change? If so, it would definitely make me more suspicious. OTOH, if the change is only observable from `use` and other `Resource`-specific combinators, then it seems right to me.

PR #3081, opened 2022-07-07
Try to fix #3075

### LINKED ISSUE #3075: Break up some of the busy-waits in `interruptibleImpl`
The busy waits are necessary, but `while` loops wrapped in `IO.apply` are definitely not ideal. We can do a bit better by threading the looping through the monadic context, which should improve the granularity of tasks on the pool. This is almost certainly a minor issue in practice (if it's an issue at all) since the busy waits should be *very* short.

durban:
This PR tries to fix #3075. It is on top of #3074.

I've left alone two `while` loops:
- One of them is in a `blocking` call, so it seems ok.
- The other one loops while holding the semaphore. It seemed dangerous to break it up.

Hopefully there are no semantic changes ~~(I've moved a `manyDone.set(true)` from a `finally`, which might be a problem. But I didn't understand why it was in a `finally`...)~~.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/IOFiberPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/IOFiberPlatform.scala
@@ -72,6 +72,8 @@ private[effect] abstract class IOFiberPlatform[A] extends AtomicBoolean(false) {
                     null
 
                   case NonFatal(t) =>
+                    canInterrupt.acquire()
+                    manyDone.set(true) // in this case, we weren't interrupted
                     Left(t)
                 } finally {
                   canInterrupt.tryAcquire()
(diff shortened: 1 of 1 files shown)


armanbilge:
I cancelled CI because it was hanging. But this is a problem you've inherited from https://github.com/typelevel/cats-effect/pull/3074 :)

durban:
Ok, thank you. Do we know why that happens?

armanbilge:
I don't, but maybe @djspiewak has a hunch?

durban:
Yeah, it's a real deadlock. Let's see if this fixes it...

armanbilge:
Hey, looking good, nice work! 😃

djspiewak:
Ah, I just got to this one! :-) You found the same bug that I found. It's fixed in the upstream PR now. I *think* that fixes everything? Let's see.

armanbilge:
@durban do you think you could merge the latest https://github.com/typelevel/cats-effect/pull/3074 into your fix for https://github.com/typelevel/cats-effect/issues/3075? Thanks!

durban:
@djspiewak Yes, I think so too.

@armanbilge Done.

djspiewak (approved):

Issue #2536, opened 2021-11-13
Explain what a "FFI" is or use other term?


LLCampos:
The term "FFI" is sprinkled around the website and scaladocs, but is never explained. Googling doesn't help ([1](https://discord.com/channels/632277896739946517/632278585700384799/882933023530684467), [2](https://discord.com/channels/632277896739946517/632278585700384799/904304276833177641)).

I know this sounds like a small thing, but cats-effect is already full of potentially new concepts - introducing a unknown, unnecessary and unexplained new one in the docs might be really frustrating for people on the learning journey. 

Should "FFI" be changed to a more understandable term? Or should it be explained, maybe in the [concepts page](https://typelevel.org/cats-effect/docs/concepts)? 

I would say changing the term is better, at least in the website, since its not a concept you really need to know to learn how to use the library

armanbilge:
> I would say changing the term is better, at least in the website, since its not a concept you really need to know to learn how to use the library

I agree. From what I can tell, "FFI" is not a need-to-know like you say, but rather a quick/familiar analogy that's supposed to help _explain_ Cats Effect. You can see this dating back to the original [blog post](https://typelevel.org/blog/2017/05/02/io-monad-for-cats.html) introducing CE:
> You can think of this sort of like an FFI that converts impure code (like println) into pure code (like putStrLn)

Except, for those of us who _don't_ readily know what an FFI is, we're left out in the cold. I'd vote to minimize use of FFI as the primary analogy.

otto-dev:
I ran into this - googled what FFI stands for while reading the docs, and still don't know to this day.

timshen24:
So what does FFI really stand for?

otto-dev:
By now I know it's **Foreign Function Interface**. However, from what I can see, the term is not used true to definition, or at least used very loosely.

An FFI bridges the barrier between two different programming languages (according to Wikipedia, Python, PHP, Go, stackoverflow etc). E.g. `C` bindings in `Python` or `Go`. In Cats Effect, however, it merely bridges the gap between different API's or paradigms in the same host language.

I think the term should replaced. Maybe just scrap the `Foreign` and spell it out. Eg. `Sync/Blocking/Async/Future/Callback/... Function Interface`.

(Edited for brevity)

PR #3082, opened 2022-07-08
Introduce the FFI acronym before use

benhutchison:
I don't think it's a safe assumption that all readers of Cats Effect docs will automatically know what "FFI" means or implies. FWIW I'm not super enthused by explaining Async in terms of another rather abstract concept, an FFI, but given that's the approach then at minimum we should unpack the acronym so unfamiliar users can get background context.

I work in a corporate environment where I'm constantly shown slides covered in acronyms where I have no idea what they refer to, and it gets tiring having to regularly ask for unpacking. My rule of thumb is, always unpack an acronym upon first use in any document/page.

### DIFF
--- a/docs/typeclasses/async.md
+++ b/docs/typeclasses/async.md
@@ -3,7 +3,7 @@ id: async
 title: Async
 ---
 
-`Async` is the asynchronous FFI for suspending side-effectful operations that
+`Async` is the asynchronous [Foreign Function Interface](https://en.wikipedia.org/wiki/Foreign_function_interface) (FFI) for suspending side-effectful operations that
 are completed elsewhere (often on another threadpool via a future-like API).
 This typeclass allows us to sequence asynchronous operations without stumbling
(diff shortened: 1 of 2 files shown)


armanbilge:
Linking to:
- https://github.com/typelevel/cats-effect/issues/2536

djspiewak:
Agreed on all counts. This is an improvement, but we shouldn't use FFI *in general* as a way of describing this stuff in the docs.

nessus42:
I just stumbled upon this thread because I was confounded with the terminology of calling `Sync` a "Foreign Function Interface" in the documentation, and subsequently ended up here after a bit of Googling.

I've used and understood the term "foreign function interface" for decades, and to me, seeing `Sync` described as an FFI is utterly confusing to me. It's as if I've been thrown into a group of people speaking a different dialect of English that I don't understand.

So, I'd like to chime in with those who agree that FFI is not the right concept to use for describing this stuff in the docs.

PR #3090, opened 2022-07-11
CPU starvation checker

### LINKED ISSUE #2762: Experiment with starvation detecter
Idea shamelessly stolen from Akka. Starting a background fiber in `IOApp` which does something like this would be an interesting thing to explore:

```scala
val checkInterval = 1.second
val initialDelay = 10.seconds
val threshold = 100.millis

val checker = IO.monotonic flatMap { now =>
  IO.sleep(checkInterval) >> IO.monotonic.map(_ - now) flatMap { delta =>
    if (delta >= threshold)
      Console[IO].errorln("[WARNING] you're probably starving")
    else
      IO.unit
  }
}

checker.foreverM
```

Coefficients should be tunable.

The idea here is pretty simple: just try to wake up and run an action once per second and make sure that the drift on this execution is within acceptable bounds. If it takes too long for the runtime to pick up the action, it strongly suggests that starvation is in play. Note that this test will actually get *less* accurate when we redo the timer backend due to the fact that we won't be going through the external queue in the above.

TimWSpence:
Run a background fiber that detects CPU starvation by looking for clock drift when sleeping. Writes a warning to stderr if starvation is detected

It can be configured via the new system properties
- `cats.effect.cpu.starvation.check.interval: FiniteDuration`: how long it should sleep for on each loop
- `cats.effect.cpu.starvation.check.initialDelay: Duration`: how long to wait before starting the check (set to `Duration.Inf` to disable entirely)
- `cats.effect.cpu.starvation.check.threshold: Double`: how much clock drift can be tolerated before warning. The max drift tolerated is `(1 + threshold) * interval`

Resolves https://github.com/typelevel/cats-effect/issues/2762


core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeConfigCompanionPlatform.scala
```diff
@@ -44,11 +44,33 @@ private[unsafe] abstract class IORuntimeConfigCompanionPlatform { this: IORuntim
         .map(Duration(_))
         .getOrElse(DefaultShutdownHookTimeout)
 
+    val cpuStarvationCheckInterval =
+      Try(System.getProperty("cats.effect.cpu.starvation.check.interval"))
+        .map(Duration(_))
+        .flatMap { d => Try(d.asInstanceOf[FiniteDuration]) }
```
armanbilge, line 54: I think you can just do `case fd: FiniteDuration => ...` here, instead of ... this :P

armanbilge: It might help to transform the `Try` to an `Option`.

TimWSpence: haha yup, either of those would be preferable 😂

TimWSpence:
@armanbilge are you still planning to do https://github.com/typelevel/cats-effect/pull/3098? Or should I add another parameter to toggle this on/off (instead of allowing `initialDelay = Duration.Inf`)?

armanbilge:
Yup, feel free to base off https://github.com/typelevel/cats-effect/pull/3098.

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeConfigCompanionPlatform.scala
```diff
@@ -44,11 +44,33 @@ private[unsafe] abstract class IORuntimeConfigCompanionPlatform { this: IORuntim
...
+        .getOrElse(DefaultCpuStarvationCheckInterval)
+
+    val cpuStarvationCheckInitialDelay =
+      Try(System.getProperty("cats.effect.cpu.starvation.check.initial.delay"))
```
djspiewak: I'm not sure I like the paradigm of splitting up words in property names with `.`. :-) Something like `cats.effect.starvation.check.initialDelay` makes a bit more sense in terms of the hierarchy.

djspiewak (changes requested):
Does this *have* to build on top of the `FiniteDuration` => `Duration` PR? That has some subtleties so I wanted to untangle these things ideally.

djspiewak:
Oh additional question: can we do this in a common location so it can be shared across the `IOApp`s? This is even more useful on JavaScript than on the JVM!

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeConfigCompanionPlatform.scala
TimWSpence: sure! And no it doesn't have to be based on top of that at all - I was just trying to avoid an additional parameter to disable it entirely. But I can rebase and do that - would be nice to unblock this!

TimWSpence:
> Oh additional question: can we do this in a common location so it can be shared across the `IOApp`s? This is even more useful on JavaScript than on the JVM!

Yes, good point!

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeConfigCompanionPlatform.scala
armanbilge: https://github.com/typelevel/cats-effect/pull/3098 already merged in so this point is moot :)

TimWSpence: Haha great!!

core/shared/src/main/scala/cats/effect/CpuStarvationCheck.scala
```diff
@@ -0,0 +1,22 @@
...
+private[effect] object CpuStarvationCheck {
+  def run(runtimeConfig: IORuntimeConfig): IO[Unit] =
+    IO.monotonic
+      .flatMap { now =>
+        IO.sleep(runtimeConfig.cpuStarvationCheckInterval) >> IO
+          .monotonic
+          .map(_ - now)
+          .flatMap { delta =>
+            Console[IO]
+              .errorln("[WARNING] you're CPU threadpool is probably starving")
```
armanbilge: 😛 
```suggestion
              .errorln("[WARNING] your CPU threadpool is probably starving")
```

TimWSpence: Already spotted it! 😛

tests/shared/src/main/scala/catseffect/examples.scala
```diff
@@ -107,4 +107,15 @@ package examples {
...
+
+  object CpuStarvation extends IOApp.Simple {
+
+    def fib(n: Long): Long = n match {
+      case 0 => 1
+      case 1 => 1
+      case n => fib(n - 1) + fib(n - 2)
+    }
+
+    val run = IO.delay(fib(1000)).replicateA_(50)
```
djspiewak: Fwiw, `Thread.sleep` also works :-P

core/shared/src/main/scala/cats/effect/CpuStarvationCheck.scala
```diff
@@ -0,0 +1,38 @@
...
+  def run(runtimeConfig: IORuntimeConfig): IO[Unit] =
+    IO.monotonic
+      .flatMap { now =>
+        IO.sleep(runtimeConfig.cpuStarvationCheckInterval) >> IO
+          .monotonic
+          .map(_ - now)
+          .flatMap { delta =>
+            Console[IO]
+              .errorln("[WARNING] your CPU threadpool is probably starving")
+              .whenA(delta > runtimeConfig.cpuStarvationCheckThreshold)
```
djspiewak: Couple thoughts on this…

First, I think we should probably have the starvation threshold be a *multiple* of the interval. In other words, the check threshold should be a `Double` like `0.1` or `0.2` (representing 10% of 20%, respectively). Then the calculation itself would be something like this:

```suggestion
      .flatMap { now =>
        IO.sleep(runtimeConfig.cpuStarvationCheckInterval) >> IO
          .monotonic
          .map(_ - now)
          .flatMap { delta =>
            Console[IO]
              .errorln("[WARNING] Your CPU is probably starving. Consider reducing the granularity of your `delay`s or adding more `cede`s. This may also be a sign that you are unintentionally running blocking I/O operations (such as File or InetAddress) without the `blocking` combinator.")
              .whenA(delta >= runtimeConfig.cpuStarvationCheckInterval.toDouble * (1 + runtimeConfig.cpuStarvationCheckThreshold))
```

My second thought is that we should be more explanatory in the error message. I took a stab at it in the above. Probably needs line wrapping at the very least.

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeConfig.scala
```diff
@@ -93,6 +119,9 @@ object IORuntimeConfig extends IORuntimeConfigCompanionPlatform {
   private[unsafe] def DefaultEnhancedExceptions = true
   private[unsafe] def DefaultTraceBufferSize = 16
   private[unsafe] def DefaultShutdownHookTimeout = Duration.Inf
+  private[unsafe] def DefaultCpuStarvationCheckInterval = 1.second
+  private[unsafe] def DefaultCpuStarvationCheckInitialDelay = 10.millis
```
djspiewak: Is `10.millis` a sufficient initial delay? That's usually well within warmup times. I would think more like `10.seconds` would be better.

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeConfig.scala
```diff
@@ -93,6 +119,9 @@ object IORuntimeConfig extends IORuntimeConfigCompanionPlatform {
...
+  private[unsafe] def DefaultCpuStarvationCheckThreshold = 100.millis
```
djspiewak: If we accept my suggestion from above, this would become `0.1d`

tests/shared/src/main/scala/catseffect/examples.scala
```diff
@@ -61,19 +62,18 @@ package examples {
...
     }
 
     foo()
 
-    // indirect assertion that we *don't* use the custom config
-    override def runtimeConfig = sys.error("boom")
-
     def run(args: List[String]): IO[ExitCode] =
-      IO.pure(ExitCode.Success)
+      IO.println("boom").whenA(!r.eq(runtime)) >> IO.pure(ExitCode.Success)
```
djspiewak: Why did this change?

djspiewak (changes requested):

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeConfigCompanionPlatform.scala
```diff
@@ -44,11 +44,32 @@ private[unsafe] abstract class IORuntimeConfigCompanionPlatform { this: IORuntim
...
+    val cpuStarvationCheckInitialDelay =
+      Try(System.getProperty("cats.effect.cpu.starvation.check.initialDelay"))
+        .map(Duration(_))
+        .getOrElse(DefaultCpuStarvationCheckInitialDelay)
+
+    val cpuStarvationCheckThreshold =
+      Try(System.getProperty("cats.effect.cpu.starvation.check.threshold"))
+        .map(Duration(_))
+        .flatMap { d => Try(d.asInstanceOf[FiniteDuration]) }
+        .getOrElse(DefaultCpuStarvationCheckThreshold)
```
armanbilge: We need to add these for JS and Native too 😅

core/shared/src/main/scala/cats/effect/CpuStarvationCheck.scala
TimWSpence: Yep, thank you!

tests/shared/src/main/scala/catseffect/examples.scala
TimWSpence: haha good point!

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeConfig.scala
TimWSpence: Yeah this was an entirely arbitrary value I stuck in whilst getting things to compile 😅 `10.seconds` sounds reasonable 👍

tests/shared/src/main/scala/catseffect/examples.scala
TimWSpence: Sorry, I hadn't finished getting this to work properly. But the problem is that we are now using the config for something other than instantiating the runtime so the way the test previously worked (`override def runtimeConfig = sys.error("boom")`) has to change

TimWSpence:
Apologies @djspiewak and @armanbilge for being so slow on this. Will try to get it over the line this week

core/js/src/main/scala/cats/effect/unsafe/IORuntimeConfigCompanionPlatform.scala
```diff
@@ -53,12 +53,35 @@ private[unsafe] abstract class IORuntimeConfigCompanionPlatform { this: IORuntim
       .flatMap(x => Try(x.toBoolean).toOption)
       .getOrElse(DefaultReportUnhandledFiberErrors)
 
+    val cpuStarvationCheckInterval =
+      process
+        .env("CATS_EFFECT_CPU_STARVATION_CHECK_INTERNAL")
```
djspiewak: Typo!

```suggestion
        .env("CATS_EFFECT_CPU_STARVATION_CHECK_INTERVAL")
```

core/js/src/main/scala/cats/effect/unsafe/IORuntimeConfigCompanionPlatform.scala
```diff
@@ -53,12 +53,35 @@ private[unsafe] abstract class IORuntimeConfigCompanionPlatform { this: IORuntim
...
+        .map(Duration(_))
+        .flatMap { d => Try(d.asInstanceOf[FiniteDuration]).toOption }
+        .getOrElse(DefaultCpuStarvationCheckInterval)
+
+    val cpuStarvationCheckInitialDelay =
+      process
+        .env("CATS_EFFECT_CPU_STARVATION_CHECK_INITIALDELAY")
```
djspiewak:
```suggestion
        .env("CATS_EFFECT_CPU_STARVATION_CHECK_INITIAL_DELAY")
```

core/native/src/main/scala/cats/effect/unsafe/IORuntimeConfigCompanionPlatform.scala
```diff
@@ -50,12 +50,32 @@ private[unsafe] abstract class IORuntimeConfigCompanionPlatform { this: IORuntim
         .flatMap(x => Try(x.toBoolean).toOption)
         .getOrElse(DefaultReportUnhandledFiberErrors)
 
+    val cpuStarvationCheckInterval =
+      Option(System.getenv("CATS_EFFECT_CPU_STARVATION_CHECK_INTERNAL"))
```
djspiewak:
```suggestion
      Option(System.getenv("CATS_EFFECT_CPU_STARVATION_CHECK_INTERVAL"))
```

core/native/src/main/scala/cats/effect/unsafe/IORuntimeConfigCompanionPlatform.scala
```diff
@@ -50,12 +50,32 @@ private[unsafe] abstract class IORuntimeConfigCompanionPlatform { this: IORuntim
...
+        .map(Duration(_))
+        .flatMap { d => Try(d.asInstanceOf[FiniteDuration]).toOption }
+        .getOrElse(DefaultCpuStarvationCheckInterval)
+
+    val cpuStarvationCheckInitialDelay =
+      Option(System.getenv("CATS_EFFECT_CPU_STARVATION_CHECK_INITIALDELAY"))
```
djspiewak:
```suggestion
      Option(System.getenv("CATS_EFFECT_CPU_STARVATION_CHECK_INITIAL_DELAY"))
```

This makes me feel like we should find a way to share some of this between JS and Scala Native. Probably not for this PR but something to think about for a follow-up.

core/shared/src/main/scala/cats/effect/CpuStarvationCheck.scala
```diff
@@ -0,0 +1,47 @@
...
+      }
+
+    IO.monotonic.flatMap(go(_)).delayBy(runtimeConfig.cpuStarvationCheckInitialDelay)
+  }
+
+  private[this] val warning =
+    """[WARNING] Your CPU is probably starving. Consider increasing the granularity
+      |of your `delay`s or adding more `cede`s. This may also be a sign that you are
+      |unintentionally running blocking I/O operations (such as File or InetAddress)
+      |without the `blocking` combinator.""".stripMargin
```
djspiewak:
```suggestion
    """[WARNING] Your CPU is probably starving. Consider increasing the granularity
      |of your delays or adding more cedes. This may also be a sign that you are
      |unintentionally running blocking I/O operations (such as File or InetAddress)
      |without the blocking combinator.""".stripMargin
```

Had a look at `TestControl` and it seems that we skip the backticks usually.

djspiewak (changes requested):
Just need a couple adjustments and then I think we're good to go. I changed the base so we're now pointing to 3.4.x

core/js/src/main/scala/cats/effect/unsafe/IORuntimeConfigCompanionPlatform.scala
TimWSpence: ouch! Sorry about that!!

TimWSpence:
> Just need a couple adjustments and then I think we're good to go. I changed the base so we're now pointing to 3.4.x

Thank you!!

core/native/src/main/scala/cats/effect/unsafe/IORuntimeConfigCompanionPlatform.scala
TimWSpence: Yes, would be nice!

djspiewak (approved):
As another follow-up, we should document the new envars and properties.

PR #3091, opened 2022-07-11
Add `Async#asyncCheckAttempt` for #3087

### LINKED ISSUE #3087: Consider an `Async` utility to encapsulate double-check registration
A relatively common pattern in lock-free structures is something like the following:

1. Check a condition. If true, then return a value
2. Register a callback for when the condition becomes true
3. Check condition again. If true, unregister the callback and return the value
4. If false, asynchronously suspend

The problem with doing this using `async`/`async_` is there are two race conditions here. First, at step 3, the callback may be invoked at the same moment you try to short-circuit. In order to resolve this race, you must consider whether your callback condition is idempotent. If it is *not*, you have a problem and may lose data. (note that the callback will drop subsequent invocations, so it is idempotent *in a sense*, but subsequent data passed to the callback will be lost) This is not really something that can be directly resolved in this type of structure though, since the callback does not return a `Boolean`. If you need to artificially construct idempotency in this scenario, it is recommended that your registration itself contain an `AtomicBoolean` which can be subjected to a CAS prior to the invocation of the callback.

The second race condition, which is even subtler, is the fact that the conventional way of achieving this when the condition at step 3 is true (manually invoking the callback within the `async` block) creates a "gap" in cancelability. If the `async` registration is canceled somewhere in these steps, and the condition is true in step 3, that cancelation will be observed at the end of the registration block. This cancelation observation is quite intuitive in the "condition false" case, where we asynchronously suspend (step 4), and will generally be handled correctly (often by unregistering the callback). However, in the case where we manually completed and we already have a value, this type of thing can create deadlocks where none are expected.

The solution to this, at present, is to simply use `cont` directly. Since you then have control over the `get` action (the async suspension), you can branch around it and only suspend if the condition in step 3 is false. If the condition in step 3 is true, then you don't need to sequence `get` at all, which means you never suspend the fiber, and in turn also means that your cancelation properties remain air-tight. I didn't really think too much about this pattern until working on the `Queue` reworks, where it came up *constantly*, and now that I know to look for it, I see it all over the existing Cats Effect codebase (e.g. take a look at `Dispatcher`'s use of `async_`). I think we should codify this pattern into something more usable than `cont` (though still built on it, obviously).

I'm thinking of a function signature something like the following:

```scala
// come up with a good name
def bippy[A](k: (Either[Throwable, A] => Unit) => F[Either[Option[F[Unit]], A]]): F[A]
```

In a sense, this generalizes the existing `async` a bit. If you produce a `Right`, the `get` won't be sequenced and the result is simply immediately returned. If you produce a `Left` then it behaves exactly like `async`. In fact, `async` itself should be reimplemented in terms of this slight generalization.

seigert:
Implementation of proposed utility method in `Async` for double-check registration. Also implemented in `IO` itself and some tests.

I've chosen `asyncPoll` as a name as method itself reminds me about Rust's [Future `poll` method](https://doc.rust-lang.org/std/future/trait.Future.html#tymethod.poll) with also returns computed result or provides waker to schedule another poll sometime later.


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1064,6 +1064,76 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
...
+            case Right(a) => G.pure(a)
+            case Left(Some(fin)) => G.onCancel(poll(get), lift(fin))
+            case Left(None) => poll(get)
+          }
+        }
+      }
+    }
+
+    IOCont(body, Tracing.calculateTracingEvent(k))
+  }
```
armanbilge: Sorry if I missed it—do we get an optimization be re-implementing this for `IO`? Otherwise probably we can delegate to the `Async` implementation?

seigert: I've added it because a) only tests for `async` I've found were in `IOSpec.scala` and b) IO has `async & async_` reimplemented too (and only visible bonus is Tracking + IOCont). If that not a concern, I could delete it.

Maybe some law should be added to AsyncLaws instead? Like `(F.asyncPool[A](k => F.pure(Right(a))) <* F.unit) <-> F.pure(a))`?

seigert:
Also, should docs for `cont` be updated about this method?  "If you are an implementor, and you have `async` (or `asyncPool`), `Async.defaultCont` provides an implementation of `cont` in terms of `async/asyncPool`."

core/shared/src/main/scala/cats/effect/IO.scala
armanbilge: Ah right, thanks, I missed the `Tracing`. Actually my main concern was that if we override the implementation in `IO`, then the generic implementation in `Async` is not being tested anywhere, since as you point out the only are were in `IOSpec`.

seigert: Well, it was this way before -- `Async.async & Async.async_` a re-implemented in `IO` almost word by word, so the status quo stays the same. :)

Anyway, I've added three more laws to `AsyncLaws` about `asyncPoll <-> (async, pure)` (hope they make some sense) and added some synthetic test for some 'basic' `Async` variant based off `IO` which reuses all method implementations in `Async` itself.

seigert:
@armanbilge, do you have any insight why check may randomly flick? Yesterday it was IoAppSpec, not it is WorkerThreadNameSpec -- both should be unaffected by my changes. :(

armanbilge:
Darn, I just restarted CI for you. `IOAppSpec` is a well known flake. `WorkerThreadNameSpec` was only recently added, but I guess it's a flaky one too :(

tests/shared/src/test/scala/cats/effect/kernel/AsyncSpec.scala
```diff
@@ -0,0 +1,149 @@
...
+import cats.laws.discipline.arbitrary._
+
+import org.scalacheck.{Arbitrary, Cogen, Prop}
+import org.scalacheck.Arbitrary.arbitrary
+import org.typelevel.discipline.specs2.mutable.Discipline
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration._
+
+class AsyncSpec extends BaseSpec with Discipline {
```
djspiewak, line 32: Why did this have to be pulled out?

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1064,6 +1064,75 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
...
+   *       }
+   *     }
+   *   }
+   * }
+   * }}}
+   *
+   * @see
+   *   [[async]] for a simplified variant without an option for immediate result
+   */
+  def asyncPoll[A](
```
djspiewak: The one problem I see with the name is we *already* use the word `Poll` for something quite different! This might be worth bikeshedding a bit to avoid user confusion.

djspiewak (commented):
This is great! Thank you for taking this on, and I'm sorry it took me so long to review. Left some comments.

tests/shared/src/test/scala/cats/effect/kernel/AsyncSpec.scala
seigert: As @armanbilge mentioned above, `Async` version of `async_` (and new generic version of `async` itself) was not tested anywhere, as IOSpec.scala only tests `Async[IO]` instance that, in turn, uses redefined `IO.async & IO.async_`.

They are almost literal copies of one another, but still -- as there is two methods in `Async` that both should provide backward compatible generic implementations via `asyncPoll` I've decided to pull out test spec for `Async` itself.

The problem is that Async does not have any instances except for `IO`, and `IO` overrides most methods, so spec provides another `Async` instance that only defines necessary abstract methods.

core/shared/src/main/scala/cats/effect/IO.scala
seigert: I don't think my English is good enough to provide a meaningful name, but after some thoughts there are my suggestions:

1. `asyncAttempt/tryAsync` -- as in 'we'll try to schedule async operation but if results are already available, then so be it'. Both with 'error handling' connotations, sadly.
2. `resolve/resolveEventually` -- as in 'computed value will be resolved one way or another, right now or in future';
3. `registerAsyncCallback` -- intentionally enterprise-like name to emphasis that this is a low-level method like `cont`.

tests/shared/src/test/scala/cats/effect/kernel/AsyncSpec.scala
djspiewak: Oh gotcha. Nice catch!

core/shared/src/main/scala/cats/effect/IO.scala
djspiewak: I'll give this some thought and report back in a day or two! I actually rather like `asyncAttempt`, but you're right it does have that negative "error" context.

seigert: @djspiewak, did you had time to think about a name?

djspiewak: Ah, I haven't! Thank you for the reminder. I'll get to this, I apologize.

djspiewak: I think `asyncCheckAttempt` is probably the best I can come up with. I agree with you that we don't necessarily *want* it to be a really concise name.

```suggestion
  def asyncCheckAttempt[A](
```

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1064,6 +1064,75 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
...
+   *     dropped);
+   *   - left side `Option[IO[Unit]] `is an optional finalizer to be run in the event that the
+   *     fiber running `async(k)` is canceled.
+   *
+   * For example, here is a simplified version of `IO.fromCompletableFuture`:
+   *
+   * {{{
+   * def fromCompletableFuture[A](fut: IO[CompletableFuture[A]]): IO[A] = {
+   *   fut.flatMap { cf =>
+   *     IO.asyncPoll { cb =>
```
```suggestion
   *     IO.asyncCheckAttempt { cb =>
```

kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
```diff
@@ -62,22 +62,28 @@ import java.util.concurrent.atomic.AtomicReference
...
+   *   - right side `A` signals immediately available result;
+   *   - left side `Option[F[Unit]]` is an optional finalizer to be run in the event that the
+   *     fiber running `async(k)` is canceled.
+   *
+   * In case where `k` returns left side the execution of `async(k)` is semantically blocked
+   * until the callback is invoked. In case of a right side callback invocation is silently
+   * dropped.
    */
-  def async[A](k: (Either[Throwable, A] => Unit) => F[Option[F[Unit]]]): F[A] = {
+  def asyncPoll[A](k: (Either[Throwable, A] => Unit) => F[Either[Option[F[Unit]], A]]): F[A] = {
```
djspiewak:
```suggestion
  def asyncCheckAttempt[A](k: (Either[Throwable, A] => Unit) => F[Either[Option[F[Unit]], A]]): F[A] = {
```

kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
```diff
@@ -62,22 +62,28 @@ import java.util.concurrent.atomic.AtomicReference
 trait Async[F[_]] extends AsyncPlatform[F] with Sync[F] with Temporal[F] {
 
   /**
-   * The asynchronous FFI.
+   * The asynchronous FFI with option for immediate result.
```
djspiewak: Can we unify this with the `IO.asyncCheckAttempt` scaladoc?

core/shared/src/main/scala/cats/effect/IO.scala
seigert: Done.

kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
seigert: Not sure that I understood you correctly, but I've tried to sync `IO` and `Async` `async*` method descriptions with exception of 'real-world examples'.

djspiewak (approved):
Thank you!! This has been a very much appreciated effort.

Issue #3094, opened 2022-07-12
Error Building Native Image on 3.3.13


ChristopherDavenport:
Building results in the following error.

```
Error: Error parsing dynamic proxy configuration in jar:file:/Users/chrisdavenport/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/typelevel/cats-effect_2.13/3.3.13/cats-effect_2.13-3.3.13.jar!/META-INF/native-image/org.typelevel/cats-effect/sun-misc-signal-proxy-config.json:
second level of document must be a lists of objects
Verify that the configuration matches the schema described in the -H:PrintFlags=+ output for option DynamicProxyConfigurationResources.
```

armanbilge:
Chris didn't include it here but this is the config he pasted on Discord:
```scala
nativeImageJvm := "graalvm-ce-java11",
nativeImageVersion := "21.1.0",
nativeImageJvmIndex := "jabba"
```

armanbilge:
Ok, confirmed to work in 3.3.12. So it's a regression due to https://github.com/typelevel/cats-effect/pull/3053. Probably we mucked up the format. Annoying but we probably want to build an image in CI to avoid these sorts of errors...

ChristopherDavenport:
I rolled back to 3.3.12 and it did not hit this issue, meaning the likely culprit is https://github.com/typelevel/cats-effect/pull/3053

vasilmkd:
Damn it, it's literally written as in the website. 😡

vasilmkd:
I was able to build a native image using `cats-effect` `3.3.14` using GraalVM 22.1.0. The config is correct. It could be the case that it only works with latest GraalVM. Sorry about that. I guess #3053 should be reverted.

djspiewak:
Can GraalVM 22.1.0 build a native image using 3.3.12? I would rather bias in favor of more recent versions.

arixmkii:
There is another issue with native image. Reported in:
https://github.com/oracle/graal/issues/4530
https://github.com/lampepfl/dotty/issues/15091

This is fixed in Scala 3.2.0-RC1, but I have to rebuild CE libraries locally to have it working.

At least it hints, that if you want to test native image - you should test both Scala 2 and Scala 3 separately, which looks really unfortunate.

Should I open a separate issue to track this?

armanbilge:
[...]

Issue #3030, opened 2022-06-14
GraalVM native image errors out with `TypeNotPresentException` on Scala 3


zetashift:
I'm trying to run a small `http4s` example with `scala-cli` and for fun I was trying to package it through GraalVM, but I'm running into an error into (what looks like) a cats-effect error. 
### Steps to reproduce

If one has `scala-cli` installed, the following can be run to reproduce the error:

```
scala-cli package --native-image https://gist.github.com/zetashift/bdefce3d692e6a626e6bf2ecaf218f48 -f -- --no-fallback
```

Code is located here: https://gist.github.com/zetashift/bdefce3d692e6a626e6bf2ecaf218f48

### Error

Short: 

```
Fatal error: java.lang.TypeNotPresentException: Type cats.effect.kernel.Par$ParallelFImpl.T not present
at java.base/sun.reflect.generics.factory.CoreReflectionFactory.makeNamedType(CoreReflectionFactory.java:117)
```

The long version: http://ix.io/40aH

Still having a lot of fun with Cats Effect and `scala-cli` so thank you all for this awesome library!

armanbilge:
Here's a minimization. Switching to Scala 2.13.8 makes the problem go away, so it seems to be specific to the Scala 3 encoding. If further minimized might even be a dotty issue.

```scala
//> using scala "3.1.2"
//> using lib "org.typelevel::cats-effect:3.3.12"

import cats._
import cats.effect._
import cats.syntax.all._

object MyApp extends IOApp.Simple {

  def run = foo[IO]

  def foo[F[_]: Async: Parallel] = Async[F].unit &> Async[F].unit

}
```

rmgk:
Note, I had a similar issue involving https://ben.kirw.in/decline/ and cats.Applicative.F.
I do not have a minimization, but for me the issue only occurred on native image 22.1.0 (not on 22.0.0.2) so I considered it a regression in native image (which are pretty common).
All on Scala 3.1.2

armanbilge:
I still wonder if this can be classified as a regression in Scala 3 relative to Scala 2. Specifically that the backend is emitting some bytecode Graal doesn't like. There are many backend fixes in 2 that need to be ported to 3 so this could be one of them.

armanbilge:
So there was a very similar issue reported and fixed in dotty:
- https://github.com/lampepfl/dotty/issues/15091
- https://github.com/lampepfl/dotty/pull/15093

[...]

PR #3097, opened 2022-07-13
Test GraalVM Native Image in CI

armanbilge:
Debatable whether this fixes https://github.com/typelevel/cats-effect/issues/3094, but it covers our ass a bit 😃


build.sbt
```diff
@@ -162,6 +167,12 @@ ThisBuild / githubWorkflowBuild := Seq(
     name = Some("Test Example JavaScript App Using Node"),
     cond = Some(s"matrix.ci == 'ciJS' && matrix.os == '$PrimaryOS'")
   ),
+  WorkflowStep.Sbt(
+    List("exampleJVM/nativeImage"),
+    name = Some("Test GraalVM Native Image"),
+    cond = Some(
+      s"matrix.scala == '$Scala213' && matrix.java == '${GraalVM.render}' && matrix.os == '$PrimaryOS'")
+  ),
```
armanbilge: @vasilmkd is it safe to assume that if something goes wrong, the `nativeImage` task will fail?

build.sbt
```diff
@@ -162,6 +167,12 @@ ThisBuild / githubWorkflowBuild := Seq(
     name = Some("Test Example JavaScript App Using Node"),
     cond = Some(s"matrix.ci == 'ciJS' && matrix.os == '$PrimaryOS'")
   ),
+  WorkflowStep.Sbt(
+    List("exampleJVM/nativeImage"),
+    name = Some("Test GraalVM Native Image"),
+    cond = Some(
+      s"matrix.scala == '$Scala213' && matrix.java == '${GraalVM.render}' && matrix.os == '$PrimaryOS'")
```
arixmkii, line 174: It might not be enough to only do this for Scala213. Details in this comment https://github.com/typelevel/cats-effect/issues/3094#issuecomment-1183250990

armanbilge: Yep, definitely agree about that. Although I think it may require us to add another job to matrix...

build.sbt
djspiewak: Should we run the resultant image as well?

build.sbt
djspiewak: I think we have to add another job. :-( Sucks.

build.sbt
armanbilge: Ideally, and I wanted to, but it's a bit annoying. I'm bootstrapping using the existing JVM example app which is designed to end with `ExitCode(2)` or something. We could make another, more innocent example app, to use for the Graal test?

build.sbt
armanbilge: This will have to wait for Scala 3.2 anyway.

djspiewak: Okies, so I think we merge this now then and add a job for 3.2 once we update.

build.sbt
djspiewak: Can either make another innocent test (probably ideal), or the test can use bash's `!`, but then I'm not sure what it's testing.

graalvm-example/src/main/scala/cats/effect/example/Example.scala
```diff
@@ -0,0 +1,30 @@
...
+import cats.effect._
+import cats.syntax.all._
+
+object GraalVMExample extends IOApp.Simple {
+
+  def run = foo[IO] *> IO.println("grrrraal")
+
+  def foo[F[_]: Async: Parallel] = Async[F].unit &> Async[F].unit
+
+}
```
armanbilge, line 30: Adapted from my minimization in https://github.com/typelevel/cats-effect/issues/3030#issuecomment-1155643750. This currently fails on Scala 3.

armanbilge:
Let's leave https://github.com/typelevel/cats-effect/issues/3030 open until we add Scala 3.2.0 + NativeImage to CI.

djspiewak (approved):

PR #3098, opened 2022-07-13
Accept `Duration` for all time-based combinators

armanbilge:
Follow-up to https://github.com/typelevel/cats-effect/pull/2954.

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
@@ -287,14 +287,14 @@ object GenTemporal {
         instantiateGenTemporalForWriterT(temporal)
     }
 
-  private[effect] def handleDuration[A](duration: Duration, ifInfinite: => A)(
+  private[effect] def handleDuration[A](duration: Duration, ifInf: => A)(
       ifFinite: FiniteDuration => A): A =
     duration match {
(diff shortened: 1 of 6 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
```diff
@@ -278,13 +287,15 @@ object GenTemporal {
...
+      ifFinite: FiniteDuration => A): A =
     duration match {
-      case _: Duration.Infinite => fa
-      case finite: FiniteDuration => f(finite)
+      case fd: FiniteDuration => ifFinite(fd)
+      case Duration.Inf => ifInf
+      case d =>
+        throw new IllegalArgumentException(
+          s"Duration must be either a `FiniteDuration` or `Duration.Inf`, but got: $d")
     }
```
armanbilge, line 298: While we're here, should we also do something about negative `FiniteDuration`s?

armanbilge:
Something I'm leery about is that these changes introduce a lot of branches that aren't currently tested by our laws which I think work in terms of `FiniteDuration`. We should change the laws as well, and if not then at least add specific test coverage for this.

djspiewak (dismissed):
This seems good if we can get it to build. :-)

armanbilge:
> This seems good if we can get it to build. :-)

Heh, I was waiting for a positive sign before continuing :) I'm on it!

PR #3102, opened 2022-07-15
Remove the `IO.defer` from `IOFiber#cancel`

vasilmkd:
I know it doesn't win any awards for writing principled `IO` code, but if we can do a bit less work in the run loop, I think that's good.

The change slightly improves race performance (through doing less work on `cancel`). The `race` benchmark uses `IO.race` itself, and the `racePair` benchmark uses `racePair` followed by a manual `cancel` of the loser.

`series/3.x`:
```
Benchmark                (size)   Mode  Cnt      Score      Error  Units
AsyncBenchmark.race         100  thrpt   10  20190.843 ± 1213.201  ops/s
AsyncBenchmark.racePair     100  thrpt   10  39544.775 ± 1526.123  ops/s
```

`This PR`:
```
Benchmark                (size)   Mode  Cnt      Score      Error  Units
AsyncBenchmark.race         100  thrpt   10  21199.357 ±  746.625  ops/s
AsyncBenchmark.racePair     100  thrpt   10  40978.626 ± 1115.885  ops/s
```

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -132,39 +132,37 @@ private final class IOFiber[A](
 
   /* this is swapped for an `IO.unit` when we complete */
   private[this] var _cancel: IO[Unit] = IO uncancelable { _ =>
-    IO defer {
-      canceled = true
+    canceled = true
 
-      // println(s"${name}: attempting cancelation")
+    // println(s"${name}: attempting cancelation")
 
-      /* check to see if the target fiber is suspended */
-      if (resume()) {
-        /* ...it was! was it masked? */
-        if (isUnmasked()) {
-          /* ...nope! take over the target fiber's runloop and run the finalizers */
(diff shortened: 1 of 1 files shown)


armanbilge:
Seems like we can target to 3.3.x? :)

djspiewak:
@vasilmkd Doesn't this preclude optimizations in `uncancelable`? Then again, this is all internal to `IOFiber`, so maybe it doesn't matter.

vasilmkd:
What optimizations in `uncancelable` exactly? We've only ever done stuff to the `Poll[IO]` instance that is constructed there, but that's just an argument to the `body` function.

Essentially, this change only runs the `canceled = true` part one run loop iteration earlier. Instead of being suspended in the thunk in `IO.defer`, it's suspended directly in the `body: Poll[IO] => IO[A]`, which shouldn't be a bad thing.

Again, this won't win any awards for writing principled `IO` code, but then people shouldn't be looking into the run loop as how things need to be done. 😆

PR #3103, opened 2022-07-16
Experiment: subclass `IO.Delay`, skip `thunk` allocation

armanbilge:
This is an experiment that uses `inline` to transparently subclass `IO.Delay`. Inspired by a Discord question.

The results are really good, since effectively it reduces the overhead of `IO.Delay` to match the overhead of `IO.Pure`.

#### This PR

```
[info] DeepBindBenchmark.async   10000  thrpt   20  1150.461 ±  33.122  ops/s
[info] DeepBindBenchmark.delay   10000  thrpt   20  4636.526 ±  95.142  ops/s
[info] DeepBindBenchmark.pure    10000  thrpt   20  4578.873 ± 182.701  ops/s
```

#### `series/3.x`

```
[info] DeepBindBenchmark.async   10000  thrpt   20  1056.442 ±  27.546  ops/s
[info] DeepBindBenchmark.delay   10000  thrpt   20  3374.185 ± 422.683  ops/s
[info] DeepBindBenchmark.pure    10000  thrpt   20  4642.148 ±  87.353  ops/s
```

Caveats:
1. Scala 3 only, obviously, and JVM only (see caveat 4).
2. I don't see how this can work with typeclasses 😕 
3. Sub-classes of `Delay` would now be living in downstream bytecode. So that invokes bincompatibility constraints.
4. Bytecode size, since now every invocation of `Delay` generates its own class. If we make `Delay` a SAM by changing `val event` to `def event`, then that would curtail this. However, JS doesn't have SAMs and is extremely sensitive to artifact size, so we wouldn't want to do this there.

We could also use a similar trick for `IO.FlatMap` I think.

Sadly I didn't fully think through caveat (2) before embarking on this experiment (conveniently fooled myself into thinking we can just slap an `inline` modifier on `Sync#delay`). So unless we care specifically about optimizing for concrete `IO`, this is a lot of complexity for no gain.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -1040,7 +1040,7 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
    *
    * Alias for [[IO.delay]].
    */
-  def apply[A](thunk: => A): IO[A] = delay(thunk)
+  inline def apply[A](inline thunk: => A): IO[A] = delay(thunk)
 
   /**
    * Suspends a synchronous side effect in `IO`. Use [[IO.delay]] if your side effect is not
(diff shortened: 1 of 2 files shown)


armanbilge:
I think the trade-offs involved (no performance improvement for tagless final, no tracing and/or binary bloat, no JS) point to this happening outside of Cats Effect. Basically if someone wants to implement a performant `IO` monad for the JVM.

PR #3110, opened 2022-07-18
Handle ClosedByInterruptException properly in interruptible

### LINKED ISSUE #3076: Handle `ClosedByInterruptException` in `interruptible`
Minor loophole, but this is something we don't handle correctly today. The exception should be treated the same as an `InterruptedException`, if I understand correctly.

durban:
Fixes #3076.


djspiewak (approved):
TIL!

PR #3114, opened 2022-07-21
Remove "asynchronous" from ref doc

bplommer:
Follow-up to https://github.com/typelevel/cats-effect/pull/1890


armanbilge (approved):
Thanks!

Issue #3119, opened 2022-07-31
The "Only Once" sample code in the Deferred documentation is outdated


floating-cat:
Page: https://typelevel.org/cats-effect/docs/std/deferred#only-once
Parts:
> Two processes will try to complete at the same time but only one will succeed, completing the deferred primitive exactly once. **The loser one will raise an error when trying to complete a deferred already completed and automatically be canceled by the IO.race mechanism, that’s why we call attempt on the evaluation.**

```
import cats.effect.{IO, Deferred}
import cats.syntax.all._

def start(d: Deferred[IO, Int]): IO[Unit] = {
  val attemptCompletion: Int => IO[Unit] = n => d.complete(n).attempt.void

  List(
    IO.race(attemptCompletion(1), attemptCompletion(2)),
    d.get.flatMap { n => IO(println(show"Result: $n")) }
  ).parSequence.void
}

val program: IO[Unit] =
  for {
    d <- Deferred[IO, Int]
    _ <- start(d)
  } yield ()
```

As far as I know, the latter one calls the `complete` and it would just return false wrapped in `F` and there is no error throws and it seems the code doesn't need to use the  `.attempt.void` there too.
Looks like this documentation is incorrect here.
From my checking, the CE2's `complete` [raise an error](https://typelevel.org/cats-effect/api/2.x/cats/effect/concurrent/Deferred.html#complete(a:A):F[Unit]) which is different in the CE3 now.

PR #3122, opened 2022-08-08
Improve documentation

### LINKED ISSUE #3118: The `factorial` sample code in the Thread Model documentation is a little misleading
Page: https://typelevel.org/cats-effect/docs/thread-model#fibers
Documentation parts:
> It is important to note that this is [cooperative multi-tasking](https://en.wikipedia.org/wiki/Cooperative_multitasking) (as opposed to pre-emptive) so it is the responsibility of a fiber to yield control of the CPU by suspending its runloop periodically. In practice this is rarely an issue as fibers automatically yield at asynchronous boundaries (eg I/O) but it does means that it is actually possible for a fiber to take control of a CPU core and never give it back if it executes a tight CPU-bound loop like

```
def factorial(n: BigInt): IO[BigInt] = n match {
   case 0 => IO.pure(1)
   case n => factorial(n-1).flatMap {
      m => IO.pure(m * n)
   }
}

factorial(10000).unsafeRunSync()
```
> If you have such a loop then you can insert a fairness boundary via `IO.shift` (CE2 but has other potential side-effects) or `IO.cede` (CE3), which will give another fiber an opportunity to run on the thread.

The documentation says the above code won't yield control of the CPU which is not true.
In CE3 with the above code sample, there is no need to insert the fairness boundary with `IO.cede` because the CE3 will insert an `IO.cede` every fixed number of iterations of the runloop mentioned in the [later documentation](https://typelevel.org/cats-effect/docs/thread-model#work-stealing-pool).

Maybe it is better to use another example which actually needs to use `IO.cede` or add a note to let the users know the CE3 will add the fairness boundary in the example code.

floating-cat:
Fix #3118 and #3119

### DIFF
--- a/docs/std/deferred.md
+++ b/docs/std/deferred.md
@@ -35,14 +35,14 @@ Finally, the blocking mentioned above is semantic only, no actual threads are bl
 
 Whenever you are in a scenario when many processes can modify the same value but you only care about the first one in doing so and stop processing, then this is a great use case of `Deferred[F, A]`.
 
-Two processes will try to complete at the same time but only one will succeed, completing the deferred primitive exactly once. The loser one will raise an error when trying to complete a deferred already completed and automatically be canceled by the `IO.race` mechanism, that’s why we call attempt on the evaluation.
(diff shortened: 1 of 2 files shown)


djspiewak (approved):
Thank you!

PR #3123, opened 2022-08-09
Add `GenConcurrent#forceStart` that guarantees a fiber will start

bplommer:
The possibility of an uncancelable fiber being canceled before it starts executing is, I think, a surprising footgun for many people. If nothing else, adding this `forceStart` method will help raise awareness of that possibility.

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
@@ -31,6 +31,21 @@ trait GenConcurrent[F[_], E] extends GenSpawn[F, E] {
   def deferred[A]: F[Deferred[F, A]]
 
   /**
+   * Like [[start]], but cancellation of the resulting fiber cannot occur before it starts
+   * running.
+   */
+  def forceStart[A](fa: F[A]): F[Fiber[F, E, A]] =
+    flatMap(deferred[Unit]) { gate =>
+      map(start(flatMap(gate.complete(()))(_ => fa))) { fiber =>
+        new Fiber[F, E, A] {
+          override def cancel: F[Unit] = flatMap(gate.get)(_ => fiber.cancel)
+
+          override def join: F[Outcome[F, E, A]] = fiber.join
+        }
+      }
+    }
+
+  /**
    * Caches the result of `fa`.
    *
    * The returned inner effect, hence referred to as `get`, when sequenced, will evaluate `fa`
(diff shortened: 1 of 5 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/syntax/GenConcurrentSyntax.scala
```diff
@@ -17,10 +17,17 @@
...
 
 trait GenConcurrentSyntax {
 
+  implicit def genConcurrentOps[F[_], A, E](
+      wrapped: F[A]
+  )(implicit F: GenConcurrent[F, E]): GenConcurrentOps[F, A, E] = {
+    val _ = F
+    new GenConcurrentOps(wrapped)
+  }
+
```
bplommer, line 30: This pattern is copied from `GenSpawnOps`

bplommer:
The test failure doesn't appear to be related to the contents of this PR:
```
[info] WorkerThreadNameSpec
[info] WorkerThread should
[error]   x rename itself when entering and exiting blocking region
[error]    the value is false (WorkerThreadNameSpec.scala:73)
```

kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
```diff
@@ -30,6 +30,21 @@ trait GenConcurrent[F[_], E] extends GenSpawn[F, E] {
...
 
+  /**
+   * Like [[start]], but cancellation of the resulting fiber cannot occur before it starts
+   * running.
+   */
+  def forceStart[A](fa: F[A]): F[Fiber[F, E, A]] =
+    flatMap(deferred[Unit]) { gate =>
+      map(start(flatMap(gate.complete(()))(_ => fa))) { fiber =>
+        new Fiber[F, E, A] {
+          override def cancel: F[Unit] = flatMap(gate.get)(_ => fiber.cancel)
```
durban, line 41: This creates a fiber with a cancelable `cancel`, which is not allowed.

durban (commented):
Since `start` also have other footguns, I'm not convinced this specific one is worth "fixing". There are other, higher level alternatives. In any case, I've left an implementation comment.

kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
djspiewak: This actually in and of itself illustrates one of the perils of this concept: what would you expect to happen if the parent fiber is canceled, while calling `cancel`, before the child fiber has a chance to run? Similarly, what if the child fiber is taking a really long time to start? Should you wait for it?

djspiewak:
I'm kind of on the fence about this one, conceptually. Fibers are, fundamentally, three state objects: **unstarted**, **started**, and **completed**. Cancelation has different effects depending on which state the fiber is in (atomically), and this is just fundamental in the model. What this PR does is attempt to eliminate the first state from the calculus by blocking `cancel` until the fiber moves into the second state. This does *work* but it raises a lot of other questions around what you should do if the requisite blocking starts to become noticeably long. In other words, it's still a footgun, just of a different sort.

PR #3127, opened 2022-08-17
#3125 - resume interrupt of calling thread during shutdown of the Wor…

### LINKED ISSUE #3125: WorkStealingThreadPool#shutdown interrupts a calling thread
`WorkStealingThreadPool#shutdown` interrupts a calling thread by calling `Thread.currentThread().interrupt()` as a last call in the method. 

This might not be an issue if the main class is an `IOApp` but when called outside of it, It sets an interrupted flag of the calling thread which might not be desirable as the pool does not own it. 
The interrupt for instance wakes up any latches later on.

blondacz:
…kStealingThreadPool only when already interrupted at the start of the shutdown.


vasilmkd:
I like this, thank you!

vasilmkd (approved):

PR #3128, opened 2022-08-17
Add scope to IOLocal

wunderk1nd-e:
Thought it could be useful to have a way to update the value of an `IOLocal` within the scope of some task `IO[A]`

Motivation was being able to set log context here: https://github.com/typelevel/log4cats/pull/676

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOLocal.scala
+++ b/core/shared/src/main/scala/cats/effect/IOLocal.scala
@@ -233,4 +233,22 @@ object IOLocal {
       }
     }
 
+  /**
+   * Alters the `T` value that is observed by an `IO[G]` value using [[IOLocal]].
+   *
+   * The updated value can only be observed from within that `IO[G]` value.
+   *
+   * @param local
+   *   The [[IOLocal]] to update
+   * @param value
+   *   The new value to use as the context for task
+   * @param task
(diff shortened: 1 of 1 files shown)


wunderk1nd-e:
I think the CI failure might be an unrelated issue (flakiness?)

bplommer:
Could be nice to have this as an instance method on `IOLocal`?

iRevive:
What if we use `Resource` instead? From my point of view, it offers better flexibility and composability
```scala
trait IOLocal[A] {
  def scope[A](value: A): Resource[IO, A] =
    Resource.make(local.getAndSet(value))(local.set)
}
```

That way, the scope can be managed for `Resource`, `fs2.Stream`, and `IO`:
```scala
val local: IOLocal[String] = ???

val stream: fs2.Stream[IO, ...] = fs2.Stream.resource(local.scope("my-scope")).evalMap(_ => ???)
val resource: Resource[IO, ...] = local.scope("external").flatMap(_ => ??? >> local.scope("internal"))
val io: IO[...] = local.scope("my-scope").use(_ => ???)
```

djspiewak (changes requested):
Agreed this should probably be `Resource`, and also would offer better ergonomics as an instance method. Overall though I like this concept!

PR #3129, opened 2022-08-19
Fix vertically repeating GitHub logo on mobile nav bar

tsegarra:


djspiewak:
Thank you! My CSS knowledge is literally negative. Like I make other people worse at CSS just by existing. Thank you so much.

PR #3134, opened 2022-08-29
Documentation: use `IO` in examples

### LINKED ISSUE #3133: A little more `IO` and a little less `F[_]`
I think we've all converged on the idea that beginners should definitely start with concrete `IO`. And that there's no harm in continuing to use concrete `IO` indefinitely; indeed many teams are doing this.

So this is all well and good. Except our docs are full of stuff like this:

https://github.com/typelevel/cats-effect/blob/3da61a59438da30f9bb01192fec9e690aacc3ce4/std/jvm/src/main/scala/cats/effect/std/Console.scala#L30-L39

Which invariably leads to beginner questions: what is `F[_]`? To which a quick answer is "replace every `F` you see with `IO`".

Well, woops 😛 
```scala
def myProgram[IO[_]: Console : Monad]: IO[Unit] = ???
```

I think we should prioritize writing examples with concrete `IO`, since that is exactly what we are pitching. Providing a second example written in terms of `F[_]` (possibly via a fancy tabbed switcher on the website) would be reserved for bonus points.

iRevive:
Closes #3133. 

I went through the Scaladoc and that's what I found. Give me a hint if I missed something.

### DIFF
--- a/std/js/src/main/scala/cats/effect/std/Console.scala
+++ b/std/js/src/main/scala/cats/effect/std/Console.scala
@@ -32,6 +32,19 @@ import java.nio.charset.Charset
  *
  * @example
  *   {{{
+ *  import cats.effect.IO
+ *  import cats.effect.std.Console
+ *
+ *  def myProgram: IO[Unit] =
+ *    for {
+ *      _ <- Console[IO].println("Please enter your name: ")
+ *      n <- Console[IO].readLine
+ *      _ <- if (n.nonEmpty) Console[IO].println("Hello, " + n) else Console[IO].errorln("Name is empty!")
+ *    } yield ()
+ *   }}}
+ *
+ * @example
+ *   {{{
  *  import cats.Monad
  *  import cats.effect.std.Console
  *  import cats.syntax.all._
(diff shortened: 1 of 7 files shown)


std/js/src/main/scala/cats/effect/std/Console.scala
```diff
@@ -32,6 +32,19 @@ import java.nio.charset.Charset
...
+ *  import cats.effect.IO
+ *  import cats.effect.std.Console
+ *
+ *  def myProgram: IO[Unit] =
+ *    for {
+ *      _ <- Console[IO].println("Please enter your name: ")
+ *      n <- Console[IO].readLine
+ *      _ <- if (n.nonEmpty) Console[IO].println("Hello, " + n) else Console[IO].errorln("Name is empty!")
+ *    } yield ()
+ *   }}}
```
armanbilge, line 44: There's an interesting question here, since I believe `IO` has defined `IO.println` and `IO.readLine` methods, so we could demonstrate that instead.

But I think the `Console[IO]` pattern is more relevant for this scaladoc, and is also a pattern that will _always_ work regardless of whether `IO` has defined aliases for those methods.

armanbilge: Oh, also just realized this is the JavaScript docs. Not your problem, but actually `readLine` is deprecated on JS because it's broken 😂

armanbilge:
Overall looks good, thanks for this!

std/js/src/main/scala/cats/effect/std/Console.scala
iRevive: Yeah, I decided to keep `Console[IO]` for consistency's sake.

samspills (approved):
Nice!! I'm happy to see both examples in the docs. I built locally and confirmed that everything looked good (in particular the switch from `Example:` to `@example` is nice 😎)

PR #3135, opened 2022-09-04
Retry functionality

### LINKED ISSUE #1459: Fold cats-retry into CE 3?
Would there be any interest in folding the features of [cats-retry](https://github.com/cb372/cats-retry) (i.e. a few combinators for retrying effectful actions + a DSL for building retry policies) into CE 3? Would this be a good fit for the `std` module?

I have to admit up front that I probably wouldn't have time to lead this, but I'd be happy to support such an effort.

SystemFw:
Fixes #1459 

I had originally started with a direct port of cats-retry, but since then I've changed the design significantly enough to warrant a proper review of the api before I port tests and docs. 
Everything is in `Retry.scala`, here are some of the relevant decisions:

- The time scheduling part of the problem is mostly the same as cats-retry, except with fewer types.
- The retry api itself has changed significantly, cats-retry has several combinators, each of which takes several callbacks, whereas I've folded that into the Retry abstraction itself. I've found the deluge a similar combinators in cats-retry to be a bit hard to navigate, plus you can't have default implementations for the callbacks because they are of shape `X => F[Y]`.
- I've removed the pretty printing of retries, we don't do that anywhere else in cats-effect.
- I've removed the ability to retry with a predicate on the result of the action, with the rationale that you can use failures and `ensure` instead.
- The errors are typed at Throwable (could be given the Gen* treatment though I'm not especially keen).

TODO:

- [x] Syntax. Should be minimal given the reduction to a single retry combinator. ~~I'm afraid usage with `IO` would require a syntax import since this lives in `std`.~~
- [ ] Pass at the scaladoc.
- [ ] Porting tests.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/OldRetry.scala
+++ b/std/shared/src/main/scala/cats/effect/std/OldRetry.scala
@@ -1,0 +1,666 @@
+package cats.effect.std
+package retry
+
+import cats._
+import cats.syntax.all._
+import cats.arrow.FunctionK
+import cats.kernel.BoundedSemilattice
+import cats.effect.kernel.Temporal
+import retry.PolicyDecision._
+import scala.concurrent.duration.{Duration, FiniteDuration}
+import scala.util.Random
+import java.util.concurrent.TimeUnit
+
+
+final case class RetryStatus(
+    retriesSoFar: Int,
+    cumulativeDelay: FiniteDuration,
+    previousDelay: Option[FiniteDuration]
+) {
+  def addRetry(delay: FiniteDuration): RetryStatus = RetryStatus(
+    retriesSoFar = this.retriesSoFar + 1,
+    cumulativeDelay = this.cumulativeDelay + delay,
+    previousDelay = Some(delay)
+  )
+}
+
+object RetryStatus {
+  val NoRetriesYet = RetryStatus(0, Duration.Zero, None)
+}
+
+sealed trait PolicyDecision
+
+object PolicyDecision {
+  case object GiveUp extends PolicyDecision
+
+  final case class DelayAndRetry(
+      delay: FiniteDuration
+  ) extends PolicyDecision
+}
+
+case class RetryPolicy[M[_]](
+    decideNextRetry: RetryStatus => M[PolicyDecision]
+) {
+  def show: String = toString
+
+  def followedBy(rp: RetryPolicy[M])(implicit M: Apply[M]): RetryPolicy[M] =
+    RetryPolicy.withShow(
+      status =>
+        M.map2(decideNextRetry(status), rp.decideNextRetry(status)) {
+          case (GiveUp, pd) => pd
+          case (pd, _) => pd
+        },
+      show"$show.followedBy($rp)"
+    )
+
+  /**
+   * Combine this schedule with another schedule, giving up when either of the schedules want to
+   * give up and choosing the maximum of the two delays when both of the schedules want to delay
+   * the next retry. The dual of the `meet` operation.
+   */
+  def join(rp: RetryPolicy[M])(implicit M: Apply[M]): RetryPolicy[M] =
+    RetryPolicy.withShow[M](
+      status =>
+        M.map2(decideNextRetry(status), rp.decideNextRetry(status)) {
+          case (DelayAndRetry(a), DelayAndRetry(b)) => DelayAndRetry(a max b)
+          case _ => GiveUp
+        },
+      show"$show.join($rp)"
+    )
+
+  /**
+   * Combine this schedule with another schedule, giving up when both of the schedules want to
+   * give up and choosing the minimum of the two delays when both of the schedules want to delay
+   * the next retry. The dual of the `join` operation.
+   */
+  def meet(rp: RetryPolicy[M])(implicit M: Apply[M]): RetryPolicy[M] =
+    RetryPolicy.withShow[M](
+      status =>
+        M.map2(decideNextRetry(status), rp.decideNextRetry(status)) {
+          case (DelayAndRetry(a), DelayAndRetry(b)) => DelayAndRetry(a min b)
+          case (s @ DelayAndRetry(_), GiveUp) => s
+          case (GiveUp, s @ DelayAndRetry(_)) => s
+          case _ => GiveUp
+        },
+      show"$show.meet($rp)"
+    )
+
+  def mapDelay(
+      f: FiniteDuration => FiniteDuration
+  )(implicit M: Functor[M]): RetryPolicy[M] =
+    RetryPolicy.withShow(
+      status =>
+        M.map(decideNextRetry(status)) {
+          case GiveUp => GiveUp
+          case DelayAndRetry(d) => DelayAndRetry(f(d))
+        },
+      show"$show.mapDelay(<function>)"
+    )
+
+  def flatMapDelay(
+      f: FiniteDuration => M[FiniteDuration]
+  )(implicit M: Monad[M]): RetryPolicy[M] =
+    RetryPolicy.withShow(
+      status =>
+        M.flatMap(decideNextRetry(status)) {
+          case GiveUp => M.pure(GiveUp)
+          case DelayAndRetry(d) => M.map(f(d))(DelayAndRetry(_))
+        },
+      show"$show.flatMapDelay(<function>)"
+    )
+
+  def mapK[N[_]](nt: FunctionK[M, N]): RetryPolicy[N] =
+    RetryPolicy.withShow(
+      status => nt(decideNextRetry(status)),
+      show"$show.mapK(<FunctionK>)"
+    )
+}
+
+object RetryPolicy {
+  def lift[M[_]](
+      f: RetryStatus => PolicyDecision
+  )(implicit M: Applicative[M]): RetryPolicy[M] =
+    RetryPolicy[M](decideNextRetry = retryStatus => M.pure(f(retryStatus)))
+
+  def withShow[M[_]](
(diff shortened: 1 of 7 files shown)


armanbilge:
> I'm afraid usage with `IO` would require a syntax import since this lives in `std`.

The core module depends on `std`, so syntax imports shouldn't be necessary, we can add methods directly to `IO`. Unless I misunderstood what you are getting at :)

std/shared/src/main/scala/cats/effect/std/Retry.scala
```diff
@@ -0,0 +1,297 @@
...
+    Retry.lift[F](_ => DelayAndRetry(delay))
+
+  /**
+   * Each delay is twice as long as the previous one. Never give up.
+   */
+  def exponentialBackoff[F[_]: Monad](
+      baseDelay: FiniteDuration
+  ): Retry[F] =
+    Retry.lift[F] { status =>
+      val delay = safeMultiply(baseDelay, Math.pow(2, status.retriesSoFar).toLong)
```
ybasket: Maybe make the base (`2`) configurable? It could still be the default as many people will want to use it as their base.

SystemFw: I can probably add it as a param with a default, yeah

armanbilge: An overload is better for bincompat 🙏

SystemFw:
By the way, I'm explicitly seeking feedback on the interaction between `selectError` and `and/or` with timed retry policies. I think in common usage it's fine, but I also think it's possible to build confusing behaviour in certain cases. However, the alternative is writing lots of overloads for `retry` combinators with callbacks, and forcing the user to bundle timing logic and error selection logic manually

armanbilge:
> I also think it's possible to build confusing behaviour in certain cases

would you mind showing an example of this?

SystemFw:
> would you mind showing an example of this?

For example `capDelay` used to be implemented as `_.or(constantDelay(cap))`. But if you use that with a policy that works on a specific error, e.g.
```scala
exponentialBackoff(1.ms)
  .selectError { 
       case NullPointerException => true
       case _ => false
    }.capDelay(500.ms) // with the old implementation of capDelay
```

the resulting policy will retry all errors, not just `NullPointerException`. Another example would be `and`ing policies that select non-overlapping errors, like `constantDelay.selectError(..ExA..).and(constantDelay.selectError(...ExB...)`, which means you never retry (though this one should be quite obvious with the rename from `join` to `and`).
All in all I think common usage is fine, since I'd imagine you first write the code that specifies the "scheduling" , and then `selectError` at the end, and also by following the simple rule of always having non-overlapping error selection for `or`, and overlapping error selection for `and`, but I think it's worth getting other people to think about it.

std/shared/src/main/scala/cats/effect/std/Retry.scala
bplommer: Or a builder? Overloads tend to lead to combinatorial explosion

std/shared/src/main/scala/cats/effect/std/Retry.scala
```diff
@@ -0,0 +1,297 @@
...
+  def selectError(f: Throwable => Boolean): Retry[F]
+
+  def selectErrorWith(f: Throwable => F[Boolean]): Retry[F]
+
+  def flatTap(f: (Throwable, Retry.Decision, Retry.Status) => F[Unit]): Retry[F]
+
+  def mapK[G[_]: Monad](f: F ~> G): Retry[G]
+}
+object Retry {
+  final case class Status(
```
bplommer, line 61: Case classes tend to be fiddly from a bincompat perspective - could this be package-private and abstracted behind a sealed trait?

std/shared/src/main/scala/cats/effect/std/OldRetry.scala
```diff
@@ -0,0 +1,666 @@
+package cats.effect.std
+package retry
```
djspiewak, line 2: I'm assuming `OldRetry` is only here for comparison purposes?

std/shared/src/main/scala/cats/effect/std/Retry.scala
```diff
@@ -0,0 +1,305 @@
...
+    nextRetry: (Retry.Status, Throwable) => F[Retry.Decision]
+  ): Retry[F] =
+    new RetryImpl[F](nextRetry)
+
+  def liftF[F[_]: Monad](
+    nextRetry: Retry.Status => F[Retry.Decision]
+  ): Retry[F] =
+    Retry((status, _) => nextRetry(status))
+
+  def lift[F[_]: Monad](
```
djspiewak, line 109: This feels like an odd use of the `lift` nomenclature. I probably would have gone with `apply` or something like that, though I'm not sure what to do about the non-functor variant.

std/shared/src/main/scala/cats/effect/std/Retry.scala
```diff
@@ -0,0 +1,305 @@
...
+  /**
+   * Delay by a constant amount before each retry. Never give up.
+   */
+  def constantDelay[F[_]: Monad](delay: FiniteDuration): Retry[F] =
+    Retry.lift[F](_ => DelayAndRetry(delay))
+
+  /**
+   * Each delay is twice as long as the previous one. Never give up.
+   */
+  def exponentialBackoff[F[_]: Monad](
```
djspiewak, line 130: TIL people actually use exponential backoff *without* jitter. I had no idea that was even a useful thing to do…

std/shared/src/main/scala/cats/effect/std/Retry.scala
```diff
@@ -0,0 +1,305 @@
...
+          safeMultiply(baseDelay, fibonacci(status.retriesSoFar + 1))
+        DelayAndRetry(delay)
+    }
+  }
+
+  /**
+   * "Full jitter" backoff algorithm. See
+   * https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
+   */
+  def fullJitter[F[_]: Monad](baseDelay: FiniteDuration): Retry[F] =
```
djspiewak, line 183: Continuing on my thought from above… It feels like this should be the algorithm that we're encouraging people to reach for first, but the name definitely does not hint in that direction. I had to actually look up the AWS Engineering blog to figure out that this is actually the algorithm that I generally identify with "exponential backoff".

std/shared/src/main/scala/cats/effect/std/Retry.scala
```diff
@@ -0,0 +1,305 @@
...
+  object Decision {
+    case object GiveUp extends Decision
+    final case class DelayAndRetry(delay: FiniteDuration) extends Decision
+  }
+
+  import Decision._
+
+  val noRetriesYet = Retry.Status(0, Duration.Zero, None)
+
+  def retry[F[_]: Temporal, A](r: Retry[F], action: F[A]): F[A] = {
```
djspiewak, line 83: Why is this not a member of `Retry`?

std/shared/src/main/scala/cats/effect/std/syntax/RetrySyntax.scala
```diff
@@ -0,0 +1,15 @@
...
+
+trait RetrySyntax {
+  implicit def retryOps[F[_], A](wrapped: F[A]): RetryOps[F, A] =
+    new RetryOps(wrapped)
+}
+
+final class RetryOps[F[_], A] private[syntax] (private[syntax] val wrapped: F[A])
+    extends AnyVal {
+  def retry(r: Retry[F])(implicit F: Temporal[F]): F[A] =
+    Retry.retry(r, wrapped)
```
djspiewak, line 14: Should we also add some convenience syntax for exponential backoff?

djspiewak (commented):
General thoughts…

I like where this is headed. It does feel slightly odd to me that we have all these combinators on `Retry` itself, but at the same time the only way to make a *stateful* decision chain is to do so using `Retry.lift` (which should be renamed but that's another topic). It feels like the set of combinators is incomplete somehow. Maybe the decision calculus is what's incomplete? I haven't given it too much thought.

std/shared/src/main/scala/cats/effect/std/Retry.scala
```diff
@@ -0,0 +1,305 @@
...
+
+  /**
+   * "Full jitter" backoff algorithm. See
+   * https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
+   */
+  def fullJitter[F[_]: Monad](baseDelay: FiniteDuration): Retry[F] =
+    Retry.lift[F] { status =>
+      val e = Math.pow(2, status.retriesSoFar.toDouble).toLong
+      val maxDelay = safeMultiply(baseDelay, e)
+      val delayNanos = (maxDelay.toNanos * Random.nextDouble()).toLong
```
kubukoz, line 187: Should this use `cats.effect.std.Random`?

std/shared/src/main/scala/cats/effect/std/Retry.scala
Daenyth: Maybe `def byStatus`?

std/shared/src/main/scala/cats/effect/std/Retry.scala
Daenyth: It feels a little odd to me that these are methods rather than some kind of `Strategy` that encapsulates the `(previousBackoff: FiniteDuration) => (nextBackoff: F[FiniteDuration])`

If they were structured in that way, we could provide easier guidance by having the strategies be available on `object Strategy` with `def default` being this recommended one

PR #3136, opened 2022-09-05
Bump to Scala 3.2.0, add extra GraalVM job

### LINKED ISSUE #3030: GraalVM native image errors out with `TypeNotPresentException` on Scala 3

armanbilge:
Fixes https://github.com/typelevel/cats-effect/issues/3030. I add a new Scala 3 + GraalVM Native Image job to CI to track this going forward.

Note that this is targeted at 3.3.x. Assumptions:

1. Cats Effect 3.4.0 is blocked for the foreseeable future (unfortunately due to an ambitious changeset paired with lack of maintainer time).
2. There is no workaround in user-land. We literally need to publish new artifacts with fixed bytecode.
3. Native Image support is important.
4. The official advice from Scala 3 team is to upgrade.
    https://www.scala-lang.org/blog/2022/09/05/scala-3.2.0-released.html#should-i-update-to-scala-320

We could also just eat the 3.4.0 bump now, just for this, and the original plans for that can become 3.5.0. Honestly I'd feel a lot more comfortable doing that.

Full disclaimer: getting this upgrade in would also unblock my Scala Native work.

@vasilmkd what do you think?


armanbilge:
One more reason to go to 3.2.0: we can make an `@iomain` annotation! h/t Anton

https://scala-lang.org/api/3.x/scala/annotation/MainAnnotation.html#

arixmkii:
Looking forward for this change! Thank you @armanbilge for working on this.

Adding my 2 cents on GraalVM native image importance. If you have a requirement to use it - there is no alternative. So, in this case people would consider different stack, but to stick with GraalVM.

As a workaround. Within the corporate environments one can do rebuilds of the tagged release with a supported compiler, but this doesn't scale for OSS projects ecosystem.

djspiewak:
If we can retarget this to 3.4.0, I'll merge :-) This looks great!

djspiewak:
@armanbilge *nudging this*

armanbilge:
Haven't forgotten! Since it's going to be retargeted at 3.4 anyway I'm waiting for the Scala Native CI changes to land since the CI changes here are bound to conflict.

djspiewak:
I think this rebase is unblocked now?

PR #3139, opened 2022-09-10
`Unique` doc - make explicit that it isn't a business identifier

bplommer:
Also removed the incorrect suggestion that there is *any* guarantee about uniqueness of the hash code.


valencik:
I like how explicit and clear this is, nicely done!

djspiewak (approved):

Issue #3077, opened 2022-07-06
`IO.readLine.timeoutTo(...)` not cancelling the `readLine` after timeout.


VzxPLnHqr:
This might be related to the decision made here regarding `readLine` https://github.com/typelevel/cats-effect/issues/1793 

In the following simple program, the expected behavior (I think?) is that it should still terminate after the timeout, since a default value has been supplied. However, instead it seems to hang waiting for something to be input by the user (only to ignore it later). I am not sure if this is a bug or if this is the expected behavior?
 
```
object Main extends IOApp.Simple {
    val run = for {
        _ <- IO.println("10 seconds to enter something:")
        v <- IO.readLine.timeoutTo(10.seconds,IO("a default value"))
        _ <- IO.println(s"you said $v")
    } yield ()
}
```

It seems like it is may be a bug since `readLine` is wrapped in interruptable here: https://github.com/typelevel/cats-effect/blob/386c27be0596935bccb404e75799c61470c5e8e8/std/shared/src/main/scala/cats/effect/std/ConsoleCrossPlatform.scala#L178

armanbilge:
Ah, thanks for linking to https://github.com/typelevel/cats-effect/issues/1793, I wasn't aware of that issue. Indeed it seems like `IO.readLine` is not always cancelable, for reasons outside of our control.

This can be observed in many ways: here, by `timeoutTo` hanging since it is awaiting cancelation of an uncancelable op.

We do have a `timeoutAndForget` combinator that wouldn't hang, but it would leak a `Thread` and probably cause other issues if you try and `readLine` again later.

https://github.com/typelevel/cats-effect/blob/7887b34b1453131f630ded107540ed11d6fe21ab/core/shared/src/main/scala/cats/effect/IO.scala#L605-L623

VzxPLnHqr:
Thanks. I can confirm that the following "works" properly but then starts doing weird things on subsequent calls to `readLine` as you said:
```
object Main extends IOApp.Simple {
    val run = for {
        _ <- IO.println("enter something in 3 seconds:")
        v <- IO.readLine.timeoutAndForget(3.seconds).handleError(_ => "hello")
        _ <- IO.println(s"you entered $v")
    } yield ()
}
```

armanbilge:
Thanks for giving that a try! Yeah, I wouldn't recommend it anyway due to the whole leaking-a-thread thing.

[...]

PR #3141, opened 2022-09-12
update Console doc string

samspills:
After [a brief Discord convo](https://discord.com/channels/632277896739946517/895394320100761731/1018891967523983382) with @armanbilge, I thought it would be good to make the Console API docs a bit more explicit in two ways:
1. There's an issue with cancellation in `readLine`, related to #3077
2. b/c of that, Console should only be used for simple and **trivial** applications (to clarify that we don't just mean not-complex applications, but not-complex not-in-production applications)


armanbilge (approved):
Thank you very much! :)

PR #3143, opened 2022-09-13
Add Mutex

### LINKED ISSUE #3142: Add `Mutex` to cats-effect-std
Following a [Discord tutorial on mutex](https://discord.com/channels/632277896739946517/632278585700384799/1019252614589841449) by Fabio :)

The interface would be:
```scala
trait Mutex[F[_]] {
  def permit: Resource[F, Unit]
} 
```

And it can be implemented in terms of a semaphore with a single permit.

Ok, more ideas incoming 😛  with this in place we can get a `MutexRef[F[_], A]` which enables effectual modification:
```scala
def modify(f: A => F[(A, B)]): F[B]
```

BalmungSan:
Fixes #3142

I am open to any suggestions, especially I am looking for feedback on naming, docs and tests.
I am also open to feedback on the implementations but I believe those are pretty straightforward.

-----

> BTW, orthogonal to this, I noticed that `SempahoreSpec` is in the `kernel` folder but in the `std` package, should I move the file to the right folder?

In **Discord**, Arman recommended that we should indeed move the file.


std/shared/src/main/scala/cats/effect/std/Mutex.scala
```diff
@@ -0,0 +1,79 @@
...
+package std
+
+import cats.effect.kernel._
+import cats.syntax.all._
+
+/**
+ * A purely functional mutex.
+ *
+ * A mutex is a concurrency primitive that can be used to give access to a resource to only one
+ * fiber at a time; e.g. a [[cats.effect.kernel.Ref]]`
```
djspiewak, line 28: Is it worth documenting that this is not a reentrant lock in its current form? For example: `mutex.lock.surround(mutex.lock.use_)` will deadlock when, arguably, it shouldn't.

BalmungSan: I would say yes.

However, before adding that note, I wonder how hard would it be to make it rentrant? Do we have a way to identify the calling `Fiber`? If so, it should be as simple as having a `Ref` with the id of the current owner.

djspiewak: Fibers don't have ids, but you could have an `IOLocal` for any `LiftIO[F]`. That wouldn't be fully generic though. The lack of ability to define a reentrant lock is one of the classical objections to the lack of fiber identity in the CE calculus.

BalmungSan: @djspiewak so, if I am understanding correctly, there is no easy way to make it reentrant. Thus, the best is just to document, right?

djspiewak: That would be my recommendation. The only other option would be to make it `LiftIO`-specific and leverage locals.

SystemFw: I think eventually we'll have to make peace with the idea that this functionality has to be added :)

djspiewak: I generally would like to, but the implications of general fiber identity are so vast and terrifying that it's really difficult to square with the current calculus. I'd rather find a more confined version of the requirements, but so far that has proven elusive.

djspiewak (approved):

PR #3153, opened 2022-09-19
Documentation: add `Scala Native` details

### LINKED ISSUE #3146: Scala Native. Document platform limitations
I'm experimenting with Scala Native and cats-effect. @armanbilge thank you for your work to make it possible!

~TLDR: `unsafeRunSync()` is missing~
The Scala Native platform has limitations, for example, the runtime is single-threaded. It would be nice to document that.

**scala-cli**: 0.1.12
**Scala**: 3.2.0, 3.1.3
**CE**: 3.1.3
**Scala Native**: 0.4.7

The snippet:
```scala
//> using scala "3.2.0"
//> using lib "org.typelevel::cats-effect::3.3.14"

import cats.effect.*

@main
def hello() = {
  IO.println("hello world").unsafeRunSync()
}
```

`compile` and `run` commands are working fine:
```bash
$ scala-cli run main.scala
hello world
```

But `scala-cli --native --native-version 0.4.7 main.scala` does not work:
```bash
$ scala-cli --native --native-version 0.4.7 main.scala
Compiling project (Scala 3.2.0, Scala Native)
[error] ./main.scala:9:3: value unsafeRunSync is not a member of cats.effect.IO[Unit] - did you mean cats.effect.IO[Unit].unsafeRunAsync?
[error]   IO.println("hello world").unsafeRunSync()
[error]   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error compiling project (Scala 3.2.0, Scala Native)
Compilation failed
```

`IOApp.Simple` works fine:
```scala
//> using scala "3.2.0"
//> using lib "org.typelevel::cats-effect::3.3.14"

import cats.effect._

  def run: IO[Unit] = IO.println("hello world")
}
```

```bash
$  scala-cli --native --native-version 0.4.7 main.scala
[info] Linking (1163 ms)
[info] Discovered 1522 classes and 9336 methods
[info] Optimizing (debug mode) (1441 ms)
[info] Generating intermediate code (1300 ms)
[info] Produced 8 files
[info] Compiling to native code (1991 ms)
[info] Linking native code (immix gc, none lto) (131 ms)
[info] Total (6074 ms)
hello world
```

I assume, there is something wrong with the linkage of `unsafeRunSync()` method?

iRevive:
Closes #3146.

I've added three items:
- Scala Native + sbt
- Scala Native + scala-cli
- Scala Native runtime limitations

### DIFF
--- a/docs/core/scala-native.md
+++ b/docs/core/scala-native.md
@@ -1,0 +1,66 @@
+---
+id: scala-native
+title: Scala Native
+---
+
+Cats Effect supports [Scala Native](https://github.com/scala-native/scala-native) since `3.3.14`.
+
+## Scala Native example
+
+Here's an example of a Hello world project compiled to a native executable 
+using [sbt-scala-native plugin](https://github.com/scala-native/scala-native).
+
+```scala
+// project/plugins.sbt
+addSbtPlugin("org.scala-native" % "sbt-scala-native" % "0.4.7")
+
+// build.sbt
+ThisBuild / organization := "com.example"
+ThisBuild / scalaVersion := "2.13.8"
+
+lazy val root = project.in(file("."))
+  .enablePlugins(ScalaNativePlugin)
+  .settings(
+    name                := "cats-effect-3-hello-world",
+    libraryDependencies += "org.typelevel" %%% "cats-effect" % "3.3.14",
+    Compile / mainClass := Some("com.example.Main")
+ )
+
+// src/main/scala/com/example/Main.scala
+package com.example
+
+import cats.effect.{IO, IOApp}
+
+object Main extends IOApp.Simple {
+  def run: IO[Unit] = IO.println("Hello Cats Effect!")
+}
+```
+
+The code can be compiled using `sbt nativeLink` and a native executable can then
+be found under `target/scala-2.13/cats-effect-3-hello-world-out`, and executed as any native
+executable with the benefit of a really fast startup time ([hyperfine](https://github.com/sharkdp/hyperfine) 
+is a command line benchmarking tool written in Rust)
+
+```sh
+$ ./target/scala-2.13/cats-effect-3-hello-world-out
+Hello Cats Effect!
+
+$ hyperfine ./target/scala-2.13/cats-effect-3-hello-world-out
+Benchmark 1: ./target/native-image/cats-effect-3-hello-world-out
+  Time (mean ± σ):       7.5 ms ±   1.2 ms    [User: 3.6 ms, System: 2.3 ms]
+  Range (min … max):     6.0 ms …  17.9 ms    141 runs
+```
+
+Another way to get your cats effect app compiled to a native executable is to leverage
+the package command of scala-cli, like in the [example](../faq.md#Scala-Native-Example)
+
+## Limitations
+
+The Scala Native runtime is [single-threaded](https://scala-native.org/en/latest/user/lang.html#multithreading), similarly to ScalaJS. That's why the `IO#unsafeRunSync` is not available.
+
+Due to platform limitations, some functionality is provided by third-party libraries:
+- TCP support requires [epollcat](https://github.com/armanbilge/epollcat) in the classpath
+- TLS support requires [s2n-tls](https://github.com/aws/s2n-tls) installed in the system
+
+[epollcat](https://github.com/armanbilge/epollcat) is the I/O-integrated runtime for Cats Effect on Scala Native, implemented with the [epoll](https://man7.org/linux/man-pages/man7/epoll.7.html) API on Linux and the [kqueue](https://en.wikipedia.org/wiki/Kqueue) API on macOS.
+The library offers drop-in replacements for `IOApp` and `IOApp.Simple`. Use `EpollApp` or `EpollApp.Simple` respectively.
--- a/docs/core/scala-native.md
+++ b/docs/core/scala-native.md
@@ -56,11 +56,10 @@ the package command of scala-cli, like in the [example](../faq.md#Scala-Native-E
 
 ## Limitations
 
-The Scala Native runtime is [single-threaded](https://scala-native.org/en/latest/user/lang.html#multithreading), similarly to ScalaJS. That's why the `IO#unsafeRunSync` is not available.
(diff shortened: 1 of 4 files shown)


armanbilge:
This is amazing, thanks for doing this!! Will take a proper look very soon.

docs/schedulers.md
```diff
@@ -68,13 +68,24 @@ Ultimately, though, `IO` is able to achieve this with exceptionally high perform
...
+## Scala Native
+
+The [Scala Native](https://github.com/scala-native/scala-native) runtime is [single-threaded](https://scala-native.org/en/latest/user/lang.html#multithreading), similarly to ScalaJS. That's why the `IO#unsafeRunSync` is not available.
+
+Due to platform limitations, some functionality is provided by third-party libraries:
+- TCP support requires [epollcat](https://github.com/armanbilge/epollcat) in the classpath
+- TLS support requires [s2n-tls](https://github.com/aws/s2n-tls) installed in the system
+
+[epollcat](https://github.com/armanbilge/epollcat) is the I/O-integrated runtime for Cats Effect on Scala Native, implemented with the [epoll](https://man7.org/linux/man-pages/man7/epoll.7.html) API on Linux and the [kqueue](https://en.wikipedia.org/wiki/Kqueue) API on macOS.
+The library offers drop-in replacements for `IOApp` and `IOApp.Simple`. Use `EpollApp` or `EpollApp.Simple` respectively.
```
armanbilge: We just published a blog post explaining how the Native runtime works :)

https://typelevel.org/blog/2022/09/19/typelevel-native.html#how-does-it-work

Instead of repeating some of this information from the other page maybe you can link to there.

docs/schedulers.md
```diff
@@ -68,13 +68,24 @@ Ultimately, though, `IO` is able to achieve this with exceptionally high perform
...
+
 ### Yielding
 
 The primary mechanism which needs to be provided for `IO` to successfully implement fibers is often referred to as a *yield*. In JavaScript terms, this means the ability to submit a callback to the event loop which will be invoked at some point in the future when the event loop has capacity. In a sense, it's submitting a callback which "goes to the back of the queue", waiting for everything else to have its turn before running again.
```
armanbilge: This sub-section belonged under "JavaScript" actually. But actually the details here are relevant to how Scala Native works as well, so maybe that can be adjusted to reflect that.

docs/core/scala-native.md
```diff
@@ -0,0 +1,66 @@
...
+  Time (mean ± σ):       7.5 ms ±   1.2 ms    [User: 3.6 ms, System: 2.3 ms]
+  Range (min … max):     6.0 ms …  17.9 ms    141 runs
+```
+
+Another way to get your cats effect app compiled to a native executable is to leverage
+the package command of scala-cli, like in the [example](../faq.md#Scala-Native-Example)
+
+## Limitations
+
+The Scala Native runtime is [single-threaded](https://scala-native.org/en/latest/user/lang.html#multithreading), similarly to ScalaJS. That's why the `IO#unsafeRunSync` is not available.
```
armanbilge: Another important limitation is that if you use `IO.blocking(...)` for something there is no blocking threadpool, it actually just blocks the thread 😅

docs/core/scala-native.md
```diff
@@ -0,0 +1,66 @@
...
+
+Due to platform limitations, some functionality is provided by third-party libraries:
+- TCP support requires [epollcat](https://github.com/armanbilge/epollcat) in the classpath
+- TLS support requires [s2n-tls](https://github.com/aws/s2n-tls) installed in the system
+
+[epollcat](https://github.com/armanbilge/epollcat) is the I/O-integrated runtime for Cats Effect on Scala Native, implemented with the [epoll](https://man7.org/linux/man-pages/man7/epoll.7.html) API on Linux and the [kqueue](https://en.wikipedia.org/wiki/Kqueue) API on macOS.
+The library offers drop-in replacements for `IOApp` and `IOApp.Simple`. Use `EpollApp` or `EpollApp.Simple` respectively.
```
armanbilge: You may want to clarify that these libraries are mostly needed (and best used with) FS2, http4s, Skunk, etc. _Technically_ they aren't directly related to Cats Effect itself, and they aren't strictly needed.

For example using only `IOApp` @keynmol cross-built the Langoustine LSP for Native.

https://github.com/neandertech/langoustine/blob/a1831b6387e02b0fde666add1cf705d41b4faa8f/modules/app/src/main/scalanative/LangoustineAppPlatform.scala

armanbilge (commented):
Thank you so much for doing this!! Just a few notes.

docs/core/scala-native.md
keynmol: Yeah, the notion around EpollApp should really go to the docs on http4s and fs2 in big red letters, technically entirety of CE-only functionality can be run via IOApp, right?

armanbilge: Yes, exactly. Nothing in CE needs `EpollApp`, nor does `fs2-core`, or a bunch of auxiliary libs like Vault, Keypool, log4cats, fs2-data, Weaver, etc. etc. etc.

Also there is http4s-curl which exposes its own `CurlApp`.

`EpollApp` is only necessary to use `fs2.io.net`, or anything which uses that. Same with the TLS stuff.

Also, I'm hoping to introduce a `UringApp` into the mix too.

keynmol: > Also, I'm hoping to introduce a UringApp into the mix too.

"introduce into the _ring_" surely? 🙄 🥊

armanbilge: Right, here are two more apps using only `IOApp`:
- https://github.com/scodec/hexdump4s
- https://github.com/mpilquist/tcpdump4s

So there's still plenty that can be done with it :) just no async network I/O

iRevive: The more I look into a paragraph more misleading it feels. As you mentioned, a user can enjoy cats-effect with Scala Native without extra dependencies. 

I removed mentions of `EpollApp`. We can keep these details in suitable projects like fs2/http4s.

docs/core/scala-native.md
iRevive: Makes sense. I mentioned `IO.blocking` too.

docs/schedulers.md
iRevive: Good point

armanbilge (approved):
Very nice addition! Thank you!

PR #3156, opened 2022-09-20
Fixed Dispatcher cancelation interactions with new modes

djspiewak:
It's been so long I can't even remember what I did here, but it works! At least, thus say the unit tests.

---

**Edit** Okay so a longer explanation now that I've had some time to reflect. This fixes a few bugs, including some fundamentally squirrely things about the new `Dispatcher` modes. For example, if you `unsafeRunAndForget` on `F.canceled`, what would you expect to happen? In `parallel` this is fairly obvious: it's already on a forked fiber, that fiber dies, end of story. In `sequential` it becomes much more complex, and in the current implementation on **series/3.x**, it actually murders the whole dispatcher instance.

Additionally, the interactions between `await` and the parallel/sequential modes are not straightforward at all. I tried to spell things out really clearly in the tests (I recommend you start with the tests in review!) but I may have missed some things.

As you can probably see by the commit log (which I have left bare for all to peruse), this is where I got stuck for the past few months.

### DIFF
--- a/tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
@@ -274,7 +274,9 @@ class DispatcherSpec extends BaseSpec {
             _ <- cdl.await // make sure the execution of fiber has started
           } yield ()
         }.start
-        _ <- releaseInner.await.timeoutTo(2.seconds, IO(false must beTrue)) // release process has started
+        _ <- releaseInner
+          .await
+          .timeoutTo(2.seconds, IO(false must beTrue)) // release process has started
         released1 <- fiber.join.as(true).timeoutTo(200.millis, IO(false))
         _ <- fiberLatch.release
         released2 <- fiber.join.as(true).timeoutTo(200.millis, IO(false))
--- a/tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
@@ -252,6 +252,24 @@ class DispatcherSpec extends BaseSpec {
         }
       }
     }
+
+    "respect self-cancelation" in real {
+      dispatcher use { runner =>
+        for {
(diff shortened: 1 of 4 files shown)


tests/shared/src/test/scala/cats/effect/std/SupervisorSpec.scala
```diff
@@ -19,61 +19,215 @@ package std
...
+    "cancel inner fiber and ignore restart if outer canceled" in real {
+      val test = IO.deferred[Unit] flatMap { latch =>
+        constructor(true, Some(_.fold(true, _ => false, _ => false))).use { supervisor =>
+          supervisor.supervise(latch.complete(()) >> IO.canceled) >> latch.get >> IO.canceled
+        }
       }
+
+      // if this doesn't work properly, the test will hang
+      test.start.flatMap(_.join).as(ok).timeoutTo(2.seconds, IO(false must beTrue))
+    }
```
armanbilge, line 217: Honestly I'm confused what this test is doing. IIUC, the test will hang if the latch is not completed. But in what situation would the latch not be completed?

armanbilge: Oh right, the test will hang because of the `await = true`? Then it makes sense :)

tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
```diff
@@ -252,6 +252,24 @@ class DispatcherSpec extends BaseSpec {
         }
       }
     }
+
+    "respect self-cancelation" in real {
```
armanbilge, line 256: Do we need to add a test that this self-cancellation doesn't kill the `Dispatcher` i.e. that it is possible to submit another action to it?

armanbilge (dismissed):

tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
djspiewak: Good call. I'll add that case.

PR #3165, opened 2022-09-24
Enabled configurability of unhandled fiber exceptions

### LINKED ISSUE #3151: Allow configuration of unhandled fiber error messages
This has proven to be quite annoying, and while the annoyance is probably healthy in some sense, it's also very annoying in a lot of common frameworks (notably, Ember). The easiest and most direct fix, particularly prior to 3.4, is to introduce a system property which makes it possible for users to disable this globally.

djspiewak:
Fixes #3151

Hopefully this should make the Fs2 build a bit nicer. I'm still defaulting the reporting to `true` since it's definitely good hygiene to avoid provoking the reporter in most cases, but having a global kill switch should help for situations where people are aggressively error shifting using `Deferred`.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeConfigCompanionPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeConfigCompanionPlatform.scala
@@ -44,11 +44,16 @@ private[unsafe] abstract class IORuntimeConfigCompanionPlatform { this: IORuntim
         .map(Duration(_))
         .getOrElse(DefaultShutdownHookTimeout)
 
+    val reportUnhandledFiberErrors =
+      Try(System.getProperty("cats.effect.report.unhandledFiberErrors").toBoolean)
+        .getOrElse(DefaultReportUnhandledFiberErrors)
+
     apply(
       cancelationCheckThreshold,
       autoYieldThreshold,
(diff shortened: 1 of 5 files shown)


armanbilge:
Windows CI seems to have hung :(

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeConfigCompanionPlatform.scala
```diff
@@ -44,11 +44,16 @@ private[unsafe] abstract class IORuntimeConfigCompanionPlatform { this: IORuntim
         .map(Duration(_))
         .getOrElse(DefaultShutdownHookTimeout)
 
+    val reportUnhandledFiberErrors =
+      Try(System.getProperty("cats.effect.report.unhandledFiberErrors").toBoolean)
+        .getOrElse(DefaultReportUnhandledFiberErrors)
```
armanbilge, line 49: We need to add this for the other platforms as well. edit: right? I'm confused how it compiled without them.

djspiewak: Oh that's a good point… I think it only worked because of the defaulting.

armanbilge (approved):

PR #3192, opened 2022-10-05
Add withXXX methods for easy IORuntime configuration

TimWSpence:


TimWSpence:
@armanbilge how about this? Also I don't know why CI didn't run...

armanbilge (approved):
In the long run I think `IORuntimeConfig` needs to no longer be a `case class`, because otherwise we are paying the cost of public `copy`s and `with` methods.

djspiewak:
Is this really better than just using `copy` though?

armanbilge:
Yes, if we de-case-classify `IORuntime`. `copy` methods are worse for bincompat since they need to be overloaded.

djspiewak:
Right but if we keep `IORuntime` as a case class, then this is kind of redundant, isn't it? Definitely get why `copy` is a pain for bincompat.

armanbilge:
> Right but if we keep IORuntime as a `case class`

Which is terrible for bincompat :) For example `case class`es generate mirrors on Scala 3 which we would have to be hand-maintained each time we change it.

Indeed, I suspect we've already broken it and our over-aggressive filters are just not telling us.

armanbilge:
Hmm, I take that back. I guess the mirrors can evolve bincompatibly without intervention (edit: although they may throw runtime errors 🤔 ). Also I am concerned about the `unapply` methods, but either those are fine or are slipping through due to type erasure. In any case, we don't have any filters for `IORuntimeConfig` /shrug

TimWSpence:
Sorry to only just get to this. Where did we land with this? Shall I just close it?

djspiewak:
Okay coming back to this… (finally)

I think let's close this for now. If we un-case-class-ify `IORuntimeConfig`, we'll revive this approach.

PR #3195, opened 2022-10-08
Tracing for Scala Native

### LINKED ISSUE #3167: Tracing for Scala Native
Like, serious bounty for whoever figures this out 😛 cause I'm getting hella annoyed trying to debug stack traces like this:

```
java.io.IOException: -97
    at java.lang.StackTrace$.currentStackTrace(Unknown Source)
    at java.lang.Throwable.fillInStackTrace(Unknown Source)
    at fs2.io.uring.Uring$$anon$1.$anonfun$apply$8(Unknown Source)
    at fs2.io.uring.Uring$$anon$1$$Lambda$4.apply(Unknown Source)
    at cats.ApplicativeError.$anonfun$raiseWhen$1(Unknown Source)
    ...
```

The lack of line numbers hurts too, but Anton is already working on that with his LSPs 💪 

In all honesty I have not really investigated this. The important pieces are:
1. Some strategy for uniquely identifying lambdas. On the JVM we use its class, on JS its source code. It is fully possible we could use the class on Native too, I am not sure how they are represented in the runtime. Figuring this out is the most important thing.
2. Some data structure for caching traces. The JVM uses [`ClassValue`](https://docs.oracle.com/javase/8/docs/api/java/lang/ClassValue.html) which is currently not available on Native. I do not know if that's a fundamental limitation or just that nobody has bothered to implement it. Also a vanilla map may work perfectly fine here: unlike the JVM I do not think there is a notion of lazy class loading/unloading at runtime.
3. Filtering stack traces. Good luck with that 😇 this isn't hard so much as it is annoying. For Scala Native it is complicated by the fact there are a hundred-and-one runtimes out in the wild.

In fact, if we figure out tracing, I'm pretty sure we will get fiber dumps for free since all that machinery is already in place. Edit: except for the signal handlers, lol.

armanbilge:
Closes https://github.com/typelevel/cats-effect/issues/3167.

Waiting on:
- https://github.com/scala-native/scala-native/pull/2886

I'm pretty confident we could do something more efficient here, by directly using libunwind to precisely find the information we want instead of building entire stack traces. However, at this point I'm happier to optimize for code reuse and like JS completely disable and elide tracing in release mode.


core/native/src/main/scala/cats/effect/unsafe/FiberMonitorCompanionPlatform.scala
```diff
@@ -0,0 +1,26 @@
...
+package cats.effect.unsafe
+
+import scala.concurrent.ExecutionContext
+
+private[unsafe] trait FiberMonitorCompanionPlatform {
+  def apply(compute: ExecutionContext): FiberMonitor = {
+    val _ = compute
+    new NoOpFiberMonitor
+  }
+}
```
armanbilge, line 26: The only reason we don't enable this on Native is because we have no way to trigger a fiber dump. If we bake the epollcat I/O integrated runtime directly into Cats Effect, we should be able to listen on signals as well and implement this functionality.

core/shared/src/main/scala/cats/effect/tracing/Tracing.scala
```diff
@@ -80,6 +84,9 @@ private[effect] object Tracing extends TracingPlatform {
       val callSiteMethodName = callSite.getMethodName
       val callSiteFileName = callSite.getFileName
 
+      if (callSiteClassName == "scala.scalanative.runtime.package$")
+        return null // short-circuit, effective end of stack
```
armanbilge: On JVM and JS, all stack traces ultimately end at the `ExecutionContext` that ran that task.

On Native, the stack traces can extend far beyond that. Everything is driven by `scala.scalanative.runtime.package$.loop`, which in turn is directly invoked by test frameworks such as specs2, munit, and Weaver. Rather than attempt to filter the frames of every test framework, we simply short circuit and stop traversing the stack at this point.

tests/shared/src/test/scala/cats/effect/IOFiberSpec.scala
```diff
@@ -39,9 +41,9 @@ class IOFiberSpec extends BaseSpec with DetectPlatform {
       }
 
       "toString a suspended fiber" in real {
-        def foreverNever = IO.async_[Unit](_ => ())
+        def foreverNever = IO.async_[Unit](_ => "this is a unique lambda".pure[Eval].void.value)
```
armanbilge: This is interesting :) without this change we were somehow sharing a cached trace for some other `_ => ()` out there.

armanbilge:
✨ immense frustration ✨

djspiewak:
This is pretty dope tbh. Do we have a warm fuzzy on the performance impact?

armanbilge:
I published `3.4.1-35-50f0fec` for Native/Scala 3 and benchmarked the [ember example](https://github.com/armanbilge/scala-native-ember-example) in debug mode.

| version | tracing | rps |
|-|-|-|
| `3.4.0` | n/a | 2344.2556 |
| `3.4.1-35-50f0fec` | none | 2353.4016 |
| `3.4.1-35-50f0fec` | cached | 1981.1968 |
| `3.4.1-35-50f0fec` | full | 35.0624 |

No benchmarks for release mode because tracing is dead-code-eliminated.

djspiewak:
The CI failure actually looks like a legit need to adjust `IOFiberSpec` on native.

armanbilge:
> The CI failure actually looks like a legit need to adjust IOFiberSpec on native.

But the kicker is, if I run that spec in isolation it actually passes as-written :) there's something weird going on when running the entire test suite that causes it to change.

armanbilge:
Okayyy figured it out. In the end no weirdness, no bugs, nothing. Just had to tweak the filters. Same old story as always 🤦

djspiewak (approved):

PR #3211, opened 2022-10-22
Fallback to concurrent queue for large capacities

armanbilge:


djspiewak (changes requested):
Can we add a quick test which attempts to create a `Queue.bounded(Int.MaxValue)`?

djspiewak (approved):

PR #3212, opened 2022-10-24
Add `IOLocal#scope`

iRevive:
The follow-up to https://github.com/typelevel/cats-effect/pull/3128.  The original PR has been stale for some time :(

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOLocal.scala
+++ b/core/shared/src/main/scala/cats/effect/IOLocal.scala
@@ -16,6 +16,8 @@
 
 package cats.effect
 
+import cats.syntax.functor._
+
 /**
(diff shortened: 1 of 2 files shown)


iRevive:
It seems this change is binary breaking. 

Is it safe to add `ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.effect.IOLocal.scope")` to MiMa exclusions? Or do I need to add `trait IOLocalBincompat1` instead?

armanbilge:
It is safe to add the exclusion, because `IOLocal` is a `sealed trait`. That means we can add abstract methods to it without breaking implementors, since we control all the implementations.

iRevive:
Hm, for some reason, GitHub does not recognize my changes in this branch anymore. I made a new PR instead - #3214.

PR #3214, opened 2022-10-25
Add `IOLocal#scope`

iRevive:
The follow-up to https://github.com/typelevel/cats-effect/pull/3128. The original PR has been stale for some time :(

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOLocal.scala
+++ b/core/shared/src/main/scala/cats/effect/IOLocal.scala
@@ -16,6 +16,8 @@
 
 package cats.effect
 
+import cats.syntax.functor._
+
 /**
  * [[IOLocal]] provides a handy way of manipulating a context on different scopes.
  *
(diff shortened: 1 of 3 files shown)


djspiewak (approved):
Thank you!

core/shared/src/main/scala/cats/effect/IOLocal.scala
```diff
@@ -188,6 +190,27 @@ sealed trait IOLocal[A] {
...
+   *     _     <- local.get // returns 42
+   *     _     <- local.scope(0).surround(local.getAndSet(1)) // returns 0
+   *     _     <- local.get // returns 42, even though 1 was set inside of the resource
+   *   } yield ()
+   * }}}
+   *
+   * @param value
+   *   the value to make a scope with
+   */
+  def scope(value: A): Resource[IO, Unit]
```
armanbilge, line 212: Since this is as-yet unreleased: what if we made it `scope(f: A => A)` so it's a bit more general?

iRevive: Sounds good. I can update the signature of the method. The PR should target `series/3.x` branch, right?

armanbilge: Yep, thanks!

PR #3215, opened 2022-10-26
Fixed race condition in which canceled offers could result in lost takes

djspiewak:
I believe this fixes the issue identified in typelevel/fs2#2856. Basically the situation here is where an `offer` was canceled *after* `take` passed its `Deferred` but before that `Deferred` was completed, you could end up in a situation where the `Deferred` was lost entirely and the `take` was stuck indefinitely.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Queue.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Queue.scala
@@ -189,11 +189,22 @@ object Queue {
     def offer(a: A): F[Unit] =
       F.deferred[Deferred[F, A]] flatMap { latch =>
         F uncancelable { poll =>
-          val cleanupF = stateR update {
+          val cleanupF = stateR modify {
             case SyncState(offerers, takers) =>
-              SyncState(offerers.filter(_ ne latch), takers)
+              val fallback = latch.tryGet flatMap {
+                case Some(offerer) => offerer.complete(a).void
+                case None => F.unit
+              }
+
+              SyncState(offerers.filter(_ ne latch), takers) -> fallback
           }
 
+          /*
+           * Okay there's a bug here. If an offerer is canceled during
+           * `latch.get`, and then a taker completes the Deferred
+           * *before* `cleanupF` finishes, then the taker is lost and
+           * the result is a deadlock.
+           */
           val modificationF = stateR modify {
             case SyncState(offerers, takers) if takers.nonEmpty =>
               val (taker, tail) = takers.dequeue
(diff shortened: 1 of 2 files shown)


djspiewak:
Published as `3.4-7154d08`

std/shared/src/main/scala/cats/effect/std/Queue.scala
```diff
@@ -189,19 +189,30 @@ object Queue {
     def offer(a: A): F[Unit] =
       F.deferred[Deferred[F, A]] flatMap { latch =>
         F uncancelable { poll =>
-          val cleanupF = stateR update {
+          val cleanupF = stateR modify {
```
armanbilge, line 192: Total aside: I can always tell when you wrote some code because literally everything is infixed 😛 

But note that this is deprecated in Scala 3.
> Infix operations involving alphanumeric operators are deprecated

https://dotty.epfl.ch/docs/reference/changed-features/operators.html

So it's all good and well for now, but someday someone is going to have to clean all of this up 😩  at the very least, it will add noise to a git blame.

std/shared/src/main/scala/cats/effect/std/Queue.scala
```diff
@@ -189,19 +189,30 @@ object Queue {
...
+
+              SyncState(offerers.filter(_ ne latch), takers) -> fallback
           }
 
+          /*
+           * Okay there's a bug here. If an offerer is canceled during
+           * `latch.get`, and then a taker completes the Deferred
+           * *before* `cleanupF` finishes, then the taker is lost and
+           * the result is a deadlock.
+           */
```
armanbilge: Um? 😅

std/shared/src/main/scala/cats/effect/std/Queue.scala
djspiewak: It frustrates me that it's deprecated in Scala 3 because that *forces* you to do something like `}.map { ...}` when you chain things, and the associativity of that `.` varies depending on whether it's a block or a lambda, which in turn breaks some of the symmetry of the call-site syntax previously in the chain. Overall it's a mess, and I disagree with the deprecation decision, so I'm expressing my rebellion in this fashion.

std/shared/src/main/scala/cats/effect/std/Queue.scala
djspiewak: Oh, stale comment. I need to remove that. I write comments to myself when I have to stop working on something right after I figure something out.

std/shared/src/main/scala/cats/effect/std/Queue.scala
armanbilge: Ooh yeah I very much dislike the `}.map` as well. In theory we can mark all these methods as `infix`.

Otherwise, if you want to rebel in a compatible way, consider writing `` `map` `` and `` `modify` `` and such 😇

armanbilge (approved):

PR #3218, opened 2022-10-28
Implement immutable-Map backed `MapRef`s with `Ref.lens`

armanbilge:
I think this is a nice simplification that also allows us to relax the typeclass constraints in numerous places.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/MapRef.scala
+++ b/std/shared/src/main/scala/cats/effect/std/MapRef.scala
@@ -221,9 +221,15 @@ object MapRef extends MapRefCompanionPlatform {
    *
    * This uses universal hashCode and equality on K.
    */
-  def fromSingleImmutableMapRef[F[_]: Concurrent, K, V](
+  def fromSingleImmutableMapRef[F[_]: Functor, K, V](
       ref: Ref[F, Map[K, V]]): MapRef[F, K, Option[V]] =
-    new ShardedImmutableMapImpl[F, K, V](_ => ref)
+    k => Ref.lens(ref)(_.get(k), m => _.fold(m - k)(v => m + (k -> v)))
+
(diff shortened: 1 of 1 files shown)


std/shared/src/main/scala/cats/effect/std/MapRef.scala
```diff
@@ -211,19 +119,32 @@ object MapRef extends MapRefCompanionPlatform {
...
+  def fromSingleImmutableMapRef[F[_]: Functor, K, V](
       ref: Ref[F, Map[K, V]]): MapRef[F, K, Option[V]] =
-    new ShardedImmutableMapImpl[F, K, V](_ => ref)
+    k => Ref.lens(ref)(_.get(k), m => _.fold(m - k)(v => m + (k -> v)))
+
+  @deprecated("Use override with Functor constraint", "3.4.0")
+  def fromSingleImmutableMapRef[F[_], K, V](
+      ref: Ref[F, Map[K, V]],
+      F: Concurrent[F]): MapRef[F, K, Option[V]] =
+    fromSingleImmutableMapRef(ref)(F)
```
armanbilge: Technically we can just remove these, but personally I think it is poor taste to break bincompat with an RC.

Since 3.4.0 is frozen there is also an argument to retarget these changes to `main`.

iRevive:
Being curious: is it necessary to go through deprecate->remove cycle if the method has never been released?

`MapRef` does not exist in the 3.3.x series. And I assume, it does not exist in any major release yet.

djspiewak:
First off, this is a very good change. :-) Second off, it's not worth the bother to `@deprecate` the overloads. Normally I do agree that breaking bincompat with RCs is bad, but this is such a minor case that it isn't worth the long-term baggage.

djspiewak (approved):

PR #3219, opened 2022-10-30
Integrated timers

djspiewak:
Dusted off @vasilmkd's old branch (first PRed in #2252) and merged it with the latest head. Still needs some work:

- The fallback `sleep` implementation in `WorkStealingThreadPool` references `IORuntime.global` just to get things to compile. This doesn't strike me as strictly necessary given the way that `sleepInternal` works, but trivially removing it got me into trouble.
- The sleep cancelation action *appears* to just be a `lazySet(false)` on `SleepCallback`, which is concerning because it leaks memory. I believe this is why the fiber dump specs are failing.

The runtime has changed meaningfully since Vasil's original implementation, so this definitely needs a bit of careful thought to ensure that we aren't doing anything weird or conflicty. Notably, this implementation simply doesn't implement timer stealing at all, and instead all timers are held local to their scheduling thread. We're going to need to work to validate the hypothesis that we can get away with this. If we need to implement theft, things become a *lot* more complex.

This is the first step in moving us toward a fully integrated runtime.


armanbilge:
> * The sleep cancelation action _appears_ to just be a `lazySet(false)` on `SleepCallback`, which is concerning because it leaks memory. I believe this is why the fiber dump specs are failing.

Btw, I have an alternative `SleepCallback` implementation in the native implementation. Might be interesting if we can unify those.

https://github.com/typelevel/cats-effect/blob/73d63c8f301df317efe39dcc7d12b79914aa9455/core/native/src/main/scala/cats/effect/unsafe/PollingExecutorScheduler.scala#L139-L152

djspiewak:
Published as `3.5-639ac01`

djspiewak:
Ran a quick test on Ember. With this change, peak RPS on a basic GET request improved by a little over 25%.

armanbilge:
Um ... holy smokes!

djspiewak:
Did a little more testing. A slightly less trivial test involving a POST body and some JSON parsing (using Circe) shows peak RPS improvements around 14%, which makes intuitive sense since that test is going to be a little more bounded by the body processing than by the pure connection overhead. Still, 25% improvements in trivial GET peak RPS is pretty damn great.

P99 latencies in both cases were improved by 13.5%. I think this number is probably a bit more trustworthy, and also still very impressive IMO.

djspiewak:
Did a quick run through the `WorkStealingBenchmark`. It hasn't regressed even a small amount. I suspect this is because the JIT is simply inlining away the `sleepers.nonEmpty` check (since it's always false in those benchmarks). In a real case with real timers, I would expect some regression in straight-line performance, but obviously any such case would also benefit commensurately from this very change. Either way, I think it's safe to say this is ready for serious review.

We can do a lot better than the implementation as it stands, but that can be incrementally layered onto this once it lands.

djspiewak:
TODO

- [x] Figure out why the CI keeps hanging (was able to reproduce locally but not minimize)
- [x] Make this work with `blocking` (and add tests)

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -514,6 +522,52 @@ private[effect] final class WorkStealingThreadPool(
    */
   override def reportFailure(cause: Throwable): Unit = reportFailure0(cause)
 
+  override def monotonicNanos(): Long = System.nanoTime()
+
+  override def nowMillis(): Long = System.currentTimeMillis()
```
He-Pin, line 537: Can this be plugble ,as what if I want to make use of https://github.com/OpenHFT/Chronicle-Ticker

armanbilge: @He-Pin yes, you can always replace the `Scheduler` in the `IORuntime` with your own implementation. See also the Cats Effect test runtime, which supports time mocking.
https://typelevel.org/cats-effect/docs/core/test-runtime#mocking-time

djspiewak:
Released as snapshot `3.5-01f5b3a`

djspiewak:
A few simple canaries were promising. At least, nothing exploded. We should look more closely. In the meantime, I think this is ready for more serious review. Note that there are definitely follow-ups we can explore which will absolutely micro-optimize this further. We can do that once this lands in `series/3.x`, since this is, by itself, already a significant improvement.

durban:
> Notably, this implementation simply doesn't implement timer stealing at all, and instead all timers are held local to their scheduling thread.

I think one scenario affected by this is when (1) a compute thread is blocked/spinwaiting on a condition (_without_ using `IO.blocking`), and (2) for that condition to become true, a timer must fire. In this case, if the timer is coincidentally on the same thread as (1), the result is a deadlock. Of course, the solution is "don't do that". True, (1) should "never" happen, but I'm pretty sure it does. Together with (2)... I'm not sure how common it is...

In any case, this is a possible (if contrived) scenario, from which the current system can recover eventually (the timers are independent), but with this PR it can deadlock. (With timer stealing, I think it would also recover eventually, because the timer in question would get stolen.) I'm not sure how important supporting badly behaved code like this is, I just wanted to mention it.

djspiewak:
@durban So you're thinking about something like this?

```scala
val flag = new AtomicBoolean(false)
IO(flag.set(true)).delayBy(2.seconds) &> IO(while (!flag.get()) {})
```

I agree that, without timer stealing, the above can hang forever if the timer happens to be on the same thread as the `while` loop. However, just to be clear, if you `replicateA_(100)` the above, you will create a livelock even on the current version of Cats Effect. Locking up worker threads in this fashion is just… very bad. :-)

durban:
@djspiewak Yeah, something like that. The `replicateA_` example is good, because it shows that currently it just _seems_ to work. So yeah, "don't do that" is the solution.

core/jvm/src/main/scala/cats/effect/unsafe/SleepCallback.scala
```diff
@@ -0,0 +1,79 @@
...
+        now + delayNanos
+      else
+        now + overflowFree(delayNanos, now)
+    }
+
+    new SleepCallback(triggerTime, callback)
+  }
+
+  implicit val sleepCallbackReverseOrdering: Ordering[SleepCallback] =
+    Ordering.fromLessThan(_.triggerTime > _.triggerTime)
```
durban, line 78:
```suggestion
    Ordering.fromLessThan(_.triggerTime - _.triggerTime > 0)
```

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -368,6 +392,27 @@ private final class WorkerThread(
...
+      val sleepers = sleepersQueue
+
+      if (sleepers.nonEmpty) {
+        val now = System.nanoTime()
+
+        var cont = true
+        while (cont) {
+          val head = sleepers.head()
+
+          if (head.triggerTime <= now) {
```
durban, line 404:
```suggestion
          if (head.triggerTime - now <= 0) {
```

durban (changes requested):

armanbilge:
Btw, I had to make a small patch on my JVM polling branch in https://github.com/typelevel/cats-effect/pull/3332/commits/e44a8023780da347a4e74a0b78ba581847e118eb.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -317,6 +327,19 @@ private final class WorkerThread(
       }
     }
 
+    def parkUntilNextSleeper(): Unit = {
+      if (!isInterrupted()) {
+        val now = System.nanoTime()
+        val head = sleepersQueue.head()
+        val nanos = head.triggerTime - now
```
armanbilge, line 334: cherry-pick https://github.com/typelevel/cats-effect/pull/3332/commits/e44a8023780da347a4e74a0b78ba581847e118eb
```suggestion
        val nanos = Math.max(head.triggerTime - now, 0)
```

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -903,7 +903,15 @@ private final class IOFiber[A](
...
             IO {
-              val cancel = runtime.scheduler.sleep(cur.delay, () => cb(RightUnit))
+              val scheduler = runtime.scheduler
+              val delay = cur.delay
+
+              val cancel =
+                if (scheduler.isInstanceOf[WorkStealingThreadPool])
+                  scheduler.asInstanceOf[WorkStealingThreadPool].sleepInternal(delay, cb)
+                else
+                  scheduler.sleep(delay, () => cb(RightUnit))
```
He-Pin, line 913: use match?

core/jvm/src/main/scala/cats/effect/unsafe/SleepCallback.scala
```diff
@@ -0,0 +1,79 @@
...
+ * limitations under the License.
+ */
+
+package cats.effect.unsafe
+
+import scala.concurrent.duration.FiniteDuration
+
+import java.util.concurrent.atomic.AtomicBoolean
+
+private final class SleepCallback private (
```
He-Pin, line 23: ScheduledCallback?

core/shared/src/main/scala/cats/effect/IOFiber.scala
djspiewak: Actually `match` is often slower because of how it gets compiled. The JIT fast-paths the sequential combination of a conditional jump branching on `isInstanceOf`, followed *immediately* by a dynamic cast, and it turns that into a single operation on most architectures. `match` is more declarative but can easily generate bytecode which messes up this JIT optimization, so in very hot-path code we tend to be more explicit about it.

core/jvm/src/main/scala/cats/effect/unsafe/SleepCallback.scala
djspiewak: `ScheduledCallback` might be a better name here.

PR #3223, opened 2022-11-03
Optimize `NonFatal(...)` checks

armanbilge:
See bytecode analysis in:
- https://github.com/typelevel/cats/pull/4282

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/IOApp.scala
+++ b/core/jvm/src/main/scala/cats/effect/IOApp.scala
@@ -208,7 +208,7 @@ trait IOApp {
     new ExecutionContext {
       def reportFailure(t: Throwable): Unit =
         t match {
-          case NonFatal(t) =>
+          case t if NonFatal(t) =>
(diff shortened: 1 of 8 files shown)


djspiewak (approved):
Nice catch! I'll run some benchmarks on this quickly before we merge.

djspiewak:
Did a pair of benchmark runs ([before](https://www.dropbox.com/sh/x2rbviejkpm2q5t/AABdAxAXonfg75VKE-Ienlbfa?dl=1), [after](https://www.dropbox.com/s/gfhe5y1skjwibel/1-after.json?dl=1)). You can toss these into https://jmh.morethan.io/ to see things visually, but the tldr is that this PR results in absolutely zero change in performance outside the margin of error (there's some noise from run to run, and notably the `parTraverse` benchmark appears to have been bad in the first run). I find this fascinating, because it suggests that the JVM's escape analysis is eliding the allocation cost of the `NonFatal` extractor once warmed up.

PR #3225, opened 2022-11-05
Introduce a `BatchingMacrotaskExecutor`

armanbilge:
This one will definitely require some kind of benchmarking :)

This is inspired by the timer+I/O-integrated runtime work, where we only check for outstanding timers and I/O every 64 iterations of the runloop. Meanwhile the `MacrotaskExecutor` is based on `setImmediate`, which gives priority to I/O events.

> Schedules the "immediate" execution of the callback after I/O events' callbacks.

https://nodejs.org/api/timers.html#setimmediatecallback-args

Cats Effect and downstream libraries are very fiber-happy so it seems a bit outrageous that every started fiber must wait for an entire event-loop cycle before it runs. This is compounded by the fact that in browsers we generally can't even submit directly to the macrotask queue and rely on hacks such as `postMessage`.

Indeed, this is the usual performance vs fairness tradeoff. But I am still unconvinced about the importance of fairness in JS.

1. In a JS lambda, you are limited to processing one request at a time. So it's not really obvious to me what there is to be unfair to, and how that might be observed.

2. In the browser, lack of fairness is most easily observed as delayed rendering. But if you are doing so much work in response to some event (e.g. user input or websocket message) that it affects rendering you are probably in trouble anyway. Increasing fairness will at best enable progressive but glitchy rendering and at worst make no difference.

Ironically the `MacrotaskExecutor` is about as fair as it gets 😅


core/js/src/main/scala/cats/effect/unsafe/BatchingMacrotaskExecutor.scala
```diff
@@ -0,0 +1,63 @@
...
+import scala.concurrent.ExecutionContextExecutor
+import scala.util.control.NonFatal
+
+import java.util.ArrayDeque
+
+private final class BatchingMacrotaskExecutor(pollEvery: Int) extends ExecutionContextExecutor {
+
+  private[this] var needsReschedule: Boolean = true
+
+  private[this] val executeQueue: ArrayDeque[Runnable] = new ArrayDeque
```
armanbilge: Performance will be sub-optimal until we get a Scala.js release with this.
- https://github.com/scala-js/scala-js/pull/4736

core/js/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
```diff
@@ -18,18 +18,18 @@ package cats.effect.unsafe
...
-import org.scalajs.macrotaskexecutor.MacrotaskExecutor
-
 import scala.concurrent.ExecutionContext
 import scala.scalajs.LinkingInfo
 
 private[unsafe] abstract class IORuntimeCompanionPlatform { this: IORuntime.type =>
 
-  def defaultComputeExecutionContext: ExecutionContext =
+  def defaultComputeExecutionContext: ExecutionContext = {
+    val ec = new BatchingMacrotaskExecutor(64)
```
armanbilge: We should make this configurable.

djspiewak:
I'm skeptical! I'll take a closer look soon, but in general, I'd rather tune the fairness/throughput coefficient within `IOFiber` itself, rather than at the executor level.

armanbilge:
> I'm skeptical! I'll take a closer look soon, but in general, I'd rather tune the fairness/throughput coefficient within IOFiber itself, rather than at the executor level.

I don't see how that's possible here.

djspiewak:
Okay I have an idea: what if we simply let `IOFiber` handle this? We're basically saying that there's a certain class of operations (e.g. `start`) which we would like to execute with microtask semantics, up to some bound, and we want to do this just on JS platforms. In particular, we can either make `scheduleFiber` platform-specific (which could allow us to remove WSTP from the JS binary), or we could make a special variant of it which is itself platform-specific. JVM and Native behavior should be unaffected.

armanbilge:
> We're basically saying that there's a certain class of operations (e.g. start) which we would like to execute with microtask semantics, up to some bound

Yes, exactly. but how do we implement this bound? It seems like it requires some "global" state (at least from the perspective of an individual fiber). Which brings us to the `ExecutionContext`.

djspiewak:
Oh yes, we do implement a wrapping EC like this one, but we don't batch all `execute`s. Instead, we have an additional method which adds to the batch and is called by the `Start` interpreter.

armanbilge:
Ahhh, I gotcha now. Makes sense, thanks!

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -1272,9 +1272,12 @@ private final class IOFiber[A](
...
-    if (ec.isInstanceOf[WorkStealingThreadPool]) {
+    if (Platform.isJvm && ec.isInstanceOf[WorkStealingThreadPool]) {
       val wstp = ec.asInstanceOf[WorkStealingThreadPool]
       wstp.execute(fiber)
+    } else if (Platform.isJs && ec.isInstanceOf[BatchingMacrotaskExecutor]) {
+      val bmte = ec.asInstanceOf[BatchingMacrotaskExecutor]
+      bmte.schedule(fiber)
     } else {
       scheduleOnForeignEC(ec, fiber)
     }
```
armanbilge: So I know this is not really what you suggested but on the other hand this lets us avoid unintentionally de-optimizing a `private[this]` method. Here's what this decompiles to on the JVM:

```java
    private void scheduleFiber(ExecutionContext ec, IOFiber<?> fiber) {
        if (true && ec instanceof WorkStealingThreadPool) {
            WorkStealingThreadPool wstp = (WorkStealingThreadPool)ec;
            wstp.execute(fiber);
        } else if (false && ec instanceof BatchingMacrotaskExecutor) {
            BatchingMacrotaskExecutor bmte = (BatchingMacrotaskExecutor)ec;
            bmte.schedule(fiber);
        } else {
            this.scheduleOnForeignEC(ec, fiber);
        }
    }
```

The downside is now `BatchingMacrotaskExecutor` is living in JVM/Native binaries ...

armanbilge:
So I'm giving this a try. Still need to take another pass adding some internal scaladocs and maybe bikeshedding the names.

I think every operation that `scheduleFiber()` is used for makes sense for batching via the microtask (promises) executor, namely: `start`, `racePair`, and resuming an `async`. Meanwhile, (auto-)ceding is implemented with `rescheduleFiber()` and this should definitely go through the macrotask executor, so that it actually cedes to the event loop :)

So I think this should work well for the typical scenario where an incoming UI/IO event is picked up by either an `async` or `Dispatcher`, and starts/resumes a handful of short-running fibers to process the event.

core/js/src/main/scala/cats/effect/unsafe/BatchingMacrotaskExecutor.scala
```diff
@@ -0,0 +1,45 @@
...
+  def execute(runnable: Runnable): Unit =
+    MacrotaskExecutor.execute(runnable)
+
+  def schedule(runnable: Runnable): Unit = {
+    if (counter % batchSize == 0)
+      MacrotaskExecutor.execute(runnable)
+    else
+      MicrotaskExecutor.execute(runnable)
+    counter += 1
+  }
```
armanbilge: Actually this logic is not right 🤔 maybe I need to keep the original implementation that maintains its own queue.

core/js/src/main/scala/cats/effect/Platform.scala
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+
+private object Platform {
```
djspiewak, line 19: This is an anti-pattern that I would very much rather avoid in **main** sources (it's fine, and very convenient, in tests). The problem is that this results in bytecode (and ultimately, assembly code) which is less efficient on all platforms because of the existence of others. Relying on the `FooPlatform` and `FooCompanionPlatform` idiom instead is more efficient and more uniform with the rest of the codebase, even though it does require some adjustments to encoding sometimes.

core/js/src/main/scala/cats/effect/unsafe/BatchingMacrotaskExecutor.scala
```diff
@@ -20,44 +20,26 @@ package unsafe
...
 
-import java.util.ArrayDeque
+private[effect] final class BatchingMacrotaskExecutor(batchSize: Int)
+    extends ExecutionContextExecutor {
 
-private final class BatchingMacrotaskExecutor(pollEvery: Int) extends ExecutionContextExecutor {
+  private[this] val MicrotaskExecutor = QueueExecutionContext.promises()
 
-  private[this] var needsReschedule: Boolean = true
+  private[this] var counter = 0
```
djspiewak: I'm almost positive it would be faster if we just have an explicit buffer with a static size. Promises are fast, but it's hard to be faster than an `Array[Runnable]` that you just iterate over.

core/jvm-native/src/main/scala/cats/effect/unsafe/BatchingMacrotaskExecutor.scala
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+package unsafe
+
+import scala.concurrent.ExecutionContext
+
+private[effect] sealed abstract class BatchingMacrotaskExecutor private ()
```
djspiewak, line 22: Another disadvantage of the `Platform` encoding used above is this type now leaks across to JVM and Native.

core/shared/src/main/scala/cats/effect/IOFiber.scala
djspiewak: This is a pretty big deoptimization since it makes the branching much more complex. It's *possible* that the constants together with JIT inlining cause the ultimately-emitted assembly to elide the impossible branch, but I would want to check that before we really trust it. We can avoid this by relying on `IOFiberPlatform` instead.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -1272,9 +1272,12 @@ private final class IOFiber[A](
   }
 
   private[this] def scheduleFiber(ec: ExecutionContext, fiber: IOFiber[_]): Unit = {
```
djspiewak, line 1281: Isn't `scheduleFiber` used in other contexts? Or is it *only* used in `start`?

djspiewak (changes requested):

core/shared/src/main/scala/cats/effect/IOFiber.scala
armanbilge: Ok, I rearranged the platform checks and now the emitted bytecode is completely identical to what it was before. I understand objections based on style / increased binary surface, but I do not see how this is not the maximally optimized implementation.

```java
    private void scheduleFiber(ExecutionContext ec, IOFiber<?> fiber) {
        if (ec instanceof WorkStealingThreadPool) {
            WorkStealingThreadPool wstp = (WorkStealingThreadPool)ec;
            wstp.execute(fiber);
        } else {
            this.scheduleOnForeignEC(ec, fiber);
        }
    }
```

core/js/src/main/scala/cats/effect/Platform.scala
armanbilge: > The problem is that this results in bytecode (and ultimately, assembly code) which is less efficient on all platforms because of the existence of others.

This is not true at all :) the scala _compiler_ completely dead-code-eliminates the false branches. They never even reach the bytecode.

Meanwhile, I am skeptical that inheriting methods from `FooPlatform` is nearly as efficient. For example, said methods can no longer be `private[this]`.

core/shared/src/main/scala/cats/effect/IOFiber.scala
armanbilge: See my justification in https://github.com/typelevel/cats-effect/pull/3225#issuecomment-1328617151.

core/js/src/main/scala/cats/effect/unsafe/BatchingMacrotaskExecutor.scala
armanbilge: Yeah, I don't think this was doing the right thing anyway. I went back to my original implementation which is based on an `ArrayDeque`.

armanbilge:
Ok, I think this is ready for another look, thanks for all the pointers. No rush, since it has to wait until 3.5.x because it needs a Scala.js upgrade. (Unless there's something wrong with my approach, and we can do this without the `ArrayDeque`. Update: I ended up removing the `ArrayDeque` because it has issues ...)

The only outstanding issue is whether to do the platforming with compile-time conditionals or platform traits.

armanbilge:
I benchmarked a warmed-up Ember.js "hello world" server with these changes.

Under concurrent load, performance is roughly similar.

When considering a single connection, the improvement is prominent: 60% higher RPS.

This makes sense: in the single connection case, there are never any other I/O events besides the one being currently handled. So ceding to the event loop during processing of the I/O event is pointless.

Meanwhile, in the concurrent connection case, nothing suggests that fairness has been compromised. This is consistent with the pattern where reacting to an I/O event starts a small number of short-lived fibers.

### 50 concurrent connections

<details>

#### 3.4.2
```
Summary:
  Total:	30.0618 secs
  Slowest:	8.6726 secs
  Fastest:	0.0125 secs
  Average:	0.0966 secs
  Requests/sec:	517.2019
```

#### 3.5-88faeed
```
Summary:
  Total:	30.0469 secs
  Slowest:	6.6952 secs
  Fastest:	0.0022 secs
  Average:	0.0884 secs
  Requests/sec:	564.7503
```

</details>

### 1 connection

<details>

#### 3.4.2
```
Summary:
  Total:	30.0012 secs
  Slowest:	0.0561 secs
  Fastest:	0.0022 secs
  Average:	0.0031 secs
  Requests/sec:	322.8537
```

#### 3.5-88faeed
```
Summary:
  Total:	30.0012 secs
  Slowest:	0.0244 secs
  Fastest:	0.0013 secs
  Average:	0.0020 secs
  Requests/sec:	499.9141
```

</details>

core/js/src/main/scala/cats/effect/unsafe/JSArrayQueue.scala
```diff
@@ -0,0 +1,72 @@
...
+package cats.effect.unsafe
+
+import scala.scalajs.js
+
+/**
+ * A JS-Array backed circular buffer FIFO queue. It is careful to grow the buffer only using
+ * `push` to avoid creating "holes" on V8 (this is a known shortcoming of the Scala.js
+ * `j.u.ArrayDeque` implementation).
+ */
+private final class JSArrayQueue[A] {
```
armanbilge, line 26: - https://github.com/scala-js/scala-js/pull/4736#discussion_r1070168777

armanbilge:
Besides the happy-path performance improvements, it's worth pointing out that this new EC helps mitigate other performance issues:

- We no longer have to wrap every fiber to support fiber dumps, since the EC natively supports this.
- If you somehow are in an environment where macrotasks are subject to the 4ms+ clamping, then there is no longer a clamping overhead _for each fiber_. Indeed, if each UI or I/O event spawns a small batch of short-lived fibers, clamping would not even be triggered.

djspiewak (approved):
Still very uncomfortable with the `Platform` trait but I'll try to hit that in a follow-up.

PR #3226, opened 2022-11-06
Release loser eagerly in `Resource.race`

### LINKED ISSUE #3111: Resource#race: late looser release
Consider a snippet:
```scala
import cats.syntax.all._
import cats.effect._
import cats.effect.std._

object ResourceAllocRace extends IOApp.Simple {
  private val con = Console[IO]

  private def mkResource(name: String, random: Random[IO]): Resource[IO, String] =
    Resource.make {
      con.println(s"... [$name]: connecting") >>
        random.nextIntBounded(3000).flatMap(ms => IO.sleep(ms.millis)) >>
        con.println(s">>> [$name]: CONNECTED")
    }(_ => con.println(s"<<< [$name]: CLOSED"))
      .as(name)

  override def run: IO[Unit] =
    Random.scalaUtilRandom[IO].flatMap { random =>
      (
        ((
          mkResource("AAA", random) race mkResource("BBB", random)
        ) race (
          mkResource("CCC", random) race mkResource("DDD", random)
        )) race ((
          mkResource("EEE", random) race mkResource("FFF", random)
        ) race (
          mkResource("GGG", random) race mkResource("HHH", random)
        ))
      ).map(_.merge.merge.merge).flatMap { winner =>
        Resource.eval(con.println(s"!!! [$winner]: WON"))
      }.use { _ =>
        con.println("... working") >>
          IO.sleep(10.seconds) >> // can be increased to any value
          con.println("... DONE")
      }
    }
}
``` 
When I run it, I may get the following output:
```
[info] running (fork) ResourceAllocRace 
[info] ... [CCC]: connecting
[info] ... [AAA]: connecting
[info] ... [EEE]: connecting
[info] ... [BBB]: connecting
[info] ... [DDD]: connecting
[info] ... [FFF]: connecting
[info] ... [GGG]: connecting
[info] ... [HHH]: connecting
[info] >>> [HHH]: CONNECTED
[info] >>> [EEE]: CONNECTED
[info] >>> [AAA]: CONNECTED
[info] >>> [FFF]: CONNECTED
[info] <<< [FFF]: CLOSED
[info] >>> [CCC]: CONNECTED
[info] >>> [DDD]: CONNECTED
[info] <<< [DDD]: CLOSED
[info] >>> [BBB]: CONNECTED
[info] <<< [BBB]: CLOSED
[info] >>> [GGG]: CONNECTED
[info] <<< [GGG]: CLOSED
[info] !!! [CCC]: WON
[info] ... working
[info] ... DONE
[info] <<< [HHH]: CLOSED
[info] <<< [EEE]: CLOSED
[info] <<< [CCC]: CLOSED
[info] <<< [AAA]: CLOSED
```

* **Expected** behavior:
only a winner resource is kept until the end of the `use` block.
* **Actual** behavior:
some of looser resources may be kept (not every time though but quite often) until the end of the `use` block along with the winner one.
* Build info:
  - Scala: 2.13.8
  - Cats Effect: 3.3.14
* System props:
  ```
  os.arch: x86_64
  os.name: Mac OS X
  os.version: 12.4
  java.vm.name: OpenJDK 64-Bit Server VM
  java.vm.vendor: Homebrew
  java.vm.version: 11.0.15+0
  ```

armanbilge:
Fixes https://github.com/typelevel/cats-effect/issues/3111

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -303,7 +303,67 @@ sealed abstract class Resource[F[_], +A] extends Serializable {
   def race[B](
       that: Resource[F, B]
   )(implicit F: Concurrent[F]): Resource[F, Either[A, B]] =
-    Concurrent[Resource[F, *]].race(this, that)
+    Resource.applyFull { poll =>
+      def cancelLoser[C](f: Fiber[F, Throwable, (C, ExitCase => F[Unit])]): F[Unit] =
+        f.cancel *>
+          f.join
+            .flatMap(
+              _.fold(
+                F.unit,
+                F.raiseError[Unit](_),
+                _.flatMap(_._2.apply(ExitCase.Canceled))
+              )
+            )
+
+      poll(F.racePair(this.allocatedCase, that.allocatedCase)).flatMap {
+        case Left((oc, f)) =>
+          oc match {
+            case Outcome.Succeeded(fa) =>
+              cancelLoser(f).start.flatMap { f =>
+                fa.map {
+                  case (a, fin) =>
+                    (
+                      Either.left[A, B](a),
+                      fin(_: ExitCase).guarantee(f.join.flatMap(_.embedNever)))
+                }
+              }
+            case Outcome.Errored(ea) =>
+              cancelLoser(f) *> F.raiseError[(Either[A, B], ExitCase => F[Unit])](ea)
+            case Outcome.Canceled() =>
+              poll(f.join).onCancel(f.cancel).flatMap {
+                case Outcome.Succeeded(fb) =>
+                  fb.map { case (b, fin) => (Either.right[A, B](b), fin) }
+                case Outcome.Errored(eb) =>
+                  F.raiseError[(Either[A, B], ExitCase => F[Unit])](eb)
+                case Outcome.Canceled() =>
+                  poll(F.canceled) *> F.never[(Either[A, B], ExitCase => F[Unit])]
+              }
+          }
+        case Right((f, oc)) =>
+          oc match {
+            case Outcome.Succeeded(fb) =>
+              cancelLoser(f).start.flatMap { f =>
+                fb.map {
+                  case (b, fin) =>
+                    (
+                      Either.right[A, B](b),
+                      fin(_: ExitCase).guarantee(f.join.flatMap(_.embedNever)))
+                }
+              }
+            case Outcome.Errored(eb) =>
+              cancelLoser(f) *> F.raiseError[(Either[A, B], ExitCase => F[Unit])](eb)
+            case Outcome.Canceled() =>
+              poll(f.join).onCancel(f.cancel).flatMap {
+                case Outcome.Succeeded(fa) =>
+                  fa.map { case (a, fin) => (Either.left[A, B](a), fin) }
+                case Outcome.Errored(ea) =>
+                  F.raiseError[(Either[A, B], ExitCase => F[Unit])](ea)
+                case Outcome.Canceled() =>
+                  poll(F.canceled) *> F.never[(Either[A, B], ExitCase => F[Unit])]
+              }
+          }
+      }
+    }
 
   /**
    * Implementation for the `flatMap` operation, as described via the `cats.Monad` type class.
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -310,7 +310,7 @@ sealed abstract class Resource[F[_], +A] extends Serializable {
             .flatMap(
               _.fold(
                 F.unit,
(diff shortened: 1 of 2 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -303,7 +303,67 @@ sealed abstract class Resource[F[_], +A] extends Serializable {
...
+          oc match {
+            case Outcome.Succeeded(fa) =>
+              cancelLoser(f).start.flatMap { f =>
+                fa.map {
+                  case (a, fin) =>
+                    (
+                      Either.left[A, B](a),
+                      fin(_: ExitCase).guarantee(f.join.flatMap(_.embedNever)))
+                }
+              }
```
armanbilge, line 329: Unlike racing `IO` where you don't get access to the result of the winner until after cancelation of the loser completes, here we can provide the result immediately while canceling the loser concurrently. This is possible because we can hitch the backpressure on loser cancelation to the resource finalizer for the winner.

djspiewak:
So the thing that makes me very nervous here is the fact that `race` was the motivating case for the behavior of `start` on `Resource`, so if `race`'s semantics *don't* fall out naturally, it feels like we missed something. I definitely get why this PR fixes things, but do we have an intuition for *why* the behavior in #3111 is happening?

armanbilge:
> so if `race`'s semantics _don't_ fall out naturally

But what _are_ race's semantics :)

They do fall out naturally, if you don't know about resources. Specifically, that resources have a release phase. And `Spawn` does not know this.

djspiewak:
Going to punt this to 3.4.1 so I can think about it a bit more. :-)

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -303,7 +303,67 @@ sealed abstract class Resource[F[_], +A] extends Serializable {
...
+              poll(f.join).onCancel(f.cancel).flatMap {
+                case Outcome.Succeeded(fb) =>
+                  fb.map { case (b, fin) => (Either.right[A, B](b), fin) }
+                case Outcome.Errored(eb) =>
+                  F.raiseError[(Either[A, B], ExitCase => F[Unit])](eb)
+                case Outcome.Canceled() =>
+                  poll(F.canceled) *> F.never[(Either[A, B], ExitCase => F[Unit])]
+              }
+          }
+        case Right((f, oc)) =>
```
filipwiech, line 342: I wonder if the code for the `Left` and `Right` cases could be shared and extracted to a common function? 🙂

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -303,7 +303,67 @@ sealed abstract class Resource[F[_], +A] extends Serializable {
   def race[B](
       that: Resource[F, B]
   )(implicit F: Concurrent[F]): Resource[F, Either[A, B]] =
-    Concurrent[Resource[F, *]].race(this, that)
+    Resource.applyFull { poll =>
```
filipwiech, line 306: Might be worth it to describe loser release behaviour (eager and concurrent) in the scaladoc, similarly to how it's done for the `Resource#both`. 🙂

djspiewak (approved):
This genuinely makes me sad, but ultimately we don't have any type linearity to lean on to do any better.

For posterity, the argument here effectively reduces to the fourth case in `join`'s finalization semantics:

```scala
/*
   * 1. If the scope containing the *fiber* terminates:
   *   a) if the fiber is incomplete, run its inner finalizers when it completes
   *   b) if the fiber is succeeded, run its finalizers
   * 2. If the fiber is canceled or errored, finalize
   * 3. If the fiber succeeds and .cancel won the race, finalize eagerly and
   *    `join` results in `Canceled()`
   * 4. If the fiber succeeds and .cancel lost the race or wasn't called,
   *    finalize naturally when the containing scope ends, `join` returns
   *    the value
   */
```

An alternative choice that could have been made here is that the first `join` becomes the primary owner of the scope. If we had that semantic, then simply swapping `*>` for `!>` in the implementation of `race` (in `GenSpawn`) would have resolved this issue entirely, since then the losing `join` scope would be closed before the `race` itself returns.

IIRC, @wemrysi and I actually explored this a bit when we designed this semantic. A lot of the problems with it stem from the fact that `Fiber` is *effectively* a sort of `Deferred`, memoizing its result. Put more bluntly: you can call `join` repeatedly, and we don't really have any idea if the first one is the only one or the last one.

Since we have no guarantees here, `Resource#start` takes the safest possible approach, which involves a race condition between `cancel` and the inner fiber. This is deeply unfortunately, but it at least guarantees two properties: 1) you can never `use` after `free`, and 2) all results are guaranteed to be `free`ed eventually, though the exact timing on this `free` is nondeterministic in general. In a sense, this is the "garbage collector" tradeoff.

This PR effectively tightens up these guarantees in the special case of `race`.

PR #3227, opened 2022-11-06
Why do we need two `Resource` interpreters?

armanbilge:
`allocatedCase` is awesome, so I had to ask :) Still, this change doesn't pass all the `Resource` tests.
```
[info]   both
[error]     x safety: lhs error during rhs interruptible region
[error]      the value is false (ResourceSpec.scala:531)
[error]     x safety: rhs error during lhs uninterruptible region
[error]      the value is true (ResourceSpec.scala:583)
```

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -155,44 +155,17 @@ sealed abstract class Resource[F[_], +A] extends Serializable {
   private[effect] def fold[B](
       onOutput: A => F[B],
       onRelease: F[Unit] => F[Unit]
-  )(implicit F: MonadCancel[F, Throwable]): F[B] = {
-    sealed trait Stack[AA]
-    case object Nil extends Stack[A]
-    final case class Frame[AA, BB](head: AA => Resource[F, BB], tail: Stack[BB])
-        extends Stack[AA]
-
-    // Indirection for calling `loop` needed because `loop` must be @tailrec
-    def continue[C](current: Resource[F, C], stack: Stack[C]): F[B] =
-      loop(current, stack)
-
-    // Interpreter that knows how to evaluate a Resource data structure;
-    // Maintains its own stack for dealing with Bind chains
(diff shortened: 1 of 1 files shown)


djspiewak:
My guess is we don't, but I don't entirely remember.

armanbilge:
Ok, I understood the fundamental difference between the `fold` interpreter and the `allocatedCase` interpreter.

https://github.com/typelevel/cats-effect/blob/52b5a3ba1623ff40fc48318c9a0bfae558f0e5ed/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala#L155-L158

https://github.com/typelevel/cats-effect/blob/52b5a3ba1623ff40fc48318c9a0bfae558f0e5ed/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala#L387-L388

Suppose you have a nested resource e.g.
```scala
Resource.make(allocate)(release).evalMap(IO.raiseError(...))
```

When attempting to acquire this resource, you will encounter an error. If you are using the `fold` interpreter, it will invoke the user-provided callback `onRelease` with `release`. However, if you are using the `allocatedCase` interpreter, it will invoke `release` on its own. That is to say, when `allocatedCase` does not complete successfully, the user cannot get access to a "partial" finalizer, unlike `fold`.

`Resource#both` takes advantage of `fold` to store the (partial) finalizers and run them concurrently:
https://github.com/typelevel/cats-effect/blob/52b5a3ba1623ff40fc48318c9a0bfae558f0e5ed/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala#L281-L297

However, `Resource#both` is currently broken because it does not pass the `ExitCase` to the finalizers correctly. See the tests I added in https://github.com/typelevel/cats-effect/commit/8d70133170a0eeea46a3f6e38c0a6ab7d06183fc. My attempted solution there encountered the same test failures as this PR.

I'm going to try and fix `fold` so that it can work with `ExitCase`.

armanbilge:
tl;dr: we do need both interpreters. Or maybe someone cleverer can figure out how to combine them.

PR #3228, opened 2022-11-06
Optimized `ByteStack` for JS

armanbilge:
Same idea as https://github.com/typelevel/cats-effect/pull/3222, to delegate to the JavaScript-native Array type which already behaves like a stack.

The V8 engine should in theory optimize this.

> While running JavaScript code, V8 keeps track of what kind of elements each array contains. This information allows V8 to optimize any operations on the array specifically for this type of element.

https://v8.dev/blog/elements-kinds

I'm going to see about benchmarks 😇

### DIFF
--- a/core/jvm-native/src/main/scala/cats/effect/ByteStack.scala
+++ b/core/jvm-native/src/main/scala/cats/effect/ByteStack.scala
@@ -18,6 +18,8 @@ package cats.effect
 
 private object ByteStack {
 
+  type T = Array[Int]
+
   final def toDebugString(stack: Array[Int], translate: Byte => String = _.toString): String = {
     val count = size(stack)
     ((count - 1) to 0 by -1)
--- a/core/js/src/main/scala/cats/effect/ByteStack.scala
+++ b/core/js/src/main/scala/cats/effect/ByteStack.scala
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2020-2022 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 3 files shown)


armanbilge:
~~Ok, so I used this thing. I still want us to do better, but at least it's something.~~
https://japgolly.github.io/scalajs-benchmark/res/scala-3.0.1-sjs-1.7.0.html#/shootouts/async_effects

Edit: blah, I think the benchmark as-written is not really testing this change.

```
Benchmark     Library                       Size     Cnt     ms/op ± error       ops/s ±   error
                                                                                          
FlatMap       Cats Effect v3.4.0-RC2        500      10     1.034 ± 0.085      969.614 ± 74.181
FlatMap       Cats Effect v3.4-70144ff      500      10     927.004 ± 16.460     1,079 ±    19
```

djspiewak:
I mean, in practice we aren't bounded by the `ByteStack`. :-) It's still very fun to optimize though.

core/jvm-native/src/main/scala/cats/effect/ByteStack.scala
```diff
@@ -18,6 +18,8 @@ package cats.effect
 
 private object ByteStack {
 
+  type T = Array[Int]
```
djspiewak, line 21: Can we sanity-check the bytecode quickly to ensure we haven't accidentally regressed something into a polymorphic case?

armanbilge: Sure. Note that `ByteStack` was already represented as a type alias.
https://github.com/typelevel/cats-effect/blob/6d63602dceb151a9cecc9524f4f8ff06669963ef/core/shared/src/main/scala/cats/effect/package.scala#L84

djspiewak: Oh right. It was. Okay less worried now.

PR #3229, opened 2022-11-07
Rewrite intro with four key concepts

valencik:
_This PR targets the `docs` branch_

This PR rewrites the homepage introduction to be shorter and focus on four important concepts.
Namely speed, reliability, the ecosystem, and our community :slightly_smiling_face:.

Preview:
Screenshot from 2022-11-07 06-59-09

Special thanks to @jducoeur, @SethTisue, and @djspiewak for considerable feedback on these changes. Trying to be descriptive and clear in few words is tricky.

Also thank you to all that filled out the informal survey to help identify the four important concepts and how we should be presenting ourselves.

### DIFF
--- a/website/pages/en/index.js
+++ b/website/pages/en/index.js
@@ -108,12 +108,16 @@ for {
   _ <- english.cancel >> french.cancel >> spanish.cancel
 } yield ()
  */
+        const fast = "**Fast**: Cats Effect provides lightweight fibers for asynchronous, highly concurrent applications. Your code stays fast even under extreme load.";
+        const reliable = "**Reliable**: Keep your applications up and running under high resource contention with automatic resource safety, backpressure, and cancellation of unnecessary work.";
+        const ecosystem = "**Ecosystem**: Cats Effect powers a thriving ecosystem of streaming frameworks, database layers, HTTP servers, and much more!";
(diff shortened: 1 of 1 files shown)


jducoeur (approved):

zmccoy (approved):
Looks great!

SystemFw (approved):
The wording looks great :)
It's not a part of this PR but I actually worry about the code snippet a bit. It looks "familiar" due to `for`, so that's an advantage, but it also promotes code that actually isn't resource safe/automatically cancels, so it conflicts with point 2 unless we make structured concurrency the default

valencik:
> I actually worry about the code snippet a bit. It looks "familiar" due to `for`, so that's an advantage, but it also promotes code that actually isn't resource safe/automatically cancels, so it conflicts with point 2 unless we make structured concurrency the default

I agree. Tackling the code snippet is next on my plate :)
I'll be sure to ping you for feedback 😁

armanbilge (approved):
So glad these are the four :)

djspiewak (approved):
This is great! I do think we should have an unobtrusive image of some sort for each one, but that can be a next step. Also eliminating or rewriting hte code snippet would be good.

PR #3232, opened 2022-11-07
Introduce `raceOutcomeBoth`

armanbilge:
This allows us to generalize over `raceOutcome` and `race`. The motivation for this is to be able to handle the race condition where _both_ fibers succeed, even though the loser is canceled. More on that in a follow-up PR.

If this merges then I think I need to re-do https://github.com/typelevel/cats-effect/pull/3226 in terms of `raceOutcomeBoth` instead.

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
@@ -315,9 +315,19 @@ trait GenSpawn[F[_], E] extends MonadCancel[F, E] with Unique[F] {
     uncancelable { poll =>
       poll(racePair(fa, fb)).flatMap {
         case Left((oca, f)) =>
-          f.cancel.whenA(!oca.isCanceled) *> f.join.map(ocb => Left((oca, ocb)))
+          val joined =
+            if (oca.isCanceled)
+              poll(f.join)
+            else
+              f.cancel *> f.join
+          joined.map(ocb => Left((oca, ocb)))
         case Right((f, ocb)) =>
-          f.cancel.whenA(!ocb.isCanceled) *> f.join.map(oca => Right((oca, ocb)))
+          val joined =
+            if (ocb.isCanceled)
+              poll(f.join)
+            else
+              f.cancel *> f.join
+          joined.map(oca => Right((oca, ocb)))
       }
     }
 
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
@@ -301,6 +301,27 @@ trait GenSpawn[F[_], E] extends MonadCancel[F, E] with Unique[F] {
       : F[Either[(Outcome[F, E, A], Fiber[F, E, B]), (Fiber[F, E, A], Outcome[F, E, B])]]
 
   /**
+   * Races the evaluation of two fibers, cancels the loser, and returns the [[Outcome]] of both.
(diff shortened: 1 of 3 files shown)


armanbilge:
Crap, this was a false generalization. `raceOutcome` and `race` vary in how they deal with the scenario that the winning fiber completed in cancelation.
- `raceOutcome` always cancels the loser, regardless of how the winner completed
- `race` joins the loser, if the winner completed in cancelation

I've currently implemented `raceOutcomeBoth` like `race`. But then `raceOutcome` cannot be derived from it (and indeed this is causing the JS tests to hang).

I have to think about this.

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -300,6 +300,37 @@ trait GenSpawn[F[_], E] extends MonadCancel[F, E] with Unique[F] {
   def racePair[A, B](fa: F[A], fb: F[B])
       : F[Either[(Outcome[F, E, A], Fiber[F, E, B]), (Fiber[F, E, A], Outcome[F, E, B])]]
 
+  /**
+   * Races the evaluation of two fibers, cancels the loser, and returns the [[Outcome]] of both.
+   * If the race is canceled before one or both participants complete, then then whichever ones
+   * are incomplete are canceled.
```
durban, line 306: I may be misunderstanding the intention, but this doesn't seem to be true: if one is canceled, and then the race is canceled while trying the join the other, then the other will not be canceled.

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -300,6 +300,37 @@ trait GenSpawn[F[_], E] extends MonadCancel[F, E] with Unique[F] {
...
+   *   the effect for the second racing fiber
+   */
+  def raceOutcomeBoth[A, B](fa: F[A], fb: F[B])
+      : F[Either[(Outcome[F, E, A], Outcome[F, E, B]), (Outcome[F, E, A], Outcome[F, E, B])]] =
+    uncancelable { poll =>
+      poll(racePair(fa, fb)).flatMap {
+        case Left((oca, f)) =>
+          val joined =
+            if (oca.isCanceled)
+              poll(f.join)
```
durban: See above, if we're canceled here, `f` will not be canceled.

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -300,6 +300,37 @@ trait GenSpawn[F[_], E] extends MonadCancel[F, E] with Unique[F] {
...
+            else
+              f.cancel *> f.join
+          joined.map(ocb => Left((oca, ocb)))
+        case Right((f, ocb)) =>
+          val joined =
+            if (ocb.isCanceled)
+              poll(f.join)
```
durban: As above.

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
armanbilge: Ah, thanks. I think the scaladoc was expressing my intent, I just screwed up the implementation 😅

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
armanbilge: You're right, I think this needs an `.onCancel(f.cancel)`. And same below.

armanbilge:
- https://github.com/typelevel/cats-effect/issues/3456

PR #3236, opened 2022-11-11
Record CPU starvation metrics

janstenpickle:
Builds on the work by @TimWSpence that adds CPU starvation detection, adding an MBean in the JVM with metrics on how many times starvation has been detected, the maximum clock drift and the current clock drift.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/metrics/CpuStarvationMbeanImpl.scala
+++ b/core/jvm/src/main/scala/cats/effect/metrics/CpuStarvationMbeanImpl.scala
@@ -16,11 +16,11 @@
 
 package cats.effect.metrics
 
-import java.util.concurrent.atomic.AtomicLong
-
 import cats.effect.IO
 
 import scala.concurrent.duration.FiniteDuration
+
+import java.util.concurrent.atomic.AtomicLong
 
 private[metrics] class CpuStarvationMbeanImpl private (
     counter: AtomicLong,
--- a/core/jvm/src/main/scala/cats/effect/metrics/CpuStarvationMbeanImpl.scala
+++ b/core/jvm/src/main/scala/cats/effect/metrics/CpuStarvationMbeanImpl.scala
@@ -1,0 +1,29 @@
+package cats.effect.metrics
+
+import java.util.concurrent.atomic.AtomicLong
+
+import cats.effect.IO
+
+import scala.concurrent.duration.FiniteDuration
+
+private[metrics] class CpuStarvationMbeanImpl private (
+    counter: AtomicLong,
+    maxClockDrift: AtomicLong)
+    extends CpuStarvationMbean {
+  override def getCpuStarvationCount(): Long = counter.get()
+
+  override def getMaxClockDriftMs(): Long = maxClockDrift.get()
+
+  def incStarvationCount: IO[Unit] = IO.delay(counter.incrementAndGet()).void
+
+  def recordDrift(drift: FiniteDuration): IO[Unit] =
+    IO.delay(maxClockDrift.updateAndGet(_.max(drift.toMillis))).void
+
+}
+
+object CpuStarvationMbeanImpl {
+  private[metrics] def apply(): IO[CpuStarvationMbeanImpl] = for {
+    counter <- IO.delay(new AtomicLong(0))
+    maxClockDrift <- IO.delay(new AtomicLong(0))
+  } yield new CpuStarvationMbeanImpl(counter, maxClockDrift)
+}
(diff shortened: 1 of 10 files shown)


core/jvm/src/main/scala/cats/effect/metrics/CpuStarvationMbeanImpl.scala
```diff
@@ -0,0 +1,56 @@
...
+
+  override def getCurrentClockDriftMs(): Long = currentClockDrift.get()
+
+  def incStarvationCount: IO[Unit] = IO.delay(counter.incrementAndGet()).void
+
+  def recordDrift(drift: FiniteDuration): IO[Unit] = {
+    lazy val driftMs = drift.toMillis
+
+    val maxDrift =
+      if (driftMs > 0) IO.delay(maxClockDrift.updateAndGet(_.max(driftMs))).void
```
janstenpickle: `Ref` seems to reimplement the Java "spin" for updating an atomic reference, I think to avoid conversion of a function to `UnaryOperator` on every modification.

Here the modification will only be performed if drift is detected, so shouldn't happen often, but I'm happy to copy the "spin" in `Ref` if needed!

core/jvm/src/main/scala/cats/effect/metrics/CpuStarvationMbeanImpl.scala
```diff
@@ -0,0 +1,56 @@
...
+  override def getCpuStarvationCount(): Long = counter.get()
+
+  override def getMaxClockDriftMs(): Long = maxClockDrift.get()
+
+  override def getCurrentClockDriftMs(): Long = currentClockDrift.get()
+
+  def incStarvationCount: IO[Unit] = IO.delay(counter.incrementAndGet()).void
+
+  def recordDrift(drift: FiniteDuration): IO[Unit] = {
+    lazy val driftMs = drift.toMillis
```
djspiewak: Minor point, but I think the `lazy` modifier here is meant as a performance optimization, but the allocation of the thunk is definitely more expensive than simply calling the function. :-)

core/jvm/src/main/scala/cats/effect/metrics/CpuStarvationMbeanImpl.scala
djspiewak: I don't think there's any need to port `Ref`'s spin implementation in this case, but I *would* like if you would use `Math.max` instead of `.max`. The performance difference between the two is several orders of magnitude.

core/jvm/src/main/scala/cats/effect/metrics/JvmCpuStarvationMetrics.scala
```diff
@@ -0,0 +1,65 @@
...
+import javax.management.{MBeanServer, ObjectName}
+
+private[effect] class JvmCpuStarvationMetrics private (mbean: CpuStarvationMbeanImpl)
+    extends CpuStarvationMetrics {
+  override def incCpuStarvationCount: IO[Unit] = mbean.incStarvationCount
+
+  override def recordClockDrift(drift: FiniteDuration): IO[Unit] = mbean.recordDrift(drift)
+}
+
+object JvmCpuStarvationMetrics {
```
djspiewak: Minor nit, but can we make this `private[effect]` as well?

core/shared/src/main/scala/cats/effect/metrics/CpuStarvationMetrics.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import scala.concurrent.duration.FiniteDuration
+
+private[effect] trait CpuStarvationMetrics {
+  def incCpuStarvationCount: IO[Unit]
+
+  def recordClockDrift(drift: FiniteDuration): IO[Unit]
+}
+
+object CpuStarvationMetrics {
```
djspiewak: Should also be `private[effect]`

core/shared/src/main/scala/cats/effect/metrics/CpuStarvationMetrics.scala
```diff
@@ -0,0 +1,35 @@
...
+  private[effect] def noOp: CpuStarvationMetrics = new CpuStarvationMetrics {
```
djspiewak: This kind of confused me when I first looked at it. In general, we try to make platform-specific code confined to that particular platform, rather than available everywhere. This might also avoid the `trait` penalty.

djspiewak (changes requested):
Some minor adjustments! This is really great though.

core/jvm/src/main/scala/cats/effect/metrics/CpuStarvationMbeanImpl.scala
janstenpickle: Good to know! I just realised I use it immediately below, so I may as well just change it to a val

djspiewak (dismissed):
Thank you!

PR #3240, opened 2022-11-12
Updated Scala and Cats

djspiewak:


djspiewak:
Too much yak shaving for now…

PR #3248, opened 2022-11-13
Updated to Scala 2.13.10 and 3.2.1

djspiewak:


build.sbt
```diff
@@ -767,6 +767,7 @@ lazy val core = crossProject(JSPlatform, JVMPlatform, NativePlatform)
       } else Seq()
     }
   )
+  .nativeSettings(tlFatalWarnings := false)
```
armanbilge, line 765: Hm, I'd rather nowarn these than lose my warnings for Native code. I can do that in a followup.

djspiewak: Yeah I think it's worth a follow-up. The problem is `nowarn` *itself* isn't working.

PR #3253, opened 2022-11-16
Site: add `Atomic Cell` page

iRevive:


std/shared/src/main/scala/cats/effect/std/AtomicCell.scala
```diff
@@ -22,24 +22,24 @@ import cats.effect.kernel._
 import cats.syntax.all._
 
 /**
- * A fiber-safe, concurrent, mutable reference.
+ * A synchronized, concurrent, mutable reference.
```
iRevive, line 25: Or is it better to keep the `fiber-safe`?

std/shared/src/main/scala/cats/effect/std/AtomicCell.scala
```diff
@@ -22,24 +22,24 @@ import cats.effect.kernel._
...
  *
  * Provides safe concurrent access and modification of its contents, by ensuring only one fiber
  * can operate on them at the time. Thus, '''all''' operations may semantically block the
  * calling fiber.
  *
  * {{{
- *  final class ParkingLot(data: AtomicCell[IO, ArraySeq[Boolean]], rnd: Random[IO]) {
```
iRevive, line 32: The example didn't compile. So I updated it.

djspiewak (approved):
This is fantastic, thank you!

PR #3254, opened 2022-11-16
Workaround for `IOApp` deadlock

armanbilge:
Fixes:
- https://github.com/bkirwi/decline/issues/461

Works around:
- https://github.com/scala/bug/issues/12692
- https://github.com/lampepfl/dotty/issues/16352

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/IOApp.scala
+++ b/core/jvm/src/main/scala/cats/effect/IOApp.scala
@@ -205,24 +205,29 @@ trait IOApp {
    * calling thread (for example, LWJGL). In these scenarios, it is recommended that the
    * absolute minimum possible amount of work is handed off to the main thread.
    */
-  protected lazy val MainThread: ExecutionContext =
-    new ExecutionContext {
-      def reportFailure(t: Throwable): Unit =
-        t match {
-          case t if NonFatal(t) =>
-            t.printStackTrace()
+  protected def MainThread: ExecutionContext =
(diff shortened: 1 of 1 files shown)


core/jvm/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -367,6 +372,9 @@ trait IOApp {
 
     val ioa = run(args.toList)
 
+    // workaround for scala#12692, dotty#16352
+    val queue = this.queue
```
armanbilge, line 376: By binding this, at least the same `queue` instance will be used below. It doesn't help with `MainThread`, but it fixes the worst of the issues.

core/jvm/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -205,24 +205,29 @@ trait IOApp {
...
+
+        def execute(r: Runnable): Unit =
+          if (!queue.offer(r)) {
+            runtime.blocking.execute(() => queue.put(r))
+          }
+      }
+    else
+      throw new UnsupportedOperationException(
+        "Your IOApp's super class has not been recompiled against Cats Effect 3.4.0+."
+      )
```
armanbilge, line 230: Not great but I don't have a better idea. 😕

djspiewak (approved):
Yeah this is unfortunate. This does at least avoid the worst pain.

PR #3257, opened 2022-11-16
Site: add `Env` page

iRevive:


djspiewak (dismissed):

docs/std/env.md
```diff
@@ -0,0 +1,32 @@
...
+  } yield ()  
+```
+
+### Why `def get(name): F[Option[String]]` rather than `def get(name): Option[String]`?
+
+Technically, environment variables can be set within the **current** process on a POSIX system.  
+Java doesn't provide that functionality in a public API, but it's possible via JNA API or in Scala Native world.
+
+Since the library supports multiple platforms (JVM, Scala Native, ScalaJS), the implementation should take
+platform-specific peculiarities into the account.
```
armanbilge: Also, I think we can get a security error on the JVM, and we would need an effect to capture that.

iRevive: Indeed. I will add this detail tomorrow.

armanbilge: Also, I just remembered: the AWS Lambda runtime actually _does_ change env variables while the process is running.

https://github.com/aws/aws-lambda-java-libs/blob/04ef611147fba47202c376da56741bf857b1edba/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/AWSLambda.java#L267-L271

iRevive: "Wake up, samurai. We have _env variables_ to change."

Now I'm 100% convinced it should be effectful.

armanbilge: We even rely on this in natchez :)
https://github.com/tpolecat/natchez/blob/ffa7ef08d40cb1dcf4a27e4e28ff2d66c45b3aa1/modules/xray/src/main/scala/natchez/xray/XRayEnvironment.scala#L33-L34

iRevive: I updated the paragraph.

armanbilge (approved):
Thanks for all these docs PRs! :)

PR #3258, opened 2022-11-16
Update to sbt-typelevel 0.5.0-M8

armanbilge:


.github/workflows/ci.yml
```diff
@@ -17,40 +17,49 @@ on:
 env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+
+concurrency:
+  group: ${{ github.workflow }} @ ${{ github.ref }}
+  cancel-in-progress: true
```
armanbilge, line 23: Yay / nay ?

.github/workflows/ci.yml
```diff
@@ -118,12 +128,19 @@ jobs:
...
         with:
           distribution: jdkfile
           java-version: 8
           jdkFile: ${{ steps.download-java-temurin-8.outputs.jdkFile }}
+          cache: sbt
+
+      - name: sbt update
+        if: matrix.java == 'temurin@8' && !steps.setup-java-temurin-8.outputs.cache-hit
+        shell: bash
+        run: sbt '++ ${{ matrix.scala }}' reload +update
```
armanbilge: As requested: on cache miss run `+update`.

.github/workflows/ci.yml
```diff
@@ -91,13 +100,14 @@ jobs:
...
-            scala: 2.12.17
+            scala: 2.12
           - os: macos-latest
             ci: ciNative
-            scala: 3.2.1
+            scala: 3
           - os: windows-latest
             java: graalvm@11
     runs-on: ${{ matrix.os }}
+    timeout-minutes: 60
```
armanbilge, line 106: Another yay/nay? I think this should be sufficient.

.github/workflows/ci.yml
armanbilge: ~Hmm, I don't think this is working correctly.~ Update: fixed

build.sbt
```diff
@@ -117,6 +117,7 @@ val Scala213 = "2.13.10"
 val Scala3 = "3.2.1"
 
 ThisBuild / crossScalaVersions := Seq(Scala3, Scala212, Scala213)
+ThisBuild / githubWorkflowScalaVersions := crossScalaVersions.value
```
djspiewak, line 120: Why did this have to be added? Is it no longer the default upstream?

.github/workflows/ci.yml
djspiewak: Yay to both. :-)

djspiewak (approved):

build.sbt
armanbilge: It's not anymore. As of sbt 1.8.0 it's possible to write `++ 2.13` or `++ 3` so that's what the default is instead. This is helpful since the workflow is stable to Scala updates, and thus the CI job names are stable to updates, and thus mergify configs and branch protection settings are stable to updates.

I tried to make it work for CE but I got stuck in the scripts for our example apps which need the full Scala 3 version to find the executable. Actually that's probably solvable, I guess I just lost my patience by that point lol.
https://github.com/typelevel/cats-effect/blob/dad00c83668e4b967c773b614d976c680548613f/example/test-native.sh#L15

PR #3264, opened 2022-11-17
Defer instance for Resource without Sync requirement

Odomontois:


tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
```diff
@@ -1095,6 +1097,27 @@ class ResourceSpec extends BaseSpec with ScalaCheck with Discipline {
     )
   }
 
+  // local newsupertype around IO, does not have a Sync instance, but still can be used in Resource tests
+  type AlmostIO[+A]
```
armanbilge: Instead of adding this, perhaps take a look at `PureConc`?

https://github.com/typelevel/cats-effect/blob/7a8fd8cb951939d35ef6b539f6a11b8fc7304481/tests/shared/src/test/scala/cats/effect/kernel/ParallelFSpec.scala#L64-L68

Odomontois: Honestly, I don't think it worked better at the end

tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
```diff
@@ -1095,6 +1095,27 @@ class ResourceSpec extends BaseSpec with ScalaCheck with Discipline {
...
   }
 
+  {
+    import cats.effect.kernel.testkit.pure._
+    import cats.effect.kernel.testkit.PureConcGenerators.arbitraryPureConc
+    import org.scalacheck.Arbitrary
+
+    type F[A] = PureConc[Throwable, A]
+
+    val arbitraryPurConcResource: Arbitrary[Resource[F, Int]] =
```
catostrophe:
```suggestion
    val arbitraryPureConcResource: Arbitrary[Resource[F, Int]] =
```

djspiewak (approved):
Nice catch! Thank you!

PR #3267, opened 2022-11-18
Site: add `Console` page

iRevive:


docs/std/console.md
```diff
@@ -0,0 +1,50 @@
...
+The problem arises from the underlying implementation of `read` in the `System.in` input stream. 
+The only way to complete the `read` is to supply some actual data or close the input.
+
+### Scala.JS
+
+`readLine` is not implemented for Scala.js. On Node.js consider using `fs2.io.stdin`.
+
+### Scala Native
+
+Scala Native handles cancelation of `readLine` correctly.
```
armanbilge: Does it? I never actually tried, but I assumed it's really broken since Scala Native doesn't have a blocking pool.

iRevive: My bad. The termination of the program does not hang, but `IO.readline.timeoutTo` hangs indefinitely. 

```scala
//> using scala "2.13.10"
//> using lib "org.typelevel::cats-effect::3.4.1"

import cats.effect._

  def run: IO[Unit] =
    IO.readLine.flatMap(IO.println).timeoutTo(3.seconds, IO.println("done")) 
}
```

```shell
$ scala-cli --native readline.scala # can be terminated with ctrl + c
$ scala-cli readline.scala # cannot be terminated
```

armanbilge: Ah yes, that's because of another Cats Effect Native shortcoming: ctrl+c doesn't run cancellation finalizers, because we don't have signal handlers. Win some, lose some I guess 😂

djspiewak (approved):

PR #3275, opened 2022-11-21
Fix 'Dispatcher' deprecaton annotation 'message' and 'since'

seigert:
Currently `Dispatcher.apply` deprecation annotation have it's fields swapped, leading to not so readable errors:

```console
method apply in object Dispatcher is deprecated (since use parallel or sequential instead; the former corresponds to the current semantics of this method): 3.4.0
```


armanbilge (dismissed):
Woops, thanks!

armanbilge (approved):

PR #3281, opened 2022-11-25
Implement `Ref` without wrapping `AtomicReference` on JS/Native

armanbilge:
Some low-hanging fruit I noticed. This _should_ be a win by removing a layer of allocation and indirection.

On the JVM it uses `AtomicReferenceFieldUpdater`. It took some shenanigans to get the compiler happy with mixed Java/Scala sources for this. One day in Scala 3 utopia when we can have true `@static` fields that should no longer be necessary.

On JS/Native it just uses a `var`. Single-threading ftw 😁

When Native gets multi-threading we can use a similar trick as the JVM to operate atomically on an a volatile field. Without the pain of going through Java 😅


tests/shared/src/test/scala/cats/effect/kernel/RefSpec.scala
```diff
@@ -103,23 +103,24 @@ class RefSpec extends BaseSpec { outer =>
...
-          r <- Ref[IO].of(0)
-          result <- r.tryUpdate { currentValue =>
-            updateRefUnsafely(r)
-            currentValue + 1
+    if (!isJS && !isNative) // concurrent modification impossible
+      "tryUpdate - should fail to update if modification has occurred" in ticked {
+        implicit ticker =>
+          val updateRefUnsafely: Ref[IO, Int] => Unit = { (ref: Ref[IO, Int]) =>
+            unsafeRun(ref.update(_ + 1))
+            ()
```
armanbilge, line 111: While _technically_ possible (as this test demonstrates) this is serious cheating on JS/Native. Not supporting it simplifies the implementation since it is always happy-path.

kernel/shared/src/main/scala/cats/effect/kernel/Ref.scala
```diff
@@ -243,7 +239,7 @@ object Ref {
    * }}}
    */
   def unsafe[F[_], A](a: A)(implicit F: Sync[F]): Ref[F, A] =
-    new SyncRef[F, A](new AtomicReference[A](a))
+    new SyncRef(F, a).asInstanceOf[Ref[F, A]]
```
armanbilge: \*cough\* shenanigans \*cough\* I mean, it _does_ say "unsafe" 😇

armanbilge:
Huh.

#### This PR
```
[info] Benchmark                   Mode  Cnt     Score    Error  Units
[info] RefBenchmark.getAndUpdate  thrpt   20  2391.638 ± 52.331  ops/s
[info] RefBenchmark.modify        thrpt   20  2272.606 ± 53.699  ops/s
```

#### series/3.4.x
```
[info] Benchmark                   Mode  Cnt     Score     Error  Units
[info] RefBenchmark.getAndUpdate  thrpt   20  2479.282 ±  52.203  ops/s
[info] RefBenchmark.modify        thrpt   20  2220.954 ± 129.366  ops/s
```

armanbilge:
Did an experiment in 80cec891555f30091af380d121824348109fabf7 where I implemented `Ref` in Java instead of extending an abstract Scala implementation. That seems to help, but it's not very fun.

```
[info] Benchmark                   Mode  Cnt     Score    Error  Units
[info] RefBenchmark.getAndUpdate  thrpt   20  2488.491 ± 52.568  ops/s
[info] RefBenchmark.modify        thrpt   20  2292.940 ± 70.478  ops/s
```

armanbilge:
Even with the latest iteration the `AtomicReference` still somehow has an edge.

#### This PR
```
[info] Benchmark                   Mode  Cnt     Score    Error  Units
[info] RefBenchmark.getAndUpdate  thrpt   20  2518.817 ± 38.037  ops/s
[info] RefBenchmark.modify        thrpt   20  2434.156 ± 38.291  ops/s
```

#### series/3.4.x
```
[info] Benchmark                   Mode  Cnt     Score    Error  Units
[info] RefBenchmark.getAndUpdate  thrpt   20  2609.786 ± 16.483  ops/s
[info] RefBenchmark.modify        thrpt   20  2460.229 ± 93.724  ops/s
```

djspiewak:
My guess is that the allocation just isn't dominating, and the JIT probably inlines `AtomicReference` very slightly better than the alternative.

armanbilge:
Well, the benchmarks aren't really measuring the allocation anyway. But yeah, it might be these additional access checks in the updater implementation that the `AtomicReference` is able to elide. Meh.

https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/19fb8f93c59dfd791f62d41f332db9e306bc1422/src/java.base/share/classes/java/util/concurrent/atomic/AtomicReferenceFieldUpdater.java#L439-L443

I'll revert the JVM changes, but I'd still like to keep it for JS. On Native we should be able to do the unsafe thing in the near future.

djspiewak (approved):

PR #3284, opened 2022-11-26
Added a specialized version of `Deferred` based on `IOFiber`'s machinery

### LINKED ISSUE #3271: Specialize `Deferred` for `IO` to use `CallbackStack`
The `get` function on `Deferred` relies on notifying listeners of the fulfillment of a write-once memory slot. This is startlingly similar to the implementation of `join` in `IOFiber`, except that `IOFiber` uses a dramatically more optimized data structure than what we find in `Deferred`. In theory, it should be possible for us experiment with overriding `Concurrent[IO].deferred` with a custom implementation (`IODeferred`?) which in turn uses the `CallbackStack` data structure from `IOFiber` itself. We should benchmark whether this is actually faster (it should be!) and by how much.

djspiewak:
Experimental reimplementation of `Deferred` based on `IOFiber`'s internal machinery rather than the more generic `LongMap` approach. Benchmarks incoming. One unfortunate discovery as part of working on this is the fact that the `Deferred` API is… distressingly open. `unsafe` is quite annoying, but `AsyncDeferred` is even worse. May have to put more thought into fixing this.

Fixes #3271


djspiewak:
Performance results are in! Looks like this is around 20-40% faster than the original. The higher count tests are more accurate than the lower count ones since they factor out the overhead a bit better. The `complete` benchmark is… bad… and I need to find a better way to do it. Unfortunately, this one is sort of fundamentally a race condition, which I suspect is part of why we don't see a major improvement in this scenario (~5%), and also why one of the results is wildly out of step with the others.

## Before

```
[info] Benchmark                   (count)   Mode  Cnt     Score    Error  Units
[info] DeferredBenchmark.cancel         10  thrpt   10    90.653 ±  1.991  ops/s
[info] DeferredBenchmark.cancel        100  thrpt   10    10.547 ±  0.188  ops/s
[info] DeferredBenchmark.cancel       1000  thrpt   10     0.834 ±  0.039  ops/s
[info] DeferredBenchmark.complete       10  thrpt   10    62.500 ±  0.385  ops/s
[info] DeferredBenchmark.complete      100  thrpt   10     3.766 ±  0.038  ops/s
[info] DeferredBenchmark.complete     1000  thrpt   10     2.743 ±  0.041  ops/s
[info] DeferredBenchmark.get            10  thrpt   10  3369.843 ± 20.579  ops/s
[info] DeferredBenchmark.get           100  thrpt   10   533.762 ±  2.963  ops/s
[info] DeferredBenchmark.get          1000  thrpt   10    56.230 ±  0.297  ops/s
```

## After

```
[info] Benchmark                   (count)   Mode  Cnt     Score    Error  Units
[info] DeferredBenchmark.cancel         10  thrpt   10    97.689 ±  1.219  ops/s
[info] DeferredBenchmark.cancel        100  thrpt   10    12.487 ±  0.292  ops/s
[info] DeferredBenchmark.cancel       1000  thrpt   10     1.077 ±  0.053  ops/s
[info] DeferredBenchmark.complete       10  thrpt   10    65.497 ±  0.245  ops/s
[info] DeferredBenchmark.complete      100  thrpt   10    15.663 ±  0.897  ops/s
[info] DeferredBenchmark.complete     1000  thrpt   10     2.833 ±  0.046  ops/s
[info] DeferredBenchmark.get            10  thrpt   10  4019.782 ± 13.285  ops/s
[info] DeferredBenchmark.get           100  thrpt   10   743.074 ±  3.311  ops/s
[info] DeferredBenchmark.get          1000  thrpt   10    80.084 ±  0.312  ops/s
```

armanbilge:
> The `complete` benchmark is… bad… and I need to find a better way to do it.

Sorry, why is it bad exactly?

core/shared/src/main/scala/cats/effect/IODeferred.scala
```diff
@@ -0,0 +1,82 @@
...
+
+import java.util.concurrent.atomic.AtomicReference
+
+private final class IODeferred[A] extends Deferred[IO, A] {
+  import IODeferred.Sentinel
+
+  private[this] val cell = new AtomicReference[AnyRef](Sentinel)
+  private[this] val callbacks = new CallbackStack[Right[Nothing, A]](null)
+
+  def complete(a: A): IO[Boolean] = IO {
```
filipwiech, line 31: Based on the #3290 this should probably be uncancelable. 🙂

filipwiech:
Some ideas for further optimizations based on @armanbilge's recent work (#3281 & #3300):

- Try to avoid wrapping `AtomicReference`.
- Specialize `Deferred` for `Boolean`, `Int`, `Long`.

Not sure how much sense they make and most likely you've already thought of them, but wanted to mention it just in case. 🙂

armanbilge:
> * Try to avoid wrapping `AtomicReference`.

I tried this as part of https://github.com/typelevel/cats-effect/pull/3281 using a `AtomicReferenceFieldUpdater` but unfortunately it had worse performance, see https://github.com/typelevel/cats-effect/pull/3281#issuecomment-1327055441.

filipwiech:
You're right, the idea might indeed be not-so-great. Although it could be interesting to benchmark the impact on memory allocation. There's also `AtomicIntegerFieldUpdater` and `AtomicLongFieldUpdater` that may be of interest in the specialized cases (there isn't one for `Boolean`?), but I admit that it looks like a long shot, sorry for the noise. 👍

armanbilge:
Yeah, it's a memory / performance trade-off. But since most refs are get/set/update many more times than they are allocated, probably better to bias towards performance.

The specializations are also interesting, except my other PR also found no performance improvement from doing that 😂 see https://github.com/typelevel/cats-effect/pull/3300#issuecomment-1336407466.

Optimization is hard :) @filipwiech it's hard to know what's a good idea or not until you try it and benchmark it.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1434,7 +1434,7 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
   def ref[A](a: A): IO[Ref[IO, A]] = IO(Ref.unsafe(a))
 
-  def deferred[A]: IO[Deferred[IO, A]] = IO(Deferred.unsafe)
+  def deferred[A]: IO[Deferred[IO, A]] = IO(new IODeferred[A])
```
filipwiech, line 1437: Would it be a good idea to add a test which verifies that `IO.deferred` returns custom implementation and not the "standard" one? 🙂

djspiewak: I think it would! Probably in `IOSpec`

core/shared/src/main/scala/cats/effect/IODeferred.scala
durban: It already is (it's a single `delay` block).

filipwiech: Thank you for pointing this out, you're absolutely right! 👍 To be honest I wonder if the documentation should be more clear and explicit about cancelation semantics and how they relate to the actual code. Right now I think one of the tutorials mentions offhandedly something about "cancelation kicking in at each `flatMap`". I remember having different mental model about this in the past, especially since the docs say elsewhere that "fibers are cancelable at all points during their execution". I think this could be somewhat confusing, or maybe it's just me. 😄

djspiewak: I agree that the cancelation documentation is *really* confusing. I think we've tried to do better on this in recent years but it could use some consolidation and cleanup.

djspiewak:
@armanbilge 

> Sorry, why is it bad exactly?

Sorry didn't reply to this. The benchmark is as follows:

```scala
  @Benchmark
  def complete(): Unit = {
    IO.deferred[Unit]
      .flatMap { d => d.get.parReplicateA_(count) &> d.complete(()) }
      .replicateA_(1000)
      .unsafeRunSync()
  }
```

So the problem here is the fact that it just reduces to the `get` benchmark very, very quickly:

```scala
  @Benchmark
  def get(): Unit = {
    IO.deferred[Unit]
      .flatMap(d => d.complete(()) *> d.get.replicateA_(count))
      .replicateA_(1000)
      .unsafeRunSync()
  }
```

Notice the only difference here is the race condition on `complete`'s scheduling!

What I'm *trying* to measure is the cost when a large number of fibers are enqueued on `get` and then released simultaneously by a single `complete`. The problem is that Cats Effect's concurrency model makes this a very difficult thing to model deterministically, which in turn makes it immensely hard to measure from a performance standpoint.

armanbilge:
> The problem is that Cats Effect's concurrency model makes this a very difficult thing to model deterministically, which in turn makes it immensely hard to measure from a performance standpoint.

Ah, indeed! Thanks for the explanation :)

core/shared/src/main/scala/cats/effect/IODeferred.scala
```diff
@@ -0,0 +1,82 @@
...
+      false
+    }
+  }
+
+  def get: IO[A] = IO defer {
+    val back = cell.get()
+
+    if (back eq Sentinel) {
+      IO.cont[A, A](new Cont[IO, A, A] {
+        def apply[G[_]: MonadCancelThrow] = {
```
armanbilge, line 45: Can this be written with `asyncCheckAtempt`? Which I believe is not on this branch (although in theory we could backport it as a package-private).

core/shared/src/main/scala/cats/effect/IODeferred.scala
```diff
@@ -0,0 +1,82 @@
...
+  def tryGet: IO[Option[A]] = IO {
+    val back = cell.get()
+    if (back eq Sentinel)
+      None
+    else
+      Some(back.asInstanceOf[A])
+  }
+}
+
+object IODeferred {
```
armanbilge:
```suggestion
private object IODeferred {
```

armanbilge (commented):
Just a visibility issue, otherwise LGTM :)

Now that `CallbackStack` has become even more important I'd like to follow-up with a specialized implementation for JS.

core/shared/src/main/scala/cats/effect/IODeferred.scala
djspiewak: I was thinking of following up on `series/3.x` to replace it with `asyncCheckAttempt`. IMO it isn't worth the backport.

core/shared/src/main/scala/cats/effect/IODeferred.scala
djspiewak: Oh good catch

djspiewak:
> Now that CallbackStack has become even more important I'd like to follow-up with a specialized implementation for JS.

I'm curious if you can actually beat the performance on JS. From my understanding of how `AtomicReference` compiles, it should be effectively equivalent to the best you can do even taking advantage of the single-threaded runtime.

armanbilge:
> I'm curious if you can actually beat the performance on JS. From my understanding of how `AtomicReference` compiles, it should be effectively equivalent to the best you can do even taking advantage of the single-threaded runtime.

Really? Isn't this basically a thread-safe linked list? On JS couldn't we just use a native stack aka `js.Array` ?

armanbilge (dismissed):

djspiewak:
Oh right it's a stack not a queue. Yeah you could totally beat it.

djspiewak:
Depends on #3323

armanbilge (approved):

PR #3287, opened 2022-11-27
properties can override environment variables

massimosiani:
Hi!
Wondering whether this may be useful to others too.

### DIFF
--- a/tests/js/src/test/scala/cats/effect/std/EnvPlatformSpecification.scala
+++ b/tests/js/src/test/scala/cats/effect/std/EnvPlatformSpecification.scala
@@ -17,20 +17,12 @@
 package cats.effect
 package std
 
-class EnvSpec extends BaseSpec with EnvPlatformSpecification {
-
-  "Env" should {
-    "retrieve a variable from the environment" in real {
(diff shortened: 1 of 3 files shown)


tests/js/src/test/scala/cats/effect/std/EnvPlatformSpecification.scala
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package cats.effect
+package std
+
+trait EnvPlatformSpecification { self: BaseSpec =>
+  def platformSpecs = {
+    "platform" should {
+      "pass" in real {
+        IO("1" must be("1"))
```
massimosiani, line 24: Never used specs2, I couldn't find a way to use a keyword like Success, but probably it's me. Any hint?

armanbilge:
Thanks for the PR! The purpose of `Env` is to be a faithful wrapper of `sys.env` / `System.getenv`. I think custom behavior like this would be highly surprising for users. It does however seem useful, perhaps as a feature in some sort of config library? :)

djspiewak:
I think this is definitely more of a config library-type change. `Env` itself should definitely just mirror the environment variables. This was a helpful experiment though!

PR #3296, opened 2022-12-01
Yet another polling system sketch

armanbilge:
The idea of this sketch is to define a `trait EventLoop[Registrar] extends ExecutionContext` that provides access to a "registrar" (e.g. a `Selector` or `io_uring`) for registering callbacks on sockets etc., but only if you are currently running on that `ExecutionContext`.

Then, every `Network` is bound to a specific `EventLoop` and must `.evalOn(...)` to that loop whenever it needs to interact with the registrar. In the ideal scenario this `evalOn` will be a no-op (i.e. already running on the WSTP which the desired polling system installed).

My goals were to:
1. Provide a unified interface for working with polling systems, that can be implemented by both the WSTP and an I/O dispatcher thread. This saves FS2 from having to maintain two separate implementations depending on whether the polling system is available in the WSTP or not.
2. Address the problem of being on the wrong EC when trying to interact with the polling system, by providing a straightforward mechanism to get on the right EC.

Usual caveats: rough sketch, names subject to bikeshed, bla bla :)

Example usage:

```scala
object Network {
  def apply[F[_]](implicit F: Async[F]): Resource[F, Network[F]] =
    Resource.eval(F.executionContext).flatMap {
      case EventLoop[Selector](loop) =>
        // we're on an EC with the desired polling system
        Resource.pure(loop)
      case _ =>
        Resource.make {
          // make a single-thread I/O dispatcher for the desired polling system
          F.delay(EventLoop.fromPollingSystem(SelectorSystem))
        } { case (_, release) => F.delay(release()) }
       .map { case (loop, _) => loop }
    }.map { loop =>
       new Network[F] {
         // ...
         def doSomeIO(): F[Unit] =
           F.delay {
             // do stuff ...
             val selector = loop.registrar()
             selector.registerCallback(...)
           }.evalOn(loop) // but do it on the loop, no matter who calls us
       }
    }
}
```

Note: although here I've demoed it at the `Network` level, this can also be done at the `Socket` level, which would be more friendly to FS2's current encoding of `Network` as a global implicit.


core/jvm/src/main/scala/cats/effect/unsafe/EventLoop.scala
```diff
@@ -0,0 +1,78 @@
...
+object EventLoop {
+  def unapply[R](loop: EventLoop[Any])(ct: ClassTag[R]): Option[EventLoop[R]] =
+    if (ct.runtimeClass.isAssignableFrom(loop.registrarTag.runtimeClass))
+      Some(loop.asInstanceOf[EventLoop[R]])
+    else
+      None
+
+  def fromPollingSystem(
+      name: String,
+      system: PollingSystem): (EventLoop[system.Poller], () => Unit) = {
```
armanbilge, line 43: A single-thread dispatcher from a polling system. I just tossed this together, I doubt it's any good.

I suspect that there will be objections to including this in CE. My argument for why we should have this is:
1. implementing a good event dispatcher thread is non-trivial
2. in the case when the WSTP does not have the desired polling system installed, to provide a fallback implementation a dispatcher thread will be necessary.

djspiewak:
Haven't had a chance to read through this yet, but just at a high level, I do think that `evalOn` is probably the right approach to resolving the "you might be on the wrong thing". It doesn't *fully* solve the problem unless you also include a `LiftIO` constraint, but it helps.

core/jvm/src/main/scala/cats/effect/unsafe/EventLoop.scala
```diff
@@ -0,0 +1,78 @@
...
+  def registrar(): Registrar
+
+}
+
+object EventLoop {
+  def unapply[R](loop: EventLoop[Any])(ct: ClassTag[R]): Option[EventLoop[R]] =
+    if (ct.runtimeClass.isAssignableFrom(loop.registrarTag.runtimeClass))
+      Some(loop.asInstanceOf[EventLoop[R]])
+    else
+      None
```
armanbilge, line 39: So instead of this pattern match maybe we should expose this as `IO.eventLoop` or something.

PR #3299, opened 2022-12-04
Add `IO#voidError`

### LINKED ISSUE #3152: Add a combinator which eats unhandled errors
Ultimately this would need to be added in Cats as well as in Cats Effect, but we can start here with `IO`. The motivating use-case is when someone has a fiber with the following sort of definition:

```scala
IO.deferred[OutcomeIO[A]] flatMap { oc =>
  fa.guaranteeCase(oc.complete(_).void).start
  // ...
}
```

If `fa` produces an error, that error will be correctly shifted into the `Deferred` and (presumably) handled, but this indirect process is invisible to Cats Effect, and thus it will consider any errors in `fa` to be "unhandled". This in turn triggers the much-maligned side-channeling (via the `ExecutionContext`). This pattern is particularly common in Fs2, which makes for a very unnecessarily chatty stderr out of the box.

A simple combinator could help resolve this problem at the definition site:

```scala
// name tbd
def swallowErrors(action: IO[Unit]): IO[Unit] =
  action.handleError(_ => ())
```

This would allow the definition site to be changed relatively easily:

```scala
IO.deferred[OutcomeIO[A]] flatMap { oc =>
  fa.guaranteeCase(oc.complete(_).void).swallowErrors.start
  // ...
}
```

In a sense, this is much more prescriptively correct formulation, and it completely avoids the issue of the overly chatty error reporting, but it will take some time to percolate through the ecosystem. A simple combinator can help this process along.

armanbilge:
Closes https://github.com/typelevel/cats-effect/issues/3152

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -802,6 +802,9 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   def void: IO[Unit] =
     map(_ => ())
 
+  def voidError(implicit ev: A <:< Unit): IO[Unit] =
+    asInstanceOf[IO[Unit]].handleError(_ => ())
+
   /**
    * Converts the source `IO` into any `F` type that implements the [[LiftIO]] type class.
    */
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -802,8 +802,10 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   def void: IO[Unit] =
     map(_ => ())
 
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -802,6 +802,9 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   def void: IO[Unit] =
     map(_ => ())
 
+  def voidError(implicit ev: A <:< Unit): IO[Unit] =
+    asInstanceOf[IO[Unit]].handleError(_ => ())
```
```suggestion
  def voidError(implicit ev: A <:< Unit): IO[Unit] = {
    val _ = ev
    asInstanceOf[IO[Unit]].handleError(_ => ())
  }
```

djspiewak (approved):

PR #3300, opened 2022-12-04
Specialize `Ref` for `Boolean`, `Int`, `Long`

### LINKED ISSUE #3181: Specialize primitive `Ref`s during construction
If we check `getClass` on the initial value passed to the constructor, we should be able to sort out whether it's actually a primitive ref, and then use `AtomicInteger`/`AtomicLong`/`AtomicBoolean`/etc instead. Once we do *that*, we can further specialize operations like `update` since `Function1` is specialized on primitives, meaning that we can get fully unboxed `Ref` operations in the normal case.

Whoever does this should add some benchmarks.

armanbilge:
Closes https://github.com/typelevel/cats-effect/issues/3181.

I think this covers everything, except the benchmarks.

### DIFF
--- a/kernel/js-native/src/main/scala/cats/effect/kernel/kernelplatform.scala
+++ b/kernel/js-native/src/main/scala/cats/effect/kernel/kernelplatform.scala
@@ -14,22 +14,10 @@
  * limitations under the License.
  */
 
-package cats.effect
+package cats.effect.kernel
 
-package object kernel extends kernelplatform {
-
-  type MonadCancelThrow[F[_]] = MonadCancel[F, Throwable]
-  val MonadCancelThrow = MonadCancel
-
-  type Spawn[F[_]] = GenSpawn[F, Throwable]
-  val Spawn = GenSpawn
-
-  type Temporal[F[_]] = GenTemporal[F, Throwable]
-  val Temporal = GenTemporal
-
-  type Concurrent[F[_]] = GenConcurrent[F, Throwable]
-  val Concurrent = GenConcurrent
-
-  type ParallelF[F[_], A] = Par.ParallelF[F, A]
-  val ParallelF = Par.ParallelF
+private[effect] trait kernelplatform {
+  private[kernel] type SyncBooleanRef[F[_]] = SyncRef[F, Boolean]
+  private[kernel] type SyncIntRef[F[_]] = SyncRef[F, Int]
+  private[kernel] type SyncLongRef[F[_]] = SyncRef[F, Long]
 }
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
@@ -194,7 +194,7 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
         if (offset1 < offset2) cacheValues1 else cacheValues2
     }
 
-    minCacheValuesF.flatMap(ref).map { cacheValuesRef =>
+    minCacheValuesF.flatMap(ref(_)).map { cacheValuesRef =>
       monotonic.flatMap { timeNow =>
         cacheValuesRef.get.flatMap {
           case (cacheRefreshTime, offset) =>
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Ref.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Ref.scala
@@ -132,6 +132,9 @@ object Ref {
     "Cannot find an instance for Ref.Make. Add implicit evidence of Concurrent[${F}, _] or Sync[${F}] to scope to automatically derive one.")
   trait Make[F[_]] {
     def refOf[A](a: A): F[Ref[F, A]]
+    def refOf(b: Boolean): F[Ref[F, Boolean]] = refOf[Boolean](b)
+    def refOf(i: Int): F[Ref[F, Int]] = refOf[Int](i)
+    def refOf(l: Long): F[Ref[F, Long]] = refOf[Long](l)
   }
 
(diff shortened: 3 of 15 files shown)


kernel/js-native/src/main/scala/cats/effect/kernel/kernelplatform.scala
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+package cats.effect.kernel
+
+private[effect] trait kernelplatform {
+  private[kernel] type SyncBooleanRef[F[_]] = SyncRef[F, Boolean]
+  private[kernel] type SyncIntRef[F[_]] = SyncRef[F, Int]
+  private[kernel] type SyncLongRef[F[_]] = SyncRef[F, Long]
+}
```
armanbilge, line 23: There's no point specializing these on JS.

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
```diff
@@ -194,7 +194,7 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
         if (offset1 < offset2) cacheValues1 else cacheValues2
     }
 
-    minCacheValuesF.flatMap(ref).map { cacheValuesRef =>
+    minCacheValuesF.flatMap(ref(_)).map { cacheValuesRef =>
```
armanbilge, line 197: Unfortunately this is going to be source-breaking.

kernel/shared/src/main/scala/cats/effect/kernel/Ref.scala
```diff
@@ -245,6 +269,15 @@ object Ref {
...
   def unsafe[F[_], A](a: A)(implicit F: Sync[F]): Ref[F, A] = new SyncRef(a)
 
+  private[effect] def unsafe[F[_]](b: Boolean)(implicit F: Sync[F]): Ref[F, Boolean] =
+    new SyncBooleanRef(b)
+
+  private[effect] def unsafe[F[_]](i: Int)(implicit F: Sync[F]): Ref[F, Int] =
+    new SyncIntRef(i)
+
+  private[effect] def unsafe[F[_]](l: Long)(implicit F: Sync[F]): Ref[F, Long] =
+    new SyncLongRef(l)
```
armanbilge, line 279: Since we are sour on these, I kept them private. Also there's various constructors here I didn't specialize. I guess I can take another look at those ...

armanbilge:
```
[info] Benchmark                              Mode  Cnt     Score    Error  Units
[info] RefBenchmark.getAndUpdateGeneric      thrpt   20  2464.489 ± 15.611  ops/s
[info] RefBenchmark.getAndUpdateSpecialized  thrpt   20  2442.461 ± 15.476  ops/s
[info] RefBenchmark.modifyGeneric            thrpt   20  2260.022 ± 86.474  ops/s
[info] RefBenchmark.modifySpecialized        thrpt   20  2179.681 ± 78.504  ops/s
```

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
djspiewak: Am very okay with that :-)

djspiewak:
Wait… there's no performance difference at all? What are we missing?

armanbilge:
I'm not sure. Just for the record, the boxing _is_ going away.

Here's how `spin` decompiles for `getAndUpdate`:
```java
    private final long spin$1(Function1 f$1) {
        long u;
        long l;
        while (!this.al.compareAndSet(l = this.al.get(), u = f$1.apply$mcJJ$sp(l))) {
        }
        return l;
    }
```

And here is the specialized lambda passed to it in the benchmark.
```java
    private static final IO loop$2(int remaining, long acc, Ref ref$2) {
        if (remaining == 0) {
            return IO$.MODULE$.apply((Function0)(JFunction0.mcJ.sp & Serializable)() -> acc);
        }
        return ((IO)ref$2.getAndUpdate((Function1)(JFunction1.mcJJ.sp & Serializable)x$1 -> x$1 + 1L)).flatMap((Function1 & Serializable)prev -> RefBenchmark$.loop$2(remaining - 1, acc + BoxesRunTime.unboxToLong((Object)prev), ref$2));
    }
```

tests/shared/src/test/scala/cats/effect/kernel/RefSpec.scala
```diff
@@ -36,86 +80,90 @@ class RefSpec extends BaseSpec with DetectPlatform { outer =>
 
     // }
 
+    "specialize" in ticked { implicit ticker =>
```
filipwiech, line 83: Would it make sense to add more specialization tests for other constructors introduced in this PR (for example in `Ref` and `Resource`)? 🙂

djspiewak:
I did some bytecode spelunking. I still can't figure out why your new versions aren't faster, but it did occur to me that there are some meaningful limitations to this technique without leveraging `@specialized` on `Ref` itself, which comes with its own problems. In particular, `set` really should be specialized, but can't be without that step. Not tested here though.

I'm going to stare at the bytecode some more. Genuinely this doesn't make sense.

armanbilge:
Yep, I realized that as well. I don't think we can specialize `Ref` anyway because it's an `abstract class` and not a `trait`.

However, we could introduce `IntRef extends Ref` and friends, that expose a specialized overload of `set`. We'd have to expose this type in `Concurrent` and the default implementation would have to be a terrible wrapper around generic `Ref`, however we would override with the optimized version on all the implementations we control.

djspiewak:
Having pondered this for a while… I think this whole idea just isn't going to work. We can't get rid of the boxing on `get` or `set`, which is really where the value would be.

PR #3302, opened 2022-12-05
Add 'IOLocal.lens' method to produce lens 'A <=> B'

seigert:


tests/shared/src/test/scala/cats/effect/IOLocalSpec.scala
```diff
@@ -93,4 +95,86 @@ class IOLocalSpec extends BaseSpec {
     }
   }
 
+  "IOLocal.lens" should {
```
filipwiech: Nice test suite. 👍 Was just wondering, since lens tests are a mirror of the `IOLocal` ones, would it be an option to share them for both classes through some parametrization (both `IOLocalImpl` and `IOLocalLens` extend `IOLocalBase` and `IOLocal`)? 🙂

seigert: Yeah, I've thought of it and was too lazy to do this right away. :)

Anyway, now it's done. Still, I'm not sure that the result is easier to understand than copypasted variant.

filipwiech: Thanks! 👍 
By the way, I just noticed that for the `Ref` lens, the tests do seem to be copied and even kept in a separate suite (there is `RefSpec` and `LensRefSpec`). Maybe there's already a precedent set to favour readability over code duplication in such cases (@armanbilge)? 🤔 I'm very sorry if I just caused a commotion for nothing. 🙁

core/shared/src/main/scala/cats/effect/IOLocal.scala
```diff
@@ -226,36 +259,72 @@ object IOLocal {
...
+    final def scope(value: A): Resource[IO, Unit] =
+      Resource.make(getAndSet(value))(p => set(p)).void
 
-        override def modify[B](f: A => (A, B)): IO[B] =
-          get.flatMap { a =>
-            val (a2, b) = f(a)
-            set(a2).as(b)
-          }
+    final def lens[B](get: A => B)(set: A => B => A): IOLocal[B] = self match {
+      case lens: IOLocalLens[aa, A] =>
```
filipwiech: Just to make sure, is this just an optimization to reduce indirection, or is it required for correctness? Sorry if it should be obvious, just trying to understand. 🙂

seigert: No, it should not be necessary, it's just a small pre-optimization: our getters and setters are pure, so why should we bear the cost of `.map/.flatMap` for every method of inlaid lens `A <=> B <=> C` if the same effect could be observed via functional composition of B & C getters/setters over the same A instance?

filipwiech: Thank you for the explanation! 👍 I would be in favor of including it as a comment right next to this implementation for posterity, if it wouldn't be too much of a troble. 🙂 
Related question: do you think it would be worth it to have some tests for the "lens of a lens" case, or would it be an overkill? 🙂

seigert: ~Done.~ I've added a comment.

As for additional tests, I think it will mostly test type system and `AndThen` implementation. For the former I think we may trust `scalac` and the latter is tested in `cats`. So I think they are unnecessary, but I can add them if needed.

filipwiech: Thanks, much appreciated! 🙂

filipwiech (dismissed):
LGTM. 👍

seigert:
I've rebased this MR onto @armanbilge changes in #3310.

@djspiewak, @armanbilge, any feedback about this MR will greatly appreciated. If such lensing functionality feels out of scope of cats-effects itself, maybe it's better to reject it?

armanbilge:
Thanks, sorry for the conflicts!

> If such lensing functionality feels out of scope of cats-effects itself, maybe it's better to reject it?

`Ref` already offers a similar functionality. And since `IOLocal` is `sealed` this cannot be implemented in userland. Suffice to say I think it is a worthwhile addition.

build.sbt
```diff
@@ -599,7 +599,8 @@ lazy val core = crossProject(JSPlatform, JVMPlatform, NativePlatform)
       // introduced by #3284
       // internal API change
       ProblemFilters.exclude[IncompatibleMethTypeProblem]("cats.effect.CallbackStack.apply"),
-      ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.effect.IOLocal.scope")
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.effect.IOLocal.scope"),
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.effect.IOLocal.lens")
```
armanbilge: Usually we add a small comment here. Seems we forgot for `IOLocal.scope` 🤔

core/shared/src/main/scala/cats/effect/IOLocal.scala
```diff
@@ -226,39 +259,88 @@ object IOLocal {
...
    *   the type of the local value
    */
-  def apply[A](default: A): IO[IOLocal[A]] =
-    IO {
-      new IOLocal[A] { self =>
-        private[this] def getOrDefault(state: IOLocalState): A =
-          state.getOrElse(self, default).asInstanceOf[A]
+  def apply[A](default: A): IO[IOLocal[A]] = IO(new IOLocalImpl(default))
+
+  private[IOLocal] abstract class IOLocalBase[A] extends IOLocal[A] { self =>
```
armanbilge: I wonder if we should just define these methods on `IOLocal` itself?

core/shared/src/main/scala/cats/effect/IOLocal.scala
```diff
@@ -226,39 +259,88 @@ object IOLocal {
...
+        // `set`, `update` or `modify` of resulting lens.
+        // After all, our getters and setters are pure,
+        // so `AndThen` allows us to safely compose them and
+        // proxy calls to the 'original' `IOLocal` independent of
+        // current nesting level.
 
-        override def set(value: A): IO[Unit] =
-          IO.Local(state => (state.updated(self, value), ()))
+        val getter = AndThen(lens.getter).andThen(get)
+        val setter = AndThen((v: aa) => (lens.getter(v), lens.setter(v))).andThen {
```
armanbilge: Are there any stack-safety issues here? Since we are directly invoking `getter(...)` and `setter(...)`

build.sbt
seigert: I've moved both 'scope' and 'lens' directly into 'IOLocal', so this lines become unnecessary...

core/shared/src/main/scala/cats/effect/IOLocal.scala
seigert: ...but, I think it's rather uncommon for cats-effect traits to provide default implementations.

On the other hand, IOLocal is quite special anyway as it uses knowledge about IO internals.

core/shared/src/main/scala/cats/effect/IOLocal.scala
seigert: Yeah, I think you were right, so I've rewrote it w/o direct invocation on first step. 

Sadly, I've could not invent how to do it without additional `Tuple2` instance (as there is no `AndThen` for functions of arbitrary arity).

Also I've added test for it. Current stack height is 16384, which I think is quite enough and it passes in a reasonable amount of time.

djspiewak (dismissed):
Outstanding. Thank you!

Issue #3288, opened 2022-11-27
Blocked worker thread detector


djspiewak:
Vert.x does something a bit like this.

The starvation detector added by @TimWSpence does a really good job of detecting when the compute runtime is starving for *some* reason, but that reason could be anything from poor task granularity to hard-blocked threads and anything in between, and there simply is not enough information to take action. There is no way to drive to a root cause, much less a solution.

In general, *most* such starvation cases need to be solved in one of two ways:

- If it's a CPU bound task, additional `cede`s and smaller task granularity (breaking a single `delay` into multiple) is likely to help
- If it's a hard-blocking task, `blocking` or `interruptible` are the solution

In both cases, this may not be enough, since some semantic gating or batching (such as a semaphore or queue) will likely also be required, but this is the first step. Determining which step to take, and where to take it, is where we should be concentrating our efforts.

To that end, it would be beneficial to have an off-by-default blocked thread detector. This should be controlled with a boolean in `IOApp` which in turn is passed to the `WorkStealingThreadPool`. When `true`, any worker thread which enters state `0`, in addition to polling the external queue, should call `Thread#getStatus` on a random *other* worker thread in the pool. If that thread is blocked, call `getStackTrace()` and print a warning to stderr reporting the blocked state along with the stack trace. Additionally, it would be helpful to add some informative text. Telling users that they should swap a `delay` for `blocking`/`interruptible` is almost always going to be the right course of action.

This implementation is going to impose meaningful overhead on the workers, which is part of why it should be *disabled* by default. We can add to the existing starvation checker's informational text to instruct users to try enabling this mode as a diagnostic tool.

TimWSpence:
@djspiewak You assume that the problem isn't just excessive GC and nobody noticed 😅 Hpyothetically peaking of course 😉 I can have a go at this if people want? It would be super useful for us as we wrap a load of Java libraries for various things and its not always obvious when they block

djspiewak:
Go for it!

TimWSpence:
So sorry for the delay @djspiewak! Finally got to looking at this and wanted to check that [this](https://github.com/typelevel/cats-effect/compare/series/3.x...TimWSpence:cats-effect:thread-blocking-detection?expand=1) is roughly what you envisioned?

PR #3305, opened 2022-12-05
More detailed warning for starvation checker

armanbilge:
The warning is now prefixed with:
> Your app's responsiveness to a new asynchronous event (such as a new connection, an upstream response, or a timer) was in excess of $threshold.

I think this will help with understanding what this warning actually _means_.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/CpuStarvationCheck.scala
+++ b/core/shared/src/main/scala/cats/effect/CpuStarvationCheck.scala
@@ -21,28 +21,36 @@ import cats.effect.std.Console
 import cats.effect.unsafe.IORuntimeConfig
 import cats.syntax.all._
 
-import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.duration.{Duration, FiniteDuration}
 
 private[effect] object CpuStarvationCheck {
 
   def run(runtimeConfig: IORuntimeConfig, metrics: CpuStarvationMetrics): IO[Nothing] = {
+    import runtimeConfig._
+
+    val threshold = cpuStarvationCheckInterval * (1 + cpuStarvationCheckThreshold)
+
+    val warning = mkWarning(cpuStarvationCheckInterval * cpuStarvationCheckThreshold)
+
     def go(initial: FiniteDuration): IO[Nothing] =
-      IO.sleep(runtimeConfig.cpuStarvationCheckInterval) >> IO.monotonic.flatMap { now =>
+      IO.sleep(cpuStarvationCheckInterval) >> IO.monotonic.flatMap { now =>
         val delta = now - initial
 
-        metrics.recordClockDrift(delta - runtimeConfig.cpuStarvationCheckInterval) >>
-          (Console[IO].errorln(warning) >> metrics.incCpuStarvationCount).whenA(delta >=
-            runtimeConfig.cpuStarvationCheckInterval * (1 + runtimeConfig.cpuStarvationCheckThreshold)) >>
+        metrics.recordClockDrift(delta - cpuStarvationCheckInterval) >>
+          (Console[IO].errorln(warning) *> metrics.incCpuStarvationCount)
+            .whenA(delta >= threshold) >>
           go(now)
       }
 
-    IO.monotonic.flatMap(go(_)).delayBy(runtimeConfig.cpuStarvationCheckInitialDelay)
+    IO.monotonic.flatMap(go(_)).delayBy(cpuStarvationCheckInitialDelay)
   }
 
-  private[this] val warning =
-    """[WARNING] Your CPU is probably starving. Consider increasing the granularity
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/effect/CpuStarvationCheck.scala
```diff
@@ -21,28 +21,36 @@ import cats.effect.std.Console
...
+    IO.monotonic.flatMap(go(_)).delayBy(cpuStarvationCheckInitialDelay)
   }
 
-  private[this] val warning =
-    """[WARNING] Your CPU is probably starving. Consider increasing the granularity
-      |of your delays or adding more cedes. This may also be a sign that you are
-      |unintentionally running blocking I/O operations (such as File or InetAddress)
-      |without the blocking combinator.""".stripMargin
+  private[this] def mkWarning(threshold: Duration) =
+    s"""|[WARNING] Your app's responsiveness to a new asynchronous event (such as a
```
armanbilge, line 49: I worry if "responsiveness" can be confusing. It's not how long your app took to respond to an event; rather it's how long your app took to even notice that event.

core/shared/src/main/scala/cats/effect/CpuStarvationCheck.scala
```diff
@@ -21,28 +21,36 @@ import cats.effect.std.Console
...
+        |new connection, an upstream response, or a timer) was in excess of $threshold.
```
armanbilge, line 50: If we wanted we could even report the measured latency.

core/shared/src/main/scala/cats/effect/CpuStarvationCheck.scala
filipwiech: Maybe it would be a good idea to introduce a documentation page that explains the starvation checker, how to use it, how it works, what does this message mean and how the user can try to remedy the problem, especially with the upcoming blocked worker thread detector (#3288)? 🙂

armanbilge: Indeed, Daniel is working on some docs right now :)

core/shared/src/main/scala/cats/effect/CpuStarvationCheck.scala
```diff
@@ -21,28 +21,36 @@ import cats.effect.std.Console
...
+import scala.concurrent.duration.{Duration, FiniteDuration}
 
 private[effect] object CpuStarvationCheck {
 
   def run(runtimeConfig: IORuntimeConfig, metrics: CpuStarvationMetrics): IO[Nothing] = {
+    import runtimeConfig._
+
+    val threshold = cpuStarvationCheckInterval * (1 + cpuStarvationCheckThreshold)
+
+    val warning = mkWarning(cpuStarvationCheckInterval * cpuStarvationCheckThreshold)
```
filipwiech, line 33: Just in case: should we pass the `threshold` here? The argument name in the `mkWarning(...)` and the `whenA(delta >= threshold)` check would suggest so, but maybe I'm just misunderstaning. 🙂

tests/jvm/src/test/scala/cats/effect/IOAppSpec.scala
```diff
@@ -200,7 +200,7 @@ class IOAppSpec extends Specification {
           h.awaitStatus()
           val err = h.stderr()
           err must not(contain("[WARNING] Failed to register Cats Effect CPU"))
-          err must contain("[WARNING] Your CPU is probably starving")
+          err must contain("[WARNING] Your app's responsiveness")
```
filipwiech, line 203: Would it be possible to check that the message contains computed threshold with a reasonable value? 🙂

filipwiech:
General comment: in the #3288 @TimWSpence had a good observation: that excessive GC activity under memory pressure can be one potential cause of the CPU starvation. Since this is still within the bounds of the application behaviour (and not completely external factor out of our scope) maybe it would make sense to add this reason to the diagnostic message, WDYT? 🙂

core/shared/src/main/scala/cats/effect/CpuStarvationCheck.scala
filipwiech: I'd say it wouldn't hurt and could be a helpful data point. 👍

buntec:
This is really nitpicking, but as discussed offline with @armanbilge, saying that the "CPU is probably starving" seems not very accurate. In fact, on the JVM your CPU could be at 100% utilization if you are doing CPU bound work, and on JS platforms this terminology is meaningless because we are talking about a single OS thread. Perhaps it would be better to say that the CE runtime is unable to schedule fibers in a fair and timely manner because it is "starving for OS threads", or something along those lines.

djspiewak (approved):
This is a good improvement in and of itself. We should continue to iterate. @filipwiech had some excellent ideas in here that are worth including in follow-up PRs.

PR #3307, opened 2022-12-06
Fix propagation of `ExitCase` in `Resource#{both,combineK}`

armanbilge:
Noticed this while looking into https://github.com/typelevel/cats-effect/pull/3227. `ExitCase`s were not being propagated at all to the finalizers.

### DIFF
--- a/tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
@@ -756,10 +756,31 @@ class ResourceSpec extends BaseSpec with ScalaCheck with Discipline {
           got mustEqual ExitCase.Succeeded
         }
 
+        "left errored, use errored, test right" >> ticked { implicit ticker =>
+          var got: ExitCase = null
+          val ex = new Exception
+          val r = Resource.onFinalizeCase(ec => IO { got = ec })
+          Resource
+            .eval(IO.raiseError(new Exception))
+            .combineK(r)
+            .use(_ => IO.raiseError(ex)) must failAs(ex)
+          got mustEqual ExitCase.Errored(ex)
+        }
+
         "use canceled, test left" >> ticked { implicit ticker =>
           var got: ExitCase = null
           val r = Resource.onFinalizeCase(ec => IO { got = ec })
           r.combineK(Resource.unit).use(_ => IO.canceled) must selfCancel
+          got mustEqual ExitCase.Canceled
+        }
+
+        "left errored, use canceled, test right" >> ticked { implicit ticker =>
+          var got: ExitCase = null
+          val r = Resource.onFinalizeCase(ec => IO { got = ec })
+          Resource
+            .eval(IO.raiseError(new Exception))
+            .combineK(r)
+            .use(_ => IO.canceled) must selfCancel
           got mustEqual ExitCase.Canceled
         }
       }
--- a/tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
@@ -599,7 +599,7 @@ class ResourceSpec extends BaseSpec with ScalaCheck with Discipline {
         rightReleased must beTrue
       }
 
-      "passes along the exit case" in {
+      "propagate the exit case" in {
         import Resource.ExitCase
 
         "use succesfully, test left" >> ticked { implicit ticker =>
@@ -721,7 +721,7 @@ class ResourceSpec extends BaseSpec with ScalaCheck with Discipline {
         }
       }
 
-      "passes along the exit case" in {
+      "propagate the exit case" in {
         import Resource.ExitCase
 
         "use succesfully, test left" >> ticked { implicit ticker =>
--- a/tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
@@ -598,6 +598,84 @@ class ResourceSpec extends BaseSpec with ScalaCheck with Discipline {
         leftReleased must beTrue
         rightReleased must beTrue
       }
+
+      "passes along the exit case" in ticked { implicit ticker =>
+        import Resource.ExitCase
+
+        { // use successfully, test left
+          var got: ExitCase = null
+          val r = Resource.onFinalizeCase(ec => IO { got = ec })
+          r.both(Resource.unit).use(_ => IO.unit) must completeAs(())
+          got mustEqual ExitCase.Succeeded
+        }
+
+        { // use successfully, test right
+          var got: ExitCase = null
+          val r = Resource.onFinalizeCase(ec => IO { got = ec })
+          Resource.unit.both(r).use(_ => IO.unit) must completeAs(())
+          got mustEqual ExitCase.Succeeded
+        }
+
+        { // use errored, test left
+          var got: ExitCase = null
+          val ex = new Exception
+          val r = Resource.onFinalizeCase(ec => IO { got = ec })
+          r.both(Resource.unit).use(_ => IO.raiseError(ex)) must failAs(ex)
+          got mustEqual ExitCase.Errored(ex)
+        }
+
+        { // use errored, test right
+          var got: ExitCase = null
+          val ex = new Exception
+          val r = Resource.onFinalizeCase(ec => IO { got = ec })
+          Resource.unit.both(r).use(_ => IO.raiseError(ex)) must failAs(ex)
+          got mustEqual ExitCase.Errored(ex)
+        }
+
+        { // right errored, test left
+          var got: ExitCase = null
+          val ex = new Exception
+          val r = Resource.onFinalizeCase(ec => IO { got = ec })
+          r.both(Resource.eval(IO.sleep(1.second) *> IO.raiseError(ex))).use_ must failAs(ex)
+          got mustEqual ExitCase.Errored(ex)
+        }
+
+        { // left errored, test right
+          var got: ExitCase = null
+          val ex = new Exception
+          val r = Resource.onFinalizeCase(ec => IO { got = ec })
+          Resource.eval(IO.sleep(1.second) *> IO.raiseError(ex)).both(r).use_ must failAs(ex)
+          got mustEqual ExitCase.Errored(ex)
+        }
+
+        { // use canceled, test left
+          var got: ExitCase = null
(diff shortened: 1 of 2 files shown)


tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
```diff
@@ -642,6 +720,49 @@ class ResourceSpec extends BaseSpec with ScalaCheck with Discipline {
...
+        }
+
+        "left errored, test left" >> ticked { implicit ticker =>
+          var got: ExitCase = null
+          val ex = new Exception
+          val r = Resource.onFinalizeCase(ec => IO { got = ec }) *>
+            Resource.eval(IO.raiseError(ex))
+          r.combineK(Resource.unit).use_ must completeAs(())
+          got mustEqual ExitCase.Succeeded
+        }
```
armanbilge, line 749: This semantic is kind of counter-intuitive. Even though the left `Resource` errors, ultimately the `combineK`ed `Resource` is acquired and used successfully. So the left finalizer is passed `Succeeded`.

This is a by-product of the fact that the left `Resource` is not released eagerly even if it errors. Maybe this semantic is undesirable?

filipwiech: After mulling this over I think it makes sense and is expected, although I could be wrong (I went back and forth on this in my mind a couple times before settling on the conclusion). It would be consistent with what `Resource.both` is doing, the reasoning being as follows:

`Resource.both` requires both Resources to be acquired successfully - if any of them errors out (or gets canceled), then all of them get an `ExitCase.Errored` (or `ExitCase.Canceled` respectively), even the one that succeded, because the operation as a whole has failed.

Conversely, `Resource.combineK` needs any of the combined Resources to be acquired successfully - if just one of them succedes, then all of them (well, the ones that were actually evaluated) get an `ExitCase.Succeeded`, even the ones that failed, because the operation as a whole has succeded.

The fact that `ExitCase` from the `use` operation propagates to all the input Resources, regardless of their own, seems to be aligned as well.

I hope it makes sense. In any case, whichever semantic will be introduced, it could be a good idea to document it in the scaladoc, especially if it seems to be counter-intuitive. 🙂 

However, I'm not sure if any sort of consistency between `Reource.both` and `combineK` is desired. There's also the case of `Resource.race` which will release loser eagerly in #3226.

tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
```diff
@@ -598,6 +598,84 @@ class ResourceSpec extends BaseSpec with ScalaCheck with Discipline {
...
+
+        "use errored, test right" >> ticked { implicit ticker =>
+          var got: ExitCase = null
+          val ex = new Exception
+          val r = Resource.onFinalizeCase(ec => IO { got = ec })
+          Resource.unit.both(r).use(_ => IO.raiseError(ex)) must failAs(ex)
+          got mustEqual ExitCase.Errored(ex)
+        }
+
+        "right errored, test left" >> ticked { implicit ticker =>
```
filipwiech, line 635: Would it be a good idea to describe exit case propagation semantics in the scaladoc for the `Resource#both`? 🙂

tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
```diff
@@ -642,6 +720,49 @@ class ResourceSpec extends BaseSpec with ScalaCheck with Discipline {
...
+
+        "left errored, test right" >> ticked { implicit ticker =>
+          var got: ExitCase = null
+          val ex = new Exception
+          val r = Resource.onFinalizeCase(ec => IO { got = ec })
+          Resource.eval(IO.raiseError(ex)).combineK(r).use_ must completeAs(())
+          got mustEqual ExitCase.Succeeded
+        }
+
+        "use canceled, test left" >> ticked { implicit ticker =>
```
filipwiech, line 770: Would it make sense to add cases for "use errored/canceled, test right" (for when left Resource fails to acquire, but right one succedes)? 🙂

tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
filipwiech: > The fact that ExitCase from the use operation propagates to all the input Resources, regardless of their own, seems to be aligned as well.

One more thought regarding this: one could say that in the case of `combineK` only one Resource is used in the end (similarly to the winner of `Resource.race` and unlike `Resource.both` which utilizes all input Resources), so there's no need to propagate `ExitCase`  from `use` to the one that wasn't. So I guess that could be the reason to maybe release eagerly in `combineK`. Sorry for the rambling, perhaps I'm not as convinced as I was before. 😄

tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
```diff
@@ -598,6 +598,84 @@ class ResourceSpec extends BaseSpec with ScalaCheck with Discipline {
...
+        }
+
+        "use canceled, test right" >> ticked { implicit ticker =>
+          var got: ExitCase = null
+          val r = Resource.onFinalizeCase(ec => IO { got = ec })
+          Resource.unit.both(r).use(_ => IO.canceled) must selfCancel
+          got mustEqual ExitCase.Canceled
+        }
+
+        "right canceled, test left" >> ticked { implicit ticker =>
```
filipwiech, line 665: What would happen if right was canceled and left errored out (and vice versa)? Would they both get `ExitCase.Errored`? Maybe it could be a good idea to add tests for those cases. 🙂

tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
```diff
@@ -642,6 +720,49 @@ class ResourceSpec extends BaseSpec with ScalaCheck with Discipline {
...
+        "left errored, test left" >> ticked { implicit ticker =>
+          var got: ExitCase = null
+          val ex = new Exception
+          val r = Resource.onFinalizeCase(ec => IO { got = ec }) *>
+            Resource.eval(IO.raiseError(ex))
+          r.combineK(Resource.unit).use_ must completeAs(())
+          got mustEqual ExitCase.Succeeded
+        }
+
+        "left errored, test right" >> ticked { implicit ticker =>
```
filipwiech, line 751: In case one of the Resources errors out during acqusition and the other one self-cancels, will the `ExitCase` of the right Resource propagate/overwrite the result on the left? Could be worth it to add tests for those scenarios. 🙂

tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
```diff
@@ -642,6 +720,49 @@ class ResourceSpec extends BaseSpec with ScalaCheck with Discipline {
...
+
+        "use errored, test left" >> ticked { implicit ticker =>
+          var got: ExitCase = null
+          val ex = new Exception
+          val r = Resource.onFinalizeCase(ec => IO { got = ec })
+          r.combineK(Resource.unit).use(_ => IO.raiseError(ex)) must failAs(ex)
+          got mustEqual ExitCase.Errored(ex)
+        }
+
+        "left errored, test left" >> ticked { implicit ticker =>
```
filipwiech, line 742: Would it make sense to add tests for "left canceled, test left/right" too? 🙂

djspiewak (changes requested):
Adding a Request Changes for @filipwiech's suggestions, which are all excellent

armanbilge:
@djspiewak does that mean you think we should change the `combineK` semantic?

tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
armanbilge: If left cancels, no resources are acquired, so no finalizers are run.

tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
armanbilge: > What would happen if right was canceled and left errored out (and vice versa)?

Depends which happens first. If right cancels first, it's covered by this test. If left errors first it's covered by "left errored, test right"

tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
armanbilge: If a resource errors out or cancels during acquisition, then it has no exit case because the finalizer never runs.

tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
armanbilge: Yes it would!

tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
armanbilge: Yes, even in two weeks I've forgotten what it was 😂

tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
djspiewak: I feel like this is somewhat similar to the following:

```scala
case object TestException extends RuntimeException

Resource.raiseError[IO, Unit, Throwable](TestException)
  .onFinalizeCase(IO.println(_))
  .voidError
```

Do we print out `Succeeded` here? If so, then this `combineK` semantic is reasonable. If not, then we're being inconsistent.

armanbilge: The following program doesn't print.

```scala
//> using lib "org.typelevel::cats-effect::3.4.2"

import cats.effect._
import cats.syntax.all._

object App extends App {
  case object TestException extends RuntimeException

  def run = Resource
    .raiseError[IO, Unit, Throwable](TestException)
    .onFinalizeCase(IO.println(_))
    .voidError
    .use_

}
```

djspiewak: It prints for me? `Succeeded`. Here's an even more interesting example:

```scala
(Resource.unit.onFinalizeCase(IO.println(_)) *> Resource.raiseError[IO, Unit, Throwable](TestException)).voidError.use_
```

Prints `Succeeded`. Meanwhile:

```scala
(Resource.unit.onFinalizeCase(IO.println(_)) *> Resource.raiseError[IO, Unit, Throwable](TestException)).use_ 
```

Prints `Errored(...)`.

I think `combineK` is fine as-is. It's rather weird, but we're at least consistent with the backpropagation of final state.

djspiewak (approved):

PR #3310, opened 2022-12-06
`IOLocal` micro-optimizations

armanbilge:
All methods can be expressed as a single `IO.Local(state => ...)` op.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOLocal.scala
+++ b/core/shared/src/main/scala/cats/effect/IOLocal.scala
@@ -16,6 +16,8 @@
 
 package cats.effect
 
+import scala.collection.immutable.Map
+
 /**
  * [[IOLocal]] provides a handy way of manipulating a context on different scopes.
  *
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/cats/effect/IOLocal.scala
```diff
@@ -206,29 +208,32 @@ object IOLocal {
   def apply[A](default: A): IO[IOLocal[A]] =
     IO {
       new IOLocal[A] { self =>
+        private[this] def getOrDefault(state: Map[IOLocal[_], Any]): A =
```
filipwiech: Would it be better to use `IOLocalState` type here, rather than a `Map`? 🙂

armanbilge: Oh, this thing! I didn't realize we had an alias, thanks :)
https://github.com/typelevel/cats-effect/blob/832079aecbbf956fb18f11392cfeec6cdaa52915/core/shared/src/main/scala/cats/effect/package.scala#L82

filipwiech: Wow, your replies are lightning fast. 😄 
Yep, it's used in the `IO.Local`. 👍

filipwiech (approved):
It's an interesting trade-off between an optimization and, for the lack of better word, simplicity (defining a couple of basic operations and implementing the rest through composition). Nevertheless I'd say it looks good, it would be interesting to maybe get some benchmark numbers. 👍

core/shared/src/main/scala/cats/effect/IOLocal.scala
```diff
@@ -206,29 +206,32 @@ object IOLocal {
...
+            (state.updated(self, a2), b)
           }
 
         override def getAndSet(value: A): IO[A] =
-          get <* set(value)
+          IO.Local(state => (state.updated(self, value), getOrDefault(state)))
 
         override def getAndReset: IO[A] =
-          get <* reset
+          IO.Local(state => (state - self, getOrDefault(state)))
```
seigert, line 234: Maybe replace `-` with `removed` then? It's more like `.updated` and wont depend on `-` inlining?.

armanbilge: Thanks, but I don't think such method exists (at least in 2.12).
https://www.scala-lang.org/api/2.12.17/scala/collection/immutable/Map.html

djspiewak (approved):
Would like to see some benchmarks, but this is good either way!

PR #3315, opened 2022-12-12
Use configurable `reportFailure` for `MainThread`

armanbilge:


filipwiech (dismissed):
Would it be possible to introduce a simple unit test for this change? 🙂

djspiewak (changes requested):
What @filipwiech said :-)

tests/jvm/src/main/scala/catseffect/examplesplatform.scala
```diff
@@ -75,4 +76,17 @@ package examples {
...
+    val exitCode = new AtomicReference[ExitCode](ExitCode.Error)
+
+    override def reportFailure(err: Throwable): IO[Unit] =
+      IO(exitCode.set(ExitCode.Success))
+
+    def run(args: List[String]): IO[ExitCode] =
+      IO.raiseError(new Exception).startOn(MainThread) *>
+        IO.sleep(100.millis) *> IO(exitCode.get)
+
+  }
```
armanbilge, line 91: I wouldn't consider this a simple test exactly (testing `IOApp` never is), but here we are :)

djspiewak (approved):

PR #3317, opened 2022-12-16
Implement `IORuntimeMetrics`

iRevive:
There are a few notable notes regarding this PR:
- The existing implementation of `CpuStarvation` MBean somewhat breaks the consistency. All previous MBeans were under `cats.effect.unsafe.metrics` package, while `CpuStarvation` is available under `cats.effect.metrics`. I didn't update the MBean name yet.
- Platform-specific metrics can be defined in the `IORuntimeMetricsPlatform`. For example, JVM `IORuntimeMetrics` offers `compute` and `cpuStarvation` metrics. While JS and Native have only `cpuStarvation`.

Personally, the whole implementation feels clunky, hopefully we can find the right direction together.

This PR will not pass MiMa checks. Once we agree on a direction, I will address the bin compat issue.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
+++ b/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
@@ -18,6 +18,7 @@ package cats.effect
 package unsafe
 
 import cats.effect.Platform.static
+import cats.effect.metrics.IORuntimeMetrics
 
 import scala.concurrent.ExecutionContext
 
@@ -43,7 +44,8 @@ final class IORuntime private[unsafe] (
     private[effect] val pollers: List[Any],
     private[effect] val fiberMonitor: FiberMonitor,
     val shutdown: () => Unit,
-    val config: IORuntimeConfig
+    val config: IORuntimeConfig,
+    val metrics: IORuntimeMetrics
 ) {
 
   private[effect] val fiberErrorCbs: StripedHashtable = new StripedHashtable()
@@ -70,6 +72,7 @@ object IORuntime extends IORuntimeCompanionPlatform {
       config: IORuntimeConfig): IORuntime = {
     val fiberMonitor = FiberMonitor(compute)
     val unregister = registerFiberMonitorMBean(fiberMonitor)
+    val metrics = IORuntimeMetrics(compute)
     def unregisterAndShutdown: () => Unit = () => {
       unregister()
       shutdown()
@@ -84,7 +87,8 @@ object IORuntime extends IORuntimeCompanionPlatform {
         pollers,
         fiberMonitor,
         unregisterAndShutdown,
-        config)
+        config,
+        metrics)
     allRuntimes.put(runtime, runtime.hashCode())
     runtime
   }
@@ -104,14 +108,19 @@ object IORuntime extends IORuntimeCompanionPlatform {
       scheduler: Scheduler,
       fiberMonitor: FiberMonitor,
       shutdown: () => Unit,
-      config: IORuntimeConfig): IORuntime =
-    new IORuntime(compute, blocking, scheduler, Nil, fiberMonitor, shutdown, config)
+      config: IORuntimeConfig): IORuntime = {
+    val metrics = IORuntimeMetrics(compute)
+    new IORuntime(compute, blocking, scheduler, Nil, fiberMonitor, shutdown, config, metrics)
+  }
 
   def builder(): IORuntimeBuilder =
     IORuntimeBuilder()
 
-  private[effect] def testRuntime(ec: ExecutionContext, scheduler: Scheduler): IORuntime =
-    new IORuntime(ec, ec, scheduler, Nil, new NoOpFiberMonitor(), () => (), IORuntimeConfig())
+  private[effect] def testRuntime(ec: ExecutionContext, scheduler: Scheduler): IORuntime = {
+    val config = IORuntimeConfig()
+    val metrics = IORuntimeMetrics(ec)
+    new IORuntime(ec, ec, scheduler, Nil, new NoOpFiberMonitor(), () => (), config, metrics)
+  }
 
   @static private[effect] final val allRuntimes: ThreadSafeHashtable[IORuntime] =
     new ThreadSafeHashtable(4)
--- a/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
+++ b/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
@@ -18,7 +18,7 @@ package cats.effect
 package unsafe
 
 import cats.effect.Platform.static
-import cats.effect.metrics.IORuntimeMetrics
+import cats.effect.unsafe.metrics.IORuntimeMetrics
 
 import scala.concurrent.ExecutionContext
 
--- a/build.sbt
+++ b/build.sbt
@@ -671,7 +671,12 @@ lazy val core = crossProject(JSPlatform, JVMPlatform, NativePlatform)
       ProblemFilters.exclude[DirectMissingMethodProblem]("cats.effect.IOLocal.lens"),
       // internal API change, makes CpuStarvationMetrics available on all platforms
       ProblemFilters.exclude[MissingClassProblem](
-        "cats.effect.metrics.JvmCpuStarvationMetrics$NoOpCpuStarvationMetrics")
+        "cats.effect.metrics.JvmCpuStarvationMetrics$NoOpCpuStarvationMetrics"),
+      ProblemFilters.exclude[MissingClassProblem]("cats.effect.metrics.CpuStarvationMetrics"),
+      // package-private classes moved to the `cats.effect.unsafe.metrics` package
+      ProblemFilters.exclude[MissingClassProblem]("cats.effect.metrics.CpuStarvation"),
+      ProblemFilters.exclude[MissingClassProblem]("cats.effect.metrics.CpuStarvation$"),
+      ProblemFilters.exclude[MissingClassProblem]("cats.effect.metrics.CpuStarvationMBean")
     ) ++ {
       if (tlIsScala3.value) {
         // Scala 3 specific exclusions
--- a/build.sbt
+++ b/build.sbt
@@ -668,7 +668,10 @@ lazy val core = crossProject(JSPlatform, JVMPlatform, NativePlatform)
       ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.effect.IOLocal.reset"),
       ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.effect.IOLocal.lens"),
       // this filter is particulary terrible, because it can also mask real issues :(
-      ProblemFilters.exclude[DirectMissingMethodProblem]("cats.effect.IOLocal.lens")
+      ProblemFilters.exclude[DirectMissingMethodProblem]("cats.effect.IOLocal.lens"),
+      // internal API change, makes CpuStarvationMetrics available on all platforms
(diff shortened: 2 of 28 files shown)


core/jvm/src/main/scala/cats/effect/unsafe/metrics/CpuStarvationSamplerMBean.scala
```diff
@@ -0,0 +1,87 @@
...
+    s"""[WARNING] Failed to register Cats Effect CPU starvation MBean, proceeding with
+       |no-operation versions. You will not see MBean metrics for CPU starvation.
+       |Exception follows: \n ${exceptionWriter.toString}
+       |""".stripMargin
+  }
+
+  private[effect] def register(sampler: CpuStarvationSampler): Resource[IO, Unit] = {
+    val acquire: IO[MBeanServer] =
+      for {
+        mBeanServer <- IO.delay(ManagementFactory.getPlatformMBeanServer)
```
iRevive: Should we register **all** mbeans at once somewhere in the IOApp? It will require quite a lot of changes, though.

core/jvm/src/main/scala/cats/effect/unsafe/metrics/CpuStarvationSamplerPlatform.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  final def getCpuStarvationCount(): Long = this.cpuStarvationCount()
+  final def getMaxClockDriftMs(): Long = this.maxClockDriftMs()
+  final def getCurrentClockDriftMs(): Long = this.currentClockDriftMs()
+}
+
+private[metrics] trait CpuStarvationSamplerCompanionPlatform {
+  this: CpuStarvationSampler.type =>
+
+  private[effect] def create(): CpuStarvationSampler = {
```
iRevive, line 36: All platforms offer identical implementations. Therefore it can be in `CpuStarvationSampler` in the `shared` module. 

Will we benefit from having platform-specific implementations? Perhaps JS or Native can be somewhat optimized?

core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
```diff
@@ -40,7 +42,9 @@ final class IORuntime private[unsafe] (
     val scheduler: Scheduler,
     private[effect] val fiberMonitor: FiberMonitor,
     val shutdown: () => Unit,
-    val config: IORuntimeConfig
+    val config: IORuntimeConfig,
+    private[effect] val cpuStarvationSampler: CpuStarvationSampler,
```
iRevive: The ugly thing I don't like. Alternatively, `cpuStarvationSampler` can be exposed by `IORuntimeMetrics`.

core/shared/src/main/scala/cats/effect/unsafe/IORuntimeMetrics.scala
```diff
@@ -0,0 +1,49 @@
...
+
+  private[effect] object CPUStarvationMetrics {
+    def fromMetrics(metrics: CpuStarvationSampler): CPUStarvationMetrics =
+      new CPUStarvationMetrics {
+        def starvationCount(): Long = metrics.cpuStarvationCount()
+        def maxClockDriftMs(): Long = metrics.maxClockDriftMs()
+        def currentClockDriftMs(): Long = metrics.currentClockDriftMs()
+      }
+
+    def noop: CPUStarvationMetrics =
```
iRevive: `noop` is used only by `IORuntime.testRuntime`

iRevive:
And a small demo:
<details>
 <summary>Demo</summary>
 
```scala
import cats.effect.{IO, IOApp}
import cats.effect.std.Random
import cats.effect.unsafe.IORuntimeMetrics
import cats.syntax.foldable._

object Test extends IOApp.Simple {

  def run: IO[Unit] = {
    IO.delay(print(runtime.metrics)).flatMap(IO.println).delayBy(500.millis).foreverM.background.surround {
      Random.scalaUtilRandom[IO].flatMap { random =>
        val a = random.betweenInt(10, 3000).flatMap { int =>
          if (int % 2 == 0) IO.blocking(Thread.sleep(int)) else IO.delay(Thread.sleep(int))
        }.start.replicateA(100)

        a.flatMap(_.traverse_(_.join))
      } >> IO.never
    }
  }
  
  def print(metrics: IORuntimeMetrics): String = {
    s"""
      |Compute:
      |workerThreadCount: ${metrics.compute.map(_.workerThreadCount())}
      |activeThreadCount: ${metrics.compute.map(_.activeThreadCount())}
      |searchingThreadCount: ${metrics.compute.map(_.searchingThreadCount())}
      |blockedWorkerThreadCount: ${metrics.compute.map(_.blockedWorkerThreadCount())}
      |localQueueFiberCount: ${metrics.compute.map(_.localQueueFiberCount())}
      |suspendedFiberCount: ${metrics.compute.map(_.suspendedFiberCount())}
      |
      |CPU Starvation:
      |starvationCount: ${metrics.cpuStarvation.starvationCount()}
      |maxClockDriftMs: ${metrics.cpuStarvation.maxClockDriftMs()}
      |currentClockDriftMs: ${metrics.cpuStarvation.currentClockDriftMs()}
      |""".stripMargin
  }
}
```

Output:
```
Compute:
workerThreadCount: Some(8)
activeThreadCount: Some(8)
searchingThreadCount: Some(0)
blockedWorkerThreadCount: Some(13)
localQueueFiberCount: Some(0)
suspendedFiberCount: Some(2)

CPU Starvation:
starvationCount: 0
maxClockDriftMs: 0
currentClockDriftMs: 0

....

Compute:
workerThreadCount: Some(8)
activeThreadCount: Some(1)
searchingThreadCount: Some(0)
blockedWorkerThreadCount: Some(13)
localQueueFiberCount: Some(0)
suspendedFiberCount: Some(2)

CPU Starvation:
starvationCount: 0
maxClockDriftMs: 11
currentClockDriftMs: 3
```

</details>

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeMetricsCompanionPlatform.scala
```diff
@@ -0,0 +1,66 @@
...
+  ): IORuntimeMetrics =
+    new IORuntimeMetrics {
+      val cpuStarvation: CPUStarvationMetrics = CPUStarvationMetrics.fromMetrics(starvation)
+      val compute: Option[ComputeMetrics] = computeMetrics(ec)
+    }
+
+  private[unsafe] def noop: IORuntimeMetrics =
+    new IORuntimeMetrics {
+      val compute: Option[ComputeMetrics] = None
+      val cpuStarvation: CPUStarvationMetrics = CPUStarvationMetrics.noop
```
armanbilge: Maybe `CPUStarvationMetrics` should be an `Option` too?

iRevive: That's debatable 😬

If I get it right, the `CPUStarvationMetrics` is always present in the production mode (even with a custom runtime). Unless you decide to use test runtime explicitly.

On the other hand, Option offers better integrity.

iRevive: According to this discussion https://github.com/typelevel/cats-effect/issues/3321, the Option approach is justified

iRevive:
Mima is sad:
```
[error] cats-effect: Failed binary compatibility check against org.typelevel:cats-effect_3:3.4.2 (e:info.apiURL=https://typelevel.org/cats-effect/api/3.x/, e:info.versionScheme=early-semver)! Found 7 potential problems (filtered 2)
[error]  * class cats.effect.metrics.CpuStarvation does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("cats.effect.metrics.CpuStarvation")
[error]  * object cats.effect.metrics.CpuStarvation does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("cats.effect.metrics.CpuStarvation$")
[error]  * interface cats.effect.metrics.CpuStarvationMBean does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("cats.effect.metrics.CpuStarvationMBean")
[error]  * interface cats.effect.metrics.CpuStarvationMetrics does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("cats.effect.metrics.CpuStarvationMetrics")
[error]  * class cats.effect.metrics.JvmCpuStarvationMetrics does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("cats.effect.metrics.JvmCpuStarvationMetrics")
[error]  * object cats.effect.metrics.JvmCpuStarvationMetrics does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("cats.effect.metrics.JvmCpuStarvationMetrics$")
[error]  * class cats.effect.metrics.JvmCpuStarvationMetrics#NoOpCpuStarvationMetrics does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("cats.effect.metrics.JvmCpuStarvationMetrics$NoOpCpuStarvationMetrics")
```

Since these classes were private to `metrics` package, I assume it's safe to add an exclusion.

iRevive:
With https://github.com/typelevel/cats-effect/pull/3526 the linking issue in ScalaJS is solved.

djspiewak (commented):
Hey it only took me two years to get to this. New record?

Sorry. :( So the implementation overall looks good. I think this sort of thing is always somewhat clunky so that's fine. We should probably make the mbean paths consistent and just publish the change in the release notes (so people know to update any monitoring keys). The unsafe packages are similarly annoying but I think the optimal solution to this.

So long as all the mbean paths are made consistent, I'm pretty okay merging this. Anyone have any strong objections?

iRevive:
@djspiewak thanks for the review! Better late than never :)

I changed the namespace of the CPUStarvation MBean to `cats.effect.unsafe.metrics`.

The names are the following:
- CPU starvation: `cats.effect.unsafe.metrics:type=CpuStarvation`
- WSTP compute pool: `cats.effect.unsafe.metrics:type=ComputePoolSampler-$hash`
- WSTP local queue: `cats.effect.unsafe.metrics:type=LocalQueueSampler-$hash-$i`
- Live fiber dump: `cats.effect.unsafe.metrics:type=LiveFiberSnapshotTrigger-$hash`

djspiewak:
Looks good! Once we sort the merge conflicts we can land it.

iRevive:
Done. ~~Waiting for CI.~~ Nvm, waiting for https://github.com/typelevel/cats-effect/pull/4180 first 😅

armanbilge:
I think we should add metrics for the `TimerHeap` as well (number of outstanding timers, next one due to fire, maybe last one due to fire?).

I'd also like to add polling system metrics. All polling systems should be able to support # of outstanding I/O operations.

iRevive:
May we do this in the follow-up PR? The current PR is already big, and I would prefer to operate smaller changesets.

_______

Getting stats (at least the number of timers) from the `TimerHeap` seems _relatively_ simple. However, `PollingSystem` could require some changes 
The naive idea is:
```scala
abstract class PollingSystem {
  def stats: PollingSystem.Stats
}
object PollingSystem {
  sealed trait Stats { 
    def outstandingOperations(): Long
  }
}
```

That way, we can enforce implementations to provide metrics.

I also found this one: https://github.com/typelevel/cats-effect/issues/3686.

djspiewak (approved):

PR #3322, opened 2022-12-23
Documented a series of recommended approaches for addressing starvation

djspiewak:
This is also a decent list of "tune your Cats Effect application for scale" recommendations, particularly for cluster replicated microservices. Hopefully it helps some folks!


docs/core/starvation-and-tuning.md
```diff
@@ -0,0 +1,501 @@
...
+import scala.concurrent.duration._
+
+val Threshold = 0.1d
+
+val tick: IO[Unit] =
+  // grab a time offset from when we started
+  IO.monotonic flatMap { start =>
+    // sleep for some long interval, then measure when we woke up
+    // get the difference (this will always be a bit more than 10.seconds)
+    IO.sleep(10.seconds) *> IO.monotonic.map(_ - start) flatMap { delta =>
```
armanbilge: Why 10 seconds, isn't the default one second?
https://github.com/typelevel/cats-effect/blob/52b5a3ba1623ff40fc48318c9a0bfae558f0e5ed/core/shared/src/main/scala/cats/effect/unsafe/IORuntimeConfig.scala#L167

docs/core/starvation-and-tuning.md
```diff
@@ -0,0 +1,501 @@
...
+
+Let's imagine for the sake of example that you were able to partially consolidate: 6 of the 8 active rogue threads are now consolidated with Cats Effect, but 2 remain. In this case, you have to try to minimize and make space for the remaining threads using configuration. If the library spawning these rogue threads is configurable, try to reduce the 2 to a 1 by passing some configuration parameters which reduce the size of its internal thread pool. This is not *always* possible, but it is often enough that it's worth looking for. Note that you won't be able to use this approach to go from two to zero, but you can at least reduce down to one.
+
+Now you have 8 compute worker threads and 1 active rogue thread. You need these two numbers to sum to the number of CPUs while keeping as many compute worker threads as possible, and unfortunately you have run out of options to reduce the active rogue thread count. In this case, you have to simply reduce the number of Cats Effect compute workers. This can be done with the following override in your `IOApp`:
+
+```scala
+override def computeWorkerThreadCount =
+  Math.max(2, super.computeWorkerThreadCount - 1)
```
armanbilge, line 397: I thought this max is enforced elsewhere, but it can't hurt.

docs/core/starvation-and-tuning.md
```diff
@@ -0,0 +1,501 @@
...
+
+  override def run(args: List[String]) = ???
+}
+```
+
+> If you have any expectations or SLA around latency, you should not be disabling the checker, but instead should adjust its sensitivity to match your expectations. Only disable the checker if you either 1) do not have *any* latency-related requirements, or 2) have good reason to believe that your environment is such that timers are unreliable.
+
+### Increasing/Decreasing Sensitivity
+
+By default, the checker will warn whenever the compute latency exceeds 100 milliseconds. This is calculated based on the `cpuStarvationCheckInterval` (default: `10.seconds`) multiplied by the `cpuStarvationCheckThreshold` (default: `0.1d`). In general, it is recommended that if you want to increase or decrease the sensitivity of the checker, you should do so by adjusting the interval (meaning that a more sensitive check will run more frequently):
```
armanbilge: Default `1.second`? Anyway that math doesn't seem right `10 s * 0.1 != 100 millis` 
https://github.com/typelevel/cats-effect/blob/52b5a3ba1623ff40fc48318c9a0bfae558f0e5ed/core/shared/src/main/scala/cats/effect/unsafe/IORuntimeConfig.scala#L167

armanbilge: > In general, it is recommended that if you want to increase or decrease the sensitivity of the checker, you should do so by adjusting the interval

This contradicts the FAQ.
https://github.com/typelevel/cats-effect/blob/26630c04e60afe7bba43bb1506ecfc4f773ad927/docs/faq.md?plain=1#L66

docs/core/starvation-and-tuning.md
```diff
@@ -0,0 +1,501 @@
...
+
+At this point, though, it may be helpful to bring in [Fs2](https://fs2.io), since these types of patterns are much safer and simpler with that library than base Cats Effect.
+
+### Not Enough CPUs
+
+One of the most insidious trends in modern server-deployed software is the strategy of squeezing application resources down to the bare minimum, wrapping those applications up in containers, and attempting to recover the lost capacity by replicating a large number of instances. While there are absolutely good reasons to containerize applications and design them to be deployed in a replicated topology, abstracted from the underlying host systems, it is very much possible to have too much of a good thing.
+
+In particular, it is very common to attempt to allocate only a single CPU (or less!), to an application. This in turn means that the application *itself* has very too few compute resources to schedule across, which creates bottlenecks. Effectively, this forces the underlying system (which is to say, the operating system kernel) to be the *only* form of scheduling, making systems like Cats Effect much less relevant.
```
armanbilge: > very too few

over-modified or intentional? 😁

armanbilge (commented):
This is so so awesome :) added a few minor notes.

docs/core/starvation-and-tuning.md
djspiewak: Oh. We should fix mine lol

docs/core/starvation-and-tuning.md
djspiewak: Is it? We should probably condense that. I kinda feel like the enforcement should be in `IOApp`, where it's overridable.

docs/core/starvation-and-tuning.md
djspiewak: Nice catch! I'll ix that. I think the FAQ is what I would prefer to adjust.

docs/core/starvation-and-tuning.md
djspiewak: Very exceptionally and excessively too few(er)

docs/core/starvation-and-tuning.md
armanbilge: Ah sorry I was wrong. We should condense that. Maybe?
https://github.com/typelevel/cats-effect/blob/52b5a3ba1623ff40fc48318c9a0bfae558f0e5ed/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala#L36-L38

docs/faq.md
```diff
@@ -48,3 +48,22 @@ Hello world
...
+Cats Effect 3.4.0 introduced a default-enabled *starvation checker*, which produces warnings like the following:
+
+```
+[WARNING] Your CPU is probably starving. Consider increasing the granularity
+of your delays or adding more cedes. This may also be a sign that you are
+unintentionally running blocking I/O operations (such as File or InetAddress)
+without the blocking combinator.
+```
+
+If you're seeing this warning and have not changed any of the default configurations, it means that your application is taking at least *100 milliseconds* to respond to an external asynchronous event. In this case, the runtime is measuring this fact using a timer, but external events are also things such as new connections, request bodies, upstream responses, and such. In other words, **if you're seeing this morning, it means your response latencies are *at least* this long.**
```
armanbilge: 🌄 

```suggestion
If you're seeing this warning and have not changed any of the default configurations, it means that your application is taking at least *100 milliseconds* to respond to an external asynchronous event. In this case, the runtime is measuring this fact using a timer, but external events are also things such as new connections, request bodies, upstream responses, and such. In other words, **if you're seeing this warning, it means your response latencies are *at least* this long.**
```

armanbilge (approved):
🚀

PR #3326, opened 2022-12-24
Backport for 3.4.3

djspiewak:


armanbilge (approved):

core/shared/src/main/scala/cats/effect/IODeferred.scala
```diff
@@ -0,0 +1,82 @@
...
+    } else {
+      false
+    }
+  }
+
+  def get: IO[A] = IO defer {
+    val back = cell.get()
+
+    if (back eq Sentinel) {
+      IO.cont[A, A](new Cont[IO, A, A] {
```
armanbilge, line 44: Oh, are we changing this in a follow-up?

djspiewak: Yep :-) I just blindly backport most of the time unless there are hard conflicts

PR #3330, opened 2022-12-25
Run Scalafix only on current platform in CI matrix

armanbilge:
Currently every platform is compiling every other platform, to run a redundant scalafix check. It's also being run on the entire JDK/OS matrix which is very unnecessary. This was causing some jobs to take more than an hour to run and thus get timed out.

FTR: if we didn't completely override the generated steps, sbt-typelevel would do this automatically. Not just for scalafix, but also for scalafmt, MiMa (edit: although MiMa on JDK 17 did catch that dotty bug 🤔 ), scaladoc, anything else that really only needs to be run on one JDK and one OS.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -228,17 +228,17 @@ jobs:
         run: sbt githubWorkflowCheck
 
       - name: Check that scalafix has been run on JVM
-        if: matrix.ci == 'ciJVM' && matrix.scala != '3.2.1' && matrix.os != 'windows-latest'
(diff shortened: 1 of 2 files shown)


djspiewak (approved):
It would definitely be nice if the surface area between sbt-typelevel and the CE build were a bit more compositional. It feels like we override almost everything. Some of that is just bikeshedding for sure, but it feels like our build has drifted over time into a very bespoke place.

armanbilge:
Yeah, it's annoying. Part of the problem is that the settings don't compose very well. But also the CE build does have some legitimately bespoke needs particularly for the CI matrix which just keeps growing. I have some new ideas how to organize it that we can try.

PR #14, opened 2017-04-22
Updated readme with snapshot release info and stability disclaimer

djspiewak:

PR #3332, opened 2022-12-26
Polling system

armanbilge:
Builds on top of
- https://github.com/typelevel/cats-effect/pull/3219
- https://github.com/typelevel/cats-effect/pull/3278
- https://github.com/typelevel/cats-effect/pull/3296
- https://github.com/typelevel/cats-effect/pull/3314

Downstream integrations:
- https://github.com/armanbilge/fs2-io_uring/pull/35
- https://github.com/typelevel/fs2/pull/3087
- https://github.com/typelevel/fs2/pull/3091


core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
```diff
@@ -0,0 +1,55 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+package unsafe
+
+import scala.concurrent.ExecutionContext
+
+abstract class PollingSystem {
```
armanbilge, line 20: As promised: this now lives in `jvm-native/` shared sources.

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
```diff
@@ -0,0 +1,55 @@
...
+   *   the maximum duration for which to block, where `nanos == -1` indicates to block
+   *   indefinitely. ''However'', if `nanos == -1` and there are no remaining events to poll
+   *   for, this method should return `false` immediately. This is unfortunate but necessary so
+   *   that the `EventLoop` can yield to the Scala Native global `ExecutionContext` which is
+   *   currently hard-coded into every test framework, including MUnit, specs2, and Weaver.
+   *
+   * @return
+   *   whether poll should be called again (i.e., there are more events to be polled)
+   */
+  def poll(data: PollData, nanos: Long, reportFailure: Throwable => Unit): Boolean
```
armanbilge: So on Native this method returns a `boolean` indicating if there are more pending events on the polling system. This is necessary for knowing when to terminate the event loop.

On JVM it's not currently necessary, but perhaps it could be useful for future optimizations. For example, maybe we can fallback to higher-resolution (and probably more efficient?) `parkNanos()` when there are no outstanding events.

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
```diff
@@ -0,0 +1,55 @@
...
+
+  def interrupt(targetThread: Thread, targetData: PollData): Unit
```
armanbilge: On Native this is currently a no-op.

core/jvm/src/main/scala/cats/effect/SelectorPoller.scala
```diff
@@ -0,0 +1,36 @@
...
+   */
+  def provider: SelectorProvider
+
+  /**
+   * Fiber-block until a [[java.nio.channels.SelectableChannel]] is ready on at least one of the
+   * designated operations. The returned value will indicate which operations are ready.
+   */
+  def select(ch: SelectableChannel, ops: Int): IO[Int]
+
+}
```
armanbilge: I feel cheerful about this interface as a consumer and I believe the `select` operation is rather primitive 😉

I do still think that these interfaces should be in terms of `IO` but that might be a whole other discussion :)

core/jvm/src/main/scala/cats/effect/unsafe/SelectorSystem.scala
```diff
@@ -0,0 +1,142 @@
...
+
+import scala.concurrent.ExecutionContext
+
+import java.nio.channels.SelectableChannel
+import java.nio.channels.spi.SelectorProvider
+import java.nio.channels.spi.AbstractSelector
+
+import SelectorSystem._
+
+final class SelectorSystem private (provider: SelectorProvider) extends PollingSystem {
```
armanbilge, line 27: Overall I think I'm happy with how this came out. Most notably, `Channel`s do not get stuck to a particular thread and may end up registered with many/all thread-local `Selector`s. This allows for dynamic load balancing as fibers get moved between threads.

I spent some time studying the epoll and kqueue `Selector` implementations in the JDK. Two notable constraints/implementation details:
- the JDK warns that any interaction with a `Selector` and `SelectionKeys` registered on it may be synchronized and blocking, so this only ever happens on the thread that owns the `Selector`
- unfortunately interactions with `Selector` and `SelectionKeys` such as setting the interest ops generally involve a syscall. This is even the case for the kqueue implementation, which could support batching. So every registration/callback pair involves a handful of syscalls.

core/jvm/src/main/scala/cats/effect/unsafe/SelectorSystem.scala
```diff
@@ -0,0 +1,142 @@
...
+          }
+        }
+      }.evalOn(ec)
+    }
+
+  }
+
+  final class PollData private[SelectorSystem] (
+      private[SelectorSystem] val selector: AbstractSelector
+  )
```
armanbilge: Another design issue we can punt for the moment, but start thinking about:

Technically we should not need this wrapper, we could just alias `type PollData = AbstractSelector`. However, because `PollData` is exposed as a public API, we'd be in bincompat trouble if we ever needed to change the representation.

I would _really_ like to find a way to make all `SelectorSystem` implementation methods/classes `protected` and/or `private`, so that we can evolve them without worrying about stuff like this.

armanbilge:
Published `3.5-01c4a03`.

armanbilge:
I've put up a draft PR trying this out in FS2.
- https://github.com/typelevel/fs2/pull/3091

This discovered a bug when attempting to access the polling system from a worker thread that has become a blocker thread. This makes me wonder: if you have `IO.blocking(...) *> IO(...).evalOn(WSTP)`, currently this doesn't shift the fiber back to a compute thread, but maybe it should?

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -75,6 +79,33 @@ private[effect] final class WorkStealingThreadPool(
...
+
+          if (thread.isInstanceOf[WorkerThread]) {
+            val worker = thread.asInstanceOf[WorkerThread]
+            if (worker.isOwnedBy(pool)) // we're good
+              ioa
+            else // possibly a blocking worker thread, possibly on another wstp
+              IO.cede *> ioa.evalOn(pool)
+          } else ioa.evalOn(pool)
+        }
+      }
```
armanbilge: This guarantees we'll always get where we need to be to access a thread-local polling data structure :)

armanbilge:
Published `3.5-6581dc4`.

core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
```diff
@@ -141,4 +144,11 @@ private[effect] abstract class IOCompanionPlatform { this: IO.type =>
    */
   def readLine: IO[String] =
     Console[IO].readLine
+
+  def poller[Poller](implicit ct: ClassTag[Poller]): IO[Option[Poller]] =
+    IO.executionContext.map {
+      case wstp: WorkStealingThreadPool if ct.runtimeClass.isInstance(wstp.poller) =>
+        Some(wstp.poller.asInstanceOf[Poller])
+      case _ => None
+    }
```
djspiewak: I don't think it's all that helpful to put this on `IO`. It also hard-codes a very specific selection criterion (typecasing) and doesn't allow for anything more robust. I'd rather invert this and have the polling system in control of determining whether an arbitrary `ExecutionContext` is compatible.

core/jvm/src/main/scala/cats/effect/SelectorPoller.scala
```diff
@@ -0,0 +1,36 @@
...
+   * The [[java.nio.channels.spi.SelectorProvider]] that should be used to create
+   * [[java.nio.channels.SelectableChannel]]s that are compatible with this polling system.
+   */
+  def provider: SelectorProvider
+
+  /**
+   * Fiber-block until a [[java.nio.channels.SelectableChannel]] is ready on at least one of the
+   * designated operations. The returned value will indicate which operations are ready.
+   */
+  def select(ch: SelectableChannel, ops: Int): IO[Int]
```
djspiewak: Why does this return in `IO`? Can that `IO` be async? Can it thread-shift?

core/jvm/src/main/scala/cats/effect/unsafe/SelectorSystem.scala
```diff
@@ -0,0 +1,140 @@
...
+  def makePoller(delayWithData: (PollData => *) ~> IO): Poller =
+    new Poller(delayWithData, provider)
+
+  def makePollData(): PollData = new PollData(provider.openSelector())
+
+  def closePollData(data: PollData): Unit =
+    data.selector.close()
+
+  def poll(data: PollData, nanos: Long, reportFailure: Throwable => Unit): Boolean = {
+    val millis = if (nanos >= 0) nanos / 1000000 else -1
```
djspiewak, line 42: This is actually really annoying, because it means that timers (any `IO.sleep`) would be clamped to millisecond granularity at most. Also this rounds down, so anything less than 1 ms would be considered 0 ms. That's a somewhat meaningful regression in granularity.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -79,6 +80,32 @@ private[effect] final class WorkStealingThreadPool(
...
+        IO.defer {
+          val thread = Thread.currentThread()
+          val pool = WorkStealingThreadPool.this
+
+          if (thread.isInstanceOf[WorkerThread]) {
+            val worker = thread.asInstanceOf[WorkerThread]
+            if (worker.isOwnedBy(pool)) // we're good
+              ioa
+            else // possibly a blocking worker thread, possibly on another wstp
+              IO.cede *> ioa.evalOn(pool)
```
djspiewak: UGGGGGGGGGGGGGGHHHHHHHHHHHHHHHHHHHHHHHHH ;-)

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -64,6 +65,8 @@ private final class WorkerThread(
   // Index assigned by the `WorkStealingThreadPool` for identification purposes.
   private[this] var _index: Int = idx
 
+  private[this] var _pollData: system.PollData = __pollData.asInstanceOf[system.PollData]
```
djspiewak: I love that we needed not just one but *two* prefix underscores.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -79,6 +80,32 @@ private[effect] final class WorkStealingThreadPool(
   private[unsafe] val parkedSignals: Array[AtomicBoolean] = new Array(threadCount)
   private[unsafe] val fiberBags: Array[WeakBag[Runnable]] = new Array(threadCount)
   private[unsafe] val sleepersQueues: Array[SleepersQueue] = new Array(threadCount)
+  private[unsafe] val pollDatas: Array[AnyRef] = new Array[AnyRef](threadCount)
+
+  private[effect] val poller: Any =
+    system.makePoller(new ((system.PollData => *) ~> IO) {
```
djspiewak: A *lot* of the shenanigans in this block are only necessitated because of the `IO`. If this were a lower-level side-effecting signature, we wouldn't need any of this.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -247,7 +280,7 @@ private[effect] final class WorkStealingThreadPool(
         // impossible.
         workerThreadPublisher.get()
         val worker = workerThreads(index)
-        LockSupport.unpark(worker)
+        system.interrupt(worker, pollDatas(index).asInstanceOf[system.PollData])
```
djspiewak: IMO this is where we start to see some of the limitations associated with using a function/state encoding, as opposed to actual virtual classes.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -598,6 +631,7 @@ private[effect] final class WorkStealingThreadPool(
       var i = 0
       while (i < threadCount) {
         workerThreads(i).interrupt()
+        system.closePollData(pollDatas(i).asInstanceOf[system.PollData])
```
djspiewak: LOTS of casting

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -111,6 +114,8 @@ private final class WorkerThread(
     setName(s"$prefix-$nameIndex")
   }
 
+  private[unsafe] def pollData(): Any = _pollData
```
djspiewak: I think we can do better about typing this if we make it a type parameter on WSTP.

core/native/src/main/scala/cats/effect/unsafe/EventLoopExecutorScheduler.scala
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+}
+
+private object EventLoopExecutorScheduler {
+  lazy val global = {
+    val system =
+      if (LinktimeInfo.isLinux)
+        EpollSystem
+      else if (LinktimeInfo.isMac)
```
djspiewak, line 162: Something to think about: if some third party implements a `UringSystem`, they won't be able to get it in here.

core/native/src/main/scala/cats/effect/unsafe/EventLoopExecutorScheduler.scala
```diff
@@ -0,0 +1,163 @@
...
+      sleepQueue.offer(sleepTask)
+      sleepTask
+    }
+
+  def reportFailure(t: Throwable): Unit = t.printStackTrace()
+
+  def nowMillis() = System.currentTimeMillis()
+
+  override def nowMicros(): Long =
+    if (LinktimeInfo.isFreeBSD || LinktimeInfo.isLinux || LinktimeInfo.isMac) {
```
djspiewak, line 74: There really should be a better way to detect "is POSIX-capable".

core/jvm/src/main/scala/cats/effect/SelectorPoller.scala
armanbilge: Yes, that `IO` _is_ async. This is where we register the read/write ready callback and wait to be notified.

core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
armanbilge: > I don't think it's all that helpful to put this on `IO`

This was your idea originally 😜 yeah, we'll revisit this :)

core/jvm/src/main/scala/cats/effect/unsafe/SelectorSystem.scala
armanbilge: Yeah I've been thinking about this as well. The underlying problem is epoll, kqueue and io_uring support finer granularity.

I wondered if in the case there are no outstanding I/O tasks if we can delegate to `Thread.sleep()` instead, but then interruption becomes more complicated since you have to know _how_ the thread is parked.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
armanbilge: I just pushed a replacement in terms of your beloved `register` function 😉 but IMHO it involves a similar amount of shenanigans :)

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
armanbilge: You mean the casting? Yeah, fair, but this encoding is also more flexible to the polling system i.e. we don't force it to implement specific classes. IMO it's like OOP vs typeclasses.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
armanbilge: yeah yeah 😛

core/native/src/main/scala/cats/effect/unsafe/EventLoopExecutorScheduler.scala
armanbilge: ~~Not sure what you mean? They can override this. That third-party is me, and I already did it 😉~~

https://github.com/armanbilge/fs2-io_uring/blob/770c91a033b55dc7ca4ee4c19853c9f3c3551fd8/uring/src/main/scala/fs2/io/uring/UringApp.scala#L24

Oh my bad, this is the global. Yeah, we're stuck with this.

core/native/src/main/scala/cats/effect/unsafe/EventLoopExecutorScheduler.scala
armanbilge: `!LinktimeInfo.isWindows` :trollface:

core/jvm/src/main/scala/cats/effect/IOCompanionPlatform.scala
djspiewak: It was my idea originally. :-) I kind of think it was a bad idea though.

core/jvm/src/main/scala/cats/effect/unsafe/SelectorSystem.scala
djspiewak: That also seems unlikely to be the common case in many such applications. :-(

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
djspiewak: Yeah this is composition vs subtyping. I think we can get the best of both by using type parameters rather than member types, but I'm still noodling.

wjoel:
There is some kind of problem. With `cats-effect 3.5-6581dc4` and `fs2-io 3.5-ac5c657`, ember's performance is much, much worse, 70% worse in some cases. The silver lining is that it's worse across the board, so hopefully it's something that is easy to find with some profiling.

ember 1.0.0-M39 https://www.techempower.com/benchmarks/#section=test&shareid=6af7baa9-195c-4a47-a33a-f8b3a9bd4f4a
https://github.com/wjoel/FrameworkBenchmarks/tree/2023-02-http4s-ember

ember JVM polling https://www.techempower.com/benchmarks/#section=test&shareid=28e63fce-6374-487c-bc01-f4240e2b7759
https://github.com/wjoel/FrameworkBenchmarks/tree/2023-02-http4s-ember-jvm-polling

To run the benchmarks, check out one of the FrameworkBenchmarks branches and `./tfb --mode benchmark --test http4s`. plaintext might be the most relevant one for performance optimization, so `./tfb --mode benchmark --type plaintext --test http4s` could be enough.

wjoel:
I've been digging this into some more, and simplified things a bit. It's sufficient to focus only the `plaintext` benchmark which, no matter what you might think about TechEmpower benchmarks in general, is quite reasonable (and doesn't involve anything like vert.x pg-client) for benchmarking this change to cats-effect. It hammers this endpoint:
```scala
  private val helloWorld = ByteVector("Hello, World!".getBytes(StandardCharsets.UTF_8))
  private val contentTypePlain = `Content-Type`(MediaType.text.plain, Some(Charset.`UTF-8`))

  // ...

      case GET -> Root / "plaintext" =>
        IO.pure(
          Response(
            entity = Entity.strict(helloWorld),
            headers = Headers(contentTypePlain)
          )
        )
```
We can run something similar to what TechEmpower does manually with `tfb.wrk`, over and over again, with 4096 concurrent connections (a fixed number, unlike what the full TechEmpower suite does). The following is what I get (Ryzen 5900X, 12 physical cores, 24 virtual threads).

It seems like there is a memory "leak", overzealous work stealing, something wrong with the different queues or their implementations, or something else entirely, but *something* is certainly not right and it's not because of the benchmark being flawed.

### jvm-polling
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 144331.43
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 147982.57
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 139646.84
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 127439.78
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 144567.55
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 117952.88
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 122469.12
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 118869.97
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 108606.01
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 119105.82
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 118123.92

That's pretty interesting already, it varies wildly and goes down a lot over time, recovers a bit, then gets even worse. For comparison, 3.4.6 is always close to 200k req/s:

### Baseline, CE 3.4.6
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 199480.48
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 200568.95
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 201935.35
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 201005.14
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 192537.04
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 188990.24
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 190384.52
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 200301.72
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 198351.62
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 199537.18
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 199509.45
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 198845.79

Even more interesting, the CPU usage stayed high on jvm-polling for a good while even *after* I stopped running tfb.wrk. The red circle is approximately when I stopped. The profiling results are from running JFR in IntelliJ.
jvm-polling

On CE 3.4.6 the CPU usage went to zero immediately after the last run of `tfb.wrk`.
no-jvm-polling

Looking at CPU profiling results in a bit more detail, jvm-polling spends a lot of time in `WorkStealingThreadPool.stealFromOtherWorkerThread`, `SleepersQueue.popHead()`, `PriorityQueue.dequeue()`, `cats.effect.unsafe.ScalQueue.poll`, `scala.collection.mutable.PriorityQueue.fixDown`, `java.util.concurrent.ConcurrentLinkedQueue.poll()`, `LocalQueue.stealInto`, which just don't show up anywhere near the top on CE 3.4.6 (or not at all).
jvm-polling-method-cpu-samples

For comparison, the most CPU heavy non-CE method is `org.http4s.ember.core.Parser$MessageP$.recurseFind` and it's #8 on CE 3.4.6, but #14 on jvm-polling.
ce-3 4 6-cpu-samples

In terms of memory allocation, the flame graphs look very similar, except that `cats.effect.unsafe.WorkerThread.parkUntilNextSleeper$1` is 13.19% of all allocations on jvm-polling, and doesn't show up on CE 3.4.6.
jvm-polling-allocations

You should probably open the JFR files yourself, so here they are.
[flight-recordings.tar.gz](https://github.com/typelevel/cats-effect/files/10700688/flight-recordings.tar.gz)

WebServer_2023_02_09_184302.jfr is CE 3.4.6
WebServer_2023_02_09_183429.jfr is jvm-polling

djspiewak:
@wjoel So thinking tactical vs strategic on this a bit… are you seeing similar issues with the timers change? I realize it's relatively time-consuming to run tests like this so no worries if you don't know yet.

wjoel:
@djspiewak Running this single benchmark doesn't take long, when I can find the time, so no worries. With only the timers change, it's much better, but still a bit worse (and with much higher variance) than CE 3.4.6.

After timers: 3.5-be6325d (and fs2 3.5.0)
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 189554.81
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 190422.15
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 187845.44
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 193121.19
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 190935.09
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 172647.72
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 181103.28
docker run --network=host techempower/tfb.wrk wrk -H 'Host: tfb-server' -H 'Accept: text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7' -H 'Connection: keep-alive' --latency -d 15 -c 4096 --timeout 8 -t 24 http://localhost:8080/plaintext -s pipeline.lua -- 16 | grep Requests/sec
Requests/sec: 184499.51

CPU usage looks fine, and dropped to zero immediately after I stopped running `tfb.wrk`, but heap memory usage is still closer to what jvm-polling got (ie. much higher):
after-timers-cpu-heap

But nothing stands out in the allocation flame graph, and in terms of CPU usage only `WorkStealingThreadPool.stealFromOtherWorkerThread` stands out. Well, maybe `LocalQueue.stealInto` as well, but the others (`ScalQueue.poll` etc.) seem to be under control, much better than in jvm-polling.
after-timers-cpu-method-list

JFR: [after-timers-flight-recording.tar.gz](https://github.com/typelevel/cats-effect/files/10701132/after-timers-flight-recording.tar.gz)

armanbilge:
I wonder if there is any possibility we are hitting the "infamous epoll 100% CPU bug" that plagues `Selector` 🤔 

https://github.com/netty/netty/blob/d010e63bf5bf744f2ab6d0fc4386611efe7954e6/transport/src/main/java/io/netty/channel/nio/NioEventLoop.java#L354-L369

armanbilge:
> Even more interesting, the CPU usage stayed high on jvm-polling for a good while even _after_ I stopped running tfb.wrk.

I have a new hypothesis about what is happening here.

As noted on [Discord](https://discord.com/channels/632277896739946517/839263556754472990/1080269655177965678), a short-coming of the current integrated timers implementation is that the worker thread park-loop wakes up even for canceled timers. Specifically, it wakes up at the time the timer was scheduled to run, only to discover that it has since been canceled.

In the happy-path, Ember server will generate _a lot_ of canceled timers. So even after the benchmark is completed, the runtime will spend the next minute waking up to process those canceled timers.

Furthermore, as noted above in https://github.com/typelevel/cats-effect/pull/3332#discussion_r1089820306, the smallest sleep granularity supported by `Selector` is 1 millisecond. For any sleeps smaller than that, the park degrades into a spin-loop.

I believe that's why CPU use stayed so high after the jvm-polling benchmark ended. It still had to work through all the canceled timers, which were spaced less than 1ms apart, forcing it to degrade into a spinloop.

wjoel:
Now that https://github.com/typelevel/cats-effect/pull/3499 has been merged, perhaps we could try rebasing and see if that helped resolve some of the performance issues related to timers in this PR?

armanbilge:
@wjoel I did the merge but expect the logic to still be broken, I think some of it was existing bugs. I'm going to follow-up with some fixes for that, I believe it will require changes to the `PollingSystem` API.

armanbilge:
@wjoel published `3.6-1f95fd7`. Thanks again for patiently benchmarking and re-benchmarking every step of the way, you've been amazing!

Unfortunately I'm still not entirely sure if it's all working correctly 🤔 I'm updating https://github.com/typelevel/fs2/pull/3091 as well and the TCP suite is passing, but not the TLS for some reason. Once I get it working I'll publish that as well.

Also, I see that the Scala Native ARM CI jobs have failed here. If anyone who has an ARM can replicate that locally, I'd be much obliged ☺️

wjoel:
Nice, with with this PR and https://github.com/typelevel/fs2/pull/3091 the results are quite close to https://github.com/typelevel/cats-effect/pull/3499, sometimes a few percentage points worse, sometimes a bit better, and no longer has any apparent regressions: https://www.techempower.com/benchmarks/#section=test&shareid=1d3ff9e7-d698-4156-9e2c-b122c22d108f

I haven't done any profiling with JFR, but the performance issues in https://github.com/typelevel/cats-effect/pull/3332#issuecomment-1416853455 are definitely fixed. :partying_face:

armanbilge:
> Also, I see that the Scala Native ARM CI jobs have failed here. If anyone who has an ARM can replicate that locally, I'd be much obliged ☺️

I figured out what this is. The `epoll_event` struct is packed differently on x86 and ARM, this is a bug that we inherited from epollcat. Haven't decided how to fix it yet ...

durban:
I think I'll continue here from https://github.com/typelevel/fs2/pull/3091#issuecomment-1526222761, because it seems like a Cats Effect issue.

So the important part seems to be this:
```
Exception in thread "io-compute-0" java.nio.channels.CancelledKeyException
	at java.base/sun.nio.ch.SelectionKeyImpl.ensureValid(SelectionKeyImpl.java:75)
	at java.base/sun.nio.ch.SelectionKeyImpl.interestOps(SelectionKeyImpl.java:98)
	at cats.effect.unsafe.SelectorSystem.poll(SelectorSystem.scala:77)
	at cats.effect.unsafe.SelectorSystem.poll(SelectorSystem.scala:25)
	at cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:513)
```

Locally I've got something slightly different:
```
Exception in thread "io-compute-2" java.nio.channels.CancelledKeyException
  | => iat java.base/sun.nio.ch.SelectionKeyImpl.ensureValid(SelectionKeyImpl.java:75)
        at java.base/sun.nio.ch.SelectionKeyImpl.interestOps(SelectionKeyImpl.java:98)
        at cats.effect.unsafe.SelectorSystem.poll(SelectorSystem.scala:77)
        at cats.effect.unsafe.SelectorSystem.poll(SelectorSystem.scala:25)
        at cats.effect.unsafe.WorkerThread.parkLoop$1(WorkerThread.scala:367)
        at cats.effect.unsafe.WorkerThread.park$1(WorkerThread.scala:328)
        at cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:652)
```

This exception probably kills a worker (`io-compute-2` in this case). The other workers in the WSTP are parked.

What's strange is that according to the WSTP's state, there is 1 searching thread. So it's possible there is still some bug in the state manipulation.

Also, I'm not sure what causes this exception, somehow the `SelectionKey` is invalid (cancelled?), but I don't really understand these polling systems...

armanbilge:
> Also, I'm not sure what causes this exception, somehow the `SelectionKey` is invalid (cancelled?), but I don't really understand these polling systems...

Oh, good catch. Probably the issue is that the socket channel has been closed but it was not removed from the selector 🤔 this might be annoying to fix 🤔 edit: nevermind, should be easy 😇 

> What's strange is that according to the WSTP's state, there is 1 searching thread. So it's possible there is still some bug in the state manipulation.

Right, that seems strange 😅

durban:
Regarding the 1 searching thread: that was a false alarm. I think that can totally happen: _before_ a thread is unparked by `notifyParked`, the number of searching (and active) threads is incremented. So the "searching" thread is the one killed by the exception. And that's why this results in a not working WSTP: the other threads think that there is already 1 searching thread, so won't try to search.

armanbilge:
> And that's why this results in a not working WSTP: the other threads think that there is already 1 searching thread, so won't try to search.

Right, I think this makes sense. IIRC at most only half the worker threads are allowed to be searching and in CI we have two threads.

I think I was able to reproduce the issue. Fixed with a try-catch since in general the channel can always be closed concurrently, so there is always a possibility of an exception ...

Thanks for all your help, I really appreciate it 😊

armanbilge:
Hmm, the Native job on Apple Silicon seems to have hung on some sort of spin-loop, since it was consuming a roughly constant, non-trivial amount of CPU.

https://cirrus-ci.com/task/5883281208508416

Update: lovely, studying the logs suggests that it's the `FileDescriptorPollerSpec` that hung 😅 

Update 2: and again! Very specifically Scala 3 on Apple Silicon (combinations with Scala 2 or macOS Intel seem fine).

Would appreciate any help reproducing this locally 😇 
```
sbt ++3.2.2 testsNative/testOnly *.FileDescriptorPollerSpec
```

armanbilge:
Published `3.6-0142603`.

TonioGela:
I can confirm that the test suite is hanging locally on an Apple Silicon (M2 Pro).

The test that hangs, in particular, is in `cats.effect.FileDescriptorPollerSpec`, and it is `FileDescriptorPoller should handle lots of simultaneous events`

Commenting that test out makes the 2 other work fine
```
[info] FileDescriptorPollerSpec
[info] FileDescriptorPoller should
[info]   + notify read-ready events
[info]   + hang if never ready
[info] Total for specification FileDescriptorPollerSpec
[info] Finished in 1 second, 253 ms
[info] 2 examples, 0 failure, 0 error
```

Feel free to ask for further local tests and/or to run reproducible examples for inspection ( I'm not offering my help with the PR content since this is one of those PRs that are WAY over my skills :D )

[EDIT] I can run all the tests for all three platforms if this helps, let me know.

armanbilge:
Thanks Tonio, narrowing it down to that test was extremely helpful!! 🙏 

> Update 2: and again! Very specifically Scala 3 on Apple Silicon (combinations with Scala 2 or macOS Intel seem fine).

So, I was wrong about this 😅 it turns out that we are missing the Scala 3 Native + macOS Intel combo in the CI matrix. Similar issues with this test suite also reproduce there, so I will have to add it 🙃

armanbilge:
<details>

I got a clue 🤔 it seems to be tripping on one of the switch statements in the fiber runloop. But still, why only on Scala 3?

```
scala.MatchError: -115 (of class java.lang.Byte)
  | => cat java.lang.StackTrace$.currentStackTrace$$anonfun$1(Unknown Source)
	at java.lang.StackTrace$$$Lambda$2.applyVoid(Unknown Source)
	at scala.runtime.function.JProcedure1.apply(Unknown Source)
	at scala.scalanative.unsafe.Zone$.apply(Unknown Source)
	at java.lang.StackTrace$.currentStackTrace(Unknown Source)
	at java.lang.Throwable.fillInStackTrace(Unknown Source)
	at cats.effect.IOFiber.run(Unknown Source)
	at cats.effect.unsafe.EventLoopExecutorScheduler.loop(Unknown Source)
	at cats.effect.unsafe.EventLoopExecutorScheduler.scheduleIfNeeded$$anonfun$1(Unknown Source)
	at cats.effect.unsafe.EventLoopExecutorScheduler$$Lambda$2.run(Unknown Source)
	at scala.scalanative.runtime.ExecutionContext$.loop(Unknown Source)
	at scala.scalanative.runtime.package$.loop(Unknown Source)
	at org.specs2.concurrent.package$.awaitResult(Unknown Source)
```

Update: another stacktrace.

```
[\u001b[0;31merror\u001b[0;0m] Fatal signal Segmentation fault: 11 caught
[\u001b[0;31merror\u001b[0;0m]   at scala.scalanative.testinterface.signalhandling.SignalConfig$.asyncSafePrintStackTrace(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at scala.scalanative.testinterface.signalhandling.SignalConfig$.$anonfun$1(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at <none>._SM70scala.scalanative.testinterface.signalhandling.SignalConfig$$$Lambda$1G17$extern$forwarder(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at <none>._sigtramp(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at java.util.HashMap.growTable(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at java.util.HashMap.put0(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at java.util.HashMap.put0(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at java.util.HashMap.put(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.unsafe.KqueueSystem$Poller.evSet(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.unsafe.KqueueSystem$PollHandle.pollReadRec$$anonfun$1$$anonfun$1$$anonfun$1$$anonfun$1$$anonfun$1(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.unsafe.KqueueSystem$PollHandle$$Lambda$4.applyVoid(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at scala.runtime.function.JProcedure1.apply(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.unsafe.IORuntimeCompanionPlatform.createEventLoop$$anonfun$1(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.unsafe.IORuntimeCompanionPlatform$$Lambda$1.applyVoid(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at scala.runtime.function.JProcedure1.apply(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.unsafe.KqueueSystem$PollHandle.pollReadRec$$anonfun$1$$anonfun$1$$anonfun$1$$anonfun$1(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.unsafe.KqueueSystem$PollHandle$$Lambda$5.applyVoid(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at scala.runtime.function.JProcedure1.apply(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.IO$$anon$5.apply$$anonfun$3$$anonfun$1$$anonfun$1(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.IO$$anon$5$$Lambda$2.apply(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.IOFiber.runLoop(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.IOFiber.execR(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.IOFiber.run(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.unsafe.EventLoopExecutorScheduler.loop(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.unsafe.EventLoopExecutorScheduler.scheduleIfNeeded$$anonfun$1(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.unsafe.EventLoopExecutorScheduler$$Lambda$2.run(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at scala.scalanative.runtime.ExecutionContext$.loop(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at scala.scalanative.runtime.package$.loop(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at org.specs2.concurrent.package$.awaitResult(Unknown Source)
```

Another:
```
[\u001b[0;31merror\u001b[0;0m] Fatal signal Segmentation fault: 11 caught
[\u001b[0;31merror\u001b[0;0m]   at scala.scalanative.testinterface.signalhandling.SignalConfig$.asyncSafePrintStackTrace(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at scala.scalanative.testinterface.signalhandling.SignalConfig$.$anonfun$1(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at <none>._SM70scala.scalanative.testinterface.signalhandling.SignalConfig$$$Lambda$1G17$extern$forwarder(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at <none>._sigtramp(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.IOFiber.succeeded(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.IOFiber.runLoop(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.IOFiber.asyncContinueSuccessfulR(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.IOFiber.run(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.unsafe.EventLoopExecutorScheduler.loop(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.unsafe.EventLoopExecutorScheduler.scheduleIfNeeded$$anonfun$1(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.unsafe.EventLoopExecutorScheduler$$Lambda$2.run(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at scala.scalanative.runtime.ExecutionContext$.loop(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at scala.scalanative.runtime.package$.loop(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at org.specs2.concurrent.package$.awaitResult(Unknown Source)
```

and another 🤔 
```
[\u001b[0;31merror\u001b[0;0m]   at scala.scalanative.testinterface.signalhandling.SignalConfig$.asyncSafePrintStackTrace(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at scala.scalanative.testinterface.signalhandling.SignalConfig$.$anonfun$1(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at <none>._SM70scala.scalanative.testinterface.signalhandling.SignalConfig$$$Lambda$1G17$extern$forwarder(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at <none>._sigtramp(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.unsafe.KqueueSystem$Poller.processEvents$1(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.unsafe.KqueueSystem$Poller.poll(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.unsafe.KqueueSystem$.poll(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.unsafe.KqueueSystem$.poll(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.unsafe.EventLoopExecutorScheduler.loop(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.unsafe.EventLoopExecutorScheduler.scheduleIfNeeded$$anonfun$1(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.unsafe.EventLoopExecutorScheduler$$Lambda$2.run(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at scala.scalanative.runtime.ExecutionContext$.loop(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at scala.scalanative.runtime.package$.loop(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at org.specs2.concurrent.package$.awaitResult(Unknown Source)
```

...

```
[\u001b[0;31merror\u001b[0;0m] Fatal signal Segmentation fault: 11 caught
[\u001b[0;31merror\u001b[0;0m]   at scala.scalanative.testinterface.signalhandling.SignalConfig$.asyncSafePrintStackTrace(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at scala.scalanative.testinterface.signalhandling.SignalConfig$.$anonfun$1(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at <none>._SM70scala.scalanative.testinterface.signalhandling.SignalConfig$$$Lambda$1G17$extern$forwarder(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at <none>._sigtramp(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.IOFiber.succeeded(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.IOFiber.runLoop(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.IOFiber.asyncContinueSuccessfulR(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.IOFiber.run(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.unsafe.EventLoopExecutorScheduler.loop(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.unsafe.EventLoopExecutorScheduler.scheduleIfNeeded$$anonfun$1(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at cats.effect.unsafe.EventLoopExecutorScheduler$$Lambda$2.run(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at scala.scalanative.runtime.ExecutionContext$.loop(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at scala.scalanative.runtime.package$.loop(Unknown Source)
[\u001b[0;31merror\u001b[0;0m]   at org.specs2.concurrent.package$.awaitResult(Unknown Source)
```

</details>

Update: 9ba870f3059de3b83696cb8e9f0b60ac715fc44f ~~appears to be working~~ gahh, it's broken all along. 3d1cd9698b174caa5971784d7d5d7998dce41738 is broken according to CI. ~~So something in-between 🤔~~

TonioGela:
Spoiler alert: the last @armanbilge commit fixed the issue :D
image

core/js-native/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -23,7 +23,7 @@ import scala.concurrent.duration.FiniteDuration
 // Can you imagine a thread pool on JS? Have fun trying to extend or instantiate
 // this class. Unfortunately, due to the explicit branching, this type leaks
 // into the shared source code of IOFiber.scala.
-private[effect] sealed abstract class WorkStealingThreadPool private ()
+private[effect] sealed abstract class WorkStealingThreadPool[Poller] private ()
```
djspiewak: I'm not a huge fan of using full-word names for type parameters, just because it then gets somewhat annoying to track down where the type is coming from. Also isn't `Poller` an actual type elsewhere that we would be shadowing?

core/js-native/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -38,12 +38,12 @@ private[effect] sealed abstract class WorkStealingThreadPool private ()
   private[effect] def canExecuteBlockingCode(): Boolean
   private[unsafe] def liveTraces(): (
       Map[Runnable, Trace],
-      Map[WorkerThread, (Thread.State, Option[(Runnable, Trace)], Map[Runnable, Trace])],
+      Map[WorkerThread[_], (Thread.State, Option[(Runnable, Trace)], Map[Runnable, Trace])],
```
djspiewak: Shouldn't this be `WorkerThread[Poller]`?

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
```diff
@@ -0,0 +1,62 @@
...
+   */
+  def poll(poller: Poller, nanos: Long, reportFailure: Throwable => Unit): Boolean
+
+  /**
+   * @return
+   *   whether poll should be called again (i.e., there are more events to be polled)
+   */
+  def needsPoll(poller: Poller): Boolean
+
+  def interrupt(targetThread: Thread, targetPoller: Poller): Unit
```
djspiewak, line 56: I'd really like to pull this out into a `Poller` trait of some sort, similar to how I encoded it on my branch, rather than putting it all into a single trait like this.

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
```diff
@@ -0,0 +1,62 @@
...
+
+}
+
+private object PollingSystem {
+  type WithPoller[Poller0] = PollingSystem {
+    type Poller = Poller0
+  }
```
djspiewak: I feel like putting the `Poller` type into the parameters of `PollingSystem` rather than as a type member would help reduce the boilerplate quite a bit.

core/jvm/src/main/scala/cats/effect/unsafe/LocalQueue.scala
```diff
@@ -337,7 +337,7 @@ private final class LocalQueue extends LocalQueuePadding {
    * @return
    *   a fiber to be executed directly
    */
-  def enqueueBatch(batch: Array[Runnable], worker: WorkerThread): Runnable = {
+  def enqueueBatch(batch: Array[Runnable], worker: WorkerThread[_]): Runnable = {
```
djspiewak, line 340: Wouldn't it be easier to add the type parameter to `LocalQueue`?

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
armanbilge: If we did that, then wouldn't it leak into `IOApp`? i.e. `IOApp` would become parameterized 🤔

core/jvm/src/main/scala/cats/effect/unsafe/LocalQueue.scala
armanbilge: Would it? 😆 what's wrong with `_`, that seems easy enough. the type parameter isn't important here

core/js-native/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
armanbilge: Sure, I guess we can use `P` :)

> Also isn't `Poller` an actual type elsewhere that we would be shadowing?

Not at the moment. We have stuff like `Selector`, `FileDescriptorPoller`, and `Uring`.

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
armanbilge: but i swore off oop 😭 fine I guess, that's how I ended up implementing them anyway ... but I do prefer this encoding, for whatever reason

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
djspiewak: Oh interesting point! Hmmmmmmm

core/jvm/src/main/scala/cats/effect/unsafe/LocalQueue.scala
djspiewak: I guess it's fine. We never actually pull out anything which contains a poller so it's really just boilerplate either way.

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
armanbilge: hmm, I tried this change and I'm having trouble making it work with the `WithPoller[P]` aux type, which I think is necessary (see https://github.com/typelevel/cats-effect/pull/3332#discussion_r1222285506). Let me have a look at your branch

armanbilge: Hmmm 🤔 
https://github.com/typelevel/cats-effect/blob/32d881f622bae633c74c72431401d0086ad36746/core/jvm/src/main/scala/cats/effect/IOApp.scala#L168

armanbilge:
Daniel and I discussed on Discord earlier today. Our working plan is to merge this PR into `series/3.x` more-or-less as-is, very soon. Then we will continue revising as follow-up PRs, particularly the encoding. The hope is that this strategy will make it easier to collaborate :)

armanbilge:
I published `3.6-e9aeb8c`. I'll work on updating my FS2 PR next and then publish a snapshot of that as well. Together that should be sufficient for testing Ember.

armanbilge:
I published FS2 `3.8-1580d81`. Anyone should be able to try this today with http4s Ember, Skunk, etc.

PR #3333, opened 2022-12-28
Avoid leaks in IODeferred

durban:


djspiewak (approved):
Oh! Nice catch. This is a relatively serious issue. Release coming.

PR #3342, opened 2023-01-02
Move the mbean initialization out of blocking and into delay

### LINKED ISSUE #3316: effects specified with IO.apply are unexpectedly running on the blocker pool
I'm running printing the current thread name via `println` delayed via `IO.apply`--*not* `IO.println`!--and I unexpectedly see the printing happening on the blocker threads:
```scala
object Blocking extends IOApp {
    for {
      // _ <- IO.cede
      _ <- IO(println(s"${Thread.currentThread.getName} default"))
      _ <- IO.blocking(println(s"${Thread.currentThread.getName} blocking"))
      _ <- IO(println(s"${Thread.currentThread.getName} default"))
    } yield ExitCode.Success
}
```
prints
```
io-compute-blocker-3 default
io-compute-blocker-3 blocking
io-compute-blocker-3 default
```
(the blocker thread number various across runs; this is cats-effect 3.4.2)

I expect the printing of `default` on the `io-compute` pool, not the `io-compute-blocker` pool.

If I uncomment the `cede`, I get the following output:
```
io-compute-1 default
io-compute-blocker-1 blocking
io-compute-blocker-1 default
```
which correctly prints the first `default`, but again the second one is unexpectedly on the blocker pool.

This reminds me of the "callback running on the external thread pool" problem of CE2, requiring an `IO.shift`, but it may only be superficially related.

mox692:
Fix #3316.

This PR moves the mbean initialization out of `IO.blocking` and into `IO.delay`, to allow user-defined program can use the compute pool from the beginning.

This is my first PR so give me a hint if I missed something :)

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/metrics/JvmCpuStarvationMetrics.scala
+++ b/core/jvm/src/main/scala/cats/effect/metrics/JvmCpuStarvationMetrics.scala
@@ -56,7 +56,6 @@ private[effect] object JvmCpuStarvationMetrics {
     val acquire: IO[(MBeanServer, JvmCpuStarvationMetrics)] = for {
       mBeanServer <- IO.delay(ManagementFactory.getPlatformMBeanServer)
       mBean <- CpuStarvation()
-      _ <- IO.blocking(mBeanServer.registerMBean(mBean, mBeanObjectName))
       // To allow user-defined program to use the compute pool from the beginning,
       // here we use `IO.delay` rather than `IO.blocking`.
       _ <- IO.delay(mBeanServer.registerMBean(mBean, mBeanObjectName))
--- a/core/jvm/src/main/scala/cats/effect/metrics/JvmCpuStarvationMetrics.scala
+++ b/core/jvm/src/main/scala/cats/effect/metrics/JvmCpuStarvationMetrics.scala
@@ -57,6 +57,8 @@ private[effect] object JvmCpuStarvationMetrics {
       mBeanServer <- IO.delay(ManagementFactory.getPlatformMBeanServer)
       mBean <- CpuStarvation()
       _ <- IO.blocking(mBeanServer.registerMBean(mBean, mBeanObjectName))
+      // To allow user-defined program to use the compute pool from the beginning,
+      // here we use `IO.delay` rather than `IO.blocking`.
       _ <- IO.delay(mBeanServer.registerMBean(mBean, mBeanObjectName))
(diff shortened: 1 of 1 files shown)


djspiewak:
Perfect! Two requests though. First, can you retarget this PR at `series/3.4.x`? This is a source- and binary forward-compatible change, so targeting at the minor branch allows us to get it out as soon as possible. Second, would you mind adding a quick comment explaining why `delay`? Otherwise I guarantee in a few years all of us will have forgotten and someone will PR a "fix" to change it back to `blocking`. :-)

armanbilge:
I am actually not sure why we marked it a `blocking` method to begin with. The JavaDoc doesn't say anything about blocking.
https://docs.oracle.com/javase/8/docs/api/javax/management/MBeanServer.html#registerMBean-java.lang.Object-javax.management.ObjectName-

~~In fact, it almost sounds like we should implement this using `async_`.~~ Edit: no, this notification is in addition to the method returning successfully.
> If this method successfully registers an MBean, a notification is sent as described [above](https://docs.oracle.com/javase/8/docs/api/javax/management/MBeanServer.html#notif).

core/jvm/src/main/scala/cats/effect/metrics/JvmCpuStarvationMetrics.scala
```diff
@@ -57,6 +57,9 @@ private[effect] object JvmCpuStarvationMetrics {
       mBeanServer <- IO.delay(ManagementFactory.getPlatformMBeanServer)
       mBean <- CpuStarvation()
       _ <- IO.blocking(mBeanServer.registerMBean(mBean, mBeanObjectName))
```
alexandrustana: Shouldn't this be removed, in order to avoid registering the bean twice? 🤔

mox692: Thanks, that was my mistake!
Fix: [e4b03ee](https://github.com/typelevel/cats-effect/pull/3342/commits/e4b03ee8bb5a8f8fd85f91c3d8f88bfcbac92f99)

mox692:
Test is failing now, so I'll see that.

armanbilge:
@mox692 thanks, that's not your fault. It looks like the test I added in https://github.com/typelevel/cats-effect/pull/3315 is flaking 😕

mox692:
Oh, I see. Thank you for your quick reply!

djspiewak (approved):

djspiewak:
Thank you for the fix!!

PR #3346, opened 2023-01-05
Optimize `Mutex` & `AtomicCell`

BalmungSan:
# Optimize `Mutex` & `AtomicCell`

This PR adds two alternative implementations of `Mutex` & `AtomicCell` based on `Async` rather than just `Concurrent`, these implementations should be more efficient.
This is a binary and source-compatible change. Since users don't need to do anything special to get those new implementations, rather if the underlying data type they use support `Async`; e.g. like `IO`, they will get the optimized versions automatically.

-----

### Configuration

+ forks: 2
+ threads: 1
+ warm-up iterations: 10
+ iterations: 10

### `Mutex`

**Concurrent**

| Benchmark | (fibers) | (iterations) | Mode | Cnt | Score | Error | Units |
| - | - | - | - | - | - | - | - |
| MutexBenchmark.happyPathConcurrent      |  10 | 1000 | thrpt | 20 | 192.475 | ± 2.296 | ops/s |
| MutexBenchmark.happyPathConcurrent      |  50 | 1000 | thrpt | 20 |  41.280 | ± 0.486 | ops/s |
| MutexBenchmark.happyPathConcurrent      | 100 | 1000 | thrpt | 20 |  19.415 | ± 0.289 | ops/s |
| - | - | - | - | - | - | - | - |
| MutexBenchmark.highContentionConcurrent |  10 | 1000 | thrpt | 20 |  26.426 | ± 0.330 | ops/s |
| MutexBenchmark.highContentionConcurrent |  50 | 1000 | thrpt | 20 |   6.963 | ± 0.204 | ops/s |
| MutexBenchmark.highContentionConcurrent | 100 | 1000 | thrpt | 20 |   3.750 | ± 0.055 | ops/s |
| - | - | - | - | - | - | - | - |
| MutexBenchmark.cancellationConcurrent   |  10 | 1000 | thrpt | 20 |  20.768 | ± 1.628 | ops/s |
| MutexBenchmark.cancellationConcurrent   |  50 | 1000 | thrpt | 20 |   6.835 | ± 0.065 | ops/s |
| MutexBenchmark.cancellationConcurrent   | 100 | 1000 | thrpt | 20 |   3.653 | ± 0.128 | ops/s |

**Async**

| Benchmark | (fibers) | (iterations) | Mode | Cnt | Score | Error | Units |
| - | - | - | - | - | - | - | - |
| MutexBenchmark.happyPathAsync      |  10 | 1000 | thrpt | 20 | 212.839 | ± 1.069 | ops/s |
| MutexBenchmark.happyPathAsync      |  50 | 1000 | thrpt | 20 |  46.925 | ± 1.366 | ops/s |
| MutexBenchmark.happyPathAsync      | 100 | 1000 | thrpt | 20 |  23.266 | ± 0.127 | ops/s |
| - | - | - | - | - | - | - | - |
| MutexBenchmark.highContentionAsync |  10 | 1000 | thrpt | 20 |  33.913 | ± 2.316 | ops/s |
| MutexBenchmark.highContentionAsync |  50 | 1000 | thrpt | 20 |   8.227 | ± 0.082 | ops/s |
| MutexBenchmark.highContentionAsync | 100 | 1000 | thrpt | 20 |   4.577 | ± 0.160 | ops/s |
| - | - | - | - | - | - | - | - |
| MutexBenchmark.cancellationAsync   |  10 | 1000 | thrpt | 20 |  24.807 | ± 0.694 | ops/s |
| MutexBenchmark.cancellationAsync   |  50 | 1000 | thrpt | 20 |   8.832 | ± 0.086 | ops/s |
| MutexBenchmark.cancellationAsync   | 100 | 1000 | thrpt | 20 |   5.034 | ± 0.027 | ops/s |

### `AtomicCell` _(using the respective `Mutex`)_

**Concurrent**

| Benchmark | (fibers) | (iterations) | Mode | Cnt | Score | Error | Units |
| - | - | - | - | - | - | - | - |
| AtomicCellBenchmark.happyPathConcurrent      |  10 | 1000 | thrpt | 20 | 164.129 | ± 0.711 | ops/s |
| AtomicCellBenchmark.happyPathConcurrent      |  50 | 1000 | thrpt | 20 |  34.864 | ± 0.628 | ops/s |
| AtomicCellBenchmark.happyPathConcurrent      | 100 | 1000 | thrpt | 20 |  17.276 | ± 0.024 | ops/s |
| - | - | - | - | - | - | - | - |
| AtomicCellBenchmark.highContentionConcurrent |  10 | 1000 | thrpt | 20 |  27.247 | ± 1.409 | ops/s |
| AtomicCellBenchmark.highContentionConcurrent |  50 | 1000 | thrpt | 20 |   6.315 | ± 0.073 | ops/s |
| AtomicCellBenchmark.highContentionConcurrent | 100 | 1000 | thrpt | 20 |   2.893 | ± 0.034 | ops/s |
| - | - | - | - | - | - | - | - |
| AtomicCellBenchmark.cancellationConcurrent   |  10 | 1000 | thrpt | 20 |  20.799 | ± 0.854 | ops/s |
| AtomicCellBenchmark.cancellationConcurrent   |  50 | 1000 | thrpt | 20 |   7.363 | ± 0.195 | ops/s |
| AtomicCellBenchmark.cancellationConcurrent   | 100 | 1000 | thrpt | 20 |   3.723 | ± 0.025 | ops/s |

**Async**

| Benchmark | (fibers) | (iterations) | Mode | Cnt | Score | Error | Units |
| - | - | - | - | - | - | - | - |
| AtomicCellBenchmark.happyPathAsync      |  10 | 1000 | thrpt | 20 | 192.392 | ± 0.191 | ops/s |
| AtomicCellBenchmark.happyPathAsync      |  50 | 1000 | thrpt | 20 |  40.638 | ± 0.139 | ops/s |
| AtomicCellBenchmark.happyPathAsync      | 100 | 1000 | thrpt | 20 |  18.958 | ± 0.139 | ops/s |
| - | - | - | - | - | - | - | - |
| AtomicCellBenchmark.highContentionAsync |  10 | 1000 | thrpt | 20 |  29.978 | ± 0.128 | ops/s |
| AtomicCellBenchmark.highContentionAsync |  50 | 1000 | thrpt | 20 |   8.026 | ± 0.366 | ops/s |
| AtomicCellBenchmark.highContentionAsync | 100 | 1000 | thrpt | 20 |   4.632 | ± 0.045 | ops/s |
| - | - | - | - | - | - | - | - |
| AtomicCellBenchmark.cancellationAsync   |  10 | 1000 | thrpt | 20 |  23.468 | ± 0.547 | ops/s |
| AtomicCellBenchmark.cancellationAsync   |  50 | 1000 | thrpt | 20 |   8.633 | ± 0.073 | ops/s |
| AtomicCellBenchmark.cancellationAsync   | 100 | 1000 | thrpt | 20 |   5.680 | ± 0.101 | ops/s |

-----

## Conclusions

IMHO the improvements are noticeable, being around a `20%` increase in most situations.
Also, `AtomicCell` results are pretty equivalent; but still slightly better, than the `Mutex` one. Suggesting that most of the increase comes from the `Mutex`, still I think the `AsyncAtomicCell` is worth it.

-----

[AtomicCellBenchmarkResults.txt](https://github.com/typelevel/cats-effect/files/10676136/AtomicCellBenchmarkResults.txt)
[MutexBenchmarkResults.txt](https://github.com/typelevel/cats-effect/files/10676137/MutexBenchmarkResults.txt)


std/shared/src/main/scala/cats/effect/std/Mutex.scala
```diff
@@ -64,22 +64,53 @@ object Mutex {
...
 
   // TODO: In case in a future cats-effect provides a way to identify fibers,
   // then this implementation can be made reentrant.
   // Or, we may also provide an alternative implementation using LiftIO + IOLocal
-  private final class Impl[F[_]](sem: Semaphore[F]) extends Mutex[F] {
+  private final class Impl[F[_]](state: Ref[F, LockChain])(implicit F: GenConcurrent[F, _])
+      extends Mutex[F] {
     override final val lock: Resource[F, Unit] =
-      sem.permit
+      Resource.apply[F, Unit] {
```
BalmungSan, line 81: I believe I should use `Resource.applyFull` instead here.

std/shared/src/main/scala/cats/effect/std/Mutex.scala
```diff
@@ -64,22 +64,53 @@ object Mutex {
...
+          def loop(chain: LockChain): F[(Unit, Resource.ExitCase => F[Unit])] = chain match {
+            case Cons(otherLock) =>
+              otherLock.asInstanceOf[Deferred[F, LockChain]].get.flatMap(loop)
+
+            case Empty =>
+              F.pure(() -> (_ => lock.complete(Empty).void))
+          }
+
+          state.getAndSet(Cons[F](lock)).flatMap { chain =>
+            poll(F.onCancel(loop(chain), lock.complete(chain).void))
```
BalmungSan: Is this the right way to use `poll`?

BalmungSan:
How do I fix the MIMA problem? Do I keep the old implementation private and deprecated?
Is it really possible to call that at runtime? Since only `apply` calls it I guess the error is a false positive... but I don't understand much of binary compatibility.

armanbilge:
Yes, it's a false positive since it's a private class/constructor. It should be safe to filter.

std/shared/src/main/scala/cats/effect/std/Mutex.scala
```diff
@@ -64,22 +64,53 @@ object Mutex {
...
 
   // TODO: In case in a future cats-effect provides a way to identify fibers,
   // then this implementation can be made reentrant.
   // Or, we may also provide an alternative implementation using LiftIO + IOLocal
-  private final class Impl[F[_]](sem: Semaphore[F]) extends Mutex[F] {
+  private final class Impl[F[_]](state: Ref[F, LockChain])(implicit F: GenConcurrent[F, _])
+      extends Mutex[F] {
     override final val lock: Resource[F, Unit] =
-      sem.permit
+      Resource.applyFull[F, Unit] { poll =>
```
armanbilge: I'm not sure when to choose between `makeFull` and `applyFull`, but personally I find `make`-variants better for readability (it's easy to distinguish the acquire/release phases).

BalmungSan: @armanbilge in general I would totally agree. However, in this case I feel it was easier to use `apply`
With `make` I would need to make `acquire` return a `Deferred[LockChain]` and then the release would be `_.complete(Empty).void` and then I would need to `void` the whole thing.

However, I have a feeling I am using `poll` wrong and maybe the split may help?

armanbilge:
> I said _"improve"_ but I am not totally sure this version is better than just delegating to `Semaphore`

Here's an idea, that I think could be material improvement. What if we extracted the `UnsafeUnbounded` datastructure introduced for the async `Queue` into a common place, so that we can use it here?

https://github.com/typelevel/cats-effect/blob/130fdcc7af75ed769f001d9f0afbff6ccf8b9488/std/shared/src/main/scala/cats/effect/std/Queue.scala#L1042

We can use the `Semaphore`-based implementation for `Concurrent` constraint, but use a runtime check for `Async` to upgrade to the `UnsafeUnbounded` implementation.

armanbilge:
> We can use the `Semaphore`-based implementation for `Concurrent` constraint

Or the implementation proposed in this PR, which is good too :)

---

Btw: I believe this PR can be targeted at 3.4.x, since it's just internal optimizations (no new APIs)

std/shared/src/main/scala/cats/effect/std/Mutex.scala
```diff
@@ -63,23 +65,65 @@ object Mutex {
...
+
+            poll(wait.map(_ => lock))
+          }
+        }
+      } { lock =>
+        F.delay {
+          val cb = lock.get()
+          if (cb ne null) cb(RUnit) else state.set(null)
+        }
+      }.void
```
BalmungSan, line 105: So, the idea of this new implementation is to save the `Deferred` instantiation, based on the assumption that at any point in time a `Fiber` is waiting for only one other `Fiber`, and it is being waited by at most another one.

So, the idea is that we will have:
+ `state` which is a mutable reference to the last cell in the `Chain`.
+ `lock` which is the `Chain` cell of the `Fiber` trying to access the `Mutex`.
+ `chain` which is the previous cell in the `Chain`.

The, we use `F.async` to wait for the callback to be called and we register the callback in the `chain` box.
When the `Mutex` is released, we call the callback that was registered in our cell and call it; of course, considering the possibility that maybe no one is actually waiting for us.
We also have to consider the possibility that the `Mutex` is already free when acquiring it; which is what happens when the previous cell was `null`.

In theory, this only allocates the bare minimum.
However, I am not totally sure if it is safe.

std/shared/src/main/scala/cats/effect/std/Mutex.scala
```diff
@@ -63,23 +65,77 @@ object Mutex {
...
+            // If the state contained any other cell different than ourselves,
+            // then we can be sure that our cell is not empty
+            // and we must awake whatever Fiber is waiting for us.
+            // Similarly, if it did contained our own cell,
+            // then it means nobody was waiting for the Mutex,
+            // and thus it can be put on a free state again.
+            thisCell.get().apply(RUnit)
+          }
+        }
+      }.void
```
BalmungSan, line 117: So, the idea of this new implementation is to save the `Deferred` instantiation, based on the assumption that at any point in time a `Fiber` is waiting for only one other `Fiber`, and it is being waited by at most another one.

So, the idea is that we will have:
+ `state` which is a mutable reference to the last `Cell` in the chain.
+ `thisCell` which is the chain `Cell` of the `Fiber` trying to access the `Mutex`.
+ `previousCell` which is the previous `Cell` in the chain.

Then, we use `F.async` to wait for the callback to be called and we register the callback in the `previousCell`.
When the `Mutex` is released, we call the callback that was registered in our `Cell` and call it. Considering the possibility that maybe no one is actually waiting for us.
We also have to consider the possibility that the `Mutex` is already free when acquiring it; which is what happens when the `previousCell` was `null`.
And, in case we where canceled while waiting for the `Mutex`, then we need to unregister our callback from the `previousCell` and register whatever was registered in our `Cell`; which may be empty.

In theory, this only allocates the bare minimum. However, I am not totally sure if it is safe. Especially around cancelation.

```scala
previousCell.set(thisCell.get()))
```
Can we be sure that this will happen before `previousCell` checks its contents?

std/shared/src/main/scala/cats/effect/std/Mutex.scala
```diff
@@ -63,23 +65,119 @@ object Mutex {
...
+        // then it means nobody was waiting for the Mutex,
+        // and thus it can be put on a free state again.
+        if (!state.compareAndSet(thisCell, null)) {
+          // Otherwise,
+          // our cell is probably not empty,
+          // we must awake whatever Fiber is waiting for us.
+          val nextCB = thisCell.getAndSet(Sentinel)
+          if (nextCB ne null) nextCB.apply(RUnit)
+        }
+      }
```
BalmungSan, line 154: So, the idea of this new implementation is to save the `Deferred` instantiation, based on the assumption that at any point in time a `Fiber` is waiting for only one other `Fiber`, and it is being waited by at most another one.

So, the idea is that we will have:
+ `state` which is a mutable reference to the last `Cell` in the chain.
+ `thisCell` which is the chain `Cell` of the `Fiber` trying to access the `Mutex`.
+ `previousCell` which is the previous `Cell` in the chain; maybe the one currently holding the `Mutex`.

**Acquire**:
First, we create our `thisCell` and replace it in the `state`, retrieving the `previousCell`; atomically.
We then use `F.asyncCheckAttempt` to check if the `Mutex` is currently in use. If that is the case, we register the callback in the `previousCell` and wait for it. Otherwise, we can just acquire it.

**Release**:
We call the callback that was registered in our `Cell` and call it. Considering the possibility that maybe no one is actually waiting for us.

**Cancelation**:
In the case we were canceled while waiting for the `Mutex`:
 + Situation A: No one is waiting for us.
Thus, we only need to ensure that any Fiber that follows awaits the previous cell; they will know how to handle the case if it is already released.
 + Situation B: Someone is waiting for us. 
Then we need to unregister our callback from the previous cell and change it to the callback of the Fiber waiting for us; we need to double check in case the `Mutex` was released at that moment to rather notify whoever is waiting for us.

----

I believe this allocates the bare minimum and is concurrent safe.

std/shared/src/main/scala/cats/effect/std/Mutex.scala
```diff
@@ -63,23 +65,141 @@ object Mutex {
...
```
BalmungSan: So, the idea of this new implementation is to save the `Deferred` instantiation, based on the assumption that at any point in time a `Fiber` is waiting for only one other `Fiber`, and it is being waited by at most another one.

So, the idea is that we will have:
+ `state` which is a mutable reference to the last `Cell` in the chain.
+ `thisCell` which is the chain `Cell` of the `Fiber` trying to access the `Mutex`.
+ `previousCell` which is the previous `Cell` in the chain; maybe the one currently holding the `Mutex`.

**Acquire**:
First, we create our `thisCell` and replace it in the `state`, retrieving the `previousCell`; atomically.
We then use `F.asyncCheckAttempt` to check if the `Mutex` is currently in use. If that is the case, we register the callback in the `previousCell` and wait for it. Otherwise, we can just acquire it.

**Release**:
We call the callback that was registered in our `Cell` and call it. Considering the possibility that maybe no one is actually waiting for us.

**Cancelation**:
In the case we were canceled while waiting for the `Mutex`:
 + Situation A: No one is waiting for us.
Thus, we only need to ensure that any Fiber that follows awaits the previous cell; they will know how to handle the case if it is already released.
 + Situation B: Someone is waiting for us. 
Then we need to unregister our callback from the previous cell and change it to the callback of the Fiber waiting for us; we need to double check in case the `Mutex` was released at that moment to rather notify whoever is waiting for us.

----

I believe this allocates the bare minimum and is concurrent safe.

----

PS: Thanks a lot to Arman (@armanbilge) for brainstorming with me on this implementation :)

benchmarks/src/main/scala/cats/effect/benchmarks/AtomicCellBenchmark.scala
```diff
@@ -0,0 +1,66 @@
...
+        data.evalModify { spots =>
+          val availableSpots = spots.zipWithIndex.collect { case (true, idx) => idx }
+          rnd.shuffleVector(availableSpots).map { shuffled =>
+            val acquired = shuffled.headOption
+            val next = acquired.fold(spots)(a => spots.updated(a, false)) // Mark the chosen spot as taken.
+            (next, shuffled.headOption)
+          }
+        }.replicateA_(iterations)
+    }.unsafeRunSync()
+  }
```
BalmungSan: Is this a good benchmark for `AtomicCell`?

armanbilge: Actually I'd love to see a benchmark for `Mutex`. Some of the things to measure would be:
1. happy-path, where only one fiber repeatedly acquires and releases the mutex
2. contention, when many fibers want it, and are potentially canceling as well

BalmungSan: @armanbilge oh yeah I also want to add a benchmark for `Mutex` but that felt harder than just `AtomicCell` so I wanted to first start here.

-----

If the test feels right for `AtomicCell` I will run it and comment the results tomorrow.

BalmungSan:
@djspiewak @armanbilge before running the benchmarks I want to confirm with both of you if you think they are appropriate.

benchmarks/src/main/scala/cats/effect/benchmarks/AtomicCellBenchmark.scala
```diff
@@ -0,0 +1,88 @@
...
+
+  @Param(Array("10000"))
+  var iterations: Int = _
+
+  def initialSpots(): Vector[Boolean] = {
+    Vector.tabulate(size) { i => (i % 2) == 0 }
+  }
+
+  private def evalModifyImpl(cell: IO[AtomicCell[IO, Vector[Boolean]]]): Unit = {
+    (cell, Random.scalaUtilRandomSeedLong[IO](seed = 135L))
```
djspiewak: Normally you don't want to use randomness in benchmarks due to instability, but since the seed is fixed I think this is okay.

djspiewak (dismissed):
Nice work!

djspiewak:
Let's see the benchmark results but overall lgtm

benchmarks/src/main/scala/cats/effect/benchmarks/AtomicCellBenchmark.scala
BalmungSan: @djspiewak
Yeah, I was also afraid of the `Random` but couldn't think of a better example for the test, so that is why I used a fixed seed.
I guess we may also just use a shallow example like `data.evalMapModify(i => IO.unit.as((i + 1) -> ()))`

PS: If we prefer to keep the `Random` does it makes sense to make the `seed` a `@Parameter`?

djspiewak: Nah I think it's fine as-is. We would probably never configure that parameter externally.

benchmarks/src/main/scala/cats/effect/benchmarks/AtomicCellBenchmark.scala
```diff
@@ -0,0 +1,88 @@
...
+@BenchmarkMode(Array(Mode.Throughput))
+@OutputTimeUnit(TimeUnit.SECONDS)
+class AtomicCellBenchmark {
+  @Param(Array("1000"))
+  var size: Int = _
+
+  @Param(Array("10", "100", "1000"))
+  var count: Int = _
+
+  @Param(Array("10000"))
```
djspiewak: Actually on testing, I think we might want to lower this a bit... :-) My rule of thumb is that a benchmark is at a decent level of granularity if it's around 100-1000 ops/s (order of magnitude). At least on my machine, 10 count, 10000 iterations gives me about 0.653 ops/s, which would mean the whole benchmark suite would take a day or so to run.

BalmungSan:
@djspiewak @armanbilge benchmark results added to the description of the PR! :D

djspiewak:
Benchmarks look compelling! Thank you!

PR #3347, opened 2023-01-05
ConcurrentAtomicCell

BalmungSan:
Adds a `Concurrent` based implementation of `AtomicCell` while leaving the current `Async` based one as a fallback.
In theory, the `Async` one should be more performant; especially if #3346 also improves the performance of `Mutex`.

However, it is not clear if such supposed improvement is worth the trickery did in this PR.
If not, we may just delete the `Async` implementation.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -884,7 +884,11 @@ lazy val std = crossProject(JSPlatform, JVMPlatform, NativePlatform)
       ProblemFilters.exclude[DirectMissingMethodProblem](
         "cats.effect.std.Queue#CircularBufferQueue.onOfferNoCapacity"),
       ProblemFilters.exclude[DirectMissingMethodProblem](
-        "cats.effect.std.Queue#DroppingQueue.onOfferNoCapacity")
+        "cats.effect.std.Queue#DroppingQueue.onOfferNoCapacity"),
+      // introduced by #3347
+      // private stuff
+      ProblemFilters.exclude[MissingClassProblem](
+        "cats.effect.std.AtomicCell$Impl")
     )
   )
   .jsSettings(
--- a/std/shared/src/main/scala/cats/effect/std/AtomicCell.scala
+++ b/std/shared/src/main/scala/cats/effect/std/AtomicCell.scala
@@ -132,7 +132,10 @@ object AtomicCell {
      * Initializes the `AtomicCell` using the provided value.
      */
     def of[A](init: A)(implicit F: Async[F]): F[AtomicCell[F, A]] =
-      Mutex[F].map(mutex => new Impl(init, mutex))
+      Mutex[F].map(mutex => new SyncImpl(init, mutex))
+
+    def concurrent[A](init: A)(implicit F: Concurrent[F]): F[AtomicCell[F, A]] =
+      (Ref.of[F, A](init), Mutex[F]).mapN { (ref, m) => new ConcurrentImpl(ref, m) }
 
     /**
      * Initializes the `AtomicCell` using the default empty value of the provided type.
@@ -141,7 +144,40 @@ object AtomicCell {
       of(init = M.empty)
   }
 
-  private final class Impl[F[_], A](
+  private final class ConcurrentImpl[F[_], A](
+      ref: Ref[F, A],
+      mutex: Mutex[F]
+  )(
+      implicit F: Concurrent[F]
+  ) extends AtomicCell[F, A] {
+    override def get: F[A] =
+      mutex.lock.surround(ref.get)
+
+    override def set(a: A): F[Unit] =
+      mutex.lock.surround(ref.set(a))
+
+    override def modify[B](f: A => (A, B)): F[B] =
+      evalModify(a => F.pure(f(a)))
+
+    override def evalModify[B](f: A => F[(A, B)]): F[B] =
+      mutex.lock.surround {
+        ref.get.flatMap(f).flatMap {
+          case (a, b) =>
+            ref.set(a).as(b)
+        }
+      }
+
+    override def evalUpdate(f: A => F[A]): F[Unit] =
+      evalModify(a => f(a).map(aa => (aa, ())))
+
+    override def evalGetAndUpdate(f: A => F[A]): F[A] =
+      evalModify(a => f(a).map(aa => (aa, a)))
+
+    override def evalUpdateAndGet(f: A => F[A]): F[A] =
+      evalModify(a => f(a).map(aa => (aa, aa)))
+  }
+
+  private final class SyncImpl[F[_], A](
       init: A,
       mutex: Mutex[F]
   )(
--- a/std/shared/src/main/scala/cats/effect/std/AtomicCell.scala
+++ b/std/shared/src/main/scala/cats/effect/std/AtomicCell.scala
@@ -131,13 +131,7 @@ object AtomicCell {
      * Initializes the `AtomicCell` using the provided value.
      */
     def of[A](init: A)(implicit F: Concurrent[F]): F[AtomicCell[F, A]] =
-      F match {
-        case f: Async[F] =>
-          AtomicCell.async(init)(f)
-
-        case _ =>
-          AtomicCell.concurrent(init)
-      }
+      AtomicCell.concurrent(init)
 
     @deprecated("Use the version that only requires Concurrent", since = "3.5.0")
     private[std] def of[A](init: A, F: Async[F]): F[AtomicCell[F, A]] =
@@ -154,10 +148,8 @@ object AtomicCell {
       of(M.empty)(F)
   }
 
-  private[effect] def async[F[_], A](init: A)(implicit F: Async[F]): F[AtomicCell[F, A]] =
-    Mutex[F].map(mutex => new AsyncImpl(init, mutex))
-
-  private[effect] def concurrent[F[_], A](init: A)(implicit F: Concurrent[F]): F[AtomicCell[F, A]] =
+  private[effect] def concurrent[F[_], A](init: A)(
+      implicit F: Concurrent[F]): F[AtomicCell[F, A]] =
     (Ref.of[F, A](init), Mutex[F]).mapN { (ref, m) => new ConcurrentImpl(ref, m) }
 
   private final class ConcurrentImpl[F[_], A](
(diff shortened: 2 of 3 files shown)


std/shared/src/main/scala/cats/effect/std/AtomicCell.scala
```diff
@@ -113,6 +113,35 @@ abstract class AtomicCell[F[_], A] {
...
+  private[std] trait MakeInstances extends MakeLowPriorityInstances {
+    implicit def concurrentInstance[F[_]](implicit F: Concurrent[F]): Make[F] =
+      new Make[F] {
+        override def atomicCellOf[A](init: A): F[AtomicCell[F, A]] =
+          F match {
+            case f: Async[_] =>
+              AtomicCell.async(init)(f)
+
+            case _ =>
+              AtomicCell.concurrent(init)
```
BalmungSan: Is this the right way to also fallback to the `Async` based implementation when the concrete type also has an `Async` instance?

armanbilge: All this `Make` stuff? No, we don't need it.

armanbilge: Ah, but the pattern match is right 👍

BalmungSan: > All this Make stuff? No, we don't need it.

That is how `Ref` does it, why it is not needed in this case?

> Ah, but the pattern match is right 👍

**Scala 3** says otherwise :(

armanbilge: > That is how `Ref` does it, why it is not needed in this case?

You can make a `Ref` with either a `Concurrent` or `Sync`. Note that `Sync` may not be `Concurrent`.

But in this case we are choosing between `Concurrent` or `Async`, but every `Async` is `Concurrent`, so really all we need is `Concurrent` :)

> Scala 3 says otherwise :(

Dare I ask ... what's the error 😅

BalmungSan: > You can make a Ref with either a Concurrent or Sync. Note that Sync may not be Concurrent.

🤦‍♂️

I feel stupid now.

> Dare I ask ... what's the error 😅

Sure:
```
[error] -- [E007] Type Mismatch Error: /home/runner/work/cats-effect/cats-effect/std/shared/src/main/scala/cats/effect/std/AtomicCell.scala:130:37 
[error] 130 |              AtomicCell.async(init)(f)
[error]     |                                     ^
[error]     |Found:    cats.effect.kernel.Concurrent[F] & cats.effect.kernel.Async[?1.CAP]
[error]     |Required: cats.effect.kernel.Async[F]
```

BalmungSan: Uhm, it seems matching `Async[F]` will work... Yes, it does :D

build.sbt
```diff
@@ -884,7 +884,10 @@ lazy val std = crossProject(JSPlatform, JVMPlatform, NativePlatform)
       ProblemFilters.exclude[DirectMissingMethodProblem](
         "cats.effect.std.Queue#CircularBufferQueue.onOfferNoCapacity"),
       ProblemFilters.exclude[DirectMissingMethodProblem](
-        "cats.effect.std.Queue#DroppingQueue.onOfferNoCapacity")
+        "cats.effect.std.Queue#DroppingQueue.onOfferNoCapacity"),
+      // introduced by #3347
+      // private stuff
+      ProblemFilters.exclude[MissingClassProblem]("cats.effect.std.AtomicCell$Impl")
```
BalmungSan: This exclusion is safe, right?

std/shared/src/main/scala/cats/effect/std/AtomicCell.scala
```diff
@@ -131,17 +131,62 @@ object AtomicCell {
...
-      of(init = M.empty)
+    def empty[A](implicit M: Monoid[A], F: Concurrent[F]): F[AtomicCell[F, A]] =
+      of(M.empty)
+  }
+
+  def async[F[_], A](init: A)(implicit F: Async[F]): F[AtomicCell[F, A]] =
+    Mutex[F].map(mutex => new AsyncImpl(init, mutex))
+
+  def concurrent[F[_], A](init: A)(implicit F: Concurrent[F]): F[AtomicCell[F, A]] =
+    (Ref.of[F, A](init), Mutex[F]).mapN { (ref, m) => new ConcurrentImpl(ref, m) }
```
armanbilge: Do we have a precedent for leaving these public?

std/shared/src/main/scala/cats/effect/std/AtomicCell.scala
```diff
@@ -131,17 +131,62 @@ object AtomicCell {
     /**
      * Initializes the `AtomicCell` using the provided value.
      */
-    def of[A](init: A)(implicit F: Async[F]): F[AtomicCell[F, A]] =
-      Mutex[F].map(mutex => new Impl(init, mutex))
+    def of[A](init: A)(implicit F: Concurrent[F]): F[AtomicCell[F, A]] =
```
armanbilge, line 133: I'm surprised if MiMa is not yelling about this. We need to add a deprecated, source-incompatible signature for bincompat:
```scala
def of[A](init: A, F: Async[F]): ...
```

Same for `empty`.

tests/shared/src/test/scala/cats/effect/std/AtomicCellSpec.scala
```diff
@@ -20,80 +20,92 @@ package std
...
 
-    "get and update successfully" in real {
-      val op = for {
-        cell <- AtomicCell[IO].of(0)
-        getAndUpdateResult <- cell.getAndUpdate(_ + 1)
-        getResult <- cell.get
-      } yield getAndUpdateResult == 0 && getResult == 1
+  "ConcurrentAtomicCell" should {
+    tests(AtomicCell.concurrent)
+  }
```
armanbilge, line 30: Nice 👍

std/shared/src/main/scala/cats/effect/std/AtomicCell.scala
BalmungSan: I was thinking the same.
On one side, I think it is a good idea to let users be explicit if they want.
On the other hand, this may be an internal API subject to change.

std/shared/src/main/scala/cats/effect/std/AtomicCell.scala
BalmungSan: Shouldn't matter, as you said, an `Async` is a `Concurrent`

armanbilge: It doesn't matter for source-compatibility, but it does matter for binary-compatibility. Binary-compatibility doesn't care about inheritance relationships.

Is MiMa really not warning about this? 😓

BalmungSan: > Binary-compatibility doesn't care about inheritance relationships

Why wouldn't it? How would dynamic dispatch work otherwise?

> Is MiMa really not warning about this? 😓

Nop.

armanbilge: > Why wouldn't it? How would dynamic dispatch work otherwise?

Not sure what to tell you, that's just how it is :) on the JVM, the binary signature is defined by the precise fully-qualified classnames of the arguments. There is no dynamic dispatch on method arguments, only on `this`.

---

I figured out why MiMa is not warning. It's because of this.
https://github.com/typelevel/cats-effect/blob/df9f0ef7cb431c69d3e360e19159e1341704ccbd/std/shared/src/main/scala/cats/effect/std/AtomicCell.scala#L128

It's package-private, so MiMa ignores it even though it is actually public API. The fact this even works is probably due to https://github.com/scala/bug/issues/6794.

I learned this the hard way in FS2, when MiMa didn't warn me and I broke bincompat. See https://github.com/typelevel/fs2/pull/2940#discussion_r916698532.

IMO the only safe thing to do is to make this public.

std/shared/src/main/scala/cats/effect/std/AtomicCell.scala
```diff
@@ -125,23 +125,75 @@ object AtomicCell {
...
      * Initializes the `AtomicCell` using the default empty value of the provided type.
      */
-    def empty[A](implicit M: Monoid[A], F: Async[F]): F[AtomicCell[F, A]] =
-      of(init = M.empty)
+    def empty[A](implicit M: Monoid[A], F: Concurrent[F]): F[AtomicCell[F, A]] =
+      of(M.empty)
+
+    @deprecated("Use the version that only requires Concurrent", since = "3.5.0")
+    private[std] def empty[A](M: Monoid[A], F: Async[F]): F[AtomicCell[F, A]] =
+      of(M.empty)(F)
```
djspiewak, line 148: I'm pretty sure we don't need this since `Async <: Concurrent`

std/shared/src/main/scala/cats/effect/std/AtomicCell.scala
djspiewak: I would definitely make these private. :-) They're very internal.

djspiewak (changes requested):
Some minor changes. Also can we add benchmarks to decide whether it's worth having the `Async` version?

std/shared/src/main/scala/cats/effect/std/AtomicCell.scala
armanbilge: We do, bin-compatiblity doesn't care about inheritance.

BalmungSan:
> Also can we add benchmarks to decide whether it's worth having the Async version?

I really would like to benchmark this `Async` version plus the `Async` `Mutex` of #3346 with the `Concurrent` version of both.
However, I have never done proper benchmarking before, is there some guide / example you would recommend? Also, I guess it would make sense to add those benchmarks to the repo itself.

std/shared/src/main/scala/cats/effect/std/AtomicCell.scala
BalmungSan: We already discussed this in one of the _resolved_ reviews. I also thought it wasn't necessary.

std/shared/src/main/scala/cats/effect/std/AtomicCell.scala
BalmungSan: Done :)

armanbilge:
Discussed offline with Luis: I propose we remove the `Async` variant for now, so we can land this PR, which is blocking 3.5.0.

Afterwards we can play with benchmarks and optimizations forward-compatibly, esp. since the performance of an `Async` `AtomicCell` hinges on the performance of an `Async` mutex.

std/shared/src/main/scala/cats/effect/std/AtomicCell.scala
djspiewak: This surprises me immensely. I thought that `invokevirtual` respected contravariance.

djspiewak:
Agree with the removing the `Async` version for now.

Regarding how to benchmark… I always copy/paste from the existing benchmarks dir. :-) It's pretty self explanatory.

djspiewak (approved):

PR #3351, opened 2023-01-07
Fix cancelation handling in `memoize`

### LINKED ISSUE #3348: `memoize` doc and implementation out of sync
The `memoize` doc says

> If all gets are canceled prior to fa completing, it will be canceled and evaluated again the next time get is sequenced.

But that's not what `memoize` really does. It will reevaluate `fa` if the first `get` that started is cancelled before completion.

```scala
scala> (IO.println("START") *> IO.sleep(2.seconds) *> IO.println("FINISH")).memoize.flatMap(io =>
     |   io.start.flatMap(f => f.cancel.delayBy(1.second)).start *> IO.sleep(200.millis) *> io.start *> IO.sleep(3.seconds)
     | ).unsafeRunSync()
START
START
FINISH
```

armanbilge:
In the old implementation, the first fiber to access the memo would take responsibility for evaluating the effect, and all other fibers would subscribe to its result. However, this meant that if this fiber got canceled, so would the effect evaluation, even if there are active subscribers.

In this new implementation, the first fiber to access the memo setups a background fiber to evaluate the effect. Then, all active subscribers take responsibility for it.

Fixes https://github.com/typelevel/cats-effect/issues/3348.

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
@@ -61,14 +61,19 @@ trait GenConcurrent[F[_], E] extends GenSpawn[F, E] {
                       other
                   }
 
-                  fa.guaranteeCase {
-                    case Outcome.Canceled() =>
-                      tryComplete(Finished(Right(productR(canceled)(never))))
-                    case Outcome.Errored(err) =>
-                      tryComplete(Finished(Left(err)))
-                    case Outcome.Succeeded(fa) =>
-                      tryComplete(Finished(Right(fa)))
-                  }
+                  fa
+                    // hack around functor law breakage
+                    .flatMap(F.pure(_))
+                    .handleErrorWith(F.raiseError(_))
+                    // end hack
+                    .guaranteeCase {
+                      case Outcome.Canceled() =>
+                        tryComplete(Finished(Right(productR(canceled)(never))))
+                      case Outcome.Errored(err) =>
+                        tryComplete(Finished(Left(err)))
+                      case Outcome.Succeeded(fa) =>
+                        tryComplete(Finished(Right(fa)))
+                    }
                 }
 
                 val eval = go.start.flatMap { fiber =>
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
@@ -63,7 +63,7 @@ trait GenConcurrent[F[_], E] extends GenSpawn[F, E] {
 
                   fa.guaranteeCase {
                     case Outcome.Canceled() =>
-                      tryComplete(Finished(Right(never)))
+                      tryComplete(Finished(Right(productR(canceled)(never))))
                     case Outcome.Errored(err) =>
                       tryComplete(Finished(Left(err)))
                     case Outcome.Succeeded(fa) =>
@@ -73,7 +73,7 @@ trait GenConcurrent[F[_], E] extends GenSpawn[F, E] {
 
                 val eval = go.start.flatMap { fiber =>
                   deferredFiber.complete(fiber) *>
-                    poll(fiber.join.flatMap(_.embedNever))
+                    poll(fiber.join.flatMap(_.embed(productR(canceled)(never))))
                 }
 
                 Evaluating(deferredFiber, 1) -> eval
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
@@ -45,40 +45,65 @@ trait GenConcurrent[F[_], E] extends GenSpawn[F, E] {
     implicit val F: GenConcurrent[F, E] = this
 
     ref[Memoize[F, E, A]](Unevaluated()) map { state =>
-      def eval: F[A] =
-        deferred[Unit] flatMap { latch =>
+      // start running the effect, or subscribe if it already is
+      def evalOrSubscribe: F[A] =
+        deferred[Fiber[F, E, A]] flatMap { deferredFiber =>
           uncancelable { poll =>
             state.modify {
               case Unevaluated() =>
-                val go =
-                  poll(fa).guaranteeCase { outcome =>
-                    val stateAction = outcome match {
-                      case Outcome.Canceled() =>
-                        state.set(Unevaluated())
-                      case Outcome.Errored(err) =>
-                        state.set(Finished(Left(err)))
-                      case Outcome.Succeeded(fa) =>
-                        state.set(Finished(Right(fa)))
-                    }
-                    stateAction <* latch.complete(())
+                // run the effect, and if this fiber is still relevant set its result
+                val go = {
+                  def tryComplete(result: Memoize[F, E, A]): F[Unit] = state.update {
+                    case Evaluating(fiber, _) if fiber eq deferredFiber =>
+                      // we are the blessed fiber of this memo
+                      result
+                    case other => // our outcome is no longer relevant
+                      other
                   }
 
-                Evaluating(latch.get) -> go
+                  fa.guaranteeCase {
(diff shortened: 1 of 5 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
```diff
@@ -45,40 +45,66 @@ trait GenConcurrent[F[_], E] extends GenSpawn[F, E] {
...
+                    case Outcome.Errored(err) =>
+                      tryComplete(Finished(Left(err)))
+                    case Outcome.Succeeded(fa) =>
+                      tryComplete(Finished(Right(fa)))
+                  }
+                }
+
+                val eval = go.start.flatMap { fiber =>
+                  deferredFiber.complete(fiber) *>
+                    poll(fiber.join.flatMap(_.embed(productR(canceled)(never))))
```
armanbilge, line 81: I wonder if we should have a combinator for:
- `productR(canceled)(never)`
- `embed(productR(canceled)(never))`

kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
```diff
@@ -45,40 +45,71 @@ trait GenConcurrent[F[_], E] extends GenSpawn[F, E] {
...
+                    case other => // our outcome is no longer relevant
+                      other
+                  }
+
+                  fa
+                    // hack around functor law breakage
+                    .flatMap(F.pure(_))
+                    .handleErrorWith(F.raiseError(_))
+                    // end hack
+                    .guaranteeCase {
```
armanbilge, line 69: Well, I finally came face to face with it 🤠

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -1380,12 +1380,15 @@ abstract private[effect] class ResourceConcurrent[F[_]]
     fa.race(fb)
 
   override def memoize[A](fa: Resource[F, A]): Resource[F, Resource[F, A]] = {
-    Resource.eval(F.ref(false)).flatMap { allocated =>
-      val fa2 = F.uncancelable(poll => poll(fa.allocatedCase) <* allocated.set(true))
+    Resource.eval(F.ref(List.empty[Resource.ExitCase => F[Unit]])).flatMap { release =>
+      val fa2 = F.uncancelable { poll =>
+        poll(fa.allocatedCase).flatMap { case (a, r) => release.update(r :: _).as(a) }
+      }
```
armanbilge, line 1386: In some awkward situations, the resource allocation could be canceled but still succeed due to uncancelability, in which case it's not actually memoized (something something functor outlaws). So we need to be prepared to store multiple finalizers, since the memo may be access again, thus restarting the allocation process.

If there is more than one finalizer they run sequentially when the memo is released, but I could change this to parallel. Thoughts?

kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
```diff
@@ -45,40 +45,71 @@ trait GenConcurrent[F[_], E] extends GenSpawn[F, E] {
...
                       case Outcome.Canceled() =>
-                        state.set(Unevaluated())
+                        tryComplete(Finished(Right(productR(canceled)(never))))
```
durban, line 71: If I'm reading it correctly, this caches a "cancelled result" if `fa` self-cancels. This seems like a semantic change. (Which is not necessarily a problem, but maybe worth mentioning in release notes?)

armanbilge: Good eye!

> This seems like a semantic change.

Well, it's not entirely clear if it is a semantic change or it is part of the bug fix. Or both?

1. In the current implementation, self-cancelation propagates to the fiber running the memoized effect. I assume we want to keep this semantic.
2. The fix in this PR is about removing the "special" status of the first fiber to access the memo, which suggests that all actively subscribed fibers should see the same result when the memo completes.
3. When a new fiber accesses the memo, there is a race condition with the completion of the effect. To ensure a consistent result whether it subscribes before or after the completion, the cancelation must be memoized.

kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
djspiewak: Why do we need to guarantee that cancelation is fully surfaced? Is your concern something like `F.canceled.uncancelable.memoize`?

kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
djspiewak: Originally this was `embedNever`, but the semantics were implemented incorrectly, and then it turned out that people preferred the incorrect semantics.

kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
djspiewak: So in the event that the inner `F[A]` is canceled on its first evaluation, we want to *re-evaluate* the effect next time, since in most cases this is not caused by self-cancelation, but rather by external cancelation. In the event that it's self-cancelation, the re-evaluation will likely produce the same result. In the event that it's external cancelation, re-evaluating will likely complete successfully, so this is a rather large semantic change.

Cancelation is just special in this case. Here's what I think we should do:

- If the result of the evaluation is `Canceled`, that result should *not* be memoized and we should reevaluate the effect
- The fiber who is actually evaluating the effect should be canceled. This is indeed making them "special", and that first fiber will see a different result than the others. This is justified because, in 99% of the cases, the *reason* the outcome was `Canceled` is because that very fiber was canceled.
- If there are other waiters on the current evaluation, they can race for the reevaluation.

djspiewak (changes requested):

kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
armanbilge: Because our tests want us to 😅 I added this to fix the `ResourceSpec` failure.

https://github.com/typelevel/cats-effect/blob/0079d5d14a7c325ef69e89a6c5e6e0f8e37c216d/tests/shared/src/test/scala/cats/effect/MemoizeSpec.scala#L99-L101

https://github.com/typelevel/cats-effect/blob/0079d5d14a7c325ef69e89a6c5e6e0f8e37c216d/tests/shared/src/test/scala/cats/effect/ResourceSpec.scala#L1126-L1128

kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
armanbilge: > since in most cases this is not caused by self-cancelation, but rather by external cancelation.

I've already handled this case here. If the effect is canceled externally, then its outcome is no longer relevant.

https://github.com/typelevel/cats-effect/blob/dbfa07289a25dd56cf355f6e4d8bffb4c8f22f32/kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala#L56-L62

https://github.com/typelevel/cats-effect/blob/dbfa07289a25dd56cf355f6e4d8bffb4c8f22f32/kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala#L98-L104

kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
armanbilge: \*shakes fist\*

kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
armanbilge: Added a test in https://github.com/typelevel/cats-effect/pull/3351/commits/8f27753529a13742360b29161174d26a3838937c to demonstrate this.

djspiewak (approved):

PR #3353, opened 2023-01-09
Add two new combinators: `cedeMap` and `intercede`

### LINKED ISSUE #3318: "`cedeMap`" and "`intercede`" combinators
Explicitly `cede`-ing between expensive compute-bound operations is one of the strategies for addressing CPU-starvation and is suggested in the warning.

https://github.com/typelevel/cats-effect/blob/832079aecbbf956fb18f11392cfeec6cdaa52915/core/shared/src/main/scala/cats/effect/CpuStarvationCheck.scala#L43-L44

How to do this is described in the scaladocs for `cede`.

https://github.com/typelevel/cats-effect/blob/52b5a3ba1623ff40fc48318c9a0bfae558f0e5ed/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala#L263-L269

Implementing that correctly is not completely trivial (\*cough\* https://github.com/typelevel/cats-effect/pull/3166 😜) and noisy enough that it probably deserves its own combinator.

In fact I think we need a couple combinators, depending on whether or not the `expensiveWork()` is in `F[_]` or not.

```scala
def cedeMap[A, B](fa: F[A])(f: A => B): F[B] =
  (fa <* cede).map(a => f(a)).guarantee(cede)

def intercede[A](fa: F[A]): F[A] =
  cede *> fa.guarantee(cede) 
```

(Names subject to bikeshed.) These should also be added as syntax and on `IO` itself.

I think both variants are important, because for example the following would _not_ achieve the desired semantics.
```scala
fa.flatMap(data => intercede(F.pure(expensiveWork(data))))
```
There, `expensiveWork()` would be computed eagerly when the `pure(...)` is _constructed_ (before the `cede`), not when it is _interpreted_ (after the `cede`).

biuld:
This pull request resolves #3318.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -1317,6 +1317,18 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
   def cede: IO[Unit] = Cede
 
   /**
+   * Functor map, but causes a reschedule before and after `f`
+   */
+  def cedeMap[A, B](fa: IO[A])(f: A => B): IO[B] =
+    (fa <* cede).map(a => f(a)).guarantee(cede)
+
+  /**
+   * causes a reschedule before and after `fa`
+   */
+  def intercede[A](fa: IO[A]): IO[A] =
+    cede *> fa.guarantee(cede)
+
+  /**
    * This is a low-level API which is meant for implementors, please use `background`, `start`,
    * `async`, or `Deferred` instead, depending on the use case
    */
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -498,6 +498,20 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   def map[B](f: A => B): IO[B] = IO.Map(this, f, Tracing.calculateTracingEvent(f))
 
   /**
+   * Functor map, but causes a reschedule before and after `f`. For more information, checkout
+   * `cede` in the companion object.
+   */
+  def cedeMap[B](f: A => B): IO[B] =
+    (this <* IO.cede).map(a => f(a)).guarantee(IO.cede)
+
+  /**
+   * Causes a reschedule before and after `fa`. For more information, checkout `cede` in the
+   * companion object.
+   */
+  def intercede: IO[A] =
+    IO.cede *> this.guarantee(IO.cede)
+
+  /**
    * Applies rate limiting to this `IO` based on provided backpressure semantics.
    *
    * @return
@@ -1315,18 +1329,6 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
    * turn may be adjusted by overriding [[IOApp.runtimeConfig]].
    */
   def cede: IO[Unit] = Cede
-
-  /**
-   * Functor map, but causes a reschedule before and after `f`
-   */
-  def cedeMap[A, B](fa: IO[A])(f: A => B): IO[B] =
-    (fa <* cede).map(a => f(a)).guarantee(cede)
-
-  /**
-   * causes a reschedule before and after `fa`
-   */
-  def intercede[A](fa: IO[A]): IO[A] =
-    cede *> fa.guarantee(cede)
 
   /**
    * This is a low-level API which is meant for implementors, please use `background`, `start`,
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
@@ -286,22 +286,13 @@ trait GenSpawn[F[_], E] extends MonadCancel[F, E] with Unique[F] {
    * Functor map, but causes a reschedule before and after `f`
    */
   def cedeMap[A, B](fa: F[A])(f: A => B): F[B] =
-    for {
-      a <- fa
-      _ <- cede
-      b = f(a)
-      _ <- cede
-    } yield b
+    (fa <* cede).map(a => f(a)).guarantee(cede)
 
   /**
    * Causes a reschedule before and after `fa`
    */
   def intercede[A](fa: F[A]): F[A] =
-    for {
-      _ <- cede
-      a <- fa
-      _ <- cede
-    } yield a
+    cede *> fa.guarantee(cede)
 
   /**
    * A low-level primitive for racing the evaluation of two fibers that returns the [[Outcome]]
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
@@ -283,6 +283,27 @@ trait GenSpawn[F[_], E] extends MonadCancel[F, E] with Unique[F] {
   def cede: F[Unit]
 
   /**
+   * Functor map, but causes a reschedule before and after `f`
+   */
+  def cedeMap[A, B](fa: F[A])(f: A => B): F[B] =
+    for {
+      a <- fa
+      _ <- cede
+      b = f(a)
+      _ <- cede
+    } yield b
+
+  /**
+   * Causes a reschedule before and after `fa`
+   */
+  def intercede[A](fa: F[A]): F[A] =
+    for {
+      _ <- cede
+      a <- fa
+      _ <- cede
+    } yield a
+
+  /**
    * A low-level primitive for racing the evaluation of two fibers that returns the [[Outcome]]
    * of the winner and the [[Fiber]] of the loser. The winner of the race is considered to be
    * the first fiber that completes with an outcome.
--- a/kernel/shared/src/test/scala/cats/effect/kernel/SyntaxSpec.scala
+++ b/kernel/shared/src/test/scala/cats/effect/kernel/SyntaxSpec.scala
@@ -104,7 +104,8 @@ class SyntaxSpec extends Specification {
     }
   }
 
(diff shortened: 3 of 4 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1316,6 +1316,18 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
...
+   * Functor map, but causes a reschedule before and after `f`
+   */
+  def cedeMap[A, B](fa: IO[A])(f: A => B): IO[B] =
+    (fa <* cede).map(a => f(a)).guarantee(cede)
+
+  /**
+   * causes a reschedule before and after `fa`
+   */
+  def intercede[A](fa: IO[A]): IO[A] =
+    cede *> fa.guarantee(cede)
```
armanbilge: Thanks for opening the PR! What would you think about moving these methods onto the `IO` instance, so that instead of
```scala
IO.cedeMap(ioa)(a => foo(a))
IO.intercede(ioa)
```
users can write
```
ioa.cedeMap(a => foo(a))
ioa.intercede
```

--- 

However, the methods as written here would be an excellent addition to `GenSpawn` in the `kernel` module.

Let me know if you have any questions!

biuld: True! `cedeMap` and `intercede` as instance methods would be less boilerplate.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -497,6 +497,20 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
...
+   */
+  def cedeMap[B](f: A => B): IO[B] =
+    (this <* IO.cede).map(a => f(a)).guarantee(IO.cede)
+
+  /**
+   * Causes a reschedule before and after `fa`. For more information, checkout `cede` in the
+   * companion object.
+   */
+  def intercede: IO[A] =
+    IO.cede *> this.guarantee(IO.cede)
```
armanbilge, line 512: Nice!

Would you mind adding versions of these methods to `GenSpawn` in the kernel module? These would be more similar to your original definitions, which are in `F[_]` and take the effect as an argument (instead of instance methods).

https://github.com/typelevel/cats-effect/blob/130fdcc7af75ed769f001d9f0afbff6ccf8b9488/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala#L283

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -282,6 +282,27 @@ trait GenSpawn[F[_], E] extends MonadCancel[F, E] with Unique[F] {
...
+  /**
+   * Functor map, but causes a reschedule before and after `f`
+   */
+  def cedeMap[A, B](fa: F[A])(f: A => B): F[B] =
+    for {
+      a <- fa
+      _ <- cede
+      b = f(a)
+      _ <- cede
+    } yield b
```
armanbilge: Thanks for adding these! Is there a particular reason you wrote them as `for ... yield`? Unfortunately it's not quite correct, because it's missing the `guarantee(...)`. Also, in my personal opinion, it's a bit harder to read :)

biuld: > Is there a particular reason you wrote them as for ... yield?

No, it's I just can't find \*> and <\* combinators here. 🤦

armanbilge: Huh, were they not compiling? They come from here:
https://github.com/typelevel/cats-effect/blob/d6f9f685bc4c63ce452857c5c466410c15c40092/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala#L22

biuld: Oh, I missed the fact that MonadCancel is cats Monad! Will change that in a minute.

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -282,6 +282,18 @@ trait GenSpawn[F[_], E] extends MonadCancel[F, E] with Unique[F] {
...
+   * Functor map, but causes a reschedule before and after `f`
+   */
+  def cedeMap[A, B](fa: F[A])(f: A => B): F[B] =
+    (fa <* cede).map(a => f(a)).guarantee(cede)
+
+  /**
+   * Causes a reschedule before and after `fa`
+   */
+  def intercede[A](fa: F[A]): F[A] =
+    cede *> fa.guarantee(cede)
```
armanbilge, line 295: Excellent! :)

If you don't mind, there is _one_ more place we need to add these methods, which is as syntax here:
https://github.com/typelevel/cats-effect/blob/130fdcc7af75ed769f001d9f0afbff6ccf8b9488/kernel/shared/src/main/scala/cats/effect/kernel/syntax/GenSpawnSyntax.scala#L45

These should delegate directly to the typeclass instance, instead of duplicating the implementation.

biuld: No problem!

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -282,6 +282,18 @@ trait GenSpawn[F[_], E] extends MonadCancel[F, E] with Unique[F] {
    */
   def cede: F[Unit]
 
+  /**
+   * Functor map, but causes a reschedule before and after `f`
+   */
```
armanbilge, line 287:
```suggestion
  /**
   * Like functor [[map]], but inserts a [[cede]] before and after `f`. Use this when `f` is a long-running, compute-bound operation.
   *
   * @see [[cede]] for more details
   */
```

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -282,6 +282,18 @@ trait GenSpawn[F[_], E] extends MonadCancel[F, E] with Unique[F] {
...
+  def cedeMap[A, B](fa: F[A])(f: A => B): F[B] =
+    (fa <* cede).map(a => f(a)).guarantee(cede)
+
+  /**
+   * Causes a reschedule before and after `fa`
+   */
```
armanbilge, line 293:
```suggestion
  /**
   * Insert a [[cede]] before and after `fa`. Use this when `fa` involves a long-running, compute-bound operation outside of the monadic context.
   *
   * @see [[cede]] for more details
   */
```

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -282,6 +282,18 @@ trait GenSpawn[F[_], E] extends MonadCancel[F, E] with Unique[F] {
    */
   def cede: F[Unit]
```
armanbilge, line 283:
```suggestion
   *
   * @see [[cedeMap]], [[intercede]]
   */
  def cede: F[Unit]
```

kernel/shared/src/test/scala/cats/effect/kernel/SyntaxSpec.scala
```diff
@@ -139,6 +139,11 @@ class SyntaxSpec extends Specification {
       val result = target.bothOutcome(another)
       result: F[(Outcome[F, E, A], Outcome[F, E, B])]
     }
+
+    {
+      val result = target.intercede
+      result: F[A]
+    }
```
armanbilge, line 152: Excellent, thanks for adding this test! Would you mind adding a test for `cedeMap` as well?

armanbilge (commented):
Looking great! Just a couple docs tweaks.

kernel/shared/src/test/scala/cats/effect/kernel/SyntaxSpec.scala
biuld: Actually the reason that I leave out the test for `cedeMap` is I don't know how to obtain `f: A => B` in test. I might need some time on getting familiar with how specs2 work. 😂

armanbilge: @biuld just add `f: A => B` to the function arguments here:
https://github.com/typelevel/cats-effect/blob/ed15013511e27c5879c65ca8c0267b2f96e18924/kernel/shared/src/test/scala/cats/effect/kernel/SyntaxSpec.scala#L107

biuld: Done.

armanbilge (approved):
👍 thanks for all your good work on this! But someone else should bikeshed the names because I just made them up :)

djspiewak:
So I genuinely love the idea of these combinators. I'm not 100% sold on the API, but I'm thinking about it.

djspiewak:
Going to punt these from 3.5 just to give us a bit more time to bikeshed. Like I said, I absolutely love the concept I just want to make sure we get the API right (not just the names, but also the ergonomics). Outstanding work, again!

BalmungSan (commented):
Leaving my two cents:

1. Personally I would prefer a more intuitive name like `computeMap` over `cedeMap`. Of course, the **Scaladoc** should mention and link `cede`.
2. In a similar vein, I would like to see two other static combinators like:

```scala
object IO {
  def compute[A](thunk: => A): IO[A] =
      IO.cede >> IO.pure(thunk) >> IO.cede // Not sure if delay + guarantee would be better.

  def attemptCompute[A](thunk: => Either[Throwable, A]): IO[A] =
    IO.cede >> IO.fromEither(thunk) >> IO.cede // ditto delay + guarantee.
}
```

> These may be in another PR tho.
> Especially since those should also be added to `Async`

PR #3354, opened 2023-01-11
update the ref example

samspills:
While looking at `Ref` today I noticed the example was a bit dated. I've made the following changes:
- use scala-cli + `IOApp.Simple` for a complete runnable example
- minor updates to the example code
  - more obvious difference between worker ID and the ref count in
    `putStrLn`
  - switch to `updateAndGet` from modify, I think it's a bit more
    intuitive to see 0,0,0,1,2,3 in the output vs. 0,0,0,0,1,2
- add `updateAndGet` to class reference, since I'm using it in the example
- shift example output to below the code

~One concern is that the `using` statements in scala-cli are hardcoded to scala 2.13.10 and cats-effect 3.4.4. Ideally these would be set as mdoc variables, but I'm not sure where/how to go about doing that.~ Update: Dropped the scala/jvm usings, and the cats-effect version isn't adding much to the existing toil of an update (a future consideration though)

### DIFF
--- a/docs/std/ref.md
+++ b/docs/std/ref.md
@@ -9,6 +9,7 @@ A concurrent mutable reference.
 abstract class Ref[F[_], A] {
   def get: F[A]
   def set(a: A): F[Unit]
+  def updateAndGet(f: A => A): F[A]
   def modify[B](f: A => (A, B)): F[B]
   // ... and more
 }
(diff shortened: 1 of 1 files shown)


armanbilge:
Thanks, would you mind retargeting at 3.4.x so it can go live sooner? :)

docs/std/ref.md
```diff
@@ -27,48 +28,57 @@ upon object reference equality.
...
-#3 >> 2
-#2 >> 1
-```
+In this example, the workers will concurrently run and update the value of the `Ref`.
 
 ```scala mdoc:reset:silent
-import cats.effect.{IO, Sync, Ref}
+//> using scala "2.13.10"
+//> using jvm "temurin:11"
+//> using lib "org.typelevel::cats-effect:3.4.4"
```
armanbilge: I feel like we can drop the scala/jvm `using`s, since this code should work on practically any scala and any jvm ... right? :)

Regarding the CE version, we already manually update those on every release, so it's just a drop in the bucket pending proper automation ...

armanbilge: e.g. see:
- https://github.com/typelevel/cats-effect/pull/3338

samspills: > I feel like we can drop the scala/jvm usings, since this code should work on practically any scala and any jvm ... right? :)

You are right 😄 Made that switch!

armanbilge (approved):
Thanks, I really like what you did! :)

Issue #3321, opened 2022-12-23
Disable starvation checking by default in browsers


djspiewak:
Due to factors like timer clamping, background tabs, and such, I'm not sure the starvation checker as currently formulated makes that much sense in a web browser environment. It's still *useful*, but it's going to yield false negatives under reasonably expected circumstances, so it seems better to default it to "off".

To be clear, I think it should still default to "on" for other environments, including Node.js

Related to #3320

armanbilge:
Well, maybe. I'd like to hear more from users about it first. So far in https://github.com/typelevel/cats-effect/issues/3320#issue-1505471485 it actually identified a legitimate issue, although it was unclear what mitigating action can be taken.

djspiewak:
Doesn't timer clamping in background scenarios make it almost a guarantee that *every* Cats Effect browser application will hit the warning, though?

armanbilge:
Hmm, fair point.

https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#timeouts_in_inactive_tabs

> Firefox Desktop and Chrome both have a minimum timeout of 1 second for inactive tabs.

Which coincidentally matches our default :)

But also:

> Firefox for Android has a minimum timeout of 15 minutes for inactive tabs and may unload them entirely.

So, yeah, fair point ...

armanbilge:
We might also dynamically enable/disable the checker depending on if the window/tab is active or not.

https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API

armanbilge:
Following https://github.com/typelevel/cats-effect/issues/1959#issuecomment-1379266594 my suggestion is that most browser applications probably don't want to use `IOApp` anyway, so there will be no starvation checker.

PR #3355, opened 2023-01-11
Encourage alternatives to `IOApp` for browser apps

armanbilge:
While `IOApp` can be used to make a browser application, in practice it brings little value over `unsafeRunAndForget()`.

Various caveats of using `IOApp` in the browser:
- there is no need for a "keep alive" fiber
- there are no command line args and no exit codes
- there are no signals to setup for cancellation/fiber dumps
- indeed, there is [no reliable notion of life cycle](https://developer.mozilla.org/en-US/docs/Web/API/Window/unload_event#usage_notes) at all
- it interacts poorly with hot module reloading
  https://github.com/typelevel/cats-effect/issues/1959
- starvation checking might be annoying[^1]
  https://github.com/typelevel/cats-effect/issues/3321

Furthermore, web frameworks such as Calico or Tyrian generally offer a dedicated API for defining the application entrypoint.

[^1]: although I still think it's valuable

### DIFF
--- a/core/js/src/main/scala/cats/effect/IOApp.scala
+++ b/core/js/src/main/scala/cats/effect/IOApp.scala
@@ -127,6 +127,12 @@ import scala.util.Try
  * number of compute worker threads to "make room" for the I/O workers, such that they all sum
  * to the number of physical threads exposed by the kernel.
  *
+ * @note
+ *   While [[IOApp]] works perfectly fine for browser applications, in practice it brings little
+ *   value over calling [[IO.unsafeRunAndForget]]. You can use your UI framework's preferred API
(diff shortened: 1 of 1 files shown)


Jasper-M:
Should you `import cats.effect.unsafe.implicits.global` for the alternatives or is there something better?

armanbilge:
@Jasper-M good question. So the "first-class" alternatives are using `IOWebApp` in Calico and `TyrianApp` in Tyrian, which are certainly better than importing the global runtime. If your framework doesn't provide such an entrypoint, then `unsafeRunAndForget()` with the global runtime will work fine.

djspiewak (approved):

PR #3357, opened 2023-01-12
Fix null pointer in tracing on javascript

Jasper-M:
As seen here: https://discord.com/channels/632277896739946517/632278585700384799/1062698867172311090

### DIFF
--- a/core/js/src/main/scala/cats/effect/tracing/TracingPlatform.scala
+++ b/core/js/src/main/scala/cats/effect/tracing/TracingPlatform.scala
@@ -104,11 +104,12 @@ private[tracing] abstract class TracingPlatform { self: Tracing.type =>
   private[tracing] def applyStackTraceFilter(
       callSiteClassName: String,
       callSiteMethodName: String,
-      callSiteFileName: Option[String]): Boolean = {
+      callSiteFileName: String): Boolean = {
 
     // anonymous lambdas can only be distinguished by Scala source-location, if available
     def isInternalScalaFile =
-      callSiteFileName.exists(fileName => !fileName.endsWith(".js") && isInternalFile(fileName))
+      callSiteFileName != null && !callSiteFileName.endsWith(".js") && isInternalFile(
+        callSiteFileName)
 
     // this is either a lambda or we are in Firefox
     def isInternalJSCode = callSiteClassName == "<jscode>" &&
--- a/core/js/src/main/scala/cats/effect/tracing/TracingPlatform.scala
+++ b/core/js/src/main/scala/cats/effect/tracing/TracingPlatform.scala
@@ -108,7 +108,7 @@ private[tracing] abstract class TracingPlatform { self: Tracing.type =>
 
     // anonymous lambdas can only be distinguished by Scala source-location, if available
     def isInternalScalaFile =
-      callSiteFileName ne null && !callSiteFileName.endsWith(".js") && isInternalFile(
+      (callSiteFileName ne null) && !callSiteFileName.endsWith(".js") && isInternalFile(
         callSiteFileName)
 
     // this is either a lambda or we are in Firefox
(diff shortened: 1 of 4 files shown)


armanbilge:
Thanks! Instead of using `Option` here, what would you think about just doing a `null` check? It's not nice code, but it avoids the allocation and will be better for performance.

core/js/src/main/scala/cats/effect/tracing/TracingPlatform.scala
```diff
@@ -104,11 +104,12 @@ private[tracing] abstract class TracingPlatform { self: Tracing.type =>
   private[tracing] def applyStackTraceFilter(
       callSiteClassName: String,
       callSiteMethodName: String,
-      callSiteFileName: Option[String]): Boolean = {
+      callSiteFileName: String): Boolean = {
 
     // anonymous lambdas can only be distinguished by Scala source-location, if available
     def isInternalScalaFile =
-      callSiteFileName.exists(fileName => !fileName.endsWith(".js") && isInternalFile(fileName))
+      callSiteFileName != null && !callSiteFileName.endsWith(".js") && isInternalFile(
```
armanbilge:
```suggestion
      callSiteFileName ne null && !callSiteFileName.endsWith(".js") && isInternalFile(
```

core/js/src/main/scala/cats/effect/tracing/TracingPlatform.scala
```diff
@@ -107,8 +107,9 @@ private[tracing] abstract class TracingPlatform { self: Tracing.type =>
       callSiteFileName: String): Boolean = {
 
     // anonymous lambdas can only be distinguished by Scala source-location, if available
-    def isInternalScalaFile = !callSiteFileName.endsWith(".js") &&
-      isInternalFile(callSiteFileName)
+    def isInternalScalaFile =
+      callSiteFileName ne null && !callSiteFileName.endsWith(".js") && isInternalFile(
```
armanbilge:
```suggestion
      (callSiteFileName ne null) && !callSiteFileName.endsWith(".js") && isInternalFile(
```

core/js/src/main/scala/cats/effect/tracing/TracingPlatform.scala
Jasper-M: The compiler emits the same bytecode for both 😛 I hope Scala.js does the same.

armanbilge: > I hope Scala.js does the same.

Seems not:

```
  $dp_equals__O__Z($m_Ltest$package$().whatever__O(), null);
  ($m_Ltest$package$().whatever__O() !== null)
```

armanbilge: Neither does Scala JVM.

```
this.whatever().equals(null);
this.whatever() != null;
```

> The compiler emits the same bytecode for both

What is this based on? 🤔

Jasper-M: I tested in the REPL
```scala
scala> :paste -raw
// Entering paste mode (ctrl-D to finish)

class Foo{ def foo(a: String) = a != null; def bar(a: String) = a ne null }

// Exiting paste mode, now interpreting.

scala> :javap -cp Foo
Compiled from "<pastie>"
public class Foo {
  public boolean foo(java.lang.String);
    Code:
       0: aload_1
       1: ifnull        8
       4: iconst_1
       5: goto          9
       8: iconst_0
       9: ireturn

  public boolean bar(java.lang.String);
    Code:
       0: aload_1
       1: ifnull        8
       4: iconst_1
       5: goto          9
       8: iconst_0
       9: ireturn
```

armanbilge: sorry, you are completely right!! I think I was doing something dumb

Jasper-M: Doesn't matter 😛 your change makes sense either way. I wasn't completely sure myself.

Issue #3359, opened 2023-01-13
Memory leak in v3.4.4 for cats.effect.CallbackStack


floating-morality:
CallbackStack leaks badly in v3.4.4
![image](https://user-images.githubusercontent.com/89297528/212270460-cbf3133f-de47-4e81-a23c-0cee7cbebf9c.png)
After downgrading to v3.4.2 the issue disappeared.
![screenshot_2023-01-13_at_8 22 15_am](https://user-images.githubusercontent.com/89297528/212270618-2da60161-4887-4d94-b36a-dd66bb560eea.png)

djspiewak:
Very concerning.

Do you have any more details on what reproduces this? A heap dump would be the most helpful thing, or some subset of code. In lieu of that, are you doing anything interesting with Deferred?

durban:
Are there long-lived `Deferred`s, which are not `complete`d, and repeatedly `get`ted and cancelled? (If I remember correctly, when cancelling a `get`, the linked list node itself is not removed; I guess this could cause problems.)

djspiewak:
Yeah this is absolutely true (regarding the linked structure not being removed). Basically it trades off some memory in the `cancel` case because it's assumed that the "long lived, lots of cancels, never completed" case is very rare.

TimWSpence:
> Yeah this is absolutely true (regarding the linked structure not being removed). Basically it trades off some memory in the `cancel` case because it's assumed that the "long lived, lots of cancels, never completed" case is very rare.

Sadly I think [Ember would like a word with you](https://gist.github.com/TimWSpence/89ca8b8904dbbd397767a1dbefaac777) there. Connecting via JMX I can see the number of `CallbackStack`s monotonically increasing there.

From a very quick glance at the Ember server code, I would guess that `shutdown.signal` is a problem at least, as it's a global `Deferred` that is `get`ted by every connection-handling fiber

TimWSpence:
> > Yeah this is absolutely true (regarding the linked structure not being removed). Basically it trades off some memory in the `cancel` case because it's assumed that the "long lived, lots of cancels, never completed" case is very rare.
> 
> Sadly I think [Ember would like a word with you](https://gist.github.com/TimWSpence/89ca8b8904dbbd397767a1dbefaac777) there. Connecting via JMX I can see the number of `CallbackStack`s monotonically increasing there.
> 
[...]

PR #3361, opened 2023-01-14
Fix `CallbackStack` leak on JS

armanbilge:
Fixes the JS version of https://github.com/typelevel/cats-effect/issues/3359, since `CallbackStack` is specialized on JS.

`clearCurrent()` is now implemented by invoking the JS `delete` operator on that array index. This transforms the `js.Array` into a sparse (aka "holey") array. This does de-optimize the data structure compared to a packed array, but hopefully this is good enough for our purposes (and better than a userland data structure).

> For real-world coding patterns, the performance difference between accessing holey or packed arrays is usually too small to matter or even be measurable.

https://v8.dev/blog/elements-kinds#avoid-creating-holes

### DIFF
--- a/core/js/src/main/scala/cats/effect/CallbackStack.scala
+++ b/core/js/src/main/scala/cats/effect/CallbackStack.scala
@@ -36,24 +36,22 @@ private final class CallbackStackOps[A](private val callbacks: js.Array[A => Uni
    * Invokes *all* non-null callbacks in the queue, starting with the current one. Returns true
    * iff *any* callbacks were invoked.
    */
-  @inline def apply(oc: A, _invoked: Boolean): Boolean = {
-    var i = callbacks.length - 1
-    var invoked = _invoked
-    while (i >= 0) {
-      val cb = callbacks(i)
-      if (cb ne null) {
-        cb(oc)
-        invoked = true
-      }
-      i -= 1
-    }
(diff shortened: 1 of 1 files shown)


djspiewak (approved):
This definitely staunches the bleeding, but I'd also be curious as to how this compares to the old `LongMap` approach, since that's effectively what a sparse array is.

armanbilge:
> but I'd also be curious as to how this compares to the old `LongMap` approach, since that's effectively what a sparse array is.

You mean this `LongMap`?

https://github.com/typelevel/cats-effect/blob/130fdcc7af75ed769f001d9f0afbff6ccf8b9488/kernel/shared/src/main/scala/cats/effect/kernel/Deferred.scala#L92-L94

That's all immutable state there, which means allocations, which means this will blast it out of the water 😛 

Edit: also note that JS doesn't natively support 64 bit ints, so anything with `Long` has lots of overhead.

armanbilge:
> This transforms the `js.Array` into a sparse (aka "holey") array. This does de-optimize the data structure compared to a packed array, but hopefully this is good enough for our purposes

Tangential, but while investigating this I discovered that Scala.js is (unintentionally) de-optimizing all uses of `js.Array` in its stdlib/javalib, so performance can't be _that_ bad.

Issue #160, opened 2018-03-16
Provide way to create an `IO[T]` from a `CompletableFuture[T]`


danielkarch:
It is sometimes necessary to work with Java libraries that return `CompletableFuture`s. It would be nice to have an easy way to turn them into `IO` values. 
This is adapted from [Monix](https://github.com/monix/monix/blob/master/monix-java/src/main/scala/monix/java8/eval/package.scala#L41):

```scala
 def fromJavaFuture[A](cf: CompletableFuture[A]): IO[A] =
    IO.cancelable(cb => {
      cf.handle[Unit](new BiFunction[A, Throwable, Unit] {
        override def apply(result: A, err: Throwable): Unit = {
          err match {
            case null =>
              cb(Right(result))
            case _: CancellationException =>
              ()
            case ex: CompletionException if ex.getCause ne null =>
              cb(Left(ex.getCause))
            case ex =>
              cb(Left(ex))
          }
        }
      })
      IO(cf.cancel(true))
    })
```

Do you think it would make sense to have this in cats-effect itself?

alexandru:
I encouraged @danielkarch to add this suggestion, as I wanted to hear other people's opinions.

I don't like the idea of adding lots of utilities to `cats-effect`, if they aren't used much, because it increases the API surface, becoming harder to maintain and with more potential for API breakage in the future.

Leaning towards no right now, as personally I haven't interacted much with `CompletableFuture` and I don't mind dumping my own utilities in my own `utils` package either. But if other people need this, then it might be wroth adding.

rossabaker:
I've implemented this as a util twice: once pre-cancellation, and once in scalaz-concurrent.  But the points about surface area are well taken.  I think I want this if and only if a few other people speak up for it.

oleg-py:
I believe it does not belong in cats-effect, whose purpose is to provide a common protocol on which Scala effect types cooperate. But I would love to see something like `cats-effect-contrib` which would provide the set of such utilities (including the one for regular java Future with polling we recently talked about on Gitter). I believe there's no use having people to reinvent those utils every time.

danielkarch:
I agree with @oleg-py. It would be nice to have an "official solution" for things like this, but `cats-effect` itself is not the right place, so perhaps a `contrib` project would be a good place for it.

alexandru:
Another project or a sub-project was what I thought of in https://github.com/typelevel/cats-effect/issues/117

[...]

PR #3374, opened 2023-01-18
Add `fromFutureCancelable` and friends

armanbilge:
This is a follow-up to:
- https://github.com/typelevel/cats-effect/pull/3205

I was working on http4s-dom and encountered roughly this:

```scala
F.fromPromise(F.delay(Fetch.fetch(...))).onCancel(F.delay(abortController.abort()))
```

The problem is that `fromPromise` is implemented via `async_` which will be uncancelable under the new semantics. Without a `fromPromiseCancelable` variant I'd have to re-implement the conversion logic or run it on another fiber, so here we are.

Along these lines I also added `fromFutureCancelable`. I did not add a variant for `CompletableFuture` because its API natively supports cancelation (go Java! :trollface:). As a plus side, these methods offer an escape hatch for anyone who is bitten by the new semantics.

The method names and signatures feel a bit unergonomic so definitely open to bikeshed. I didn't try too hard to share implementations since it would also be messy.

### DIFF
--- a/tests/shared/src/test/scala/cats/effect/IOSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/IOSpec.scala
@@ -1416,6 +1416,16 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
         }
       }
 
+      "round trip cancelable through s.c.Future" in ticked { implicit ticker =>
+        forAll { (ioa: IO[Int]) =>
+          ioa eqv IO.fromFutureCancelable(
+            IO(ioa.unsafeToFutureCancelable()).map {
+              case (fut, fin) => (fut, IO.fromFuture(IO(fin())))
+            }
+          )
+        }
+      }
+
       "canceled through s.c.Future is errored" in ticked { implicit ticker =>
         val test =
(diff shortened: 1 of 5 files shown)


tests/shared/src/test/scala/cats/effect/IOSpec.scala
```diff
@@ -1416,6 +1416,16 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
...
 
+      "round trip cancelable through s.c.Future" in ticked { implicit ticker =>
+        forAll { (ioa: IO[Int]) =>
+          ioa eqv IO.fromFutureCancelable(
+            IO(ioa.unsafeToFutureCancelable()).map {
+              case (fut, fin) => (fut, IO.fromFuture(IO(fin())))
+            }
+          )
+        }
+      }
```
armanbilge: This test is broken and I'm not sure why. Will follow-up (or would appreciate an assist :).

djspiewak (approved):
The nesting on this pretty much prevents anything with better ergonomics. I think this is probably the best we can do, unfortunately.

He-Pin:
lol:) refs: https://github.com/typelevel/cats-effect/issues/160 too

PR #3375, opened 2023-01-18
Refactor `fromCompletableFuture` to use `cont`

armanbilge:
In the previous implementation, to ensure backpressure in the case that the `CompletableFuture` could not be canceled, a second callback was registered on it. By re-implementing via `cont` we can fallback to awaiting the already installed callback.

### DIFF
--- a/kernel/jvm/src/main/scala/cats/effect/kernel/AsyncPlatform.scala
+++ b/kernel/jvm/src/main/scala/cats/effect/kernel/AsyncPlatform.scala
@@ -14,7 +14,8 @@
  * limitations under the License.
  */
 
-package cats.effect.kernel
+package cats
+package effect.kernel
 
 import java.util.concurrent.{CompletableFuture, CompletionException, CompletionStage}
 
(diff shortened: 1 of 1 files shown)


armanbilge:
I wonder if this is a common enough pattern that it deserves its own combinator (or maybe this is what `async` should have been). Specifically, an `async` where the finalizer is an `F[Boolean]`, that indicates whether it canceled, or whether to continue waiting for the completion callback before moving on.

kernel/jvm/src/main/scala/cats/effect/kernel/AsyncPlatform.scala
```diff
@@ -29,28 +30,36 @@ private[kernel] trait AsyncPlatform[F[_]] extends Serializable { this: Async[F]
...
+                      case _ => t
+                    }))
+                }
               }
-            ))
+
+              val await = G.onCancel(
+                poll(get),
+                // if cannot cancel, fallback to get
+                G.ifM(lift(delay(cf.cancel(false))))(G.unit, G.void(get))
```
djspiewak, line 55: Oh this is absolutely fascinating! It definitely works but it's 100% not a thing that we considered when writing `cont`.

djspiewak (approved):

kernel/jvm/src/main/scala/cats/effect/kernel/AsyncPlatform.scala
armanbilge: I use this same pattern to interop with io_uring.

https://github.com/armanbilge/fs2-io_uring/blob/4b676319a43bebd317d6f978b356b97915921a14/uring/src/main/scala/fs2/io/uring/Uring.scala#L60-L63

PR #3378, opened 2023-01-19
Add FAQ entry about `IO(...)` on blocking thread

armanbilge:
Since it really is an FAQ :)

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -65,7 +65,7 @@ If the `IO(...)` in the middle was run on the compute pool this would require:
 2. shifting back to the compute pool, for the non-blocking op
 3. requisitioning another blocking thread, for the second blocking op
 
-So in this case, the intermediate `IO(...)` is highly likely to run on a blocking thread. This enables the entire sequence to run on the same blocking thread, without any shifting.
+So in this case, the intermediate `IO(...)` is highly likely to run on a blocking thread. This enables the entire sequence to run on the same blocking thread, without any shifting. As soon as any asynchronous operation (or an auto-cede boundary) is encountered, the fiber will be shifted back to a compute thread. The tradeoff here is to create a small amount of unnecessary thread contention in exchange for avoiding unnecessary context shifts in many cases.
 
 Note that this is just a specific example to demonstrate why running an non-blocking `IO(...)` on a blocking thread can be beneficial, but it is not the only situation in which you may observe this behavior.
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -61,9 +61,9 @@ IO.blocking(...) *> IO(...) *> IO.blocking(...)
 
 If the `IO(...)` in the middle was run on the compute pool this would require:
 
-1. recquisitioning a blocking thread, and running the first blocking op
+1. requisitioning a blocking thread, and running the first blocking op
 2. shifting back to the compute pool, for the non-blocking op
-3. recquisitioning another blocking thread, for the second blocking op
+3. requisitioning another blocking thread, for the second blocking op
 
 So in this case, the intermediate `IO(...)` is highly likely to run on a blocking thread. This enables the entire sequence to run on the same blocking thread, without any shifting.
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -49,6 +49,24 @@ Hello world
 
 See [here](core/scala-native.md) for details.
 
+## Why is my `IO(...)` running on a blocking thread?
+
+Cats Effect guarantees that `IO.blocking(...)` will run on a blocking thread. However, in many situations it may also run a non-blocking operation `IO(...)` (or `IO.delay(...)`) on a blocking thread. Do not be alarmed! This is an optimization.
+
+Consider the following program:
+
+```scala
+IO.blocking(...) *> IO(...) *> IO.blocking(...)
+```
+
+If the `IO(...)` in the middle was run on the compute pool this would require:
+
+1. recquisitioning a blocking thread, and running the first blocking op
+2. shifting back to the compute pool, for the non-blocking op
+3. recquisitioning another blocking thread, for the second blocking op
+
+So in this case, the intermediate `IO(...)` is highly likely to run on a blocking thread. This enables the entire sequence to run on the same blocking thread, without any shifting.
+
 ## Dealing with Starvation
 
 Cats Effect 3.4.0 introduced a default-enabled *starvation checker*, which produces warnings like the following:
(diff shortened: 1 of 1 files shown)


decoursin:
In my opinion, the essential "problem" is that we (or, you, or cats-effect) doesn't have a two thread-pool system (not counting the scheduler/'non-blocking IO polling' thread-pool), as described wonderfully [here](https://gist.github.com/djspiewak/46b543800958cf61af6efa8e072bfd5c). The `WorkStealingThreadPool` is really a one thread-pool system, I believe.

From my understanding, the WorkStealingThreadPool detects when a fiber is fiber blocked and temporarily parks the underlying WorkerThread to prevent from utilizing CPU resources, until the fiber is finished blocking, then the WorkerThread is unparked. Or more precisely said: when a fiber enters a `scala.concurrent.blocking` call, it triggers a `blockOn` invocation of the WorkerThread, and it's exactly at this point that the WorkerThread parks itself.

Before the WorkerThread parks itself, it creates a new WorkerThread, because the current one is about to be blocked, parked, and unavailable for any unknown amount of time. The new WorkerThread is not connected with the one to be parked in anyway.

The WorkStealingThreadPool doesn't use the built in `Executors.newCachedThreadPool` nor the `Executors.newFixedThreadPool`, it creates and manages Threads itself.

This in my opinion is exactly how it should be, and is actually superior to the two thread-pool system described in the link above.

The essential purpose, after all, is only to make sure that the blocked Threads aren't wastefully using CPU resources and that there's a minimum number of Threads always available (which almost always equals the amount of CPUs). It doesn't matter if the about-to-be-blocked fiber runs on a dedicated Blocking Thread Pool, only as long as it doesn't block the underlying the Thread.

That's why I think the naming distinction of `io-compute-idx` vs `io-compute-blocker-idx` here is rather superfluous, since the fiber doesn't (or isn't) actually running on either a Compute Thread Pool, or a Blocker Thread Pool, it's only running in one thread pool.

I believe, after looking into the code, that it _is_ possible to make a naming distinction for when the Fiber is or is not running inside of a `scala.concurrent.blocking` chunk of code, this though would have to be initiated within `IOFiber.scala`, not within the `WorkerThread.scala`. (That's necessary in order to determine when during the Fiber Run Loop is inside a `scala.concurrent.blocking` chunk or has just left one. The `WorkerThread.scala` cannot determine when a fiber has left a `scala.concurrent.blocking` chunk, because it doesn't have visibility into the details of the fiber execution.)

However, as I previously just alluded to, that doesn't make sense anyways, and I would instead recommend completely getting rid of this distinction, and renaming `io-compute-idx` to something that doesn't mislead, like just `io-idx` or `io-pool-idx`; and updating the documentation in many places, like [here](https://typelevel.org/cats-effect/docs/thread-model), of the new Thread Model.

armanbilge:
> temporarily parks the underlying WorkerThread to prevent from utilizing CPU resources, until the fiber is finished blocking, then the WorkerThread is unparked

Well, the blocking task itself "parks" the thread, so that it is not utilizing CPU resources. When it is finished blocking, it will continue running the fiber up to the next `IO.async` or `IO.cede`, at which point it will be rescheduled back on an ordinary compute thread.

_Then_ this `WorkerThread` will park itself. It will either be unparked and revived, to replace a compute thread when a new blocking task starts. Or if that doesn't happen within some time (default 1 minute) it will be shutdown and cleaned up.

---

> However, as I previously just alluded to, that doesn't make sense anyways, and I would instead recommend completely getting rid of this distinction

I think the distinction is still important, since these are two different "modes" of the thread, which is important diagnostic information.

For example, suppose you are looking a thread dump, and you see that a `io-pool-idx` thread is blocked on some task. Without knowing if it is a "compute mode" or "blocking mode" thread, how do you know if you didn't accidentally do `IO.delay(blockingTask())` instead of `IO.blocking(blockingTask())`?

---

> doesn't have a two thread-pool system (not counting the scheduler/'non-blocking IO polling' thread-pool)

those threadpools are going to be absorbed into the `WorkStealingThreadPool` as well, see https://github.com/typelevel/cats-effect/discussions/3070 :)

decoursin:
Hi @armanbilge thanks so much for the corrections and additional information - I will read that soon 

> For example, suppose you are looking a thread dump, and you see that a io-pool-idx thread is blocked on some task. Without knowing if it is a "compute mode" or "blocking mode" thread, how do you know if you didn't accidentally do IO.delay(blockingTask()) instead of IO.blocking(blockingTask())?

Good point.

In my opinion, then, you have to do something at [L960 in IOFiber](https://github.com/typelevel/cats-effect/blob/series/3.x/core/shared/src/main/scala/cats/effect/IOFiber.scala#L960) like: ```Thread.currentThread().asInstanceOf[WorkerThread].resetNameOutOfBlocking()```

docs/faq.md
```diff
@@ -49,6 +49,26 @@ Hello world
...
+
+```scala
+IO.blocking(...) *> IO(...) *> IO.blocking(...)
+```
+
+If the `IO(...)` in the middle was run on the compute pool this would require:
+
+1. recquisitioning a blocking thread, and running the first blocking op
+2. shifting back to the compute pool, for the non-blocking op
+3. recquisitioning another blocking thread, for the second blocking op
```
mox692: `recquisitioning` -> `requisitioning` ?

djspiewak:
> This in my opinion is exactly how it should be, and is actually superior to the two thread-pool system described in the link above.

Quickly replying to this point… I completely agree. :-) The gist you linked is correct in so far as it goes, but it lacks ambition. In particular, I never realized we could collapse event dispatchers into the compute pool (as libuv does). You still end up with two separate *semantic* pools, with the second being the blocking workers, but the whole thing can be managed by a single scheduler.

docs/faq.md
```diff
@@ -49,6 +49,26 @@ Hello world
...
+IO.blocking(...) *> IO(...) *> IO.blocking(...)
+```
+
+If the `IO(...)` in the middle was run on the compute pool this would require:
+
+1. requisitioning a blocking thread, and running the first blocking op
+2. shifting back to the compute pool, for the non-blocking op
+3. requisitioning another blocking thread, for the second blocking op
+
+So in this case, the intermediate `IO(...)` is highly likely to run on a blocking thread. This enables the entire sequence to run on the same blocking thread, without any shifting.
```
djspiewak:
```suggestion
So in this case, the intermediate `IO(...)` is highly likely to run on a blocking thread. This enables the entire sequence to run on the same blocking thread, without any shifting. As soon as any asynchronous operation (or an auto-cede boundary) is encountered, the fiber will be shifted back to a compute thread. The tradeoff here is to create a small amount of unnecessary thread contention in exchange for avoiding unnecessary context shifts in many cases.
```

djspiewak (dismissed):
Minor suggestion to clarify that we will very quickly get off the blocking thread. Otherwise LGTM!

decoursin:
By the way, thanks very much for cats 3, it's awesome 🤩

PR #3379, opened 2023-01-19
Fixed example code for CPU starvation checking in documentation.

baber:
**What**
Fixed a bug in the example code that shows how CPU starvation checking can be implemented.

**Why**
To ensure the documentation is correct.

**Details**
The existing code will always conclude CPU starvation has occurred and print out a warning message because the `delta` value is assumed to be "the time between requesting CPU access and actually getting it" but it is actually the **sum** of the sleep interval and "the time between requesting CPU access and actually getting it".

### DIFF
--- a/docs/core/starvation-and-tuning.md
+++ b/docs/core/starvation-and-tuning.md
@@ -49,20 +49,20 @@ Before diving into how we can correct issues with CPU starvation, it is worth un
 ```scala mdoc:silent
 import scala.concurrent.duration._
 
-val Threshold = 0.1d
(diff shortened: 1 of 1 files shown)


armanbilge:
Thanks! Would you mind re-targeting your PR at `series/3.4.x` branch? So we can publish the fixes sooner.

https://github.com/typelevel/cats-effect/blob/series/3.x/CONTRIBUTING.md#process

armanbilge:
superseded by https://github.com/typelevel/cats-effect/pull/3380

PR #3380, opened 2023-01-19
Fixed example code for CPU starvation checking in docs.

baber:
**What**
Fixed a bug in the example code that shows how CPU starvation checking can be implemented.

**Why**
To ensure the documentation is correct.

**Details**
The existing code will always conclude CPU starvation has occurred and print out a warning message because the `delta` value is assumed to be "the time between requesting CPU access and actually getting it" but it is actually the **sum** of the sleep interval and "the time between requesting CPU access and actually getting it".


armanbilge (approved):
Thank you so much!

PR #3384, opened 2023-01-26
Starvation clock timestamp

### LINKED ISSUE #3308: print timestamp for CPU starvation warning in System.err
Please print timestamp for CPU starvation warning in System.err in addition to warning.

With current implementation we can observe few such warnings, but it is impossible to correlate the warning with application's log.

mtomko:
This pull request is intended to address https://github.com/typelevel/cats-effect/issues/3308

I have updated the relevant documentation with the new expected message and a brief clarification of the meaning of the timestamp. Tests are adjusted slightly, but due to the myriad of corner cases time introduces, my assertions are not as precise as I'd want in other cases.


mtomko:
Okay, I see now where to hook my tests in 😛 

I have not done anything like this with cats-effect, but I seem to recall there's a way to have some control over the time measured by the app, so maybe the assertion can be simple. Otherwise, I can adjust the assertion such that it either skips over the timestamp or matches against it somehow.

core/js/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -167,6 +168,14 @@ trait IOApp {
    */
   protected def runtimeConfig: unsafe.IORuntimeConfig = unsafe.IORuntimeConfig()
 
+  /**
+   * The formatter used to display timestamps when CPU starvation is detected. It is defined
+   * here to allow various target platforms (e.g., JVM, JS) to utilize a convenient class.
+   */
+  private def finiteDurationShow: Show[FiniteDuration] = Show.show[FiniteDuration] { fd =>
+    new js.Date(fd.toMillis.toDouble).toString
+  }
```
armanbilge: Nice! Is there a particular reason we need to define this in `IOApp`? I was thinking we can introduce a `trait CpuStarvationCheckPlatform` where we define the platform-specific methods.

Also in that case, I feel like a `def formatFiniteDuration(fd: FiniteDuration): String` would be fine.

armanbilge (commented):
We have an existing test harness here, we just need to update the assertion to verify the presence of a timestamp :)
https://github.com/typelevel/cats-effect/blob/9f7abed314cbfd83728a195219ae666beb4e09bd/tests/jvm/src/test/scala/cats/effect/IOAppSpec.scala#L198-L204

armanbilge:
> Okay, I see now where to hook my tests in 😛

Oh, I see, CI shed the light for you there 😂

core/js/src/main/scala/cats/effect/IOApp.scala
mtomko: I'm looking at adding the `trait` and having it use a self type for `CpuStarvationCheck`, using the example of `IOFiberPlatform` and `IOFiber[A]`, but for some reason scalac doesn't like it:

```scala
core/js/src/main/scala/cats/effect/CpuStarvationCheckPlatform.scala:22:67
[E]      not found: type CpuStarvationCheck
[E]      L22: private[effect] abstract class CpuStarvationCheckPlatform { this: CpuStarvationCheck =>
```

The pattern followed by `IOFiber` et. al suggests that scalac can do multiple passes to resolve the circularity - ie, the platform has `IOFiber` as a self type, but `IOFiber` is defined in shared and the platform is defined only in the various platform-specific directories (actually, it's not in all of them...). 

I chose `IOFiber` as a pattern because it is a class not an object, and most of the things that extend a `*Platform` are objects.

I'm a bit out of my element, but it seems like my definition should either work as-is, or neither it nor `IOFiber` should work. I'll take a fresh look at it later, I guess.

armanbilge: Try this:
```scala
this: CpuStarvationCheck.type =>
```
Because `CpuStarvationCheck` is an `object`.

mtomko: 🤦🏼 thank you

mtomko:
I have tests passing now. I will double check whether there is any documentation that should change and see if there is a reasonable way to make some assertions about the printed dates, then remove the draft tag.

I can rebase before that time if you would prefer to have the messy early commits removed, or leave them in place as history.

armanbilge:
> I can rebase before that time if you would prefer to have the messy early commits removed

Ah, good that you mention rebasing: I forgot to ask that you target your PR at series/3.4.x, that way your changes be included in the next patch release. So if you don't mind rebasing onto that, and then re-targeting your PR. Thanks!

Regarding "messy" commits, we like to keep the history here :) so no need to clean that up. Thanks for asking!

Regarding docs, I believe we have it in a couple places e.g. here
https://github.com/typelevel/cats-effect/blob/9f7abed314cbfd83728a195219ae666beb4e09bd/docs/faq.md?plain=1#L56-L61

mtomko:
Apparently this kind of rebasing is beyond my ken. Working on it.

mtomko:
I have cherry picked my commits over to another branch with the correct base branch. If I can't figure out how to fix this PR, I'll just make a fresh one with my changes.

mtomko:
Looks like that last check just timed out fetching a JVM.

core/native/src/main/scala/cats/effect/CpuStarvationCheckPlatform.scala
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package cats.effect
+
+import scala.concurrent.duration.FiniteDuration
+
+import java.time.Instant
+import java.time.format.DateTimeFormatter
+
+private[effect] abstract class CpuStarvationCheckPlatform { this: CpuStarvationCheck.type =>
```
armanbilge: Instead of duplicating this file between JVM/Native, we can place one copy in the `jvm-native` source folder.

docs/core/starvation-and-tuning.md
```diff
@@ -3,12 +3,14 @@ id: starvation-and-tuning
 title: Starvation and Tuning
 ---
 
-All Cats Effect applications constructed via `IOApp` have an automatic mechanism which periodically checks to see if the application runtime is starving for compute resources. If you ever see warnings which look like the following, they are the result of this mechanism automatically detecting that the responsiveness of your application runtime is below the configured threshold.
+All Cats Effect applications constructed via `IOApp` have an automatic mechanism which periodically checks to see if the application runtime is starving for compute resources. If you ever see warnings which look like the following, they are the result of this mechanism automatically detecting that the responsiveness of your application runtime is below the configured threshold. Note that the timestamp is the time when the starvation was detected, which is not precisely the time when starvation (or the task that is responsible) began.
```
armanbilge, line 6: 👍

armanbilge (commented):
Overall LGTM, just a minor comment. Thanks for all of your work on this, including the doc updates!

armanbilge (approved):

djspiewak (approved):
Thanks for taking this on!

PR #3387, opened 2023-01-27
Thread blocking detection

TimWSpence:
Attempt to detect erroneous use of `IO.apply` or `IO.delay` to suspend blocking operations, instead of using `IO.blocking` or `IO.interruptible`.

Addresses https://github.com/typelevel/cats-effect/issues/3288.


core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -370,6 +370,23 @@ private final class WorkerThread(
...
+            // TODO prefetch pool.workerThread or Thread.State.BLOCKED ?
+            // TODO check that branch elimination makes it free when off
+            var otherIdx = random.nextInt(pool.workerThreads.length)
+            while (otherIdx == idx) {
+              otherIdx = random.nextInt(pool.workerThreads.length)
+            }
+            val thread = pool.workerThreads(otherIdx)
+            val state = thread.getState()
+            val parked = thread.parked.get()
+            if (!parked && (state == Thread.State.BLOCKED || state == Thread
```
TimWSpence, line 431: This is the most insane scalafmt I've ever seen 😂

build.sbt
```diff
@@ -507,6 +507,7 @@ lazy val core = crossProject(JSPlatform, JVMPlatform, NativePlatform)
       // introduced by #2256, Hide the package private constructor for `IORuntime`
       // changes to `cats.effect.unsafe` package private code
       ProblemFilters.exclude[DirectMissingMethodProblem]("cats.effect.unsafe.IORuntime.this"),
+      ProblemFilters.exclude[DirectMissingMethodProblem]("cats.effect.unsafe.IORuntime.createWorkStealingComputeThreadPool"),
```
armanbilge: This exclusion is not safe, it's a real bincompat issue :)

TimWSpence: Haha yeah I did wonder about that. What's the right way to handle this again? I tried adding a new method with the extra flag and having the old one delegate to that but I got errors about multiple methods with default arguments

armanbilge: > having the old one delegate to that but I got errors about multiple methods with default arguments

You did it right, just remove the default arguments from the old, deprecated one :)

TimWSpence: 🤦 thanks!

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
```diff
@@ -33,8 +33,7 @@ private[unsafe] abstract class IORuntimeCompanionPlatform { this: IORuntime.type
 
   private[this] final val DefaultBlockerPrefix = "io-compute-blocker"
 
-  @deprecated("Preserved for binary-compatibility", "3.4.6")
-  private[effect] def createWorkStealingComputeThreadPool(
+  def createWorkStealingComputeThreadPool(
```
armanbilge: I think we should deprecated it, but keep it public. Also, should be deprecated since 3.5.0.

tests/js/src/main/scala/catseffect/examplesplatform.scala
```diff
@@ -115,4 +117,10 @@ package examples {
     def run(args: List[String]): IO[ExitCode] = IO.never
   }
 
+  // stub
+  object BlockedThreads extends IOApp.Simple {
+    val run = Console[IO].errorln(
+      "[WARNING] A Cats Effect worker thread was detected to be in a blocked state")
+  }
```
armanbilge: I've been defining the stubs as `IO.never` (see examples above), and just not running them for JS, since this test doesn't make sense on JS anyway.

TimWSpence: Ah that's why those tests said `IO.never` 🤦 Yeah that's much better 👍

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
TimWSpence: Makes sense 👍 Is `3.5.0` the right version to put? And while we're on the topic, is `series/3.x` the right branch to target?

armanbilge: Yes, we never add new APIs in a minor branch e.g. 3.4.x, only bug fixes.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -415,6 +415,23 @@ private final class WorkerThread(
 
       ((state & ExternalQueueTicksMask): @switch) match {
         case 0 =>
+          if (pool.blockedThreadDetectionEnabled) {
+            // TODO prefetch pool.workerThread or Thread.State.BLOCKED ?
+            // TODO check that branch elimination makes it free when off
+            var otherIdx = random.nextInt(pool.workerThreads.length)
+            while (otherIdx == idx) {
```
durban: If the pool size is 1, this is an infinite loop.

TimWSpence: Good point!

core/jvm/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -239,6 +239,20 @@ trait IOApp {
...
+   * production.
+   *
+   * If enabled, the compute pool will attempt to detect when blocking operations have been
+   * erroneously wrapped in `IO.apply` or `IO.delay` instead of `IO.blocking` or
+   * `IO.interruptible` and will report stacktraces of this to stderr.
+   *
+   * This may be of interest if you've been getting warnings about CPU starvation printed to
+   * stderr. [[https://typelevel.org/cats-effect/docs/core/starvation-and-tuning]]
+   */
+  protected def blockedThreadDetectionEnabled: Boolean = false
```
djspiewak, line 254: IMO we should *also* read this from a system property, similar to a lot of the other runtime configuration parameters

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -64,7 +64,8 @@ private[effect] final class WorkStealingThreadPool(
     private[unsafe] val threadPrefix: String, // prefix for the name of worker threads
     private[unsafe] val blockerThreadPrefix: String, // prefix for the name of worker threads currently in a blocking region
     private[unsafe] val runtimeBlockingExpiration: Duration,
-    reportFailure0: Throwable => Unit
+    reportFailure0: Throwable => Unit,
+    private[unsafe] val blockedThreadDetectionEnabled: Boolean
```
djspiewak: Just aesthetically, I kinda prefer `reportFailure` to be last, but that's just me.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -74,7 +75,7 @@ private[effect] final class WorkStealingThreadPool(
   /**
    * References to worker threads and their local queues.
    */
-  private[this] val workerThreads: Array[WorkerThread] = new Array(threadCount)
+  private[unsafe] val workerThreads: Array[WorkerThread] = new Array(threadCount)
```
djspiewak: This has performance implications. Why was the visibility changed?

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -652,6 +669,26 @@ private final class WorkerThread(
...
+          sb.append("\n")
+        i += 1
+      }
+      sb.toString()
+    }
+    s"""|[WARNING] A Cats Effect worker thread was detected to be in a blocked state ($state)
+        |${formatTrace(stackTrace)}
+        |This is very likely to be due to suspending a blocking call in IO via
+        |`IO.delay` or `IO.apply`. If this is the case then you should use
+        |`IO.blocking` or `IO.interruptible` instead.""".stripMargin
```
djspiewak, line 692: Interestingly, in a follow-up PR, we can probably additionally detect if the blocking is coming from an `apply`/`delay` and actually point users at the exact line of code that needs to change.

tests/jvm/src/main/scala/catseffect/examplesplatform.scala
```diff
@@ -89,4 +89,16 @@ package examples {
...
+  object BlockedThreads extends IOApp.Simple {
+
+    override protected def blockedThreadDetectionEnabled = true
+
+    // Loop prevents other worker threads from being parked and hence not
+    // performing the blocked check
+    lazy val loop: IO[Unit] = IO.unit >> loop
+
+    val run =
+      loop.start >> IO(Thread.sleep(2.seconds.toMillis))
```
djspiewak: ⛳ 

```suggestion
      IO.unit.foreverM.start >> IO(Thread.sleep(2.seconds.toMillis))
```

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -415,6 +415,23 @@ private final class WorkerThread(
 
       ((state & ExternalQueueTicksMask): @switch) match {
         case 0 =>
+          if (pool.blockedThreadDetectionEnabled && pool.workerThreads.length > 1) {
+            // TODO prefetch pool.workerThread or Thread.State.BLOCKED ?
+            // TODO check that branch elimination makes it free when off
+            var otherIdx = random.nextInt(pool.workerThreads.length)
+            while (otherIdx == idx) {
+              otherIdx = random.nextInt(pool.workerThreads.length)
```
djspiewak: Rather than looping, can we go with a more constant-time approach here? Something like:

```suggestion
              otherIdx = (idx + Math.max(1, random.nextInt(pool.workerThreads.length - 1))) % pool.workerThreads.length
```

This would guarantee that `otherIdx` always ends up *not* on `idx` on our second try, so long as there is more than one thread.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -415,6 +415,23 @@ private final class WorkerThread(
 
       ((state & ExternalQueueTicksMask): @switch) match {
         case 0 =>
+          if (pool.blockedThreadDetectionEnabled && pool.workerThreads.length > 1) {
```
djspiewak: We should simplify this branch prediction by simply checking the number of worker threads on pool init and overriding `blockedThreadDetectionEnabled` to `false` if it's `1`.

djspiewak (changes requested):

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
TimWSpence: It was changed so that the worker threads could get a handle to the other threads to check the thread state. What are the implications sorry? Are they mitigated by having a `private[effect] def workerThread(idx: Int)` method or something instead?

On a related performance note, I wasn't sure what the logic is for what is worth pre-fetching? The worker thread currently accesses a few things via the reference to the pool and I wasn't sure whether to pre-fetch any of them?

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
TimWSpence: Good point! Should I do that as part of this change?

tests/jvm/src/main/scala/catseffect/examplesplatform.scala
TimWSpence: I'm thoroughly ashamed of myself 😂

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
TimWSpence: Nice! Is that sufficiently more expensive that we don't want to do this on the first attempt?

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
armanbilge: > What are the implications sorry? Are they mitigated by having a `private[effect] def workerThread(idx: Int)` method or something instead?

Yup, see bytecode comparison in https://github.com/typelevel/cats/pull/4314#discussion_r988196572.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
djspiewak: I just worry about pathological cases that are hard to track down. I'd rather be constant time if at all possible. If we don't want to do the fancy modular arithmetic thing, IMO the next-best thing would be to just turn the `while` into an `if` above and, if we happen to randomly select our own index, just bail out and get it next time.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
djspiewak: Let's do it in a follow-up so we don't block this. :-)

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
djspiewak: > What are the implications sorry?

The tldr is that bytecode-`private` is more efficient than bytecode-`public` because of the effect it has on the generated assembly (e.g. `private` methods can't be overridden, so it's a static dispatch rather than a dynamic one). Scala takes this a step further because `private[this] val`/`var` turns into a *field* access in the bytecode, meaning you avoid the `invoke`, which is even faster still.

> On a related performance note, I wasn't sure what the logic is for what is worth pre-fetching? The worker thread currently accesses a few things via the reference to the pool and I wasn't sure whether to pre-fetch any of them?

Pre-fetching is usually not really worth it unless the value is mutable and volatile (or other volatile access is expected within the current scope). We actually do it more than we need to throughout Cats Effect. There's a bit of a misconception about JVM performance that locals are somehow faster than fields, but when the fields are private and non-volatile and there are no *other* memory fences floating around in the method, then a local and a field both turn into a register in the assembly (in other words, HotSpot figures this out and does the pre-fetching for you), so it ends up being the exact same thing.

TimWSpence: Awesome, many thanks both as always ❤️ This week is quite busy in work but I'll try to turn this around ASAP

core/jvm/src/main/scala/cats/effect/IOApp.scala
TimWSpence: Thus far, the system properties all feed into `IORuntimeConfig` which is shared across all platforms but this is only relevant for the JVM. Is that OK?

djspiewak (approved):

core/jvm/src/main/scala/cats/effect/IOApp.scala
armanbilge: Still, even if its configured here, we can read it from a system property. That way it can be enabled without recompiling an app?

Yeah, sorry, a bit late to the party 😝 

Also this will be relevant to Native in the near future.

PR #3388, opened 2023-01-29
Protect timers against Long overflow

durban:
The `System.nanoTime` docs say this:

> To compare two nanoTime values ... one should use `t1 - t0 < 0`, not `t1 < t0`, because of the possibility of numerical overflow.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/SleepCallback.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/SleepCallback.scala
@@ -75,5 +75,5 @@ private object SleepCallback {
   }
 
   implicit val sleepCallbackReverseOrdering: Ordering[SleepCallback] =
-    Ordering.fromLessThan(_.triggerTime > _.triggerTime)
+    Ordering.fromLessThan(_.triggerTime - _.triggerTime > 0)
 }
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
@@ -401,7 +401,7 @@ private final class WorkerThread(
         while (cont) {
           val head = sleepers.head()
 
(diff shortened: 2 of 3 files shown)


armanbilge:
Dumb question: how can `t1 < t0` overflow? I don't get it 😅 

Edit: is it basically saying we should treat `nanoTime()` as returning an unsigned long value? In which case, shouldn't we prefer `Long.compareUnsigned`?

durban:
If I'm reading the docs correctly, `nanoTime` itself is allowed to overflow. For example:

1. The JVM starts up; `nanoTime == Long.MAX - 100L` (it can start from an arbitrary value).
2. Quite quickly we do `val t1 = System.nanoTime()`; let's say `t1 == Long.MAX - 50L`.
3. 100ns elapses, and then we do `val t2 = System.nanoTime()`; so now `t2 == Long.MIN + 49L` due to overflow.

So if we do `t1 < t2`, that will be `false`, even though `t1` is from an _earlier_ invocation of `nanoTime`. This is bad.

However, if we do `t1 - t2 < 0`, that will be `true`; correctly.

(For "normal" values, the two things are the same.)

durban:
@armanbilge Regarding your edit: I'd have to think about it, but `compareUnsigned` would probably be also correct.

armanbilge:
> I'd have to think about it, but `compareUnsigned` would probably be also correct.

Although, the implementation is different.

https://github.com/openjdk/jdk/blob/af564e46b006fcd57ec7391cd1438b3b9311b1d6/src/java.base/share/classes/java/lang/Long.java#L1644-L1647

djspiewak (approved):
Nice catch! That would have been an absolutely horrifying bug to track down in the wild.

durban:
FTR: It seems, that `compareUnsigned` would not be the same. A counterexample is `0L - (-1000L) == 1000L > 0L` but `java.lang.Long.compareUnsigned(0L, -1000L) < 0L`.

PR #3389, opened 2023-01-29
Handle if async callback receives null

durban:
If an async callback is called with `null`, instead of hanging, behave as if it was called with `Left(new NullPointerException)`.

This is a change in behavior. But I would be surprised, if anybody would prefer the current behavior.

See also #1178.

### DIFF
--- a/tests/shared/src/test/scala/cats/effect/IOSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/IOSpec.scala
@@ -514,10 +514,11 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
           latch <- Deferred[IO, Unit]
           fib <- IO.async[Int] { cb => cbp.complete(cb) *> latch.get.as(None) }.start
           cb <- cbp.get
-          (_, r) <- IO.both(
+          _r <- IO.both(
             latch.complete(()) *> IO.sleep(0.1.second) *> IO(cb(null)),
             fib.joinWithNever.attempt
           )
+          (_, r) = _r
           _ <- IO(r must beLeft(beAnInstanceOf[NullPointerException]))
         } yield ok
       }
--- a/tests/shared/src/test/scala/cats/effect/IOSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/IOSpec.scala
@@ -319,12 +319,12 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
         ioa must completeAs(44)
       }
 
-      "result in an null if lifting a pure null value" in ticked { implicit ticker =>
+      "result in a null if lifting a pure null value" in ticked { implicit ticker =>
         // convoluted in order to avoid scalac warnings
         IO.pure(null).map(_.asInstanceOf[Any]).map(_ == null) must completeAs(true)
       }
 
-      "result in an NPE if delaying a null value" in ticked { implicit ticker =>
+      "result in a null if delaying a null value" in ticked { implicit ticker =>
         IO(null).map(_.asInstanceOf[Any]).map(_ == null) must completeAs(true)
         IO.delay(null).map(_.asInstanceOf[Any]).map(_ == null) must completeAs(true)
       }
@@ -334,7 +334,6 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
           .attempt
           .map(_.left.toOption.get.isInstanceOf[NullPointerException]) must completeAs(true)
       }
-
     }
 
     "fibers" should {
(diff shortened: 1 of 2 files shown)


tests/shared/src/test/scala/cats/effect/IOSpec.scala
```diff
@@ -319,12 +319,12 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
...
       }
 
-      "result in an null if lifting a pure null value" in ticked { implicit ticker =>
+      "result in a null if lifting a pure null value" in ticked { implicit ticker =>
         // convoluted in order to avoid scalac warnings
         IO.pure(null).map(_.asInstanceOf[Any]).map(_ == null) must completeAs(true)
       }
 
-      "result in an NPE if delaying a null value" in ticked { implicit ticker =>
+      "result in a null if delaying a null value" in ticked { implicit ticker =>
```
djspiewak, line 327: Amusing that the test name incorrectly described the behavior.

djspiewak (commented):
I'm a little confused. Why would we do this? Threading `null` through an asynchronous callback is very valid behavior, and actually something that happens reasonably often when interacting with external asynchronous APIs.

djspiewak:
Also I'm a huge fan of my entirely unjustified statement in that linked issue:

> I haven't tried it, but I suspect that it hangs forever. **It should be a `NullPointerException`.**

Dear me-of-three-years-ago, why did you expect an NPE?

durban:
> Threading null through an asynchronous callback is very valid behavior, and actually something that happens reasonably often when interacting with external asynchronous APIs.

But... it doesn't work: if I do `cb(null)`, the `async` is never completed (and there is an unhandled exception in `IOFiber`). (Without this PR.)

To be clear, I'm _not_ talking about `cb(Right(null))`; that works fine (as it should).

djspiewak:
Oooooooh! Okay I missed that. Yeah alright agreed. :-)

tests/shared/src/test/scala/cats/effect/IOSpec.scala
```diff
@@ -477,6 +476,52 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
         test.attempt.flatMap { n => IO(n mustEqual Right(42)) }
       }
 
+      "calling async callback with null during registration (ticked)" in ticked {
+        implicit ticker =>
+          IO.async[Int] { cb => IO(cb(null)).as(None) }
+            .map(_ + 1)
+            .attempt
+            .map(_.left.toOption.get.isInstanceOf[NullPointerException]) must completeAs(true)
```
djspiewak:
```suggestion
            .map(_ must beLeft(beAnInstanceOf[NullPointerException])) must completeAs(ok)
```

tests/shared/src/test/scala/cats/effect/IOSpec.scala
```diff
@@ -477,6 +476,52 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
...
+            cbp <- Deferred[IO, Either[Throwable, Int] => Unit]
+            fib <- IO.async[Int] { cb => cbp.complete(cb).as(None) }.start
+            _ <- IO(ticker.ctx.tickAll())
+            cb <- cbp.get
+            _ <- IO(ticker.ctx.tickAll())
+            _ <- IO(cb(null))
+            e <- fib.joinWithNever.attempt
+          } yield e.left.toOption.get.isInstanceOf[NullPointerException]
+
+          test must completeAs(true)
```
djspiewak:
```suggestion
          } yield e must beLeft(beAnInstanceOf[NullPointerException])

          test must completeAs(ok)
```

tests/shared/src/test/scala/cats/effect/IOSpec.scala
```diff
@@ -477,6 +476,52 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
...
+      }
+
+      "calling async callback with null during registration (real)" in real {
+        IO.async[Int] { cb => IO(cb(null)).as(None) }
+          .map(_ + 1)
+          .attempt
+          .map(_.left.toOption.get.isInstanceOf[NullPointerException])
+          .flatMap { r => IO(r must beTrue) }
```
djspiewak:
```suggestion
          .flatMap(e => IO(e must beLeft(beAnInstanceOf[NullPointerException])))
```

tests/shared/src/test/scala/cats/effect/IOSpec.scala
```diff
@@ -477,6 +476,52 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
...
+          latch <- Deferred[IO, Unit]
+          fib <- IO.async[Int] { cb => cbp.complete(cb) *> latch.get.as(None) }.start
+          cb <- cbp.get
+          r <- IO.both(
+            latch.complete(()) *> IO.sleep(0.1.second) *> IO(cb(null)),
+            fib.joinWithNever.attempt
+          )
+        } yield r._2.left.toOption.get.isInstanceOf[NullPointerException]
+
+        test.flatMap { r => IO(r must beTrue) }
```
djspiewak:
```suggestion
        for {
          cbp <- Deferred[IO, Either[Throwable, Int] => Unit]
          latch <- Deferred[IO, Unit]
          fib <- IO.async[Int] { cb => cbp.complete(cb) *> latch.get.as(None) }.start
          cb <- cbp.get
          (_, r) <- IO.both(
            latch.complete(()) *> IO.sleep(0.1.second) *> IO(cb(null)),
            fib.joinWithNever.attempt)

          _ <- IO(r must beLeft(beAnInstanceOf[NullPointerException]))
        } yield ok
```

djspiewak (changes requested):

tests/shared/src/test/scala/cats/effect/IOSpec.scala
durban: `completeAs(ok)` is a compile error, but I tried to use the `beLeft(beAnInstanceOf...)` thing.

tests/shared/src/test/scala/cats/effect/IOSpec.scala
durban: `completeAs(ok)` is a compile error, but I tried to use the `beLeft(beAnInstanceOf...)` thing.

durban:
Ok, I've fixed the compile error: the `(_, r) <- ...` wasn't working in dotty. Actually, I'm not sure how did it work in Scala 2...

djspiewak:
> Actually, I'm not sure how did it work in Scala 2...

Oh right better monadic for… I was actually expecting Scala 3 to do better on that front.

djspiewak (approved):

armanbilge:
>  I was actually expecting Scala 3 to do better on that front.

You need `-source:future`, which we can't do because cross-compiling.

PR #3393, opened 2023-02-01
Simplify ContState

durban:
##### This is the commit message:

Instead of having 4 separate states (initial, waiting, winner and result) now we only have 3: initial, waiting and result. Instead of an `AtomicInteger` storing tags, now we use an `AtomicReference` for the result, and 2 sentinels: `null` for "initial", and a private object for "waiting".

This allows us to publish the result with a single CAS (instead of a CAS and a volatile write). We can also get rid of the spinning in `Get` (since we don't have to wait for the transition between "winner" and "result").

There should be no observable change in behavior (except maybe a small performance improvement).

##### Other comments:

I'm not sure about this, but I already did the work, so here it is.

What I think is good: somewhat simpler code; no spinning  in `Get`.

But the "waiting" sentinel is not exactly pretty...

##### Performance:

I've run 2 seemingly relevant benchmarks in `AsyncBenchmark`, here are the results. The difference is really small (like 1.5%), so I'm not sure if it's real. But if it's real, then it's slightly faster :-)

```
1263a9017 (series/3.4.x):

[info] Benchmark                  (size)   Mode  Cnt      Score     Error  Units
[info] AsyncBenchmark.async          100  thrpt   25  47030.178 ± 252.288  ops/s
[info] AsyncBenchmark.cancelable     100  thrpt   25  42272.219 ± 174.570  ops/s


f8c656413 (this PR):

[info] Benchmark                  (size)   Mode  Cnt      Score     Error  Units
[info] AsyncBenchmark.async          100  thrpt   25  47769.803 ± 236.728  ops/s
[info] AsyncBenchmark.cancelable     100  thrpt   25  42964.563 ± 122.446  ops/s
```

### DIFF
--- a/core/shared/src/main/scala/cats/effect/ContState.scala
+++ b/core/shared/src/main/scala/cats/effect/ContState.scala
@@ -42,7 +42,7 @@ private final class ContState(var wasFinalizing: Boolean)
   var handle: WeakBag.Handle = _
 }
 
-private final object ContState {
+private object ContState {
 
   /**
    * This is a sentinel, signifying a "waiting" state of `ContState`; only its identity is
--- a/core/shared/src/main/scala/cats/effect/ContState.scala
+++ b/core/shared/src/main/scala/cats/effect/ContState.scala
@@ -18,11 +18,37 @@ package cats.effect
 
 import cats.effect.unsafe.WeakBag
 
-import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.atomic.AtomicReference
 
-// `result` is published by a volatile store on the atomic integer extended by this class.
-// `wasFinalizing` and `handle` are published in terms of the `suspended` atomic variable in `IOFiber`
-private final class ContState(var wasFinalizing: Boolean) extends AtomicInteger(0) {
-  var result: Either[Throwable, Any] = _
+/**
+ * Possible states (held in the `AtomicReference`):
+ *   - "initial": `get() == null`
+ *   - "waiting": `get() == waiting` (sentinel)
+ *   - "result": `get()` is the result from the callback
+ *
+ * Note: A `null` result is forbidden, so "initial" is always different from "result". Also,
+ * `waitingSentinel` is private, so that can also be differentiated from a result (by object
+ * identity).
+ *
+ * `wasFinalizing` and `handle` are published in terms of the `suspended` atomic variable in
+ * `IOFiber`
+ */
+private final class ContState(var wasFinalizing: Boolean)
+    extends AtomicReference[Either[Throwable, Any]] {
+
+  val waiting: Either[Throwable, Any] =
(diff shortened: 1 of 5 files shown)


durban:
Huh... `ContState` is private, but mima says I can't change it...

armanbilge:
> mima says I can't change it...

MiMa is wrong sometimes :) you can add an exclusion.

durban:
Thanks... I literally have no idea what I'm doing. I mean, I'd understand if it were public, but it's private...

armanbilge:
It crops up often enough that we should probably file a bug report in MiMa or even better fix it. Would save a lot of time in the long run 😂

durban:
ContState is public in the bytecode... :man_facepalming:

armanbilge:
Yes, unfortunately many Scala-level private things end up public in bytecode. Sometimes MiMa can recognize that, but not always. That's why we have such a big list of exclusions 😅

djspiewak (approved):
Okay this is *definitely* nicer. :-) I think we were really fixated on using ints for the state machine that we didn't consider hiding things in `null` with an `AtomicReference`. Mad respect for taking on the scariest and most historically-problematic lines of code in this entire codebase and making them genuinely better!

core/shared/src/main/scala/cats/effect/ContState.scala
```diff
@@ -18,11 +18,37 @@ package cats.effect
...
+ * identity).
+ *
+ * `wasFinalizing` and `handle` are published in terms of the `suspended` atomic variable in
+ * `IOFiber`
+ */
+private final class ContState(var wasFinalizing: Boolean)
+    extends AtomicReference[Either[Throwable, Any]] {
+
+  val waiting: Either[Throwable, Any] =
+    ContState.waitingSentinel // this is just a reference to the sentinel
```
armanbilge, line 40: @durban I'm a bit late, but I'm curious about this encoding. Does this really provide an optimization despite increasing the memory footprint of `ContState` (or so I assume)?

durban: Well, it's hard to say. The benchmarks I did were already inconclusive (hard to say if the improvement is more than noise). Although, I admit, I did not run them without this "optimization".

The idea is simply to avoid fetching the sentinel from a static (e.g., an `object` field). I seem to remember something about statics (or static fields?) needing a memory barrier. Although I'm sure the JIT is pretty smart about that... so again, I'm not sure. Of course a proper way of doing this would be to measure, but it's so noisy, that I don't have big hopes.

(Somewhat, but not really, related: I did run benchmarks for scala `object` methods vs. java static methods. Usually there is no difference, but sometimes, when the number of threads approaches the number of physical threads, the java static methods seem to be faster. I don't really know why, but the scala bytecode have +1 `getstatic` I think.)

PR #3403, opened 2023-02-05
Fix `CallbackStack` leak, restore specialized `IODeferred`

### LINKED ISSUE #3366: Restore specialized `Deferred`
Requires fix to `CallbackStack` packing, which I have mostly done locally.

Relates to #3365

armanbilge:
Closes https://github.com/typelevel/cats-effect/issues/3366.

### This PR
```
[info] Benchmark                   (count)   Mode  Cnt     Score    Error  Units
[info] DeferredBenchmark.cancel         10  thrpt   20    38.933 ±  0.777  ops/s
[info] DeferredBenchmark.cancel        100  thrpt   20     5.323 ±  0.401  ops/s
[info] DeferredBenchmark.cancel       1000  thrpt   20     0.395 ±  0.028  ops/s
[info] DeferredBenchmark.complete       10  thrpt   20    35.443 ±  4.403  ops/s
[info] DeferredBenchmark.complete      100  thrpt   20    10.200 ±  2.016  ops/s
[info] DeferredBenchmark.complete     1000  thrpt   20     1.687 ±  0.058  ops/s
[info] DeferredBenchmark.get            10  thrpt   20  1880.096 ± 12.066  ops/s
[info] DeferredBenchmark.get           100  thrpt   20   277.012 ±  2.158  ops/s
[info] DeferredBenchmark.get          1000  thrpt   20    28.974 ±  0.091  ops/s
```

### 3.4.x
```
[info] Benchmark                   (count)   Mode  Cnt    Score    Error  Units
[info] DeferredBenchmark.cancel         10  thrpt   20   42.868 ±  1.431  ops/s
[info] DeferredBenchmark.cancel        100  thrpt   20    4.399 ±  0.254  ops/s
[info] DeferredBenchmark.cancel       1000  thrpt   20    0.361 ±  0.025  ops/s
[info] DeferredBenchmark.complete       10  thrpt   20   25.492 ±  2.345  ops/s
[info] DeferredBenchmark.complete      100  thrpt   20    8.211 ±  1.625  ops/s
[info] DeferredBenchmark.complete     1000  thrpt   20    1.216 ±  0.271  ops/s
[info] DeferredBenchmark.get            10  thrpt   20  906.738 ± 83.481  ops/s
[info] DeferredBenchmark.get           100  thrpt   20  166.403 ±  4.774  ops/s
[info] DeferredBenchmark.get          1000  thrpt   20   19.388 ±  1.687  ops/s
```

### DIFF
--- a/core/js/src/main/scala/cats/effect/CallbackStack.scala
+++ b/core/js/src/main/scala/cats/effect/CallbackStack.scala
@@ -57,6 +57,8 @@ private final class CallbackStackOps[A](private val callbacks: js.Array[A => Uni
 
   @inline def clear(): Unit =
     callbacks.length = 0 // javascript is crazy!
+
+  @inline def pack(bound: Int): Int = bound
 }
 
 private object CallbackStack {
--- a/core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
+++ b/core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
@@ -80,35 +80,40 @@ private final class CallbackStack[A](private[this] var callback: A => Unit)
 
   /**
    * It is intended that `bound` be tracked externally and incremented on each clear(). Whenever
-   * pack is called, the number of empty cells removed from the stack is produced. It is expected
-   * that this value should be subtracted from `bound` for the subsequent pack/clear calls. It is
-   * permissible to pack on every clear() for simplicity, though it may be more reasonable to delay
-   * pack() calls until bound exceeds some reasonable threshold.
+   * pack is called, the number of empty cells removed from the stack is produced. It is
+   * expected that this value should be subtracted from `bound` for the subsequent pack/clear
+   * calls. It is permissible to pack on every clear() for simplicity, though it may be more
+   * reasonable to delay pack() calls until bound exceeds some reasonable threshold.
    *
-   * The observation here is that it is cheapest to remove empty cells from the front of the list,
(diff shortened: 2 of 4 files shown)


djspiewak (dismissed):
image

durban:
It is not closely related to this PR, but I think the `CallbackStack` in `IOFiber` will also need to do this. (If I understand it correctly, cancelling a lot of `join`s would cause the same leak.)

djspiewak:
> It is not closely related to this PR, but I think the CallbackStack in IOFiber will also need to do this. (If I understand it correctly, cancelling a lot of joins would cause the same leak.)

There's a space/time tradeoff here. In order to do that, `IOFiber` would need to track some additional state (the `AtomicInteger` isn't free!), and `join` cancelation would become proportionally more expensive. In essence, we would need to deoptimize the common case of `IOFiber` (which is to say, not having many `join`ers, and not `cancel`ing them very often when we do) in order to optimize the *uncommon* case. I think we can be pretty confident in the uncommonness by virtue of the fact that no one has complained about this in the last two and a half years, despite the fact that it's been like this the whole time.

Really it's the observation that `Deferred#get` and `Fiber#join` are just used in slightly different patterns, and so `Deferred` needs to be a bit more space-efficient in the face of cancelation, while `IOFiber` can trade off some wasted space in order to get a more efficient fiber footprint and faster `cancel`.

Issue #3385, opened 2023-01-27
`IOLocal`: MTL `Local` instance & semantics


armanbilge:
@bpholt recently brought up the issue of providing an MTL `Local` instance for `IOLocal` in https://github.com/armanbilge/oxidized/issues/9#issuecomment-1405824000. That repo exists because Cats MTL can't provide it (since it creates a dep cycle with CE), and Cats Effect hasn't provided it (so far). Here are some reasons we should reconsider.

1. As noted, Cats Effect already has a dependency on MTL, via the testkit. So it's not technically increasing our bincompat vulnerability.[^1]
2. http4s is pursuing MTL-based middleware for 1.0. So MTL will likely end up on the classpath of all practical Typelevel applications anyway.

Furthermore: there is a lot of frustration about the semantics to expect from `IOLocal` e.g. https://github.com/typelevel/cats-effect/issues/3100, https://github.com/typelevel/fs2/issues/2842.

The interesting thing is, if you use `IOLocal` _only_ as a `Local` (i.e. none of its lower-level methods) I am pretty sure you won't run into weird behavior like that. `Local` encodes semantics that make it safe to use `IOLocal`, without leaking implementation details about how `Stream` or whatever uses `Fiber`s under-the-hood. Of course, it is less powerful when used this way, but that's par-for-the-course.

So I think there's an argument to push `Local` as the blessed way to interact with an `IOLocal`, unless you are specifically doing low-level stuff. Think `.background` to `.start`. And that could be another good reason to provide such an instance directly.

[^1]: although I suppose in the current state we _could_ break the testkit if MTL _had_ to break, and it would be less bad than breaking core.

rossabaker:
Would you construct it from an extant `IOLocal` or create an encapsulated one from an initial state so nobody else can do weird shit with it?

armanbilge:
> create an encapsulated one from an initial state so nobody else can do weird shit with it?

IMO this should be the "blessed" way, but I think the alternative constructor is good to have around too.

rossabaker:
An attractive feature of MTL local is that it surfaces the inherent limitations at compile time.  For example, scoping the use of a resource.  By signature, you can't!

```scala
    def localK[F[_], E](f: E => E)(implicit L: Local[F, E]) =
      new (F ~> F) {
        def apply[A](fa: F[A]): F[A] =
          L.local(fa)(f)
      }

[...]

Issue #634, opened 2019-09-12
Cats Effect 3.0 Proposal


djspiewak:
*(moved from https://gist.github.com/djspiewak/fb7851cc4f804f851e70d15ba7c94bb1)*

*If you're someone for whom code speaks louder than words, here's a branch to ponder: https://github.com/typelevel/cats-effect/tree/ce3*

The following is a proposal for Cats Effect 3, collaboratively devised over the last few months by several of the maintainers and contributors to the library. I will frequently use first-person in this document (because I'm lazy), which is posted under my name (ditto), but it really was a joint effort. As with any joint effort, there has definitely been some healthy debate about several of the concepts in this proposal. For reference, these are the things we're the least sure about:

- `Concurrent` now takes two type parameters, and thus is somewhat less ergonomic
- `Region` has been added, along with some complex associated machinery
  + One component of that machinery is the `Safe.Case` type
- Several existing laws are retained (such as `Sync` being stack-safe)
- `ExecutionContext` appears in the API
- `ExitCase` is a really terrible name (the names in general need some polishing)

None of this is final! At all! Really. Now is the time for questions and debate and concerns and bike-shedding. Put forward your own ideas. Find fundamental problems with what we've done! The sky's the limit!

## Preamble

Before diving into the exact changes that we're proposing, I want to briefly talk about Cats Effect, what it is and where it fits in the ecosystem. Cats Effect is really two things at once:

- A set of abstractions for characterizing functional effectful runtimes in Scala
- A baseline, practical implementation of those abstractions

Both of these elements of the project are extremely important, though the success of the former in the ecosystem and in end-user applications has rendered the latter *less* important than it was at the conception of the project. Note that, from an ecosystem strategic standpoint, this implies that Cats Effect has three primary stakeholders:

- **Runtime Systems**. Otherwise known as "concrete datatypes". These are the *implementors* of the Cats Effect abstractions. The current (known) major members of this category are Monix and ZIO, in addition to Cats Effect itself, though technically monad transformers mean that there are an infinite number of inductive concrete types.
- **Middleware**. These are things like http4s, Doobie, Monix, and many many others. Basically any Scala code which uses Cats Effect and *doesn't* include a `main` function.
- **End Users**. Most of this code is closed-source, but there are some bits and pieces which are public, such as [quasar](https://github.com/slamdata/quasar). These are projects which are delivered to customers, deployed on servers, etc. They always include a `main`, or at least something like `IOApp`.

[...]

PR #3405, opened 2023-02-05
Remove `IOLocal#scope`, revert #3214

armanbilge:
Reverts #3214, with apologies to @iRevive. IIRC I even encouraged that PR, so my bad.

Basically, for `scope` to work correctly, the `acquire` and `release` effects must run on the same fiber. `Resource` cannot guarantee this, so IMO it's a leaky abstraction.  Furthermore Ross and I demonstrated various weird/surprising behaviors when mixed with other combinators starting from https://github.com/typelevel/cats-effect/pull/3360#issuecomment-1416883633.

So yeah, I feel very unsure about exposing an API like this now. If we want to offer "higher-level" APIs that can be used safely, my vote is for something like https://github.com/typelevel/cats-effect/issues/3385.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -906,8 +906,7 @@ lazy val std = crossProject(JSPlatform, JVMPlatform, NativePlatform)
         "cats.effect.std.Queue#DroppingQueue.onOfferNoCapacity"),
       // introduced by #3347
       // private stuff
-      ProblemFilters.exclude[MissingClassProblem](
-        "cats.effect.std.AtomicCell$Impl")
+      ProblemFilters.exclude[MissingClassProblem]("cats.effect.std.AtomicCell$Impl")
     )
   )
   .jsSettings(
--- a/core/shared/src/main/scala/cats/effect/IOLocal.scala
+++ b/core/shared/src/main/scala/cats/effect/IOLocal.scala
@@ -17,7 +17,6 @@
 package cats.effect
 
 import cats.data.AndThen
-import cats.syntax.functor._
 
 /**
  * [[IOLocal]] provides a handy way of manipulating a context on different scopes.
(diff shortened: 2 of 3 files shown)


rossabaker:
I have incidentally talked myself into it being fine.  There are pointy bits, but I think the behavior is correct according to spec, and there are lots of other pointy bits.  Why not remove `.set`, which also does weird shit in races?

I agree that higher-level APIs are needed for safe use.  #3385 is excellent for "static monadic regions", but not up to the task of "dynamic monadic regions", to borrow terms from #634.  I'm still pursuing that holy grail for otel4s.

armanbilge:
> I think the behavior is correct according to spec

Is it? It says this.

```
The original value is restored upon the finalization of a resource.
```

There's no way it can guarantee this, without guaranteeing that it can run those on the same fiber. Which it cannot, because `allocated`.

rossabaker:
Okay, that's fair.  But `allocated` is "an advanced and potentially unsafe api which can cause a resource leak if not used correctly".  Is there an adversarial function we can pass to `use` that would break it?

Stream gets into trouble not by calling `allocated`, but pattern matching to destructure `Resource.Allocated`, which I consider morally the same thing.

armanbilge:
> But `allocated` is "an advanced and potentially unsafe api which can cause a resource leak if not used correctly"

Yes, but whose job is it to use correctly? If I compose my `local.scope` resource with some other resources and hand it off to something that does `.allocated` (e.g. a pool), is it their fault that it all went wrong?

armanbilge:
The problem is, that even if `allocated` is an advanced API, you _can_ use it correctly even if you run the acquire and release on different fibers. That's allowed.

Which is why this API doesn't work: to deliver what it promises, it relies on behavior which is not guaranteed.

rossabaker:
That's fair.  FS2 is using `Allocated` appropriately: it releases what it acquires.  We are getting bad behavior with FS2 and IOLocal not because FS2 is buggy, but because IOLocal imposes constraints FS2 isn't designed for.

djspiewak:
I think there are enough footguns here that I'd rather err on the side of caution. We can always bring it back in a later release.

wunderk1nd-e:
@djspiewak @rossabaker Would going back to the originally proposed implementation using bracket dodge the issues with how the `Resource` might end up getting used?
https://github.com/typelevel/cats-effect/pull/3128

armanbilge:
@wunderk1nd-e thanks for chiming in! Yes, I believe that would address the issues, since the acquire and release phases of `bracket` are guaranteed to run on the same fiber.

Issue #3370, opened 2023-01-16
RFC: add uncancelable `Ref#flatModify` and `flatModifyFull(poll => ...)`


armanbilge:
We added `Ref#flatModify` in https://github.com/typelevel/cats-effect/pull/3273. But then after concerns about its uncancelability semantics in https://github.com/typelevel/cats-effect/issues/3306, we ended up removing it in https://github.com/typelevel/cats-effect/pull/3311.

Personally, I think `flatModify` is a useful combinator to have, particularly when it is uncancelable. It's just too easy to forget e.g.
https://github.com/typelevel/fs2/blob/6318ca2673c98c522e7d092ed6ea447f1ac2175a/core/shared/src/main/scala/fs2/concurrent/Signal.scala#L256

Similar to how the `acquire` phase in `Resource.make` is uncancelable (and we offer the `makeFull` variant with `poll`) I would be in favor of adding uncancelable `flatModify` and `flatModifyFull`.

PR #3408, opened 2023-02-07
Add 'flatModify', 'flatModifyFull' and corresponding 'State' methods

seigert:
My attempt at #3370.

I'm not exactly sure about `*Full` method signatures, it could also be `f: Poll[F] => A => (A, F[B])`. I've tried `f: A => (A, Poll[F] => F[B])` and `State[A, Poll[F] => F[B]]` but it does not plays well with type inference.


djspiewak (approved):
Thanks for taking this on!

PR #3409, opened 2023-02-09
Even faster async mutex

armanbilge:
This PR tries my idea from https://github.com/typelevel/cats-effect/pull/3346#issuecomment-1374040239 of re-using the `UnsafeUnbounded` datastructure from the async queue to implement the async mutex. /cc @BalmungSan 

#### this PR
```
[info] Benchmark                           (fibers)  (iterations)   Mode  Cnt    Score   Error  Units
[info] MutexBenchmark.cancellationAsync          10          1000  thrpt   20   15.434 ± 1.058  ops/s
[info] MutexBenchmark.cancellationAsync         100          1000  thrpt   20    3.203 ± 0.464  ops/s
[info] MutexBenchmark.happyPathAsync             10          1000  thrpt   20  145.271 ± 5.933  ops/s
[info] MutexBenchmark.happyPathAsync            100          1000  thrpt   20   13.297 ± 1.656  ops/s
[info] MutexBenchmark.highContentionAsync        10          1000  thrpt   20   22.955 ± 2.932  ops/s
[info] MutexBenchmark.highContentionAsync       100          1000  thrpt   20    2.953 ± 0.479  ops/s
```
#### series/3.x
```
[info] Benchmark                           (fibers)  (iterations)   Mode  Cnt    Score   Error  Units
[info] MutexBenchmark.cancellationAsync          10          1000  thrpt   20   14.627 ± 1.514  ops/s
[info] MutexBenchmark.cancellationAsync         100          1000  thrpt   20    3.234 ± 0.658  ops/s
[info] MutexBenchmark.happyPathAsync             10          1000  thrpt   20  124.522 ± 7.762  ops/s
[info] MutexBenchmark.happyPathAsync            100          1000  thrpt   20   11.987 ± 0.379  ops/s
[info] MutexBenchmark.highContentionAsync        10          1000  thrpt   20   16.659 ± 2.072  ops/s
[info] MutexBenchmark.highContentionAsync       100          1000  thrpt   20    2.131 ± 0.127  ops/s
```

### DIFF
--- a/benchmarks/src/main/scala/cats/effect/benchmarks/MutexBenchmark.scala
+++ b/benchmarks/src/main/scala/cats/effect/benchmarks/MutexBenchmark.scala
@@ -51,10 +51,7 @@ class MutexBenchmark {
   var iterations: Int = _
 
   private def happyPathImpl(mutex: IO[Mutex[IO]]): Unit = {
-    mutex
-      .flatMap { m => m.lock.use_.replicateA_(fibers) }
-      .replicateA_(iterations)
-      .unsafeRunSync()
+    mutex.flatMap { m => m.lock.use_.replicateA_(fibers * iterations) }.unsafeRunSync()
   }
 
   @Benchmark
--- a/std/shared/src/main/scala/cats/effect/std/Mutex.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Mutex.scala
@@ -21,7 +21,7 @@ package std
 import cats.effect.kernel._
 import cats.syntax.all._
 
-import java.util.concurrent.atomic.AtomicReference
+import java.util.concurrent.atomic.AtomicBoolean
 
 /**
  * A purely functional mutex.
@@ -84,9 +84,7 @@ object Mutex {
    * Creates a new `Mutex`. Like `apply` but initializes state using another effect constructor.
    */
   def in[F[_], G[_]](implicit F: Sync[F], G: Async[G]): F[Mutex[G]] =
-    F.delay(
-      new AtomicReference[LockCell]()
-    ).map(state => new AsyncImpl[G](state)(G))
+    F.delay(new AsyncImpl[G])
 
   private final class ConcurrentImpl[F[_]](sem: Semaphore[F]) extends Mutex[F] {
     override final val lock: Resource[F, Unit] =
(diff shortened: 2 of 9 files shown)


std/shared/src/main/scala/cats/effect/std/package.scala
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+
+package object std {
+
+  private[std] val FailureSignal: Throwable = new RuntimeException
+    with scala.util.control.NoStackTrace
```
armanbilge: Instead of placing this in the package object we could also make a top-level `object FailureSignal extends RuntimeException with NoStackTrace`.

std/shared/src/main/scala/cats/effect/std/UnsafeUnbounded.scala
```diff
@@ -0,0 +1,112 @@
...
+ * limitations under the License.
+ */
+
+package cats.effect.std
+
+import scala.annotation.tailrec
+
+import java.util.concurrent.atomic.AtomicReference
+
+private final class UnsafeUnbounded[A] {
```
armanbilge, line 23: Didn't make any changes here, just copy-pasta.

std/shared/src/main/scala/cats/effect/std/Mutex.scala
```diff
@@ -96,92 +94,65 @@ object Mutex {
...
+        while (waiter eq null) waiter = waiters.take()
+        waiter(RightTrue) // pass the buck
+      } catch { // no waiter found
+        case FailureSignal =>
+          locked.set(false) // release
+          try {
+            var waiter = waiters.take()
+            while (waiter eq null) waiter = waiters.take()
+            waiter(RightFalse) // waken any new waiters
+          } catch {
```
armanbilge, line 137: There's some fairness corruption here under contention, where an acquirer may cut-in-line of an acquirer that had placed itself in the queue.

BalmungSan: Are the FIFO semantics of the current `Mutex` something we would like to preserve? If so, should we be louder about it on the docs?

BTW, does the `Semaphore` based one guarantee that as well?

armanbilge: Overall it's still FIFO (or should be, we should add a test if we don't have already). It's just under contention it can be slightly corrupted—but under contention, who was really "first" anyway?

I think this was the long-running debate Daniel and Fabio had for the async queue :)

benchmarks/src/main/scala/cats/effect/benchmarks/MutexBenchmark.scala
```diff
@@ -51,10 +51,7 @@ class MutexBenchmark {
   var iterations: Int = _
 
   private def happyPathImpl(mutex: IO[Mutex[IO]]): Unit = {
-    mutex
-      .flatMap { m => m.lock.use_.replicateA_(fibers) }
-      .replicateA_(iterations)
-      .unsafeRunSync()
+    mutex.flatMap { m => m.lock.use_.replicateA_(fibers * iterations) }.unsafeRunSync()
```
BalmungSan, line 54: I wonder what is the rationale for this change? And why only to the happy path?

armanbilge: Maybe I misunderstood the purpose of the benchmark, but we want to replicate many acquire/releases of the same mutex from the same fiber—we don't need to allocate a new mutex in each iteration, and "fibers" is not really accurate term. It's just iterations in the end.

Actually you are right, we can probably make a similar change to the other benchmarks.

std/shared/src/main/scala/cats/effect/std/UnsafeUnbounded.scala
armanbilge: Oh, another advantage of pulling this out into a top-level class it is much easier to provide an specialized implementation for JS without `Platform` shenanigans.

std/shared/src/main/scala/cats/effect/std/Mutex.scala
```diff
@@ -96,92 +94,65 @@ object Mutex {
...
-      }
+  private final class AsyncImpl[F[_]](implicit F: Async[F]) extends Mutex[F] {
+    import AsyncImpl._
 
-    // Awaits until the Mutex is free.
-    private def await(thisCell: LockCell): F[Unit] =
-      F.asyncCheckAttempt[Unit] { thisCB =>
-        F.delay {
-          val previousCell = state.getAndSet(thisCell)
+    private[this] val locked = new AtomicBoolean(false)
```
armanbilge, line 100: If we swap this `AtomicBoolean` for an `AtomicInteger` then do we basically have an `AsyncSemaphore`?

std/shared/src/main/scala/cats/effect/std/UnsafeUnbounded.scala
djspiewak: I originally planned on pulling it out to the top-level but got lazy.

std/shared/src/main/scala/cats/effect/std/package.scala
djspiewak: I think that *either* approach introduces a memory barrier unless we locally cache the value. If we locally cache the value, then they're the same. I would prefer the top-level object approach, tbh, but we really need to make sure it's pre-cached within scopes.

djspiewak (changes requested):
Minor change

std/shared/src/main/scala/cats/effect/std/Mutex.scala
```diff
@@ -99,6 +99,7 @@ object Mutex {
 
     private[this] val locked = new AtomicBoolean(false)
     private[this] val waiters = new UnsafeUnbounded[Either[Throwable, Boolean] => Unit]
+    private[this] val failureSignal = FailureSignal // prefetch
```
djspiewak: Super minor thing but I'd prefer if we shadow `FailureSignal` here by prefetching the fully-qualified name. This avoids the need for the backticks, and also prevents people from accidentally introducing memory barriers by referring to the non-prefetched thing.

djspiewak (changes requested):

std/shared/src/main/scala/cats/effect/std/Mutex.scala
armanbilge: Yeah I thought about this as well. I was more afraid that I would _forget_ to prefetch it, unless I always referred to it under a name only available as a prefetch.

djspiewak (approved):

std/shared/src/main/scala/cats/effect/std/Mutex.scala
djspiewak: Well now this way your *only* option is to prefetch.

armanbilge: If you remembered to shadow it in the first place 😆 if the shadow doesn't exist in your current scope, you have no way of knowing that you are going through the memory barrier.

djspiewak: I feel like we introduce new scopes less often than we introduce new code in scopes. :-P

Issue #3410, opened 2023-02-10
Create "Recipes" node in the documentation


mtomko:
I was recently looking for a way to solve a simple problem ("start a supervised task that outlives an http request") and ended up asking on discord. I got an answer (thanks @armanbilge) but in the ensuing discussion we decided it might be nice to curate a list of example solutions to simple problems like this one and publish it in the documentation.

I am willing to start writing out the documentation node and will include an example solution for the problem I originally had, but will gladly include other simple examples anyone feels like contributing or requesting. I think it'll be nice for the page to launch with more than one example. If nothing else, I can mine our codebase for common patterns.

armanbilge:
Another name idea: "recipes" :)

I bet a lot of stuff in cats-effect-std can be covered in these recipes. I'll try to keep adding as I think of them.
- [ ] "how to atomically update a ref with the result of an effect" aka `AtomicCell` :)

samspills:
I mentioned in the discord, but I coincidentally wrote a bit yesterday about scheduling an update to a `Ref`, using an fs2 `Stream` and combining it with an http4s api+server to also interact with the ref. There's not a lot of explaining but the [post is here](https://www.blerf.ca/words/refs-and-streams-2/) if people want to check it out as a potential example to include :)

arosien:
The examples can be a kind of "reverse-index" of the standard library, as you often start with a problem and are trying to find the components of the solution. Only having the list of standard library components doesn't help a user unless they already know what the components do.

mtomko:
I've made a draft PR with an example. I may work on some other examples but I wanted to start by getting something into a PR state. I agree with the idea of using this as a reverse index, and I think many of the members of the standard library deserve something here.

mtomko:
It's very tempting to make the example just:

```scala
import scala.concurrent.duration._

import cats.effect.{IO, IOApp}
import cats.effect.std.Supervisor
import cats.syntax.all._

object Joke extends IOApp.Simple {

  val run =
    Supervisor[IO](await = true).use { supervisor =>
      for {
        _ <- supervisor.supervise(IO.sleep(100.millis) >> IO.print("MOO!"))
        _ <- IO.println("Q: Knock, knock!")
        _ <- IO.println("A: Who's there?")
        _ <- IO.println("Q: Interrupting cow.") >> IO.sleep(60.millis)
        _ <- "A: Interrupting cow who?".toList.traverse_(IO.print) >> IO.println("")
      } yield ()
    }

}
```

[...]

PR #3412, opened 2023-02-14
Add Recipes node with simple example for `Supervisor`

mtomko:
This is intended to (begin to) address https://github.com/typelevel/cats-effect/issues/3410

I haven't figured out how to run a docs server and see the resulting code, so I'm flying a bit blind. Submodules in `git` are new to me, so I wasn't sure how to check out the submodules while still targeting the 3.4.x branch.

### DIFF
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -1,0 +1,58 @@
+---
+id: recipes
+title: Recipes
+---
+
+```scala mdoc:invisible:reset-object
+import scala.concurrent.duration._
+
+import cats.effect._
+
+final case class Request(path: String, paramaters: Map[String, List[String]])
+
+sealed trait Response extends Product with Serializable
+final case object NotFound extends Response
+final case class Ok(payload: String) extends Response
+
+// dummy case class representing the bound server
+final case class IpAddress()
+
+// an HTTP server is a function from request to IO[Response] and is managed within a Resource
+final case class HttpServer(handler: Request => IO[Response]) {
+  def resource: Resource[IO, IpAddress] = Resource.eval(IO.never.as(IpAddress()))
+}
+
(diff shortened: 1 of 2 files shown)


mtomko:
I may add https://scastie.scala-lang.org/ewn7NUzwSoSPwNSoaL9KBA as an example too, unless people think it's too silly

docs/recipes.md
```diff
@@ -0,0 +1,92 @@
...
+  def handler(supervisor: Supervisor[IO]): Request => IO[Response] = {
+    case Request("start", params) => 
+      supervisor.supervise(longRunningTask(params)).void >> IO.pure(Ok("started"))
+    case Request(_, _) => IO.pure(NotFound)
+  }
+
+  val run =
+    Supervisor[IO](await = true).use { supervisor =>
+      HttpServer(handler(supervisor)).resource.use(_ => IO.never)
+    }
```
djspiewak:
```suggestion
    Supervisor[IO](await = true).flatMap { supervisor =>
      HttpServer(handler(supervisor)).resource
    }.useForever
```

Could also futz with `ResourceApp` to golf this down even further, but I do think that using `flatMap` rather than `use` here is important.

djspiewak (changes requested):
My largest concern is this sets a high bar for people who want to add more recipes! 😃 Outstanding work. Minor change but otherwise good to go.

docs/recipes.md
mtomko: This looks great, thanks for the suggestion. I wouldn't have thought of `useForever`. Was `use` incorrect or simply unhelpful?

mtomko: I've committed your suggestion. I'll glance at `ResourceApp`, too, though.

djspiewak (approved):

PR #3417, opened 2023-02-14
Fix `fromCompletableFuture` cancelation leak

TimWSpence:
Previously we could observe cancelation on the `flatMap` boundary before we register the finalizer.

### DIFF
--- a/kernel/jvm/src/main/scala/cats/effect/kernel/AsyncPlatform.scala
+++ b/kernel/jvm/src/main/scala/cats/effect/kernel/AsyncPlatform.scala
@@ -32,37 +32,36 @@ private[kernel] trait AsyncPlatform[F[_]] extends Serializable { this: Async[F] 
    */
   def fromCompletableFuture[A](fut: F[CompletableFuture[A]]): F[A] = uncancelable { poll =>
     flatMap(poll(fut)) { cf =>
-      onCancel(
-        poll(cont {
-          new Cont[F, A, A] {
-            def apply[G[_]](implicit G: MonadCancelThrow[G])
-                : (Either[Throwable, A] => Unit, G[A], F ~> G) => G[A] = {
-              (resume, get, lift) =>
-                G.uncancelable { poll =>
-                  val go = delay {
-                    cf.handle[Unit] {
-                      case (a, null) => resume(Right(a))
-                      case (_, t) =>
-                        resume(Left(t match {
-                          case e: CompletionException if e.getCause ne null => e.getCause
-                          case _ => t
-                        }))
-                    }
-                  }
-
-                  val await = G.onCancel(
-                    poll(get),
-                    // if cannot cancel, fallback to get
-                    G.ifM(lift(delay(cf.cancel(false))))(G.unit, G.void(get))
-                  )
-
-                  G.productR(lift(go))(await)
+      poll(cont {
+        new Cont[F, A, A] {
+          def apply[G[_]](implicit G: MonadCancelThrow[G])
+              : (Either[Throwable, A] => Unit, G[A], F ~> G) => G[A] = { (resume, get, lift) =>
+            G.uncancelable { poll =>
+              val go = delay {
+                cf.handle[Unit] {
+                  case (a, null) => resume(Right(a))
+                  case (_, t) =>
+                    resume(Left(t match {
+                      case e: CompletionException if e.getCause ne null => e.getCause
+                      case _ => t
+                    }))
                 }
+              }
+
+              // if cannot cancel, fallback to get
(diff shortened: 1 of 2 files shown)


TimWSpence:
Is this correct or am I going crazy? 😅

armanbilge (commented):
No, this seems right. Ooph, good catch, my bad, thanks!

Actually I think this may need to be fixed in 3.4.x 🤔

armanbilge:
If you can, I wonder how difficult it would be to add a test case for this with the test runtime. So I don't screw it up next time I decide to refactor 😇

TimWSpence:
> If you can, I wonder how difficult it would be to add a test case for this with the test runtime. So I don't screw it up next time I decide to refactor 😇

Not sure but I'll give it a go! I'm pretty sure I've fixed something very similar to this before 😂

TimWSpence:
> No, this seems right. Ooph, good catch, my bad, thanks!
> 
> Actually I think this may need to be fixed in 3.4.x 🤔

Yeah I think it does. Do you want me to change target branch? Not sure if we have a preferred direction in which to cherry-pick

armanbilge:
Yes, please re-target :)

The test runtime should make it easier to test this sort of thing, since it emulates non-deterministic orderings of effects. So we don't need to try and make a real race condition.

TimWSpence:
@armanbilge just added a test that seems to fail reliably on the old version. However, I think my fix has potentially removed the backpressure and restoring it would involve introducing a second callback again. I can't see a way to mess with the `cont` node to make it safe but maybe you can?

TimWSpence:
Ok pushed something that looks vaguely sensible (I think 😅). I thought it failed the test at first but now I can't get it to fail again 😭

armanbilge:
@TimWSpence I adapted an existing test in 74bb273af22d78a99064d290e014b20d219706f9 which demonstrates the leak. It uses `TestControl` instead of relying on runtime race conditions, which is what I meant in https://github.com/typelevel/cats-effect/pull/3417#issuecomment-1430204853 :)

TimWSpence:
> @TimWSpence I adapted an existing test in [74bb273](https://github.com/typelevel/cats-effect/commit/74bb273af22d78a99064d290e014b20d219706f9) which demonstrates the leak. It uses `TestControl` instead of relying on runtime race conditions, which is what I meant in [#3417 (comment)](https://github.com/typelevel/cats-effect/pull/3417#issuecomment-1430204853) :)

Oh nice! I couldn't see if that is on a branch? Should I move it into this PR in place of the current test? Also we choose a random number between 0 and 2048 and run 1000 iterations. Should we just run every value in 0-1024 instead?

Also, any thoughts in why it still fails after the changes to `fromCompletableFuture` that I made? 😅

armanbilge:
> I couldn't see if that is on a branch?

https://github.com/armanbilge/cats-effect/tree/bug/completable-future-cancellation-leak

> Should I move it into this PR in place of the current test?

So you got the hint 😜 

> Also we choose a random number between 0 and 2048 and run 1000 iterations. Should we just run every value in 0-1024 instead?

No, because there is other randomization happening as well within the test runtime, which schedules the fibers in various orders.

> Also, any thoughts in why it still fails after the changes to `fromCompletableFuture` that I made? 😅

What fails exactly? Not sure, but I can take another look today. Thanks for cleaning up my mess!

kernel/jvm/src/main/scala/cats/effect/kernel/AsyncPlatform.scala
```diff
@@ -30,13 +30,12 @@ private[kernel] trait AsyncPlatform[F[_]] extends Serializable { this: Async[F]
...
-  def fromCompletableFuture[A](fut: F[CompletableFuture[A]]): F[A] = flatMap(fut) { cf =>
-    cont {
-      new Cont[F, A, A] {
-        def apply[G[_]](
-            implicit
-            G: MonadCancelThrow[G]): (Either[Throwable, A] => Unit, G[A], F ~> G) => G[A] = {
-          (resume, get, lift) =>
+  def fromCompletableFuture[A](fut: F[CompletableFuture[A]]): F[A] = uncancelable { poll =>
+    flatMap(poll(fut)) { cf =>
+      poll(cont {
```
armanbilge, line 35: Ok, so the problem is here ... basically if you do
```scala
uncancelable { poll =>
  poll(fa).flatMap(a => poll(f(a))
}
```
this is equivalent to just doing
```scala
fa.flatMap(a => f(a))
```
:trollface:

This was tricky, but after a couple tries I made it work by evaluating the `fut` _within_ the `G.uncancelable` below.

PR #3421, opened 2023-02-16
Improved `Ref` scaladoc

durban:
I've tried to improve the scaladoc of `Ref`. English is not my first language, so review carefully :-)

See some explanations as inline comments.

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Ref.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Ref.scala
@@ -31,8 +31,8 @@ import cats.syntax.all._
  *
  * The default implementation is nonblocking and lightweight, consisting essentially of a purely
  * functional wrapper over an `AtomicReference`. Consequently it ''must not'' be used to store
- * mutable data as `AtomicReference#compareAndSet` and friends are not threadsafe and are
- * dependent upon object reference equality.
+ * mutable data as `AtomicReference#compareAndSet` and friends are dependent upon object
+ * reference equality.
  *
  * @see
  *   [[cats.effect.std.AtomicCell]]
@@ -62,14 +62,14 @@ abstract class Ref[F[_], A] extends RefSource[F, A] with RefSink[F, A] {
     }
 
   /**
-   * Obtains a snapshot of the current value, and a setter for updating it. The setter may noop
-   * (in which case `false` is returned) if another concurrent call to `access` uses its setter
-   * first.
+   * Obtains a snapshot of the current value, and a setter for updating it.
    *
-   * Once it has noop'd a setter will never succeed.
+   * The setter attempts to modify the contents from the snapshot to the new value (and return
+   * `true`). If it cannot do this (because the contents changed since taking the snapshot), the
+   * setter is a noop and returns `false`.
    *
-   * Satisfies: `r.access.map(_._1) == r.get` `r.access.flatMap { case (v, setter) =>
-   * setter(f(v)) } == r.tryUpdate(f).map(_.isDefined)`
+   * Satisfies: `r.access.map(_._1) == r.get` and `r.access.flatMap { case (v, setter) =>
+   * setter(f(v)) } == r.tryUpdate(f).map(_.isDefined)`.
    */
   def access: F[(A, A => F[Boolean])]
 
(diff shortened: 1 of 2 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/Ref.scala
```diff
@@ -31,8 +31,8 @@ import cats.syntax.all._
  *
  * The default implementation is nonblocking and lightweight, consisting essentially of a purely
  * functional wrapper over an `AtomicReference`. Consequently it ''must not'' be used to store
- * mutable data as `AtomicReference#compareAndSet` and friends are not threadsafe and are
- * dependent upon object reference equality.
+ * mutable data as `AtomicReference#compareAndSet` and friends are dependent upon object
+ * reference equality.
```
durban, line 35: Removed the remark about CAS not being threadsafe, since... it's point is that it is threadsafe. The first sentence literally says "thread-safe".

kernel/shared/src/main/scala/cats/effect/kernel/Ref.scala
```diff
@@ -62,14 +62,14 @@ abstract class Ref[F[_], A] extends RefSource[F, A] with RefSink[F, A] {
     }
 
   /**
-   * Obtains a snapshot of the current value, and a setter for updating it. The setter may noop
-   * (in which case `false` is returned) if another concurrent call to `access` uses its setter
-   * first.
+   * Obtains a snapshot of the current value, and a setter for updating it.
    *
-   * Once it has noop'd a setter will never succeed.
```
durban, line 69: Since #2901 this was technically incorrect (the CAS can succeed a second time). Instead of this sentence, I've wrote more about what the setter actually does. (Another option would be to just... tell what it is: a CAS.)

kernel/shared/src/main/scala/cats/effect/kernel/Ref.scala
```diff
@@ -96,16 +96,16 @@ abstract class Ref[F[_], A] extends RefSource[F, A] with RefSink[F, A] {
   def update(f: A => A): F[Unit]
 
   /**
-   * Like `tryModify` but does not complete until the update has been successfully made.
+   * Like `tryModify` but retries until the update has been successfully made.
```
durban, line 99: This explains more clearly, what happens.

kernel/shared/src/main/scala/cats/effect/kernel/Ref.scala
```diff
@@ -62,14 +62,14 @@ abstract class Ref[F[_], A] extends RefSource[F, A] with RefSink[F, A] {
...
    *
-   * Once it has noop'd a setter will never succeed.
+   * The setter attempts to modify the contents from the snapshot to the new value (and return
+   * `true`). If it cannot do this (because the contents changed since taking the snapshot), the
+   * setter is a noop and returns `false`.
    *
-   * Satisfies: `r.access.map(_._1) == r.get` `r.access.flatMap { case (v, setter) =>
-   * setter(f(v)) } == r.tryUpdate(f).map(_.isDefined)`
+   * Satisfies: `r.access.map(_._1) == r.get` and `r.access.flatMap { case (v, setter) =>
+   * setter(f(v)) } == r.tryUpdate(f).map(_.isDefined)`.
```
durban, line 72: Cosmetic changes.

kernel/shared/src/main/scala/cats/effect/kernel/Ref.scala
```diff
@@ -96,16 +96,16 @@ abstract class Ref[F[_], A] extends RefSource[F, A] with RefSink[F, A] {
...
   /**
-   * Update the value of this ref with a state computation.
+   * Update the value of this `Ref` with a state computation.
    *
-   * The current value of this ref is used as the initial state and the computed output state is
-   * stored in this ref after computation completes. If a concurrent modification occurs, `None`
-   * is returned.
+   * The current value of this `Ref` is used as the initial state and the computed output state
+   * is stored in this `Ref` after computation completes. If a concurrent modification occurs,
+   * `None` is returned.
```
durban, line 108: Cosmetic changes.

kernel/shared/src/main/scala/cats/effect/kernel/Ref.scala
```diff
@@ -419,8 +419,6 @@ trait RefSink[F[_], A] extends Serializable {
    * Sets the current value to `a`.
    *
    * The returned action completes after the reference has been successfully set.
-   *
-   * Satisfies: `r.set(fa) *> r.get == fa`
```
durban, line 423: This was confusing, and ignoring the actual effect of setting the contents.

djspiewak (approved):
Great changes, thank you!

Issue #3423, opened 2023-02-16
`IO.println` never returns for long strings in nodejs runtime


cornerman:
We are experiencing an issue with `IO.println` in the nodejs runtime.
If we provide a long string to `IO.println`, it never actually prints anything and never returns.

The following is a little reproduction, which triggers this issue for me:

```scala
for {
   _ <- IO.unit
   _ = println("aal\n" * 500000) // works
   _ = println("DONE")
   _ <- IO(println("bal\n" * 500000)) // works
   _ = println("DONE")
   _ <- IO.println("cal\n" * 500000) // nothing printed, never returns
   _ = println("DONE")
} yield ()
```

Tested Versions:

cats-effect: 3.4.6 and 3.4.7
nodejs: 16.18.1 and 18.13.0

armanbilge:
Thanks for reporting. I believe the bug is here. Because it is "corked" it is not draining, but `write` will not proceed until it has drained.
https://github.com/typelevel/cats-effect/blob/0f050d1d31b6efc3ebaa3d66edb4a061a92a6939/std/js/src/main/scala/cats/effect/std/Console.scala#L123-L136

Replacing `IO.println(...)` with `IO.print(...)` appears to fix it.

The other uses of `println(...)` are working because they are using `console.log(...)` which does not properly backpressure I/O.

cornerman:
Thank you for the fast response and explanation <3

Though, I do not really understand why the drain event apparently works for small strings  - but I also do not know much about node-stream semantics. (Edit: ok, probably because it is so small, it needs no backpressure).

armanbilge:
Yes, exactly, `write(...)` returns a `Boolean` indicating if you should wait for the `'drain'` event. No need for small writes.

PR #3424, opened 2023-02-16
Fix corking when writing large strings to stdout/stderr on Node.js

armanbilge:
Fixes https://github.com/typelevel/cats-effect/issues/3423.

As mentioned in the issue, the bug was that corking forces buffering, which in turn prevents draining, so there was no `'drain'` event being emitted. Corking is used here to buffer the first write until we follow-up with the `write("\n")`, to minimize syscalls.

This fix is to uncork _before_ waiting on the `'drain'` callback ... not after 😅 

While revisiting this code, I also fixed a couple other issues related to corking. Namely, between corking/uncorking there should not be any `cede`s (e.g. due to concurrent `println(...)`) so it should happen in an atomic `delay` block. And the uncork should be in a `finally` block.

### DIFF
--- a/tests/js/src/test/scala/cats/effect/std/ConsoleJSSpec.scala
+++ b/tests/js/src/test/scala/cats/effect/std/ConsoleJSSpec.scala
@@ -23,6 +23,9 @@ class ConsoleJSSpec extends BaseSpec {
     "work in any JS environment" in real {
       Console[IO].println("printing") *> Console[IO].errorln("erroring") *> IO(true)
     }
+    "println should not hang for large strings" in real {
+      Console[IO].println("foo" * 10000).as(true)
+    }
   }
 
 }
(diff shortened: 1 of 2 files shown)


tests/js/src/test/scala/cats/effect/std/ConsoleJSSpec.scala
```diff
@@ -23,6 +23,9 @@ class ConsoleJSSpec extends BaseSpec {
     "work in any JS environment" in real {
       Console[IO].println("printing") *> Console[IO].errorln("erroring") *> IO(true)
     }
+    "println should not hang for large strings" in real {
+      Console[IO].println("foo" * 10000).as(true)
+    }
```
armanbilge, line 28: Unfortunately this test prints a massive `foofoofoo....` to the test logs 😂

djspiewak: Not sure how to avoid it in this case.

djspiewak (approved):

PR #3431, opened 2023-02-18
Nix shell -> Nix flake

rossabaker:
Everything else in the Typelevel org is a flake these days.  This gets things refreshed so the scala-native toolchain is available.


rossabaker:
flake.lock should be updated after https://github.com/typelevel/typelevel-nix/pull/90 merges.

rossabaker:
Canceled the CI pending the flake update to save some compute.

djspiewak:
Out of curiosity, what happens to those of us who are using `direnv` but *not* using Nix?

rossabaker:
I have wondered about that, but never seen an existence proof of such people.  I'm content to leave it out, but most Nix users will end up adding it locally, and no non-Nix user has ever complained about it.

(I would be highly interested in a global direnv setting that said "if there's no .envrc but there's a flake, use that", but that's not to be solved here...)

djspiewak:
```
direnv: using nix
environment:516: nix-shell: command not found
```

Apparently you just get this persistent error. Or you can block the `.envrc`, which appears to be what I was doing on http4s before I ran this experiment.

rossabaker:
I won't fight removing the .envrc.  A tiny handful of Nix users will find it handy, and a tinier handful of Nixless Direnv users will want to keep it blocked.  I'd find it much more compelling to remove it if anyone were creating a _different_ .envrc.  Otherwise, it's blocked by default, and what I believe to be the vast majority of direnv users will get a useful definition when they unblock it.

djspiewak:
Yeah I think it's fine since I don't have any compelling need for direnv in CE specifically. I won't deny you the convenience that I'm not using.

djspiewak (approved):

PR #3432, opened 2023-02-18
Fixed issues in the timer handling state machine integration

djspiewak:
This may fix @armanbilge's Fs2 issue. The essence here is the fact that `parkUntilNextSleeper` breaks an implicit assumption in the state machine, which is that the local queue cannot get *new* work items while the thread is parked. This is totally reasonable, but the timers stuff breaks this since new local work can come from timers expiring. The solution is to detect that some timer expired and then reroute directly to local work handling. In the rare case where the worker is awakened by external work *and* a timer expires simultaneously, the timer takes precedence, but the worker will eventually loop back around to state `0` which should poll externally or steal.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
@@ -327,21 +327,41 @@ private final class WorkerThread(
       }
     }
 
-    def parkUntilNextSleeper(): Unit = {
-      if (!isInterrupted()) {
+    // returns true if timed out, false if unparked
+    def parkUntilNextSleeper(): Boolean = {
+      while (!done.get()) {
         val now = System.nanoTime()
         val head = sleepersQueue.head()
         val nanos = head.triggerTime - now
+
+        // System.err.println(s"parking for $nanos nanos")
         LockSupport.parkNanos(pool, nanos)
 
-        if (parked.getAndSet(false)) {
-          pool.doneSleeping()
+        if (isInterrupted()) {
+          pool.shutdown()
+        } else {
+          if (parked.get()) {
+            // we were either awakened spuriously, or we timed out
+
+            if (head.triggerTime - System.nanoTime() <= 0) {
+              // we timed out
+              if (parked.getAndSet(false)) {
+                pool.doneSleeping()
+              }
+
+              return true
+            }
+          } else {
+            // we were awakened
+            return false
+          }
         }
       }
+
+      false
     }
 
     while (!done.get()) {
-
       if (blocking) {
         // The worker thread was blocked before. It is no longer part of the
         // core pool and needs to be cached.
@@ -393,25 +413,6 @@ private final class WorkerThread(
       }
 
       val sleepers = sleepersQueue
-
-      if (sleepers.nonEmpty) {
-        val now = System.nanoTime()
-
-        var cont = true
-        while (cont) {
-          val head = sleepers.head()
-
-          if (head.triggerTime - now <= 0) {
-            if (head.get()) {
-              head.callback(RightUnit)
-            }
-            sleepers.popHead()
-            cont = sleepers.nonEmpty
-          } else {
-            cont = false
-          }
-        }
-      }
 
       ((state & ExternalQueueTicksMask): @switch) match {
         case 0 =>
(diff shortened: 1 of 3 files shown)


djspiewak:
## Before

```
[info] Benchmark                                              (size)   Mode  Cnt     Score   Error    Units
[info] SleepBenchmark.sleep                                    10000  thrpt   10  1033.501 ± 2.298  ops/min
[info] WorkStealingBenchmark.alloc                           1000000  thrpt   10    13.890 ± 0.079  ops/min
[info] WorkStealingBenchmark.manyThreadsSchedulingBenchmark  1000000  thrpt   10    49.484 ± 1.040  ops/min
[info] WorkStealingBenchmark.runnableScheduling              1000000  thrpt   10  2693.215 ± 4.480  ops/min
[info] WorkStealingBenchmark.runnableSchedulingScalaGlobal   1000000  thrpt   10  2226.544 ± 4.643  ops/min
[info] WorkStealingBenchmark.scheduling                      1000000  thrpt   10    51.038 ± 2.970  ops/min
```

## After

```
[info] Benchmark                                              (size)   Mode  Cnt     Score   Error    Units
[info] SleepBenchmark.sleep                                    10000  thrpt   10  1045.840 ± 2.665  ops/min
[info] WorkStealingBenchmark.alloc                           1000000  thrpt   10    13.560 ± 0.135  ops/min
[info] WorkStealingBenchmark.manyThreadsSchedulingBenchmark  1000000  thrpt   10    44.617 ± 2.187  ops/min
[info] WorkStealingBenchmark.runnableScheduling              1000000  thrpt   10  2956.029 ± 4.047  ops/min
[info] WorkStealingBenchmark.runnableSchedulingScalaGlobal   1000000  thrpt   10  2228.159 ± 8.080  ops/min
[info] WorkStealingBenchmark.scheduling                      1000000  thrpt   10    50.958 ± 1.996  ops/min
```

So this actually *improves* performance, I think mostly due to the fact that we only look at `sleepers` in `state >= 4` now.

tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
```diff
@@ -321,6 +321,34 @@ trait IOPlatformSpecification { self: BaseSpec with ScalaCheck =>
...
+        val (pool, shutdown) = IORuntime.createWorkStealingComputeThreadPool(threads = 1)
+
+        implicit val runtime: IORuntime = IORuntime.builder().setCompute(pool, shutdown).build()
+
+        try {
+          val test = IO.sleep(1.milli) *> IO.pure(true)
+          test.unsafeRunTimed(1.second) must beSome(true)
+        } finally {
+          runtime.shutdown()
+        }
```
armanbilge: Sorry, I am a bit confused by the tests. Are they identical, except for the choice of durations? In which case, that suggests these are "special numbers", but I don't really see why 😕

armanbilge:
FS2 appears to be fixed! :)

armanbilge (commented):
As far as I can tell the fix looks good and is actually pretty straightforward to understand. FS2 is happy too. So mostly I'm just confused about the tests.

tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
djspiewak: Ah I should write some comments explaining the differences here. The numbers aren't particularly *magical*, they're just attempting to reproduce conditions inside the state machine. The shorter sleep is more likely to be caught by the state machine *before* parking, while the longer sleep is guaranteed to park the worker. Since these are technically two subtly independent code paths, I wanted to test them independently. For a time, one of these tests was failing while the other passed, and then I was able to invert that state, so they do indeed reproduce different scenarios.

armanbilge (approved):

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -584,13 +592,20 @@ private final class WorkerThread(
...
+
+              // After the worker thread has been unparked, look for work in the
+              // external queue.
+              state = 3
             } else {
-              parkUntilNextSleeper()
+              if (parkUntilNextSleeper()) {
+                // we made it to the end of our sleeping, so go straight to local queue stuff
+                pool.transitionWorkerFromSearching(rnd)
+                state = 4
```
mr-git, line 603: random reader here: why `state` is "named" with magical numbers instead of constants? I found descriptions in header, but it doesn't have as good mnemonics as names would.

armanbilge: Yeah, when I was reviewing this I was wishing the same. It's a technique used in the codebase to make sure that optimal bytecode is generated.
https://github.com/typelevel/cats-effect/blob/c8db3da04768571755a5b4c26fe551d3ec2556ff/core/shared/src/main/scala/cats/effect/IOFiber.scala#L240-L244

mr-git: I understand that it requires more work, but something like:
```
state = 3 // UNPARKED_AND_LOOKING_FOR_WORK
state = 4 // LOOKING_FOR_FIBERS_TO_EXECUTE
```
could help next reader, but would required a bit of perfectionism to implement.

I didn't notice these lines, I didn't know about so sever effect.
https://github.com/djspiewak/cats-effect/blob/0dd30f9c7f8929a9ec0db4598270f0daad71a8ab/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala#L306-L309

thanks for clarification!

PR #3444, opened 2023-02-20
Demonstrate and try to fix a possible bug with a "lost" fiber

### LINKED ISSUE #3392: CPU starvation causes problems with scheduling fairness (compared to CE2)
How to reproduce: run this code on CE2 and CE3 multiple times and measure how much it takes to complete:

```scala
import cats.effect._
import cats.syntax.all._

import java.time.{Instant, ZoneId, ZoneOffset}
import java.time.format.DateTimeFormatter

  val formatter = DateTimeFormatter.ofPattern("HH:mm:ss:SSS").withZone(ZoneId.from(ZoneOffset.UTC))

  def log(msg: String): IO[Unit] = IO.delay {
    val time = formatter.format(Instant.now)
    println(s"$time $msg")
  }

  val workflow1: IO[Unit] = for {
    _ <- log("workflow 1 started")
    _ <- (1 to 7_500).toVector.traverse { _ => IO.sleep(1.milliseconds) }
    _ <- log("workflow 1 completed")
  } yield ()

  val workflow2: IO[Unit] = for {
    _ <- (1 to 500_000).toVector.parTraverse { _ => IO.sleep(1.second) }
  } yield ()

  val run: IO[Unit] = for {
    f <- workflow1.start
    _ <- workflow2.foreverM.start
    _ <- f.join
  } yield ()
}
```

On my machine it takes 20-30 seconds to complete with CE2 and 5 minutes with CE3.

durban:
This might be related to #3392, ~~or may be something else entirely~~ (I couldn't reproduce that one, so I can't check).

#### The problem:

First 2 commits (f2044679c and b39cdd0a5) demonstrate the possible problem with a quick and dirty program; run it with `coreJVM/Test/run` (recommended to `set core.jvm/Test/fork := true` before, because it needs killing). This will print a number every second. In some runs it will be always increasing (this is correct):
```
[info] running (fork) cats.effect.unsafe.WstpSpec 
[info] Hello!
[info] 9682305
[info] 21700806
[info] 33978091
...
```

But sometimes it will always print the same number (this is incorrect):
```
[info] running (fork) cats.effect.unsafe.WstpSpec 
[info] Hello!
[info] 120
[info] 120
[info] 120
[info] 120
[info] 120
[info] 120
...
```

There is a fiber `fib1` which should _forever_ increment the number. So this probably means, that for some reason this fiber is not scheduled (any more). Heapdump shows references to this fiber from the main thread (the `fib1` local) and a `_join` (probably the joining fiber in line 33) and a circle through probably its own `_cancel`.

#### What I _think_ happens:

After the forever incrementing fiber autocedes (filling `cedeBypass`), the `WorkerThread` can go to state `0`, and deque a `blocking` task from the external queue. It runs this `blocking`, and passes its internal structures to a new/cached thread to take its place. However: its `cedeBypass` is "logically" part of the local queue, but it is not passed to the replacement thread. After doing the blocking, it becomes cached, and (in the test after 3 seconds) it shuts itself down. Thus, its `cedeBypass` becomes forever "lost".

#### A possible/incomplete fix:
The 3rd and 4th commit (b9dabc59e and 9e106d9c9) tries to fix this problem. The `WorkerThread` before passing its things to the replacement thread, checks if there is a `cedeBypass`, if yes, it enqueues it to the local queue, thus safely(?) passing it onto the replacement thread.

This _seems_ to fix the problem in the simple demonstration, ~~but I did not do a lot of testing otherwise.~~

One possible problem with the fix, that this doesn't make the `cedeBypass` having higher priority than the external queue; I don't know if that's needed.


core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -633,6 +633,14 @@ private final class WorkerThread(
       // Spawn a new `WorkerThread` to take the place of this thread, as the
       // current thread prepares to execute a blocking action.
 
+      // We'll transfer our local queue to the new/cached thread;
+      // don't forget to also transfer our cede bypass (if any):
+      val bypass = cedeBypass
+      if (bypass ne null) {
+        queue.enqueue(bypass, external, random)
+        cedeBypass = null
+      }
```
vasilmkd, line 642: Does this branch ever get executed?

The reason why I'm asking is the following.

`cedeBypass` should only be populated _after_ a fiber's `IOFiber#run` method _exits_, and it should already be set to `null` again _before_ running `IOFiber#run` again. At least, this should be true in theory.

On the other hand, `WorkerThread#blockOn` could _only_ be executed during a `IOFiber#run`, so at this point, `cedeBypass` would always be `null`.

I'm pretty sure if this weren't the case, we would be seeing other bugs as well, but I cannot think of any examples at the moment, I'm a bit rusty on all details.

durban: I think yes. As far as I can tell, a `Fiber#run` is executed here without setting `cedeBypass` to `null`: https://github.com/typelevel/cats-effect/blob/series/3.4.x/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala#L389. I think this happens in the example program.

durban: (Or maybe line 403 instead.)

durban: @vasilmkd I've verified (with a quick `println`) that the `if` block is indeed executed when running the example program.

vasilmkd: That's great, thanks.

vasilmkd:
I'm inclined to say this is the root of the bug, yes. I'm trying to think where to fix it exactly.

The other alternative is to modify `WorkerThread#reschedule` to also take into consideration the `state & ExternalQueueTicksMask` value and only mutate the `cedeBypass` if safe to do so (if we're sure we're not about to inspect the other queues).

vasilmkd:
Great work @durban. Thank you so much.

vasilmkd:
> The other alternative is to modify WorkerThread#reschedule to also take into consideration the state & ExternalQueueTicksMask value and only mutate the cedeBypass if safe to do so (if we're sure we're not about to inspect the other queues).

This most likely won't work because `state` is a local variable in `run`.

core/jvm/src/test/scala/cats/effect/unsafe/WstpSpec.scala
```diff
@@ -0,0 +1,51 @@
+package cats.effect
+package unsafe
```
djspiewak: Just to keep sanity, can we move this to `testsJVM`? We have a few `main` classes there.

djspiewak (dismissed):
Outstanding work. Genuinely. I'm publishing a snapshot right now to test with the OG reproduction.

vasilmkd:
The other _real_ alternative solution is to keep `reschedule` as it is, and explicitly boot `cedeBypass` to the local queue in `case 0`. All other cases should be covered because they originate from a failed dequeue in `case _` or after being woken up from sleep.

djspiewak:
I think the performance impact of that in the happy path would be slightly higher than this approach, which only penalizes `blockOn`

djspiewak:
Published this branch as `3.4.7-34-defc3a4`

vasilmkd:
This fix is great and we most likely should go through with it, pending some more execution time just in case in an ec2 instance.

However, we should probably reexamine whether `cedeBypass` is actually worth the maintenance headaches, compared to the small benchmark win that it provides in return. I'm personally not comfortable that it is knowingly breaking (my personal) expectations of when it is set and unset.

Furthermore, the worker thread run loop has gained a few new features, which could make the benchmark win with `cedeBypass` irrelevant.

vasilmkd (dismissed):

core/jvm/src/test/scala/cats/effect/unsafe/WstpSpec.scala
durban: I've moved the file, added a header and a comment.

djspiewak:
#3445

durban:
The CI failure seems unrelated (`IOAppSpec`).

djspiewak:
> The CI failure seems unrelated (IOAppSpec).

Yeah we have some sneaky-flaky tests right now. Need to fix.

djspiewak:
Okay so thanks to @vasilmkd's creative experimentation, we have really good evidence that this bug is indeed happening in #3392. I have a couple EC2 instances churning on reproductions with various patches, but we're pretty certain this is the problem. Outstanding work, @durban!

djspiewak (dismissed):

djspiewak:
I swapped the `main` for an automated test. It actually wasn't that hard to reproduce in this fashion, though notably if you formulate the whole thing as an `IO`, it doesn't work.

djspiewak (approved):

CremboC:
We're pretty sure this bug caused many of our production services to constantly freeze up. The services were `fs2-kafka` consumers, but notably they would only freeze up after implementing a `groupWithin` and a `commitBatch` logic. The symptoms were suddenly drop to 0% CPU, but no error raised whatsoever. Even Kafka itself reported the consumer as "healthy" (implying heartbeats were still being sent).

We've deployed this CE version to multiple consumers and report back.

djspiewak:
@CremboC That definitely sounds like a smoking gun to me! Please do report back with your experiences, positive or (even more importantly!) negative.

CremboC:
@djspiewak unfortunately it seems like it happened again, even with 3.4.8. It just stopped consuming from Kafka and didn't recover (although I have implemented an automate restarting if it does get stuck, so maybe it would have recovered..?)

djspiewak:
Concerning. There may be another issue here. Can you open a new ticket? The PR definitely fixes this reproducer (and a legit bug). I think the first step would be to try to get a fiber dump while your consumer is hung

PR #3446, opened 2023-02-21
Replace ArrayBlockingQueue with LinkedTransferQueue when transferring…

vasilmkd:
… data structures between blocked and cached worker threads

This change is functionally identical, but better conveys the intent of the data structure transfer procedure.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
@@ -26,7 +26,7 @@ import scala.concurrent.{BlockContext, CanAwait}
 import scala.concurrent.duration.Duration
 import scala.util.control.NonFatal
 
-import java.util.concurrent.{ArrayBlockingQueue, ThreadLocalRandom}
+import java.util.concurrent.{LinkedTransferQueue, ThreadLocalRandom}
 import java.util.concurrent.atomic.AtomicBoolean
 import java.util.concurrent.locks.LockSupport
 
@@ -95,7 +95,7 @@ private final class WorkerThread(
    */
   private[this] var _active: Runnable = _
 
-  private val indexTransfer: ArrayBlockingQueue[Integer] = new ArrayBlockingQueue(1)
+  private val indexTransfer: LinkedTransferQueue[Integer] = new LinkedTransferQueue()
   private[this] val runtimeBlockingExpiration: Duration = pool.runtimeBlockingExpiration
 
   val nameIndex: Int = pool.blockedWorkerThreadNamingIndex.incrementAndGet()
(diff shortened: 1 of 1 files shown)


armanbilge:
Does this have any performance implications? Scala Native multi-threading already supports `ArrayBlockingQueue`, but not `LinkedTransferQueue` yet, so this would be one more concurrent data structure to re-implement (or workaround) before that can become Real.:tm: Yeah, yeah, annoying but 🥺

vasilmkd:
I didn't know about that limitation. Sorry. This PR is mostly cosmetic and not necessary.

armanbilge:
No problem! Honestly, I'm amazed how soon SN multithreading will be landing. If you are curious, here is current progress on porting `java.util.concurrent`.
- https://github.com/scala-native/scala-native/issues/3165

djspiewak:
We can revert to ABQ if SN hasn't caught up on porting `TransferQueue` when we port WSTP to native. I'd rather keep it on JVM at present though.

djspiewak (approved):

armanbilge:
@vasilmkd @djspiewak I think we might want to revert this change.

> This change is functionally identical, but better conveys the intent of the data structure transfer procedure.

I'm not sure that this is correct. The problem is that `LinkedTransferQueue#transfer` blocks until there is a matching `take`. This means that the transitioning thread cannot begin working on the blocking task until its replacement worker comes fully online and takes over its state.

Meanwhile, `ArrayBlockingQueue#offer` can stash the state in the queue and immediately move on. The replacement thread can retrieve it at its leisure.

> Does this have any performance implications?

So, it seems like it probably does 😕 

h/t @CJSmith-0141 whose testing on https://github.com/typelevel/cats-effect/commit/6e9a1213febcabd0ff29c2cac0d84fd597d47f95 pointed us at this issue.

He-Pin:
Iirc, the linked transfer queue is unbouned

armanbilge:
@He-Pin that's correct, but the behavior is that `transfer` blocks until there is a recipient.

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/LinkedTransferQueue.html#transfer-E-

> Transfers the element to a consumer, waiting if necessary to do so.
More precisely, transfers the specified element immediately if there exists a consumer already waiting to receive it (in [`take()`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/LinkedTransferQueue.html#take--) or timed [`poll`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/LinkedTransferQueue.html#poll-long-java.util.concurrent.TimeUnit-)), else inserts the specified element at the tail of this queue and waits until the element is received by a consumer.

Easy to test with the following code:
```scala
import java.util.concurrent.*

@main def main =
  val queue = new LinkedTransferQueue
  queue.transfer(new Object)
  println("done!")
```

He-Pin:
then transfer will never work as arrayblockqueue with size 1

I think ce can implemented one based on aqs , as ce only use one or two methods

CJSmith-0141:
By `aqs` do you mean [AbstractQueuedSynchronizer](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/locks/AbstractQueuedSynchronizer.html)?

He-Pin:
Yes

PR #3451, opened 2023-02-22
Add Recipe presenting real world use case

Wosin:
I was thinking about something that presents a concept in a simple manner but also captures the essence of update based on the effectful result, but not too sure if this example is the best.

### DIFF
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -121,11 +121,11 @@ To better present real-life use-case scenario, let's imagine that we have a `Ser
 
 ```scala mdoc:silent
 case class Response(exchangeRate: Double)
-trait Service{
+trait Service {
   def query(): IO[Response]
 }
 
-object StubService extends Service{
+object StubService extends Service {
   override def query(): IO[Response] = Random
     .scalaUtilRandom[IO]
     .flatMap(random => random.nextDouble)
@@ -140,8 +140,8 @@ Now, say that we want to have a cache that holds the highest exchange rate that 
 class MaxProxy(atomicCell: AtomicCell[IO, Double], requestService: Service) {
 
   def queryCache(): IO[Response] = {
-    atomicCell evalModify (current => {
-      requestService.query().map(result => 
+    atomicCell evalModify { current => 
+      requestService.query().map { result => 
         if (result.exchangeRate > current) 
           (result.exchangeRate, result) 
         else 
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -117,51 +117,40 @@ class Server(atomicCell: AtomicCell[IO, Int]) {
 }
 ```
 
-Here is a larger example that shows multiple `Worker` instances modifying `AtomicCell` in concurrent fashion:
+To better present real-life use-case scenario, let's imagine that we have a `Service` that performs some HTTP request to external service holding exchange rates over time:
 
 ```scala mdoc:silent
-class Worker(workerId: Int, atomicCell: AtomicCell[IO, Int]) {
-  def update(input: Int): IO[Int] =
-    IO.pure(input + 1 )
+case class Response(exchangeRate: Double)
+trait Service{
+  def query(): IO[Response]
+}
 
-  private def putStrValue(value: Int): IO[Unit] =
-   IO.blocking(println(s"Worker #$workerId >> $value"))
+object StubService extends Service{
+  override def query(): IO[Response] = Random
+    .scalaUtilRandom[IO]
+    .flatMap(random => random.nextDouble)
+    .map(Response(_))
+}
+```
+To simplify we model the `StubService` to just return some random `Double` values.
 
-  def performUpdate: IO[Int] = {
-    for {
-      after <- atomicCell.evalUpdateAndGet{ current =>
-        putStrValue(current) *> update(current)
-      }
-      _ <- putStrValue(after)
-    } yield  after
+Now, say that we want to have a cache that holds the highest exchange rate that is ever returned by our service, we can have the proxy implementation based on `AtomicCell` like below:
+
+```scala mcdoc:silent
+class MaxProxy(atomicCell: AtomicCell[IO, Double], requestService: Service) {
+
+  def queryCache(): IO[Response] = {
+    atomicCell.evalModify(current => {
+      requestService.query().map(result => 
+        if(result.exchangeRate > current) 
+          (result.exchangeRate, result) 
+        else 
+          (current, result))
+    })
   }
 
-}
-object RefExample extends IOApp.Simple {
-
-  val run: IO[Unit] =
-    for {
-      cell <- AtomicCell[IO].of(0)
-      w1 = new Worker(1, cell)
-      w2 = new Worker(2, cell)
-      w3 = new Worker(3, cell)
-      _ <- List(
-        w1.performUpdate,
-        w2.performUpdate,
-        w3.performUpdate,
-      ).parSequence.void
-    } yield ()
+  def getHistoryMax(): IO[Double] = atomicCell.get
 }
 
-```
-
-The above code should yield results that look like below, note that the order of Workers may differ, but the actual results not:
-```
-Worker #1 >> 0
-Worker #1 >> 1
-Worker #2 >> 1
-Worker #2 >> 2
-Worker #3 >> 2
-Worker #3 >> 3
 ```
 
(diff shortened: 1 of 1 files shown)


docs/recipes.md
```diff
@@ -90,3 +90,61 @@ object Server extends IOApp.Simple {
...
+However, if we want to update our ref using result of an effect `Ref` will usually not be enough and we need a more powerful construct to achieve that. 
+In cases like that we can use the [AtomicCell](std/atomic-cell.md) that can be viewed as a synchronized `Ref`.
+
+The most typical example for `Ref` is the concurrent counter, but what if the update function for our counter would be effectful?
+
+Assume we have the following function:
+
+```scala mdoc:silent
+def update(input: Int): IO[Int] =
+  IO.defer(input + 1 )
```
djspiewak:
```suggestion
  IO(input + 1 )
```

:-) I think you meant `delay`, but `apply` is even more concise.

docs/recipes.md
```diff
@@ -90,3 +90,61 @@ object Server extends IOApp.Simple {
...
+```
+If we want to concurrently update a variable using our `update` function it's not possible to do it directly with `Ref`, luckily `AtomicCell` has `evalUpdate`:
+
+```scala mdoc:silent
+class Server(atomicCell: AtomicCell[IO, Int]) {
+  def update(input: Int): IO[Int] =
+    IO.defer(input + 1 )
```
djspiewak:
```suggestion
    IO(input + 1 )
```

docs/recipes.md
```diff
@@ -90,3 +90,61 @@ object Server extends IOApp.Simple {
...
+
+  def performUpdate(input: Int): IO[Int] = 
+    atomicCell.evalUpdate(i => update(i))
+}
+```
+So, let's assume that we want to implement very simple rate limiting proxy, that will only allow some specific number of requests and then return failure if this limit is surpassed.
+
+The code could look like below:
+```scala mdoc:silet
```
djspiewak:
```suggestion
```scala mdoc:silent
```

docs/recipes.md
```diff
@@ -90,3 +90,61 @@ object Server extends IOApp.Simple {
...
+}
+
+object StubService extends Service {
+  def query(request: Request): IO[Response] = IO.pure(Response("Request Succeeded."))
+}
+
+class RateLimiterProxy(limit: Int, cell: AtomicCell[IO, Int], service: Service){
+  def limit[A](request: IO[Request]): IO[Response] = {
+    cell.evalModify{ requests =>
+      if(requests >= limit)
```
djspiewak:
```suggestion
      if (requests >= limit)
```

docs/recipes.md
```diff
@@ -90,3 +90,61 @@ object Server extends IOApp.Simple {
...
+  def query(request: Request): IO[Response]
+}
+
+object StubService extends Service {
+  def query(request: Request): IO[Response] = IO.pure(Response("Request Succeeded."))
+}
+
+class RateLimiterProxy(limit: Int, cell: AtomicCell[IO, Int], service: Service){
+  def limit[A](request: IO[Request]): IO[Response] = {
+    cell.evalModify{ requests =>
```
djspiewak:
```suggestion
    cell evalModify { requests =>
```

djspiewak (changes requested):
Should we discuss the `modify` + `flatten` idiom as well?

Wosin:
Hey, I've actually found simpler example, let me know if that works, I think RateLimiter was a little farfetched.

docs/recipes.md
```diff
@@ -90,3 +90,67 @@ object Server extends IOApp.Simple {
...
+  def performUpdate(input: Int): IO[Int] = 
+    atomicCell.evalUpdate(i => update(i))
+}
+```
+
+To better present real-life use-case scenario, let's imagine that we have a `Service` that performs some HTTP request to external service holding exchange rates over time:
+
+```scala mdoc:silent
+case class Response(exchangeRate: Double)
+trait Service{
```
iRevive: To keep things consistent. 

```suggestion
trait Service {
```

docs/recipes.md
```diff
@@ -90,3 +90,67 @@ object Server extends IOApp.Simple {
...
+  def query(): IO[Response]
+}
+
+object StubService extends Service{
```
iRevive:
```suggestion
object StubService extends Service {
```

docs/recipes.md
```diff
@@ -90,3 +90,67 @@ object Server extends IOApp.Simple {
...
+```
+To simplify we model the `StubService` to just return some random `Double` values.
+
+Now, say that we want to have a cache that holds the highest exchange rate that is ever returned by our service, we can have the proxy implementation based on `AtomicCell` like below:
+
+```scala mcdoc:silent
+class MaxProxy(atomicCell: AtomicCell[IO, Double], requestService: Service) {
+
+  def queryCache(): IO[Response] = {
+    atomicCell evalModify (current => {
```
iRevive:
```suggestion
    atomicCell evalModify { current => 
```

docs/recipes.md
```diff
@@ -90,3 +90,67 @@ object Server extends IOApp.Simple {
...
+      requestService.query().map(result =>
```
iRevive: I do not have a strong opinion there. A curly bracket feels more idiomatic for a multiline block, but both options are valid.

```suggestion
      requestService.query().map { result => 
```

docs/recipes.md
```diff
@@ -90,3 +90,75 @@ object Server extends IOApp.Simple {
 
 In this example, `longRunningTask` is started in the background.
 The server returns to the client without waiting for the task to finish.
+
+## Atomically update a Ref with result of an effect
+
+Cats effect provides [Ref](std/ref.md), that we can use to model mutable concurrent reference.
```
djspiewak:
```suggestion
Cats Effect provides [Ref](std/ref.md), that we can use to model mutable concurrent reference. 
```

docs/recipes.md
```diff
@@ -90,3 +90,75 @@ object Server extends IOApp.Simple {
...
+However, if we want to update our ref using result of an effect `Ref` will usually not be enough and we need a more powerful construct to achieve that. 
+In cases like that we can use the [AtomicCell](std/atomic-cell.md) that can be viewed as a synchronized `Ref`.
+
+The most typical example for `Ref` is the concurrent counter, but what if the update function for our counter would be effectful?
+
+Assume we have the following function:
+
+```scala mdoc:silent
+def update(input: Int): IO[Int] =
+  IO(input + 1 )
```
djspiewak:
```suggestion
  IO(input + 1)
```

docs/recipes.md
```diff
@@ -90,3 +90,75 @@ object Server extends IOApp.Simple {
...
+To simplify we model the `StubService` to just return some random `Double` values.
+
+Now, say that we want to have a cache that holds the highest exchange rate that is ever returned by our service, we can have the proxy implementation based on `AtomicCell` like below:
+
+```scala mcdoc:silent
+class MaxProxy(atomicCell: AtomicCell[IO, Double], requestService: Service) {
+
+  def queryCache(): IO[ServiceResponse] = {
+    atomicCell evalModify { current =>
+      requestService.query().map { result =>
```
djspiewak: Just trying to be consistent on this one.

```suggestion
      requestService.query() map { result =>
```

djspiewak (changes requested):
Looks very good! Just a couple minor cleanups and then ready to go

docs/recipes.md
```diff
@@ -90,3 +90,75 @@ object Server extends IOApp.Simple {
...
+  IO(input + 1)
+```
+If we want to concurrently update a variable using our `update` function it's not possible to do it directly with `Ref`, luckily `AtomicCell` has `evalUpdate`:
+
+```scala mdoc:silent
+import cats.effect.std.AtomicCell
+
+class Server(atomicCell: AtomicCell[IO, Int]) {
+  def update(input: Int): IO[Int] =
+    IO(input + 1 )
```
iRevive:
```suggestion
    IO(input + 1)
```

docs/recipes.md
```diff
@@ -90,3 +90,75 @@ object Server extends IOApp.Simple {
...
+    .scalaUtilRandom[IO]
+    .flatMap(random => random.nextDouble)
+    .map(ServiceResponse(_))
+}
+```
+To simplify we model the `StubService` to just return some random `Double` values.
+
+Now, say that we want to have a cache that holds the highest exchange rate that is ever returned by our service, we can have the proxy implementation based on `AtomicCell` like below:
+
+```scala mcdoc:silent
```
iRevive: One last change I believe :) 
```suggestion
```scala mdoc:silent
```

djspiewak (approved):
I like this example quite a bit. Nicely done!

PR #3453, opened 2023-02-22
Corrected handling of self-cancelation within `timeout`

### LINKED ISSUE #3396: Timeout on a canceled IO
If you set a timeout on a task that somehow cancels itself, the end of the timeout will be awaited instead of handing back control immediately. Even more surprising (to me) is that `timeoutAndForget` never hands back control in that case.

```scala
scala> IO.canceled.unsafeRunTimed(4.seconds) // errors immediately
java.util.concurrent.CancellationException: The fiber was canceled

scala> IO.canceled.timeout(3.seconds).unsafeRunTimed(4.seconds) // errors after 3 seconds
java.util.concurrent.TimeoutException: 3 seconds

scala> IO.canceled.timeoutAndForget(3.seconds).unsafeRunTimed(4.seconds) // returns after 4 seconds
val res63: Option[Unit] = None
```

In order to avoid this I guess you would have to surface the cancellation as an `Exception`.

djspiewak:
This has to be targeted at 3.5.0 since it requires a law change. In particular, this adjusts the default race semantic such that `race(canceled, fa)` is now `canceled` rather than being equal to `fa`. I think this actually makes a lot more sense.

Fixes #3396

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
@@ -152,9 +152,9 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
     uncancelable { poll =>
       implicit val F: GenTemporal[F, E] = this
 
-      racePair(fa, sleep(duration)) flatMap {
+      poll(racePair(fa, sleep(duration))) flatMap {
         case Left((oc, f)) =>
-          poll(f.cancel *> oc.embedNever)
+          poll(f.cancel *> oc.embed(poll(F.canceled) *> F.never))
 
         case Right((f, _)) =>
           start(f.cancel) *> raiseError[A](ev(new TimeoutException(duration.toString)))
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
@@ -352,7 +352,7 @@ trait GenSpawn[F[_], E] extends MonadCancel[F, E] with Unique[F] {
             case Outcome.Succeeded(fa) => f.cancel *> fa.map(Left(_))
             case Outcome.Errored(ea) => f.cancel *> raiseError(ea)
             case Outcome.Canceled() =>
-              poll(f.join).onCancel(f.cancel).flatMap {
+              f.cancel *> f.join flatMap {
                 case Outcome.Succeeded(fb) => fb.map(Right(_))
                 case Outcome.Errored(eb) => raiseError(eb)
                 case Outcome.Canceled() => poll(canceled) *> never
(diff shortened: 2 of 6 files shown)


armanbilge (approved):
I wasn't sure about this initially, but https://github.com/typelevel/cats-effect/issues/3456 and the linked Discord discussion changed my mind.

This also addresses the problem I encountered in https://github.com/typelevel/cats-effect/pull/3232#issuecomment-1306535113. Embedding the cancelation does seem like the right default, with `racePair` available for more customized semantics.

kernel/shared/src/main/scala/cats/effect/kernel/GenTemporal.scala
```diff
@@ -152,9 +152,9 @@ trait GenTemporal[F[_], E] extends GenConcurrent[F, E] with Clock[F] {
     uncancelable { poll =>
       implicit val F: GenTemporal[F, E] = this
 
-      racePair(fa, sleep(duration)) flatMap {
+      poll(racePair(fa, sleep(duration))) flatMap {
         case Left((oc, f)) =>
-          poll(f.cancel *> oc.embedNever)
+          poll(f.cancel *> oc.embed(poll(F.canceled) *> F.never))
```
alexandru, line 157: I was just about to open a bug for the current version (3.4.8), as `timeoutAndForget` was uncancelable, but then saw that you already fixed it.

I'd like to ask, however, what is the meaning of this line here? And does it work on the current version?

This change is confusing. What's the difference between `oc.embed(poll(F.canceled) *> F.never)` and `oc.embedNever`? It feels like this change is needed due to some unintuitive interaction between the operations involved (`race`, `embed` and `poll`).

I am asking because we are migrating a big codebase from Cats-Effect v2, to Cats-Effect v3, and I need custom semantics for our timeout operations, which means our custom timeout implementation. And we need something workable right now.

Thanks for any insight you can give.

armanbilge: > What's the difference between `oc.embed(poll(F.canceled) *> F.never)` and `oc.embedNever`?

- `oc.embedNever` will hang indefinitely if `oc` is canceled
- `oc.embed(poll(F.canceled) *> F.never)` will self-cancel if `oc` is canceled

alexandru: Thank you kindly for your answer @armanbilge,

But what's the problem if `embedNever` hangs? Given it's in the context of `poll`, `IO.never` should be cancelable. This is the part that I'm missing.

alexandru: Sorry, if I'm generating noise on an issue right now, maybe this should have been asked on the forum.

I wondered if this isn't some mistake, or an API issue. But I may have gaps in how that `embed` / `fold` works. Intuitively, `embedNever` should have worked, but I guess it doesn't have access to that `poll`.

armanbilge: > But what's the problem if `embedNever` hangs?

Then we end up with problems like https://github.com/typelevel/cats-effect/issues/3396.

> Given it's in the context of `poll`, `IO.never` should be cancelable. This is the part that I'm missing.

Right, it's cancelable, but it will hang unless it is canceled. The semantic we decided we want here is for cancelation to _propagate_.

alexandru: @armanbilge thanks a lot, I think I get it.

PR #3454, opened 2023-02-22
Ensure that deadlock detection considers results availability

### LINKED ISSUE #3349: `isDeadLocked` on `TestControl` returns true if result is available
```scala
scala> TestControl.execute(IO.unit).flatMap(tc => tc.tickAll *> tc.isDeadlocked).unsafeRunSync()
val res9: Boolean = true
```

djspiewak:
Fixes #3349

### DIFF
--- a/testkit/shared/src/main/scala/cats/effect/testkit/TestControl.scala
+++ b/testkit/shared/src/main/scala/cats/effect/testkit/TestControl.scala
@@ -273,7 +273,7 @@ final class TestControl[A] private (
    * some interaction with an external asynchronous scheduler (such as another thread pool).
    */
   val isDeadlocked: IO[Boolean] =
-    IO(ctx.state.tasks.isEmpty)
+    IO(!_results.get().isDefined && ctx.state.tasks.isEmpty)
 
   /**
    * Returns the base64-encoded seed which governs the random task interleaving during each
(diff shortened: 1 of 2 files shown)


testkit/shared/src/main/scala/cats/effect/testkit/TestControl.scala
```diff
@@ -273,7 +273,7 @@ final class TestControl[A] private (
    * some interaction with an external asynchronous scheduler (such as another thread pool).
    */
   val isDeadlocked: IO[Boolean] =
-    IO(ctx.state.tasks.isEmpty)
+    IO(!_results.get().isDefined && ctx.state.tasks.isEmpty)
```
armanbilge, line 276:
```suggestion
    IO(_results.get().isEmpty && ctx.state.tasks.isEmpty)
```

armanbilge (approved):

testkit/shared/src/main/scala/cats/effect/testkit/TestControl.scala
djspiewak: I thought about that, but I actually kinda dislike the fact that there are two variants of `isDefined` on `Option`, and I tend to avoid the negative one.

PR #3455, opened 2023-02-22
Expand docs for Ref flatModify methods, and use them when appropriate

bplommer:
There's a lot of indentation changes, so easiest to review with "hide whitespace" enabled.

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
@@ -48,50 +48,48 @@ trait GenConcurrent[F[_], E] extends GenSpawn[F, E] {
       // start running the effect, or subscribe if it already is
       def evalOrSubscribe: F[A] =
         deferred[Fiber[F, E, A]] flatMap { deferredFiber =>
-          uncancelable { poll =>
-            state.modify {
-              case Unevaluated() =>
-                // run the effect, and if this fiber is still relevant set its result
-                val go = {
-                  def tryComplete(result: Memoize[F, E, A]): F[Unit] = state.update {
-                    case Evaluating(fiber, _) if fiber eq deferredFiber =>
-                      // we are the blessed fiber of this memo
-                      result
-                    case other => // our outcome is no longer relevant
-                      other
+          state.flatModifyFull {
+            case (poll, Unevaluated()) =>
+              // run the effect, and if this fiber is still relevant set its result
+              val go = {
+                def tryComplete(result: Memoize[F, E, A]): F[Unit] = state.update {
+                  case Evaluating(fiber, _) if fiber eq deferredFiber =>
+                    // we are the blessed fiber of this memo
+                    result
+                  case other => // our outcome is no longer relevant
+                    other
+                }
+
+                fa
+                  // hack around functor law breakage
+                  .flatMap(F.pure(_))
+                  .handleErrorWith(F.raiseError(_))
+                  // end hack
+                  .guaranteeCase {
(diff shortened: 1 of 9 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/Ref.scala
```diff
@@ -116,9 +122,12 @@ abstract class Ref[F[_], A] extends RefSource[F, A] with RefSink[F, A] {
   /**
    * Like [[modify]] but schedules resulting effect right after modification.
    *
-   * Unlike [[flatModify]] finalizer cancellation could be masked via supplied `Poll`.
+   * Unlike [[flatModify]] finalizer cancellation could be unmasked via supplied `Poll`.
    * Modification itself is still uncancelable.
    *
+   * When used as part of a state machine, cancelable regions should usually have an `onCancel`
+   * finalizer to update the state to reflect that the effect will not be performed.
```
armanbilge, line 129: Hm, interesting, I've never seen this pattern. Do you have any examples you can point to?

armanbilge (commented):
Thanks looks good, nice cleanup! Just a question about the docs :)

kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
```diff
@@ -48,50 +48,48 @@ trait GenConcurrent[F[_], E] extends GenSpawn[F, E] {
...
+              Evaluating(deferredFiber, 1) -> eval
 
-              case Evaluating(fiber, subscribers) =>
-                Evaluating(fiber, subscribers + 1) ->
-                  poll(fiber.get.flatMap(_.join).flatMap(_.embed(productR(canceled)(never))))
-                    .onCancel(unsubscribe(fiber))
+            case (poll, Evaluating(fiber, subscribers)) =>
+              Evaluating(fiber, subscribers + 1) ->
+                poll(fiber.get.flatMap(_.join).flatMap(_.embed(productR(canceled)(never))))
+                  .onCancel(unsubscribe(fiber))
```
armanbilge, line 89: Oh lol wait, this is an example isn't it. Embarrassingly I think I wrote this in https://github.com/typelevel/cats-effect/pull/3351 😇 go figure.

armanbilge (approved):

kernel/shared/src/main/scala/cats/effect/kernel/GenConcurrent.scala
bplommer: There's a lot more but I didn't want this PR to get too big - maybe I'll do a followup. (Actually, I'm curious if there are any cases where that pattern *isn't* used)

armanbilge: Sure, here we have a cancelable `flatModify` without an `onCancel`.
https://github.com/typelevel/fs2/blob/f6085648752a191ad72b0ad759f85d8ed6412631/core/shared/src/main/scala/fs2/concurrent/Signal.scala#L273-L283

bplommer: Interesting - ~I guess the idea is that it's more efficient to leave a "dead" listener registered than to unregister it?~ [EDIT: no, unregistration happens further up the call stack in `Stream.bracket`] I wonder if it's worth having a `flatModifyCancelable` method for cases like this (the `// cancelable` comment makes me think yes) - it's also applicable for cases such as finalizers that don't need explicit uncancelability because they're running in an uncancelable context.

I'm still curious if there are cases where `poll` is used without `onCancel` within an uncancelable region, though.

PR #3460, opened 2023-02-25
Added `cancelable`

djspiewak:
The unit test does a pretty good job of demonstrating the use case. Reproducing here in more condensed form:

```scala
IO.deferred[Unit] flatMap { latch =>
  val badness = latch.get.uncancelable
  badness.cancelable(latch.complete(()).void
}
```

Under normal circumstances, the above would be a fancy version of `never`. The use of `cancelable` converts this into something which has the natural semantics: canceling the effect completes the `Deferred` and unlocks the fiber. More usefully, this can be applied to effects constructed with `blocking` or `delay` to make them safely cancelable in the case where some state can be mutated to prematurely terminate the effect.

Generalizes #3459, as well as #3374


armanbilge (dismissed):
Thanks, this is _so_ much nicer than what I had 😁 🤦

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -363,6 +363,51 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
...
+   * Note that there is no free lunch here. If an effect truly cannot be prematurely terminated,
+   * `cancelable` will not allow for cancelation. As an example, if you attempt to cancel
+   * `uncancelable(_ => never)`, the cancelation will hang forever (in other words, it will be
+   * itself equivalent to `never`). Applying `cancelable` will not change this in any way. Thus,
+   * attempting to cancel `cancelable(uncancelable(_ => never), unit)` will ''also'' hang
+   * forever. As in all cases, cancelation will only return when all finalizers have run and the
+   * fiber has fully terminated.
+   *
+   * If the `IO` self-cancels and the `cancelable` itself is uncancelable, the resulting fiber
+   * will be equal to `never` (similar to [[race]]. Under normal circumstances, if `IO`
```
durban: There seem to be a `)` missing here.

kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala
```diff
@@ -216,6 +216,58 @@ trait GenSpawn[F[_], E] extends MonadCancel[F, E] with Unique[F] {
...
+   * Note that there is no free lunch here. If an effect truly cannot be prematurely terminated,
+   * `cancelable` will not allow for cancelation. As an example, if you attempt to cancel
+   * `uncancelable(_ => never)`, the cancelation will hang forever (in other words, it will be
+   * itself equivalent to `never`). Applying `cancelable` will not change this in any way. Thus,
+   * attempting to cancel `cancelable(uncancelable(_ => never), unit)` will ''also'' hang
+   * forever. As in all cases, cancelation will only return when all finalizers have run and the
+   * fiber has fully terminated.
+   *
+   * If `fa` self-cancels and the `cancelable` itself is uncancelable, the resulting fiber will
+   * be equal to `never` (similar to [[race]]. Under normal circumstances, if `fa` self-cancels,
```
durban: There seem to be a `)` missing here.

docs/faq.md
```diff
@@ -89,3 +89,79 @@ If you're seeing this warning and have not changed any of the default configurat
...
+    bytes.toArray
+  }
+
+IO.bracket(
+  IO(new FileInputStream(inputFile)))(
+  readBytes(_))(
+  fis => IO(fis.close()))
+```
+
+In the above snippet, swapping `blocking` for `interruptible` won't actually help, since `fis.read()` ignores `Thread` interruption! However, we can still *partially* resolve this by introducing a bit of external state:
```
durban, line 139: It might be worth mentioning, why this is a *partial* solution. (Is it because when blocked in `fis.read()`, it still can't be cancelled?)

durban (commented):
Related discussion: https://github.com/typelevel/cats-effect/discussions/2671.

SystemFw (dismissed):
This makes perfect sense to me. In fact, I have a vague memory of double checking this was encodable with `uncancelable` : the shape of `cancelable` being added here was the main combinator for cancelation in early versions of cats-effect

core/shared/src/main/scala/cats/effect/IO.scala
djspiewak: Nice catch!

docs/faq.md
djspiewak: Yeah I think this is worth diving into

docs/faq.md
```diff
@@ -164,4 +164,6 @@ IO(new AtomicBoolean(false)) flatMap { flag =>
 
 This is *almost* the same as the previous snippet except for the introduction of an `AtomicBoolean` which is checked on each iteration of the `while` loop. We then set this flag by using `cancelable` (right at the end), which makes the whole thing (safely!) cancelable, despite the fact that it was defined with `blocking` (note this also works with `delay` and absolutely everything else).
 
+It is still worth keeping in mind that this is only a partial solution. Whenever `fis.read()` is blocked, cancelation will wait for that blocking to complete regardless of how long it takes. In other words, we still can't interrupt the `read()` function any more than we could with `interruptible`. All that `cancelable` is achieving here is allowing us to encode a more bespoke cancelation protocol for a particular effect, in this case in terms of `AtomicBoolean`.
```
armanbilge, line 167: So maybe I missed the point here a bit ... but `readAll.cancelable(IO(fis.close()))` would make this a full solution wouldn't it?

djspiewak: …wow. I missed that. Yes it would. And actually this would be a meaningful upgrade for `Files`, which I believe right now has no mechanism for this. The downside is it makes this a somewhat less compelling example of `cancelable` *specifically*.

armanbilge: Yes, that's exactly what I was planning. That's why I wanted this whole thing to begin with 😛

docs/faq.md
```diff
@@ -166,4 +166,6 @@ This is *almost* the same as the previous snippet except for the introduction of
 
 It is still worth keeping in mind that this is only a partial solution. Whenever `fis.read()` is blocked, cancelation will wait for that blocking to complete regardless of how long it takes. In other words, we still can't interrupt the `read()` function any more than we could with `interruptible`. All that `cancelable` is achieving here is allowing us to encode a more bespoke cancelation protocol for a particular effect, in this case in terms of `AtomicBoolean`.
 
+> Note: It is actually possible to implement a *proper* solution for cancelation here simply by applying the `cancelable` to the inner `blocking` effect and defining the handler to be `IO(fis.close())`, in addition to adding some error handling logic to catch the corresponding exception. This is a special case however, specific to `FileInputStream`, and doesn't make for as nice of an example. :-)
```
armanbilge: Well maybe `IO.blocking(fis.close())` to be on the safe side :)

Also question:

> adding some error handling logic to catch the corresponding exception

Why is this needed, for something that is being canceled anyway? To avoid triggering the unhandled exception thingy?

djspiewak: > To avoid triggering the unhandled exception thingy?

Exactly. It also *feels* more composable. You're assuming less about external circumstances.

armanbilge (approved):

docs/faq.md
```diff
@@ -89,3 +89,83 @@ If you're seeing this warning and have not changed any of the default configurat
...
+IO.interuptible(Thread.sleep(1000)).timeout(100.millis)
+```
+
+The above will return in 100 milliseconds, raising a `TimeoutException` as expected.
+
+However, not *all* effects respect thread interruption. Notably, most things involving `java.io` file operations (e.g. `FileReader`) ignore interruption. When working with this type of effect, we must turn to other means. A simple and naive example:
+
+```scala
+def readBytes(fis: FileInputStream) = 
+  IO blocking {
```
johnynek, line 121: nit: seems you usually use `IO.blocking` no?

PR #3462, opened 2023-02-27
Added `traverse` FAQ entry

djspiewak:
I'm genuinely not sure how we didn't have this FAQ entry previously.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -65,7 +65,7 @@ val program: IO[List[Unit]] = names.traverse(name => IO.println(s"Hi, $name!"))
 
 In the above example, `traverse` will iterate over every element of the `List`, running the function `String => IO[Unit]` which it was passed, assembling the results into a resulting `List` wrapped in a single outer `IO`.
 
-> You *must* have `import cats.syntax.all_` (or `cats.syntax.traverse._`) in scope, otherwise this will not work! `traverse` is a method that is implicitly enriched onto collections like `List`, `Vector`, and such, meaning that it must be brought into scope using an import (unfortunately!).
+> You *must* have `import cats.syntax.all_` (or `import cats.syntax.traverse._`) in scope, otherwise this will not work! `traverse` is a method that is implicitly enriched onto collections like `List`, `Vector`, and such, meaning that it must be brought into scope using an import (unfortunately!).
 
 Of course, in the above example, we don't really care about the results, since our `IO` is constantly producing `Unit`. This pattern is so common that we have a special combinator for it: `traverse_`.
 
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -49,6 +49,68 @@ Hello world
...
+import cats.effect._
+import cats.syntax.all._
+
+val names = List("daniel", "chris", "joseph", "renee", "bethany", "grace")
+val program: IO[List[Unit]] = names.traverse(name => IO.println(s"Hi, $name!"))
+```
+
+In the above example, `traverse` will iterate over every element of the `List`, running the function `String => IO[Unit]` which it was passed, assembling the results into a resulting `List` wrapped in a single outer `IO`.
+
+> You *must* have `import cats.syntax.all_` (or `import cats.syntax.traverse._`) in scope, otherwise this will not work! `traverse` is a method that is implicitly enriched onto collections like `List`, `Vector`, and such, meaning that it must be brought into scope using an import (unfortunately!).
```
durban: Missing `.` (in `all._`).

docs/faq.md
```diff
@@ -49,6 +49,68 @@ Hello world
...
+val program: IO[Unit] = names.traverse_(name => IO.println(s"Hi, $name!"))
+```
+
+This efficiently discards the results of each `IO` (which are all `()` anyway), producing a single `IO[Unit]` at the end.
+
+This type of pattern generalizes extremely well, both to actions which *do* return results, as well as to more advanced forms of execution, such as parallelism. For example, let's define and use a `fetchUri` method which uses [Http4s](https://http4s.org) to load the contents of a URI. We will then iterate over our list of `names` *in parallel*, fetching all of the contents into a single data structure:
+
+```scala
+import cats.effect._
+import cats.syntax.all_
```
durban: `all._`

docs/faq.md
```diff
@@ -49,6 +49,68 @@ Hello world
...
+
+```scala mdoc:reset:silent
+import cats.effect._
+import cats.syntax.all._
+
+val names = List("daniel", "chris", "joseph", "renee", "bethany", "grace")
+val program: IO[List[Unit]] = names.traverse(name => IO.println(s"Hi, $name!"))
+```
+
+In the above example, `traverse` will iterate over every element of the `List`, running the function `String => IO[Unit]` which it was passed, assembling the results into a resulting `List` wrapped in a single outer `IO`.
```
diesalbla: Somewhere it should mention that results are assembled _preserving the order_, in the sense that each value in the result list is itself the result of applying the function to the value from the same position at the input list.

PR #3465, opened 2023-02-28
Make `Console#readLine` cancelable

### LINKED ISSUE #3077: `IO.readLine.timeoutTo(...)` not cancelling the `readLine` after timeout.

armanbilge:
Fixes https://github.com/typelevel/cats-effect/issues/3077. This change is source-breaking, because it upgrades the constraint on `Console` from `Sync` to `Async`.

The fix is inspired by https://github.com/typelevel/fs2/pull/3093#issuecomment-1366256942. We now run all `readLineWithCharset` operations on a dedicated daemon thread, that is lazily started.

This `stdinReader` runs a simple state machine. It waits for a `ReadLineRequest`, then it reads a line from stdin, then it completes the request callback. If the request was canceled in the meantime, then it saves this line for the next request, which might already be pending.

This implementation never drops a line. If it has a line cached from a previous call that was canceled, the next request will either be completed with said line, or it will be failed with an `IllegalStateException` if the charset does not match the charset the cached line was read with.

I manually verified that this fixes the example given in https://github.com/typelevel/cats-effect/issues/3077, as well as a couple other examples that I played with. I also verified that the `cats-effect-stdin` thread only appears if you specifically touch `readLine`.

### DIFF
--- a/std/jvm/src/main/scala/cats/effect/std/ConsoleCompanionPlatform.scala
+++ b/std/jvm/src/main/scala/cats/effect/std/ConsoleCompanionPlatform.scala
@@ -1,0 +1,103 @@
+/*
+ * Copyright 2020-2023 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.std
+
+import cats.effect.kernel.{Async, Sync}
+import cats.syntax.all._
+
+import java.nio.charset.Charset
+import java.util.concurrent.LinkedTransferQueue
+
+private[std] trait ConsoleCompanionPlatform extends ConsoleCompanionCrossPlatform {
+
+  /**
+   * Constructs a `Console` instance for `F` data types that are [[cats.effect.kernel.Async]].
+   */
+  def make[F[_]](implicit F: Async[F]): Console[F] =
+    new AsyncConsole[F]
+
+  @deprecated("Use overload with Async constraint", "3.5.0")
+  def make[F[_]](F: Sync[F]): Console[F] = F match {
+    case async: Async[F] => make(async)
+    case _ => new SyncConsole()(F)
+  }
+
+  private final class AsyncConsole[F[_]](implicit F: Async[F]) extends SyncConsole[F] {
+    override def readLineWithCharset(charset: Charset): F[String] =
+      F.async[String] { cb =>
+        F.delay(stdinReader.readLineWithCharset(charset, cb)).map { cancel =>
+          Some(F.delay(cancel.run()))
+        }
+      }
+  }
+
+  private object stdinReader extends Thread {
+
+    private final class ReadLineRequest(
+        val charset: Charset,
+        @volatile var callback: Either[Throwable, String] => Unit
+    ) extends Runnable {
+      def run() = callback = null
+    }
+
+    private[this] val requests = new LinkedTransferQueue[ReadLineRequest]
+
+    def readLineWithCharset(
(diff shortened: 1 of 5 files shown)


std/jvm/src/main/scala/cats/effect/std/ConsoleCompanionPlatform.scala
```diff
@@ -0,0 +1,103 @@
...
+private[std] trait ConsoleCompanionPlatform extends ConsoleCompanionCrossPlatform {
+
+  /**
+   * Constructs a `Console` instance for `F` data types that are [[cats.effect.kernel.Async]].
+   */
+  def make[F[_]](implicit F: Async[F]): Console[F] =
+    new AsyncConsole[F]
+
+  @deprecated("Use overload with Async constraint", "3.5.0")
+  def make[F[_]](F: Sync[F]): Console[F] = F match {
```
armanbilge, line 34: I know, I know, v3.5.0 shop is closed. But ... we could fix this beginner stumbling block once and for all 😇 

No strong feelings 😁

armanbilge: Oh actually, we _can_ fix this source-compatibly and get it into 3.5.x. We can't expose the new instance publicly without a source-breaking change, but we _can_ replace the instance for `IO` right now, and do the source-breaking change in 3.6. This gets us most of the win.

std/jvm/src/main/scala/cats/effect/std/ConsoleCompanionPlatform.scala
```diff
@@ -0,0 +1,103 @@
...
+        if (line eq null) { // need a line for current request
+          charset = request.charset // remember the charset we used
+          line = Either.catchNonFatal(readLineWithCharsetImpl(charset))
+          // we just blocked, so loop to possibly freshen current request
+        } else { // we have a request and a line!
+          if (request.charset != charset) { // not the charset we have :/
+            callback(Left(new IllegalStateException(s"Next read must be for $charset line")))
+            request = null // loop to get a new request that can handle this charset
+          } else { // happy days!
+            callback(line)
```
armanbilge, line 91: Something that is nagging at me is that this is a leak here. The callback may have already be canceled, but we are invoking it anyway. There seems no reliable way to know for sure.

I don't see any way to plug this without changing the callback in `async { cb => ` to return a `Boolean` indicating whether it completed successfully or not. I feel like we've discussed this before.

std/jvm/src/main/scala/cats/effect/std/ConsoleCompanionPlatform.scala
```diff
@@ -0,0 +1,103 @@
...
+
+import java.nio.charset.Charset
+import java.util.concurrent.LinkedTransferQueue
+
+private[std] trait ConsoleCompanionPlatform extends ConsoleCompanionCrossPlatform {
+
+  /**
+   * Constructs a `Console` instance for `F` data types that are [[cats.effect.kernel.Async]].
+   */
+  def make[F[_]](implicit F: Async[F]): Console[F] =
```
djspiewak, line 30: Isn't this now the same constraint on both JS and JVM? Do we *need* the platform split?

djspiewak (approved):
Life is pain.

std/jvm/src/main/scala/cats/effect/std/ConsoleCompanionPlatform.scala
armanbilge: Well, at the moment there's still Native with the `Sync` constraint. I can revisit that when we get polling system and/or Scala Native multithreading.

PR #3467, opened 2023-02-28
IOFiber#toString

### LINKED ISSUE #3463: `IOFiberSpec` is still flaky in CI
```
[error]   x toString a completed fiber
[2135](https://github.com/typelevel/cats-effect/actions/runs/4284923854/jobs/7462682986#step:23:2136)
[error]    'cats.effect.IOFiber@4ad52419 RUNNING' doesn't match 'cats.effect.IOFiber@[0-9a-f][0-9a-f]+ COMPLETED' (IOFiberSpec.scala:68)
[2136](https://github.com/typelevel/cats-effect/actions/runs/4284923854/jobs/7462682986#step:23:2137)
```

We should consider disabling it in CI or something because we can't have jobs randomly failing like this.

durban:
Hopefully fixes #3463.

### DIFF
--- a/tests/shared/src/test/scala/cats/effect/IOFiberSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/IOFiberSpec.scala
@@ -49,7 +49,7 @@ class IOFiberSpec extends BaseSpec with DetectPlatform {
         val pattern =
           raw"cats.effect.IOFiber@[0-9a-f][0-9a-f]+ SUSPENDED(: release @ fiber.IOFiberSpec.completeLatch\$$[0-9]\(((.*IOFiberSpec.scala:[0-9]{2})|(Unknown Source))\))?"
         for {
-          latch <- CountDownLatch[IO](1)
+          latch <- CountDownLatch[IO](1) // `Deferred` makes the trace non-deterministic
           f <- foreverNever(latch).start
           _ <- latch.await
           _ <- IO.sleep(1.milli)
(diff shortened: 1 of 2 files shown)


tests/shared/src/test/scala/cats/effect/IOFiberSpec.scala
```diff
@@ -39,14 +41,19 @@ class IOFiberSpec extends BaseSpec with DetectPlatform {
       }
 
       "toString a suspended fiber" in real {
-        def foreverNever = IO.async_[Unit](_ => ())
+        // these are separate methods to have them in the trace:
+        def completeLatch(latch: CountDownLatch) =
+          IO.apply(latch.countDown())
+        def foreverNever(latch: CountDownLatch) =
+          IO.async[Unit](_ => completeLatch(latch).as(Some(IO.unit)))
```
armanbilge: Can we use a `cats.effect.std.CountdownLatch`?

durban: We could even use a `Deferred`. But it messes up the trace. But CDL clearly doesn't work (CI failure), so... I'll try `std.CountDownLatch`...

armanbilge: Oh right! How does `Deferred` mess up the trace? As long as it still says `SUSPENDED` that's what matters, the current test may be too particular.

durban: Yeah, the `SUSPENDED` is okay; the method name is non-deterministically incorrect. But I've removed the latch completely, it doesn't seem to actually help...

armanbilge (approved):
Nice, thanks so much for quashing this one!!

djspiewak (approved):

Issue #3468, opened 2023-02-28
`unsafeRunAndForget()` doesn't log errors


armanbilge:
https://github.com/typelevel/cats-effect/blob/a9eac6cf8685a43bb45db2d4ce35a387a7ec8c52/core/shared/src/main/scala/cats/effect/IO.scala#L937-L940

```scala
//> using lib "org.typelevel::cats-effect::3.5.0-RC3"

import cats.effect.*
import cats.effect.unsafe.implicits.*

@main def main =
  IO.raiseError(new RuntimeException).unsafeRunAndForget()
  Thread.sleep(1000)
```

```
[Hello darkness, my old friend]
```

PR #3472, opened 2023-03-01
debugging notes

samspills:
- #3468 
## Running
I've been testing with scala 3. I've been running this by:
- changing the version in build.sbt to `3.5-sam-SNAPSHOT`
- in sbt change the scala version to 3.2.2
- running `rootJVM/publishLocal` 
- using scala-cli to run Arman's example from the issue
```scala
//> using scala "3.2.2"
//> using lib "org.typelevel::cats-effect::3.5-sam-SNAPSHOT"

import cats.effect.*
import cats.effect.unsafe.implicits.*

@main def main =
  IO.raiseError(new RuntimeException).unsafeRunAndForget()
  Thread.sleep(1000)
```

## Notes
- added printlns all over the place
- tried to use them to follow my way down the worker thread
  - inside `run` on worker thread we end up in case 3: worker looking for work in external queue
    - in case 3 we find an element, which is a single runnable (vs. array of runnable)
    - the runnable is an IOFiber, and WT calls `fiber.run()`
- however the `try fiber.run()` DOES NOT cause me to see the println that I put in the fiber.run method
- also don't see anything coming out of fiber `done` method

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
@@ -510,17 +510,21 @@ private[effect] final class WorkStealingThreadPool(
    *   the runnable to be executed
    */
   override def execute(runnable: Runnable): Unit = {
+    println(s"SRP WSTP: inside execute")
     val pool = this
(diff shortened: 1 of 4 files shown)


samspills:
sorry meant to just open a fork PR :)

PR #3475, opened 2023-03-02
report failures during unsafeRunAndForget

### LINKED ISSUE #3468: `unsafeRunAndForget()` doesn't log errors

samspills:
We noticed in #3468 that errors were being swallowed by `unsafeRunAndForget()`. After exploring, I found that there is a callback invoked for the fiber in this case, which causes it to bypass the error reporting in `IOFiber.done`:
https://github.com/typelevel/cats-effect/blob/a9eac6cf8685a43bb45db2d4ce35a387a7ec8c52/core/shared/src/main/scala/cats/effect/IOFiber.scala#L1024

This PR updates `unsafeRunAndForget` to report non-fatal errors, if the runtime is configured to do so, and also prints the stack trace of fatal errors.

resolves #3468

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -865,7 +865,17 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
    * should never be totally silent.
    */
   def unsafeRunAndForget()(implicit runtime: unsafe.IORuntime): Unit = {
-    val _ = unsafeRunFiber((), _ => (), _ => ())
+    val _ = unsafeRunFiber(
+      (),
+      t => {
+        if (NonFatal(t) && runtime.config.reportUnhandledFiberErrors) {
+          runtime.compute.reportFailure(t)
+        } else {
+          // fatal failure
+          t.printStackTrace()
+        }
+      },
+      _ => ())
     ()
   }
 
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -868,11 +868,10 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
     val _ = unsafeRunFiber(
       (),
       t => {
-        if (NonFatal(t) && runtime.config.reportUnhandledFiberErrors) {
-          runtime.compute.reportFailure(t)
-        } else {
-          t.printStackTrace()
-        }
+        if (NonFatal(t)) {
+          if (runtime.config.reportUnhandledFiberErrors)
+            runtime.compute.reportFailure(t)
+        } else { t.printStackTrace() }
       },
       _ => ())
     ()
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -871,7 +871,6 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
         if (NonFatal(t) && runtime.config.reportUnhandledFiberErrors) {
           runtime.compute.reportFailure(t)
         } else {
-          // fatal failure
           t.printStackTrace()
         }
       },
--- a/tests/shared/src/test/scala/cats/effect/IOSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/IOSpec.scala
@@ -1346,7 +1346,7 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
         test.flatMap(_ => IO(canceled)) must completeAs(true)
       }
 
-      "handle errors raised during unsafeRunAndForget" in ticked { implicit ticker =>
+      "report errors raised during unsafeRunAndForget" in ticked { implicit ticker =>
         import cats.effect.unsafe.IORuntime
         import scala.concurrent.Promise
         val test = for {
--- a/tests/shared/src/test/scala/cats/effect/IOSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/IOSpec.scala
@@ -303,6 +303,16 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
 
         action must completeAs(Nil)
       }
+
+      "report errors raised during unsafeRunAndForget" in {
+        import cats.effect.unsafe.IORuntime
+        import scala.concurrent.Promise
+        val errorReporter = Promise[Boolean]()
+        val customRuntime =
+          IORuntime.builder().setFailureReporter(_ => errorReporter.success(true)).build()
+        IO.raiseError(new RuntimeException).unsafeRunAndForget()(customRuntime)
+        errorReporter.future
+      }
     }
 
     "suspension of side effects" should {
@@ -1344,21 +1354,6 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
         }
 
         test.flatMap(_ => IO(canceled)) must completeAs(true)
-      }
-
-      "report errors raised during unsafeRunAndForget" in ticked { implicit ticker =>
-        import cats.effect.unsafe.IORuntime
-        import scala.concurrent.Promise
-        val test = for {
-          errorHandler <- IO(Promise[Boolean]())
-          customRuntime = IORuntime
-            .builder()
-            .setFailureReporter(_ => errorHandler.success(true))
-            .build()
-          _ = IO.raiseError(new RuntimeException).unsafeRunAndForget()(customRuntime)
-          handled <- IO.fromFuture(IO(errorHandler.future))
-        } yield handled
-        test must completeAs(true)
       }
     }
 
--- a/tests/shared/src/test/scala/cats/effect/IOSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/IOSpec.scala
@@ -320,7 +320,7 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
             .builder()
             .setCompute(ec2(ec, errorReporter), () => ())
             .build()
-          _ = IO.raiseError(new RuntimeException).unsafeRunAndForget()(customRuntime)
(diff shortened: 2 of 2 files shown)


armanbilge:
Awesome! 🚀

As a test, I was thinking you can build a custom `IORuntime` with an unhandled-error-reporter installed that e.g. completes a `Promise` or something. Then you can `unsafeRunForget()(customRuntime)` and wait for that promise to be completed.

Here's an existing test that shows how to build a custom runtime.
https://github.com/typelevel/cats-effect/blob/59f12d1fffeadda1fe88ec1781ba7889596dbca1/tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala#L325-L330

tests/shared/src/test/scala/cats/effect/IOSpec.scala
```diff
@@ -1599,6 +1599,21 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
 
         test.flatMap(_ => IO(canceled)) must completeAs(true)
       }
+
+      "handle errors raised during unsafeRunAndForget" in ticked { implicit ticker =>
```
samspills: @armanbilge I wasn't quite sure where to put this test so I stuck it in the misc. section of IOSpec? Is that reasonable??

tests/shared/src/test/scala/cats/effect/IOSpec.scala
```diff
@@ -1599,6 +1599,21 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
...
+          errorHandler <- IO(Promise[Boolean]())
+          customRuntime = IORuntime
+            .builder()
+            .setFailureReporter(_ => errorHandler.success(true))
+            .build()
+          _ = IO.raiseError(new RuntimeException).unsafeRunAndForget()(customRuntime)
+          handled <- IO.fromFuture(IO(errorHandler.future))
+        } yield handled
+        test must completeAs(true)
+      }
```
armanbilge: `ticked` is the mock runtime, which thinks the world revolves around it, so it will be in denial about the other runtime in this test 😁 

Actually we can just write this test mostly unsafely, something like this I reckon.
```scala
"report errors raised during unsafeRunAndForget" in {
  import cats.effect.unsafe.IORuntime
  import scala.concurrent.Promise
  val errorHandler = Promise[Boolean]()
  val customRuntime = IORuntime
    .builder()
    .setFailureReporter(_ => errorHandler.success(true))
    .build()
  IO.raiseError(new RuntimeException).unsafeRunAndForget()(customRuntime)
  errorHandler.future
}
```

tests/shared/src/test/scala/cats/effect/IOSpec.scala
armanbilge: Sure seems fine :) I also saw a section of tests for unhandled error reporting further above, it might fit in there as well.

tests/shared/src/test/scala/cats/effect/IOSpec.scala
samspills: ooh that is nicer!!

samspills: Still waiting on CI but: the test passes locally with the fix in `unsafeRunAndForget`. If I remove the fix then the whole thing hangs b/c the future is never completed. I just wanted to check, is that an "okay" failure mode for the test?

edit: test is failing on native so I'm still looking at it!

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -938,7 +938,17 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
...
    */
   def unsafeRunAndForget()(implicit runtime: unsafe.IORuntime): Unit = {
-    val _ = unsafeRunFiber((), _ => (), _ => ())
+    val _ = unsafeRunFiber(
+      (),
+      t => {
+        if (NonFatal(t) && runtime.config.reportUnhandledFiberErrors) {
+          runtime.compute.reportFailure(t)
+        } else {
+          // fatal failure
```
durban: `t` can be non-fatal here.

samspills: Oh yup 🤦🏻‍♀️ Didn't update the comment after adding the other condition. Thanks!

tests/shared/src/test/scala/cats/effect/IOSpec.scala
samspills: I think I've figured out the issue with the native test. The runtime builder `.build()` method is platform specific. In the JVM version, the compute EC is initialized with the failure reporter from the runtime:
https://github.com/typelevel/cats-effect/blob/7af6add4ada8152acbf30c48d0deca108e982a9c/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeBuilderPlatform.scala#L23-L25

However that's not the case for native/js
https://github.com/typelevel/cats-effect/blob/0f050d1d31b6efc3ebaa3d66edb4a061a92a6939/core/js-native/src/main/scala/cats/effect/unsafe/IORuntimeBuilderPlatform.scala#L23-L24

I'm thinking I can create a custom compute for the test, with the error handler defined there explicitly. Does that make sense??

armanbilge: > I'm thinking I can create a custom compute for the test, with the error handler defined there explicitly.

Oh yeah, that's a good idea. In fact if you do that, you may be able to switch back to a `ticked` test, which is preferable because then the failure mode wouldn't be hanging!

https://github.com/typelevel/cats-effect/blob/2bd09fca083f972bc8d43a64e5ca27b223a388f3/tests/shared/src/test/scala/cats/effect/IOSpec.scala#L262-L269

Also we should fix JS and Native, but that be another PR :)

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -944,7 +944,6 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
         if (NonFatal(t) && runtime.config.reportUnhandledFiberErrors) {
           runtime.compute.reportFailure(t)
         } else {
```
armanbilge: I thought this was Daniel Urban's point about this :)

```suggestion
        if (NonFatal(t)) {
          if (runtime.config.reportUnhandledFiberErrors)
            runtime.compute.reportFailure(t)
        } else {
```

tests/shared/src/test/scala/cats/effect/IOSpec.scala
```diff
@@ -324,14 +324,26 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
...
+        }
+
+        val test = for {
+          ec <- IO.executionContext
+          errorReporter <- IO(Promise[Boolean]())
+          customRuntime = IORuntime
+            .builder()
+            .setCompute(ec2(ec, errorReporter), () => ())
+            .build()
+          _ = IO.raiseError(new RuntimeException).unsafeRunAndForget()(customRuntime)
```
armanbilge: Bit weird, but ...

```suggestion
          _ <- IO(IO.raiseError(new RuntimeException).unsafeRunAndForget()(customRuntime))
```

core/shared/src/main/scala/cats/effect/IO.scala
samspills: ohhhhhh 🤦🏻‍♀️ 🤦🏻‍♀️ 🤦🏻‍♀️

core/shared/src/main/scala/cats/effect/IO.scala
samspills: Sorry, I totally missed your point here at first 😅 The gating logic is updated now via https://github.com/typelevel/cats-effect/pull/3475/commits/f4f1e917ccc811f00f4bacbe44bb0b6b54d8fa62

armanbilge (dismissed):
LGTM. Thank you for all your work and perseverance on this!!!

armanbilge (dismissed):

armanbilge (approved):

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -865,7 +865,15 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
...
    */
   def unsafeRunAndForget()(implicit runtime: unsafe.IORuntime): Unit = {
-    val _ = unsafeRunFiber((), _ => (), _ => ())
+    val _ = unsafeRunFiber(
+      (),
+      t => {
+        if (NonFatal(t)) {
+          if (runtime.config.reportUnhandledFiberErrors)
+            runtime.compute.reportFailure(t)
+        } else { t.printStackTrace() }
```
djspiewak, line 874: Should we be using `IOFiber.onFatalFailure(t)` instead?

armanbilge: Maybe, then I guess it should be changed for all the other `unsafeRun*` as well.
https://github.com/typelevel/cats-effect/blob/eb307822dad0e91b3ab0388f8292f4e85d724b3c/core/shared/src/main/scala/cats/effect/IO.scala#L833-L834

djspiewak: Ah fair point. Okay I retract

samspills: I could explore switching them all to `IOFiber.onFatalFailure` in a follow up PR if that would be interesting?

djspiewak: I'm not sure whether it's the right thing to do. I have to think a bit more about that feature. :-)

PR #3476, opened 2023-03-02
Don't shadow `println` in `IOFiber`

armanbilge:
Fool me once, shame on you;
Fool me twice, shame on me.
h/t @samspills


djspiewak (approved):
I genuinely have no idea why we did this. It's horribly confusing.

PR #3478, opened 2023-03-02
Fix `IORuntimeBuilder` `failureReporter` config on JS

armanbilge:


tests/shared/src/test/scala/cats/effect/unsafe/IORuntimeBuilderSpec.scala
```diff
@@ -0,0 +1,33 @@
...
+
+package cats.effect
+package unsafe
+
+class IORuntimeBuilderSpec extends BaseSpec with DetectPlatform {
+
+  "IORuntimeBuilder" should {
+    if (isNative) "configure the failure reporter" in pending
+    else
+      "configure the failure reporter" in {
```
armanbilge, line 25: We _could_ fix it on Native, but all the useful runtimes are custom anyway, so it's kind of pointless at this stage.

djspiewak (approved):

PR #3479, opened 2023-03-03
Fix Resource scaladoc

durban:


kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -866,8 +866,7 @@ object Resource extends ResourceFOInstances0 with ResourceHOInstances0 with Reso
     applyCase[F, A](acquire.map(a => (a, e => release(a, e))))
 
   /**
-   * Creates a resource from an acquiring effect and a release function that can discriminate
-   * between different [[ExitCase exit cases]].
+   * Creates a resource from a possibly interruptible acquiring effect and a release function.
```
armanbilge: I wonder if we should prefer the `cancelable` terminology here? It's all a bit fuzzy: fiber cancelation, thread interruption, and FS2 `Stream` interruption.

durban: Probably yeah...

armanbilge (approved):
thanks :)

PR #3480, opened 2023-03-05
Make `HotSwap` safe to concurrent access

### LINKED ISSUE #2050: `Hotswap` exposes problems with resource leaks on concurrent access
Firstly, thanks for the amazing work on `CE3`, it really is a joy to use.

While working on [Trace4Cats](https://github.com/trace4cats/trace4cats/), I had the need to swap resources at runtime, but split the functionality of swapping and access of `Hotswap` where the resource may be swapped concurrently to access. `Hotswap` seemed perfect for this, but quickly threw up a problem: what if something is using the resource `R`, while it is being swapped? Secondly, what if two fibers try to swap the resource simultaneously?

Admittedly my use case is not the same as in the [documentation](https://typelevel.org/cats-effect/docs/std/hotswap), but I believe it exposes a problem that others may encounter without realising. I also believe that the [example](https://typelevel.org/cats-effect/docs/std/hotswap) is not completely safe, as you could theoretically have two fibers hitting the character limit in two separate operations and swap the resource almost simultaneously, with no guard around whether it is still in use in another fiber.

To this end @catostrophe and I created [`HotswapRef`](https://github.com/janstenpickle/hotswap-ref) to separate the actions of swapping and access safer for concurrent calls, using a semaphore to protect calls to swap and reference counting when accessing the resource `R`, which is returned in a `Resource` to model acquistion/release of a handle. While this does build on `Hotswap` using other `cats-effect` primitives and might not belong in the standard library itself, I do believe that at least some of this functionality should belong in `Hotswap`, like the semaphore around calls to `swap`.

Edit: just to be clear we deliberately designed [`HotswapRef`](https://github.com/janstenpickle/hotswap-ref) so that calls to `access` would _never_ block, only calls to `swap`. This does present its own challenges whereby calling `swap` on the `HotswapRef` within `use` on the `Resource` returned by `access` will cause a deadlock, this is highlighted in [`ConditionalHotswapRefConstructor`](https://github.com/janstenpickle/hotswap-ref/blob/master/modules/core/src/main/scala/io/janstenpickle/hotswapref/ConditionalHotswapRefConstructor.scala#L17-L26).

### LINKED ISSUE #3473: `Hotswap` type parameter isn't really used
Currently we have:

```scala
sealed trait Hotswap[F[_], R] {
  def swap(next: Resource[F, R]): F[R]
  def clear: F[Unit]
}
```

But a more flexible implementation would be:
```scala
sealed trait Hotswap[F[_]] {
  def swap[R](next: Resource[F, R]): F[R]
  def clear: F[Unit]
}
```

That would allow you to swap in different types of resources.

We could make that change if we wanted. It would be bin-compatible, but source-breaking.

Alternatively, we could consider making the type parameter actually useful. e.g.
```scala
sealed trait Hotswap[F[_], R] {
  def get: F[Option[R]]
  def swap(next: Resource[F, R]): F[R]
  def clear: F[Unit]
}
```

armanbilge:
Fixes https://github.com/typelevel/cats-effect/issues/2050. Closes https://github.com/typelevel/cats-effect/issues/3473.

The current resource in `Hotswap` is now guarded by a mutex, which must be acquired to swap or clear it. A new method `get` also provides safe access to the current resource, by acquiring the mutex while it is in use.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
@@ -101,31 +101,37 @@ object Hotswap {
    * Creates a new [[Hotswap]], which represents a [[cats.effect.kernel.Resource]] that can be
    * swapped during the lifetime of this [[Hotswap]].
    */
-  def create[F[_], R](implicit F: Concurrent[F]): Resource[F, Hotswap[F, R]] = {
-    sealed abstract class State
-    case object Cleared extends State
-    case class Acquired(r: R, fin: F[Unit]) extends State
-    case object Finalized extends State
+  def create[F[_], R](implicit F: Concurrent[F]): Resource[F, Hotswap[F, R]] =
+    Resource.eval(Semaphore[F](Long.MaxValue)).flatMap { semaphore =>
+      sealed abstract class State
+      case object Cleared extends State
+      case class Acquired(r: R, fin: F[Unit]) extends State
+      case object Finalized extends State
 
-    def initialize: F[Ref[F, State]] =
-      F.ref(Cleared)
+      def initialize: F[Ref[F, State]] =
+        F.ref(Cleared)
 
-    def finalize(state: Ref[F, State]): F[Unit] =
-      state.getAndSet(Finalized).flatMap {
-        case Acquired(_, finalizer) => finalizer
-        case Cleared => F.unit
-        case Finalized => raise("Hotswap already finalized")
-      }
+      def finalize(state: Ref[F, State]): F[Unit] =
+        state.getAndSet(Finalized).flatMap {
+          case Acquired(_, finalizer) => finalizer
+          case Cleared => F.unit
+          case Finalized => raise("Hotswap already finalized")
+        }
 
-    def raise(message: String): F[Unit] =
-      F.raiseError[Unit](new RuntimeException(message))
+      def raise(message: String): F[Unit] =
+        F.raiseError[Unit](new RuntimeException(message))
 
-    Resource.eval(Mutex[F]).flatMap { mutex =>
+      def shared: Resource[F, Unit] = semaphore.permit
+
+      def exclusive: Resource[F, Unit] =
+        Resource.makeFull[F, Unit](poll => poll(semaphore.acquireN(Long.MaxValue)))(_ =>
+          semaphore.releaseN(Long.MaxValue))
+
       Resource.make(initialize)(finalize).map { state =>
         new Hotswap[F, R] {
 
           override def swap(next: Resource[F, R]): F[R] =
-            mutex.lock.surround {
+            exclusive.surround {
               F.uncancelable { poll =>
                 poll(next.allocated).flatMap {
                   case (r, fin) =>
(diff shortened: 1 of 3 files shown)


std/shared/src/main/scala/cats/effect/std/Hotswap.scala
```diff
@@ -64,11 +64,20 @@ sealed trait Hotswap[F[_], R] {
    * is not used thereafter. Failure to do so may result in an error on the _consumer_ side. In
    * any case, no resources will be leaked.
    *
+   * For safer access to the current resource see [[get]], which guarantees that it will not be
+   * released while it is being used.
+   *
    * If [[swap]] is called after the lifetime of the [[Hotswap]] is over, it will raise an
    * error, but will ensure that all resources are finalized before returning.
    */
   def swap(next: Resource[F, R]): F[R]
```
armanbilge, line 73: This API is a bit unsafey since it directly returns the `R` but I think we'll just have to live with that.

std/shared/src/main/scala/cats/effect/std/Hotswap.scala
```diff
@@ -64,11 +64,20 @@ sealed trait Hotswap[F[_], R] {
...
 
+  /**
+   * Gets the current resource, if it exists. The returned resource is guaranteed to be
+   * available for the duration of the returned resource.
+   */
+  def get: Resource[F, Option[R]]
```
armanbilge, line 79: Maybe a little bit annoying UX that this is an `Option[R]`, alternatively we could raise an error. But the user can decide to do that too.

armanbilge:
Woops, I didn't think this through :) we need to make sure that multiple readers don't block each other.

h/t https://github.com/janstenpickle/hotswap-ref

std/shared/src/main/scala/cats/effect/std/Hotswap.scala
armanbilge: Bikeshed the name. `access` a la `HotswapRef` ?

std/shared/src/main/scala/cats/effect/std/Hotswap.scala
```diff
@@ -92,45 +101,72 @@ object Hotswap {
...
+        }
+
+      def raise(message: String): F[Unit] =
+        F.raiseError[Unit](new RuntimeException(message))
+
+      def shared: Resource[F, Unit] = semaphore.permit
+
+      def exclusive: Resource[F, Unit] =
+        Resource.makeFull[F, Unit](poll => poll(semaphore.acquireN(Long.MaxValue)))(_ =>
+          semaphore.releaseN(Long.MaxValue))
```
armanbilge, line 128: Stole this from here:
https://github.com/janstenpickle/hotswap-ref/blob/34781a0af15784981cd007b74f0f5c31836d73f3/modules/core/src/main/scala/io/janstenpickle/hotswapref/Lock.scala

It seems like a good addition to `std` on its own merits.
```scala
trait Lock[F[_]] {
  def shared: Resource[F, Unit]
}
```

armanbilge: Have a related API here.
https://github.com/armanbilge/fs2-dom/blob/cb0044a3fc8c8ec576072c84dce5b93bcbc07530/dom/src/main/scala/fs2/dom/LockManager.scala#L27-L37

djspiewak: `Lock` would probably be a good addition, but the tricky bit is that it isn't reentrant, so we have to be really careful.

djspiewak (approved):

PR #3484, opened 2023-03-08
Allow that the renamed blocker thread is terminated

### LINKED ISSUE #3216: `WorkerThreadNameSpec` is sometimes flaky in CI
```
WorkerThreadNameSpec
WorkerThread should
  x rename itself when entering and exiting blocking region
   the value is false (WorkerThreadNameSpec.scala:73)
Total for specification WorkerThreadNameSpec
```

aeons:
In `WorkerThreadNameSpec`, if the thread is readded to the compute pool, we still check that it has been renamed, but if the thread has been terminated in the mean time, that is ok as well.

Closes #3216


djspiewak (approved):
Thank you!

PR #3485, opened 2023-03-08
Update starvation documentation with clearer sensitivity explanation

matelaszlo:
- Clearer docs in the Increasing/Decreasing Sensitivity section
- Fixing small mistake in comment

### DIFF
--- a/docs/core/starvation-and-tuning.md
+++ b/docs/core/starvation-and-tuning.md
@@ -485,14 +485,14 @@ object MyMain extends IOApp {
 
 ### Increasing/Decreasing Sensitivity
 
-By default, the checker will warn whenever the compute latency exceeds 100 milliseconds. This is calculated based on the `cpuStarvationCheckInterval` (default: `1.second`) multiplied by the `cpuStarvationCheckThreshold` (default: `0.1d`). In general, it is recommended that if you want to increase or decrease the sensitivity of the checker, you should do so by adjusting the interval (meaning that a more sensitive check will run more frequently):
+By default, the checker will warn whenever the compute latency exceeds 100 milliseconds. This is calculated based on the `cpuStarvationCheckInterval` (default: `1.second`) multiplied by the `cpuStarvationCheckThreshold` (default: `0.1d`). In general, it is recommended that if you want to increase or decrease the sensitivity of the checker, you should do so by adjusting the interval. Decreasing the interval results in a more sensitive check running more frequently, while increasing the interval results in a less sensitive check running less frequently:
 
 ```scala mdoc:silent
 import scala.concurrent.duration._
 
 object MyOtherMain extends IOApp {
 
-  // adjust threshold to 50 milliseconds
+  // relax threshold to 500 milliseconds
   override def runtimeConfig = 
     super.runtimeConfig.copy(cpuStarvationCheckInterval = 5.seconds)
 


djspiewak (approved):
Awesome change, thank you!

docs/core/starvation-and-tuning.md
```diff
@@ -485,14 +485,14 @@ object MyMain extends IOApp {
...
-By default, the checker will warn whenever the compute latency exceeds 100 milliseconds. This is calculated based on the `cpuStarvationCheckInterval` (default: `1.second`) multiplied by the `cpuStarvationCheckThreshold` (default: `0.1d`). In general, it is recommended that if you want to increase or decrease the sensitivity of the checker, you should do so by adjusting the interval (meaning that a more sensitive check will run more frequently):
+By default, the checker will warn whenever the compute latency exceeds 100 milliseconds. This is calculated based on the `cpuStarvationCheckInterval` (default: `1.second`) multiplied by the `cpuStarvationCheckThreshold` (default: `0.1d`). In general, it is recommended that if you want to increase or decrease the sensitivity of the checker, you should do so by adjusting the interval. Decreasing the interval results in a more sensitive check running more frequently, while increasing the interval results in a less sensitive check running less frequently:
 
 ```scala mdoc:silent
 
 object MyOtherMain extends IOApp {
 
-  // adjust threshold to 50 milliseconds
+  // relax threshold to 500 milliseconds
```
durban, line 495: ~~There is still one `0` missing. (Or maybe: change the text to `5 seconds` too.)~~

docs/core/starvation-and-tuning.md
```diff
@@ -485,14 +485,14 @@ object MyMain extends IOApp {
 
 ### Increasing/Decreasing Sensitivity
 
-By default, the checker will warn whenever the compute latency exceeds 100 milliseconds. This is calculated based on the `cpuStarvationCheckInterval` (default: `1.second`) multiplied by the `cpuStarvationCheckThreshold` (default: `0.1d`). In general, it is recommended that if you want to increase or decrease the sensitivity of the checker, you should do so by adjusting the interval (meaning that a more sensitive check will run more frequently):
+By default, the checker will warn whenever the compute latency exceeds 100 milliseconds. This is calculated based on the `cpuStarvationCheckInterval` (default: `1.second`) multiplied by the `cpuStarvationCheckThreshold` (default: `0.1d`). In general, it is recommended that if you want to increase or decrease the sensitivity of the checker, you should do so by adjusting the interval. Decreasing the interval results in a more sensitive check running more frequently, while increasing the interval results in a less sensitive check running less frequently:
```
durban, line 488: ~~Maybe it would be clearer like this: "Decreasing the interval results in a more sensitive check (i.e., running more frequently), while increasing the interval results in a less sensitive check (i.e., running less frequently)". But english is not my first language, so don't listen to me...~~

armanbilge: Hmm, not sure. The sensitivity of the check, relates to the threshold. The frequency of the check, relates to how often it is running. Due to the way configuration is working, they are intertangled. But increasing the frequency is not the only way that the sensitivity is increasing.

docs/core/starvation-and-tuning.md
durban: Ignore this, I'm wrong :man_facepalming:

docs/core/starvation-and-tuning.md
durban: You're right, I've completely misunderstood how this thing works :man_facepalming:

matelaszlo: I think the key thing to note is:
When increasing the interval from 1 to 5 seconds the check will not just run five times less frequently but will also be five times less sensitive each time (it will accept latency up to 500 milliseconds instead of 100 milliseconds).

That is what the phrasing change above and the code example are highlighting.

armanbilge (approved):

PR #3487, opened 2023-03-10
Dispatcher error reporting

### LINKED ISSUE #3131: Dispatcher.unsafeToFuture reporting errors
CE3/Dispatcher/Future/Exceptions. 

I have some `IO`s and convert them Futures (Enforced by Play) using `Dispatcher.unsafeToFuture`. In certain cases I raise errors using `IO.raiseError` to be handled in the unsafe region. The issue is that I see unwanted stacktraces in the console for these potentially expected errors printed by CE runtime(`WorkStealingThreadPool`) as they are rightfully considered as unhandled errors. 

The following example may produce the case where an exception will be printed
```scala
package example

import cats.effect.std.{ Dispatcher, Queue }

import scala.concurrent.{ Await, ExecutionContext, Future }
import scala.util.control.NoStackTrace

object MainDispatcher {

  val (dispatcher, dispatcherShutdown) = Dispatcher[IO].allocated.unsafeRunSync()

  def main(args: Array[String]): Unit = {
    case class SomeError(message: String) extends NoStackTrace
    val m: Future[Unit] =
      dispatcher
        .unsafeToFuture(IO.raiseError(SomeError("nice")))
        .recover {
          case SomeError(message) =>
            println(message)
            ()
        }(ExecutionContext.global)

    Await.result(m, 10.seconds)
  }

}
```

samspills:
This is my attempt to resolve #3131 

My investigation (with Arman's help <3) branched out to two different problems:
1. The original issue about unwanted stack traces being logged when using `Dispatcher#unsafeToFuture`, in the parallel case. Exceptions are currently being reraised in the IO, via `onError`, and triggering the fiber unhandled error reporting. 
2. While investigating that, we realized that sequential dispatchers would have the opposite issue: because all exceptions are being handled with a [`handleError` in the fork](https://github.com/typelevel/cats-effect/blob/series/3.x/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala#L214), you wouldn't know about an exception in the case of `Dispatcher#unsafeRunAndForget` 

This PR attempts to resolve both. Respectively:
1. 66857f4b069faa533e7f607353d80c01406c924f replaces `onError` with `handleErrorWith`
~2. b21e122dd8a74250b143c794e3e3a3c0ea0ecd6e moves the `unsafeRunAndForget` definition into the apply, so we can print the stack trace if the future completes with an exception.~ 
2. e844c59ac8b0eec8ef39d14adc7c612ed6ca7136 redefines `unsafeRunAndForget` using `unsafeRunAsync` + callback to print a stack trace 

I haven't figured out how to test this yet so I'm still working on that, but also looking for feedback on the 2nd solution. I'm not sure if its okay to just move things around like that 😅 

Edit: hahaha, I guess I got my answer about just moving things around
```
java.lang.RuntimeException: Failed binary compatibility check against org.typelevel:cats-effect-std_2.13:3.3.11
```

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
@@ -312,7 +312,7 @@ object Dispatcher {
 
           val action = fe
             .flatMap(e => F.delay(promise.success(e)))
-            .onError { case t => F.delay(promise.failure(t)) }
+            .handleErrorWith(t => F.delay(promise.failure(t)))
             .void
 
           val cancelState = new AtomicReference[CancelState](CancelInit)
--- a/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
@@ -52,6 +52,11 @@ trait Dispatcher[F[_]] extends DispatcherPlatform[F] {
   def unsafeToFutureCancelable[A](fa: F[A]): (Future[A], () => Future[Unit])
 
   /**
+   * Submits an effect to be executed with fire-and-forget semantics.
+   */
+  def unsafeRunAndForget[A](fa: F[A]): Unit
+
+  /**
    * Submits an effect to be executed, returning a `Future` that holds the result of its
    * evaluation.
    */
@@ -64,14 +69,6 @@ trait Dispatcher[F[_]] extends DispatcherPlatform[F] {
    */
   def unsafeRunCancelable[A](fa: F[A]): () => Future[Unit] =
     unsafeToFutureCancelable(fa)._2
-
-  /**
-   * Submits an effect to be executed with fire-and-forget semantics.
-   */
-  def unsafeRunAndForget[A](fa: F[A]): Unit = {
-    unsafeToFutureCancelable(fa)
-    ()
-  }
 
   // package-private because it's just an internal utility which supports specific implementations
   // anyone who needs this type of thing should use unsafeToFuture and then onComplete
@@ -307,6 +304,15 @@ object Dispatcher {
       }
     } yield {
       new Dispatcher[F] {
+        def unsafeRunAndForget[A](fa: F[A]): Unit = {
+          unsafeToFutureCancelable(fa)
+            ._1
+            .onComplete {
+              case Success(_) => ()
+              case Failure(t) => t.printStackTrace()
+            }(ec)
+        }
+
         def unsafeToFutureCancelable[E](fe: F[E]): (Future[E], () => Future[Unit]) = {
           val promise = Promise[E]()
 
--- a/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
@@ -52,11 +52,6 @@ trait Dispatcher[F[_]] extends DispatcherPlatform[F] {
   def unsafeToFutureCancelable[A](fa: F[A]): (Future[A], () => Future[Unit])
 
   /**
-   * Submits an effect to be executed with fire-and-forget semantics.
-   */
-  def unsafeRunAndForget[A](fa: F[A]): Unit
-
-  /**
    * Submits an effect to be executed, returning a `Future` that holds the result of its
    * evaluation.
    */
@@ -69,6 +64,16 @@ trait Dispatcher[F[_]] extends DispatcherPlatform[F] {
    */
   def unsafeRunCancelable[A](fa: F[A]): () => Future[Unit] =
     unsafeToFutureCancelable(fa)._2
+
+  /**
+   * Submits an effect to be executed with fire-and-forget semantics.
+   */
+  def unsafeRunAndForget[A](fa: F[A]): Unit = {
+    unsafeRunAsync(fa) {
+      case Left(t) => t.printStackTrace()
+      case Right(_) => ()
+    }
+  }
 
   // package-private because it's just an internal utility which supports specific implementations
   // anyone who needs this type of thing should use unsafeToFuture and then onComplete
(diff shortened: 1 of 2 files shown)


armanbilge:
Btw you can target this at 3.4.x :)

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -307,12 +304,21 @@ object Dispatcher {
...
     } yield {
       new Dispatcher[F] {
+        def unsafeRunAndForget[A](fa: F[A]): Unit = {
+          unsafeToFutureCancelable(fa)
+            ._1
+            .onComplete {
+              case Success(_) => ()
+              case Failure(t) => t.printStackTrace()
+            }(ec)
+        }
```
armanbilge: I think if you write this in terms of `unsafeRunAsync` you can move it back to `trait Dispatcher` :)

samspills: Ohhhh yes, that's an excellent idea!

tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
```diff
@@ -257,6 +258,47 @@ class DispatcherSpec extends BaseSpec with DetectPlatform {
...
+        errorReporter <- Resource.eval(IO(Promise[Boolean]()))
+        customEc = ec2(ec, errorReporter)
+        _ <- dispatcher
+          .evalOn(customEc)
+          .flatMap(runner =>
+            Resource.eval(IO(runner.unsafeToFuture(IO.raiseError(new Exception("boom"))))))
+        reported <- Resource.eval(IO.fromFuture(IO(errorReporter.future)))
+      } yield reported
+
+      test.use(t => IO(t).mustFailWith[TimeoutException])
```
samspills: This is not actually doing what I want, b/c tests are timing out but not showing up as passing (or failing, to be fair). 

Might need to rethink how to test the absence of error reporting

armanbilge: I would try doing the `unsafeToFuture` and `fromFuture` within the `use` or (`surround`) block. Then you can just slap a `.timeout(1.second)` on it or something.

tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
```diff
@@ -272,10 +272,11 @@ class DispatcherSpec extends BaseSpec with DetectPlatform {
...
           .flatMap(runner =>
             Resource.eval(IO(runner.unsafeRunAndForget(IO.raiseError(new Exception("boom"))))))
-        reported <- Resource.eval(IO.fromFuture(IO(errorReporter.future)))
-      } yield reported
+      } yield errorReporter
 
-      test.use(t => IO(t mustEqual true))
+      test
+        .use(t => IO.fromFuture(IO(t.future)).timeout(1.second).handleError(_ => false))
+        .flatMap(t => IO(t mustEqual true))
```
samspills: In the sad path, if errors aren't reported, this test will fail after 1 second with 
```
[error]     x report exception if raised during unsafeRunAndForget
[error]      false != true (DispatcherSpec.scala:279)
```

(the previous version would just timeout, which wasn't actually a failing test which is definitely not ideal)

tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
```diff
@@ -293,10 +294,10 @@ class DispatcherSpec extends BaseSpec with DetectPlatform {
           .evalOn(customEc)
           .flatMap(runner =>
             Resource.eval(IO(runner.unsafeToFuture(IO.raiseError(new Exception("boom"))))))
-        reported <- Resource.eval(IO.fromFuture(IO(errorReporter.future)))
-      } yield reported
+      } yield errorReporter
 
-      test.use(t => IO(t).mustFailWith[TimeoutException])
+      test.use(t =>
+        IO.fromFuture(IO(t.future)).timeout(1.second).mustFailWith[TimeoutException])
```
samspills, line 300: In the sad path, if errors ARE reported, this test will fail fast with
```
[error]     x do not treat exception in unsafeRunToFuture as unhandled
[error]      java.lang.Exception: boom doesn't have class 'java.util.concurrent.TimeoutException'. It has class 'java.lang.Exception' (Runners.scala:96)
```

I can also construct a situation[^1] where the test fails like that, but there is _also_ a stacktrace for `java.lang.IllegalStateException: Promise already completed.` printed. This seems _not correct_ but I'm not sure why this is happening yet. Anyway just a heads up 😅 

[^1]: By trying to get the `unsafeRunAndForget` test passing, but letting this test fail. I kept the update to `unsafeRunAndForget` but switched back to `onError` in `unsafeToFutureCancelable`, that caused the 'promise already completed' to be reported.

tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
armanbilge:
```suggestion
      test
        .use(t => IO.fromFuture(IO(t.future)).timeoutTo(1.second, IO.pure(false)))
        .flatMap(t => IO(t mustEqual true))
```

samspills: oh look at that!

tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
samspills: I have figured this out! It actually wasn't this test that was causing the stack trace, it was the `unsafeRunAndForget` test, and it makes sense (imo). 

By trying to make only this test fail, I ended up in a situation where `unsafeRunAndForget` had a callback to report the failure, and ALSO `unsafeToFutureCancelable` was using `onError` which would _also_ report the failure. So there were two different, and legit, attempts to complete the `errorReporter` promise.

But part of my fix is that `unsafeToFutureCancelable` _shouldn't_ be reporting the error like that, so I think this extra stack trace is okay because it will always accompany a failing test for `unsafeToFuture`

armanbilge (approved):
LGTM, nice job investigating this one!

Ideally the tests wouldn't rely on timeouts, but after thinking about it some more I'm pretty sure it's the only option with `real` tests. Since unhandled error reporting is the last thing a fiber does, there's no good "hook" available to be confident about when that should/shouldn't have happened by. So we just have to wait and see :)

PR #3490, opened 2023-03-10
Fix "support re-enablement via cancelable" test

armanbilge:
The fiber was being canceled before it was started, so the test wasn't actually invoking `cancelable` at all 😁


djspiewak (approved):
😬

PR #3471, opened 2023-03-01
Try using a field for `IO#tag`

armanbilge:
All the talk about unsafe shenanigans when I thought we may as well try the obvious thing: `final byte tag;`.

(De-)compiles down to this in the runloop:
```java
byte by = cur0.tag;
switch (by) {
```

Threw a couple of benchmarks at it, seems to be a win? `delay` came out a bit weird.

#### This PR
```
Benchmark                (size)   Mode  Cnt     Score    Error  Units
DeepBindBenchmark.async           10000  thrpt   20  1270.544 ±   8.526  ops/s
DeepBindBenchmark.delay           10000  thrpt   20  3699.730 ± 215.698  ops/s
DeepBindBenchmark.pure            10000  thrpt   20  4707.829 ± 181.234  ops/s
RedeemWithBenchmark.errorRaised   10000  thrpt   20  1147.630 ±   6.235  ops/s
RedeemWithBenchmark.happyPath     10000  thrpt   20  1693.848 ±  13.185  ops/s
```

#### series/3.x
```
Benchmark                (size)   Mode  Cnt     Score    Error  Units
DeepBindBenchmark.async           10000  thrpt   20  1076.945 ± 67.990  ops/s
DeepBindBenchmark.delay           10000  thrpt   20  3819.385 ± 37.620  ops/s
DeepBindBenchmark.pure            10000  thrpt   20  4346.132 ± 60.918  ops/s
RedeemWithBenchmark.errorRaised   10000  thrpt   20  1071.057 ± 13.123  ops/s
RedeemWithBenchmark.happyPath     10000  thrpt   20  1604.361 ±  9.081  ops/s
```

PR #3492, opened 2023-03-12
Removed unneeded prefetch in `IOFiber`

djspiewak:
This was an artifact of our `-Xcheckinit` days. The net result here is reducing the memory footprint of `IOFiber` by two full pointer widths, which is actually not a small amount. Surprisingly though, the results are almost exactly a wash for reasons I don't understand... except for `parTraverse`, which is mysteriously 20% faster now. I reproduced the performance delta across repeated runs and negative tests, so I think it's real, but I do find it fascinatingly bizarre.

Even more interestingly, I would have expected the performance results here to be essentially the inverse of #3471, which saw a statistically-significant degradation from shifting from a megamorphic `invokevirtual` to a `getfield` (at the cost of a slightly larger object map). The only explanation I have for why this PR has such minimal impact while *that* PR was so very noticeable (aside from experimental error) is that perhaps we're right up against an L1 page boundary on my test machine. If that's the case, rebasing #3471 on top of this PR and then re-running should produce a noticeable improvement.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -196,7 +196,7 @@ private final class IOFiber[A](
      * either because the entire IO is done, or because this branch is done
      * and execution is continuing asynchronously in a different runloop invocation.
      */
-    if (_cur0 eq IOEndFiber) {
+    if (_cur0 eq IO.EndFiber) {
       return
     }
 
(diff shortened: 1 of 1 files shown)


vasilmkd:
I was planning to do this almost a year ago. I'm glad to see them go. I might have some more ideas for reducing the size of IOFiber even further.

As far as #3471, do be careful about increasing the memory footprint of each IO, it might be a noticeable change for some huge users.

vasilmkd (dismissed):

armanbilge (approved):

Issue #3495, opened 2023-03-15
Make cpu starvation warning's logging destination configurable


manuelcueto:
As mentioned [here](https://github.com/typelevel/cats-effect/issues/3308#issuecomment-1339712130_) 
This warning log logs directly to System.err, which might be undesirable if the application leverages some custom Logging (i.e Logback, log4cats). 
the `CpuStarvationCheck` should contain a configurable Console(most likely through `IORuntimeConfig`) .

manuelcueto:
Working on it :).
However at a first glance, a Console is too big an abstraction for this functionality. It probably is out of scope, but Something like 
`trait Console extends ConsoleReader with ConsoleWriter`
would make sense, and then CpuStarvationCheck should only need a ConsoleWriter in reality

PR #3496, opened 2023-03-15
Add console as config in ioRuntimeConfig, pass it to CPUStarvation

manuelcueto:
Addresses #3495 
Change CpuStarvation to take a function obtaining the metrics when there's a clock drift, to give the ability to the user to do as they want with the information. Defaults to the previous mechanism of logging into Sys.err

### DIFF
--- a/core/js/src/main/scala/cats/effect/IOApp.scala
+++ b/core/js/src/main/scala/cats/effect/IOApp.scala
@@ -179,7 +179,7 @@ trait IOApp {
     Console[IO].printStackTrace(err)
 
   /**
-   * Defines what to do when [[CpuStarvationCheck]] is triggered. Defaults to log a warning to
+   * Defines what to do when CpuStarvationCheck is triggered. Defaults to log a warning to
    * System.err.
    */
   protected def onCpuStarvationWarn: CpuStarvationWarningMetrics => IO[Unit] =
--- a/core/js/src/main/scala/cats/effect/IOApp.scala
+++ b/core/js/src/main/scala/cats/effect/IOApp.scala
@@ -16,7 +16,7 @@
 
 package cats.effect
 
-import cats.effect.metrics.JsCpuStarvationMetrics
+import cats.effect.metrics.{CpuStarvationWarningMetrics, JsCpuStarvationMetrics}
 import cats.effect.std.Console
 import cats.effect.tracing.TracingConstants._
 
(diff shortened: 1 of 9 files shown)


core/shared/src/main/scala/cats/effect/unsafe/IORuntimeConfig.scala
```diff
@@ -28,7 +30,8 @@ final case class IORuntimeConfig private (
     val reportUnhandledFiberErrors: Boolean,
     val cpuStarvationCheckInterval: FiniteDuration,
     val cpuStarvationCheckInitialDelay: Duration,
-    val cpuStarvationCheckThreshold: Double) {
+    val cpuStarvationCheckThreshold: Double,
+    val cpuStarvationConsole: Console[IO]) {
```
armanbilge: If we are going to put a `Console` into the `IORuntimeConfig`, I wonder if it shouldn't be specific to the cpu starvation check. It might be a good, general purpose logging mechanism to use in other places as well.

armanbilge: Eh, nevermind, seems both Daniel and I think that an ad-hoc callback hook makes more sense than a `Console` :)

djspiewak (approved):

core/js/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -178,6 +178,13 @@ trait IOApp {
   protected def reportFailure(err: Throwable): IO[Unit] =
     Console[IO].printStackTrace(err)
 
+  /**
+   * Defines what to do when CpuStarvationCheck is triggered. Defaults to log a warning to
+   * System.err.
+   */
+  protected def onCpuStarvationWarn: CpuStarvationWarningMetrics => IO[Unit] =
```
djspiewak: Actually one quick nit: shouldn't this be a method, rather than a function?

```suggestion
  protected def onCpuStarvationWarn(metrics: CpuStarvationWarningMetrics): IO[Unit] =
```

It means we can't define it in a point-free fashion, but the users are much more likely to define this in a non-point-free fashion, similar to `reportFailure`.

djspiewak (changes requested):

core/js/src/main/scala/cats/effect/IOApp.scala
manuelcueto: Will look into this tomorrow morning.  👍  Thanks for the review

djspiewak (approved):

djspiewak:
Thank you for taking this on!

PR #3498, opened 2023-03-15
Add flush to print functions in std.Console

### LINKED ISSUE #1736: CE3 IOApp finishes early before all finalizers
Could be shown with a simple test case (several run may be needed):
```scala
import cats.effect.std.Console
import cats.syntax.all._

object GuaranteeBroken extends IOApp.Simple {
  val workers: Vector[IO[Nothing]] = Vector.tabulate(10) { i =>
    IO.never.guarantee(Console[IO].println(s"Worker #$i is done"))
  }

  val run: IO[Unit] = IO.race(IO.sleep(3.seconds), workers.parSequence_).void
}
```

Usually, especially when run from IDE or under load only around 7 finalizers are run
(or maybe they all run, but `stdout` cache not flushed?).

Similar code but in form of Ammonite script always runs all finalizers:

```scala
import $ivy.`org.typelevel::cats-effect:3.0.0-M5`

import cats.effect._
import cats.effect.std.Console
import cats.syntax.all._

val workers = Vector.tabulate(10)(i => IO.never.guarantee(Console[IO].println(s"Worker #$i is done")))

IO.race(IO.sleep(3.seconds), workers.parSequence_).void.unsafeRunSync()
```

zetashift:
This time picked off from the right branch!

From discord:

>  I think we should probably just add it on each call. That's kind of what people intuitively expect anyway. If someone wants more control over buffering, they should use Fs2. Console is meant for quick-and-dirty things

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/ConsoleCrossPlatform.scala
+++ b/std/shared/src/main/scala/cats/effect/std/ConsoleCrossPlatform.scala
@@ -239,12 +239,18 @@ private[std] abstract class ConsoleCompanionCrossPlatform {
 
     def print[A](a: A)(implicit S: Show[A] = Show.fromToString[A]): F[Unit] = {
       val text = a.show
-      F.blocking(System.out.print(text))
+      F.blocking {
+        System.out.print(text)
+        System.out.flush()
+      }
     }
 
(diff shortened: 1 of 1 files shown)


djspiewak (approved):
Thank you!

Issue #3501, opened 2023-03-18
Closing a `Dispatcher` from within it seems to enter invalid state (?)


armanbilge:
Consider this snippet.
```scala
//> using lib "org.typelevel::cats-effect::3.5.0-RC3"

import cats.effect._
import cats.effect.std.Dispatcher

import scala.concurrent.duration._

object App extends IOApp.Simple {
  // same result with sequential
  def run = Dispatcher.parallel[IO].allocated.flatMap {
    case (dispatcher, release) =>
      IO(dispatcher.unsafeRunAndForget(release)) *>
        IO.sleep(100.millis) *>
        IO(dispatcher.unsafeRunAndForget(IO.println("please don't print"))) *>
        IO.sleep(100.millis) *>
        IO.println("huh")
  }
}
```

What that's doing is starting a `Dispatcher`, and then attempting to close that dispatcher by submitting the close action to it.

When I tried this I wasn't entirely sure what to expect: `release` will block until the `Dispatcher` cancels all ongoing tasks. But one of those ongoing tasks is canceling the `Dispatcher`, which is blocked ... 🤔 so I thought this would either deadlock or the release would be `canceled`.

Well, it doesn't deadlock. But also the `Dispatcher` no long appears to work e.g. the `IO.println(...)` doesn't print. _But_ it still allows to submit tasks to it, instead of raising an `java.lang.IllegalStateException: dispatcher already shutdown`.

So it's in this weird state, where nothing is running but shutdown somehow hasn't completed.

armanbilge:
Oh, nevermind. Just turns out that `100.millis` was not sufficient, which is surprising. `1000.millis` works. Maybe a cold JVM.

armanbilge:
Actually, the `parallel` dispatcher is fine, but seems to be an issue still with `sequential` 🤔

djspiewak:
@armanbilge Is this fixed now?

samspills:
@djspiewak oh whoops, I said I would look at this 😅 I think this might have been resolved with the `alive` change in #3510, but I haven't written a proper test for it yet

durban:
Something still doesn't seem okay with this: https://github.com/typelevel/cats-effect/actions/runs/4866933748/jobs/8678979906?pr=3592#step:23:4698

samspills:
Thank you for reopening @durban!

I can replicate the flake easily with the following setup:

```scala
"issue 3501: reject new tasks after release action is submitted as a task" in real {
[...]

PR #3510, opened 2023-03-22
Dispatcher: check for outstanding actions before release

### LINKED ISSUE #3506: Dispatcher#unsafeRunAndForget seems not to await despite (await = true)
CE 3.4.8

```scala
Dispatcher.parallel[IO](await = true).use { dispatcher =>
  for {
    _ <- IO(dispatcher.unsafeRunAndForget(IO.print(" Scalar 2023").delayBy(5.millis)))
    _ <- IO(dispatcher.unsafeRunAndForget(IO.print("Hello")))
    //        _ <- IO.sleep(100.millis)
  } yield ()
} *> IO.blocking(System.out.flush())
```

does not print anything.

samspills:
resolves #3506

c0ed15a88eaff0c1d1073551e9cdf74693578030 adds a test demonstrating that a `Dispatcher`, with `await=true`, will finish without completing the effect (thus failing the test) 

Ultimately this is caused by the dispatcher finalizer running and releasing, even though there may have been new actions enqueued since the worker last checked. 

Paired with @djspiewak and came to the solution in 0b6eaa446e352573f8f64acb8a3d3137d0b951b8 in which we can now call `step` one last time, after the dispatcher is done but before releasing, so that those last-minute actions are forked appropriately. With this change the new test passes 🎉

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
@@ -247,12 +247,8 @@ object Dispatcher {
       supervisor <- Supervisor[F](await, Some((_: Outcome[F, Throwable, _]) => true))
 
       _ <- {
-        def dispatcher(
-            doneR: AtomicBoolean,
-            latch: AtomicReference[() => Unit],
-            state: Array[AtomicReference[List[Registration]]]): F[Unit] = {
-
-          val step = for {
+        def step(state: Array[AtomicReference[List[Registration]]], await: F[Unit]): F[Unit] =
+          for {
             regs <- F delay {
               val buffer = mutable.ListBuffer.empty[Registration]
               var i = 0
@@ -269,12 +265,7 @@ object Dispatcher {
 
             _ <-
               if (regs.isEmpty) {
-                F.async_[Unit] { cb =>
-                  if (!latch.compareAndSet(Noop, () => cb(Completed))) {
-                    // state was changed between when we last set the latch and now; complete the callback immediately
-                    cb(Completed)
-                  }
-                }
+                await
               } else {
                 regs traverse_ {
                   case r @ Registration(action, prepareCancel) =>
@@ -287,10 +278,23 @@ object Dispatcher {
               }
           } yield ()
 
+        def dispatcher(
+            doneR: AtomicBoolean,
+            latch: AtomicReference[() => Unit],
+            state: Array[AtomicReference[List[Registration]]]): F[Unit] = {
+
+          val await =
+            F.async_[Unit] { cb =>
+              if (!latch.compareAndSet(Noop, () => cb(Completed))) {
+                // state was changed between when we last set the latch and now; complete the callback immediately
+                cb(Completed)
+              }
+            }
+
           F.delay(latch.set(Noop)) *> // reset latch
             // if we're marked as done, yield immediately to give other fibers a chance to shut us down
             // we might loop on this a few times since we're marked as done before the supervisor is canceled
-            F.delay(doneR.get()).ifM(F.cede, step)
+            F.delay(doneR.get()).ifM(F.cede, step(state, await))
         }
 
         0.until(workers).toList traverse_ { n =>
@@ -300,7 +304,7 @@ object Dispatcher {
             val release = F.delay(latch.getAndSet(Open)())
             Resource.make(supervisor.supervise(worker)) { _ =>
               // published by release
-              F.delay(doneR.lazySet(true)) *> release
+              F.delay(doneR.lazySet(true)) *> step(states(n), F.unit) *> release
             }
           }
         }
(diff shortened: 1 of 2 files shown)


tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
```diff
@@ -310,6 +310,15 @@ class DispatcherSpec extends BaseSpec with DetectPlatform {
       }
     }
 
+    "issue #3506: await unsafeRunAndForget" in real {
+      for {
+        resultR <- IO.ref(false)
+        _ <- dispatcher.use { runner => IO(runner.unsafeRunAndForget(resultR.set(true))) }
+        result <- resultR.get
```
armanbilge, line 344: Because of the race condition this test does not fail reliably for the `parallel` dispatcher. We might want to run it for several iterations or used the `ticked` runtime.

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -300,7 +304,7 @@ object Dispatcher {
             val release = F.delay(latch.getAndSet(Open)())
             Resource.make(supervisor.supervise(worker)) { _ =>
               // published by release
-              F.delay(doneR.lazySet(true)) *> release
+              F.delay(doneR.lazySet(true)) *> step(states(n), F.unit) *> release
```
armanbilge: I'm not sure if `lazySet` makes sense anymore, since it's not immediately published by `release`.

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -287,10 +278,23 @@ object Dispatcher {
...
+                // state was changed between when we last set the latch and now; complete the callback immediately
+                cb(Completed)
+              }
+            }
+
           F.delay(latch.set(Noop)) *> // reset latch
             // if we're marked as done, yield immediately to give other fibers a chance to shut us down
             // we might loop on this a few times since we're marked as done before the supervisor is canceled
-            F.delay(doneR.get()).ifM(F.cede, step)
+            F.delay(doneR.get()).ifM(F.cede, step(state, await))
```
armanbilge, line 296: Tangential, but I wonder if we can complete this `done` loop by returning a sentinel `F[Unit]` value that the supervisor can use as indication that it should no longer restart this fiber.

armanbilge (commented):
Andrew, Sam, and I tripled to review this PR 😂 added some notes below.

We also produced a new test in https://github.com/typelevel/cats-effect/commit/4374e98a3c884f4343e7cee9a61e24d17452ad8a that verifies that dispatchers `"reject new tasks while shutting down"`. It is currently failing for some permutations, but may be fixable by setting `alive` to `false` before doing the final `step(...)` in the release process.

This issue is relevant to this PR, since we want to be sure that the outstanding actions can't keep growing once we start the release process.

tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
samspills: I reran the test (without fix) a bunch of times but I couldn't reproduce the flake. I still rewrote using `ticked` and also didn't see any flakes there so hopefully that is enough without running multiple times?

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
samspills: I agree, I don't think it makes sense. Now that we're doing work before the release, I think we want the `doneR` value to be immediately available

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
samspills: Oh I forgot to answer here :) When we talked about this it made sense to me, so I do want to try experimenting with that as a follow up!

samspills:
Heads up: I have a draft PR up (#3521) with the `alive` test and a potential fix, it includes the commits here but it's a different PR to keep the concerns from getting muddled :)

samspills:
The work on `alive` turned out to be pretty modular and not terribly hairy so I've merged it into this PR 🎉 

More details are in the writeup for #3521, but tldr:
I added a test that submits a 1.second task to a dispatcher, and then concurrently releases the dispatcher while submitting a second (rogue) task. This test failed for sequential dispachers always, but highlighted a race condition with `alive` for parallel dispatchers:
- wait long enough before submitting the rogue task (100 milliseconds): the dispatcher is not alive and the rogue task isn't submitted at all
- wait not at all: the dispatcher is alive and kicking and the rogue task is submitted _and_ completed
- wait a goldilocks period of 500-900ish nanos: the rogue task can be submitted but not completed. This case shouldn't be allowed to happen.

The fix is moving the `alive` resource so that it is the last thing initialized. This means it will be the first to release, setting alive to false even before the workers start releasing. I think this is good, it means that the dispatcher will stop accepting new tasks before the workers start releasing, so no new tasks can be enqueued while the workers are running their final `step`s

armanbilge:
Sam points out that the `alive` changes appear to have fixed https://github.com/typelevel/cats-effect/issues/3501 as well 🎉 we should capture that in a test, I'm happy to do that in a follow-up PR as well.

djspiewak (approved):
Super sorry for taking so long to get back to this! Trying to climb out of the hole I fell into…

Thank you for taking this on!

PR #3512, opened 2023-03-25
Add acquire & release to Mutex

BalmungSan:
Yes, we all know using `acquire` & `release` is error-prone compared to `lock`.
However, sometimes you need to be able to split the steps; like here: https://github.com/BalmungSan/programs-as-values/blob/main/src/main/scala/lambda/io/priority/HighLowPriorityRunner.scala#L26-L27

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Mutex.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Mutex.scala
@@ -49,6 +49,28 @@ import java.util.concurrent.atomic.AtomicBoolean
 abstract class Mutex[F[_]] {
 
   /**
+   * Acquires the lock.
+   *
+   * @note
+   *   In general prefer to use [[lock]] which returns a `Resource` that ensures the lock is
+   *   properly released.
+   * @see
+   *   [[lock]] [[release]]
+   */
+  def acquire: F[Unit]
+
+  /**
+   * Releases the lock.
+   *
+   * @note
+   *   In general prefer to use [[lock]] which returns a `Resource` that ensures the lock is
+   *   properly released.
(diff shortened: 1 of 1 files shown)


std/shared/src/main/scala/cats/effect/std/Mutex.scala
```diff
@@ -48,6 +48,28 @@ import java.util.concurrent.atomic.AtomicBoolean
...
+  /**
+   * Acquires the lock.
+   *
+   * @note
+   *   In general prefer to use [[lock]] which returns a `Resource` that ensures the lock is
+   *   properly released.
+   * @see
+   *   [[lock]] [[release]]
+   */
+  def acquire: F[Unit]
```
armanbilge, line 60: Unfortunately we can't add abstract methods backwards-compatibly.

I feel like there was a rationale for why we didn't add these APIs to `Mutex`, related to typical usage patterns. Meanwhile `Semaphore` offers what you want.

armanbilge: Found it:

> if you need code that has a more complicated pattern of locks/unlocks, it's probably not a mutex

https://discord.com/channels/632277896739946517/632278585700384799/1019261725394096230

BalmungSan: Yeah, actually, thinking about it, I don't think you need `acquire` and `release` for a `Mutex`

SystemFw: It's good that you found that because I was about to repeat it :)

PR #3516, opened 2023-03-29
Add basic tests for RandomSpec

### LINKED ISSUE #2909: Add tests to `RandomSpec`, audit `Random` for platform-specific methods
See https://github.com/typelevel/cats-effect/pull/2906#issue-1179909799.

This is similar to the `Console` issues in https://github.com/typelevel/cats-effect/pull/2604. _However_, in that case the method _would_ link and only fail at _runtime_.

In this case, these methods would not link _at all_. So arguably we could completely remove them from the JS side without breaking binary-compatibility: they did not link, and they still won't link. However, this would be source-breaking of course.

There's still the caveat if someone was making these methods usable by shimming the JDK themselves in their application so that everything would link. This would be a not-so-nice surprise for them 😅

antoniojimeneznieto:
This PR aims to enhance the test coverage for the Random functionality by providing some initial test cases. As a first-time contributor, I apologize in advance for any mistakes and welcome any feedback. Here are some key points about the changes in this PR:

- The existing code has not been modified, ensuring that the changes in this PR are solely focused on adding tests.
- The tests for randomness algorithms like shuffle have been implemented with a low probability of failure (1/10000!). Although this approach might not be ideal for testing the randomness of these functions, it serves as a starting point for further discussion and improvement.

Closes https://github.com/typelevel/cats-effect/issues/2909.

### DIFF
--- a/tests/shared/src/test/scala/cats/effect/std/RandomSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/std/RandomSpec.scala
@@ -32,9 +32,7 @@ class RandomSpec extends BaseSpec {
         val numIterations: Int = 1000
         for {
           random <- Random.scalaUtilRandom[IO]
-          randDoubles <- (1 to numIterations)
-            .toList
-            .traverse(_ => random.betweenDouble(min, max))
+          randDoubles <- random.betweenDouble(min, max).replicateA(numIterations)
         } yield randDoubles.forall(randDouble => randDouble >= min && randDouble <= max)
       }
     }
@@ -46,7 +44,7 @@ class RandomSpec extends BaseSpec {
         val numIterations: Int = 1000
         for {
           random <- Random.scalaUtilRandom[IO]
-          randFloats <- (1 to numIterations).toList.traverse(_ => random.betweenFloat(min, max))
+          randFloats <- random.betweenFloat(min, max).replicateA(numIterations)
         } yield randFloats.forall(randFloat => randFloat >= min && randFloat <= max)
       }
     }
@@ -58,7 +56,7 @@ class RandomSpec extends BaseSpec {
         val numIterations: Int = 1000
         for {
           random <- Random.scalaUtilRandom[IO]
-          randInts <- (1 to numIterations).toList.traverse(_ => random.betweenInt(min, max))
+          randInts <- random.betweenInt(min, max).replicateA(numIterations)
         } yield randInts.forall(randInt => randInt >= min && randInt <= max)
       }
     }
@@ -70,7 +68,7 @@ class RandomSpec extends BaseSpec {
         val numIterations: Int = 1000
         for {
           random <- Random.scalaUtilRandom[IO]
-          randLongs <- (1 to numIterations).toList.traverse(_ => random.betweenLong(min, max))
+          randLongs <- random.betweenLong(min, max).replicateA(numIterations)
         } yield randLongs.forall(randLong => randLong >= min && randLong <= max)
       }
     }
@@ -81,7 +79,7 @@ class RandomSpec extends BaseSpec {
         val numIterations: Int = 1000
         for {
           random <- Random.scalaUtilRandom[IO]
-          randomChars <- (1 to numIterations).toList.traverse(_ => random.nextAlphaNumeric)
+          randomChars <- random.nextAlphaNumeric.replicateA(numIterations)
         } yield randomChars.forall(randomChar => alphaNumeric.contains(randomChar))
       }
     }
@@ -91,7 +89,7 @@ class RandomSpec extends BaseSpec {
         for {
           random <- Random.scalaUtilRandom[IO]
           randomBoolean <- random.nextBoolean
-        } yield randomBoolean == true || randomBoolean == false
+        } yield ok
       }
     }
 
@@ -129,7 +127,7 @@ class RandomSpec extends BaseSpec {
         val numIterations: Int = 1000
         for {
           random <- Random.scalaUtilRandom[IO]
-          randomDoubles <- (1 to numIterations).toList.traverse(_ => random.nextDouble)
+          randomDoubles <- random.nextDouble.replicateA(numIterations)
         } yield randomDoubles.forall(double => double >= 0.0 && double < 1.0)
       }
     }
@@ -139,7 +137,7 @@ class RandomSpec extends BaseSpec {
         val numIterations: Int = 1000
         for {
           random <- Random.scalaUtilRandom[IO]
-          randomFloats <- (1 to numIterations).toList.traverse(_ => random.nextFloat)
+          randomFloats <- random.nextFloat.replicateA(numIterations)
         } yield randomFloats.forall(float => float >= 0.0f && float < 1.0f)
       }
     }
@@ -149,7 +147,7 @@ class RandomSpec extends BaseSpec {
         val sampleSize = 1000
         for {
           random <- Random.scalaUtilRandom[IO]
-          gaussians <- (1 to sampleSize).toList.traverse(_ => random.nextGaussian)
+          gaussians <- random.nextGaussian.replicateA(sampleSize)
           mean = gaussians.sum / sampleSize
           variance = gaussians.map(x => math.pow(x - mean, 2)).sum / sampleSize
           stddev = math.sqrt(variance)
@@ -162,10 +160,7 @@ class RandomSpec extends BaseSpec {
         for {
           random <- Random.scalaUtilRandom[IO]
           int <- random.nextInt
-        } yield int match {
-          case _: Int => true
-          case _ => false
-        }
+        } yield ok
       }
     }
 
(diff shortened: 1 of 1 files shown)


djspiewak:
This looks like a great start! Random test failure rates do concern me though. Even 0.01% is still comparatively frequent given the number of test runs we do on a regular basis, and random CI failures are the *worst*. I'm definitely not one to put up my nose at more tests though. Is there a way we can simply detect the collision case and retry, reducing the odds to something like 1 - (1 - 0.0001)^k, where *k* is the maximum number of retries?

antoniojimeneznieto:
Thank you for the feedback, Daniel. I appreciate it :) It seems that I may have misunderstood the probability of failure in this particular test. Given a list composed of numbers between 1 and 10,000, where each number is different and the order matters, there are 10,000! (factorial of 10,000) possible combinations. Assuming the shuffle is truly random, the probability of obtaining the exact same list is 1 / factorial(10,000), resulting in a failure probability of approximately 2.846 x 10^(-35659) (I may be missing something in my reasoning). Concerning your idea of detecting collisions and retrying sounds like an excellent alternative to me

tests/shared/src/test/scala/cats/effect/std/RandomSpec.scala
```diff
@@ -17,38 +17,238 @@
...
+    "betweenDouble" >> {
+      "generate a random double within a range" in real {
+        val min: Double = 0.0
+        val max: Double = 1.0
+        val numIterations: Int = 1000
+        for {
+          random <- Random.scalaUtilRandom[IO]
+          randDoubles <- (1 to numIterations)
+            .toList
+            .traverse(_ => random.betweenDouble(min, max))
```
armanbilge:
```suggestion
          randDoubles <- random.betweenDouble(min, max).replicateA(numIterations)
```

tests/shared/src/test/scala/cats/effect/std/RandomSpec.scala
```diff
@@ -17,38 +17,238 @@
...
+
+    "nextInt" >> {
+      "generate random int value" in real {
+        for {
+          random <- Random.scalaUtilRandom[IO]
+          int <- random.nextInt
+        } yield int match {
+          case _: Int => true
+          case _ => false
+        }
```
armanbilge: Is this really asserting something that is not guaranteed by the type system? :)

A no-op assert is fine for this. Just executing `nextInt` asserts that it links and runs on Scala.js and Native which is always useful.

```suggestion
          int <- random.nextInt
        } yield ok
```

tests/shared/src/test/scala/cats/effect/std/RandomSpec.scala
antoniojimeneznieto: Cleaner, more elegant and efficient. I am going to update the tests accordingly. Thank you Arman for the feedback :)

tests/shared/src/test/scala/cats/effect/std/RandomSpec.scala
antoniojimeneznieto: I see, it makes completely sense. I am going update it!

armanbilge:
Fatal warnings in CI :)

https://github.com/typelevel/cats-effect/actions/runs/4599225154/jobs/8124359599?pr=3516#step:23:148

```
[error] /home/runner/work/cats-effect/cats-effect/tests/shared/src/test/scala/cats/effect/std/RandomSpec.scala:91:11: parameter value randomBoolean in anonymous function is never used
[error]           randomBoolean <- random.nextBoolean
[error]           ^
[error] /home/runner/work/cats-effect/cats-effect/tests/shared/src/test/scala/cats/effect/std/RandomSpec.scala:162:11: parameter value int in anonymous function is never used
[error]           int <- random.nextInt
[error]           ^
[error] /home/runner/work/cats-effect/cats-effect/tests/shared/src/test/scala/cats/effect/std/RandomSpec.scala:182:11: parameter value long in anonymous function is never used
[error]           long <- random.nextLong
[error]           ^
[error] No warnings can be incurred under -Werror.
[error] four errors found
```

antoniojimeneznieto:
It would be a good idea to use must beTrue ? Or if the tests are not adding much value, deleting them could be another option ?

armanbilge:
Sorry, not sure I understood your question. But even if the tests are not asserting anything very interesting, they are still useful just for testing that the code can link and run.

antoniojimeneznieto:
Sorry, I should have explained it better, the problem seems to be that randomBoolean, int and long are not being utilised. Maybe instead of just yielding ok we could use the following matchers:

randomBoolean must beOneOf(true, false)
int must beBetween(minVal, maxVal)
long must beBetween(minVal, maxVal)

armanbilge:
Oh, I see!

You can assign the value to `_` to indicate that it is unused, we are only running it for the effect. e.g.
```
_ <- random.nextBoolean
```

tests/shared/src/test/scala/cats/effect/std/RandomSpec.scala
```diff
@@ -89,7 +89,7 @@ class RandomSpec extends BaseSpec {
         for {
           random <- Random.scalaUtilRandom[IO]
           randomBoolean <- random.nextBoolean
-        } yield ok
+        } yield randomBoolean must beOneOf(true, false)
```
armanbilge: This is fine too :)

tests/shared/src/test/scala/cats/effect/std/RandomSpec.scala
```diff
@@ -17,38 +17,228 @@
 package cats.effect
 package std
 
+import cats.implicits._
```
armanbilge: Nit: `cats.implicits` import is soft-deprecated, the syntax import can fully replace it :)
```suggestion
import cats.syntax.all._
```

tests/shared/src/test/scala/cats/effect/std/RandomSpec.scala
```diff
@@ -17,38 +17,228 @@
...
-        random2 <- Random.javaSecuritySecureRandom[IO](2): @nowarn("cat=deprecation")
-        bytes2 <- random2.nextBytes(256)
-      } yield bytes1.length == 128 && bytes2.length == 256
+    "betweenDouble" >> {
+      "generate a random double within a range" in real {
+        val min: Double = 0.0
+        val max: Double = 1.0
+        val numIterations: Int = 1000
+        for {
+          random <- Random.scalaUtilRandom[IO]
```
armanbilge: What do you think about abstracting these tests over the `Random` implementation? E.g. moving them inside of a `def(random: IO[Random[IO]])`.

Then we can run all the tests with each of the different random implementations (scala util, thread local, secure, etc.) since this is where we've found platform-specific bugs before (e.g. secure was not implemented on JS). I think we've also had bugs with the thread local random before as well, using the wrong thread local ... 😅

armanbilge (commented):
Looking very good!

I had one last idea about testing the different `Random` implementations—that way we can maximize the value of your work :)

tests/shared/src/test/scala/cats/effect/std/RandomSpec.scala
antoniojimeneznieto: That's a brilliant idea, I am going to implement it :)

antoniojimeneznieto: There are still a couple of methods to be abstracted, but what do you think so far?

tests/shared/src/test/scala/cats/effect/std/RandomSpec.scala
antoniojimeneznieto: Noted :)

tests/shared/src/test/scala/cats/effect/std/RandomSpec.scala
armanbilge: Thanks! Sorry, I should have given an example. Here's how we are doing it for `Dispatcher`:
https://github.com/typelevel/cats-effect/blob/397ddd109b0463955191789892aa8cc9b886fa7c/tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala#L220-L222

antoniojimeneznieto: My bad, I should have checked dispatcher before

antoniojimeneznieto:
Could we reutilise the tests for SecureRandom? 

`"SecureRandom" should {
  testRandom(SecureRandom.javaSecuritySecureRandom[IO])
}`

tests/shared/src/test/scala/cats/effect/std/RandomSpec.scala
```diff
@@ -22,39 +22,262 @@ import org.specs2.specification.core.Fragments
...
-        random1 <- Random.javaSecuritySecureRandom[IO]: @nowarn("cat=deprecation")
-        bytes1 <- random1.nextBytes(128)
-        random2 <- Random.javaSecuritySecureRandom[IO](2): @nowarn("cat=deprecation")
-        bytes2 <- random2.nextBytes(256)
-      } yield bytes1.length == 128 && bytes2.length == 256
+    val R = Random.scalaUtilRandom[IO]
+
+    testRandom(R)
+
+    elementOfTests[Int, List[Int]](
```
armanbilge: Nice job! Is it possible to move these tests inside of `testRandom` as well?

Then as you point out, we should run these tests for all the different random constructors we have: `scalaUtilRandom`, `javaUtilRandom`, `javaUtilConcurrentThreadLocalRandom`, and `javaSecuritySecureRandom` I think are all of them

antoniojimeneznieto: Thanks :) Yes!

djspiewak (approved):
This is great! Thank you!

PR #3521, opened 2023-04-05
Dispatcher: set `alive` to false when done

samspills:
I branched off of #3510, with the intention to try to keep the alive work separate from the `step` work, but ultimately it might make sense to merge together. 

The differences to #3510 are in the last 3 commits:
- 27c55c344dde191396d67f60476a794d96c332c8 initial test
    - failing for sequential dispatchers b/c `rogueSubmitResult` is Right
    - was passing for parallel dispatchers though
- 804fdb90b043fd480d80594396178984f4a197c6 test with reduced sleep
    - failing for parallel dispatchers, sometimes b/c `rogueSubmitResult` is Right, sometimes because the task actually gets done (a done task fails the test, but if the task finished then we're not actually in the failure situation we're concerned about. So this is still flaky, just a different kind of flaky)
    - Edit: I wrote a better explanation of the test situation and the impact of changing the sleep period:
        - wait long enough before submitting the task: the dispatcher is not alive and the task isn't submitted at all
        - wait not at all: the dispatcher is alive and kicking and the task is submitted and completed
        - wait a goldilocks period of 500-900ish nanos: the task can be submitted but not completed
    - The problem is that the first two cases _are_ reasonable behaviours and fine. The third case can currently happen, and that's the problematic behaviour.
- 24d1bdb035cba91ae4860294e28b45b0b69e5389 set alive to false as part of the worker shutdown
    - setting this before we do the final loop through `step` so we can be sure that no new tasks are added 
    - the test passes

I'm concerned that I don't understand the potential implications of the fix. It makes sense to me for a sequential dispatcher, but for a parallel dispatcher one worker will toggle `alive` for the whole dispatcher. It doesn't seem like a single worker _should_ release on its own but I'm still a bit anxious about bringing this shared dispatcher scope into the worker loop?

### update
After chatting with Arman:
- 090aad0e765177cdc036f75b823ff8ffcaf17a52 instead of having workers handle the `alive` setting, I moved the `alive` resource so that it is the very last thing in the for comprehension. This way when closing the dispatcher, the very first thing that happens is that `alive` is set to false, even before we've set workers to done and run their final `step` call.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
@@ -247,12 +247,8 @@ object Dispatcher {
       supervisor <- Supervisor[F](await, Some((_: Outcome[F, Throwable, _]) => true))
 
       _ <- {
-        def dispatcher(
-            doneR: AtomicBoolean,
-            latch: AtomicReference[() => Unit],
-            state: Array[AtomicReference[List[Registration]]]): F[Unit] = {
-
-          val step = for {
+        def step(state: Array[AtomicReference[List[Registration]]], await: F[Unit]): F[Unit] =
+          for {
             regs <- F delay {
               val buffer = mutable.ListBuffer.empty[Registration]
               var i = 0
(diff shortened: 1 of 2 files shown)


armanbilge (approved):
The fix is so clean and straightforward I'm in favor of merging this back into the other PR. Nice work!

samspills:
Merged this branch into #3510, and closing this one

PR #3542, opened 2023-04-17
Add Lock to std

### LINKED ISSUE #3535: Add `Lock` to std
Spinning out of https://github.com/typelevel/cats-effect/pull/3480#discussion_r1125597170.

The API would look something like:

```scala
abstract class Lock[F[_]] {
  def shared: Resource[F, Unit]
}
```

kevchuang:
Resolves #3535

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
@@ -51,7 +51,7 @@ import cats.syntax.all._
  *
  * Ported from https://github.com/typelevel/fs2.
  */
-sealed trait Hotswap[F[_], R] extends Lock[F] {
+sealed trait Hotswap[F[_], R] {
 
   /**
    * Allocates a new resource, closes the previous one if it exists, and returns the newly
@@ -121,6 +121,12 @@ object Hotswap {
       def raise(message: String): F[Unit] =
         F.raiseError[Unit](new RuntimeException(message))
 
+      def shared: Resource[F, Unit] = semaphore.permit
+
+      def exclusive: Resource[F, Unit] =
+        Resource.makeFull[F, Unit](poll => poll(semaphore.acquireN(Long.MaxValue)))(_ =>
+          semaphore.releaseN(Long.MaxValue))
+
       Resource.make(initialize)(finalize).map { state =>
         new Hotswap[F, R] {
 
@@ -144,12 +150,6 @@ object Hotswap {
 
           override def clear: F[Unit] =
             exclusive.surround(swapFinalizer(Cleared).uncancelable)
-
-          override def shared: Resource[F, Unit] = semaphore.permit
-
-          override def exclusive: Resource[F, Unit] =
-            Resource.makeFull[F, Unit](poll => poll(semaphore.acquireN(Long.MaxValue)))(_ =>
-              semaphore.releaseN(Long.MaxValue))
 
           private def swapFinalizer(next: State): F[Unit] =
             state.modify {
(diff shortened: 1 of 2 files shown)


std/shared/src/main/scala/cats/effect/std/Lock.scala
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+package cats.effect.std
+
+import cats.effect.kernel.Resource
+
+abstract class Lock[F[_]] {
+  def shared: Resource[F, Unit]
+  def exclusive: Resource[F, Unit]
+}
```
djspiewak, line 25: I suspect we should have an implementation here as well, apart from just `HotSwap`. People are almost certainly going to try to do `Lock[IO].flatMap(lock => ...)` and wonder why it doesn't work. :-D

djspiewak (changes requested):
Forgot to note the requested changes in the PR state :-)

std/shared/src/main/scala/cats/effect/std/Hotswap.scala
```diff
@@ -51,7 +51,7 @@ import cats.syntax.all._
  *
  * Ported from https://github.com/typelevel/fs2.
  */
-sealed trait Hotswap[F[_], R] {
+sealed trait Hotswap[F[_], R] extends Lock[F] {
```
durban: I'm not sure `Hotswap` should extend `Lock` – it seems more like an implementation detail of `Hotswap`.

djspiewak (approved):
LGTM!

std/shared/src/main/scala/cats/effect/std/Lock.scala
```diff
@@ -0,0 +1,39 @@
...
+
+package cats.effect.std
+
+import cats.effect.kernel.{Concurrent, Poll, Resource}
+import cats.syntax.functor._
+
+abstract class Lock[F[_]] {
+  def shared: Resource[F, Unit]
+  def exclusive: Resource[F, Unit]
+}
```
armanbilge, line 25: We may want to consider adding some other APIs here. For example:
```scala
  def tryShared: Resource[F, Boolean]
  def tryExclusive: Resource[F, Boolean]
```

Also should check Chris's lib for inspiration.

std/shared/src/main/scala/cats/effect/std/Lock.scala
```diff
@@ -0,0 +1,39 @@
...
+
+  def apply[F[_]: Concurrent](maxShared: Long): F[Lock[F]] =
+    Semaphore[F](maxShared).map { semaphore =>
+      new Lock[F] {
+        override def shared: Resource[F, Unit] = semaphore.permit
+        override def exclusive: Resource[F, Unit] =
+          Resource.makeFull((poll: Poll[F]) => poll(semaphore.acquireN(maxShared)))(_ =>
+            semaphore.releaseN(maxShared))
+      }
+    }
```
armanbilge, line 38: We should add a test for this implementation. For example to verify that the exclusive lock is indeed exclusive.

armanbilge (changes requested):

std/shared/src/main/scala/cats/effect/std/Lock.scala
kevchuang: Understood ! I'm working on that !

std/shared/src/main/scala/cats/effect/std/Lock.scala
djspiewak: Definitely should be tested. It occurred to me the other day that I just didn't catch that; thanks for being on it, Arman. :-)

ChristopherDavenport:
Not sure this is enough for folks that want traditional locking.

Like its an approach, but locks are decently diverse.

PR #3546, opened 2023-04-20
Cpu-starvation warnings one line

### LINKED ISSUE #3540: log CPU starvation message as single line
Currently on Cats Effect 3.4.8, the CPU starvation message gets logged like:
```
2023-04-17T10:18:37.277Z [WARNING] Your app's responsiveness to a new asynchronous 
event (such as a new connection, an upstream response, or a timer) was in excess
of 5 milliseconds. Your CPU is probably starving. Consider increasing the 
granularity of your delays or adding more cedes. This may also be a sign that you
are unintentionally running blocking I/O operations (such as File or InetAddress)
without the blocking combinator.
```
As it gets written as multi-line entry, it breaks logging.

Please consider to print it as single line.

Probably something like following would work:
```scala
  private[this] def mkWarning(threshold: Duration)(when: FiniteDuration) =
    s"""|${format(when)} [WARNING] Your app's responsiveness to a new asynchronous
        | event (such as a new connection, an upstream response, or a timer) was in excess
        | of $threshold. Your CPU is probably starving. Consider increasing the
        | granularity of your delays or adding more cedes. This may also be a sign that you
        | are unintentionally running blocking I/O operations (such as File or InetAddress)
        | without the blocking combinator.""".stripMargin.replaceAll("\n", "")
```

mox692:
resolves https://github.com/typelevel/cats-effect/issues/3540

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/metrics/JvmCpuStarvationMetrics.scala
+++ b/core/jvm/src/main/scala/cats/effect/metrics/JvmCpuStarvationMetrics.scala
@@ -41,9 +41,9 @@ private[effect] object JvmCpuStarvationMetrics {
     th.printStackTrace(new PrintWriter(exceptionWriter))
 
     s"""[WARNING] Failed to register Cats Effect CPU starvation MBean, proceeding with
-       |no-operation versions. You will not see MBean metrics for CPU starvation.
-       |Exception follows: \n ${exceptionWriter.toString}
-       |""".stripMargin
+       | no-operation versions. You will not see MBean metrics for CPU starvation.
+       | Exception follows: ${exceptionWriter.toString}
+       |""".stripMargin.replaceAll(System.lineSeparator(), "")
   }
 
   private[this] class NoOpCpuStarvationMetrics extends CpuStarvationMetrics {
--- a/core/jvm/src/main/scala/cats/effect/metrics/JvmCpuStarvationMetrics.scala
+++ b/core/jvm/src/main/scala/cats/effect/metrics/JvmCpuStarvationMetrics.scala
@@ -43,7 +43,7 @@ private[effect] object JvmCpuStarvationMetrics {
     s"""[WARNING] Failed to register Cats Effect CPU starvation MBean, proceeding with
        | no-operation versions. You will not see MBean metrics for CPU starvation.
        | Exception follows: ${exceptionWriter.toString}
-       |""".stripMargin.replaceAll(System.lineSeparator(), "")
+       |""".stripMargin.replaceAll("\n", "")
   }
 
   private[this] class NoOpCpuStarvationMetrics extends CpuStarvationMetrics {
--- a/core/jvm/src/main/scala/cats/effect/metrics/JvmCpuStarvationMetrics.scala
+++ b/core/jvm/src/main/scala/cats/effect/metrics/JvmCpuStarvationMetrics.scala
@@ -41,9 +41,9 @@ private[effect] object JvmCpuStarvationMetrics {
     th.printStackTrace(new PrintWriter(exceptionWriter))
 
     s"""[WARNING] Failed to register Cats Effect CPU starvation MBean, proceeding with
-       | no-operation versions. You will not see MBean metrics for CPU starvation.
-       | Exception follows: ${exceptionWriter.toString}
-       |""".stripMargin.replaceAll("\n", "")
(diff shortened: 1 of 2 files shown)


core/jvm/src/main/scala/cats/effect/metrics/JvmCpuStarvationMetrics.scala
```diff
@@ -41,9 +41,9 @@ private[effect] object JvmCpuStarvationMetrics {
     th.printStackTrace(new PrintWriter(exceptionWriter))
 
     s"""[WARNING] Failed to register Cats Effect CPU starvation MBean, proceeding with
-       |no-operation versions. You will not see MBean metrics for CPU starvation.
-       |Exception follows: \n ${exceptionWriter.toString}
-       |""".stripMargin
+       | no-operation versions. You will not see MBean metrics for CPU starvation.
+       | Exception follows: ${exceptionWriter.toString}
+       |""".stripMargin.replaceAll(System.lineSeparator(), "")
```
armanbilge: Question: when we have a multi-line `""" ... """` block is it actually using the `System.lineSeparate()` to assembly that at runtime? Or is it assembled at compile time with whatever line separator is actually used in the source file?

Thanks for the PR btw! :)

mox692: @armanbilge 
I think the formar, `System.lineSeparate()` emits the line separator accourding to the runtime system propaty.

```scala
  def example = 
    s"""| some
        | text
        | here""".stripMargin.replaceAll(System.lineSeparator(), "")

  def run = IO.print(s"separator: ${System.lineSeparator()}\n") >> IO.print(example)
}

// running on the unix terminal with `-Dline.separator=@` flag
separator: @
 some   // line break happens
 text
 here
```

But, it might be a bit confusing🤔
would it be better to write it as a single line in the source code as well?

armanbilge: So if I compile this code:
```scala
class Foo {
  def bar = """
            """
}
```
and run it through decompiler, it becomes this:
```java
public class Foo {
    public String bar() {
        return "\n            ";
    }
}
```

So it looks like the scala compiler is using `\n` here (it is using the contents of the `""" ... """` verbatim, it is not using `lineSeparator()`. So, we should use `'\n'`.

mox692: I see now, sorry for my failings!

armanbilge: Not at all! I myself had no idea how it works until today 😁

armanbilge (dismissed):

core/jvm/src/main/scala/cats/effect/metrics/JvmCpuStarvationMetrics.scala
```diff
@@ -41,9 +41,9 @@ private[effect] object JvmCpuStarvationMetrics {
     th.printStackTrace(new PrintWriter(exceptionWriter))
 
     s"""[WARNING] Failed to register Cats Effect CPU starvation MBean, proceeding with
-       |no-operation versions. You will not see MBean metrics for CPU starvation.
-       |Exception follows: \n ${exceptionWriter.toString}
-       |""".stripMargin
+       | no-operation versions. You will not see MBean metrics for CPU starvation.
+       | Exception follows: ${exceptionWriter.toString}
+       |""".stripMargin.replaceAll("\n", "")
```
durban: This makes the whole exception stacktrace print on one line. Is that really useful?

Also, another complication is that `printStackTrace` _does_ use `System.lineSeparator()`...

armanbilge: Ooph, good catch 🤔 maybe we should just leave this one alone? You will get this warning at most once per run, but the CPU starvation warning may occur multiple times.

mox692: Sorry, I missed the comment.
yes, it's terrible to output the stack trace in just one line...

> maybe we should just leave this one alone?

I agree, I'll modify it to apply only to `StarvationCheck`.

durban (approved):

armanbilge (approved):

Issue #3503, opened 2023-03-20
Timed out tests doesn't fail the sbt task


durban:
This might be a specs2 thing, but it seems unlikely that it's intentional.

- One example is [here](https://github.com/typelevel/cats-effect/issues/3463#issuecomment-1448605353)
- Another one [here](https://github.com/typelevel/cats-effect/actions/runs/4443535679/jobs/7800886809) (the CI execution was cancelled, not failed; I don't know what and why cancelled it, might it be related?):
```
[info] DrainBatchSpec
[info] Batch draining should
[info]   o work correctly in the presence of concurrent stealers
[info] timeout exception
```

armanbilge:
> * CI execution was cancelled, not failed; I don't know what and why cancelled

We have a 1 hour time limit on CI to prevent hanging. Maybe it's not sufficient?

https://github.com/typelevel/cats-effect/blob/fb35b3f85d03cb5d33eb5e12ceb4d3f179346783/.github/workflows/ci.yml#L106

durban:
Hm, okay, that's what cancelled it :-) There might be a real problem there (most JVM runs finished in around 15 minutes, so 60 should be plenty), unclear if related to the timed out specs2 test...

durban:
It _might_ be intentional, but I still don't understand _why_ they make a timeout a skipped test:
https://github.com/etorreborre/specs2/blob/SPECS2-4.19.2/core/shared/src/main/scala/org/specs2/specification/core/Execution.scala#L156-L160

Also, there is this here:
https://github.com/typelevel/cats-effect/blob/series/3.x/tests/shared/src/test/scala/cats/effect/Runners.scala#L103-L117
So it might be an unintentional interaction...

armanbilge:
> It _might_ be intentional, but I still don't understand _why_ they make a timeout a skipped test:

Also, there is this commit: https://github.com/etorreborre/specs2/commit/24766396e5643900b76a039c15c218e5bd0d8fec 🤔

durban:
Should be fixed by  #3551.

PR #3549, opened 2023-04-24
Fix mutex cancelled acquire

durban:
Fix Async mutex acquire to be cancellation-safe.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Mutex.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Mutex.scala
@@ -116,8 +116,7 @@ object Mutex {
             }
           }
         }
-      })
-      .flatMap { acquired =>
+      }).flatMap { acquired =>
         if (acquired) F.unit // home free
         else acquire // wokened, but need to acquire
       }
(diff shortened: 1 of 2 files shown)


armanbilge:
Unfortunately this doesn't seem to have fixed the Skunk issue 😕  tried it in https://github.com/armanbilge/skunk/commit/7a2ab6a0b8971a633cc635dae79562f777e4a36c.

armanbilge:
The test you added is good but unfortunately it's affected by https://github.com/typelevel/cats-effect/issues/3503 😕

armanbilge:
Is this cancelation safe? Can the `async` which created this `waiter` be canceled before it is invoked?

https://github.com/typelevel/cats-effect/blob/c760a55de5bbeaf2e19b774b0a144a20bec42105/std/shared/src/main/scala/cats/effect/std/Mutex.scala#L127-L129

durban:
Yeah, I think that's the same problem as the one was with `Queue`. Probably could be fixed the same way...

armanbilge:
Oh right. How was that one fixed? 😆

djspiewak (approved):

PR #3551, opened 2023-04-24
Throw a custom exception for test timeouts

### LINKED ISSUE #3503: Timed out tests doesn't fail the sbt task

armanbilge:
Fixes https://github.com/typelevel/cats-effect/issues/3503, which is getting to be ridiculous and quite harmful actually.

This PR is going to be fun:tm: ... now we find out how many of our tests have been secretly failing 😕

### DIFF
--- a/tests/js/src/main/scala/cats/effect/DetectPlatform.scala
+++ b/tests/js/src/main/scala/cats/effect/DetectPlatform.scala
@@ -36,5 +36,6 @@ trait DetectPlatform {
   }
 
   def isJS: Boolean = true
+  def isJVM: Boolean = false
   def isNative: Boolean = false
 }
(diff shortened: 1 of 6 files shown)


armanbilge:
Ermm 🥴 
```
[error]     x not lose cedeing threads from the bypass when blocker transitioning
[error]      List(60, 60, 60) == 'List(60, 60, 60)' (IOPlatformSpecification.scala:296)
```
https://github.com/typelevel/cats-effect/actions/runs/4782639212/jobs/8502159804?pr=3551#step:23:3257

durban:
That's strange... I guess, in theory, if one thread is absurdly slow, it could happen(?). I'm not sure...

djspiewak (approved):

PR #3552, opened 2023-04-24
Add Apple Silicon to CI matrix

armanbilge:
We're already testing both JVM and Native on Linux ARM. Do we want to test JVM on Apple Silicon as well?


djspiewak:
If it doesn't cost us anything, yes. Apple Silicon is subtly different from other ARM64 chips in a few ways, so it's worth at least verifying.

armanbilge:
> If it doesn't cost us anything, yes.

It'll probably "cost" us CI time, not sure what the concurrency limits are for macOS. But so far the Cirrus jobs always seem to complete before the GHA stuff anyway.

PR #3555, opened 2023-04-24
Fix mutex cancelled acquire even more

### LINKED ISSUE #3554: In theory, waiters can be lost in `Mutex`
If a waiter is notified at the exact moment that it is canceled, it's possible for that wake-up to be lost, which can leave the overall mutex in an inconsistent state.

durban:
Hopefully should fix #3554. Same trick as for `Queue`.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Mutex.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Mutex.scala
@@ -21,6 +21,8 @@ package std
 import cats.effect.kernel._
 import cats.syntax.all._
 
+import scala.annotation.tailrec
+
 import java.util.concurrent.atomic.AtomicBoolean
 
 /**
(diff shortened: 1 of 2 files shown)


djspiewak:
Btw should probably use `.timeoutTo` inside of the test where you're expecting not-a-hang, just to sidestep the Specs issue

djspiewak:
Btw, I've been seeing a lot of `MutexSpec` failures on #3557. I'm thinking we merge *this* PR and then we can work on trying to bring that one in. Unfortunately, I suspect we've had a lot more CI failures than we thought masked by this Specs2 bug.

armanbilge:
Once this lands if we publish a snapshot I can test skunk again.

djspiewak:
@durban Is this ready to go? It's still marked as draft so I was holding off on doing things with it.

durban:
@djspiewak Yeah, sorry, I've kept it as draft because I was expecting CI failures (I had to figure out an appropriate timeout length). Should be good now.

In theory the longer timeout here could help with the `MutexSpec` failures in #3557 (I suspect the existing "not deadlock when highly contended" test was secretly already timing out in CI).

djspiewak (approved):

armanbilge:
Nice, 3f61234 appears to have fixed Skunk. Thank you!! 🙏

PR #3556, opened 2023-04-24
Fix problem with nextGaussian test

antoniojimeneznieto:
The probability of test error is considerably high. Expecting the mean and variance of all Gaussians to be simply finite should solve the problems.

### DIFF
--- a/tests/shared/src/test/scala/cats/effect/std/RandomSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/std/RandomSpec.scala
@@ -173,7 +173,7 @@ class RandomSpec extends BaseSpec {
           mean = gaussians.sum / sampleSize
           variance = gaussians.map(x => math.pow(x - mean, 2)).sum / sampleSize
           stddev = math.sqrt(variance)
-        } yield math.abs(mean) < 0.1 && math.abs(stddev - 1.0) < 0.1
+        } yield math.abs(mean).isFinite && math.abs(stddev - 1.0).isFinite
       }
     }
 
@@ -243,7 +243,7 @@ class RandomSpec extends BaseSpec {
 
     "shuffleList" >> {
       "shuffle a list" in real {
-        val sampleSize: Integer = 10000
+        val sampleSize: Int = 10000 // In case of modification, consider the probability of error
         val list: List[Int] = (1 to sampleSize).toList
         for {
           random <- randomGen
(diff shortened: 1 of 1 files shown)


tests/shared/src/test/scala/cats/effect/std/RandomSpec.scala
```diff
@@ -173,7 +173,7 @@ class RandomSpec extends BaseSpec {
           mean = gaussians.sum / sampleSize
           variance = gaussians.map(x => math.pow(x - mean, 2)).sum / sampleSize
           stddev = math.sqrt(variance)
-        } yield math.abs(mean) < 0.1 && math.abs(stddev - 1.0) < 0.1
+        } yield math.abs(mean).isFinite && math.abs(stddev - 1.0).isFinite
```
djspiewak: Is `isFinite` actually right though? My understanding of the test was that `0.1` was the tolerance?

armanbilge: I mean, depends what we want to test here. If we want to test these are actual Gaussians, then sure we need to find a way to balance tolerance with an acceptable probability of failure.

But if we just want to test that this code runs and links and produces not completely broken results (e.g. `Double.NaN`), then `isFinite` should be sufficient and should virtually never fail.

antoniojimeneznieto: Correct! The idea is, we have a Gaussian of mean 0 and variance 1. We ask it to give us 1000 random values following the distribution of that Gaussian (so the values will usually be around 0 with a variance of 1). 

Therefore, if this function really gives number with that distribution, the mean of those 1000 numbers should be around 0 and the variance around 1. 

We compute the overall mean and variance and we check that the value is not more than 0.1 away from the expected mean (0) and expected variance (1)

But the probability that the mean is greater than 0.1 is quiet high especially with only 1000 values

antoniojimeneznieto: Screenshot 2023-04-24 at 23 19 16

antoniojimeneznieto: As @armanbilge says, if we want to prove that they are gaussian we would have to define that exact tolerance

djspiewak:
Looks like `isFinite` was introduced in 2.13. Might want to do something more explicit, such as `!= Double.NaN`

armanbilge:
`!= Double.NaN` is always true 😛 we can use `java.lang.Double.isFinite(...)` instead.

```
scala> Double.NaN != Double.NaN
val res0: Boolean = true
```

armanbilge (approved):
Thanks a bunch!

antoniojimeneznieto:
Thanks to both of you for the help and feedback :)

PR #10, opened 2017-04-22
Improvements for IO#shift, expose TrampolinedContext

alexandru:
Sending `Runnable` instances in thread-pools is expensive. The implementation of `shift` ends up sending 2 runnable instances in that `ExecutionContext`, which is wasteful.

This PR introduces a `cats.effect.internals.TrampolineContext`:

1. We can avoid the second async boundary in `IO#shift` (when calling the resulting callback).
2. ~~IMO it's useful for users as well, because async stuff needs async boundaries, especially because we expose a stack-unsafe `Async.async` - if you want we can make this private, but I think it's a useful implementation to expose to users - Scala also has its own private implementation used for Scala's `Future`, there's a proliferation of private implementations going around.~~ EDIT: the implementation has been marked private for now

Thus this introduces an optimization for:

1. `IO#shift`
2. `Effect#shift`

Other changes:

- ~~exposed `NonFatal` in `cats.effect.util.NonFatal`, because we keep doing this all over the place and we might as well depend on one implementation~~
- renamed `implicit EC: ExecutionContext` to `implicit ec: ExecutionContext` (lowercase), because that's the naming convention everywhere else in *Scala Land*
- introduced a `cats.effect.laws.util.TestContext`, which can be used for cross-platform asynchronous tests

~~Also guys, the default implementation of `Effect#shift` was wrong and couldn't work on top of Javascript. Stop making use of `unsafeRunSync` in the implementation.~~

PR #3562, opened 2023-04-25
New `AsyncMutex` implementation

### LINKED ISSUE #3564: `Async` mutex corrupts FIFO on happy(ish) path
I want to capture this test from https://github.com/typelevel/cats-effect/pull/3562 as a separate issue.

What this tests is, if you have a bunch of waiters queued up on an unavailable mutex, what happens if one of those waiters cancels?

In the current implementation, the next waiter will be moved to the back of the queue. More specifically, in this test when waiter 0 cancels, waiter 1 is moved after waiter 10 (!!). Arguably this is all happy(ish) path: there are no race conditions at play, just a cancelation, but the result is a significant corruption of FIFO, IMHO.

https://github.com/typelevel/cats-effect/blob/2c355db7162894faa58923cf9e196753f02e1193/tests/shared/src/test/scala/cats/effect/std/MutexSpec.scala#L172-L195

BalmungSan:
I was half-following the recent discussions about the `Mutex` issue and on an internal conversation with Arman, we both concluded that the **current** version of `AsyncMutex` was hard to follow, whereas the two **previous** versions were "easier" _(it still requires a lot of comments and some mental force but still it is more straightforward IMHO)_. There is also the point that the **previous** versions always preserved fairness _(FIFO semantics)_, contrary to **current**.

Thus, considering that the main rationale for the implementation change was performance _(especially on the happy path)_ I was curious and decided to re-run the benchmarks. Since the version Arman introduced on #3409 was heavily modified, I expected some changes.

---

For clarification, during the lifetime of this PR, we have had the following versions:
+ **current** is the version currently on _main_ `series/3.x`
+ **previous** is the final version presented on #3346 _(this version has been discarded since a new test shows it has a serious bug which would allow multiple acquires of the `Mutex`)_
+ **deferred** is the original version presented on #3346 _(this version only requires `Concurrent` so it can also replace **semaphore** implementation)_
+ **new** is the final version presented on this PR _(it is based on **deferred** but taking advantage of `Async`)_

---

Right now we _(Arman & Luis)_ propose to use either **deferred** or **new**, since both pass all the current tests _(including the FIFO order one)_ as well as being easier to understand _(in our humble opinion)_.

---

### Current

| Benchmark | (fibers) | (iterations) | Mode | Cnt | Score | Error | Units |
| - | - | - | - | - | - | - | - |
| MutexBenchmark.happyPathAsync      |  10 | 1000 | thrpt | 20 | 219.723 | ± 0.334 | ops/s |
| MutexBenchmark.happyPathAsync      |  50 | 1000 | thrpt | 20 |  44.040 | ± 0.563 | ops/s |
| MutexBenchmark.happyPathAsync      | 100 | 1000 | thrpt | 20 |  22.485 | ± 0.152 | ops/s |
| - | - | - | - | - | - | - | - |
| MutexBenchmark.highContentionAsync |  10 | 1000 | thrpt | 20 |  31.041 | ± 1.161 | ops/s |
| MutexBenchmark.highContentionAsync |  50 | 1000 | thrpt | 20 |   9.442 | ± 0.129 | ops/s |
| MutexBenchmark.highContentionAsync | 100 | 1000 | thrpt | 20 |   5.477 | ± 0.131 | ops/s |
| - | - | - | - | - | - | - | - |
| MutexBenchmark.cancellationAsync   |  10 | 1000 | thrpt | 20 |  21.953 | ± 0.809 | ops/s |
| MutexBenchmark.cancellationAsync   |  50 | 1000 | thrpt | 20 |   7.899 | ± 0.202 | ops/s |
| MutexBenchmark.cancellationAsync   | 100 | 1000 | thrpt | 20 |   4.844 | ± 0.034 | ops/s |

### Previous / Deferred

| Benchmark | (fibers) | (iterations) | Mode | Cnt | Score | Error | Units |
| - | - | - | - | - | - | - | - |
| MutexBenchmark.happyPathAsync      |  10 | 1000 | thrpt | 20 | 213.143 | ± 1.696 | ops/s |
| MutexBenchmark.happyPathAsync      |  50 | 1000 | thrpt | 20 |  44.325 | ± 0.215 | ops/s |
| MutexBenchmark.happyPathAsync      | 100 | 1000 | thrpt | 20 |  22.191 | ± 0.215 | ops/s |
| - | - | - | - | - | - | - | - |
| MutexBenchmark.highContentionAsync |  10 | 1000 | thrpt | 20 |  34.244 | ± 1.025 | ops/s |
| MutexBenchmark.highContentionAsync |  50 | 1000 | thrpt | 20 |   8.179 | ± 0.230 | ops/s |
| MutexBenchmark.highContentionAsync | 100 | 1000 | thrpt | 20 |   4.424 | ± 0.028 | ops/s |
| - | - | - | - | - | - | - | - |
| MutexBenchmark.cancellationAsync   |  10 | 1000 | thrpt | 20 |  22.864 | ± 0.901 | ops/s |
| MutexBenchmark.cancellationAsync   |  50 | 1000 | thrpt | 20 |   7.927 | ± 0.066 | ops/s |
| MutexBenchmark.cancellationAsync   | 100 | 1000 | thrpt | 20 |   4.895 | ± 0.045 | ops/s |

### New

| Benchmark | (fibers) | (iterations) | Mode | Cnt | Score | Error | Units |
| - | - | - | - | - | - | - | - |
| MutexBenchmark.happyPathAsync      |  10 | 1000 | thrpt | 20 | 183.778 | ± 0.156 | ops/s |
| MutexBenchmark.happyPathAsync      |  50 | 1000 | thrpt | 20 |  37.519 | ± 0.046 | ops/s |
| MutexBenchmark.happyPathAsync      | 100 | 1000 | thrpt | 20 |  18.778 | ± 0.121 | ops/s |
| - | - | - | - | - | - | - | - |
| MutexBenchmark.highContentionAsync |  10 | 1000 | thrpt | 20 |  32.455 | ± 1.579 | ops/s |
| MutexBenchmark.highContentionAsync |  50 | 1000 | thrpt | 20 |   8.192 | ± 0.276 | ops/s |
| MutexBenchmark.highContentionAsync | 100 | 1000 | thrpt | 20 |   4.486 | ± 0.075 | ops/s |
| - | - | - | - | - | - | - | - |
| MutexBenchmark.cancellationAsync   |  10 | 1000 | thrpt | 20 |  24.626 | ± 1.514 | ops/s |
| MutexBenchmark.cancellationAsync   |  50 | 1000 | thrpt | 20 |   8.150 | ± 0.073 | ops/s |
| MutexBenchmark.cancellationAsync   | 100 | 1000 | thrpt | 20 |   5.046 | ± 0.089 | ops/s |

---

For context, here are the results of the **semaphore** implementation for all runs.

### Current

```
[info] MutexBenchmark.happyPathConcurrent             10          1000  thrpt   20  198.903 ± 0.638  ops/s
[info] MutexBenchmark.happyPathConcurrent             50          1000  thrpt   20   40.689 ± 0.348  ops/s
[info] MutexBenchmark.happyPathConcurrent            100          1000  thrpt   20   20.214 ± 0.116  ops/s
[info] MutexBenchmark.highContentionConcurrent        10          1000  thrpt   20   31.392 ± 1.198  ops/s
[info] MutexBenchmark.highContentionConcurrent        50          1000  thrpt   20    7.466 ± 0.156  ops/s
[info] MutexBenchmark.highContentionConcurrent       100          1000  thrpt   20    4.014 ± 0.067  ops/s
[info] MutexBenchmark.cancellationConcurrent          10          1000  thrpt   20   21.554 ± 0.651  ops/s
[info] MutexBenchmark.cancellationConcurrent          50          1000  thrpt   20    7.455 ± 0.049  ops/s
[info] MutexBenchmark.cancellationConcurrent         100          1000  thrpt   20    3.619 ± 0.056  ops/s
```

### Previous / Deferred

```
[info] MutexBenchmark.happyPathConcurrent             10          1000  thrpt   20  198.903 ± 0.638  ops/s
[info] MutexBenchmark.happyPathConcurrent             50          1000  thrpt   20   40.689 ± 0.348  ops/s
[info] MutexBenchmark.happyPathConcurrent            100          1000  thrpt   20   20.214 ± 0.116  ops/s
[info] MutexBenchmark.highContentionConcurrent        10          1000  thrpt   20   31.392 ± 1.198  ops/s
[info] MutexBenchmark.highContentionConcurrent        50          1000  thrpt   20    7.466 ± 0.156  ops/s
[info] MutexBenchmark.highContentionConcurrent       100          1000  thrpt   20    4.014 ± 0.067  ops/s
[info] MutexBenchmark.cancellationConcurrent          10          1000  thrpt   20   21.554 ± 0.651  ops/s
[info] MutexBenchmark.cancellationConcurrent          50          1000  thrpt   20    7.455 ± 0.049  ops/s
[info] MutexBenchmark.cancellationConcurrent         100          1000  thrpt   20    3.619 ± 0.056  ops/s
```

### New

```
[info] MutexBenchmark.happyPathConcurrent             10          1000  thrpt   20  199.503 ± 0.725  ops/s
[info] MutexBenchmark.happyPathConcurrent             50          1000  thrpt   20   40.464 ± 0.115  ops/s
[info] MutexBenchmark.happyPathConcurrent            100          1000  thrpt   20   20.236 ± 0.263  ops/s
[info] MutexBenchmark.highContentionConcurrent        10          1000  thrpt   20   32.125 ± 1.281  ops/s
[info] MutexBenchmark.highContentionConcurrent        50          1000  thrpt   20    7.467 ± 0.117  ops/s
[info] MutexBenchmark.highContentionConcurrent       100          1000  thrpt   20    3.934 ± 0.026  ops/s
[info] MutexBenchmark.cancellationConcurrent          10          1000  thrpt   20   21.359 ± 0.838  ops/s
[info] MutexBenchmark.cancellationConcurrent          50          1000  thrpt   20    7.438 ± 0.102  ops/s
[info] MutexBenchmark.cancellationConcurrent         100          1000  thrpt   20    3.432 ± 0.017  ops/s
```

---

> Lies, Damn Lies, and Benchmarks

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Mutex.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Mutex.scala
@@ -106,8 +106,8 @@ object Mutex {
             if (currentCell eq ConcurrentImpl.Empty) F.pure(ourCell)
             else {
               F.onCancel(
-                poll(currentCell.asInstanceOf[ConcurrentImpl.Next[F]].cell.get),
-                ourCell.cell.complete(currentCell).void
+                poll(currentCell.asInstanceOf[ConcurrentImpl.Next[F]].get),
+                ourCell.complete(currentCell).void
               ).flatMap { nextCell => loop(currentCell = nextCell) }
             }
 
@@ -126,7 +126,7 @@ object Mutex {
           if (lastCell eq thisCell) setter(ConcurrentImpl.Empty)
           else F.pure(false)
       } flatMap {
-        case false => thisCell.cell.complete(ConcurrentImpl.Empty).void
+        case false => thisCell.complete(ConcurrentImpl.Empty).void
         case true => F.unit
       }
 
@@ -139,14 +139,15 @@ object Mutex {
 
   private object ConcurrentImpl {
     // Represents a queue of waiters for the mutex.
-    private[Mutex] sealed abstract class LockQueue
+    private[Mutex] final type LockQueue = AnyRef
     // Represents the first cell of the queue.
-    private[Mutex] object Empty extends LockQueue
+    private[Mutex] final type Empty = LockQueue
+    private[Mutex] final val Empty: Empty = null
     // Represents a cell in the queue of waiters.
-    private[Mutex] final class Next[F[_]](val cell: Deferred[F, LockQueue]) extends LockQueue
+    private[Mutex] final type Next[F[_]] = Deferred[F, LockQueue]
 
     private[Mutex] def LockQueueCell[F[_]](implicit F: Concurrent[F]): F[Next[F]] =
-      Deferred[F, LockQueue].map(df => new Next(cell = df))
+      Deferred[F, LockQueue]
   }
 
   private final class TransformedMutex[F[_], G[_]](
--- a/std/shared/src/main/scala/cats/effect/std/Mutex.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Mutex.scala
@@ -21,8 +21,6 @@ package std
 import cats.effect.kernel._
 import cats.syntax.all._
 
-import java.util.concurrent.atomic.AtomicReference
-
 /**
  * A purely functional mutex.
  *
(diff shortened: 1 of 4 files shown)


tests/shared/src/test/scala/cats/effect/std/MutexSpec.scala
```diff
@@ -147,5 +147,51 @@ final class MutexSpec extends BaseSpec with DetectPlatform {
...
+            _ <- f2.cancel
+            _ <- f1Release
+            _ <- t.join
+            r <- ref.get
+          } yield r.reverse
+        }
+      }
+
+      p must completeAs(numbers)
+    }
```
armanbilge: This test passes for this PR but fails with the current implementation on `series/3.x` because waiter #1 moves to the back of the line after #10. IMHO that is a pretty significant corruption of fairness esp. considering there are no race conditions at all here, this test is fully `ticked`.

BalmungSan:
# DO NOT MERGE!

I just realized this version has a serious bug

If you have: F1 holding the Mutex and F2. F3, F4 in line waiting.
And then you cancel F2, and then F3, then that last one would think the `Mutex` is free and awake F4.

std/shared/src/main/scala/cats/effect/std/Mutex.scala
```diff
@@ -96,86 +96,115 @@ object Mutex {
...
+      }
 
-    private[this] val _release: F[Unit] = F.delay {
-      locked.set(false) // release
-      notifyOne() // try to wake someone
-    }
+    // Acquires the Mutex.
+    private def acquire(poll: Poll[F]): F[AsyncImpl.LockCell] =
+      F.delay(new AtomicReference[AsyncImpl.CB]()).flatMap { thisCell =>
+        poll(await(thisCell).map(_ => thisCell))
```
durban: If this is cancelled after the `asyncCheckAttempt` (in `await`), but before the `.map`, I think it leaves the mutex in an invalid state.

durban:
@BalmungSan I've set this as "draft" due to your "do not merge" comment. I've also left another comment.

std/shared/src/main/scala/cats/effect/std/Mutex.scala
BalmungSan: I don't think that is the case. If we get canceled during that point, it should call the `cancel` method rather than the `release` one. And the `Mutex` was not acquired at all.

Anyways, this whole implementation will be trashed and I will start again based on this one: https://github.com/typelevel/cats-effect/pull/3346/commits/d7c32347dbb734830a11ab8e2295514a0ccc178d

durban: `cancel` is called by the finalizer of the `asyncCheckAttempt`, so if cancellation happens _after_ the `asyncCheckAttempt`, it will not be called. (But of course if you replace the whole implementation, it won't matter.)

BalmungSan:
As expected, this PR version of `Mutex` fails the new test.
FWIW, the current version on `series/3.x` does pass the test! 🎉
cc @durban @armanbilge @djspiewak

BalmungSan:
So, the _"original"_ `Deferred` based lock-chain idea pass all the current tests.
I am pushing it as a milestone, I think I can optimize it a little bit but just in case.
If we decide to keep this version, since it only needs `Concurrent` we can just replace the `Semaphore` one with this.

BalmungSan:
I just updated the PR description with the benchmark results for **current**, **previous** _(**deferred**)_ & **new**. And it seems the performance of all versions is very similar.

Thus, I think the best is to just use the **deferred** one and remove the **semaphore** one.

cc: @durban @armanbilge @djspiewak

djspiewak:
Yeah this is super-duper close. It definitely feels like it's best to pick the simplest one with the most reasonable semantics.

BalmungSan:
Okay, so I reverted and unified both the `Async` & `Concurrent` implementations to the original `LockChain` idea.
I fleshed out the naming a bit, added a couple of comments, and made some minor improvements like using `eq`, avoiding wrapper allocation, and the like. Plus, a minor optimization on `release` for when there is no one waiting for the `Mutex`.

### Final version benchmark results

| Benchmark | (fibers) | (iterations) | Mode | Cnt | Score | Error | Units |
| - | - | - | - | - | - | - | - |
| MutexBenchmark.happyPathConcurrent      |  10 | 1000 | thrpt | 20 | 213.946 | ± 1.116 | ops/s |
| MutexBenchmark.happyPathConcurrent      |  50 | 1000 | thrpt | 20 |  44.099 | ± 0.356 | ops/s |
| MutexBenchmark.happyPathConcurrent      | 100 | 1000 | thrpt | 20 |  21.624 | ± 0.089 | ops/s |
| - | - | - | - | - | - | - | - |
| MutexBenchmark.highContentionConcurrent |  10 | 1000 | thrpt | 20 |  33.514 | ± 1.624 | ops/s |
| MutexBenchmark.highContentionConcurrent |  50 | 1000 | thrpt | 20 |   7.655 | ± 0.363 | ops/s |
| MutexBenchmark.highContentionConcurrent | 100 | 1000 | thrpt | 20 |   4.203 | ± 0.024 | ops/s |
| - | - | - | - | - | - | - | - |
| MutexBenchmark.cancellationConcurrent   |  10 | 1000 | thrpt | 20 |  22.853 | ± 1.031 | ops/s |
| MutexBenchmark.cancellationConcurrent   |  50 | 1000 | thrpt | 20 |   7.895 | ± 0.176 | ops/s |
| MutexBenchmark.cancellationConcurrent   | 100 | 1000 | thrpt | 20 |   4.997 | ± 0.080 | ops/s |

c.c. @djspiewak @armanbilge @durban

djspiewak (approved):

durban (approved):

PR #3563, opened 2023-04-26
fixing up additional dispatcher flakey tests

### LINKED ISSUE #3312: `DispatcherSpec` flake
```
[error]     x wait for the completion of the active fibers
[error]      the value is false (DispatcherSpec.scala:303)
```

samspills:
resolves #3312 

- fixed the flaky `reject new tasks while shutting down` test by approaching the test case in a slightly different way, see my comment for more details
- the other test mentioned in #3312, `wait for the completion of the active fibers` appears to no longer be flaky?
  - I refactored the test lightly to add a `.replicateA(1000)` and then ran that a few times and it did not flake
  - I ran the test on each version of CE, in reverse, and I can't replicate a flake until 3.4.4
    - I don't see anything in the [release notes for 3.4.5](https://github.com/typelevel/cats-effect/releases/tag/v3.4.5) that would have obviously affected this particular flake (and I saw it flake in March but maybe I can't remember, maybe I was accidentally on an old branch?)

@armanbilge have you seen the `wait for the completion of the active fibers` test flake anywhere lately?

### DIFF
--- a/tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
@@ -326,30 +326,35 @@ class DispatcherSpec extends BaseSpec with DetectPlatform {
     }
 
     "reject new tasks while shutting down" in real {
-      (IO.ref(false), IO.ref(false)).flatMapN { (resultR, rogueResultR) =>
-        dispatcher
-          .allocated
-          .flatMap {
-            case (runner, release) =>
-              IO(runner.unsafeRunAndForget(
-                IO.sleep(1.second).uncancelable.guarantee(resultR.set(true)))) *>
-                IO.sleep(100.millis) *>
-                release.both(
-                  IO.sleep(500.nanos) *>
-                    IO(runner.unsafeRunAndForget(rogueResultR.set(true))).attempt
-                )
-          }
-          .flatMap {
-            case (_, rogueSubmitResult) =>
-              for {
-                result <- resultR.get
-                rogueResult <- rogueResultR.get
-                _ <- IO(result must beTrue)
-                _ <- IO(rogueResult must beFalse)
-                _ <- IO(rogueSubmitResult must beLeft)
-              } yield ok
-          }
-      }
+      (IO.ref(false), IO.ref(false))
+        .flatMapN { (resultR, rogueResultR) =>
+          dispatcher
+            .allocated
+            .flatMap {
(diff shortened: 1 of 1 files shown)


djspiewak:
Well, the one problem with `ticked` is you can't usually use it to reproduce race conditions. I haven't had enough coffee yet to figure out whether your test is actually testing a race condition. If it isn't, then `ticked` is definitely fine and will be a lot more deterministic.

samspills:
Yeahhh :( I was chatting with Arman and we were thinking that its less about a race condition and we could do this in ticked if we increase the interleaving between the release and the rogue task with cedes

That said, I still can't get the test to fail with the pre-3.4.9 version of Dispatcher (where `alive` is _not_ the first resource released) so we might be wrong about that and it does need to be in `real`

tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
```diff
@@ -326,30 +326,35 @@ class DispatcherSpec extends BaseSpec with DetectPlatform {
...
+              case (_, rogueSubmitResult) =>
+                for {
+                  result <- resultR.get
+                  rogueResult <- rogueResultR.get
+                  _ <- IO(result must beTrue)
+                  _ <- IO(rogueResult match {
+                    // if the rogue task is not completed then we _must_ have failed to submit it
+                    case false => rogueSubmitResult must beLeft
+                    case true => rogueSubmitResult must beRight
+                  })
```
samspills, line 353: The original flake that Arman pointed out in the #3312 is caused by the `rogueResult` being successfully submitted **and** completing correctly. We kind of discussed this in the original PR for this work, but to recap: I don't think that situation is actually a problem. If a task gets in under the wire, that's fine. What's very-much-not-fine is if the task is submitted and never done. 

So with that in mind I rewrote this check to require that if the task isn't completed, then we must have failed to submit the task (and if it is completed then we obviously succeeded in submitting the result).

HOWEVER this does have the affect of moving the flake into the other direction. If I move the `alive` resource in `Dispatcher` back to its pre-3.4.9 location this test can occasionally pass on the parallel case, simply by luck. To counteract this I've added a `.replicateA(5)`. With the repeated attempts, we reduce the chance of the test passing incorrectly by .... some amount 😅 I've rerun many times and I haven't yet seen wayward test success since adding the replication.

armanbilge (approved):
Excellent investigation and analysis, nice work!

PR #3565, opened 2023-04-26
Add raceN / firstCompletedOf recipe

BalmungSan:
Motivated by a recent discussion on **Discord**.

Since this method can have many variations and is not very common, I guess a recipe is the best place to add it.

### DIFF
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -187,7 +187,7 @@ def raceN[A](ios: List[IO[A]]): IO[Option[A]] =
 
     val fallbackResult = d.complete(None)
 
-    (runAll >> fallbackResult).background.surround(d.get)
+    runAll.guarantee(fallbackResult).background.surround(d.get)
   }
 ```
 
@@ -207,7 +207,7 @@ def raceN[A](ios: List[IO[A]]): IO[Either[Throwable, A]] =
     // You may use an especial custom exception, or combine the Either + Option.
     val fallbackResult = d.complete(new Exception("No IO completed"))
 
-    (runAll >> fallbackResult).background.surround(d.get)
+    runAll.guarantee(fallbackResult).background.surround(d.get)
   }
 ```
 
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -5,7 +5,7 @@ title: Recipes
 
 ## Start a supervised task that outlives the creating scope
 
-If you need to run an action in a fiber in a "start-and-forget" manner, you'll want to use [Supervisor](std/supervisor.md). 
+If you need to run an action in a fiber in a "start-and-forget" manner, you'll want to use [Supervisor](std/supervisor.md).
 This lets you safely evaluate an effect in the background without waiting for it to complete and ensuring that the fiber and all its resources are cleaned up at the end.
 You can configure a`Supervisor` to  wait for all supervised fibers to complete at the end its lifecycle, or to simply cancel any remaining active fibers.
 
@@ -74,7 +74,7 @@ import cats.effect.std.Supervisor
 object Server extends IOApp.Simple {
 
   def handler(supervisor: Supervisor[IO]): Request => IO[Response] = {
-    case Request("start", params) => 
+    case Request("start", params) =>
       supervisor.supervise(longRunningTask(params)).void >> IO.pure(Ok("started"))
     case Request(_, _) => IO.pure(NotFound)
   }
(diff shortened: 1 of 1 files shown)


docs/recipes.md
```diff
@@ -153,10 +153,66 @@ class MaxProxy(atomicCell: AtomicCell[IO, Double], requestService: Service) {
...
+```scala mdoc:silent
+def raceN[A](ios: List[IO[A]]): IO[Option[A]] =
+  Deferred[IO, Option[A]].flatMap { d =>
+    val runAll = ios.parTraverse_ { io =>
+      io.flatMap(a => d.complete(Some(a)))
+    }
+
+    val fallbackResult = d.complete(None)
+
+    (runAll >> fallbackResult).background.surround(d.get)
```
armanbilge: right?

```suggestion
    (runAll.guarantee(fallbackResult)).background.surround(d.get)
```

docs/recipes.md
```diff
@@ -153,10 +153,66 @@ class MaxProxy(atomicCell: AtomicCell[IO, Double], requestService: Service) {
...
+def raceN[A](ios: List[IO[A]]): IO[Either[Throwable, A]] =
+  Deferred[IO, Either[Throwable, A]].flatMap { d =>
+    val runAll = ios.parTraverse_ { io =>
+      io.attempt.flatMap(d.complete)
+    }
+
+    // You may use an especial custom exception, or combine the Either + Option.
+    val fallbackResult = d.complete(new Exception("No IO completed"))
+
+    (runAll >> fallbackResult).background.surround(d.get)
```
armanbilge:
```suggestion
    (runAll.guarantee(fallbackResult)).background.surround(d.get)
```

docs/recipes.md
```diff
@@ -153,10 +153,66 @@ class MaxProxy(atomicCell: AtomicCell[IO, Double], requestService: Service) {
...
+
+**We want the first completed instead of the first success**
+
+What if instead of waiting for the first success,
+you want the first completed `IO`, even if it completes with an error.
+Similar that the previous case, we can just use an `Either` & `attempt`:
+
+```scala mdoc:silent
+def raceN[A](ios: List[IO[A]]): IO[Either[Throwable, A]] =
+  Deferred[IO, Either[Throwable, A]].flatMap { d =>
```
ChristopherDavenport, line 202: Should stop using this type generally and prefer the use of Outcome. This ends in a lot of weird deadlocks if folks use this without considering cancelation. So generally I say outcome or not at all.

BalmungSan: Well, the example mentions that the hypothetical user only cares about either success or error.
We may add a third example using `Outcome` or just being more explicit in the final note that another option is to consider cancelation _(I prefer this one)_.

docs/recipes.md
```diff
@@ -153,10 +153,66 @@ class MaxProxy(atomicCell: AtomicCell[IO, Double], requestService: Service) {
...
-  
+
   def getHistoryMax(): IO[Double] = atomicCell.get
 }
+```
 
+## How to implement raceN / firstCompletedOf
 
+Assume you have a bunch of `IOs` that you want to run all in parallel and get the first success;
+canceling all the others.
```
armanbilge, line 164: Can I throw one more implementation into the ring?

```scala
def raceN[A](ios: List[IO[A]]): IO[A] = ios match {
  case Nil => IO.raiseError[A](new NoSuchElementException)
  case head :: tail => tail.foldLeft(head)(_.race(_).map(_.merge))
}
```

Edit: fixed 😇

BalmungSan: I think this was OP's original implementation _(or something similar)_
It really doesn't feel natural, at least to me.

docs/recipes.md
```diff
@@ -153,10 +153,66 @@ class MaxProxy(atomicCell: AtomicCell[IO, Double], requestService: Service) {
...
+
+```scala mdoc:silent
+def raceN[A](ios: List[IO[A]]): IO[A] =
+  Deferred[IO, A].flatMap { d =>
+    ios.parTraverse_(io => io.flatMap(d.complete)).background.surround(d.get)
+  }
```
djspiewak, line 170: Playing more golf:

```suggestion
  IO.deferred[Outcome[IO, Throwable, A]] flatMap { d =>
    ios.traverse(io => io.guaranteeCase(d.complete(_).void).background)
      .surround(d.get)
      .flatMap(_.embedNever)
  }
```

Though maybe I'm putting the suggestion in the wrong spot, since this isn't in any way the simplest implementation.

PR #3567, opened 2023-04-26
Make blockedThreadDetectionEnabled configurable via a system property

### LINKED ISSUE #3561: Make thread-blocking detector configurable via system property
Currently it's configured in `IOApp`, but it would be even better if this value can also be loaded from a system property at runtime.

https://github.com/typelevel/cats-effect/blob/538d50df86beb23e09b9a176e556b168beec6b5c/core/jvm/src/main/scala/cats/effect/IOApp.scala#L254

We have a bunch of examples of system properties here e.g.
https://github.com/typelevel/cats-effect/blob/538d50df86beb23e09b9a176e556b168beec6b5c/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeConfigCompanionPlatform.scala#L63

chunjef:
Resolves #3561.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/IOApp.scala
+++ b/core/jvm/src/main/scala/cats/effect/IOApp.scala
@@ -250,8 +250,14 @@ trait IOApp {
    *
    * This may be of interest if you've been getting warnings about CPU starvation printed to
    * stderr. [[https://typelevel.org/cats-effect/docs/core/starvation-and-tuning]]
+   *
+   * Can also be configured by setting the `cats.effect.detectBlockedThreads` system property.
    */
-  protected def blockedThreadDetectionEnabled: Boolean = false
+  protected def blockedThreadDetectionEnabled: Boolean =
+    Option(System.getProperty("cats.effect.detectBlockedThreads")).map(_.toLowerCase()) match {
+      case Some(value) => value.equalsIgnoreCase("true")
+      case None => false // default to disabled
+    }
 
   /**
    * Controls whether non-daemon threads blocking application exit are logged to stderr when the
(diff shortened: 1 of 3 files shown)


core/jvm/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -250,8 +250,14 @@ trait IOApp {
...
    * stderr. [[https://typelevel.org/cats-effect/docs/core/starvation-and-tuning]]
+   *
+   * Can also be configured by setting the `cats.effect.detectBlockedThreads` system property.
    */
-  protected def blockedThreadDetectionEnabled: Boolean = false
+  protected def blockedThreadDetectionEnabled: Boolean =
+    Option(System.getProperty("cats.effect.detectBlockedThreads")).map(_.toLowerCase()) match {
+      case Some(value) => value.equalsIgnoreCase("true")
+      case None => false // default to disabled
+    }
```
armanbilge: I wonder if we simplify this with `java.lang.Boolean.getBoolean` method.

https://docs.oracle.com/javase/8/docs/api/java/lang/Boolean.html#getBoolean-java.lang.String-

docs/core/io-runtime-config.md
```diff
@@ -25,6 +25,7 @@ This can be done for example with the [EnvironmentPlugin for Webpack](https://we
 
 | System Property (JVM) / ENV Variable (JS)                                                         | Value (default)    | Description                                                                                                                       |
 |---------------------------------------------------------------------------------------------------|--------------------|-----------------------------------------------------------------------------------------------------------------------------------|
+| `cats.effect.detectBlockedThreads` <br/> N/A                                                      | `Boolean` (`false`) | Whether or not we should detect blocked threads. |
```
armanbilge, line 28: 👍

armanbilge (commented):
Thanks for the PR!

core/jvm/src/main/scala/cats/effect/IOApp.scala
chunjef: It looks like using `getBoolean` would indeed be simpler. Updating the branch.

armanbilge (approved):
LGTM, thank you for your contribution!

Issue #3568, opened 2023-04-26
`IO.cede.foreverM.timeoutTo(...)` timeout flake


armanbilge:
Observed in e61c058efae982e7716c19c2d0f8a4c69eaed985.

```
[info]     timeoutTo should
[info]       + succeed
[error]       ! use fallback
[error]        cats.effect.TestTimeoutException: null (Runners.scala:109)
[error] cats.effect.Runners.$anonfun$timeout$1(Runners.scala:109)
[error] cats.effect.unsafe.WorkStealingThreadPool.$anonfun$sleep$1(WorkStealingThreadPool.scala:647)
[error] cats.effect.unsafe.WorkStealingThreadPool.$anonfun$sleep$1$adapted(WorkStealingThreadPool.scala:647)
[error] cats.effect.unsafe.WorkStealingThreadPool.stealTimers(WorkStealingThreadPool.scala:238)
[error] cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:599)
```

https://github.com/typelevel/cats-effect/actions/runs/4811306795/jobs/8565136011?pr=3565#step:23:3612

Is it just me, or I don't see a reasonable explanation for why this test should timeout after several seconds 🤔 

https://github.com/typelevel/cats-effect/blob/e61c058efae982e7716c19c2d0f8a4c69eaed985/tests/shared/src/test/scala/cats/effect/IOSpec.scala#L1783-L1793

djspiewak:
I think it's just that the runners are very slow and we're starting to see it a lot now. Honestly I want to just bump up the global timeout.

armanbilge:
Still flaking :(

https://github.com/typelevel/cats-effect/actions/runs/4839630759/jobs/8625021373?pr=3579#step:23:3613

PR #3579, opened 2023-04-29
dispatcher releasing itself rejects new tasks

### LINKED ISSUE #3501: Closing a `Dispatcher` from within it seems to enter invalid state (?)

samspills:
resolves #3501 

- adds a test to confirm that an illegal state exception is raised when attempting to submit tasks to a dispatcher after we've already submitted its own release action as a task
- confirmed that if I move the `alive` release in `Dispatcher` from line 315 up to 245, this test fails for sequential dispatchers
- I replicated the test 1000 times and it passed so I think 100 milliseconds is enough time for the parallel case and this won't be flaky. 

(I wasn't sure about release plans so I branched off of 3.x? This is just a test so not super important for releasing but let me know if I should switch the branch)

### DIFF
--- a/tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
@@ -346,10 +346,9 @@ class DispatcherSpec extends BaseSpec with DetectPlatform {
                   result <- resultR.get
                   rogueResult <- rogueResultR.get
                   _ <- IO(result must beTrue)
-                  _ <- IO(rogueResult match {
-                    // if the rogue task is not completed then we _must_ have failed to submit it
-                    case false => rogueSubmitResult must beLeft
-                    case true => rogueSubmitResult must beRight
+                  _ <- IO(if (rogueResult == false) {
+                    // if the rogue task is not completed then we must have failed to submit it
+                    rogueSubmitResult must beLeft
                   })
                 } yield ok
             }
--- a/tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
@@ -356,6 +356,15 @@ class DispatcherSpec extends BaseSpec with DetectPlatform {
         }
         .replicateA(5)
     }
+
+    "issue 3501: reject new tasks after release action is submitted as a task" in real {
+      dispatcher.allocated.flatMap {
+        case (runner, release) =>
+          IO(runner.unsafeRunAndForget(release)) *>
+            IO.sleep(100.millis) *>
+            IO(runner.unsafeRunAndForget(IO(ko)) must throwAn[IllegalStateException])
+      }
+    }
   }
 
   private def awaitTermination(dispatcher: Resource[IO, Dispatcher[IO]]) = {


armanbilge (dismissed):
Thank you Sam!

armanbilge:
I'm confused about this "reject new tasks while shutting down" flake ... seems to have flaked with the exact exception we are looking for?

https://cirrus-ci.com/task/4538803671531520?logs=main#L3252

samspills:
The failing test in CI is the IO timeoutTo test, which I think should already be fixed in 3.x. I'll double check that my branch is up to date before rerunning tests again (just going to be a little bit as I'm afk right now)

samspills:
Oh I didn't even look at that one!! Thanks for pointing that out Arman, I'll investigate that as well. It does seem weird...

armanbilge:
> The failing test in CI is the IO timeoutTo test, which I think should already be fixed in 3.x. I'll double check that my branch is up to date

Yeah, I just commented on https://github.com/typelevel/cats-effect/issues/3568#issuecomment-1528831131. You're not behind—CI runs on a synthetic merge commit so it is always up-to-date with the target branch at time of opening the PR. (This is also why the close-reopen trick re-triggers CI with a fresh merge that includes any new changes on the target branch.)

samspills:
I was able to reproduce the cirrus failure locally:
```
[error]     x reject new tasks while shutting down
[error]      Left(java.lang.IllegalStateException: dispatcher already shutdown) is not Right (DispatcherSpec.scala:352)
```

it seems pretty rare. The issue is that the rogue task is enqueued but then the second `alive` check in `unsafeToFutureCancelable` fails, which tries to cancel the task and raises the exception: https://github.com/typelevel/cats-effect/blob/38736ca1e35cf79eb6345bc65768b2d2ab5bc375/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala#L414-L420

Since the task here is so trivial, it's already finished so we end up in a situation where the task was completed even though the submission of it failed. I'm still working on addressing it.

tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
```diff
@@ -346,10 +346,9 @@ class DispatcherSpec extends BaseSpec with DetectPlatform {
                   result <- resultR.get
                   rogueResult <- rogueResultR.get
                   _ <- IO(result must beTrue)
-                  _ <- IO(rogueResult match {
-                    // if the rogue task is not completed then we _must_ have failed to submit it
-                    case false => rogueSubmitResult must beLeft
-                    case true => rogueSubmitResult must beRight
+                  _ <- IO(if (rogueResult == false) {
+                    // if the rogue task is not completed then we must have failed to submit it
+                    rogueSubmitResult must beLeft
```
samspills, line 351: As mentioned in my earlier comment, the previous version of this test could fail because the task was enqueued+completed but the submission of it actually failed during the final `alive` check. 

To deal with this I've tried to further reduce the scope of this test. I think all we want to test is that a task is not successfully submitted during the shutdown process. This was still the case for the previous failure (the submission failed! it's just that our task is so trivial that it was already finished instead of being cancelled) but the way the test was written treated it as a failure.

Arman and I are still chatting about the alive check and the task enqueueing, but that's definitely a different PR

djspiewak (approved):

PR #3583, opened 2023-05-01
Restore a tiny bit of type information for IOFiber#resumeIO

vasilmkd:
- Some time ago I was experimenting with making this a more general data field, but it didn't work out.


djspiewak (changes requested):
I've actually always wondered why it was an `AnyRef`. Shouldn't we be able to remove the cast in `execR` now?

djspiewak (approved):

PR #3591, opened 2023-05-02
Fix worker thread name index

durban:
Strangely, the _initial_ name of the `WorkerThreads` is one higher than their (zero-based) index in the WSTP (i.e., `io-compute-1`, `io-compute-2`, ... instead of `io-compute-0`, `io-compute-1`, ...). After a worker is converted to a blocker, and then returns to the WSTP, its name will have the _correct_ index (i.e., zero-based). Confusingly, this can cause two worker threads to have the _same_ name. E.g., originally `io-compute-2` (with index `1`) becomes `io-compute-1`, but there could be _another_ `io-compute-1` (with index `0`).

This PR fixes this inconsistency.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
@@ -98,7 +98,7 @@ private final class WorkerThread(
   private val indexTransfer: LinkedTransferQueue[Integer] = new LinkedTransferQueue()
   private[this] val runtimeBlockingExpiration: Duration = pool.runtimeBlockingExpiration
 
-  val nameIndex: Int = pool.blockedWorkerThreadNamingIndex.incrementAndGet()
+  val nameIndex: Int = pool.blockedWorkerThreadNamingIndex.getAndIncrement()
 
   // Constructor code.
   {
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
@@ -668,6 +668,9 @@ private final class WorkerThread(
         val idx = index
         val clone =
           new WorkerThread(idx, queue, parked, external, fiberBag, pool)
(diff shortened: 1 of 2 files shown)


armanbilge:
> Strangely, the _initial_ name of the `WorkerThreads` is one higher than their (zero-based) index in the WSTP

I have some recollection that the `nameIndex` and WSTP index are kind of independent, but I haven't really thought about it. See https://github.com/typelevel/cats-effect/pull/2687#discussion_r776535347, sorry if this is just noise :)

durban:
Hm, that's interesting. I'm not sure (any more) what was the original intention. I thought, that working workers _should_ have the WSTP index in their names, because when a cached thread becomes a worker, it is carefully set to that: https://github.com/typelevel/cats-effect/blob/series/3.x/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala#L845.

But apparently this is not true for new (replacement) workers. So now I made it so (my 2 new commits).

It is completely possible, that I _imagined_ this whole "name should have the WSTP index" thing. But seeing 2 threads with the same name in visualvm is pretty confusing (it did happen, that's how I found this). So this PR should fix that...

durban:
It just occurred to me, that it's completely possible (even likely), that the call in `init` is the wrong one (should be `setName(s"$prefix-$nameIdx")`), and the other 2 places I've "fixed" in this PR should be left alone. That would also fix the duplicate name problem (I think).

My approach (currently in this PR) has the advantage of having the WSTP workers name end with their WSTP index.

The (probably) originally intended approach has the advantage of threads having a stable postfix in their names, and it's just the prefix which changes when they move in/out of the WSTP.

armanbilge:
> The (probably) originally intended approach has the advantage of threads having a stable postfix in their names, and it's just the prefix which changes when they move in/out of the WSTP.

Yeah, fwiw I like the sound of this.

vasilmkd:
WSTP index did not need to be the same as the name index, but if it helps debugging, go for it.

djspiewak (approved):
Definitely a fan of anything that improves introspectibility.

PR #3596, opened 2023-05-04
Fix Mutex memory leak

BalmungSan:
Previously, when we canceled the last waiter of the `Mutex`, then its cell would remain in memory until a new waiter would arrive; which was pointless since that waiter would immediately get the next cell on the queue.
Thus, we now first check if we are the last waiter, and if that is the case then we set the state as the next cell; allowing GC to free our cell. Note, we were already doing something very similar con `release`.

I also couldn't resist the need to rename a couple of things, hoping to make the implementation more readable; I did this in a separate commit.


armanbilge (approved):
Nice, I like that `awakeCell` is shared now 👍

PR #3599, opened 2023-05-05
Revised `Queue.synchronous` internals to simplify concurrent hand-off

### LINKED ISSUE #3587: `synchronous queue should not lose takers when offerer is canceled and there are no other takers` flake
Observed in 53a9f5634d9824689f37cc4078e23125fb175308.
https://github.com/typelevel/cats-effect/actions/runs/4856733694/jobs/8656584247?pr=3586#step:23:4083

Is it legit or just a flake 😅 

https://github.com/typelevel/cats-effect/blob/53a9f5634d9824689f37cc4078e23125fb175308/tests/shared/src/test/scala/cats/effect/std/QueueSpec.scala#L75-L99

djspiewak:
Fixes #3587 

This revises the state machine to ensure that `take` gets the value at the same time that it gets the `offer` latch, rather than splitting that into two linearization points. That split was fundamentally unsound (and also over-complicated). By moving things into a single linearization point (via the CAS on `Ref`), we should have fully eliminated the last of the race conditions here.

Previous race conditions in this area will manifest as a `take` which gets a result from an `offer` *as* that `offer` is canceled. This is fine though, because `take` has its value, `offer` is unblocked (because it's canceled), and the `Queue` itself will be empty.

### DIFF
--- a/tests/shared/src/test/scala/cats/effect/std/QueueSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/std/QueueSpec.scala
@@ -94,7 +94,7 @@ class BoundedQueueSpec extends BaseSpec with QueueTests[Queue] with DetectPlatfo
         _ <- IO.race(taker.joinWithNever, q.offer(()).delayBy(500.millis))
       } yield ()
 
-      test.parReplicateA(if (isJS || isNative) 1 else 500).as(ok)
+      test.parReplicateA(if (isJS || isNative) 1 else 1000).as(ok)
     }
   }
 
--- a/std/shared/src/main/scala/cats/effect/std/Queue.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Queue.scala
@@ -187,26 +187,21 @@ object Queue {
       extends Queue[F, A] {
 
     def offer(a: A): F[Unit] =
-      F.deferred[Deferred[F, A]] flatMap { latch =>
+      F.deferred[Unit] flatMap { latch =>
         F uncancelable { poll =>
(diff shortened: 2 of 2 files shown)


djspiewak:
CI failure is fixed by #3602

tests/shared/src/test/scala/cats/effect/std/QueueSpec.scala
```diff
@@ -94,7 +94,7 @@ class BoundedQueueSpec extends BaseSpec with QueueTests[Queue] with DetectPlatfo
         _ <- IO.race(taker.joinWithNever, q.offer(()).delayBy(500.millis))
       } yield ()
 
-      test.parReplicateA(if (isJS || isNative) 1 else 500).as(ok)
+      test.parReplicateA(if (isJS || isNative) 1 else 1000).as(ok)
```
durban, line 97: It's already there, but is `parReplicateA` actually useful here? Couldn't it just be `replicateA`?

durban (approved):
It just occurred to me: doesn't this (and also the current version) has the "lost item" problem? When a `take` is cancelled while suspended, and an `offer` completes the deferred...

But none of my comments should block this PR, as it definitely fixes #3587.

djspiewak:
Oh that's a good point. We should address that in a follow up.

tests/shared/src/test/scala/cats/effect/std/QueueSpec.scala
djspiewak: Doing it in parallel just makes the test complete faster.

PR #3601, opened 2023-05-05
Added test to ensure inner worker awaits are canceled

### LINKED ISSUE #3534: `Dispatcher` internal await is uncancelable in 3.5.x
Relevant wayfinder:

```scala
              if (regs.isEmpty) {
                // TODO this needs to be modified to be cancelable
                F.async_[Unit] { cb =>
                  if (!latch.compareAndSet(Noop, () => cb(Completed))) {
                    // state was changed between when we last set the latch and now; complete the callback immediately
                    cb(Completed)
                  }
                }
```

This code predates the `async` cancelability change and I'm relatively certain that it is intended to be cancelable. We don't have any tests which exercise that codepath though, so this qualifies as an invisibug.

djspiewak:
Closes #3534

Turns out this isn't actually a problem because we sequence `release` immediately thereafter, which *completes* the callback rather than canceling the `async`.

### DIFF
--- a/tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
@@ -152,6 +152,13 @@ class DispatcherSpec extends BaseSpec with DetectPlatform {
           _ <- gate2.acquireN(2)
         } yield ok
       }
+
+      "cancel inner awaits when canceled" in ticked { implicit ticker =>
+        val work = Dispatcher.parallel[IO](await = false).useForever
+        val test = work.background.use(_ => IO.sleep(100.millis))
+
+        test must completeAs(())
+      }
     }
   }
 
--- a/tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
@@ -366,7 +366,7 @@ class DispatcherSpec extends BaseSpec with DetectPlatform {
     }
 
     "cancel inner awaits when canceled" in ticked { implicit ticker =>
-      val work = Dispatcher.parallel[IO](await = false).useForever
+      val work = dispatcher.useForever
       val test = work.background.use(_ => IO.sleep(100.millis))
 
       test must completeAs(())
--- a/tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
@@ -152,13 +152,6 @@ class DispatcherSpec extends BaseSpec with DetectPlatform {
           _ <- gate2.acquireN(2)
         } yield ok
       }
-
-      "cancel inner awaits when canceled" in ticked { implicit ticker =>
-        val work = Dispatcher.parallel[IO](await = false).useForever
-        val test = work.background.use(_ => IO.sleep(100.millis))
-
-        test must completeAs(())
-      }
     }
   }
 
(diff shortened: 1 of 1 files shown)


tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
```diff
@@ -152,6 +152,13 @@ class DispatcherSpec extends BaseSpec with DetectPlatform {
           _ <- gate2.acquireN(2)
         } yield ok
       }
+
+      "cancel inner awaits when canceled" in ticked { implicit ticker =>
+        val work = Dispatcher.parallel[IO](await = false).useForever
```
samspills: Just curious, why make a new dispatcher to use forever instead of using the passed in `dispatcher` resource?

(Disclosure: I'm on my phone on a streetcar and it's hard to read the whole file 😅 I might be missing something obvious)

djspiewak: What I'm trying to test here is the situation where at least one of the workers is quiesced and in `await`, and then we cancel the overall dispatcher fiber. The easiest way to do that is just `useForever` on the resource and chill out waiting to be canceled, since that will guarantee that all the worker fibers are awaiting (so long as we waited long enough on the external cancelation, which is what the `sleep` + `ticker` guarantees for us).

armanbilge: Is this test not relevant for `await = true` or `sequential` dispatcher? IIUC they should all behave the same here, so we can add it as a generic test like Sam suggests.

samspills: Ah, in my original comment I thought this test was in the `parallel` function (line 165), so that's why I thought it would be a reference to `dispatcher`. Instead though this test would have access to the `D` defined on line 118. So, just in the interest of keeping the pattern of using the pre-created dispatcher, this line could be `val work = D.useForever`

Unless this test were to move to the `common` section, as Arman suggests. My understanding is the same, we'd expect the same behaviour for all cases so that could be good.

djspiewak: We wouldn't have this behavior for the `await = true` case, though we *would* for `sequential`. So we actually need a slightly different variant of `common` which *only* covers `await = false`, which doesn't exist at present and I was lazy. I can add that though.

armanbilge: > We wouldn't have this behavior for the `await = true` case

Sorry, you're going to have to explain more 😅  Are you saying when `await = true` this doesn't complete normally? why?

```scala
val work = Dispatcher.parallel[IO](await = true).useForever
work.background.use(_ => IO.sleep(100.millis))
test must completeAs(())
```

djspiewak: Oh actually that's a good point, it's just kind of an irrelevant test. Alrighty, into `common` it goes.

tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
```diff
@@ -364,6 +364,13 @@ class DispatcherSpec extends BaseSpec with DetectPlatform {
             IO(runner.unsafeRunAndForget(IO(ko)) must throwAn[IllegalStateException])
       }
     }
+
+    "cancel inner awaits when canceled" in ticked { implicit ticker =>
+      val work = Dispatcher.parallel[IO](await = false).useForever
```
armanbilge:
```suggestion
      val work = dispatcher.useForever
```

djspiewak: 🤦🏼‍♂️

armanbilge (approved):

PR #3602, opened 2023-05-05
Make the new dispatcher release test deterministic

djspiewak:


tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
```diff
@@ -356,13 +356,16 @@ class DispatcherSpec extends BaseSpec with DetectPlatform {
...
+    "issue 3501: reject new tasks after release action is submitted as a task" in ticked {
+      implicit ticker =>
+        val test = dispatcher.allocated.flatMap {
+          case (runner, release) =>
+            IO(runner.unsafeRunAndForget(IO.sleep(50.millis) *> release)) *>
+              IO.sleep(100.millis) *>
+              IO(runner.unsafeRunAndForget(IO(ko)) must throwAn[IllegalStateException])
+        }
+
+        test.void must completeAs(())
```
samspills, line 368: Ohh `test.void` til!

samspills (approved):

tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
samspills: I tried to figure this out with a `test must FailAs(...)` but I couldn't quite get it.

djspiewak: Yeah I was going to do a `failAs` but it's really restricted in how it matches exceptions, so I just bailed and left the Specs matcher in place.

durban (approved):

PR #3604, opened 2023-05-06
Reset the global runtime when it is shutdown.

### LINKED ISSUE #1818: Global runtime installation mechanism not working in Scastie
Reported by @oleg-py [here](https://gitter.im/typelevel/cats-effect-dev?at=605af4e7f07ffa1eb5608da3).

image

https://scastie.scala-lang.org/gLdvaPQsRpu9Gokfz9D7FA

armanbilge:
Fixes https://github.com/typelevel/cats-effect/issues/1818. (`publishLocal`ed and confirmed manually)

If the global runtime is shutdown, then it should be reset so that another global can be installed. This prevents the awkward situation where tasks are submitted to a global that's actually not alive, which was causing the hang in https://github.com/typelevel/cats-effect/issues/1818#issuecomment-1536987188.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
@@ -176,7 +176,7 @@ private[unsafe] abstract class IORuntimeCompanionPlatform { this: IORuntime.type
     (Scheduler.fromScheduledExecutor(scheduler), { () => scheduler.shutdown() })
   }
 
-  private[this] var _global: IORuntime = null
+  @volatile private[this] var _global: IORuntime = null
 
   // we don't need to synchronize this with IOApp, because we control the main thread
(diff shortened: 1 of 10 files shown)


core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
```diff
@@ -176,7 +176,7 @@ private[unsafe] abstract class IORuntimeCompanionPlatform { this: IORuntime.type
     (Scheduler.fromScheduledExecutor(scheduler), { () => scheduler.shutdown() })
   }
 
-  private[this] var _global: IORuntime = null
+  @volatile private[this] var _global: IORuntime = null
```
armanbilge, line 179: If we _really_ care we'll need some more serious synchronization than this to prevent double-installations.

djspiewak (approved):
Very nice! Also I think the `@volatile` is probably sufficient.

PR #3605, opened 2023-05-07
Fix #3594

### LINKED ISSUE #3594: flaky `DefaultContSpec` in CI [JVM]
On this commit: https://github.com/typelevel/cats-effect/commit/3cc51888d594cf797a9856b7b879d68249d0d642

Failure here: https://github.com/typelevel/cats-effect/actions/runs/4877781790/jobs/8702776117?pr=3593#step:23:4739
for (macos-latest, 2.13.10, graalvm@11, ciJVM)

```
[error] x get within onCancel - 1
[error]  false -> false != true -> true (ContSpec.scala:130)
[error] x get within onCancel - 2
[error]  false -> false != true -> true (ContSpec.scala:151)
```

durban:
~~This won't fix #3594, but hopefully will make it clearer what happens when these tests fail in CI. (I think the 2 second timer sometimes fires _before_ the 1 second timer, but I'm not sure.)~~

It seems to be impossible to reproduce #3594 in CI any more. My best guess for what happened is that the 2 second timer probably fired _before_ the 1 second timer. (Since thread-local timers, this is unlikely, but possible. I've actually seen it once locally in an intentionally constrained environment.) So the "fix" is to make the 2 second sleep longer :-)


durban:
Hilariously, I can't get CI to _fail_ on this PR :-)

djspiewak:
IT'S FIXED! jk

durban:
Should we merge this? Or why bother, seems to be resolved anyway?

samspills:
I don't think anyone else has seen this again since I reported, right? If so, I think it's probably fine to close this for now and resurrect if the failure happens again?

armanbilge:
I have seen some other issues with that spec, specifically on Apple Silicon.
- https://github.com/typelevel/cats-effect/issues/3658

Issue #3474, opened 2023-03-01
`cancelable` may leak resources on unsuccessful cancelation


armanbilge:
Spinning this out of a [Discord observation](https://discord.com/channels/632277896739946517/839263556754472990/1079945167651471461) so that it doesn't get lost. It relates to the new `cancelable` combinator added in https://github.com/typelevel/cats-effect/pull/3460.

This issue is important, because it has led me to believe we cannot currently implement the non-leaking `race` proposed in https://github.com/typelevel/cats-effect/issues/3456.

For motivation, consider this case:
```scala
Resource.makeFull { poll =>
  poll(IO.blocking(server.accept()).cancelable(server.shutdown()))
}(socket => IO(socket.close()))
```

There, we have a blocking call to `accept` a socket on a server. We can cancel it by shutting down the server. However, if the blocking call succeeds (perhaps in a race condition with `shutdown()`), and we do get an open socket, we really want to be sure that we close it.

To demonstrate the leak I have added a test in 84e7ef10c48a8b697dbcaf27709e74d5705caf4a.
https://github.com/typelevel/cats-effect/blob/84e7ef10c48a8b697dbcaf27709e74d5705caf4a/tests/shared/src/test/scala/cats/effect/IOSpec.scala#L1094-L1100

In this case, "cancelation" is simply causing the `Deferred` to complete normally. So that means the `get` is completing with `()`. If that's a resource, we don't want to lose that, so the outcome should be successful, not canceled.

It's a bit weird, but conceptually this is equivalent to a race condition where you _request_ cancellation, but the effect in question succeeds anyway. In that case, it didn't really cancel, and if you pretend like it did, you can leak a resource.

Now, what worries me is that I don't see how we can fix this. And if we can't fix this, I don't see how we can fix https://github.com/typelevel/cats-effect/issues/3456.

https://github.com/typelevel/cats-effect/blob/7a5311b18bfb6c553a15227492be5c6f80d9d941/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala#L264-L269

So, the problem is that if cancelation occurs while this fiber is in `poll(fiber.join)`, then _it is assumed that it canceled!_ That means that `cancelable` can no longer have a successful outcome.

[...]

PR #3606, opened 2023-05-07
Adjusted `Queue.synchronous` to include a two-phase commit

### LINKED ISSUE #3603: `Queue.synchronous` may suffer from the "lost item" problem
Specifically, when a `take` is canceled at the moment an `offer` completes, the value may be lost.

djspiewak:
Fixes #3603 

Okay I think I finally got it. `synchronous` preserves the "at least once" guarantee while also preserving the "only one value at a time", which means it boils down to "exactly once". The only thing that makes this modestly not-insane is the fact that we can lose values on the `offer` side, just not on the `take` side, meaning that we don't need commits in both directions. Thus, the fix here is to introduce a signaling mechanism from `take` to `offer` which effectively releases `offer` and confirms that we have the `A` in `take` and we are no longer cancelable. There are some tradeoffs to this, but given that we're going for exactly-once delivery, tradeoffs should absolutely be expected.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Queue.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Queue.scala
@@ -245,9 +245,9 @@ object Queue {
                 val removeListener = stateR modify {
                   case SyncState(offerers, takers) =>
                     // like filter, but also returns a Boolean indicating whether it was found
-                    def filterFound[A <: AnyRef](
-                        in: ScalaQueue[A],
-                        out: ScalaQueue[A]): (Boolean, ScalaQueue[A]) = {
+                    def filterFound[Z <: AnyRef](
+                        in: ScalaQueue[Z],
+                        out: ScalaQueue[Z]): (Boolean, ScalaQueue[Z]) = {
 
                       if (in.isEmpty) {
                         (false, out)
@@ -255,7 +255,7 @@ object Queue {
                         val (head, tail) = in.dequeue
 
                         if (head eq latch)
-                          (true, out.enqueueAll(tail))
+                          (true, out ++ tail)
                         else
                           filterFound(tail, out.enqueue(head))
                       }
--- a/std/shared/src/main/scala/cats/effect/std/Queue.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Queue.scala
@@ -242,18 +242,37 @@ object Queue {
 
             case SyncState(offerers, takers) =>
               val cleanupF = {
-                val removeListener = stateR update {
+                val removeListener = stateR modify {
                   case SyncState(offerers, takers) =>
-                    SyncState(offerers, takers.filter(_ ne latch))
+                    // like filter, but also returns a Boolean indicating whether it was found
+                    def filterFound[A <: AnyRef](
+                        in: ScalaQueue[A],
+                        out: ScalaQueue[A]): (Boolean, ScalaQueue[A]) = {
+
+                      if (in.isEmpty) {
+                        (false, out)
+                      } else {
+                        val (head, tail) = in.dequeue
+
+                        if (head eq latch)
+                          (true, out.enqueueAll(tail))
+                        else
+                          filterFound(tail, out.enqueue(head))
+                      }
+                    }
+
+                    val (found, takers2) = filterFound(takers, ScalaQueue())
+                    SyncState(offerers, takers2) -> found
                 }
 
-                val failCommit = latch.tryGet flatMap {
+                val failCommit = latch.get flatMap {
                   // this is where we *fail* the 2-phase commit up in offer
-                  case Some((_, commitLatch)) => commitLatch.complete(false).void
-                  case None => F.unit
+                  case (_, commitLatch) => commitLatch.complete(false).void
                 }
 
-                removeListener *> failCommit
+                // if we *don't* find our latch, it means an offerer has it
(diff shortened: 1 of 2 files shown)


std/shared/src/main/scala/cats/effect/std/Queue.scala
```diff
@@ -187,21 +187,27 @@ object Queue {
       extends Queue[F, A] {
 
     def offer(a: A): F[Unit] =
-      F.deferred[Unit] flatMap { latch =>
+      F.deferred[Boolean] flatMap { latch =>
         F uncancelable { poll =>
+          // this is the 2-phase commit
+          // we don't need an onCancel for latch.get since if *we* are canceled, it's okay to drop the value
+          val checkCommit = poll(latch.get).ifM(F.unit, poll(offer(a)))
```
durban, line 194: By retrying `offer(a)`, we lose FIFO ordering, right? (Which is probably fine... just to be clear. And no, I don't know how _not_ to lose ordering...)

std/shared/src/main/scala/cats/effect/std/Queue.scala
```diff
@@ -212,49 +218,85 @@ object Queue {
...
-              val cleanupF = stateR update {
-                case SyncState(offerers, takers) =>
-                  SyncState(offerers, takers.filter(_ ne latch))
+              val cleanupF = {
+                val removeListener = stateR update {
+                  case SyncState(offerers, takers) =>
+                    SyncState(offerers, takers.filter(_ ne latch))
+                }
+
+                val failCommit = latch.tryGet flatMap {
```
durban: Doesn't this `tryGet` has the same problem the "old" implementation had? A race between this `tryGet` and the `taker.complete((a, latch))` in `(try)Offer`: if this `tryGet` is faster, it will not see the `commitLatch`, and will not actually fail the commit.

durban (commented):
I'm not completely convinced that the other direction (offerer cancellation) can be ignored:

- This direction (taker cancellation): the problem is that `offer` thinks it successfully passed an item, but the `take` was cancelled; so the item is _lost_.
- The other direction (offerer cancellation):  the opposite problem is that `take` successfully receives the item, without `offer` actually passing that item (because `offer` was successfully cancelled); so it's a _"phantom"_ item.

(Of course I realize making both directions work would make it very convoluted. And the taker cancellation definitely seems like the more important problem.)

std/shared/src/main/scala/cats/effect/std/Queue.scala
djspiewak: FIFO is definitely lost. I'm pretty sure that we can't preserve FIFO *and* guaranteed delivery. There's probably some connection we can draw to the CAP theorem to prove it rigorously.

std/shared/src/main/scala/cats/effect/std/Queue.scala
djspiewak: Good catch! I missed this. I think I fixed it now though.

std/shared/src/main/scala/cats/effect/std/Queue.scala
```diff
@@ -212,49 +218,104 @@ object Queue {
...
 
             case SyncState(offerers, takers) =>
-              val cleanupF = stateR update {
-                case SyncState(offerers, takers) =>
-                  SyncState(offerers, takers.filter(_ ne latch))
+              val cleanupF = {
+                val removeListener = stateR modify {
+                  case SyncState(offerers, takers) =>
+                    // like filter, but also returns a Boolean indicating whether it was found
+                    def filterFound[Z <: AnyRef](
```
durban, line 249: `@tailrec`?

durban (dismissed):

std/shared/src/main/scala/cats/effect/std/Queue.scala
djspiewak: Probably a good idea :-)

durban (approved):

djspiewak:
> The other direction (offerer cancellation): the opposite problem is that take successfully receives the item, without offer actually passing that item (because offer was successfully cancelled); so it's a "phantom" item.

@durban Sorry for the delayed response on this. I think this is an interesting point, but IMO we can ignore it because the data itself is not lost. Put another way, it is externally indistinguishable from the situation where the offerer is canceled shortly *after* completing the `offer` (with control flow having entirely exited `Queue`), and we're obviously not preventing that.

armanbilge:
> Put another way, it is externally indistinguishable from the situation where the offerer is canceled shortly _after_ completing the `offer` (with control flow having entirely exited `Queue`)

Is that really true? Would not `offer(...).onCancel(...)` distinguish between those situations?

durban:
@djspiewak Thinking about this some more, I'm pretty sure it is "distinguishable" in the sense that it is _possible_ to write a test, which detects if it happens (e.g., as @armanbilge wrote, with `onCancel`). However, the question is: does this matter in practice? I admit, the answer is possibly "not really"...

Rephrasing the problem: data (as in a queue element) is not lost. Rather, a "success signal" of the `offer` is lost. I'm wondering what would happen if the elements are things which require cleanup (#3474 :-)... `offer(x).onCancel(cleanup_x)` would run the cleanup, while `take` received the `x`. (Although I've no idea, why would anybody pass resources through a synchronous queue...)

PR #3610, opened 2023-05-11
Catch stray exceptions in `uncancelable` body

armanbilge:
Fixes https://github.com/typelevel/cats-effect/issues/3609.

PR #3615, opened 2023-05-11
Fixed issue in which failing `uncancelable` would remain masked for one stage

djspiewak:


armanbilge (approved):
Linking to https://github.com/typelevel/cats-effect/pull/3610#discussion_r1191714245.

PR #3617, opened 2023-05-12
Drafted version update script

djspiewak:
Eventually we can grow this into more automation around the release chores.


scripts/update-versions.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
```
armanbilge, line 1: Can we add a comment with example usage?

armanbilge (approved):
🎉

durban (approved):

PR #3622, opened 2023-05-12
Fix issue #3589

### LINKED ISSUE #3589: Update migration guide for newer CE3 stuff
For example, https://typelevel.org/cats-effect/docs/migration-guide#where-does-the-blocking-pool-come-from
mentions stuff that isn't correct in the current CE3, because of the WSTP

durban:
~~I have no idea how .md docs work, so this is untested code.~~ Also, english is not my first language, so review carefully :-)

I've run `docs/mdoc`, and there were no errors.


djspiewak:
Aw nuts. I tried to just change the base and it looks like it pulled in an extra commit. Would you mind rebasing to sort that out? We have a `series/3.5.x` now. Otherwise, thank you though! I'll take a closer copyedit review in a bit

durban:
Ok, I've rebased on series/3.5.x.

djspiewak (approved):
Content looks great!

PR #3629, opened 2023-05-13
Implemented script for auto-updating the site following a release

djspiewak:


scripts/data/scastie.sbt
```diff
@@ -0,0 +1,8 @@
+scalacOptions ++= Seq(
+  "-deprecation",
+  "-encoding", "UTF-8",
+  "-feature",
+  "-unchecked"
+)
+
+turbo := false
```
armanbilge, line 8: Actually we no longer need this after:
- https://github.com/typelevel/cats-effect/pull/3604

armanbilge (approved):

scripts/data/scastie.sbt
djspiewak: Oh nice. We should PR that removal separately. It's all controlled here now.

PR #3637, opened 2023-05-17
add evalOn(executor)

### LINKED ISSUE #3633: Add `Async#evalOn(executor)` or `Async#evalOnExecutor(executor)`
`Async#evalOn` takes a Scala `ExecutionContext` but frequently this functionality is needed for running on a Java `Executor`.

- `ExecutionContext.fromExecutor(...)` can be used to get an `ExecutionContext` from an `Executor`

- We can do even better, by using the current `ExecutionContext`s `reportFailure` method for the new `ExecutionContext` built from the `Executor`. This is better than the default reporter.

homycdev:
https://github.com/typelevel/cats-effect/issues/3633

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -40,6 +40,7 @@ import cats.effect.kernel.CancelScope
 import cats.effect.kernel.GenTemporal.handleDuration
 import cats.effect.std.{Backpressure, Console, Env, Supervisor, UUIDGen}
 import cats.effect.tracing.{Tracing, TracingEvent}
+import cats.effect.unsafe.IORuntime
 import cats.syntax.all._
 
 import scala.annotation.unchecked.uncheckedVariance
@@ -1485,6 +1486,11 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
   def trace: IO[Trace] =
     IOTrace
 
+  private[effect] def runtime: IO[IORuntime] = ReadRT
+
+  def pollers: IO[List[Any]] =
+    IO.runtime.map(_.pollers)
+
   def uncancelable[A](body: Poll[IO] => IO[A]): IO[A] =
     Uncancelable(body, Tracing.calculateTracingEvent(body))
 
@@ -2085,6 +2091,10 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
   private[effect] case object IOTrace extends IO[Trace] {
     def tag = 23
+  }
+
+  private[effect] case object ReadRT extends IO[IORuntime] {
+    def tag = 24
   }
 
   // INTERNAL, only created by the runloop itself as the terminal state of several operations
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,12 +42,16 @@ jobs:
             java: graalvm@11
           - os: windows-latest
             scala: 3.3.0
+            ci: ciJVM
           - os: macos-latest
             scala: 3.3.0
+            ci: ciJVM
           - os: windows-latest
             scala: 2.12.18
+            ci: ciJVM
           - os: macos-latest
             scala: 2.12.18
+            ci: ciJVM
           - ci: ciFirefox
             scala: 3.3.0
           - ci: ciChrome
@@ -97,9 +101,6 @@ jobs:
           - os: macos-latest
             ci: ciNative
             scala: 2.12.18
-          - os: macos-latest
-            ci: ciNative
-            scala: 3.3.0
           - os: windows-latest
             java: graalvm@11
     runs-on: ${{ matrix.os }}
--- a/benchmarks/src/main/scala/cats/effect/benchmarks/WorkStealingBenchmark.scala
+++ b/benchmarks/src/main/scala/cats/effect/benchmarks/WorkStealingBenchmark.scala
@@ -165,12 +165,13 @@ class WorkStealingBenchmark {
       (ExecutionContext.fromExecutor(executor), () => executor.shutdown())
     }
 
-    val compute = new WorkStealingThreadPool(
+    val compute = new WorkStealingThreadPool[AnyRef](
       256,
       "io-compute",
       "io-blocker",
       60.seconds,
       false,
+      SleepSystem,
       _.printStackTrace())
 
     val cancelationCheckThreshold =
--- a/build.sbt
+++ b/build.sbt
@@ -41,7 +41,7 @@ ThisBuild / git.gitUncommittedChanges := {
   }
 }
 
-ThisBuild / tlBaseVersion := "3.5"
+ThisBuild / tlBaseVersion := "3.6"
 ThisBuild / tlUntaggedAreSnapshots := false
 
 ThisBuild / organization := "org.typelevel"
@@ -224,8 +224,8 @@ ThisBuild / githubWorkflowBuildMatrixExclusions := {
   val windowsAndMacScalaFilters =
     (ThisBuild / githubWorkflowScalaVersions).value.filterNot(Set(Scala213)).flatMap { scala =>
       Seq(
-        MatrixExclude(Map("os" -> Windows, "scala" -> scala)),
-        MatrixExclude(Map("os" -> MacOS, "scala" -> scala)))
+        MatrixExclude(Map("os" -> Windows, "scala" -> scala, "ci" -> CI.JVM.command)),
+        MatrixExclude(Map("os" -> MacOS, "scala" -> scala, "ci" -> CI.JVM.command)))
     }
 
   val jsScalaFilters = for {
@@ -254,9 +254,7 @@ ThisBuild / githubWorkflowBuildMatrixExclusions := {
 
     javaFilters ++ Seq(
       MatrixExclude(Map("os" -> Windows, "ci" -> ci)),
-      MatrixExclude(Map("os" -> MacOS, "ci" -> ci, "scala" -> Scala212)),
-      // keep a native+2.13+macos job
-      MatrixExclude(Map("os" -> MacOS, "ci" -> ci, "scala" -> Scala3))
+      MatrixExclude(Map("os" -> MacOS, "ci" -> ci, "scala" -> Scala212))
     )
   }
 
@@ -640,7 +638,10 @@ lazy val core = crossProject(JSPlatform, JVMPlatform, NativePlatform)
         "cats.effect.IOFiberConstants.ExecuteRunnableR"),
       ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.effect.IOLocal.scope"),
       ProblemFilters.exclude[DirectMissingMethodProblem](
-        "cats.effect.IOFiberConstants.ContStateResult")
+        "cats.effect.IOFiberConstants.ContStateResult"),
+      // introduced by #3332, polling system
+      ProblemFilters.exclude[DirectMissingMethodProblem](
+        "cats.effect.unsafe.IORuntimeBuilder.this")
     ) ++ {
       if (tlIsScala3.value) {
         // Scala 3 specific exclusions
@@ -823,6 +824,14 @@ lazy val core = crossProject(JSPlatform, JVMPlatform, NativePlatform)
         )
       } else Seq()
     }
+  )
+  .nativeSettings(
+    mimaBinaryIssueFilters ++= Seq(
+      ProblemFilters.exclude[MissingClassProblem](
+        "cats.effect.unsafe.PollingExecutorScheduler$SleepTask"),
+      ProblemFilters.exclude[MissingClassProblem]("cats.effect.unsafe.QueueExecutorScheduler"),
+      ProblemFilters.exclude[MissingClassProblem]("cats.effect.unsafe.QueueExecutorScheduler$")
+    )
   )
   .disablePlugins(JCStressPlugin)
 
--- a/core/js-native/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
+++ b/core/js-native/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
@@ -23,7 +23,7 @@ import scala.concurrent.duration.FiniteDuration
 // Can you imagine a thread pool on JS? Have fun trying to extend or instantiate
 // this class. Unfortunately, due to the explicit branching, this type leaks
 // into the shared source code of IOFiber.scala.
-private[effect] sealed abstract class WorkStealingThreadPool private ()
+private[effect] sealed abstract class WorkStealingThreadPool[P] private ()
     extends ExecutionContext {
   def execute(runnable: Runnable): Unit
   def reportFailure(cause: Throwable): Unit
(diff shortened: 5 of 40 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
```diff
@@ -401,6 +407,9 @@ object Async {
     def evalOn[A](fa: OptionT[F, A], ec: ExecutionContext): OptionT[F, A] =
       OptionT(F.evalOn(fa.value, ec))
 
+    def evalOn[A](fa: OptionT[F, A], executor: Executor): OptionT[F, A] =
+      OptionT(F.evalOn(fa.value, executor))
```
armanbilge: Instead of implementing it for every `Async` ...

kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
```diff
@@ -166,6 +166,12 @@ trait Async[F[_]] extends AsyncPlatform[F] with Sync[F] with Temporal[F] {
    */
   def evalOn[A](fa: F[A], ec: ExecutionContext): F[A]
 
+  /**
+   * [[Async.evalOn]] with provided java.util.concurrent.Executor
+   */
+  def evalOn[A](fa: F[A], executor: Executor): F[A]
```
armanbilge: ... we should just add a default implementation here, that everyone can use :)

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1821,6 +1821,11 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
     def evalOn[A](fa: IO[A], ec: ExecutionContext): IO[A] =
       fa.evalOn(ec)
 
+    def evalOn[A](fa: IO[A], executor: Executor): IO[A] =
+      IO.executionContext
+        .flatMap(ec => IO(ExecutionContext.fromExecutor(executor, ec.reportFailure)))
+        .flatMap(ec => fa.evalOn(ec))
```
armanbilge: Nice! This looks exactly like what I imagined. We can make this the default implementation in `Async`.

I have one more suggestion: we can pattern match and check if the `Executor` is already an `ExecutionContext`, and skip the conversion in that case.

We do something similar here.
https://github.com/typelevel/cats-effect/blob/7e24bb96d4921f4bb50bf439c3429bcff998f457/kernel/shared/src/main/scala/cats/effect/kernel/Async.scala#L200-L206

armanbilge (commented):
Thanks!

There is one thing I didn't think of, that we should add a test for: Scala has an `ExecutionContextExecutor` type that is both an `ExecutionContext` and an `Executor`. What happens if someone passes that to `evalOn`, how does it disambiguate?

https://www.scala-lang.org/api/2.13.10/scala/concurrent/ExecutionContextExecutor.html

kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
durban: Wait, how does adding a new abstract method _not_ fail the mima check?

armanbilge: It does :)

durban: Okay... then how is CI green? :-)

armanbilge: It's not.
Screen Shot 2023-05-17 at 6 21 48

core/shared/src/main/scala/cats/effect/IO.scala
homycdev: Hope, I got you right. Please, check new commit when you can.

kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
```diff
@@ -166,6 +166,20 @@ trait Async[F[_]] extends AsyncPlatform[F] with Sync[F] with Temporal[F] {
    */
   def evalOn[A](fa: F[A], ec: ExecutionContext): F[A]
 
+  /**
+   * [[Async.evalOn]] with provided java.util.concurrent.Executor
+   */
+  def evalOn[A](fa: F[A], executor: Executor): F[A] =
+    executor match {
+      case ec: ExecutionContextExecutor =>
```
armanbilge: It's possible it implements `ExecutionContext` without implementing `ExecutionContextExecutor`.
```suggestion
      case ec: ExecutionContext =>
```

kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
```diff
@@ -166,6 +166,20 @@ trait Async[F[_]] extends AsyncPlatform[F] with Sync[F] with Temporal[F] {
    */
   def evalOn[A](fa: F[A], ec: ExecutionContext): F[A]
 
+  /**
+   * [[Async.evalOn]] with provided java.util.concurrent.Executor
```
armanbilge:
```suggestion
   * [[Async.evalOn]] with provided [[java.util.concurrent.Executor]]
```

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1820,7 +1820,7 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
     def evalOn[A](fa: IO[A], ec: ExecutionContext): IO[A] =
       fa.evalOn(ec)
-
+
```
armanbilge: Spurious change

kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
```diff
@@ -610,7 +624,6 @@ object Async {
 
     def evalOn[A](fa: WriterT[F, L, A], ec: ExecutionContext): WriterT[F, L, A] =
       WriterT(F.evalOn(fa.run, ec))
-
```
armanbilge: Spurious change

homycdev: sorry for these changes. I will fix asap

homycdev:
@armanbilge  is there everything alright? or something's missing?

kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
```diff
@@ -166,6 +166,21 @@ trait Async[F[_]] extends AsyncPlatform[F] with Sync[F] with Temporal[F] {
...
+   * [[Async.evalOn]] with provided [[java.util.concurrent.Executor]]
+   */
+  def evalOn[A](fa: F[A], executor: Executor): F[A] =
+    executor match {
+      case ec: ExecutionContext =>
+        evalOn[A](fa, ec: ExecutionContext)
+      case executor =>
+        flatMap(executionContext) { refEc =>
+          val newEc: ExecutionContext =
+            ExecutionContext.fromExecutor(executor, refEc.reportFailure)
```
durban: This is a strange case: if `executor` is non-`null`, `fromExecutor` is pure; however, if it's `null`, then it creates a new threadpool. I would suggest rejecting `null` here.

durban (changes requested):
I think what's mainly missing is the test suggested in [this comment](https://github.com/typelevel/cats-effect/pull/3637#pullrequestreview-1430640959).

(@armanbilge I've figured out what's happening with the "green" CI: in the PR list, and besides the commit there is a green checkmark. However, if I log in, I see the yellow warning at the bottom. This is probably normal, just a bit strange...)

kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
homycdev: how do you see this rejection in the context of the library?

homycdev: I can throwNPE as it is thrown in the default evalOn for example... but not sure whether this approach would be considered as good one, since this low-level rejection is going to be within high-level trait method

core/shared/src/main/scala/cats/effect/IO.scala
homycdev: regarding tests, i have found tests only on laws for Async's methods. 
But from the comment you wrote me, i shall rather check whether the executionContext was executed through provided executor... correct me if I am wrong. 

Thank you in advance!

kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
armanbilge: > If null, a new Executor is created with default configuration.

What the hell ?? 😅 Good catch. Yeah, I think raising a NPE is the right thing to do. We could also no-op, but that seems arbitrary 🤔

homycdev: > There is one thing I didn't think of, that we should add a test for: Scala has an ExecutionContextExecutor type that is both an ExecutionContext and an Executor. What happens if someone passes that to evalOn, how does it disambiguate?

@armanbilge 

regarding tests, i have found tests only on laws for Async's methods.
But from the comment you wrote me, i shall rather check whether the executionContext was executed through provided executor... correct me if I am wrong.

Thank you in advance!

armanbilge: > i shall rather check whether the executionContext was executed through provided executor... correct me if I am wrong.

@homycdev actually, that's not what I concerned about. My concern is this. We now have methods:
```scala
def evalOn(fa: F[A], ec: ExecutionContext): F[A]
def evalOn(fa: F[A], exec: Executor): F[A]
```
so what will the compiler do if I write code like this:
```scala
def ecexec: ExecutionContextExecutor = ???
evalOn(fa, ecexec)
```
how will it choose which method to use? My guess is it won't be able to disambiguate and compilation will fail. If that's the case, we'll have to think about how to handle this.

durban: I usually prefer an `IllegalArgumentException` (with `require`) instead of an NPE, but honestly, either one is fine.

homycdev: > how will it choose which method to use? My guess is it won't be able to disambiguate and compilation will fail. If that's the case, we'll have to think about how to handle this.

can this approach be considered? 
```scala
def evalOn[A, Exec <: ExecutionContextExecutor](fa: F[A], exec: Exec) = {
        require(executor != null, "Cannot pass undefined Executor as an argument")
// the rest of the code 
}

```

armanbilge: I am not sure if I understand. If the user has an ordinary `Executor` (i.e. not an `ExecutionContextExecutor`) then that method won't work.

homycdev: I meant
```scala
def evalOn[A, Exec >: ExecutionContextExecutor](fa: F[A], exec: Exec) = {

```

but nvm, i think it is better just to change the name of the method, to avoid ambiguity 

```scala
  def evalOnExecutor[A](fa: F[A], executor: Executor): F[A] = {
    require(executor != null, "Cannot pass undefined Executor as an argument")
    executor match {
      case ec: ExecutionContext =>
        evalOn[A](fa, ec: ExecutionContext)
      case executor =>
        flatMap(executionContext) { refEc =>
          val newEc: ExecutionContext =
            ExecutionContext.fromExecutor(executor, refEc.reportFailure)
          evalOn[A](fa, newEc)
        }
    }
  }
```

armanbilge: > but nvm, i think it is better just to change the name of the method, to avoid ambiguity

Yeah, probably 😕 

the other option I can think of would be to have _three_ overloads since I think it will choose the most specific one, but I'm not sure if that's a good idea 😅 

```scala
 def evalOn[A](fa: F[A], ec: ExecutionContext): F[A]
 def evalOn[A](fa: F[A], exec: Executor): F[A]
 def evalOn[A](fa: F[A], ecexec: ExecutionContextExecutor): F[A]
```

homycdev: so what shall we decide on? Keep two methods with different names, or do as you proposed with three overloads?

armanbilge: I think Gavin has 👍 the triple-overload idea. I wonder what Daniel thinks (both of them 😉). Sorry, bike-shedding!

homycdev: could you tag them pls?) I'd tag them myself, but not sure what their gh aliases 

I am still waiting to implement final solution 😅

durban: I see 3 options:

- Do nothing (i.e., do not add this method). I think this is an acceptable (if not particularly helpful) solution; everybody can continue to call `fromExecutor` themselves.
- Have the 3 methods, which I don't like, but it could still be the least bad solution. (Actually, I think instead of `ExecutionContextExecutor` we could have `ExecutionContext with Executor`, but that might be overdoing it.)
- Have 2 methods, with a different name. It's at least simpler than 3 methods... I also don't particularly like this one...

durban: Okay, thought about it more, and I think calling it `evalOnExecutor` is probably best: it's a little like the `from*` methods, which have the same prefix (so autocomplete works). Three overloads could be somewhat confusing to anyone reading the source.

homycdev: what do we do with the ExecutionContextExecutor? We ignore it? Or also add method something like def evalOnExecutionContextExectuor?

armanbilge: An `ExecutionContextExecutor` is both an `ExecutionContext` and an `Executor`. So a user can pass it to either method and it will work fine :)

homycdev: i know, but asked just in case you'd need specific method for that one too. Sorry for inconvenience

armanbilge: no problem, thanks for asking!

kernel/shared/src/main/scala/cats/effect/kernel/syntax/AsyncSyntax.scala
```diff
@@ -30,6 +32,9 @@ final class AsyncOps[F[_], A] private[syntax] (private[syntax] val wrapped: F[A]
   def evalOn(ec: ExecutionContext)(implicit F: Async[F]): F[A] =
     Async[F].evalOn(wrapped, ec)
 
+  def evalOn(executor: Executor)(implicit F: Async[F]): F[A] =
+    Async[F].evalOnExecutor(wrapped, executor)
```
armanbilge:
```suggestion
  def evalOnExecutor(executor: Executor)(implicit F: Async[F]): F[A] =
    Async[F].evalOnExecutor(wrapped, executor)
```

Would you mind also adding this method directly on `IO`? Thanks!

homycdev: @armanbilge hello! Could you please check?)

djspiewak (dismissed):
I think this is good to go. @armanbilge any thoughts?

Thank you for tackling this!

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -358,6 +353,16 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
...
   def evalOn(ec: ExecutionContext): IO[A] = IO.EvalOn(this, ec)
 
+  /**
+   * Shifts the execution of the current IO to the specified [[java.util.concurrent.Executor]].
+   * @param executor
+   * @return
+   */
+  def evalOnExecutor(executor: Executor): IO[A] =
+    executionContext
+      .flatMap(ec => IO(ExecutionContext.fromExecutor(executor, ec.reportFailure)))
```
durban: With this, `IO#evalOnExecutor` has a different behavior from `Async#evalOnExecutor`. I think they should behave the same.

durban (changes requested):

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -35,6 +35,7 @@ import cats.{
   Traverse
 }
 import cats.data.Ior
+import cats.effect.IO.executionContext
```
armanbilge: Instead of importing this can we call it as `IO.executionContext`? Sometimes unqualified methods can cause confusion 😅

armanbilge (commented):
Thanks, yes, sorry! There are a few more APIs I think it would be good to add, for consistency: `startOnExecutor`, `backgroundOnExecutor`, `evalOnExecutorK`

core/shared/src/main/scala/cats/effect/IO.scala
djspiewak: They should indeed! I missed that in my review.

homycdev: Sorry for this. I hope, I understood you correctly now!

PR #3644, opened 2023-05-25
Point out cancelation semantics in `fromCompletableFuture` scaladoc

danicheg:


kernel/jvm/src/main/scala/cats/effect/kernel/AsyncPlatform.scala
```diff
@@ -27,6 +27,12 @@ private[kernel] trait AsyncPlatform[F[_]] extends Serializable { this: Async[F]
   /**
    * Suspend a `java.util.concurrent.CompletableFuture` into the `F[_]` context.
    *
+   * Note that cancelation is cooperative, and it's up to the `CompletableFuture` to respond to
+   * interruption requests and handle cancelation appropriately. This means that if the
+   * `CompletableFuture` computation is uncancellable (for example, it doesn't handle `Thread`
+   * interruptions), there will be no 'fire and forget' semantics and by satisfying backpressure
+   * guarantees, cancelation will wait for computation to complete.
```
armanbilge: I'm not sure about the comment about "thread interruptions". The JavaDoc for `CompletableFuture#cancel` says this:

> mayInterruptIfRunning - this value has no effect in this implementation because interrupts are not used to control processing.

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#cancel-boolean-

WDYT about this?

```suggestion
   * @note Cancelation is implemented by invoking the `CompletableFuture`'s `cancel()` method.
   * It is up to the `CompletableFuture` to handle cancelation appropriately and indicate it has done so.
   * This means that if the `CompletableFuture` indicates that it did not cancel, there will be no
   * "fire-and-forget" semantics. Instead, to satisfy backpressure guarantees, the `CompletableFuture`
   * will be treated as if it is uncancelable and the fiber will fallback to waiting for it to complete.
```

armanbilge (commented):
Sorry, we should retarget this change at `series/3.5.x` branch so it will be released sooner. Thanks btw!

kernel/jvm/src/main/scala/cats/effect/kernel/AsyncPlatform.scala
danicheg: Uh-oh, probably the part on thread interruptions is really wrong. I like your rephrasing generally, but I would like to leave the reference that cancelation is cooperative. There are a few more mentions about it within CE's scaladocs. And maybe we need to remove the part on using the `cancel` method on CF since it's the only way to cancel it, and very unlike smth will change for Java's API.

danicheg:
@armanbilge do you merge changes from `series/3.x` to `series/3.5.x`? Or it's done conversely?

armanbilge:
We merge from `series/3.5.x` -> `series/3.x`.

kernel/jvm/src/main/scala/cats/effect/kernel/AsyncPlatform.scala
armanbilge: > but I would like to leave the reference that cancelation is cooperative. There are a few more mentions about it within CE's scaladocs.

Ahh, I'm sorry, you are right! This is used in many places. 

https://github.com/typelevel/cats-effect/blob/6a6db8fdd88cc5800d34bac64d7476bda5cff1bc/docs/concepts.md?plain=1#L102

I was confusing with "cooperative polling" 😅 TIL.

Ok, how about this? (Feel free to make your own revision and push to the branch :)

```suggestion
   * @note Cancelation is cooperative and it is up to the `CompletableFuture` to respond to the request
   * by handling cancelation appropriately and indicating that it has done so.
   * This means that if the `CompletableFuture` indicates that it did not cancel, there will be no
   * "fire-and-forget" semantics. Instead, to satisfy backpressure guarantees, the `CompletableFuture`
   * will be treated as if it is uncancelable and the fiber will fallback to waiting for it to complete.
```

danicheg: Seems great to me. Thanks for adjusting and pushing this forward.

armanbilge (approved):
Thank you, this is an excellent thing to note!

durban (approved):

Issue #3519, opened 2023-04-01
The Random docs skip over the hardest part ..and other doubts 


benhutchison:
- The [docs for `std.Random`](https://typelevel.org/cats-effect/docs/std/random) do not include a working example; ie if you concat the snippets on the page, it will not compile. And on closer inspection, you'll discover that the missing part is omitted from the docs not because it's too trivial to mention, but rather because it's the most difficult piece!  

The problem lies in the section "Creating a Random instance".  

```
Obtaining an instance of Random can be as simple as:

import cats.effect.IO
import cats.effect.std.Random

Random.scalaUtilRandom[IO]
``` 

Unfortunately, the user will discover that creating a random instance is not so _simple_ at all. The expression given is of type `IO[Random[IO]]`, so that's not actually an instance. 

In fact, the user will likely need to execute that expression within the program and then hoist the result into given/implicit scope, for later sections of the program to use. This is a slightly non-typical way to create instances and I couldn't find any such example in the docs.

- Secondly, the use of the word "instance" here is problematic. It has range of meanings. For people of Haskell persuasion, to call something that's created during the running program for a specific lifetime an "instance" would confuse them.

- And this poses another library design question: _why_ is Random an _instance_, that is, passed implicitly? Other objects in `std` like `Ref` and `Queue` are passed explicitly, but `Random` is passed implicitly. Is there a clear principle that specifies why a particular thing is passed explicitly vs implicitly?

armanbilge:
> The expression given is of type `IO[Random[IO]]`, so that's not actually an instance.
> 
> In fact, the user will likely need to execute that expression within the program

Yeah, this is a recurring shortcoming of our documentation, which I attempted to cover in https://github.com/typelevel/cats-effect/issues/3132. Basically, how to work with mutable state. Essentially you should treat it the same way as a `Ref`, which is also allocated as an `IO[Ref[A]]`

--- 

> but `Random` is passed implicitly. Is there a clear principle that specifies why a particular thing is passed explicitly vs implicitly?

This is also an excellent, recurring question 😁 one good rule of thumb is that unless there is a single, canonical instance, it probably should not be treated `implicit`ly. So we might want to reconsider this one.

benhutchison:
> > The expression given is of type `IO[Random[IO]]`, so that's not actually an instance.
> > In fact, the user will likely need to execute that expression within the program
> 
[...]

PR #3647, opened 2023-05-25
 Make `UUIDGen` more performant

### LINKED ISSUE #2882: Make `UUIDGen` more performant

diogocanut:
On this PR I'm trying to address #2882 .


std/jvm/src/main/scala/cats/effect/std/UUIDGenCompanionPlatform.scala
```diff
@@ -17,12 +17,44 @@
...
+    new UUIDGen[F] {
+      override final val randomUUID: F[UUID] =
+        for {
+          jsr <- SecureRandom.javaSecuritySecureRandom[F]
+          nBytes <- jsr.nextBytes(16)
+          bArray <- Sync[F].delay(setUUIDVersion(nBytes))
+          uuid <- Sync[F].delay(UUIDBuilder(bArray))
+        } yield uuid
+
+      private def setUUIDVersion(randomBytes: Array[Byte]) = {
```
diogocanut: Just copied those two unsafe methods from Java `UUID`. Not sure where is the best place to fit it.

std/shared/src/main/scala/cats/effect/std/SecureRandom.scala
```diff
@@ -136,8 +168,62 @@ object SecureRandom extends SecureRandomCompanionPlatform {
    * on Linux, macOS, and BSD. Unsupported platforms such as Windows will encounter link-time
    * errors.
    */
-  def javaSecuritySecureRandom[F[_]: Sync]: F[SecureRandom[F]] =
-    Sync[F]
-      .delay(new JavaSecureRandom())
+  def javaSecuritySecureRandom[F[_]: Sync]: F[SecureRandom[F]] = {
+    // This is a known, non-blocking, threadsafe algorithm
+    def happyRandom = Sync[F].delay(JavaSecureRandom.getInstance("NativePRNGNonBlocking"))
```
diogocanut: Can't access `getInstace` method from private `SecureRandomCompanionPlatform`

std/shared/src/main/scala/cats/effect/std/SecureRandom.scala
```diff
@@ -136,8 +168,62 @@ object SecureRandom extends SecureRandomCompanionPlatform {
...
+          )
+
+        case Some(_) | None =>
+          // We can't guarantee we're not stuck in a mutex.
+          fallbackPool
+      }
       .map(r => new ScalaRandom[F](Applicative[F].pure(r)) with SecureRandom[F] {})
+  }
+
+  private def javaMajorVersion[F[_]: Sync]: F[Option[Int]] =
```
diogocanut: Also not sure where to fit this method.

std/shared/src/main/scala/cats/effect/std/SecureRandom.scala
armanbilge: Sorry if I misunderstood. I think you should move this code to `SecureRandomCompanionPlatform`, because it will only work on the JVM.

std/shared/src/main/scala/cats/effect/std/SecureRandom.scala
armanbilge: Same with this one, it should go to `SecureRandomCompanionPlatform` because it is only useful on the JVM.

armanbilge (commented):
Nice start!!

std/jvm/src/main/scala/cats/effect/std/UUIDGenCompanionPlatform.scala
```diff
@@ -17,12 +17,44 @@
...
 
 private[std] trait UUIDGenCompanionPlatform {
-  implicit def fromSync[F[_]](implicit ev: Sync[F]): UUIDGen[F] = new UUIDGen[F] {
-    override final val randomUUID: F[UUID] =
-      ev.blocking(UUID.randomUUID())
+  implicit def fromSync[F[_]](implicit ev: Sync[F]): UUIDGen[F] = {
+    new UUIDGen[F] {
+      override final val randomUUID: F[UUID] =
+        for {
+          jsr <- SecureRandom.javaSecuritySecureRandom[F]
```
armanbilge: Making a new `SecureRandom` for every UUID will be expensive. Instead I think we can use an existing `SecureRandom` instance and define the method signature like this:

```scala
implicit def fromSecureRandom[F[_]: Functor: SecureRandom]: UUIDGen[F] =
```

armanbilge: Oh, and we'll have to keep the old one for source- and binary-compatibility. We'll have to move it to `UUIDGenCompanionPlatformLowPriority` and do `UUIDGenCompanionPlatform extends UUIDGenCompanionPlatformLowPriority`.

diogocanut: Moved to `UUIDGenCompanionPlatformLowPriority`

std/shared/src/main/scala/cats/effect/std/SecureRandom.scala
diogocanut: Also moved to `SecureRandomCompanionPlatform`

std/jvm/src/main/scala/cats/effect/std/SecureRandomCompanionPlatform.scala
```diff
@@ -14,8 +14,107 @@
...
+  private def javaUtilRandomBlocking[F[_]: Sync](random: JavaSecureRandom): F[SecureRandom[F]] =
+    Sync[F]
+      .blocking(random)
+      .map(r => new ScalaRandom[F](Applicative[F].pure(r)) with SecureRandom[F] {})
+
+  /**
+   * Creates a SecureRandom instance. On most platforms, it will be non-blocking. If a
+   * non-blocking instance can't be guaranteed, falls back to a blocking implementation.
+   */
+  def nonBlockingJavaSecuritySecureRandom[F[_]: Sync]: F[SecureRandom[F]] = {
```
diogocanut: Now I created this method here that will work only on JVM, but it doesn't affect `javaSecuritySecureRandom`. Could I do something there inside `javaSecuritySecureRandom` like this?

```
if (isJvm) nonBlockingJavaSecuritySecureRandom
else ...
```

armanbilge: Hm, I'm not sure if that will work. We may have to move `def javaSecuritySecureRandom` into the platform trait as well.

std/jvm/src/main/scala/cats/effect/std/UUIDGenCompanionPlatform.scala
```diff
@@ -20,7 +20,7 @@ import cats.effect.kernel.Sync
 
 import java.util.UUID
 
-private[std] trait UUIDGenCompanionPlatform {
+private[std] trait UUIDGenCompanionPlatform extends UUIDGenCompanionPlatformLowPriority {
   implicit def fromSync[F[_]](implicit ev: Sync[F]): UUIDGen[F] = new UUIDGen[F] {
```
armanbilge, line 31: Actually, we should move this one into the `LowPriority` trait

std/jvm/src/main/scala/cats/effect/std/UUIDGenCompanionPlatformLowPriority.scala
```diff
@@ -0,0 +1,55 @@
...
+
+package cats.effect.std
+
+import cats.Functor
+import cats.syntax.all._
+
+import java.util.UUID
+
+private[std] trait UUIDGenCompanionPlatformLowPriority {
+  implicit def fromSecureRandom[F[_]: Functor: SecureRandom]: UUIDGen[F] = {
```
armanbilge: ... and this one should move into the (high priority) trait. Because, we want this new one you added to be higher-priority than the old, blocking one :)

Also, it's ok to keep `UUIDGenCompanionPlatformLowPriority` in the same file as `UUIDGenCompanionPlatform`.

std/jvm/src/main/scala/cats/effect/std/UUIDGenCompanionPlatformLowPriority.scala
```diff
@@ -0,0 +1,55 @@
...
+        for (i <- 0 until 8) {
+          msb = (msb << 8) | (data(i) & 0xff)
+        }
+        for (i <- 8 until 16) {
+          lsb = (lsb << 8) | (data(i) & 0xff)
+        }
+        val mostSigBits = msb
+        val leastSigBits = lsb
+        new UUID(mostSigBits, leastSigBits)
+      }
```
armanbilge: For performance, we should do both of these in the same `map` call. Also mutating the `Array[Byte]` is technically a side-effect. Maybe we should share the code from here?

https://github.com/typelevel/cats-effect/blob/6a6db8fdd88cc5800d34bac64d7476bda5cff1bc/std/js-native/src/main/scala/cats/effect/std/UUIDGenCompanionPlatform.scala#L49-L59

diogocanut: I created a new file in `shared` named `UnsafeUUIDBuilder` with this logic, WDYT?

std/jvm/src/main/scala/cats/effect/std/UUIDGenCompanionPlatformLowPriority.scala
diogocanut: done, also put both in same file.

std/jvm/src/main/scala/cats/effect/std/SecureRandomCompanionPlatform.scala
diogocanut: I removed  `javaSecuritySecureRandom` from `SecureRandom` and renamed this method. I got two issues now, the first is `ambiguous reference to overloaded definition` , because we also have `javaSecuritySecureRandom` in object `SecureRandom` of type (n: Int). 

And the second is that will require a implementation of `javaSecuritySecureRandom` for native and js.

Logs:

```
[error] /Users/diogocanut/Projects/cats-effect/std/shared/src/main/scala/cats/effect/std/Random.scala:331:42: ambiguous reference to overloaded definition,
[error] both method javaSecuritySecureRandom in object SecureRandom of type (n: Int)(implicit evidence$16: cats.effect.kernel.Sync[F]): F[cats.effect.std.SecureRandom[F]]
[error] and  method javaSecuritySecureRandom in trait SecureRandomCompanionPlatform of type (implicit evidence$3: cats.effect.kernel.Sync[F]): F[cats.effect.std.SecureRandom[F]]
[error] match expected type ?
[error]     SecureRandom.javaSecuritySecureRandom[F].widen[Random[F]]
```

std/shared/src/main/scala/cats/effect/std/UnsafeUUIDBuilder.scala
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package cats.effect.std
+
+import java.util.UUID
+
+trait UnsafeUUIDBuilder {
+
+  def build(buffer: Array[Byte]): UUID
+}
```
armanbilge: Sorry if I missed it, I don't think that this `trait` is used?

std/shared/src/main/scala/cats/effect/std/UnsafeUUIDBuilder.scala
```diff
@@ -0,0 +1,40 @@
...
+
+object UnsafeUUIDBuilder {
+  def build(buffer: Array[Byte]): UUID = {
```
armanbilge: Actually I think we can make this a private method in `SecureRandom.scala`. Then the JS/Native `UUIDGen` can actually be based on the method you added `SecureRandom[F] => UUIDGen[F]`.

std/jvm/src/main/scala/cats/effect/std/SecureRandomCompanionPlatform.scala
armanbilge: > And the second is that will require a implementation of `javaSecuritySecureRandom` for native and js.

Sure, we can add it, based on the old implementation that we already had :)

> I got two issues now, the first is `ambiguous reference to overloaded definition` , because we also have `javaSecuritySecureRandom` in object `SecureRandom` of type (n: Int).

Sorry, I think I'm a bit confused. If I understand correctly, I don't think we should be adding any new public methods to `SecureRandom`. We are just improving the implementation of the existing method.

diogocanut: Added old implementation to native and js and compilation worked.

std/shared/src/main/scala/cats/effect/std/UnsafeUUIDBuilder.scala
diogocanut: not sure if I understood it correctly. Should we have something like this inside `SecureRandom.scala`?

```
  def UUIDGen[F[_]: SecureRandom: Functor]: UUIDGen[F] =
    new UUIDGen[F] {
      override final val randomUUID: F[UUID] =
        SecureRandom[F].nextBytes(16).map(unsafeUUIDBuilder)
    }
```

Not sure what you means by having this method private inside `SecureRandom`, how should I call it from JS/Native, is it something like de code above?

armanbilge: > Not sure what you means by having this method private inside `SecureRandom`

Woops, I'm sorry! I meant private inside of `UUIDGen.scala`. Yes, exactly, I was imagining exactly what you wrote there.

std/shared/src/main/scala/cats/effect/std/SecureRandom.scala
```diff
@@ -122,4 +122,8 @@ object SecureRandom extends SecureRandomCompanionPlatform {
       def selectRandom = incrGet.map(array(_))
       new ScalaRandom[F](selectRandom) with SecureRandom[F] {}
     }
+
+  override def javaSecuritySecureRandom[F[_]: Sync]: F[SecureRandom[F]] =
```
diogocanut, line 139: This was needed due to error
```
[error] /Users/diogocanut/Projects/cats-effect/std/shared/src/main/scala/cats/effect/std/Random.scala:331:42: ambiguous reference to overloaded definition,
[error] both method javaSecuritySecureRandom in object SecureRandom of type (n: Int)(implicit evidence$16: cats.effect.kernel.Sync[F]): F[cats.effect.std.SecureRandom[F]]
[error] and  method javaSecuritySecureRandom in trait SecureRandomCompanionPlatform of type (implicit evidence$1: cats.effect.kernel.Sync[F]): F[cats.effect.std.SecureRandom[F]]
[error] match expected type ?
[error]     SecureRandom.javaSecuritySecureRandom[F].widen[Random[F]]
```

std/js-native/src/main/scala/cats/effect/std/UUIDGenCompanionPlatform.scala
```diff
@@ -32,21 +32,34 @@ import cats.effect.kernel.Sync
...
-        /* We use nextBytes() because that is the primitive of most secure RNGs,
-         * and therefore it allows to perform a unique call to the underlying
-         * secure RNG.
-         */
-        csprng.nextBytes(randomUUIDBuffer)
-
-        UnsafeUUIDBuilder.build(buffer)
-      }
-  }
+private[std] trait UUIDGenCompanionPlatform extends UUIDGenCompanionPlatformLowPriority
```
diogocanut, line 33: I left old method unchanged and add it into `UUIDGenCompanionPlatformLowPriority`. Let me know if it makes sense.

std/js-native/src/main/scala/cats/effect/std/UUIDGenCompanionPlatform.scala
```diff
@@ -32,21 +32,34 @@ import cats.effect.kernel.Sync
...
+
+private[std] trait UUIDGenCompanionPlatformLowPriority {
+  implicit def fromSync[F[_]](implicit ev: Sync[F]): UUIDGen[F] =
+    new UUIDGen[F] {
```
armanbilge: Can we implement this using `fromSecureRandom`? We'll have to wrap the `JavaSecureRandom` into a `SecureRandom[F]`, but then we don't need to duplicate the UUID code.

diogocanut: done.

.gitignore
```diff
@@ -20,3 +20,5 @@ metals.sbt
 
 # npm
 node_modules/
+
+.direnv/
```
durban, line 25: Do we need this? If yes, can we add a comment about what is this?

std/js-native/src/main/scala/cats/effect/std/UUIDGenCompanionPlatform.scala
```diff
@@ -28,35 +28,19 @@
...
 
+import cats.Applicative
 import cats.effect.kernel.Sync
+import cats.effect.std.Random.ScalaRandom
+
+private[std] trait UUIDGenCompanionPlatform extends UUIDGenCompanionPlatformLowPriority
+
+private[std] trait UUIDGenCompanionPlatformLowPriority {
+
+  private implicit def secureRandom[F[_]](implicit ev: Sync[F]): SecureRandom[F] =
```
durban: If this is just used here locally, I recommend making it a `private[this] def`, and passing explicitly.

std/jvm/src/main/scala/cats/effect/std/SecureRandomCompanionPlatform.scala
```diff
@@ -14,8 +14,119 @@
...
+   *
+   * @param random
+   *   a non-blocking instance of java.util.Random. It is the caller's responsibility to assure
+   *   that the instance is non-blocking. If the instance blocks during seeding, it is the
+   *   caller's responsibility to seed it.
+   */
+  private def javaUtilRandomNonBlocking[F[_]: Sync](
+      random: JavaSecureRandom): F[SecureRandom[F]] =
+    Sync[F]
+      .delay(random)
```
durban: I'm not sure what the intention was, but this line seems to do approximately nothing(?).

std/jvm/src/main/scala/cats/effect/std/SecureRandomCompanionPlatform.scala
```diff
@@ -14,8 +14,119 @@
...
+      .map(r => new ScalaRandom[F](Applicative[F].pure(r)) with SecureRandom[F] {})
+
+  /**
+   * Creates a blocking Random instance. All calls to nextBytes are shifted to the blocking
+   * pool. This is safer, but less effecient, than [[javaUtilRandomNonBlocking]].
```
durban: "... are shifted to the blocking pool" -> this doesn't seem to be true (at least I can't see where that would happen).

std/jvm/src/main/scala/cats/effect/std/SecureRandomCompanionPlatform.scala
```diff
@@ -14,8 +14,119 @@
...
+   *
+   * @param random
+   *   a potentially blocking instance of java.util.Random
+   */
+  private def javaUtilRandomBlocking[F[_]: Sync](random: JavaSecureRandom): F[SecureRandom[F]] =
+    Sync[F]
+      .blocking(random)
```
durban: This line also seems to be doing nothing.

durban (dismissed):

.gitignore
armanbilge: Yes, it's because of the `.envrc` file added in https://github.com/typelevel/cats-effect/pull/3431. Ross didn't add the `.gitignore` in that PR because he believes in global `.gitignore`s for these sorts of things.

https://github.com/typelevel/cats-effect/blob/1e7fa937fa9b79d7d733b3876556a60ce2e6dbf4/.envrc#L1

std/shared/src/main/scala/cats/effect/std/SecureRandom.scala
```diff
@@ -123,21 +123,7 @@ object SecureRandom extends SecureRandomCompanionPlatform {
...
-   * In browsers, delegates to the
-   * [[https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API Web Crypto API]].
-   *
-   * In Node.js, delegates to the [[https://nodejs.org/api/crypto.html crypto module]].
-   *
-   * On Native, delegates to
-   * [[https://man7.org/linux/man-pages/man3/getentropy.3.html getentropy]] which is supported
-   * on Linux, macOS, and BSD. Unsupported platforms such as Windows will encounter link-time
-   * errors.
-   */
```
armanbilge: We should keep this Scaladoc :)

std/jvm/src/main/scala/cats/effect/std/SecureRandomCompanionPlatform.scala
```diff
@@ -14,8 +14,119 @@
...
+        )
+
+      case Some(_) | None =>
+        // We can't guarantee we're not stuck in a mutex.
+        fallbackPool
+    }
+  }
+
+  private def javaMajorVersion[F[_]: Sync]: F[Option[Int]] =
+    Sync[F].delay(sys.props.get("java.version")).map(_.flatMap(parseJavaMajorVersion))
```
armanbilge:
```suggestion
    Sync[F].delay(Option(System.getProperty("java.version"))).map(_.flatMap(parseJavaMajorVersion))
```

std/js-native/src/main/scala/cats/effect/std/UUIDGenCompanionPlatform.scala
```diff
@@ -28,35 +28,19 @@
...
+    new ScalaRandom[F](Applicative[F].pure(new SecureRandom.JavaSecureRandom()))
+      with SecureRandom[F] {}
+
+  implicit def fromSync[F[_]](implicit ev: Sync[F]): UUIDGen[F] =
```
armanbilge: I propose we mark this as `@deprecated` in favor of the `SecureRandom` version.

std/jvm/src/main/scala/cats/effect/std/UUIDGenCompanionPlatform.scala
```diff
@@ -20,9 +20,13 @@ import cats.effect.kernel.Sync
 
 import java.util.UUID
 
-private[std] trait UUIDGenCompanionPlatform {
+private[std] trait UUIDGenCompanionPlatform extends UUIDGenCompanionPlatformLowPriority
+
+private[std] trait UUIDGenCompanionPlatformLowPriority {
+
   implicit def fromSync[F[_]](implicit ev: Sync[F]): UUIDGen[F] = new UUIDGen[F] {
```
armanbilge, line 31: Same with this one, `@deprecated`. This will nudge users to the higher performance option.

std/js-native/src/main/scala/cats/effect/std/UUIDGenCompanionPlatform.scala
diogocanut: Like this?

```
  private[this] def secureRandom[F[_]](implicit ev: Sync[F]): SecureRandom[F] =
    new ScalaRandom[F](Applicative[F].pure(new SecureRandom.JavaSecureRandom()))
      with SecureRandom[F] {}

  implicit def fromSync[F[_]](implicit ev: Sync[F]): UUIDGen[F] = {
    implicit val secureRandom: SecureRandom[F] = secureRandom
    UUIDGen.fromSecureRandom[F]
  }
```

std/jvm/src/main/scala/cats/effect/std/UUIDGenCompanionPlatform.scala
diogocanut: Just wondering what message should we use, maybe something like this?

`@deprecated("Call SecureRandom.javaSecuritySecureRandom and UUIDGen.fromSecureRandom", "3.6.0")`

armanbilge: Hmm, good question. What about something like this, since most users will be getting it implicitly?
```scala
@deprecated(
  "Put an implicit `SecureRandom.javaSecuritySecureRandom` into scope to get a more efficient `UUIDGen`, or directly call `UUIDGen.fromSecureRandom`",
  "3.6.0"
)

```

diogocanut: That's a good one.

std/js-native/src/main/scala/cats/effect/std/UUIDGenCompanionPlatform.scala
durban: No, you don't need an `implicit` at all (I mean besides `fromSync`) – an implicit argument can be passed explicitly.

diogocanut: got it, pushing my changes.

std/jvm/src/main/scala/cats/effect/std/SecureRandomCompanionPlatform.scala
diogocanut: what's your suggestion for here and the blocking one? I need to wrap it in F here:

```
happyRandom.redeemWith(
  _ =>
    fallback.flatMap {
      case rnd if isThreadsafe(rnd) =>
        // We avoided the mutex, but not the blocking.  Use a
        // shared instance from the blocking pool.
        javaUtilRandomBlocking(rnd)
      case _ =>
        // We can't prove the instance is threadsafe, so we need
        // to pessimistically fall back to a pool.  This should
        // be exceedingly uncommon.
        fallbackPool
    },
  // We are thread safe and non-blocking.  This is the
  // happy path, and happily, the common path.
  rnd => javaUtilRandomNonBlocking(rnd)
        )
```

diogocanut:
After deprecating old `UUIDGen` `fromSync` methods, there's one pipeline check failing, because it's running in no warning mode, and there's a warning related to this line of code in `IO.scala`.

```scala
  def randomUUID: IO[UUID] = UUIDGen[IO].randomUUID
``` 

Warning:

```
[warn] /Users/diogocanut/Projects/cats-effect/core/shared/src/main/scala/cats/effect/IO.scala:1447:37: method fromSync in trait UUIDGenCompanionPlatformLowPriority is deprecated (since 3.6.0): Put an implicit `SecureRandom.javaSecuritySecureRandom` into scope to get a more efficient `UUIDGen`, or directly call `UUIDGen.fromSecureRandom`
[warn]   def randomUUID: IO[UUID] = UUIDGen[IO].randomUUID
[warn]                                     ^
[warn] one warning found
```

Should we provide an implicit for `SecureRandom.javaSecuritySecureRandom` somehow?

std/jvm/src/main/scala/cats/effect/std/SecureRandomCompanionPlatform.scala
```diff
@@ -14,8 +14,118 @@
...
+   * [[https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API Web Crypto API]].
+   *
+   * In Node.js, delegates to the [[https://nodejs.org/api/crypto.html crypto module]].
+   *
+   * On Native, delegates to
+   * [[https://man7.org/linux/man-pages/man3/getentropy.3.html getentropy]] which is supported
+   * on Linux, macOS, and BSD. Unsupported platforms such as Windows will encounter link-time
+   * errors.
+   */
+  def javaSecuritySecureRandom[F[_]: Sync]: F[SecureRandom[F]] = {
```
armanbilge: Can we move this implementation into a `private[effect] def unsafeSecuritySecureRandom[F[_]: Sync]: SecureRandom[F]`? Then we can use that method in `IO.randomUUID` to get a secure random.

diogocanut: Sure. Just not sure if I got it right, are we going to keep this implementation here and create a new unsafe method using similar implementation? Or the intent is to remove the code here at `SecureRandomCompanionPlatform`, have old `javaSecuritySecureRandom` inside `SecureRandom` and move this implementation as a unsafe version to something similar as `IOCompanionPlatform`?

armanbilge: Basically something like this:

```scala
def javaSecuritySecureRandom[F[_]: Sync]: F[SecureRandom[F]] =
  Sync[F].delay(unsafeJavaSecuritySecureRandom())

private[effect] def unsafeJavaSecuritySecureRandom[F[_]: Sync]: SecureRandom[F] =
  ...
```

diogocanut: is it ok to implement our fallback pool without `Ref`?

```scala
  def javaSecuritySecureRandom[F[_]: Sync](n: Int): F[SecureRandom[F]] =
    for {
      ref <- Ref[F].of(0)
      array <- Sync[F].delay(Array.fill(n)(new SRandom(new JavaSecureRandom())))
    } yield {
      def incrGet = ref.modify(i => (if (i < (n - 1)) i + 1 else 0, i))
      def selectRandom = incrGet.map(array(_))
      new ScalaRandom[F](selectRandom) with SecureRandom[F] {}
    }
```

armanbilge: yes, you can actually use an `AtomicInteger` and will have a lot better performance 😅

std/jvm/src/main/scala/cats/effect/std/SecureRandomCompanionPlatform.scala
```diff
@@ -14,8 +14,111 @@
...
+    javaMajorVersion match {
+      case Some(major) if major > 8 =>
+        try {
+          // We are thread safe and non-blocking.  This is the
+          // happy path, and happily, the common path.
+          javaUtilRandom(happyRandom)
+        } catch {
+          case _: Throwable =>
+            fallback match {
+              case rnd if isThreadsafe(rnd) =>
```
diogocanut, line 81: Not sure if this still make sense without `Sync[F].blocking`. Maybe we can remove this match and call fallbackPool directly?

armanbilge: You're right, it doesn't. If the `SecureRandom` is blocking we _must_ using `Sync[F].blocking(...)` similar to how it's done here.
https://github.com/http4s/http4s/blob/9f83a48f1b46f7baa1b96ec053cb0be5e640cd15/core/src/main/scala/org/http4s/internal/Random.scala#L53-L65

diogocanut: Got it, in our case it's only relevant to have a blocking implementation of `nextBytes`?

armanbilge: Hmm, no, we'd have to do it for all the `next` methods 😕 
https://github.com/diogocanut/cats-effect/blob/1c6b6ecb604c3701fe58bb334e8352adb0375da9/std/shared/src/main/scala/cats/effect/std/Random.scala#L437

Maybe instead of `delay` or `blocking` we should use `suspend` and take the `Sync.Type` as a constructor parameter. Then at least we don't need to duplicate the code.

diogocanut: Nice, done.

std/shared/src/main/scala/cats/effect/std/Random.scala
```diff
@@ -434,17 +434,20 @@ object Random extends RandomCompanionPlatform {
 
   }
 
-  private[std] abstract class ScalaRandom[F[_]: Sync](f: F[SRandom]) extends RandomCommon[F] {
+  private[std] abstract class ScalaRandom[F[_]: Sync](
+      f: F[SRandom],
+      hint: Sync.Type = Sync.Type.Delay)
+      extends RandomCommon[F] {
```
armanbilge: Something like this should fix the CI failure.

```suggestion
  private[std] abstract class ScalaRandom[F[_]: Sync](
      f: F[SRandom],
      hint: Sync.Type)
      extends RandomCommon[F] {
      
      def this(f: F[SRandom]) = this(f, Sync.Type.Delay)
```

armanbilge (commented):
I think we're almost there ... this PR turned out to be incredibly complicated to figure out how to organize everything. Suffice to say, very very nicely done with this, thanks for persevering 💯

std/shared/src/main/scala/cats/effect/std/Random.scala
diogocanut: Nice! Now CI is working.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1963,6 +1963,9 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
   // later in the file
   private[this] val _never: IO[Nothing] = asyncForIO.never
 
+  implicit def secureRandom: SecureRandom[IO] =
```
armanbilge: Otherwise it will create a new one each time :)

```suggestion
  lazy val secureRandom: SecureRandom[IO] =
```

---

Call-out for other reviewers: are we okay about unsafely creating a global secure random like this? I think having this would greatly enhance UX, but it creates a muddy precedent to use a pattern that has gone wrong sometimes e.g. FS2 `Network` ...

std/js/src/main/scala/cats/effect/std/SecureRandomCompanionPlatform.scala
```diff
@@ -110,4 +114,10 @@ private[std] trait SecureRandomCompanionPlatform {
     }
   }
 
+  def javaSecuritySecureRandom[F[_]: Sync]: F[SecureRandom[F]] =
+    Sync[F].delay(unsafeJavaSecuritySecureRandom)
+
+  private[effect] def unsafeJavaSecuritySecureRandom[F[_]: Sync]: SecureRandom[F] =
```
armanbilge: It's convention for unsafe, side-effecting methods to always use parentheses to emphasize that they are side-effecting.

```suggestion
  private[effect] def unsafeJavaSecuritySecureRandom[F[_]: Sync](): SecureRandom[F] =
```

std/jvm/src/main/scala/cats/effect/std/SecureRandomCompanionPlatform.scala
```diff
@@ -14,8 +14,120 @@
...
+  private def javaUtilRandom[F[_]: Sync](random: JavaSecureRandom): SecureRandom[F] =
+    new ScalaRandom[F](Applicative[F].pure(random)) with SecureRandom[F] {}
+
+  /**
+   * Creates a blocking Random instance.
+   *
+   * @param random
+   *   a potentially blocking instance of java.util.Random
+   */
+  def javaUtilRandomBlocking[F[_]: Sync](random: JavaSecureRandom): SecureRandom[F] =
```
armanbilge: This can be private like `javaUtilRandom` above.

```suggestion
  private def javaUtilRandomBlocking[F[_]: Sync](random: JavaSecureRandom): SecureRandom[F] =
```

std/jvm/src/main/scala/cats/effect/std/SecureRandomCompanionPlatform.scala
```diff
@@ -14,8 +14,120 @@
...
+import cats.effect.kernel._
+import cats.effect.std.Random.ScalaRandom
+
+import scala.util.{Random => SRandom, Try}
+
+import java.util.concurrent.atomic.AtomicInteger
 
 private[std] trait SecureRandomCompanionPlatform {
   private[std] type JavaSecureRandom = java.security.SecureRandom
+  private[std] def getInstance: JavaSecureRandom =
```
armanbilge:
```suggestion
  private def getInstance: JavaSecureRandom =
```

core/shared/src/main/scala/cats/effect/IO.scala
diogocanut: Hmm, how to add the implicit to scope having it as lazy val?

```
[warn] /Users/diogocanut/Projects/cats-effect/core/shared/src/main/scala/cats/effect/IO.scala:1447:37: method fromSync in trait UUIDGenCompanionPlatformLowPriority is deprecated (since 3.6.0): Put an implicit `SecureRandom.javaSecuritySecureRandom` into scope to get a more efficient `UUIDGen`, or directly call `UUIDGen.fromSecureRandom`
[warn]   def randomUUID: IO[UUID] = UUIDGen[IO].randomUUID
```

armanbilge: Oh sorry, I meant to write `implicit lazy val` :)

std/jvm/src/main/scala/cats/effect/std/SecureRandomCompanionPlatform.scala
```diff
@@ -14,8 +14,120 @@
...
+    }
+
+    javaMajorVersion match {
+      case Some(major) if major > 8 =>
+        try {
+          // We are thread safe and non-blocking.  This is the
+          // happy path, and happily, the common path.
+          javaUtilRandom(happyRandom)
+        } catch {
+          case _: Throwable =>
```
armanbilge:
```suggestion
          case ex if NonFatal(ex) =>
```

std/jvm/src/main/scala/cats/effect/std/SecureRandomCompanionPlatform.scala
```diff
@@ -14,8 +14,120 @@
...
+   * In browsers, delegates to the
+   * [[https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API Web Crypto API]].
+   *
+   * In Node.js, delegates to the [[https://nodejs.org/api/crypto.html crypto module]].
+   *
+   * On Native, delegates to
+   * [[https://man7.org/linux/man-pages/man3/getentropy.3.html getentropy]] which is supported
+   * on Linux, macOS, and BSD. Unsupported platforms such as Windows will encounter link-time
+   * errors.
+   */
```
armanbilge: We retained this doc in the shared sources.
```suggestion
```

armanbilge (commented):
One last minor thing, I think!

core/shared/src/main/scala/cats/effect/IO.scala
durban: @armanbilge What was the problem in FS2 `Network`?

armanbilge: Well, it's a rather involved discussion :) so the problem is that FS2 promises an implicit `Network[IO]`, but this is technically encapsulating some global resources (namely, I/O threads). This was working well enough, until we started the polling system work. It created a significant challenge, since now we have to thread the `IOApp`s polling system into this globally promised `Network[IO]`.

If instead FS2 had been offering a `Resource[IO, Network[IO]]` that you had to construct explicitly, it would have been a lot easier to work with.

The fundamental question here is, when is it okay to "cheat" and unsafely allocate global shared state or resources, instead of forcing it to be explicitly allocated in `IO` or `Resource` by the user. It's some kind of UX trade-off.

See a similar place where we faced this in:
- https://github.com/typelevel/cats-effect/pull/3465

durban: Thanks, I think I understand. Btw, I'm not convinced a `Random[IO]` must be implicitly available: there isn't "one obvious preferable" `Random[IO]`. However, for `SecureRandom`, there could be: in fact, this PR tries to do that... so I think I'm okay with it in this specific case. (But it's possible i just don't appreciate the possible future pain this could cause :-)

std/shared/src/main/scala/cats/effect/std/SecureRandom.scala
```diff
@@ -136,8 +136,7 @@ object SecureRandom extends SecureRandomCompanionPlatform {
    * on Linux, macOS, and BSD. Unsupported platforms such as Windows will encounter link-time
    * errors.
    */
-  def javaSecuritySecureRandom[F[_]: Sync]: F[SecureRandom[F]] =
-    Sync[F]
-      .delay(new JavaSecureRandom())
-      .map(r => new ScalaRandom[F](Applicative[F].pure(r)) with SecureRandom[F] {})
+  override def javaSecuritySecureRandom[F[_]: Sync]: F[SecureRandom[F]] =
+    super.javaSecuritySecureRandom[F]
```
durban, line 140: The other overload above (line 114) refers to this method. However, the behavior of the 2 are now quite different. I think the other overload could also benefit from the new improvements, although the `n: Int` does not make a lot of sense for the best case (nonblocking) PRNG.

std/js-native/src/main/scala/cats/effect/std/UUIDGenCompanionPlatform.scala
```diff
@@ -28,35 +28,24 @@
...
-        val i4 = intFromBuffer(12)
-        val msb = (i1.toLong << 32) | (i2.toLong & 0xffffffffL)
-        val lsb = (i3.toLong << 32) | (i4.toLong & 0xffffffffL)
-        new UUID(msb, lsb)
-      }
+private[std] trait UUIDGenCompanionPlatform extends UUIDGenCompanionPlatformLowPriority
+
+private[std] trait UUIDGenCompanionPlatformLowPriority {
+
+  private[this] def secureRandom[F[_]](implicit ev: Sync[F]): SecureRandom[F] =
```
durban: Isn't this method exactly the same as `unsafeJavaSecuritySecureRandom`?

std/jvm/src/main/scala/cats/effect/std/SecureRandomCompanionPlatform.scala
```diff
@@ -14,8 +14,104 @@
...
+        fallbackPool
+    }
+  }
+
+  private def pool[F[_]: Sync](n: Int): SecureRandom[F] = {
+    val index = new AtomicInteger(0)
+    val array = Array.fill(n)(new SRandom(new SecureRandom.JavaSecureRandom))
+
+    def selectRandom: F[SRandom] = Sync[F].delay {
+      val currentIndex = index.getAndUpdate(i => if (i < (n - 1)) i + 1 else 0)
```
durban: Isn't this `(i + 1) % n`?

std/shared/src/main/scala/cats/effect/std/SecureRandom.scala
diogocanut: Yes, for the best case `n: Int`. It will only sense in the worst case, when we fall back to a pool.

```scala
  private def pool[F[_]: Sync](n: Int): SecureRandom[F] = {
    val index = new AtomicInteger(0)
    val array = Array.fill(n)(new SRandom(new SecureRandom.JavaSecureRandom))

    def selectRandom: F[SRandom] = Sync[F].delay {
      val currentIndex = index.getAndUpdate(i => if (i < (n - 1)) i + 1 else 0)
      array(currentIndex)
    }

    new ScalaRandom[F](selectRandom) with SecureRandom[F] {}
  }
```

std/js-native/src/main/scala/cats/effect/std/UUIDGenCompanionPlatform.scala
diogocanut: Not sure if I got your question, you mean by method name? Because `unsafeJavaSecuritySecureRandom` approach will only work in `JVM`.

durban: I meant that the body of this method seems exactly the same as the body of `unsafeJavaSecuritySecureRandom`.

diogocanut: Yes, you are right. Should we rename it?

armanbilge: Instead of renaming, can we remove it and just use `unsafeJavaSecuritySecureRandom()`?

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1963,6 +1963,9 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
   // later in the file
   private[this] val _never: IO[Nothing] = asyncForIO.never
 
+  implicit lazy val secureRandom: SecureRandom[IO] =
+    SecureRandom.unsafeJavaSecuritySecureRandom[IO]()
```
djspiewak, line 2192: This suddenly introduces an implicit `Random[IO]`, which doesn't exist right now, and means that `SecureRandom` will end up being the de facto `Random` used in most situations. Is that what we want?

std/js-native/src/main/scala/cats/effect/std/UUIDGenCompanionPlatform.scala
```diff
@@ -30,33 +30,16 @@ package cats.effect.std
...
-        val msb = (i1.toLong << 32) | (i2.toLong & 0xffffffffL)
-        val lsb = (i3.toLong << 32) | (i4.toLong & 0xffffffffL)
-        new UUID(msb, lsb)
-      }
+  @deprecated(
+    "Put an implicit `SecureRandom.javaSecuritySecureRandom` into scope to get a more efficient `UUIDGen`, or directly call `UUIDGen.fromSecureRandom`",
+    "3.6.0"
+  )
+  implicit def fromSync[F[_]](implicit ev: Sync[F]): UUIDGen[F] = {
+    UUIDGen.fromSecureRandom[F](ev, SecureRandom.unsafeJavaSecuritySecureRandom())
```
djspiewak, line 42: If we're deprecating this method, shouldn't we make it non-`implicit`?

std/jvm/src/main/scala/cats/effect/std/UUIDGenCompanionPlatform.scala
```diff
@@ -20,9 +20,17 @@ import cats.effect.kernel.Sync
...
-private[std] trait UUIDGenCompanionPlatform {
+private[std] trait UUIDGenCompanionPlatform extends UUIDGenCompanionPlatformLowPriority
+
+private[std] trait UUIDGenCompanionPlatformLowPriority {
+
+  @deprecated(
+    "Put an implicit `SecureRandom.javaSecuritySecureRandom` into scope to get a more efficient `UUIDGen`, or directly call `UUIDGen.fromSecureRandom`",
+    "3.6.0"
+  )
   implicit def fromSync[F[_]](implicit ev: Sync[F]): UUIDGen[F] = new UUIDGen[F] {
```
djspiewak, line 31: Same comment as earlier regarding removing the `implicit` modifier

core/shared/src/main/scala/cats/effect/IO.scala
armanbilge: Oh, good catch, I totally missed this because of the inheritance relationship, and maybe this is what @durban was getting at in https://github.com/typelevel/cats-effect/pull/3647#discussion_r1220241936.

Well, it's tricky. To be honest I'm not entirely sure if I see the harm, but these sorts of implicits have bitten us in the past. In fact, `UUIDGen` being an implicit is biting us already ...

std/js-native/src/main/scala/cats/effect/std/UUIDGenCompanionPlatform.scala
armanbilge: No. If we take away the `implicit`, then user code will just stop compiling. If we keep a lower-priority deprecated implicit, then users will get migration help.

std/jvm/src/main/scala/cats/effect/std/UUIDGenCompanionPlatform.scala
armanbilge: Same issue :)

core/shared/src/main/scala/cats/effect/IO.scala
durban: I didn't think about this when I wrote that coment, but I agree that it _could_ be a problem. Although, the worst problem I can think of right now is that the `SecureRandom` is probably slower than a non-secure `Random`, which is not the end of the world... but it is still not clear to me how `Random` is expected to be used: a user creates an instance themselves, then passes that instance implicitly?

std/shared/src/main/scala/cats/effect/std/SecureRandom.scala
durban: I wasn't very clear: I'm thinking about the fact that the method on line 116 is creating an rng which is "sharded", but calls (potentially) blocking code in `delay`. The latter is solved by this PR for the other overload. I think it should be solved for the overload in line 116 too.

durban (commented):
A general comment: does this PR contain copied code from http4s? If yes, maybe that should be mentioned somewhere...

core/shared/src/main/scala/cats/effect/IO.scala
armanbilge: Yeah it's not the end of the world. Much better to default to `SecureRandom` than the other way around lol ...

> but it is still not clear to me how `Random` is expected to be used: a user creates an instance themselves, then passes that instance implicitly?

See also:
- https://github.com/typelevel/cats-effect/issues/3519

std/shared/src/main/scala/cats/effect/std/SecureRandom.scala
diogocanut: Hmm, what's your suggestion for implementing method on line 116 with our new implementation of `javaSecuritySecureRandom`?

djspiewak (dismissed):

djspiewak:
Conflicts need to be resolved but otherwise g2g

djspiewak (approved):

PR #3650, opened 2023-05-28
Remove workaround for `Scala3` in `Supervisor`

danicheg:
This is a follow-up to https://github.com/lampepfl/dotty/issues/15546.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -907,50 +907,58 @@ lazy val std = crossProject(JSPlatform, JVMPlatform, NativePlatform)
       "org.scalacheck" %%% "scalacheck" % ScalaCheckVersion % Test,
       "org.specs2" %%% "specs2-scalacheck" % Specs2Version % Test
     ),
-    mimaBinaryIssueFilters ++= Seq(
-      // introduced by #2604, Fix Console on JS
-      // changes to `cats.effect.std` package private code
-      ProblemFilters.exclude[MissingClassProblem]("cats.effect.std.Console$SyncConsole"),
-      // introduced by #2951
-      // added configurability to Supervisor's scope termination behavior
-      // the following are package-private APIs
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -947,7 +947,9 @@ lazy val std = crossProject(JSPlatform, JVMPlatform, NativePlatform)
       ProblemFilters.exclude[MissingClassProblem]("cats.effect.std.Queue$UnsafeUnbounded$Cell"),
       // introduced by #3480
       // adds method to sealed Hotswap
-      ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.effect.std.Hotswap.get")
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.effect.std.Hotswap.get"),
+      ProblemFilters.exclude[DirectMissingMethodProblem](
+        "cats.effect.std.Supervisor.apply$default$2")
```
armanbilge: This filter is highly unfortunate 🤔 if we ever add default parameters to the public API in the future then it will mask bincompat issues there.

Is it possible that this filter is needed only for Scala 3?

armanbilge (commented):
Thanks! This was on my mind but I was too lazy to go back to fix it 😛

build.sbt
danicheg: Yep, it makes sense only for Scala3, I will fix it

danicheg: Also, sorry for stealing your fruit 😄

danicheg:
Smth strange is going on with CI 😞 @armanbilge could you please trigger CI run once again?

danicheg:
Argh, I've done smth stupid. Sorry!

armanbilge (approved):
Sorry that was a bit annoying, thank you! 😁

PR #3670, opened 2023-06-06
Catch exceptions in IOCont

durban:


armanbilge (approved):
oops, I'd wondered about this too :) nice catch!

djspiewak (approved):
Good catch! (no pun intended)

PR #3676, opened 2023-06-09
Draft: Laws testing for `Queue` instances

diogocanut:
Work in progress!

Adressing #1618


tests/shared/src/test/scala/cats/effect/std/DequeueLawsSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+
+  def genDequeue[A: Arbitrary](implicit ticker: Ticker): Gen[Dequeue[IO, A]] =
+    for {
+      list <- Arbitrary.arbitrary[List[A]]
+      queue = fromList(list)
+      outcome = unsafeRun(queue) match {
+        case Outcome.Succeeded(a) => a
+        case _ => None
+      }
+    } yield outcome.get
```
diogocanut, line 38: Need to refactor this code.

tests/shared/src/test/scala/cats/effect/std/QueueLawsSpec.scala
```diff
@@ -0,0 +1,83 @@
...
+  implicit def arbQueue[A: Arbitrary](implicit ticker: Ticker): Arbitrary[Queue[IO, A]] =
+    Arbitrary(genQueue)
+
+  {
+    implicit val ticker = Ticker()
+    checkAll("QueueInvariantLaws", InvariantTests[Queue[IO, *]].invariant[Int, Int, String])
+  }
+
+  // implicit def eqForQueueSink[A: Eq](implicit ticker: Ticker): Eq[QueueSink[IO, A]] =
+  //   Eq.by(toListSink)
```
diogocanut, line 70: I was not able to transform `Sink` implementation into a `List` since it only has `offer` methods.

PR #3678, opened 2023-06-12
Directly get secure random data to array on native

armanbilge:
Now that we can directly access the pointer backing an array, we no longer need a buffer.


durban:
Out of curiosity: what does it mean "now that we can"? I found [this PR](https://github.com/scala-native/scala-native/pull/234) which seems to add the `at` method, but that is quite old...

armanbilge:
I'm not sure what that PR is. But this wasn't exposed as a public method on `Array` until this PR:
- https://github.com/scala-native/scala-native/pull/2888

djspiewak (approved):

Issue #3544, opened 2023-04-19
Avoid per-iteration read barriers in worker loop


armanbilge:
Spinning out of https://github.com/typelevel/cats-effect/pull/3499#discussion_r1170622173. These are the ones I'm aware of:

https://github.com/typelevel/cats-effect/blob/ab5d56c6a6cb2ef7484e5ecb881ecf6b85ff61dc/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala#L419

https://github.com/typelevel/cats-effect/blob/ab5d56c6a6cb2ef7484e5ecb881ecf6b85ff61dc/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala#L689-L694

djspiewak:
Just to preserve some context…

There are two things here that are both independently challenging. The first thing that is challenging is the `done.get()` call, which is hard to fake around. We use this to detect when the pool is shutting down, so we should murderize the worker threads. In this case, the best we can do might end up being something a bit like what we do for `canceled` in the fiber.

The second thing here is the `sleepers` polling. This *has* to be a memory barrier in the present encoding because we allow theft of timers by other workers, meaning that we need to be atomic with respect to that theft. This is a little different than how we handle it in the work queue, because the work queue is stolen from one end while we dequeue from the other, which in turn means that we can play games with the memory barriers. When we poll and steal from the same side, there's just not a lot that can be done: we have to cross a barrier to ensure atomicity.

armanbilge:
If I may suggest the dumb thing: why don't we just check both at a lesser granularity? e.g. every 64 iterations, instead of every iteration.

djspiewak:
> If I may suggest the dumb thing: why don't we just check both at a lesser granularity? e.g. every 64 iterations, instead of every iteration.

This is definitely where we should start. It has some slightly subtle implications for `done.get()` but nothing that should be problematic, I think.

armanbilge:
> When we poll and steal from the same side, there's just not a lot that can be done: we have to cross a barrier to ensure atomicity.

[...]

PR #3690, opened 2023-06-13
Cache `now` in `WorkerThread` in states 4-63

### LINKED ISSUE #3677: Possible performance regression in 3.5.0
There's some discussion regarding the possible performance regression between 3.4.x and 3.5.0 found by Flavio Brasil. More information (including flame charts) can be found in his tweets here and replies in threads underneath it:

https://twitter.com/fbrasisil/status/1667798762788646912

Is it a regression or is it a part of the plan for I/O polling?

djspiewak:
I still want to see numbers which demonstrate the meaning of these types of things in realistic scenarios, but here's a quick draft that takes inspiration from libuv and minimizes the syscall. In particular, this caches the value of `nanoTime()` for states 4-63 (so, most of the time) and only refreshes it when either stealing, polling the external queue, or when the user evaluates `monotonic`. This does trade off timer granularity a bit, but it's still strictly better than what we had pre-3.5 (where timers would reenter through the external queue).

Fixes #3677

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/SchedulerCompanionPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/SchedulerCompanionPlatform.scala
@@ -52,6 +52,15 @@ private[unsafe] abstract class SchedulerCompanionPlatform { this: Scheduler.type
         now.getEpochSecond * 1000000 + now.getLong(ChronoField.MICRO_OF_SECOND)
       }
 
-      def monotonicNanos() = System.nanoTime()
+      def monotonicNanos() = {
+        val back = System.nanoTime()
+
+        val thread = Thread.currentThread()
+        if (thread.isInstanceOf[WorkerThread]) {
+          thread.asInstanceOf[WorkerThread].now = back
+        }
+
+        back
+      }
     }
 }
--- a/core/jvm/src/main/scala/cats/effect/unsafe/SchedulerCompanionPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/SchedulerCompanionPlatform.scala
@@ -52,15 +52,6 @@ private[unsafe] abstract class SchedulerCompanionPlatform { this: Scheduler.type
         now.getEpochSecond * 1000000 + now.getLong(ChronoField.MICRO_OF_SECOND)
       }
 
-      def monotonicNanos() = {
-        val back = System.nanoTime()
-
-        val thread = Thread.currentThread()
-        if (thread.isInstanceOf[WorkerThread]) {
-          thread.asInstanceOf[WorkerThread].now = back
-        }
-
-        back
-      }
+      def monotonicNanos() = System.nanoTime()
     }
 }
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
@@ -90,6 +90,12 @@ private final class WorkerThread(
   private[this] var blocking: Boolean = false
 
   /**
+   * Contains the current monotonic time. Updated by the main worker loop as well as the
+   * `Scheduler` functions when accessing system time in userspace.
+   */
+  private[unsafe] var now: Long = System.nanoTime()
+
+  /**
    * Holds a reference to the fiber currently being executed by this worker thread. This field
    * is sometimes published by the `head` and `tail` of the [[LocalQueue]] and sometimes by the
    * `parked` signal of this worker thread. Threads that want to observe this value should read
@@ -340,7 +346,8 @@ private final class WorkerThread(
         // if we find an already expired one, we go
         // immediately to state 4 (local queue stuff):
         val nextTrigger = sleepers.peekFirstTriggerTime()
-        if ((nextTrigger != MIN_VALUE) && (nextTrigger - System.nanoTime() <= 0L)) {
+        now = System.nanoTime()
+        if ((nextTrigger != MIN_VALUE) && (nextTrigger - now <= 0L)) {
           pool.transitionWorkerFromSearching(rnd)
           4
         } else {
@@ -378,7 +385,7 @@ private final class WorkerThread(
           parkLoop()
           false
         } else {
-          val now = System.nanoTime()
+          now = System.nanoTime()
           val nanos = triggerTime - now
 
           if (nanos > 0L) {
@@ -390,7 +397,8 @@ private final class WorkerThread(
             } else {
               if (parked.get()) {
                 // we were either awakened spuriously, or we timed out
-                if (triggerTime - System.nanoTime() <= 0) {
+                now = System.nanoTime()
+                if (triggerTime - now <= 0) {
                   // we timed out
                   if (parked.getAndSet(false)) {
                     pool.doneSleeping()
@@ -530,6 +538,9 @@ private final class WorkerThread(
             }
           }
 
+          // update the current time
+          now = System.nanoTime()
+
           // Transition to executing fibers from the local queue.
           state = 4
 
@@ -596,7 +607,8 @@ private final class WorkerThread(
 
         case 2 =>
           // First try to steal some expired timers:
-          if (pool.stealTimers(System.nanoTime(), rnd)) {
+          now = System.nanoTime()
+          if (pool.stealTimers(now, rnd)) {
             // some stolen timer created new work for us
             pool.transitionWorkerFromSearching(rnd)
             state = 4
@@ -693,7 +705,6 @@ private final class WorkerThread(
 
         case _ =>
           // Call all of our expired timers:
-          val now = System.nanoTime()
           var cont = true
           while (cont) {
             val cb = sleepers.pollFirstIfTriggered(now)
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
@@ -341,12 +341,13 @@ private final class WorkerThread(
         }
       }
 
+      now = System.nanoTime()
+
       if (nextState != 4) {
         // after being unparked, we re-check sleepers;
         // if we find an already expired one, we go
         // immediately to state 4 (local queue stuff):
         val nextTrigger = sleepers.peekFirstTriggerTime()
-        now = System.nanoTime()
         if ((nextTrigger != MIN_VALUE) && (nextTrigger - now <= 0L)) {
           pool.transitionWorkerFromSearching(rnd)
           4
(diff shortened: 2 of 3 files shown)


core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -693,7 +705,6 @@ private final class WorkerThread(
 
         case _ =>
           // Call all of our expired timers:
-          val now = System.nanoTime()
           var cont = true
           while (cont) {
             val cb = sleepers.pollFirstIfTriggered(now)
```
armanbilge, line 719: The optimization discussed in https://github.com/typelevel/cats-effect/issues/3544#issuecomment-1519355866 would become even more meaningful after this change, if `now` may remain constant for several iterations. So crossing a read-barrier on every iteration is pointless 😅

djspiewak: Yeah I thought about that. It's not horribly complicated to do something like that here, it just introduces a bit more state and a bit more branching. I wanted to start small.

core/jvm/src/main/scala/cats/effect/unsafe/SchedulerCompanionPlatform.scala
```diff
@@ -52,6 +52,15 @@ private[unsafe] abstract class SchedulerCompanionPlatform { this: Scheduler.type
...
+      def monotonicNanos() = {
+        val back = System.nanoTime()
+
+        val thread = Thread.currentThread()
+        if (thread.isInstanceOf[WorkerThread]) {
+          thread.asInstanceOf[WorkerThread].now = back
+        }
+
+        back
+      }
```
armanbilge: Hmm, I think you meant to make this change in the WSTP. This one here is effectively dead code :)

https://github.com/typelevel/cats-effect/blob/469e402d84566f82452a366e48863abd9522e2df/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala#L601

djspiewak: Ah you're right! I'll adjust

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -340,7 +346,8 @@ private final class WorkerThread(
         // if we find an already expired one, we go
         // immediately to state 4 (local queue stuff):
         val nextTrigger = sleepers.peekFirstTriggerTime()
-        if ((nextTrigger != MIN_VALUE) && (nextTrigger - System.nanoTime() <= 0L)) {
+        now = System.nanoTime()
```
durban: When we're returning from `park`, it might make sense to update `now` even if we're not going directly to `4`, as we don't know how long we might've been parked. As an alternative, it might make sense to do the update in state `3` (like in `0`). And also maybe in `1`.

durban (commented):
I've left a comment, otherwise looks reasonable. The hard part is (as you say) knowing what is the effect of this in real life...

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
djspiewak: Very much agreed!

djspiewak:
I used a t2.large instance to measure three scenarios: **series/3.4.x**, **series/3.5.x**, and this PR (as of the current **HEAD**). I ran the `WorkStealingThreadPool` benchmarks (`-wi 10 -i 10 -f 1`) as well as the timer drift measurements (so we can measure granularity impacts loaded and unloaded). The results are as follows. Tldr is at the bottom.

## series/3.4.x

### Granularity

```
Warming up...
Measuring unloaded...
Unloaded overhead: 0.1542258647750001
Measuring heavily loaded...
Loaded overhead: 1.5743448329583334
Measuring unloaded 100x...
Unloaded overhead 100x: 0.11572495005833328
Killed
```

### Benchmarks

```
[info] Benchmark                                              (size)   Mode  Cnt    Score   Error    Units
[info] WorkStealingBenchmark.alloc                           1000000  thrpt   10    0.917 ± 0.003  ops/min
[info] WorkStealingBenchmark.manyThreadsSchedulingBenchmark  1000000  thrpt   10    3.862 ± 0.225  ops/min
[info] WorkStealingBenchmark.runnableScheduling              1000000  thrpt   10  249.878 ± 3.349  ops/min
[info] WorkStealingBenchmark.runnableSchedulingScalaGlobal   1000000  thrpt   10  227.841 ± 0.476  ops/min
[info] WorkStealingBenchmark.scheduling                      1000000  thrpt   10    4.423 ± 0.136  ops/min
```

## series/3.5.x

### Granularity

```
Warming up...
Measuring unloaded...
Unloaded overhead: 0.11346367834166671
Measuring heavily loaded...
Loaded overhead: 2.6429889117916665
Measuring unloaded 100x...
Unloaded overhead 100x: 0.1209033832166666
Loaded overhead: 2.912850151016667
```

### Benchmarks

```
[info] Benchmark                                              (size)   Mode  Cnt    Score   Error    Units
[info] WorkStealingBenchmark.alloc                           1000000  thrpt   10    0.858 ± 0.001  ops/min
[info] WorkStealingBenchmark.manyThreadsSchedulingBenchmark  1000000  thrpt   10    0.832 ± 0.013  ops/min
[info] WorkStealingBenchmark.runnableScheduling              1000000  thrpt   10   17.278 ± 0.180  ops/min
[info] WorkStealingBenchmark.runnableSchedulingScalaGlobal   1000000  thrpt   10  225.010 ± 0.537  ops/min
[info] WorkStealingBenchmark.scheduling                      1000000  thrpt   10    0.883 ± 0.006  ops/min
```

## `HEAD` of this PR

### Granularity

```
Warming up...
Measuring unloaded...
Unloaded overhead: 0.11598775258333327
Measuring heavily loaded...
Loaded overhead: 0.5139900563083333
Measuring unloaded 100x...
Unloaded overhead 100x: 0.12293377584999998
Loaded overhead: 0.7911155032249999
```

### Benchmarks

```
[info] Benchmark                                              (size)   Mode  Cnt    Score     Error    Units
[info] WorkStealingBenchmark.alloc                           1000000  thrpt   10    0.922 ±   0.002  ops/min
[info] WorkStealingBenchmark.manyThreadsSchedulingBenchmark  1000000  thrpt   10    1.923 ±   1.738  ops/min
[info] WorkStealingBenchmark.runnableScheduling              1000000  thrpt   10   64.816 ±   0.578  ops/min
[info] WorkStealingBenchmark.runnableSchedulingScalaGlobal   1000000  thrpt   10  135.160 ± 121.636  ops/min
[info] WorkStealingBenchmark.scheduling                      1000000  thrpt   10    2.322 ±   2.148  ops/min
```

## Tldr

This PR reduces scheduling overhead by around 2-3x in most fiber-heavy workloads, and around 3.5-4x in pure `Runnable` workloads. It also *improves* timer granularity in heavily loaded scenarios by about 5x (which is quite a surprise, since we expected a regression here).

Additionally we saw some interesting edge effects in some of the benchmark runs, most clearly seen in the `WorkStealingBenchmark.runnableSchedulingScalaGlobal` benchmark run on **HEAD**, which exhibits a vast standard deviation and a significantly lower mean. The fun thing is that this particular benchmark is *entirely unaffected* by the changes in any of these branches, since it is simply benchmarking `ExecutionContext.global` and bypassing Cats Effect altogether! In other words, we're measuring contention within the AWS EC2 hypervisor layer. I'm almost entirely certain that we wouldn't see these types of effects if I had used an EC2 metal instance, but I'm not paying that just to validate a hypothesis. :-P

**tldr** This PR improves everything pretty dramatically (at least so far as synthetic benchmarks go; jury's still out on real world stuff). We can make things even better, but I think we should absolutely merge and release this once I have addressed @durban's comment.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
djspiewak: Turns out we were already updating `now` on *most* exits from `park`, though oddly not on natural exits from `parkNanos` (which is the most important one!). Fixed.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -147,9 +153,13 @@ private final class WorkerThread(
   }
 
   def sleep(delay: FiniteDuration, callback: Right[Nothing, Unit] => Unit): Runnable = {
+    // take the opportunity to update the current time, just in case other timers can benefit
+    val _now = System.nanoTime()
```
durban, line 157: Out of curiosity: here you introduce a local `_now`, probably to avoid reading back `now`, but in other places (e.g., state `2`) don't do that. What is the reason for that?

durban (approved):

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
djspiewak: Because I didn't want to write `now = this.now` and then justify it in the inevitable PR review. 😛

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -378,7 +390,7 @@ private final class WorkerThread(
           parkLoop()
           false
```
armanbilge, line 391: Sorry if I missed it. When does `now` get updated after this park?

durban: In `park`, line 348.

durban:
I think there is something wrong with CI: some cirrus jobs didn't get scheduled(?) for more than 12 hours...

armanbilge:
> I think there is something wrong with CI

Cirrus CI is using pre-emptible VMs to be cost-efficient. So under demand machines may not be available.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
armanbilge: Got it, thanks!

djspiewak:
> So under demand machines may not be available.

This is one concern I have with expanding things to fs2, unless it's more of a global Cirrus demand and not github org-specific?

armanbilge:
> unless it's more of a global Cirrus demand

It's this one I'm afraid.

Edit: to clarify, besides global demand there are [some other limits](https://cirrus-ci.org/faq/#are-there-any-limits) for OSS. But in FS2 I have only added four very short-running jobs, it is nowhere comparable to the CE matrix.

wjoel:
Baseline (@armanbilge's 3.6 pre-release): https://www.techempower.com/benchmarks/#section=test&shareid=59f51434-e97e-4b76-91bb-78ba913a676e&test=plaintext
This PR (but rebased on `series/3.x` to include the JVM polling): https://www.techempower.com/benchmarks/#section=test&shareid=8e948333-cd1d-4e35-b03b-50d91b9b1491&test=plaintext

tl;dr 15% more reqs/sec on plaintext, small improvement (1-2%) on JSON.

armanbilge (approved):
Really nice stuff here 😃

PR #3693, opened 2023-06-15
Fix `queueMicrotask` polyfill

armanbilge:
Dang dynamic languages! Don't think we can test this without a really old Node.js or a crappy browser. Maybe if we could rig up a raw V8, but meh.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve

h/t @jilen

### DIFF
--- a/core/js/src/main/scala/cats/effect/unsafe/BatchingMacrotaskExecutor.scala
+++ b/core/js/src/main/scala/cats/effect/unsafe/BatchingMacrotaskExecutor.scala
@@ -46,7 +46,7 @@ private[effect] final class BatchingMacrotaskExecutor(
     if (js.typeOf(js.Dynamic.global.queueMicrotask) == "function")
       js.Dynamic.global.queueMicrotask.asInstanceOf[js.Function1[js.Function0[Any], Any]]
     else {
-      val resolved = js.Dynamic.global.Promise.resolved(())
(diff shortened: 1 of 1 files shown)


core/js/src/main/scala/cats/effect/unsafe/BatchingMacrotaskExecutor.scala
```diff
@@ -46,7 +46,7 @@ private[effect] final class BatchingMacrotaskExecutor(
     if (js.typeOf(js.Dynamic.global.queueMicrotask) == "function")
       js.Dynamic.global.queueMicrotask.asInstanceOf[js.Function1[js.Function0[Any], Any]]
     else {
-      val resolved = js.Dynamic.global.Promise.resolved(())
+      val resolved = js.Dynamic.global.Promise.resolve(())
```
durban, line 49: Nitpicking: `val resolved` could be `val resolve` instead.

durban (approved):

core/js/src/main/scala/cats/effect/unsafe/BatchingMacrotaskExecutor.scala
armanbilge: I was going for verb vs adjective: "I will resolve the promise, then I will have a resolved promise".

Issue #3662, opened 2023-06-05
Multi-threading on Native


armanbilge:
Scala Native 0.5 will support multithreading 🎉 in fact it has already been implemented in a series of PRs starting from https://github.com/scala-native/scala-native/pull/3114 and there are reports of early success in https://github.com/scala-native/scala-native/pull/3286#issuecomment-1561442170.

This is a super-issue to track and discuss efforts to support multi-threading on Cats Effect Native. The primary goal is to move the `WorkStealingThreadPool` from `jvm`-only to `jvm-native` shared sources. If we get this right, I believe the rest of the Typelevel ecosystem should Just Work:tm: on multi-threaded native without requiring any source changes.

### Questions

1. Should we continue to have dedicated support for single-threading? By "dedicated support" I mean retaining the event-loop execution context and using link-time switches to fallback to efficient, non-thread safe data structures.

    On the one hand, a true single-threaded runtime is a myth. Blocking threads are always going to be needed for file I/O and typically DNS as well.

    On the other hand, APIs such as libuv (and to a lesser extent io_uring) can take responsibility for the blocking thread pool, so that userland can be single-threaded. It's quite possible that this will have higher performance than managing our own blocking threads, but considering our own optimizations e.g. to avoid thread-shifting blocking tasks, I'm not sure if this is so cleancut.

    So at this stage I'm not sure if it is worth the added complexity. Without a libuv polling system there's no viable single-threaded Cats Effect application without a blocking pool.

    Afterthought: a single-threaded runtime may be required to support WASM 🤔 

### Tasks

- [ ] Implement JDK concurrent data structures used in Cats Effect. @mox692 is working on this 🙏 
    https://github.com/scala-native/scala-native/issues/3165

- [ ] Move WSTP into shared `jvm-native` sources. Some portions of it are implemented in Java so these will have to be ported. For example some padding tricks are used in the `LocalQueue` to avoid "false-sharing". We need to figure out how to achieve the same on Native.

- [ ] Upgrade various APIs to use concurrency primitives. For example `Ref`, which is currently using an ordinary mutable field. We also need to fixup the implementations of `IO.blocking` 😅 

- [ ] Add support for multi-threading to the polling systems. Notably, interruption and thread-safe cancelation.
    https://github.com/typelevel/cats-effect/pull/3332

- [ ] Audit test suite for disabled or otherwise special-cased tests that now should be run on Native.

[...]

PR #3695, opened 2023-06-16
Native signal handling and safe `IOApp` cancelation

armanbilge:
1. Refactor `IOAppSpec` so that using it to test 3 platforms is less insane

2. Setup signal handling infrastructure with file descriptor polling and pipes

3. Add handlers/listeners to enable safe `IOApp` cancelation ~~and fiber dumps~~


armanbilge:
Still diagnosing the build failures. I wonder if weak references have some issues on Scala Native so enabling fiber monitoring has compromised CI ...

Update:
- https://github.com/scala-native/scala-native/issues/3341
- https://github.com/scala-native/scala-native/pull/3347

build.sbt
```diff
@@ -876,27 +884,52 @@ lazy val tests: CrossProject = crossProject(JSPlatform, JVMPlatform, NativePlatf
...
+      Test / unmanagedSourceDirectories += (LocalRootProject / baseDirectory).value / "ioapp-tests" / "src" / "test" / "scala",
+      libraryDependencies += "org.specs2" %%% "specs2-core" % Specs2Version % Test,
+      buildInfoPackage := "cats.effect",
+      buildInfoKeys ++= Seq(
+        "jsRunner" -> (tests.js / Compile / fastOptJS / artifactPath).value,
+        "nativeRunner" -> (tests.native / Compile / nativeLink / artifactPath).value
+      )
+    )
+
+lazy val ioAppTestsJVM =
```
djspiewak, line 911: Can't we shortcut a lot of this by just using `crossProject`?

core/native/src/main/scala/cats/effect/Signal.scala
```diff
@@ -0,0 +1,133 @@
...
+    action.sa_handler = handler
+    sigaddset(action.at2, 13) // mask SIGPIPE
+    if (sigaction(signum, action, null) != 0)
+      throw new IOException(fromCString(strerror(errno)))
+  }
+
+  private[this] final val SIGINT = 2
+  private[this] final val SIGTERM = 15
+  private[this] final val SIGUSR1Linux = 10
+  private[this] final val SIGUSR1Mac = 30
```
djspiewak: Wouldn't it be better to do the `isMac` branching here so we could have a single `SIGUSR1` value?

core/native/src/main/scala/cats/effect/unsafe/EventLoopExecutorScheduler.scala
```diff
@@ -152,6 +152,15 @@ private[effect] final class EventLoopExecutorScheduler[P](
 
   def shutdown(): Unit = system.close()
 
+  def liveTraces(): Map[IOFiber[_], Trace] = {
```
djspiewak, line 157: This reminds me that one of the important cleanup tasks for our first "Scala Native with real threads" release will need to be making sure that we've deduplicated as much as possible/relevant between JVM and Native.

core/native/src/main/scala/cats/effect/unsafe/FiberMonitor.scala
```diff
@@ -0,0 +1,112 @@
...
+ * limitations under the License.
+ */
+
+package cats.effect
+package unsafe
+
+import scala.collection.mutable
+import scala.concurrent.ExecutionContext
+
+private[effect] sealed abstract class FiberMonitor extends FiberMonitorShared {
```
djspiewak: So in the category of cross-build confusion, this is now a *third* paradigm for cross-build branching. We have `isNative`/`isJVM`/etc, we have the `*Platform` variants, and now we have `*Shared` variants. Can we at least not have that third one? ;-) Would it be possible to reorganize this to have `FiberMonitorPlatform` instead?

djspiewak (changes requested):

build.sbt
armanbilge: We can't. All three of these are actually JVM projects. They just happens to be JVM projects that invoke JVMs, Node.js, or Native binaries.

core/native/src/main/scala/cats/effect/Signal.scala
armanbilge: Then it's not an inlinable constant. `final val` with a literal is treated specially by the compiler.

Ok yeah, that's a dumb reason 😂

core/native/src/main/scala/cats/effect/unsafe/FiberMonitor.scala
armanbilge: Not really, no. The issue is that the _constructors_ are different, because they either take a WSTP, a BMTE, or an ELES. You can't platform a constructor.

Unless we pull out some common fiber-monitorable EC interface. Then they can all share a common constructor.

core/native/src/main/scala/cats/effect/unsafe/EventLoopExecutorScheduler.scala
armanbilge: I added it to the checklist in:
- https://github.com/typelevel/cats-effect/issues/3662#issue-1741007577

core/native/src/main/scala/cats/effect/Signal.scala
djspiewak: LOL. Yeah I think in this case it doesn't matter so much.

core/native/src/main/scala/cats/effect/unsafe/FiberMonitor.scala
djspiewak: It might be worth pulling out the interface, if only for internal use. I feel like we need to be really relentless about trying to simplify and ensure consistency of our platform-split code, given how much of it we have.

armanbilge: I was able to make some small improvement: we can share an interface between JS and Native because they are similar, but the WSTP is too different to participate in a shared abstraction (e.g. it has a notion of thread-local weakbags for fibers).

djspiewak (dismissed):

djspiewak:
Ready to go once deconflicted

PR #3701, opened 2023-06-22
Dispatcher: remove `alive` in favour of nullifying queues

samspills:
This idea came from conversations with Arman!

If workers close their own state queues as they're marked done, then we can remove the double check on alive when submitting work. If a task is submitted to a queue successfully it _will_ be seen by that worker. Or if the worker has already closed the queue, then submitting the task will fail entirely. If workers are controlling their own queues in this way, I'm not convinced we need the alive ref at all! So I've gone so far as to remove it entirely.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
@@ -258,12 +258,10 @@ object Dispatcher {
               var i = 0
               while (i < workers) {
                 val st = state(i)
-                if ((st.get() ne Nil) & (st.get() ne null)) {
+                if (st.get() ne null) {
                   val list = if (done) st.getAndSet(null) else st.getAndSet(Nil)
-                  if (list ne null) {
+                  if ((list ne null) && (list ne Nil)) {
                     buffer ++= list.reverse // FIFO order here is a form of fairness
-                  } else {
-                    println("SRP :: st.getAndSet returned null") // TODO
                   }
                 }
                 i += 1
--- a/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
@@ -247,16 +247,24 @@ object Dispatcher {
       supervisor <- Supervisor[F](await, Some((_: Outcome[F, Throwable, _]) => true))
 
       _ <- {
-        def step(state: Array[AtomicReference[List[Registration]]], await: F[Unit]): F[Unit] =
+        def step(
+            state: Array[AtomicReference[List[Registration]]],
+            await: F[Unit],
+            doneR: AtomicBoolean): F[Unit] =
           for {
+            done <- F.delay(doneR.get())
             regs <- F delay {
               val buffer = mutable.ListBuffer.empty[Registration]
               var i = 0
               while (i < workers) {
                 val st = state(i)
-                if (st.get() ne Nil) {
-                  val list = st.getAndSet(Nil)
-                  buffer ++= list.reverse // FIFO order here is a form of fairness
+                if ((st.get() ne Nil) & (st.get() ne null)) {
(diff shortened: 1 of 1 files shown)


std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -247,16 +247,22 @@ object Dispatcher {
       supervisor <- Supervisor[F](await, Some((_: Outcome[F, Throwable, _]) => true))
 
       _ <- {
-        def step(state: Array[AtomicReference[List[Registration]]], await: F[Unit]): F[Unit] =
+        def step(
+            state: Array[AtomicReference[List[Registration]]],
+            await: F[Unit],
+            doneR: AtomicBoolean): F[Unit] =
           for {
+            done <- F.delay(doneR.get())
```
armanbilge, line 255: I wonder, does this need to be an `AtomicBoolean`, what if we just directly pass a `done: Boolean` parameter? The final invocation of `step` can set it to `true`.

armanbilge: Oh maybe not ... 🤔

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -247,16 +247,22 @@ object Dispatcher {
...
               val buffer = mutable.ListBuffer.empty[Registration]
               var i = 0
               while (i < workers) {
                 val st = state(i)
-                if (st.get() ne Nil) {
-                  val list = st.getAndSet(Nil)
-                  buffer ++= list.reverse // FIFO order here is a form of fairness
+                if (st.get() ne null) {
+                  val list = if (done) st.getAndSet(null) else st.getAndSet(Nil)
+                  if ((list ne null) && (list ne Nil)) {
```
armanbilge, line 263: ... because looking at this here, I realize there still must be a race condition. If I remember correctly, there is still a race between the `step`s running on the worker fiber, and the final `step` taken when the `Dispatcher` closes. I guess it's for another PR, but I think getting rid of that race condition would also help with reasoning about this implementation.

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -361,65 +363,57 @@ object Dispatcher {
...
+                  promise.future
+                case CancelToken(cancelToken) =>
+                  cancelToken()
               }
-
-              loop()
             }
 
-            // double-check after we already put things in the structure
-            if (alive.get()) {
```
armanbilge, line 413: 💯

armanbilge (commented):
Looks great! I got carried away with some more thoughts but that can be another PR ...

I forgot, did we have a test for this? I have a vague recollection we sort of discovered this issue when investigating a test failure. So it would be good to add/restore that test, whatever it was 😅

djspiewak (approved):
Agreed that this definitely improves things. Thank you!

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
samspills: Yes you're exactly correct! I agree about trying to remove that race, but I recall that things got tricky with that so it will likely take more staring and more thinking 😆

samspills:
> I forgot, did we have a test for this? I have a vague recollection we sort of discovered this issue when investigating a test failure. So it would be good to add/restore that test, whatever it was 😅

@armanbilge I will have to re-remember things, but I think this came up while we were trying to diagnose the flakiness in the `issue 3501: reject new tasks after release action is submitted as a task` test. We fixed the flakiness at the time by adding an extra sleep in there. I also wonder if this change means we can simplify the `reject new tasks while shutting down` test? I'll look at both of them and follow up :D

armanbilge:
Thanks! Yes, perhaps related to this issue.
- https://github.com/typelevel/cats-effect/issues/3501

PR #3724, opened 2023-06-29
Optimize `liveTraces()` on JS

armanbilge:
This was a silly one: we already have a queue of fibers, so it's unnecessary to add/remove them from a separate fiber bag for monitoring. We can just directly iterate over the queue.

### DIFF
--- a/tests/js/src/test/scala/cats/effect/unsafe/JSArrayQueueSpec.scala
+++ b/tests/js/src/test/scala/cats/effect/unsafe/JSArrayQueueSpec.scala
@@ -69,7 +69,7 @@ class JSArrayQueueSpec extends BaseSpec with ScalaCheck {
             }
         }
 
-        true must beTrue
+        ok
       }
     }
   }
--- a/tests/js/src/test/scala/cats/effect/unsafe/JSArrayQueueSpec.scala
+++ b/tests/js/src/test/scala/cats/effect/unsafe/JSArrayQueueSpec.scala
@@ -20,7 +20,7 @@ package unsafe
 import org.scalacheck.Prop.forAll
 import org.specs2.ScalaCheck
 
-import scala.collection.mutable.ListBuffer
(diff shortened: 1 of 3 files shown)


tests/js/src/test/scala/cats/effect/unsafe/JSArrayQueueSpec.scala
```diff
@@ -41,6 +41,37 @@ class JSArrayQueueSpec extends BaseSpec with ScalaCheck {
         taken.toList must beEqualTo(stuff.flatten)
       }
     }
+
+    "iterate over contents in foreach" in {
+      forAll { (stuff: List[Option[Int]]) =>
+        val queue = new JSArrayQueue[Int]
+        val shadow = new Queue[Int]
```
durban, line 48: It's good to see I'm not the only one who tends to call these things "shadows" :-) (Although I don't remember where I've got the idea from.)

tests/js/src/test/scala/cats/effect/unsafe/JSArrayQueueSpec.scala
```diff
@@ -41,6 +41,37 @@ class JSArrayQueueSpec extends BaseSpec with ScalaCheck {
...
+          case None =>
+            if (!shadow.isEmpty) {
+              val got = queue.take()
+              val expected = shadow.dequeue()
+              got must beEqualTo(expected)
+              checkContents()
+            }
+        }
+
+        true must beTrue
```
durban: Isn't this the same as `ok`?

durban (dismissed):

tests/js/src/test/scala/cats/effect/unsafe/JSArrayQueueSpec.scala
armanbilge: oh lol good call
```suggestion
        ok
```

durban (approved):

PR #3728, opened 2023-06-30
Allow that the renamed blocker thread is terminated

### LINKED ISSUE #3704: `WorkerThreadNameSpec` seems a bit flaky in CI
```
[info] WorkerThread should
[error]   x rename itself when entering and exiting blocking region
[error]    blocker thread not found after reset (WorkerThreadNameSpec.scala:84)
```

aeons:
In WorkerThreadNameSpec, if the thread is readded to the compute pool, we still check that it has been renamed and has the correct index, but if the thread has been terminated in the mean time, that is ok as well.

Closes #3704

### DIFF
--- a/tests/jvm/src/test/scala/cats/effect/unsafe/WorkerThreadNameSpec.scala
+++ b/tests/jvm/src/test/scala/cats/effect/unsafe/WorkerThreadNameSpec.scala
@@ -20,6 +20,8 @@ import cats.effect.{BaseSpec, IO}
 import cats.effect.testkit.TestInstances
 import cats.syntax.all._
 
+import scala.concurrent.duration._
+
 class WorkerThreadNameSpec extends BaseSpec with TestInstances {
 
   override def runtime(): IORuntime = {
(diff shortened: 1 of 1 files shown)


armanbilge:
Thanks! Can we actually apply this patch to the 3.5.x branch? Then we'll merge it up into 3.x. Hopefully that should avoid it being clobbered by a merge again...

aeons:
Changed branch and cherrypicked.

durban (dismissed):
It is not clear to me how could the blocker thread exit: the `runtimeBlockingExpiration` is 1 minute by default. Could this test take such a long time?

armanbilge:
> the `runtimeBlockingExpiration` is 1 minute by default. Could this test take such a long time?

That's an interesting point. Since we are making a custom `IORuntime` for this test anyway, what if we just increased that expiration time? I can't remember if the configuration is exposed or hard-coded.

durban:
Yeah, that's exactly what I was thinking about. It's exposed here: https://github.com/typelevel/cats-effect/blob/series/3.5.x/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala#L57

aeons:
Switched it to expire threads after 10 minutes. Let's see if this fixes it.

armanbilge (approved):
Thanks!

PR #3730, opened 2023-07-02
ResourceApp extends IOApp (breaking change)

gvolpe:
## Reason

I would like `ResourceApp` to have access to all the runtime configurations available in `IOApp`, such as `computeWorkerThreadCount`, `reportFailure` and `runtimeConfig`, which are currently not exposed.

This is mainly to start the conversation more than expecting this PR to be merged, as this is a *breaking change* (`run` is now named `resource` in `ResourceApp`).

I am not sure if this can be done in any better way, perhaps a package-private constraint instead of `protected`?

### DIFF
--- a/core/shared/src/main/scala/cats/effect/ResourceApp.scala
+++ b/core/shared/src/main/scala/cats/effect/ResourceApp.scala
@@ -98,7 +98,8 @@ object ResourceApp {
      */
     def resource: Resource[IO, Unit]
 
-    final def resource(args: List[String]): Resource[IO, ExitCode] = resource.as(ExitCode.Success)
+    final def resource(args: List[String]): Resource[IO, ExitCode] =
+      resource.as(ExitCode.Success)
   }
 
   /**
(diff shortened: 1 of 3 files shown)


armanbilge:
> This is mainly to start the conversation

Thanks! Well, in the spirit of conversation: why not just use `IOApp`, especially if you need this much customization?

I'm genuinely curious when/why folks prefer to use `ResourceApp`. Personally, I'm not crazy about `ResourceApp`, I think it's better to use `IOApp` with the appropriate combinators on the `Resource`.

gvolpe:
> why not just use IOApp

We have our own `ForeverApp extends IOApp` basically, but we're having to do this over and over, thought I'd bring it up here  :) 

> I'm genuinely curious when/why folks prefer to use ResourceApp.

All our apps are built as a sequence of `Resource`s, for which it would be nice to leverage `ResourceApp`. Doesn't seem very useful by hiding all the runtime goodies from `IOApp`.

Using `IOApp` with resources is inconvenient because for-comprehensions end up having to be wrapped in `()`, which I find aesthetically unpleasant 😄 Best thing is to split it up in two different functions/methods, and then have `def run = resource.useForever`.

PR #3740, opened 2023-07-09
update `Deferred#complete` type

### LINKED ISSUE #3720: Incorrect signature in documentation
The signature of `complete` is incorrect in the docs here: https://typelevel.org/cats-effect/docs/thread-model#fiber-blocking-previously-semantic-blocking (it returns `F[Boolean]`).

ag91:
Fixes https://github.com/typelevel/cats-effect/issues/3720

### DIFF
--- a/docs/thread-model.md
+++ b/docs/thread-model.md
@@ -121,7 +121,7 @@ purely functional promise that can only be completed once.
 trait Deferred[F[_], A] {
   def get: F[A]
 
-  def complete(a: A): F[Unit]
+  def complete(a: A): F[Boolean]
 }
 ```
 


durban:
Thanks! I think this should go into`series/3.5.x` instead. Could you retarget the PR?

ag91:
just rebasing now, I will ping when this is ready

ag91:
right I did rebase. Note though: also the 3.x series has the problem in the docs. 
(I am assuming you suggested to move the change to the 3.5.x because it is more pragmatic to be correct on the most recent version: did I get you right?)

armanbilge:
Yes, that's right. We fix in 3.5.x because that's the current published series, and we regularly merge changes from 3.5.x into 3.x.

Further reading:
https://github.com/typelevel/cats-effect/blob/series/3.x/CONTRIBUTING.md#process

durban (approved):
Thanks!

PR #3748, opened 2023-07-13
Make useForever polymorphic rather than speciailising it to Nothing

george-wilson-rea:
I'd prefer if I could pick a type for this, eg. `ExitCode`.

I write apps like this:

```
class MyApp extends IOApp {
  def run(args: Args): IO[ExitCode] =
    for {
      config <- readConfig
      otherStuff <- setUpOtherStuff(config)
      ec <- httpServer.resource.useForever
    } yield ec
}
```

and with the `Nothing` type, the Scala compiler warns me about dead code, but if I use `httpServer.resource.use(_ => IO.never[ExitCode])` then it works fine. Hence, I propose changing `useForever` to work like this.

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -220,8 +220,8 @@ sealed abstract class Resource[F[_], +A] extends Serializable {
    *
    * The finalisers run when the resulting program fails or gets interrupted.
    */
-  def useForever(implicit F: Spawn[F]): F[Nothing] =
-    use[Nothing](_ => F.never)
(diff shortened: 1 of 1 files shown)


durban:
I might be missing something obvious, but I tried your example code, and this change did not make the warning disappear: the compiler still inferred `ec: Nothing` . (On the other hand, using https://github.com/oleg-py/better-monadic-for _does_ make the warning go away, probably because it removes the final `map` call.)

djspiewak:
I concur with @durban: I'm not sure this really helps. The type will still be inferred to `Nothing` in unascribed cases, which in turn recreates the problem. There are also a number of cases that I can't quite remember (I think only on Scala 2.12?) where the `Nothing` encoding actually works a lot better than `[A] A`, even though they should be equivalent.

PR #3749, opened 2023-07-14
Dispatcher: test cleanup

samspills:
Just removing the extra sleep that was added here: https://github.com/typelevel/cats-effect/pull/3593. The sleep was added as part of a short-term fix for flakiness and now that the second check on alive (and possible cancellation of the submitted task) is gone, this test shouldn't flake. 

I added a `.replicateA(20000)` on line 366 to convince myself that this was stable 🤞🏻

### DIFF
--- a/tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
@@ -360,7 +360,7 @@ class DispatcherSpec extends BaseSpec with DetectPlatform {
       implicit ticker =>
         val test = dispatcher.allocated.flatMap {
           case (runner, release) =>
-            IO(runner.unsafeRunAndForget(IO.sleep(50.millis) *> release)) *>
+            IO(runner.unsafeRunAndForget(release)) *>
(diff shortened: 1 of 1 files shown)


tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
```diff
@@ -360,7 +360,7 @@ class DispatcherSpec extends BaseSpec with DetectPlatform {
       implicit ticker =>
         val test = dispatcher.allocated.flatMap {
           case (runner, release) =>
-            IO(runner.unsafeRunAndForget(IO.sleep(50.millis) *> release)) *>
+            IO(runner.unsafeRunAndForget(release)) *>
```
armanbilge, line 363: I was briefly confused why we added this sleep in the first place, for a `ticked` test ... and it's because in https://github.com/typelevel/cats-effect/pull/3593 it was originally a `real` test and the sleep helped avoid the race condition where the `release` started while the `unsafeRunAndForget` was still executing.

tl;dr this test is a lot more robust now.

armanbilge (approved):
Thanks for following-up!

PR #3750, opened 2023-07-14
Add note to starvation docs re: cpus in k8s

Daenyth:
This is a common "gotcha" when running in k8s.
Ideally we could automatically detect this environment in the CE runtime itself on initialization and warn the user proactively, but at least this gives people something to ctrl-f for in the doc

### DIFF
--- a/docs/core/starvation-and-tuning.md
+++ b/docs/core/starvation-and-tuning.md
@@ -350,7 +350,7 @@ Of course, it's never as simple as doubling the number of vCPUs and halving the 
 
 #### Not Enough Threads - Running in Kubernetes
 
-One cause of "not enough threads" can be that the application is running inside kubernetes with a cpu_quota not configured. When the cpu limit is not configured, the jvm detects the number of available processors as 1, which will severely restrict what the runtime is able to do.
+One cause of "not enough threads" can be that the application is running inside kubernetes with a `cpu_quota` not configured. When the cpu limit is not configured, the JVM detects the number of available processors as 1, which will severely restrict what the runtime is able to do.
 
 This guide on [containerizing java applications for kubernetes](https://learn.microsoft.com/en-us/azure/developer/java/containers/kubernetes#understand-jvm-available-processors) goes into more detail on the mechanism involved.
 
--- a/docs/core/starvation-and-tuning.md
+++ b/docs/core/starvation-and-tuning.md
@@ -442,7 +442,7 @@ If this is not possible, the next-best approach is to rely on a circular buffer.
 
 > This scenario is specific to Kubernetes, Amazon ECS, and similar resource-controlled containerized deployments.
 
-In many Docker clusters, it is relatively standard practice to over-provision CPU credits by some factor (even 100% over-provisioning is quite common). What this effectively means is that the container environment will promise (or at least *allow*) *m* applications access to *n* vCPUs each, despite only *(m* n) / (1 + k)*CPUs being physically present across the underlying hardware cluster. In this equation, *k* is the over-provisioning factor, often referred to as "burst credits".
+In many Docker clusters, it is relatively standard practice to over-provision CPU credits by some factor (even 100% over-provisioning is quite common). What this effectively means is that the container environment will promise (or at least *allow*) *m* applications access to *n* vCPUs each, despite only *(m * n) / (1 + k)* CPUs being physically present across the underlying hardware cluster. In this equation, *k* is the over-provisioning factor, often referred to as "burst credits".
 
 This is a standard strategy because many applications are written in a fashion which is only loosely coupled to the underlying CPU count, to a large degree because many applications are simply not optimized to that extent. Cats Effect *is* optimized to take advantage of the precise number of underlying CPUs, and well-written Cats Effect applications inherit this optimization, meaning that they are inherently much more sensitive to this hardware factor than many other applications.
 
--- a/docs/core/starvation-and-tuning.md
+++ b/docs/core/starvation-and-tuning.md
@@ -6,9 +6,9 @@ title: Starvation and Tuning
 All Cats Effect applications constructed via `IOApp` have an automatic mechanism which periodically checks to see if the application runtime is starving for compute resources. If you ever see warnings which look like the following, they are the result of this mechanism automatically detecting that the responsiveness of your application runtime is below the configured threshold. Note that the timestamp is the time when the starvation was detected, which is not precisely the time when starvation (or the task that is responsible) began.
 
 ```
-2023-01-28T00:16:24.101Z [WARNING] Your app's responsiveness to a new asynchronous 
+2023-01-28T00:16:24.101Z [WARNING] Your app's responsiveness to a new asynchronous
 event (such as a new connection, an upstream response, or a timer) was in excess
-of 40 milliseconds. Your CPU is probably starving. Consider increasing the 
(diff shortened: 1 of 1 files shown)


docs/core/starvation-and-tuning.md
```diff
@@ -434,7 +442,7 @@ If this is not possible, the next-best approach is to rely on a circular buffer.
 
 > This scenario is specific to Kubernetes, Amazon ECS, and similar resource-controlled containerized deployments.
 
-In many Docker clusters, it is relatively standard practice to over-provision CPU credits by some factor (even 100% over-provisioning is quite common). What this effectively means is that the container environment will promise (or at least *allow*) *m* applications access to *n* vCPUs each, despite only *(m * n) / (1 + k)* CPUs being physically present across the underlying hardware cluster. In this equation, *k* is the over-provisioning factor, often referred to as "burst credits".
+In many Docker clusters, it is relatively standard practice to over-provision CPU credits by some factor (even 100% over-provisioning is quite common). What this effectively means is that the container environment will promise (or at least *allow*) *m* applications access to *n* vCPUs each, despite only *(m* n) / (1 + k)*CPUs being physically present across the underlying hardware cluster. In this equation,*k* is the over-provisioning factor, often referred to as "burst credits".
```
samspills: I don't think we want the spaces removed from `*(m * n) / (1 + k)* CPUs`, it looks like the formatting takes over the multiplication

Daenyth: oh drat, yeah I didn't even touch that - it's just the file formatting. I'll fix it

Daenyth:
Before this gets merged I want to add some wording about overriding the thread count in `IOApp` for cases where you really do want your k8s pod to have <=1 cpu

armanbilge:
drive-by-dropping a link to more material we should incorporate

https://discord.com/channels/632277896739946517/839263556754472990/1135955773613146152

samspills:
going to close and reopen b/c I think that's what's needed to get the updated CI workflows

samspills:
nope I appear to have been wrong about that _womp womp_

armanbilge:
> nope I appear to have been wrong about tha

We merged your fix to 3.5.x, but this PR is targeting 3.x. Actually this PR could target 3.5.x.

TimWSpence:
Woops just changing the target branch is not sufficient 🤦‍♂️

docs/core/starvation-and-tuning.md
```diff
@@ -348,6 +348,14 @@ The problem with "going wide" is it restricts the resources available within use
...
 
+#### Not Enough Threads - Running in Kubernetes
+
+One cause of "not enough threads" can be that the application is running inside kubernetes with a cpu_quota not configured. When the cpu limit is not configured, the jvm detects the number of available processors as 1, which will severely restrict what the runtime is able to do.
+
+This guide on [containerizing java applications for kubernetes](https://learn.microsoft.com/en-us/azure/developer/java/containers/kubernetes#understand-jvm-available-processors) goes into more detail on the mechanism involved.
+
+**All cats-effect applications running in kubernetes should have either a cpu_quota configured or use the jvm `-XX:ActiveProcessorCount` argument to explicitly tell the jvm how many cores to use.**
```
djspiewak:
```suggestion
**All Cats Effect applications running in kubernetes should have either a `cpu_quota` configured or use the jvm `-XX:ActiveProcessorCount` argument to explicitly tell the jvm how many cores to use.**
```

docs/core/starvation-and-tuning.md
```diff
@@ -348,6 +348,14 @@ The problem with "going wide" is it restricts the resources available within use
 
 Of course, it's never as simple as doubling the number of vCPUs and halving the number of instances. Scaling is complicated, and you'll likely need to adjust other resources such as memory, connection limits, file handle counts, autoscaling signals, and such. Overall though, a good rule of thumb is to consider 8 vCPUs to be the minimum that should be available to a Cats Effect application at scale. 16 or even 32 vCPUs is likely to improve performance even further, and it is very much worth experimenting with these types of tuning parameters.
 
+#### Not Enough Threads - Running in Kubernetes
+
+One cause of "not enough threads" can be that the application is running inside kubernetes with a cpu_quota not configured. When the cpu limit is not configured, the jvm detects the number of available processors as 1, which will severely restrict what the runtime is able to do.
```
djspiewak:
```suggestion
One cause of "not enough threads" can be that the application is running inside kubernetes with a `cpu_quota` not configured. When the cpu limit is not configured, the JVM detects the number of available processors as 1, which will severely restrict what the runtime is able to do.
```

docs/core/starvation-and-tuning.md
```diff
@@ -434,7 +442,7 @@ If this is not possible, the next-best approach is to rely on a circular buffer.
 
 > This scenario is specific to Kubernetes, Amazon ECS, and similar resource-controlled containerized deployments.
 
-In many Docker clusters, it is relatively standard practice to over-provision CPU credits by some factor (even 100% over-provisioning is quite common). What this effectively means is that the container environment will promise (or at least *allow*) *m* applications access to *n* vCPUs each, despite only *(m * n) / (1 + k)* CPUs being physically present across the underlying hardware cluster. In this equation, *k* is the over-provisioning factor, often referred to as "burst credits".
+In many Docker clusters, it is relatively standard practice to over-provision CPU credits by some factor (even 100% over-provisioning is quite common). What this effectively means is that the container environment will promise (or at least *allow*) *m* applications access to *n* vCPUs each, despite only *(m* n) / (1 + k)*CPUs being physically present across the underlying hardware cluster. In this equation, *k* is the over-provisioning factor, often referred to as "burst credits".
```
djspiewak:
```suggestion
In many Docker clusters, it is relatively standard practice to over-provision CPU credits by some factor (even 100% over-provisioning is quite common). What this effectively means is that the container environment will promise (or at least *allow*) *m* applications access to *n* vCPUs each, despite only *(m * n) / (1 + k)* CPUs being physically present across the underlying hardware cluster. In this equation, *k* is the over-provisioning factor, often referred to as "burst credits".
```

docs/core/starvation-and-tuning.md
```diff
@@ -446,6 +454,16 @@ The solution is to eliminate this over-provisioning. If a scheduled container is
...
 
+#### Kubernetes CPU Pinning
+
+Even if you have followed the above advice and avoided over-provisioning, the linux kernel scheduler is unfortunately not aware of the Cats Effect scheduler and will likely actively work against the Cats Effect scheduler by moving Cats Effect worker threads between different CPUs, thereby destroying CPU cache-locality. In certain environments we can prevent this by configuring Kubernetes to pin an application to a gviven set of CPUs:
```
djspiewak:
```suggestion
Even if you have followed the above advice and avoided over-provisioning, the Linux kernel scheduler is unfortunately not aware of the Cats Effect scheduler and will likely actively work against the Cats Effect scheduler by moving Cats Effect worker threads between different CPUs, thereby destroying CPU cache-locality. In certain environments we can prevent this by configuring Kubernetes to pin an application to a gviven set of CPUs:
```

Is this true? The kernel uses a work-stealing algorithm to balance tasks across compute units, so it has the same long-term convergence as Cats Effect itself: when fully loaded, threads remain where they are and affinity is maximized.

djspiewak (changes requested):

djspiewak (approved):

PR #3752, opened 2023-07-15
Add Documentation to `PollingSystem`

antoniojimeneznieto:
This PR adds documentation to the `PollingSystem` abstract class and its associated methods.

### DIFF
--- a/core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
+++ b/core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
@@ -90,7 +90,7 @@ abstract class PollingSystem {
    *   callback that handles any failures that occur during polling.
    *
    * @return
-   *   whether any events were polled
+   *   whether any events were polled. e.g. if the method returned due to timeout, this should be `false`.
    */
   def poll(poller: Poller, nanos: Long, reportFailure: Throwable => Unit): Boolean
 
--- a/core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
+++ b/core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
@@ -80,7 +80,7 @@ abstract class PollingSystem {
 
   /**
    * @param poller
-   *   the poller used to poll events.
+   *   the thread-local [[Poller]] used to poll events.
    *
    * @param nanos
    *   the maximum duration for which to block, where `nanos == -1` indicates to block
--- a/core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
+++ b/core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
@@ -18,7 +18,7 @@ package cats.effect
 package unsafe
 
 /**
- * Encapsulates the methods required for managing and interacting with a polling system. Polling
+ * Represents a stateful system for managing and interacting with a polling system. Polling
  * systems are typically used in scenarios such as handling multiplexed blocking I/O or other
  * event-driven systems, where one needs to repeatedly check (or "poll") some condition or
  * state, blocking up to some timeout until it is ready.
@@ -30,9 +30,6 @@ package unsafe
  *   - The lifecycle management methods, such as creating and closing the polling system and its
  *     components
  *   - The runtime interaction methods, such as polling events and interrupting the process
- *
- * A concrete implementation of `PollingSystem` should provide specific implementations for all
- * these components and actions.
  */
 abstract class PollingSystem {
 
@@ -54,12 +51,12 @@ abstract class PollingSystem {
   /**
    * Creates a new instance of the user-facing interface.
    *
-   * @param register
+   * @param access
    *   callback to obtain a thread-local `Poller`.
    * @return
    *   an instance of the user-facing interface `Api`.
    */
-  def makeApi(register: (Poller => Unit) => Unit): Api
+  def makeApi(access: (Poller => Unit) => Unit): Api
 
   /**
    * Creates a new instance of the thread-local data structure used for polling.
@@ -70,7 +67,7 @@ abstract class PollingSystem {
   def makePoller(): Poller
 
   /**
-   * Close a specific poller.
+   * Closes a specific poller.
    *
    * @param poller
    *   the poller to be closed.
@@ -89,7 +86,8 @@ abstract class PollingSystem {
    *   callback that handles any failures that occur during polling.
    *
    * @return
-   *   whether any events were polled. e.g. if the method returned due to timeout, this should be `false`.
+   *   whether any events were polled. e.g. if the method returned due to timeout, this should
+   *   be `false`.
    */
   def poll(poller: Poller, nanos: Long, reportFailure: Throwable => Unit): Boolean
 
@@ -100,7 +98,7 @@ abstract class PollingSystem {
   def needsPoll(poller: Poller): Boolean
 
   /**
-   * Interrupt a specific target poller running on a specific target thread.
+   * Interrupts a specific target poller running on a specific target thread.
    *
    * @param targetThread
    *   is the thread where the target poller is running.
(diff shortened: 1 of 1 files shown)


core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
```diff
@@ -17,6 +17,23 @@
 package cats.effect
 package unsafe
 
+/**
+ * Encapsulates the methods required for managing and interacting with a polling system. Polling
```
armanbilge: "encapsulates the methods" sounds like this is a typeclass i.e. just a collection of methods that is otherwise stateless.

Actually, in the current design, an instance of this class _is_ the live polling system i.e. it is stateful, which is why it has a no-arg `close()` method.

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
```diff
@@ -17,6 +17,23 @@
...
+ * systems are typically used in scenarios such as handling asynchronous I/O or other
```
armanbilge: So this is a really interesting one, depending on what precisely we mean by "asynchronous" :)

_Technically_ if you have truly async I/O I would say you _don't_ need polling, because the I/O is already async and you are just waiting for a callback from somewhere.

Arguably, polling is actually a form of _blocking_ I/O, because you have to block on the poll method. The magic is that (1) we can multiplex polling of multiple sockets into a single blocking call and (2) by integrating this blocking directly into the runtime worker loop it actually becomes "free" i.e. the runtime would have had to block _anyway_ and the user gets an experience of 100% async, non-blocking I/O.

So I wonder if we can use a term like "multiplexed blocking I/O". I just made this up haha so I'm open to better ideas.
```suggestion
 * systems are typically used in scenarios such as handling multiplexed blocking I/O or other
```

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
```diff
@@ -17,6 +17,23 @@
...
+ * event-driven systems, where one needs to repeatedly check (or "poll") some condition or
+ * state.
```
armanbilge: An important feature of the `poll` method is that it can block to wait for the condition. If it isn't able to block, then the runtime degrades into a busy-loop when it is idle.

```suggestion
 * event-driven systems, where one needs to repeatedly check (or "poll") some condition or
 * state, blocking up to some timeout until it is ready.
```

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
```diff
@@ -17,6 +17,23 @@
...
+ *
+ * This class abstracts the general components and actions of a polling system, such as:
+ *   - The user-facing interface (API) which interacts with the outside world
+ *   - The thread-local data structure used for polling, which keeps track of the internal state
+ *     of the system and its events
+ *   - The lifecycle management methods, such as creating and closing the polling system and its
+ *     components
+ *   - The interaction methods, such as polling events and interrupting the process
```
armanbilge: These are the methods that the runtime (e.g. the work-stealing threadpool or single-thread event loop) uses to interact with the polling system. But they are not intended for user interaction.

```suggestion
 *   - The runtime interaction methods, such as polling events and interrupting the process
```

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
```diff
@@ -29,19 +46,49 @@ abstract class PollingSystem {
    */
   type Poller <: AnyRef
 
+  /**
+   * Provides the functionality to close the polling system.
```
armanbilge:
```suggestion
   * Closes the polling system.
```

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
```diff
@@ -29,19 +46,49 @@ abstract class PollingSystem {
...
+   */
   def close(): Unit
 
+  /**
+   * Creates a new instance of the user-facing interface.
+   *
+   * @param register
+   *   callback that takes a function from `Poller` to `Unit` The function is used to register a
+   *   new poller.
```
armanbilge: > The function is used to register a new poller

This is not correct, it's used to obtain an existing poller, not to register a new one. I can see why it's confusing, because it's named "register". I think it was originally Daniel's idea in one of the older prototypes, maybe he can explain it, or we should consider a different name?

```suggestion
   * @param register
   *   callback to obtain a thread-local `Poller`.
```

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
```diff
@@ -29,19 +46,49 @@ abstract class PollingSystem {
...
+  /**
+   * Creates a new instance of the thread-local data structure used for polling.
+   *
+   * @return
+   *   an instance of the poller `Poller`.
+   */
   def makePoller(): Poller
 
+  /**
+   * Provides the functionality to close a specific poller.
```
armanbilge:
```suggestion
   * Close a specific poller.
```

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
```diff
@@ -29,19 +46,49 @@ abstract class PollingSystem {
...
+   *
+   * @param poller
+   *   the poller to be closed.
+   */
   def closePoller(poller: Poller): Unit
 
   /**
+   * @param poller
+   *   the poller used to poll events.
```
armanbilge:
```suggestion
   *   the thread-local [[Poller]] used to poll events.
```

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
```diff
@@ -29,19 +46,49 @@ abstract class PollingSystem {
...
+   *
    * @param nanos
    *   the maximum duration for which to block, where `nanos == -1` indicates to block
    *   indefinitely.
    *
+   * @param reportFailure
+   *   callback that handles any failures that occur during polling.
+   *
    * @return
    *   whether any events were polled
```
armanbilge:
```suggestion
   *   whether any events were polled. e.g. if the method returned due to timeout, this should be `false`.
```

armanbilge (commented):
Thank you so much for this!! Nice job.

This is actually a really great opportunity to elaborate and discuss some of these design choices, so I've added a few nitpicky comments about some specific details and terminology. Curious what you and @djspiewak think!

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
antoniojimeneznieto: I find the distinction between truly async I/O and polling operations very interesting and clarifying. 

I must say that the term 'multiplexed blocking I/O' really resonates with the concept. I like it :)

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
antoniojimeneznieto: I understand, maybe a more fitting description could be 'Represents a stateful system for managing ...'

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
antoniojimeneznieto: I had a conversation with Daniel and he clarified me how it works. We all agrees that the name is confusing and we thought about renaming `register` to `access`. What do you think ? :)

armanbilge: Thanks for discussing that! I like `access` 👍

armanbilge (approved):
This is excellent. Thanks for the work and thoughtful discussion!

Screen Shot 2023-07-24 at 10 38 35 AM

PR #3760, opened 2023-07-24
fix typo in tutorial

vgrigoriu:

PR #3761, opened 2023-07-26
Changed tutorial so consumers are not overwhelmed by producers

lrodero:
The second section of the tutorial describes how to use fibers to solve the producer-consumer problem. Thing is, as the code is now, producers outpace consumers so much that consumers seem to be 'frozen'. The explication seems to be the `Ref.modify` run by consumers is way slower than the `Ref.getAndUpdate` call run by producers. This forces consumers to retry once and again to run the `modify` call.

This PR is to add text [at the end of the first subsection](https://lrodero.github.io/cats-effect/docs/tutorial#issue-1-the-producer-outpaces-the-consumer) (the one introducing a 'naive' solution to the producers-consumers problem) to describe the problem and to propose several solutions: using a `.sleep` call (the one used in the code samples); using `AtomicCell` instead of `Ref`, or using a bounded queue (which is implemented in another section afterwards).

The new version of the tutorial can be read online [here](https://lrodero.github.io/cats-effect/docs/tutorial). All code samples are available [here](https://github.com/lrodero/cats-effect-tutorial).

Other minor changes:
* Typo as in PR #3760
* In copy file section merge `transfer` and `transmit` functions in a single one `transfer` in order to simplify code and ease understanding
* In code samples use `Sync[F].whenA` instead of `if ... else Sync[F].unit`
* Updated sbt version (1.9.3)
* Changed wrong path `catseffecttutorial.CopyFile` to correct one `catseffecttutorial.copyfile.CopyFile`

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -42,11 +42,13 @@ running the code snippets in this tutorial, it is recommended to use the same
 ```scala
 name := "cats-effect-tutorial"
 
-version := "3.5.1"
+val ceVersion = "3.5.1"
 
-scalaVersion := "2.13.6"
+version := ceVersion
 
-libraryDependencies += "org.typelevel" %% "cats-effect" % "3.5.1" withSources() withJavadoc()
+scalaVersion := "2.13.11"
+
+libraryDependencies += "org.typelevel" %% "cats-effect" % ceVersion withSources() withJavadoc()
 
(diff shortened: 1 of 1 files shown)


lrodero:
Sorry, I have to close this PR as I was using my company 'id' in github for some contributions (although I'm not using their resources nor working time to work on this). I'll re-PR it asap.

PR #3762, opened 2023-07-26
Changed tutorial so consumers are not overwhelmed by producers

lrodero:
The second section of the tutorial describes how to use fibers to solve the producer-consumer problem. Thing is, as the code is now, producers outpace consumers so much that consumers seem to be 'frozen'. The explication seems to be the `Ref.modify` run by consumers is way slower than the `Ref.getAndUpdate` call run by producers. This forces consumers to retry once and again to run the `modify` call.

This PR is to add text [at the end of the first subsection](https://lrodero.github.io/cats-effect/docs/tutorial#issue-1-the-producer-outpaces-the-consumer) (the one introducing a 'naive' solution to the producers-consumers problem) to describe the problem and to propose several solutions: using a `.sleep` call (the one used in the code samples); using `AtomicCell` instead of `Ref`, or using a bounded queue (which is implemented in another section afterwards).

The new version of the tutorial can be read online [here](https://lrodero.github.io/cats-effect/docs/tutorial). All code samples are available [here](https://github.com/lrodero/cats-effect-tutorial).

Other minor changes:
* Typo as in PR #3760
* In copy file section merge `transfer` and `transmit` functions in a single one `transfer` in order to simplify code and ease understanding
* In code samples use `Sync[F].whenA` instead of `if ... else Sync[F].unit`
* Updated sbt version (1.9.3)
* Changed wrong path `catseffecttutorial.CopyFile` to correct one `catseffecttutorial.copyfile.CopyFile`

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -44,7 +44,7 @@ name := "cats-effect-tutorial"
 
 version := "3.5.2"
 
-scalaVersion := "2.13.6"
+scalaVersion := "2.13.11"
 
 libraryDependencies += "org.typelevel" %% "cats-effect" % "3.5.2" withSources() withJavadoc()
 
@@ -56,11 +56,11 @@ scalacOptions ++= Seq(
 )
 ```
 
-Also make sure that you use a recent version of `sbt`, at least `1.4.2`. You can
-set the `sbt` version in `project/build.properties` file:
+Also make sure that you use a recent version of `sbt`, _e.g._ `1.9.7`. You can set
+the `sbt` version in `project/build.properties` file:
 
 ```scala
-sbt.version=1.4.2
+sbt.version=1.9.7
 ```
 
 Almost all code snippets in this tutorial can be pasted and compiled right in
@@ -162,7 +162,7 @@ import cats.effect.{IO, Resource}
 import java.io.{File, FileInputStream}
 
 def inputStream(f: File): Resource[IO, FileInputStream] =
-  Resource.fromAutoCloseable(IO(new FileInputStream(f)))
+  Resource.fromAutoCloseable(IO.blocking(new FileInputStream(f)))
 ```
 
 That code is way simpler, but with that code we would not have control over what
@@ -181,11 +181,11 @@ import java.io._
 def inputOutputStreams(in: File, out: File): Resource[IO, (InputStream, OutputStream)] = ???
 
 // transfer will do the real work
-def transfer(origin: InputStream, destination: OutputStream): IO[Long] = ???
+def transfer(origin: InputStream, destination: OutputStream, buffer: Array[Byte], acc: Long): IO[Long] = ???
 
 def copy(origin: File, destination: File): IO[Long] =
   inputOutputStreams(origin, destination).use { case (in, out) =>
-    transfer(in, out)
+    transfer(in, out, new Array[Byte](1024 * 10), 0)
   }
 ```
 
@@ -220,17 +220,17 @@ import java.io._
 // function inputOutputStreams not needed
 
 // transfer will do the real work
-def transfer(origin: InputStream, destination: OutputStream): IO[Long] = ???
+def transfer(origin: InputStream, destination: OutputStream, buffer: Array[Byte], acc: Long): IO[Long] = ???
 
 def copy(origin: File, destination: File): IO[Long] = {
-  val inIO: IO[InputStream]  = IO(new FileInputStream(origin))
-  val outIO:IO[OutputStream] = IO(new FileOutputStream(destination))
+  val inIO: IO[InputStream]  = IO.blocking(new FileInputStream(origin))
+  val outIO:IO[OutputStream] = IO.blocking(new FileOutputStream(destination))
 
   (inIO, outIO)              // Stage 1: Getting resources
     .tupled                  // From (IO[InputStream], IO[OutputStream]) to IO[(InputStream, OutputStream)]
     .bracket{
       case (in, out) =>
-        transfer(in, out)    // Stage 2: Using resources (for copying data, in this case)
+        transfer(in, out, new Array[Byte](1024 * 10), 0L) // Stage 2: Using resources (for copying data, in this case)
     } {
       case (in, out) =>      // Stage 3: Freeing resources
         (IO(in.close()), IO(out.close()))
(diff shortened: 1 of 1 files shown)


docs/tutorial.md
```diff
@@ -539,33 +535,34 @@ integers (`1`, `2`, `3`...), consumer will just read that sequence.  Our shared
...
 Now, our `producer` method will be:
-
 ```scala mdoc:compile-only
 import cats.effect._
 import cats.effect.std.Console
 import cats.syntax.all._
-import collection.immutable.Queue
+import scala.collection.immutable.Queue
 
 def producer[F[_]: Sync: Console](queueR: Ref[F, Queue[Int]], counter: Int): F[Unit] =
```
durban, line 550: `Sync` seems unnecessarily strong here.

docs/tutorial.md
```diff
@@ -580,14 +577,14 @@ queue but it must be aware that the queue can be empty:
 import cats.effect._
 import cats.effect.std.Console
 import cats.syntax.all._
-import collection.immutable.Queue
+import scala.collection.immutable.Queue
 
 def consumer[F[_]: Sync: Console](queueR: Ref[F, Queue[Int]]): F[Unit] =
```
durban, line 580: I also don't think we need `Sync` here.

docs/tutorial.md
```diff
@@ -685,29 +679,87 @@ have some specific and unusual requirements you should prefer to use higher
...
+   to be able to use `sleep` now `F` requires an implicit `Async[F]` instance. The
+   new producer will look like this:
+   ```scala mdoc:compile-only
+   import cats.effect._
+   import cats.effect.std.Console
+   import cats.syntax.all._
+   import scala.collection.immutable.Queue
+   import scala.concurrent.duration.DurationInt
+   
+   def producer[F[_]: Async: Console](queueR: Ref[F, Queue[Int]], counter: Int): F[Unit] =
```
durban, line 706: `Async` seems unnecessarily strong for a simple `sleep`.

docs/tutorial.md
```diff
@@ -685,29 +679,87 @@ have some specific and unusual requirements you should prefer to use higher
...
+     for {
+       _ <- Async[F].whenA(counter % 10000 == 0)(Console[F].println(s"Produced $counter items"))
+       _ <- Async[F].sleep(1.microsecond) // To prevent overwhelming consumers
+       _ <- queueR.getAndUpdate(_.enqueue(counter + 1))
+       _ <- producer(queueR, counter + 1)
+     } yield ()
+   ```
+2. Replace `Ref` with `AtomicCell` to keep the `Queue` instance. `AtomicCell`,
+   as `Ref`, is a concurrent data structure to keep a reference to some data.
```
durban: "as Ref" -> "like Ref"? (But I'm not sure, english is not my first language.)

docs/tutorial.md
```diff
@@ -685,29 +679,87 @@ have some specific and unusual requirements you should prefer to use higher
...
+   But unlike `Ref` it ensures that only one fiber can operate on that reference
+   at any given time. Thus the consumer won't have to try once and again to modify
+   its content. Otoh `AtomicCell` is slower than `Ref`. This is because `Ref` is
```
durban: "Otoh": I would recommend spelling out the abbreviation.

docs/tutorial.md
```diff
@@ -685,29 +679,87 @@ have some specific and unusual requirements you should prefer to use higher
...
+
+#### Issue 2: consumer runs even if there are no elements in the queue
+The consumer will be continually running regardless if there are elements in the
+queue, which is far from ideal. If we have several consumers competing for the
+data the problem gets even worse. We will address this problem in the next
+section by using [Deferred](std/deferred.md).
+
+The full implementation of the naive producer consumer we have just created in
+this section is available 
+[here](https://github.com/lrodero/cats-effect-tutorial/blob/series/3.x/src/main/scala/catseffecttutorial/producerconsumer/InefficientProducerConsumer.scala).
```
durban, line 738: This have the risk of becoming a dead link later (although I don't have better idea).

docs/tutorial.md
```diff
@@ -749,7 +801,7 @@ def consumer[F[_]: Async: Console](id: Int, stateR: Ref[F, State[F, Int]]): F[Un
 
   for {
     i <- take
-    _ <- if(i % 10000 == 0) Console[F].println(s"Consumer $id has reached $i items") else Async[F].unit
+    _ <- Async[F].whenA(i % 10000 == 0)(Console[F].println(s"Consumer $id has reached $i items"))
```
durban, line 802: `Async` as above.

docs/tutorial.md
```diff
@@ -761,40 +813,45 @@ we will have now several producers and consumers running in parallel). The
...
+import cats.effect.{Async, Deferred, Ref}
 import cats.effect.std.Console
 import cats.syntax.all._
 import scala.collection.immutable.Queue
+import scala.concurrent.duration.DurationInt
 
 case class State[F[_], A](queue: Queue[A], takers: Queue[Deferred[F,A]])
 
-def producer[F[_]: Sync: Console](id: Int, counterR: Ref[F, Int], stateR: Ref[F, State[F,Int]]): F[Unit] = {
+def producer[F[_]: Async: Console](id: Int, counterR: Ref[F, Int], stateR: Ref[F, State[F,Int]]): F[Unit] = {
```
durban, line 829: `Async` as above.

docs/tutorial.md
```diff
@@ -842,18 +899,12 @@ are started in their own fiber by the call to `parSequence`, which will wait for
...
-
-
 ### Producer consumer with bounded queue
-Having a bounded queue implies that producers, when queue is full, will wait (be
-'semantically blocked') until there is some empty bucket available to be filled.
-So an implementation needs to keep track of these waiting producers. To do so we
-will add a new queue `offerers` that will be added to the `State` alongside
-`takers`.  For each waiting producer the `offerers` queue will keep a
+Having a bounded queue implies that producers, when the queue is full, will wait
+(be 'semantically blocked') until there is some empty bucket available to be
```
durban: I think the new terminology is "fiber blocked".

docs/tutorial.md
lrodero: It's not strictly required, that is true, we could just use `cats.Monad`. But I preferred to use CE type classes, as in the other code snippets that use `Sync` and `Async`. I'm ok with changing it but please let's be certain that is what we want :slightly_smiling_face: (implies changes in snippets, text and linked [repo](https://github.com/lrodero/cats-effect-tutorial)).

docs/tutorial.md
lrodero: You are absolutely right, changing it :+1: .

docs/tutorial.md
lrodero: It's a repo of my own and I confirm that I intend to keep it open and cared of, but I totally understand the concern :slightly_frowning_face: . I don't mind contributing the code to the main CE repo, but I don't know if there is a proper place to fit it in. IIRC this was commented in a previous PR but no other solution was found.

docs/tutorial.md
lrodero: Nor mine :sweat_smile: .  But I think you might be right and it sounds ok, changing it.

docs/tutorial.md
lrodero: While a decision is taken about this I have addressed the other comments, thanks for the input!

docs/tutorial.md
```diff
@@ -901,7 +901,7 @@ previous example this program shall run forever until the user presses CTRL-C.
 
 ### Producer consumer with bounded queue
 Having a bounded queue implies that producers, when the queue is full, will wait
-(be 'semantically blocked') until there is some empty bucket available to be
+(be 'filter blocked') until there is some empty bucket available to be
```
durban: filter -> fiber?

docs/tutorial.md
durban: I definitely think `Sync`/`Async` is too much here. But since it's already there, I don't think there is any need to change them _in this PR_. It's fine to leave it for a followup.

docs/tutorial.md
durban: Okay, as I've said, I don't have a better idea :man_shrugging:

docs/tutorial.md
lrodero: Clumsy fingers, sorry about that.

djspiewak (changes requested):
Needs some fairly trivial conflict resolution and I think it's ready to merge! Thank you!

djspiewak (approved):

PR #3773, opened 2023-08-04
update CI to use graalvm@17

samspills:
graalvm for jdk 11 was removed here: https://github.com/typelevel/jdk-index/pull/197


armanbilge (approved):
thank you ❤️

PR #3774, opened 2023-08-04
Print warning when we detect a non-main-thread main

Daenyth:
This should help people not run facefirst into https://github.com/sbt/sbt/issues/6242
It's a frequent question for new users


core/jvm/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -314,6 +314,29 @@ trait IOApp {
   protected def onCpuStarvationWarn(metrics: CpuStarvationWarningMetrics): IO[Unit] =
     CpuStarvationCheck.logWarning(metrics)
 
+  /**
+   * Defines what to do when IOApp detects that `main` is being invoked on a `Thread` which
+   * isn't the main process thread. This condition can happen when we are running inside of an
+   * `sbt run` with `fork := false`
+   */
+  private def onNonMainThreadDetected(): Unit = {
```
Daenyth, line 322: This can probably become `protected` in the next API break, though honestly I'm skeptical that it's valuable to expose at all

core/jvm/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -314,6 +314,29 @@ trait IOApp {
...
+        .map(_.equalsIgnoreCase("true"))
+        .getOrElse(true)
+    if (shouldPrint)
+      System
+        .err
+        .println(
+          """|Warning: IOApp `main` is running on a thread other than the main thread.
+             |This may prevent correct resource cleanup after `main` completes.
+             |This condition could be caused by executing `run` in an interactive sbt session with `fork := false`.
+             |Set `Compile / run / fork := true` in this project to resolve this.
```
armanbilge, line 334:
```suggestion
             |Set `Compile / run / fork := true` in this project to resolve this.
             |
             |To silence this warning set the system property:
             |`-Dcats.effect.warnOnNonMainThreadDetected=false`.
```

core/jvm/src/main/scala/cats/effect/IOApp.scala
djspiewak: Yeah I'd rather not expose an API like this in `IOApp`. :-) Better to keep it private.

core/jvm/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -314,6 +314,32 @@ trait IOApp {
...
+  private def onNonMainThreadDetected(): Unit = {
+    val shouldPrint =
+      Option(System.getProperty("cats.effect.warnOnNonMainThreadDetected"))
+        .map(_.equalsIgnoreCase("true"))
+        .getOrElse(true)
+    if (shouldPrint)
+      System
+        .err
+        .println(
+          """|Warning: IOApp `main` is running on a thread other than the main thread.
```
djspiewak:
```suggestion
          """|[WARNING] IOApp `main` is running on a thread other than the main thread.
```

We use this convention in the starvation checker

core/jvm/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -314,6 +314,32 @@ trait IOApp {
...
+             |This may prevent correct resource cleanup after `main` completes.
+             |This condition could be caused by executing `run` in an interactive sbt session with `fork := false`.
+             |Set `Compile / run / fork := true` in this project to resolve this.
+             |
+             |To silence this warning set the system property:
+             |`-Dcats.effect.warnOnNonMainThreadDetected=false`.
```
djspiewak, line 337: To remain consistent, we should also (in the next minor release) add an overridable `Boolean` field in `IOApp` which controls this behavior (defaulted to read from the system property). Can you toss that in an issue so we don't forget?

djspiewak (approved):

djspiewak (changes requested):
Wrong button on previous review sorry

core/jvm/src/main/scala/cats/effect/IOApp.scala
armanbilge: - https://github.com/typelevel/cats-effect/issues/3806

djspiewak (approved):

PR #3781, opened 2023-08-08
Timer stealing without a concurrent data structure

armanbilge:
In https://github.com/typelevel/cats-effect/pull/3499 we introduced a custom `ConcurrentSkipListMap`-like data structure. Notably this enabled timers to be stolen between threads (and also for canceled timers to be removed on cancelation).

Here I explore if we can support timer stealing without a specialized concurrent data structure. The observations are that:

1. in practice, most if not all timers will be some kind of `IO.sleep`
2. `IO.sleep` is implemented in terms of `IO.async`
3. the callback in `IO.async` is already a synchronization point: it can be invoked many times, but only the first will win

So if Thread B steals and invokes a timer from Thread A, it actually doesn't need to notify Thread A of this at all. When Thread A later tries to invoke that timer's callback it will safely no-op. (We can do even better if Thread B lazily nulls out that callback and this is fortuitously published to Thread A, then it can avoid invoking the callback at all without any explicit synchronization.)

So this PR introduces a `TimersHeap` that is intended to be written only by its owner thread. However, it may still be read by other threads (guarding against inconsistent state due to race conditions). This makes progress towards https://github.com/typelevel/cats-effect/issues/3544 since accessing the sleepers queue in the worker loop no longer crosses any memory barriers.

Daniel summarized the potential benefits of this strategy on Discord.
> In the case where you don't have any theft at all, it's strictly faster. When you have theft, those timers effectively impose an amortized burden which is a bit higher than necessary on the "steal-ee" worker, but it won't be actively incorrect.

### Fiddly things

- timer cancelation. Without thread-safety we can't immediately remove it from the data structure, so they have to be cleaned up during other traversals for triggering/inserting timers. I wonder if we could be clever and detect if the cancelation request came from the worker thread that owns the `TimerHeap`, in which case it would be okay to clean it up immediately.

- handling non-`IO.sleep` timers i.e. scheduled via the `Scheduler` interface. Presumably the contract is that we will not invoke these more than once, so we need to setup an `AtomicBoolean` to guarantee this.

- scheduling timers from external threads. Currently this involves submitting a `Runnable` via the external queue that then schedules the timer.

### Benchmarks

Somehow `sleepRace` got worse which suggests we need smarter cancelation. Also there is a lot of error in the measurement.

#### series/3.5.x
```
[info] Benchmark                 (size)   Mode  Cnt    Score    Error    Units
[info] SleepBenchmark.sleep       10000  thrpt   20  200.754 ± 57.792  ops/min
[info] SleepBenchmark.sleepRace   10000  thrpt   20   60.868 ± 18.988  ops/min

info] Warming up...
[info] Measuring unloaded...
[info] Unloaded overhead: 0.09038918301666676
[info] Measuring heavily loaded...
[info] Loaded overhead: 0.8053270754333333
[info] Measuring unloaded 100x...
[info] Unloaded overhead 100x: 0.08938118594166666
[info] Measuring heavily loaded 100x...
[info] Loaded overhead: 1.5771005562083333
```

#### This PR
```
[info] Benchmark                 (size)   Mode  Cnt    Score    Error    Units
[info] SleepBenchmark.sleep       10000  thrpt   20  251.271 ± 94.770  ops/min
[info] SleepBenchmark.sleepRace   10000  thrpt   20   38.113 ± 11.084  ops/min


[info] Warming up...
[info] Measuring unloaded...
[info] Unloaded overhead: 0.11180908160833325
[info] Measuring heavily loaded...
[info] Loaded overhead: 1.0175761078166667
[info] Measuring unloaded 100x...
[info] Unloaded overhead 100x: 0.09586685694166674
[info] Measuring heavily loaded 100x...
[info] Loaded overhead: 1.444556097975
```


armanbilge:
Benchmarks after latest changes.

#### series/3.5.x
```
[info] Benchmark                 (size)   Mode  Cnt    Score    Error    Units
[info] SleepBenchmark.sleep       10000  thrpt   20  314.807 ? 49.493  ops/min
[info] SleepBenchmark.sleepRace   10000  thrpt   20   74.017 ?  8.415  ops/min

[info] Warming up...
[info] Measuring unloaded...
[info] Unloaded overhead: 0.08945369533333336
[info] Measuring heavily loaded...
[info] Loaded overhead: 0.4166134899416667
[info] Measuring unloaded 100x...
[info] Unloaded overhead 100x: 0.0899092131999999
[info] Measuring heavily loaded 100x...
[info] Loaded overhead: 0.40755935104166663
```

#### This PR
```
[info] Benchmark                 (size)   Mode  Cnt    Score    Error    Units
[info] SleepBenchmark.sleep       10000  thrpt   20  324.368 ? 47.451  ops/min
[info] SleepBenchmark.sleepRace   10000  thrpt   20   89.657 ?  3.758  ops/min

[info] Warming up...
[info] Measuring unloaded...
[info] Unloaded overhead: 0.0938145233583334
[info] Measuring heavily loaded...
[info] Loaded overhead: 0.43363560199166673
[info] Measuring unloaded 100x...
[info] Unloaded overhead 100x: 0.08244931874166661
[info] Measuring heavily loaded 100x...
[info] Loaded overhead: 0.3595659251250001
```

durban:
This is an interesting approach. I agree, that the "fiddly things" should be solved for this to work.

Specifically, cancelling timers seems critical. I'm not sure, that cleaning them up during other inserts will be enough. (But it might be, as it's not a skiplist any more...) But actually freeing cancelled timers seems important.

Another question: are we okay with timer stealing being a no-op? Because (if I'm reading it correctly), on certain platforms it definitely can be...

armanbilge:
> This is an interesting approach. I agree, that the "fiddly things" should be solved for this to work.

Thanks! I've pushed solutions for all of these in the latest commits.

> Specifically, cancelling timers seems critical. I'm not sure, that cleaning them up during other inserts will be enough.

Based on the first round of benchmarks it did seem like this was not enough. The latest commits I pushed changed it so that if the canceling thread owns the timer heap (i.e. it was the same thread that scheduled it) then it is removed immediately. This appears to have made a difference in the second benchmarks.

I just realized my justification for this approach depends a lot on the I/O-integrated runtime ... so my thinking is, timeouts are typically associated with some I/O op e.g. when the read/write succeeds the timeout can be canceled. In the new I/O integrated runtime, both the timer and the I/O op will be specifically associated with that thread. So when the I/O completes, it will continue on the same thread that started it and also started the timer.

_However_ this is not currently the case where I/O relies on an external selector thread ... in that case, we'd be going through the external queue and we'll end up some random worker thread that probably doesn't own our timer. Hmm.

**tl;dr** this approach heavily optimizes for the happy path where in practice we are staying local to a single worker thread.

> But actually freeing cancelled timers seems important.

Freeing what specifically, the callback? This is definitely freed regardless of which thread does the canceling, by `null`ing out the callback. What is non-deterministic is when this is visible/published (since the update is not synchronized/volatile). But it should happen eventually ...

However, the heap datastructure itself could grow quite large holding all these `null`ed entries. 🤔

armanbilge:
> Another question: are we okay with timer stealing being a no-op? Because (if I'm reading it correctly), on certain platforms it definitely can be...

Oh, forgot to respond to this. I'm not sure what you mean? Even in the current implementation it can be a no-op, if there are no timers to steal 😉 

I guess the relevant question is, are we okay with timer stealing missing some currently stealable timers? Still, in the current approach this is a possibility, since it is just choosing threads at random, you could keep getting very unlucky. But the idea is _eventually_ someone should steal from that thread.

Which I think is the same idea here: _eventually_ the data should be published to the other thread ... right? I'm not enough of an expert particularly when it comes to ARM. Maybe there are some pathological situations where the timers would never be published, but I feel like because of GC and stuff it has to happen eventually ...

durban:
Sorry, I probably didn't look at the newer commits.

Yeah, actually freeing "memory". So not just the callback, but also other memory which was allocated when the callback was inserted. (So yeah, I'm thinking exactly about the "datastructure itself could grow quite large".)

About stealing being no-op: 
> are we okay with timer stealing missing some currently stealable timers?

Exactly. I don't think this can happen currently. Specifically: if there is at least one stealable timer, it will not finish stealing without actually stealing something. (Currently it doesn't steal _everything_ it can, but it could.)

Eventual memory visibility: as far as I know, _plain_ reads (which this PR seems to use) absolutely _don't_ guarantee that. (That would be a property of _opaque_, if I remember correctly, which is JVM 9+ I think.)

armanbilge:
> So yeah, I'm thinking exactly about the "datastructure itself could grow quite large".

Yeah, this is a legitimate danger when too much cancelation is originating on other threads. I think the owner thread is just going to have to do some periodic self-cleanup ... this is annoying b/c scanning the heap for canceled timers is O(n) ...

---

> Eventual memory visibility: as far as I know, _plain_ reads (which this PR seems to use) absolutely _don't_ guarantee that.

Sure, generally speaking plain reads definitely don't guarantee that. But what I'm wondering is if this is effectively guaranteed in practice by virtue of the fact that JVM is a GC runtime. So all allocations need to become visible to other threads eventually, in order for GC to work correctly. Right?

> (That would be a property of _opaque_, if I remember correctly, which is JVM 9+ I think.)

Interesting. I read a bit about it and I'm not entirely sure:

> Returns the value of a variable, accessed in program order, but with no assurance of memory ordering effects with respect to other threads.

https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/invoke/VarHandle.html#getOpaque(java.lang.Object...)

If there's no assurance of ordering wrt other threads, then isn't it lacking the same guarantees? What is "program order"?

durban:
> So all allocations need to become visible to other threads eventually, in order for GC to work correctly. Right?

I'm not sure (the JMM doesn't seem to mention the GC very much). I think it's enough if they become visible to _the GC_; other threads don't have to care.

_Opaque:_ I think this is where I've read that opaque accesses eventually happen: https://gee.cs.oswego.edu/dl/html/j9mm.html#opaquesec (this is _not_ the spec, but unlike the spec, it actually explains some things).

> If there's no assurance of ordering wrt other threads, then isn't it lacking the same guarantees?

I mean, a plain read is not guaranteed to _happen at all_... so an opaque read still has more guarantees.

> What is "program order"?

Here you go: https://docs.oracle.com/javase/specs/jls/se17/html/jls-17.html#jls-17.4.3 (I'm not saying I understand it, but it is very clearly defined :-).

armanbilge:
> I'm not sure (the JMM doesn't seem to mention the GC very much). I think it's enough if they become visible to _the GC_; other threads don't have to care.

Oh yeah, I am not trying to make a formal argument based on JMM or specifications or anything. I'm just saying in practice, because of how things work in real programs on real JVMs on real CPUs, timers will eventually be visible to other threads, probably sooner rather than later.

And in the grand scheme of things I think that's good enough. IMO programs should not rely on stealing to work correctly; it should really only be most useful in pathological scenarios, which are already violating several other "happy-path" assumptions in CE.

Finally, we do have a test for stealing, so we will see for ourselves if that ever fails, particularly on ARM.

---

Aha, thanks for those links about opacity! That's a much more helpful explanation :)

"Writes are eventually visible." Well maybe One Day:tm: we can use that /shrug

durban:
Okay, so you're basically saying, that we _are_ okay with timer stealing missing some currently stealable timers. I also think that's _probably_ fine (at least I can't rellay think of a non-pathological case which would be hurt by it).

armanbilge:
> we _are_ okay with timer stealing missing some currently stealable timers

Yes. We _are_ okay with this, and also my hypothesis is that in practice we won't be detectably missing currently stealable timers _anyway_ e.g. they will be fortuitously published by some other mechanism.

durban:
I think this might have a problem.

It's important, that a timer (which is not cancelled) is actually triggered (its callback is called). I don't think this is guaranteed here:

- Let's see a case, when the owner thread calls `pollFirstIfTriggered`, and concurrently another thread calls `steal`.
- Both of these will call `getAndClear`, which is essentially:
  - plain read of callback
  - plain write of callback (`null`)
- _Previously_ the owner thread initialized callback to non-`null`, also with a plain write.
- It might happen, that the non-owner thread does not "see" the initializing write, _and also_ the owner thread "sees" the `null` the other thread wrote. Thus, nobody will call the callback.

This _might_ be fixed with an `if` in `getAndClear`, but I think a better thing would be to safely initialize `Node`, so everybody can "see" the initializing write.

armanbilge:
> but I think a better thing would be to safely initialize `Node`, so everybody can "see" the initializing write.

That's a good point, we should probably just do that.

Thanks for bringing this up. I had thought through this exact scenario  as well but I convinced myself it was okay based on a Discord thread.

Screen Shot 2023-09-10 at 1 53 28 AM

https://discord.com/channels/632277896739946517/839263556754472990/1087781250284130375

armanbilge:
Relevant reading: https://shipilev.net/blog/2014/all-fields-are-final/

armanbilge:
More relevant reading: https://shipilev.net/blog/2014/safe-public-construction/#_safe_initialization

It describes the implementation "quirk" we are currently relying on.

> **A final field was written.** Notice we do not care about **what** field was actually written, we unconditionally emit the barrier before exiting the (initializer) method. _That means if you have at least one final field write, the final fields semantics extend to every other field written in constructor._

So essentially we are using the `final`ity of `triggerTime` to publish the callback.

https://github.com/typelevel/cats-effect/blob/76b9e8310c9b16ec06849f3ee7e27fedd31bbede/core/jvm/src/main/scala/cats/effect/unsafe/TimerHeap.scala#L327-L331

So at least on HotSpot I don't think it's broken. Whether we want to rely on this quirk however is a much different question. Especially since in this case it's not too hard to fix.

durban:
Hm...

Yeah, I've remembered something like that (except I've misremembered all the important details :-). So yeah, it seems it'll work on hotspot (for now).

> Especially since in this case it's not too hard to fix.

How? Now that I've actually read what the spec says, I'm no sure how to fix it...

armanbilge:
> How? Now that I've actually read what the spec says, I'm no sure how to fix it...

Really? Your idea above seemed reasonable, does it have some flaw?

> This _might_ be fixed with an `if` in `getAndClear`

Also we could make the `callback` a `val` so it's a final field and you use a boolean var to track whether its been invoked. We can't `null` it in that case, but it's not the end of the world.

durban:
Ah, yes, the `if`. That should fix this specific issue. (I was thinking about fixing the whole "`var` might not be visible to other threads" thing, and forgot about the specific problem. In fact, we should be sure that there are no other similar problems.)

> Also we could make the callback a val so it's a final field and you use a boolean var to track whether its been invoked. We can't null it in that case, but it's not the end of the world.

Hm... freeing memory on cancel seems important. Although, the `packIfNeeded` might already ensure that.

armanbilge:
> Hm... freeing memory on cancel seems important

Yes, but to properly cleanup we need to completely remove the node anyway. `null`ing the callback frees a few allocations.

core/jvm/src/main/scala/cats/effect/unsafe/TimerHeap.scala
```diff
@@ -0,0 +1,381 @@
...
+        callback = null
+      back
+    }
+
+    def get(): Right[Nothing, Unit] => Unit = callback
+
+    /**
+     * Cancel this timer.
+     */
+    def apply(): Unit = if (callback ne null) { // if we're not already deleted
```
durban: I think if the cancelling thread doesn't see `callback`, it will not enter the body of this `if`, and so cancellation will simply not happen. (Similarly to the issue previously discussed, this should also be fine on hotspot.)

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -776,3 +767,32 @@ private[effect] final class WorkStealingThreadPool(
   private[unsafe] def getSuspendedFiberCount(): Long =
     workerThreads.map(_.getSuspendedFiberCount().toLong).sum
 }
+
+private object WorkStealingThreadPool {
+
+  /**
+   * A wrapper for a cancelation callback that is created asynchronously. Best-effort: does not
+   * explicitly publish.
```
durban: I think the plain writes here also mean, that cancellation may not happen at all. And this is not in a constructor, so the hotspot barrier will not save us.

Is it okay, if a timer cancellation doesn't happen? It could be argued, that since it's essentially a race with the clock, it's fine if it doesn't. (Maybe that's what the comment says here?) But I don't think I'm convinced by that argument.

tests/jvm/src/test/scala/cats/effect/unsafe/SleepersSpec.scala
```diff
@@ -74,22 +81,26 @@ class SleepersSpec extends Specification {
...
     }
 
     "be ordered correctly even if Long overflows" in {
-      val sleepers = new TimerSkipList
+      val sleepers = new TimerHeap
 
       val now1 = Long.MaxValue - 20L
       val delay1 = 10.nanos.toNanos
-      val expected1 = Long.MaxValue - 10L // no overflow yet
+      // val expected1 = Long.MaxValue - 10L // no overflow yet
```
durban, line 103: Commented code.

tests/jvm/src/test/scala/cats/effect/unsafe/TimerHeapSpec.scala
```diff
@@ -66,65 +71,83 @@ class TimerSkipListSpec extends Specification {
...
+      val r4 = m.insert(0L, 5L, cb4, out)
+      out(0) must beNull
+      val r2 = m.insert(0L, 3L, cb2, out)
+      out(0) must beNull
+      val r3 = m.insert(0L, 4L, cb3, out)
+      out(0) must beNull
 
       m.peekFirstQuiescent() mustEqual cb0
-      m.peekFirstTriggerTime() mustEqual 0L
+      m.peekFirstTriggerTime() mustEqual 1L
```
durban, line 90: What changed here? What's the reason for the +1 offset?

durban (commented):
The CI failed due to some cirrus limit. (Does this mean, that from now on, every CI run will be red?)

armanbilge:
> The CI failed due to some cirrus limit. (Does this mean, that from now on, every CI run will be red?)

At the moment, but we're working on fixing it.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
armanbilge: > Is it okay, if a timer cancellation doesn't happen?

Yes.

This entire PR is premised on the idea that invoking the callback when you are not supposed to is okay, because the callback itself already tracks this state. So that might be invoking the callback a second time, or invoking it a first time when it was supposedly canceled. In either case, the first invocation/cancelation of the callback already won, so all subsequent calls are no-ops.

Although now that you raise this, I think I didn't think through the cancelation case for an arbitrary `Runnable` and not a `Fiber`. But I can fix that.

tests/jvm/src/test/scala/cats/effect/unsafe/SleepersSpec.scala
armanbilge: It became unused but it's still useful for documentation ...

core/jvm/src/main/scala/cats/effect/unsafe/TimerHeap.scala
armanbilge: Hmm, good point. I think we can live with this as justified in https://github.com/typelevel/cats-effect/pull/3781#discussion_r1328134075.

Or we just have to eat the idea that `callback` should be a `val`.

armanbilge: Hm, actually we can fix this by using a different `Boolean` to track if it was removed from the local data structure, when we take that local-removing path.

tests/jvm/src/test/scala/cats/effect/unsafe/TimerHeapSpec.scala
armanbilge: The original scenario was:

```scala
m.insert(0L, 0L, cb0, out)
m.insert(0L, 1L, cb1, out)
```

In that case the first timer is being returned via `out` when scheduling the second timer. Doing a "replace" instead of "remove/push" saves an iteration over the heap and also avoids growing it.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
durban: Okay, so the idea is that since the fiber will be cancelled anyway, it's not a problem if the callback is not removed from the data structure. But this still causes a memory leak, right?

armanbilge: Hmm you're right, it would be a memory leak. Fair enough, let's just use an atomic here to guarantee cancelation propagates.

core/jvm/src/main/scala/cats/effect/unsafe/TimerHeap.scala
armanbilge: Instead of boolean we can just set `index = -1`. fb465db36715616348873dff80dd20116e3987ef

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
armanbilge: > I think I didn't think through the cancelation case for an arbitrary `Runnable`

e0c644cb9d1d3e6e4f34814b6a87131bbf5ed9c4

> let's just use an atomic here to guarantee cancelation propagates

8d08a32c69cb3db1944a900ca2176b72727029b3

djspiewak:
So I futzed around and got my benchmark environment running again, but unfortunately `SleepBenchmark` doesn't run on my machine on this branch. It produces a `NullPointerException` (looping) every time I tried. I can try to get a stack trace tomorrow.

armanbilge:
Thanks, sorry. I also get the stack trace but not entirely sure what I messed up yet. I think it's a regression in fb465db36715616348873dff80dd20116e3987ef because the benchmarks were running before.

armanbilge:
@djspiewak NPEs fixed now, if you don't mind giving benchmarking another try 🙏 thanks!

armanbilge:
Curiosity got the better of me so I did some benchmarks on ARM with a Google Cloud t2a-standard-8 machine (sorry Daniel).

#### this PR
```
[info] Benchmark                 (size)   Mode  Cnt    Score   Error    Units
[info] SleepBenchmark.sleep       10000  thrpt   20  568.314 ± 3.611  ops/min
[info] SleepBenchmark.sleepRace   10000  thrpt   20  143.809 ± 2.512  ops/min

[info] running (fork) cats.effect.benchmarks.SleepDrift 
[info] Warming up...
[info] Measuring unloaded...
[info] Unloaded overhead: 0.0782825053666667
[info] Measuring heavily loaded...
[info] Loaded overhead: 0.19834489787499998
[info] Measuring unloaded 100x...
[info] Unloaded overhead 100x: 0.07358983100833338
[info] Measuring heavily loaded 100x...
[info] Loaded overhead: 0.21740777570833325
```

#### series/3.5.x
```
[info] Benchmark                 (size)   Mode  Cnt    Score     Error    Units
[info] SleepBenchmark.sleep       10000  thrpt   20  329.971 ± 108.372  ops/min
[info] SleepBenchmark.sleepRace   10000  thrpt   20  129.704 ±   0.936  ops/min

[info] running (fork) cats.effect.benchmarks.SleepDrift 
[info] Warming up...
[info] Measuring unloaded...
[info] Unloaded overhead: 0.07882860818333337
[info] Measuring heavily loaded...
[info] Loaded overhead: 0.19550042234166676
[info] Measuring unloaded 100x...
[info] Unloaded overhead 100x: 0.07500671731666664
[info] Measuring heavily loaded 100x...
[info] Loaded overhead: 0.22464878354999995
```

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -152,20 +158,53 @@ private final class WorkerThread(
...
+    val cancel = sleepers.insert(
+      now = now,
+      delay = delay,
       callback = callback,
-      tlr = random
+      out = out
     )
+
+    val cb = out(0)
+    if (cb ne null) cb(RightUnit)
```
durban, line 208: Could we avoid the `out` parameter of `insert` by directly calling `out(0)` (if any) in `insert`? (I might be missing something, but it seems to me we're returning something in `out` just to immediately call it.)

core/jvm/src/main/scala/cats/effect/unsafe/TimerHeap.scala
```diff
@@ -0,0 +1,386 @@
...
+
+/*
+ * Scala.js (https://www.scala-js.org/)
+ *
+ * Copyright EPFL.
+ *
+ * Licensed under Apache License 2.0
+ * (https://www.apache.org/licenses/LICENSE-2.0).
+ *
+ * See the NOTICE file distributed with this work for
```
durban, line 25: I think (part of?) the contents of the NOTICE file of Scala.js should be added to the NOTICE file of Cats Effect.

core/jvm/src/main/scala/cats/effect/unsafe/TimerHeap.scala
```diff
@@ -0,0 +1,386 @@
...
+    } else { // insert at the end
+      val heap = growIfNeeded() // new heap array if it grew
+      size += 1
+      val node = new Node(triggerTime, callback, size)
+      heap(size) = node
+      fixUp(size)
+      node
+    }
+  } else {
+    val node = new Node(now + delay, callback, 1)
```
durban, line 205:
```suggestion
    val node = new Node(computeTriggerTime(now, delay), callback, 1)
```

core/jvm/src/main/scala/cats/effect/unsafe/TimerHeap.scala
```diff
@@ -0,0 +1,386 @@
...
+  }
+
+  private[this] def isExpired(node: Node, now: Long): Boolean =
+    cmp(node.triggerTime, now) <= 0 // triggerTime <= now
+
+  private[this] def growIfNeeded(): Array[Node] = {
+    val heap = this.heap // local copy
+    if (size >= heap.length - 1) {
+      val newHeap = new Array[Node](heap.length * 2)
+      System.arraycopy(heap, 1, newHeap, 1, heap.length - 1)
```
durban:
```suggestion
      val newHeap = Arrays.copyOf(heap, heap.length * 2, classOf[Node])
```

core/jvm/src/main/scala/cats/effect/unsafe/TimerHeap.scala
```diff
@@ -0,0 +1,386 @@
...
+ */
+
+package cats.effect
+package unsafe
+
+import scala.annotation.tailrec
+
+import java.util.concurrent.atomic.AtomicBoolean
+
+private final class TimerHeap extends AtomicBoolean { needsPack =>
```
durban: Can we put a comment somewhere in this file that this is _intentionally_ full of data races?

durban (commented):
I've left a few comments, but overall LGTM. A few general comments:

- I haven't looked into the fix up/down part of the heap (I'm assuming that it is copied from Scala.js, and I'm assuming it is correct).
- I'm still slightly concerned by depending on a Hotspot implementation detail, but don't want to block the PR (and I'm assuming there are other parts of CE which depend on this).
- As this is _intentionally_ full of data races, it is quite hard to review... but I can't find more specific problems right now.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
armanbilge: > it seems to me we're returning something in `out` just to immediately call it.)

That's exactly the case.

---

> Could we avoid the `out` parameter of `insert` by directly calling `out(0)` (if any) in `insert`?

Yes, actually this is how I originally coded it, but then I had a difficult time adapting the existing tests for that, so I switched to this encoding.

core/jvm/src/main/scala/cats/effect/unsafe/TimerHeap.scala
armanbilge: First we need a NOTICE file ... yeah, it's the right thing to do.

armanbilge:
> * I haven't looked into the fix up/down part of the heap (I'm assuming that it is copied from Scala.js

Yes.

---

> * I'm still slightly concerned by depending on a Hotspot implementation detail,

We shouldn't be depending on this anymore, I believe I fixed the ones you pointed out. But let me know if I missed one.

core/jvm/src/main/scala/cats/effect/unsafe/TimerHeap.scala
armanbilge: Actually it's not needed. In this branch this is the one-and-only `Node` in the heap, so you can't overflow.

core/jvm/src/main/scala/cats/effect/unsafe/TimerHeap.scala
armanbilge: 74c92ae94f0446b8cedcd4038d49083696756709

core/jvm/src/main/scala/cats/effect/unsafe/TimerHeap.scala
armanbilge: 04e54eaf96337ce5c0884b64f3d12607519cd8cd

core/jvm/src/main/scala/cats/effect/unsafe/TimerHeap.scala
armanbilge: c001d4002acd35c0be0040b8e7a0c5ef3a02748b

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
durban: I think those tests are what I've written for TimerSkipList. Feel free to restructure them as needed :-) (But it's up to you, it's also fine as is.)

durban (dismissed):
> We shouldn't be depending on this anymore, I believe I fixed the ones you pointed out. But let me know if I missed one.

No, you're right. (I'm slightly bothered by the fact that "I can't find anything wrong with this" instead of "I can show this is correct". But I'll just have to live with that :-)

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -640,26 +611,33 @@ private[effect] final class WorkStealingThreadPool(
...
+
+    scheduleExternal { () =>
+      val worker = Thread.currentThread().asInstanceOf[WorkerThread]
+      cancel.setCallback(worker.sleepLate(scheduledAt, delay, callback))
+    }
+
     cancel
   }
 
   override def sleep(delay: FiniteDuration, task: Runnable): Runnable = {
```
armanbilge, line 664: Here's an interesting question: should we assume that `task` is non-throwing? If we cannot assume that, then we should wrap it in a `try`/`catch` and report non-fatal errors with `reportFailure`.

I think it's safe to assume that the callback in `sleepInternal` is not throwing.

djspiewak:
From a performance standpoint, this is nearly a wash. It does make `sleep` a bit faster, but it also reduces timer granularity under load by about 31%, which definitely isn't awesome (though not the end of the world). I'm not entirely sure *why* timer granularity is made worse by this change; I would have expected the opposite?

## Before

```
[info] Benchmark                 (size)   Mode  Cnt    Score   Error    Units
[info] SleepBenchmark.sleep       10000  thrpt   20  958.362 ± 0.636  ops/min
[info] SleepBenchmark.sleepRace   10000  thrpt   20  198.568 ± 0.277  ops/min

Warming up...
Measuring unloaded...
Unloaded overhead: 0.07872363459999998
Measuring heavily loaded...
Loaded overhead: 0.28008827346666676
Measuring unloaded 100x...
Unloaded overhead 100x: 0.07937270549166664
Loaded overhead: 0.2589938195999999
```

## After

```
[info] Benchmark                 (size)   Mode  Cnt     Score   Error    Units
[info] SleepBenchmark.sleep       10000  thrpt   20  1046.615 ± 1.359  ops/min
[info] SleepBenchmark.sleepRace   10000  thrpt   20   201.767 ± 0.359  ops/min

[info] Warming up...
[info] Measuring unloaded...
[info] Unloaded overhead: 0.07740299239999993
[info] Measuring heavily loaded...
[info] Loaded overhead: 0.25821024226666656
[info] Measuring unloaded 100x...
[info] Unloaded overhead 100x: 0.07822806469166665
[info] Measuring heavily loaded 100x...
[info] Loaded overhead: 0.33923368308333335
```

armanbilge:
I did my own x86 benchmarks and they are comparable to the ARM benchmarks. Both were 8 core linux machines.

#### this PR
```
[info] Benchmark                 (size)   Mode  Cnt    Score   Error    Units
[info] SleepBenchmark.sleep       10000  thrpt   20  397.111 ± 2.954  ops/min
[info] SleepBenchmark.sleepRace   10000  thrpt   20   88.611 ± 2.623  ops/min

[info] running (fork) cats.effect.benchmarks.SleepDrift 
[info] Warming up...
[info] Measuring unloaded...
[info] Unloaded overhead: 0.10907944324999996
[info] Measuring heavily loaded...
[info] Loaded overhead: 0.4034487338499999
[info] Measuring unloaded 100x...
[info] Unloaded overhead 100x: 0.11182439151666657
[info] Measuring heavily loaded 100x...
[info] Loaded overhead: 0.4696231997083333
```

#### series/3.5.x
```
[info] Benchmark                 (size)   Mode  Cnt    Score    Error    Units
[info] SleepBenchmark.sleep       10000  thrpt   20  266.710 ± 14.059  ops/min
[info] SleepBenchmark.sleepRace   10000  thrpt   20   75.928 ±  1.305  ops/min

[info] Warming up...
[info] Measuring unloaded...
[info] Unloaded overhead: 0.11620415939166673
[info] Measuring heavily loaded...
[info] Loaded overhead: 0.40317077278333335
[info] Measuring unloaded 100x...
[info] Unloaded overhead 100x: 0.11014420795833324
[info] Measuring heavily loaded 100x...
[info] Loaded overhead: 0.4731014630083332
```

armanbilge:
[Quoting myself from Discord](https://discord.com/channels/632277896739946517/839263556754472990/1195873042925093025):

> I'd love to see it merged and it seems good but I'm willing to admit its risky for 3.5.x
the risks are that its a complete change of implementation, and that the implementation works best with happy paths that polling promises
e.g. my canonical example is: schedule timeout, start I/O, complete I/O, cancel timeout
in 3.5 the I/O comes back in through external pool, so the cancel is from a random thread
in 3.6 the I/O comes back on the same thread, so the cancel is also on the same thread

Daniel S responds:
> since it's optimizing for a happy path that is very improbable in 3.5, and since it's relatively risky… I think we're better off punting it to 3.6 where the risk has a payoff attached to it 
IMO, let's retarget it and merge it

@durban since you previously approved for merging into 3.5.x, I'm curious if you have any opinion.

durban:
@armanbilge I don't have a strong opinion regarding 3.5 or 3.6. What I think is important though, is to figure out if we're really okay with this "best effort" stealing. Basically the discussion in https://github.com/typelevel/cats-effect/discussions/3873.

armanbilge:
Based on discussion in https://github.com/typelevel/cats-effect/pull/3781#issuecomment-1890812172 and https://github.com/typelevel/cats-effect/discussions/3873#discussioncomment-8434327 I'm going to re-target to `series/3.x` and then it should be good to merge.

NOTICE.txt
```diff
@@ -0,0 +1,8 @@
+cats-effect
+Copyright 2020-2023 Typelevel
```
durban: 2024

core/jvm/src/main/scala/cats/effect/unsafe/TimerHeap.scala
```diff
@@ -0,0 +1,486 @@
...
+          if (invoked) cb(RightUnit)
+
+          val leftInvoked = go(heap, size, 2 * m)
+          val rightInvoked = go(heap, size, 2 * m + 1)
+
+          invoked || leftInvoked || rightInvoked
+        } else false
+      } else false
+
+    val heap = this.heap // local copy
```
durban, line 166: If I understand it correctly, the fact that we never read `null` here depends on the WSTP initializing and starting threads in a particular order (first init everything, then start everything). This seems a little bit fragile. Would it be worth doing a null check here to be sure?

durban (dismissed):
I've left a few minor comments.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
durban: Yeah, I think we should do the `try` to be sure.

core/jvm/src/main/scala/cats/effect/unsafe/TimerHeap.scala
armanbilge: I agree. Good catch.

djspiewak (approved):

PR #3787, opened 2023-08-21
Migrate kernel and law tests to munit

armanbilge:
This was the low-hanging fruit, just to dip our toe in the water.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -312,6 +312,8 @@ val Specs2Version = "4.20.0"
 val ScalaCheckVersion = "1.17.0"
 val DisciplineVersion = "1.4.0"
 val CoopVersion = "1.2.0"
+val MUnitVersion = "1.0.0-M8"
+val DisciplineMUnitVersion = "2.0.0-M3"
 
 val MacrotaskExecutorVersion = "1.1.1"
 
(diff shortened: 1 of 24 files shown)


djspiewak (changes requested):
Solid start. Nothing in here needs changing, just using the Request Changes to hold the merge.

I think we should do two things. First, munit somewhat infamously circumvents sbt's output buffers for essentially no reason, and it occasionally causes weirdness with other tasks. I always have to look up how @etspaceman fixed it and cargo cult from him, since I can never remember it. Second, I'd prefer not to be in a state where we have two simultaneous test frameworks in use within the project, so I think it's worth driving this to completion and merging it all at once.

armanbilge:
> First, munit somewhat infamously circumvents sbt's output buffers for essentially no reason, and it occasionally causes weirdness with other tasks.

Oh right that's extremely annoying. I've cargo-culted that too in various places, I'll make sure to bring that in.

> Second, I'd prefer not to be in a state where we have two simultaneous test frameworks in use within the project, so I think it's worth driving this to completion and merging it all at once.

Fair enough!

iRevive:
Can we close this one?

PR #3788, opened 2023-08-22
Remove `Thunk.asFunction0` utility

armanbilge:
We were discussing by-names on the Scala Discord and I mentioned this hack we had in CE. Seth points out that the Scala 3 bug that originally motivated this was fixed in 3.1.3. Since we now require Scala 3.2+ it is safe to remove it and avoid the unnecessary indirection. With a minor rewrite we can even retain the original test.


build.sbt
```diff
@@ -648,7 +648,9 @@ lazy val core = crossProject(JSPlatform, JVMPlatform, NativePlatform)
       ProblemFilters.exclude[IncompatibleResultTypeProblem](
         "cats.effect.unsafe.TimerSkipList.insert"),
       ProblemFilters.exclude[IncompatibleResultTypeProblem](
-        "cats.effect.unsafe.WorkerThread.sleep")
+        "cats.effect.unsafe.WorkerThread.sleep"),
+      // #3787, internal utility that was no longer needed
+      ProblemFilters.exclude[MissingClassProblem]("cats.effect.Thunk")
```
armanbilge:
```suggestion
      ProblemFilters.exclude[MissingClassProblem]("cats.effect.Thunk"),
      ProblemFilters.exclude[MissingClassProblem]("cats.effect.Thunk$")
```

durban (approved):

PR #3791, opened 2023-08-24
more detail to random.md documentation

wonk1132:
Adds more details about how to use `Random`, including how to do so referentially transparent. Includes a unit test example.


armanbilge:
Thanks! Would you mind re-targeting to `series/3.5.x` branch, so it can be published sooner?

PR #3792, opened 2023-08-24
adds more details to random.md documentation

wonk1132:
Adds more detailed documentation for the `Random` type, including a unit test example.


djspiewak (dismissed):
This is great! Can we retarget at series/3.5.x though so it gets out sooner?

docs/std/random.md
```diff
@@ -36,23 +36,77 @@ on the JVM:
...
-Random.scalaUtilRandom[IO]
+object Magic {
+  def getMagicNumber[F[_] : Monad](
+    mult: Int
+  )(implicit randomizer: F[Random[F]]): F[String] =
+    for
+      rand <- randomizer.flatMap(random => random.betweenInt(1, 11)) // 11 is excluded
+      number = rand * mult
+      msg = s"the magic number is: $number"
+    yield msg
```
armanbilge:
```suggestion
    for {
      rand <- randomizer.flatMap(random => random.betweenInt(1, 11)) // 11 is excluded
      number = rand * mult
      msg = s"the magic number is: $number"
    } yield msg
```

docs/std/random.md
```diff
@@ -36,23 +36,76 @@ on the JVM:
...
 import cats.effect.std.Random
+import cats.implicits.*
+import cats.Monad
+import cats.effect.unsafe.implicits.global
+
+object BusinessLogic {
+
+  // use the standard implementation of Random backed by java.util.Random()
+  // (the same implementation as Random.javaUtilRandom(43))
+  implicit val r: IO[Random[IO]] = Random.scalaUtilRandom[IO]
```
durban: Is having an implicit `IO[A]` a common pattern? I don't think I've ever seen it. I would expect an implicit `Random[IO]` here (but of course that opens the question of "should `Random` be passed implicitly").

docs/std/random.md
```diff
@@ -36,23 +36,76 @@ on the JVM:
...
+
+```scala mdoc
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatest.matchers.must.Matchers.*
+
+class MagicSpec extends AnyFunSuite {
+
+  // for testing, create a Random instance that gives back the same number every time. With
+  // this version of the Random type class, we can test our business logic works as intended.
+  implicit val r: IO[Random[IO]] = IO(
```
durban:
```suggestion
  implicit val r: IO[Random[IO]] = IO.pure(
```

docs/std/random.md
```diff
@@ -36,23 +36,76 @@ on the JVM:
...
+    new Random[IO] {
+      def betweenInt(minInclusive: Int, maxExclusive: Int): IO[Int] =
+        IO(7) // gives back 7 every call
```
durban:
```suggestion
        IO.pure(7) // gives back 7 every call
```

docs/std/random.md
wonk1132: ah. yup.

docs/std/random.md
wonk1132: double yup

docs/std/random.md
wonk1132: I agree this seems odd and it doesn't make sense in this context why it's used implicitly. I changed it to be an explicit param.

djspiewak:
Looks like the build failures are legitimate

wonk1132:
sorry, I don't get what I need to do to fix this issue. Can someone point me the right direction?

armanbilge:
Looks like import errors. Use `._` instead of `.*`. Also `cats.implicits` is legacy, use `cats.syntax.all`.

Do you know how to run it locally? `docs/mdoc` in sbt console. Then you can see and fix the errors before you push.

Thanks for your work!

docs/std/random.md
```diff
@@ -36,23 +36,75 @@ on the JVM:
...
 
-## Creating a `Random` instance
+## Creating and using a `Random` instance
 
-Obtaining an instance of `Random` can be as simple as:
+The following example shows the usage of `Random` in a way that is *referentially transparent*, meaning that you can know from the 'outside' what the result will be by knowing the inputs:
 ```scala mdoc:silent
-import cats.effect.IO
 import cats.effect.std.Random
+import cats.syntax.all
```
armanbilge:
```suggestion
import cats.syntax.all._
```

sshark:
I feel we should retain the example of using `Random` in the `dieRoll` generic function in https://typelevel.org/cats-effect/docs/std/random#using-random

```
def dieRoll[F[_]: Functor: Random]: F[Int] =
  Random[F].betweenInt(0, 6).map(_ + 1) 
```

I would like to expand this example into a working copy-and-paste example. This example at current state does not show how `dieRoll` is used together with an implicit `Random` instance. Thank you

PR #3796, opened 2023-08-29
Cirrus CI cutbacks ...

armanbilge:
I missed this, but apparently last month Cirrus CI announced that it is no longer allowing unlimited use for free 😕 

https://cirrus-ci.org/blog/2023/07/17/limiting-free-usage-of-cirrus-ci/

Here's our usage over the last few months, measured in "credits". Green is macOS, blue Linux. Our new allowance will be 50 credits per month 😕 

<details>

Screen Shot 2023-08-28 at 6 42 53 PM

</details>

So we need to do some serious belt-tightening, esp. wrt to the macOS CI jobs.

So far the changes I propose are:
1. Drop the JCStress tests, these take a _really_ long time.
2. Drop 2.12 jobs. It's just not worth it.
3. Only run Apple Silicon jobs for Scala 3.
4. Only run Scala Native jobs for Scala 3.

Meanwhile I restored the Linux ARM CI jobs for 2.13 (JVM) and 3 (JVM, Native). We really do need those ... hopefully they will not be so flaky this time.

Not sure if this is enough. We shall see ...

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -4,68 +4,50 @@ jvm_highcore_task:
     cpu: 4
     memory: 8G
   matrix:
-    - name: JVM high-core-count 2.12
-      script: sbt '++ 2.12' testsJVM/test
     - name: JVM high-core-count 2.13
-      script: sbt '++ 2.13' testsJVM/test stressTests/Jcstress/run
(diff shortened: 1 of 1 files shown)


armanbilge:
Administrative merge, since this comes into effect on Sept 1.

PR #3799, opened 2023-08-29
Fix Hotswap.get being blocked by Hotswap.swap

tibetansingingcat:
`Hotswap#swap` currently acquires all permits for its Semaphore during allocation of the `next` Resource, blocking `Hotswap#get` until the new Resource has been instantiated. In many use cases this will not be perceptible but if the Resource has a long initialization time, any calls to `get` will block until the new Resource is ready... effectively undermining the very premise of Hotswap.


armanbilge:
Btw we should target these changes at series/3.5.x so we can get it into a patch release. Thanks!

PR #3800, opened 2023-08-29
Fix `Hotswap#get` being blocked by `Hotswap#swap`

tibetansingingcat:
`Hotswap#swap` currently acquires all permits for its Semaphore during allocation of the `next` Resource, blocking `Hotswap#get` until the new Resource has been instantiated. In many use cases this will not be perceptible but if the Resource has a long initialization time, any calls to `get` will block until the new Resource is ready... effectively undermining the very premise of Hotswap.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
@@ -136,7 +136,7 @@ object Hotswap {
                 case (r, fin) =>
                   exclusive.mapK(poll).onCancel(Resource.eval(fin)).surround {
                     swapFinalizer(Acquired(r, fin)).as(r)
-                  })
+                  }
               }
             }
 
--- a/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
@@ -134,9 +134,9 @@ object Hotswap {
             F.uncancelable { poll =>
               poll(next.allocated).flatMap {
                 case (r, fin) =>
-                  poll(exclusive.surround {
+                  poll(exclusive.onCancel(Resource.eval(fin)).surround {
                     swapFinalizer(Acquired(r, fin)).uncancelable.as(r)
-                  }).onCancel(fin)
+                  })
               }
             }
 
--- a/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
@@ -131,12 +131,12 @@ object Hotswap {
         new Hotswap[F, R] {
 
           override def swap(next: Resource[F, R]): F[R] =
-            exclusive.surround {
-              F.uncancelable { poll =>
-                poll(next.allocated).flatMap {
-                  case (r, fin) =>
-                    swapFinalizer(Acquired(r, fin)).as(r)
-                }
+            F.uncancelable { poll =>
+              poll(next.allocated).flatMap {
+                case (r, fin) =>
+                  poll(exclusive.surround {
+                    swapFinalizer(Acquired(r, fin)).uncancelable.as(r)
+                  }).onCancel(fin)
               }
             }
 
--- a/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
@@ -134,8 +134,8 @@ object Hotswap {
             F.uncancelable { poll =>
               poll(next.allocated).flatMap {
                 case (r, fin) =>
-                  poll(exclusive.onCancel(Resource.eval(fin)).surround {
-                    swapFinalizer(Acquired(r, fin)).uncancelable.as(r)
+                  exclusive.mapK(poll).onCancel(Resource.eval(fin)).surround {
+                    swapFinalizer(Acquired(r, fin)).as(r)
                   })
               }
             }
--- a/tests/shared/src/test/scala/cats/effect/std/HotswapSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/std/HotswapSpec.scala
@@ -119,13 +119,14 @@ class HotswapSpec extends BaseSpec { outer =>
     "successfully cancel during swap and run finalizer if cancelation is requested while waiting for get to release" in ticked {
       implicit ticker =>
         val go = Ref.of[IO, List[String]](List()).flatMap { log =>
-          Hotswap[IO, Unit](logged(log, "a")).use { case (hs, _) =>
-            for {
-              _ <- hs.get.evalMap(_ => IO.sleep(1.minute)).use_.start
-              _ <- IO.sleep(2.seconds)
-              _ <- hs.swap(logged(log, "b")).timeoutTo(1.second, IO.unit)
-              value <- log.get
-            } yield value
+          Hotswap[IO, Unit](logged(log, "a")).use {
+            case (hs, _) =>
+              for {
+                _ <- hs.get.evalMap(_ => IO.sleep(1.minute)).use_.start
+                _ <- IO.sleep(2.seconds)
+                _ <- hs.swap(logged(log, "b")).timeoutTo(1.second, IO.unit)
+                value <- log.get
+              } yield value
           }
         }
 
--- a/tests/shared/src/test/scala/cats/effect/std/HotswapSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/std/HotswapSpec.scala
@@ -109,7 +109,7 @@ class HotswapSpec extends BaseSpec { outer =>
     "not block current resource while swap is instantiating new one" in ticked {
       implicit ticker =>
         val go = Hotswap.create[IO, Unit].use { hs =>
-          hs.swap(Resource.eval(IO.sleep(1.minute) *> IO.unit)).start *>
+          hs.swap(IO.sleep(1.minute).toResource).start *>
             IO.sleep(5.seconds) *>
             hs.get.use_.timeout(1.second) *> IO.unit
         }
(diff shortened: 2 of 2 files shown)


std/shared/src/main/scala/cats/effect/std/Hotswap.scala
```diff
@@ -131,12 +131,12 @@ object Hotswap {
...
-                poll(next.allocated).flatMap {
-                  case (r, fin) =>
-                    swapFinalizer(Acquired(r, fin)).as(r)
-                }
+            F.uncancelable { poll =>
+              poll(next.allocated).flatMap {
+                case (r, fin) =>
+                  poll(exclusive.surround {
+                    swapFinalizer(Acquired(r, fin)).uncancelable.as(r)
+                  }).onCancel(fin)
```
armanbilge: This could lead to the finalizer being called twice I think. We only want to run it if acquiring the lock was canceled. Once we have the lock and we store the finalizer, even if we are canceled somewhere afterwards we no longer need to call it.

tests/shared/src/test/scala/cats/effect/std/HotswapSpec.scala
```diff
@@ -104,6 +105,15 @@ class HotswapSpec extends BaseSpec { outer =>
 
       go must completeAs(())
     }
+
+    "not block current resource while swap is instantiating new one" in ticked {
+      implicit ticker =>
+        val go = Hotswap.create[IO, Unit].use { hs =>
+          hs.swap(Resource.eval(IO.sleep(1.minute) *> IO.unit)).start *>
+            hs.get.use_.timeout(1.second) *> IO.unit
```
armanbilge: We should add a small sleep between these steps so that the `swap` fiber definitely `start`s before the `get`.

Also I realized another "bug", or undesirable behavior: `get` acquires a shared lock even if there is currently no resource (as is the case in this test) and it returns `None`. This is sub-optimal since it can block installation of a new resource for no reason.

armanbilge (commented):
Can we add another test that `swap` is cancelable if while waiting for a `get` to release the old resource and also that the finalizer is run in this case?

tests/shared/src/test/scala/cats/effect/std/HotswapSpec.scala
tibetansingingcat: Right, `get` could be called and some long-running operation could be happening despite the Resource containing a None and `swap` is blocked because of it. Seems unlikely that it would happen in practice but I'll think about how it could be addressed in this PR.

tests/shared/src/test/scala/cats/effect/std/HotswapSpec.scala
```diff
@@ -104,6 +105,32 @@ class HotswapSpec extends BaseSpec { outer =>
...
+    "successfully cancel during swap and run finalizer if cancelation is requested while waiting for get to release" in ticked {
+      implicit ticker =>
+        val go = (for {
+          log <- Ref.of[IO, List[String]](List()).toResource
+          (hs, _) <- Hotswap[IO, Unit](logged(log, "a"))
+          _ <- hs.get.evalMap(_ => IO.sleep(1.minute)).use_.start.toResource
+          _ <- IO.sleep(3.seconds).toResource
+          swapFib <- hs.swap(logged(log, "b")).start.toResource
+          _ <- IO.sleep(3.seconds).toResource
+          _ <- swapFib.cancel.toResource
```
armanbilge: Instead of starting and canceling I think we can use a timeout like you did for your last test.

Also, sorry, completely my pet peeve 😅 ... but I personally really dislike sequential programs written in `Resource`. Instead can we `Hotswap[...].use { ... ` and run the entire test inside of that? Like you did for the other test. Then we don't need all this `toResource` stuff.

tests/shared/src/test/scala/cats/effect/std/HotswapSpec.scala
```diff
@@ -104,6 +105,32 @@ class HotswapSpec extends BaseSpec { outer =>
 
       go must completeAs(())
     }
+
+    "not block current resource while swap is instantiating new one" in ticked {
+      implicit ticker =>
+        val go = Hotswap.create[IO, Unit].use { hs =>
+          hs.swap(Resource.eval(IO.sleep(1.minute) *> IO.unit)).start *>
```
armanbilge: Nit

```suggestion
          hs.swap(IO.sleep(1.minute).toResource).start *>
```

armanbilge (commented):
Tests are looking good, sorry just some style nits 😅

tests/shared/src/test/scala/cats/effect/std/HotswapSpec.scala
tibetansingingcat: So I thought about doing this None-blocks-swap improvement but I might prefer to create an issue for it rather than attempt it here.

This was what I came up with, I wasn't sure how to avoid expressing the method with `.toResource` while still being able to add additional finalization if we haven't already released the permit.

```scala
          override def get: Resource[F, Option[R]] = for {
            _ <- semaphore.acquire.toResource
            currState <- state.get.toResource
            res <-
              currState match {
                case Acquired(r, _) => Some(r).pure[F].toResource.onFinalize(semaphore.release)
                case _ => semaphore.release.as(None).toResource
              }
          } yield res
```

There also could be some sort of issue with this that I'm not realizing.

I've added an additional test,

```scala
    "not block swap if hotswap is empty" in ticked {
      implicit ticker =>
        val go = Hotswap.create[IO, Boolean].use { hs =>
          for {
            _ <- hs.get.evalMap(o => IO.sleep(1.minute)).use_.start
            _ <- IO.sleep(5.seconds)
            next = Resource.eval(IO.pure(true))
            _ <- hs.swap(next).timeout(10.seconds)
          } yield ()
        }

        go must completeAs(())
    }
```

I haven't pushed these changes because if I'm way off base I would rather just merge the PR as it is and file an issue for improving this behavior.

Let me know if I happen to be on the right track and I'll push these additions.

armanbilge: > I might prefer to create an issue for it rather than attempt it here.
> I would rather just merge the PR as it is and file an issue for improving this behavior.

I agree, this is probably better :) I didn't take a detailed look yet, will circle back later. Thanks for all your work!

tibetansingingcat: No problem! Sounds good, thank you so much for all your guidance and your patience as I found my way to a fix. I greatly appreciate being afforded the space and assistance to learn how to solve it.

I went ahead and created an issue https://github.com/typelevel/cats-effect/issues/3802 for the remaining improvement :)

std/shared/src/main/scala/cats/effect/std/Hotswap.scala
```diff
@@ -131,12 +131,12 @@ object Hotswap {
...
-              F.uncancelable { poll =>
-                poll(next.allocated).flatMap {
-                  case (r, fin) =>
-                    swapFinalizer(Acquired(r, fin)).as(r)
-                }
+            F.uncancelable { poll =>
+              poll(next.allocated).flatMap {
+                case (r, fin) =>
+                  poll(exclusive.onCancel(Resource.eval(fin)).surround {
+                    swapFinalizer(Acquired(r, fin)).uncancelable.as(r)
```
armanbilge: I think there is a cancelation bug here. If we are canceled after `exclusive.onCancel(...)` and before `swapFinalizer` then `fin` is lost. I just push a test in d528afbd95078ba7e8c55ef79efbe3095cbbf63f that demonstrates the problem.

armanbilge: I wonder if something like this can work.

```suggestion
                  exclusive.mapK(poll).onCancel(Resource.eval(fin)).surround {
                    swapFinalizer(Acquired(r, fin)).as(r)
```

tests/shared/src/test/scala/cats/effect/std/HotswapSpec.scala
```diff
@@ -104,6 +107,46 @@ class HotswapSpec extends BaseSpec { outer =>
...
+
+        go must completeAs(List("open a", "open b", "close b"))
+    }
+
+    "swap is safe to concurrent cancelation" in ticked { implicit ticker =>
+      val go = IO.ref(false).flatMap { open =>
+        Hotswap[IO, Unit](Resource.unit)
+          .use {
+            case (hs, _) =>
+              hs.swap(Resource.make(open.set(true))(_ => open.set(false)))
```
armanbilge: Even better, let's make sure we are only closing it if it is open ...

```suggestion
              hs.swap(Resource.make(open.set(true))(_ => open.getAndSet(false).map(_ should beTrue).void))
```

armanbilge (dismissed):
Nice catch and excellent work!

tibetansingingcat:
> Nice catch and excellent work!

Thanks! Didn't want this PR to get lost to the sands of time. Thank you for those tweaks! That use of `.mapK` was enlightening to me.

tests/shared/src/test/scala/cats/effect/std/HotswapSpec.scala
```diff
@@ -104,6 +107,47 @@ class HotswapSpec extends BaseSpec { outer =>
 
       go must completeAs(())
     }
+
+    "not block current resource while swap is instantiating new one" in ticked {
+      implicit ticker =>
+        val go = Hotswap.create[IO, Unit].use { hs =>
+          hs.swap(IO.sleep(1.minute).toResource).start *>
+            IO.sleep(5.seconds) *>
+            hs.get.use_.timeout(1.second) *> IO.unit
```
durban:
```suggestion
            hs.get.use_.timeout(1.second).void
```

durban (dismissed):
LGTM. I've left a comment, but it's really not that important.

armanbilge (approved):

Issue #3806, opened 2023-09-02
Add `def warnOnNonMainThreadDetected: Boolean` configuration to `IOApp`


armanbilge:
Follow-up to:
- https://github.com/typelevel/cats-effect/pull/3774#discussion_r1313875233

diogocanut:
I want to take this one if not taken

armanbilge:
@diogocanut thanks!

diogocanut:
Should we have the same logic as in `onNonMainThreadDetected`, like:

```scala
def warnOnNonMainThreadDetected: Boolean =
  Option(System.getProperty("cats.effect.warnOnNonMainThreadDetected"))
    .map(_.equalsIgnoreCase("true"))
    .getOrElse(true)
```

or default it as disabled?

```scala
java.lang.Boolean.getBoolean("cats.effect.warnOnNonMainThreadDetected")
```

armanbilge:
I think the default should be enabled.

armanbilge:
- https://github.com/typelevel/cats-effect/pull/3815

PR #3814, opened 2023-09-08
Add warnOnNonMainThreadDetected configuration to IOApp

diogocanut:
Solves #3806

PR #3815, opened 2023-09-09
Add `warnOnNonMainThreadDetected` configuration to `IOApp`

diogocanut:
Solves #3806


armanbilge:
Thanks, but this is missing the other changes from https://github.com/typelevel/cats-effect/pull/3814 where you actually use `warnOnNonMainThreadDetected` :) everything you did there was perfect, we just need to target it at this branch.

armanbilge:
Oh of course 🤦 that's why you couldn't make the change, because we didn't merge `series/3.5.x` branch into `series/3.x` branch yet. Sorry I forgot about that. We should wait until that happens, then this PR makes sense.

armanbilge:
@diogocanut we just merged `series/3.5.x` into `series/3.x` so you should be able to finish this PR now. Sorry for my mistake before!

diogocanut:
@armanbilge done, thanks for noticing.

core/jvm/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -337,12 +337,14 @@ trait IOApp {
    * isn't the main process thread. This condition can happen when we are running inside of an
    * `sbt run` with `fork := false`
    */
+
```
armanbilge: Stray space?
```suggestion
```

diogocanut: ops, removed.

Daenyth:
bump

armanbilge (approved):

PR #3816, opened 2023-09-10
add explicit type

xuwei-k:
prepare Scala 2.13.12

- https://github.com/scala/scala/pull/10439
- https://github.com/scala/community-build/blob/12866dc60bf80ea569f8a7e4d229a8afe04d84a1/proj/cats-effect.conf#L15-L16

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -1223,7 +1223,7 @@ private[effect] trait ResourceHOInstances0 extends ResourceHOInstances1 {
   implicit def catsEffectAsyncForResource[F[_]](implicit F0: Async[F]): Async[Resource[F, *]] =
     new ResourceAsync[F] {
       def F = F0
-      override def applicative = this
+      override def applicative: ResourceAsync[F] = this
     }
 
   implicit def catsEffectSemigroupKForResource[F[_], A](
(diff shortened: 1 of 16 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -1223,7 +1223,7 @@ private[effect] trait ResourceHOInstances0 extends ResourceHOInstances1 {
   implicit def catsEffectAsyncForResource[F[_]](implicit F0: Async[F]): Async[Resource[F, *]] =
     new ResourceAsync[F] {
       def F = F0
-      override def applicative = this
+      override def applicative: ResourceAsync[F] = this
```
armanbilge: I'm not sure if it matters, but can this be:
```suggestion
      override def applicative: Async[Resource[F, *]] = this
```

armanbilge (commented):
Thanks! Just one question.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
xuwei-k:
```
typelevel/cats-effect/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala:1226:20: incompatible type in overriding
[error] override def applicative: cats.effect.kernel.ResourceAsync[F] (defined in class ResourceAsync);
[error]  found   : cats.effect.kernel.Async[[β$26$]cats.effect.kernel.Resource[F,β$26$]]
[error]  required: cats.effect.kernel.ResourceAsync[F]
[error]       override def applicative: Async[Resource[F, *]] = this
[error]                    ^
[error] one error found
[error] (Compile / compileIncremental) Compilation failed
```

🤔

armanbilge: Thanks. I was surprised some of the other ones could use the more general type. 🤔

PR #3821, opened 2023-09-11
Allocate one `parasitic` EC per `Dispatcher`

armanbilge:
h/t @kamilkloch for pointing this out

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
@@ -75,16 +75,16 @@ trait Dispatcher[F[_]] extends DispatcherPlatform[F] {
     }
   }
 
+  private[this] lazy val parasitic: ExecutionContext = new ExecutionContext {
+    def execute(runnable: Runnable) = runnable.run()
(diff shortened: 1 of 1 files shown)


std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -75,16 +75,16 @@ trait Dispatcher[F[_]] extends DispatcherPlatform[F] {
     }
   }
 
+  private[this] lazy val parasitic: ExecutionContext = new ExecutionContext {
```
armanbilge, line 78: Adding a `val` to a `trait` is not backwards-compatible.

Technically adding a `lazy val` is also not backwards-compatible ... but it _is_ backwards binary-compatible. The catch is that clients who don't recompile will get semantics of `def` instead of `lazy val` (this has burned us before in `IOApp`). But in this particular case that's actually equivalent to the current status quo.

armanbilge:
Wait, this is dumb. One second 😂

armanbilge:
Aha, https://github.com/typelevel/cats-effect/pull/3822 was much smarter than me. Closing!

PR #3822, opened 2023-09-11
Optimize `Dispatcher#{unsafeRunAsync,unsafeRunAndForget}`

kamilkloch:
It looks like the parasitic EC need not be instantiated on every `unsafeRunAsync` call.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
@@ -70,17 +70,15 @@ trait Dispatcher[F[_]] extends DispatcherPlatform[F] {
    * Submits an effect to be executed with fire-and-forget semantics.
    */
   def unsafeRunAndForget[A](fa: F[A]): Unit =
-    unsafeToFutureCancelable(fa)
-      ._1
-      .onComplete {
-        case Failure(ex) => ex.printStackTrace()
-        case _ => ()
-      }(parasiticEC)
+    unsafeToFuture(fa).onComplete {
+      case Failure(ex) => ex.printStackTrace()
+      case _ => ()
+    }(parasiticEC)
 
   // package-private because it's just an internal utility which supports specific implementations
   // anyone who needs this type of thing should use unsafeToFuture and then onComplete
   private[std] def unsafeRunAsync[A](fa: F[A])(cb: Either[Throwable, A] => Unit): Unit =
-    unsafeToFutureCancelable(fa)._1.onComplete(t => cb(t.toEither))(parasiticEC)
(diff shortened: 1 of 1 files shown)


std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -68,28 +68,28 @@ trait Dispatcher[F[_]] extends DispatcherPlatform[F] {
...
-    // this is safe because the only invocation will be cb
-    implicit val parasitic: ExecutionContext = new ExecutionContext {
-      def execute(runnable: Runnable) = runnable.run()
-      def reportFailure(t: Throwable) = t.printStackTrace()
-    }
-
-    unsafeToFuture(fa).onComplete(t => cb(t.toEither))
-  }
+  private[std] def unsafeRunAsync[A](fa: F[A])(cb: Either[Throwable, A] => Unit): Unit =
+    unsafeToFutureCancelable(fa)._1.onComplete(t => cb(t.toEither))(parasiticEC)
```
armanbilge: Isn't this just inlining `unsafeToFuture`?

```suggestion
    unsafeToFuture(fa).onComplete(t => cb(t.toEither))(parasiticEC)
```

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -68,28 +69,28 @@ trait Dispatcher[F[_]] extends DispatcherPlatform[F] {
...
-      case Right(_) => ()
-    }
-  }
+  def unsafeRunAndForget[A](fa: F[A]): Unit =
+    unsafeToFutureCancelable(fa)
+      ._1
+      .onComplete {
+        case Failure(ex) => ex.printStackTrace()
+        case _ => ()
+      }(parasiticEC)
```
armanbilge: Same thing here ... isn't this just inlining `unsafeRunAsync`? Instead of repeating the implementation let's just use that.

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
kamilkloch: It is. If the speed up is negligible, let's skip it.

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
kamilkloch: Isn't it an extra allocation of `Either`?

armanbilge: Oh I see, good catch. Well at least let's use `unsafeToFuture` here as well.

armanbilge (approved):
Great catches 👍

PR #3829, opened 2023-09-15
Improve `MapRef` docs

BalmungSan:


std/shared/src/main/scala/cats/effect/std/MapRef.scala
```diff
@@ -26,8 +26,12 @@ import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.atomic.AtomicBoolean
 
 /**
- * This is a total map from K to Ref[F, V]. This allows us to use the Ref API backed by a
- * ConcurrentHashMap or similar.
+ * This is a total map from `K` to `Ref[F, V]`.
+ *
+ * It is conceptually similar to a `Ref[F, Map[K, V]]`, but with better ergonomics when working
+ * on a per key basis. It also has less contention since the key-value pairs are sharded; thus
```
armanbilge: This is not true in general, it depends on the implementation. A `MapRef` is just an interface, sharding is an implementation detail. Although Fabio points out we can simplify the construction as well.

docs/std/mapref.md
```diff
@@ -0,0 +1,60 @@
...
+      shardCount = 5 // Arbitrary number of shards just for demonstration.
+    ).map { mapRef =>
+      new DatabaseClient[IO] {
+        override def getDataById(id: Id): F[Option[Data]] =
+          mapRef(id).get
+
+        override def upsertData(id: Id, data: Data): F[Unit] =
+          mapRef(id).update(_ => Some(data))
+      }
+    }
```
amumurst:
```suggestion
      new DatabaseClient[IO] {
        override def getDataById(id: Id): IO[Option[Data]] =
          mapRef(id).get

        override def upsertData(id: Id, data: Data): IO[Unit] =
          mapRef(id).update(_ => Some(data))
      }
    }
```

docs/std/mapref.md
```diff
@@ -0,0 +1,60 @@
...
+
+It is conceptually similar to a `Ref[F, Map[K, V]]`,
+but with better ergonomics when working on a per key basis.
+Note, however, that it does not support atomic updates to multiple keys.
+
+Additionally, some implementations also provide less contention,
+since all operations are performed over individual key-value pairs,
+those could be sharded.
+Thus, multiple concurrent updates may be executed independently to each other;
+as long as their keys belong to different shards.
```
armanbilge:
```suggestion
Additionally, some implementations also provide less contention:
since all operations are performed on individual key-value pairs,
the pairs can be sharded by key.
Thus, multiple concurrent updates may be executed independently to each other,
as long as their keys belong to different shards.
```

docs/std/mapref.md
```diff
@@ -0,0 +1,60 @@
...
+
+### In-Memory database
+
+This is probably one of the most common uses of this tool.
```
armanbilge:
```suggestion
This is probably one of the most common uses of this datatype.
```

docs/std/mapref.md
```diff
@@ -0,0 +1,60 @@
...
+
+```scala mdoc:reset:silent
+//> using lib "org.typelevel::cats-effect:3.4.11"
```
armanbilge:
```suggestion
//> using lib "org.typelevel::cats-effect:3.5.1"
```

armanbilge (dismissed):
Thanks for this addition! This was much overdue 😅

djspiewak:
The build failures are legitimate

armanbilge (approved):

PR #3841, opened 2023-09-21
Acquire exclusive lock before finalizing `Hotswap`

### LINKED ISSUE #3840: Hotswap#get scaladoc/impl mismatch
The scaladoc of `Hotswap#get` says that "_The returned resource is guaranteed to be available for the duration of the returned resource_". I don't think this is true: if the whole `Hotswap` is finalized, the "returned resource" will be finalized too (here: https://github.com/typelevel/cats-effect/blob/a6c5b155be3b729af49eb26204e101f2d527fa80/std/shared/src/main/scala/cats/effect/std/Hotswap.scala#L116).

I think either the scaladoc or the implementation should be changed. (Probably the scaladoc, although I've never used `Hotswap`.)

armanbilge:
Fixes https://github.com/typelevel/cats-effect/issues/3840.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
@@ -117,13 +117,12 @@ object Hotswap {
       def initialize: F[Ref[F, State]] =
         F.ref(Cleared)
 
-      def finalize(state: Ref[F, State]): F[Unit] = exclusive.surround {
+      def finalize(state: Ref[F, State]): F[Unit] =
         state.getAndSet(Finalized).flatMap {
-          case Acquired(_, finalizer) => finalizer
+          case Acquired(_, finalizer) => exclusive.surround(finalizer)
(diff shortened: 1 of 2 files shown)


std/shared/src/main/scala/cats/effect/std/Hotswap.scala
```diff
@@ -111,22 +117,17 @@ object Hotswap {
...
         F.ref(Cleared)
 
-      def finalize(state: Ref[F, State]): F[Unit] =
+      def finalize(state: Ref[F, State]): F[Unit] = exclusive.surround {
         state.getAndSet(Finalized).flatMap {
           case Acquired(_, finalizer) => finalizer
           case Cleared => F.unit
           case Finalized => raise("Hotswap already finalized")
         }
+      }
```
armanbilge: Actually we should probably keep this more tightly scoped.

djspiewak (approved):

PR #3842, opened 2023-09-22
Run Cirrus only for labeled PRs or tags

armanbilge:


samspills (approved):
🙌

djspiewak (dismissed):
Can we also make them run unconditionally for release tags? Also we should probably create the labels. :-)

.cirrus.yml
```diff
@@ -1,4 +1,6 @@
 jvm_highcore_task:
+  only_if: $CIRRUS_TAG != '' || $CIRRUS_PR != ''
```
armanbilge, line 2: @djspiewak they already run unconditionally for release tags. Unless I'm missing your point.

djspiewak: I misunderstood the diff. lgtm then

PR #3847, opened 2023-09-25
Try quoting labels in release notes config

armanbilge:
Not sure if this will fix it, but something seems to be broken 😕 

```
Could not parse .github/release.yml: Tried to load unspecified class: Symbol
```

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -2,14 +2,14 @@ changelog:
   categories:
     - title: Features
       labels:
-        - :mushroom: enhancement
+        - ':mushroom: enhancement'
     - title: Bug Fixes
       labels:
-        - :beetle: bug
+        - ':beetle: bug'
     - title: Behind the Scenes
       labels:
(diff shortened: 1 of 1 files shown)


armanbilge:
Nice, that fixed it :)

.github/release.yml
```diff
@@ -2,14 +2,14 @@ changelog:
   categories:
     - title: Features
       labels:
-        - :mushroom: enhancement
+        - ':mushroom: enhancement'
```
djspiewak, line 5: Oh yeah, the problem here is the `:`. Quoting is the correct fix.

armanbilge: Other YAML parsers like it just fine. /shrug

djspiewak: I'm pretty sure that's out of spec, but I'd have to check. My recollection is that any unquoted string which contains a non-leading `:` should be considered a key/value pair.

PR #3711, opened 2023-06-24
Added script for automatically posting release to discord

djspiewak:

PR #3855, opened 2023-09-26
Post release to Discord in CI

armanbilge:
Follow-up to https://github.com/typelevel/cats-effect/pull/3711 that uses the configured secret.

### DIFF
--- a/scripts/make-release-prs.sh
+++ b/scripts/make-release-prs.sh
@@ -7,12 +7,12 @@ cd "$(dirname $0)/.."
 primary_base="$(pwd)"
 
 if [[ $# -ne 2 ]] || [[ "$1" == "--help" ]]; then
-  echo "usage: $0 old-version new-version"
+  echo "usage: $0 old-tag new-tag"
   exit 1
 fi
 
-old_version="$1"
-new_version="$2"
+old_version="${1#v}"
+new_version="${2#v}"
 
 minor_base=series/$(echo $new_version | sed -E 's/([0-9]+).([0-9]+).[0-9]+/\1.\2.x/')
 major_base=series/$(echo $new_version | sed -E 's/([0-9]+).[0-9]+.[0-9]+/\1.x/')
--- a/scripts/post-release-discord.sh
+++ b/scripts/post-release-discord.sh
@@ -14,6 +14,6 @@ fi
 #
 # this url should be considered a secret and handled with appropriate care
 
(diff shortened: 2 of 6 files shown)


scripts/post-release-discord.sh
```diff
@@ -14,6 +14,6 @@ fi
 #
 # this url should be considered a secret and handled with appropriate care
 
-data="{\"content\":\"https://github.com/typelevel/cats-effect/releases/tag/v$1\"}"
+data="{\"content\":\"https://github.com/typelevel/cats-effect/releases/tag/$1\"}"
```
armanbilge, line 17: Seems easier to just pass the tag.

djspiewak: Well, the reason I *didn't* do it that way is because all the other scripts take a version number unprefixed by `v`, and it's going to get very confusing if we make it inconsistent.

djspiewak (changes requested):
Definitely easier *here* to just pass tag, but it's also inconsistent with the other scripts.

armanbilge:
Presumably we eventually want to run the other scripts in CI too? We're just going to butt heads with this again. It's easier to pass a tag in GHA than it is to do string manipulation.

Meanwhile, the scripts themselves can easily strip the `v` prefix.

djspiewak:
> Presumably we eventually want to run the other scripts in CI too? We're just going to butt heads with this again. It's easier to pass a tag in GHA than it is to do string manipulation.

That's fair. Can we change the other scripts then at the same time? This is exactly the sort of stuff that gets *really* confusing if we aren't strict about it.

scripts/make-release-prs.sh
```diff
@@ -7,12 +7,12 @@ cd "$(dirname $0)/.."
...
 if [[ $# -ne 2 ]] || [[ "$1" == "--help" ]]; then
-  echo "usage: $0 old-version new-version"
+  echo "usage: $0 old-tag new-tag"
   exit 1
 fi
 
-old_version="$1"
-new_version="$2"
+old_version="${1#v}"
+new_version="${2#v}"
```
djspiewak, line 15: Oooooh, fancy.

djspiewak (approved):

scripts/make-release-prs.sh
armanbilge: Fancy? I feel dirty 😩

PR #3857, opened 2023-09-28
Update versions for 3.5.2

armanbilge:


docs/std/mapref.md
```diff
@@ -32,7 +32,7 @@ as long as their keys belong to different shards.
 This is probably one of the most common uses of this datatype.
 
 ```scala mdoc:reset:silent
-//> using lib "org.typelevel::cats-effect:3.5.1"
+//> using lib "org.typelevel::cats-effect::3.5.2"
```
armanbilge, line 35: The update script is still missing some of these.

PR #3859, opened 2023-09-28
Release script improvements

armanbilge:


scripts/make-release-prs.sh
```diff
@@ -20,7 +20,7 @@ minor_branch="release/$new_version-minor"
 major_branch="release/$new_version-major"
 
 cd "$(mktemp -d)"
-git clone git@github.com:typelevel/cats-effect.git
+gh repo clone typelevel/cats-effect
```
armanbilge, line 23: ssh cloning is BESPOKE 😛

scripts/update-versions.sh
```diff
@@ -23,3 +23,4 @@ new_version="$2"
 # the quote reduce the false positive rate
 find . -type f -name '*.md' -exec perl -pi -e "s/\"$old_version\"/\"$new_version\"/g" {} \;
 find . -type f -name '*.md' -exec perl -pi -e "s/\\*\\*$old_version\\*\\*/\\*\\*$new_version\\*\\*/g" {} \;
+find . -type f -name '*.md' -exec perl -pi -e "s/:$old_version/:$new_version/g" {} \;
```
djspiewak, line 26: We should also make sure we've caught any stray places where there's an old old version, otherwise this won't work going forward.

djspiewak (approved):

scripts/update-versions.sh
armanbilge: Yep, fixed in 7dfe45de87ba0da7b9f760133a0e79311db1db5a.

PR #3864, opened 2023-10-04
Check supervisor status before submitting a task

### LINKED ISSUE #3394: Closed `Supervisor` strangeness

```scala

  def run: IO[Unit] = {
    }.flatMap { sup =>
      )
    }
  }
}
```

This prints:

```
started 1
cancelled 1
started 2
done
```

alexandrustana:
Closes #3394

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Supervisor.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Supervisor.scala
@@ -227,6 +227,10 @@ object Supervisor {
           }
 
           for {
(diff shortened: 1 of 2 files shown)


armanbilge:
Thanks for the PR! Would you mind re-targeting to `series/3.5.x`, since it's a bug fix?

alexandrustana:
I've used the wrong base branch 😅 I'll recreate the MR from `series/3.5.x` as suggested.

PR #3865, opened 2023-10-05
Check supervisor status before submitting a task

### LINKED ISSUE #3394: Closed `Supervisor` strangeness

alexandrustana:
Fixes #3394 

Tried to encapsulate the `status` of the `Supervisor.State` within itself. This way, whenever `joinAll` or `cancellAll` are called, the `Supervisor` will be considered closed and not accept any new tasks. 

Since supervising a task isn't an atomic operation, 2 checks have been added to ensure that the `Supervisor` is not closed. One check right before running the `Fiber` and one check after the task has been registered (which also implies cleanup).

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Supervisor.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Supervisor.scala
@@ -23,6 +23,7 @@ import cats.syntax.all._
 import scala.collection.mutable.ListBuffer
 
 import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.atomic.AtomicBoolean
 
 /**
  * A fiber-based supervisor that monitors the lifecycle of all fibers that are started via its
@@ -144,6 +145,7 @@ object Supervisor {
     // run all the finalizers
     val joinAll: F[Unit]
     val cancelAll: F[Unit]
+    val isClosed: F[Boolean]
   }
 
   private def supervisor[F[_]](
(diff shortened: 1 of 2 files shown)


std/shared/src/main/scala/cats/effect/std/Supervisor.scala
```diff
@@ -242,19 +257,25 @@ object Supervisor {
...
-          stateRef.get.map(_.values.toList)
-
-        val joinAll: F[Unit] = allFibers.flatMap(_.traverse_(_.join.void))
-        val cancelAll: F[Unit] = allFibers.flatMap(_.parUnorderedTraverse(_.cancel).void)
-      }
-    }
+    val mkState = F
+      .ref(false)
+      .flatMap(closed =>
+        F.ref[Map[Unique.Token, Fiber[F, Throwable, _]]](Map.empty).map { stateRef =>
```
armanbilge, line 263: Instead of having two refs here, we can just reuse the `Ref` holding the `Map` and set it to `null` when it is closed. Then there is no race condition, since there is a single source-of-truth for the status of the `Supervisor`.

djspiewak:
Is this superseded by #3972?

durban:
Hm, yes, I think it is.

PR #3870, opened 2023-10-09
Block in-place if running on a virtual thread

### LINKED ISSUE #3869: Virtual Threads (Project Loom) in IO.blocking / IO.interruptible
Hello,

Right now, Cats-Effect uses a cached thread pool for blocking I/O, used in operations like `IO.blocking`, `IO.interruptible` and `IO.interruptibleMany`. It would be great if, when Cats-Effect ran on top of Java 21, that thread pool would switch to usage of virtual threads. Personally, I'd prefer for this to just work out of the box. Users that upgrade to Java 21 should just get the benefits without configuring `IOApp`.

The biggest win isn't performance, but *correctness*. Some limitations are lifted for virtual threads, such as network socket operations becoming interruptible. So for example, when we have a JDBC query executed via `IO.interruptible`, the interruption will just work. Read the details from here: https://openjdk.org/jeps/444#Networking

The challenge is in creating an implementation that is able to use virtual threads when available. Currently, Cats-Effect is not compiled for JDK 21, and probably won't be for some time. Therefore, a solution would have to either:

1. use the reflection API to call `Thread.ofVirtual` when available (e.g., MethodHandle), otherwise fallback on a cached thread-pool, or;
2. create a micro-library that builds a [multi-release JAR](https://www.baeldung.com/java-multi-release-jar), but it's not a widely used approach; libraries like CE most often choose reflection.

What do you think? 
I can follow up with a PR if you agree to try it out.

armanbilge:
Closes https://github.com/typelevel/cats-effect/issues/3869, h/t Daniel for the idea.

If a fiber detects that it is running on a virtual thread then it will invoke a blocking operation in-place (similar to the `scala.concurrent.blocking` mechanism). This allows to do stuff like:

```scala
IO.interruptible(...).evalOnExecutor(Executors.newVirtualThreadPerTaskExecutor())
```

This PR is targeted at `series/3.x` because we are pinned to Scala 3.2 in `series/3.5.x` and that doesn't support JDK 21. We could backport the implementation changes here to `3.5.x` but we cannot test them.

@alexandru wdyt?


tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
```diff
@@ -543,7 +543,10 @@ trait IOPlatformSpecification extends DetectPlatform { self: BaseSpec with Scala
           val loomEc = ExecutionContext.fromExecutor(loomExec)
 
           IO.blocking {
-            Thread.currentThread().asInstanceOf[{ def isVirtual(): Boolean }].isVirtual()
+            classOf[Thread]
+              .getDeclaredMethod("isVirtual")
+              .invoke(Thread.currentThread())
+              .asInstanceOf[Boolean]
```
He-Pin, line 549: why not used method handle too?

armanbilge: This is test code.

wb14123:
Will `evalOnExecutor` shift both blocking and unblocking IOs to the virtual threads pool?

armanbilge:
@wb14123 yes. To be more precise: 
1. `evalOn` (and `evalOnExecutor`) modify the compute execution context for a task, but not the blocking EC.
2. After the changes in this PR, a fiber running on a virtual thread will run blocking operations in-place i.e. on the same virtual thread that it is already running on.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -996,6 +996,20 @@ private final class IOFiber[A](
               } else {
                 blockingFallback(cur)
               }
+            } else if (isVirtualThread(Thread.currentThread())) {
+              var error: Throwable = null
+              val r =
+                try {
+                  cur.thunk()
```
wb14123, line 1003: Shall we also use `scala.concurrent.blocking(cur.thunk())` here so that if `thunk()` is an operation that blocks the OS thread, JVM can spawn another OS thread? If it doesn't block the OS thread, since `val ec = currentCtx` is virtual thread pool, JVM will still use the same OS thread.

armanbilge: `scala.concurrent.blocking` doesn't do anything for virtual threads, because virtual threads do not extend [`scala.concurrent.BlockContext`](https://www.scala-lang.org/api/2.13.12/scala/concurrent/BlockContext.html) :)

> if `thunk()` is an operation that blocks the OS thread, JVM can spawn another OS thread

The entire point of using virtual threads is that the JVM will manage all of this for you. [Further reading.](https://docs.oracle.com/en/java/javase/21/core/virtual-threads.html#GUID-704A716D-0662-4BC7-8C7F-66EE74B1EDAD)

armanbilge: Hmm, maybe I am wrong about that 🤔 

> The scheduler does not compensate for pinning by expanding its parallelism

https://openjdk.org/jeps/425

wb14123: Yeah I think it only creates a virtual thread if you submit the task to virtual thread executor. So maybe something like `currentCtx.execute(cur.thunk)` (suppose currentCtx is the virtual thread executor)

armanbilge: Oh, I read the wrong section.

> However, some blocking operations in the JDK do not unmount the virtual thread, and thus block both its carrier and the underlying OS thread. This is because of limitations either at the OS level (e.g., many filesystem operations) or at the JDK level (e.g., [`Object.wait()`](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/lang/Object.html#wait())). The implementation of these blocking operations will compensate for the capture of the OS thread by temporarily expanding the parallelism of the scheduler. Consequently, the number of platform threads in the scheduler's `ForkJoinPool` may temporarily exceed the number of available processors.

It's very confusing. But the point is that the JVM is handling it 😅

armanbilge: > Yeah I think it only creates a virtual thread if you submit the task to virtual thread executor. So maybe something like `currentCtx.execute(cur.thunk)` (suppose currentCtx is the virtual thread executor)

Hmm, there seems to be some confusion. If we reach this branch, we are _already_ running on a virtual thread, because this fiber has been submitted to the virtual thread executor e.g. via 
```scala
.evalOnExecutor(Executors.newVirtualThreadPerTaskExecutor())
```

wb14123: Yes but this virtual thread is shared with other IOs, right? Like the computing pool.

So if `cur.thunk` is a potential real blocking operation, you need to create another virtual thread for it so JVM can put it on another OS thread to unblock other IOs if needed.

armanbilge: > Yes but this virtual thread is shared with other IOs, right? Like the computing pool.

Do you mean, this virtual thread is shared with other fibers? It's not actually.

If we check the docs for `Executors.newVirtualThreadPerTaskExecutor()`:

> Creates an Executor that starts a new virtual Thread for each task. The number of threads created by the Executor is unbounded.

armanbilge: Ah, I see your point now 😕 

It's possible that someone builds an `ExecutionContext` based on virtual threads, that is actually not using `newVirtualThreadPerTaskExecutor`. In that case, yes, blocking the thread could cause a problem. Hmm 🤔 

Thanks for raising this, sorry it took me so long to catch on.

wb14123: Even with `newVirtualThreadPerTaskExecutor`, if I have this code:

```
val myApp = IO.blocking(blockingOp) >> IO(nonblockingOp)
myApp.evalOn(Executors.newVirtualThreadPerTaskExecutor())
```

Is it possible that `blockingOp` and `nonblockingOp` share the same virtual thread? If not, where does it create a new task for each of the IO fiber?

> Creates an Executor that starts a new virtual Thread for each task

We need a task for it to starts a new virtual thread, right?

armanbilge: > Is it possible that `blockingOp` and `nonblockingOp` share the same virtual thread?

Yes they will share the same virtual thread. But that is not a problem: the second `IO` should not run until the first `IO` completes, so it does not matter if the virtual thread is blocked.

wb14123: Sorry my bad... I meant to run them in parallel: `IO.blocking(blockingOp) &> IO(nonblockingOp)`

armanbilge: In that case, an independent fiber will be created for each `IO`, and those fibers will be submitted to the execution context. If you are using `newVirtualThreadPerTaskExecutor` that means that each of those fibers will run on its own virtual thread.

wb14123: Oh I see, in this case it should be fine then, thanks for the patient explanation!

armanbilge, line 1003: Very important to call out: there is a big assumption here that if we are running on a virtual thread, the model being used is that each fiber is running on a dedicated virtual thread. While this _seems_ like the most sensible way to use virtual threads, it's not the only way.

For example, someone could build an execution context with a fixed number of virtual threads that poll from a shared task queue. In that case, blocking the virtual thread as we are doing here would be very problematic for them.

durban: > someone could build an execution context with a fixed number of virtual threads

I think that is looking for trouble. The documentation literally says "_Never Pool Virtual Threads_"...

core/jvm/src/main/java/cats/effect/IOFiberConstants.java
```diff
@@ -43,4 +47,28 @@ final class IOFiberConstants {
...
+
+  static {
+    final MethodHandles.Lookup lookup = MethodHandles.publicLookup();
+    final MethodType mt = MethodType.methodType(boolean.class);
+    MethodHandle mh;
+    try {
+      mh = lookup.findVirtual(Thread.class, "isVirtual", mt);
+    } catch (Throwable t) {
+      mh =
+          MethodHandles.dropArguments(
```
alexandru, line 69: Seems wasteful to produce a method handle that always returns `false` when invoked. Maybe I'm missing something, maybe this ends up as being inlined, and it doesn't matter, but I would see this method more like:

```java
  static boolean isVirtualThread(final Thread thread) {
    if (THREAD_IS_VIRTUAL_HANDLE != null)
        try {
          return (boolean) THREAD_IS_VIRTUAL_HANDLE.invokeExact(thread);
        } catch (Throwable t) {}
    return false;
  }
```

core/shared/src/main/scala/cats/effect/IOFiber.scala
alexandru: There might be cases in which limiting the number of possible virtual threads is desirable. For example, for limiting parallelism in certain instances. But I don't see the problem, because it would work *as the user intended*.

The only question is if we should integrate with `BlockContext`. I think we can integrate, as I don't think you necessarily need to inherit Threads from it. But for now, I don't think it's necessary; mostly because I can't think of any reason for wanting to limit virtual threads while allowing more threads in case of blocking operations.

alexandru: Indeed, the JVM is handling it. Also, some limits may be lifted in time, such as `Object.wait`.

armanbilge: > There might be cases in which limiting the number of possible virtual threads is desirable.

~~To be clear: there is actually no issue with limiting the number of virtual threads. The concern is if a virtual thread may be responsible for running more than one fiber. So long as every fiber has its own virtual thread, then there is no problem.~~

edit: I wrote that too late at night and it doesn't make sense, sorry :)

--- 

> I think we can integrate, as I don't think you necessarily need to inherit Threads from it.

You can also install the `BlockContext` as a thread-local, but making this work will add some allocations and overhead.

alexandru (commented):
Looking good 👍

core/jvm/src/main/java/cats/effect/IOFiberConstants.java
armanbilge: Yeah, I mean I have no idea without studying the assembly and/or benchmarking 😄 

In your version, we depend on the JVM optimizing away the conditional and branching. In my version there is no branching, just a static method call, which would ideally be inlined.

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -996,6 +996,20 @@ private final class IOFiber[A](
...
+                } catch {
+                  case t if NonFatal(t) =>
+                    error = t
+                  case t: Throwable =>
+                    onFatalFailure(t)
+                }
+
+              val next = if (error eq null) succeeded(r, 0) else failed(error, 0)
+              runLoop(next, nextCancelation, nextAutoCede)
```
alexandru, line 1012: The optimization is good to have in the current implementation, but there's something I don't like...

I would have expected the ability to override `runtime.blocking`, and `evalOn` doesn't do that. Also, `WorkStealingThreadPool` is special-cased, and this forces virtual threads to be special cased, too. And in both cases, `runtime.blocking` is ignored, which smells like a semantic issue that causes confusion.

An alternative would have been to look at the `runtime.blocking` thread-pool and notice that we are already on it. In which case no thread shift would be necessary.

armanbilge: > An alternative would have been to look at the `runtime.blocking` thread-pool and notice that we are already on it.

It's an interesting idea, but this would apply only to virtual threads anyway: this is the only pool on which it is safe to submit/run tasks whether they are blocking or not. There is no other `ExecutionContext` that can _directly_ handle both: the `WorkStealingThreadPool` and `ForkJoinPool` can handle blocking but only if it is appropriately wrapped in `scala.concurrent.blocking(...)` so in general it is not safe to submit arbitrary blocking tasks directly to these `ExecutionContext`s.

> And in both cases, `runtime.blocking` is ignored, which smells like a semantic issue that causes confusion.

I agree, it's confusing 😕 currently, `runtime.blocking` operates as a fallback, when you are not on the WSTP or a virtual thread.

> Also, `WorkStealingThreadPool` is special-cased, and this forces virtual threads to be special cased, too.

We could relax the special-casing a little bit by detecting if the current thread implements [`BlockContext`](https://www.scala-lang.org/api/2.13.12/scala/concurrent/BlockContext.html), in which case it would be safe to use the `blocking(...)` mechanism. Besides encompassing the WSTP this would allow us to support the FJP as well.

But it doesn't help with virtual threads which do not and cannot implement `BlockContext` so is out-of-scope for this PR.

tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
```diff
@@ -531,6 +532,25 @@ trait IOPlatformSpecification { self: BaseSpec with ScalaCheck =>
...
+            .invoke(null)
+            .asInstanceOf[ExecutorService]
+
+          val loomEc = ExecutionContext.fromExecutor(loomExec)
+
+          IO.blocking {
+            classOf[Thread]
+              .getDeclaredMethod("isVirtual")
+              .invoke(Thread.currentThread())
+              .asInstanceOf[Boolean]
```
djspiewak, line 549:
```suggestion
            Thread.currentThread().asInstanceOf[{ def isVirtual: Boolean }].isVirtual
```

Is a bit more concise :)

djspiewak (dismissed):
Ty for tackling this!

PR #3871, opened 2023-10-10
Add JDK 21 to CI matrix

armanbilge:
Trying to do this separately from https://github.com/typelevel/cats-effect/pull/3870.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -120,6 +120,7 @@ ThisBuild / crossScalaVersions := Seq(Scala3, Scala212, Scala213)
 ThisBuild / githubWorkflowScalaVersions := crossScalaVersions.value
 ThisBuild / tlVersionIntroduced := Map("3" -> "3.1.1")
 ThisBuild / tlJdkRelease := Some(8)
+ThisBuild / javacOptions += "-Xlint:-options" // --release 8 is deprecated on 21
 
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -135,11 +136,17 @@ ThisBuild / githubWorkflowPublishPreamble +=
 val OldGuardJava = JavaSpec.temurin("8")
 val LTSJava = JavaSpec.temurin("11")
 val LatestJava = JavaSpec.temurin("17")
+val LoomJava = JavaSpec.oracle("21")
```
durban, line 139: Is there no temurin 21? On adoptium.net there seems to be one...

durban (approved):

build.sbt
armanbilge: For some reason not yet, I tried that first and changed in https://github.com/typelevel/cats-effect/pull/3870/commits/aeda41c9c4c81a4f2429d240a3433a4eed35c77c.

vasilmkd: I haven't updated the index yet. It wasn't available as of couple of days ago when I last checked. I'll do it later tonight.

armanbilge: @vasilmkd actually we are not using the jdk index anymore 😅 

If it's changed recently let me check again.

PR #3881, opened 2023-10-23
Fix starvation demo app in tuning guide

armanbilge:
The demo actually wasn't starving because the fiber remained on the blocking thread it was shunted to for the preceding `IO.println`.

Further reading: https://typelevel.org/cats-effect/docs/faq#why-is-my-io-running-on-a-blocking-thread

h/t @felher

### DIFF
--- a/docs/core/starvation-and-tuning.md
+++ b/docs/core/starvation-and-tuning.md
@@ -29,7 +29,10 @@ import cats.syntax.all._
 object StarveThyself extends IOApp.Simple {
   val run = 
     0.until(100).toList parTraverse_ { i =>
-      IO.println(s"running #$i") >> IO(Thread.sleep(10000))
+      IO {
+        println(s"running #$i")
+        Thread.sleep(10000)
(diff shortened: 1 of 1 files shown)


docs/core/starvation-and-tuning.md
```diff
@@ -29,7 +29,10 @@ import cats.syntax.all._
 object StarveThyself extends IOApp.Simple {
   val run = 
     0.until(100).toList parTraverse_ { i =>
-      IO.println(s"running #$i") >> IO(Thread.sleep(10000))
+      IO {
+        println(s"running #$i")
+        Thread.sleep(10000)
+      }
```
armanbilge, line 35: Alternatively we could insert an `IO.cede` in-between the two `IO`s.

kamilkloch: Leaving out `IO.cede` keeps the example simpler for the broader public.

durban (approved):

PR #3890, opened 2023-11-08
Add a `memoizedAcquire` method to `Resource`

### LINKED ISSUE #3513: Add specialized signature of `memoize` to `Resource`
So currently `memoize` on `Resource` returns a `Resource[IO, Resource[IO, A]]`. The thing is, that inner resource is not really necessary (the implementation is just a `Resource.eval(...)`) and it makes the ergonomics a lot worse. The ideal signature would look more like `Resource[IO, IO[A]]`.

Besides `memoize`, I wonder if we can rescue some of the other method signatures as well, that return extremely hairy nested `Resource`s. It's pretty cool that `Resource` implements `Async` with all these bells-and-whistles, but the method signatures aren't always very ergonomic when working directly with concrete `Resource`.

Daenyth:
Ref #3513

I'm not sure how to unit test this exactly, I'm open to guidance or commits for that


kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -811,6 +811,24 @@ object Resource extends ResourceFOInstances0 with ResourceHOInstances0 with Reso
...
+     * The resource is released as normal at the end of `use` (whether normal termination, error, or cancelled), if it was acquired.
+     */
+    def memoizedAcquire(implicit F: Concurrent[F]): Resource[F, F[A]] =
+      // NB this uses syntax instead of an instance method because of variance on `A` within the class
+      Concurrent[Resource[F, *]].memoize[A](self).map {
+        case Resource.Eval(fa) => fa
+        case unexpected =>
+          throw new IllegalStateException(
+            s"Memoized Resource is not Resource.Eval: $unexpected")
+      }
```
armanbilge: Actually, we can avoid this silliness if we go the other way: put the juicy implementation in here, and then `memoize` can delegate to `memoizedAcquire` and wrap it up in `Resource.eval`.

armanbilge: Also then, the existing `memoize` tests seem like sufficient coverage to me.

SystemFw:
To test this I would write to Refs in your acquire and use, and assert on the state of those refs to check that it's behaving as you expect. I wouldnt worry about concurrent testing since memoize is pulling most of the weight

armanbilge:
I don't think we need to add any new tests specifically because of this PR, unless there are existing gaps in the existing `memoize` tests.

To be clear, what I'm proposing is that we refactor so it looks like this:
```scala
def memoizedAcquire(ra: Resource[F, A]): Resource[F, F[A]] =
  /* move existing memoize implementation here */

def memoize(ra: Resource[F, A]): Resource[F, Resource[F, A]] = 
  memoizedAcquire.map(Resource.eval(_))
```

Then all the existing `memoize` tests can just be in terms of `memoizedAcquire`.

SystemFw:
Yeah fair enough, with that refactor I would agree that we don't need any new tests

Daenyth:
I'm a huge fan of not having to add new tests. I will get to that refactor when/as I have time but if anyone wants to beat me to it, the branch is open for maintainer commits ;)

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -751,6 +751,24 @@ sealed abstract class Resource[F[_], +A] extends Serializable {
...
+   * upon `use()`. If the `F[A]` value is executed multiple times, acquire happens once only and
+   * the acquired resource is shared to all callers. The resource is released as normal at the
+   * end of `use` (whether normal termination, error, or cancelled), if it was acquired.
+   */
+  def memoizedAcquire[B >: A](implicit F: Concurrent[F]): Resource[F, F[B]] = {
+    Resource.eval(F.ref(List.empty[Resource.ExitCase => F[Unit]])).flatMap { release =>
+      val fa2 = F.uncancelable { poll =>
+        poll(allocatedCase).flatMap { case (a, r) => release.update(r :: _).as(a) }
+      }
+      Resource.makeCaseFull[F, F[B]](poll => poll(F.memoize(fa2).map(_.widen))) { (_, exit) =>
```
durban, line 766:
```suggestion
      Resource.makeCaseFull[F, F[B]](poll => poll(F.memoize(fa2)).map(_.widen)) { (_, exit) =>
```

Otherwise I think it can get cancelled right after `F.memoize(fa2)`, and I think that would be not good.

Daenyth:
I don't have the time to update this branch with the changes but I encourage anyone motivated to take over

satabin:
Hi @Daenyth, I would be interested in taking this over if that's ok. I'd love to see this feature land in the 3.6 release :-)

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
armanbilge: Oh I see, good catch. Probably we should add a test case.

djspiewak:
@satabin Go for it! I think you'll need to pull this branch locally and open a new pull request so you can push to it. Please be sure to refrain from rebasing the commits; we can carry the conversation more easily from this PR to that one if the shas are not rewritten.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
satabin: I will look into adding a test case for this, let's see what I can come up with.

satabin: I tried to write a test for this but failed to find a reliable way to ensure acquisition is canceled right after the call to `F.memoize(fa2)` and before the end of `.map(_,widen)`. Is there some tool in the testkit that would allow me to do so and that I am not aware of?

satabin:
I opened #4105 to continue the work on this PR.

Issue #3891, opened 2023-11-09
Leaky `fromFutureCancelable`


TimWSpence:
https://github.com/typelevel/cats-effect/blob/series/3.x/kernel/shared/src/main/scala/cats/effect/kernel/Async.scala#L268 has the potential to leak the future on cancelation due to the lack of `uncancelable`/`poll`

armanbilge:
Thanks, good catch. I think `fromFuture` also needs an `uncancelable`, and we should also check the `js.Promise` variants as well ...

TimWSpence:
> Thanks, good catch. I think `fromFuture` also needs an `uncancelable`, and we should also check the `js.Promise` variants as well ...

I have a vague of fixing things like this before by floating the `async` to the top as the `async` registration at least used to be uncancelable. I'll push a draft PR to demonstrate and try to confirm that that still works 😅

TimWSpence:
OK pushed. Tomorrow I'll try to confirm that it actually does what I think it does 😅

TimWSpence:
Resolved by https://github.com/typelevel/cats-effect/pull/3892

PR #3892, opened 2023-11-09
Fix cancelation leak in `fromFuture`, `fromFutureCancelable`

TimWSpence:
Fix leak in `fromFutureCancelable` and `fromFuture`. Addresses https://github.com/typelevel/cats-effect/issues/3891

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
@@ -212,9 +212,9 @@ trait Async[F[_]] extends AsyncPlatform[F] with Sync[F] with Temporal[F] {
    *   [[fromFutureCancelable]] for a cancelable version
    */
   def fromFuture[A](fut: F[Future[A]]): F[A] =
-    flatMap(fut) { f =>
-      flatMap(executionContext) { implicit ec =>
-        async_[A](cb => f.onComplete(t => cb(t.toEither)))
+    async_[A] { cb =>
+      flatMap(fut) { f =>
+        map(executionContext) { implicit ec => f.onComplete(t => cb(t.toEither)) }
       }
     }
 
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
@@ -222,11 +222,13 @@ trait Async[F[_]] extends AsyncPlatform[F] with Sync[F] with Temporal[F] {
    * Like [[fromFuture]], but is cancelable via the provided finalizer.
    */
   def fromFutureCancelable[A](futCancel: F[(Future[A], F[Unit])]): F[A] =
-    flatMap(futCancel) {
-      case (fut, fin) =>
-        flatMap(executionContext) { implicit ec =>
-          async[A](cb => as(delay(fut.onComplete(t => cb(t.toEither))), Some(fin)))
-        }
+    async[A] { cb =>
+      flatMap(futCancel) {
+        case (fut, fin) =>
+          flatMap(executionContext) { implicit ec =>
+            as(delay(fut.onComplete(t => cb(t.toEither))), Some(fin))
+          }
+      }
     }
 
   /**
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
@@ -212,22 +212,18 @@ trait Async[F[_]] extends AsyncPlatform[F] with Sync[F] with Temporal[F] {
    *   [[fromFutureCancelable]] for a cancelable version
    */
   def fromFuture[A](fut: F[Future[A]]): F[A] =
-    async_[A] { cb =>
-      flatMap(fut) { f =>
-        map(executionContext) { implicit ec => f.onComplete(t => cb(t.toEither)) }
-      }
+    flatMap(executionContext) { implicit ec =>
+      flatMap(fut) { f => async_[A](cb => f.onComplete(t => cb(t.toEither))) }
     }
 
   /**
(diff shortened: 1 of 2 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/Async.scala
```diff
@@ -212,21 +212,23 @@ trait Async[F[_]] extends AsyncPlatform[F] with Sync[F] with Temporal[F] {
...
    * Like [[fromFuture]], but is cancelable via the provided finalizer.
    */
   def fromFutureCancelable[A](futCancel: F[(Future[A], F[Unit])]): F[A] =
-    flatMap(futCancel) {
-      case (fut, fin) =>
-        flatMap(executionContext) { implicit ec =>
-          async[A](cb => as(delay(fut.onComplete(t => cb(t.toEither))), Some(fin)))
-        }
+    async[A] { cb =>
+      flatMap(futCancel) {
```
armanbilge: I think we have to use `uncancelable`+`poll`. Otherwise starting the `Future` with `futCancel` is not cancelable, and it can be. It's only after we've started it, that we need guarantee the `async` is setup appropriately.

armanbilge: I hope that is sufficient anyway, otherwise we'd have to drop down to `cont` 😬

TimWSpence: Well I did only say that I'd fix the leak 😛 

Yeah I was wondering about that as I wrote it. I'll give it a go with `uncancelable`/`poll`. Although I did notice that in `AsyncPlatform#fromCompletableFuture` we changed it to use `cont`. I don't suppose you remember why?

armanbilge: > I don't suppose you remember why?

That's because the cancelation protocol for `CompletableFuture` is more complex: even if you request cancelation, it may indicate to you that it _hasn't_ canceled in which case you need to fallback to waiting for the result. The best way to express this is with `cont`.

(Aside, but CE doesn't really handle this well, we still have leaks even in the `cont`-based implementation https://github.com/typelevel/cats-effect/issues/3474).

TimWSpence:
@armanbilge I've updated now. It looks a bit weird to me but I _think_ this has the correct behaviour? The construction of the future should not prohibit cancelation (by being `uncanelable`) if some part of the provided future construction supports it so we just want to make the `flatMap` boundary before we register the future with the `async` node uncancelable.

I'm not sure about `fromFuture` - we can't cancel there so what is the desired behaviour? Leak the future on cancelation or make it uncancelable?

armanbilge:
> I've updated now. It looks a bit weird to me but I _think_ this has the correct behaviour?

We can add some tests, that will help us understand if we got it right :)

TimWSpence:
> > I've updated now. It looks a bit weird to me but I _think_ this has the correct behaviour?
> 
> We can add some tests, that will help us understand if we got it right :)

pfft I don't write bugs 😛 

Yeah I'll do that. What do you think about the semantics of `fromFuture`? Should we leak it or block till it completes?

armanbilge:
> Should we leak it or block till it completes?

Absolutely never leak, `Future` is not cancelable so we just have to block until it completes. This was the major bugfix released in CE v3.5.0.

- https://github.com/typelevel/cats-effect/pull/3205

TimWSpence:
Sorry @armanbilge I realized I wasn't sure where I should put these tests. They don't really feel like law tests but I can't find any other existing `Async` tests. Should they just go in `IOSpec`?

armanbilge:
> I can't find any other existing Async tests

https://github.com/typelevel/cats-effect/blob/44545879b453c9a9b91d5d3c472b58b4ab04adb1/tests/shared/src/test/scala/cats/effect/kernel/AsyncSpec.scala

https://github.com/typelevel/cats-effect/blob/44545879b453c9a9b91d5d3c472b58b4ab04adb1/tests/jvm/src/test/scala/cats/effect/kernel/AsyncPlatformSpec.scala

> Should they just go in `IOSpec`

I guess that's fine too. It's all kind of a mess, in theory there is a distinction between tests of the `Async` typeclass itself and its default implementations and `IO`-specific tests (e.g. we overwrote `Async` default implementations).

TimWSpence:
> > I can't find any other existing Async tests
> 
> https://github.com/typelevel/cats-effect/blob/44545879b453c9a9b91d5d3c472b58b4ab04adb1/tests/shared/src/test/scala/cats/effect/kernel/AsyncSpec.scala
> 
> https://github.com/typelevel/cats-effect/blob/44545879b453c9a9b91d5d3c472b58b4ab04adb1/tests/jvm/src/test/scala/cats/effect/kernel/AsyncPlatformSpec.scala
> 
> > Should they just go in `IOSpec`
> 
> I guess that's fine too. It's all kind of a mess, in theory there is a distinction between tests of the `Async` typeclass itself and its default implementations and `IO`-specific tests (e.g. we overwrote `Async` default implementations).

Oh yeah `AsyncSpec` isn't actually tied to the laws tests, it just currently doesn't run any other tests 🤦

tests/shared/src/test/scala/cats/effect/kernel/AsyncSpec.scala
```diff
@@ -78,6 +80,33 @@ class AsyncSpec extends BaseSpec with Discipline {
     }
   }
 
+  "fromFuture" should {
+    "backpressure on cancelation" in real {
+      // a non-cancelable, never-completing Future
+      def mkf() = Promise[Unit]().future
```
armanbilge, line 99:
```suggestion
      def mkf() = Future.never
```

djspiewak (approved):
Nice catch!

TimWSpence:
> Nice catch!

Thanks Daniel! :)

PR #3900, opened 2023-11-20
Fix #3898 for `Dispatcher.parallel`

### LINKED ISSUE #3898: Cancelation hangs in `Dispatcher#unsafeToFutureCancelable`
h/t @TimWSpence for original minimization.

```scala
//> using scala "3.3.1"
//> using dep "org.typelevel::cats-effect::3.5.2"

import cats.*
import cats.syntax.all.*
import cats.effect.*
import scala.concurrent.duration.*

object Repro extends IOApp.Simple:

  // also sequential
  val run = Dispatcher.parallel[IO].use { dispatcher =>
    IO.fromFuture {
      IO {
        println("starting")
        val (_, cancel) = dispatcher.unsafeToFutureCancelable(IO.never) // broken
        // val (_, cancel) = IO.never.unsafeToFutureCancelable()(runtime) // works
        println("started, now canceling")
        cancel()
      }
    }.flatMap(_ => IO(println("canceled"))).replicateA_(1000)
  }
```

durban:
This seems to fix #3898 only for `Dispatcher.parallel`. (`Dispatcher.sequential` seems to be a completely different problem.)

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
@@ -278,7 +278,7 @@ object Dispatcher {
                       fork(action).flatMap(cancel => F.delay(prepareCancel(cancel)))
 
                     // Check for task cancelation before executing.
-                    F.delay(r.get()).ifM(supervise, F.unit)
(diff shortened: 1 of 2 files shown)


djspiewak (approved):
This is getting so hairy…

TimWSpence:
Nice work! Dispatcher is scary 😅

PR #3903, opened 2023-11-24
Execute `IO.blocking` on WSTP without `BlockContext` indirection

armanbilge:
We already have all the fantastic machinery to execute blocking code on a `WorkerThread` and it is pointless to indirect through the `BlockContext` interface when we can just invoke this mechanism directly.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
@@ -492,12 +492,12 @@ private[effect] final class WorkStealingThreadPool(
     val thread = Thread.currentThread()
     if (thread.isInstanceOf[WorkerThread]) {
       val worker = thread.asInstanceOf[WorkerThread]
-      if (worker.canExecuteBlockingCodeOn(this))
+      if (worker.canExecuteBlockingCodeOn(this)) {
         worker.prepareBlocking()
-      true
-    } else {
-      false
+        return true
+      }
     }
+    false
   }
 
   /**
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
@@ -492,12 +492,20 @@ private[effect] final class WorkStealingThreadPool(
     val thread = Thread.currentThread()
     if (thread.isInstanceOf[WorkerThread]) {
       val worker = thread.asInstanceOf[WorkerThread]
-      if (worker.canExecuteBlockingCodeOn(this)) {
-        worker.prepareBlocking()
-        return true
-      }
+      worker.canExecuteBlockingCodeOn(this)
+    } else {
+      false
     }
-    false
+  }
+
+  /**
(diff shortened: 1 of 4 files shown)


core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -797,7 +797,7 @@ private final class WorkerThread(
    *   There is no reason to enclose any code in a `try/catch` block because the only way this
    *   code path can be exercised is through `IO.delay`, which already handles exceptions.
    */
-  override def blockOn[T](thunk: => T)(implicit permission: CanAwait): T = {
+  def prepareBlocking(): Unit = {
     val rnd = random
 
     pool.notifyParked(rnd)
```
armanbilge, line 799: Unrelated, but I wonder why we invoke `notifyParked()` unconditionally, even if this thread is already `blocking = true`.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -485,14 +485,16 @@ private[effect] final class WorkStealingThreadPool(
...
+   * for blocking. Only returns true for worker threads that belong to this execution context.
    */
   private[effect] def canExecuteBlockingCode(): Boolean = {
     val thread = Thread.currentThread()
     if (thread.isInstanceOf[WorkerThread]) {
       val worker = thread.asInstanceOf[WorkerThread]
-      worker.canExecuteBlockingCodeOn(this)
+      if (worker.canExecuteBlockingCodeOn(this))
+        worker.prepareBlocking()
+      true
```
djspiewak: Can we move this logic up into `IOFiber` instead? I don't think there's any particular reason for it to be here, and you're making something that looks like a boolean check into a side-effecting method.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
djspiewak: Good question. I'm guessing there's no reason and we can probably guard that notification as you say.

djspiewak (changes requested):

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
armanbilge: Well conceptually I liked that if `canExecuteBlockingCode()` returns `true` then you _can_ run blocking code, immediately.

I do see your point. We can expose this as a separate method which `IOFiber` calls. This requires:

1. Doing the whole `Thread.currentThread()` dance a second time. How should we handle the case where the current thread is not a `WorkerThread`, or does not belong to the current pool? Do we care about those cases?

2. Also exposing whatever new API we invent on the JS `WorkStealingThreadPool` shim. Not a big deal, just annoying.

Worth it?

armanbilge: Oh lol, also I just realized this is bugged as-written 😅 fixed in c2ef8bc67f1cc6862d5c9c737f6e107406c27c40.

djspiewak (dismissed):

PR #3906, opened 2023-11-26
Shaving a few quick build yaks

djspiewak:
Mostly because my system auto-updated to Java 21


armanbilge:
I think this effort is futile, I already went through it with Loom support.

To support Java 21 you need Scala 3.3.x, which we cannot update to on this branch.

Instead, I encourage to shave yaks that let us EOL this branch 😉

PR #3917, opened 2023-12-17
Experiment with #3553

### LINKED ISSUE #3553: `resume`/`callback` in `cont`, `async`, etc. should return a `Boolean`
This `Boolean` should be similar to the `Boolean` returned when completing a `Deferred` i.e. it should indicate that this invocation of the `resume`/`callback` is the one and only one that actually completed the callback. Everyone else should get `false`.

This information about whether the callback succeeded would help avoid leaks in various concurrent state machines. Here are two that I encountered recently:
- https://github.com/typelevel/cats-effect/pull/3465#discussion_r1133340355
- https://github.com/typelevel/cats-effect/pull/3549#issuecomment-1519335094

durban:
Obviously don't merge this.


durban:
This PR is really old now; if I ever do this, I'll just rebase and create a new PR.

PR #3921, opened 2023-12-25
Only run mdoc on Temurin 17

armanbilge:
[It's been so broken we didn't realize.](https://github.com/typelevel/cats-effect/actions/runs/7317327789/job/19932728325#step:23:22)

Latest mdoc releases only support JDK 11+. I chose JDK 17 here because we use it for both Scala 2.13 and Scala 3 in the CI matrix and we want to run mdoc for both of those.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -249,7 +249,7 @@ jobs:
       - shell: bash
         run: sbt '++ ${{ matrix.scala }}' '${{ matrix.ci }}'
 
-      - if: (matrix.scala == '2.13.12' || matrix.scala == '3.3.1') && matrix.ci == 'ciJVM'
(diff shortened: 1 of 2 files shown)


durban (dismissed):
(Why did no github CI run for this PR? I only see cirrus tasks...)

armanbilge:
Ugh, good catch.
```
The workflow is not valid. .github/workflows/ci.yml (Line: 252, Col: 13): Unexpected symbol: ''temurin@17'. Located at position 98 within expression: (matrix.scala == '2.13.12' || matrix.scala == '3.3.1') && matrix.ci == 'ciJVM' && matrix.java == 'temurin@17
```
https://github.com/typelevel/cats-effect/actions/runs/7317684688

durban (approved):

PR #3922, opened 2023-12-28
Prevent `Hotswap#get` from acquiring a lock when there is no `Resource` present

### LINKED ISSUE #3802: Prevent `Hotswap#get` from acquiring a lock when there is no `Resource` present
Currently:

https://github.com/typelevel/cats-effect/blob/290f49864236d73f10d2b60d546158b5c1d845e1/std/shared/src/main/scala/cats/effect/std/Hotswap.scala#L143-L149

Every time `get` is called, we acquire a permit from the `Semaphore[F]` inside of `Hotswap`. If any permits are currently acquired, `Hotswap#swap` will be blocked.

Conceivably, a long running operation could occur inside of `Hotswap#get` despite no `Resource` being present, resulting in blocking to prevent replacing a non-existent `Resource` and during that time, any other invocations of `Hotswap#get` will also be forced to operate on `None`, despite any possible invocations of `swap` waiting in the wings with a real, live `Resource`.

josgarmar28:
Resolves #3802

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
@@ -123,10 +123,11 @@ object Hotswap {
 
       def shared(state: Ref[F, State]): Resource[F, Boolean] =
         Resource.makeFull[F, Boolean] { poll =>
-          state.get.flatMap {
-            case Acquired(_, _) => poll(semaphore.acquire.as(true))
-            case _ => F.pure(false)
-          }
+          poll(semaphore.acquire).flatMap(_ =>
+            state.get.flatMap {
+              case Acquired(_, _) => F.pure(true)
+              case _ => semaphore.release.as(false)
+            })
         } { r => if (r) semaphore.release else F.unit }
 
       def exclusive: Resource[F, Unit] =
--- a/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
@@ -121,15 +121,13 @@ object Hotswap {
       def raise(message: String): F[Unit] =
         F.raiseError[Unit](new RuntimeException(message))
 
-      def shared(state: Ref[F, State]): Resource[F, Unit] =
-        Resource.makeFull[F, Unit] { poll =>
-          poll(
-            state.get.flatMap {
-              case Acquired(_, _) => semaphore.acquire
-              case _ => F.unit
-            }
-          )
-        } { _ => semaphore.release }
+      def shared(state: Ref[F, State]): Resource[F, Boolean] =
+        Resource.makeFull[F, Boolean] { poll =>
+          state.get.flatMap {
+            case Acquired(_, _) => poll(semaphore.acquire.as(true))
+            case _ => F.pure(false)
+          }
+        } { r => if (r) semaphore.release else F.unit }
 
       def exclusive: Resource[F, Unit] =
         Resource.makeFull[F, Unit](poll => poll(semaphore.acquireN(Long.MaxValue)))(_ =>
--- a/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
@@ -141,7 +141,7 @@ object Hotswap {
 
           override def get: Resource[F, Option[R]] =
             Resource.makeFull[F, Option[R]] { poll =>
-              poll(semaphore.acquire) *>
+              poll(semaphore.acquire) *> // acquire shared lock
                 state.get.flatMap {
                   case Acquired(r, _) => F.pure(Some(r))
                   case _ => semaphore.release.as(None)
--- a/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
@@ -121,7 +121,15 @@ object Hotswap {
       def raise(message: String): F[Unit] =
         F.raiseError[Unit](new RuntimeException(message))
 
-      def shared: Resource[F, Unit] = semaphore.permit
+      def shared(state: Ref[F, State]): Resource[F, Unit] =
+        Resource.makeFull[F, Unit] { poll =>
+          poll(
+            state.get.flatMap {
+              case Acquired(_, _) => semaphore.acquire
+              case _ => F.unit
+            }
+          )
+        } { _ => semaphore.release }
 
       def exclusive: Resource[F, Unit] =
         Resource.makeFull[F, Unit](poll => poll(semaphore.acquireN(Long.MaxValue)))(_ =>
@@ -141,7 +149,7 @@ object Hotswap {
             }
 
           override def get: Resource[F, Option[R]] =
-            shared.evalMap { _ =>
+            shared(state).evalMap { _ =>
               state.get.map {
                 case Acquired(r, _) => Some(r)
                 case _ => None
--- a/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
@@ -73,9 +73,8 @@ sealed trait Hotswap[F[_], R] {
   def swap(next: Resource[F, R]): F[R]
 
   /**
-   * Acquires a shared lock to retrieve the current resource, if it exists. The returned
-   * resource is guaranteed to be available for its duration. The lock is released if the
-   * current resource does not exist.
+   * Gets the current resource, if it exists. The returned resource is guaranteed to be
+   * available for the duration of the returned resource.
    */
   def get: Resource[F, Option[R]]
 
--- a/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Hotswap.scala
@@ -121,14 +121,6 @@ object Hotswap {
       def raise(message: String): F[Unit] =
         F.raiseError[Unit](new RuntimeException(message))
 
-      def shared(state: Ref[F, State]): Resource[F, Boolean] =
-        Resource.makeFull[F, Boolean] { poll =>
-          state.get.flatMap {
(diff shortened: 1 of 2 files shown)


std/shared/src/main/scala/cats/effect/std/Hotswap.scala
```diff
@@ -121,7 +121,15 @@ object Hotswap {
...
-      def shared: Resource[F, Unit] = semaphore.permit
+      def shared(state: Ref[F, State]): Resource[F, Unit] =
+        Resource.makeFull[F, Unit] { poll =>
+          poll(
+            state.get.flatMap {
+              case Acquired(_, _) => semaphore.acquire
+              case _ => F.unit
+            }
+          )
+        } { _ => semaphore.release }
```
josgarmar28: I'm not sure if there are any cases where this new `shared` method releases a permit that hasn't been acquired.

durban: Yes, I think there are (when state is not `Acquired`).

std/shared/src/main/scala/cats/effect/std/Hotswap.scala
```diff
@@ -121,7 +121,15 @@ object Hotswap {
       def raise(message: String): F[Unit] =
         F.raiseError[Unit](new RuntimeException(message))
 
-      def shared: Resource[F, Unit] = semaphore.permit
+      def shared(state: Ref[F, State]): Resource[F, Unit] =
+        Resource.makeFull[F, Unit] { poll =>
+          poll(
+            state.get.flatMap {
+              case Acquired(_, _) => semaphore.acquire
+              case _ => F.unit
```
armanbilge: We can use a boolean to track whether the semaphore should be released :)

```suggestion
              case Acquired(_, _) => semaphore.acquire.as(true)
              case _ => F.pure(false)
```

armanbilge: Oh, also in this case, the scope of the `poll` should be restricted to `semaphore.acquire`

std/shared/src/main/scala/cats/effect/std/Hotswap.scala
josgarmar28: Understood, thanks!

std/shared/src/main/scala/cats/effect/std/Hotswap.scala
josgarmar28: Done, thank you!

std/shared/src/main/scala/cats/effect/std/Hotswap.scala
```diff
@@ -141,7 +147,7 @@ object Hotswap {
             }
 
           override def get: Resource[F, Option[R]] =
-            shared.evalMap { _ =>
+            shared(state).evalMap { _ =>
               state.get.map {
```
armanbilge: There's a race condition here: we are running `state.get` in the `shared` method, and then again afterwards. What if it has changed in the meantime?

josgarmar28: That's true. I've committed a version without using the `shared` method in order to run `state.get` and return `Some(r)` in one go. Does it have any downsides?

armanbilge: Yes, I think there is still a race condition here. Consider this sequence of steps.
1. Fiber A: runs `state.get` and finds the state is `Acquired` ...
2. Fiber B: exclusively acquires the semaphore and `clear`s the state and releases the resource, then releases the semaphore
3. Fiber A: ... acquires the semaphore, and proceeds to return the (released) resource 💥

armanbilge: Basically, it's not safe to read the state until we've acquired the semaphore, because it may be changed unless we guard it. But depending on what state we get, we may want to return the semaphore early.

josgarmar28: Now, I'm running state.get only when we acquire the semaphore.

std/shared/src/main/scala/cats/effect/std/Hotswap.scala
```diff
@@ -121,7 +121,13 @@ object Hotswap {
       def raise(message: String): F[Unit] =
         F.raiseError[Unit](new RuntimeException(message))
 
-      def shared: Resource[F, Unit] = semaphore.permit
+      def shared(state: Ref[F, State]): Resource[F, Boolean] =
+        Resource.makeFull[F, Boolean] { poll =>
+          state.get.flatMap {
+            case Acquired(_, _) => poll(semaphore.acquire.as(true))
```
armanbilge: Unfortunately this still has the same race condition I described in https://github.com/typelevel/cats-effect/pull/3922#discussion_r1438374423, we are reading the state before acquiring the semaphore.

josgarmar28: Okay, I hope I finally understood that comment.

std/shared/src/main/scala/cats/effect/std/Hotswap.scala
```diff
@@ -121,7 +121,14 @@ object Hotswap {
       def raise(message: String): F[Unit] =
         F.raiseError[Unit](new RuntimeException(message))
 
-      def shared: Resource[F, Unit] = semaphore.permit
+      def shared(state: Ref[F, State]): Resource[F, Boolean] =
+        Resource.makeFull[F, Boolean] { poll =>
+          poll(semaphore.acquire).flatMap(_ =>
```
armanbilge:
```suggestion
          poll(semaphore.acquire) *>
```

std/shared/src/main/scala/cats/effect/std/Hotswap.scala
```diff
@@ -121,7 +121,14 @@ object Hotswap {
...
+            state.get.flatMap {
+              case Acquired(_, _) => F.pure(true)
+              case _ => semaphore.release.as(false)
```
armanbilge: Instead of a `Boolean` actually we can just pass the `Option[R]` like you did before, then we don't need to do `state.get` twice.

josgarmar28: Should I merge both the `get` and `shared` methods into a single `get` method?

armanbilge: Yes I think that makes sense, "`shared`" is no longer independent. We can add a code comment documenting that we are effectively acquiring a shared lock.

josgarmar28: Great. Is the comment clear enough?

std/shared/src/main/scala/cats/effect/std/Hotswap.scala
```diff
@@ -141,12 +140,13 @@ object Hotswap {
...
 
           override def get: Resource[F, Option[R]] =
-            shared.evalMap { _ =>
-              state.get.map {
-                case Acquired(r, _) => Some(r)
-                case _ => None
-              }
-            }
+            Resource.makeFull[F, Option[R]] { poll =>
+              poll(semaphore.acquire) *>
```
armanbilge: Thanks, this implementation is looking good! Since we no longer have the `shared` method I was thinking we can add a comment here.
```suggestion
              poll(semaphore.acquire) *> // acquire shared lock
```

josgarmar28: @armanbilge Should I revert the changes made to the description comment as well?

std/shared/src/main/scala/cats/effect/std/Hotswap.scala
```diff
@@ -73,8 +73,9 @@ sealed trait Hotswap[F[_], R] {
   def swap(next: Resource[F, R]): F[R]
 
   /**
-   * Gets the current resource, if it exists. The returned resource is guaranteed to be
-   * available for the duration of the returned resource.
+   * Acquires a shared lock to retrieve the current resource, if it exists. The returned
+   * resource is guaranteed to be available for its duration. The lock is released if the
+   * current resource does not exist.
    */
```
armanbilge: I think the original comment was okay, the lock is an implementation detail.

tests/shared/src/test/scala/cats/effect/std/HotswapSpec.scala
```diff
@@ -165,6 +165,16 @@ class HotswapSpec extends BaseSpec { outer =>
 
       TestControl.executeEmbed(go, IORuntimeConfig(1, 2)).replicateA_(1000) must completeAs(())
     }
-  }
 
+    "get should not acquire a lock when there is no resource present" in ticked {
+      implicit ticker =>
+        val go = Hotswap.create[IO, Unit].use { hs =>
+          hs.get.useForever.start *>
+            hs.swap(IO.sleep(1.second).toResource) *>
```
armanbilge: We should insert a `sleep` between these two steps to be sure that the fiber `start`s. Also I think it's okay to just swap in a `Resource.unit`, basically we are checking that the swap succeeds and does not hang. If the `get` above had acquired a shared lock, then the `swap` would hang due to be unable to obtain an exclusive lock.

tests/shared/src/test/scala/cats/effect/std/HotswapSpec.scala
```diff
@@ -165,6 +165,16 @@ class HotswapSpec extends BaseSpec { outer =>
...
+            IO.sleep(2.seconds) *>
+            hs.get.use_.timeout(1.second).void
```
armanbilge: I'm not sure if the rest tests anything meaningful.

tests/shared/src/test/scala/cats/effect/std/HotswapSpec.scala
josgarmar28: Agreed!

armanbilge (approved):
LGTM, thanks for all your work on this!

durban (approved):

josgarmar28:
> LGTM, thanks for all your work on this!

@armanbilge Thank you! I appreciate your patience :)

PR #3923, opened 2023-12-30
Rewrote `Dispatcher`

djspiewak:
I have fixed many things. I possibly also made new problems. I think it's more maintainable now, but who am I to say?

I wrote a long comment in the middle which kind of explains the implementation, but the tldr is we're now very explicit about the impure submission queue and its associated race conditions. Those race conditions are fully resolved out by `Worker`, including the cancelation state machine. On the "actually do stuff" side is `Executor`, which is quite simple for the existing modes and quite complex for the new mode: `sequentialCancelable`.

Speaking of which, I added a third mode because I didn't realize it wasn't already the semantics of the second mode (`sequential`). Whoops. Anyway, we might not want to expose this one since I'm not sure anyone wants it, but I did all the work so here you go. Mostly I wanted to prove that it was possible. And it is… just hard.

Oh, I also removed and rewrote a bunch of tests which either weren't right, weren't reliable, or weren't interesting. I added a few new tests covering cases we probably should have always had. I don't *think* I missed anything, but who can say? Please enjoy.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
@@ -521,12 +521,12 @@ object Dispatcher {
   private object Executor {
 
     // default sequential executor (ignores cancelation)
-    def inplace[F[_]: Applicative]: Executor[F] =
+    def inplace[F[_]: Concurrent]: Executor[F] =
       new Executor[F] {
         def apply(task: F[Unit])(registerCancel: F[Unit] => F[Unit]): F[Unit] = {
-          // we can use unit as a cancel action here since it must always sequence *after* the task
-          // thus, the task must complete before the cancel action will be picked up
-          registerCancel(Applicative[F].unit) *> task
+          Concurrent[F].deferred[Unit].flatMap { d =>
+            (registerCancel(d.get) *> task).guarantee(d.complete(()).void)
+          }
         }
       }
 
--- a/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
@@ -300,11 +300,11 @@ object Dispatcher {
                       worker.queue.unsafeOffer(reg)
 
                       def cancel(): Future[Unit] = {
-                        reg.action = null.asInstanceOf[F[Unit]]
-
                         stateR.get() match {
                           case RegState.Unstarted =>
                             val latch = Promise[Unit]()
+
+                            reg.action = null.asInstanceOf[F[Unit]]
 
                             if (stateR.compareAndSet(RegState.Unstarted, RegState.CancelRequested(latch)))
                               latch.future
--- a/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
@@ -184,18 +184,22 @@ object Dispatcher {
    *   - true - wait for the completion of the active fiber
    *   - false - cancel the active fiber
    */
-  def sequential[F[_]: Async](await: Boolean, cancelable: Boolean = false): Resource[F, Dispatcher[F]] =
-    impl[F](false, await, cancelable)
+  def sequential[F[_]: Async](await: Boolean): Resource[F, Dispatcher[F]] =
+    impl[F](false, await, false)
+
+  // TODO decide if we want other people to use this
+  private[std] def sequentialCancelable[F[_]: Async](await: Boolean): Resource[F, Dispatcher[F]] =
+    impl[F](false, await, true)
 
   /*
-   * There are two fundamental modes here: sequential and parallel. There is very little overlap
-   * in semantics between the two apart from the submission side. The whole thing is split up into
+   * There are three fundamental modes here: sequential, parallel, and sequential-cancelable. There
+   * is very little overlap in semantics between the three apart from the submission side. The whole thing is split up into
    * a submission queue with impure enqueue and cancel functions which is drained by the `Worker` and an
    * internal execution protocol which also involves a queue. The `Worker` encapsulates all of the
    * race conditions and negotiations with impure code, while the `Executor` manages running the
    * tasks with appropriate semantics. In parallel mode, we shard the `Worker`s according to the
    * number of CPUs and select a random queue (in impure code) as a target. This reduces contention
-   * at the cost of ordering, which is not guaranteed in parallel mode. With sequential mode, there
+   * at the cost of ordering, which is not guaranteed in parallel mode. With the sequential modes, there
    * is only a single worker.
    *
    * On the impure side, the queue bit is the easy part: it's just a `LinkedBlockingQueue` which
@@ -207,10 +211,15 @@ object Dispatcher {
    * which is in the process of being taken off the queue, and those race conditions are negotiated
    * between the impure code and the `Worker`.
    *
-   * On the pure side, the two different `Executor`s are very distinct. In parallel mode, it's easy:
+   * On the pure side, the three different `Executor`s are very distinct. In parallel mode, it's easy:
    * we have a separate `Supervisor` which doesn't respawn actions, and we use that supervisor to
    * spawn a new fiber for each task unit. Cancelation in this mode is easy: we just cancel the fiber.
-   * For sequential mode, we spawn a *single* executor fiber on the main supervisor (which respawns).
+   *
+   * Sequential mode is the simplest of all: all work is executed in-place and cannot be canceled.
+   * The cancelation action in all cases is simply `unit` because the impure submission will not be
+   * seen until after the work is completed *anyway*, so there's no point in being fancy.
+   *
+   * For sequential-cancelable mode, we spawn a *single* executor fiber on the main supervisor (which respawns).
    * This fiber is paired with a pure unbounded queue and a shutoff latch. New work is placed on the
    * queue, which the fiber takes from in order and executes in-place. If the work self-cancels or
    * errors, the executor will be restarted. In the case of external cancelation, we shut off the
@@ -218,14 +227,6 @@ object Dispatcher {
    * executor fiber in-place so long as we're sure it's actively working on the target task. Once
    * that cancelation completes (which will ultimately restart the executor fiber), we re-fill the
    * queue and unlock the latch to allow new work (from the `Worker`).
-   *
-   * Note that a third mode is possible but not implemented: sequential *without* cancelation. In
-   * this mode, we execute each task directly on the worker fiber as it comes in, without the
-   * added indirection of the executor queue. This reduces overhead considerably, but the price is
-   * we can no longer support external (impure) cancelation. This is because the worker fiber is
-   * *also* responsible for dequeueing from the impure queue, which is where the cancelation tasks
-   * come in. The worker can't observe a cancelation action while it's executing another action, so
-   * cancelation cannot then preempt and is effectively worthless.
    */
   private[this] def impl[F[_]: Async](
       parallel: Boolean,
--- a/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
@@ -214,7 +214,7 @@ object Dispatcher {
    * at the cost of ordering, which is not guaranteed in parallel mode. With the sequential modes, there
    * is only a single worker.
    *
-   * On the impure side, the queue bit is the easy part: it's just a `LinkedBlockingQueue` which
+   * On the impure side, the queue bit is the easy part: it's just a `UnsafeUnbounded` (queue) which
    * accepts Registration(s). It's easiest to think of this a bit like an actor model, where the
    * `Worker` is the actor and the enqueue is the send. Whenever we send a unit of work, that
    * message has an `AtomicReference` which allows us to back-propagate a cancelation action. That
--- a/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
@@ -269,22 +269,19 @@ object Dispatcher {
 
               launchAll.as(new Dispatcher[F] {
                 def unsafeToFutureCancelable[A](fa: F[A]): (Future[A], () => Future[Unit]) = {
-                  def inner[E](fe: F[E], checker: Option[F[Boolean]])
-                      : (Future[E], () => Future[Unit]) = {
+                  def inner[E](fe: F[E], result: Promise[E], finalizer: Boolean): () => Future[Unit] = {
                     if (doneR.get()) {
                       throw new IllegalStateException("Dispatcher already closed")
                     }
 
-                    val p = Promise[E]()
-                    val cancelR = new AtomicReference[F[Unit]]()
-                    val invalidateCancel = Sync[F].delay(cancelR.set(null.asInstanceOf[F[Unit]]))
+                    val stateR = new AtomicReference[TaskState[F]](TaskState.Unstarted)
 
                     // forward atomicity guarantees onto promise completion
                     val promisory = MonadCancel[F] uncancelable { poll =>
                       // invalidate the cancel action when we're done
-                      poll(fe.guarantee(invalidateCancel)).redeemWith(
-                        e => Sync[F].delay(p.failure(e)),
-                        a => Sync[F].delay(p.success(a)))
+                      poll(fe.guarantee(Sync[F].delay(stateR.set(TaskState.Completed)))).redeemWith(
+                        e => Sync[F].delay(result.failure(e)),
+                        a => Sync[F].delay(result.success(a)))
                     }
 
                     val worker =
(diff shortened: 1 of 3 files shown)


djspiewak:
Hard for me to tell if this is a legit failure or not, but there's at least *one* legit bug here somewhere. Got this locally just now:

```
[error]     ! not hang when cancelling
[error]      cats.effect.TestTimeoutException: null (Runners.scala:109)
[error] cats.effect.unsafe.WorkStealingThreadPool.$anonfun$sleep$1(WorkStealingThreadPool.scala:685)
[error] cats.effect.unsafe.WorkStealingThreadPool.$anonfun$sleep$1$adapted(WorkStealingThreadPool.scala:685)
[error] cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:740)
```

JVM, 2.12.18, ARM64 (parallel, await = false)

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -190,240 +196,444 @@ object Dispatcher {
...
+   * is very little overlap in semantics between the three apart from the submission side. The whole thing is split up into
+   * a submission queue with impure enqueue and cancel functions which is drained by the `Worker` and an
+   * internal execution protocol which also involves a queue. The `Worker` encapsulates all of the
+   * race conditions and negotiations with impure code, while the `Executor` manages running the
+   * tasks with appropriate semantics. In parallel mode, we shard the `Worker`s according to the
+   * number of CPUs and select a random queue (in impure code) as a target. This reduces contention
+   * at the cost of ordering, which is not guaranteed in parallel mode. With the sequential modes, there
+   * is only a single worker.
+   *
+   * On the impure side, the queue bit is the easy part: it's just a `LinkedBlockingQueue` which
```
durban: The LinkedBlockingQueue have been already replaced by UnsafeUnbounded, but this comment still mentions it.

djspiewak: Good callout thank you!

tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
```diff
@@ -201,37 +310,67 @@ class DispatcherSpec extends BaseSpec with DetectPlatform {
...
+              gate.get *> IO(cancel())
+            }
+          }
+          clogUp.parReplicateA_(1000) *>
+            // now try to run a new task
+            IO.fromFuture(IO(runner.unsafeToFuture(IO.unit)))
+        }
+        .replicateA_(if (isJVM) 1000 else 1)
+        .as(ok)
+    }
```
armanbilge: This test verifies that the worker doesn't get blocked here:

https://github.com/typelevel/cats-effect/blob/abf680d64527489b41067b51224ca2629eedc21d/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala#L449-L450

tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala
```diff
@@ -201,37 +310,67 @@ class DispatcherSpec extends BaseSpec with DetectPlatform {
...
+            IO(cancel())
+          }
+
+          clogUp.parReplicateA_(1000) *>
+            // now try to run a new task
+            IO.fromFuture(IO(runner.unsafeToFuture(IO.unit)))
+        }
+        .replicateA_(if (isJVM) 1000 else 1)
+        .as(ok)
     }
```
armanbilge: This test verifies that the worker doesn't get blocked here.
https://github.com/typelevel/cats-effect/blob/abf680d64527489b41067b51224ca2629eedc21d/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala#L427-L429

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -190,240 +197,474 @@ object Dispatcher {
...
+
+  private object Executor {
+
+    // default sequential executor (ignores cancelation)
+    def inplace[F[_]: Applicative]: Executor[F] =
+      new Executor[F] {
+        def apply(task: F[Unit])(registerCancel: F[Unit] => F[Unit]): F[Unit] = {
+          // we can use unit as a cancel action here since it must always sequence *after* the task
+          // thus, the task must complete before the cancel action will be picked up
+          registerCancel(Applicative[F].unit) *> task
```
armanbilge:
```suggestion
          // we can use unit as a cancel action here since it must always sequence *after* the task
          // thus, the task must complete before the cancel action will be picked up and we do not need to backpressure on its termination as is expected of cancel
          registerCancel(Applicative[F].unit) *> task
```

djspiewak:
Fixed all the things, I think. The first test was more deterministically expressible via `TestControl`. The second test was a deadlock even with corrected functionality, so I fixed that as well. Both tests rely on the latch release running on the dispatcher, which implies the workers are unblocked.

I am now seeing a rather odd issue though, which is occasionally something like this prints:

```
[info]     + cancelation does not block a worker
cats.effect.std.FailureSignal$
[info]     + cancelation race does not block a worker
[info]     + cancel all inner effects when canceled
[info] Total for specification DispatcherSpec
```

This suggests that `UnsafeUnbounded` is being taken while empty and the error is *not* handled, so it's falling through to the fiber default handler. This isn't ideal, obviously, and might be related to @armanbilge's changes.

armanbilge:
Yes, that was my fault, sorry! I've got a fix locally, will push it up a bit later.

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -190,240 +198,456 @@ object Dispatcher {
...
+    impl[F](false, await, true)
+
+  /*
+   * There are three fundamental modes here: sequential, parallel, and sequential-cancelable. There
+   * is very little overlap in semantics between the three apart from the submission side. The whole thing is split up into
+   * a submission queue with impure enqueue and cancel functions which is drained by the `Worker` and an
+   * internal execution protocol which also involves a queue. The `Worker` encapsulates all of the
+   * race conditions and negotiations with impure code, while the `Executor` manages running the
+   * tasks with appropriate semantics. In parallel mode, we shard the `Worker`s according to the
+   * number of CPUs and select a random queue (in impure code) as a target. This reduces contention
```
samspills, line 216: I'm not sure I understand this comment. In the existing (v3.5.3) implementation, a dispatcher has N workers and each worker has N queues (where N = number of CPUs). But in the rewrite there are N workers and each worker has just 1 queue (so we've lost a degree of sharding). I don't see where we're selecting a random queue, just where we select the random worker. Maybe this comment could be clarified? 

And then relatedly: do you have a feeling for the practical consequences of the reduced sharding? Is it okay to just shard across workers?

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -190,240 +198,456 @@ object Dispatcher {
...
+          val workersF =
+            if (parallel)
+              workerF.replicateA(Cpus).map(_.toArray)
+            else
+              workerF.map(w => Array(w))
+
+          workersF evalMap { workers =>
+            Async[F].executionContext flatMap { ec =>
+              val launchAll = 0.until(workers.length).toList traverse_ { i =>
+                supervisor.supervise(workers(i).run).void
```
durban:
```suggestion
                supervisor.supervise(workers(i).run)
```

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -190,240 +198,456 @@ object Dispatcher {
...
+                    if (finalizer) {
+                      worker.queue.unsafeOffer(Registration.Finalizer(promisory.void))
+
+                      // cannot cancel a cancel
+                      () => Future.failed(new UnsupportedOperationException)
+                    } else {
+                      val reg = new Registration.Primary(promisory.void, stateR)
+                      worker.queue.unsafeOffer(reg)
+
+                      def cancel(): Future[Unit] = {
```
durban, line 335:
```suggestion
                      @tailrec
                      def cancel(): Future[Unit] = {
```

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -190,240 +198,456 @@ object Dispatcher {
...
+                    if (doneR.get()) {
+                      throw new IllegalStateException("Dispatcher already closed")
+                    }
+
+                    val stateR = new AtomicReference[RegState[F]](RegState.Unstarted)
+
+                    // forward atomicity guarantees onto promise completion
+                    val promisory = MonadCancel[F] uncancelable { poll =>
+                      // invalidate the cancel action when we're done
+                      poll(fe.guarantee(Sync[F].delay(stateR.set(RegState.Completed))))
```
durban: I think there could be a problem here: if `fe` finishes very quickly, this will unconditionally set the state to completed, even if there is a `CancelRequested` already there. If the `Worker` checks the state _after_ this, it won't find the `CancelRequested`, and this way the cancel will hang (the promise will never be completed).

At least I _think_ that's what happens. I have a branch, where there is a test which (sometimes) hangs; also, in the next commit, I'm trying to fix it: https://github.com/durban/cats-effect/tree/dispatcher-bug. (Seems to work, or at least the test doesn't hang any more.)

Feel free to merge my branch into this one if you think it's useful.

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -190,240 +198,456 @@ object Dispatcher {
...
-                  if (!cancelState.compareAndSet(state, CancelToken(cancelToken))) {
-                    loop()
-                  } else {
-                    cancelToken().onComplete {
-                      case Success(_) => promise.success(())
-                      case Failure(ex) => promise.failure(ex)
-                    }(ec)
+                } else {
+                  val withCompletion =
+                    action.guarantee(Sync[F].delay(reg.stateR.set(RegState.Completed)))
```
durban: Also I think, this is redundant (we're already setting the state to completed above). (This is also removed in my branch.)

durban (commented):
I've left a few comments. ([One of them](https://github.com/typelevel/cats-effect/pull/3923#discussion_r1470187188) is I think about a bug and a fix.)

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
djspiewak: This is a good question. I actually don't know. This suggests that we really do need `Dispatcher` benchmarks and I should stop putting it off. My intuition about the cost of the reduced sharding is that it's likely to be minimal. It's basically going to put a bit more load on the enqueueing side (in the form of contention) while simplifying and taking load off of the dequeue side (the worker).

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
djspiewak: Hmmmmmmmmmmmmmmmmmmmmmmmmm! I'll take a look! Thank you!

djspiewak: Ah I agree this is a problem. I stared at it for a little while before I got it, but the issue manifests when the action completes before the `execute` lambda has a chance to run, meaning that the state is still `Unstarted`. So the exact sequencing:

1. We start running the action in `execute` (`stateR == Unstarted`)
2. We cancel the action in `cancel()` (`stateR == CancelRequested(latch)`)
3. The action completes and we unconditionally flip the state (`stateR == Completed`)
4. The `execute` lambda finally has a chance to run. The CAS fails but the state is `Completed` so we fall through to `Applicative[F].unit`
5. *`latch` is still waiting*

I think the solution is to `getAndSet` and complete the latch if present.

durban: Yes, I think that's what I did on my branch (if I understand you correctly).

djspiewak: You do and you did! I merged your changes. Interestingly, I can't get your test to *ever* fail on my M2 Max. My guess is the hardware timing is such that it's too hard to reproduce that narrow timeslice. Very nice catch.

durban: Yeah, when I was playing with it, I noticed that the test was _really_ sensitive to _very_ small things.

djspiewak:
Hold on this until we can track down the uncaught supervisor state exceptions.

durban:
I believe I've fixed _one_ of the sources of the uncaught exceptions: #3991 (feel free to merge that branch into this one). The test affected was `complete / cancel race`. There was a race with multiple completions of a `CancelRequested(latch)` (with `trySuccess`), and one of these would also submit a `Fiber#cancel` to the cancelling-supervisor. But others would've already completed the `latch`, so the cancel-future would be completed, so the test could finish, so the supervisor could shut down. The fix in #3991 is to handle the state properly, and make sure there is only one completion of the latch.

There remain other sources of the same exception:

- `run multiple IOs in parallel`: I don't really understand this test; it seems to intentionally shut down a dispatcher while things are still going on?
- `cancelation race does not block a worker`: I think (but I'm not sure) this causes a `Worker` to finalize itself (with enqueing `PoisonPill`), then right after that the cancelling-supervisor finalizes itself (because it's the next `Resource` to close), but the queue of the worker is not empty, there could still be `Registration.Finalizer`s in it, which are then submitted to the cancelling-supervisor.

durban:
For the sequential, non-cancelable dispatcher, is `cancel()` supposed to be a no-op? Or is the future returned by `cancel()` supposed to be completed only once the task is finished?

Because currently (i.e., in this PR) it nondeterministically behaves either way, based on whether `cancel()` wins or loses a race. I.e., the following test sometimes hangs, sometimes completes for `Dispatcher.sequential[IO](await = true)`:

```scala
      "foobar" in real {
        D.use { dispatcher =>
          IO.fromFuture(IO {
            val (_, cancel) = dispatcher.unsafeToFutureCancelable(IO.never[Unit])
            val cancelFut = cancel()
            cancelFut
          })
        }.replicateA_(6).as(ok)
      }
```

djspiewak:
> For the sequential, non-cancelable dispatcher, is cancel() supposed to be a no-op? Or is the future returned by cancel() supposed to be completed only once the task is finished?

I think the latter is the behavior I would expect. `cancel()` should basically behave as if the whole task is wrapped in `uncancelable`. Nondeterministically behaving in one way or the other sounds like a gloriously weird bug.

Also I'm chewing on your branch right now. Will merge and then get to work on some of the other bits.

djspiewak:
I fixed the weird parallel IO test to avoid dependence on await/non-await semantics. I still see unhandled exceptions though. It's really difficult for me to reproduce those exceptions on this machine. Also because of parallel specs execution it's very possible the exceptions are coming from somewhere other than those tests, but I'm not sure.

**Edit:** Actually now that I think about it, the nondeterminism in sequential `cancel()` *is* consistent with what would happen if the action was just wrapped in `uncancelable`. Consider:

```scala
IO.never.uncancelable.start.flatMap(_.cancel)
```

This will nondeterministically hang depending on whether the `cancel` gets in before the fiber begins running. Same idea here. So I *think* it's kind of okay?

durban:
> This will nondeterministically hang depending on whether the cancel gets in before the fiber begins running. Same idea here. So I think it's kind of okay?

Except in the case of `start`/`cancel`, the uncancelable task either (1) doesn't even start executing (and `cancel` finishes quickly), or (2) executes "completely", i.e., hangs in `never` (and `cancel` also hangs).

With this dispatcher (note, we're still talking about sequential, cancelable = false, await = true) I think there is a (3): the task starts executing, but `cancel` finishes before the task is completed.

This test fails:

```scala
      "foobar2" in real {
        D.use { dispatcher =>
          IO.ref(0).flatMap { ctr1 =>
            IO.ref(0).flatMap { ctr2 =>
              IO.fromFuture(IO {
                val (_, cancel) = dispatcher.unsafeToFutureCancelable(IO.uncancelable { _ =>
                  ctr1.update(_ + 1) *> IO.sleep(0.1.second) *> ctr2.update(_ + 1)
                })
                val cancelFut = cancel()
                cancelFut
              }).flatMap { _ =>
                // if we're here, `cancel()` finished, so
                // either the task didn't run at all (i.e.,
                // it was cancelled before starting), or
                // it ran and already finished completely:
                (ctr1.get, ctr2.get).flatMapN { (v1, v2) =>
                  IO(v1 mustEqual v2)
                }
              }
            }
          }
        }.replicateA_(10000).as(ok)
      }
```

djspiewak:
Ah, okay I buy that this is a problem then.

djspiewak:
Minor update: I've seen this test fail once, but it's very difficult to get it to fail at all on my laptop. Something about the timing I suspect. I may need some help tracking this one down.

durban:
I think I've fixed it. See #4012 (feel free to merge my branch directly into this PR).

The problem is with this comment:

```scala
// we can use unit as a cancel action here since it must always sequence *after* the task
// thus, the task must complete before the cancel action will be picked up
```

This is not (always) true:

1. It _is_ true if `registerCancel` in `Worker` modifies the state `Unstarted` -> `Running(pure(()))`. In that case, `cancel()` will find the `pure(())` in `Running`, and submit it to the _same queue_.
2. However, if `cancel()` is faster, changes state `Unstarted` -> `CancelRequested`. Then `registerCancel` in `Worker` changes state `CancelRequested` -> `Running(pure(()))`, then submits the `pure(())` to _the separate cancel-supervisor_. The cancel-supervisor "executes" the `pure(())`, then completes the `latch` in `CancelRequested` (`cancel()` is waiting for this). What the cancel-supervisor does, is (or can be) parallel to `Executor` running the orignal task.

My fix (in [6a9cfa6](https://github.com/typelevel/cats-effect/pull/4012/commits/6a9cfa617676887dd34ae450aff05dac0ddb5072)) is to instead of the synthetic `pure(())` cancellation action, wait for a `Deferred` for the task to complete. (Since we're ignoring cancellation, this should be fine in either case.) Without the fix, I could reliably reproduce the bug with the test in [a3c285d](https://github.com/typelevel/cats-effect/pull/4012/commits/a3c285dcbe9134394efe71d70957949d23de18ac). With the fix, I can't any more.

djspiewak:
Magnificent fix @durban! Sorry I had been too busy to get to this. Snagged it now. Thank you!

djspiewak:
Let's call this good. I got the 👍 from @armanbilge just now to merge without waiting for him, since @durban in particular has been all up in this PR's business. If nothing else, this should make the build a lot more stable!

Issue #3313, opened 2022-12-11
[Feature request] Add a wrapper for Properties


massimosiani:
Hi! Thinking about adding a wrapper for system props, similar to what `Env` is for env vars.
I came up with
```
trait Prop[F[_]] {
  def get(name: String): F[Option[String]]
  def modify(name: String, f: String => (String, String)): F[Option[String]]
  def set(name: String, value: String): F[Unit]
  def update(name: String, f: String => String): F[Unit]
}
```
(don't really like the name).

I can provide a PR if it sounds useful.

bpholt:
I could have used this today. It seems like a helpful addition.

iRevive:
Same. I have several use cases where `Prop` would be helpful.

PR #3924, opened 2024-01-01
Add `SystemProperties`

massimosiani:
Addresses #3313.


djspiewak:
Thank you for looking at this! An interesting question for us to answer though is what we want to do with Native and JavaScript. This is a bit different from `Env`, which works on all three platforms (with the exception of in-browser JS).

armanbilge:
> Native and JavaScript

They both support System properties. It's basically just a global mutable map.

durban:
A question: the `getAndSet`, `modify`, and similar methods look like the ones on `Ref`. However, here they are not atomic, right? So having them could be somewhat misleading. (While having only get and set would make it clear, that there is no atomicity for modification.)

massimosiani:
I would agree. TBH, I wondered whether we should try and make them atomic, but that's probably not something you'd expect by system props.

durban:
I don't think we can make them atomic. Or do you know a way?

massimosiani:
Right, not sure how could be made atomic. Citing Arman, it's basically just a global mutable map.
I'll remove `getAndSet` and `getAndUpdate` but I think that `update` is still useful if we make clear that there are no guarantees about concurrent modifications. It's a handy wrapper for a `get` followed by `set`. I also like `modify` as a wrapper but would like to hear other opinions.

armanbilge:
> I think that `update` is still useful if we make clear that there are no guarantees about concurrent modifications

Unfortunately I think it's still confusing. Anyway, I expect 99% of usecases to just be reading system properties. APIs that require setting/updating system properties seem really dicey 🙄

durban:
I'd just go with get, set and remove. If that's the functionality offered by the platform, and we can't provide atomicity, its best simply to wrap those.

armanbilge:
Similar to `Env`, we can also add something to list/iterate the keys/entries.

std/shared/src/main/scala/cats/effect/std/Prop.scala
```diff
@@ -0,0 +1,135 @@
...
+   * same key.
+   */
+  def set(key: String, value: String): F[Unit]
+
+  /**
+   * Removes the property.
+   */
+  def unset(key: String): F[Unit]
+
+  def entries: F[Properties]
```
durban: I'm not sure I like returning a mutable object here. (I think we usually try to avoid that?)

armanbilge: For `Env` we decided to return an `Iterable[(String, String)]`, but that was because of case-sensitivity issues which I don't think apply here. So probably makes most sense to return an immutable `Map[String, String]`.

std/shared/src/main/scala/cats/effect/std/Prop.scala
```diff
@@ -0,0 +1,142 @@
...
+  ]: Prop[ReaderWriterStateT[F, E, L, S, *]] =
+    Prop[F].mapK(ReaderWriterStateT.liftK)
+
+  private[std] final class SyncProp[F[_]](implicit F: Sync[F]) extends Prop[F] {
+
+    def get(key: String): F[Option[String]] =
+      F.delay(Option(System.getProperty(key))) // thread-safe
+
+    def set(key: String, value: String): F[Unit] =
+      F.void(F.delay(System.setProperty(key, value)))
```
durban: If I'm reading the JDK source correctly, this eventually calls a `synchronized` method. So, technically, it should be `F.blocking`. I don't know if we care about this though...

std/shared/src/main/scala/cats/effect/std/Prop.scala
```diff
@@ -0,0 +1,142 @@
...
+
+    def unset(key: String): F[Unit] = F.void(F.delay(System.clearProperty(key)))
+
+    @nowarn213("cat=deprecation")
+    @nowarn3("cat=deprecation")
+    def entries: F[Map[String, String]] = {
+      import scala.collection.JavaConverters._
+      F.delay(Map.empty ++ System.getProperties().asScala)
```
durban: Similarly, I think this calls into a `Collections.synchronizedSet`, so technically should be `blocking`. (Although this should be uncontended, I think. So we probably care even less...)

tests/shared/src/test/scala/cats/effect/std/PropSpec.scala
```diff
@@ -0,0 +1,45 @@
...
+      Random.javaUtilConcurrentThreadLocalRandom[IO].nextString(12).flatMap { key =>
+        Prop[IO].set(key, "bar") *> Prop[IO]
+          .unset(key) *> Prop[IO].get(key).flatMap(x => IO(x must beNone))
+      }
+    }
+    "retrieve the system properties" in real {
+      for {
+        _ <- Prop[IO].set("some property", "the value")
+        props <- Prop[IO].entries
+        assertion <- IO(props mustEqual System.getProperties())
```
durban: There is a test failure here, I think it's just because we're comparing a Scala `Map` with a Java hashtable.

durban (commented):
I think the `entries` is not an atomic snapshot of the system properties. it would be nice if we could do that, but I don't think it's possible with the available JDK API.

std/shared/src/main/scala/cats/effect/std/Prop.scala
```diff
@@ -0,0 +1,142 @@
...
+import cats.{~>, Applicative, Functor}
+import cats.data.{EitherT, IorT, Kleisli, OptionT, ReaderWriterStateT, StateT, WriterT}
+import cats.effect.kernel.Sync
+import cats.kernel.Monoid
+
+import org.typelevel.scalaccompat.annotation._
+
+import scala.collection.immutable.Map
+
+trait Prop[F[_]] { self =>
```
durban: Maybe we should make this `sealed`, just for future-proofing?

TonioGela:
[BIKESHEDDING] I'm late to the party to comment probably, but I fear that `Prop[F]` as a name may clash with scalacheck or one of its integrations.

std/shared/src/main/scala/cats/effect/std/Prop.scala
armanbilge: On the other hand, this prevents mocking. Our current practice is not really to `sealed` anything, for better or for worse 🤷

armanbilge:
> but I fear that `Prop[F]` as a name may clash with scalacheck or one of its integrations.

I actually also dislike the name `Prop[F]`, without even considering the Scalacheck issues (good point).

The reason that I dislike it is because this API is very specifically abstracting over the (global singleton) _system_ properties, not the generic JDK `Properties` datatype.

If we want to keep the name short we could rename to `SysProps[F]` perhaps. Or just go for `SystemProperties[F]` which I think is my preference.

std/shared/src/main/scala/cats/effect/std/Prop.scala
armanbilge: Yes, that's right. I think we should do that.

armanbilge:
```suggestion
      F.void(F.blocking(System.setProperty(key, value)))
```

armanbilge:
> I think the `entries` is not an atomic snapshot of the system properties. it would be nice if we could do that, but I don't think it's possible with the available JDK API.

The JDK suggests there is some notion of "bulk" operations.

```
    /**
     * Properties does not store values in its inherited Hashtable, but instead
     * in an internal ConcurrentHashMap.  Synchronization is omitted from
     * simple read operations.  Writes and bulk operations remain synchronized,
     * as in Hashtable.
     */
```

https://github.com/openjdk/jdk/blob/388fcf03c02c41bb690733e8565642c24ead56e0/src/java.base/share/classes/java/util/Properties.java#L161-L167

I am still trying to figure out what the bulk operations are exactly. If serializing the properties to a stream is one such bulk operation, perhaps we can use that to make an atomic copy first ... 🤔

armanbilge:
> I think the `entries` is not an atomic snapshot of the system properties. it would be nice if we could do that, but I don't think it's possible with the available JDK API.

It is possible, I think I was able to fix it. Thanks for catching :)

std/shared/src/main/scala/cats/effect/std/Prop.scala
```diff
@@ -134,9 +134,13 @@ object Prop {
...
-    def entries: F[Map[String, String]] = {
-      import scala.collection.JavaConverters._
-      F.delay(Map.empty ++ System.getProperties().asScala)
-    }
+    def entries: F[Map[String, String]] =
+      F.blocking {
+        import scala.collection.JavaConverters._
+        val props = System.getProperties
+        val back = new java.util.HashMap[String, String](props.size())
+        props.putAll(back)
```
armanbilge: Aw crap, I think I got this backwards 😕 it should be `back.putAll(props)` but then it's no longer atomic which was the point of copying. Hmm.

massimosiani:
> Or just go for `SystemProperties[F]` which I think is my preference.

That's my favorite too.

std/shared/src/main/scala/cats/effect/std/SystemProperties.scala
```diff
@@ -0,0 +1,117 @@
...
+
+  /**
+   * Constructs a `SystemProperties` instance for `F` data types that are
+   * [[cats.effect.kernel.Sync]].
+   */
+  def make[F[_]](implicit F: Sync[F]): SystemProperties[F] = new SyncSystemProperties[F]
+
+  /**
+   * [[Prop]] instance built for `cats.data.EitherT` values initialized with any `F` data type
+   * that also implements `Prop`.
```
iRevive:
```suggestion
   * [[SystemProperties]] instance built for `cats.data.EitherT` values initialized with any `F` data type
   * that also implements `SystemProperties`.
```

armanbilge:
After going through a few iterations, I think we should just stick to `get`/`set`/`clear`.

Technically `Properties` extends `Hashtable`, and it is possible to implement a `MapRef` for `Hashtable`. But actually it extends `Hashtable[Object, Object]`, and the `Properties`-specific APIs add special handling for non-`String` values and defaults. Anyway, it's a bit of a mess, and I really don't see a compelling usecase for atomic modification of `Properties`.

djspiewak (approved):

PR #3733, opened 2023-07-04
fix: Added priority queue sink and source extending queue

lukecollier:
**Motivation**
Following the discussion on [this issue](https://github.com/typelevel/cats-effect/issues/3721) this ticket add's quality of live improvements for using priority queues with various functions.

Issue #3776, opened 2023-08-06
`Dispatcher` "raise an error on leaked runner" flake


armanbilge:
55044195c8d6bdfdadcd7411aa6ebe9e97bcf045

https://cirrus-ci.com/task/5809014697623552?logs=main#L3024

```
[info] DispatcherSpec
[info] sequential dispatcher should
[info]   await = true
...
[error]     x raise an error on leaked runner
[error]      Expected: java.lang.IllegalStateException. Got nothing (DispatcherSpec.scala:256)
```

https://github.com/typelevel/cats-effect/blob/55044195c8d6bdfdadcd7411aa6ebe9e97bcf045/tests/shared/src/test/scala/cats/effect/std/DispatcherSpec.scala#L253-L259

armanbilge:
Flaked again.

1bbe7f619125302131b0d6c94a67ad05aa31e33a

https://cirrus-ci.com/task/4850274011971584?logs=main#L3010

armanbilge:
By adding replicates to the test in efd0c23c30b2e7c0c3161f74b4ae931a2216ff58 I was able to reproduce more easily. It's not specific to any `Dispatcher`, just seems generally broken 😕 

```
sbt:cats-effect> testsJVM/testOnly cats.effect.std.DispatcherSpec -- ex "raise an error on leaked runner"
[info] DispatcherSpec
[info] sequential dispatcher should
[info]   await = true
[error]     x raise an error on leaked runner
[error]      Expected: java.lang.IllegalStateException. Got nothing (DispatcherSpec.scala:256)
[info]   await = false
[error]     x raise an error on leaked runner
[error]      Expected: java.lang.IllegalStateException. Got nothing (DispatcherSpec.scala:256)
[info] parallel dispatcher should
[info]   await = true
[error]     x raise an error on leaked runner
[error]      Expected: java.lang.IllegalStateException. Got nothing (DispatcherSpec.scala:256)
[info]   await = false
[error]     x raise an error on leaked runner
[error]      Expected: java.lang.IllegalStateException. Got nothing (DispatcherSpec.scala:256)
[info] Total for specification DispatcherSpec
```

durban:
Another example: https://github.com/typelevel/cats-effect/actions/runs/6497507058/job/17646786126#step:22:4783

PR #3930, opened 2024-01-04
Added `PriorityQueue{Sink,Source}` extending `Queue{Sink,Source}`

neomaclin:
trying to continue the work #3733 by @lukecollier
@armanbilge

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/PQueue.scala
+++ b/std/shared/src/main/scala/cats/effect/std/PQueue.scala
@@ -270,6 +270,7 @@ trait PQueueSink[F[_], A] extends QueueSink[F, A] {
 }
 
 object PQueueSink {
+
   implicit def catsContravariantForPQueueSink[F[_]]: Contravariant[PQueueSink[F, *]] =
     new Contravariant[PQueueSink[F, *]] {
       override def contramap[A, B](fa: PQueueSink[F, A])(f: B => A): PQueueSink[F, B] =
--- a/std/shared/src/main/scala/cats/effect/std/PQueue.scala
+++ b/std/shared/src/main/scala/cats/effect/std/PQueue.scala
@@ -231,7 +231,7 @@ trait PQueueSource[F[_], A] extends QueueSource[F, A] {
    * dequeued is undefined.
    */
   override def tryTakeN(maxN: Option[Int])(implicit F: Monad[F]): F[List[A]] =
-    super.tryTakeN(maxN)
+    QueueSource.tryTakeN[F, A](maxN, tryTake)
 
 }
 
@@ -265,7 +265,7 @@ trait PQueueSink[F[_], A] extends QueueSink[F, A] {
    *   an effect that contains the remaining valus that could not be offered.
    */
   override def tryOfferN(list: List[A])(implicit F: Monad[F]): F[List[A]] =
-    super.tryOfferN(list)
+    QueueSink.tryOfferN(list, tryOffer)
 
 }
 
--- a/std/shared/src/main/scala/cats/effect/std/PQueue.scala
+++ b/std/shared/src/main/scala/cats/effect/std/PQueue.scala
@@ -214,9 +214,48 @@ object PQueue {
     else ()
 }
 
-trait PQueueSource[F[_], A] extends QueueSource[F, A]
+trait PQueueSource[F[_], A] extends QueueSource[F, A] {
+
+  /**
+   * Attempts to dequeue elements from the PQueue, if they are available without semantically
+   * blocking. This is a convenience method that recursively runs `tryTake`. It does not provide
+   * any additional performance benefits.
+   *
+   * @param maxN
+   *   The max elements to dequeue. Passing `None` will try to dequeue the whole queue.
+   *
+   * @return
+   *   an effect that contains the dequeued elements from the PQueue
+   *
+   * Note: If there are multiple elements with least priority, the order in which they are
+   * dequeued is undefined.
+   */
+  override def tryTakeN(maxN: Option[Int])(implicit F: Monad[F]): F[List[A]] = {
+    PQueueSource.assertMaxNPositive(maxN)
+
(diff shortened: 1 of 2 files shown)


durban:
It seems to me, that `tryTakeN` and `tryOfferN` has the same implementation in `QueueSource`/`QueueSink`. So why are we overriding?

neomaclin:
> It seems to me, that `tryTakeN` and `tryOfferN` has the same implementation in `QueueSource`/`QueueSink`. So why are we overriding?

actually, come to think of it. they are actually identical in almost everyway. a simple extends without should just do., pushed

neomaclin:
insteresting seeing the error in Mima check, maybe that is need of override?

armanbilge:
> insteresting seeing the error in Mima check, maybe that is need of override?

yes, I think you can add an override call that delegates to the super so you do not need to repeat the implementation.

neomaclin:
> > insteresting seeing the error in Mima check, maybe that is need of override?
> 
> yes, I think you can add an override call that delegates to the super so you do not need to repeat the implementation.

looks like Mima hates either way....(override with super call triggered the ReversedMissingMethodProblem)

neomaclin:
why DispatcherSpec failed is a bit beyound me...

neomaclin:
> It seems to me, that `tryTakeN` and `tryOfferN` has the same implementation in `QueueSource`/`QueueSink`. So why are we overriding?

I guess the attempts were telling us MiMa doesn't agree...

armanbilge:
> looks like Mima hates either way....(override with super call triggered the ReversedMissingMethodProblem)

hmm. can you share the full MiMa error message? I don't understand why it's not happy.

neomaclin:
> > looks like Mima hates either way....(override with super call triggered the ReversedMissingMethodProblem)
> 
> hmm. can you share the full MiMa error message? I don't understand why it's not happy.

ah. in order to fix the build, I have force pushed and squashed that commit (not sure how to find the actual error log in that case). but bascially MiMa is reported the "ReversedMissingMethodProblem" along with the message "super.tryOfferN is present only in the current version" if I have only

`override def tryOfferN(list: List[A])(implicit F: Monad[F]): F[List[A]] = super.tryOfferN(list)`

neomaclin:
> > looks like Mima hates either way....(override with super call triggered the ReversedMissingMethodProblem)
> 
> hmm. can you share the full MiMa error message? I don't understand why it's not happy.

or, I could break the build again to generate the error.

neomaclin:
> > looks like Mima hates either way....(override with super call triggered the ReversedMissingMethodProblem)
> 
> hmm. can you share the full MiMa error message? I don't understand why it's not happy.

here they are

```log
[error] cats-effect-std: Failed binary compatibility check against org.typelevel:cats-effect-std_2.13:3.3.11 (e:info.apiURL=https://typelevel.org/cats-effect/api/3.x/, e:info.versionScheme=early-semver)! Found 2 potential problems (filtered 18)
[error]  * abstract synthetic method cats$effect$std$PQueueSink$$super$tryOfferN(scala.collection.immutable.List,cats.Monad)java.lang.Object in interface cats.effect.std.PQueueSink is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.effect.std.PQueueSink.cats$effect$std$PQueueSink$$super$tryOfferN")
[error]  * abstract synthetic method cats$effect$std$PQueueSource$$super$tryTakeN(scala.Option,cats.Monad)java.lang.Object in interface cats.effect.std.PQueueSource is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.effect.std.PQueueSource.cats$effect$std$PQueueSource$$super$tryTakeN")
```

neomaclin:
[fb356b7](https://github.com/typelevel/cats-effect/pull/3930/commits/fb356b75ea75b13dcd0e9f775989ea298f0fdd96) with the implementaion on the overrided method was ok with MiMa (and other build checks). let me know if I should go back to that for succcess build of the PR. oh. and since I don't have the write access, I wouldn't be able to merge anything even if the build is/was ok.

armanbilge:
We should avoid duplicating the implementation. Since the `super` trick is not possible, we can extract the implementation to a method on the companion object, and then both traits can delegate to that.

neomaclin:
some how the DispatcherSpec comes back

```
[error]     x raise an error on leaked runner
[error]      Expected: java.lang.IllegalStateException. Got nothing (DispatcherSpec.scala:269)
```

samspills:
> some how the DispatcherSpec comes back
> 
> ```
> [error]     x raise an error on leaked runner
> [error]      Expected: java.lang.IllegalStateException. Got nothing (DispatcherSpec.scala:269)
> ```

@neomaclin that's unrelated to your PR, don't worry :) It's a known flake: https://github.com/typelevel/cats-effect/issues/3776

neomaclin:
if this is good to go in, please go ahead, ( I can't do it myself..)

std/shared/src/main/scala/cats/effect/std/PQueue.scala
```diff
@@ -302,54 +251,26 @@ object PQueueSource {
...
-   * method that recursively runs `tryOffer` and does not offer any additional performance
-   * benefits.
+   * Attempts to enqueue the given elements at the back of the queue without semantically
+   * blocking. If an item in the list cannot be enqueued, the remaining elements will be
+   * returned. This is a convenience method that recursively runs `tryOffer` and does not offer
+   * any additional performance benefits.
    *
    * @param list
-   *   the elements to be put in the PQueue
+   *   the elements to be put at the back of the queue
```
durban: "at the back of the queue" is not exactly correct here, as this is a priority queue.

neomaclin: > "at the back of the queue" is not exactly correct here, as this is a priority queue.

reverting the related docs for tryOfferN in PQueueSink

durban (approved):

djspiewak:
Thank you for taking this on!

PR #3934, opened 2024-01-08
Apply `@static` annotation internally

armanbilge:
Liberally applies the Scala 3 [`@static` annotation](https://www.scala-lang.org/api/current/scala/annotation/static.html) to CE internals (via the scalac-compat shim). These `val`s and `def`s now compile to and are accessed as true Java `static`s. In theory this could give us a performance boost, at worst it shouldn't hurt.

So far I have not figured out how to apply this annotation binary-compatibly for public-facing APIs.
- https://github.com/lampepfl/dotty/issues/19394

### DIFF
--- a/core/jvm-native/src/main/scala/cats/effect/ByteStack.scala
+++ b/core/jvm-native/src/main/scala/cats/effect/ByteStack.scala
@@ -18,7 +18,7 @@ package cats.effect
 
 import org.typelevel.scalaccompat.annotation._
 
-private final class ByteStack private ()
+private[effect] final class ByteStack
 
 private object ByteStack {
 
--- a/core/jvm-native/src/main/scala/cats/effect/ByteStack.scala
+++ b/core/jvm-native/src/main/scala/cats/effect/ByteStack.scala
@@ -16,7 +16,7 @@
 
 package cats.effect
 
-import org.typelevel.scalaccompat.annotation._
+import Platform.static
 
 private[effect] final class ByteStack
 
@@ -24,7 +24,7 @@ private object ByteStack {
 
   type T = Array[Int]
 
-  @static3 final def toDebugString(
+  @static final def toDebugString(
       stack: Array[Int],
       translate: Byte => String = _.toString): String = {
     val count = size(stack)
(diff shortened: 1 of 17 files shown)


core/jvm-native/src/main/scala/cats/effect/ByteStack.scala
```diff
@@ -16,11 +16,17 @@
 
 package cats.effect
 
+import org.typelevel.scalaccompat.annotation._
+
+private final class ByteStack private ()
+
 private object ByteStack {
```
armanbilge: To have statics we need the corresponding `class` to exist.

core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
```diff
@@ -149,7 +151,7 @@ private final class CallbackStack[A](private[this] var callback: A => Unit)
 }
 
 private object CallbackStack {
-  def apply[A](cb: A => Unit): CallbackStack[A] =
+  @static3 def of[A](cb: A => Unit): CallbackStack[A] =
```
armanbilge: You cannot have both an instance and static method of the same name.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1109,6 +1111,9 @@ private[effect] trait IOLowPriorityImplicits {
 
 object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
 
+  @static3 private[this] val _alignForIO = new IOAlign
+  @static3 private[this] val _asyncForIO: kernel.Async[IO] = new IOAsync
```
armanbilge: `@static val`s must be declared first in the `object`.

armanbilge:
Here's a diff of a decompiled `IOFiber` before/after.

https://gist.github.com/armanbilge/c5f9241d4f51b525312353803291a513/revisions#diff-0f933bd31e78a61734764d53158dd85a52ba5cd21c5e688eef4dd219a5fbe955

armanbilge:
- https://github.com/lampepfl/dotty/issues/19396

smarter:
`@static` is basically an experimental feature that was never properly spec-ed or tested, I'd be wary about relying on it. In particular note that a lambda inside an `@static` method will become a non-static method, which can lead to interesting initialization order issues.

armanbilge:
> I'd be wary about relying on it

Thanks, good to know. At least if we can use it internally to optimize our bytecode seems worthwhile (worst case, we rip it back out). But seems like we better not expose any public static APIs at this time.

armanbilge:
- https://github.com/lampepfl/dotty/issues/19441

djspiewak (approved):

Issue #3940, opened 2024-01-09
`CallbackStack#pack` can double-count removals when called concurrently


armanbilge:
h/t @mernst-github in https://github.com/typelevel/cats-effect/issues/3935#issuecomment-1882084648. I've captured this as a test case in 6e2e87d57c7f7db7295dd7cdb5a8f2af9963c0a9, which may non-deterministically fail with:
```
[error]   x handle race conditions in pack
[error]    3 != 2 (CallbackStackSpec.scala:49)
```

mernst-github:
Probably worth carrying over from #3935 that this corrupts the IODeferred `clearCounter` and leads to unreliable `pack` invocations, effectively a memory leak.
I don't have deep insights into the design of the callback stack, but on first glance I would try to avoid concurrent `pack`s altogether (code bails anyway when it detects one), I find it hard to reason about its safety. Guarding `pack` at the stack _root_ with an atomic (without atomics for the actual linked stack) sounds more robust to me.

armanbilge:
> but on first glance I would try to avoid concurrent `pack`s altogether (code bails anyway when it detects one) ... Guarding `pack` at the stack _root_ with an atomic (without atomics for the actual linked stack) sounds more robust to me.

Thanks, this is a really interesting idea!! @samspills and I are giving it a try :)

durban:
After some reading, it turns out, that it is "well-known" (LOL), that removing from a linked list with a single CAS is incorrect; we just didn't notice it :-). There is some reading about it [here](https://www.liblfds.org/downloads/white%20papers/%5BLinked%20List%5D%20-%20%5BHarris%5D%20-%20A%20Pragmatic%20Implementation%20of%20Non-Blocking%20Linked-Lists.pdf) and in the gigantic comment in the openjdk ConcurrentSkipListMap. The short version is that fixing it requires marking the `next` pointer with a CAS before unlinking the element. (Fun fact: our [TimerSkipList](https://github.com/typelevel/cats-effect/blob/series/3.x/core/jvm/src/main/scala/cats/effect/unsafe/TimerSkipList.scala) already does this, as it is a port of (some of) CSLM.)

We can do the marking for `CallbackStack` if we need to. But doing something simpler/faster by using the fact that the `CallbackStack` is _not_ a general purpose linked list is probably a good idea (like what you discussed).

mernst-github:
This is gone in 3.5.3, thanks!

armanbilge:
Thanks for all your help!

PR #3942, opened 2024-01-10
Implement `Fiber#join` via `asyncCheckAttempt`

armanbilge:
Sam and I noticed this while working on https://github.com/typelevel/cats-effect/issues/3940. Probably the old implementation was bugged due to the second race condition described in https://github.com/typelevel/cats-effect/issues/3087?


armanbilge:
[Discussed on Discord](https://discord.com/channels/632277896739946517/839263556754472990/1194702926191546574), there is no bug, but this is more "proper" 😅

durban (approved):
Yeah, I don't know how this could cause any actual problems, but it's definitely more proper this way.

Issue #3584, opened 2023-05-01
Can we push to make `IOFiber` fit in 64 bytes?


vasilmkd:
As of Cats Effect 3.5.0-RC5, this is the memory layout of `IOFiber`:

```
cats.effect.IOFiber object internals:
OFF  SZ                                TYPE DESCRIPTION               VALUE
  0   8                                     (object header: mark)     N/A
  8   4                                     (object header: class)    N/A
 12   4                                 int AtomicBoolean.value       N/A
 16   4                                 int IOFiber.masks             N/A
 20   1                                byte IOFiber.resumeTag         N/A
 21   1                             boolean IOFiber.canceled          N/A
 22   1                             boolean IOFiber.finalizing        N/A
 23   1                                     (alignment/padding gap)   
 24   4      scala.collection.immutable.Map IOFiber.localState        N/A
 28   4   scala.concurrent.ExecutionContext IOFiber.currentCtx        N/A
 32   4              cats.effect.ArrayStack IOFiber.objectState       N/A
 36   4              cats.effect.ArrayStack IOFiber.finalizers        N/A
 40   4           cats.effect.CallbackStack IOFiber.callbacks         N/A
 44   4                    java.lang.Object IOFiber.resumeIO          N/A
 48   4        cats.effect.unsafe.IORuntime IOFiber.runtime           N/A
 52   4      cats.effect.tracing.RingBuffer IOFiber.tracingEvents     N/A
 56   4                               int[] IOFiber.conts             N/A
 60   4          cats.effect.kernel.Outcome IOFiber.outcome           N/A
 64   4                      cats.effect.IO IOFiber._cancel           N/A
 68   4                      cats.effect.IO IOFiber._join             N/A
Instance size: 72 bytes
Space losses: 1 bytes internal + 0 bytes external = 1 bytes total
```

This is already amazing, considering that we were pushing well over 100 bytes around CE 3.3. Does it even make sense to try to get it any lower? My idea is that some x86 CPUs still have 64 byte cache line sizes. This is probably not true for ARM silicon.

Some ideas on the data structures. `localState` seems very rarely used, IMO, it doesn't need to be a direct field. Could it hold a permanent place as the 0th index of `objectState`? Could this trick also be used to store other data? On the surface, the only problem I can foresee is that it would force the backing array of `objectState` to be `init`ialized earlier than it is now. Now, it is initialized when we first run the fiber, not when it is allocated. That could be a problem.

vasilmkd:
[...]

PR #3943, opened 2024-01-10
Refactor `CallbackStack` based on new "pack locking" strategy

### LINKED ISSUE #3940: `CallbackStack#pack` can double-count removals when called concurrently

armanbilge:
Fixes https://github.com/typelevel/cats-effect/issues/3940. Very fun collaboration with @samspills but credit to @mernst-github for proposing the idea of guarding `pack` in https://github.com/typelevel/cats-effect/issues/3940#issuecomment-1883123013.

The new callback stack is made up of two data types:
1. `CallbackStack.Node`, a singly-linked list of callbacks
2. `CallbackStack` itself, which holds:
    - an `AtomicReference` to the head `Node`, and
    - an `AtomicBoolean` lock `allowedToPack`, used for guarding `pack`

Pushing to the stack is a straighforward CAS of the head.

`pack` is the only method that modifies the list.
- the `AtomicBoolean` lock guarantees that only one `pack` operation is running at a time, and is also used to publish the modifications to subsequent packs
- if a call to `pack` is unable to acquire the lock, it gives up immediately

Invoking the callbacks is not guarded by the lock. This means that the modifications by `pack` may not be visible, and in fact `pack`ing can even occur concurrently while iterating over the list. This is actually okay:
- the list is only ever modified to _remove_ `Node`s, since additions happen by CASing the pointer to the head `Node`
- thus, if while iterating over the list we follow a stale pointer to a next `Node` that was actually removed, we take a "detour" but do eventually visit all the `Node`s in the list

### DIFF
--- a/core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
+++ b/core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
@@ -18,17 +18,94 @@ package cats.effect
 
 import scala.annotation.tailrec
 
+import java.util.concurrent.atomic.AtomicBoolean
 import java.util.concurrent.atomic.AtomicReference
+
+import CallbackStack.Node
+
+private final class CallbackStack[A](private[this] var callback: A => Unit)
+    extends AtomicReference[Node[A]] {
+  head =>
+
+  private[this] val allowedToPack = new AtomicBoolean(true)
+
+  def push(cb: A => Unit): Node[A] = {
+    val newHead = new Node(cb)
+
+    @tailrec
+    def loop(): CallbackStack[A] = {
+      val currentHead = head.get()
+      newHead.next = currentHead
+
+      if (!head.compareAndSet(currentHead, newHead))
+        loop()
+      else
+        newHead
+    }
+
+    loop()
+  }
+
+  def unsafeSetCallback(cb: A => Unit): Unit = {
+    callback = cb
+  }
+
+  /**
+   * Invokes *all* non-null callbacks in the queue, starting with the current one. Returns true
+   * iff *any* callbacks were invoked.
+   */
+  def apply(a: A): Boolean = {
+    while (!allowedToPack.compareAndSet(true, false)) {
+      // spinloop
+    }
+
+    val cb = callback
+    var invoked = if (cb != null) {
+      cb(a)
+      true
+    } else {
+      false
+    }
+    var currentNode = head.get()
+
+    while (currentNode ne null) {
+      val cb = currentNode.getCallback()
+      if (cb != null) {
+        cb(a)
+        invoked = true
+      }
+      currentNode = currentNode.next
+    }
+
+    invoked
+  }
+}
+
+private object CallbackStack {
+  private[CallbackStack] final class Node[A](
+      private[this] var callback: A => Unit,
+  ) {
+    var next: Node[A] = _
+
+    def getCallback(): A => Unit = callback
+
+    def clear(): Unit = {
+      callback = null
+    }
+  }
+}
 
 private final class CallbackStack[A](private[this] var callback: A => Unit)
     extends AtomicReference[CallbackStack[A]] {
+
+  val allowedToPack = new AtomicBoolean(true)
 
   def push(next: A => Unit): CallbackStack[A] = {
     val attempt = new CallbackStack(next)
 
     @tailrec
     def loop(): CallbackStack[A] = {
-      val cur = get()
+      val cur = head.get()
       attempt.lazySet(cur)
 
       if (!compareAndSet(cur, attempt))
@@ -106,14 +183,20 @@ private final class CallbackStack[A](private[this] var callback: A => Unit)
    * (amortized). This still biases the optimizations towards the head of the list, but ensures
    * that packing will still inevitably reach all of the garbage cells.
    */
-  def pack(bound: Int): Int = {
-    // the first cell is always retained
-    val got = get()
-    if (got ne null)
-      got.packInternal(bound, 0, this)
-    else
+  def pack(bound: Int): Int =
+    if (allowedToPack.compareAndSet(true, false)) {
+      // the first cell is always retained
+      val got = get()
+      val rtn =
+        if (got ne null)
+          got.packInternal(bound, 0, this)
+        else
+          0
+      allowedToPack.set(true)
+      rtn
+    } else {
       0
-  }
+    }
 
   @tailrec
   private def packInternal(bound: Int, removed: Int, parent: CallbackStack[A]): Int = {
--- a/core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
+++ b/core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
@@ -181,8 +181,10 @@ private object CallbackStack {
           val prev = root.get()
           if (prev.next eq this) { // prev is our new parent, we are its tail
             this.packTail(bound, removed, prev)
-          } else { // we were unable to remove ourselves, but we can still pack our tail
+          } else if (next != null) { // we were unable to remove ourselves, but we can still pack our tail
             next.packTail(bound - 1, removed, this)
+          } else {
+            removed
           }
         }
       } else {
--- a/core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
+++ b/core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
@@ -84,10 +84,12 @@ private final class CallbackStack[A](private[this] var callback: A => Unit)
   }
 
   /**
-   * Removes the callback referenced by a handle.
+   * Removes the callback referenced by a handle. Returns `true` if the data structure was
+   * cleaned up immediately, `false` if a subsequent call to [[pack]] is required.
    */
-  def clearHandle(handle: CallbackStack.Handle[A]): Unit = {
+  def clearHandle(handle: CallbackStack.Handle[A]): Boolean = {
     handle.clear()
+    false
   }
 
   /**
@@ -139,6 +141,8 @@ private final class CallbackStack[A](private[this] var callback: A => Unit)
     } else {
       0
     }
+
+  override def toString(): String = s"CallbackStack($callback, ${get()})"
 
 }
 
@@ -231,5 +235,7 @@ private object CallbackStack {
         }
       }
     }
+
+    override def toString(): String = s"Node($callback, $next)"
   }
 }
--- a/core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
+++ b/core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
@@ -197,7 +197,7 @@ private object CallbackStack {
           0
         } else {
           if (bound > 0)
-            next.packTail(bound - 1, 0, this)
+            next.packTail(bound - 1, removed, this)
           else
             0
         }
--- a/core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
+++ b/core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
@@ -172,10 +172,10 @@ private object CallbackStack {
             removed + 1
           } else {
             // note this can cause the bound to go negative, which is fine
-            root.get().packHead(bound - 1, removed + 1, root)
+            next.packHead(bound - 1, removed + 1, root)
           }
-        } else { // get the new top of the stack and start over
-          root.get().packHead(bound, removed, root)
+        } else { // we were unable to remove ourselves, but we can still pack our tail
+          packTail(bound - 1, removed, this)
         }
       } else {
         if (next == null) {
--- a/core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
+++ b/core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
@@ -59,7 +59,7 @@ private final class CallbackStack[A](private[this] var callback: A => Unit)
    * iff *any* callbacks were invoked.
    */
   def apply(a: A): Boolean = {
-    // TODO should we read allowedToPack for memory effect?
+    // see also note about data races in Node#packTail
 
     val cb = callback
     var invoked = if (cb != null) {
--- a/core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
+++ b/core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
@@ -21,6 +21,7 @@ import scala.annotation.tailrec
 import java.util.concurrent.atomic.AtomicBoolean
 import java.util.concurrent.atomic.AtomicReference
 
+import CallbackStack.Handle
 import CallbackStack.Node
 
 private final class CallbackStack[A](private[this] var callback: A => Unit)
@@ -29,11 +30,15 @@ private final class CallbackStack[A](private[this] var callback: A => Unit)
 
   private[this] val allowedToPack = new AtomicBoolean(true)
 
-  def push(cb: A => Unit): Node[A] = {
+  /**
+   * Pushes a callback to the top of the stack. Returns a handle that can be used with
+   * [[clearHandle]] to clear the callback.
+   */
+  def push(cb: A => Unit): Handle[A] = {
     val newHead = new Node(cb)
 
     @tailrec
-    def loop(): CallbackStack[A] = {
+    def loop(): Handle[A] = {
       val currentHead = head.get()
       newHead.next = currentHead
 
@@ -55,9 +60,7 @@ private final class CallbackStack[A](private[this] var callback: A => Unit)
    * iff *any* callbacks were invoked.
    */
   def apply(a: A): Boolean = {
-    while (!allowedToPack.compareAndSet(true, false)) {
-      // spinloop
-    }
+    // TODO should we read allowedToPack for memory effect?
 
     val cb = callback
     var invoked = if (cb != null) {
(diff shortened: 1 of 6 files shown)


core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
```diff
@@ -48,35 +58,46 @@ private final class CallbackStack[A](private[this] var callback: A => Unit)
    * Invokes *all* non-null callbacks in the queue, starting with the current one. Returns true
    * iff *any* callbacks were invoked.
    */
-  @tailrec
-  def apply(oc: A, invoked: Boolean): Boolean = {
-    val cb = callback
+  def apply(a: A): Boolean = {
+    // TODO should we read allowedToPack for memory effect?
```
armanbilge: As explained above, it's not strictly necessary, but if we did read `allowedToPack` then at least some of the `pack`ing would be guaranteed to be visible. But I'm not sure if this additional synchronization is worth it: the modifications may already be visible fortuitously and we still need to be prepared for concurrent `pack`ing anyway.

core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
```diff
@@ -48,35 +58,46 @@ private final class CallbackStack[A](private[this] var callback: A => Unit)
...
-  def currentHandle(): CallbackStack.Handle = 0
-
-  def clear(): Unit = lazySet(null)
+  /**
+   * Nulls all references in this callback stack.
+   */
+  def clear(): Unit = {
+    callback = null
+    head.lazySet(null)
+  }
```
armanbilge, line 100: I wonder if we need to be more aggressive about iterating the list and `null`ing all the callbacks and `next`s. The concern is if someone holds onto a `Handle` for cancelation, they could actually be keeping the entire structure in memory.

core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
```diff
@@ -18,23 +18,33 @@ package cats.effect
 
 import scala.annotation.tailrec
 
-import java.util.concurrent.atomic.AtomicReference
+import java.util.concurrent.atomic.{AtomicBoolean, AtomicReference}
+
+import CallbackStack.Handle
+import CallbackStack.Node
 
 private final class CallbackStack[A](private[this] var callback: A => Unit)
```
mernst-github, line 26: This "main" `callback` variable is somewhat of an oddball and complicates api and implementation:
- it can be installed after the fact via `unsafe...` instead of using `push`.
- it cannot be cleared
- it is only used by IOFiber and always null for IODeferred
- case in point: you don't mention it in the PR description ;-)

I'm wondering if it would balance things if it were moved into IOFiber.

core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
mernst-github: Good point, I didn't consider concurrent `pack` and `apply`. If there were any concerns regarding correctness (boils down to: is it ok for `pack` to rewrite the edges pointing to already published data without synchronization?), could even consider spinning on the `allowedToPack` lock and own the list during `apply`? That might even amortize because apply won't take the packed "detour"?

mernst-github (commented):
Looks great! Just a couple of thoughts:

core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
armanbilge: > This "main" `callback` variable is somewhat of an oddball and complicates api and implementation:

Eh, you're not wrong, but it's fine. `CallbackStack` is not a general purpose data structure, it's a specialized data structure designed specifically for our needs. It's allowed to be odd.

> I'm wondering if it would balance things if it were moved into IOFiber.

This would increase the footprint of `IOFiber`, see https://github.com/typelevel/cats-effect/issues/3584.

core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
armanbilge: > could even consider spinning on the `allowedToPack` lock and own the list during `apply`

Sam and I considered this idea but decided against it. We could be spinning for an unbounded time while we wait for `pack` to complete iterating over a very long list. Or what if we lose the race and another `pack` acquires the lock? For sure, these are worst case pathologies, but since owning the lock is not required to safely iterate the list, spinning until we acquire doesn't seem worth the risk.

core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
djspiewak: For posterity's sake, we're obsessed about `IOFiber`'s memory footprint because we've discovered that this is actually one of the primary performance factors for the whole library. We create so many new fibers all of the time under perfectly normal circumstances that even a few extra bytes in the object header can have a drastic impact on the top-line numbers. In fact, as that issue notes, we're pretty sure that if we can squeeze the footprint down just a bit more, we could probably ladder up a significant step in overall performance due to cache alignment.

tests/shared/src/test/scala/cats/effect/CallbackStackSpec.scala
```diff
@@ -21,10 +21,40 @@ class CallbackStackSpec extends BaseSpec {
...
       val stack = CallbackStack[Unit](null)
-      val pushed = stack.push(_ => ())
-      val handle = pushed.currentHandle()
-      pushed.clearCurrent(handle)
-      stack.pack(1) must beEqualTo(1)
+      val handle = stack.push(_ => ())
+      stack.push(_ => ())
+      val removed = stack.clearHandle(handle)
+      if (removed)
+        stack.pack(1) must beEqualTo(0)
```
djspiewak: Fwiw, `mustEqual` is shorter :)

tests/shared/src/test/scala/cats/effect/CallbackStackSpec.scala
```diff
@@ -21,10 +21,40 @@ class CallbackStackSpec extends BaseSpec {
...
+
+      IO(CallbackStack[Unit](null)).flatMap { stack =>
+        val pushClearPack = for {
+          handle <- IO(stack.push(_ => ()))
+          removed <- IO(stack.clearHandle(handle))
+          packed <- IO(stack.pack(1))
+        } yield (if (removed) 1 else 0) + packed
+
+        pushClearPack
+          .both(pushClearPack)
```
djspiewak: This is only a two-way race. Is it worth tossing a few thousand more fibers at this?

tests/shared/src/test/scala/cats/effect/CallbackStackSpec.scala
```diff
@@ -21,10 +21,40 @@ class CallbackStackSpec extends BaseSpec {
...
+          .product(IO(stack.pack(1)))
+          .flatMap { case ((x, y), z) => IO((x + y + z) must beEqualTo(2)) }
+          .replicateA_(1000)
```
djspiewak:
```suggestion
          .parReplicateA_(1000)
```

Runs faster, particularly locally.

tests/shared/src/test/scala/cats/effect/CallbackStackSpec.scala
```diff
@@ -21,10 +21,40 @@ class CallbackStackSpec extends BaseSpec {
...
+      }
+    }
+
+    "pack runs concurrently with clear" in real {
+      IO {
+        val stack = CallbackStack[Unit](null)
+        val handle = stack.push(_ => ())
+        stack.clearHandle(handle)
+        stack
+      }.flatMap(stack => IO(stack.pack(1)).both(IO(stack.clear()))).replicateA_(1000).as(ok)
```
djspiewak: Same comment as above regarding parallelism factor.

```suggestion
      }.flatMap(stack => IO(stack.pack(1)).both(IO(stack.clear()))).parReplicateA_(1000).as(ok)
```

djspiewak (changes requested):
Love this. Some trivial changes and a couple simple questions but otherwise this is great.

core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
durban: I'd say yes, read `allowedToPack` here, because: what will we do here? We'll call all the `cb`s in the list, right? Which are cont's `resume`s, so when we call them, each will do 1-2 CASes. If we can avoid calling some of them (the cleared ones) by doing a volatile read, that seems like a win. (Of course we might already not call them, because we might already see that they're null... so I'm not sure actually...)

core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
durban: I think that's a valid concern. Maybe we could do that during `apply`? It's a little strange, because `apply` and `clear` are separate, but a call to `clear` always follows `apply`, doesn't it?

core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
```diff
@@ -106,51 +127,121 @@ private final class CallbackStack[A](private[this] var callback: A => Unit)
...
-    else
+  def pack(bound: Int): Int =
+    if (allowedToPack.compareAndSet(true, false)) {
+      val got = head.get()
+      val rtn =
+        if (got ne null)
+          got.packHead(bound, 0, this)
+        else
+          0
+      allowedToPack.set(true)
```
durban: Maybe do this in a  `finally`, since we're releasing a lock?

core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
armanbilge: That's a really interesting point :) it's essentially gambling:
- we do a volatile read now, but get a pretty good guarantee we won't do unnecessary CASes
- we take our chances and if we are lucky we got away with doing the minimum synchronization necessary

core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
armanbilge: I assume `finally` isn't free, and in theory the code between acquiring/releasing should not throw unless something fatal happened 😅

tests/shared/src/test/scala/cats/effect/CallbackStackSpec.scala
armanbilge: I think I'm going to keep this one sequential. Each replicate itself now involves thousands of fibers. Also, by running sequentially we are able to use the same `CallbackStack` instance for every replicate as an additional test of robustness.

core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
armanbilge: > It's a little strange, because `apply` and `clear` are separate, but a call to `clear` always follows `apply`, doesn't it?

I agree. I'd be happy to see `clear` go away and we do the clearing as part of `apply`.

Having a separate `clear` seems important here. This code surprised me.

https://github.com/typelevel/cats-effect/blob/02076374689dcced923d10938117b182cf76d580/core/shared/src/main/scala/cats/effect/IOFiber.scala#L1075-L1084

I think the only callback that can throw is one installed at construction e.g. via `IO#unsafeRunAsync`. But it seems better to `try`/`catch` where the callback is defined, and only install non-throwing callbacks in `IOFiber`. It especially strikes me as odd if the other callbacks are not evaluated, because the first one threw.

However, these changes seem involved enough to warrant their own PR ...

core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
djspiewak: The cleared ones are nulled out though, so we shouldn't have any penalty there besides the node read, or am I missing something?

armanbilge: > The cleared ones are nulled out though

Yes, but currently there is no guarantee that this is published either. They are ordinary writes/reads.

core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
djspiewak: The cost of a `finally` is really small, especially since you only need one here. Agreed this shouldn't really happen, but if it happens it's going to be a hell of a bug to track down. Worth being safe imo.

djspiewak:
Looks like a perfect wash (within margin for error) in terms of performance. Good to go on that front!

## Before

```
[info] Benchmark                                             (count)  (cpuTokens)   (size)   Mode  Cnt      Score     Error    Units
[info] DeepBindBenchmark.async                                   N/A          N/A    10000  thrpt   10   2935.967 ± 102.359    ops/s
[info] DeepBindBenchmark.delay                                   N/A          N/A    10000  thrpt   10   9742.384 ±  42.545    ops/s
[info] DeepBindBenchmark.pure                                    N/A          N/A    10000  thrpt   10  11626.005 ± 170.858    ops/s
[info] DeferredBenchmark.cancel                                   10          N/A      N/A  thrpt   10    104.204 ±   0.456    ops/s
[info] DeferredBenchmark.cancel                                  100          N/A      N/A  thrpt   10     13.537 ±   0.074    ops/s
[info] DeferredBenchmark.cancel                                 1000          N/A      N/A  thrpt   10      1.312 ±   0.008    ops/s
[info] DeferredBenchmark.complete                                 10          N/A      N/A  thrpt   10    110.424 ±   0.836    ops/s
[info] DeferredBenchmark.complete                                100          N/A      N/A  thrpt   10     26.084 ±   0.109    ops/s
[info] DeferredBenchmark.complete                               1000          N/A      N/A  thrpt   10      5.073 ±   0.287    ops/s
[info] DeferredBenchmark.getAfter                                 10          N/A      N/A  thrpt   10   7059.728 ±  19.859    ops/s
[info] DeferredBenchmark.getAfter                                100          N/A      N/A  thrpt   10   2033.584 ±  10.383    ops/s
[info] DeferredBenchmark.getAfter                               1000          N/A      N/A  thrpt   10    247.342 ±   0.501    ops/s
[info] DeferredBenchmark.getBefore                                10          N/A      N/A  thrpt   10   3994.309 ±  15.618    ops/s
[info] DeferredBenchmark.getBefore                               100          N/A      N/A  thrpt   10    798.325 ±   1.832    ops/s
[info] DeferredBenchmark.getBefore                              1000          N/A      N/A  thrpt   10     81.027 ±   0.163    ops/s
[info] ParallelBenchmark.parTraverse                             N/A        10000     1000  thrpt   10    878.708 ±  25.632    ops/s
[info] ParallelBenchmark.traverse                                N/A        10000     1000  thrpt   10     70.062 ±   0.050    ops/s
[info] ShallowBindBenchmark.async                                N/A          N/A    10000  thrpt   10   2157.575 ±   3.120    ops/s
[info] ShallowBindBenchmark.delay                                N/A          N/A    10000  thrpt   10  10052.708 ±  54.888    ops/s
[info] ShallowBindBenchmark.pure                                 N/A          N/A    10000  thrpt   10  10539.875 ± 135.470    ops/s
[info] WorkStealingBenchmark.alloc                               N/A          N/A  1000000  thrpt   10     14.151 ±   0.111  ops/min
[info] WorkStealingBenchmark.manyThreadsSchedulingBenchmark      N/A          N/A  1000000  thrpt   10     50.544 ±   0.753  ops/min
[info] WorkStealingBenchmark.runnableScheduling                  N/A          N/A  1000000  thrpt   10   2958.068 ±  15.379  ops/min
[info] WorkStealingBenchmark.runnableSchedulingScalaGlobal       N/A          N/A  1000000  thrpt   10   2195.662 ±   5.718  ops/min
[info] WorkStealingBenchmark.scheduling                          N/A          N/A  1000000  thrpt   10     53.854 ±   2.195  ops/min
```

## After

```
[info] Benchmark                                             (count)  (cpuTokens)   (size)   Mode  Cnt      Score     Error    Units
[info] DeepBindBenchmark.async                                   N/A          N/A    10000  thrpt   10   3068.624 ±  16.441    ops/s
[info] DeepBindBenchmark.delay                                   N/A          N/A    10000  thrpt   10   9853.126 ± 124.913    ops/s
[info] DeepBindBenchmark.pure                                    N/A          N/A    10000  thrpt   10  11664.990 ±  94.674    ops/s
[info] DeferredBenchmark.cancel                                   10          N/A      N/A  thrpt   10    107.015 ±   0.686    ops/s
[info] DeferredBenchmark.cancel                                  100          N/A      N/A  thrpt   10     14.352 ±   0.083    ops/s
[info] DeferredBenchmark.cancel                                 1000          N/A      N/A  thrpt   10      1.326 ±   0.016    ops/s
[info] DeferredBenchmark.complete                                 10          N/A      N/A  thrpt   10    110.982 ±   1.658    ops/s
[info] DeferredBenchmark.complete                                100          N/A      N/A  thrpt   10     26.111 ±   0.183    ops/s
[info] DeferredBenchmark.complete                               1000          N/A      N/A  thrpt   10      5.129 ±   0.271    ops/s
[info] DeferredBenchmark.getAfter                                 10          N/A      N/A  thrpt   10   6990.249 ±  47.301    ops/s
[info] DeferredBenchmark.getAfter                                100          N/A      N/A  thrpt   10   2042.712 ±  21.909    ops/s
[info] DeferredBenchmark.getAfter                               1000          N/A      N/A  thrpt   10    246.581 ±   1.930    ops/s
[info] DeferredBenchmark.getBefore                                10          N/A      N/A  thrpt   10   3941.678 ±  24.476    ops/s
[info] DeferredBenchmark.getBefore                               100          N/A      N/A  thrpt   10    744.944 ±   5.293    ops/s
[info] DeferredBenchmark.getBefore                              1000          N/A      N/A  thrpt   10     80.747 ±   0.431    ops/s
[info] ParallelBenchmark.parTraverse                             N/A        10000     1000  thrpt   10    887.059 ±  23.448    ops/s
[info] ParallelBenchmark.traverse                                N/A        10000     1000  thrpt   10     69.816 ±   0.752    ops/s
[info] ShallowBindBenchmark.async                                N/A          N/A    10000  thrpt   10   2169.835 ±  22.053    ops/s
[info] ShallowBindBenchmark.delay                                N/A          N/A    10000  thrpt   10   9872.628 ± 116.515    ops/s
[info] ShallowBindBenchmark.pure                                 N/A          N/A    10000  thrpt   10  10410.052 ±  75.004    ops/s
[info] WorkStealingBenchmark.alloc                               N/A          N/A  1000000  thrpt   10     14.018 ±   0.133  ops/min
[info] WorkStealingBenchmark.manyThreadsSchedulingBenchmark      N/A          N/A  1000000  thrpt   10     48.573 ±   2.040  ops/min
[info] WorkStealingBenchmark.runnableScheduling                  N/A          N/A  1000000  thrpt   10   2997.308 ±  10.291  ops/min
[info] WorkStealingBenchmark.runnableSchedulingScalaGlobal       N/A          N/A  1000000  thrpt   10   2239.119 ±  12.325  ops/min
[info] WorkStealingBenchmark.scheduling                          N/A          N/A  1000000  thrpt   10     52.694 ±   1.716  ops/min
```

djspiewak (dismissed):
I'm still leaning toward rolling the dice on *not* fencing in `apply`. Let's see how that treats us though, and if we're seeing problems there, we'll add the fence.

PR #3947, opened 2024-01-12
Fix ordering issue in sequential dispatcher

### LINKED ISSUE #3945: sequential dispatcher: race condition in release `step` causes disorder
Change the `"strictly sequentialize multiple IOs"` test by moving `gate.await` outside of the `use`: [2b9cc73](https://github.com/samspills/cats-effect/pull/3/commits/2b9cc7334054c0e7f352050e3d1bf97fc40864fd) (h/t @rossabaker)

and result ordering is lost b/c the final release `step` races the worker `step`

This is the same race condition that @armanbilge pointed out here: https://github.com/typelevel/cats-effect/pull/3701#discussion_r1240911678

rossabaker:
Passes tests to fix #3945.  Needs a bit more refinement, but I understand there's more discussion behind the Discord Curtain about the rewrite being bumped to 3.5.  I don't care which solution is taken as long as it gets fixed, but pausing here before more time gets wasted on duplicate effort.

### DIFF
--- a/example/jvm/src/main/scala/cats/effect/example/Example.scala
+++ b/example/jvm/src/main/scala/cats/effect/example/Example.scala
@@ -17,8 +17,24 @@
 package cats.effect
 package example
 
+import cats.effect.std._
+
+import scala.concurrent.duration._
+
+import java.util.concurrent.atomic.AtomicReference
+
 object Example extends IOApp.Simple {
 
-  def run: IO[Unit] =
-    (IO(println("started")) >> IO.never).onCancel(IO(println("canceled")))
+  def run: IO[Unit] = {
+    val x = new AtomicReference("")
+    Dispatcher.sequential[IO].use { dispatcher =>
+      IO(
+        (1 to 25).toList.foreach { i =>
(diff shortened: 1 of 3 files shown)


samspills:
@djspiewak Should we keep working on this fix for 3.5.x? That would give #3923 more time + the 3.6 RC cycle and I think that approach makes sense, since I believe the rewrite still has the `not hang when cancelling` issue

djspiewak:
So this effectively retrofits (via convergent evolution) some of the state machine I had to implement for the rewritten `Dispatcher` but on top of the existing leaning tower that is the current implementation. Absent any other factors, this is reasonable, but holy hell it makes me nervous. I fixed so much stuff (including tests that were specifying incorrect or even contradictory behavior) in the rewrite branch.

Given that there's at least two distinct things that need serious hammock thinking time for 3.6, I'm tempted to say I should rebase #3923 and land it in 3.5.x. I'll try to get to the outstanding issue on it this weekend. Also help very much welcome! I honestly think that the version I have in that PR is a lot easier to build on and fix than the one which is in `HEAD`.

PR #3956, opened 2024-01-16
Fix Discord post step in release job

armanbilge:
Really sad screw up 😢


djspiewak (approved):
Agggggghhhhhhhhh so close!

PR #3961, opened 2024-01-16
Refine steps about post-release scripts

armanbilge:
Now that I've actually done a release 😂

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -14,13 +14,13 @@
 1. Publish the release.
 1. Wait for all the CI madness to happen, for the release to announced to Discord, and for the artifacts to sync to Maven Central.
 1. Make sure you're locally updated and on the right major/minor branch (this is the same branch as step 3).
-1. Open a PR to merge the minor branch into the major branch. This is only necessary for patch releases.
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -14,13 +14,13 @@
 1. Publish the release.
 1. Wait for all the CI madness to happen, for the release to announced to Discord, and for the artifacts to sync to Maven Central.
 1. Make sure you're locally updated and on the right major/minor branch (this is the same branch as step 3).
-1. Open a PR to merge the minor branch into the major branch. This is only necessary for patch releases.
+1. Run the following script to open two PRs to (1) update the version in the minor branch and then (2) merge the minor branch into the major branch. This script currently works only for patch releases and not for minor releases, milestones, or release candidates, which need special handling.
```
armanbilge: > This script currently works only for patch releases and not for minor releases, milestones, or release candidates, which need special handling.

- For minor releases, we are not merging the minor branch into the major branch, but we are updating the version.
- For Ms and RCs, we not updating the version, and probably not merging branches either (unless we M/RC a patch, but that seems weird).

samspills (approved):

RELEASING.md
armanbilge: > For Ms and RCs ... and probably not merging branches either

Actually that's not right. Once we hit RCs we are in feature freeze, so it's likely we already have the minor/major branch distinction at that point and need to do the merges.

PR #3968, opened 2024-01-22
Add explanatory comment to Mutex implementation

durban:
Apparently, this thing has a name: `Craig, Landin, and Hagersten lock`.

Further reading (unfortunately I can't link a PDF):
Magnussen, P., A. Landin, and E. Hagersten. Queue locks on cache coherent multiprocessors. 8th Intl. Parallel Processing Symposium, Cancun, Mexico, Apr. 1994.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Mutex.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Mutex.scala
@@ -87,6 +87,12 @@ object Mutex {
   )(
       implicit F: Concurrent[F]
   ) extends Mutex[F] {
+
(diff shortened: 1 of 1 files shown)


armanbilge:
cc @BalmungSan :)

BalmungSan (approved):
It is always reassuring to know that I reinvented the wheel, especially if such wheel was originally done by someone smarter than me.

I will try to find some time to read it, just to make sure they aren't doing anything better than us.

PS: We probably should rename cells to use the same names as them.

armanbilge (approved):
fantastic find!

djspiewak:
Outstanding.

PR #3972, opened 2024-01-26
Fix a few Supervisor bugs

### LINKED ISSUE #3394: Closed `Supervisor` strangeness

durban:
This PR fixes a few bugs in `Supervisor` (and creates an unknown number of new ones ;-).

Fixed bugs (at least the ones I remember spotting; I tried to write tests for them, couldn't for all of them):

- #3394
- With `checkRestart`, `cancel` allowed and didn't wait for one last restart.
- `supervise`/close race (although, due to #3394, close wasn't effective anyway).
- With `checkRestart`, if `cancel` cancelled the working fiber before it did anything, the finalizer (which removes the fiber from the map) did not run. (This was effectively a memory leak.)

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Supervisor.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Supervisor.scala
@@ -329,8 +329,12 @@ object Supervisor {
             case map => (map.updated(token, fiber), true)
           }
 
-        private[this] val allFibers: F[List[Fiber[F, Throwable, _]]] =
+        private[this] val allFibers: F[List[Fiber[F, Throwable, _]]] = {
+          // we're closing, so we won't need the state any more,
+          // so we're using `null` as a sentinel to reject later
+          // insertions in `add`:
           stateRef.getAndSet(null).map(_.values.toList)
+        }
 
         val joinAll: F[Unit] = allFibers.flatMap(_.traverse_(_.join.void))
 
--- a/std/shared/src/main/scala/cats/effect/std/Supervisor.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Supervisor.scala
@@ -19,6 +19,7 @@ package cats.effect.std
 import cats.effect.kernel._
 import cats.effect.kernel.implicits._
 import cats.syntax.all._
+import cats.MonadError
 
 import scala.collection.mutable.ListBuffer
 
@@ -140,14 +141,14 @@ object Supervisor {
 
   private trait State[F[_]] {
     def remove(token: Unique.Token): F[Unit]
-    def add(token: Unique.Token, fiber: Fiber[F, Throwable, _]): F[Unit]
+    def add(token: Unique.Token, fiber: Fiber[F, Throwable, _]): F[Boolean]
     // run all the finalizers
     val joinAll: F[Unit]
     val cancelAll: F[Unit]
   }
 
   private def supervisor[F[_]](
-      mkState: F[State[F]],
+      mkState: Ref[F, Boolean] => F[State[F]],
       await: Boolean,
       checkRestart: Option[Outcome[F, Throwable, _] => Boolean])(
       implicit F: Concurrent[F]): Resource[F, Supervisor[F]] = {
@@ -155,7 +156,7 @@ object Supervisor {
     // intertwined with resource management
     for {
       doneR <- Resource.eval(F.ref(false))
-      state <- Resource.makeCase(mkState) {
+      state <- Resource.makeCase(mkState(doneR)) {
         case (st, Resource.ExitCase.Succeeded) if await => doneR.set(true) >> st.joinAll
         case (st, _) =>
           doneR.set(true) >> { /*println("canceling all!");*/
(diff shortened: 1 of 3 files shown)


tests/shared/src/test/scala/cats/effect/std/SupervisorSpec.scala
```diff
@@ -17,11 +17,17 @@
...
+import cats.syntax.all._
+
 import org.specs2.specification.core.Fragments
 
 import scala.concurrent.duration._
 
-class SupervisorSpec extends BaseSpec {
+class SupervisorSpec extends BaseSpec with DetectPlatform {
+
+  sequential
```
djspiewak: Do we need this?

tests/shared/src/test/scala/cats/effect/std/SupervisorSpec.scala
```diff
@@ -17,11 +17,17 @@
...
+
+  override def executionTimeout = 60.seconds
```
djspiewak: Oooof. Can we lower iteration counts instead?

tests/shared/src/test/scala/cats/effect/std/SupervisorSpec.scala
```diff
@@ -229,5 +235,68 @@ class SupervisorSpec extends BaseSpec {
...
+
+    def superviseFinalizeRace(mkSupervisor: Resource[IO, Supervisor[IO]], task: IO[Unit]) = {
+      val tsk = IO.uncancelable { poll =>
+        mkSupervisor.allocated.flatMap {
+          case (supervisor, close) =>
+            supervisor.supervise(IO.never[Unit]).replicateA(100).flatMap { fibers =>
+              val tryFork = supervisor.supervise(task).map(Some(_)).recover {
+                case ex: IllegalStateException
+                    if ex.getMessage == "supervisor already shutdown" =>
+                  None
```
djspiewak: Might be a bit easier to diagnose an eventual failure here if we convert this into a side-effecting test assertion.

```suggestion
                case ex: IllegalStateException =>
                  ex.getMessage mustEqual "supervisor already shutdown"
                  None
```

tests/shared/src/test/scala/cats/effect/std/SupervisorSpec.scala
```diff
@@ -229,5 +235,68 @@ class SupervisorSpec extends BaseSpec {
...
+    }
+
+    "restart / cancel race" in real {
+      val tsk = constructor(false, Some(_ => true)).use { supervisor =>
+        IO.ref(0).flatMap { counter =>
+          supervisor.supervise(counter.update(_ + 1) *> IO.canceled).flatMap { adaptedFiber =>
+            IO.sleep(100.millis) *> adaptedFiber.cancel *> adaptedFiber.join *> (
+              (counter.get, IO.sleep(100.millis) *> counter.get).flatMapN {
+                case (v1, v2) =>
+                  IO((v1: Int) mustEqual (v2: Int))
```
djspiewak: Are the type ascriptions needed?

tests/shared/src/test/scala/cats/effect/std/SupervisorSpec.scala
```diff
@@ -229,5 +235,68 @@ class SupervisorSpec extends BaseSpec {
...
+              }
+            )
+          }
+        }
+      }
+
+      tsk.parReplicateA_(if (isJVM) 10000 else 5).as(ok)
```
djspiewak: Why do we need to replicate it at all on non-JVM platforms?

Also, I'm assuming this is the test that takes most of the time?

std/shared/src/main/scala/cats/effect/std/Supervisor.scala
```diff
@@ -244,44 +317,60 @@ object Supervisor {
...
+
+        def add(token: Unique.Token, fiber: Fiber[F, Throwable, _]): F[Boolean] =
+          stateRef.modify {
+            case null => (null, false)
+            case map => (map.updated(token, fiber), true)
+          }
 
         private[this] val allFibers: F[List[Fiber[F, Throwable, _]]] =
-          stateRef.get.map(_.values.toList)
+          stateRef.getAndSet(null).map(_.values.toList)
```
djspiewak, line 342: So `null` is the signal that we've decided to kill ourselves? Basically, the invariant here is that `joinAll`/`cancelAll` can only be called at the end of the world, when the `Supervisor` is shutting down. Can we toss that in a comment so we don't forget it?

djspiewak (changes requested):
Ouch this is hairy. Nice work. :) Some minor nits.

djspiewak:
Also it occurs to me that, with this, we might not need the double-check on closure state in #3923

tests/shared/src/test/scala/cats/effect/std/SupervisorSpec.scala
durban: With lowering the iteration counts, I think we will be able to avoid this.

tests/shared/src/test/scala/cats/effect/std/SupervisorSpec.scala
durban: Nope, I don't even remember why I did those :slightly_smiling_face:

tests/shared/src/test/scala/cats/effect/std/SupervisorSpec.scala
durban: Good point; I think I copied the `if (isJVM) ... else 5` part from somewhere.

For me, it was usually `supervise / finalize race with checkRestart` that timed out.

std/shared/src/main/scala/cats/effect/std/Supervisor.scala
durban: There is already a comment in `abstract class State` about `joinAll`/`cancelAll`, but I'll note it here too.

durban:
FTR there was a `Dispatcher` failure (https://github.com/typelevel/cats-effect/actions/runs/7678308491/job/20928046255#step:24:4779), but seemed unrelated to this PR, so I just restarted the CI jobs.

djspiewak (approved):
Awesome!

PR #3973, opened 2024-01-26
Clear the CallbackStack more aggressively

samspills:
I'm still seeing a CallbackStack leak at $work with the most recent 3.5.3 release 😓 I chatted it over with Arman, and we think it's the situation he outlined here: https://github.com/typelevel/cats-effect/pull/3943#discussion_r1448077903

This PR updates `apply` to clean up the stack as it goes, ~and the `clear` method to null everything out much more aggressively~ Actually, I've removed the `clear` method on the stack, and added a `try` / `finally` to the apply so that `apply` will finish tearing down if a callback throws an exception. This is necessary because by tearing down as `apply` goes, `clear` would have no way to access the node that `apply` got to, as the links between `head` and that node would have been broken.

### DIFF
--- a/core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
+++ b/core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
@@ -77,7 +77,9 @@ private final class CallbackStack[A](private[this] var callback: A => Unit)
         currentNode.clear()
         invoked = true
       }
-      currentNode = currentNode.getNext()
+      val nextNode = currentNode.getNext()
+      currentNode.setNext(null)
+      currentNode = nextNode
     }
 
     invoked
@@ -99,8 +101,10 @@ private final class CallbackStack[A](private[this] var callback: A => Unit)
     callback = null
     var currentNode = head.get()
     while (currentNode ne null) {
+      val nextNode = currentNode.getNext()
       currentNode.clear()
-      currentNode = currentNode.getNext()
+      currentNode.setNext(null)
+      currentNode = nextNode
     }
     head.lazySet(null)
   }
--- a/core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
+++ b/core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
@@ -74,6 +74,7 @@ private final class CallbackStack[A](private[this] var callback: A => Unit)
       val cb = currentNode.getCallback()
       if (cb != null) {
         cb(a)
+        currentNode.clear()
         invoked = true
       }
       currentNode = currentNode.getNext()
--- a/core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
+++ b/core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
@@ -97,6 +97,11 @@ private final class CallbackStack[A](private[this] var callback: A => Unit)
    */
   def clear(): Unit = {
     callback = null
+    var currentNode = head.get()
+    while (currentNode ne null) {
+      currentNode.clear()
+      currentNode = currentNode.getNext()
+    }
     head.lazySet(null)
   }
 
--- a/core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
+++ b/core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
@@ -60,29 +60,42 @@ private final class CallbackStack[A](private[this] var callback: A => Unit)
    */
   def apply(a: A): Boolean = {
     // see also note about data races in Node#packTail
-
-    val cb = callback
-    var invoked = if (cb != null) {
-      cb(a)
-      true
-    } else {
-      false
-    }
     var currentNode = head.get()
-
-    while (currentNode ne null) {
-      val cb = currentNode.getCallback()
-      if (cb != null) {
+    try {
+      val cb = callback
+      var invoked = if (cb != null) {
         cb(a)
-        currentNode.clear()
-        invoked = true
+        true
+      } else {
+        false
       }
-      val nextNode = currentNode.getNext()
-      currentNode.setNext(null)
-      currentNode = nextNode
-    }
 
-    invoked
+      // note that we're tearing down the callback stack structure behind us as we go
+      while (currentNode ne null) {
+        val cb = currentNode.getCallback()
+        if (cb != null) {
+          cb(a)
+          currentNode.clear()
+          invoked = true
+        }
+        val nextNode = currentNode.getNext()
+        currentNode.setNext(null)
+        currentNode = nextNode
+      }
+      head.lazySet(null)
+
+      invoked
+    } finally {
+      // if a callback throws, we stop invoking remaining callbacks
+      // but we continue the process of tearing down the stack to prevent memory leaks
+      while (currentNode ne null) {
+        val nextNode = currentNode.getNext()
+        currentNode.clear()
+        currentNode.setNext(null)
+        currentNode = nextNode
+      }
+      head.lazySet(null)
+    }
   }
 
   /**
@@ -92,21 +105,6 @@ private final class CallbackStack[A](private[this] var callback: A => Unit)
   def clearHandle(handle: CallbackStack.Handle[A]): Boolean = {
     handle.clear()
     false
-  }
-
-  /**
-   * Nulls all references in this callback stack.
-   */
-  def clear(): Unit = {
-    callback = null
-    var currentNode = head.get()
-    while (currentNode ne null) {
-      val nextNode = currentNode.getNext()
-      currentNode.clear()
-      currentNode.setNext(null)
-      currentNode = nextNode
-    }
-    head.lazySet(null)
   }
 
   /**
--- a/build.sbt
+++ b/build.sbt
@@ -662,7 +662,9 @@ lazy val core = crossProject(JSPlatform, JVMPlatform, NativePlatform)
       // #3943, refactored internal private CallbackStack data structure
       ProblemFilters.exclude[IncompatibleResultTypeProblem]("cats.effect.CallbackStack.push"),
       ProblemFilters.exclude[DirectMissingMethodProblem](
-        "cats.effect.CallbackStack.currentHandle")
+        "cats.effect.CallbackStack.currentHandle"),
+      // #3973, remove clear from internal private CallbackStack
+      ProblemFilters.exclude[DirectMissingMethodProblem]("cats.effect.CallbackStack.clear")
     ) ++ {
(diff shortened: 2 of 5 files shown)


core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
```diff
@@ -60,26 +60,42 @@ private final class CallbackStack[A](private[this] var callback: A => Unit)
...
+        currentNode.setNext(null)
+        currentNode = nextNode
+      }
+      head.lazySet(null)
+
+      invoked
+    } finally {
+      // if a callback throws, we stop invoking remaining callbacks
+      // but we continue the process of tearing down the stack to prevent memory leaks
+      while (currentNode ne null) {
```
djspiewak, line 91: Why do we need to tear down the full stack? Isn't nulling out the current tail pointer sufficient to break the reference chain?

core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
```diff
@@ -60,26 +60,42 @@ private final class CallbackStack[A](private[this] var callback: A => Unit)
...
+        true
+      } else {
+        false
       }
-      currentNode = currentNode.getNext()
-    }
 
-    invoked
+      // note that we're tearing down the callback stack structure behind us as we go
+      while (currentNode ne null) {
```
djspiewak, line 74: Same comment. Can't we just null out the head and leave the rest alone?

core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
samspills: It doesn't seem so, but I could be wrong? I was looking at a heap dump today with 7.7 million callback stack nodes, only 436 of which had a non-null callback. I chose a random null-callback-having node, and it's GC Root showed a chain of 4.8 million callback stack nodes, and it looked like all of that was held as a reference for an IOFiber finalizer. My thinking is that the fiber is keeping a reference to a `Handle`, then that is keeping everything "under" that node in memory, which turns out to be quite a lot (and the head being nulled out wouldn't break this chain, I don't think, because the head is "above" this node -- at least that's how I see it in my mind)

durban:
Btw, if a callback throws an exception, why aren't we calling the rest of the callbacks? (I realize this is also the current behavior.) I see @armanbilge also already asked this [here](https://github.com/typelevel/cats-effect/pull/3943#discussion_r1451633273).

core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
djspiewak: Well, I was a bit hand-wavy. What I meant by nulling out the head is nulling the tail pointer of the head node, just to break the reference chain. Can't we take that approach?

djspiewak:
> Btw, if a callback throws an exception, why aren't we calling the rest of the callbacks?

Oh that very much seems like a bug. It's harmless right now since `cont`'s `resume` will never throw, but I would prefer to be defensive about it unless it costs us performance to do so.

core/jvm-native/src/main/scala/cats/effect/CallbackStack.scala
armanbilge: No, that's not sufficient. When you install a callback in the stack, you get handed a reference to that `Node` that you can use to clear it it later. If for whatever reason you hang onto that `Node` indefinitely, you are holding a hard reference to the tail of the stack as see by that `Node`.

djspiewak: Oooooh I see. This is very tedious.

djspiewak:
I'll run some benchmarks to sanity check.

djspiewak:
Essentially a wash on performance, at least so far as our benchmarks can measure.

## Before

```
[info] Benchmark                                             (count)  (cpuTokens)   (size)   Mode  Cnt     Score    Error    Units
[info] DeferredBenchmark.cancel                                   10          N/A      N/A  thrpt   10   110.427 ±  0.496    ops/s
[info] DeferredBenchmark.cancel                                  100          N/A      N/A  thrpt   10    14.109 ±  0.055    ops/s
[info] DeferredBenchmark.cancel                                 1000          N/A      N/A  thrpt   10     1.318 ±  0.005    ops/s
[info] DeferredBenchmark.complete                                 10          N/A      N/A  thrpt   10   108.645 ±  0.414    ops/s
[info] DeferredBenchmark.complete                                100          N/A      N/A  thrpt   10    25.782 ±  0.136    ops/s
[info] DeferredBenchmark.complete                               1000          N/A      N/A  thrpt   10     5.018 ±  0.245    ops/s
[info] DeferredBenchmark.getAfter                                 10          N/A      N/A  thrpt   10  6915.081 ± 24.629    ops/s
[info] DeferredBenchmark.getAfter                                100          N/A      N/A  thrpt   10  1929.931 ±  7.287    ops/s
[info] DeferredBenchmark.getAfter                               1000          N/A      N/A  thrpt   10   230.685 ±  1.069    ops/s
[info] DeferredBenchmark.getBefore                                10          N/A      N/A  thrpt   10  3929.464 ±  5.893    ops/s
[info] DeferredBenchmark.getBefore                               100          N/A      N/A  thrpt   10   745.665 ±  1.358    ops/s
[info] DeferredBenchmark.getBefore                              1000          N/A      N/A  thrpt   10    86.061 ±  0.183    ops/s
[info] ParallelBenchmark.parTraverse                             N/A        10000     1000  thrpt   10   875.136 ± 24.879    ops/s
[info] ParallelBenchmark.traverse                                N/A        10000     1000  thrpt   10    70.208 ±  0.165    ops/s
[info] WorkStealingBenchmark.alloc                               N/A          N/A  1000000  thrpt   10    13.987 ±  0.173  ops/min
[info] WorkStealingBenchmark.manyThreadsSchedulingBenchmark      N/A          N/A  1000000  thrpt   10    48.534 ±  2.015  ops/min
[info] WorkStealingBenchmark.runnableScheduling                  N/A          N/A  1000000  thrpt   10  2972.862 ± 20.806  ops/min
[info] WorkStealingBenchmark.runnableSchedulingScalaGlobal       N/A          N/A  1000000  thrpt   10  2229.858 ± 10.108  ops/min
[info] WorkStealingBenchmark.scheduling                          N/A          N/A  1000000  thrpt   10    49.855 ±  3.070  ops/min
```

## After

```
[info] Benchmark                                             (count)  (cpuTokens)   (size)   Mode  Cnt     Score    Error    Units
[info] DeferredBenchmark.cancel                                   10          N/A      N/A  thrpt   10   111.104 ±  0.618    ops/s
[info] DeferredBenchmark.cancel                                  100          N/A      N/A  thrpt   10    13.589 ±  0.070    ops/s
[info] DeferredBenchmark.cancel                                 1000          N/A      N/A  thrpt   10     1.413 ±  0.007    ops/s
[info] DeferredBenchmark.complete                                 10          N/A      N/A  thrpt   10   107.398 ±  0.614    ops/s
[info] DeferredBenchmark.complete                                100          N/A      N/A  thrpt   10    25.566 ±  0.153    ops/s
[info] DeferredBenchmark.complete                               1000          N/A      N/A  thrpt   10     5.072 ±  0.249    ops/s
[info] DeferredBenchmark.getAfter                                 10          N/A      N/A  thrpt   10  6792.078 ± 57.712    ops/s
[info] DeferredBenchmark.getAfter                                100          N/A      N/A  thrpt   10  1920.488 ±  8.127    ops/s
[info] DeferredBenchmark.getAfter                               1000          N/A      N/A  thrpt   10   231.931 ±  0.958    ops/s
[info] DeferredBenchmark.getBefore                                10          N/A      N/A  thrpt   10  3916.262 ±  7.243    ops/s
[info] DeferredBenchmark.getBefore                               100          N/A      N/A  thrpt   10   745.866 ±  1.469    ops/s
[info] DeferredBenchmark.getBefore                              1000          N/A      N/A  thrpt   10    85.573 ±  0.235    ops/s
[info] ParallelBenchmark.parTraverse                             N/A        10000     1000  thrpt   10   867.002 ± 25.050    ops/s
[info] ParallelBenchmark.traverse                                N/A        10000     1000  thrpt   10    70.059 ±  0.122    ops/s
[info] WorkStealingBenchmark.alloc                               N/A          N/A  1000000  thrpt   10    14.005 ±  0.094  ops/min
[info] WorkStealingBenchmark.manyThreadsSchedulingBenchmark      N/A          N/A  1000000  thrpt   10    49.554 ±  2.297  ops/min
[info] WorkStealingBenchmark.runnableScheduling                  N/A          N/A  1000000  thrpt   10  2730.891 ± 86.788  ops/min
[info] WorkStealingBenchmark.runnableSchedulingScalaGlobal       N/A          N/A  1000000  thrpt   10  2228.855 ±  9.531  ops/min
[info] WorkStealingBenchmark.scheduling                          N/A          N/A  1000000  thrpt   10    50.375 ±  5.190  ops/min
```

djspiewak (approved):

aartigao:
@samspills Did you test 3.5.4 to see if the leak was fixed?

We're having a leak too, and Eclipse MAT points to `CallbackStack.Node`. I'm trying to figure out if it's something we're doing wrong and the leak of `Node`s is just a bad usage of the CE API.

image
image

samspills:
@aartigao I tested a snapshot with this commit and the leak was _not_ fixed. I've been poking at it when I can but I haven't been able to dedicate myself to it (kubernetes shields us from most of the OOM pain currently)

I've found that my leak seems to be caused by IOFiber joins, and in that use case the CallbackStack is not `pack`ed the way it is in the Deferred case. To me this says that it is indeed a bad usage of the CE API with IOFiber, but I haven't had time to nail down where (no usage of IOFiber in my app directly). From your class histogram, can I guess that you're using fs2-kafka? My leaking app is also an fs2-kafka app. I have other fs2-kafka apps that are behaving well but if that is a commonality in our cases then that is interesting

aartigao:
>From your class histogram, can I guess that you're using fs2-kafka?

Yup! I've been digging and our leak matches in time when our traffic descends to 0 and there are no more Kafka records to consume. It can be an `fs2-kafka` thing, but we're suspicious about a part of our code (some really complex code) that may be leaking fibers in the wild. In our case it's possible to misuse the CE Fiber API because that code makes use of `racePair` and might be not cancelling the loser fiber.

Let's see, we'll run another test this afternoon. Tomorrow I'll have news 🙃

aartigao:
I can confirm that, in our case, the leak was caused due to a misuse of `IO.racePair`.

PR #3976, opened 2024-01-28
Updated copyright year

djspiewak:


djspiewak:
Okay apparently `headerCreate` can actually just totally bork files? Idk. Did it using search replace this time.

PR #3988, opened 2024-02-03
Updated macos GHA runners to use the new M series

djspiewak:


djspiewak:
Ah we need to update https://github.com/typelevel/download-java. Okay I'll come back to yak shaving another day unless someone else picks it up…

I think we should try to get to this sooner rather than later though, if only to get the higher performance macos builds. The intel runners are incredibly slow.

armanbilge:
> Ah we need to update https://github.com/typelevel/download-java

This action is deprecated, the new sbt-typelevel series doesn't use it anymore. We couldn't easily do that upgrade on this branch, but it's already done on `series/3.x`.

armanbilge:
- https://github.com/actions/runner-images/issues/9369

PR #3999, opened 2024-02-13
Use `dep` instead of `lib` in FAQ docs

ChidiRnweke:
The directive "lib" has been deprecated in favour of the directive "dep". Using  the lib directive "lib" results in a deprecation warning. See https://scala-cli.virtuslab.org/docs/release_notes#rename-using-lib-to-using-dep for more information.


djspiewak (approved):
Thank you!

PR #4001, opened 2024-02-16
Fix issue #3998

### LINKED ISSUE #3998: Cancelling `Async` queue `take` makes other `take` hang
This test fails:

```scala
    "foobar" in real {
      val mkQueue = Queue.unboundedForAsync[IO, String] // this fails
      // val mkQueue = Queue.unboundedForConcurrent[IO, String] // this is ok
      val test = mkQueue.flatMap { q =>
        val Item = "foobar"
        def taker(ref: Ref[IO, String]): IO[Unit] = IO.uncancelable { poll =>
          poll(q.take).flatMap(ref.set)
        }

        for {
          // 2 takers:
          ref1 <- IO.ref("")
          ref2 <- IO.ref("")
          fibers <- IO.both(taker(ref1).start, taker(ref2).start)
          (fib1, fib2) = fibers
          _ <- IO.sleep(0.1.second)
          // race cancelling one of them
          // and pushing one item;
          // whatever happens, exactly one
          // of them must receive that item:
          _ <- fib1.cancel.both(q.offer(Item)) // this fails
          // _ <- fib1.cancel.productR(q.offer(Item)) // this is ok
          item1 <- ref1.get
          _ <- if (item1.nonEmpty) {
            // fib1 must've took the item
            IO(item1 mustEqual Item) *> {
              // and fib2 is not yet finished
              ref2.get.flatMap { item2 =>
                IO(item2.isEmpty must beTrue)
              } *> {
                // make fib2 not hang by pushing another item:
                q.offer("x") *> fib2.joinWithNever *> ref2.get.flatMap { item2 =>
                  IO(item2 mustEqual "x")
                }
              }
            }
          } else {
            IO.ref(false).flatMap { timedOut =>
              // fib1 (it's already completed)
              // didn't take the item; fib2 must:
              fib2.joinWithNever.timeoutTo(
                4.seconds,
                timedOut.set(true) *> q.offer("y") *> fib2.joinWithNever // fallback to not make the test hang
              ) *> ref2.get.flatMap { item2 =>
                IO(item2 mustEqual Item) *> timedOut.get.flatMap { timedOut =>
                  // if we used the fallback, make the test fail:
                  IO(timedOut mustEqual false) // fails HERE
                }
              }
            }
          }

        } yield ()
      }

      test.parReplicateA_(10000).as(ok)
    }
```

Observations:
- Using `unboundedForConcurrent` (instead of `unboundedForAsync`) fixes it.
- Using `fib1.cancel.productR(q.offer(Item))` instead of `both` (i.e., no race between `cancel` and `offer`) fixes it.
- When it fails, it fails at the line I marked with `HERE`; I think this means, that the item wasn't lost (`fib2` received it), just the wakeup was lost, then the fallback offer waked it up.

durban:
Fix `take` cancelling for unbounded like it already works for bounded.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Queue.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Queue.scala
@@ -820,52 +820,58 @@ object Queue {
 
     val size: F[Int] = F.delay(buffer.size())
 
-    val take: F[A] = F defer {
-      try {
-        // attempt to take from the buffer. if it's empty, this will raise an exception
-        F.pure(buffer.take())
-      } catch {
-        case FailureSignal =>
-          // buffer was empty
-          // capture the fact that our retry succeeded and the value we were able to take
-          var received = false
-          var result: A = null.asInstanceOf[A]
+    val take: F[A] = F uncancelable { poll =>
+      F defer {
+        try {
+          // attempt to take from the buffer. if it's empty, this will raise an exception
+          F.pure(buffer.take())
+        } catch {
+          case FailureSignal =>
+            // buffer was empty
+            // capture the fact that our retry succeeded and the value we were able to take
+            var received = false
+            var result: A = null.asInstanceOf[A]
 
-          // a latch to block until some offerer wakes us up
-          val wait = F.async[Unit] { k =>
-            F delay {
-              // register ourselves as a listener for offers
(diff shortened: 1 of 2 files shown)


djspiewak (approved):
Nice catch! Thank you!

PR #4003, opened 2024-02-17
Fix issue #4002

### LINKED ISSUE #4002: `IO#asyncCheckAttempt` is inconsistent with `Async#asyncCheckAttempt`
`IO` has `case Left(None) => poll(get)`, so it's cancellable. I think the `Async` one is the correct one.

durban:
This PR makes `IO#asyncCheckAttempt` work the same as `Async#asyncCheckAttempt`. So both will be uncancelable when suspended with a `None` finalizer. Closes #4002.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -1206,7 +1206,7 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
           lift(k(resume)) flatMap {
             case Right(a) => G.pure(a)
             case Left(Some(fin)) => G.onCancel(poll(get), lift(fin))
-            case Left(None) => poll(get)
+            case Left(None) => get
           }
         }
       }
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1206,7 +1206,7 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits {
           lift(k(resume)) flatMap {
             case Right(a) => G.pure(a)
             case Left(Some(fin)) => G.onCancel(poll(get), lift(fin))
-            case Left(None) => poll(get)
+            case Left(None) => get
```
samspills, line 1209: Will this cause the same kind of semantic break as the `async_` change in 3.5.0?

durban: Yes.

As far as I can tell, `asyncCheckAttempt` should have been added like this in 3.5.0 (and for `Async`, it was). It's only due to an unlucky git merge that it wasn't. (And I guess due to the lack of a unit test.)

djspiewak: I suspect the impact of this is much lower than the `async_` change, since this is a far less widely used constructor. Also we override the `Async` implementation in `Async[IO]`, so this bug fix is actually entirely shadowed unless you're using monad transformers or a third-party effect. I think it's fair to consider this PR to be a bug fix rather than a breaking change.

**Edit:** Oh my shadowing argument is backwards. Rip. Either way, bug fix IMO.

djspiewak (approved):

Issue #3993, opened 2024-02-06
Allow overriding how fatal errors are printed


fredfp:
The handling of some uncaught fatal errors can be customized by overriding `cats.effect.IOApp.reportFailure`.

However, such customization is currently not possible in many cases (all usages of `java.lang.Throwable.printStackTrace()`, for example [here](https://github.com/typelevel/cats-effect/pull/3008)).

Would it be possible to handle similar situations like in the fs2 reactive-streams integration [here](https://github.com/typelevel/fs2/blob/08cb8d41738833d2a5fa327eeed071f2bc19daa4/reactive-streams/src/main/scala/fs2/interop/reactivestreams/StreamSubscriber.scala#L147) i.e., by using (if defined) java thread's default `UncaughtExceptionHandler`?

This would allow one to redirect uncaught errors to a custom logger, so that it gets properly ingested in a logging stack/system.

djspiewak:
I think there are only a couple areas (at the edge of the world, basically) where we forcibly print the stack trace. In most cases, we rely on `ExecutionContext#reportFailure`, which defaults to delegating to `IOApp#reportFailure`, which can be overridden to whatever you want. I think it would be totally reasonable to swap these last `printStackTrace()`s for `reportFailure` on the `IORuntime#compute` executor, which would link all of the machinery back together again.

fredfp:
Usual failure reporter (i.e., `ExecutionContext#reportFailure`) have the following signature: `Throwable => Unit`, whereas for `IOApp#reportFailure` it is `Throwable => IO[Unit]`. Did you have that in mind in your comment above?

I also wouldn't rely on (or require) an IOApp instance to be available and in scope.

Wouldn't it be better to do it the other way round? i.e., rely on an `UncaughtExceptionHandler`, and call it from everywhere (via `ExecutionContext#reportFailure`) and from `IOApp#reportFailure`'s default implementation?

djspiewak:
> Usual failure reporter (i.e., ExecutionContext#reportFailure) have the following signature: Throwable => Unit, whereas for IOApp#reportFailure it is Throwable => IO[Unit]. Did you have that in mind in your comment above?

[...]

PR #4010, opened 2024-02-19
Add setters for the failure and starvation reporters in IOApp

BalmungSan:
As promised in **Discord**.
c.c. @armanbilge @djspiewak 

-----

I will fix tests if we decide to move forward with this approach.
For now, I am more interested in gathering feedback about the proposal.

### DIFF
--- a/core/native/src/main/scala/cats/effect/IOApp.scala
+++ b/core/native/src/main/scala/cats/effect/IOApp.scala
@@ -146,6 +146,14 @@ trait IOApp {
 
   private[this] var _runtime: unsafe.IORuntime = null
 
+  @volatile
+  private[this] var _failureReporter: Throwable => IO[Unit] =
+    err => IO.consoleForIO.printStackTrace(err)
+
+  @volatile
+  private[this] var _onCpuStarvationReporter: CpuStarvationWarningMetrics => IO[Unit] =
+    metrics => onCpuStarvationWarn(metrics)
+
   /**
    * The runtime which will be used by `IOApp` to evaluate the [[IO]] produced by the `run`
    * method. This may be overridden by `IOApp` implementations which have extremely specialized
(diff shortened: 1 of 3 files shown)


core/native/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -167,12 +174,57 @@ trait IOApp {
    */
   protected def runtimeConfig: unsafe.IORuntimeConfig = unsafe.IORuntimeConfig()
 
+  // Currently unused!
+  private def reportFailure(err: Throwable): IO[Unit] =
+    IO(_failureReporter).flatMap(_.apply(err))
```
BalmungSan, line 179: `IORuntime.createEventLoop` does not receive a `reportFailure` argument.
Thus, I guess **native** either doesn't have this problem or the reporting functionality has not been implemented yet. I am assuming the later, thus I decide to leave this ready to be called once that functionality is implemented.

core/native/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -204,6 +253,12 @@ trait IOApp {
     val installed = if (runtime == null) {
       import unsafe.IORuntime
 
+      // Eventually this will be needed:
+      // reportFailure = t =>
+      //   IO(_failureReporter)
+      //     .flatMap(_.apply(t))
+      //     .unsafeRunAndForgetWithoutCallback()(runtime)
```
BalmungSan, line 260: `IORuntime.createEventLoop` does not receive a `reportFailure` argument.
Thus, I guess **native** either doesn't have this problem or the reporting functionality has not been implemented yet. I am assuming the latter, thus I decide to leave this ready to be called once that functionality is implemented.

fredfp:
As reporter of https://github.com/typelevel/cats-effect/issues/3993 I'm wondering if the starvation reporter is anything special here.

Errors are usually reported by printing to sdterr (e.g., `cats.effect.IOApp.onNonMainThreadDetected`), I was wondering if we should allow overriding the part that logs to the console/stderr (i.e., for redirection to a logger).

It is in that sense that I'm asking if starvation reporter is anything special, would it make sense to rather add (and allow overriding) something like `def reportError(error: String): IO[Unit]` and call this to report starvation, non main thread detected and other issues?

djspiewak:
I think I'm still pretty uncomfortable with this approach. I definitely agree the current API is clunky and forces users to deal with some annoying unsafety, but I'm not sure this type of blunt stateful approach is right either. Punting to 3.7 to give us more time to think about it.

PR #4025, opened 2024-03-04
Removes language:postfixOps from tutorial as it is not used.

froth:
I basically pasted the tutorial into a new sbt project and it compiles
fine without.

Also updated scala and sbt version.


froth:
ci-failure seems very unreladed, idk...

djspiewak (approved):
Yeah it's unrelated. We have some flakiness in the CI right now

PR #4026, opened 2024-03-04
Removes language:postfixOps from tutorial as it is not used. (3.x)

froth:
I basically pasted the tutorial into a new sbt project and it compiles
fine without.


djspiewak (approved):
Good point! I think we've been cargo culting this for a long time.

djspiewak:
Closing since #4025 will be copied over on release.

PR #4044, opened 2024-03-19
Use `IOApp#reportFailure` instead of `printStackTrace()` for blocking EC

### LINKED ISSUE #3993: Allow overriding how fatal errors are printed

fredfp:
fixes #3993

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
@@ -151,7 +151,14 @@ private[unsafe] abstract class IORuntimeCompanionPlatform { this: IORuntime.type
     createDefaultComputeThreadPool(self(), threads, threadPrefix)
 
   def createDefaultBlockingExecutionContext(
-      threadPrefix: String = "io-blocking"): (ExecutionContext, () => Unit) = {
+      threadPrefix: String = "io-blocking"
+  ): (ExecutionContext, () => Unit) =
+    createDefaultBlockingExecutionContext(threadPrefix, _.printStackTrace())
+
+  private[effect] def createDefaultBlockingExecutionContext(
+      threadPrefix: String,
+      reportFailure: Throwable => Unit
+  ): (ExecutionContext, () => Unit) = {
     val threadCount = new AtomicInteger(0)
     val executor = Executors.newCachedThreadPool { (r: Runnable) =>
       val t = new Thread(r)
@@ -159,7 +166,7 @@ private[unsafe] abstract class IORuntimeCompanionPlatform { this: IORuntime.type
       t.setDaemon(true)
       t
     }
-    (ExecutionContext.fromExecutor(executor), { () => executor.shutdown() })
+    (ExecutionContext.fromExecutor(executor, reportFailure), { () => executor.shutdown() })
   }
 
   def createDefaultScheduler(threadPrefix: String = "io-scheduler"): (Scheduler, () => Unit) = {
--- a/core/jvm/src/main/scala/cats/effect/IOApp.scala
+++ b/core/jvm/src/main/scala/cats/effect/IOApp.scala
@@ -388,7 +388,11 @@ trait IOApp {
           )
 
         val (blocking, blockDown) =
-          IORuntime.createDefaultBlockingExecutionContext()
+          IORuntime.createDefaultBlockingExecutionContext(
+            threadPrefix = "io-blocking",
+            reportFailure =
+              (t: Throwable) => reportFailure(t).unsafeRunAndForgetWithoutCallback()(runtime)
+          )
 
         IORuntime(
           compute,


core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
```diff
@@ -174,7 +174,7 @@ private[unsafe] abstract class IORuntimeCompanionPlatform { this: IORuntime.type
       threadPrefix,
       blockerThreadPrefix,
       60.seconds,
-      _.printStackTrace(),
+      self.compute.reportFailure,
```
armanbilge: I'm suspicious if this does the right thing. It might need to be written as `ex => self.compute.reportFailure(ex)`.

armanbilge (commented):
Thanks for the PR! We can probably target this at `series/3.5.x` so it can be released sooner than later :)

fredfp:
> Thanks for the PR! We can probably target this at `series/3.5.x` so it can be released sooner than later :)

That would be great, is there anything I could/should do about it?

armanbilge:
Yes, you can rebase your branch and retarget the PR on GitHub.

fredfp:
I'm not sure how to handle bincompat, I've pushed a 2nd commit with an attempt to address it.

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
```diff
@@ -150,16 +154,24 @@ private[unsafe] abstract class IORuntimeCompanionPlatform { this: IORuntime.type
...
   def createDefaultBlockingExecutionContext(
-      threadPrefix: String = "io-blocking"): (ExecutionContext, () => Unit) = {
+      threadPrefix: String
+  ): (ExecutionContext, () => Unit) =
+    createDefaultBlockingExecutionContext(threadPrefix, _.printStackTrace())
+
+  def createDefaultBlockingExecutionContext(
+      threadPrefix: String = "io-blocking",
+      reportFailure: Throwable => Unit = _.printStackTrace()
+  ): (ExecutionContext, () => Unit) = {
```
armanbilge: Oops, I'm sorry. I missed that you added a new public API. We can't add new APIs or deprecate existing APIs in the 3.5.x line. That would force this PR to go to 3.x.

However, I still think these changes are good for 3.5.x and we can accomplish them without changing the public API. To do so, we can make this new method `private[effect]` and keep the existing public method un-deprecated. Then we can do the publication/deprecation in 3.x. Thoughts?

fredfp: It's done for 3.5.x, should I ignore the residual mima failures as they are not related to public apis?

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -88,22 +87,21 @@ trait Dispatcher[F[_]] extends DispatcherPlatform[F] {
...
-    t.printStackTrace()
+  protected def reportFailure(t: Throwable): Unit
 
   // package-private because it's just an internal utility which supports specific implementations
   // anyone who needs this type of thing should use unsafeToFuture and then onComplete
   private[std] def unsafeRunAsync[A](fa: F[A])(cb: Either[Throwable, A] => Unit): Unit =
     unsafeToFuture(fa).onComplete(t => cb(t.toEither))(parasiticEC)
-}
-
-object Dispatcher {
```
armanbilge: You cannot make these changes compatibly. However, I don't think this particular change is important anyway. The `parasiticEC` is used in a very specific place to invoke a callback that we control i.e. a callback that should never fail. So I don't expect this `reportFailure` method to ever be invoked.

fredfp: Ok, removed for 3.5.x. I'll keep those on the side for 3.x.

armanbilge: Sorry, it's not compatible for 3.5.x or 3.x. Is there a situation where you think it is important?

fredfp: I agree, I found them looking for instances of `printStackTrace` and tried to remove them, but those can stay as no user code should ever get executed on the parasiticEC.

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
```diff
@@ -141,7 +141,11 @@ private[unsafe] abstract class IORuntimeCompanionPlatform { this: IORuntime.type
       threadPrefix: String = "io-compute",
       blockerThreadPrefix: String = DefaultBlockerPrefix)
       : (WorkStealingThreadPool, () => Unit) =
-    createWorkStealingComputeThreadPool(threads, threadPrefix, blockerThreadPrefix)
+    createWorkStealingComputeThreadPool(
+      threads,
+      threadPrefix,
+      blockerThreadPrefix,
+      reportFailure = t => self.compute.reportFailure(t))
```
armanbilge: Hang on, sorry. I'm just a little bit confused by this. Does this actually work?

Here's why I'm confused: this method is building the `WorkStealingComputeThreadPool` which will be installed as `compute` in `self.compute`. Thus we are implementing `reportFailure` of this `ExecutionContext` by delegating recursively to itself. Is that right or am I mixed up somehow?

fredfp: You are absolutely right, thanks for spotting this. I went to check pre-3.4 source code and `cats.effect.IOApp#reportFailure` didn't exist at the time, so I reverted to the previous behaviour which is the best we can do.

core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
```diff
@@ -151,15 +151,22 @@ private[unsafe] abstract class IORuntimeCompanionPlatform { this: IORuntime.type
     createDefaultComputeThreadPool(self(), threads, threadPrefix)
 
   def createDefaultBlockingExecutionContext(
-      threadPrefix: String = "io-blocking"): (ExecutionContext, () => Unit) = {
+      threadPrefix: String = "io-blocking"
+  ): (ExecutionContext, () => Unit) =
+    createDefaultBlockingExecutionContext(threadPrefix, _.printStackTrace())
+
+  private[effect] def createDefaultBlockingExecutionContext(
```
armanbilge, line 158: We should create a follow-up to make this public in 3.x.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -899,7 +899,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
       cb(Left(new CancellationException("The fiber was canceled"))),
       t => {
         if (!NonFatal(t)) {
-          t.printStackTrace()
+          runtime.compute.reportFailure(t)
```
armanbilge: Hmm, I'm look at this and suddenly I'm not so sure about this change. Here's the problem.

`!NonFatal(t)` "not non-fatal" i.e. it _is_ a fatal error. When there is a fatal error (e.g. `OutOfMemoryError`) the JVM is in an inconsistent state and we cannot assume that it is safe to allocate.

The `runtime.compute.reportFailure(t)` may involve allocating an `IO`/`Fiber` and therefore is not safe to do in general.

Meanwhile `t.printStackTrace()` is know to be always safe.

armanbilge (approved):

djspiewak (approved):

PR #4052, opened 2024-04-16
`Dispatcher`: avoid spinning

durban:
This is a small internal refactoring of the "new" `Dispatcher`.

Currently during cancellation there is a racy write/read of `action` (a plain `var`), and this is resolved by spinning on `stateR` (an `AtomicReference`). The idea of this PR is that since we already have an `AtomicReference`, which we're reading _anyway_, we can avoid the data race, and thus the spinning, by putting `action` inside the `AtomicReference`.

Besides this, there are 2 very small changes: removing an outdated comment, and an unnecessary type parameter.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
@@ -16,7 +16,7 @@
 
 package cats.effect.std
 
-import cats.{Applicative, MonadThrow}
+import cats.Applicative
 import cats.effect.kernel.{
   Async,
   Concurrent,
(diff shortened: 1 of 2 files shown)


std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -383,24 +382,22 @@ object Dispatcher {
...
-    final class Primary[F[_]](var action: F[Unit], val stateR: AtomicReference[RegState[F]])
-        extends Registration[F]
+    final class Primary[F[_]](val stateR: AtomicReference[RegState[F]]) extends Registration[F]
 
     final case class Finalizer[F[_]](action: F[Unit]) extends Registration[F]
 
     final case class PoisonPill[F[_]]() extends Registration[F]
   }
 
-  // the signal is just a skolem for the atomic references; we never actually run it
```
durban, line 403: This comment probably referred to an older version of the code(?).

std/shared/src/main/scala/cats/effect/std/Dispatcher.scala
```diff
@@ -383,24 +382,22 @@ object Dispatcher {
   private sealed abstract class RegState[+F[_]] extends Product with Serializable
 
   private object RegState {
-    case object Unstarted extends RegState[Nothing]
+    final case class Unstarted[F[_]](action: F[Unit]) extends RegState[F]
     final case class Running[F[_]](cancel: F[Unit]) extends RegState[F]
-    final case class CancelRequested[F[_]](latch: Promise[Unit]) extends RegState[F]
```
durban, line 388: Unrelated, but this type param seems unnecessary.

djspiewak (approved):
This is a really nice simplification

PR #4054, opened 2024-04-17
Document usage of `-Wnonunit-statement`

### LINKED ISSUE #3646: Add FAQ for `unused value of type cats.effect.IO[Unit] (add : Unit to discard silently)`
This is the warning raised by the `-Wnonunit-statement`, at least on Scala 2 (someone should check Scala 3).
- https://github.com/scala/scala/pull/9893
- https://github.com/lampepfl/dotty/pull/16936

We should:
1. document and recommend this very useful compiler option
2. get the error into our docs, so it is indexed and searchable for when folks hit it

froth:
fixes #3646

I will open another pull request against the g8 template adding the warning.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -243,7 +243,7 @@ For example the following code emits the warning: `unused value of type cats.eff
 ```scala
   def example(input: String): IO[String] = {
     IO.println(input)
-    IO(input)
+    IO.pure(input)
   }
 ```
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -247,7 +247,7 @@ For example the following code emits the warning: `unused value of type cats.eff
   }
 ```
 
-This warning tells you, that the `IO.println` statement is created and then discarded without being used. Consequently this code will not print anything. To fix this you have to connect both IO instances.
+This warning tells you, that the `IO.println` instance is created and then discarded without being used. Consequently this code will not print anything. To fix this you have to connect both IO instances.
 
 Working version of the above example (emitting no warnings):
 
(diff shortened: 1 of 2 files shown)


froth:
CI failure seems unrelated

docs/faq.md
```diff
@@ -233,3 +233,28 @@ It is still worth keeping in mind that this is only a partial solution. Whenever
...
+## What do non-unit statement warnings mean?
+
+We recommend to enable the compiler warning `-Wnonunit-statement`.  If you do so you get compiler warnings for some common bug patterns.
+
+For example the following code emits the warning: `unused value of type cats.effect.IO[Unit]`.
+
+```scala
+  def example(input: String): IO[String] = {
+    IO.println(input)
+    IO(input)
```
durban:
```suggestion
    IO.pure(input)
```

docs/faq.md
```diff
@@ -233,3 +233,28 @@ It is still worth keeping in mind that this is only a partial solution. Whenever
...
+  }
+```
+
+This warning tells you, that the `IO.println` statement is created and then discarded without being used. Consequently this code will not print anything. To fix this you have to connect both IO instances.
```
durban: "statement" -> "instance" or "object"

docs/faq.md
```diff
@@ -233,3 +233,28 @@ It is still worth keeping in mind that this is only a partial solution. Whenever
...
+
+Working version of the above example (emitting no warnings):
+
+```scala
+  def example(input: String): IO[String] = {
+    IO.println(input)
+      .flatMap(_ => IO(input))
```
durban, line 257: This is (mostly) just style, so feel free to ignore it, but I'd use `*>` here instead of `flatMap`.

docs/getting-started.md
```diff
@@ -15,6 +15,12 @@ Naturally, if you're using ScalaJS, you should replace the double `%%` with a tr
 addCompilerPlugin("com.olegpy" %% "better-monadic-for" % "0.3.1")
 ```
 
+It is remommended to enable warnings for non-unit statements(see [FAQ](./faq.md#what-do-non-unit-statement-warnings-mean)):
```
durban: "remommended" -> "recommended"

Also, a space is missing before the opening `(`.

durban (commented):
Thanks, I've left a few small comments.

docs/faq.md
froth: I was unsure here, a for comprehension would also be possible instead. I opted for flatMap because it is something a beginner would most likely be familiar with and looks similar to the "broken" version. But *> is nicer. I think I would leave it as if that's ok.

durban: Sure, as I've said, it's just style.

froth:
Thanks for the review, I pushed changes :)

durban (approved):
Thanks for your work.

froth:
@durban is there anything else to be done here from my side? Just to make sure I don't forget that I opened this ;)

durban:
@froth Not at all. We're just slow ;-)

In fact, this seems like a simple case, so I'll merge this PR this weekend, unless another maintainer objects (or merges it first).

Issue #3456, opened 2023-02-22
Introduce "inclusive" `race` as new default semantic


djspiewak:
Full discussion here: https://discord.com/channels/632277896739946517/839263556754472990/1078054700932419695

The short version is that `race`, in its current form, represents a resource leak. If there's a tie (i.e. `f.cancel *> f.join` does *not* produce `Outcome.Canceled()`), then it's ambiguous as to what result should be chosen. Right now, we technically bias to the `Left`, but this is just an implementation detail made manifest. Even worse, if a race is doubly-successful in this way, it's very possible that a resource was acquired which will now just disappear. In most cases, resources which are acquired and then later exposed to an error or self-cancelation will trigger finalizers, but often not in success cases, so this is particularly dangerous.

**In a very real sense, `race` should never have returned `Either`.** The more correct return type would have been `Ior`.

We can address this in one of two ways. We can either introduce a new variant, like `raceInclusive`, which returns `Ior` and then deprecate `race`. Alternatively, we can play tricks with `Dummy` implicits and break everyone's sources:

```scala
def race[A, B](fa: F[A], fb: F[B])(implicit D: Dummy): F[Ior[A, B]] = ???
private[effect] def race[A, B](fa: F[A], fb: F[B]): F[Either[A, B]] = ???
```

I'm very tempted to say this is better because it avoids significant problems and cleans up the API (in some sense), but this is definitely a request for comments.

This is all orthogonal to the fiber *isolation* semantics of `race`. This affects a few things, but the most significant being self-cancelation semantics: do we treat self-cancelation as "losing the race", or do we treat it as *cancelation* and embed that result into the parent fiber? The former semantic leads to issues like #3396, while the latter semantic opens up some interesting possibilities (such as a computation "giving up" and allowing the alternative to complete naturally).

`racePair` is capable of handling all of this generality, and so for any use-cases which are *not* extremely common, we should push people in that direction, despite the fact that it's very low-level and somewhat hard to work with correctly. We should focus the pretty parts of the API on the most common semantics, which is likely fiber *non*-isolation (i.e. `timeout` semantics).

Jasper-M:
> do we treat self-cancelation as "losing the race", or do we treat it as cancelation and embed that result into the parent fiber? The former semantic leads to issues like #3396, while the latter semantic opens up some interesting possibilities (such as a computation "giving up" and allowing the alternative to complete naturally).

Aren't both of those things related to the _former_ semantic?

[...]

Issue #3430, opened 2023-02-17
PureConc not running finalizers


durban:
I might be missing something completely obvious, but `PureConc` doesn't seem to run finalizers:

```scala
    "foo" in {
      val t: PureConc[Int, Int] = for {
        c <- F.ref(0)
        latch <- F.deferred[Unit]
        fib <- F.start((latch.complete(()) *> F.never[Unit]).onCancel(c.update(_ + 1)))
        _ <- latch.get
        _ <- fib.cancel
        v <- c.get
      } yield v

      pure.run(t) mustEqual Outcome.Succeeded(Some(1))
    }
```

If I put this in `PureConcSpec`, it fails with:
```
Succeeded(Some(0)) != Succeeded(Some(1))
```

djspiewak:
Well that absolutely looks like a bug…

durban:
Yeah, something is definitely wrong here; possibly more than one thing.

What I'm (almost) sure is wrong is `PureFiber#cancel`:
https://github.com/typelevel/cats-effect/blob/series/3.x/kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/pure.scala#L391. It just writes `Outcome.Canceled()` into an mvar. So if the fiber will not be scheduled any more (e.g., because it went into a `never`), who will run the finalizers?

There is also something strange going on with pushing/popping the finalizers, but I'm not sure what (if anything).

PR #4059, opened 2024-04-19
Make `timeout`/`timeoutTo` always return the outcome of the effect

biochimia:
`timeout*` methods are implemented in terms of a race between a desired effect and the timeout. In the case that both effects complete simultaneously, it could happen that the timeout would win the race, a `TimeoutException` be raised, and the outcome of the desired effect lost.

As is noted in #3456, this is a general problem with the `race*` methods, and can't be addressed in the general case without breaking the current interfaces.

This change is a more narrow take on the problem specifically focusing on the `timeout` and `timeoutTo` methods. As these methods inherently wait for both racing effects to complete, the implementation is changed to always take into account the outcome of the desired effect, only raising a `TimeoutException` if the timeout won the race *and* the desired effect was effectively canceled. Similarly, errors from the desired effect are preferentially propagated over the generic `TimeoutException`.

The `timeoutAndForget` methods are left unchanged, as they explicitly avoid waiting for the losing effect to finish.

This change allows for `timeout` and `timeoutTo` methods to be safely used on effects that acquire resources, such as `Semaphore.acquire`, ensuring that successful outcomes are always propagated back to the user.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -779,17 +779,27 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
     andWait(duration: Duration)
 
   /**
-   * Returns an IO that either completes with the result of the source within the specified time
-   * `duration` or otherwise raises a `TimeoutException`.
+   * Returns an IO that either completes with the result of the source or otherwise raises a
+   * `TimeoutException`.
    *
-   * The source is canceled in the event that it takes longer than the specified time duration
-   * to complete. Once the source has been successfully canceled (and has completed its
-   * finalizers), the `TimeoutException` will be raised. If the source is uncancelable, the
-   * resulting effect will wait for it to complete before raising the exception.
+   * The source is raced against the timeout `duration`, and its cancelation is triggered if the
+   * source doesn't complete within the specified time. The resulting effect will always wait
+   * for the source effect to complete (and to complete its finalizers), and will return the
+   * source's outcome over raising a `TimeoutException`.
+   *
+   * In case source and timeout complete simultaneously, the result of the source will be
+   * returned over raising a `TimeoutException`.
+   *
+   * If the source effect is uncancelable, a `TimeoutException` will never be raised.
    *
    * @param duration
-   *   is the time span for which we wait for the source to complete; in the event that the
-   *   specified time has passed without the source completing, a `TimeoutException` is raised
+   *   is the time span for which we wait for the source to complete before triggering its
+   *   cancelation; in the event that the specified time has passed without the source
+   *   completing, a `TimeoutException` is raised
+   *
+   * @see
+   *   [[timeoutAndForget]] for a variant which does not wait for cancelation of the source
+   *   effect to complete.
    */
   def timeout[A2 >: A](duration: Duration): IO[A2] =
     handleDuration(duration, this) { finiteDuration =>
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -819,9 +819,12 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
...
   def timeoutTo[A2 >: A](duration: Duration, fallback: IO[A2]): IO[A2] = {
     handleDuration[IO[A2]](duration, this) { finiteDuration =>
-      race(IO.sleep(finiteDuration)).flatMap {
-        case Right(_) => fallback
-        case Left(value) => IO.pure(value)
+      IO.uncancelable { poll =>
+        poll(racePair(IO.sleep(finiteDuration))) flatMap {
+          case Left((oc, f)) => f.cancel *> oc.embed(poll(IO.canceled) *> IO.never)
+          case Right((f, _)) =>
+            f.cancel *> f.join.flatMap { oc => oc.fold(fallback, IO.raiseError, identity) }
```
durban, line 842: Could the `oc.fold` here be `oc.embed` instead?

durban (commented):
Can we write a unittest for this? I realize that it's not easy due to the race, but maybe it's possible?

A general comment about the `F.canceled *> F.never` pattern (I realize that `race` already has it, so this is not strictly about this PR): is this a case of "it can never happen (unless CE has a bug)"? If so, would it be better to `raiseError` instead of `never`?

armanbilge:
> Can we write a unittest for this? I realize that it's not easy due to the race, but maybe it's possible?

I think something like `IO.sleep(2.seconds).uncancelable.timeout(1.second)` could work.

core/shared/src/main/scala/cats/effect/IO.scala
biochimia: `oc.embed` does not apply here because the return type of `IO.timeoutTo` is `IO[A2]`, and not `IO[A]`, and is thus incompatible with `Outcome.embed`.

Note that the signature of `timeoutTo` in `IO` differs from that in `GenTemporal`. The latter returns `F[A]`, keeping the same effect type.

biochimia (commented):
> Can we write a unittest for this? I realize that it's not easy due to the race, but maybe it's possible?

I added a couple of tests in `IOSpec`, making use of @armanbilge's suggestion.

I'm not sure if the tests should go elsewhere. I'm also not sure of where would be appropriate place to specifically test the `GenTemporal` implementation.

> A general comment about the F.canceled *> F.never pattern (I realize that race already has it, so this is not strictly about this PR): is this a case of "it can never happen (unless CE has a bug)"? If so, would it be better to raiseError instead of never?

On this one, I tried to follow existing practice. I don't have an informed opinion of which alternative is better. If this were to be changed, I suppose it should be done separately and cover the different places the pattern is used in, to keep the codebase consistent.

Other than unit tests. If there's alignment on moving these changes forward, I can also update the documentation to make explicit the semantics, namely with regards to the interaction between `timeout*` methods and `uncancelable` effects.

armanbilge:
> I'm not sure if the tests should go elsewhere.

You could add them here as well.

https://github.com/typelevel/cats-effect/blob/e8669bc3bcffffc15ce0a9d5156ad33c5d6ac16f/laws/shared/src/test/scala/cats/effect/laws/GenTemporalSpec.scala#L31

biochimia:
Are there any guidelines on what sort of tests should go under `laws/` and which should go under `tests/`?

For instance, I notice that the `IOSpec` tests include support for different "runners" (i.e., `real` vs `ticked`), while this is missing in the `GenTemporalSpec` that lives under `laws/`.

armanbilge:
`real` and `ticked` are the "real" and "test" runtimes for `IO`, specifically.

> while this is missing in the `GenTemporalSpec`

That's because these tests are not run in `IO`. These tests are run with `PureConc` "pure concurrent" transformed with `TimeT` which doesn't have a runtime and is designed primarily for pure testing, without side-effects.

> Are there any guidelines on what sort of tests should go under `laws/` and which should go under `tests/`?

If it is not testing something specific to `IO` or the runtime, then I prefer to test it with `PureConc`.

biochimia:
I've added more tests under `GenTemporalSpec`. I also tried to add tests for cancelation semantics, but I seem to get weird semantics in the tests:

- I tried using `onCancel` to observe the cancellation of a timed out action, but the timeout gets triggered without the finaliser being invoked.
- I tried using `F.canceled` so as to test the behavior with regards to self-cancellation, but that also seemed to get ignored in the tests, triggering a timeout anyway. This one, I also tried to add with the help of `.pendingUntilFixed`, but the cancelled action seems to propagate and the test ends up generating an error, anyway.

These seem to be artifacts of the test setup, as I didn't observe this behaviour in the test cases I originally had for the issue.

armanbilge:
> * I tried using `onCancel` to observe the cancellation of a timed out action, but the timeout gets triggered without the finaliser being invoked.

Erm, there is this outstanding issue 😳 

- https://github.com/typelevel/cats-effect/issues/3430

biochimia:
I've updated docs for `timeout` and `timeoutTo` methods, to reflect the updated semantics. Tests were previously added. Any feedback is welcome.

How do we move forward on this?

djspiewak (approved):

djspiewak:
Let's not block this on #3430. Ideally I would like to, but fixing that has been a multi-tiered yak ziggurat, so better to move forward on this one independently.

fommil:
A git bisect confirmed that this introduced the bug documented over in #4489, tl;dr Resource doesn't run its finalizer.

```scala
  val run: IO[Unit] =
    IO.ref(false).flatMap { ref =>
      val res = Resource.make(ref.set(true))(_ => ref.set(false))
      val timedRes = res.timeout(1.hour)
      val check = timedRes.use_ *>
        ref.get.ifM(IO.raiseError(new Exception("not released")), IO.unit)
      check.replicateA_(10000)
    }
```

PR #4065, opened 2024-04-24
Add `KeyedMutex`

BalmungSan:
Extention to the current `Mutex` class to support one look per `key`.

I extracted the `LockQueue` and `lock` implementations into their own file to allow reuse.

I added a basic suite of tests, mostly focused on multiple key interactions.
Since each key is, in essence, its own `Mutex` I don't think we need to replicate all the complex `Mutex` tests... since those are actually tests for the `LockQueue`.

### DIFF
--- a/docs/core/native-image.md
+++ b/docs/core/native-image.md
@@ -67,4 +67,4 @@ Benchmark 1: ./target/native-image/cats-effect-3-hello-world
 ```
 
 Another way to get your cats effect app compiled with native-image is to leverage
-the package command of scala-cli, like in the [example](../faq.md#Native-Image-Example)
+the package command of scala-cli, like in the [example](../faq.md#native-image-example).
--- a/build.sbt
+++ b/build.sbt
@@ -1123,7 +1123,13 @@ lazy val std = crossProject(JSPlatform, JVMPlatform, NativePlatform)
         ProblemFilters.exclude[FinalMethodProblem](
           "cats.effect.std.Dispatcher#RegState#Unstarted.toString"),
         ProblemFilters.exclude[DirectMissingMethodProblem](
-          "cats.effect.std.Dispatcher#Registration#Primary.*")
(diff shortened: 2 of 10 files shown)


iRevive:
> Should we use a MapRef instead of a Ref[Map]? Could that help reduce contention?

I have observed lower memory footprint while using `MapRef` instead of `Ref[F, Map[K, V]]` in some cases.

armanbilge:
> I have observed lower memory footprint while using `MapRef` instead of `Ref[F, Map[K, V]]` in some cases.

@iRevive One of the implementations of `MapRef` is literally a `Ref[Map]`, maybe your observations are related to a `ConcurrentHashMap`-backed `MapRef` ?

iRevive:
> > I have observed lower memory footprint while using `MapRef` instead of `Ref[F, Map[K, V]]` in some cases.
> 
> @iRevive One of the implementations of `MapRef` is literally a `Ref[Map]`, maybe your observations are related to a `ConcurrentHashMap`-backed `MapRef` ?

Yes, it was the `MapRaf` that uses `ConcurrentHashMap` under the hood.

docs/core/native-image.md
```diff
@@ -67,4 +67,4 @@ Benchmark 1: ./target/native-image/cats-effect-3-hello-world
 ```
 
 Another way to get your cats effect app compiled with native-image is to leverage
-the package command of scala-cli, like in the [example](../faq.md#Native-Image-Example)
+the package command of scala-cli, like in the [example](../faq.md#native-image-example).
```
BalmungSan, line 70: Unrelated change, but I noticed this warning while working on the docs.
If you prefe,r we may cherry pick this commit to a different PR.

djspiewak (approved):
You know this is probably going to be quite galling after sitting open for a year, but honestly I think this is good as-is! Thank you for being so patient

Issue #4055, opened 2024-04-17
unsafeToFuture fails to raise a second InterruptedException if the exception is handled and retried


swapdonkey:
The code below catches a ExecutionException which wraps the Interrupted exception on the first iteration. On the second iteration no ExecutionException is thrown and the main thread blocks on the Await. 
 ```scala
 while(true){
   val f = IO{
        throw new InterruptedException("Test") 
     }.attempt.unsafeToFuture()(IORuntime.global)

   try{
     Await.result(f, Duration.Inf)
   } catch {
      case t:ExecutionException => println(s"Underlyng exception ${t.getCause}")
    }
}
```

I think the issue is in ```IOFiber.onFatalFailure``` specifically the line 
```if (IORuntime.globalFatalFailureHandled.compareAndSet(false, true))``` 
On the first iteration ```globalFatalFailureHandled``` is initialised to false on the second iteration its now true so the code 
block is never executed and the exception isn't raised. 

If that is the case would adding a function like ```resetFatalFailureHandled```  be added to IORuntime to reset it if the future returned from ```unsafeToFuture``` has its ExecutionException handled and retried in a loop?

durban:
It seems there are multiple things going on here. The essential thing is that when a _fatal_ exception is detected (which `InterruptedException` is), we're shutting down the whole world (meaning _all_ the `IORuntime`s in the current JVM).

There are incorrect things I see in the behavior of CE with the code above:

- The `IORuntime` doesn't seem to shut down cleanly.
- There is a _second_ `IORuntime` created, which also doesn't shut down cleanly.

These should be fixed, but fixing these won't make the code example "work", at least I don't think so: the idea is, that if there is a fatal exception, the most we can hope for is a clean shutdown.

What would you consider "correct behavior" for this code example?

djspiewak:
So the tricky thing here is that `InterruptedException`, like all fatal errors, fully torpedoes the runtime. By the time you get to the second iteration, the threads are all shut down and gone, so it's impossible to use `global` to execute anything. The design expectation is that you're not going to *try* to execute anything after this point, and you'll basically just shut down the VM.

Of course, that's not what you want *here* since you're trying to run new things after we torpedo the runtime. The answer is probably for you to not use `global` and instead make your own `IORuntime` (probably using the builder). Whenever you catch and recover from a fatal error, build a new runtime for yourself and discard the old one.

armanbilge:
[...]

PR #4067, opened 2024-04-25
Fix #4066: shut down executors when `IORuntime.global` shuts down

### LINKED ISSUE #4066: When `IORuntime.global` shuts down, it doesn't shut down the WSTP
... or other `Executor`s. I've noticed this when debugging  #4055. PR incoming...

durban:
Fixes #4066. Related to #4055.

### DIFF
--- a/tests/jvm/src/test/scala/cats/effect/IOAppSpec.scala
+++ b/tests/jvm/src/test/scala/cats/effect/IOAppSpec.scala
@@ -285,13 +285,6 @@ class IOAppSpec extends Specification {
         h.stderr() must not(contain("Promise already completed"))
       }
 
-      "shut down WSTP on fatal error without IOApp" in {
-        val h = platform(FatalErrorShutsDownRt, List.empty)
-        h.awaitStatus()
-        h.stdout() must not(contain("sadness"))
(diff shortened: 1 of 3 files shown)


armanbilge (approved):
thanks!

tests/jvm/src/test/scala/cats/effect/IOAppSpec.scala
```diff
@@ -335,6 +335,13 @@ class IOAppSpec extends Specification {
           ok
         }
 
+        "shut down WSTP on fatal error without IOApp" in {
+          val h = platform(FatalErrorShutsDownRt, List.empty)
+          h.awaitStatus()
+          h.stdout() must not(contain("sadness"))
```
djspiewak, line 341: lol

djspiewak (approved):
Oh I forgot that we completely suppressed the shutdown. So we kinda did this to ourselves. Nice catch!

tests/jvm/src/test/scala/cats/effect/IOAppSpec.scala
durban: That wasn't me, I've copied it from another test :smile:

PR #4090, opened 2024-06-24
Remove `IORuntime` from `allRuntimes` on shutdown

RafalSierkiewicz:
The PR originated from question on discord https://discord.com/channels/632277896739946517/632278585700384799/1254741830042652772 

Whenever the IORuntime is created through `apply` method it is being added to the `allRuntimes` collection.
Unfortunately it seems that is not removed anywhere. 

This PR tries to remove the created `IORuntime` from `allRuntimes` collection. 
I tried to keep the api as close as it was, but as the `allRuntimes` collection require `runtime` and the `hashCode` to be supplied to remove function I struggled a bit. 

I am open for comments and suggestions. Let me know which branch I should target, not sure if change is backward and forward compatible

### DIFF
--- a/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
+++ b/core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
@@ -18,6 +18,7 @@ package cats.effect
 package unsafe
 
 import scala.concurrent.ExecutionContext
+import scala.util.chaining._
 
 import java.util.concurrent.atomic.AtomicBoolean
 
@@ -42,7 +43,7 @@ final class IORuntime private[unsafe] (
     val scheduler: Scheduler,
     private[effect] val pollers: List[Any],
     private[effect] val fiberMonitor: FiberMonitor,
-    val shutdown: () => Unit,
+    defaultShutdown: () => Unit,
     val config: IORuntimeConfig
 ) {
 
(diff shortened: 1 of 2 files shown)


djspiewak (changes requested):
I think it would make more sense to do this in the same local scope where we *add* to `allRuntimes`, both because it avoids this parameter juggle in the main class, but also because it keeps the logic a bit more localized. The forward references should be resolvable with some type ascription and strategic use of the `lazy` modifier.

RafalSierkiewicz:
> I think it would make more sense to do this in the same local scope where we _add_ to `allRuntimes`, both because it avoids this parameter juggle in the main class, but also because it keeps the logic a bit more localized. The forward references should be resolvable with some type ascription and strategic use of the `lazy` modifier.

Yes, I totally I agree. I was trying to achieve it, but somehow realized that might be hard 🙈 I changed it to use `lazy` as well as `def` and added test

core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
```diff
@@ -70,12 +70,13 @@ object IORuntime extends IORuntimeCompanionPlatform {
       config: IORuntimeConfig): IORuntime = {
     val fiberMonitor = FiberMonitor(compute)
     val unregister = registerFiberMonitorMBean(fiberMonitor)
-    val unregisterAndShutdown = () => {
+    def unregisterAndShutdown: () => Unit = () => {
```
djspiewak, line 73: Moderate pedantry:

```suggestion
    def unregisterAndShutdown(): Unit = {
```

djspiewak (changes requested):

core/shared/src/main/scala/cats/effect/unsafe/IORuntime.scala
RafalSierkiewicz: I had to revert the change because compilation is failing in such a solution 
`An unapplied 0-arity method was eta-expanded (due to the expected type () => Unit), rather than applied to `()`.
[error] Write unregisterAndShutdown() to invoke method unregisterAndShutdown, or change the expected type.`

djspiewak (approved):

PR #4092, opened 2024-06-26
Add timed variant for IO[Unit]

geirolz:
Hello folks! 
I'm not sure about the solution but I encountered this "problem" enough times to submit this PR.
Basically when you want to use `timed` on an `IO[Unit]` you have to map the result to just keep the left side of the tuple. 
It would be nice to have a method to this directly within `IO`. 

My proposal here introduces breaking changes, other more conservative solutions would be to use a dedicated name for this case such as:
- `timed_`
- `timedIgnore` (with or without `and` )
- `timedL` or `timedR` it depends on which side you see it 
- `timedOnly`

Let me know what do you think! 
Thank :)

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -853,6 +853,9 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   private[effect] def timeoutAndForget(duration: FiniteDuration): IO[A] =
     timeoutAndForget(duration: Duration)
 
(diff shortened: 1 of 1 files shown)


ybasket:
There's functor syntax in `cats.syntax.all` that lets you write `myUnitIO.timed._1F` to achieve the same, would that work for you? I'm no maintainer and this is no direct opinion on the PR, just wondered whether you are aware as it's syntactically very close to what you suggest.

Scastie example: https://scastie.scala-lang.org/5LQicwTuSf2e2Fkrds2qvw

PR #4100, opened 2024-07-08
Fix typos

bwignall:
Fix some small typos I found. These should be non-semantic.


djspiewak (approved):
Thank you for the catches! Rip my spelling

Issue #3566, opened 2023-04-26
`executeEmbed` should auto-cede between every step by default


armanbilge:
Although the test runtime and `executeEmbed` cannot be used to exercise race conditions, it is still very useful for being able to exercise non-deterministic scheduling of fibers. In fact we can get even closer to simulating parallel execution by randomly interleaving the individuals _steps_ of concurrently running fibers, by configuring the test runtime to auto-cede between every step.

Arguably there is no harm to auto-cede between every step. Because fiber scheduling is entirely random, the fiber that auto-ceded may end up being immediately rescheduled _anyway_ (as if it never ceded at all). So we are not ruling out any possible scenarios, we are only expanding them to include more permutations of the various interleavings between concurrently running fibers.

There are certain bugs which cannot be tested without this, for example:
https://github.com/typelevel/cats-effect/blob/8d8097815594be405af2fcfddb4c9b599a1d04d4/tests/jvm/src/test/scala/cats/effect/kernel/AsyncPlatformSpec.scala#L46-L69

PR #3425, opened 2023-02-16
Fix `fromCompletableFuture` cancelation leak

armanbilge:
I inadvertently took over @TimWSpence's PR https://github.com/typelevel/cats-effect/pull/3417 😅

PR #4105, opened 2024-07-18
Add a `memoizedAcquire` method to `Resource`

satabin:
As mentioned in https://github.com/typelevel/cats-effect/pull/3890#issuecomment-2227063576, this is the PR taking over #3890 to add a specialized version of `memoized` for `Resource`.

Ref #3513


kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -751,6 +751,24 @@ sealed abstract class Resource[F[_], +A] extends Serializable {
...
+   * upon `use()`. If the `F[A]` value is executed multiple times, acquire happens once only and
+   * the acquired resource is shared to all callers. The resource is released as normal at the
+   * end of `use` (whether normal termination, error, or cancelled), if it was acquired.
+   */
+  def memoizedAcquire[B >: A](implicit F: Concurrent[F]): Resource[F, F[B]] = {
+    Resource.eval(F.ref(List.empty[Resource.ExitCase => F[Unit]])).flatMap { release =>
+      val fa2 = F.uncancelable { poll =>
+        poll(allocatedCase).flatMap { case (a, r) => release.update(r :: _).as(a) }
+      }
+      Resource.makeCaseFull[F, F[B]](poll => poll(F.memoize(fa2)).map(_.widen)) { (_, exit) =>
```
satabin, line 789: @durban @armanbilge I applied the change here, to ensure acquisition will not be cancelled once `F.memoize(fa2)` has been successfully run. I tried to come up with a test to ensure the behavior but could not find any reliable way to write something that tries to cancel it at the right time.

Do you have an idea on how I could write such a test?

armanbilge: > Do you have an idea on how I could write such a test?

https://github.com/typelevel/cats-effect/pull/3425 is an example of a similar PR. Notice the special use of the test runtime.
```scala
TestControl.executeEmbed(go, IORuntimeConfig(1, 2))
```

Further reading:
- https://github.com/typelevel/cats-effect/issues/3566

satabin: Thanks, I will have a look at this.

satabin:
I did not forget about this PR. Testing it is currently blocked by the `PureConc` finalizer problem (see #3430). As soon as there is a fix, I will design a test using it.

djspiewak:
Sadly I think this one might be *really* blocked on the `PureConc` stuff. It's hard to be totally comfortable with this absent tests.

armanbilge:
To refresh my memory: the reason we couldn't write the test in `IO` (using the test runtime) is because I think that `IO` actually doesn't reproduce this bug thanks to the optimization introduced in https://github.com/typelevel/cats-effect/pull/1742. Basically, the `map` is able to execute immediately without requiring an additional iteration of the runloop and thus skips the cancelation check.

armanbilge:
> It's hard to be totally comfortable with this absent tests.

Also point of clarification, there _are_ several tests for `memoizedAcquire` (which is just a refactoring of `memoize`). During refactoring I introduced [a bug, which Daniel U caught in review](https://github.com/typelevel/cats-effect/pull/3890#discussion_r1626638893), and this test is intended to catch that.

satabin:
> > It's hard to be totally comfortable with this absent tests.
> 
> Also point of clarification, there _are_ several tests for `memoizedAcquire` (which is just a refactoring of `memoize`). During refactoring I introduced [a bug, which Daniel U caught in review](https://github.com/typelevel/cats-effect/pull/3890#discussion_r1626638893), and this test is intended to catch that.

Yes, the currently committed test does not reflect this, it is still an attempt I wrote before noticing I couldn't reproduce it with `IO`. I will update the test with what I think should work, once the `PureConc` behavior is fixed.

djspiewak (dismissed):
Alright, let's go for it.

djspiewak:
Needs `prePR`

satabin:
> Needs `prePR`

Yes, I should have time to work on this this weekend.
Regarding the test I wanted to add to check that acquisition cancellation does not occur after the resource has been actually acquired and before it is put in the `release` list. Should I just not add it, or add and ignore it, until `PureConc` is fixed? What do you think is the best approach?

djspiewak:
> Regarding the test I wanted to add to check that acquisition cancellation does not occur after the resource has been actually acquired and before it is put in the release list. Should I just not add it, or add and ignore it, until PureConc is fixed? What do you think is the best approach?

Ideally add it and then mark `pendingUntilFixed`.

tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
```diff
@@ -1170,6 +1170,31 @@ class ResourceSpec extends BaseSpec with ScalaCheck with Discipline {
         }.void must completeAs(())
       }
     }
+
+    // TODO enable once `PureConc` finalizer bug is fixed. Test is indefinitely waiting for finalizers as of now
+    /*
```
armanbilge: Instead of commenting the test you can use `pendingUntilFixed` :)

https://github.com/typelevel/cats-effect/blob/28f8f294070412fe829f41c79e864e49a03a6299/tests/js/src/test/scala/cats/effect/IOPlatformSpecification.scala#L31-L34

satabin: True, I made a mistake in my test design. It should be ok now. See cd6103377

satabin: @armanbilge I took your comment into account and added a pending test. Do you think this PR is ready to be merged now?

djspiewak (approved):

PR #4106, opened 2024-07-19
Add `IORuntimeMetrics` and `CpuStarvationMetrics`

iRevive:
The first step towards cross-platform `IORuntimeMetrics`. This PR also introduces a cross-platform `CpuStarvationMetrics`. 

### Motivation

It would be nice to have an option to access runtime metrics without relying on MBeans. For example, you can use [Scala OpenTelemetry SDK](https://typelevel.org/otel4s/sdk/overview.html) to collect starvation metrics on all platforms (JVM, JS, Native).

### Usage

```scala
IO.delay(runtime.metrics.cpuStarvation.maxDrift())
  .flatMap(drift => IO.println(s"current drift: $drift"))
  .delayBy(2.seconds)
  .foreverM
```

### Next steps

Add WSTP compute and local queue metrics. The draft - https://github.com/typelevel/cats-effect/pull/3317.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/metrics/CpuStarvationMetrics.scala
+++ b/core/shared/src/main/scala/cats/effect/metrics/CpuStarvationMetrics.scala
@@ -16,12 +16,47 @@
 
 package cats.effect.metrics
 
-import cats.effect.IO
+import scala.concurrent.duration._
 
-import scala.concurrent.duration.FiniteDuration
+sealed trait CpuStarvationMetrics {
 
-private[effect] trait CpuStarvationMetrics {
-  def incCpuStarvationCount: IO[Unit]
+  /**
+   * Returns the current number of times CPU starvation has occurred.
+   *
+   * @note
+   *   the value may differ between invocations
+   */
+  def starvationCount(): Long
 
-  def recordClockDrift(drift: FiniteDuration): IO[Unit]
+  /**
+   * Returns the current (last) observed clock drift.
+   *
+   * @note
+   *   the value may differ between invocations
+   */
+  def clockDriftCurrent(): FiniteDuration
+
+  /**
+   * Returns the maximum clock drift observed since the launch.
+   *
+   * @note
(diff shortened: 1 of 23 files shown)


core/shared/src/main/scala/cats/effect/metrics/CpuStarvationMetrics.scala
```diff
@@ -16,12 +16,47 @@
...
 
-private[effect] trait CpuStarvationMetrics {
-  def incCpuStarvationCount: IO[Unit]
+  /**
+   * Returns the current number of times CPU starvation has occurred.
+   *
+   * @note
+   *   the value may differ between invocations
+   */
+  def starvationCount(): Long
```
iRevive: It's a good question whether the return type should be `IO[Long]`. While `IO` indicates potential side effects, it would be unhandy in case of integration with Java-based metrics collectors (prometheus, micrometer, etc).

core/shared/src/main/scala/cats/effect/metrics/CpuStarvationMetrics.scala
```diff
@@ -16,12 +16,47 @@
...
+
+  /**
+   * Returns the current (last) observed clock drift.
+   *
+   * @note
+   *   the value may differ between invocations
+   */
+  def clockDriftCurrent(): FiniteDuration
```
iRevive: `Long` could be more suitable and more memory efficient.

djspiewak:
What do you think about moving this stuff to `cats.effect.unsafe.metrics`? That feels like a more appropriate place, since we really don't want to wrap the impure interfaces in `IO`, but at the same time it's absolutely not a "safe" set of APIs in the sense we normally use the word.

iRevive:
Good point. I moved classes to the `cats.effect.unsafe.metrics` package.

iRevive:
Closing this one in favor of https://github.com/typelevel/cats-effect/pull/3317.

PR #4111, opened 2024-08-01
fixes and attempts to improve scaladoc for Outcome

satorg:
The specified type in the `Outcome` docs is incorrect – it cannot be just `OutcomeIO` out there.
The PR attempts to fix and further clarify the docs,
but any suggestions on how to make it clearer are very welcome.


djspiewak (approved):
Excellent!

PR #4112, opened 2024-08-03
Add `Stack`

BalmungSan:
- [x] Interface
- [x] Docs
- [x] Tests
- [x] Implementation

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Stack.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Stack.scala
@@ -18,7 +18,8 @@ package cats
 package effect
 package std
 
-import cats.effect.kernel.{Async, GenConcurrent, Sync}
+import cats.effect.kernel._
+import cats.syntax.all._
 
 /**
  * A purely functional, concurrent data structure which allows insertion and retrieval of
@@ -38,19 +39,19 @@ abstract class Stack[F[_], A] { self =>
   /**
    * Pushes the given element to the top of the `Stack`.
    *
-   * @param a
+   * @param element
    *   the element to push at the top of the `Stack`.
    */
-  def push(a: A): F[Unit]
+  def push(element: A): F[Unit]
 
   /**
    * Pushes the given elements to the top of the `Stack`, the last element will be the final
    * top.
    *
-   * @param as
+   * @param elements
    *   the elements to push at the top of the `Stack`.
    */
-  def pushN(as: A*): F[Unit]
+  def pushN(elements: A*): F[Unit]
 
   /**
    * Takes the top element of `Stack`, if there is none it will semantically block until one is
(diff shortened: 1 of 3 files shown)


reardonj:
Queue, PQueue and Dequeue are split into Source/Sink interfaces to allow for Functor and Contravariant interfaces. This should probably have the same treatment?

BalmungSan:
> Queue, PQueue and Dequeue are split into Source/Sink interfaces to allow for Functor and Contravariant interfaces. This should probably have the same treatment?

I thought about doing something like that, but TBH, not sure if it is really worth the effort.
IIRC, there is a general idea that the split of `Queue` didn't help that much, but I could be wrong.

Personally, I think that if you want to hide part of the interface and add extra functionality like `map`, that is better done in userland with a custom capability that encapsulates the `Queue / Stack`.

But, would love to hear what others think.

tests/shared/src/test/scala/cats/effect/std/StackSpec.scala
```diff
@@ -0,0 +1,234 @@
...
+        (
+          s.pop.parReplicateA(numbers.size).map(_.sorted),
+          numbers.parTraverse_(s.push)
+        ).parTupled
+      }
+
+      p must completeAs((numbers, ()))
+    }
+
+    "not lost elements when concurrently canceling a pop with a push" in real {
```
BalmungSan, line 209: I was hopping we could ensure this. But, I think it is actually impossible.

While the `push` and `pop` processes are uncancellable minos waiting on the `Deferred`, there is a race condition between completing the `df` and its `get` being cancelled. Thus, we can lose an element there.

std/shared/src/main/scala/cats/effect/std/Stack.scala
```diff
@@ -0,0 +1,291 @@
...
+                case true =>
+                  // If we managed to invalidate our waiter,
+                  // we try to remove it from the waiters queue.
+                  state.update(_.removeWaiter(waiter)).void
+
+                case false =>
+                  // But, if we didn't managed to invalidate it.
+                  // Then, that means we managed to receive a pushed element.
+                  // Thus, we have to push it again to avoid it getting lost.
+                  waiter.get.flatMap(element => this.push(element))
```
reardonj, line 274: Couldn't this let elements slip under the top the stack if a push happens after the pop then before the push back? ie. stack of A,B,C:

1. pop A but cancellation happens
2. before push is called, another thread pushes D
3. finalizer pushes A back

Now you'd have A,D,B,C, which should be impossible.

BalmungSan: Yes, concurrent `pop` cancellation and `push` can change the order of elements in the `Stack`, I don't think there is much we can do here tho, we would need to store more information and the implementation would be very complex.

Also, note that for it to happen you actually need to wait, that can't happen on a non-empty `Stack`, only in an empty one.

So the real situation is something more akin to:
`Stack` is empty.
One fiber is waiting on `pop`
Another fiber pushes `A, B, C`, concurrently with `pop` is cancelled, and to another `push(D)`
The end result may be something like: `C, D, B, A`
Which, yes, feels wrong, but in a highly concurrent scenario, I'm not sure what else to do, at least the element didn't got lost and the overall priority is mostly preserved.

reardonj: Is your scenario possible with pushN(A,B,C), or just a series of push()?

BalmungSan: It is possible with `pushN` sadly, for series for `pushs` I don't mind since if they are all concurrent then the right order is undefined anyway AFAIK.
But for `pushN` I am indeed sad that it can lead to that error, the thing is that `pushN` is atomic in deciding which waiters to await, but it can't wait for them to be awakened before modifying the `State`.

PR #4113, opened 2024-08-06
Add best-effort stealing API to `PollingSystem`

### LINKED ISSUE #4008: Add (best-effort) stealing API to polling system
In https://github.com/typelevel/cats-effect/discussions/3873#discussioncomment-8356851 we decided that we will support stealing for load-balancing. Thus, we should add an API to the polling system that allows completed I/O to be stolen. Not all polling systems will be able to implement this, but hopefully some of them can, such as io_uring, see https://github.com/typelevel/cats-effect/discussions/3873#discussioncomment-7608898.

armanbilge:
Closes https://github.com/typelevel/cats-effect/issues/4008.

I sketched out a possible stealing implementation for `Selector` along with a test but ended up reverting it because it's not possible to threadsafely interact with the selected keys set (without relying on synchronization, but this could mean the stealing thread being blocked indefinitely if the owner thread happens to be polling concurrently).

Since we don't support multi-threading on Native yet, stealing for epoll/kqueue is implemented as no-op for now. I will investigate if we can implement stealing for those systems later on.

I'm still optimistic that we can support stealing for io_uring.


core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
```diff
@@ -91,6 +91,23 @@ abstract class PollingSystem {
...
+   *   the thread-local [[Poller]] used to poll events.
+   *
+   * @param reportFailure
+   *   callback that handles any failures that occur during stealing.
+   *
+   * @return
+   *   whether any events were stolen. e.g. if the method returned due to timeout, this should
+   *   be `false`.
+   */
+  def steal(poller: Poller, reportFailure: Throwable => Unit): Boolean
```
djspiewak, line 109: Something that's a bit unclear in this API is whether `poller` is what you're stealing *from* or *to*.

djspiewak (commented):
One thing I'm a *bit* concerned about with this approach is, without a concrete implementation and usage, it's hard to be 100% certain that this is actually the API we want. I would almost feel like we're better off deferring this change until we have the ability to actually use it. Since it's already an abstract class (rather than a trait), and it has a default implementation, we can do things like this compatibly after 3.6.

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
armanbilge: Noted, although you always need somewhere to steal _from_. Also similar to timers we only steal completed events, so they don't need a destination heap / poller to be stolen to, this would be pretty challenging to implement anyway.

djspiewak:
So it occurs to me that, because this method is abstract, it's going to be binary breaking if we do it post 3.6 unless we provide a default implementation. The fact that it's an abstract class rather than a trait doesn't really save us.

armanbilge:
Right, but because the API only promises best-effort stealing the default implementation can just be a no-op.

PR #4118, opened 2024-08-16
Fix `cancelable` to `guarantee` cancelation

armanbilge:
I think this was an oversight. A non-happy-path finalizer could cause cancelation and backpressure to be circumvented.


durban (approved):
Nice, this definitely seems like it was an oversight.

Issue #3757, opened 2023-07-18
`Resource#attempt` holds onto acquired resources even in case of error


armanbilge:
```scala
//> using dep org.typelevel::cats-effect::3.5.1

import cats.effect.*
import cats.syntax.all.*

object Bug extends IOApp.Simple:
  def run = IO.ref(0).flatMap { ref =>
    val resource = Resource.make(ref.update(_ + 1))(_ => ref.update(_ + 1))
    val error = Resource.raiseError[IO, Unit, Throwable](new Exception)

    val attempt = (resource *> error).attempt

    attempt.use { r =>
      ref.get.flatMap { i =>
        IO {
          // acquiring the resource as a whole failed, so it should be released in entirety
          assert(r.isLeft, r.toString)
          assert(i == 2, i.toString) // assertion failed: 1
        }
      }
    }
  }
```

durban:
Hm, if it failed, why is the body of `use` called at all? But, I don't think it failed: it succeeded with `Left(new Exception)`.

armanbilge:
Yes you are exactly right. However although the resource acquisition failed (prior to the attempt) it has not run all the finalizers yet.

durban:
So, you're basically saying, that since `resource *> error` failed, it should run the finalizer of `resource` even before continuing to the `.attempt`, right? I guess that makes sense, (even if it seems a little strange to me)...

armanbilge:
> even if it seems a little strange to me

why does that feel strange? I'm drawing a distinction between `(resource *> error).attempt` and `resource *> error.attempt`. Currently they behave the same.

durban:
Yes, exactly: it seems strange to me that those two would be different. (To be clear, we're obviously assuming `resource` does not fail here.) Based on [this comment](https://github.com/typelevel/cats/pull/3241#issuecomment-571398273) by @djspiewak, it seems I might not be alone. At least I think that law would imply that they should be the same...

armanbilge:
> To be clear, we're obviously assuming `resource` does not fail here

👍 

> At least I think that law would imply that they should be the same...

Hmm, yes, I think I interpret that law the same way. Thanks for the pointer, let me think about this a bit 🤔

armanbilge:
Since I haven't mentioned it yet, the problem with the current semantics is that if you try to implement some sort of retry then failed resources don't get released in a timely fashion.

For example, here's a rough sketch of a connection pool as used in Ember client or Skunk.

```scala
[...]

PR #4121, opened 2024-08-20
Make `IO#onError` consistent with `ApplicativeError`

### LINKED ISSUE #4058: IO.onError is inconsistent with ApplicativeError
The `onError` that you get from `ApplicativeError` has this signature:

```
def onError[A](fa: F[A])(pf: PartialFunction[E, F[Unit]]): F[A]
```

While the "native" one in `IO` doesn't take a `PartialFunction`:

```
def onError(f: Throwable => IO[Unit]): IO[A]
```

That could lead to some surprises if you would rewrite a piece of code from tagless final to `IO`, or if you are just used to using the `onError` extension method in other code.

lenguyenthanh:
This also deprecates the old `IO.onError` function and replace it's usage with the new one.

Close #4058

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -585,7 +585,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
 
   @deprecated("Use onError with PartialFunction argument", "3.6.0")
   def onError(f: Throwable => IO[Unit]): IO[A] = {
-    val pf: PartialFunction[Throwable, IO[Unit]] = { case t => f(t) }
+    val pf: PartialFunction[Throwable, IO[Unit]] = { case t => f(t).reportError }
     onError(pf)
   }
 
@@ -596,8 +596,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
    * Implements `ApplicativeError.onError`.
    */
   def onError(pf: PartialFunction[Throwable, IO[Unit]]): IO[A] =
-    handleErrorWith(t =>
-      pf.applyOrElse(t, (_: Throwable) => IO.unit).reportError *> IO.raiseError(t))
+    handleErrorWith(t => pf.applyOrElse(t, (_: Throwable) => IO.unit) *> IO.raiseError(t))
 
   /**
    * Like `Parallel.parProductL`
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -1995,6 +1995,9 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits with TuplePara
     override def handleError[A](fa: IO[A])(f: Throwable => A): IO[A] =
       fa.handleError(f)
 
+    override def onError[A](fa: IO[A])(pf: PartialFunction[Throwable, IO[Unit]]): IO[A] =
+      fa.onError(pf)
+
     override def timeout[A](fa: IO[A], duration: FiniteDuration)(
         implicit ev: TimeoutException <:< Throwable): IO[A] = {
       fa.timeout(duration)
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -590,8 +590,10 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
     IO.OnCancel(this, fin)
 
   @deprecated("Use onError with PartialFunction argument", "3.6.0")
-  def onError(f: Throwable => IO[Unit]): IO[A] =
-    handleErrorWith(t => f(t).voidError *> IO.raiseError(t))
+  def onError(f: Throwable => IO[Unit]): IO[A] = {
+    val pf: PartialFunction[Throwable, IO[Unit]] = { case t => f(t) }
+    onError(pf)
+  }
 
   /**
    * Execute a callback on certain errors, then rethrow them. Any non matching error is rethrown
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -602,7 +602,8 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
    * Implements `ApplicativeError.onError`.
    */
   def onError(pf: PartialFunction[Throwable, IO[Unit]]): IO[A] =
-    handleErrorWith(t => pf.applyOrElse(t, (_: Throwable) => IO.unit) *> IO.raiseError(t))
+    handleErrorWith(t =>
+      pf.applyOrElse(t, (_: Throwable) => IO.unit).reportError *> IO.raiseError(t))
 
   /**
    * Like `Parallel.parProductL`
@@ -941,6 +942,19 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   def void: IO[Unit] =
     map(_ => ())
 
+  /**
+   * Similar to [[IO.voidError]], but also reports the error.
+   */
+  private[effect] def reportError(implicit ev: A <:< Unit): IO[Unit] = {
+    val _ = ev
+    asInstanceOf[IO[Unit]].handleErrorWith { t =>
+      IO.executionContext.flatMap(ec => IO(ec.reportFailure(t)))
+    }
+  }
+
+  /**
+   * Discard any error raised by the source.
+   */
   def voidError(implicit ev: A <:< Unit): IO[Unit] = {
     val _ = ev
     asInstanceOf[IO[Unit]].handleError(_ => ())
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -493,7 +493,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
         IO.executionContext.flatMap(ec => IO(ec.reportFailure(t)))
       }
 
-      poll(this).onCancel(finalizer).onError(_ => handled).flatTap(_ => finalizer)
+      poll(this).onCancel(finalizer).onError { case _ => handled }.flatTap(_ => finalizer)
     }
 
   /**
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -588,9 +589,19 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   def onCancel(fin: IO[Unit]): IO[A] =
     IO.OnCancel(this, fin)
 
+  @deprecated("Use onError with PartialFunction argument", "3.6.0")
   def onError(f: Throwable => IO[Unit]): IO[A] =
     handleErrorWith(t => f(t).voidError *> IO.raiseError(t))
```
armanbilge: Can we delegate this to the other method?
```suggestion
    onError(f: PartialFunction[Throwable, IO[Unit]])
```

lenguyenthanh: mm, I'm not sure I understand, I think your suggestion shouldn't be compiled?
Or do you mean use the new `onError`?
btw, I'm about to go to bed, will think more and answer tomorrow.

armanbilge: > I think your suggestion shouldn't be compiled? Or do you mean use the new `onError`?

Oops, I made a typo. But I think you are right, this won't compile as-is 🤔 

Basically we should avoid to duplicate the `onError` implementation. We can share the implementation by delegating one `onError` to the other `onError`, by converting `Function` to `PartialFunction` or vice versa.

lenguyenthanh: Fixed this in [f4f505c](https://github.com/typelevel/cats-effect/pull/4121/commits/f4f505c8e70adb930662d33a22e0415ff6018e07)

Another way would be 
```diff
-    handleErrorWith(t => pf.applyOrElse(t, (_: Throwable) => IO.unit) *> IO.raiseError(t))
+    onError(t => pf.applyOrElse(t, (_: Throwable) => IO.unit))
```

but it feels awkward as We used a function that We just deprecated it.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -588,8 +589,20 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
...
+  }
+
+  /**
+   * Execute a callback on certain errors, then rethrow them. Any non matching error is rethrown
+   * as well.
+   *
+   * Implements `ApplicativeError.onError`.
+   */
+  def onError(pf: PartialFunction[Throwable, IO[Unit]]): IO[A] =
+    handleErrorWith(t => pf.applyOrElse(t, (_: Throwable) => IO.unit) *> IO.raiseError(t))
```
armanbilge, line 599: To match the other implementation we need this `voidError`. But this raises an interesting question about the semantics: if the `onError` callback raises an error, which error should be propagated, that error or the original one? Also the `voidError` completely discards the error without reporting it, which seems really weird 😕 

```suggestion
    handleErrorWith(t => pf.applyOrElse(t, (_: Throwable) => IO.unit).voidError *> IO.raiseError(t))
```

armanbilge: I think what we should do is extract this into a helper method.
https://github.com/typelevel/cats-effect/blob/7168625ee8f960f6873f2692e73afba75b22d774/core/shared/src/main/scala/cats/effect/IO.scala#L493

Then instead of `voidError` we should redirect errors within the `onError` callback to that? And then re-raise the original error.

lenguyenthanh: thanks @armanbilge, I'm completely missed that `voidError`. I added `reportError` helper method as your suggestion. I think this is a great improvement. The name is subjecting for bike-shedding. We can also use this `reportError` method in the two other places. I can do that in this pr.

armanbilge: > I'm completely missed that `voidError`

Yes, I almost missed it as well. We should add a test to cover this case (what happens if `onError` callback raises an error), that would have helped us. Actually we might even consider adding this as a law to Cats, so that `onError` semantics are the same across all implementations.

> We can also use this `reportError` method in the two other places. I can do that in this pr.

🚀 it looks good! I think name is reasonable and it is private internal anyway.

lenguyenthanh: cool, I'll add a test case for in this first. And think about a law for cats later on.

armanbilge: Hmm, actually, now I am hesitating 😅 

The goal of this PR is to make `IO#onError` consistent with `ApplicativeError#onError`. If we use `voidError`/`reportError` then the semantics are inconsistent again with `ApplicativeError` 😕 

https://github.com/typelevel/cats/blob/8e8724a29881a394bc39357a7f0cd21124573f30/core/src/main/scala/cats/ApplicativeError.scala#L261-L262

Unless we change it in Cats (and this may be too drastic of a change) I think we just have to let the callback's error propagate instead.

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -588,8 +589,21 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
...
+
+  /**
+   * Execute a callback on certain errors, then rethrow them. Any non matching error is rethrown
+   * as well.
+   *
+   * Implements `ApplicativeError.onError`.
+   */
+  def onError(pf: PartialFunction[Throwable, IO[Unit]]): IO[A] =
+    handleErrorWith(t =>
+      pf.applyOrElse(t, (_: Throwable) => IO.unit).reportError *> IO.raiseError(t))
```
armanbilge: I guess we should use `reportError` for the deprecated `onError`, but for the `onError` consistent with `ApplicativeError` we have to match the semantics defined in Cats.

```suggestion
  @deprecated("Use onError with PartialFunction argument", "3.6.0")
  def onError(f: Throwable => IO[Unit]): IO[A] = {
    val pf: PartialFunction[Throwable, IO[Unit]] = { case t => f(t).reportError }
    onError(pf)
  }

  /**
   * Execute a callback on certain errors, then rethrow them. Any non matching error is rethrown
   * as well.
   *
   * Implements `ApplicativeError.onError`.
   */
  def onError(pf: PartialFunction[Throwable, IO[Unit]]): IO[A] =
    handleErrorWith(t =>
      pf.applyOrElse(t, (_: Throwable) => IO.unit)*> IO.raiseError(t))
```

core/shared/src/main/scala/cats/effect/IO.scala
lenguyenthanh: ahhhh, you're right. So, my first implemenaton was accidentally correct :p.

core/shared/src/main/scala/cats/effect/IO.scala
lenguyenthanh: fixed in: https://github.com/typelevel/cats-effect/pull/4121/commits/868451b090e407094d01eaa953ee8b4fde69ac5c

armanbilge (dismissed):
Nice, this looks good to me. It turns out the old `onError` was inconsistent both in signature and in behavior, so makes sense to deprecate.

armanbilge:
@lenguyenthanh if you want a small follow-up PR, I think it would be worthwhile to add a law to Cats testing this behavior when `onError` raises an error.

lenguyenthanh:
thanks @armanbilge for your guidance, It helped me a lot!

For the `onError` laws, iiuc, We already have it: https://github.com/typelevel/cats/blob/8e8724a29881a394bc39357a7f0cd21124573f30/laws/src/main/scala/cats/laws/ApplicativeErrorLaws.scala#L70-L71C63

Not sure why We didn't catch it here, maybe because of the generator?

armanbilge:
> For the `onError` laws, iiuc, We already have it:

@lenguyenthanh ah yes, you're right. Even though there is no explicit test for this case, it should be covered by the generator.

> Not sure why We didn't catch it here, maybe because of the generator?

I was confused about this as well, but I think I know why. We actually need to do one more thing in your PR: we need to add an `onError` override to the `IOAsync` instances, because otherwise your implementation is actually not being law tested.

https://github.com/typelevel/cats-effect/blob/868451b090e407094d01eaa953ee8b4fde69ac5c/core/shared/src/main/scala/cats/effect/IO.scala#L1965

Actually, we should make sure that every custom method on `IO` is overridden in the `IOAsync` instances. If we had done this at the beginning for `onError` we would have realized that the signature and implementation are wrong.

lenguyenthanh:
> If we had done this at the beginning for onError we would have realized that the signature and implementation are wrong.

@armanbilge added the `onError` override to `IOAsync`. But it seems tests are passed even with the old `onError` method. I guess it's because a partial function is a function and our test suite doesn't work well with `throwable` 🤷. I'll dig more into this after #3757

armanbilge:
> But it seems tests are passed even with the old `onError` method.

Hmm, if I restore the semantics mismatch/bug, then the laws fail. So seems to be working.

```diff
diff --git a/core/shared/src/main/scala/cats/effect/IO.scala b/core/shared/src/main/scala/cats/effect/IO.scala
index ee3a9da1f..408b2a5e9 100644
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -596,7 +596,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
    * Implements `ApplicativeError.onError`.
    */
   def onError(pf: PartialFunction[Throwable, IO[Unit]]): IO[A] =
-    handleErrorWith(t => pf.applyOrElse(t, (_: Throwable) => IO.unit) *> IO.raiseError(t))
+    handleErrorWith(t => pf.applyOrElse(t, (_: Throwable) => IO.unit).voidError *> IO.raiseError(t))
 
   /**
    * Like `Parallel.parProductL`
```

```
failing seed for async.applicativeError onError raise is 7OQiZWdn0skINxhWMBdm5dKBPdF49wcLmELPfFNLY_D=
[error] x async.applicativeError onError raise
[error]  Falsified after 4 passed tests.
[error]  > ARG_1: java.lang.Exception
[error]  > ARG_2: IO(...)
[error]  The seed is 7OQiZWdn0skINxhWMBdm5dKBPdF49wcLmELPfFNLY_D=
```

armanbilge (approved):

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -588,8 +583,20 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
   def onCancel(fin: IO[Unit]): IO[A] =
     IO.OnCancel(this, fin)
 
-  def onError(f: Throwable => IO[Unit]): IO[A] =
-    handleErrorWith(t => f(t).voidError *> IO.raiseError(t))
+  @deprecated("Use onError with PartialFunction argument", "3.6.0")
```
joroKr21, line 586: Hmm, this deprecation might be a bit confusing, because lambda syntax `foo => bar` is a valid partial function but we can't use it with onError since it picks the deprecated overload.

armanbilge: Yes, that was an oversight ...
- https://github.com/typelevel/cats-effect/issues/4329

PR #4122, opened 2024-08-20
link to `MapRef` docs in site sidebar

samspills:
Currently not in the sidebar so no one can find it! I think this is all that's necessary


armanbilge (approved):
thanks!

PR #4127, opened 2024-08-28
Add lifecycle diagrams to `Supervisor` web docs

samspills:
I have been referencing these a lot, they're good docs!, and I thought they should be in the web documentation as well. 

I didn't write much of anything original I just copied the explanation and the diagrams over as it made sense. I tried to maintain the `motivation` and `supervisor` separation because I like it, but happy to rearrange as well if people disagree.

### DIFF
--- a/docs/std/supervisor.md
+++ b/docs/std/supervisor.md
@@ -11,6 +11,38 @@ title: Supervisor
 
 `Spawn[F]#background`: ties the lifecycle of the spawned fiber to that of the fiber that invoked `background`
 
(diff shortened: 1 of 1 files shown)


zarthross (approved):

durban (approved):

armanbilge (approved):
Thanks! As a follow-up task I wonder if we can render these more nicely with Mermaid 🧜‍♀️

samspills:
Oh yeah, mermaid is a good idea!

PR #4128, opened 2024-09-02
Implement a safer `Resource.attempt` which releases acquired resource(s) in case of error

lenguyenthanh:
This is my solution for #3757, 

As I discussed with @armanbilge, the new method requires `MonadCancelThrow` in order to use `Resource.allocatedCase` which handles releasing acquired resource(s) in case of error. This pr also removed implicit `ApplicativeError` and deprecated the current attempt method.

~~This is a draft as it still needs to add tests and address some issues that I'll comment in the pr. But please give me early feedback.~~

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/metrics/JvmCpuStarvationMetrics.scala
+++ b/core/jvm/src/main/scala/cats/effect/metrics/JvmCpuStarvationMetrics.scala
@@ -66,7 +66,7 @@ private[effect] object JvmCpuStarvationMetrics {
         case (mbeanServer, _) => IO.blocking(mbeanServer.unregisterMBean(mBeanObjectName))
       }
       .map(_._2)
-      .handleErrorWith[CpuStarvationMetrics, Throwable] { th =>
+      .handleErrorWith[CpuStarvationMetrics] { th =>
         Resource.eval(Console[IO].errorln(warning(th))).map(_ => new NoOpCpuStarvationMetrics)
       }
   }
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -731,7 +731,7 @@ sealed abstract class Resource[F[_], +A] extends Serializable {
     Resource.applyFull[F, Either[Throwable, A]] { poll =>
       poll(allocatedCase).attempt.map {
         case Right((a, r)) => (a.asRight[Throwable], r)
-        case error => (error, _ => F.unit)
+        case error => (error.asInstanceOf[Either[Throwable, A]], _ => F.unit)
       }
     }
 
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -1303,7 +1303,7 @@ private[effect] trait ResourceHOInstances3 extends ResourceHOInstances4 {
 
 private[effect] trait ResourceHOInstances4 extends ResourceHOInstances5 {
   @deprecated("Use MonadCancelThrow instances", "3.6.0")
-  def catsEffectMonadErrorForResource[F[_], E](
+  implicit def catsEffectMonadErrorForResource[F[_], E](
       implicit F0: MonadError[F, E]): MonadError[Resource[F, *], E] =
     new ResourceMonadError[F, E] {
       def F = F0
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -1302,7 +1302,9 @@ private[effect] trait ResourceHOInstances3 extends ResourceHOInstances4 {
 }
 
 private[effect] trait ResourceHOInstances4 extends ResourceHOInstances5 {
-  @deprecated("Use MonadCancelThrow instances", "3.6.0")
+  @deprecated(
+    "Bring an implicit MonadCancelThrow[F] into scope to get the fixed Resource instance",
+    "3.6.0")
   implicit def catsEffectMonadErrorForResource[F[_], E](
       implicit F0: MonadError[F, E]): MonadError[Resource[F, *], E] =
     new ResourceMonadError[F, E] {
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -724,8 +724,8 @@ sealed abstract class Resource[F[_], +A] extends Serializable {
   }
 
   def attempt(implicit F: MonadCancelThrow[F]): Resource[F, Either[Throwable, A]] =
-    Resource.applyCase[F, Either[Throwable, A]] {
-      allocatedCase.attempt.map {
+    Resource.applyFull[F, Either[Throwable, A]] { poll =>
+      poll(allocatedCase).attempt.map {
         case Left(error) => (error.asLeft[A], _ => ().pure[F])
         case Right((a, r)) => (a.asRight[Throwable], r)
       }
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -731,8 +731,16 @@ sealed abstract class Resource[F[_], +A] extends Serializable {
       }
     }
 
+  @deprecated("Use overload with MonadCancelThrow", "3.6.0")
   def handleErrorWith[B >: A, E](f: E => Resource[F, B])(
-      implicit F: ApplicativeError[F, E]): Resource[F, B] =
+      F: ApplicativeError[F, E]): Resource[F, B] =
+    attempt(F).flatMap {
+      case Right(a) => Resource.pure(a)
+      case Left(e) => f(e)
+    }
+
+  def handleErrorWith[B >: A](f: Throwable => Resource[F, B])(
+      implicit F: MonadCancelThrow[F]): Resource[F, B] =
     attempt(F).flatMap {
       case Right(a) => Resource.pure(a)
       case Left(e) => f(e)
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -700,7 +700,9 @@ sealed abstract class Resource[F[_], +A] extends Serializable {
       }
     }
 
-  def attempt[E](implicit F: ApplicativeError[F, E]): Resource[F, Either[E, A]] =
+  @deprecated("Use overload with MonadCancelThrow", "3.6.0")
+  def attempt[E](F: ApplicativeError[F, E]): Resource[F, Either[E, A]] = {
+    implicit val x: ApplicativeError[F, E] = F
     this match {
       case Allocate(resource) =>
         Resource.applyFull { poll =>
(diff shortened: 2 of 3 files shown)


kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -1462,7 +1473,7 @@ abstract private[effect] class ResourceMonadError[F[_], E]
   implicit protected def F: MonadError[F, E]
 
   override def attempt[A](fa: Resource[F, A]): Resource[F, Either[E, A]] =
-    fa.attempt
+    fa.attempt(F)
```
lenguyenthanh, line 1501: this still used the old `attempt` but We can't use the new one because the error type is not fixed to `Throwable`.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -710,19 +712,28 @@ sealed abstract class Resource[F[_], +A] extends Serializable {
...
+      allocatedCase.attempt.map {
+        case Left(error) => (error.asLeft[A], _ => ().pure[F])
+        case Right((a, r)) => (a.asRight[Throwable], r)
+      }
+    }
 
   def handleErrorWith[B >: A, E](f: E => Resource[F, B])(
       implicit F: ApplicativeError[F, E]): Resource[F, B] =
-    attempt.flatMap {
+    attempt(F).flatMap {
```
lenguyenthanh, line 741: imo, We should add a "safe" variant of `handleErrorWith` as well to avoid inconsistent behaviors between this and `attempt`.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
armanbilge: We should mark this instance as deprecated and create a new instance that requires `MonadCancelThrow`.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
armanbilge: Good catch. You are right, we definitely should do that.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -700,7 +700,9 @@ sealed abstract class Resource[F[_], +A] extends Serializable {
       }
     }
 
-  def attempt[E](implicit F: ApplicativeError[F, E]): Resource[F, Either[E, A]] =
+  @deprecated("Use overload with MonadCancelThrow", "3.6.0")
+  def attempt[E](F: ApplicativeError[F, E]): Resource[F, Either[E, A]] = {
+    implicit val x: ApplicativeError[F, E] = F
```
armanbilge: We can do a trick here to automatically fix more callsites without requirng them to recompile: we can pattern match on `F` and if it is `Sync` or `Async`  then we can use the fixed implementation instead.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -710,19 +712,28 @@ sealed abstract class Resource[F[_], +A] extends Serializable {
...
       case p @ Pure(_) =>
         Resource.pure(p.a.asRight)
       case e @ Eval(_) =>
         Resource.eval(e.fa.attempt)
     }
+  }
+
+  def attempt(implicit F: MonadCancelThrow[F]): Resource[F, Either[Throwable, A]] =
+    Resource.applyCase[F, Either[Throwable, A]] {
+      allocatedCase.attempt.map {
```
kubukoz: **question:** shouldn't the allocation itself be interruptible?

like `Resource.applyFull { poll => poll(allocatedCase).attempt.map...`

armanbilge: Good catch, this looks right. Let's make sure to add a test case for this.

lenguyenthanh: I though about using `poll` but didn't know for sure. I added tests and fix this in https://github.com/typelevel/cats-effect/pull/4128/commits/07fa387a2d555f37196b9fdb0a9a7419b20f5129 and https://github.com/typelevel/cats-effect/pull/4128/commits/28394460eadd78805ccead379d9dbe77ef58c6a2

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
lenguyenthanh: fixed in https://github.com/typelevel/cats-effect/pull/4128/commits/cb1a293b49af10ac9127ba23fbb69d801465ea79. But now https://github.com/typelevel/cats-effect/blob/cb1a293b49af10ac9127ba23fbb69d801465ea79/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala#L1300-L1306 complains because it use the deprecated `ResourceMonadError`.

I'm not sure what the best way to fix this. I feel like We can remove it completely. But maybe it broke client code.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
lenguyenthanh: fixed in https://github.com/typelevel/cats-effect/pull/4128/commits/3c2d6af91df66c3d71088c9c6ebad884c06615e2

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
lenguyenthanh: I'm not sure how to implement your suggestion.

I tried:

```scala
  @deprecated("Use overload with MonadCancelThrow", "3.6.0")
  def attempt[E](F: ApplicativeError[F, E]): Resource[F, Either[E, A]] =
    F match {
      case x: Async[_] =>
        implicit val x: MonadCancelThrow[F] = x
        attempt
      case x: Sync[_] => attempt(x)
        implicit val x: MonadCancelThrow[F] = x
        attempt
      case _ => ???
   }
```

But it doesn't work as `attempt` with `MonadCancelThrow` returns `Resource[F, Either[Throwable, A]]` but not `Resource[F, Either[E, A]]`

core/jvm/src/main/scala/cats/effect/metrics/JvmCpuStarvationMetrics.scala
```diff
@@ -66,7 +66,7 @@ private[effect] object JvmCpuStarvationMetrics {
         case (mbeanServer, _) => IO.blocking(mbeanServer.unregisterMBean(mBeanObjectName))
       }
       .map(_._2)
-      .handleErrorWith[CpuStarvationMetrics, Throwable] { th =>
+      .handleErrorWith[CpuStarvationMetrics] { th =>
```
lenguyenthanh, line 69: This change because `handleErrorWith[E, A]` now requires explicit `ApplicativeError`

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
armanbilge: You can cast unsafely. If `F` is `Sync` then `E` must be `Throwable`. Also I just realized since `Async <: Sync` we only need to match for `Sync`.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
armanbilge: We deprecate this one too :)

lenguyenthanh: We deprecate this one but keep the trait hierarchy?

armanbilge: That's a good question. Maybe we should try to deprioritize this instance by moving it to the bottom of the hierarchy. Alternatively, we could just remove the `implicit` modifier.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
lenguyenthanh: Fixed it https://github.com/typelevel/cats-effect/pull/4128/commits/3b7d0ee83c4ecfeb639b1fa75493ca6b73c1b823

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
lenguyenthanh: I kept the hierachy the same but removed implicit modifier and deprecated `catsEffectMonadErrorForResource` function.

armanbilge: I thought about it more and actually I think it's okay to leave the `implicit` there. We don't need to change any prioritization, because we aren't introducing any new instances, and the instance derived from `MonadCancelThrow` (which is required for the fixed `attempt`) is already higher priority than this one.

By leaving it `implicit`, users will get an actionable deprecation instead of non-compiling code.

tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
```diff
@@ -1172,6 +1172,29 @@ class ResourceSpec extends BaseSpec with ScalaCheck with Discipline {
...
   }
 
+  "attempt" >> {
+
+    "releases resource on error" in ticked { implicit ticker =>
+      IO.ref(0)
+        .flatMap { ref =>
+          val resource = Resource.make(ref.update(_ + 1))(_ => ref.update(_ + 1))
+          val error = Resource.raiseError[IO, Unit, Throwable](new Exception)
+          (resource *> error).attempt.use { _ => ref.get.map { _ must be_==(2) } }
```
armanbilge: instead of ignoring the value `_` we can check that it's a `Left`.

tests/shared/src/test/scala/cats/effect/ResourceSpec.scala
```diff
@@ -1172,6 +1172,29 @@ class ResourceSpec extends BaseSpec with ScalaCheck with Discipline {
...
+        }
+        .void must completeAs(())
+    }
+
+    "acquire is interruptible" in ticked { implicit ticker =>
+      val sleep = IO.sleep(1.second)
```
armanbilge: Could also be:
```suggestion
      val sleep = IO.never
```

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -700,29 +700,52 @@ sealed abstract class Resource[F[_], +A] extends Serializable {
...
-        Resource.pure(p.a.asRight)
-      case e @ Eval(_) =>
-        Resource.eval(e.fa.attempt)
     }
 
+  def attempt(implicit F: MonadCancelThrow[F]): Resource[F, Either[Throwable, A]] =
+    Resource.applyFull[F, Either[Throwable, A]] { poll =>
+      poll(allocatedCase).attempt.map {
+        case Left(error) => (error.asLeft[A], _ => ().pure[F])
+        case Right((a, r)) => (a.asRight[Throwable], r)
```
armanbilge: Some shuffling to avoid allocations and improve performance.

```suggestion
        case Right((a, r)) => (a.asRight[Throwable], r)
        case error => (error, _ => F.unit)
```

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -1330,10 +1354,19 @@ abstract private[effect] class ResourceFOInstances1 {
 }
 
 abstract private[effect] class ResourceMonadCancel[F[_]]
-    extends ResourceMonadError[F, Throwable]
+    extends ResourceMonad[F]
     with MonadCancel[Resource[F, *], Throwable] {
   implicit protected def F: MonadCancel[F, Throwable]
 
+  override def attempt[A](fa: Resource[F, A]): Resource[F, Either[Throwable, A]] =
+    fa.attempt(F)
```
armanbilge: Unless it's necessary for some reason? 🤔 

```suggestion
    fa.attempt
```

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -700,29 +700,52 @@ sealed abstract class Resource[F[_], +A] extends Serializable {
...
-    attempt.flatMap {
+      F: ApplicativeError[F, E]): Resource[F, B] =
+    attempt(F).flatMap {
+      case Right(a) => Resource.pure(a)
+      case Left(e) => f(e)
+    }
+
+  def handleErrorWith[B >: A](f: Throwable => Resource[F, B])(
+      implicit F: MonadCancelThrow[F]): Resource[F, B] =
+    attempt(F).flatMap {
```
armanbilge: Unless it's necessary for some reason?

```suggestion
    attempt.flatMap {
```

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -1279,6 +1302,7 @@ private[effect] trait ResourceHOInstances3 extends ResourceHOInstances4 {
 }
 
 private[effect] trait ResourceHOInstances4 extends ResourceHOInstances5 {
+  @deprecated("Use MonadCancelThrow instances", "3.6.0")
```
armanbilge: Maybe a message like this can be more helpful to the user?

```suggestion
  @deprecated("Bring an implicit MonadCancelThrow[F] into scope to get the fixed Resource instance", "3.6.0")
```

armanbilge (commented):
Looking good, one last comment :)

armanbilge (approved):
nice work :)

djspiewak (approved):

PR #4131, opened 2024-09-08
Make `joinWithNever` attempting self cancelling

### LINKED ISSUE #4093: Potential inconsistency between joinWithNever and docs
Hi there,

In the [docs](https://typelevel.org/cats-effect/docs/typeclasses/spawn) for the the `Spawn` typeclass at the very end it states:

> In English, the semantics of this are as follows:
> 
> If the child fiber completed successfully, produce its result
> If it errored, re-raise the error within the current fiber
> If it canceled, attempt to self-cancel, and if the self-cancelation fails, deadlock
> Sometimes this is an appropriate semantic, and the cautiously-verbose joinWithNever function implements it for you. 

However, the `joinWithNever` implementation doesn't look like it actually makes any attempts at self cancellation and I think I've confirmed this with a small test app.

```
object FunTest extends IOApp.Simple {
  override val run: IO[Unit] = for {
    _ <- IO.println("Starting")
    child = IO.println("Starting child...") >>
      IO.sleep(1.second) >>
      IO.canceled >>
      IO.sleep(1.second) >>
      IO.println("Child completed")
    childFiber     <- child.start
    dependentChild <- childFiber.joinWithNever.as("I managed to return something").start
    result <- dependentChild
      .joinWith(IO.pure("I was cancelled"))
      .timeoutTo(3.seconds, "I waited too long :(".pure[IO])
    _ <- IO.println(s"result is $result")
  } yield ()
}
```

which outputs the following to console

```
Starting
Starting child...
result is I waited too long :(

Process finished with exit code 0
```

lenguyenthanh:
Close #4093 by attempting self-cancel before `F.never`

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Fiber.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Fiber.scala
@@ -68,10 +68,11 @@ trait Fiber[F[_], E, A] extends Serializable {
    *
    * If the fiber completes with [[Outcome.Succeeded]], the successful value is returned. If the
    * fiber completes with [[Outcome.Errored]], the error is raised. If the fiber completes with
-   * [[Outcome.Canceled]], the caller is indefinitely suspended without termination.
+   * [[Outcome.Canceled]], attempt to self-cancel, and if the self-cancelation fails, the caller
+   * is indefinitely suspended without termination.
    */
   def joinWithNever(implicit F: GenSpawn[F, E]): F[A] =
-    joinWith(F.never)
+    joinWith(F.canceled >> F.never)
 
   /**
    * Awaits the completion of the bound fiber and returns its result once it completes.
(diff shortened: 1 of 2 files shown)


djspiewak:
So the concern I have with this is that there's a lot of call-sites which kind of implicitly assume it *isn't* self-canceling. This is a pretty significant semantic to change. The inconsistency with the docs was actually reported back before 3.0.0 went final, and we decided that the mistaken semantic (the one presently in HEAD) was more expected, but we never went back and fixed the docs.

It think it might be better to fix the docs to match the semantic, rather than changing something like this which can have really far-reaching consequences.

armanbilge:
See also a similar confusion with `embedNever` in https://github.com/typelevel/cats-effect/pull/3351#discussion_r1070651800.

lenguyenthanh:
that makes sense. I'll submit another pr for updating the docs.

PR #4133, opened 2024-09-10
Add `ownPoller` 

### LINKED ISSUE #4009: Add API to polling system to attempt to get current poller without shifting
This would mostly be to support cancelation with a similar mechanism used in the new timer heap introduced in https://github.com/typelevel/cats-effect/pull/3781. Specifically, using this API you can check if your current thread owns the poller you need to send the cancel to. If so, it can be done on a fast/simple happy-path, otherwise it will need to go through some concurrent-safe mechanism.

armanbilge:
This PR introduces a mechanism for checking if a given poller is "owned" by the current thread and thus is safe to interact with. This is useful primarily for cancelation, which may be requested from any thread, but can happy-path additional cleanup operations when on the owning thread.

The existing `accessPoller` and new `ownPoller` methods are now bundled up in a `PollerProvider` API.

Finally, `SelectorSystem` is improved to use `ownPoller` to attempt to remove the linked list node for a canceled selection operation, instead of simply clearing it.

Closes https://github.com/typelevel/cats-effect/issues/4009.

### DIFF
--- a/core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
+++ b/core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
@@ -104,7 +104,7 @@ abstract class PollingSystem {
 
 }
 
-trait PollingContext[P] {
+sealed trait PollingContext[P] {
 
   /**
    * Register a callback to obtain a thread-local `Poller`
@@ -118,6 +118,8 @@ trait PollingContext[P] {
    */
   def ownPoller(poller: P): Boolean
 }
+
+private[unsafe] trait UnsealedPollingContext[P] extends PollingContext[P]
 
 object PollingSystem {
 
--- a/core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
+++ b/core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
@@ -51,7 +51,7 @@ abstract class PollingSystem {
   /**
    * Creates a new instance of the user-facing interface.
    */
-  def makeApi(provider: PollerProvider[Poller]): Api
+  def makeApi(ctx: PollingContext[Poller]): Api
 
   /**
    * Creates a new instance of the thread-local data structure used for polling.
@@ -104,7 +104,7 @@ abstract class PollingSystem {
 
 }
 
-trait PollerProvider[P] {
+trait PollingContext[P] {
 
   /**
    * Register a callback to obtain a thread-local `Poller`
--- a/core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
+++ b/core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
@@ -50,13 +50,8 @@ abstract class PollingSystem {
 
   /**
    * Creates a new instance of the user-facing interface.
-   *
-   * @param access
-   *   callback to obtain a thread-local `Poller`.
-   * @return
-   *   an instance of the user-facing interface `Api`.
    */
-  def makeApi(access: (Poller => Unit) => Unit): Api
+  def makeApi(provider: PollerProvider[Poller]): Api
 
   /**
    * Creates a new instance of the thread-local data structure used for polling.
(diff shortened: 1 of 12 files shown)


durban (dismissed):
LGTM

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
```diff
@@ -109,7 +104,20 @@ abstract class PollingSystem {
...
-private object PollingSystem {
+trait PollerProvider[P] {
+
+  /**
+   * Register a callback to obtain a thread-local `Poller`
+   */
+  def accessPoller(cb: P => Unit): Unit
+
+  /**
+   * Returns `true` if it is safe to interact with this `Poller`
```
djspiewak: Can we add a note that it's always safe to return `false` here? So in other words, we tolerate false negatives, but we cannot tolerate false positives.

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
```diff
@@ -109,7 +104,20 @@ abstract class PollingSystem {
 
 }
 
-private object PollingSystem {
+trait PollerProvider[P] {
+
+  /**
+   * Register a callback to obtain a thread-local `Poller`
+   */
+  def accessPoller(cb: P => Unit): Unit
```
djspiewak, line 112: Bikeshed: remove the `Poller` suffix from both methods, since it's redundant.

core/jvm/src/main/scala/cats/effect/unsafe/SelectorSystem.scala
```diff
@@ -161,9 +150,55 @@ object SelectorSystem {
 
   def apply(): SelectorSystem = apply(SelectorProvider.provider())
 
-  private final class CallbackNode(
-      var interest: Int,
-      var callback: Either[Throwable, Int] => Unit,
-      var next: CallbackNode
-  )
+  private final class Callbacks {
```
djspiewak, line 153: We've built so many of these…

djspiewak (changes requested):

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
armanbilge: I don't know if it's a good enough reason for you but I chose `accessPoller` so that WSTP can directly implement `PollerProvider` without methods vaguely named `access` or `own`.

FWIW I'd rather bikeshed the name `PollerProvider`, it's possible in the future we may want to add other useful methods in here that are exposed by the WSTP to `PollingSystem` implementors.

djspiewak: Alrighty. Will ponder that one for a bit

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
```diff
@@ -109,7 +104,22 @@ abstract class PollingSystem {
 
 }
 
-private object PollingSystem {
+trait PollerProvider[P] {
```
djspiewak: What about `PollerContext` or something in that vein? Kinda sucks that everything is "poller this" and "poller that". Could also be slightly more generic like `WorkerContext` or `TLContext`. `ReferenceFrame` is not inaccurate but more obscure.

armanbilge: Renamed to `PollingContext`

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
```diff
@@ -109,7 +104,24 @@ abstract class PollingSystem {
...
+
+  /**
+   * Returns `true` if it is safe to interact with this `Poller`. Implementors of this method
+   * may be best-effort: it is always safe to return `false`, so callers must have an adequate
+   * fallback for the non-owning case.
+   */
+  def ownPoller(poller: P): Boolean
+}
+
+private[unsafe] trait UnsealedPollingContext[P] extends PollingContext[P]
```
armanbilge, line 122: Before you tell me that `private[unsafe]` is the same as `private` since we are already in package `unsafe`, it's not:

```
private trait UnsealedPollingContext escapes its defining scope as part of type cats.effect.unsafe.UnsealedPollingContext[P]
```

djspiewak: Thanks Scala!

djspiewak (dismissed):

Issue #3909, opened 2023-12-01
Dynamically resize the WSTP if `availableProcessors()` changes


armanbilge:
We currently rely on [`Runtime#availableProcessors`](https://docs.oracle.com/javase/8/docs/api/java/lang/Runtime.html#availableProcessors--) to size the work-stealing threadpool. The JavaDoc warns that
> Applications that are sensitive to the number of available processors should therefore occasionally poll this property and adjust their resource usage appropriately.

A concrete example of an environment where the number of available processors may change is Kubernetes with the [Vertical Pod Autoscaler](https://github.com/kubernetes/autoscaler/tree/85b6058131746ac9662449b56464e3300a2bf8e7/vertical-pod-autoscaler).
> The Kubernetes Vertical Pod Autoscaler automatically adjusts the CPU and memory reservations for your Pods to help "right size" your applications.

Marking as "experiment" since this would be a non-trivial enhancement :) and maybe not worth the complexity.

durban:
I'm nitpicking, but I don't think the VPA actually dynamically changes container resources. Based on the documentation, it seems to recreate pods with the changed resources.

However, it _is_ possible to dynamically change CPU requests/limits ([see here](https://kubernetes.io/blog/2023/05/12/in-place-pod-resize-alpha/)), so this issue is definitely relevant. This k8s feature is somewhat new, and behind a feature gate, but it works. (I did try it, and observed the return value of `availableProcessors()` changing in a single JVM.)

(I assume the VPA will also use this feature in the future, [see here](https://github.com/kubernetes/autoscaler/issues/4016).)

Interestingly, the `ForkJoinPool` doesn't seem to follow the recommendation in the `availableProcessors()` javadoc (it only calls it once).

djspiewak:
> Interestingly, the ForkJoinPool doesn't seem to follow the recommendation in the availableProcessors() javadoc (it only calls it once).

This is fascinating. I've never seen this note before. For posterity:

> This value may change during a particular invocation of the virtual machine. Applications that are sensitive to the number of available processors should therefore occasionally poll this property and adjust their resource usage appropriately.

So the problem is that actually *using* this information is somewhat difficult. Resizing the number of worker threads is kinda possible in theory, but I can see a whole host of objections very quickly:

- There are a number of places where we are almost certainly getting higher performance because the worker count is fixed for the lifetime of the application
[...]

PR #4139, opened 2024-09-17
Add default constructor for `MapRef`

### LINKED ISSUE #4136: Default constructor for `MapRef`
We currently have many `MapRef` implementations, based on `Ref[Map]`, sharding, `ConcurrentHashMap`, etc. We should have an `apply` constructor that chooses a reasonable default implementation, _probably_ `ConcurrentHashMap`.

yisraelU:
Adds an apply method for MapRef. If the Effect type is an instance of Async , the `ConcurrentHashMap` implementation will be used otherwise the `sharedImmutableMap` will be used with the number of shards set to the number of Processors determined by `Runtime.getRuntime.availableProcessors`.
The goal of this MR is to improve the UX for a user by having a sane default

closes #4136

P.S. as this is a first-time PR to this repo, please let me know if there are any conventions/practices, etc.. that I should be adhering to. TY

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,39 +1,5 @@
 {
-  "name": "cats-effect",
-  "lockfileVersion": 2,
-  "requires": true,
-  "packages": {
-    "": {
-      "devDependencies": {
-        "source-map-support": "^0.5.19"
-      }
-    },
-    "node_modules/buffer-from": {
-      "version": "1.1.2",
-      "resolved": "https://registry.npmjs.org/buffer-from/-/buffer-from-1.1.2.tgz",
-      "integrity": "sha512-E+XQCRwSbaaiChtv6k6Dwgc+bx+Bs6vuKJHHl5kox/BaKbhiXzqQOwK4cO22yElGp2OCmjwVhT3HmxgyPGnJfQ==",
-      "dev": true
-    },
-    "node_modules/source-map": {
-      "version": "0.6.1",
-      "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.6.1.tgz",
-      "integrity": "sha512-UjgapumWlbMhkBgzT7Ykc5YXUT46F0iKu8SGXq0bcwP5dz/h0Plj6enJqjz1Zbq2l5WaqYnrVbwWOWMyF3F47g==",
-      "dev": true,
-      "engines": {
-        "node": ">=0.10.0"
-      }
-    },
-    "node_modules/source-map-support": {
-      "version": "0.5.19",
-      "resolved": "https://registry.npmjs.org/source-map-support/-/source-map-support-0.5.19.tgz",
-      "integrity": "sha512-Wonm7zOCIJzBGQdB+thsPar0kYuCIzYvxZwlBa87yi/Mdjv7Tip2cyVbLj5o0cFPN4EVkuTwb3GDDyUx2DGnGw==",
-      "dev": true,
-      "dependencies": {
-        "buffer-from": "^1.0.0",
-        "source-map": "^0.6.0"
-      }
-    }
-  },
+  "lockfileVersion": 1,
   "dependencies": {
     "buffer-from": {
       "version": "1.1.2",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,39 @@
 {
-  "lockfileVersion": 1,
+  "name": "cats-effect",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {
+    "": {
+      "devDependencies": {
+        "source-map-support": "^0.5.19"
+      }
+    },
+    "node_modules/buffer-from": {
+      "version": "1.1.2",
+      "resolved": "https://registry.npmjs.org/buffer-from/-/buffer-from-1.1.2.tgz",
+      "integrity": "sha512-E+XQCRwSbaaiChtv6k6Dwgc+bx+Bs6vuKJHHl5kox/BaKbhiXzqQOwK4cO22yElGp2OCmjwVhT3HmxgyPGnJfQ==",
+      "dev": true
+    },
+    "node_modules/source-map": {
+      "version": "0.6.1",
+      "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.6.1.tgz",
+      "integrity": "sha512-UjgapumWlbMhkBgzT7Ykc5YXUT46F0iKu8SGXq0bcwP5dz/h0Plj6enJqjz1Zbq2l5WaqYnrVbwWOWMyF3F47g==",
+      "dev": true,
+      "engines": {
+        "node": ">=0.10.0"
+      }
+    },
+    "node_modules/source-map-support": {
+      "version": "0.5.19",
+      "resolved": "https://registry.npmjs.org/source-map-support/-/source-map-support-0.5.19.tgz",
+      "integrity": "sha512-Wonm7zOCIJzBGQdB+thsPar0kYuCIzYvxZwlBa87yi/Mdjv7Tip2cyVbLj5o0cFPN4EVkuTwb3GDDyUx2DGnGw==",
+      "dev": true,
+      "dependencies": {
+        "buffer-from": "^1.0.0",
+        "source-map": "^0.6.0"
+      }
+    }
+  },
   "dependencies": {
     "buffer-from": {
       "version": "1.1.2",
--- a/std/shared/src/main/scala/cats/effect/std/MapRef.scala
+++ b/std/shared/src/main/scala/cats/effect/std/MapRef.scala
@@ -47,7 +47,7 @@ object MapRef extends MapRefCompanionPlatform {
 
   /**
    * Default constructor for [[MapRef]]. If [[Sync]] is available, it will delegate to
-   * [[ofConcurrentHashMap], otherwise it will fallback to [[ofShardedImmutableMap]].
+   * [[ofConcurrentHashMap]], otherwise it will fallback to [[ofShardedImmutableMap]].
    */
   def apply[F[_]: Concurrent, K, V]: F[MapRef[F, K, Option[V]]] = {
     Concurrent[F] match {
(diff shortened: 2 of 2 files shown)


package-lock.json
```diff
@@ -1,39 +1,5 @@
 {
-  "name": "cats-effect",
```
BalmungSan: I am not sure why there is a `package-lock` in this repo. But, I doubt you wanted to modify it :)
I guess it is best to revert these changes.

std/shared/src/main/scala/cats/effect/std/MapRef.scala
```diff
@@ -45,6 +45,22 @@ trait MapRef[F[_], K, V] extends Function1[K, Ref[F, V]] {
...
+  /**
+   * the default Constructor for MapRef. If Async is available, it will use a ConcurrentHashMap,
+   * otherwise it will use a sharded immutable map.
+   */
+  def apply[F[_]: Concurrent, K, V](): F[MapRef[F, K, Option[V]]] = {
+
+    Concurrent[F] match {
+      case a: Async[_] =>
+        implicit val async: Async[F] = a
+        ofConcurrentHashMap()
```
BalmungSan: IIRC you can match on `Async[F]`, while yes the `F` will be erased it would preserve the right type.

BalmungSan: Nitpick: rather than creating a dummy `implicit val`, you may pass `a` explicitly to `ofConcurrentHashMap()(a)`

std/shared/src/main/scala/cats/effect/std/MapRef.scala
```diff
@@ -45,6 +45,22 @@ trait MapRef[F[_], K, V] extends Function1[K, Ref[F, V]] {
...
+      case _ =>
+        ofShardedImmutableMap[F, K, V](Runtime.getRuntime.availableProcessors())
```
BalmungSan: Nitpick: I would use named arguments and remove the empty parenthesis from `availableProcessors`

std/shared/src/main/scala/cats/effect/std/MapRef.scala
BalmungSan: Oh it seems `ofConcurrentHashMap` only requires `Sync`, thus we should only check for that rather than `Async`
Apologies for the confusion, I thought it would require `Async` 😅

std/shared/src/main/scala/cats/effect/std/MapRef.scala
```diff
@@ -45,6 +45,22 @@ trait MapRef[F[_], K, V] extends Function1[K, Ref[F, V]] {
 
 object MapRef extends MapRefCompanionPlatform {
 
+  /**
+   * the default Constructor for MapRef. If Async is available, it will use a ConcurrentHashMap,
+   * otherwise it will use a sharded immutable map.
+   */
+  def apply[F[_]: Concurrent, K, V](): F[MapRef[F, K, Option[V]]] = {
```
BalmungSan: `apply` should not have an empty parameter list.

BalmungSan (commented):
While this alone is great.
We may also request a default value in `apply` to improve UX as suggested by @Jasper-M

package-lock.json
armanbilge: We need an npm package for testing tracing on JS.

https://github.com/typelevel/cats-effect/blob/eb231ffb0f5368792957a8f955b5836556b8a3e6/package.json#L3-L5

std/shared/src/main/scala/cats/effect/std/MapRef.scala
armanbilge: > remove the empty parenthesis from `availableProcessors`

It's technically side-effectual, it may return different values at different times.

- https://github.com/typelevel/cats-effect/issues/3909

yisraelU: @BalmungSan @armanbilge 
So I am confused about this. Should this call be viewed as not referentially transparent and wrapped in IO ? 
I noticed, in other places such as https://github.com/typelevel/cats-effect/blob/eb231ffb0f5368792957a8f955b5836556b8a3e6/core/jvm/src/main/scala/cats/effect/IOApp.scala#L191 directly without being wrapped in IO and its called with the parens 

Also from the JavaDoc 
```This value may change during a particular invocation of the virtual machine. Applications that are sensitive to the number of available processors should therefore occasionally poll this property and adjust their resource usage appropriately.```

File: std/shared/src/main/scala/cats/effect/std/MapRef.scala (hunk 2 above)
Comment (yisraelU) [reply std/shared/src/main/scala/cats/effect/std/MapRef.scala]: > Nitpick: rather than creating a dummy `implicit val`, you may pass `a` explicitly to `ofConcurrentHashMap()(a)`

Cool , Ididn't know you could do that when a method is defined with a Context Bound

File: package-lock.json (hunk 1 above)
Comment (yisraelU) [reply package-lock.json]: It's coming from something autogenerated . Ill try to see how this occured

Comment (yisraelU) [reply package-lock.json]: fixed

**Comment by yisraelU** [comment]
> While this alone is great. Wwe may also request a default value in `apply` to improve UX as suggested by @Jasper-M

I have added an overloaded apply . WDYT

File: std/shared/src/main/scala/cats/effect/std/MapRef.scala (hunk 3 above)
Comment (armanbilge) [reply std/shared/src/main/scala/cats/effect/std/MapRef.scala]: > So I am confused about this. Should this call be viewed as not referentially transparent and wrapped in IO ?

Technically, you are completely right. But we have decided to cheat about this in several places as you've noticed, because it would be impractical to properly suspend this.

See also https://github.com/typelevel/cats-effect/issues/3909. tl;dr we are assuming it's a static value.

Comment (BalmungSan) [reply std/shared/src/main/scala/cats/effect/std/MapRef.scala]: Right my bad, I thought that value may not change.
Yeah, is best to leave the parenthesis then. Ideally, it should even be wrapped but here we can't wrap it so as you have it is the best :)

**Comment by BalmungSan** [comment]
> I have added an overloaded apply . WDYT

Sadly I don't think that would work well: https://scastie.scala-lang.org/BalmungSan/WHoq2ShkTCm3QSs5puHixw

I think the best would be to do what Arman suggested and rather add a `withDefault` _(better names are welcome)_ extension method to `MapRefOptionOps`. So that users can do:

```scala
val m1 = MapRef[IO, K, V] // Returns MapRef[IO, K, Option[V]
val m2 = m1.withDefault(foo) // Returns MapRef[IO, K, V]
```

yisraelU:
> > I have added an overloaded apply . WDYT
> 
> Sadly I don't think that would work well: https://scastie.scala-lang.org/BalmungSan/WHoq2ShkTCm3QSs5puHixw
> 
> I think the best would be to do what Arman suggested and rather add a `withDefault` _(better names are welcome)_ extension method to `MapRefOptionOps`. So that users can do:
> 
> ```scala
> val m1 = MapRef[IO, K, V] // Returns MapRef[IO, K, Option[V]
> val m2 = m1.withDefault(foo) // Returns MapRef[IO, K, V]
> ```

I see, it fails for certain types.
If we go back to empty parens for the apply method it should work. 
I'm happy with the extension method solution though, and I have nothing better than `withDefault` :)

BalmungSan (approved):

std/shared/src/main/scala/cats/effect/std/MapRef.scala
```diff
@@ -45,6 +45,20 @@ trait MapRef[F[_], K, V] extends Function1[K, Ref[F, V]] {
...
+  /**
+   * the default Constructor for MapRef. If Async is available, it will use a ConcurrentHashMap,
+   * otherwise it will use a sharded immutable map.
+   */
+  def apply[F[_]: Concurrent, K, V]: F[MapRef[F, K, Option[V]]] = {
+    Concurrent[F] match {
+      case s: Sync[F] =>
+        ofConcurrentHashMap()(s)
+      case _ =>
+        ofShardedImmutableMap[F, K, V](shardCount = Runtime.getRuntime.availableProcessors())
```
durban, line 57: The `availableProcessors` call is technically side-effecting, so should be inside `F`. But here we don't have `Sync`, so I don't really know what to do...

armanbilge: See https://github.com/typelevel/cats-effect/pull/4139#discussion_r1763988320

std/shared/src/main/scala/cats/effect/std/MapRef.scala
```diff
@@ -45,6 +45,20 @@ trait MapRef[F[_], K, V] extends Function1[K, Ref[F, V]] {
 
 object MapRef extends MapRefCompanionPlatform {
 
+  /**
+   * the default Constructor for MapRef. If Async is available, it will use a ConcurrentHashMap,
+   * otherwise it will use a sharded immutable map.
+   */
```
armanbilge:
```suggestion
  /**
   * Default constructor for [[MapRef]]. If [[Sync]] is available, it will delegate to [[ofConcurrentHashMap],
   * otherwise it will fallback to [[ofShardedImmutableMap]].
   */
```

yisraelU: Thanks . I forgot to update after changing to Sync

std/shared/src/main/scala/cats/effect/std/MapRef.scala
durban: Thanks, sorry I missed that.

BalmungSan (approved):

std/shared/src/main/scala/cats/effect/std/MapRef.scala
```diff
@@ -45,6 +45,20 @@ trait MapRef[F[_], K, V] extends Function1[K, Ref[F, V]] {
 
 object MapRef extends MapRefCompanionPlatform {
 
+  /**
+   * Default constructor for [[MapRef]]. If [[Sync]] is available, it will delegate to
+   * [[ofConcurrentHashMap], otherwise it will fallback to [[ofShardedImmutableMap]].
```
armanbilge:
```suggestion
   * [[ofConcurrentHashMap]], otherwise it will fallback to [[ofShardedImmutableMap]].
```

durban (approved):

armanbilge (approved):
thanks!

PR #4148, opened 2024-09-29
Update to Scala 2.13.15 and fix Any inference

mzuehlke:
Closes: #4144

Fixes:
```
[error] /home/runner/work/cats-effect/cats-effect/tests/shared/src/test/scala/cats/effect/ResourceSpec.scala:989:32: a type was inferred to be `Any`; this may indicate a programming error.
[error]               acquiredLeft.get.ifM(loserReleased.get.map(_ must beLeft[Unit]).void, IO.pure(ok))
[error]                                ^
[error] No warnings can be incurred under -Werror.
[error] two errors found

```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-14]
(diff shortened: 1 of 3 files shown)


armanbilge (approved):
thank you!! 😊

PR #4150, opened 2024-10-16
Add `snapshot` to `MapRef`

iRevive:
### Motivation

`MapRef` can be considered as a replacement for `Ref[F, Map[K, V]]`. However, currently, there is no way to retrieve all available values.

#### Why not add `snapshot` method directly to the `MapRef` trait?

Two reasons:
- We cannot implement `snapshot` for the `def fromSeqRefs[F[_]: Functor, K, V]` (`traverse` requires `Applicative`)
- For the majority of the implementations, the result type would be `Map[K, Option[V]]`, which is suboptimal

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/MapRef.scala
+++ b/std/shared/src/main/scala/cats/effect/std/MapRef.scala
@@ -45,6 +45,14 @@ trait MapRef[F[_], K, V] extends Function1[K, Ref[F, V]] {
 
 object MapRef extends MapRefCompanionPlatform {
 
+  trait Snapshot[F[_], K, V] {
+
+    /**
+     * Returns the current snapshot of the available keys and values.
+     */
+    def snapshot: F[Map[K, V]]
+  }
+
   /**
    * Creates a sharded map ref to reduce atomic contention on the Map, given an efficient and
    * equally distributed hash, the contention should allow for interaction like a general
@@ -54,7 +62,7 @@ object MapRef extends MapRefCompanionPlatform {
    */
   def ofShardedImmutableMap[F[_]: Concurrent, K, V](
       shardCount: Int
-  ): F[MapRef[F, K, Option[V]]] = {
+  ): F[MapRef[F, K, Option[V]] with Snapshot[F, K, V]] = {
     assert(shardCount >= 1, "MapRef.sharded should have at least 1 shard")
     List
       .fill(shardCount)(())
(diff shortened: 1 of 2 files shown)


std/shared/src/main/scala/cats/effect/std/MapRef.scala
```diff
@@ -54,7 +62,7 @@ object MapRef extends MapRefCompanionPlatform {
    */
   def ofShardedImmutableMap[F[_]: Concurrent, K, V](
       shardCount: Int
-  ): F[MapRef[F, K, Option[V]]] = {
+  ): F[MapRef[F, K, Option[V]] with Snapshot[F, K, V]] = {
```
iRevive, line 65: We can add an alias type `type WithSnapshot[F, K, V] = MapRef[F, K, Option[V]] with Snapshot[F, K V]` and use it instead.

std/shared/src/main/scala/cats/effect/std/MapRef.scala
```diff
@@ -84,16 +92,41 @@ object MapRef extends MapRefCompanionPlatform {
...
+    val refFunction = { (k: K) =>
+      val location = Math.abs(k.## % shardCount)
+      array(location)
+    }
+    new MapRef[F, K, Option[V]] with Snapshot[F, K, V] {
+      def apply(k: K): Ref[F, Option[V]] =
+        fromSingleImmutableMapRef(refFunction(k)).apply(k)
+
+      def snapshot: F[Map[K, V]] =
+        seq.traverse(_.get).map(_.flatten.toMap)
```
durban, line 109: Unfortunately this can return a `Map` which is not really a snapshot (in case of concurrent modifications).

std/shared/src/main/scala/cats/effect/std/MapRef.scala
```diff
@@ -237,6 +277,12 @@ object MapRef extends MapRefCompanionPlatform {
     }
 
     def apply(k: K): Ref[F, Option[V]] = new HandleRef(k)
+
+    def snapshot: F[Map[K, V]] = sync.delay {
+      val builder = Map.newBuilder[K, V]
+      chm.forEach((k, v) => builder.addOne((k, v)))
+      builder.result()
```
durban, line 284: Similarly, I believe this is not a consistent view of the CHM.

std/shared/src/main/scala/cats/effect/std/MapRef.scala
iRevive: Hm, that's a valid concern. What if we rebrand `snapshot` into `entriesUnsafe` or any other more suitable term?

durban: I'm not sure... what's the expected use case here?

(If I'm creating the `MapRef`, I can already do this.... in fact, I can do better: I can create it from a `TrieMap`, which *has* atomic snapshot functionality.)

iRevive: > If I'm creating the MapRef, I can already do this

I didn't think of this scenario, to be honest. This works, indeed:
```scala
for {
  chm <- Sync[F].delay(new java.util.concurrent.ConcurrentHashMap[K, V]())
  mapRef = MapRef.fromConcurrentHashMap[F, K, V](chm)
} yield ...
```

There is a MapRef `mapRef` and an option to retrieve all values directly from `chm`. 
It's good enough for me, and I will close this PR.

PR #4153, opened 2024-10-21
Remove GraalVM from CI matrix

mzuehlke:
GraalVM is no longer available under the free license: https://github.com/graalvm/setup-graalvm#notes-on-oracle-graalvm-for-jdk-17

sbt-typelevel already did the same: https://github.com/typelevel/sbt-typelevel/pull/763

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -137,9 +137,8 @@ val LTSJava = JavaSpec.temurin("11")
 val LatestJava = JavaSpec.temurin("17")
 val ScalaJSJava = OldGuardJava
 val ScalaNativeJava = OldGuardJava
-val GraalVM = JavaSpec.graalvm("17.0.12")
 
-ThisBuild / githubWorkflowJavaVersions := Seq(OldGuardJava, LTSJava, LatestJava, GraalVM)
+ThisBuild / githubWorkflowJavaVersions := Seq(OldGuardJava, LTSJava, LatestJava)
 ThisBuild / githubWorkflowOSes := Seq(PrimaryOS, Windows, MacOS)
 
 ThisBuild / githubWorkflowBuildPreamble ++= Seq(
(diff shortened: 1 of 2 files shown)


mzuehlke:
I guess the Scala versions in this branch are way too old

armanbilge:
Oh darn, actually we can't use GraalVM 21 here at all, because this branch is pinned to Scala 3.2 and we can't do anything about that.

mzuehlke:
So dropping GraalVM from this branch seems to be the only option, or follow the not recommended path https://github.com/graalvm/setup-graalvm#notes-on-oracle-graalvm-for-jdk-17 and use exactly java-version: '17.0.12'

armanbilge:
> use exactly java-version: '17.0.12'

Oh, good idea. Let's do that. This branch just needs to limp a little longer.

mzuehlke:
The way `sbt-typelevel` uses the version number leads to an invalid workflow file  😢 

> **[Invalid workflow file: .github/workflows/ci.yml#L206](https://github.com/typelevel/cats-effect/actions/runs/11454014522/workflow)**
The workflow is not valid. .github/workflows/ci.yml (Line: 206, Col: 20): Unexpected symbol: '0.12.outputs.jdkFile'. Located at position 32 within expression: steps.download-java-graalvm-17.0.12.outputs.jdkFile .github/workflows/ci.yml (Line: 404, Col: 20): Unexpected symbol: '0.12.outputs.jdkFile'. Located at position 32 within expression: steps.download-java-graalvm-17.0.12.outputs.jdkFile

mzuehlke:
I would propose to remove GraalVM from the build matrix then.

armanbilge:
> The way `sbt-typelevel` uses the version number leads to an invalid workflow file 😢

Crap. We should fix this bug in sbt-tl.

That probably won't help us here though. We might just have to remove GraalVM.

djspiewak:
Removing GraalVM from 3.5 feels like the right short-term strategy. We really need to get 3.6 out the door.

djspiewak:
Note to self: when we do our next 3.5 release, we need to make sure to revert this commit specifically.

djspiewak (approved):

mzuehlke:
> Note to self: when we do our next 3.5 release, we need to make sure to revert this commit specifically.

What do you want to revert ? The problem is that you can no longer download GraalVM 17 without a paid license.
Specifying the latest patch version that is still available isn't working with `sbt-typelevel-github-action`.
And excape would be to "fix the GHA `ci.yml` by hand and disable the usage of `githubWorkflowCheck`.

build.sbt
```diff
@@ -194,12 +188,6 @@ ThisBuild / githubWorkflowBuild := Seq("JVM", "JS", "Native").map { platform =>
     name = Some("Test Example JavaScript App Using Node"),
     cond = Some(s"matrix.ci == 'ciJS' && matrix.os == '$PrimaryOS'")
   ),
-  WorkflowStep.Sbt(
-    List("graalVMExample/nativeImage", "graalVMExample/nativeImageRun"),
-    name = Some("Test GraalVM Native Image"),
-    cond = Some(
-      s"matrix.scala == '$Scala213' && matrix.java == '${GraalVM.render}' && matrix.os == '$PrimaryOS'")
-  ),
```
armanbilge, line 202: > What do you want to revert ?

We should have restored this in `series/3.6.x` branch 😅

PR #4160, opened 2024-10-28
Fix `joinWithNever` docs

### LINKED ISSUE #4093: Potential inconsistency between joinWithNever and docs

lenguyenthanh:
Close #4093

content is copied/pasted from [scala doc](https://github.com/typelevel/cats-effect/blob/ecf93db298c7fdb33df3be96dbe24e8f9bf90b9c/kernel/shared/src/main/scala/cats/effect/kernel/Fiber.scala#L71)

### DIFF
--- a/docs/typeclasses/spawn.md
+++ b/docs/typeclasses/spawn.md
@@ -231,7 +231,7 @@ In English, the semantics of this are as follows:
 
 - If the child fiber completed successfully, produce its result
 - If it errored, re-raise the error within the current fiber
-- If it canceled, the caller is indefinitely suspended without termination(a.k.a **deadlock**)
+- If it canceled, the caller is indefinitely suspended without termination (a.k.a. **deadlock**)
 
 Sometimes this is an appropriate semantic, and the cautiously-verbose `joinWithNever` function implements it for you. It is worth noting that this semantic was the *default* in Cats Effect 2 (and in fact, no other semantic was possible).
 
--- a/docs/typeclasses/spawn.md
+++ b/docs/typeclasses/spawn.md
@@ -231,7 +231,7 @@ In English, the semantics of this are as follows:
 
 - If the child fiber completed successfully, produce its result
 - If it errored, re-raise the error within the current fiber
-- If it canceled, attempt to self-cancel, and if the self-cancelation fails, **deadlock**
+- If it canceled, the caller is indefinitely suspended without termination(a.k.a **deadlock**)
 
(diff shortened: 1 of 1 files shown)


docs/typeclasses/spawn.md
```diff
@@ -231,7 +231,7 @@ In English, the semantics of this are as follows:
 
 - If the child fiber completed successfully, produce its result
 - If it errored, re-raise the error within the current fiber
-- If it canceled, attempt to self-cancel, and if the self-cancelation fails, **deadlock**
+- If it canceled, the caller is indefinitely suspended without termination(a.k.a **deadlock**)
```
durban:
```suggestion
- If it canceled, the caller is indefinitely suspended without termination (a.k.a. **deadlock**)
```

durban (dismissed):

armanbilge:
Thanks! This can be targeted to `series/3.5.x`

lenguyenthanh:
mm, I think I did it wrong 😅

Edit: fixed it!

armanbilge (approved):

durban (approved):

lenguyenthanh:
thanks @durban, @armanbilge!

PR #4167, opened 2024-11-12
Propagate Java thread interruption in `Dispatcher#unsafeRunSync`

### LINKED ISSUE #4166: Propagate Java thread interruption in `unsafeRunSync` et al.
We want to handle it similarly to the `TimeoutException`

https://github.com/typelevel/cats-effect/blob/725a4cd25725d23c3aa2213c6a54e9feadb66478/std/jvm/src/main/scala/cats/effect/std/DispatcherPlatform.scala#L58-L65

kamilkloch:
Closes #4166.

### DIFF
--- a/std/jvm/src/main/scala/cats/effect/std/DispatcherPlatform.scala
+++ b/std/jvm/src/main/scala/cats/effect/std/DispatcherPlatform.scala
@@ -59,10 +59,7 @@ private[std] trait DispatcherPlatform[F[_]] { this: Dispatcher[F] =>
     val (fut, cancel) = unsafeToFutureCancelable(fa)
     try Await.result(fut, timeout)
     catch {
-      case t: TimeoutException =>
-        cancel()
-        throw t
-      case t: InterruptedException =>
+      case t @ (_: TimeoutException | _: InterruptedException) =>
         cancel()
         throw t
     }
--- a/std/jvm/src/main/scala/cats/effect/std/DispatcherPlatform.scala
+++ b/std/jvm/src/main/scala/cats/effect/std/DispatcherPlatform.scala
@@ -16,10 +16,9 @@
 
 package cats.effect.std
 
-import scala.concurrent.{Await, TimeoutException}
-import scala.concurrent.duration.Duration
-
 import java.util.concurrent.CompletableFuture
+import scala.concurrent.duration.Duration
+import scala.concurrent.{Await, TimeoutException}
 
 private[std] trait DispatcherPlatform[F[_]] { this: Dispatcher[F] =>
 
@@ -60,6 +59,9 @@ private[std] trait DispatcherPlatform[F[_]] { this: Dispatcher[F] =>
     try Await.result(fut, timeout)
     catch {
       case t: TimeoutException =>
+        cancel()
+        throw t
+      case t: InterruptedException =>
         cancel()
         throw t
     }
--- a/std/jvm/src/main/scala/cats/effect/std/DispatcherPlatform.scala
+++ b/std/jvm/src/main/scala/cats/effect/std/DispatcherPlatform.scala
@@ -16,9 +16,10 @@
 
 package cats.effect.std
 
-import java.util.concurrent.CompletableFuture
-import scala.concurrent.duration.Duration
 import scala.concurrent.{Await, TimeoutException}
+import scala.concurrent.duration.Duration
+
+import java.util.concurrent.CompletableFuture
 
 private[std] trait DispatcherPlatform[F[_]] { this: Dispatcher[F] =>
 
--- a/std/jvm/src/main/scala/cats/effect/std/DispatcherPlatform.scala
+++ b/std/jvm/src/main/scala/cats/effect/std/DispatcherPlatform.scala
@@ -44,16 +44,17 @@ private[std] trait DispatcherPlatform[F[_]] { this: Dispatcher[F] =>
   }
 
   /**
-   * Submits an effect to be executed and indefinitely blocks until a result is produced. This
-   * function will throw an exception if the submitted effect terminates with an error.
+   * Submits an effect to be executed and indefinitely blocks until a result is produced.
+   * Cancels the effect in case of Java thread interruption. This function will throw an
+   * exception if the submitted effect terminates with an error.
    */
   def unsafeRunSync[A](fa: F[A]): A =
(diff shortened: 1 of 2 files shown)


std/jvm/src/main/scala/cats/effect/std/DispatcherPlatform.scala
```diff
@@ -62,6 +62,9 @@ private[std] trait DispatcherPlatform[F[_]] { this: Dispatcher[F] =>
       case t: TimeoutException =>
         cancel()
         throw t
+      case t: InterruptedException =>
+        cancel()
+        throw t
```
armanbilge: Does this syntax work?

```suggestion
      case t: TimeoutException | InterruptedException =>
        cancel()
        throw t
```

tests/jvm/src/test/scala/cats/effect/std/DispatcherJVMSpec.scala
```diff
@@ -42,5 +44,26 @@ class DispatcherJVMSpec extends BaseSpec {
         }
       } yield ok
     }
+
+    "Propagate Java thread interruption in unsafeRunSync" in real {
+      val res = Dispatcher.parallel[IO](await = true).evalMap { dispatcher =>
```
armanbilge:
```suggestion
      val res = Dispatcher.parallel[IO](await = true).use { dispatcher =>
```

tests/jvm/src/test/scala/cats/effect/std/DispatcherJVMSpec.scala
```diff
@@ -42,5 +44,26 @@ class DispatcherJVMSpec extends BaseSpec {
...
+        for {
+          canceled <- Deferred[IO, Unit]
+          io = IO.sleep(1.second).onCancel(canceled.complete(()).void)
+          thread = new Thread(() =>
+            try dispatcher.unsafeRunSync(io)
+            catch { case _: InterruptedException => })
+          _ <- IO(thread.start).as(thread)
+          _ <- IO.sleep(100.millis)
+          _ <- IO(thread.interrupt())
+          _ <- IO(thread.join())
```
armanbilge:
```suggestion
          _ <- IO.blocking(thread.join())
```

tests/jvm/src/test/scala/cats/effect/std/DispatcherJVMSpec.scala
```diff
@@ -42,5 +44,26 @@ class DispatcherJVMSpec extends BaseSpec {
...
+
+    "Propagate Java thread interruption in unsafeRunSync" in real {
+      val res = Dispatcher.parallel[IO](await = true).evalMap { dispatcher =>
+        for {
+          canceled <- Deferred[IO, Unit]
+          io = IO.sleep(1.second).onCancel(canceled.complete(()).void)
+          thread = new Thread(() =>
+            try dispatcher.unsafeRunSync(io)
+            catch { case _: InterruptedException => })
+          _ <- IO(thread.start).as(thread)
```
armanbilge:
```suggestion
          _ <- IO(thread.start)
```

tests/jvm/src/test/scala/cats/effect/std/DispatcherJVMSpec.scala
```diff
@@ -42,5 +44,26 @@ class DispatcherJVMSpec extends BaseSpec {
...
     }
+
+    "Propagate Java thread interruption in unsafeRunSync" in real {
+      val res = Dispatcher.parallel[IO](await = true).evalMap { dispatcher =>
+        for {
+          canceled <- Deferred[IO, Unit]
+          io = IO.sleep(1.second).onCancel(canceled.complete(()).void)
+          thread = new Thread(() =>
+            try dispatcher.unsafeRunSync(io)
+            catch { case _: InterruptedException => })
```
armanbilge: Instead of using `new Thread` what if we wrote this test with `IO.interruptible`?

armanbilge (commented):
We should also make a similar change for `IO#unsafeRunSync`.

std/jvm/src/main/scala/cats/effect/std/DispatcherPlatform.scala
kamilkloch: Unfortunately, no.

kamilkloch:
> We should also make a similar change for `IO#unsafeRunSync`.

I got a bit lost trying to comprehend the nuanced https://github.com/typelevel/cats-effect/blob/725a4cd25725d23c3aa2213c6a54e9feadb66478/core/jvm/src/main/scala/cats/effect/IOPlatform.scala#L70

std/jvm/src/main/scala/cats/effect/std/DispatcherPlatform.scala
armanbilge: It's a bit more sophisticated 😉 thanks for trying.

```suggestion
      case t @ (_: TimeoutException | _: InterruptedException) =>
        cancel()
        throw t
```

armanbilge:
> I got a bit lost trying to comprehend the nuanced

Actually, maybe let's hold off on that change. I forgot that the specification for this method is a bit unusual.

https://github.com/typelevel/cats-effect/blob/725a4cd25725d23c3aa2213c6a54e9feadb66478/core/jvm/src/main/scala/cats/effect/IOPlatform.scala#L45-L48

---

But if we did decide to move forward ...

<details>

You can factor out a package-private version of `unsafeRunAsync` that returns the `Fiber`.

https://github.com/typelevel/cats-effect/blob/725a4cd25725d23c3aa2213c6a54e9feadb66478/core/shared/src/main/scala/cats/effect/IO.scala#L979-L992

Then, you can use that for cancelation in `unsafeRunTimed`.

https://github.com/typelevel/cats-effect/blob/725a4cd25725d23c3aa2213c6a54e9feadb66478/core/jvm/src/main/scala/cats/effect/IOPlatform.scala#L79-L85
</details>

kamilkloch:
> Actually, maybe let's hold off on that change. I forgot that the specification for this method is a bit unusual.

If only you are ok with the resulting discrepancy in sematics between `Dispatcher#unsafeRunSync` and `IO#unsafeRunSync`, I would hold off as well. (Actually, for `unsafeRunTimed` the semantics is already different)

armanbilge (approved):
thanks!

djspiewak (approved):

PR #4168, opened 2024-11-13
Only remove error callback if we've added it

durban:
This was probably just an oversight in #2869. It's probably not a big deal, but also, why do unnecessary work...

FIXME: is "enhancement" the correct label for this? Or should it be "backstage"?

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -1012,15 +1012,21 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
       oc =>
         oc.fold(
           {
-            runtime.fiberErrorCbs.remove(failure)
+            if (registerCallback) {
+              runtime.fiberErrorCbs.remove(failure)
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -1012,15 +1012,21 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
       oc =>
         oc.fold(
           {
-            runtime.fiberErrorCbs.remove(failure)
+            if (registerCallback) {
+              runtime.fiberErrorCbs.remove(failure)
+            }
```
armanbilge: Instead of repeating three times can we factor this out before/after the `oc.fold` ?

durban: Yeah, that makes sense.

armanbilge (approved):
You could say it's an "enhancement" b/c it's an optimization. The exact label doesn't matter too much imo :)

Issue #4170, opened 2024-11-17
`Dequeue#size` drops below 0 on concurrent `takeFront`, `offerBack`, `size`


kamilkloch:
```scala
import cats.effect.std.Dequeue
import cats.effect.{IO, IOApp}

import scala.concurrent.duration.DurationInt

object DequeueTests extends IOApp.Simple {

  def run: IO[Unit] = {
    Dequeue.bounded[IO, Int](1024).flatMap { q =>
      val supplier = q.offerBack(1).debug("offerBack").delayBy(200.millis)
      val consumer = q.takeFront.debug("takeFront")
      val checkSize = q.size.debug("size").flatMap(n => if (n < 0) IO.raiseError(new Exception(s"q.size = $n")) else IO.unit)

      consumer.background.surround(supplier >> checkSize)
    }
  }
}
```

Output:
```
offerBack: Succeeded: ()
takeFront: Succeeded: 1
size: Succeeded: -1
java.lang.Exception: q.size = -1
```
(Run on CE 3.6-ecf93db)

reardonj:
I am very suspicious of this line: https://github.com/typelevel/cats-effect/blob/2d89990e15ac0ee72bed2900fab09fcbd28ed3bf/std/shared/src/main/scala/cats/effect/std/Dequeue.scala#L161

if I understand correctly, when there is a pending taker when an offer happens, the state does not increment the size. However, when `_take` ultimately gets called, it does decrement the size in the state. So when offer is called when a taker is pending, size gets decremented by 1 once both effects have finished executing, instead of getting incremented then decremented.

durban:
Was fixed by #4171.

PR #4171, opened 2024-11-18
Fix `size` updates in `Dequeue` state machine

armanbilge:
Fixes https://github.com/typelevel/cats-effect/issues/4170, h/t @reardonj for identifying the problematic logic.

### DIFF
--- a/tests/shared/src/test/scala/cats/effect/std/QueueSpec.scala
+++ b/tests/shared/src/test/scala/cats/effect/std/QueueSpec.scala
@@ -965,5 +965,11 @@ trait QueueTests[Q[_[_], _]] { self: BaseSpec =>
         r <- IO(v2 must beEqualTo(2))
       } yield r
     }
+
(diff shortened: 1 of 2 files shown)


tests/shared/src/test/scala/cats/effect/std/QueueSpec.scala
reardonj, line 1: Does this actually test `Dequeue` somehow? This looks like tests for `Queue`. I see a separate `DequeueSpec`.

armanbilge: Yes, `DequeueSpec` inherits tests from `QueueSpec`. FWIW I was testing with this sbt command:
```
testsJVM/testOnly *.BoundedDequeueSpec -- ex precedes
```

reardonj: ahhh!

samspills (approved):

durban (approved):

PR #4180, opened 2024-11-22
Removed implicit modifiers on deprecated instance to avoid unavoidable warnings

djspiewak:
This should fix HEAD builds.

### DIFF
--- a/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
+++ b/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
@@ -1302,9 +1302,7 @@ private[effect] trait ResourceHOInstances3 extends ResourceHOInstances4 {
 }
 
 private[effect] trait ResourceHOInstances4 extends ResourceHOInstances5 {
-  @deprecated(
-    "Use catsEffectMonadCancelForResource",
-    "3.6.0")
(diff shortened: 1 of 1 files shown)


armanbilge (changes requested):
I'm not sure if this is the change we want to make. The reason we left it as an implicit is so that this instance is still available in contexts that don't have `MonadCancelThrow` available (instead of an implicit-not-found compilation error). If you receive the deprecation message, the correct fix is to bring `MonadCancelThrow` into scope.

armanbilge (dismissed):
@lenguyenthanh tried a different fix in 3f79b9d0dc589fbfa14b106acb1ba406709ca2ad (thanks!) but we couldn't get it to work. So I think this is the best option to fix things.

kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala
```diff
@@ -1305,8 +1305,8 @@ private[effect] trait ResourceHOInstances4 extends ResourceHOInstances5 {
   @deprecated(
     "Bring an implicit MonadCancelThrow[F] into scope to get the fixed Resource instance",
```
armanbilge: Maybe we can slightly revise this.

```suggestion
    "Use catsEffectMonadCancelForResource",
```

armanbilge (approved):

PR #4186, opened 2024-11-27
Use global counter as an MBean ID for WSTP

iRevive:
I cherry-picked the ID idea from #3813 since it was pushed to 3.7. 

> I guess the hash helps to make it unique, but it actually doesn't guarantee that. I wonder if instead we can use a global AtomicLong counter. That is guaranteed to be unique, and in the happy path (one global runtime) it will always be 0 which should be nice enough for easy access. WDYT?

It should be easier to use with MBeans and metrics, too.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
@@ -104,11 +104,11 @@ private[unsafe] abstract class IORuntimeCompanionPlatform { this: IORuntime.type
         if (mBeanServer ne null) {
           val registeredMBeans = mutable.Set.empty[ObjectName]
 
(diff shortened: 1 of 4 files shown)


core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
```diff
@@ -125,7 +125,7 @@ private[unsafe] abstract class IORuntimeCompanionPlatform { this: IORuntime.type
 
             try {
               val localQueueSamplerName = new ObjectName(
-                s"cats.effect.unsafe.metrics:type=LocalQueueSampler-$hash-$i")
+                s"cats.effect.unsafe.metrics:type=LocalQueueSampler-$threadPoolId-$i")
```
iRevive, line 128: Unlike #3813, there is no global counter for queues. Pool ID + queue index should be a unique combination.

djspiewak (approved):

PR #4187, opened 2024-11-27
Add `TimerHeapMetrics`

iRevive:
A follow-up for https://github.com/typelevel/cats-effect/pull/3317#issuecomment-2496305040.

I'm unfamiliar with the `TimerHeap` internals, so perhaps the direction needs to be corrected.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/TimerHeap.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/TimerHeap.scala
@@ -61,7 +61,7 @@ private final class TimerHeap extends AtomicInteger {
 
   // And this is how many of those externally canceled nodes were already removed.
   // We track this separately so we can increment on owner thread without overhead of the atomic.
-  private[this] var removedCanceledCounter = 0
+  private[this] var removedCanceledCounter: Int = 0
 
   // The index 0 is not used; the root is at index 1.
   // This is standard practice in binary heaps, to simplify arithmetics.
@@ -69,6 +69,13 @@ private final class TimerHeap extends AtomicInteger {
   private[this] var size: Int = 0
 
   private[this] val RightUnit = Right(())
+
+  // metrics
+  private[this] var totalScheduled: Long = 0L
+  private[this] var totalExecuted: Long = 0L
+  private[this] var totalCanceled: Long = 0L
+
+  private def incrementTotalCanceled() = totalCanceled += 1
 
   /**
    * only called by owner thread
@@ -81,8 +88,12 @@ private final class TimerHeap extends AtomicInteger {
       if (root.isDeleted()) { // DOA. Remove it and loop.
 
         removeAt(1)
-        if (root.isCanceled())
+        if (root.isCanceled()) {
           removedCanceledCounter += 1
+          totalCanceled += 1
+        } else {
+          totalExecuted += 1
+        }
 
         peekFirstTriggerTime() // loop
 
@@ -133,11 +144,20 @@ private final class TimerHeap extends AtomicInteger {
         heap(size) = null
         size -= 1
 
-        if (root.isCanceled())
+        if (root.isCanceled()) {
           removedCanceledCounter += 1
+          totalCanceled += 1
+        } else {
+          totalExecuted += 1
+        }
 
         val back = root.getAndClear()
-        if (rootExpired && (back ne null)) back else loop()
+        if (rootExpired && (back ne null)) {
+          totalExecuted += 1
+          back
+        } else {
+          loop()
+        }
       } else null
     } else null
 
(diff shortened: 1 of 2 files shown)


core/jvm/src/main/scala/cats/effect/unsafe/metrics/WorkStealingPoolMetrics.scala
```diff
@@ -201,6 +206,25 @@ object WorkStealingPoolMetrics {
...
+
+    /**
+     * The current number of the outstanding timers.
+     */
+    def outstandingTimers(): Int
+
+    /**
+     * Returns the next due to fire.
+     */
+    def nextTimerDue(): Option[Long]
```
armanbilge, line 260: We should figure out what this `Long` represents. Currently it is an absolute time in nano seconds relative to the (arbitrary) `nanoTime` offset. I am not sure how useful that is.

- Would it make sense to convert to a unix epoch time? I think we'd have to give up some precision.
- Or, we could return the number of nanoseconds until the next timer fires — so a relative time to the present, not absolute.

core/jvm/src/main/scala/cats/effect/unsafe/TimerHeap.scala
```diff
@@ -241,6 +241,20 @@ private final class TimerHeap extends AtomicInteger {
...
+  def outstandingTimers(): Int = size
+
+  /**
+   * Returns the next due to fire.
+   */
+  def nextTimerDue(): Option[Long] =
+    heap
+      .lift(1)
+      .filter(t => !t.isDeleted() && t.triggerTime != Long.MinValue)
+      .map(_.triggerTime)
```
armanbilge: Once we decide on the API I can help with the implementation. I assume that this method may be called from non-owning threads, in which case we need to write the implementation defensively against race conditions.

I also think we should just return the time at the root of the heap, even if it's deleted/canceled. Returning `None` is confusing when there are more active timers besides the root. Besides, the timer was active at some point, deletion/cancelation is a race condition against metrics retrieval, and it's still a lower bound on the actual next timer due. So I think it's still relevant statistically speaking.

armanbilge (commented):
thanks so much for starting this follow-up :)

iRevive:
From the OpenTelemetry perspective, I can think of the following metrics:
- `timerheap.scheduled.count` - total number of timers that have been scheduled (enqueued)
- `timerheap.executed.count` - total number of executed timers
- `timerheap.cancelled.count` - total number of canceled timers - `canceledCounter.get()` or `removedCanceledCounter`
- `timerheap.outstanding.count` - current number of pending timers - `size`
- `timerheap.next.execution.time` - the due of the next timer - `nextTimerDue `

We may need to introduce new counters to keep track of `scheduled` and `executed` timers.

core/jvm/src/main/scala/cats/effect/unsafe/metrics/WorkStealingPoolMetrics.scala
iRevive: > the number of nanoseconds until the next timer fires — so a relative time to the present, not absolute.

This option looks more versatile and suitable.

armanbilge:
@iRevive I just pushed changes to `TimerHeap` to expose the metrics you listed, passing the baton back to you 😁

core/jvm/src/main/scala/cats/effect/unsafe/TimerHeap.scala
```diff
@@ -241,6 +272,29 @@ private final class TimerHeap extends AtomicInteger {
...
+   * Returns the next due to fire.
+   */
+  def nextTimerDue(): Option[Long] = {
+    val root = heap(1)
+    if (root ne null) {
+      val now = System.nanoTime()
+      val when = root.triggerTime - now
+      Some(when)
+    } else None
+  }
```
armanbilge, line 296: Caveats:
1. May return a negative number if the next timer is overdue to fire.
2. May overflow if the next timer is too far in the future.

Thoughts?

iRevive: > May return a negative number if the next timer is overdue to fire.

It's ok, I guess. 

If the negative number prevails over time (each time we collect metrics), it means that timers are under heavy pressure, right?

> May overflow if the next timer is too far in the future.

We can use `math.multiplyExact` and return `None` in the case of overflow. But it's not ideal either, I think.

armanbilge: > If the negative number prevails over time (each time we collect metrics), it means that timers are under heavy pressure, right?

Well, not necessarily. Assuming that we are seeing the latest data, it could indicate the worker thread is overwhelmed by (long-running) tasks and not able to check/trigger timers frequently enough. It would be measuring essentially the same thing as the starvation checker.

However, `TimerHeap` is not a thread-safe data structure so it's also possible that we might be seeing stale data i.e. a timer that was already triggered on the owner thread, but the changes to the heap were not yet published to the thread that is querying metrics.

---

> We can use `math.multiplyExact` and return `None` in the case of overflow. But it's not ideal either, I think.

Actually we would need `subtractExact`. But I agree returning `None` is not ideal, but on the other hand this situation is pretty unusual. Basically it means the next timer is scheduled 292 years in the future or something like that. That's very nearly like having no timer scheduled I suppose ...

iRevive:
@armanbilge I added the discussed methods to the `TimerHeapMetrics` but didn't touch the overflow case.

Should we keep it as is, perhaps? And change the calculation logic if users complain?

armanbilge:
> but didn't touch the overflow case.

Actually, I think it is okay and will not overflow in practice. This is because there is already some special handling to prevent overflows when doing comparisons within the heap. After thinking about it more, I realize that it will also prevent overflows when computing this metric.

https://github.com/typelevel/cats-effect/blob/9ce05f2395ae642e0ae8594e5cac89ca70b6aff7/core/jvm/src/main/scala/cats/effect/unsafe/TimerHeap.scala#L394-L402

core/jvm/src/main/scala/cats/effect/unsafe/TimerHeap.scala
djspiewak: I think we can ignore the overflow case. :D Anyone scheduling timers that far in the future deserves whatever they get from the metrics. Also I'm pretty sure the implementation itself would probably handle it incorrectly *anyway* since, at some point prior-or-equal to the relative overflow value, it's going to overflow the comparison math and start looking like an expired timer rather than a far-future one. So you basically *can't* have a functional timer that far in the future.

Also the negative number feels completely fine to me. It's actually an interesting signal in and of itself (as you two discussed), modulo the publication variability Arman already noted.

**Edit**: Oh this was discussed more below. Ignore me tyty

djspiewak (approved):

Issue #2251, opened 2021-08-21
Sbt main process cancelation seems broken with timers


djspiewak:
#2250 illustrates a bit of an issue with the process cancelation in sbt when `fork := false`. For whatever reason, the fibers *don't* seem to be canceled entirely and the thread pools are not cleaned up. Needs verification and minimization.

armanbilge:
Is this still relevant now that we've integrated timers in v3.5.0?

Update: reproduced with v3.4.10 and v3.5.0-RC5. Basically running something like this via sbt, canceling it, and then getting the println right-on-schedule anyway 😅 

```scala
import cats.effect._
import scala.concurrent.duration._

object Test extends IOApp.Simple {
  def run = IO(println("huh")).delayBy(10.seconds)
}
```

armanbilge:
Seems that this isn't actually working—not sure how sbt is signaling cancelation, if at all 🤔 
https://github.com/typelevel/cats-effect/blob/c00f3d01c34f5cf2fc52614c18b627c8fda9bfad/core/jvm/src/main/scala/cats/effect/IOApp.scala#L528-L530

djspiewak:
There's a bug in sbt. I really should have written down the issue number the last time I rabbit holed on this, but sbt has genuinely stopped forwarding cancelation onto `run` tasks correctly. Basically, the line you just linked *used* to work in a prior version of Sbt, and it's supposed to still work, but it's dead code on current versions.

armanbilge:
I found:
- https://github.com/sbt/sbt/issues/6242

which seems to mention this exact problem, but you closed it as a duplicate of:
- https://github.com/sbt/sbt/issues/6192

which I think was just about the annoying stacktrace 🤔

Friendseeker:
> Is this still relevant now that we've integrated timers in v3.5.0?
> 
> Update: reproduced with v3.4.10 and v3.5.0-RC5. Basically running something like this via sbt, canceling it, and then getting the println right-on-schedule anyway 😅
> 
> ```scala
> import cats.effect._
> import scala.concurrent.duration._
> 
> object Test extends IOApp.Simple {
>   def run = IO(println("huh")).delayBy(10.seconds)
> }
> ```

Reproducible on sbt 1.0.1, not reproducible on sbt 1.0.0....

PR #4188, opened 2024-11-27
Update `onNonMainThreadDetected()` warning

Friendseeker:
Updates the warning to reflect the latest understanding of https://github.com/typelevel/cats-effect/issues/2251 from https://discord.com/channels/632277896739946517/632277897448652844/1310909184937824287

@djspiewak

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/IOApp.scala
+++ b/core/jvm/src/main/scala/cats/effect/IOApp.scala
@@ -350,7 +350,11 @@ trait IOApp {
           """|[WARNING] IOApp `main` is running on a thread other than the main thread.
              |This may prevent correct resource cleanup after `main` completes.
              |This condition could be caused by executing `run` in an interactive sbt session with `fork := false`.
-             |Set `Compile / run / fork := true` in this project to resolve this.
+             |To ensure proper cleanup, either
+             |  - set `Compile / run / fork := true` in this project
+             |  - use `fgRun` instead of `run`
+             |  - use 'bgStop` to terminiate `run`
+             |  - update sbt to a version after 1.10.5
              |
              |To silence this warning set the system property:
              |`-Dcats.effect.warnOnNonMainThreadDetected=false`.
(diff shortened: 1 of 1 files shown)


djspiewak (approved):
Is there any way we can detect the sbt version? We're in its JVM in this scenario, so in theory we might be able to putz around with its internals if need be, or at least read envars/properties.

core/jvm/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -350,7 +350,11 @@ trait IOApp {
           """|[WARNING] IOApp `main` is running on a thread other than the main thread.
              |This may prevent correct resource cleanup after `main` completes.
              |This condition could be caused by executing `run` in an interactive sbt session with `fork := false`.
-             |Set `Compile / run / fork := true` in this project to resolve this.
+             |To ensure proper cleanup, either
+             |  - set `Compile / run / fork := true` in this project
+             |  - use `fgRun` instead of `run`
+             |  - use 'bgStop` to terminate `run`
+             |  - update sbt to a version after 1.10.5
```
armanbilge, line 357: Thanks for the PR + in-depth investigation! Sorry if I missed it, what exactly is the change that's landing in sbt > 1.10.5 that fixes this?

Friendseeker: I also feel grateful for the support I received from you and Daniel during the investigation.

https://github.com/sbt/sbt/pull/7916 is intended to fix it.

The PR is not merged yet, but I did test locally that it makes ctrl-c interrupts properly and the review process seems to be smooth sailing so far.

Friendseeker:
> Is there any way we can detect the sbt version? We're in its JVM in this scenario, so in theory we might be able to putz around with its internals if need be, or at least read envars/properties.

I gave it a look and while I believe it is possible, it is not entirely obvious how to implement it.

sbt does not seem to store sbt version in system properties (I tried `sbt.version`, it has nothing in it). Also sbt seems to have no global object storing version, so reflection to access some global object is no go.

I guess we can search for `build.properties` but at this point it involves File IO, which is also undesirable.

PR #4194, opened 2024-12-10
update `LocalQueue` docs/comments

biochimia:
I was reviewing the WorkStrealingThreadPool implementation, and updated some docs along the way, so proposing them in case they're useful to others.

I'm happy to iterate further on the changes and to drop any that seem misguided.


djspiewak (approved):
Tyty!

PR #4195, opened 2024-12-10
Blocked thread detection fixes

biochimia:
There are two small things here, both related to the blocked thread detection code:
- it was using the original thread index to check for itself, while the index will be updated after the thread converts to blocking. The fix here is to rely on the updated index, instead;
- when randomly picking another thread, it would include itself in the pool, and then pick again in case it picked itself. In this case the random choice was changed to exclude self from the choice pool.


djspiewak (approved):
Very nice cleanup

PR #4196, opened 2024-12-10
Use AtomicReferenceArray to keep track of worker threads 

biochimia:
For the most part the array of worker threads is read-only, however it may be updated from any thread when a worker thread transitions to blocking or back.

Previously, a separate atomic variable was used to introduce memory barriers when accessing the array. This is discarded in favour of `AtomicReferenceArray` that specifically manages access to elements of the array.

The `WorkStealingThreadPool.getWorkerThreads` method is replaced by a more limited `getWorkerThread(index: Int)` method. This is used by `WorkerThread` when checking for blocked threads, and the new method can be used to specifically access a single worker thread.

~*Note:* This PR includes the changes also proposed in #4195, but the intent is that they may be discussed separately. I can keep the two PRs in sync with regards to the common changes. In case the former PR is rejected I will drop the rejected changes here and otherwise update this one accordingly.~

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
@@ -42,7 +42,13 @@ import java.time.Instant
 import java.time.temporal.ChronoField
 import java.util.Comparator
 import java.util.concurrent.{ConcurrentSkipListSet, ThreadLocalRandom}
-import java.util.concurrent.atomic.{AtomicBoolean, AtomicInteger, AtomicLong, AtomicReference}
+import java.util.concurrent.atomic.{
+  AtomicBoolean,
+  AtomicInteger,
+  AtomicLong,
+  AtomicReference,
+  AtomicReferenceArray
+}
 
 import WorkStealingThreadPool._
 
@@ -84,7 +90,8 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
   /**
    * References to worker threads and their local queues.
    */
-  private[this] val workerThreads: Array[WorkerThread[P]] = new Array(threadCount)
+  private[this] val workerThreads: AtomicReferenceArray[WorkerThread[P]] =
+    new AtomicReferenceArray(threadCount)
   private[unsafe] val localQueues: Array[LocalQueue] = new Array(threadCount)
   private[unsafe] val sleepers: Array[TimerHeap] = new Array(threadCount)
   private[unsafe] val parkedSignals: Array[AtomicBoolean] = new Array(threadCount)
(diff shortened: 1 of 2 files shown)


djspiewak:
This needs some careful benchmarking! My strong hunch is that this is going to be a lot slower than the previous approach, but we'll see.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -173,22 +171,19 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
           poller,
           this)
 
-      workerThreads(i) = thread
+      workerThreads.set(i, thread)
```
biochimia, line 174: Perhaps this could use `lazySet`.

djspiewak:
This is very surprising to me. Arm chair reasoning really suggests that this should be a significant performance hit, but the benchmarks say it's a wash. Will think a bit more.

## Before

```
[info] Benchmark                                             (cpuTokens)   (size)   Mode  Cnt     Score    Error    Units
[info] ParallelBenchmark.parTraverse                               10000     1000  thrpt   10   724.626 ±  1.581    ops/s
[info] ParallelBenchmark.traverse                                  10000     1000  thrpt   10    69.191 ±  0.095    ops/s
[info] WorkStealingBenchmark.alloc                                   N/A  1000000  thrpt   10    11.697 ±  0.121  ops/min
[info] WorkStealingBenchmark.manyThreadsSchedulingBenchmark          N/A  1000000  thrpt   10    28.175 ±  6.114  ops/min
[info] WorkStealingBenchmark.runnableScheduling                      N/A  1000000  thrpt   10  2399.482 ± 51.270  ops/min
[info] WorkStealingBenchmark.runnableSchedulingScalaGlobal           N/A  1000000  thrpt   10  1886.390 ±  5.104  ops/min
[info] WorkStealingBenchmark.scheduling                              N/A  1000000  thrpt   10    22.989 ±  3.496  ops/min
```

## After

```
[info] Benchmark                                             (cpuTokens)   (size)   Mode  Cnt     Score   Error    Units
[info] ParallelBenchmark.parTraverse                               10000     1000  thrpt   10   722.919 ± 1.127    ops/s
[info] ParallelBenchmark.traverse                                  10000     1000  thrpt   10    69.215 ± 0.234    ops/s
[info] WorkStealingBenchmark.alloc                                   N/A  1000000  thrpt   10    11.858 ± 0.119  ops/min
[info] WorkStealingBenchmark.manyThreadsSchedulingBenchmark          N/A  1000000  thrpt   10    28.503 ± 7.477  ops/min
[info] WorkStealingBenchmark.runnableScheduling                      N/A  1000000  thrpt   10  2518.366 ± 4.768  ops/min
[info] WorkStealingBenchmark.runnableSchedulingScalaGlobal           N/A  1000000  thrpt   10  1884.161 ± 5.135  ops/min
[info] WorkStealingBenchmark.scheduling                              N/A  1000000  thrpt   10    25.316 ± 3.531  ops/min
```

vasilmkd:
Given that the benchmarks are not showing any obvious regressions, the code is much simpler and clearer. Why not. 👍🏻

biochimia (commented):
In terms of cost, the biggest impact I'd expect would be during initialization and shutdown, as we trade a single memory barrier for one per worker thread. On the other hand, the cost of starting and joining threads probably hides that. Either way, as one time operations, I would not expect this to show up in general benchmarks.

Both `notifyParked` and `replaceWorker` should have similar costs, with a single barrier as before.

In the case of the blocked thread detection, a single barrier is introduced and I think it should be there, for correctness.

Then there is `liveTrace` and `getSuspendedFiberCount`, which again pay a higher cost. The former operation seems to be rather expensive on its own. I think that neither operation should be on the critical path for overall performance.

That was my assessment, at least.

djspiewak (approved):
Yeah staring at this some more, I get the reasoning. Okay let's go for it!

Also holy crap a @vasilmkd sighting! 🎉

PR #4197, opened 2024-12-10
use worker thread index at startup

biochimia:
Naming of active worker threads was inconsistent: when initially created, worker threads were numbered by the pool's `blockedWorkerThreadNamingIndex` in constructor code; if they later switched back from blocking state they would use the worker thread index.

~`nameIndex` is renamed to `blockedWorkerIndex`, to make explicit the link to the counter that originates it. This is still used when naming a blocked worker, as well as for ranking `cachedThreads` in the pool.~ (edit: Rename dropped to avoid binary compatibility breakage)

With this change, the (current) worker index is used for naming both newly constructed worker threads, and threads that switch back from blocking.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
@@ -115,6 +115,7 @@ private[effect] final class WorkerThread[P <: AnyRef](
     def run() = ()
   }
 
+  // Identifies blocked worker, used for ranking cached threads
   val nameIndex: Int = pool.blockedWorkerThreadNamingIndex.getAndIncrement()
 
   // Constructor code.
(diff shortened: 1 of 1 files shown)


core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -115,7 +115,8 @@ private[effect] final class WorkerThread[P <: AnyRef](
     def run() = ()
   }
 
-  val nameIndex: Int = pool.blockedWorkerThreadNamingIndex.getAndIncrement()
+  // Identifies blocked worker, used for ranking cached threads
+  val blockedWorkerIndex: Int = pool.blockedWorkerThreadNamingIndex.getAndIncrement()
```
biochimia: The variable rename spuriously breaks binary compatibility, so I'll drop it and keep only the added comment.

djspiewak (approved):
Hopefully this should make things a bit less confusing for people looking at thread dumps over time!

PR #4200, opened 2024-12-13
Fallback `IOLocal#unsafeThreadLocal` by a true `ThreadLocal`

armanbilge:
tl;dr is that for the `IOLocalContextStorage` in https://github.com/typelevel/otel4s/pull/214 to work for arbitrary Java libraries, it must properly implement the `ContextStorage` API even when used in a context outside of `IOFiber`. To accomplish this, we can make `ThreadLocal`s returned by `IOLocal#unsafeThreadLocal` behave as a true `ThreadLocal` when used outside of an `IOFiber`. This is not currently possible to implement in userland because there is no way to know if you are currently running in a fiber or not.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/IOLocalPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/IOLocalPlatform.scala
@@ -21,23 +21,32 @@ import IOFiberConstants.ioLocalPropagation
 private[effect] trait IOLocalPlatform[A] { self: IOLocal[A] =>
 
   /**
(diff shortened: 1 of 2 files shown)


djspiewak:
This is pretty evil. :( Also remember the only way you could get an `IOLocal` *in the first place* in this type of situation is if you created one inside of `IO` or `unsafe`ly and then yolo'd it out into thread land.

armanbilge:
We decided it's too evil.

> because there is no way to know if you are currently running in a fiber or not.

We are going to do this instead.

Issue #3837, opened 2023-09-18
Enable fiber dumps on Scala Native


armanbilge:
Everything is already implemented and they are working correctly AFAIK.

I made an attempt to re-enable them in https://github.com/typelevel/cats-effect/pull/3809 but we were hit by out-of-memory problems in CI.

I'm not sure if we are just stressing CI and/or Scala Native GC too much in our test suite ... or if there is still a legitimate bug with `WeakReference`s that prevents them from being collected in a timely manner.

armanbilge:
We tried again to re-enable fiber dumps in https://github.com/typelevel/cats-effect/pull/4201 and they were still causing excessive memory use :(

djspiewak:
Have we filed a bug so that @WojciechMazur tron can activate?

djspiewak:
I'm going to remove this from the milestone since it seems like a longer pole and we probably need upstream fixes. I'd rather this doesn't block 3.7

PR #4201, opened 2024-12-15
Add support for Native multithreaded execution

durban:
This is on top of @djspiewak's `wip/multithreaded-wstp` branch. ~~I did nothing so far, except default to `SleepSystem` in 489fbb9db361e6edaa6a4b56240044824588218b. This makes it possible to run `IOSpec` on scala-native; it (sometimes) passes on my machine ;-). It's probably too early to have this as a PR, but I'm doing it anyway to avoid duplicating work, and maybe to have a discussion about `EpollSystem` (see below). (@djspiewak feel free to close this PR if you have other plans with your branch.)~~


durban:
So, I'm running this on Linux, and without 489fbb9db361e6edaa6a4b56240044824588218b it tries to use `EpollSystem`. But starting the tests (e.g., `testsNative/testOnly cats.effect.IOSpec`) only leads to a hanging process. It seems to me, that the WSTP threads are waiting in `epoll_wait`, and a (the?) GC thread seems to wait for mutator threads to reach a safepoint(?). (Details: the GC is in `thread_yield`, called by `Synchronizer_acquire`.) I'm speculating here, but maybe `Thread.sleep()` does "something", which `EpollSystem` doesn't, before `epoll_wait`ing?

~~EDIT: what I wrote above is probably completely wrong...~~ :man_shrugging:

djspiewak:
This is great! Thank you for pushing this forward to the next obstacle.

One of the things that occurred to me as I poked at my branch originally is SN's tooling for introspecting thread state is really really limited so far as I understand it. Maybe this is just my ignorance and LLVM has some magic we could turn on, but I strongly suspect we're going to need better introspection to run down some of these problems.

What I'm thinking is we're probably going to end up building that, or at least leaning in heavily to do so, and that's probably a large part of what we'll need to do to get this off the ground. We should chat with the SN folks.

armanbilge:
The reason it's hanging is because we haven't implemented interruption yet for the Native I/O-polling systems. This wasn't necessary when it was single-threaded, but now it's critical :)

https://github.com/typelevel/cats-effect/blob/489fbb9db361e6edaa6a4b56240044824588218b/core/native/src/main/scala/cats/effect/unsafe/EpollSystem.scala#L67

Compare with:
https://github.com/typelevel/cats-effect/blob/9ce05f2395ae642e0ae8594e5cac89ca70b6aff7/core/jvm/src/main/scala/cats/effect/unsafe/SelectorSystem.scala#L97-L100

armanbilge:
Oh, the other reason it may be hanging is indeed related to GC. On Scala Native, blocking native calls need to be annotated explicitly with the `@blocking` annotation, so that it does the necessary book-keeping so it's possible to GC while a thread is stuck in that blocking call.

https://github.com/scala-native/scala-native/blob/c7b54a18e3ff11d8b2792f16fbb6e97780314014/nativelib/src/main/scala/scala/scalanative/unsafe/package.scala#L103-L106

For now it's fine to just mark it `@blocking`, but b/c this comes at a performance cost, we should actually make two separate `epoll_wait` bindings. One will use `@blocking` for when the timeout is > 0, and the other will not, for when the timeout == 0.

djspiewak:
> The reason it's hanging is because we haven't implemented interruption yet for the Native I/O-polling systems. This wasn't necessary when it was single-threaded, but now it's critical :)

You know, I didn't even think about this. Makes loads of sense though. Pipes time!

durban:
@armanbilge Thanks, I've tried to do the 2 things you mentioned. In 62b81410f0 I turned on the `EpollSystem` again, and tried implementing `interrupt`, and added the scala-native `blocking` annotation. This way `testsNative/testOnly cats.effect.IOSpec` passes on my machine. It obviously needs more work (e.g., I think `interrupt` is not threadsafe), but at least it's a step in the right direction.

core/native/src/main/scala/cats/effect/unsafe/EpollSystem.scala
```diff
@@ -221,6 +251,15 @@ object EpollSystem extends PollingSystem {
 
     private[EpollSystem] def needsPoll(): Boolean = !handles.isEmpty()
 
+    private[EpollSystem] def interrupt(): Unit = {
+      val buf = stackalloc[CUnsignedInt]()
+      buf(0) = 1.toUInt
+      // TODO: this is not threadsafe, we're reading `interruptFd` without synchronization:
+      if (unistd.write(this.interruptFd, buf, 8.toCSize) == -1) {
```
armanbilge: It doesn't need to be, it will be synchronized by `workerThreadPublisher` at its callsites.

https://github.com/typelevel/cats-effect/blob/8d926512b22fc917449848435b120f2554a6ee69/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala#L324-L326

djspiewak:
Interesting. So I merged your branch with `series/3.x` and now I'm getting the following:

```
[error] Unknown DWARF abbrev code: 26
[error] 
[error] STACKTRACE
[error] 
[error] java.lang.RuntimeException: Unknown DWARF abbrev code: 26
[error] 
[error] 
[error] This looks like a specs2 exception...
[error] Please report it with the preceding stacktrace at http://github.com/etorreborre/specs2/issues
[error]  
[error] Error: Total 1, Failed 0, Errors 1, Passed 0
[error] Error during tests:
[error] 	cats.effect.IOSpec
```

**Edit**: Appears to be a macOS only thing. Compiles and runs on Linux. Lovely.

djspiewak:
Okay got around the issue with Lorenzo's help. It's fixed in SN **main**, so I updated to a local snapshot (lol) on my branch and made progress. I'll dig into interruption for kqueue

djspiewak:
Update:

1. We no longer need the snapshot. I've pushed the magic compiler settings incantation on my branch
2. Also pushed is an initial stab at using `EVFILT_USER` to handle kqueue interrupts. I implemented it by basically putting that event permanently into the front of the changes array and then triggering it on the kqfd on `interrupt()`. In principle, this makes sense, but it doesn't actually work. The threads wake up but things spin forever. I'm probably being dumb. Have fun.

Will get back to this later.

djspiewak:
I pushed more. Kqueue is pretty close to working I think.

djspiewak:
Update: I've got `testsNative/test` almost entirely green on macOS.

durban:
@djspiewak Yeah, I don't know anything about macos; I can't even help testing it (I have no access to such a system). Anyway, I've merged your branch into this PR, so that your progress is visible here too.

djspiewak:
I think this PR will become *the* PR, more than likely

djspiewak:
Oh, if you want to just focus on Linux + epoll, that works perfectly. There's also more stuff to go through and pull out of jvm-specific areas and into jvm-native areas (e.g. the `IOPlatformSpecs`, which are all intermixed and weird, or `interruptibleImpl`, etc). Lots of stuff strewn around.

durban:
Okay, I'll try to look at one of those when I'll have some time. (Btw, I think you can push directly to this PR; at least I've checked that github checkbox.)

djspiewak:
Alrighty, I'm going to deprecate my fork's branch then and we can centralize on yours.

djspiewak:
Oh btw as a planning process note… Once we can get something which is mergeable, we should do that, close out this PR, and use normal PR process from that point forward

djspiewak:
Btw I can't actually push to your branch. Pushing to mine for now.

djspiewak:
Status report:

I think I've got all (but one) of the tests working and all the functionality is shifted over aside from `interruptible`-related things and `IOAppSpec` (this is worth checking though just in case I missed something). The latter is dependent on the sbt yak shave (the `BuildInfo` stuff). The `interruptible` stuff just needs to be stubbed around basically since we don't have the right exception types in SN; fairly trivial. The one failing test is `"handle lots of simultaneous events"` in `FileDescriptorPollerSpec`, which is annoying since it probably actually means something. Everything else is passing.

I took the time to factor out `kevent64` into blocking and non-blocking variants. The pattern works pretty well so we should probably do something similar for epoll.

This still doesn't compile on Scala 3 because of the `epoll_event` tag shenanigans. That's going to take some more compiler sweet talking.

One thing that has come out as part of this is our conditional source matrix is getting pretty hairy. It might be worth taking a pass over things and updating our encodings to be what we want them to be (e.g. we have `Platform` traits, but some of them are cross-platform, and we also have the `isJVM`/`isNative`/etc checks, and sometimes we mix them all together for maximal confusion).

Anyway, apart from the FD weirdness, I think kqueue is pretty much ready to go. It could be optimized more but I'm not worried about performance yet. We're very close on this branch as a whole, tbh.

**Edit**: I've confirmed that Linux and macOS are now basically in the same state, which is reassuring. I actually wonder if something might be odd about that particular test rather than the underlying polling system or WSTP?

durban:
> Btw I can't actually push to your branch

Huh, that's strange. I double checked, and the "allow edits by maintainers" _is_ selected. I don't know what's going on. Anyway, I'm sure we'll be able to solve it with some combination of merge and push...

armanbilge:
> I've confirmed that Linux and macOS are now basically in the same state,

This is not my observation. 

- On Linux, the entire `FileDescriptorPollerSpec` is broken. There seems to be a segfault when registering the pipes with the epoll.
- On macOS, only the `"handle lots of simultaneous events"` is failing. It can segfault, or for smaller numbers simply hang and timeout.

djspiewak:
> There seems to be a segfault when registering the pipes with the epoll

Do segfaults manifest as fatal errors in sbt? If so then that's exactly what I'm experiencing, not a failing test.

djspiewak:
Oh I see what you mean. Hmm. I must have just gotten a segfault in both and assumed they were the same issue. Unfortunate.

durban:
I think I've fixed the segfault in `FileDescriptorPollerSpec` for `EpollSystem` (see f4bc66acde). I think the `stackalloc`s were somehow wrong: `type epoll_event <: AnyRef` makes scala-native think it's 16 bytes. While there is a `Tag[epoll_event]` with the correct size, `stackalloc` doesn't seem to use it(?). If I use the correct size manually, the segfault disappears. (I'm sure there is a nicer solution...)

durban:
Okay, so it seems that in SN 0.4 `stackalloc` takes an implicit `Tag`; probably that's why this used to work. In 0.5 it doesn't; but then how does it know the size? :man_shrugging:

djspiewak (commented):
We definitely need to figure out how to start sharing code across platforms more sanely. :P

armanbilge:
Latest from CI ... this is reminiscent of the issues we are having in Cats, where test suites just error out 😕 

- https://github.com/typelevel/cats/issues/4647

```
[error] Error: Total 2141, Failed 0, Errors 32, Passed 2108, Skipped 1, Pending 1
[error] Error during tests:
[error] 	cats.effect.unsafe.IORuntimeConfigSpec
[error] 	cats.effect.std.internal.BinomialHeapSpec
[error] 	cats.effect.ExitCodeSpec
[error] 	cats.effect.IOPropSpec
[error] 	cats.effect.std.DroppingQueueSpec
[error] 	cats.effect.std.MapRefSpec
[error] 	cats.effect.std.BackpressureSpec
[error] 	cats.effect.std.UnboundedDequeueSpec
[error] 	cats.effect.OptionTIOSpec
[error] 	cats.effect.tracing.TracingSpec
[error] 	cats.effect.std.UnboundedPQueueSpec
[error] 	cats.effect.std.SystemPropertiesSpec
[error] 	cats.effect.KleisliIOSpec
[error] 	cats.effect.std.DeferredParallelism2Spec
[error] 	cats.effect.std.DeferredParallelism4Spec
[error] 	cats.effect.unsafe.IORuntimeBuilderSpec
[error] 	cats.effect.kernel.DeferredSpec
[error] 	cats.effect.IOSpec
[error] 	not.cats.effect.IOParImplicitSpec
[error] 	cats.effect.std.BoundedQueueSpec
[error] 	cats.effect.ContSpec
[error] 	cats.effect.std.internal.BankersQueueSpec
[error] 	cats.effect.IOMtlLocalSpec
[error] 	cats.effect.std.MutexSpec
[error] 	cats.effect.kernel.ParallelFSpec
[error] 	cats.effect.FileDescriptorPollerSpec
[error] 	cats.effect.std.EnvSpec
[error] 	cats.effect.ThunkSpec
[error] 	cats.effect.std.ConsoleSpec
[error] 	cats.effect.DefaultContSpec
[error] 	cats.effect.testkit.TestControlSpec
```

armanbilge:
Locally is a bit more promising 🤔 

```
[error] Error: Total 2860, Failed 0, Errors 2, Passed 2857, Skipped 1, Pending 2
[error] Error during tests:
[error]         cats.effect.ContSpec
[error]         cats.effect.DefaultContSpec
```

djspiewak:
I saw that happen when any individual test segfaults: it seems to cause a whole cascade of ghost errors.

djspiewak:
I no longer see segfaults locally on mac. Haven't tried Linux yet. `stackalloc` is really fucky on SN 0.5 afaict, even for really simple cases. I don't think anyone should be using the no-args version for anything ever.

Oh as an aside, the problems I had pushing to this remote were being caused by git-lfs. In case anyone else sees these types of issues, the solution is to use the `--no-verify` flag with `git push`.

armanbilge:
> I no longer see segfaults locally on mac

I still do 😕 
```
sbt:cats-effect> testsNative/testOnly *.FileDescriptorPollerSpec
[info] Build skipped: No changes detected in build configuration and class path contents since last build.
[info] Starting process '/Users/arman/code/cats-effect/tests/native/target/scala-2.13/cats-effect-tests-test' on port '63437'.
[info] Starting process '/Users/arman/code/cats-effect/tests/native/target/scala-2.13/cats-effect-tests-test' on port '63439'.
[error] Test runner interrupted by fatal signal 11
[warn] Force close java.lang.RuntimeException: Process /Users/arman/code/cats-effect/tests/native/target/scala-2.13/cats-effect-tests-test finished with non-zero value 139 (0x8b)
[info] FileDescriptorPollerSpec
[info] FileDescriptorPoller should
[info]   + notify read-ready events
[error] Error: Total 2, Failed 0, Errors 1, Passed 1
[error] Error during tests:
[error] 	cats.effect.FileDescriptorPollerSpec
[error] (testsNative / Test / testOnly) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 5 s, completed Jan 3, 2025, 9:59:12 AM
sbt:cats-effect>
```

djspiewak:
Nuts. I wonder if I just didn't run it enough.

djspiewak:
@armanbilge Braaaaavo.

djspiewak:
@armanbilge In the interest of avoiding future conflicts, rather than `dc09477` I would recommend merging your PR branch which does the same change against `series/3.x` and then sorting out the conflict. We would need to resolve that conflict eventually anyway.

armanbilge:
> rather than `dc09477` I would recommend merging your PR branch which does the same change against `series/3.x` and then sorting out the conflict.

@djspiewak that's actually not possible, because that API was only added in this branch. The WSTP only exists on Native as of this PR :)

djspiewak:
We're getting very close here. The CI failures strike me as legitimate. I don't have access to x86_64 linux (or really anything) right now, though I'll spin up an EC2 instance if no one else gets to it first. For context, the test which timed out after 100s finishes in 2.1s on my laptop (macOS ARM).

But I think the mac issues are in the rearview mirror.

durban:
So far I haven't been able to locally reproduce the timeout seen in CI (`cats.effect.IOSuite.lots of externally-canceled timers`). (I've been trying on an x86_64 Linux.)

djspiewak:
@durban Same. I spun up an EC2 instance and haven't seen it fail even once. That specific takes like 7 seconds on a t3x.large. Really not sure what's going on here.

djspiewak:
Looks like everything was green on this commit(!!) except for ubuntu-latest, 2.12, ciNative, which timed out after an hour. I'm going to try booping it to see if it's transient.

djspiewak:
I have been so far unable to reproduce the long run time in the GHA runner. A quick one-off EC2 instance has been able to run `ciNative` repeatedly with Temurin 8 in about 20 minutes, which is of course much less than what we see here.

djspiewak:
Still can't reproduce it. This might be some sort of weird GHA-only thing. It's also very strange to me that we only see this with 2.12 and temurin@8

durban:
I can reproduce _something_... (I'm not sure if it's the same as in CI):

x86_64 linux, 2.12, `ciNative`, but on a JVM 17; `cats.effect.IOAppSuite` seemingly hangs, spinning all cores. It is apparently in the `GlobalRacingInit` test. The WSTP threads are doing tracing things(?), e.g., I see `TracingPlatform.calculateTracingEvent` in the stacktrace. I thought tracing have been disabled on native? (Although I don't know if that's the reason, or a red herring.) For reference, I'm on 55c465c89c.

durban:
Ok, I see, fiber monitoring was disabled, _not_ tracing.

durban:
... a few minutes later: yeah, this doesn't seem threadsafe at all: https://github.com/durban/cats-effect/blob/55c465c89c8cf9dcca4ace81eda528ad43a310ca/core/native/src/main/scala/cats/effect/tracing/TracingPlatform.scala#L27

djspiewak:
Nice catch! I completely forgot about tracing. I bet that's the exact issue.

durban:
I've pushed 7dab7b8 yesterday, and CI seems green... but I don't believe it, so I'll just restart it...

durban:
Yeah, it failed :-) (I would've been surprised if it didn't.)

It seems the same failure as before on ARM linux...

djspiewak:
Nice catch on the event tag!

I pulled up a Linux ARM64 VM and did a bit of struct exploration in C btw. I was able to confirm that we are calculating the field offsets correctly for `epoll_event`, at least on that architecture: `data` is 8 bytes from the front, which is exactly the pointer size.

durban:
Okay, I think the segfault is gone with 2fafd1d. It seems, that the `sigaction` definition in scala-native is completely incorrect (at least on Linux; scala-native/scala-native/issues/4274). So I rewrote the method calling `sigaction` in C. It's not exactly nice, but I don't see a better option (until SN fixes the definition), since not just the size, but the field offsets are also completely wrong (in fact, I'm surprised it seems to work on x86_64 Linux). In theory we could try something with `if (LinktimeInfo.target.arch ...` (like with `epoll_event`), but I fear it would be subtly wrong with a high probability, and also, why do that, when the same information is already in signal.h...

I think the segfault happened, because `sigaction` was called with a too small bad struct, and it "picked" up also other random things from the stack, it coincidentally didn't segfault _then_, but later, when the (badly) registered handler was called, it segfaulted.

Previously I've also fixed some small things (when trying to figure out what segfaults).

The last CI failure (with 985b2ec) is something completely different:
```
Out of heap space cannot allocate more objects
	at StackTrace_PrintStackTrace
	at Heap_exitWithOutOfMemory
	at Allocator_allocSlow
	at Allocator_Alloc
	at scalanative_GC_alloc_small
	at _SM23cats.effect.IO$FlatMap$D5applyL14cats.effect.IOL15scala.Function1L32cats.effect.tracing.TracingEventL22cats.effect.IO$FlatMapEO
	at _SM14cats.effect.IOD7flatMapL15scala.Function1L14cats.effect.IOEO
	at _SM14cats.effect.IOD7flattenL22scala.$less$colon$lessL14cats.effect.IOEO
	at _SM15cats.effect.IO$D5deferL15scala.Function0L14cats.effect.IOEO
	at _SM22cats.effect.IODeferredRE
	at _SM15cats.effect.IO$D19deferred$$anonfun$1L22cats.effect.IODeferredEpT15cats.effect.IO$
	at _SM25cats.effect.IO$$$Lambda$5D5applyL16java.lang.ObjectEO
	at _SM19cats.effect.IOFiberD7runLoopL14cats.effect.IOiiuEPT19cats.effect.IOFiber
	at _SM19cats.effect.IOFiberD5execRuEPT19cats.effect.IOFiber
	at _SM19cats.effect.IOFiberD3runuEO
	at _SM31cats.effect.unsafe.WorkerThreadD3runuEO
	at _SM39scala.scalanative.runtime.NativeThread$D16threadEntryPointL38scala.scalanative.runtime.NativeThreaduEPT39scala.scalanative.runtime.NativeThread$
	at _SM39scala.scalanative.runtime.NativeThread$D24threadRoutine$$anonfun$1L28scala.scalanative.unsafe.PtrL28scala.scalanative.unsafe.PtrEPT39scala.scalanative.runtime.NativeThread$
	at _SM49scala.scalanative.runtime.NativeThread$$$Lambda$1G17$extern$forwarder
	at ProxyThreadStartRoutine
	at _pthread_start
```

djspiewak:
Incredible work! Now to figure out if it's a real memory leak or if we're just a bit over budget.

djspiewak:
Good news everyone! This one is locally reproducible. It seems legitimate.

djspiewak:
I opened https://github.com/scala-native/scala-native/issues/4275 to track the issue here. I think there's a problem cleaning up threads once allocated, at least on macOS. I set the GC limit up to 8g as a workaround (fixes it locally).

djspiewak:
In honor of our first legit green checkmark on this PR, I'm finally removing the WIP status! Sadly we can't rename the branch but whatever.

core/native/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -17,11 +17,16 @@
 package cats.effect
```
djspiewak, line 17: We should try to consolidate this with the JVM one a bit if we can. There are very minimal differences.

std/shared/src/main/scala/cats/effect/std/Random.scala
```diff
@@ -331,25 +331,41 @@ object Random extends RandomCompanionPlatform {
...
-        val next = d * (maxExclusive - minInclusive) + minInclusive
+        val diff = maxExclusive - minInclusive
+        val next = if (diff != java.lang.Double.POSITIVE_INFINITY) {
+          (d * diff) + minInclusive
+        } else { // overflow:
+          val maxHalf = maxExclusive / 2.0
+          val minHalf = minInclusive / 2.0
+          ((d * (maxHalf - minHalf)) + minHalf) * 2.0
+        }
         if (next < maxExclusive) next
```
djspiewak: @durban Can you check to make sure I didn't mess up this merge?

tests/jvm-native/src/test/scala/cats/effect/std/DispatcherParallelSuite.scala
```diff
@@ -44,26 +44,28 @@ class DispatcherJVMSuite extends BaseSuite {
     } yield ()
   }
 
-  real("propagate Java thread interruption in unsafeRunSync") {
-    Dispatcher.parallel[IO](await = true).use { dispatcher =>
-      for {
-        pre <- IO.deferred[Unit]
-        canceled <- IO.deferred[Unit]
+  if (isJVM) {
+    real("propagate Java thread interruption in unsafeRunSync") {
```
djspiewak, line 48: Oh right I forgot we turned off `interruptible`. Follow up PR.

tests/shared/src/test/scala/cats/effect/std/RandomSuite.scala
```diff
@@ -51,6 +51,27 @@ class RandomSuite extends BaseSuite {
...
+
+    real(s"$name - betweenDouble - handle underflow") {
+      for {
+        random <- randomGen
+        randDouble <- random.betweenDouble(
+          Double.MinPositiveValue,
+          java.lang.Math.nextUp(Double.MinPositiveValue))
+      } yield assert(randDouble == Double.MinPositiveValue)
+    }
+
```
djspiewak: @durban Here's the other side of it

djspiewak (commented):
Took a pass through to make sure it's all sane. Left some comments.

core/native/src/main/scala/cats/effect/IOApp.scala
djspiewak: #4333

djspiewak:
@armanbilge @durban Can you each take a pass through the diff and decide if you think it's reasonable or if there are any things we want to flag for each other? Once we're all happy, we can call this ready to go. (I think my only items are the double-check on the `Random` merge)

std/shared/src/main/scala/cats/effect/std/Random.scala
durban: The merge is fine.

tests/shared/src/test/scala/cats/effect/std/RandomSuite.scala
durban: The merge is fine.

durban (commented):
@djspiewak The Random _merge_ is fine, though there is a bug which was already in my original PR :disappointed: ~~I'll make a new PR against 3.6.x.~~

durban:
What I thought was a real bug in my original #4302, turned out to have no _real_ effect (due to the fact that `java.lang.Double.POSITIVE_INFINITY == java.lang.Float.POSITIVE_INFINITY`). That's why the tests didn't catch it. So instead of making a PR for 3.6.x, I just pushed a commit here. As far as I can tell, it should have no actual behavior change, just makes the source more readable.

@djspiewak I'll try to make a full pass over this whole PR sometime...

durban:
For posterity, this is a CI failure for 9e65eb4:
```
[error] java.nio.file.NoSuchFileException: /home/runner/work/cats-effect/cats-effect/laws/native/target/scala-3.3.5/native-test/generated/56d623b9.ll-68e811fe.o.tmp
[error] 	at sun.nio.fs.UnixException.translateToIOException(UnixException.java:86)
[error] 	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
[error] 	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)
[error] 	at sun.nio.fs.UnixFileAttributeViews$Basic.readAttributes(UnixFileAttributeViews.java:55)
[error] 	at sun.nio.fs.UnixFileSystemProvider.readAttributes(UnixFileSystemProvider.java:144)
[error] 	at sun.nio.fs.LinuxFileSystemProvider.readAttributes(LinuxFileSystemProvider.java:99)
[error] 	at java.nio.file.Files.readAttributes(Files.java:1737)
[error] 	at java.nio.file.FileTreeWalker.getAttributes(FileTreeWalker.java:225)
[error] 	at java.nio.file.FileTreeWalker.visit(FileTreeWalker.java:276)
[error] 	at java.nio.file.FileTreeWalker.next(FileTreeWalker.java:372)
[error] 	at java.nio.file.Files.walkFileTree(Files.java:2706)
[error] 	at scala.scalanative.codegen.SourceCodeCache.localSourceDirs$lzycompute(SourceCodeCache.scala:49)
[error] 	at scala.scalanative.codegen.SourceCodeCache.localSourceDirs(SourceCodeCache.scala:43)
[error] 	at scala.scalanative.codegen.SourceCodeCache.fromRelativePath$1(SourceCodeCache.scala:114)
[error] 	at scala.scalanative.codegen.SourceCodeCache.$anonfun$findSources$22(SourceCodeCache.scala:142)
[error] 	at scala.Option.orElse(Option.scala:447)
[error] 	at scala.scalanative.codegen.SourceCodeCache.$anonfun$findSources$3(SourceCodeCache.scala:142)
[error] 	at scala.collection.concurrent.TrieMap.getOrElseUpdate(TrieMap.scala:895)
[error] 	at scala.scalanative.codegen.SourceCodeCache.findSources(SourceCodeCache.scala:84)
[error] 	at scala.scalanative.codegen.llvm.MetadataCodeGen.$anonfun$toDIFile$1(MetadataCodeGen.scala:213)
[error] 	at scala.Option.flatMap(Option.scala:271)
[error] 	at scala.scalanative.codegen.llvm.MetadataCodeGen.toDIFile(MetadataCodeGen.scala:212)
[error] 	at scala.scalanative.codegen.llvm.MetadataCodeGen.toDIFile$(MetadataCodeGen.scala:210)
[error] 	at scala.scalanative.codegen.llvm.AbstractCodeGen.toDIFile(AbstractCodeGen.scala:22)
[error] 	at scala.scalanative.codegen.llvm.MetadataCodeGen$DefnScopes.getDISubprogramScope$lzycompute(MetadataCodeGen.scala:241)
[error] 	at scala.scalanative.codegen.llvm.MetadataCodeGen$DefnScopes.getDISubprogramScope(MetadataCodeGen.scala:239)
[error] 	at scala.scalanative.codegen.llvm.AbstractCodeGen.$anonfun$genFunctionDefn$6(AbstractCodeGen.scala:305)
[error] 	at scala.scalanative.codegen.llvm.MetadataCodeGen.dbgUsing(MetadataCodeGen.scala:55)
[error] 	at scala.scalanative.codegen.llvm.MetadataCodeGen.dbgUsing$(MetadataCodeGen.scala:50)
[error] 	at scala.scalanative.codegen.llvm.AbstractCodeGen.dbgUsing(AbstractCodeGen.scala:22)
[error] 	at scala.scalanative.codegen.llvm.AbstractCodeGen.genFunctionDefn(AbstractCodeGen.scala:305)
[error] 	at scala.scalanative.codegen.llvm.AbstractCodeGen.genDefn(AbstractCodeGen.scala:208)
[error] 	at scala.scalanative.codegen.llvm.AbstractCodeGen.onDefn$1(AbstractCodeGen.scala:134)
[error] 	at scala.scalanative.codegen.llvm.AbstractCodeGen.$anonfun$genDefns$4(AbstractCodeGen.scala:147)
[error] 	at scala.scalanative.codegen.llvm.AbstractCodeGen.$anonfun$genDefns$4$adapted(AbstractCodeGen.scala:146)
[error] 	at scala.collection.mutable.UnrolledBuffer$Unrolled.foreach(UnrolledBuffer.scala:242)
[error] 	at scala.collection.mutable.UnrolledBuffer.foreach(UnrolledBuffer.scala:148)
[error] 	at scala.scalanative.codegen.llvm.AbstractCodeGen.genDefns(AbstractCodeGen.scala:146)
[error] 	at scala.scalanative.codegen.llvm.AbstractCodeGen.$anonfun$gen$2(AbstractCodeGen.scala:68)
[error] 	at scala.scalanative.codegen.llvm.AbstractCodeGen.$anonfun$gen$2$adapted(AbstractCodeGen.scala:66)
[error] 	at scala.scalanative.io.VirtualDirectory$NioDirectory.write(VirtualDirectory.scala:113)
[error] 	at scala.scalanative.io.VirtualDirectory$NioDirectory.write$(VirtualDirectory.scala:109)
[error] 	at scala.scalanative.io.VirtualDirectory$LocalDirectory.write(VirtualDirectory.scala:151)
[error] 	at scala.scalanative.codegen.llvm.AbstractCodeGen.$anonfun$gen$1(AbstractCodeGen.scala:66)
[error] 	at scala.scalanative.codegen.llvm.AbstractCodeGen.$anonfun$gen$1$adapted(AbstractCodeGen.scala:61)
[error] 	at scala.scalanative.io.VirtualDirectory$NioDirectory.write(VirtualDirectory.scala:113)
[error] 	at scala.scalanative.io.VirtualDirectory$NioDirectory.write$(VirtualDirectory.scala:109)
[error] 	at scala.scalanative.io.VirtualDirectory$LocalDirectory.write(VirtualDirectory.scala:151)
[error] 	at scala.scalanative.codegen.llvm.AbstractCodeGen.gen(AbstractCodeGen.scala:61)
[error] 	at scala.scalanative.codegen.llvm.CodeGen$.$anonfun$emit$10(CodeGen.scala:132)
[error] 	at scala.concurrent.Future$.$anonfun$apply$1(Future.scala:659)
[error] 	at scala.util.Success.$anonfun$map$1(Try.scala:255)
[error] 	at scala.util.Success.map(Try.scala:213)
[error] 	at scala.concurrent.Future.$anonfun$map$1(Future.scala:292)
[error] 	at scala.concurrent.impl.Promise.$anonfun$transform$1(Promise.scala:42)
[error] 	at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:74)
[error] 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
[error] 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
[error] 	at java.lang.Thread.run(Thread.java:750)
[error] (testsNative / Compile / nativeLink) java.nio.file.NoSuchFileException: /home/runner/work/cats-effect/cats-effect/laws/native/target/scala-3.3.5/native-test/generated/56d623b9.ll-68e811fe.o.tmp
```

Honestly, I just shrugged, and restarted the CI...

djspiewak:
```
[error] (coreNative / Compile / packageSrc) java.util.zip.ZipException: duplicate entry: cats/effect/unsafe/LocalQueue.scala
```

Tried to release a snapshot. That's a new one…

durban:
To be fair, there _is_ a `LocalQueue.scala` both in `jvm-native` and `native`... I guess this is when it's trying to make source JARs?

Also, there is this (I didn't look into it yet):
```
[info] ==> X cats.effect.IOSuite.blocking work does not starve poll 1.227s munit.ComparisonFailException: tests/jvm-native/src/test/scala/cats/effect/IOConcurrencySuite.scala:536
[info] 535:
[info] 536:        assertEquals(test.unsafeRunTimed(1.second), Some(true))
[info] 537:      } finally {
[info] values are not the same
[info] => Obtained
[info] None
[info] => Diff (- expected, + obtained)
[info] -Some(
[info] -  value = true
[info] -)
[info] +None
```

djspiewak:
@durban That might be just native taking a longer time than normal. It seems to be about 5x slower than JVM.

djspiewak:
Released as `3.7-8f2b497`

djspiewak:
Saw that same error (the missing tempfile) a second time. Appears to be something in the SN compiler itself. Very odd.

iRevive:
A small change to move WSTP metrics to `jvm-native`: https://github.com/durban/cats-effect/pull/1. We can merge it now, or I can make a follow-up PR once this PR (#4201) is merged.

durban:
@iRevive I've merged it here, easier now than later. (Btw, I think you should've been able to push to this PR directly. Anyway, it's done now.)

durban:
Without b3c08f53bd, `WSTP#shutdown` interrupts each thread with `Thread#interrupt`. Now for `SleepSystem` and `SelectorSystem` this is fine. (`Thread#interrupt` is explicitly spec'd to interrupt selectors, and `LockSupport#parkNanos` is similarly spec'd to recognize `Thread#interrupt`.) However, `EpollSystem` (and I suspect also `KQueueSystem`) is just stuck in a blocking `epoll_wait` call (or `kqueue_wait`, or however it's called). This causes (1) the WSTP shutdown to wait for the full `shutdownTimeout` (1sec by default), and (2) the threads to not terminate (because why would they, they're indefinitely stuck in a blocking call), and thus continue to use memory. I believe this causes scala-native/scala-native/issues/4275 (or at least it's _one_ of the causes). The attached profile there, which shows 1003ms in `poll`, I believe it's the 1sec `shutdownTimeout`.

In b3c08f53bd I've added a `PollingSystem#interrupt` call before each `Thread#interrupt`. This seems to solve the problem on Linux. (Whoever has a macos, can try there.) Now I'm not sure, calling _both_ `interrupt`s is entirely correct. However, `Thread#interrupt` by itself is apparently not enough. `PollingSystem#interrupt` by itself wouldn't set the thread interrupted status, which might be needed. (Is it?) Also, if we call both, in which order?

A question could be: should scala-native `Thread#interrupt` also interrupt `epoll_wait`? I think it's not able to do that, but I might be missing something. (After all, it's just a call to a random C function... although it has the `blocking` annotation... but still, how could it know, what to do to interrupt it?)

djspiewak:
Ahhhhhh! Excellent catch! This makes a lot of sense and basically fully explains the behavior. I do think we should continue interrupting the threads, since that allows us to catch a few cases where we aren't blocked in a poller (especially on the JVM) and it's best practice to set those flags.

> A question could be: should scala-native Thread#interrupt also interrupt epoll_wait? I think it's not able to do that, but I might be missing something. (After all, it's just a call to a random C function... although it has the blocking annotation... but still, how could it know, what to do to interrupt it?)

It can't and it shouldn't try, IMO. The JVM cannot interrupt threads blocked in JNI calls for the same reason. The only reason interruption works the way it does with many (not all) JVM blocking intrinsics is because pipes and other bespoke signaling mechanisms are employed, very similarly to how we implement `PollingSystem#interrupt`.

armanbilge:
Indeed great catch @durban! I think we actually need b3c08f53bd45d3eb1da403bf82ff6d2ffd1e9d5e on `series/3.6.x` because the bug is not specific to Scala Native.

djspiewak:
Good call. Actually to avoid merge conflicts the ideal route would be to revert here, then cherry pick to a new branch on top of 3.6, PR that, and merge the PR back over here.

Or we can roll the dice on conflicts, but feels safer to not.

durban:
Okay, I've pushed the revert, and here is the PR: #4340.

djspiewak:
Unexpected (expected?) bonus from @durban's fix: CI is much faster now!

djspiewak (dismissed):
As noted previously, looking for all three approvals here before merge.

djspiewak:
For those of you with more time than sense, here's another snapshot: `3.7-4972921`

std/shared/src/main/scala/cats/effect/std/Random.scala
```diff
@@ -355,7 +355,7 @@ object Random extends RandomCompanionPlatform {
         f <- nextFloat
       } yield {
         val diff = maxExclusive - minInclusive
-        val next = if (diff != java.lang.Double.POSITIVE_INFINITY) {
+        val next = if (diff != java.lang.Float.POSITIVE_INFINITY) {
```
armanbilge, line 358: Is this a bugfix that's not specific to Native?

armanbilge (dismissed):
I think it's good (enough ;)! Opened a bunch of follow-up issues, but they are essentially just optimizations. We seem to be getting green CI consistently and on Cirrus as well which is amazing.

Huge props to Daniel and Daniel for driving this effort from start-to-finish.

djspiewak (dismissed):
Nice catch

core/native/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -449,7 +448,7 @@ trait IOApp {
     }
   }
 
-  private[this] def halt(status: Int): Unit = exit(status)
+  private[this] def halt(status: Int): Unit = System.exit(status)
```
armanbilge: @durban I'm confused by this change. The (undocumented, my bad) intention of `halt` is to offer semantics equivalent to `Runtime#halt`:

> Forcibly terminates the currently running Java virtual machine. This method never returns normally.
This method should be used with extreme caution. Unlike the [exit](https://docs.oracle.com/javase/8/docs/api/java/lang/Runtime.html#exit-int-) method, this method does not cause shutdown hooks to be started. If the shutdown sequence has already been initiated then this method does not wait for any running shutdown hooks to finish their work.

https://docs.oracle.com/javase/8/docs/api/java/lang/Runtime.html#halt-int-

durban: @armanbilge Okay, I understand the intention now. Except that System.exit seems to call directly stdlib.exit [here](https://github.com/scala-native/scala-native/blob/1c91ccff21c79a60e511419528189c99521f098d/javalib/src/main/scala/java/lang/Runtime.scala#L143). And I think hooks will be called [here](https://github.com/scala-native/scala-native/blob/1c91ccff21c79a60e511419528189c99521f098d/javalib/src/main/scala/java/lang/Runtime.scala#L21). :confused:

armanbilge: Oh darn, thanks for checking that. Not sure what to do then. Maybe we can just leave this to a follow-up issue.

armanbilge: I'm going to revert 9ce1dfaf923de0ffca68a1ad237b74c81d7c62e2 because even if it currently behaves the same, it distinguishes the semantic. Then open a follow-up.

durban: @armanbilge c33d8876ce

durban: Uhh, sorry, that was a race condition in pushing to this branch... I won... take a look...

armanbilge (dismissed):
@durban ready to land this? 😃

std/shared/src/main/scala/cats/effect/std/Random.scala
durban: It's not really a bugfix, more like a "stlye" thing; there is no real bug (i.e., it doesn't need to go into 3.6.x), because `Double.POSITIVE_INFINITY == Float.POSITIVE_INFINITY`.

durban:
I went over this whole PR, and I have one real concern: whether we're invoking undefined behavior in `TimerHeap` (scala-native/scala-native/issues/4288; #4357). But even if we are, that sounds like something maybe worth fixing in scala-native, and probably shouldn't block this (already long) PR. I can't github-approve, because it's "my" PR, but I'm approving.

Various things that we should look at (not in this PR, but probably before 3.7.0):

- ~~Cats 2.13.0~~ #4356
- ~~`FiberMonitor` on scala-native~~ #3837 
- ~~`TracingPlatform`: I think the `ConcurrentHashMap` grows forever, is this a memory leak?~~ (It's not, there is no class unloading on SN.)
- ~~The `Runtime#halt` thing~~ (it's already #4354).
- ~~Double check `stackalloc`s (there have been problems with them):~~ #4358
  - ~~`KqueueSystem`: `stackalloc[timespec]`~~
  - ~~`WorkStealingThreadPoolPlatform`: `stackalloc[timespec]`~~
- ~~`IOApp`: there used to be a "keepalive" on scala-native. Was this a mistake? Is this only a JS thing? If it wasn't, do we not need it any more?~~ We don't need it any more.
- ~~`IORuntimeBuilder#addPoller` seems to exist on JS, and it's silently ignored(?).~~ #4359

djspiewak:
> whether we're invoking undefined behavior in TimerHeap

Oh that's fascinating. Can you open an issue on our side to track this so we don't lose it? I wouldn't want to do a proper 3.7 with that question outstanding.

> FiberMonitor on scala-native

Yeah I think @armanbilge has an issue for this. It's going to be dependent on getting `WeakReference` working correctly on SN.

> TracingPlatform: I think the ConcurrentHashMap grows forever, is this a memory leak?

I don't think so. It's bounded by the number of distinct lambdas (not closures), no?

> Double check stackallocs (there have been problems with them):

Can you make an issue for this one too? Agreed we should look really carefully at them.

> IOApp: there used to be a "keepalive" on scala-native. Was this a mistake? Is this only a JS thing? If it wasn't, do we not need it any more?

I think it was cargo-culted from Scala.js, where in turn I cargo-culted it from the original `IOApp` in CE2, where it was originally added by @alexandru. I still don't actually know what purpose it serves, but our ancestors work in mysterious ways.

I don't think it was ever needed on Native.

> IORuntimeBuilder#addPoller seems to exist on JS, and it's silently ignored(?).

Oh, that feels like a bug. I wonder if we should pull that back?

armanbilge:
> I went over this whole PR, and I have one real concern: whether we're invoking undefined behavior in `TimerHeap` ([scala-native/scala-native/issues/4288](https://github.com/scala-native/scala-native/issues/4288))

Can you share more details (maybe a follow-up issue)? I saw your SN issue but I didn't have a chance to think about it carefully.

---

> * `FiberMonitor` on scala-native

- https://github.com/typelevel/cats-effect/issues/3837

---

> * `TracingPlatform`: I think the `ConcurrentHashMap` grows forever, is this a memory leak?

It's keyed by classes, and there's a finite number of classes in a compiled artifact.

---

> * `IOApp`: there used to be a "keepalive" on scala-native. Was this a mistake? Is this only a JS thing? If it wasn't, do we not need it any more?

It wasn't a mistake. We don't need it anymore. We needed it before because we couldn't block on completion of the main fiber in the main method. e.g. consider an application that is just `IO.never`, how to keep the runtime from terminating? This is what "keepalive" helped with. But now we can just block.

---

> * `IORuntimeBuilder#addPoller` seems to exist on JS, and it's silently ignored(?).

JS doesn't really support pollers ... you're relying on the runtime for this. Did we put `addPoller` on `IORuntimeBuilder` in this PR? If so, let's undo it, if not, let's make an issue and deal with it separately.

durban:
@djspiewak @armanbilge Okay, I promise I'll open an issue for everything you've mentioned, just not right now...

PR #4204, opened 2024-12-18
Add `IORuntime.isUnderFiberContext`

armanbilge:
Follow-up to https://github.com/typelevel/cats-effect/pull/4200. Makes more complex `IOLocal`/`ThreadLocal` interop possible in userland. cc @iRevive 

Let's bikeshed (or not...) `isUnderFiberContext` and the property `cats.effect.trackFiberContext`. Not sure how I feel about "context".


iRevive:
Here is the draft implementation: https://github.com/typelevel/otel4s/pull/870/files. 
The tests pass locally.

djspiewak (approved):
"Context" is the only bikeshed-worthy bit of this, IMO, and I think it's not a terrible noun. "Frame" also comes to mind but is more limited and makes less sense. I'll give it some more thought today but I think this is probably good as-is.

PR #4205, opened 2024-12-19
Refactor WSTP metrics

armanbilge:
So that they better match the structure of the WSTP. This will make it easier to add `WorkerThread`-specific metrics and `PollingSystem` metrics. cc @iRevive

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/metrics/WorkStealingPoolMetrics.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/metrics/WorkStealingPoolMetrics.scala
@@ -91,7 +91,7 @@ sealed trait WorkStealingPoolMetrics {
   def workerThreads: List[WorkerThreadMetrics]
 }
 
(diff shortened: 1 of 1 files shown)


djspiewak (dismissed):

core/jvm/src/main/scala/cats/effect/unsafe/metrics/WorkStealingPoolMetrics.scala
```diff
@@ -31,241 +31,226 @@ sealed trait WorkStealingPoolMetrics {
...
+  def suspendedFiberCount(): Long
+
+  /**
+   * The list of worker-specific metrics of this work-stealing thread pool.
+   */
+  def workerThreads: List[WorkerThreadMetrics]
 }
 
-object WorkStealingPoolMetrics {
+trait WorkerThreadMetrics {
```
armanbilge:
```suggestion
sealed trait WorkerThreadMetrics {
```

PR #4206, opened 2024-12-19
Add `WorkerThread`-specific metrics

armanbilge:
The idle time metric is inspired by libuv and measures the total time the worker thread has been parked. We also report the number of times it has parked.
https://docs.libuv.org/en/v1.x/metrics.html#c.uv_metrics_idle_time

The blocking / respawn metrics count the number of times the worker thread (represented by this index) became a blocking thread and was swapped out, and the number of times this swapping required spawning a new thread (instead of using a cached thread).

Lastly, we count the number of times the thread polled for I/O.

cc @iRevive

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/metrics/WorkStealingPoolMetrics.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/metrics/WorkStealingPoolMetrics.scala
@@ -104,6 +104,11 @@ trait WorkerThreadMetrics {
   def idleTime(): Long
 
   /**
+   * The total number of times that this WorkerThread has parked.
+   */
+  def parkedCount(): Long
+
+  /**
    * The total number of times that this WorkerThread has switched to a blocking thread and been
    * replaced.
    */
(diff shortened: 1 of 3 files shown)


djspiewak (dismissed):

core/jvm/src/main/scala/cats/effect/unsafe/metrics/WorkStealingPoolMetrics.scala
```diff
@@ -236,6 +263,14 @@ object WorkStealingPoolMetrics {
       idx: Int
   ): WorkerThreadMetrics = new WorkerThreadMetrics {
     val index: Int = idx
+
+    private val metrics = wstp.metrices(idx)
+    def idleTime(): Long = metrics.getIdleTime()
+    def parkedCount(): Long = metrics.getParkedCount()
+    def polledCount(): Long = metrics.getPolledCount()
+    def blockingCount(): Long = metrics.getBlockingCount()
+    def respawnCount(): Long = metrics.getBlockingCount()
```
armanbilge:
```suggestion
    def respawnCount(): Long = metrics.getRespawnCount()
```

djspiewak (approved):

PR #4208, opened 2024-12-20
Fix MapRef.fromSeqRefs vulnerability

### LINKED ISSUE #4202: MapRef.fromSeqRefs fails on Nil
It's possible to create instance of `MapRef` from empty list of `Ref`s, but implementation of `fromSeqRefs` gives an error `java.lang.ArithmeticException: / by zero` when attempting to use that `MapRef` instance. It probably needs to either allow `fromSeqRefs` to be called only on a non-empty list, or avoid rising `ArithmeticException` inside the `fromSeqRefs` with current signauture.

underskyer:
I just fix `fromSeqRefs` constructor of `MapRef`. There it was possible to call this method with an empty list of shards, which resulted in a divide by zero exception. Now you can call it with _nonempty_ sequence of shards only. That constructor is only used in two other constructors in the code base. So it no needs to fix some tests or docs, I think.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/MapRef.scala
+++ b/std/shared/src/main/scala/cats/effect/std/MapRef.scala
@@ -91,12 +91,10 @@ object MapRef extends MapRefCompanionPlatform {
       shardCount: Int
   ): G[MapRef[F, K, Option[V]]] = {
     if (shardCount >= 1) {
-      Sync[G].defer {
-        List
-          .fill(shardCount)(())
-          .traverse(_ => Ref.in[G, F, Map[K, V]](Map.empty))
-          .map(lst => fromNonEmptySeqRefs(NonEmptySeq.fromSeqUnsafe(lst)))
-      }
+      List
+        .fill(shardCount)(())
+        .traverse(_ => Ref.in[G, F, Map[K, V]](Map.empty))
+        .map(lst => fromNonEmptySeqRefs(NonEmptySeq.fromSeqUnsafe(lst)))
     } else {
       ApplicativeError[G, Throwable].raiseError(
         new IllegalArgumentException("Shards count should be greater then zero")
--- a/std/shared/src/main/scala/cats/effect/std/MapRef.scala
+++ b/std/shared/src/main/scala/cats/effect/std/MapRef.scala
@@ -114,7 +114,7 @@ object MapRef extends MapRefCompanionPlatform {
       seq: scala.collection.immutable.Seq[Ref[F, Map[K, V]]]
   ): MapRef[F, K, Option[V]] = {
     val array = seq.toArray
-    val shardCount = seq.size
+    val shardCount = array.length
     val refFunction = { (k: K) =>
       val location = Math.abs(k.## % shardCount)
       array(location)
@@ -131,7 +131,7 @@ object MapRef extends MapRefCompanionPlatform {
       seq: NonEmptySeq[Ref[F, Map[K, V]]]
   ): MapRef[F, K, Option[V]] = {
     val array = seq.toSeq.toArray
-    val shardCount = seq.length
+    val shardCount = array.length
     val refFunction = { (k: K) =>
       val location = Math.abs(k.## % shardCount)
       array(location)
--- a/std/shared/src/main/scala/cats/effect/std/MapRef.scala
+++ b/std/shared/src/main/scala/cats/effect/std/MapRef.scala
@@ -68,11 +68,16 @@ object MapRef extends MapRefCompanionPlatform {
   def ofShardedImmutableMap[F[_]: Concurrent, K, V](
       shardCount: Int
   ): F[MapRef[F, K, Option[V]]] = {
-    assert(shardCount >= 1, "MapRef.sharded should have at least 1 shard")
-    List
-      .fill(shardCount)(())
-      .traverse(_ => Concurrent[F].ref[Map[K, V]](Map.empty))
-      .map(lst => fromNonEmptySeqRefs(lst.toNeSeq.get))
+    if (shardCount >= 1) {
+      List
+        .fill(shardCount)(())
+        .traverse(_ => Concurrent[F].ref[Map[K, V]](Map.empty))
+        .map(fromNonEmptySeqRefs[F, K, V] compose NonEmptySeq.fromSeqUnsafe)
+    } else {
+      ApplicativeError[F, Throwable].raiseError(
+        new IllegalArgumentException("Shards count should be greater then zero")
+      )
+    }
   }
 
   /**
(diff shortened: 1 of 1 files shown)


underskyer:
Maybe I should leave the old method as `deprecated`?

djspiewak:
> Maybe I should leave the old method as deprecated?

I think this will definitely be required for compatibility. Speaking of which, why didn't mima flag this?

armanbilge:
> I think this will definitely be required for compatibility. Speaking of which, why didn't mima flag this?

It's actually not required for binary-compatibility, because `NonEmptySeq` is an `AnyVal` wrapper for `Seq`.

So it's really just a matter of source-compatibility.

djspiewak:
Oh, I didn't realize that. I'm totally cool with murdering source compat in 3.6 for something like this, so let's skip the deprecation.

std/shared/src/main/scala/cats/effect/std/MapRef.scala
```diff
@@ -89,19 +89,19 @@ object MapRef extends MapRefCompanionPlatform {
...
    * This uses universal hashCode and equality on K.
    */
   def fromSeqRefs[F[_]: Functor, K, V](
-      seq: scala.collection.immutable.Seq[Ref[F, Map[K, V]]]
+      seq: NonEmptySeq[Ref[F, Map[K, V]]]
   ): MapRef[F, K, Option[V]] = {
-    val array = seq.toArray
-    val shardCount = seq.size
+    val array = seq.toSeq.toArray
+    val shardCount = seq.size.toInt
```
armanbilge:
```suggestion
    val shardCount = array.length
```

std/shared/src/main/scala/cats/effect/std/MapRef.scala
```diff
@@ -72,7 +72,7 @@ object MapRef extends MapRefCompanionPlatform {
     List
       .fill(shardCount)(())
       .traverse(_ => Concurrent[F].ref[Map[K, V]](Map.empty))
-      .map(fromSeqRefs(_))
+      .map(lst => fromSeqRefs(lst.toNeSeq.get))
```
armanbilge: We should clean up the implementation ... instead of an `assert`, we should use `F.raiseError`, and instead of `.get` we could do `fromSeqUnsafe`

underskyer: What error should there be?

underskyer: Does this have to be an exception of some specific type?

armanbilge: It can be an `IllegalArgumentException`

underskyer: I'm actually not a big fan of `if' statements. But I'm not sure about the coding style here.

std/shared/src/main/scala/cats/effect/std/MapRef.scala
underskyer: done

std/shared/src/main/scala/cats/effect/std/MapRef.scala
```diff
@@ -119,7 +131,7 @@ object MapRef extends MapRefCompanionPlatform {
       seq: NonEmptySeq[Ref[F, Map[K, V]]]
   ): MapRef[F, K, Option[V]] = {
     val array = seq.toSeq.toArray
-    val shardCount = seq.size.toInt
+    val shardCount = seq.length
```
armanbilge: Sorry to be a broken record, we should really prefer `array.length` here. Calculating the length for an arbitrary seq may be O(n) e.g. need to iterate over the entire linked list.
```suggestion
    val shardCount = array.length
```

underskyer: done

std/shared/src/main/scala/cats/effect/std/MapRef.scala
```diff
@@ -84,24 +89,49 @@ object MapRef extends MapRefCompanionPlatform {
...
       shardCount: Int
-  ): G[MapRef[F, K, Option[V]]] = Sync[G].defer {
-    assert(shardCount >= 1, "MapRef.sharded should have at least 1 shard")
-    List
-      .fill(shardCount)(())
-      .traverse(_ => Ref.in[G, F, Map[K, V]](Map.empty))
-      .map(fromSeqRefs(_))
+  ): G[MapRef[F, K, Option[V]]] = {
+    if (shardCount >= 1) {
+      Sync[G].defer {
```
armanbilge: I don't think this need to be wrapped in `defer`.

underskyer: It looks like there really aren't any side effects expected inside that would make sense to deffer for later. Should it be removed?

armanbilge: Yes, it can be removed.

std/shared/src/main/scala/cats/effect/std/MapRef.scala
```diff
@@ -84,24 +89,47 @@ object MapRef extends MapRefCompanionPlatform {
...
+      )
+    }
   }
 
   /**
    * Creates a sharded map ref from a sequence of refs.
    *
    * This uses universal hashCode and equality on K.
    */
+  @deprecated("Use fromNonEmptySeqRefs instead", "3.5.0")
```
armanbilge:
```suggestion
  @deprecated("Use fromNonEmptySeqRefs instead", "3.6.0")
```

std/shared/src/main/scala/cats/effect/std/MapRef.scala
```diff
@@ -84,24 +89,47 @@ object MapRef extends MapRefCompanionPlatform {
...
       seq: scala.collection.immutable.Seq[Ref[F, Map[K, V]]]
   ): MapRef[F, K, Option[V]] = {
     val array = seq.toArray
-    val shardCount = seq.size
+    val shardCount = array.length
+    val refFunction = { (k: K) =>
+      val location = Math.abs(k.## % shardCount)
+      array(location)
+    }
+    k => fromSingleImmutableMapRef(refFunction(k)).apply(k)
```
armanbilge: Instead of repeating the implementation, we should delegate from one method to the other.

underskyer: It's not easy to unify the implementation of these functions. One requires `Sync`, while the other requires `Concurent`. These are independent type classes - there are no subtyping relations or other conversions between them. It seems that the `Concurent` variant is redundant, but we should keep it for backward compatibility.

underskyer: Are you talking about `ofShardedImmutableMap` and `inShardedImmutableMap`?))

armanbilge: No, I'm talking about the method you deprecated and the method you added, `fromSeqRefs` and `fromNonEmptySeqRefs`, since you duplicated the implementation.

underskyer: I thought the deprecated method would be removed in a future version, so I left it unchanged. Ok, I'll rewrite it so that it refers to the new implementation.

std/shared/src/main/scala/cats/effect/std/MapRef.scala
```diff
@@ -58,6 +58,9 @@ object MapRef extends MapRefCompanionPlatform {
 
   }
 
+  private def noShardsException = new IllegalArgumentException(
+    "Shards count should be greater then zero")
```
iRevive: typo:
```suggestion
    "Shards count should be greater than zero")
```

armanbilge (approved):
Nice work!

PR #4209, opened 2024-12-20
Introduce `PollerMetrics`

### LINKED ISSUE #3686: Polling system metrics
We should figure out what kinds of metrics we want to collect and how to expose them in the API. cc @iRevive

A couple ideas:

1. libuv reports an "idle time" metric. we could measure this per-thread, and actually this is not specific to the polling system at all, this is a general property of the WSTP.
    > Retrieve the amount of time the event loop has been idle in the kernel’s event provider (e.g. epoll_wait).

    https://docs.libuv.org/en/v1.x/metrics.html#c.uv_metrics_idle_time

2. number of outstanding channels/file descriptors/interest events, also per-thread. would be interesting to see if this balances across threads. on the JVM, we could break this into additional categories: # TCP sockets, # server sockets, etc.

Linking to:
- https://github.com/typelevel/cats-effect/pull/3317

Follow-up to:
- https://github.com/typelevel/cats-effect/pull/3332

armanbilge:
Closes https://github.com/typelevel/cats-effect/issues/3686. cc @iRevive

We now track:
- current outstanding operations
- total submitted operations
- total completed operations
- total errored operations
- total canceled operations

for the following:
- generic operations
- server socket accepts
- client socket connects
- reads
- writes

I think this covers the main things that are useful, interesting, and widely-supported. JDK NIO and io_uring can report on accepts and connects, but for epoll/kqueue those will be bundled under reads and writes, respectively.

In the future we can add a mechanism for polling systems to expose their own custom metrics, in addition to these.

I didn't implement metrics on the Native side yet. We can do during/following the Native multi-threading work.

Bikeshed: "completed" -> "succeeded" ? that would be more consistent with our terminology in CE, I reached for "completed" as in "I/O completions"

### DIFF
--- a/core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
+++ b/core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
@@ -17,6 +17,8 @@
 package cats.effect
 package unsafe
 
+import cats.effect.unsafe.metrics.PollerMetrics
+
 /**
  * Represents a stateful system for managing and interacting with a polling system. Polling
  * systems are typically used in scenarios such as handling multiplexed blocking I/O or other
@@ -101,6 +103,8 @@ abstract class PollingSystem {
    *   is the poller to be interrupted.
    */
   def interrupt(targetThread: Thread, targetPoller: Poller): Unit
+
+  def metrics(poller: Poller): PollerMetrics
 
 }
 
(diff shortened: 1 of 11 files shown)


djspiewak (approved):

djspiewak:
> Bikeshed: "completed" -> "succeeded" ? that would be more consistent with our terminology in CE, I reached for "completed" as in "I/O completions"

Complete seems fine since we aren't talking about *how* they completed.

armanbilge:
> Complete seems fine since we aren't talking about _how_ they completed.

Ah, but we are. It's only counting _successful_ completions, errors and cancelations are counted separately. So I think that answers my question, I should rename it 😅

PR #4214, opened 2024-12-23
Fixed off-by-one error in other index selection

djspiewak:
Btw, the fact that this only seems to fail on arm64 (not x86_64) is *fascinating*! It would seem to suggest that Linux on x86 is very very unlikely to schedule things such that consecutive thread IDs are awakened for stealing. I thought that was a uniform random though, so maybe we're just getting un/lucky on core count.


djspiewak:
So how do I make Cirrus pick up this PR?

armanbilge (approved):
Good catch. Will look into CI ...

PR #4215, opened 2024-12-26
Add MBean metrics: timer heap, poller, worker thread

### LINKED ISSUE #4213: Register new metrics as MBeans
We added several new metrics recently. They are exposed as public APIs, but not automatically registered as MBeans.

- https://github.com/typelevel/cats-effect/pull/4187
- https://github.com/typelevel/cats-effect/pull/4206
- https://github.com/typelevel/cats-effect/pull/4209

iRevive:
Closes #4213.

## TimerHeap

image

## WorkerThread

image

## Poller 

image

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/metrics/TimerHeapSampler.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/metrics/TimerHeapSampler.scala
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2020-2024 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


core/jvm/src/main/scala/cats/effect/unsafe/metrics/TimerHeapSampler.scala
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @param timer
+ *   the monitored timer heap
+ */
+private[unsafe] final class TimerHeapSampler(timer: TimerHeap) extends TimerHeapSamplerMBean {
+  def getTimersOutstandingCount(): Int = timer.outstandingTimers()
+  def getTotalTimersExecutedCount(): Long = timer.totalTimersExecuted()
+  def getTotalTimersScheduledCount(): Long = timer.totalTimersScheduled()
+  def getTotalTimersCanceledCount(): Long = timer.totalTimersCanceled()
+  def getNextTimerDue(): Long = timer.nextTimerDue().getOrElse(0L)
```
iRevive, line 32: We cannot use `Option` here. Is falling back to `0` ok? Perhaps we should use `Long.MaxValue`?

djspiewak: Honestly something like the local epoch offset - 1 would be ideal, but we don't have that. In lieu of that, we should probably just stick with 0.

djspiewak (approved):

PR #39, opened 2017-04-27
Remove unsafeRunSync from EffectLaws

alexandru:
By using `unsafeRunSync` in `EffectLaws` we're assuming a hell of a lot about the execution model and this cannot be, because:

1. these laws cannot be executed on top of Javascript
2. it's also assuming that `attempt.map(...)` keeps execution on the current thread forever

So because in Monix's test I'm using a `TestScheduler` (my own `TestContext`), these laws dead-lock, because they are freezing the current thread before `TestScheduler` has a chance to execute. And Monix does by default batched async execution, so after a certain configurable threshold, async boundaries get inserted. In other words in a loop of 10,000 cycles you're guaranteed to have async boundaries, since the threshold is configured at about 1024 or something.

Anyway, your kind reminder to stop using `unsafeRunSync` 😜

PR #4216, opened 2024-12-27
Replace specs2 with munit

iRevive:
Follow up to #3787. 

The whole PR is purely focused on the migration. There are options to build more usable test abstractions and utilities, but I don't want to do it without proper thinking.


tests/shared/src/test/scala/cats/effect/std/QueueSuite.scala
```diff
@@ -846,24 +832,24 @@ trait QueueTests[Q[_[_], _]] { self: BaseSpec =>
       } yield continue
 
       val Bound = 10 // only try ten times before skipping
-      def loop(i: Int): IO[Result] = {
+      def loop(i: Int): IO[Unit] = {
         // we try iterating a few times to get canceled in the middle
         if (i > Bound) {
-          IO.pure(skipped(s"attempted $i times and could not reproduce scenario"))
+          IO.println(s"attempted $i times and could not reproduce scenario")
```
iRevive, line 838: So, it's not a direct equivalent. But I cannot think of anything better.

djspiewak:
…holy crap

iRevive:
General 20k LOC changeset. Nothing to worry about 😅

ioapp-tests/src/test/scala/IOAppSpec.scala
```diff
@@ -145,118 +145,123 @@ class IOAppSpec extends Specification {
     platform == JVM && sys.props.get("java.version").filter(_.startsWith("1.8")).isDefined
   lazy val isWindows = System.getProperty("os.name").toLowerCase.contains("windows")
 
-  s"IOApp (${platform.id})" should {
+  {
 
     if (!isWindows) { // these tests have all been emperically flaky on Windows CI builds, so they're disabled
 
-      "evaluate and print hello world" in {
+      test(s"IOApp (${platform.id}) - evaluate and print hello world") {
```
djspiewak: Bit of bikeshedding: I think on this one (and a bunch of similars), we can simply discard the `should`. With `IOAppSpec` it just doesn't add any value.

build.sbt
```diff
@@ -302,10 +302,11 @@ ThisBuild / autoAPIMappings := true
 
 val CatsVersion = "2.11.0"
 val CatsMtlVersion = "1.3.1"
-val Specs2Version = "4.20.5"
 val ScalaCheckVersion = "1.17.1"
-val DisciplineVersion = "1.4.0"
 val CoopVersion = "1.2.0"
+val MUnitVersion = "1.0.0-M11"
+val MUnitScalaCheckVersion = "1.0.0-M11"
+val DisciplineMUnitVersion = "2.0.0-M3"
```
djspiewak, line 327: We should also add the incantation to make munit use sbt's output mechanism rather than its own bootlegged printlns. I can't remember the exact details; I think it was @etspaceman who originally showed me.

laws/shared/src/test/scala/cats/effect/BaseSuite.scala
```diff
@@ -27,7 +27,7 @@ import org.scalacheck.{Arbitrary, Gen, Prop}
 import org.scalacheck.util.Pretty
 
 // A dumping ground for random helpers for property tests
-private[effect] trait BaseSpec {
+private[effect] trait BaseSuite {
```
djspiewak, line 30: :+1:

iRevive:
Specs2 model feels more strict. With munit the following code compiles and the test is green, but it's never being executed:
```scala
test("leaky io") {
  IO(sys.error("boom"))
}
```
The compiler doesn't throw any warning, either. 

_____

I hope I didn't miss such cases.

tests/shared/src/test/scala/cats/effect/IOSuite.scala
```diff
@@ -0,0 +1,2249 @@
...
+    assertCompleteAs(
+      IO.raiseError[Unit](TestException).start.flatMap(_.join),
+      Outcome.errored[IO, Throwable, Unit](TestException))
+  }
+
+  ticked("fibers - start and ignore a non-terminating fiber") { implicit ticker =>
+    assertCompleteAs(IO.never.start.as(42), 42)
+  }
+
+  ticked("fibers - start a fiber then continue with its results") { implicit ticker =>
```
djspiewak, line 441: In some ways, this is honestly less confusing that the `in real`/`in ticked` we had with specs.

tests/shared/src/test/scala/cats/effect/IOSuite.scala
```diff
@@ -0,0 +1,2249 @@
...
+  }
+
+  ticked(
+    "asyncCheckAttempt - continue from the results of an asyncCheckAttempt suspended result produced prior to registration") {
+    implicit ticker =>
+      val fa = IO.asyncCheckAttempt[Int](cb => IO(cb(Right(42))).as(Left(None))).map(_ + 2)
+      assertCompleteAs(fa, 44)
+  }
+
+      // format: off
```
djspiewak: Double indent?

tests/shared/src/test/scala/cats/effect/testkit/TestControlSuite.scala
```diff
@@ -0,0 +1,228 @@
...
+  real("execute - tickFor - advance incrementally in minimum steps") {
+    val step = IO.sleep(1.second) *> IO.realTime
+
+    TestControl.execute((step, step).tupled) flatMap { control =>
+      for {
+        _ <- control.tickFor(1.second + 500.millis)
+        _ <- control.tickAll
+
+        r <- control.results
+        _ <- IO(assertEquals(r, Some(beSucceeded((1.second, 2.seconds)))))
```
djspiewak, line 176: Is `beSucceeded` buying us that much here?

ioapp-tests/src/test/scala/IOAppSpec.scala
```diff
@@ -327,60 +332,56 @@ class IOAppSpec extends Specification {
...
-      "gracefully ignore undefined process.exit" in {
+      test(s"IOApp (${platform.id}) - gracefully ignore undefined process.exit") {
         val h = platform("UndefinedProcessExit", List.empty)
-        h.awaitStatus() mustEqual 0
+        assertEquals(h.awaitStatus(), 0)
       }
-      ()
     }
 
-    "make specs2 happy" in ok
```
djspiewak, line 388: 🎉

djspiewak (changes requested):
NGL, I didn't read through all 11k lines. Did some spot checking around though and left some comments on things that I think probably apply more broadly. Agree we should review carefully. Overall looks good though.

build.sbt
armanbilge:
```scala
Test / testOptions += Tests.Argument("+l")
```

djspiewak:
> Specs2 model feels more strict. With munit the following code compiles and the test is green, but it's never being executed:

This one is concerning, to be sure. We've historically had a bunch of problems with this, and we do have a ton of tests which throw exceptions in various cases. We should explore our own variant of `test` probably which is stricter.

tests/shared/src/test/scala/cats/effect/IOSuite.scala
iRevive: Leftover from the `IOSpec`.

tests/shared/src/test/scala/cats/effect/testkit/TestControlSuite.scala
iRevive: Otherwise, it's more verbose since we must explicitly specify types to make `assertEquals` happy.

tests/shared/src/test/scala/cats/effect/BaseSuite.scala
```diff
@@ -0,0 +1,29 @@
...
+
+package cats.effect
+
+import munit.FunSuite
+
+trait BaseSuite extends FunSuite with Runners {
+
+  override def munitValueTransforms: List[ValueTransform] =
+    super.munitValueTransforms ++ List(
+      new ValueTransform("IO", { case _: IO[_] => sys.error("Non-evaluated IO") }),
```
iRevive: These transformers prevent us from misuse like:
```scala
test("leaky io") {
  IO(sys.error("boom"))
}
```

tests/shared/src/test/scala/cats/effect/BaseSuite.scala
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect
+
+import munit.FunSuite
+
+trait BaseSuite extends FunSuite with Runners {
```
iRevive, line 25: Majority of MUnit helpers (such as `TestTransforms`, `SuiteTransforms`, `FunFixtures`) require [BaseFunSuite](https://github.com/scalameta/munit/blob/main/munit/shared/src/main/scala/munit/FunSuite.scala#L13). 
`BaseFunSuite` defines the `test` method. 

If we want to have a custom `test` method, we must reimplement many things.

As an alternative, I added the value transformer that checks there is no non-evaluated IO/SyncIO.

iRevive:
specs2 CI time: [5h 22m](https://github.com/typelevel/cats-effect/actions/runs/12519935530/usage?pr=4217)
munit CI time: [5h 6m](https://github.com/typelevel/cats-effect/actions/runs/12548396306/usage?pr=4216)

There's not much difference, assuming we lost a spec parallelism.

djspiewak:
I think the next step on this is probably to go through and surgically break some subtle things one at a time and make sure the test suite still catches it.

tests/shared/src/test/scala/cats/effect/BaseSuite.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  @deprecated("Please use a type safe alternative, such as 'real' or 'ticked'", "3.6.0")
+  override def test(name: String)(body: => Any)(implicit loc: Location): Unit =
+    super.test(name)(body)
+
+  @deprecated("Please use a type safe alternative, such as 'real' or 'ticked'", "3.6.0")
+  override def test(options: TestOptions)(body: => Any)(implicit loc: Location): Unit =
+    super.test(options)(body)
```
iRevive: I deprecated the built-in options. Now, we are forced to use a type-safe alternative.

tests/shared/src/test/scala/cats/effect/BaseSuite.scala
```diff
@@ -0,0 +1,53 @@
...
+import munit.{FunSuite, Location, ScalaCheckSuite, TestOptions}
+import org.scalacheck.Prop
+
+trait BaseSuite extends FunSuite with Runners {
+
+  @deprecated("Please use a type safe alternative, such as 'real' or 'ticked'", "3.6.0")
+  override def test(name: String)(body: => Any)(implicit loc: Location): Unit =
+    super.test(name)(body)
+
+  @deprecated("Please use a type safe alternative, such as 'real' or 'ticked'", "3.6.0")
```
iRevive: Is it `type safe`, `type-safe`, `typsafe`?

tests/shared/src/test/scala/cats/effect/BaseSuite.scala
djspiewak: We can just boink them entirely since it's internal, rather than deprecating.

tests/shared/src/test/scala/cats/effect/BaseSuite.scala
djspiewak: I've seen all three. 😃 Which is your favorite?

armanbilge:
Sorry, I snuck a new test in here.
- https://github.com/typelevel/cats-effect/pull/4223

tests/shared/src/test/scala/cats/effect/BaseSuite.scala
```diff
@@ -0,0 +1,66 @@
...
+            sys.error("Non-evaluated Prop.")
+        }
+      ),
+      new ValueTransform(
+        "Other",
+        {
+          case r if !r.isInstanceOf[Unit] && !r.isInstanceOf[Prop] =>
+            sys.error(s"Unexpected value of type ${r.getClass.getName}: $r")
+        }
+      )
```
iRevive: This transformer ensures the result type is `Unit` or `Prop`. 
From the munit perspective, that's a valid test:
```scala
test("some test") {
  1 == 2
}
```
However, it should use `assert`:
```scala
test("some test") {
  assert(1 == 2)
}
```

______

Here are all the fixed tests that were exposed by this transformer: https://github.com/typelevel/cats-effect/pull/4216/commits/9186d4bb7f7ba2fa2681c85181e7904264c70137

tests/shared/src/test/scala/cats/effect/Runners.scala
```diff
@@ -22,86 +22,113 @@ import cats.effect.unsafe.IORuntime
...
 
-  def ticked[A: AsResult](test: Ticker => A): Execution =
-    Execution.result(test(Ticker(TestContext())))
+  def executionTimeout: FiniteDuration = 20.seconds
+  override def munitTimeout: Duration = executionTimeout
 
-  def real[A: AsResult](test: => IO[A]): Execution =
-    Execution.withEnvAsync { _ =>
-      val (fut, cancel) = test.unsafeToFutureCancelable()(runtime())
+  def ticked(options: TestOptions)(body: Ticker => Unit)(implicit loc: Location): Unit =
```
iRevive, line 39: I've tightened the body type from `Ticker => A` to `Ticker => Unit`. Spotted a few non-evaluated assertions that way.

iRevive:
> Sorry, I snuck a new test in here.
> 
> * [Report non-fatal `Runnable` failures on `MainThread` EC #4223](https://github.com/typelevel/cats-effect/pull/4223)

No worries. I've backported the test.

djspiewak:
So for some reason, `IOAppSpec` just hung for me locally. I ran `root/testQuick` and I think it was the native platform which hung:

<details>

```
[warn] Canceling execution...
[error] Test runner interrupted by fatal signal 2
[warn] Force close java.lang.RuntimeException: Process /Users/daniel/Development/Scala/cats-effect/series-3.x/example/native/target/scala-2.13/cats-effect-example-test-out finished with non-zero value 130 (0x82)
[error] Test runner interrupted by fatal signal 2
[warn] Force close java.lang.RuntimeException: Process /Users/daniel/Development/Scala/cats-effect/series-3.x/laws/native/target/scala-2.13/cats-effect-laws-test-out finished with non-zero value 130 (0x82)
[error] Test runner interrupted by fatal signal 2
[error] Test runner interrupted by fatal signal 2
[error] Test runner interrupted by fatal signal 2
[warn] Force close java.lang.RuntimeException: Process /Users/daniel/Development/Scala/cats-effect/series-3.x/tests/native/target/scala-2.13/cats-effect-tests-test-out finished with non-zero value 130 (0x82)
[error] Test runner interrupted by fatal signal 2
[warn] Force close java.lang.RuntimeException: Process /Users/daniel/Development/Scala/cats-effect/series-3.x/testkit/native/target/scala-2.13/cats-effect-testkit-test-out finished with non-zero value 130 (0x82)
[warn] Force close java.lang.RuntimeException: Process /Users/daniel/Development/Scala/cats-effect/series-3.x/tests/native/target/scala-2.13/cats-effect-tests-test-out finished with non-zero value 130 (0x82)
[warn] Force close java.lang.RuntimeException: Process /Users/daniel/Development/Scala/cats-effect/series-3.x/tests/native/target/scala-2.13/cats-effect-tests-test-out finished with non-zero value 130 (0x82)
[warn] Force close java.lang.RuntimeException: Process /Users/daniel/Development/Scala/cats-effect/series-3.x/kernel/native/target/scala-2.13/cats-effect-kernel-test-out finished with non-zero value 130 (0x82)
[warn] Force close java.lang.RuntimeException: Process /Users/daniel/Development/Scala/cats-effect/series-3.x/tests/native/target/scala-2.13/cats-effect-tests-test-out finished with non-zero value 130 (0x82)
[warn] Force close java.lang.RuntimeException: Process /Users/daniel/Development/Scala/cats-effect/series-3.x/tests/native/target/scala-2.13/cats-effect-tests-test-out finished with non-zero value 130 (0x82)
[error] Test runner interrupted by fatal signal 2
[warn] Force close java.lang.RuntimeException: Process /Users/daniel/Development/Scala/cats-effect/series-3.x/tests/native/target/scala-2.13/cats-effect-tests-test-out finished with non-zero value 130 (0x82)
[warn] Force close java.lang.RuntimeException: Process /Users/daniel/Development/Scala/cats-effect/series-3.x/tests/native/target/scala-2.13/cats-effect-tests-test-out finished with non-zero value 130 (0x82)
[warn] Force close java.lang.RuntimeException: Process /Users/daniel/Development/Scala/cats-effect/series-3.x/kernel-testkit/native/target/scala-2.13/cats-effect-kernel-testkit-test-out finished with non-zero value 130 (0x82)
[warn] Force close java.lang.RuntimeException: Process /Users/daniel/Development/Scala/cats-effect/series-3.x/std/native/target/scala-2.13/cats-effect-std-test-out finished with non-zero value 130 (0x82)
[warn] Force close java.lang.RuntimeException: Process /Users/daniel/Development/Scala/cats-effect/series-3.x/tests/native/target/scala-2.13/cats-effect-tests-test-out finished with non-zero value 130 (0x82)
[error] Test runner interrupted by fatal signal 2
[warn] Force close java.lang.RuntimeException: Process /Users/daniel/Development/Scala/cats-effect/series-3.x/tests/native/target/scala-2.13/cats-effect-tests-test-out finished with non-zero value 130 (0x82)
[error] Test runner interrupted by fatal signal 2
[error] Test runner interrupted by fatal signal 2
[error] Test runner interrupted by fatal signal 2
[error] Test runner interrupted by fatal signal 2
[error] Test runner interrupted by fatal signal 2
[error] Test runner interrupted by fatal signal 2
[error] Test runner interrupted by fatal signal 2
[error] Test runner interrupted by fatal signal 2
[error] Test runner interrupted by fatal signal 2
[warn] Force close java.lang.RuntimeException: Process /Users/daniel/Development/Scala/cats-effect/series-3.x/tests/native/target/scala-2.13/cats-effect-tests-test-out finished with non-zero value 130 (0x82)
[warn] Force close java.lang.RuntimeException: Process /Users/daniel/Development/Scala/cats-effect/series-3.x/core/native/target/scala-2.13/cats-effect-test-out finished with non-zero value 130 (0x82)
[info] cats.effect.IOAppSpec:
[error] ==> X cats.effect.IOAppSpec.evaluate and print hello world  0.159s munit.ComparisonFailException: /Users/daniel/Development/Scala/cats-effect/series-3.x/ioapp-tests/src/test/scala/IOAppSpec.scala:154
[error] 153:        val h = platform("HelloWorld", Nil)
[error] 154:        assertEquals(h.awaitStatus(), 0)
[error] 155:        assertEquals(h.stdout(), s"Hello, World!${System.lineSeparator()}")
[error] values are not the same
[error] => Obtained
[error] 1
[error] => Diff (- obtained, + expected)
[error] -1
[error] +0
[error]     at munit.Assertions.failComparison(Assertions.scala:274)
[error] ==> X cats.effect.IOAppSpec.pass all arguments to child  0.055s munit.ComparisonFailException: /Users/daniel/Development/Scala/cats-effect/series-3.x/ioapp-tests/src/test/scala/IOAppSpec.scala:161
[error] 160:        val h = platform("Arguments", expected)
[error] 161:        assertEquals(h.awaitStatus(), 0)
[error] 162:        assertEquals(
[error] values are not the same
[error] => Obtained
[error] 1
[error] => Diff (- obtained, + expected)
[error] -1
[error] +0
[error]     at munit.Assertions.failComparison(Assertions.scala:274)
[error] ==> X cats.effect.IOAppSpec.exit on non-fatal error  0.053s munit.FailException: /Users/daniel/Development/Scala/cats-effect/series-3.x/ioapp-tests/src/test/scala/IOAppSpec.scala:171 assertion failed
[error] 170:        assertEquals(h.awaitStatus(), 1)
[error] 171:        assert(h.stderr().contains("Boom!"))
[error] 172:      }
[error]     at munit.FunSuite.assert(FunSuite.scala:11)
[error]     at cats.effect.IOAppSpec.$anonfun$new$7(IOAppSpec.scala:171)
[error] ==> X cats.effect.IOAppSpec.exit with leaked fibers  0.059s munit.ComparisonFailException: /Users/daniel/Development/Scala/cats-effect/series-3.x/ioapp-tests/src/test/scala/IOAppSpec.scala:176
[error] 175:        val h = platform("LeakedFiber", List.empty)
[error] 176:        assertEquals(h.awaitStatus(), 0)
[error] 177:      }
[error] values are not the same
[error] => Obtained
[error] 1
[error] => Diff (- obtained, + expected)
[error] -1
[error] +0
[error]     at munit.Assertions.failComparison(Assertions.scala:274)
[error] ==> X cats.effect.IOAppSpec.exit on fatal error  0.055s munit.FailException: /Users/daniel/Development/Scala/cats-effect/series-3.x/ioapp-tests/src/test/scala/IOAppSpec.scala:182 assertion failed
[error] 181:        assertEquals(h.awaitStatus(), 1)
[error] 182:        assert(h.stderr().contains("Boom!"))
[error] 183:        assert(!h.stdout().contains("sadness"))
[error]     at munit.FunSuite.assert(FunSuite.scala:11)
[error]     at cats.effect.IOAppSpec.$anonfun$new$13(IOAppSpec.scala:182)
[error] ==> X cats.effect.IOAppSpec.exit on fatal error with other unsafe runs  0.059s munit.FailException: /Users/daniel/Development/Scala/cats-effect/series-3.x/ioapp-tests/src/test/scala/IOAppSpec.scala:189 assertion failed
[error] 188:        assertEquals(h.awaitStatus(), 1)
[error] 189:        assert(h.stderr().contains("Boom!"))
[error] 190:      }
[error]     at munit.FunSuite.assert(FunSuite.scala:11)
[error]     at cats.effect.IOAppSpec.$anonfun$new$19(IOAppSpec.scala:189)
[error] ==> X cats.effect.IOAppSpec.exit on raising a fatal error with attempt  0.045s munit.FailException: /Users/daniel/Development/Scala/cats-effect/series-3.x/ioapp-tests/src/test/scala/IOAppSpec.scala:195 assertion failed
[error] 194:        assertEquals(h.awaitStatus(), 1)
[error] 195:        assert(h.stderr().contains("Boom!"))
[error] 196:        assert(!h.stdout().contains("sadness"))
[error]     at munit.FunSuite.assert(FunSuite.scala:11)
[error]     at cats.effect.IOAppSpec.$anonfun$new$23(IOAppSpec.scala:195)
[error] ==> X cats.effect.IOAppSpec.exit on raising a fatal error with handleError  0.062s munit.FailException: /Users/daniel/Development/Scala/cats-effect/series-3.x/ioapp-tests/src/test/scala/IOAppSpec.scala:202 assertion failed
[error] 201:        assertEquals(h.awaitStatus(), 1)
[error] 202:        assert(h.stderr().contains("Boom!"))
[error] 203:        assert(!h.stdout().contains("sadness"))
[error]     at munit.FunSuite.assert(FunSuite.scala:11)
[error]     at cats.effect.IOAppSpec.$anonfun$new$29(IOAppSpec.scala:202)
[error] ==> X cats.effect.IOAppSpec.exit on raising a fatal error inside a map  0.044s munit.FailException: /Users/daniel/Development/Scala/cats-effect/series-3.x/ioapp-tests/src/test/scala/IOAppSpec.scala:209 assertion failed
[error] 208:        assertEquals(h.awaitStatus(), 1)
[error] 209:        assert(h.stderr().contains("Boom!"))
[error] 210:        assert(!h.stdout().contains("sadness"))
[error]     at munit.FunSuite.assert(FunSuite.scala:11)
[error]     at cats.effect.IOAppSpec.$anonfun$new$35(IOAppSpec.scala:209)
[error] ==> X cats.effect.IOAppSpec.exit on raising a fatal error inside a flatMap  0.056s munit.FailException: /Users/daniel/Development/Scala/cats-effect/series-3.x/ioapp-tests/src/test/scala/IOAppSpec.scala:216 assertion failed
[error] 215:        assertEquals(h.awaitStatus(), 1)
[error] 216:        assert(h.stderr().contains("Boom!"))
[error] 217:        assert(!h.stdout().contains("sadness"))
[error]     at munit.FunSuite.assert(FunSuite.scala:11)
[error]     at cats.effect.IOAppSpec.$anonfun$new$41(IOAppSpec.scala:216)
[error] ==> X cats.effect.IOAppSpec.warn on global runtime collision  0.046s munit.ComparisonFailException: /Users/daniel/Development/Scala/cats-effect/series-3.x/ioapp-tests/src/test/scala/IOAppSpec.scala:222
[error] 221:        val h = platform("GlobalRacingInit", List.empty)
[error] 222:        assertEquals(h.awaitStatus(), 0)
[error] 223:        assert(
[error] values are not the same
[error] => Obtained
[error] 1
[error] => Diff (- obtained, + expected)
[error] -1
[error] +0
[error]     at munit.Assertions.failComparison(Assertions.scala:274)
[error] ==> X cats.effect.IOAppSpec.reset global runtime on shutdown  0.097s munit.ComparisonFailException: /Users/daniel/Development/Scala/cats-effect/series-3.x/ioapp-tests/src/test/scala/IOAppSpec.scala:232
[error] 231:        val h = platform("GlobalShutdown", List.empty)
[error] 232:        assertEquals(h.awaitStatus(), 0)
[error] 233:        assert(
[error] values are not the same
[error] => Obtained
[error] 1
[error] => Diff (- obtained, + expected)
[error] -1
[error] +0
[error]     at munit.Assertions.failComparison(Assertions.scala:274)
[warn] ==> i cats.effect.IOAppSpec.warn on cpu starvation ignored 0.0s
[error] ==> X cats.effect.IOAppSpec.custom runtime installed as global  0.058s munit.ComparisonFailException: /Users/daniel/Development/Scala/cats-effect/series-3.x/ioapp-tests/src/test/scala/IOAppSpec.scala:258
[error] 257:        val h = platform("CustomRuntime", List.empty)
[error] 258:        assertEquals(h.awaitStatus(), 0)
[error] 259:      }
[error] values are not the same
[error] => Obtained
[error] 1
[error] => Diff (- obtained, + expected)
[error] -1
[error] +0
[error]     at munit.Assertions.failComparison(Assertions.scala:274)
[error] ==> X cats.effect.IOAppSpec.abort awaiting shutdown hooks  0.053s munit.ComparisonFailException: /Users/daniel/Development/Scala/cats-effect/series-3.x/ioapp-tests/src/test/scala/IOAppSpec.scala:264
[error] 263:          val h = platform("ShutdownHookImmediateTimeout", List.empty)
[error] 264:          assertEquals(h.awaitStatus(), 0)
[error] 265:        }
[error] values are not the same
[error] => Obtained
[error] 1
[error] => Diff (- obtained, + expected)
[error] -1
[error] +0
[error]     at munit.Assertions.failComparison(Assertions.scala:274)
[error] ==> X cats.effect.IOAppSpec.run finalizers on TERM  10.601s munit.ComparisonFailException: /Users/daniel/Development/Scala/cats-effect/series-3.x/ioapp-tests/src/test/scala/IOAppSpec.scala:301
[error] 300:        h.term()
[error] 301:        assertEquals(h.awaitStatus(), 143)
[error] 302:
[error] values are not the same
[error] => Obtained
[error] 1
[error] => Diff (- obtained, + expected)
[error] -1
[error] +143
[error]     at munit.Assertions.failComparison(Assertions.scala:274)
[info]   + exit on fatal error without IOApp 0.044s
[info]   + exit on canceled 0.055s
[error] ==> X cats.effect.IOAppSpec.live fiber snapshot  639.814s java.lang.InterruptedException: sleep interrupted
[error]     at java.lang.Thread.sleepNanos0(Native Method)
[error]     at java.lang.Thread.sleepNanos(Thread.java:496)
[error]     at java.lang.Thread.sleep(Thread.java:527)
[error]     at cats.effect.IOAppSpec.$anonfun$new$81(IOAppSpec.scala:331)
[error] ==> X cats.effect.IOAppSpec.gracefully ignore undefined process.exit  0.005s java.lang.InterruptedException: null
[error]     at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:1696)
[error]     at java.lang.ProcessImpl.waitFor(ProcessImpl.java:425)
[error]     at scala.sys.process.ProcessImpl$SimpleProcess.exitValue(ProcessImpl.scala:266)
[error]     at cats.effect.IOAppSpec$Platform$$anon$1.awaitStatus(IOAppSpec.scala:49)
[error]     at cats.effect.IOAppSpec.$anonfun$new$105(IOAppSpec.scala:386)
Cancelled: testQuick
```

</details>

iRevive:
My best guess is that `testQuick` relies on the cached outputs. For example, `testQuick` doesn't even work on a fresh run:
```
...
[error] ==> X cats.effect.IOAppSpec.exit on canceled  0.002s java.io.IOException: Cannot run program "/Users/maksym/projects/oss/public/cats-effect/tests/native/target/scala-2.13/cats-effect-tests-out": error=2, No such file or directory
[error]     at java.lang.ProcessBuilder.start(ProcessBuilder.java:1048)
[error]     at scala.sys.process.ProcessBuilderImpl$Simple.run(ProcessBuilderImpl.scala:85)
[error]     at cats.effect.IOAppSpec$Platform.apply(IOAppSpec.scala:42)
[error]     at cats.effect.IOAppSpec.$anonfun$new$79(IOAppSpec.scala:319)
[error] Caused by: java.io.IOException: error=2, No such file or directory
[error]     at java.lang.UNIXProcess.forkAndExec(Native Method)
[error]     at java.lang.UNIXProcess.<init>(UNIXProcess.java:247)
[error]     at java.lang.ProcessImpl.start(ProcessImpl.java:134)
[error]     at java.lang.ProcessBuilder.start(ProcessBuilder.java:1029)
[error]     ... 4 more
[error] Failed: Total 16, Failed 16, Errors 0, Passed 0, Ignored 1
```

If you didn't clean the project, it may use the old cached native artifacts.

ioapp-tests/src/test/scala/IOAppSpec.scala
iRevive: Done

djspiewak:
For some reason, munit seems a lot more prone to hanging tests without timeouts in some failure cases. For example, if you remove the `shouldFinalize()` check on IOFiber.scala:700, you can reproduce tests hanging more or less indefinitely without any timeouts. I did a bunch of conditional flipping throughout the runtime and the most common result was just… hanging forever. It's not entirely clear to me why this would be the case, but obviously it would be much more ideal if the tests just timed out with a proper error.

iRevive:
I did some flipping tests before, and they were timing out correctly. 
Turns out, they work fine on Java 21:
```
[info] welcome to sbt 1.10.7 (Azul Systems, Inc. Java 21.0.4)

sbt:cats-effect> testsJVM/testOnly *IOSuite
...
[error] ==> X cats.effect.IOSuite.random racing sleeps  20.004s java.util.concurrent.TimeoutException: test timed out after 20 seconds
[error] ==> X cats.effect.IOSuite.steal timers  20.005s java.util.concurrent.TimeoutException: test timed out after 20 seconds
[error] ==> X cats.effect.IOSuite.lots of externally-canceled timers  20.005s java.util.concurrent.TimeoutException: test timed out after 20 seconds
[info]   + not lose cedeing threads from the bypass when blocker transitioning 6.615s
[info]   + wake parked thread for polled events 1.307s
[error] ==> X cats.effect.IOSuite.block in-place on virtual threads  20.003s java.util.concurrent.TimeoutException: test timed out after 20 seconds
...
```


However, CPU usage skyrocketed to 1300%, and the fans went crazy.

_____

Today I decided to use nix env, and the result was different:
```
[info] welcome to sbt 1.10.7 (Azul Systems, Inc. Java 1.8.0_422)

sbt:cats-effect> testsJVM/testOnly *IOSuite
...
[info]   + parTraverseN - throw when n < 1 0.003s
[info]   + parTraverseN - propagate errors 0.011s
...
```

Hangs forever. CPU usage is normal, though.

djspiewak:
Interesting. I'm on Java 23 and for me it was a forever hang with normal CPU.

iRevive:
Do you use Temurin by any chance?
Tested locally: 
- `Eclipse Adoptium Java 21.0.4` - hangs
- `Azul Systems, Inc. Java 21.0.4` - times out correctly 

I'm testing on Darwin Aarch64.

djspiewak:
```
openjdk 23.0.1 2024-10-15
OpenJDK Runtime Environment Temurin-23.0.1+11 (build 23.0.1+11)
OpenJDK 64-Bit Server VM Temurin-23.0.1+11 (build 23.0.1+11, mixed mode, sharing)
```

Fascinating! And somewhat terrifying. I wonder why there's such a significant difference between the distributions?

iRevive:
Seems to be related: https://github.com/scalameta/munit/issues/730.
Once I switched to the global context from parasitic, the hanging was gone:
```scala
override def munitExecutionContext: ExecutionContext = ExecutionContext.global
```

iRevive:
I did some extra testing on the Azul JVM. After several reruns of the `sbt test,` it finally hung, too.

https://github.com/typelevel/cats-effect/pull/4216/commits/65dc9ad33917e6a8f7d6b5c2d196451b897a059f - this change fixes hanging behavior on all JVMs.

djspiewak:
This change still seems to hang forever for me:

```diff
diff --git a/core/shared/src/main/scala/cats/effect/IOFiber.scala b/core/shared/src/main/scala/cats/effect/IOFiber.scala
index 9e0357b734..39c44fd37e 100644
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -697,7 +697,7 @@ private final class IOFiber[A](
                    */
                   suspend()
                 }
-              } else if (finalizing == state.wasFinalizing && !shouldFinalize() && outcome == null) {
+              } else if (finalizing == state.wasFinalizing && outcome == null) {
                 /*
                  * If we aren't canceled or completed, and we're
                  * still in the same finalization state, loop on
```

Maybe that's a unique one. I'm trying out a few more which hung previously.

iRevive:
> This change still seems to hang forever for me:
> 
> ```diff
> diff --git a/core/shared/src/main/scala/cats/effect/IOFiber.scala b/core/shared/src/main/scala/cats/effect/IOFiber.scala
> index 9e0357b734..39c44fd37e 100644
> --- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
> +++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
> @@ -697,7 +697,7 @@ private final class IOFiber[A](
>                     */
>                    suspend()
>                  }
> -              } else if (finalizing == state.wasFinalizing && !shouldFinalize() && outcome == null) {
> +              } else if (finalizing == state.wasFinalizing && outcome == null) {
>                  /*
>                   * If we aren't canceled or completed, and we're
>                   * still in the same finalization state, loop on
> ```
> 
> Maybe that's a unique one. I'm trying out a few more which hung previously.

Roughly 1 in 10 runs hangs, no matter what JVM you use. 
I think I will have a gap next week to investigate this one properly.

iRevive:
I almost unveiled the mystery. 

The [`p.tryComplete(result)`](https://github.com/typelevel/cats-effect/blob/d13efdc1e42d22280c095d4fc56117680aad145c/tests/shared/src/test/scala/cats/effect/Runners.scala#L154) is stuck in some cases.

Even though the test is **completed**, the completion callback (`onComplete`) is never invoked due to a deadlock. 
The thread pool gets exhausted, and the whole suite hangs eventually. 

_______

That's the test definition:
```scala
real("parTraverseN_ - propagate errors") {
  List(1, 2, 3)
    .parTraverseN_(2) { (n: Int) =>
      if (n == 2) IO.raiseError(new RuntimeException) else n.pure[IO]
    }
    .mustFailWith[RuntimeException]
}
```

`real` unwraps into:
```scala
test("parTraverseN_ - propagate errors") {
  val (fut, cancel) = {
    List(1, 2, 3).parTraverseN(2) { (n: Int) => ??? }.mustFailWith[RuntimePermission]
  }.unsafeToFutureCancelable()(runtime())

  val p = Promise[A]()
  val r = runtime()
  implicit val ec = r.compute

  val cancelTimer = r.scheduler.sleep(duration, { () =>
    if (p.tryFailure(new TestTimeoutException)) cancel()
  })

  f.onComplete { result =>
    if (p.tryComplete(result)) // stuck forever
      cancelTimer.run() 
  } // ec = r.compute is used here

  p.future
}
```

`test` uses [PlatformCompat.waitAtMost](https://github.com/scalameta/munit/blob/v1.0.0-M11/munit/shared/src/main/scala/munit/FunSuite.scala#L37) under the hood:
```scala
munitTestsBuffer += munitTestTransform(
  new Test(
    "parTraverseN_ - propagate errors",
    { () =>
      try {
        PlatformCompat.waitAtMost(() => {
          val (fut, cancel) = {
            List(1, 2, 3).parTraverseN(2) { (n: Int) => ??? }.mustFailWith[RuntimePermission]
          }.unsafeToFutureCancelable()(runtime())

          val p = Promise[A]()
          val r = runtime()
          implicit val ec = r.compute

          val cancelTimer = r.scheduler.sleep(duration, { () =>
            if (p.tryFailure(new TestTimeoutException)) cancel()
          })

          f.onComplete { result =>
            if (p.tryComplete(result)) // stuck forever 
              cancelTimer.run()
          } // ec = r.compute is used here

          p.future
        }, munitTimeout, munitExecutionContext)
      } catch {
        case NonFatal(e) =>
          Future.failed(e)
      }
    },
    options.tags.toSet,
    loc
  )
)
```

And if we inline all calls, the test definition is:
```scala
new Test(
  "parTraverseN_ - propagate errors",
  { () =>
    try {
      // test body
      def startFuture(): Future[Unit] = {
        val (fut, cancel) = {
          List(1, 2, 3).parTraverseN(2) { (n: Int) => ??? }.mustFailWith[RuntimePermission]
        }.unsafeToFutureCancelable()(runtime())

        val p = Promise[A]()
        val r = runtime()
        implicit val ec = r.compute

        val cancelTimer = r.scheduler.sleep(duration, { () =>
          if (p.tryFailure(new TestTimeoutException)) cancel()
        })

        f.onComplete { result =>
          if (p.tryComplete(result)) // stuck forever 
            cancelTimer.run()
        } // ec = r.compute is used here

        p.future
      }

      // comes from waitForCompletion/PlatformCompat.waitAtMost
      val onComplete = Promise[T]()
      val timeout = sh.schedule[Unit](
        () => onComplete.tryFailure(new TimeoutException(s"test timed out after $duration")),
        munitTimeout.toMillis,
        TimeUnit.MILLISECONDS
      )
      munitExecutionContext.execute(new Runnable {
        def run(): Unit = {
          startFuture().onComplete { result =>
            onComplete.tryComplete(result)
            timeout.cancel(false)
          }(munitExecutionContext)
        }
      })
      onComplete.future
    } catch {
      case NonFatal(e) =>
        Future.failed(e)
    }
  },
  options.tags.toSet,
  loc
)
```

Where `munitExecutionContext` is `parasitic` by default, but `ExecutionContext.global` hangs as well.

djspiewak:
> The [p.tryComplete(result)](https://github.com/typelevel/cats-effect/blob/d13efdc1e42d22280c095d4fc56117680aad145c/tests/shared/src/test/scala/cats/effect/Runners.scala#L154) is stuck in some cases.

Well… now *that's* a puzzler. Huh.

djspiewak:
So when it's hanging, are you able to get a thread dump? Or is it asynchronously hanging? If asynchronous, can you pull the internal queue state from the executor used by munit for its futures by dumping the heap? `parasitic` would be the easiest one to introspect.

I stared at the implementation of `tryComplete` for a bit and the only way I can think of this hanging is if some downstream callback is either blocking a thread (which will show up on a heap dump) or scheduling a continuation which never gets picked up (which would show up on the EC), so that feels like the right place to start.

iRevive:
The state of the [single-threaded executor](https://github.com/scalameta/munit/blob/v1.0.0-M11/munit/jvm/src/main/scala/munit/internal/PlatformCompat.scala#L22-L32) that is responsible for the [timeouts](https://github.com/scalameta/munit/blob/v1.0.0-M11/munit/jvm/src/main/scala/munit/internal/PlatformCompat.scala#L59-L66):
```
java.util.concurrent.ScheduledThreadPoolExecutor@718be7e9[Running, pool size = 1, active threads = 1, queued tasks = 92, completed tasks = 233]
```

<details>
  <summary>The thread dump</summary>
  
  ```javascript
"munit-scheduler-1" #39 [39683] daemon prio=5 os_prio=31 cpu=172626.79ms elapsed=319.40s tid=0x0000000145bd3600 nid=39683 runnable  [0x000000017471a000]
   java.lang.Thread.State: RUNNABLE
        at cats.effect.IOFiber.loop$1(IOFiber.scala:707)
        at cats.effect.IOFiber.stateLoop$1(IOFiber.scala:747)
        at cats.effect.IOFiber.$anonfun$runLoop$1(IOFiber.scala:753)
        at cats.effect.IOFiber.$anonfun$runLoop$1$adapted(IOFiber.scala:643)
        at cats.effect.IOFiber$$Lambda/0x000000a001238208.apply(Unknown Source)
        at cats.effect.IOFiber.$anonfun$runLoop$14(IOFiber.scala:971)
        at cats.effect.IOFiber$$Lambda/0x000000a00138a580.run(Unknown Source)
        at cats.effect.kernel.testkit.TestContext.tickOne(TestContext.scala:161)
        at cats.effect.kernel.testkit.TestContext.tick(TestContext.scala:177)
        at cats.effect.kernel.testkit.TestContext.tickAll(TestContext.scala:199)
        at cats.effect.testkit.TestInstances.unsafeRun(TestInstances.scala:211)
        at cats.effect.testkit.TestInstances.unsafeRun$(TestInstances.scala:202)
        at cats.effect.IOSuite.unsafeRun(IOSuite.scala:38)
        at cats.effect.testkit.TestInstances.$anonfun$orderIoFiniteDuration$1(TestInstances.scala:161)
        at cats.effect.testkit.TestInstances$$Lambda/0x000000a0011b4a30.apply(Unknown Source)
        at cats.kernel.Order$.cats$kernel$Order$$$anonfun$by$1(Order.scala:159)
        at cats.kernel.Order$$anonfun$by$2.compare(Order.scala:159)
        at cats.kernel.Order.lteqv(Order.scala:93)
        at cats.kernel.Order.lteqv$(Order.scala:92)
        at cats.kernel.Order$$anonfun$by$2.lteqv(Order.scala:159)
        at cats.syntax.PartialOrderOps.$less$eq(partialOrder.scala:34)
        at cats.effect.laws.IsEq.toPropTolerant(IsEq.scala:39)
        at cats.effect.laws.IsEq.toPropTolerant$(IsEq.scala:27)
        at cats.effect.laws.IsEq$Assert.toPropTolerant(IsEq.scala:83)
        at cats.effect.laws.IsEq$.toPropTolerant(IsEq.scala:78)
        at cats.effect.laws.GenTemporalTests$$anon$2.$anonfun$props$20(GenTemporalTests.scala:223)
        at cats.effect.laws.GenTemporalTests$$anon$2$$Lambda/0x000000a0012585b0.apply(Unknown Source)
        at scala.Function1.$anonfun$andThen$1(Function1.scala:87)
        at scala.Function1$$Lambda/0x000000a0011f9190.apply(Unknown Source)
        at org.scalacheck.Prop$.$anonfun$forAllShrink$2(Prop.scala:868)
        at org.scalacheck.Prop$$$Lambda/0x000000a001385c00.apply(Unknown Source)
        at org.scalacheck.Prop$.secure(Prop.scala:484)
        at org.scalacheck.Prop$.result$1(Prop.scala:868)
        at org.scalacheck.Prop$.$anonfun$forAllShrink$1(Prop.scala:915)
        at org.scalacheck.Prop$$$Lambda/0x000000a0011f98a8.apply(Unknown Source)
        at org.scalacheck.Prop$.$anonfun$apply$1(Prop.scala:305)
        at org.scalacheck.Prop$$$Lambda/0x000000a0011f9e78.apply(Unknown Source)
        at org.scalacheck.PropFromFun.apply(Prop.scala:20)
        at org.scalacheck.Prop$.result$1(Prop.scala:869)
        at org.scalacheck.Prop$.$anonfun$forAllShrink$1(Prop.scala:915)
        at org.scalacheck.Prop$$$Lambda/0x000000a0011f98a8.apply(Unknown Source)
        at org.scalacheck.Prop$.$anonfun$apply$1(Prop.scala:305)
        at org.scalacheck.Prop$$$Lambda/0x000000a0011f9e78.apply(Unknown Source)
        at org.scalacheck.PropFromFun.apply(Prop.scala:20)
        at org.scalacheck.Prop$.$anonfun$delay$1(Prop.scala:491)
        at org.scalacheck.Prop$$$Lambda/0x000000a0012a8c78.apply(Unknown Source)
        at org.scalacheck.Prop$.$anonfun$apply$1(Prop.scala:305)
        at org.scalacheck.Prop$$$Lambda/0x000000a0011f9e78.apply(Unknown Source)
        at org.scalacheck.PropFromFun.apply(Prop.scala:20)
        at org.scalacheck.Prop.$anonfun$viewSeed$1(Prop.scala:39)
        at org.scalacheck.Prop$$Lambda/0x000000a0012a9048.apply(Unknown Source)
        at org.scalacheck.Prop$.$anonfun$apply$1(Prop.scala:305)
        at org.scalacheck.Prop$$$Lambda/0x000000a0011f9e78.apply(Unknown Source)
        at org.scalacheck.PropFromFun.apply(Prop.scala:20)
        at munit.ScalaCheckSuite.$anonfun$propToTry$2(ScalaCheckSuite.scala:63)
        at munit.ScalaCheckSuite$$Lambda/0x000000a00133ad98.apply(Unknown Source)
        at org.scalacheck.Prop$.$anonfun$apply$1(Prop.scala:305)
        at org.scalacheck.Prop$$$Lambda/0x000000a0011f9e78.apply(Unknown Source)
        at org.scalacheck.PropFromFun.apply(Prop.scala:20)
        at org.scalacheck.Test$.workerFun$1(Test.scala:486)
        at org.scalacheck.Test$.$anonfun$check$4(Test.scala:518)
        at org.scalacheck.Test$.$anonfun$check$4$adapted(Test.scala:518)
        at org.scalacheck.Test$$$Lambda/0x000000a00133b7f0.apply(Unknown Source)
        at org.scalacheck.Platform$.runWorkers(Platform.scala:42)
        at org.scalacheck.Test$.check(Test.scala:518)
        at munit.ScalaCheckSuite.propToTry(ScalaCheckSuite.scala:62)
        at munit.ScalaCheckSuite.$anonfun$scalaCheckPropTransform$3(ScalaCheckSuite.scala:45)
        at munit.ScalaCheckSuite$$Lambda/0x000000a001327170.apply(Unknown Source)
        at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:475)
        at munit.Suite$$anon$1.execute(Suite.scala:28)
        at scala.concurrent.impl.Promise$Transformation.submitWithValue(Promise.scala:429)
        at scala.concurrent.impl.Promise$DefaultPromise.submitWithValue(Promise.scala:338)
        at scala.concurrent.impl.Promise$DefaultPromise.dispatchOrAddCallbacks(Promise.scala:312)
        at scala.concurrent.impl.Promise$DefaultPromise.transform(Promise.scala:128)
        at munit.internal.FutureCompat$ExtensionFuture.transformCompat(FutureCompat.scala:16)
        at munit.ScalaCheckSuite.$anonfun$scalaCheckPropTransform$2(ScalaCheckSuite.scala:47)
        at munit.ScalaCheckSuite$$Lambda/0x000000a0010d4400.apply(Unknown Source)
        at munit.Test.$anonfun$withBodyMap$1(Test.scala:34)
        at munit.Test$$Lambda/0x000000a0010d8000.apply(Unknown Source)
        at munit.internal.console.StackTraces$.dropOutside(StackTraces.scala:12)
        at munit.MUnitRunner.$anonfun$runTestBody$1(MUnitRunner.scala:351)
        at munit.MUnitRunner$$Lambda/0x000000a001323488.apply(Unknown Source)
        at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:470)
        at munit.Suite$$anon$1.execute(Suite.scala:28)
        at scala.concurrent.impl.Promise$Transformation.submitWithValue(Promise.scala:429)
        at scala.concurrent.impl.Promise$DefaultPromise.submitWithValue(Promise.scala:338)
        at scala.concurrent.impl.Promise$DefaultPromise.dispatchOrAddCallbacks(Promise.scala:312)
        at scala.concurrent.impl.Promise$DefaultPromise.flatMap(Promise.scala:176)
        at munit.MUnitRunner.runTestBody(MUnitRunner.scala:347)
        at munit.MUnitRunner.runTest(MUnitRunner.scala:327)
        at munit.MUnitRunner.$anonfun$runTests$1(MUnitRunner.scala:140)
        at munit.MUnitRunner$$Lambda/0x000000a0013210b0.apply(Unknown Source)
        at scala.collection.Iterator$$anon$9.next(Iterator.scala:584)
        at munit.MUnitRunner.loop$1(MUnitRunner.scala:155)
        at munit.MUnitRunner.$anonfun$sequenceFutures$1(MUnitRunner.scala:164)
        at munit.MUnitRunner$$Lambda/0x000000a0013cdfd0.apply(Unknown Source)
        at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:470)
        at munit.Suite$$anon$1.execute(Suite.scala:28)
        at scala.concurrent.impl.Promise$Transformation.submitWithValue(Promise.scala:429)
        at scala.concurrent.impl.Promise$DefaultPromise.submitWithValue(Promise.scala:338)
        at scala.concurrent.impl.Promise$DefaultPromise.tryComplete0(Promise.scala:285)
        at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:504)
        at munit.Suite$$anon$1.execute(Suite.scala:28)
        at scala.concurrent.impl.Promise$Transformation.submitWithValue(Promise.scala:429)
        at scala.concurrent.impl.Promise$DefaultPromise.submitWithValue(Promise.scala:338)
        at scala.concurrent.impl.Promise$DefaultPromise.tryComplete0(Promise.scala:285)
        at scala.concurrent.impl.Promise$DefaultPromise.linkRootOf(Promise.scala:347)
        at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:492)
        at munit.Suite$$anon$1.execute(Suite.scala:28)
        at scala.concurrent.impl.Promise$Transformation.submitWithValue(Promise.scala:429)
        at scala.concurrent.impl.Promise$DefaultPromise.submitWithValue(Promise.scala:338)
        at scala.concurrent.impl.Promise$DefaultPromise.tryComplete0(Promise.scala:285)
        at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:504)
        at munit.Suite$$anon$1.execute(Suite.scala:28)
        at scala.concurrent.impl.Promise$Transformation.submitWithValue(Promise.scala:429)
        at scala.concurrent.impl.Promise$DefaultPromise.submitWithValue(Promise.scala:338)
        at scala.concurrent.impl.Promise$DefaultPromise.tryComplete0(Promise.scala:285)
        at scala.concurrent.impl.Promise$DefaultPromise.linkRootOf(Promise.scala:347)
        at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:478)
        at munit.Suite$$anon$1.execute(Suite.scala:28)
        at scala.concurrent.impl.Promise$Transformation.submitWithValue(Promise.scala:429)
        at scala.concurrent.impl.Promise$DefaultPromise.submitWithValue(Promise.scala:338)
        at scala.concurrent.impl.Promise$DefaultPromise.tryComplete0(Promise.scala:285)
        at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:504)
        at munit.Suite$$anon$1.execute(Suite.scala:28)
        at scala.concurrent.impl.Promise$Transformation.submitWithValue(Promise.scala:429)
        at scala.concurrent.impl.Promise$DefaultPromise.submitWithValue(Promise.scala:338)
        at scala.concurrent.impl.Promise$DefaultPromise.tryComplete0(Promise.scala:285)
        at scala.concurrent.impl.Promise$DefaultPromise.tryComplete(Promise.scala:278)
        at scala.concurrent.Promise.tryFailure(Promise.scala:117)
        at scala.concurrent.Promise.tryFailure$(Promise.scala:117)
        at scala.concurrent.impl.Promise$DefaultPromise.tryFailure(Promise.scala:104)
        at munit.internal.PlatformCompat$.$anonfun$waitAtMost$2(PlatformCompat.scala:87)
        at munit.internal.PlatformCompat$.$anonfun$waitAtMost$2$adapted(PlatformCompat.scala:82)
        at munit.internal.PlatformCompat$$$Lambda/0x000000a001323b18.call(Unknown Source)
        at java.util.concurrent.FutureTask.run(java.base@21.0.4/FutureTask.java:317)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(java.base@21.0.4/ScheduledThreadPoolExecutor.java:304)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(java.base@21.0.4/ThreadPoolExecutor.java:1144)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(java.base@21.0.4/ThreadPoolExecutor.java:642)
        at java.lang.Thread.runWith(java.base@21.0.4/Thread.java:1596)
        at java.lang.Thread.run(java.base@21.0.4/Thread.java:1583)

   Locked ownable synchronizers:
        - <0x00000006008021a0> (a java.util.concurrent.ThreadPoolExecutor$Worker)
  ```
  
</details>


So, IO execution leaks to the munit scheduler:
```
"munit-scheduler-1" #39 [39683] daemon prio=5 os_prio=31 cpu=172626.79ms elapsed=319.40s tid=0x0000000145bd3600 nid=39683 runnable  [0x000000017471a000]
 java.lang.Thread.State: RUNNABLE
      at cats.effect.IOFiber.loop$1(IOFiber.scala:707)
      at cats.effect.IOFiber.stateLoop$1(IOFiber.scala:747)
      at cats.effect.IOFiber.$anonfun$runLoop$1(IOFiber.scala:753)
```

And as a result the cancelation doesn't work anymore and the whole suite is stuck.

armanbilge:
Thanks for investigating!

> So, IO execution leaks to the munit scheduler:

Actually, it's not the execution (read: run loop) that is running on the munit scheduler, just the spinloop that is used to reschedule an `IOFiber` that is asynchronously suspended.

https://github.com/typelevel/cats-effect/blob/917d5548ddd86864a39ed949a3085188bc421a8c/core/shared/src/main/scala/cats/effect/IOFiber.scala#L652-L663

djspiewak:
@iRevive So this is really strange because it suggests that we're sequencing `get` somewhere between line 787 and line 818. This is why it's a spin wait: unless we crash we're going to get out of there quickly.

Which actually does raise an interesting question: perhaps we got an exception from `ByteStack` or similar?

djspiewak:
Oh actually I think that might make sense, given the single threaded nature of hte runtime and the nature of the change. My guess is this probably would have happened on specs2 as well.

So let's maybe put that one to bed. Popping stack back up, is there anything else that's concerning here? The rest of it seems fine to me.

djspiewak:
Okay I'm calling "TTL" on my previous comment. I think this is ready to go aside from the merge conflicts.

iRevive:
~~Now the most interesting part - ensure that I didn't wipe some tests accidentally 😅~~

Everything seems to be in place.

djspiewak (approved):

PR #4220, opened 2024-12-30
Add `UUIDGen` inductive materializers for monad transformers

bio-aeon:
This PR adds the `UUIDGen` materializer for `OptionT`.

UPD: Added materializers also for other monad transformers.

### DIFF
--- a/std/jvm/src/main/scala/cats/effect/std/UUIDGenCompanionPlatform.scala
+++ b/std/jvm/src/main/scala/cats/effect/std/UUIDGenCompanionPlatform.scala
@@ -16,8 +16,16 @@
 
 package cats.effect.std
 
-import cats.Functor
-import cats.data.OptionT
+import cats.{Applicative, Functor, Monoid}
+import cats.data.{
+  EitherT,
+  IndexedReaderWriterStateT,
+  IndexedStateT,
+  IorT,
+  Kleisli,
+  OptionT,
+  WriterT
+}
 import cats.effect.kernel.Sync
 
 import java.util.UUID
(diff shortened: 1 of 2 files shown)


std/jvm/src/main/scala/cats/effect/std/UUIDGenCompanionPlatform.scala
```diff
@@ -33,4 +43,62 @@ private[std] trait UUIDGenCompanionPlatformLowPriority {
       ev.blocking(UUID.randomUUID())
   }
 
+  /**
+   * [[UUIDGen]] instance built for `cats.data.EitherT` values initialized with any `F` data
+   * type that also implements `UUIDGen`.
+   */
+  implicit def catsEitherTUUIDGen[F[_]: UUIDGen: Functor, L]: UUIDGen[EitherT[F, L, *]] =
```
armanbilge: I don't think these need to be platform-specific to the JVM, we can put these on the `UUIDGen` companion object.

bio-aeon: Good point. Moved.

std/shared/src/main/scala/cats/effect/std/UUIDGen.scala
```diff
@@ -59,4 +137,11 @@ object UUIDGen extends UUIDGenCompanionPlatform {
         new UUID(msb, lsb)
       }
     }
+
+  private[std] abstract class TranslatedUUIDGen[F[_], G[_]](self: UUIDGen[F])(f: F ~> G)
```
armanbilge: This can probably be a `final` class right?

```suggestion
  private[std] final class TranslatedUUIDGen[F[_], G[_]](self: UUIDGen[F])(f: F ~> G)
```

armanbilge (dismissed):
Thanks!

@djspiewak is this something we'd be comfortable slipping into 3.6, even though it is source-breaking? This helps http4s address the new deprecation warnings from https://github.com/typelevel/cats-effect/pull/3647.

std/shared/src/main/scala/cats/effect/std/UUIDGen.scala
bio-aeon: Yeah, I agree with you. To be honest, not sure why the similar one `Random.TranslatedRandom` is abstract.

armanbilge (approved):

djspiewak:
@armanbilge I think this is fine for 3.6. The rules are really more like *guidelines*, anyway.

std/shared/src/main/scala/cats/effect/std/UUIDGen.scala
armanbilge: > To be honest, not sure why the similar one `Random.TranslatedRandom` is abstract.

I'm not entirely sure either, but probably related to this.

https://github.com/typelevel/cats-effect/blob/0cb794be902b39d0e43ae0ad3a44e10c94edabff/std/shared/src/main/scala/cats/effect/std/SecureRandom.scala#L43-L44

armanbilge:
🤦 I forgot to change the merge branch. @djspiewak maybe we can just fast-forward `series/3.6.x` to `series/3.x` ? There's nothing here yet that shouldn't go into 3.6.

djspiewak:
Yeah I think that's fine. We definitely haven't brought in anything yet which is breaking.

armanbilge:
Thanks. Because of the way repo permissions are setup, I think only you can do that.

PR #4222, opened 2025-01-02
Add doc for `IOApp#pollingSystem`

armanbilge:


core/jvm/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -165,6 +165,13 @@ trait IOApp {
    */
   protected def runtimeConfig: unsafe.IORuntimeConfig = unsafe.IORuntimeConfig()
 
+  /**
+   * The [[unsafe.PollingSystem]] used by the [[runtime]] which will evaluate the [[IO]]
+   * produced by `run`. It is very unlikely that users will need to override this method.
```
djspiewak, line 170: Not to bikeshed, but wouldn't it be better to talk about how the polling system specifically handles the syscall management for the scheduler? As worded, this seems to handwave at the entire WSTP + IOFiber semantic space, which obviously polling systems don't touch.

djspiewak (changes requested):

core/jvm/src/main/scala/cats/effect/IOApp.scala
armanbilge: > As worded, this seems to handwave at the entire WSTP + IOFiber semantic space

LOL yes this needs improvement. It's a couple iterations of copy pasta 🤪 

---

> Not to bikeshed, but wouldn't it be better to talk about how the polling system specifically handles the syscall management for the scheduler?

What exactly do you have in mind? Really only io_uring comes close to fully taking over syscall management.

djspiewak (approved):

PR #4224, opened 2025-01-03
Delete .envrc

armanbilge:
Sorry to anyone who finds this useful, but unfortunately it's been more annoying than helpful for @djspiewak's and my dev workflows.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,1 +1,0 @@
-use flake


valencik (approved):
The few folks that use it can add it back and git ignore it in their setup

PR #4227, opened 2025-01-06
Always report timers/events when worker unparked

### LINKED ISSUE #4225: I/O polling hangs when `computeWorkerThreadCount = 1`
First discovered this when debugging Native multithreading. It's also affecting CE 3.6.0-RC1 on the JVM, so it must be a WSTP / polling system integration bug.

```scala
//> using dep org.typelevel::cats-effect::3.6.0-RC1

import cats.effect.*
import java.nio.ByteBuffer
import java.nio.channels.Pipe
import java.nio.channels.SelectionKey._

object Bug extends IOApp.Simple {
  override def computeWorkerThreadCount = 1 // changing to 2 fixes

  def run = mkPipe.use { pipe =>
    for {
      selector <- getSelector
      buf <- IO(ByteBuffer.allocate(4))
      _ <- IO(pipe.sink.write(ByteBuffer.wrap(Array(1, 2, 3)))).background
        .surround {
          selector.select(pipe.source, OP_READ) *> IO(pipe.source.read(buf))
        }
      _ <- IO(pipe.sink.write(ByteBuffer.wrap(Array(42)))).background.surround {
        selector.select(pipe.source, OP_READ) *> IO(pipe.source.read(buf))
      }
    } yield assert(buf.array().toList == List[Byte](1, 2, 3, 42))
  }.timeout(1.second)

  def getSelector: IO[Selector] =
    IO.pollers
      .map(_.collectFirst { case selector: Selector => selector })
      .map(_.get)

  def mkPipe: Resource[IO, Pipe] =
    Resource
      .eval(getSelector)
      .flatMap { selector =>
        Resource.make(IO(selector.provider.openPipe())) { pipe =>
          IO(pipe.sink().close()).guarantee(IO(pipe.source().close()))
        }
      }
      .evalTap { pipe =>
        IO {
          pipe.sink().configureBlocking(false)
          pipe.source().configureBlocking(false)
        }
      }
}
```

armanbilge:
Fixes https://github.com/typelevel/cats-effect/issues/4225.

A parked worker may wake up for various reasons:
1. expired timer or polling event
2. external interruption / unparking, due to availability of external work
3. spurious wakeup

Due to race conditions, a worker may actually wake up for multiple reasons simultaneously. Thus, it should always check first if there are expired timers or polling events, so it can transition to handling those, before considering other cases.

---

In https://github.com/typelevel/cats-effect/issues/4225, we were able to detect this bug because the WSTP had only one worker.
1. the I/O event completes and the callback reschedules the associated fiber
2. while rescheduling the fiber, the worker notifies the pool that there is work available to steal
3. the pool unparks the one-and-only worker (another thing we should consider tweaking: don't unpark thyself)
4. the worker believes it was unparked due to external work (there is none) and meanwhile doesn't transition to handling its actual local work
5. the worker goes back to sleep until the next timer fires, at which point it properly transitions to handling local work

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
@@ -474,21 +474,18 @@ private[effect] final class WorkerThread[P <: AnyRef](
               pool.shutdown()
               false // we know `done` is `true`
             } else {
-              if (parked.get()) {
-                // we were either awakened spuriously, or we timed out or polled an event
-                if (polled || (triggerTime - now <= 0)) {
-                  // we timed out or polled an event
-                  if (parked.getAndSet(false)) {
-                    pool.doneSleeping()
-                  }
-                  true
-                } else {
-                  // awakened spuriously, re-check next sleeper
-                  parkUntilNextSleeper()
+              // no matter why we woke up, there may be timers or events ready
+              if (polled || (triggerTime - now <= 0)) {
+                // we timed out or polled an event
+                if (parked.getAndSet(false)) {
+                  pool.doneSleeping()
                 }
-              } else {
-                // awakened intentionally
-                false
+                true
+              } else { // we were either awakened spuriously or intentionally
+                if (parked.get()) // awakened spuriously, re-check next sleeper
+                  parkUntilNextSleeper()
+                else // awakened intentionally, but not due to a timer or event
+                  false
               }
             }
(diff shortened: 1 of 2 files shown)


armanbilge:
There's possibly a bigger discussion here, which is whether the fact that polling systems schedule tasks on a worker's local queue while the worker is parked violates some invariant of our runtime 🤷

djspiewak (approved):
Nice catch!

PR #4230, opened 2025-01-07
Introduce `PollResult`, `PollingSystem#processReadyEvents`

### LINKED ISSUE #4229: Workers shouldn't schedule local tasks while parked
> There's possibly a bigger discussion here, which is whether the fact that polling systems schedule tasks on a worker's local queue while the worker is parked violates some invariant of our runtime 🤷

_Originally posted by @armanbilge in https://github.com/typelevel/cats-effect/issues/4227#issuecomment-2572486150_
            
Daniel S and I talked about this on Discord, and we decided we do want to avoid scheduling tasks while a worker is parked. I'm adjusting the `PollingSystem` API so it doesn't have to work this way ...

armanbilge:
Fixes https://github.com/typelevel/cats-effect/issues/4229.

`PollingSystem#poll` is now split into `poll` and `processReadyEvents`, where `poll` does the blocking syscall (and nothing more) and `processReadyEvents` handles rescheduling of fibers on the runtime. This enables the `WorkerThread` to unpark after `poll` and before `processReadyEvents`, so that it is in the appropriate state for receiving new work on its local queue.

### DIFF
--- a/core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
+++ b/core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
@@ -72,6 +72,8 @@ abstract class PollingSystem {
   def closePoller(poller: Poller): Unit
 
   /**
+   * Blocks the thread until an event is polled, the timeout expires, or interrupted.
+   *
    * @param poller
    *   the thread-local [[Poller]] used to poll events.
    *
@@ -79,18 +81,28 @@ abstract class PollingSystem {
    *   the maximum duration for which to block, where `nanos == -1` indicates to block
    *   indefinitely.
    *
-   * @param reportFailure
-   *   callback that handles any failures that occur during polling.
+   * @return
+   *   whether any events are ready. e.g. if the method returned due to timeout, this should be
+   *   `false`. If `true`, [[processReadyEvents]] must be called before calling any other method
+   *   on this poller.
+   */
+  def poll(poller: Poller, nanos: Long): Boolean
+
+  /**
+   * Processes ready events e.g. collects their results and resumes the corresponding tasks.
+   * This method should only be called after [[poll]] and only if it returned `true`.
+   *
+   * @param poller
+   *   the thread-local [[Poller]] with ready events
    *
    * @return
-   *   whether any events were polled. e.g. if the method returned due to timeout, this should
-   *   be `false`.
+   *   whether any of the ready events caused tasks to be rescheduled on the runtime
    */
-  def poll(poller: Poller, nanos: Long, reportFailure: Throwable => Unit): Boolean
+  def processReadyEvents(poller: Poller): Boolean
 
   /**
    * @return
-   *   whether poll should be called again (i.e., there are more events to be polled)
+   *   whether [[poll]] should be called again (i.e., there are more events to be polled)
    */
   def needsPoll(poller: Poller): Boolean
 
--- a/core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
+++ b/core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
@@ -82,15 +82,14 @@ abstract class PollingSystem {
    *   indefinitely.
    *
    * @return
-   *   whether any events are ready. e.g. if the method returned due to timeout, this should be
-   *   `false`. If `true`, [[processReadyEvents]] must be called before calling any other method
-   *   on this poller.
+   *   whether any ready events were polled and should be handled with [[processReadyEvents]].
+   *   If result is incomplete, then [[poll]] should be called again after
+   *   [[processReadyEvents]].
    */
-  def poll(poller: Poller, nanos: Long): Boolean
+  def poll(poller: Poller, nanos: Long): PollResult
 
   /**
    * Processes ready events e.g. collects their results and resumes the corresponding tasks.
-   * This method should only be called after [[poll]] and only if it returned `true`.
    *
    * @param poller
    *   the thread-local [[Poller]] with ready events
@@ -148,4 +147,24 @@ object PollingSystem {
   type WithPoller[P] = PollingSystem {
     type Poller = P
   }
+}
+
+sealed abstract class PollResult
+object PollResult {
+
+  /**
+   * Polled all of the available ready events.
+   */
+  case object Complete extends PollResult
+
+  /**
+   * Polled some, but not all, of the available ready events. Poll should be called again to
+   * reap additional ready events.
+   */
+  case object Incomplete extends PollResult
+
+  /**
+   * The poll was interrupted or timed out before any events became ready.
+   */
+  case object Interrupted extends PollResult
 }
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
@@ -421,24 +421,34 @@ private[effect] final class WorkerThread[P <: AnyRef](
       }
     }
 
+    @tailrec
+    def drainReadyEvents(result: PollResult, acc: Boolean): Boolean =
+      if (result ne PollResult.Interrupted) {
+        val tasksScheduled = system.processReadyEvents(_poller) | acc
+        if (result eq PollResult.Complete) tasksScheduled
+        else drainReadyEvents(system.poll(_poller, 0), tasksScheduled)
+      } else {
+        acc
+      }
+
     // returns true if polled event, false if unparked
     def parkLoop(): Boolean = {
       while (!done.get()) {
         // Park the thread until further notice.
         val start = System.nanoTime()
         metrics.incrementPolledCount()
-        val polled = system.poll(_poller, -1)
+        val pollResult = system.poll(_poller, -1)
         now = System.nanoTime() // update now
         metrics.addIdleTime(now - start)
 
         // the only way we can be interrupted here is if it happened *externally* (probably sbt)
         if (isInterrupted()) {
           pool.shutdown()
-        } else if (polled) {
+        } else if (pollResult ne PollResult.Interrupted) {
           if (parked.getAndSet(false))
             pool.doneSleeping()
           // TODO, if no tasks scheduled could fastpath back to park?
-          val _ = system.processReadyEvents(_poller)
+          val _ = drainReadyEvents(pollResult, false)
           return true
         } else if (!parked.get()) { // Spurious wakeup check.
           return false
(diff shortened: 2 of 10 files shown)


core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
```diff
@@ -72,25 +72,37 @@ abstract class PollingSystem {
...
+   * @return
+   *   whether any events are ready. e.g. if the method returned due to timeout, this should be
+   *   `false`. If `true`, [[processReadyEvents]] must be called before calling any other method
+   *   on this poller.
+   */
+  def poll(poller: Poller, nanos: Long): Boolean
+
+  /**
+   * Processes ready events e.g. collects their results and resumes the corresponding tasks.
+   * This method should only be called after [[poll]] and only if it returned `true`.
```
djspiewak:
```suggestion
   * This method should only be called after [[poll]] and only if it returned `true`, though it is valid to call it at any time. If no events are outstanding and ready for scheduling, calling this method should be a no-op.
```

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
```diff
@@ -72,25 +72,37 @@ abstract class PollingSystem {
...
+   * @param poller
+   *   the thread-local [[Poller]] with ready events
    *
    * @return
-   *   whether any events were polled. e.g. if the method returned due to timeout, this should
-   *   be `false`.
+   *   whether any of the ready events caused tasks to be rescheduled on the runtime
    */
-  def poll(poller: Poller, nanos: Long, reportFailure: Throwable => Unit): Boolean
+  def processReadyEvents(poller: Poller): Boolean
```
djspiewak, line 100: Fascinating that we can just totally ditch the reporter. Makes sense though since we're shifting the invocation burden back to the worker.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -579,7 +583,9 @@ private[effect] final class WorkerThread[P <: AnyRef](
           sleepers.packIfNeeded()
           // give the polling system a chance to discover events
           metrics.incrementPolledCount()
-          system.poll(_poller, 0, reportFailure)
+          if (system.needsPoll(_poller) && system.poll(_poller, 0)) {
+            val _ = system.processReadyEvents(_poller)
```
djspiewak: Do we need to `try`/`catch` this? Also are we okay with the fact that an error in one callback will torpedo the whole invocation?

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -480,6 +481,9 @@ private[effect] final class WorkerThread[P <: AnyRef](
                 if (parked.getAndSet(false)) {
                   pool.doneSleeping()
                 }
+                if (polled) { // TODO, if no tasks scheduled and no timers could fastpath back to park?
+                  val _ = system.processReadyEvents(_poller)
```
djspiewak: Same comment as below.

core/native/src/main/scala/cats/effect/unsafe/KqueueSystem.scala
```diff
@@ -138,10 +141,10 @@ object KqueueSystem extends PollingSystem {
...
   final class Poller private[KqueueSystem] (kqfd: Int) {
 
-    private[this] val changelistArray = new Array[Byte](sizeof[kevent64_s].toInt * MaxEvents)
-    @inline private[this] def changelist =
-      changelistArray.atUnsafe(0).asInstanceOf[Ptr[kevent64_s]]
-    private[this] var changeCount = 0
+    private[this] val buffer = new Array[Byte](sizeof[kevent64_s].toInt * MaxEvents)
+    @inline private[this] def eventlist =
+      buffer.atUnsafe(0).asInstanceOf[Ptr[kevent64_s]]
+    private[this] var eventCount = 0
```
djspiewak: Holy incoming merge conflicts…

core/native/src/main/scala/cats/effect/unsafe/KqueueSystem.scala
```diff
@@ -231,13 +187,62 @@ object KqueueSystem extends PollingSystem {
...
+        kqfd,
+        eventlist,
+        eventCount,
+        eventlist,
+        MaxEvents,
+        flags.toUInt,
+        timeoutSpec
+      )
+
+      if (rtn >= 0) {
```
djspiewak, line 201: Note to our future selves: we'll need to do some post-processing here to remove `EV_USER`.

djspiewak (changes requested):

djspiewak:
Btw this is a much better name than what I came up with

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
armanbilge: I realized we never actually needed it. The `PollingSystem` must already have a way of rescheduling work on the runtime, which implies it has an `ExecutionContext` and therefore a `reportFailure`.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
armanbilge: > Also are we okay with the fact that an error in one callback will torpedo the whole invocation?

This is not necessarily true—it's up to the `PollingSystem`. For example, none of our current implementations will do this. We can mandate a behavior in the specification.

djspiewak: You mean none of our current implementations catch errors in the callbacks?

armanbilge: Sorry, none of current implementations will torpedo, because all of our implementations are based on `IO.async` whose callbacks never throw anyway.

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
armanbilge: So this is an interesting discussion. If this method is called, and `poll` and there are currently no events outstanding, instead of no-opping, should it make a syscall to reap for events? The implementations already do syscalls sometimes, to ensure the ready list is fully drained.

In a sense we could have `poll` and "`pollIfNeededAndProcessReadyEvents`". The former is only needed in contexts when we are parked, but the latter can be used e.g. in State 0 of the worker loop.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
djspiewak: Yeah that's true. I think it's still better to toss a try/catch around the whole thing in `WorkerThread`, just to be pessimistically safe. Also I think it's okay that a callback can fail out the whole invocation (of `processReadyEvents`), but we should probably re-do the call in the event that we hit an error so that we don't have orphaned callbacks (e.g. if we have five callbacks and the second one errors, we need to reinvoke to ensure the remaining callbacks are run).

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
djspiewak: IMO better to keep it simple in this API layer. The two steps really are orthogonal, and as we've seen, splitting them apart gives `WorkerThread` more flexibility to inject relevant semantics.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
armanbilge: > I think it's still better to toss a try/catch around the whole thing in `WorkerThread`, just to be pessimistically safe.

The polling system syscall can error itself, once we `catch` those what should we do? It's effectively a fatal error, and it can from `processReadyEvents` too, since that may rely syscalls to drain the ready list. In those cases it would not be safe to re-invoke.

djspiewak: That's going to be an exception thrown from `poll` though, no? If we want to treat that as a fatal error then we can encode that semantic in `WorkerThread` and just shut down the whole runtime.

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
armanbilge: > The two steps really are orthogonal

That's my point, currently they aren't as orthogonal as they seem, because `processReadyEvents` is still making polling syscalls to drain the ready list.

Maybe what we need to do is disallow `processReadyEvents` from making syscalls at all, and then the WSTP has to loop between `poll` and `processReadyEvents` to achieve this?

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
armanbilge: > That's going to be an exception thrown from `poll` though, no?

No:

> it can from `processReadyEvents` too, since that may rely syscalls to drain the ready list. In those cases it would not be safe to re-invoke.

djspiewak: What's a case where `processReadyEvents` would use syscalls to drain the ready list? In my mental model, `poll` should handle the burden of shifting things into userspace, so `processReadyEvents` should be syscall-free.

armanbilge: When we call epoll/kqueue/whatever, we hand them a fixed size buffer to populate with ready events. If that buffer comes back entirely full, chances are there are additional ready events still in the kernel. So we need to process this first batch and then poll again, to reap the next batch of events.

core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
djspiewak: Oooooh I see what you're getting at. Can we resolve this by having `poll` return a ternary result?

- Returned all results, no more to fetch
- Returned all results, hit buffer limit so need to be re-polled
- Returned nothing, no more to fetch

armanbilge: Yeah, I think that should work.

> * Returned all results, hit buffer limit so need to be re-polled

\* so need to be re-polled after `processReadyEvent`s clears the buffer.

djspiewak (approved):

PR #4231, opened 2025-01-08
New year who dis

djspiewak:


armanbilge (dismissed):

djspiewak:
Definitely an encouraging sign that CI failed.

std/shared/src/main/scala/cats/effect/std/MapRef.scala
```diff
@@ -486,68 +486,4 @@ object MapRef extends MapRefCompanionPlatform {
...
-        case Some(v) =>
-          val (set, out) = f(v)
-          (set.some, out.some)
-      }
-
-    def withDefaultValue(default: V)(implicit E: Eq[V], F: Functor[F]): MapRef[F, K, V] =
-      defaultedMapRef(mRef, default)
-  }
-}
+            e
```
armanbilge: spurious changes?

djspiewak:
I think every year I rediscover that sbt-header has some bugs.

Marcus-Rosti:
> I think every year I rediscover that sbt-header has some bugs.

Well, there's always next year

PR #4234, opened 2025-01-11
Shutdown the runtime only when unforked

armanbilge:
It's already explained in the code comment: shutting down the runtime is really only relevant for unforked scenarios.

https://github.com/typelevel/cats-effect/blob/a49a0286b1530047f7aaf1e0964339d88c73dce8/core/jvm/src/main/scala/cats/effect/IOApp.scala#L505-L507

By not aggressively shutting down the runtime in the shutdown handler, we fix some gotchas e.g. deadlock if you attempted to execute `System.exit` on the runtime.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/IOApp.scala
+++ b/core/jvm/src/main/scala/cats/effect/IOApp.scala
@@ -217,7 +217,6 @@ trait IOApp {
               IOApp.this.reportFailure(t).unsafeRunAndForgetWithoutCallback()(runtime)
 
             case t =>
-              runtime.shutdown()
               queue.clear()
               queue.put(t)
           }
@@ -504,7 +503,7 @@ trait IOApp {
 
       // Clean up after ourselves, relevant for running IOApps in sbt,
       // otherwise scheduler threads will accumulate over time.
-      runtime.shutdown()
+      if (!isForked) runtime.shutdown()
     }
 
     val hook = new Thread(() => handleShutdown())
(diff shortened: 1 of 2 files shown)


core/jvm/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -217,7 +217,6 @@ trait IOApp {
               IOApp.this.reportFailure(t).unsafeRunAndForgetWithoutCallback()(runtime)
 
             case t =>
-              runtime.shutdown()
               queue.clear()
               queue.put(t)
```
armanbilge, line 221: We don't have access to `isForked` here. But, I think this particular case doesn't matter, because `t` is a fatal exception, and point of fatal exceptions in that they may leave the JVM in an inconsistent state (such as stuff not being shutdown/closed). Furthermore it may never have been safe to attempt shutdown here anyway—we don't entirely control this `shutdown` code, so we couldn't be certain it's not allocating.

tests/jvm/src/main/scala/catseffect/examplesplatform.scala
```diff
@@ -35,8 +35,7 @@ package examples {
     override protected def runtimeConfig =
       super.runtimeConfig.copy(shutdownHookTimeout = Duration.Zero)
 
-    val run: IO[Unit] =
-      IO.blocking(System.exit(0)).uncancelable
+    val run: IO[Unit] = IO(System.exit(0))
```
armanbilge, line 38: Restores this test to the original state*, now that we fixed this gotcha.

\* The [original test](https://github.com/typelevel/cats-effect/pull/2394) had `uncancelable` and I'm not sure why, since I don't think it does anything.

djspiewak (approved):

PR #4236, opened 2025-01-13
update `.scalafmt.conf`. enforce new wildcard syntax

xuwei-k:


djspiewak (commented):
This is really good but would you mind retargeting against `series/3.6.x`? It touches enough lines that we're going to need to have it on both branches.

xuwei-k:
https://www.githubstatus.com/incidents/qd96yfgvmcf9 😢

djspiewak (approved):
Thank you!

PR #4238, opened 2025-01-15
fix: don't try to advance time by Zero

Jasper-M:
Correct me if that's impossible, but I suspect this caused some `requirement failed` errors to randomly occur in our CI.

### DIFF
--- a/kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/TestContext.scala
+++ b/kernel-testkit/shared/src/main/scala/cats/effect/kernel/testkit/TestContext.scala
@@ -198,7 +198,10 @@ final class TestContext private (_seed: Long) extends ExecutionContext { self =>
   def tickAll(): Unit = {
     tick()
     if (!stateRef.tasks.isEmpty) {
-      advance(nextInterval())
+      val next = nextInterval()
+      if (next > Duration.Zero) {
+        advance(next)
+      }
       tickAll()
     }
   }


djspiewak:
Huh. This situation really should be impossible, simply because `tick()` should be vacuuming up everything which is available at the current time epoch, so the next should always be strictly greater than zero. If you're seeing cases where that isn't true, then something else is definitely broken.

Jasper-M:
Huh, that does seem impossible indeed 🫤
Unfortunately as it only seems to occur sometimes, I didn't yet manage to get more information out of it than this stacktrace
```
java.lang.IllegalArgumentException: requirement failed
  at scala.Predef$.require(Predef.scala:324)
  at executeEmbed @ com.example.MySpec$$anon$1.<init>(MySpec.scala:46)
  at executeEmbed @ com.example.MySpec$$anon$1.<init>(MySpec.scala:46)
```
where 46 is the line where I call `unsafeRunSync()`.

I have to admit that we are calling an actual database in this test in combination with `TestControl`, which I've suspected from the beginning is causing this issue. But still weird that it doesn't fail consistently in that case. And I don't see which other `require` call it can be.

Jasper-M:
I can now confirm that we get this exception:
> java.lang.IllegalArgumentException: requirement failed: time must be > 0

But probably it is due to doobie doing `async.evalOn(async.delay(dataSource.getConnection()), connectEC)` which I guess in our specific case results in this race condition instead of the usual deadlock.

PR #4244, opened 2025-01-19
fix warnings in scalafix rule

xuwei-k:


djspiewak:
Thank you! This is great. Would you like this to go out in the upcoming 3.6.x releases? If so, then you'll need to retarget to `series/3.6.x`.

djspiewak (approved):

PR #4246, opened 2025-01-20
Revert documentation format changes

mzuehlke:
Scalafmt 3.8.4 introduced some formatting bugs that are reverted with this pull request.

Issue in Scalafmt:
- https://github.com/scalameta/scalafmt/issues/4716

Bad formatting got merged to cats-effect in:
- https://github.com/typelevel/cats-effect/pull/4235


djspiewak:
Nice catch thank you! Would you mind retargeting to `series/3.6.x` so we can release sooner?

armanbilge:
Actually, the changes this is reverting are only in `series/3.x`, so this PR has the right target.

I think we should add ba11dfe5b6ab7c9f45521dde7c110212629ac50f to `.git-blame-ignore-revs`.
https://github.com/typelevel/cats-effect/blob/ba11dfe5b6ab7c9f45521dde7c110212629ac50f/.git-blame-ignore-revs#L10-L11

mzuehlke:
> I think we should add https://github.com/typelevel/cats-effect/commit/ba11dfe5b6ab7c9f45521dde7c110212629ac50f to .git-blame-ignore-revs.

done 👍

.git-blame-ignore-revs
```diff
@@ -9,3 +9,6 @@ db6c201aad98cc3d19f67cc688dfa6332e2fb939
 
 # Scala Steward: Reformat with scalafmt 3.8.4
 97565e196071ba28bac911411b9c18bd0e6526e8
+
+# Revert documentation format changes
+ba11dfe5b6ab7c9f45521dde7c110212629ac50f
```
armanbilge:
```suggestion
ba11dfe5b6ab7c9f45521dde7c110212629ac50f

```

PR #4247, opened 2025-01-21
Refactor `WorkerThread` runloop; avoid pathological starvation of pollers

### LINKED ISSUE #4228: External queue can starve timers/pollers
If the local queue becomes empty, the worker transitions to checking the external queue.
https://github.com/typelevel/cats-effect/blob/a49a0286b1530047f7aaf1e0964339d88c73dce8/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala#L824-L827

If it finds work there, it will immediately transition back to crunching the local queue.
https://github.com/typelevel/cats-effect/blob/a49a0286b1530047f7aaf1e0964339d88c73dce8/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala#L668-L669

This offers no opportunity to fire timers or poll for I/O.

armanbilge:
Fixes https://github.com/typelevel/cats-effect/issues/4228.

At a high level, a `WorkerThread` is always in one of the following three states:

1. Working (primarily on the local queue).
2. Looking for work (external or stolen).
3. Parked.

Previously, (3) was a separate `parkLoop()`, but (1) and (2) were entangled in the top-level runloop. Now, (2) is refactored into a separate `lookForWork()` loop. Thus, the top-level runloop is only responsible for tracking "ticks", such that polling runs every 64 ticks.

worker-runloop-fsm

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
@@ -317,7 +317,7 @@ private[effect] final class WorkerThread[P <: AnyRef](
 
     val done = pool.done
 
-    /**
+    /*
      * This method is called when the local queue is empty, and will return only when work is
      * found.
      *
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
@@ -313,113 +313,255 @@ private[effect] final class WorkerThread[P <: AnyRef](
     random = ThreadLocalRandom.current()
     val rnd = random
 
-    /*
-     * A counter (modulo `ExternalQueueTicks`) which represents the
-     * `WorkerThread` finite state machine. The following values have special
-     * semantics explained here:
-     *
-     *   0: To increase the fairness towards fibers scheduled by threads which
-     *      are external to the `WorkStealingThreadPool`, every
-     *      `ExternalQueueTicks` number of iterations, the external queue takes
-     *      precedence over the local queue.
-     *
-     *      If a fiber is successfully dequeued from the external queue, it will
-     *      be executed immediately. If a batch of fibers is dequeued instead,
-     *      the whole batch is enqueued on the local queue and other worker
-     *      threads are notified of existing work available for stealing. The
-     *      `WorkerThread` unconditionally transitions to executing fibers from
-     *      the local queue (state value 4 and larger).
-     *
-     *      This state occurs "naturally" after a certain number of executions
-     *      from the local queue (when the state value wraps around modulo
-     *      `ExternalQueueTicks`).
-     *
-     *   1: Fall back to checking the external queue after a failed dequeue from
-     *      the local queue. Depending on the outcome of this check, the
-     *      `WorkerThread` transitions to executing fibers from the local queue
-     *      in the case of a successful dequeue from the external queue (state
-     *      value 4 and larger). Otherwise, the `WorkerThread` continues with
-     *      asking for permission to steal from other `WorkerThread`s.
-     *
-     *      Depending on the outcome of this request, the `WorkerThread` starts
-     *      looking for fibers to steal from the local queues of other worker
-     *      threads (if permission was granted, state value 2), or parks
-     *      directly. In this case, there is less bookkeeping to be done
-     *      compared to the case where a worker was searching for work prior to
-     *      parking. After the worker thread has been unparked, it transitions
-     *      to looking for work in the external queue (state value 3) while also
-     *      holding a permission to steal fibers from other worker threads.
-     *
-     *   2: The `WorkerThread` has been allowed to steal fibers from other
-     *      worker threads. If the attempt is successful, the first fiber is
-     *      executed directly and the `WorkerThread` transitions to executing
-     *      fibers from the local queue (state value 4 and larger). If the
-     *      attempt is unsuccessful, the worker thread announces to the pool
-     *      that it was unable to find any work and parks.
-     *
-     *   3: The `WorkerThread` has been unparked an is looking for work in the
-     *      external queue. If it manages to find work there, it announces to
-     *      the work stealing thread pool that it is no longer searching for
-     *      work and continues to execute fibers from the local queue (state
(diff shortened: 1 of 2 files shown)


core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -313,113 +313,255 @@ private[effect] final class WorkerThread[P <: AnyRef](
...
      * performance.
      */
-    var state = 4
+    def lookForWork(): Unit = {
+      var state = 0
+      while (!done.get()) {
+        (state: @switch) match {
+          case 0 =>
+            // Check the external queue after a failed dequeue from the local
+            // queue (due to the local queue being empty).
```
armanbilge: Actually, we could lift `case 0` out of the loop. It's always guaranteed to run exactly once, and never again. The loop just transitions between cases 1 and 2.

armanbilge:
I think the failures are legit 😕 something must be broken.

armanbilge:
Aha, all the CI failures are on JDK 21, and this is the test it's hanging on.
https://github.com/typelevel/cats-effect/blob/fc5818dd512fbc7e281f91c54f9332985313d36d/tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala#L615-L630

Edit: but ... we are getting this weirdness, on the new test I added in this PR.
```
java.lang.InterruptedException
  | => tat java.base/java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireSharedInterruptibly(AbstractQueuedSynchronizer.java:1100)
        at java.base/java.util.concurrent.CountDownLatch.await(CountDownLatch.java:230)
        at cats.effect.IOPlatformSpecification.$anonfun$platformSpecs$184(IOPlatformSpecification.scala:601)
        at cats.effect.unsafe.WorkerThread.lookForWork$1(WorkerThread.scala:391)
        at cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:819)
```

tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
```diff
@@ -582,6 +582,40 @@ trait IOPlatformSpecification extends DetectPlatform { self: BaseSpec with Scala
...
+          def mkExternalWork: Runnable = { () =>
+            val latch = new CountDownLatch(1)
+            ExecutionContext.global.execute { () =>
+              pool.execute(mkExternalWork)
+              latch.countDown()
+            }
+            try {
+              latch.await() // wait until next task is in external queue
+            } catch {
+              case _: InterruptedException => // ignore, runtime is shutting down
```
armanbilge, line 604: The stray `InterruptedException` is apparently fatal and was nuking all the runtimes, causing remaining tests to hang. We only noticed this on JDK 21 CI runners because the last test in the suite happens to be JDK 21+ only.

djspiewak:
Haven't reviewed the code yet, but I ran some benchmarks to sanity check. Looks like a very slight regression (note the `parTraverse` results in particular), but considering it fixes a number of fairness issues, I don't consider that to be an impediment.

## Before

```
[info] Benchmark                                             (cpuTokens)   (size)   Mode  Cnt     Score   Error    Units
[info] ParallelBenchmark.parTraverse                               10000     1000  thrpt   10   755.287 ± 1.087    ops/s
[info] ParallelBenchmark.traverse                                  10000     1000  thrpt   10    68.832 ± 0.112    ops/s
[info] WorkStealingBenchmark.alloc                                   N/A  1000000  thrpt   10    11.866 ± 0.089  ops/min
[info] WorkStealingBenchmark.manyThreadsSchedulingBenchmark          N/A  1000000  thrpt   10    24.042 ± 2.950  ops/min
[info] WorkStealingBenchmark.runnableScheduling                      N/A  1000000  thrpt   10  2541.428 ± 5.020  ops/min
[info] WorkStealingBenchmark.runnableSchedulingScalaGlobal           N/A  1000000  thrpt   10  1874.609 ± 6.710  ops/min
[info] WorkStealingBenchmark.scheduling                              N/A  1000000  thrpt   10    25.641 ± 3.496  ops/min
```

## After

```
[info] Benchmark                                             (cpuTokens)   (size)   Mode  Cnt     Score   Error    Units
[info] ParallelBenchmark.parTraverse                               10000     1000  thrpt   10   732.683 ± 1.429    ops/s
[info] ParallelBenchmark.traverse                                  10000     1000  thrpt   10    68.975 ± 0.100    ops/s
[info] WorkStealingBenchmark.alloc                                   N/A  1000000  thrpt   10    11.892 ± 0.135  ops/min
[info] WorkStealingBenchmark.manyThreadsSchedulingBenchmark          N/A  1000000  thrpt   10    30.569 ± 3.926  ops/min
[info] WorkStealingBenchmark.runnableScheduling                      N/A  1000000  thrpt   10  2576.645 ± 5.654  ops/min
[info] WorkStealingBenchmark.runnableSchedulingScalaGlobal           N/A  1000000  thrpt   10  1905.347 ± 2.478  ops/min
[info] WorkStealingBenchmark.scheduling                              N/A  1000000  thrpt   10    25.254 ± 4.653  ops/min
```

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -313,113 +313,255 @@ private[effect] final class WorkerThread[P <: AnyRef](
...
+
+          case 1 =>
+            // update the current time
+            now = System.nanoTime()
+
+            // First try to steal some expired timers:
+            if (pool.stealTimers(now, rnd)) {
+              // some stolen timer created new work for us
+              pool.transitionWorkerFromSearching(rnd)
+              return
```
djspiewak: We should probably steal both timers *and* fibers; why only steal one?

djspiewak (commented):
Relatively minor optimization suggestion. The rest looks quite good.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
armanbilge: I was able to fully unroll it in bddd0f30fcb0dcb1e405ba70b7fcc53dde9f14e3. No more `switch` needed.

djspiewak (approved):

PR #4255, opened 2025-01-25
Restore `ArrayBlockingQueue` for worker-to-blocking state transfer

armanbilge:
Reverts https://github.com/typelevel/cats-effect/pull/3446, see explanation in https://github.com/typelevel/cats-effect/pull/3446#issuecomment-2614103276.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
@@ -914,7 +914,7 @@ private[effect] final class WorkerThread[P <: AnyRef](
         val idx = index
         pool.replaceWorker(idx, cached)
         // Transfer the data structures to the cached thread and wake it up.
-        cached.indexTransfer.offer(idx)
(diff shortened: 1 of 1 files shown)


core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -914,7 +914,7 @@ private[effect] final class WorkerThread[P <: AnyRef](
         val idx = index
         pool.replaceWorker(idx, cached)
         // Transfer the data structures to the cached thread and wake it up.
-        cached.indexTransfer.transfer(idx)
+        cached.indexTransfer.offer(idx)
```
armanbilge: ~~If we really want to test our assumptions to ensure we never block unnecessarily, we should use `tryOffer` and fail if unable to do so. If we got it right, it `tryOffer` should always succeed.~~

Oops, that's what `offer` already does, we are just currently ignoring the result value. I guess it always worked fine 😅

djspiewak (approved):
Oh I get why this works. Yes this is vastly better.

PR #4264, opened 2025-02-02
Transfer tick state when replacing a blocked thread

armanbilge:


core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -1026,6 +1032,11 @@ private[effect] final class WorkerThread[P <: AnyRef](
 
 private[effect] object WorkerThread {
 
+  private[unsafe] final class TransferState {
```
durban, line 1035: Is `TransferState` mutable just to avoid always allocating a new instance? Or am I missing something?

(Also, a possibly stupid idea: we could pack the 2 `Int`s into one `Long`... I think that way the `transferState` field could be avoided. But also, that would cause `java.lang.Long` object allocations when transferring...)

armanbilge: > Is `TransferState` mutable just to avoid always allocating a new instance?

Yes, that's right.

> But also, that would cause `java.lang.Long` object allocations when transferring...

Yes, and they would be unlikely to be cached. The old implementation transferring `java.lang.Integer`s were likely to be cached due to their relatively small values (likely < 64, except on some very large machines).

I think if we want to do better here, we'd need to implement a custom concurrent data structure for transfers.

djspiewak (approved):

PR #4266, opened 2025-02-05
Swapped out static tag approach with `ClassValue` cache

### LINKED ISSUE #4260: use `java.lang.ClassValue` instead of `IO.tag`
I think `java.lang.ClassValue` is faster than current implementation.
 
- benchmark code https://github.com/xuwei-k/cats-effect/commit/f7c2dabfd974bcfb57cff683a360ef7c163a4487
- https://github.com/scala/scala/pull/9632
 
```
Jmh / run -i 10 -wi 10 -f 1 -t 1 cats.effect.benchmarks.TagBenchmark
```

```
[info] Benchmark                 Mode  Cnt         Score        Error  Units
[info] TagBenchmark.classValue  thrpt   10  10348652.195 ± 116363.873  ops/s
[info] TagBenchmark.tag         thrpt   10   7336411.108 ±  91988.686  ops/s
```

djspiewak:
All credit goes to @xuwei-k here. I even cribbed his code more or less directly. I haven't yet run the full set of benchmarks so hold off on merging.

Fixes #4260


djspiewak:
Ah, unfortunate discovery: `ClassValue` doesn't exist on Scala Native!

armanbilge:
Darn, unless we do some messy source-splitting, that means we won't get this in until we complete the 0.5 upgrade ... anyway I opened an issue.
- https://github.com/scala-native/scala-native/issues/4209

djspiewak:
Yeah I briefly thought about doing some sort of weird split thing but it would be bonkers. So I'll probably just push this far enough to get a full set of benchmark results so we can prioritize down the road, then leave it on ice until SN catches up.

armanbilge:
Studied the implementation of `ClassValue` a bit, it still has some fairly complicated machinery. I wonder if we can beat it by doing something similar and simpler (I will prototype this later when I have time). The one definitive advantage that `ClassValue` has is that it's implemented by a field on `Class`.

djspiewak:
> The one definitive advantage that ClassValue has is that it's implemented by a field on Class.

Ostensibly it's also something the JIT understands. I think that's the real advantage, but we would need to actually examine the hotspot outputs to be sure.

armanbilge:
> I wonder if we can beat it by doing something similar and simpler (I will prototype this later when I have time).

I made an experimental "`ClassByteMap`" in 1472f3d252eb0d79c161c9f5d01dec249ce7252b, seems to be faster than `ClassValue`:

```
Benchmark                   Mode  Cnt         Score        Error  Units
TagBenchmark.classByteMap  thrpt   20  10417078.521 ± 466789.955  ops/s
TagBenchmark.classValue    thrpt   20   8920919.389 ± 110746.064  ops/s
TagBenchmark.tag           thrpt   20   7082499.380 ± 132528.920  ops/s
```

PR #4267, opened 2025-02-06
Add NonEmptyHotswap

morgen-peschke:
Also corrects the site docs for `Hotswap` so it conforms to the `Hotswap` API

Addresses #4268


std/shared/src/main/scala/cats/effect/std/NonEmptyHotswap.scala
```diff
@@ -0,0 +1,153 @@
...
+            state.modify {
+              case Acquired(_, fin) =>
+                next -> fin
+              case Finalized =>
+                val fin = next match {
+                  case Acquired(_, fin) => fin
+                  case _ => F.unit
+                }
+                Finalized -> (fin *> raise[Unit]("Cannot swap after finalization"))
+            }.flatten
```
armanbilge:
```suggestion
            state.flatModify {
              case Acquired(_, fin) =>
                next -> fin
              case Finalized =>
                val fin = next match {
                  case Acquired(_, fin) => fin
                  case _ => F.unit
                }
                Finalized -> (fin *> raise[Unit]("Cannot swap after finalization"))
            }
```

std/shared/src/main/scala/cats/effect/std/NonEmptyHotswap.scala
```diff
@@ -0,0 +1,153 @@
...
+ * }}}
+ *
+ * [[NonEmptyHotswap]] is particularly useful when working with effects that cycle through
+ * resources, like writing bytes to files or rotating files every N bytes or M seconds. Without
+ * [[NonEmptyHotswap]], such effects leak resources: on each file rotation, a file handle or
+ * some internal resource handle accumulates. With [[NonEmptyHotswap]], the only registered
+ * resource is the [[NonEmptyHotswap]] itself, and each file is swapped in only after swapping
+ * the previous one out.
+ *
+ * Born from [[Hotswap]].
```
armanbilge: Instead of duplicating the implementation, we should either implement one in terms of the other ... I believe either direction should be possible.

std/shared/src/main/scala/cats/effect/std/NonEmptyHotswap.scala
```diff
@@ -0,0 +1,153 @@
...
+   * is not used thereafter. Failure to do so may result in an error on the _consumer_ side. In
+   * any case, no resources will be leaked.
+   *
+   * For safer access to the current resource see [[get]], which guarantees that it will not be
+   * released while it is being used.
+   *
+   * If [[swap]] is called after the lifetime of the [[NonEmptyHotswap]] is over, it will raise
+   * an error, but will ensure that all resources are finalized before returning.
+   */
+  def swap(next: Resource[F, R]): F[R]
```
armanbilge: This signature is actually unsafe, because you should never return `R` without acquiring the semaphore lock. See https://github.com/typelevel/cats-effect/pull/3480#discussion_r1125594833.

It must be either:

```suggestion
  def swap(next: Resource[F, R]): F[Unit]
```
or:
```suggestion
  def swap(next: Resource[F, R]): Resource[F, R]
```

armanbilge (commented):
Thanks for the PR!

Honestly, I wonder if we should deprecate the old `HotSwap`.

1. `Hotswap[F, A]` is just `NonEmptyHotswap[F, Option[A]]`. We don't need both.
2. The current `Hotswap` API is broken anyway, e.g. see https://github.com/typelevel/cats-effect/pull/3480#discussion_r1125594833.

@djspiewak I know you hate it, but there are so many things we could fix if we just embrace the ugly `Hotswap2`, `Queue2`, 😅 

Oh, also you should retarget to `series/3.6.x` if you want it to make the next release (@djspiewak are we still merging new APIs into this branch?)

std/shared/src/main/scala/cats/effect/std/NonEmptyHotswap.scala
morgen-peschke: I'm happy with `hs.swap(...) *> hs.get`

std/shared/src/main/scala/cats/effect/std/NonEmptyHotswap.scala
morgen-peschke: After changing `swap` to return `F[Unit]`, I don't believe it is possible to implement `Hotswap` in terms of `NonEmptyHotswap` 🤔 

As for doing it the other way around, I'll wait on the discussion about deprecating the current `Hotswap` before playing around with that one.

armanbilge: > I don't believe it is possible to implement `Hotswap` in terms of `NonEmptyHotswap` 🤔

I believe you can do `swap *> get.use(pure)`. It's leaking, but that's exactly the point.

morgen-peschke: Good point, I'd forgotten that `.use` can be used to cheat like that. It looks like that is possible, though it that does mean that `Hotswap#get` will acquire a lock even when no resource is present.

I'm not sure that's a deal-breaker though, since it's not a change to the API and should be only a very minor performance penalty.

armanbilge: > will acquire a lock even when no resource is present.

Oh right, I forgot about this subtlety.

- https://github.com/typelevel/cats-effect/pull/3922

Which means my previous assertion that `Hotswap[F, A]` is just `NonEmptyHotswap[F, Option[A]]` is actually not quite correct.

morgen-peschke: It's easy enough to roll back [8dc9fcf](https://github.com/typelevel/cats-effect/pull/4267/commits/8dc9fcf7371fa55a495e233c64df2a70944b2179) 🤷🏻

morgen-peschke: Added a note in the docs about it to avoid losing track of this difference.

std/shared/src/main/scala/cats/effect/std/Hotswap.scala
```diff
@@ -102,70 +101,14 @@ object Hotswap {
...
-      case class Acquired(r: R, fin: F[Unit]) extends State
-      case object Finalized extends State
+    NonEmptyHotswap.create[F, R].map { nes =>
+      new Hotswap[F, R] {
+        override def swap(next: Resource[F, R]): F[R] =
+          nes.swap(next.map(_.some)) *> get.use(_.get.pure[F])
 
-      def initialize: F[Ref[F, State]] =
-        F.ref(Cleared)
+        override def get: Resource[F, Option[R]] = nes.get
```
armanbilge: We don't need to rollback this commit, we just need to reuse some ideas from the old implementation. Something like this:
```scala
  override def get: Resource[F, Option[R]] =
    Resource[F, Option[R]] { poll =>
      poll(nes.get.allocated).flatMap {
        case (Some(r), fin) => F.pure((Some(r), fin))
        case (None, fin) => fin *> F.pure((None, F.unit))
      }
    }
```

morgen-peschke: Wouldn't that still acquire the lock, since that happens inside the call to `nes.get`?

If I'm reading that right it would release the lock rather quickly, I don't really understand the `Resource` internals well enough to grok if that would be quick enough to satisfy the requirement.

armanbilge: > Wouldn't that still acquire the lock, since that happens inside the call to `nes.get`? If I'm reading that right it would release the lock rather quickly,

That's correct. This is also how `Hotswap` works currently:

https://github.com/typelevel/cats-effect/blob/917d5548ddd86864a39ed949a3085188bc421a8c/std/shared/src/main/scala/cats/effect/std/Hotswap.scala#L141-L148

morgen-peschke: Works for me 👍🏻

morgen-peschke: Works for the test too :)

kamilkloch:
> @djspiewak I know you hate it, but there are so many things we could fix if we just embrace the ugly `Hotswap2`, `Queue2`, 😅

sttp has embraced this path a long time ago and it works like a charm :) https://github.com/softwaremill/sttp/tree/master/core/src/main

morgen-peschke:
@armanbilge did you want the `Hotswap2` rename?

armanbilge:
I do, but I'm not sure if @djspiewak has bought in.

morgen-peschke:
@djspiewak could we get a ping if this is something you'd like to weigh in on?

morgen-peschke:
@armanbilge it's been 6 weeks without a ping from @djspiewak, can we take that as "decline to comment" and proceed?

morgen-peschke:
@armanbilge no shade intended by this question, just want to know if I should close it: is this a dead PR?

armanbilge:
@morgen-peschke review and release cycles in Cats Effect are often very slow, sorry about that, and thanks for your patience ☺️

morgen-peschke:
@armanbilge what's the culturally acceptable frequency of comment-bumping a PR in this state? 

I'd really like to offload having to remember to check back on this to a repeating calendar event and I also don't want to be a bother by posting too often just to keep the PR from sliding completely off the radar, if the expectation is that this could be another 3+ months.

ETA: In case the above came across wrong, I want to reiterate that no slight is intended, I'm trying to figure out the best way to manage the time and energy I'm going to spend keeping track of this PR, as well as avoid creating the association in y'all's mind as that annoying dev who won't shut up about that one PR.

docs/std/hotswap.md
```diff
@@ -61,4 +59,9 @@ def rotating(n: Int): Resource[IO, Logger[IO]] =
       }
     }
   }
+}
 ```
+
+## Hotswap
+
+In cases where the managed `Resource` may be missing, `Hotswap` provides some optimizations for this use case.
```
armanbilge: Sorry if I'm missing something, what is this sentence referring to exactly?

std/shared/src/main/scala/cats/effect/std/Hotswap.scala
armanbilge, line 1: I think we can fully deprecate `Hotswap` in favor of `Hotswap2`.

std/shared/src/main/scala/cats/effect/std/Hotswap.scala
```diff
@@ -102,70 +102,20 @@ object Hotswap {
...
-              poll(next.allocated).flatMap {
-                case (r, fin) =>
-                  exclusive.mapK(poll).onCancel(Resource.eval(fin)).surround {
-                    swapFinalizer(Acquired(r, fin)).as(r)
-                  }
-              }
+    Hotswap2.create[F, R].map { nes =>
+      new Hotswap[F, R] {
+        override def swap(next: Resource[F, R]): F[R] =
+          nes.swap(next.map(_.some)) *> get.use(_.get.pure[F])
```
armanbilge, line 112: This is the "leaking" implementation, correct? Maybe we should add a comment saying this is intentional, to satisfy the mistakes of the old api.

std/shared/src/main/scala/cats/effect/std/Hotswap2.scala
```diff
@@ -0,0 +1,166 @@
...
+                  }
+              }
+            }
+
+          override def get: Resource[F, R] =
+            Resource.makeFull[F, R] { poll =>
+              poll(semaphore.acquire) *> // acquire shared lock
+                state.get.flatMap {
+                  case Acquired(r, _) => F.pure(r)
+                  case _ => raise("NonEmptyHotswap already finalized")
```
armanbilge:
```suggestion
                  case _ => raise("Hotswap already finalized")
```

std/shared/src/main/scala/cats/effect/std/Hotswap2.scala
```diff
@@ -0,0 +1,166 @@
...
+        }
+      }
+    }
+
+  /**
+   * Creates a `NonEmptyHotswap` that is largely functionally equivalent to `Hotswap`
+   *
+   * The primary difference is that [[Hotswap2.swap]] does not leak the resource.
+   */
+  def create[F[_], R](implicit F: Concurrent[F]): Resource[F, Hotswap2[F, Option[R]]] =
```
armanbilge: Bikeshed? can also update the scaladoc.

```suggestion
  def empty[F[_], R](implicit F: Concurrent[F]): Resource[F, Hotswap2[F, Option[R]]] =
```

std/shared/src/main/scala/cats/effect/std/Hotswap2.scala
```diff
@@ -0,0 +1,166 @@
...
+ *                                              r3 released
+ * }}}
+ *
+ * [[Hotswap2]] is particularly useful when working with effects that cycle through resources,
+ * like writing bytes to files or rotating files every N bytes or M seconds. Without
+ * [[Hotswap2]], such effects leak resources: on each file rotation, a file handle or some
+ * internal resource handle accumulates. With [[Hotswap2]], the only registered resource is the
+ * [[Hotswap2]] itself, and each file is swapped in only after swapping the previous one out.
+ *
+ * Born from [[Hotswap]].
```
armanbilge:
```suggestion
 * Replaces the deprecated [[Hotswap]] with a safer API.
```

std/shared/src/main/scala/cats/effect/std/Hotswap2.scala
```diff
@@ -0,0 +1,166 @@
...
+ */
+sealed trait Hotswap2[F[_], R] {
+
+  /**
+   * Allocates a new resource, closes the previous one, and returns the newly allocated `R`.
```
armanbilge:
```suggestion
   * Allocates a new resource and closes the previous one.
```

std/shared/src/main/scala/cats/effect/std/Hotswap2.scala
```diff
@@ -0,0 +1,166 @@
...
+   *
+   * When the lifetime of the [[Hotswap2]] is completed, the resource allocated by the most
+   * recent [[swap]] will be finalized.
+   *
+   * [[swap]] finalizes the previous resource immediately, so users must ensure that the old `R`
+   * is not used thereafter. Failure to do so may result in an error on the _consumer_ side. In
+   * any case, no resources will be leaked.
+   *
+   * For safer access to the current resource see [[get]], which guarantees that it will not be
+   * released while it is being used.
```
armanbilge: These docs can be updated, they are trying to massage around the shortcomings of the old API.

armanbilge (dismissed):
I spoke to Daniel offline, he is alright with moving ahead with `Hotswap2`. Thanks for your patience!

I just noticed ... we will need to retarget this PR at `series/3.x` and I think this will require rewriting the test suite using munit, sorry about that!

docs/std/hotswap.md
morgen-peschke: It's referring to some of the stuff that `Hotswap` does around avoiding blocking `swap` after a call to `get` returns a `Resource` containing a `None`.

If we're fully deprecating `Hotswap`, we can probably drop this section entirely.

morgen-peschke:
@armanbilge the test failure appears to be unrelated to this change, if everything else looks good, can you retrigger CI? If something needs changed, don't worry about it, since it should clear on the next push 🤞🏻

std/shared/src/main/scala/cats/effect/std/Hotswap.scala
```diff
@@ -51,6 +50,7 @@ import cats.syntax.all._
  *
  * Ported from https://github.com/typelevel/fs2.
  */
+@deprecated("Use Hotswap2", "3.6.3")
```
armanbilge: This will land in 3.7.0, since we only introduce new APIs in minors (and reserve patch versions for bug fixes only).

```suggestion
@deprecated("Use Hotswap2", "3.7.0")
```

morgen-peschke: Fixed, here and elsewhere 👍🏻

djspiewak (commented):
Is there a reason we moved away from `NonEmptyHotswap` as a name? Aesthetically that seems better to me than `Hotswap2`. We could still deprecate `Hotswap` even in that world.

morgen-peschke:
> Is there a reason we moved away from `NonEmptyHotswap` as a name? Aesthetically that seems better to me than `Hotswap2`. We could still deprecate `Hotswap` even in that world.

This isn't something I have strong opinions about, so you'll have to check in with @armanbilge for the reasons behind it.

Context: originally raised in [this comment](https://github.com/typelevel/cats-effect/pull/4267#pullrequestreview-2600765005), I swapped the names because it was requested by the primary reviewer, I couldn't think of a good reason not to do it, and I didn't want to hold up the PR.

armanbilge:
@djspiewak

> Aesthetically that seems better to me than `Hotswap2`.

I don't want to hold this up over aesthetic concerns 😁 if `NonEmptyHotswap` feels better, let's just do that. To be clear, I still feel strongly that the original `Hotswap` should be deprecated.

The reason I suggested `Hotswap2` is because I think this new API should _be_ `Hotswap`, not just a different flavor of it. The original `Hotswap` API is broken, and my personal opinion is the "emptiness" of it is also kind of weird. So describing this new API as the non-empty version of this broken, deprecated API did not sit right with me.

std/shared/src/main/scala/cats/effect/std/Hotswap.scala
```diff
@@ -51,6 +50,7 @@ import cats.syntax.all._
  *
  * Ported from https://github.com/typelevel/fs2.
  */
+@deprecated("Use Hotswap2", "3.7.0")
```
mzuehlke: the deprecation annotation still refer to **Hotswap2**

morgen-peschke: Thanks for the catch, fixed.

djspiewak (approved):

mzuehlke (approved):

PR #4270, opened 2025-02-08
Poll on the 0th tick, check external queue on 32nd

armanbilge:
Refactors the worker thread runloop so that I/O polling and external queue polling are no longer part of the same tick, and instead occur at opposite ends of the loop i.e. the 0th and 32nd ticks. The goal is to reduce the chance that the local queue overflows, by giving it a chance to work on tasks in-between polling these sources.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
@@ -20,7 +20,7 @@ package unsafe
 import cats.effect.tracing.Tracing.captureTrace
 import cats.effect.tracing.TracingConstants
 
-import scala.annotation.{switch, tailrec}
+import scala.annotation.tailrec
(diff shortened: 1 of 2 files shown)


core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -762,8 +762,8 @@ private[effect] final class WorkerThread[P <: AnyRef](
         }
       }
 
-      ((tick & ExternalQueueTicksMask): @switch) match {
-        case 0 =>
+      if ((tick & ExternalWorkTicksMask) == 0) {
+        if ((tick & PollingTicksMask) == 0) {
```
armanbilge, line 766: There's a couple ways we could encode this 🤷 

Also, random idea: what if we just completely unrolled the 64 ticks of the worker loop?

djspiewak (approved):

PR #4279, opened 2025-02-25
Run finalizers in `IO#unsafeRunTimed`

armanbilge:
Follow-up to https://github.com/typelevel/cats-effect/pull/4167#issuecomment-2472346690. Runs finalizers `IO#unsafeRunTimed` if it times-out or is interrupted, but doesn't backpressure on their completion.

### DIFF
--- a/tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
+++ b/tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
@@ -51,22 +51,26 @@ trait IOPlatformSpecification extends DetectPlatform { self: BaseSpec with Scala
     "platform" should {
 
       "unsafeRunTimed cancels task if timed-out" in {
-        val latch = new CountDownLatch(1)
-        val task = IO.never.onCancel(IO.sleep(100.millis) *> IO(latch.countDown()))
+        val latch1 = new CountDownLatch(1)
+        val latch2 = new CountDownLatch(1)
+        val task = IO.never.onCancel(IO.blocking(latch1.await()) *> IO(latch2.countDown()))
         task.unsafeRunTimed(100.millis)(runtime())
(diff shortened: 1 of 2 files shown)


tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
```diff
@@ -50,6 +50,26 @@ trait IOPlatformSpecification extends DetectPlatform { self: BaseSpec with Scala
   def platformSpecs = {
     "platform" should {
 
+      "unsafeRunTimed cancels task if timed-out" in {
+        val latch = new CountDownLatch(1)
+        val task = IO.never.onCancel(IO.sleep(100.millis) *> IO(latch.countDown()))
+        task.unsafeRunTimed(100.millis)(runtime())
+        latch.getCount() must beEqualTo(1) // didn't backpressure on finalizer
```
djspiewak: Can we reframe this test to avoid timing sensitivity? Using juc `CountDownLatch` would probably make that pretty easy.

tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
```diff
@@ -50,6 +50,26 @@ trait IOPlatformSpecification extends DetectPlatform { self: BaseSpec with Scala
...
+        latch.await(1, SECONDS) must beTrue // but it does eventually run
+      }
+
+      "unsafeRunSync cancels task if interrupted" in realWithRuntime { implicit rt =>
+        for {
+          latch <- IO(new CountDownLatch(1))
+          task = IO.never.as(false).onCancel(IO.sleep(100.millis) *> IO(latch.countDown()))
```
djspiewak: Ditto above

djspiewak (changes requested):

djspiewak (approved):

PR #4287, opened 2025-03-04
WIP: Add `JobManager`

BalmungSan:
Implements the `JobManager` idea proposed in #1345 

-----

## Roadmap

- [ ] Define the API.
- [ ] Implement it.
- [ ] Add tests.
- [ ] Add docs.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/JobManager.scala
+++ b/std/shared/src/main/scala/cats/effect/std/JobManager.scala
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2020-2025 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.std
+
+import cats.effect.kernel._
+
+/**
+ * A `JobManager` allows you to launch `Jobs` in the background using a unique identifier. Then
+ * you can use the identifier to query for the status of the job or cancel it.
+ */
+trait JobManager[F[_], Id, S] {
+
+  /**
+   * Creates and launches the given `Job` in the background. If another Job with the same id was
+   * already running, it will be cancelled before starting this one.
+   */
+  def startJob(id: Id, job: Resource[F, JobManager.Job[F, S]]): F[Unit]
+
+  /**
+   * Gets the status of the `Job` associated with the given `id`. If `id` doesn't exists or the
+   * `Job` already finished then the returned value will be a `None`.
+   */
+  def getJobStatus(id: Id): F[Option[S]]
+
+  /**
+   * Signals cancellation of the `Job` associated with the given `id`, and waits for its
+   * completion.
+   */
+  def cancelJob(id: Id): F[Unit]
+}
+
+object JobManager {
+
+  /**
+   * Represents a job managed by a `JobManager`.
+   */
+  trait Job[F[_], S] {
+
+    /**
+     * Starts the logic of this `Job`.
+     */
+    def run: F[Unit]
+
+    /**
+     * Gets the status of this `Job`.
+     */
+    def getStatus: F[S]
+  }
+}
--- a/std/shared/src/main/scala/cats/effect/std/JobManager.scala
+++ b/std/shared/src/main/scala/cats/effect/std/JobManager.scala
@@ -21,9 +21,10 @@ import cats.syntax.all._
 
 /**
  * A `JobManager` allows you to launch `Jobs` in the background using a unique identifier. Then
- * you can use the identifier to query for the status of the job or cancel it.
+ * you can use the identifier to query for the status of the job, wait for its completion, or
+ * cancel it.
  */
-trait JobManager[F[_], Id, S] {
+trait JobManager[F[_], Id, S, R] {
 
   /**
    * Creates and launches the given `Job` in the background. If another Job with the same id was
@@ -31,12 +32,25 @@ trait JobManager[F[_], Id, S] {
    *
    * Note: This waits until the actual job's `run` process has started. In other words, you can
    * query the `status` of the associated id immediately after.
+   *
+   * @return
+   *   the [[cats.effect.kernel.Fiber]] of the started `Job`.
    */
-  def startJob(id: Id, job: Resource[F, JobManager.Job[F, S]]): F[Unit]
+  def startJob(id: Id, job: Resource[F, JobManager.Job[F, S, R]]): F[Fiber[F, Throwable, R]]
 
   /**
-   * Gets the status of the `Job` associated with the given `id`. If `id` doesn't exists or the
-   * `Job` already finished then the returned value will be a `None`.
+   * Waits until the given `Job` finishes and returns its result.
+   *
+   * Note: If `id` does not exists or the `Job` already finished then the returned value will be
+   * a `None`.
+   */
+  def join(id: Id): F[Option[Outcome[F, Throwable, R]]]
+
+  /**
+   * Gets the status of the `Job` associated with the given `id`.
+   *
+   * Note: If `id` does not exists or the `Job` already finished then the returned value will be
+   * a `None`.
    */
   def getJobStatus(id: Id): F[Option[S]]
 
@@ -52,12 +66,12 @@ object JobManager {
   /**
    * Represents a job managed by a `JobManager`.
    */
-  trait Job[F[_], S] {
+  trait Job[F[_], S, R] {
 
     /**
      * Starts the logic of this `Job`.
      */
-    def run: F[Unit]
+    def run: F[R]
 
     /**
      * Gets the status of this `Job`.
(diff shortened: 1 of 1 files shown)


std/shared/src/main/scala/cats/effect/std/JobManager.scala
```diff
@@ -0,0 +1,96 @@
...
+ * A `JobManager` allows you to launch `Jobs` in the background using a unique identifier. Then
+ * you can use the identifier to query for the status of the job or cancel it.
+ */
+trait JobManager[F[_], Id, S] {
+
+  /**
+   * Creates and launches the given `Job` in the background. If another Job with the same id was
+   * already running, it will be cancelled before starting this one.
+   */
+  def startJob(id: Id, job: Resource[F, JobManager.Job[F, S]]): F[Unit]
```
BalmungSan: On one had, I like the idea of users being able to use their own `Ids`.
On the other, I think most users would benefit from a default using automatically generated `UUIDs`.

Should we provide such default in some way?

std/shared/src/main/scala/cats/effect/std/JobManager.scala
```diff
@@ -0,0 +1,96 @@
...
+
+  /**
+   * Gets the status of the `Job` associated with the given `id`. If `id` doesn't exists or the
+   * `Job` already finished then the returned value will be a `None`.
+   */
+  def getJobStatus(id: Id): F[Option[S]]
```
BalmungSan, line 55: I somewhat dislike the idea that both bad id and already finished return in a `None`.
But, otherwise, the map will grow indefinitely.

Any ideas?

std/shared/src/main/scala/cats/effect/std/JobManager.scala
```diff
@@ -0,0 +1,96 @@
...
+
+  /**
+   * Signals cancellation of the `Job` associated with the given `id`, and waits for its
+   * completion.
+   */
+  def cancelJob(id: Id): F[Unit]
```
BalmungSan, line 61: Should we provide a `cacelAndForget` variant where we don't wait on cancellation?

std/shared/src/main/scala/cats/effect/std/JobManager.scala
```diff
@@ -0,0 +1,96 @@
...
+    /**
+     * Starts the logic of this `Job`.
+     */
+    def run: F[Unit]
+
+    /**
+     * Gets the status of this `Job`.
+     */
+    def getStatus: F[S]
+  }
```
BalmungSan: Users would then implement this `trait` for their own `Jobs`.

std/shared/src/main/scala/cats/effect/std/JobManager.scala
```diff
@@ -0,0 +1,96 @@
...
+            jobsMap(id).set(
+              RunningJob(
+                status = job.getStatus,
+                cancel = fiber.cancel
+              ).some
+            ) >> fiber.join
+          }
+        }
+
+        supervisor.supervise(runJob).void
```
BalmungSan: We don't wait for the `Job` to start before returning to the user.
But, that means there is a brief delay between starting the `Job` and users being able to query its status.

std/shared/src/main/scala/cats/effect/std/JobManager.scala
```diff
@@ -0,0 +1,99 @@
...
+        val runJob = jobR.use { job =>
+          supervisor.supervise(job.run).flatMap { fiber =>
+            jobsMap(id)
+              .getAndSet(
+                RunningJob(
+                  status = job.getStatus,
+                  cancel = fiber.cancel
+                ).some
+              )
+              .flatMap(_.traverse_(_.cancel)) >>
```
BalmungSan: In case the same `id` was already used, we cancel the previous `Job`.

std/shared/src/main/scala/cats/effect/std/JobManager.scala
durban: Would that be essentially `cancelJob(...).start`? If yes, I don't think we should add it (everyone can `.start` for themselves).

std/shared/src/main/scala/cats/effect/std/JobManager.scala
durban: I don't think that's okay. I think the point of `std` is to solve tricky race conditions like this for users.

std/shared/src/main/scala/cats/effect/std/JobManager.scala
BalmungSan: Conceptually yes, but also, since we already have a `Dispatcher` in place, it could be used to perform that rather than raw `start`.

std/shared/src/main/scala/cats/effect/std/JobManager.scala
BalmungSan: Fair point.
What do you think would be the best way to solve that? A `Deferred` a `CountdownLatch`? Other thing?
Or maybe, simply run the `runJob` logic there rather that sending it to the `Supervisor`?

std/shared/src/main/scala/cats/effect/std/JobManager.scala
durban: Okay, so I think my point is: if it's somehow _better_ (performance, safety, whatever) than just `.start`-ing, then yeah, maybe add it. If it's not better, then definitely not.

std/shared/src/main/scala/cats/effect/std/JobManager.scala
durban: Yeah, I'm a little confused by why it's sent to the Supervisor... (I'm sure there is a reason, I just don't know it).

std/shared/src/main/scala/cats/effect/std/JobManager.scala
durban: By fully controlling IDs, we could have `startJob` _return_ a fresh ID, so "bad ID" would never happen. (But then we'd lose the ability of users to have their own IDs...)

std/shared/src/main/scala/cats/effect/std/JobManager.scala
BalmungSan: AFAIK, it is safer since its life cycle is attached to the `Supervisor`.
But I am not sure what canceling a cancel does.

std/shared/src/main/scala/cats/effect/std/JobManager.scala
BalmungSan: The main thing is that if the same `id` was already found we cancel it, which is costly.
But, I think I could just send that to the `Supervisor` as well.

std/shared/src/main/scala/cats/effect/std/JobManager.scala
durban: It does "nothing", as a `cancel` is uncancelable.

BalmungSan: Then I guess as you said, it provides no value and users who don't want to wait on the cancellation can just `start`.

std/shared/src/main/scala/cats/effect/std/JobManager.scala
BalmungSan: Oh, looking at the code, I realized that we also need to run the job setup _(`Resource.acquire`)_ before being able to run the job itself.
So that was also part of what was being done in the background, but it also means that the short delay could be bigger.

Thus, I decided to use a `Deferred` to wait until the job has been properly registered before returning. But that may take a while.
Another idea that I just had would be to have an `Initializing` status that could be used meanwhile. However, that would complicate the logic quite a bit.

djspiewak:
I'm finally reviewing this!

So I like where this is going quite a bit. One thing I'm not as fond of is the fact that it's built around `F[Unit]`. I could be persuaded otherwise, but I tend to think that it's better to allow users to return values and then pick them up later. Probably with something like `join(id: Id): F[Option[Outcome[F, Throwable, R]]]` (or `get`) where `R` is the result type.

On that note, I actually wonder about the usefulness of the status stuff. Any user who is doing something really robust there is probably going to want something like an Fs2 `Signal` or similar, while someone who isn't thinking to deeply about it is probably only going to want `Unstarted`/`Started`/`Completed`.

I do actually like the parametric ids. It imposes some limitations obviously but it keeps us a bit more honest, and in many applications, users will in fact have a natural `Id` type that they're going to want to work with. We can have a utility constructor which instantiates `Id = Uuid` for convenience or something.

Needs tests!

std/shared/src/main/scala/cats/effect/std/JobManager.scala
mtomko: Can it not return `Either[Unit, Option[S]]`, where `Left` is "unknown ID" and `Right(None)` is finished?

Alternatively, I guess you could  have an ADT with `Unknown | Completed | Running(x)`

I don't fully understand the comment about the map growing indefinitely, though, so my suggestion may not improve that situation. Also, making this distinction turns a simple API into one that has a lot of ceremony for the caller, who perhaps doesn't care or can guess. I'm wondering if what I'm missing is a detail in how `MapRef` works, that a lookup implictly fills a default value in?

std/shared/src/main/scala/cats/effect/std/JobManager.scala
```diff
@@ -73,21 +76,35 @@ object JobManager {
...
+          // 3. Register the job in the jobsMap.
+          // 4. In case there was another job already register with the same id, cancel it.
+          // 5. Wait for the job to finish.
+          // 6. Run the job cleanup (Resource release).
+          val runJob = jobR.use { job =>
+            supervisor.supervise(job.run).flatMap { jobFiber =>
+              val registerJob =
+                F.guarantee(
+                  jobsMap(id).getAndSet(
+                    RunningJob(
```
mtomko: Would it work to track the job status using a `Deferred` as well? That would let you also provide a semantically blocking API to wait for the completion of the job (under the hood it would just await the Deferred). Otherwise, callers who need `S` are going to have to implement their own polling loops and while they _could_ use a Deferred, it'll have to wake a fiber periodically to do it.

std/shared/src/main/scala/cats/effect/std/JobManager.scala
```diff
@@ -0,0 +1,116 @@
...
+          // but wait until it has been registered in the jobsMap.
+          supervisor.supervise(runJob) >> waitForJobRegistration.get
+        }
+      }
+
+      override def getJobStatus(id: Id): F[Option[S]] =
+        jobsMap(id).get.flatMap(_.traverse(_.status))
+
+      override def cancelJob(id: Id): F[Unit] =
+        jobsMap(id).getAndSet(None).flatMap(_.traverse_(_.cancel))
```
mtomko: Is there any risk of leakage here if `cancelJob` is itself cancelled before it can cancel the underlying job?

std/shared/src/main/scala/cats/effect/std/JobManager.scala
BalmungSan: > I don't fully understand the comment about the map growing indefinitely

Yeah, the question wasn't really about how to model it but rather whether it should.
In order to be able to return a `Completed` status, we would need to preserve all IDs that have run, thus making the internal `Map` _"grow indefinitely"_, which smells like a memory leak.

I guess we could let the user decide which of the two options they want. But I do not want to complicate the API and implementation without satisfying a real use case.

In other words, the question was actually:
_"Would anyone actually need to distinguish completed from bad ID. And, if so, should we provide that for them instead of letting them build that for themselves?"_

std/shared/src/main/scala/cats/effect/std/JobManager.scala
BalmungSan: Good question.
I guess this whole block should be `uncancellable`? What do the Daniels think?

std/shared/src/main/scala/cats/effect/std/JobManager.scala
BalmungSan: Sorry, not sure I follow the wording of the question?

Are you asking if a user would be able to implement `getStatus` using a `Deferred`? If so, yes, I do not see why they couldn't; I also do not see why they would need to be constantly pooling?
If you are suggesting that we should provide a helper that would use a `Deferred` for them? Then, as with any other helper discussion, the question would be if the use case is common enough to deserve being put in the library.

std/shared/src/main/scala/cats/effect/std/JobManager.scala
mtomko: My name isn't Daniel, but my instinct offhand is that we just wrap the whole thing in uncancellable and call it a day. I don't see how any one piece of it could be cancelled without the other, but maybe I'm missing something.

std/shared/src/main/scala/cats/effect/std/JobManager.scala
mtomko: I don't think I can answer that question in general. I will say that in the use case I had in mind, the universe of possible IDs is not that big (maybe 100ish?) We might cancel a job and put in a replacement that would end up with the same ID though.

std/shared/src/main/scala/cats/effect/std/JobManager.scala
mtomko: Maybe I'm missing a detail in the implementation, but currently the way `startJob` works is that it launches `job.run` in the background using a supervisor and then places a `JobStatus` in the `MapRef`; there's a `Deferred` here for tracking when job _registration_ is complete, but the job could still be running.

Callers who just want `S` can only get it by calling `getJobStatus` over and over until it is `Some`. That's the polling I was referring to.

But I think you could also do something where the `MapRef` was `MapRef[F, Id, (F[Unit], Deferred[F, S])]` where the `F[Unit]` was the cancellation program. Then the existing `getJobStatus` could be served by `tryGet` on the `Deferred`. But you could also provide a variant that used `get` on the `Deferred` which would block until `job.run` completed the `Deferred`. That semantic blocking can't be done by a caller with the current API without the user implementing their own polling, I don't think.

Sorry, I haven't actually implemented the change yet, so the shape of these might not be 100% right.

mtomko: I can try to implement it tomorrow and see if the types work out. It's possible not everything can exist in the right order the way I was thinking of it.

std/shared/src/main/scala/cats/effect/std/JobManager.scala
durban: > What do the Daniels think?

LOL, is that a thing now...?

Anyway, I'd do a `guaranteeCase` instead of the `flatMap` (or, indeed, an `uncancelable`).

std/shared/src/main/scala/cats/effect/std/JobManager.scala
BalmungSan: Sorry, I still do not follow what would be the purpose here?

`Status` is not really meant as a way to get a result out of the `Job`, although it certainly could be used for that. Rather, it is intended to be able to inspect the progress of a `Job`.

> Callers who just want S can only get it by calling getJobStatus over and over until it is Some.

The very first call to `getJobStatus` would return `Some`, unless it is already completed by that time or you run the pooling and the start in parallel.

> But you could also provide a variant that used get on the Deferred which would block until job.run completed the Deferred.

I am not sure why anyone would want that? The whole point of this is to run something in the background, not to wait for it.

-----

However, this does connect with the point Daniel Spiewak made in his original review, that maybe we should not swallow the results and rather provide the users a handle over them. Similar to how `Supervisor` gives you the started `Fiber`.

So I am thinking that the correct API is something like this:

```scala
trait JobManager[F[_], Id, S, R] {
  // Starts job in the background.
  // Semantically blocks until the job starts, but does not wait for it to complete.
  def startJob(id: Id, job: Resource[F, JobManager.Job[F, S, R]]): F[Fiber[F, Throwable, R]]

  // Semantically blocks until the job associated with the given id is completed.
  // If the id is not associated with any job, completes immediately with a `None`.
  def join(id: Id): F[Option[Outcome[F, Throwable, R]]]

  // Polls for the status of a job.
  // If the id is not associated with any job, returns a `None`.
  def getJobStatus(id: Id): F[Option[S]]

  // Semantically blocks until the job is cancelled.
  // If the id is not associated with any job, it is a noop.
  def cancelJob(id: Id): F[Unit]
}
```

WDYT @djspiewak?

BalmungSan: While thinking about this idea, I found that the previous implementation had a bug where non-cancelled jobs were never cleaned up from the `jobsMap`.
Thus, I decided to fix that and adapt the API, since I am now convinced this is the most complete and flexible one.

std/shared/src/main/scala/cats/effect/std/JobManager.scala
BalmungSan: Fixed :)

std/shared/src/main/scala/cats/effect/std/JobManager.scala
mtomko: I think the confusion here may be somewhat related to the varying purposes we may have for this kind of implementation. In the Discord I published a trait that I had implemented and a description of the scenario I had in mind. In my case I'm not worrying about running jobs in the background at all, but managing the lifecycle of Resources that are long lived, but lazily initialized - in our case this could be things like connection pools or clients of upstream services.

PR #4290, opened 2025-03-05
Don't expose private WSTP type in method signature

### LINKED ISSUE #4283: Private type in a public signature
`IORuntime.createWorkStealingComputeThreadPool` is a public method. Its return type contains `WorkStealingThreadPool`, which is private.

armanbilge:
Fixes https://github.com/typelevel/cats-effect/issues/4283.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/IORuntimeCompanionPlatform.scala
@@ -20,7 +20,7 @@ import cats.effect.tracing.TracingConstants._
 import cats.effect.unsafe.metrics._
 
 import scala.collection.mutable
-import scala.concurrent.ExecutionContext
+import scala.concurrent.{ExecutionContext, ExecutionContextExecutor}
 import scala.concurrent.duration._
 
 import java.lang.management.ManagementFactory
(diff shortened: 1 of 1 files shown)


durban (approved):
Thanks!

Is it okay "cheating" bincompat like this? I realize that due to the tuple, it will link fine, but it _will_ cause a `ClassCastException`, right?

armanbilge:
> but it _will_ cause a `ClassCastException`, right?

Hmm, in what situation? As long as we actually keep returning a `cats.effect.unsafe.WorkStealingThreadPool` at _runtime_, if some client code has hard-coded that type, it will cast safely.

I did check, and it seems that (at least for Scala 3) the type can actually leak to client bytecode i.e.
```scala
val x = IORuntime.createWorkStealingComputeThreadPool()._1
```
compiles to
```
60: invokevirtual #81                 // Method scala/Tuple3._1:()Ljava/lang/Object;
63: checkcast     #83                 // class cats/effect/unsafe/WorkStealingThreadPool
```

durban:
Oh, right, it's still a WSTP at runtime :-) that seems fine then :+1:

djspiewak (approved):

armanbilge:
Oops, I think I meant to target this to `series/3.6.x`, anyway doesn't really matter.

PR #4295, opened 2025-03-07
Use `LinkedTransferQueue` for handling cached threads

armanbilge:
Replaces the `ConcurrentSkipListSet` with a `LinkedTransferQueue` for handling cached threads.
- we don't actually need any of the set-like features of the `ConcurrentSkipListSet`
- `LinkedTransferQueue` simplifies the logic: the transfer either succeeds or fails; no add-then-remove dance
- threads are polled in order of expiration, instead of in an arbitrary order based on their index, so caching will be more effective


djspiewak (approved):
Oh yeah this is way better

djspiewak:
I'm going to merge this directly into the SN wip branch so we don't have to wait for a release cycle. Since we merge rather than rebase, this won't create any conflicts

PR #4298, opened 2025-03-15
fix: Nix development Enviroment & Readme Fix

qxrein:
./build was failing inside nix development environment.

The main issues were:

- The `gifsicle` package can't run because it's dynamically linked
- `autoreconf` is missing, which is part of the `autoconf` package

As NixOS cannot run dynamically linked executables intended for generic linux environments out of the box. For more information, see:
https://nix.dev/permalink/stub-ld

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ To build the documentation site locally, the following dependencies are needed, 
 - Node (14.x ideally)
 - Yarn (any version should work)
 
-NOTE: [Nix](https://nixos.org/) users can just run `nix-shell` at the root directory and follow along the next instructions.
+NOTE: [Nix](https://nixos.org/) users can just run `nix develop` at the root directory and follow along the next instructions.
 
 Next, check out the documentation branch along with its submodules.
 
--- a/flake.nix
+++ b/flake.nix
@@ -24,6 +24,26 @@
             native.enable = true;
             nodejs.package = pkgs.nodejs-18_x;
           };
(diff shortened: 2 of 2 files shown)


armanbilge:
> ./build was failing inside nix development environment.

Can you explain more? There is no `./build` command for this repository.

qxrein:
sorry mb, it was the typo i am talking about `./build.sh` command in the `docs` branch being used to built site(to generate `build folder`)
read: https://github.com/typelevel/cats-effect?tab=readme-ov-file#contributing

djspiewak:
Any way we can add a build stage to catch this if it regresses again?

qxrein:
we can create a function inside build.sh to handle this

armanbilge:
> Any way we can add a build stage to catch this if it regresses again?

Well, the Right™ thing to do is run our CI inside of the Nix environment. This is what http4s does. Actually, this guarantees a more stable build environment where stuff won't randomly break, because every update to CI comes via explicit Nix shell updates.

qxrein:
i agree with @armanbilge

djspiewak (approved):

djspiewak:
I don't completely object to running CI within nix so long as it doesn't become mandatory for local development.

qxrein:
we can shift CI to Nix with that we can keep the old non-Nix builds going too.
Just like the current status of Nix in the repository.

PR #4302, opened 2025-03-16
Fix betweenDouble/betweenFloat

### LINKED ISSUE #4300: `Random#betweenDouble` and `#betweenFloat` overflow
https://github.com/scala/bug/issues/13098 also affects `cats.effect.std.Random`.

durban:
Fixes #4300. Further reading: https://github.com/scala-js/scala-js/pull/5142.

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Random.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Random.scala
@@ -331,25 +331,41 @@ object Random extends RandomCompanionPlatform {
     SecureRandom.javaSecuritySecureRandom[F].widen[Random[F]]
 
   private[std] sealed abstract class RandomCommon[F[_]: Sync] extends Random[F] {
-    def betweenDouble(minInclusive: Double, maxExclusive: Double): F[Double] =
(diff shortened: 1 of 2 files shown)


durban:
Uhh... is it possible that CI can't run on `series/3.5.x` any more?

```
Error: This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v3`. Learn more: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/
```

djspiewak:
> Uhh... is it possible that CI can't run on series/3.5.x any more?

Yeah that wouldn't actually surprise me :/ There's a whole tower of yaks there. I wouldn't object to targeting at series/3.6.x since we're about to release it

durban:
Okay, I've retargeted it to series/3.6.x; it picked up an extra commit, but it seems harmless(?).

(Although, this effectively seems to mean, that 3.5 is unmaintained...)

djspiewak:
Yeah we're in an awkward spot there. We could push things through despite the broken CI if we had to (and we would for something like a security issue) but it's officially unpleasant now.

djspiewak (approved):

PR #4316, opened 2025-03-24
Add `realTimeZonedDateTime` to `Clock` JVM extensions

mmienko:
Add realTimeZonedDateTime to Clock extensions for convenience

### DIFF
--- a/kernel/jvm-native/src/main/scala/cats/effect/kernel/ClockPlatform.scala
+++ b/kernel/jvm-native/src/main/scala/cats/effect/kernel/ClockPlatform.scala
@@ -16,10 +16,17 @@
 
 package cats.effect.kernel
 
-import java.time.Instant
+import java.time.{Instant, ZoneId, ZonedDateTime}
 
 private[effect] trait ClockPlatform[F[_]] extends Serializable { self: Clock[F] =>
   def realTimeInstant: F[Instant] = {
     self.applicative.map(self.realTime)(d => Instant.EPOCH.plusNanos(d.toNanos))
   }
+
+  def realTimeZonedDateTime: F[ZonedDateTime] =
+    self.applicative.map(realTimeInstant)(d => ZonedDateTime.ofInstant(d, ClockPlatform.UTC))
+}
+
+object ClockPlatform {
+  private val UTC = ZoneId.of("UTC")
 }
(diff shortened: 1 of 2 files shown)


djspiewak (dismissed):
I like this quite a lot, but reading the diff, I just realized that we split `ClockPlatform` to JVM + Native, while `IOCompanion` is only split to JVM. We should make this consistent, though it may take a bit more refactoring. My guess is we want to revisit this after #4201 lands.

Would you mind opening an issue to track this? Then we can merge this PR.

mmienko:
Oh yes, good callout. Will create issue to track.

mmienko:
Done https://github.com/typelevel/cats-effect/issues/4323

kernel/jvm-native/src/main/scala/cats/effect/kernel/ClockPlatform.scala
```diff
@@ -16,10 +16,17 @@
...
 private[effect] trait ClockPlatform[F[_]] extends Serializable { self: Clock[F] =>
   def realTimeInstant: F[Instant] = {
     self.applicative.map(self.realTime)(d => Instant.EPOCH.plusNanos(d.toNanos))
   }
+
+  def realTimeZonedDateTime: F[ZonedDateTime] =
+    self.applicative.map(realTimeInstant)(d => ZonedDateTime.ofInstant(d, ClockPlatform.UTC))
+}
+
+object ClockPlatform {
```
armanbilge:
```suggestion
private[effect] object ClockPlatform {
```

mmienko: thanks

kernel/jvm-native/src/main/scala/cats/effect/kernel/ClockPlatform.scala
```diff
@@ -16,10 +16,17 @@
...
-import java.time.Instant
+import java.time.{Instant, ZoneId, ZonedDateTime}
 
 private[effect] trait ClockPlatform[F[_]] extends Serializable { self: Clock[F] =>
   def realTimeInstant: F[Instant] = {
     self.applicative.map(self.realTime)(d => Instant.EPOCH.plusNanos(d.toNanos))
   }
+
+  def realTimeZonedDateTime: F[ZonedDateTime] =
+    self.applicative.map(realTimeInstant)(d => ZonedDateTime.ofInstant(d, ClockPlatform.UTC))
```
armanbilge:
```suggestion
    self.applicative.map(realTimeInstant)(d => ZonedDateTime.ofInstant(d, ZoneOffset.UTC))
```

armanbilge (approved):

PR #4317, opened 2025-03-24
`unsafeRunTimed`

qxrein:
Using `unsafeRunAndForget()` means the cancellation operation is started but not waited for. This can lead to resource leaks since finalizers might not complete properly.


armanbilge:
Thanks! You are right, this can lead to resource leaks. It's actually part of the specification of this method (see the documentation). 

https://github.com/typelevel/cats-effect/blob/971a79ef8185629f2dd86c51535a0746412e4412/core/jvm/src/main/scala/cats/effect/IOPlatform.scala#L51-L63

qxrein:
sorry for making the wrong PR

PR #4320, opened 2025-03-25
Remove TODO in `IOFiber.readBarrier()`

### LINKED ISSUE #4321: Remove TODO in `IOFiber.readBarrier()`
#4320 
benchmark reference:
```
withBarrier (volatile)   : ~5.2M ops/s  
withoutBarrier (plain)   : ~5.3M ops/s  (unsafe for fibers)  
javaFullFence            : ~800K ops/s  (overkill)  
```

optimizing further makes no sense because `VarHandle/Unsafe` will introduce complexity for negligible gain. `Plain reads` break correctness. Volatile is just perfect.

qxrein:
Resolves: #4321

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -1373,10 +1373,18 @@ private final class IOFiber[A](
     }
   }
 
-  // TODO figure out if the JVM ever optimizes this away
+  /**
+   * this is specially for JDK8 compatibility, when the JDK 8 support is eventually dropped,
+   * loadfence() try/catch could be replaced with VarHandle.acquireFence()
+   */
   private[this] def readBarrier(): Unit = {
-    suspended.get()
-    ()
+    if (Platform.isJvm) {
+      // Use the most efficient available barrier for the platform
+      try sun.misc.Unsafe.getUnsafe().loadFence()
+      catch { case _: Throwable => val _ = suspended.get() } // Fallback
+    } else {
+      val _ = suspended.get() // Non-JVM platforms
+    }
   }
 
   /* Implementations of resume methods */


armanbilge:
Thanks for the PR! The TODO is not about the potential for further optimization, but rather the concern that the JVM/Hotspot would optimize away the `suspended.get()` (since we discard the result) thus losing the memory effect. If we didn't still support JDK8 we could possibly replace it with `VarHandle.acquireFence` (not a full fence).

qxrein:
ok, but seems like we still maintain some compatibility with JDK8. But we can try to use an hybrid `Unsafe.loadFence()` + `AtomicBoolean` fallback approach, this is giving me intiution that using` loadfence` could speedup over raw `AtomicBoolean` on JVMs. Let me try something.

qxrein:
https://github.com/typelevel/cats-effect/issues/4321#issuecomment-2757824624

core/shared/src/main/scala/cats/effect/IOFiber.scala
```diff
@@ -1373,10 +1373,18 @@ private final class IOFiber[A](
...
+  /**
+   * this is specially for JDK8 compatibility, when the JDK 8 support is eventually dropped,
+   * loadfence() try/catch could be replaced with VarHandle.acquireFence()
+   */
   private[this] def readBarrier(): Unit = {
-    suspended.get()
-    ()
+    if (Platform.isJvm) {
+      // Use the most efficient available barrier for the platform
+      try sun.misc.Unsafe.getUnsafe().loadFence()
```
armanbilge, line 1383: I don't think this actually works (i.e. it will always throw an exception).

Anyway, we don't use `Unsafe`, partly for the reasons described here:
https://scala-lang.org/blog/next-scala-lts.html#immediate-motivation

qxrein: hmm looks like we'd have to upgrade the JDK compatibility of cats-effect

PR #4322, opened 2025-03-25
Add Lock to std

### LINKED ISSUE #3535: Add `Lock` to std

sebastianrozyczka:
Resolves #3535

### DIFF
--- a/std/shared/src/main/scala/cats/effect/std/Lock.scala
+++ b/std/shared/src/main/scala/cats/effect/std/Lock.scala
@@ -1,0 +1,357 @@
+/*
+ * Copyright 2020-2025 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


std/shared/src/main/scala/cats/effect/std/Lock.scala
```diff
@@ -0,0 +1,312 @@
...
+
+        override def exclusive: Resource[F, Unit] = {
+          Claim.resource(identityProvider).flatMap { claim =>
+            def acquire: F[Unit] = F.uncancelable { poll =>
+              ref.modify {
+                case State(None, _, _) =>
+                  State(Current.Exclusive(claim).some, Q.empty, Q.empty) -> Right(())
+
+                // Reentrant access
+                case state @ State(Some(Current.Exclusive(running)), _, _)
```
sebastianrozyczka: not sure if we should we upgrade if it's held by Shared?

Issue #4314, opened 2025-03-24
Implement `PollerMetrics` for `EpollSystem` and `KqueueSystem`


armanbilge:

antoniojimeneznieto:
Hello, I can help with this issue! I'll start working on it and share my progress soon :)

PR #4325, opened 2025-03-26
Implement PollerMetrics for EpollSystem

antoniojimeneznieto:
Linked to: #4314

This PR implements metrics tracking for read and write operations in the Poller class for the EpollSystem. It provides basic metrics reporting such as total submitted, succeeded, errored, and canceled operations and outstanding operations count.

### DIFF
--- a/core/native/src/main/scala/cats/effect/unsafe/EpollSystem.scala
+++ b/core/native/src/main/scala/cats/effect/unsafe/EpollSystem.scala
@@ -35,6 +35,8 @@ import scala.scalanative.unsigned._
 
 import java.io.IOException
 import java.util.{Collections, IdentityHashMap, Set}
+import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.atomic.AtomicLong
 
 object EpollSystem extends PollingSystem {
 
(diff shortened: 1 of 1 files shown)


core/native/src/main/scala/cats/effect/unsafe/EpollSystem.scala
```diff
@@ -181,6 +181,122 @@ object EpollSystem extends PollingSystem {
...
+
+      override def totalOperationsSucceededCount(): Long =
+        totalReadSucceeded + totalWriteSucceeded
+
+      override def totalOperationsErroredCount(): Long = totalReadErrored + totalWriteErrored
+
+      override def totalOperationsCanceledCount(): Long = totalReadCanceled + totalWriteCanceled
+
+      override def acceptOperationsOutstandingCount(): Int =
+        throw new UnsupportedOperationException("Accept operation metrics are not supported.")
```
iRevive: An exception can break existing integrations. Perhaps we can return 0 instead? What do you think?

antoniojimeneznieto: Absolutely!

djspiewak:
Can we target this at series/3.6.x?

PR #4326, opened 2025-03-26
Add doc to metrics() in PollingSystem

antoniojimeneznieto:
This PR adds documentation to the `metrics()` method in the `PollingSystem` api


djspiewak (dismissed):
This is great! Can you retarget to `series/3.6.x` though so we can release it sooner?

PR #4334, opened 2025-03-30
Document `PollingSystem#metrics`

antoniojimeneznieto:


core/jvm-native/src/main/scala/cats/effect/unsafe/PollingSystem.scala
```diff
@@ -115,6 +115,18 @@ abstract class PollingSystem {
...
+   * Retrieves the metrics associated with the specified poller.
+   *
+   * Metrics include counts of submitted, succeeded, errored, and canceled operations for
+   * accept, connect, read, and write events. Additionally, it provides counts of outstanding
+   * operations and total operations.
+   *
+   * @param poller
+   *   The poller for which metrics are being retrieved.
+   * @return
+   *   An instance of PollerMetrics representing the current metrics of the poller.
```
armanbilge: It's subtle, but I think this description is slightly confusing. This method doesn't return the current metrics of the poller (i.e. `PollerMetrics` is not an immutable datatype of metrics). Instead, it returns a `PollerMetrics` instance that can be used to (unsafely, side-effectually) retrieve the current metrics for that poller.

armanbilge (approved):
thanks!

PR #4335, opened 2025-03-31
Implement PollerMetrics for EpollSystem

antoniojimeneznieto:
Linked to: https://github.com/typelevel/cats-effect/issues/4314

This PR implements metrics tracking for read and write operations in the Poller class for the EpollSystem. It provides basic metrics reporting such as total submitted, succeeded, errored, and canceled operations and outstanding operations count.

### DIFF
--- a/core/native/src/main/scala/cats/effect/unsafe/EpollSystem.scala
+++ b/core/native/src/main/scala/cats/effect/unsafe/EpollSystem.scala
@@ -247,6 +247,7 @@ object EpollSystem extends PollingSystem {
 
       override def totalWriteOperationsCanceledCount(): Long = totalWriteCanceled
 
+      override def toString: String = "Epoll"
     }
 
     private[EpollSystem] def metrics(): PollerMetrics = pollerMetrics
--- a/core/native/src/main/scala/cats/effect/unsafe/EpollSystem.scala
+++ b/core/native/src/main/scala/cats/effect/unsafe/EpollSystem.scala
@@ -69,7 +69,7 @@ object EpollSystem extends PollingSystem {
 
   def interrupt(targetThread: Thread, targetPoller: Poller): Unit = ()
 
-  def metrics(poller: Poller): PollerMetrics = PollerMetrics.noop
+  def metrics(poller: Poller): PollerMetrics = poller.metrics()
 
   private final class FileDescriptorPollerImpl private[EpollSystem] (
       ctx: PollingContext[Poller])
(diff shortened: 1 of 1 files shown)


core/native/src/main/scala/cats/effect/unsafe/EpollSystem.scala
```diff
@@ -181,6 +181,114 @@ object EpollSystem extends PollingSystem {
...
+
+      override def totalWriteOperationsSubmittedCount(): Long = totalWriteSubmitted
+
+      override def totalWriteOperationsSucceededCount(): Long = totalWriteSucceeded
+
+      override def totalWriteOperationsErroredCount(): Long = totalWriteErrored
+
+      override def totalWriteOperationsCanceledCount(): Long = totalWriteCanceled
+
+    }
```
iRevive: Follow-up to https://github.com/typelevel/cats-effect/pull/4365. 
```suggestion
      override def toString: String = "Epoll"
    }
```

djspiewak:
IMO this (and #4336) are very cool and look good and I don't see anything standing in the way of their merge. @armanbilge thoughts? @antoniojimeneznieto do you feel these are in a state where they can be non-WIP PRs?

core/native/src/main/scala/cats/effect/unsafe/EpollSystem.scala
```diff
@@ -181,6 +181,115 @@ object EpollSystem extends PollingSystem {
...
+    private var totalReadCanceled = 0L
+    private var readOutstanding = 0
+
+    private var totalWriteSubmitted = 0L
+    private var totalWriteSucceeded = 0L
+    private var totalWriteErrored = 0L
+    private var totalWriteCanceled = 0L
+    private var writeOutstanding = 0
+
+    private[this] val pollerMetrics: PollerMetrics = new PollerMetrics {
```
armanbilge: Instead of an internal object, we can have `Poller extends PollerMetrics` and implement these methods directly :)

antoniojimeneznieto: That makes sense! Thank you arman :)

antoniojimeneznieto:
> IMO this (and #4336) are very cool and look good and I don't see anything standing in the way of their merge. @armanbilge thoughts? @antoniojimeneznieto do you feel these are in a state where they can be non-WIP PRs?

Yes, thank you! I believe they're ready and no longer WIP

djspiewak (approved):

PR #4340, opened 2025-04-03
Also interrupt pollers when shutting down the WSTP

durban:
As discussed in #4201.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
@@ -710,6 +710,7 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
       while (i < threadCount) {
         val workerThread = workerThreads.get(i)
         if (workerThread ne currentThread) {
+          system.interrupt(workerThread, pollers(i))
           workerThread.interrupt()
         }
         i += 1
--- a/tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
+++ b/tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
@@ -44,7 +44,6 @@ import java.util.concurrent.{
   ThreadLocalRandom
 }
 import java.util.concurrent.atomic.{AtomicBoolean, AtomicInteger, AtomicLong, AtomicReference}
-import java.util.concurrent.locks.LockSupport
 
 trait IOPlatformSpecification extends DetectPlatform { self: BaseSpec with ScalaCheck =>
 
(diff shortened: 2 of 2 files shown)


core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -710,6 +710,7 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
       while (i < threadCount) {
         val workerThread = workerThreads.get(i)
         if (workerThread ne currentThread) {
+          system.interrupt(workerThread, pollers(i))
           workerThread.interrupt()
```
armanbilge, line 714: I am wondering if we can just replace `workerThread.interrupt()` entirely, and if not, if there's a specific reason to call the interrupts in this order?

I think `system.interrupt` should be sufficient, it's the only mechanism by which a worker thread can park.

armanbilge (commented):
Thanks! Want to add a test? 🤠

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
durban: I mean... `Thread#sleep` is another mechanism :man_shrugging: 

I was also wondering about the order, I asked in the native PR (I mean #4201). I don't know.

durban:
I _was_ thinking about a test... but this is only JVM (in this branch), and on the JVM it already works correctly. So what would we test?

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
armanbilge: > I mean... `Thread#sleep` is another mechanism 🤷‍♂️

Right, but that should never happen on a worker thread (outside of `SleepSystem`, but in that case `SleepSystem.interrupt` does the right thing).

armanbilge:
> and on the JVM it already works correctly

It doesn't work correctly on the JVM, it just so happens that our current polling systems happen to respect `Thread#interrupt`. But a different polling system (such as the io_uring one) would not work correctly. It should be possible to construct a polling system for the test that doesn't respect `Thread#interrupt` e.g. by swallowing the exception and looping.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
durban: Yeah, it _should_ never happen... in _theory_... @djspiewak was arguing in https://github.com/typelevel/cats-effect/pull/4201#issuecomment-2773760120, that we _should_ still call `Thread#interrupt`.

armanbilge: Ahh, sorry, I missed that. Fair enough.

> I do think we should continue interrupting the threads, since that allows us to catch a few cases where we aren't blocked in a poller (especially on the JVM) and it's best practice to set those flags.

djspiewak:
I think the best way to test this is probably to make a mock polling system that just mutates a flag when it gets interrupted. Then make a runtime, wait a second, and shut it down.

durban:
Yeah, uh... we _could_ do that. But I was already almost done with this much more convoluted test, when you wrote that comment... :-) This way at least we're testing a somewhat realistic shutdown procedure, if I didn't mess it up. (But let me know it it's too much.)

djspiewak (changes requested):
LOL! I mean, this is magnificent, but I think it'll be easier to maintain if we do the mock interrupt thing. We can even make it deterministic because we can release a latch when we "sleep", which then would trigger the test to shutdown the runtime. Or we could even shut down the runtime directly from inside the mock system, that type of thing.

djspiewak (approved):
Excellent!

PR #4341, opened 2025-04-03
Make deprecated `IO.onError` package private

### LINKED ISSUE #4329: Allow `IO.onError` to be passed a total function
#4121 deprecated `IO.onError(f: Throwable => IO[Unit])` and added `def onError(pf: PartialFunction[Throwable, IO[Unit]])` to be used instead.

[Normally total functions can be passed in place of a partial function and Scala will automatically create a partial function wrapper for it](https://scastie.scala-lang.org/4yClxhLmQ5mq8ZoO6f1Oag). In other words, we should be able to call the partial function variant of `onError` with a total function. [Here's a scastie snippet showing that expected behaviour of `onError` with `Resource`](https://scastie.scala-lang.org/JGfAkqBCTRu82iLCMDMSeA).

Unfortunately, [when calling `IO.onError` with a total function in CE 3.6.0, there is deprecation warning](https://scastie.scala-lang.org/FxT9UQcvSSyDXP06okStYQ). This is inconsistent with earlier versions or the behaviour of `ApplicativeError.onError`.

It'd be great to make `IO.onError` not generate a deprecation warning when passed a total function.

[Proposed fix from discord](https://discord.com/channels/632277896739946517/632278585700384799/1355235154351620257) is to mark the deprecated version of `onError` as package private.

CJSmith-0141:
Also adds a test to make sure that code outside the package can still compile with a total function. 

Closes #4329 

Thank you @armanbilge for the solve and @Grogs for the writeup.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -591,7 +591,7 @@ sealed abstract class IO[+A] private () extends IOPlatform[A] {
     IO.OnCancel(this, fin)
 
   @deprecated("Use onError with PartialFunction argument", "3.6.0")
-  def onError(f: Throwable => IO[Unit]): IO[A] = {
+  private[effect] def onError(f: Throwable => IO[Unit]): IO[A] = {
     val pf: PartialFunction[Throwable, IO[Unit]] = { case t => f(t).reportError }
     onError(pf)
   }
--- a/tests/shared/src/test/scala-2.13+/not/cats/effect/IOCompilationSpec.scala
+++ b/tests/shared/src/test/scala-2.13+/not/cats/effect/IOCompilationSpec.scala
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2020-2025 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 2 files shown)


CJSmith-0141:
Looks like making it package private doesn't work for `2.12`. Maybe un deprecate instead?

joroKr21:
I think on 2.12 you can't use a total function in place of a partial function. So that test would have to be only for 2.13+

CJSmith-0141:
Ok since this would be a source breaking change for Scala `2.12`, I changed the base branch for this PR to be `series/3.x`.

I also moved the test to be 2.13+.

joroKr21:
The previous change was already source breaking because Scala 2.12 can't infer the parameter type of the function passed to an overloaded method like `onError { case e => ... }` will complain about unknown function parameter type.

armanbilge (approved):
thanks!

djspiewak:
@armanbilge Are we going to retarget this to 3.6?

armanbilge (changes requested):
> Are we going to retarget this to 3.6?

Yes!

armanbilge:
superseded https://github.com/typelevel/cats-effect/pull/4343

PR #4343, opened 2025-04-03
Make deprecated `IO.onError` package private

### LINKED ISSUE #4329: Allow `IO.onError` to be passed a total function

CJSmith-0141:
This is a repeat of #4341 targeting `series/3.6.x` instead. Fine with closing this one or the other one, no worries :) 

closes #4329

This is a source breaking change for Scala 2.12

### DIFF
--- a/tests/shared/src/test/scala-2.13+/not/cats/effect/IOCompilationSpec.scala
+++ b/tests/shared/src/test/scala-2.13+/not/cats/effect/IOCompilationSpec.scala
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2020-2025 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package not.cats.effect // verifies scoping outside of CE
+
+import cats.effect.{BaseSpec, IO}
+
+class IOCompilationSpec extends BaseSpec {
+  "Can use total function in PartialFunction IO.onError" in { // compilation test
+    IO.pure(5).onError(_ => IO.unit)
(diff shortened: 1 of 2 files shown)


tests/shared/src/test/scala-2.13+/not/cats/effect/IOCompilationSpec.scala
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package not.cats.effect // verifies scoping outside of CE
+
+import cats.effect.{BaseSpec, IO}
+
+class IOCompilationSpec extends BaseSpec {
```
djspiewak, line 21: Can just put this into `IOSpec`.

djspiewak (changes requested):

tests/shared/src/test/scala-2.13+/not/cats/effect/IOCompilationSpec.scala
CJSmith-0141: When it's in IOSpec it picks up the deprecated method. Is that ok?

djspiewak: Ooooh I see, this is outside the package. Okay nvm

djspiewak (approved):

Issue #4337, opened 2025-03-31
IO.syncStep(n).unsafeRunSync() fails with StackOverflowError


migesok:
This code works for me on 3.5.7 but fails on 3.6.0 (OpenJDK 21 on M1 Mac OS):
```
import cats.effect.*
import cats.syntax.all.*

object Main extends App {
  val syncStepSize = 1000000

  val io = for {
    ref <- Ref.of[IO, Int](0)
    _ <- Vector.fill(100000)(1).traverse_[IO, Unit](n => ref.update(_ + n))
    result <- ref.get
  } yield result

  println(io.syncStep(syncStepSize).unsafeRunSync())
}
```
```
Exception in thread "main" java.lang.StackOverflowError
	at cats.effect.IO$FlatMap.ioe(IO.scala:2235)
	at cats.effect.SyncStep$.interpret(IO.scala:2355)
	at cats.effect.SyncStep$.interpret(IO.scala:2356)
	at cats.effect.SyncStep$.interpret(IO.scala:2356)
	at cats.effect.SyncStep$.interpret(IO.scala:2356)
	at cats.effect.SyncStep$.interpret(IO.scala:2356)
	at cats.effect.SyncStep$.interpret(IO.scala:2356)
	at cats.effect.SyncStep$.interpret(IO.scala:2356)
        ...
```

I'm not sure if it's a bug or an expected behaviour. It was sure unexpected for me 😄 

I tried to debug it a little bit but got lost in `syncStep` and `SyncIO` implementations.

durban:
Thanks for the report! Can you tell us the versions of Cats in the two cases?

As far as I can tell, the `syncStep` interpreter is stack-unsafe for a long time. However, I think `traverse` (in Cats) used to have extra things for stack safety, but now it detects IO (or similar), and assumes they have a stack-safe `flatMap`. And `IO#flatMap` _is_ stack-safe... if executed by the "normal" interpreter. It seems it is not, if executed by the `syncStep` interpreter. (So that's why I'm asking for Cats versions.)

In any case, we should fix this.

migesok:
In this project I didn't specify the cats version explicitly so I get the one defined by cats-effect:
- for 3.6.0: `sbt: org.typelevel:cats-core_2.13:2.11.0:jar`
- for 3.5.7: `sbt: org.typelevel:cats-core_2.13:2.9.0:jar`

Is it possible to make `syncStep` stack-safe or is it by design?
[...]

PR #4352, opened 2025-04-04
Make `IO#syncStep` interpreter stack-safe

pantShrey:
Linked to: [#4337](https://github.com/typelevel/cats-effect/issues/4337)

This PR adds stack safety to SyncStep.interpret by wrapping execution in defer every MaxSteps steps. This prevents stack overflows in large recursive computations while still respecting the original step limit as a hard cap.

Testing:
Validated locally with a 100,000-step Ref update chain and a limit = 1,000,000 (syncStepSize). The computation now completes with Right(100000), resolving the previously observed StackOverflowError.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -2334,7 +2334,7 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits with TuplePara
 }
 
 private object SyncStep {
-  val MaxSteps = 512 // 512 matches  with trampoline depth in IOFiber for stack safety
+  final val MaxSteps = 512 // 512 matches  with trampoline depth in IOFiber for stack safety
 
   def interpret[G[+_], B](io: IO[B], limit: Int, stepsUntilDefer: Int)(
       implicit G: Sync[G]): G[Either[IO[B], (B, Int)]] = {
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -2334,10 +2334,13 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits with TuplePara
 }
 
 private object SyncStep {
-  def interpret[G[+_], B](io: IO[B], limit: Int)(
+  private val MaxSteps = 512
+  def interpret[G[+_], B](io: IO[B], limit: Int, stepsUntilDefer: Int = MaxSteps)(
       implicit G: Sync[G]): G[Either[IO[B], (B, Int)]] = {
     if (limit <= 0) {
       G.pure(Left(io))
+    } else if (stepsUntilDefer <= 0) {
+      G.defer(interpret(io, limit, MaxSteps))
     } else {
       io match {
         case IO.Pure(a) => G.pure(Right((a, limit)))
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -2334,12 +2334,9 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits with TuplePara
 }
 
 private object SyncStep {
-  private val MaxSteps = 512
-  def interpret[G[+_], B](io: IO[B], limit: Int)(
-      implicit G: Sync[G]): G[Either[IO[B], (B, Int)]] = {
-    interpret(io, limit, MaxSteps)
-  }
-  def interpret[G[+_], B](io: IO[B], limit: Int, stepsUntilDefer: Int = MaxSteps)(
+  val MaxSteps = 512 // 512 matches  with trampoline depth in IOFiber for stack safety
```
armanbilge: This allows the compiler to inline it
```suggestion
  final val MaxSteps = 512 // 512 matches  with trampoline depth in IOFiber for stack safety
```

pantShrey: oh ok thank you, am adding it

core/shared/src/main/scala/cats/effect/IO.scala
```diff
@@ -2334,10 +2334,14 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits with TuplePara
 }
 
 private object SyncStep {
-  def interpret[G[+_], B](io: IO[B], limit: Int)(
+  final val MaxSteps = 512 // 512 matches  with trampoline depth in IOFiber for stack safety
+
+  def interpret[G[+_], B](io: IO[B], limit: Int, stepsUntilDefer: Int)(
```
armanbilge: Another optimization, so that it compiles as a static method.

```suggestion
  @static def interpret[G[+_], B](io: IO[B], limit: Int, stepsUntilDefer: Int)(
```

pantShrey: adding static is failing compilation as `SyncStep` is private object and `@static` requires a companion class

armanbilge:
> Validated locally with a 100,000-step Ref update chain and a limit = 1,000,000 (syncStepSize). The computation now completes with Right(100000), resolving the previously observed StackOverflowError.

Great! Can we add something similar to this as a test?

pantShrey:
@armanbilge Added the test for large Ref update chains.

tests/shared/src/test/scala/cats/effect/IOSpec.scala
```diff
@@ -2130,6 +2130,15 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
           .syncStep(Int.MaxValue)
         sio.map(_.bimap(_ => (), _ => ())) must completeAsSync(Right(()))
       }
+
+      "handle large sequence of operations without StackOverflowError #4337" in {
+        val io = IO.unit.replicateA_(100000)
+        io.syncStep(1000000).map {
+          case Left(_) => throw new RuntimeException("Expected synchronous completion")
+          case Right(()) => ()
+        } must completeAsSync(())
```
armanbilge:
```suggestion
        io.syncStep(Int.MaxValue) must completeAsSync(Right(()))
```

tests/shared/src/test/scala/cats/effect/IOSpec.scala
```diff
@@ -2130,6 +2130,15 @@ class IOSpec extends BaseSpec with Discipline with IOPlatformSpecification {
           .syncStep(Int.MaxValue)
         sio.map(_.bimap(_ => (), _ => ())) must completeAsSync(Right(()))
       }
+
+      "handle large sequence of operations without StackOverflowError #4337" in {
+        def go(depth: Int, acc: IO[Unit] = IO.unit): IO[Unit] =
+          if (depth <= 0) acc else go(depth - 1, acc.flatMap(_ => IO.unit))
+        val io = go(50000)
+        io.syncStep(Int.MaxValue).unsafeRunSync() must beRight(())
```
armanbilge:
```suggestion
        io.syncStep(Int.MaxValue) must completeAsSync(Right(()))
```

pantShrey: > 

this fails and returns error
Error: could not find implicit value for evidence parameter of type cats.Eq[Either[cats.effect.IO[Unit],Unit]]
[error]         io.syncStep(Int.MaxValue) must completeAsSync(Right(()))

pantShrey: other option i can try is 

io.syncStep(Int.MaxValue).map(_.bimap(_ => (), _ => ())) must completeAsSync(Right(()))

armanbilge: We should definitely prefer `completeAsSync` instead of `unsafeRunSync`, so let's find a way to make it work, thanks!

pantShrey: > We should definitely prefer `completeAsSync` instead of `unsafeRunSync`, so let's find a way to make it work, thanks!

io.syncStep(Int.MaxValue).map(.bimap( => (), _ => ())) must completeAsSync(Right(()))

This is working locally , took reference from the other test cases

is this fine?

armanbilge (approved):
nice work, thank you!

Issue #4331, opened 2025-03-30
Implement `interruptible` on SN


djspiewak:
Follow up from #4201 

We didn't port over the `interruptible` implementation in the initial pass. This is definitely doable, though it's not clear how relevant it'll be in practice. We should probably just have it for completeness and principle of least surprise.

PR #4360, opened 2025-04-05
`interruptible` on SN for #4331

qxrein:
#4331

### DIFF
--- a/core/native/src/main/scala/cats/effect/unsafe/EpollSystem.scala
+++ b/core/native/src/main/scala/cats/effect/unsafe/EpollSystem.scala
@@ -74,21 +74,15 @@ object EpollSystem extends PollingSystem {
     val wasInterrupted = Thread.interrupted()
 
     try {
-      if (wasInterrupted) Thread.currentThread().interrupt()
-
       try {
         thunk
       } catch {
-        case e: InterruptedException =>
-          Thread.currentThread().interrupt()
-          throw new IOException("Operation was interrupted", e)
-        case e: NoSuchElementException
-            if e.getMessage == "interrupted" || e.getMessage == "None.get" =>
-          // SN throws this for interruptions - preserve but don't wrap
+        case e: NoSuchElementException if e.getMessage == "interrupted" =>
           Thread.currentThread().interrupt()
           throw e
-      } finally {
-        val _ = Thread.interrupted()
+        case e: InterruptedException =>
+          Thread.currentThread().interrupt()
+          throw e
       }
     } finally {
       if (wasInterrupted) Thread.currentThread().interrupt()
--- a/core/native/src/main/scala/cats/effect/unsafe/EpollSystem.scala
+++ b/core/native/src/main/scala/cats/effect/unsafe/EpollSystem.scala
@@ -70,6 +70,31 @@ object EpollSystem extends PollingSystem {
     targetPoller.interrupt()
   }
 
+  def interruptible[A](thunk: => A): A = {
+    val wasInterrupted = Thread.interrupted()
+
+    try {
+      if (wasInterrupted) Thread.currentThread().interrupt()
+
+      try {
+        thunk
+      } catch {
+        case e: InterruptedException =>
+          Thread.currentThread().interrupt()
+          throw new IOException("Operation was interrupted", e)
+        case e: NoSuchElementException
+            if e.getMessage == "interrupted" || e.getMessage == "None.get" =>
+          // SN throws this for interruptions - preserve but don't wrap
+          Thread.currentThread().interrupt()
+          throw e
+      } finally {
+        val _ = Thread.interrupted()
+      }
+    } finally {
+      if (wasInterrupted) Thread.currentThread().interrupt()
+    }
+  }
+
   def metrics(poller: Poller): PollerMetrics = PollerMetrics.noop
 
   private final class FileDescriptorPollerImpl private[EpollSystem] (
--- a/core/native/src/main/scala/cats/effect/unsafe/KqueueSystem.scala
+++ b/core/native/src/main/scala/cats/effect/unsafe/KqueueSystem.scala
@@ -69,6 +69,30 @@ object KqueueSystem extends PollingSystem {
   def interrupt(targetThread: Thread, targetPoller: Poller): Unit =
     targetPoller.interrupt()
 
+  def interruptible[A](thunk: => A): A = {
+    val wasInterrupted = Thread.interrupted()
+
+    try {
+      if (wasInterrupted) Thread.currentThread().interrupt()
+
+      try {
+        thunk
+      } catch {
+        case e: InterruptedException =>
+          Thread.currentThread().interrupt()
+          throw new IOException("Operation was interrupted", e)
+        case e: NoSuchElementException
+            if e.getMessage == "interrupted" || e.getMessage == "None.get" =>
+          Thread.currentThread().interrupt()
+          throw e
+      } finally {
+        val _ = Thread.interrupted()
+      }
+    } finally {
+      if (wasInterrupted) Thread.currentThread().interrupt()
+    }
+  }
+
   def metrics(poller: Poller): PollerMetrics = PollerMetrics.noop
 
   private final class FileDescriptorPollerImpl private[KqueueSystem] (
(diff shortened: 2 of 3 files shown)


djspiewak (changes requested):
I think this is probably not what we were thinking in this case! Instead, we should port and share the existing JVM interruptible, since the point is to interoperate with any code which is leveraging thread interruption.

qxrein:
lol ok, I was initially confused for the same cause. OK i'll do that.

qxrein:
current situation for JVM `interruptible` kinda messy. In the JVM, when we call `Thread.interrupt()`, it’s super reliable: it flips the thread’s interrupt flag and, if the thread’s stuck in some blocking operation, it throws an `InterruptedException`.  But then we get to SN, Here, 'interruptions' often throw a `NoSuchElementException` instead, and the interrupt flag? It’s not consistently kept up to date. 
I believe we can’t just recreate the JVM’s interruption system in SN because the basic building blocks and guarantees we need aren’t there. yeah, we can interrupt operations, but it doesn’t stick to the JVM’s thread state rules down to the letter. It’s not really a flaw in how it’s built; it’s more like a deep-rooted limit in how SN handles concurrency. 

Let me also commit the testcase i tried to build on which this thing is happening.

qxrein:
testcase will fail due to some comments so ignore it for a while

djspiewak:
> current situation for JVM interruptible kinda messy. In the JVM, when we call Thread.interrupt(), it’s super reliable: it flips the thread’s interrupt flag and, if the thread’s stuck in some blocking operation, it throws an InterruptedException. But then we get to SN, Here, 'interruptions' often throw a NoSuchElementException instead, and the interrupt flag? It’s not consistently kept up to date.

Wow this is kind of nuts. @WojciechMazur is this something which is on your radar? (we can also file an issue if one doesn't exist) I don't mind *not* supporting `interruptible` for Scala Native, I just want to make sure that if the intention is for SN to match JVM thread interruption semantics, then we should provide the relevant support on our end.

WojciechMazur:
>  if the thread’s stuck in some blocking operation, it throws an InterruptedException 

Does it really? I don't think so. The interrupt on it's own does only flip flag and interupts threads blocked on NIO via [sun.misc.ch.Interruptible](https://github.com/openjdk/jdk/blob/cc546e7a283faad40b75183887f6a1f433a2407f/src/java.base/share/classes/sun/nio/ch/Interruptible.java) 
We can test that it would not interrupt normal, blocking, execution of the thread here 2 examples: using FFI and blocking IO 
In both cases InterruptException would not be thrown 

```scala
//> using jvm 23
//> using javaOpt --enable-preview

import java.lang.foreign.*

@main def ForeignFunctionExample = {
  val linker: Linker = Linker.nativeLinker()
  val sleep = linker.downcallHandle(
    linker.defaultLookup.find("sleep").orElseThrow(),
    FunctionDescriptor.of(ValueLayout.JAVA_INT, ValueLayout.JAVA_LONG)
  )

  val t1 = Thread.ofPlatform().start: () =>
    println("sleep")
    val length: Long = sleep.invoke(15)
    println("done")
  
  Thread.sleep(1000)
  println("interrupt")
  t1.interrupt()
  t1.join(1000)
  println(s"joined: ${t1.getState()}")
}

@main def StdioExample() = {
  val t1 = Thread.ofPlatform().start: () =>
    val result = scala.io.StdIn.readLine("readline:")
    println(s"done: $result")
  
  Thread.sleep(1000)
  println("interrupt")
  t1.interrupt()
  t1.join(1000)
  println(s"joined: ${t1.getState()}")
}
```


> Here, 'interruptions' often throw a NoSuchElementException instead, and the interrupt flag? 

It's the first time I'm reading about `NoSuchElementException` in case of interrupt. I'd like to see some reproducer for that. If it does actually lead to throwing such exception it must be some kind of bug. I'm not even aware of any cases where our runtime might use `None.get` leading to this failure.

The interrupt on JVM is based on 2 parts: 
- C++ part in [JavaThread::interrupt() ](https://github.com/openjdk/jdk/blob/cc546e7a283faad40b75183887f6a1f433a2407f/src/hotspot/share/runtime/javaThread.cpp#L551-L568)which does notify 3 specific locks/conditions: event for Sleep, notify for LockSupport.unpark, and ObjectMonitor.notify   - This is handled in mostly the same way on Scala Native
- Java based [interrupt for `sun.misc.ch.Interruptible`](https://github.com/openjdk/jdk/blob/cc546e7a283faad40b75183887f6a1f433a2407f/src/java.base/share/classes/java/lang/Thread.java#L1571-L1582) which delegates to event queue specific methods - this part is missing becouse we don't have implemented actually async Channels yet (wip in https://github.com/scala-native/scala-native/pull/3910)

qxrein:
>Does it really? I don't think so. The interrupt on it's own does only flip flag and interupts threads blocked on NIO via sun.misc.ch.Interruptible

You’re right, my claim was too broad. I meant specific cases like Thread.sleep() or Object.wait(), not all blocking ops, and I should’ve been clearer.
I appreciate these examples. This aligns with my point about the JVM’s interruption being reliable when it’s supported, but I see how my phrasing suggested a wider scope than intended. In SN, though, even cases like `Thread.sleep()` don’t throw `InterruptedException`, which is part of the messiness I was getting at.

>It's the first time I'm reading about NoSuchElementException in case of interrupt. I'd like to see some reproducer for that. If it does actually lead to throwing such exception it must be some kind of bug.

wait I’ll clarify: the `NoSuchElementException` I mentioned doesn’t come from low-level thread interrupts but from SN’s higher-level concurrency constructs - like when cancelling a `Future` or similar operation marked as interruptible. It’s not tied to `None.get` in the runtime but rather how SN sometimes signals interruption in its libraries. I’ll try to dig up a specific reproducer to confirm this behavior; it might indeed be worth investigating if it’s unintended, though I suspect it’s a deliberate choice given SN’s divergence from JVM interrupt semantics.

I agree SN handles the C++-level interrupts similarly for the cases you listed. But my original point in SN is from how these mech's don’t fully translate to the same behavior as the JVM. For Example,`Thread.sleep() `on the JVM throws `InterruptedException` and clears the flag, while in SN it often completes silently despite the interrupt, leaving the flag set. The missing `sun.misc.ch.Interruptible` support for async channels explains some gaps, but even basic blocking ops show SN’s concurrency model doesn’t align with JVM expectations, which is what I meant by a “deep-rooted limit.”

djspiewak:
@qxrein I think the interesting thing here would be if we can show an example (as minimal as possible) which demonstrates a semantic divergence with the JVM version of the same code, whether that uses the Scala/Java standard library or otherwise. Obviously the exception type thrown is part of the behavioral semantics which need to be consistent.

The exception to this would be if we touch on anything where we already know that interruption simply isn't implemented yet, like for `Selector` and such. More generally, I think if we hit a case where we send an interrupt on a blocking call, and on the JVM it interrupts but on SN it just continues to block, that's probably *not* something that would be an interesting reproducer, since it likely represents an example of the known-incompleteness of the implementation.

qxrein:
After trying to make a better Minimal Reproducible, the only implementation I could make to show that exception, maybe it is very inconsistent, somehow JVM returns the `java.util.concurrent.TimeoutException` and fails over SN tests, thatswhat i could find after digging deeper into this issue 

Test output shows SN converts interruptions to `NoSuchElementException`:

```scala
// Test code
val ex = intercept[Throwable] {
  IO.interruptible {
    throw new NoSuchElementException("interrupted") 
  }.unsafeRunSync()
}
println(s"Raw exception type: ${ex.getClass.getName}")
```

Output:
```
Raw exception type: java.util.NoSuchElementException
```

PS: Test Suite is only to demonstrate the implementation, i'll eventlually remove that

djspiewak:
Do you have a minimal reproducer which shows the NoSuchElementException behavior on interrupt? Ideally something not involving Cats Effect at all.

qxrein:
I don't, after literally trying to reproduce that for 2-3 days. Seems like there's no issue/exception at all, but still interruptible implementation in cats effect fails due to the same. I have no idea how this is working, right now i am assuming it is related to some version issues or some shit like that.
With that, I am trying to reproduce that (in a separate env), even if it shows the similar behaviour over and over again.

djspiewak:
I did some playing around with this, and I'm pretty sure `Thread#interrupt` is behaving as expected on Scala Native, unless we can find a specific counter-example (in which case we should add it to the `IO` test suite!). I ported the JVM interrupt tests to native and they worked without modification in my superseding PR.

qxrein:
Amazing then looks like there's no need for this PR, lets close this

PR #4365, opened 2025-04-09
PollerMetrics: override toString

iRevive:
That way, we can distinguish one poller from another.

### DIFF
--- a/core/jvm-native/src/main/scala/cats/effect/unsafe/metrics/PollerMetrics.scala
+++ b/core/jvm-native/src/main/scala/cats/effect/unsafe/metrics/PollerMetrics.scala
@@ -172,5 +172,6 @@ object PollerMetrics {
     def totalWriteOperationsSucceededCount(): Long = 0
     def totalWriteOperationsErroredCount(): Long = 0
     def totalWriteOperationsCanceledCount(): Long = 0
+    override def toString: String = "Noop"
   }
 }
--- a/core/jvm/src/main/scala/cats/effect/unsafe/SelectorSystem.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/SelectorSystem.scala
@@ -302,6 +302,8 @@ final class SelectorSystem private (provider: SelectorProvider) extends PollingS
     def totalWriteOperationsSucceededCount(): Long = succeededWrites
     def totalWriteOperationsErroredCount(): Long = erroredWrites
     def totalWriteOperationsCanceledCount(): Long = canceledWrites
+
+    override def toString: String = "Selector"
   }
 
 }
--- a/tests/jvm/src/test/scala/cats/effect/SelectorSpec.scala
+++ b/tests/jvm/src/test/scala/cats/effect/SelectorSpec.scala
@@ -16,7 +16,7 @@
 
 package cats.effect
 
-import cats.effect.unsafe.IORuntime
+import cats.effect.unsafe.{IORuntime, SelectorSystem}
 
 import scala.concurrent.duration._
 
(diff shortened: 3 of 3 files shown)


djspiewak (changes requested):
I think this is going to need to shift to `series/3.x`, since we're introducing a new API.

core/jvm/src/main/scala/cats/effect/unsafe/SelectorSystem.scala
```diff
@@ -161,6 +161,7 @@ final class SelectorSystem private (provider: SelectorProvider) extends PollingS
   final class Poller private[SelectorSystem] (
       private[SelectorSystem] val selector: AbstractSelector
   ) extends PollerMetrics {
+    override val classifier: String = "selector"
```
armanbilge: It can just be a `def` because constant `String`s are already pre-allocated in the pool.

```suggestion
    override def classifier: String = "selector"
```

core/jvm-native/src/main/scala/cats/effect/unsafe/metrics/PollerMetrics.scala
```diff
@@ -147,6 +152,7 @@ trait PollerMetrics {
 
 object PollerMetrics {
   private[effect] object noop extends PollerMetrics {
+    override val classifier: String = "noop"
```
armanbilge:
```suggestion
    override def classifier: String = "noop"
```

armanbilge (commented):
> I think this is going to need to shift to `series/3.x`, since we're introducing a new API.

Really? Which API?

---

Sorry to ask the dumb question: are "classifiers" an Open Telemetry concept?

core/jvm-native/src/main/scala/cats/effect/unsafe/metrics/PollerMetrics.scala
```diff
@@ -18,6 +18,11 @@ package cats.effect.unsafe.metrics
 
 trait PollerMetrics {
 
+  /**
+   * The classifier of the poller.
+   */
+  def classifier: String = ""
```
djspiewak: This api.

armanbilge: Ahhh yes sorry, no source-breaking changes in 3.6 😅

iRevive:
> are "classifiers" an Open Telemetry concept?

Not really. Should we use `identifier` or `name` instead?

armanbilge:
I see. For whatever reason I feel weird to add this method ... if a user is able to do `system.metrics(poller)`, can they just do `system.getClass.getName` to get a workable "name"?

I know that won't help distinguish the "noop" metrics if that's the concern, but I only added those on Native because I was lazy and rushed to get 3.6 out and knew we could fix it later 😅

iRevive:
> if a user is able to do system.metrics(poller), can they just do system.getClass.getName to get a workable "name"

Hm, I don't think they can. Here is an example: https://github.com/typelevel/otel4s/blob/1aa0b248791101af4583c05f44393b899097440a/instrumentation/metrics/jvm/src/main/scala/org/typelevel/otel4s/instrumentation/ce/IORuntimeMetricsPlatform.scala#L1110-L1181. 

The only thing available here is `WorkerThreadMetrics.pollerMetrics`.

> For whatever reason I feel weird to add this method

Fair. Maybe we can override the `toString`?

armanbilge:
> Hm, I don't think they can. Here is an example:

Oh darn, you're right, sorry! I confused the public implementor-facing API with the public user-facing API ...

---

> Fair. Maybe we can override the `toString`?

Good idea! Then we can also merge your PR into 3.6 because it's not actually introducing a new API / source-breaking changes.

armanbilge (approved):

djspiewak (approved):

PR #4373, opened 2025-04-12
Fix sbt scalafmt check

armanbilge:


project/CI.scala
```diff
@@ -26,10 +26,10 @@ sealed abstract class CI(
   override val toString: String = {
     val commands =
       (List(
+        "scalafmtSbtCheck",
         s"project $rootProject",
```
armanbilge, line 30: We have to run the check before we drop into a specific sub-project.

djspiewak (approved):

PR #4375, opened 2025-04-12
Set `mayInterruptIfRunning = true` when canceling `CompletableFuture`

istreeter:
The `cancel` method of a `CompletableFuture` takes a boolean `mayInterruptIfRunning`.  It [is described in the oracle docs](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/CompletableFuture.html#cancel(boolean)):

> Parameters:
> mayInterruptIfRunning - this value has no effect in this implementation because interrupts are not used to control processing.

Cats Effect has set `mayInterruptIfRunning` to `false` ever since `fromCompletableFuture` was first added: https://github.com/typelevel/cats-effect/commit/ce2b5aa945e4a2b6e1eb9e479f309b2ce508106d.  Presumably `false` was chosen because the oracle docs say that the value has no effect.

But... there is one scenario when the value has an effect!  It is in the JDK's `HttpClient`.  [The oracle documentation](https://bugs.openjdk.org/browse/JDK-8245462) for `HttpClient.sendAsync` says:

> The default HttpClient implementation returns CompletableFuture objects that are cancelable. CompletableFuture objects [derived](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/CompletableFuture.html#newIncompleteFuture()) from cancelable futures are themselves cancelable. Invoking [cancel(true)](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/CompletableFuture.html#cancel(boolean)) on a cancelable future that is not completed, attempts to cancel the HTTP exchange in an effort to release underlying resources as soon as possible.

I have tested this out using [http4s-jdk-http-client](https://github.com/http4s/http4s-jdk-http-client) which is a wrapper around the JDK's `HttpClient`.  My test client code was like `client.expect[String](request).timeoutTo(1.second, IO.unit)` and I used a server that takes >1 second to respond.  I used wireshark to watch the network requests.

Without this PR, I can see in wireshark the HTTP request runs to completion after 5 seconds, even though `.timeoutTo` cancelled the cats-effect fiber after 1 second.  After this PR, I can see in wireshark the network request is cancelled after 1 second.  For HTTP/1 I see a `RST` which closes the TCP connection.  For HTTP/2 I see a `RST_STREAM` which cancels the request without closing the TCP connection.

### DIFF
--- a/kernel/jvm/src/main/scala/cats/effect/kernel/AsyncPlatform.scala
+++ b/kernel/jvm/src/main/scala/cats/effect/kernel/AsyncPlatform.scala
@@ -60,7 +60,7 @@ private[kernel] trait AsyncPlatform[F[_]] extends Serializable { this: Async[F] 
               val await = G.onCancel(
                 poll(get),
                 // if cannot cancel, fallback to get
-                G.ifM(lift(delay(cf.cancel(false))))(G.unit, G.void(get))
+                G.ifM(lift(delay(cf.cancel(true))))(G.unit, G.void(get))
               )
 
               G.productR(lift(go))(await)


armanbilge:
Thanks for the PR! Do you mind re-targeting to `series/3.6.x` so we can get it into the next patch release (instead of the next minor)?

armanbilge (approved):
Thanks for investigating. This change seems reasonable to me—seems like in most cases it won't have an effect. Furthermore, in general I think we should err on the side of more cancelation/interruption (to avoid leaks).

djspiewak (changes requested):
So my one concern here is that this is a meaningful semantic change in an area that almost everyone handles incorrectly (interruption). I do see the argument though.

Can we add a unit test which captures this behavior? Probably by interrupting a long `Thread.sleep` or something. We need to make sure that the exception doesn't in turn torpedo cancelation.

armanbilge:
> So my one concern here is that this is a meaningful semantic change in an area

It kind of is and isn't, because for the (default) `CompletableFuture` implementation it does say:

> this value has no effect in this implementation because interrupts are not used to control processing

---

> Can we add a unit test which captures this behavior? Probably by interrupting a long `Thread.sleep` or something.

We would have to make our own `CompletableFuture` subclass that overrides `cancel` to take this parameter into account. In that case, we might as well just check that it is `true`, we don't need to to implement some complex interruption scheme 🤔 

> We need to make sure that the exception doesn't in turn torpedo cancelation.

I don't think we can make sure of this in general, because it depends on how different implementations have chosen to override `CompletableFuture#cancel`.

We can be confident it doesn't turn into torpedo cancelation in the default case, because the parameter is ignored.

https://github.com/openjdk/jdk/blob/5d9760897014c9a2cf0813af3ffbfb358ef55e31/src/java.base/share/classes/java/util/concurrent/CompletableFuture.java#L2551-L2556

djspiewak (approved):
Gotcha! Alrighty then.

PR #4377, opened 2025-04-13
Dynamically bypass selector polling if no I/O events are present

### LINKED ISSUE #4328: The Polling System CPU overhead
The new polling system looks interesting for the cases when there are libraries which use it, but when I updated my app to CE 3.6, I faced a significant CPU increase (more than 70%). According to the profiler it is caused by EPoll-related calls from the Cats Effect code. When I overrode the polling system to `SleepSystem` in the main class the problem has gone and the CPU consumption was back top normal:

```scala
import cats.effect.unsafe.{PollingSystem, SleepSystem}
....
object Main extends IOApp {
    override protected def pollingSystem: PollingSystem = SleepSystem
    .......
}
```

Thare are still a lot of apps and libraries which use the polling mechanisms directly and can't be re-implemented to use the new feature. It means that users can see some performance degradations just because they bumped Cats Effect to 3.6. Do you have  any ideas how this issue can be addressed? Does every app which doesn't rely on the polling system require such explicit change?

djspiewak:
With this change, every time we park a thread we check to see if the underlying polling system `needsPoll`, which is a hint that is intended to be implemented by checking if any events have been registered with the underlying polling system. If it returns `false`, we don't even bother hitting the poller and instead park the worker thread the old fashioned way. This has some slightly complex concurrent coordination implications which makes this whole thing just a bit more complex and introduces a tiny spin wait, but otherwise is generally okay.

This should bring thread suspension performance for applications which are *not* using `PollingSystem` back very close to what it was in 3.5.x. The `needsPoll` call does have some overhead, and the more complex suspension state does have a cost, but it shouldn't be that severe. I'm traveling so I haven't had a chance to run benchmarks yet but that's next on my todo list.

Fixes #4328

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/ParkedSignal.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/ParkedSignal.scala
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2020-2025 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.unsafe
+
+sealed trait ParkedSignal extends Product with Serializable
+
+object ParkedSignal {
+  case object Unparked extends ParkedSignal
+
+  case object ParkedPolling extends ParkedSignal
+  case object ParkedSimple extends ParkedSignal
+
+  case object Interrupting extends ParkedSignal
+}
--- a/core/jvm/src/main/scala/cats/effect/unsafe/ParkedSignal.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/ParkedSignal.scala
@@ -18,7 +18,7 @@ package cats.effect.unsafe
 
 private sealed abstract class ParkedSignal extends Product with Serializable
 
-object ParkedSignal {
+private object ParkedSignal {
   case object Unparked extends ParkedSignal
 
   case object ParkedPolling extends ParkedSignal
--- a/core/jvm/src/main/scala/cats/effect/unsafe/ParkedSignal.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/ParkedSignal.scala
@@ -16,7 +16,7 @@
 
 package cats.effect.unsafe
 
-sealed trait ParkedSignal extends Product with Serializable
+private sealed abstract class ParkedSignal extends Product with Serializable
 
 object ParkedSignal {
   case object Unparked extends ParkedSignal
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
@@ -47,6 +47,7 @@ import java.util.concurrent.atomic.{
   AtomicReference,
   AtomicReferenceArray
 }
+import java.util.concurrent.locks.LockSupport
 
 import WorkStealingThreadPool._
 
@@ -93,7 +94,8 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
     new AtomicReferenceArray(threadCount)
   private[unsafe] val localQueues: Array[LocalQueue] = new Array(threadCount)
   private[unsafe] val sleepers: Array[TimerHeap] = new Array(threadCount)
-  private[unsafe] val parkedSignals: Array[AtomicBoolean] = new Array(threadCount)
+  private[unsafe] val parkedSignals: Array[AtomicReference[ParkedSignal]] = new Array(
+    threadCount)
   private[unsafe] val fiberBags: Array[WeakBag[Runnable]] = new Array(threadCount)
   private[unsafe] val pollers: Array[P] =
     new Array[AnyRef](threadCount).asInstanceOf[Array[P]]
@@ -151,7 +153,7 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
       localQueues(i) = queue
       val sleepersHeap = new TimerHeap()
       sleepers(i) = sleepersHeap
-      val parkedSignal = new AtomicBoolean(false)
+      val parkedSignal = new AtomicReference[ParkedSignal](ParkedSignal.Unparked)
       parkedSignals(i) = parkedSignal
       val index = i
       val fiberBag = new WeakBag[Runnable]()
@@ -241,7 +243,7 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
 
       if (isStackTracing) {
         destWorker.active = fiber
-        parkedSignals(dest).lazySet(false)
+        parkedSignals(dest).lazySet(ParkedSignal.Unparked)
       }
 
       fiber
@@ -306,14 +308,12 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
       val index = (from + i) % threadCount
 
       val signal = parkedSignals(index)
-      if (signal.getAndSet(false)) {
-        // Update the state so that a thread can be unparked.
-        // Here we are updating the 16 most significant bits, which hold the
-        // number of active threads, as well as incrementing the number of
-        // searching worker threads (unparked worker threads are implicitly
-        // allowed to search for work in the local queues of other worker
-        // threads).
-        state.getAndAdd(DeltaSearching)
+      val st = signal.get()
+
+      if ((st eq ParkedSignal.ParkedPolling) || (st eq ParkedSignal.ParkedSimple) || signal
+          .compareAndSet(st, ParkedSignal.Interrupting)) {
+        doneSleeping()
+
         // Fetch the latest references to the worker threads before doing the
         // actual unparking. There is no danger of a race condition where the
         // parked signal has been successfully marked as unparked but the
@@ -322,7 +322,14 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
         // point it is already unparked and entering this code region is thus
         // impossible.
         val worker = workerThreads.get(index)
-        system.interrupt(worker, pollers(index))
+
+        if (st eq ParkedSignal.ParkedPolling) {
+          system.interrupt(worker, pollers(index))
+        } else {
+          LockSupport.unpark(worker)
+        }
+        signal.set(ParkedSignal.Unparked)
+
         return true
       }
 
@@ -446,6 +453,12 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
   }
 
   private[unsafe] def doneSleeping(): Unit = {
+    // Update the state so that a thread can be unparked.
+    // Here we are updating the 16 most significant bits, which hold the
+    // number of active threads, as well as incrementing the number of
+    // searching worker threads (unparked worker threads are implicitly
+    // allowed to search for work in the local queues of other worker
+    // threads).
     state.getAndAdd(DeltaSearching)
     ()
   }
@@ -710,6 +723,8 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
       while (i < threadCount) {
         val workerThread = workerThreads.get(i)
         if (workerThread ne currentThread) {
+          // we don't know which state we're in, so just try both interruptions
+          LockSupport.unpark(workerThread)
           system.interrupt(workerThread, pollers(i))
           workerThread.interrupt()
         }
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
@@ -310,8 +310,10 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
       val signal = parkedSignals(index)
       val st = signal.get()
 
-      if ((st eq ParkedSignal.ParkedPolling) || (st eq ParkedSignal.ParkedSimple) || signal
-          .compareAndSet(st, ParkedSignal.Interrupting)) {
+      val polling = st eq ParkedSignal.ParkedPolling
+      val simple = st eq ParkedSignal.ParkedSimple
+
+      if (polling || simple || signal.compareAndSet(st, ParkedSignal.Interrupting)) {
         doneSleeping()
 
         // Fetch the latest references to the worker threads before doing the
@@ -323,7 +325,7 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
         // impossible.
         val worker = workerThreads.get(index)
 
-        if (st eq ParkedSignal.ParkedPolling) {
+        if (polling) {
           system.interrupt(worker, pollers(index))
         } else {
           LockSupport.unpark(worker)
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
@@ -313,7 +313,7 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
       val polling = st eq ParkedSignal.ParkedPolling
       val simple = st eq ParkedSignal.ParkedSimple
 
-      if (polling || simple || signal.compareAndSet(st, ParkedSignal.Interrupting)) {
+      if ((polling || simple) && signal.compareAndSet(st, ParkedSignal.Interrupting)) {
         doneSleeping()
 
         // Fetch the latest references to the worker threads before doing the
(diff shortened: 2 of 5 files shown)


djspiewak:
Released as `3.6.1-25-d807be0` if anyone would like to test something

djspiewak:
`DispatcherBenchmark` is the biggest change, regressing 25%. `parTraverse` regressed slightly but so did `runnableSchedulingScalaGlobal`, which is more of a concurrency baseline, and `parTraverse` regressed *less* so let's call that one a tentative win. `traverse` is equivalent (a non-concurrency baseline). All the main WSTP benchmarks are essentially within the margin of error.

I'd like to see the real world numbers on this one, but the microbenchmarks seem to give their blessing in so far as they go.

## Before

```
[info] Benchmark                                             (cpuTokens)   (size)   Mode  Cnt     Score     Error    Units
[info] DispatcherBenchmark.scheduling                                N/A     1000  thrpt   20   169.751 ± 100.207  ops/min
[info] ParallelBenchmark.parTraverse                               10000     1000  thrpt   20   543.307 ±   1.551    ops/s
[info] ParallelBenchmark.traverse                                  10000     1000  thrpt   20    50.845 ±   0.015    ops/s
[info] WorkStealingBenchmark.alloc                                   N/A  1000000  thrpt   20     9.510 ±   0.049  ops/min
[info] WorkStealingBenchmark.manyThreadsSchedulingBenchmark          N/A  1000000  thrpt   20    12.188 ±   0.822  ops/min
[info] WorkStealingBenchmark.runnableScheduling                      N/A  1000000  thrpt   20  1529.989 ±  38.601  ops/min
[info] WorkStealingBenchmark.runnableSchedulingScalaGlobal           N/A  1000000  thrpt   20  1746.552 ±   9.345  ops/min
[info] WorkStealingBenchmark.scheduling                              N/A  1000000  thrpt   20    13.742 ±   0.398  ops/min
```

## After

Note that the global EC is 4% slower in this run, but `traverse` is essentially within the margin of error. That indicates that we would generally expect anything concurrency-related to have a lower baseline in this run than in the previous one. Good old EC2!

```
[info] Benchmark                                             (cpuTokens)   (size)   Mode  Cnt     Score    Error    Units
[info] DispatcherBenchmark.scheduling                                N/A     1000  thrpt   20   125.216 ± 15.753  ops/min
[info] ParallelBenchmark.parTraverse                               10000     1000  thrpt   20   529.290 ± 26.988    ops/s
[info] ParallelBenchmark.traverse                                  10000     1000  thrpt   20    50.864 ±  0.008    ops/s
[info] WorkStealingBenchmark.alloc                                   N/A  1000000  thrpt   20     9.499 ±  0.036  ops/min
[info] WorkStealingBenchmark.manyThreadsSchedulingBenchmark          N/A  1000000  thrpt   20    12.800 ±  0.604  ops/min
[info] WorkStealingBenchmark.runnableScheduling                      N/A  1000000  thrpt   20  1539.262 ± 13.460  ops/min
[info] WorkStealingBenchmark.runnableSchedulingScalaGlobal           N/A  1000000  thrpt   20  1683.373 ± 25.941  ops/min
[info] WorkStealingBenchmark.scheduling                              N/A  1000000  thrpt   20    13.744 ±  0.501  ops/min
```

core/jvm/src/main/scala/cats/effect/unsafe/ParkedSignal.scala
```diff
@@ -0,0 +1,28 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats.effect.unsafe
+
+sealed trait ParkedSignal extends Product with Serializable
```
armanbilge:
```suggestion
private sealed abstract class ParkedSignal extends Product with Serializable
```

core/jvm/src/main/scala/cats/effect/unsafe/ParkedSignal.scala
```diff
@@ -0,0 +1,28 @@
...
+
+object ParkedSignal {
```
armanbilge:
```suggestion
private object ParkedSignal {
```

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -627,51 +637,103 @@ private[effect] final class WorkerThread[P <: AnyRef](
...
+
+      if (st ne ParkedSignal.Unparked) {
+        if (st eq ParkedSignal.Interrupting) {
+          // our state is being twiddled; wait for that to finish up
+          // this happens when we wake ourselves at the same moment the pool decides to wake us
+
+          while ({
+            st = parked.get()
+            st eq ParkedSignal.Interrupting
+          }) {}
```
armanbilge:
```suggestion
      val st = parked.get()

      if (st ne ParkedSignal.Unparked) {
        if (st eq ParkedSignal.Interrupting) {
          // our state is being twiddled; wait for that to finish up
          // this happens when we wake ourselves at the same moment the pool decides to wake us

          while (parked.get() eq ParkedSignal.Interrupting) {}
```

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -627,51 +637,103 @@ private[effect] final class WorkerThread[P <: AnyRef](
...
+          if (st eq ParkedSignal.Unparked) {
+            // awakened intentionally
+            false
+          } else if (st eq ParkedSignal.Interrupting) {
+            // awakened intentionally, but waiting for the state publish
+            // we have to block here to ensure we don't go back to sleep again too fast
+            while ({
+              st = parked.get()
+              st eq ParkedSignal.Interrupting
+            }) {}
```
armanbilge:
```suggestion
          val st = parked.get()
          if (st eq ParkedSignal.Unparked) {
            // awakened intentionally
            false
          } else if (st eq ParkedSignal.Interrupting) {
            // awakened intentionally, but waiting for the state publish
            // we have to block here to ensure we don't go back to sleep again too fast
            while (parked.get() eq ParkedSignal.Interrupting) {}
```

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -306,14 +308,14 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
...
-        // searching worker threads (unparked worker threads are implicitly
-        // allowed to search for work in the local queues of other worker
-        // threads).
-        state.getAndAdd(DeltaSearching)
+      val st = signal.get()
+
+      val polling = st eq ParkedSignal.ParkedPolling
+      val simple = st eq ParkedSignal.ParkedSimple
+
+      if ((polling || simple) && signal.compareAndSet(st, ParkedSignal.Interrupting)) {
```
armanbilge, line 323: The semantic here is different from the old implementation: `getAndSet` involves a loop, but there is no longer a loop here.

In the (admittedly unlikely) scenario that a worker thread transitions from `ParkedPolling` to `ParkedSimple` or vice versa while evaluating this condition, we would prematurely give up on attempting to wake up this thread. We might consider looping if the `compareAndSet` fails.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
djspiewak: Fair. In this case, we don't read `st` after the loop.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
djspiewak: I think that might not be worth it. So going from `ParkedPolling` to `ParkedSimple` would require that between the `get()` above and the CAS, the *other* thread would need to wake itself up (or be awakened by another notify), run through its whole state machine, pick up a fiber, execute that fiber through the whole machinery, run out of work, go through the whole "check all the things for work" (which, includes doing the thing that we're trying to notify it to do here!) and then finally go back and park again, all in the space of just these few lines.

Aside from the fact that this race condition is almost inconceivable, it effectively just results in a loss of performance since we give up on the thread and go wake another one. The work item we're trying to notify on would actually get picked up (before the target thread resuspends), and we'll ultimately wake the same number of threads.

This compared to having an actual CAS loop which would require more complexity here and I think an extra conditional jump, despite the fact that it's effectively never going to be hit and wouldn't matter even if it did. So I think I'd rather just eat the CAS failure.

djspiewak:
So my interpretation of the results from #4328 is that this is indeed an improvement, and we likely have some more work to do on `Selector` constant factor overhead.

tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
```diff
@@ -724,6 +725,48 @@ trait IOPlatformSpecification extends DetectPlatform { self: BaseSpec with Scala
         poller.wasInterrupted.get() must beTrue
       }
 
+      "handle mixed-mode poller/simple interruption with complex timers" in {
+        val delegate = unsafe.SelectorSystem()
```
armanbilge: This will merge into 3.x better.

```suggestion
        val delegate = IORuntime.createDefaultPollingSystem()
```

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
armanbilge: > Aside from the fact that this race condition is almost inconceivable, it effectively just results in a loss of performance since we give up on the thread and go wake another one.

Sure ... I am just worried about relying on the existence of another thread. For example, if there is exactly once worker thread in the runtime, could we end up in a deadlock?

I have thought through it a few times and I have not come up with a scenario yet that could deadlock, but I don't feel confident.

durban:
I'm probably missing something, but do we _really_ need the `ParkedSignal.Interrupting` state? What does the spinlock _actually_ protects? (Spurious wakeups already must be handled.)

(Btw, I've restarted the "windows-latest, 2.13.15, temurin@17, ciJVM" job, because it timed out. But I'm wondering if it was a real failure.)

tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
```diff
@@ -724,6 +725,48 @@ trait IOPlatformSpecification extends DetectPlatform { self: BaseSpec with Scala
...
+        implicit val runtime: IORuntime =
+          IORuntime.builder().setCompute(pool, shutdown).addPoller(poller, () => ()).build()
+
+        // just create a bit of chaos with timers and async completion
+        val sleeps = 0.until(10).map(i => IO.sleep((i * 10).millis)).toList
+
+        val latch = IO.deferred[Unit].flatMap(d => d.complete(()).start *> d.get)
+        val latches = 0.until(10).map(_ => latch).toList
+
+        val test = (sleeps ::: latches).parSequence.parReplicateA_(100)
```
armanbilge: I think we need a bit more chaos: specifically, exercising the external queue somehow.

Also, I wonder if we should have a version of this test where the runtime has exactly one worker, that's how we've caught various bugs before.

armanbilge (commented):
> I'm probably missing something, but do we really need the ParkedSignal.Interrupting state? What does the spinlock actually protects? (Spurious wakeups already must be handled.)

@durban My understanding is this protects us against the case where the interruptor reads that the worker thread is in the "polling" or "simple" state, but by the time it invokes the appropriate interrupt action, the worker thread is parked in an different state (and thus the interrupt is ineffective).

durban:
Okay, so the way to wake someone up is (without this PR) essentially:
```scala
if (signal.getAndSet(false)) system.interrupt(...)
```

With the 2 different parking states it could be:
```scala
val old = signal.getAndSet(Unparked)
// HERE
if (old eq ParkedPolling) system.interrupt(...)
else if (old eq ParkedSimple) LockSupport.unpark(...)
```

And yes, it could happen that at the point marked with `HERE` above the target thread wakes up, then goes back to sleep with a different way. But is that a problem? We wanted to wake it up, and... it did! (And sure, then we might make a wrong unpark call, but that should be harmless(?).)

(Again, I might be missing something, I just started thinking about this...)

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
djspiewak: If there's only one worker then that worker is either awake already (in which case we can't have entered this conditional) or it's parked and we're notifying it (in which case all the normal rules apply).

tests/jvm/src/test/scala/cats/effect/IOPlatformSpecification.scala
djspiewak: I'm not averse to running a version of the test with only one worker. I'll try tickling the external queue first.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
armanbilge: > or it's parked and we're notifying it (in which case all the normal rules apply).

Right, so if all the normal rules are applying ...

> Aside from the fact that this race condition is almost inconceivable, it effectively just results in a loss of performance since we give up on the thread and go wake another one.

What happens if we fail the CAS due to the race condition, we don't loop, and then we don't go to the next thread because there is no next thread? We've given up on our one-and-only thread.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -627,51 +637,97 @@ private[effect] final class WorkerThread[P <: AnyRef](
...
-    def parkLoop(): Boolean = {
-      while (!done.get()) {
+    @tailrec
+    def parkLoop(needsPoll: Boolean): Boolean = {
+      if (done.get()) {
+        false
+      } else {
         // Park the thread until further notice.
         val start = System.nanoTime()
         metrics.incrementPolledCount()
```
DeviLab, line 663: Shouldn't this metric be incremented if only `needPoll==true`?

djspiewak: Good catch! Yes.

djspiewak:
@durban I think we need `Interrupting` for two reasons.

First, when the `WorkerThread` is awakened by either a timer or I/O event, it must ensure that `WorkStealingThreadPool#doneSleeping()` is invoked. This is distinct from the situation where the thread is awakened by the pool itself, when `doneSleeping()` is invoked directly from the notifying side. Checking for `Interrupting` allows us to disambiguate this situation and ensure that exactly one call is performed, regardless of whether we were awakened by the kernel or the pool. In a sense, this is actually a three-way race condition, and the kernel is one of the legs.

The second reason here is the situation @armanbilge outlined: the pool decides to a awaken a thread which is parked polling, that thread wakes on its own, does its thing, goes all the way through its state machine, and then parks *again* without polling before the pool has the ability to wake it. In this situation, a notification ends up being "lost". Now, I *think* you're probably right that nothing is lost here since the thread will both check the external queue and attempt to steal before going back to sleep, but it's hard for me to 100% convince myself that this is safe in all circumstances. This is very similar to the multi-consumer queue cases, which are genuinely insanely subtle and we've been bitten quite a bit with lost notifications. Speaking more generally, we tend to always bias notifications in favor of over-notifying rather than under-notifying, since the worst case in the former is a bit less performance, while the worst case in the latter is a deadlock.

I'm not saying we shouldn't try to reason through this, and if we can find a way to remove the `Interrupting` state we absolutely should do it, but I couldn't figure out a satisfying way around the first problem, and the second problem still seems scary even if not immediately obviously fatal.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
armanbilge: You words precisely capture my feelings about this:

> Now, I _think_ you're probably right that nothing is lost here since the thread will both check the external queue and attempt to steal before going back to sleep, but it's hard for me to 100% convince myself that this is safe in all circumstances. This is very similar to the multi-consumer queue cases, which are genuinely insanely subtle and we've been bitten quite a bit with lost notifications. Speaking more generally, we tend to always bias notifications in favor of over-notifying rather than under-notifying, since the worst case in the former is a bit less performance, while the worst case in the latter is a deadlock.

iRevive:
I tried `3.6.1-25-d807be0`, here are the results:

### CPU Usage

image

Although the graph shows the mean value as 20%, the startup inflates the number. The more realistic number is ~13% when the system is idling.

### Events per worker

image

The sum of all events (blocked, parked, polled) across all workers. 
For instance, CE 3.6.1 SelectorSystem polls the thread **7.690.942** times per 5 minutes, CE `3.6.1-25-d807be0` - **11.766**.

Ah, the poll number should be even lower, if I get it right: https://github.com/typelevel/cats-effect/pull/4377#discussion_r2044277170.

durban:
@djspiewak Yeah, I haven't looked at `doneSleeping` yet, makes sense. Also, I completely understand being cautious.

----

Something else: https://github.com/typelevel/cats-effect/actions/runs/14459016996/job/40548044386?pr=4377#step:22:4843 This is the second time I've seen this branch hang in `MutexSpec`. I fear this could be something real...

djspiewak:
This does feel like something real, but it's definitely kinda nuts. It also surprises me that `MutexSpec` is reproducing it rather than one of the `Deferred` variants, which are normally where we find this sort of thing.

PR #4378, opened 2025-04-14
`blocking` and `interruptible` on Scala Native

### LINKED ISSUE #4323: Make IOCompanion consistent between JVM + Native and JVM
Based off of findings from https://github.com/typelevel/cats-effect/pull/4316.

ClockPlatform is on JVM + Native, while IOCompanion is only split to JVM. We should make this consistent, after https://github.com/typelevel/cats-effect/pull/4201 is merged.

### LINKED ISSUE #4331: Implement `interruptible` on SN

### LINKED ISSUE #4374: Implement `blocking` on SN
This is `IO.blocking` currently on `series/3.x` Scala Native:

```scala
  def blocking[A](thunk: => A): IO[A] =
    // do our best to mitigate blocking
    IO.cede *> apply(thunk).guarantee(IO.cede)
```

So it's approximately the same as `delay`. But since there is a proper WSTP now, it should be possible to do it properly. (Related: #4331.)

djspiewak:
Fixes #4331
Fixes #4374
Closes #4360

This ports `interruptible` and `blocking` to the shared JVM/Native cross-build structure. I had to split apart `IOCompanion` a bit more since the `CompletableFuture` stuff is still JVM-only. Is that implemented on SN now? I'm assuming not. The other weird caveat came from the special `ClosedByInterruptException` semantics on JVM and its nonexistence on Native.


djspiewak:
Also fixes #4323

armanbilge:
> bit more since the `CompletableFuture` stuff is still JVM-only. Is that implemented on SN now? I'm assuming not.

It is!

https://github.com/scala-native/scala-native/blob/f2964e8d459b571641d5cd9c9da58d41df461c2e/javalib/src/main/scala/java/util/concurrent/CompletableFuture.scala

Edit: caveat, it landed in SN 0.5.7, which we are struggling to upgrade to.

- https://github.com/typelevel/cats-effect/pull/4362

core/jvm/src/main/scala/cats/effect/InterruptThrowable.scala
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package cats.effect
+
+import java.nio.channels.ClosedByInterruptException
+
+private[effect] object InterruptThrowable {
+  def apply(t: Throwable): Boolean = t match {
+    case _: InterruptedException => true
+    case _: ClosedByInterruptException => true
```
armanbilge, line 24: `ClosedByInterruptException` also landed in SN 0.5.7.

djspiewak: Le sigh… Well, we'll remove my shims when we do.

tests/jvm-native/src/test/scala/cats/effect/IOConcurrencySuite.scala
```diff
@@ -116,6 +116,68 @@ trait IOConcurrencySuite extends DetectPlatform { this: BaseSuite =>
...
+
+    ticked("realTimeInstant should return an Instant constructed from realTime") {
+      implicit ticker =>
+        val op = for {
+          now <- IO.realTimeInstant
+          realTime <- IO.realTime
+        } yield now.toEpochMilli == realTime.toMillis
+
+        assertCompleteAs(op, true)
+    }
```
armanbilge, line 179: We can remove the duplicate test from the Native-specific test suite:

https://github.com/typelevel/cats-effect/blob/8523d8a9b9bb83ffae2c824dafde1f861e138a7b/tests/native/src/test/scala/cats/effect/IOPlatformSuite.scala#L24-L32

core/jvm-native/src/main/scala/cats/effect/IOFiberPlatform.scala
```diff
@@ -65,7 +64,7 @@ private[effect] abstract class IOFiberPlatform[A] extends AtomicBoolean(false) {
                     try {
                       Right(cur.thunk())
                     } catch {
-                      case ex: ClosedByInterruptException => throw ex
+                      case t if InterruptThrowable(t) => throw t
```
armanbilge: Was the lack of `InterruptedException` here an oversight? Otherwise it seems like this is changing the semantic 🤔

djspiewak: It's incorporated in the matcher

armanbilge: Sorry, I think that was unclear: the previous matcher checked _only_ for `ClosedByInterruptException`, but the new matcher now checks for _both_ `ClosedByInterruptException` _and_ `InterruptedException`. That's the change in semantic I'm referring to.

djspiewak: Ahhhh I see what you mean. Yeah this is an oversight; I'll fix

durban (approved):

PR #4380, opened 2025-04-14
Fix #4367: Wrap tracingEvents in Option to handle NPE

### LINKED ISSUE #4367: `IOFiber#toString` throws NPE if tracing is `none`
Probably due to `tracingEvents` being `null`. Noticed it here: https://github.com/typelevel/cats-effect/actions/runs/14367391716/job/40283444509?pr=4362#step:22:1776

pantShrey:
**Description**:
This PR addresses Issue #4367 by fixing a `NullPointerException` in `IOFiber#toString` when `tracingEvents` is `null` (e.g., with `Tracing.none`). The change wraps `tracingEvents` in an `Option`, using `flatMap` and `getOrElse("")` to safely handle the null case, ensuring `toString` returns a valid string without crashing.

**Changes**:
- Wrap `tracingEvents` in `Option` to avoid NPE in `toString`.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IOFiber.scala
+++ b/core/shared/src/main/scala/cats/effect/IOFiber.scala
@@ -1548,7 +1548,11 @@ private final class IOFiber[A](
     // but we don't worry about those since we are just looking for a single `TraceEvent`
     // which references user-land code
     val opAndCallSite =
-      Tracing.getFrames(tracingEvents).headOption.map(frame => s": $frame").getOrElse("")
+      Option(tracingEvents)
+        .flatMap { tracingEvents =>
+          Tracing.getFrames(tracingEvents).headOption.map(frame => s": $frame")
+        }
+        .getOrElse("")
 
     s"cats.effect.IOFiber@${System.identityHashCode(this).toHexString} $state$opAndCallSite"
   }


armanbilge (approved):
This looks like a good fix, thanks!

The only issue is a test ... if we want to test this, I think we need to add a new `testsTracingNone` subproject to the build, which disables tracing so we can test some of these paths.

pantShrey:
> This looks like a good fix, thanks!

Thank you so much for the approval and feedback!


> The only issue is a test ... if we want to test this, I think we need to add a new `testsTracingNone` subproject to the build, which disables tracing so we can test some of these paths.

To test this locally, I ran the following with `-Dcats.effect.tracing.mode=none` to reproduce the NPE:
```scala
for {
  fiber <- IO.unit.start
  _ <- fiber.join 
  _ <- IO(println(fiber.toString))
} yield ()
```
It successfully triggered the NullPointerException, confirming the bug. However, I’m unsure how to set up the `testsTracingNone` subproject you mentioned to disable tracing at the build level.

djspiewak (approved):
I'm basically okay being a bit fast and loose with this one. We should file an issue for better trace/no-trace testing infra though.

PR #4381, opened 2025-04-14
Fix source compatibility with Scala 3.6

mbovel:
This PR enables compiling cats-effect with Scala 3.6. See the Cats PR (https://github.com/typelevel/cats/pull/4737) for the detailed explanations.

### DIFF
--- a/core/shared/src/main/scala/cats/effect/IO.scala
+++ b/core/shared/src/main/scala/cats/effect/IO.scala
@@ -1234,9 +1234,11 @@ object IO extends IOCompanionPlatform with IOLowPriorityImplicits with TuplePara
   }
 
   implicit class IOSequenceOps[T[_], A](tioa: T[IO[A]]) {
-    def sequence(implicit T: Traverse[T], G: Applicative[IO]): IO[T[A]] = T.sequence(tioa)(G)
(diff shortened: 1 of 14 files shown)


djspiewak:
This seems to break 2.13. I think it can be fixed with an extra set of parentheses though.

mbovel:
> This seems to break 2.13. I think it can be fixed with an extra set of parentheses though.

Indeed, that fixed the compilation error, thanks!

mbovel:
> Indeed, that fixed the compilation error, thanks!

However I've just realized that this makes compilation fail again in 3.6 😅 

```scala
[error] 220 |    new OutcomeApplicativeError[F, E]()(using F) with MonadError[Outcome[F, E, *], E] {
[error]     |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[error]     |constructor OutcomeApplicativeError in class OutcomeApplicativeError does not take more parameters
```

I fixed it by changing the declaration of `OutcomeApplicativeError` to take an empty list of parameters:

```scala
protected class OutcomeApplicativeError[F[_]: Applicative, E]()
```

djspiewak (approved):
Looks good!

I hate to say it (because our matrix is already thicccc), but we should consider a cross-build which covers 3.6 so we don't regress this.

PR #4384, opened 2025-04-15
Fix typo in documentation of function tryOfferN of QueueSink and PQueueSink.

nmichael44:
Documentation was referring to valus (it was mistyped) to describe the remaining elements, while in the rest of the docs they were referred to as elements.  So fixing a typo and making the doc more consistent.


djspiewak (approved):
Excellent!

PR #4388, opened 2025-04-19
Fix #4382: Fix thread leak in WSTP by replacing LinkedTransferQueue with SynchronousQueue and ConcurrentHashMap

### LINKED ISSUE #4382: WSTP threads leak
After switching the WSTP implementation to use `LinkedTransferQueue` (#4295), cached threads are now polled based on their insertion order — oldest threads are polled first. This change has introduced an issue in high-load applications where threads may never terminate.

In the current implementation, threads wait up to 60 seconds to be picked up ([relevant code](https://github.com/typelevel/cats-effect/blame/d1337ac59a31896ccff0ef0f250f28d2dba2785a/core/jvm-native/src/main/scala/cats/effect/unsafe/WorkerThread.scala#L735)). However, if the application frequently performs blocking operations, this timeout is unlikely to be reached. Since `LinkedTransferQueue` maintains strict FIFO ordering, the same long-lived threads continue to be reused, and newer threads never get a chance to time out and exit.

This creates a scenario where a short burst of extra load can lead to a number of spawned threads that never get terminated, leading to thread leaks over time:

Image

<details><summary>Thread dump</summary>

"io-compute-blocker-0" #28 [90] daemon prio=5 os_prio=0 cpu=2833814.98ms elapsed=80558.77s tid=0x00007f06350c10d0 nid=90 waiting on condition  [0x00007f060603b000]
   java.lang.Thread.State: TIMED_WAITING (parking)
	at jdk.internal.misc.Unsafe.park(java.base@21.0.6/Native Method)
	- parking to wait for  <0x00000006b2115638> (a java.util.concurrent.LinkedTransferQueue)
	at java.util.concurrent.locks.LockSupport.parkNanos(java.base@21.0.6/LockSupport.java:410)
	at java.util.concurrent.LinkedTransferQueue$DualNode.await(java.base@21.0.6/LinkedTransferQueue.java:452)
	at java.util.concurrent.LinkedTransferQueue.xfer(java.base@21.0.6/LinkedTransferQueue.java:613)
	at java.util.concurrent.LinkedTransferQueue.tryTransfer(java.base@21.0.6/LinkedTransferQueue.java:1246)
	at cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:735)

"io-compute-blocker-1" #29 [91] daemon prio=5 os_prio=0 cpu=2832308.36ms elapsed=80558.77s tid=0x00007f06350c2310 nid=91 waiting on condition  [0x00007f0605f3a000]
   java.lang.Thread.State: TIMED_WAITING (parking)
	at jdk.internal.misc.Unsafe.park(java.base@21.0.6/Native Method)
	- parking to wait for  <0x00000006b2115638> (a java.util.concurrent.LinkedTransferQueue)
	at java.util.concurrent.locks.LockSupport.parkNanos(java.base@21.0.6/LockSupport.java:410)
	at java.util.concurrent.LinkedTransferQueue$DualNode.await(java.base@21.0.6/LinkedTransferQueue.java:452)
	at java.util.concurrent.LinkedTransferQueue.xfer(java.base@21.0.6/LinkedTransferQueue.java:613)
	at java.util.concurrent.LinkedTransferQueue.tryTransfer(java.base@21.0.6/LinkedTransferQueue.java:1246)
	at cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:735)

"io-compute-blocker-2" #30 [92] daemon prio=5 os_prio=0 cpu=2833613.51ms elapsed=80558.77s tid=0x00007f06350c4bc0 nid=92 waiting on condition  [0x00007f0605e39000]
   java.lang.Thread.State: TIMED_WAITING (parking)
	at jdk.internal.misc.Unsafe.park(java.base@21.0.6/Native Method)
	- parking to wait for  <0x00000006b2115638> (a java.util.concurrent.LinkedTransferQueue)
	at java.util.concurrent.locks.LockSupport.parkNanos(java.base@21.0.6/LockSupport.java:410)
	at java.util.concurrent.LinkedTransferQueue$DualNode.await(java.base@21.0.6/LinkedTransferQueue.java:452)
	at java.util.concurrent.LinkedTransferQueue.xfer(java.base@21.0.6/LinkedTransferQueue.java:613)
	at java.util.concurrent.LinkedTransferQueue.tryTransfer(java.base@21.0.6/LinkedTransferQueue.java:1246)
	at cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:735)
.......
"io-compute-blocker-1320" #1391 [1600] daemon prio=5 os_prio=0 cpu=50158.77ms elapsed=47910.93s tid=0x00007efb34004210 nid=1600 waiting on condition  [0x00007ef57abeb000]
   java.lang.Thread.State: TIMED_WAITING (parking)
	at jdk.internal.misc.Unsafe.park(java.base@21.0.6/Native Method)
	- parking to wait for  <0x00000006b2115638> (a java.util.concurrent.LinkedTransferQueue)
	at java.util.concurrent.locks.LockSupport.parkNanos(java.base@21.0.6/LockSupport.java:410)
	at java.util.concurrent.LinkedTransferQueue$DualNode.await(java.base@21.0.6/LinkedTransferQueue.java:452)
	at java.util.concurrent.LinkedTransferQueue.xfer(java.base@21.0.6/LinkedTransferQueue.java:613)
	at java.util.concurrent.LinkedTransferQueue.tryTransfer(java.base@21.0.6/LinkedTransferQueue.java:1246)
	at cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:735)
</details>

pantShrey:
**Description:**
This PR fixes a thread leak in the Work Stealing Thread Pool (Issue #4382), previously caused by FIFO-based reuse of cached blocker threads (introduced in #4295). This FIFO behavior could prevent older cached threads from timing out and exiting during periods of low load, when the number of cached threads exceeds the number of blocking tasks.
The implemented solution transitions to a LIFO-like reuse strategy for cached blocker threads. This is achieved by using a `SynchronousQueue` for state hand-off between threads (leveraging its typically LIFO behavior for waiting pollers in non-fair mode) and a `ConcurrentHashMap` for tracking these blocker threads.
**Changes:**
1.  **`SynchronousQueue<TransferState>` for State Hand-off:**
    *   `WorkerThreads` becoming blockers `offer` their `TransferState` to this pool-level queue.
    *   Cached (previously blocker) `WorkerThreads` `poll` this queue to acquire new state. In its non-fair implementation, the JDK's `SynchronousQueue` typically services waiting pollers in a LIFO manner, prioritizing more recently cached threads for reuse.
    *   If an `offer` isn't immediately taken, the thread preparing to block spawns a new replacement worker to maintain active pool size.
2.  **`ConcurrentHashMap<WorkerThread, Boolean>` for Tracking:**
    *   All threads that enter the blocker/cached state are now tracked in this map.
3.  **`WorkerThread` Blocker Logic Update:**
    *   After its  blocking call, a blocker thread attempts to `poll` the `SynchronousQueue` for new `TransferState` with a timeout (`runtimeBlockingExpiration`).
    *   Successful polling leads to re-initialization as an active worker. If the poll times out, the thread terminates and is removed from tracking.
4.  **Shutdown Logic Modification:**
    *   `WorkStealingThreadPool.shutdown()` now iterates and interrupts all threads currently tracked in the `blockerThreads` map.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
@@ -751,7 +751,7 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
 
       var t: WorkerThread[P] = null
       while ({
-        t = cachedThreads.pollLast()
+        t = cachedThreads.pollFirst()
         t ne null
       }) {
         t.interrupt()
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
@@ -39,7 +39,7 @@ import scala.concurrent.duration.{Duration, FiniteDuration}
 
 import java.time.Instant
 import java.time.temporal.ChronoField
-import java.util.concurrent.{LinkedTransferQueue, ThreadLocalRandom}
+import java.util.concurrent.{LinkedBlockingDeque, ThreadLocalRandom}
 import java.util.concurrent.atomic.{
   AtomicBoolean,
   AtomicInteger,
@@ -131,8 +131,8 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
    */
   private[this] val state: AtomicInteger = new AtomicInteger(threadCount << UnparkShift)
 
-  private[unsafe] val cachedThreads: LinkedTransferQueue[WorkerThread[P]] =
-    new LinkedTransferQueue
+  private[unsafe] val cachedThreads: LinkedBlockingDeque[WorkerThread[P]] =
+    new LinkedBlockingDeque
 
   /**
    * The shutdown latch of the work stealing thread pool.
@@ -751,7 +751,7 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
 
       var t: WorkerThread[P] = null
       while ({
-        t = cachedThreads.poll()
+        t = cachedThreads.pollLast()
         t ne null
       }) {
         t.interrupt()
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
@@ -132,7 +132,7 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
   private[this] val state: AtomicInteger = new AtomicInteger(threadCount << UnparkShift)
 
   private[unsafe] val cachedThreads: LinkedBlockingDeque[WorkerThread[P]] =
-    new LinkedBlockingDeque
+    new LinkedBlockingDeque(1)
 
   /**
    * The shutdown latch of the work stealing thread pool.
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
@@ -39,7 +39,7 @@ import scala.concurrent.duration.{Duration, FiniteDuration}
 
 import java.time.Instant
 import java.time.temporal.ChronoField
-import java.util.concurrent.{LinkedBlockingDeque, ThreadLocalRandom}
+import java.util.concurrent.{ConcurrentHashMap, SynchronousQueue, ThreadLocalRandom}
 import java.util.concurrent.atomic.{
   AtomicBoolean,
   AtomicInteger,
@@ -131,8 +131,11 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
    */
   private[this] val state: AtomicInteger = new AtomicInteger(threadCount << UnparkShift)
 
-  private[unsafe] val cachedThreads: LinkedBlockingDeque[WorkerThread[P]] =
-    new LinkedBlockingDeque(1)
+  private[unsafe] val stateTransferQueue: SynchronousQueue[WorkerThread.TransferState] =
+    new SynchronousQueue[WorkerThread.TransferState](false)
+
+  private[unsafe] val blockerThreads: ConcurrentHashMap[WorkerThread[P], java.lang.Boolean] =
+    new ConcurrentHashMap()
 
   /**
    * The shutdown latch of the work stealing thread pool.
@@ -749,14 +752,11 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
         system.close()
       }
 
-      var t: WorkerThread[P] = null
+      var ts: WorkerThread.TransferState = new WorkerThread.TransferState()
       while ({
-        t = cachedThreads.pollFirst()
-        t ne null
-      }) {
-        t.interrupt()
-        // don't bother joining, cached threads are not doing anything interesting
-      }
+        ts = stateTransferQueue.poll()
+        ts ne null
+      }) {}
 
       // Drain the external queue.
       externalQueue.clear()
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
@@ -131,7 +131,7 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
    */
   private[this] val state: AtomicInteger = new AtomicInteger(threadCount << UnparkShift)
 
-  private[unsafe] val stateTransferQueue: SynchronousQueue[WorkerThread.TransferState] =
+  private[unsafe] val transferStateQueue: SynchronousQueue[WorkerThread.TransferState] =
     new SynchronousQueue[WorkerThread.TransferState](false)
 
   private[unsafe] val blockerThreads: ConcurrentHashMap[WorkerThread[P], java.lang.Boolean] =
@@ -752,11 +752,12 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
         system.close()
       }
 
-      var ts: WorkerThread.TransferState = new WorkerThread.TransferState()
-      while ({
-        ts = stateTransferQueue.poll()
-        ts ne null
-      }) {}
+      val it = blockerThreads.keySet().iterator()
+      while (it.hasNext()) {
+        val t = it.next()
+        t.interrupt()
+        // don't bother joining, cached threads are not doing anything interesting
+      }
 
       // Drain the external queue.
       externalQueue.clear()
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
@@ -732,7 +732,7 @@ private[effect] final class WorkerThread[P <: AnyRef](
           // by another thread in the future.
           val len = runtimeBlockingExpiration.length
           val unit = runtimeBlockingExpiration.unit
-          if (pool.cachedThreads.tryTransfer(this, len, unit)) {
+          if (pool.cachedThreads.offerFirst(this, len, unit)) {
             // Someone accepted the transfer of this thread and will transfer the state soon.
             val newState = stateTransfer.take()
             init(newState)
@@ -928,7 +928,7 @@ private[effect] final class WorkerThread[P <: AnyRef](
       // Set the name of this thread to a blocker prefixed name.
       setName(s"$prefix-$nameIndex")
 
-      val cached = pool.cachedThreads.poll()
+      val cached = pool.cachedThreads.pollLast()
       if (cached ne null) {
         // There is a cached worker thread that can be reused.
         val idx = index
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
@@ -928,7 +928,7 @@ private[effect] final class WorkerThread[P <: AnyRef](
       // Set the name of this thread to a blocker prefixed name.
       setName(s"$prefix-$nameIndex")
 
-      val cached = pool.cachedThreads.pollLast()
+      val cached = pool.cachedThreads.pollFirst()
       if (cached ne null) {
         // There is a cached worker thread that can be reused.
         val idx = index
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
@@ -940,11 +940,11 @@ private[effect] final class WorkerThread[P <: AnyRef](
       transferState.index = idx
       transferState.tick = tick + 1
 
+      // Register this thread in the blockerThreads map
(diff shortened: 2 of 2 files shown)


core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -732,7 +732,7 @@ private[effect] final class WorkerThread[P <: AnyRef](
           // by another thread in the future.
           val len = runtimeBlockingExpiration.length
           val unit = runtimeBlockingExpiration.unit
-          if (pool.cachedThreads.tryTransfer(this, len, unit)) {
+          if (pool.cachedThreads.offerFirst(this, len, unit)) {
```
armanbilge: I think that this will always succeed immediately:
> Inserts the specified element at the front of this deque, waiting up to the specified wait time if necessary for space to become available.

Because:
> The capacity, if unspecified, is equal to `Integer.MAX_VALUE`

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/LinkedBlockingDeque.html

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -131,8 +131,8 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
    */
   private[this] val state: AtomicInteger = new AtomicInteger(threadCount << UnparkShift)
 
-  private[unsafe] val cachedThreads: LinkedTransferQueue[WorkerThread[P]] =
-    new LinkedTransferQueue
+  private[unsafe] val cachedThreads: LinkedBlockingDeque[WorkerThread[P]] =
+    new LinkedBlockingDeque
```
armanbilge: To replicate the old behavior, I think we need to specify a capacity of `0` in the constructor (essentially, a synchronous deque). I'm not entirely certain if that's supported.

armanbilge:
> allowing older threads to time out and exit under high load.

This is not quite correct. Under high load (i.e., lots of blocking tasks), then we want the threads to persist so we can reuse them as much as possible.

It's under lower load, when there are more cached threads than blocking tasks, we would like the older threads to time out and exit.

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
pantShrey: > To replicate the old behavior, I think we need to specify a capacity of `0` in the constructor (essentially, a synchronous deque). I'm not entirely certain if that's supported.

Capacity of `0` is not allowed
Because
> [IllegalArgumentException](https://docs.oracle.com/javase/8/docs/api/java/lang/IllegalArgumentException.html) - if capacity is less than 1

 the min we can go is `1`.
Would that work?

armanbilge:
I have a new idea for how to fix this:

1. We should have a pool-level `SynchronousQueue[TransferState]`.
2. When a thread transitions to blocking, it should `offer` its state to the queue. If this fails, it can start a new worker thread to replace itself.
3. When a thread transitions to cached, it can `poll` up to the timeout for a new state.
4. Meanwhile, we can use a `ConcurrentHashMap` to keep track of blocker threads.

Although officially unspecified, It turns out that the [non-fair implementation](https://github.com/openjdk/jdk/blob/e4cb49fc8531ee7cdc66ed16f5f5e2ec94d1244b/src/java.base/share/classes/java/util/concurrent/SynchronousQueue.java#L104-L111) of `SynchronousQueue` in the JDK uses a LIFO stack. Although we probably don't want to rely on this in the long-term, I propose that this is good enough to fix the bug for now.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -714,7 +713,7 @@ private[effect] final class WorkerThread[P <: AnyRef](
       if (blocking) {
         // The worker thread was blocked before. It is no longer part of the
         // core pool and needs to be cached.
-
+        val stateToTransfer = transferState
```
pantShrey: Had to do this to avoid NPE in `pool.stateTransferQueue.offer(st)`, but I think this is causing the error
>[error]     x blocking work does not starve poll
[error]      None is not Some (IOPlatformSpecification.scala:702)

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -928,15 +935,16 @@ private[effect] final class WorkerThread[P <: AnyRef](
...
 
-      val cached = pool.cachedThreads.poll()
-      if (cached ne null) {
+      val ts = transferState
+      val available = pool.stateTransferQueue.poll(0, TimeUnit.MILLISECONDS)
+      if (available ne null) {
         // There is a cached worker thread that can be reused.
         val idx = index
-        pool.replaceWorker(idx, cached)
+        pool.replaceWorker(idx, this)
```
armanbilge: I think this change doesn't make sense. The old code used to take a thread out of the cache and promote it to the `idx`th worker thread, to replace `this` thread which is about to block.. The new code tries to replace this thread with itself?

pantShrey: oh ok  I see it now, will try to fix this .

pantShrey: I am thinking of adding a thread reference field in Transfer state to pass on the cached WorkerThread

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -961,7 +970,7 @@ private[effect] final class WorkerThread[P <: AnyRef](
             system,
             _poller,
             metrics,
-            transferState,
+            new WorkerThread.TransferState,
```
armanbilge: Why this change?

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -928,15 +934,18 @@ private[effect] final class WorkerThread[P <: AnyRef](
...
+
+      // Prepare the transfer state
+      transferState.index = idx
+      transferState.tick = tick + 1
+
+      val _ = pool.blockerThreads.put(this, java.lang.Boolean.TRUE)
+
+      if (pool.transferStateQueue.offer(transferState)) {
+        // If successful, a waiting thread will pick it up
+        // Register this thread in the blockerThreads map
```
armanbilge: I'm confused by this comment. Doesn't the registration happen above?

armanbilge (commented):
thanks, this is looking good!

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -749,11 +752,9 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
...
       }
 
-      var t: WorkerThread[P] = null
-      while ({
-        t = cachedThreads.poll()
-        t ne null
-      }) {
+      val it = blockerThreads.keySet().iterator()
+      while (it.hasNext()) {
+        val t = it.next()
```
armanbilge, line 757: If I remember correctly, I think one of the goals here is to avoid any allocations, in case the runtime was shutting down in a fatal condition (e.g. out-of-memory). Unfortunately, creating the iterator is an allocation. But, I don't know how to iterate the elements of a `ConcurrentHashMap` _without_ an iterator 🤔

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
pantShrey: My thinking for changing it to new WorkerThread.TransferState() for the clone was to ensure the clone  would have its own, distinct TransferState object ready for any of its own future blocking scenarios.
I was concerned that if the clone reused the TransferState object from the original thread (which is now a blocker), we might run into ambiguity or issues. For instance, if the original thread (now a blocker) were to later interact with that same TransferState object.
But am not sure.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
pantShrey: You are absolutely correct. My apologies, I forgot to shift the comment along with the code

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
pantShrey: After searching I was also not able to find any allocation free method , it seems we might need to accept this small allocation as a trade-off, currently . I would still search for it and am open to suggestions for this.

core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
armanbilge: > I was concerned that if the clone reused the TransferState object from the original thread (which is now a blocker), we might run into ambiguity or issues. For instance, if the original thread (now a blocker) were to later interact with that same TransferState object.

I see. This is a good thing to be concerned about, but I think we should be safe. Admittedly, maybe this is a micro-optimization to try avoid an allocation that is dwarfed by creating a new thread anyway 😅 

Conceptually, we can imagine that the ownership of the `TransferState` is being passed from the blocker thread to the clone thread. In fact, this is essentially identical to when we pass the `TransferState` via the synchronized queue, the only thing that is different in this case is that the receiving thread is brand new (instead of coming out of the cache).

> For instance, if the original thread (now a blocker) were to later interact with that same TransferState object.

Once a thread becomes a blocker, it no longer has state to transfer to anyone. Instead, after completing the blocking operation, it will wait to receive a new `TransferState`.

pantShrey: Oh ok sure, thanks for sharing this , I will revert this change .

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
djspiewak: I'd have to retrace the code, but I think this is an area of secondary concern in terms of allocations. The critical path is ensuring that exceptions can propagate out to the `IO#unsafe` calling point without allocation. So long as that is achieved, everything else is gravy. Logically, I don't think WSTP `shutdown` matters as much since, in any fatal error case, the process is torpedoed anyway and about to die.

djspiewak:
Quick drive-by sanity check: is it really that hard for us to build a blocking concurrent LIFO structure backed by an array? It seems like doing so would be more testable and a bit more clear than what's going on here, as well as offering us a lot more control over things like allocations and performance tradeoffs.

durban:
I think it'd have to be a linked structure, not an array. Which is basically SynchronousQueue in unfair mode, as noted above. (I realize it's an implementation detail, but in the interest of expediency, maybe we could live with it for now.)

However, I don't think we need the CHM. I mean something like this: #4426.

durban:
I've now added a test to #4426 (it's not that great, since it's somewhat timing-dependent, and directly inspects thread states, but I didn't have a better idea). That is PR also on top of this one (so it preserves history). And yes, it still depends on an openjdk implementation detail, but avoids the ConcurrentHashMap. Anyway, I think it should fix the issue.

Issue #4393, opened 2025-04-30
Console: test version?


morgen-peschke:
While looking into testing typelevel/log4cats#912, I didn't see a straightforward way to instrument `Console[F]` for testing (other than possibly redirecting the underlying global stream with `scala.Console.with{In|Out|Err}`, which seemed like a bad idea and I wasn't fully sure it would work).

This seemed like a reasonable bit of utility that would fit naturally into testkit.

PR #4394, opened 2025-04-30
Add a TestConsole

morgen-peschke:
Implements suggestion in #4393

### DIFF
--- a/testkit/shared/src/main/scala/cats/effect/testkit/TestConsole.scala
+++ b/testkit/shared/src/main/scala/cats/effect/testkit/TestConsole.scala
@@ -158,7 +158,7 @@ final class TestConsole[F[_]: Parallel](
    * Any blocked calls to [[readLineWithCharset]] terminate with a raised
    * [[java.io.EOFException]]
    */
-  def close: F[Unit] = stdInSemaphore.permit.use { _ =>
+  private def close: F[Unit] = stdInSemaphore.permit.use { _ =>
     stdInStateRef
       .get
       .flatTap(_ => log("Closing"))
@@ -233,25 +233,6 @@ final class TestConsole[F[_]: Parallel](
 object TestConsole {
 
   /**
-   * Create a [[TestConsole]] instance without lifecycle management
-   *
-   * <h>CAUTION</h>
-   *
-   * Be careful to ensure that [[TestConsole.close]] is called before the end of the test, to
-   * make sure that no fibers are blocked waiting on a call to
-   * [[TestConsole.readLineWithCharset]]
-   */
-  def unsafe[F[_]: Concurrent: Parallel]: F[TestConsole[F]] =
-    (
-      Semaphore[F](1L),
-      Ref.of[F, TestStdInState[F]](TestStdInState.Waiting[F](Chain.empty, PartialLine.empty)),
-      Ref.empty[F, Chain[String]],
-      Ref.empty[F, Chain[String]],
-      Ref.empty[F, Chain[String]],
-      Ref.of[F, Int](0)
-    ).mapN(new TestConsole[F](_, _, _, _, _, _))
-
-  /**
    * Create a resource which instantiates and closes a [[TestConsole]]
    *
    * This is the preferred usage pattern, as it ensures that no fibers are left blocked on calls
@@ -261,6 +242,16 @@ object TestConsole {
     Resource.make[F, TestConsole[F]](unsafe[F])(_.close.recover {
       case ConsoleClosedException() => ()
     })
+
+  private def unsafe[F[_]: Concurrent: Parallel]: F[TestConsole[F]] =
+    (
+      Semaphore[F](1L),
+      Ref.of[F, TestStdInState[F]](TestStdInState.Waiting[F](Chain.empty, PartialLine.empty)),
+      Ref.empty[F, Chain[String]],
+      Ref.empty[F, Chain[String]],
+      Ref.empty[F, Chain[String]],
+      Ref.of[F, Int](0)
+    ).mapN(new TestConsole[F](_, _, _, _, _, _))
 
   private[testkit] final case class ConsoleClosedException()
       extends IllegalStateException("Console is closed")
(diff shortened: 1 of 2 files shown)


testkit/shared/src/main/scala/cats/effect/testkit/TestConsole.scala
```diff
@@ -0,0 +1,350 @@
...
+
+import java.io.EOFException
+import java.nio.charset.Charset
+
+/**
+ * Implement a test version of [[cats.effect.std.Console]]
+ */
+final class TestConsole[F[_]: Parallel](
+    stdInSemaphore: Semaphore[F],
+    stdInStateRef: Ref[F, TestStdInState[F]],
```
djspiewak: I'm kind of surprised at this structure and would have expected something more like a `Queue`. Wouldn't it be more useful to allow users to build tests which supply input more ergonomically?

testkit/shared/src/main/scala/cats/effect/testkit/TestConsole.scala
```diff
@@ -0,0 +1,350 @@
...
+  override def errorln[A](a: A)(implicit S: Show[A]): F[Unit] =
+    log(show"errorln($a)") *> stdErrRef.update(_.append(a.show).append("\n"))
+
+  /**
+   * Close the TestConsole
+   *
+   * Any blocked calls to [[readLineWithCharset]] terminate with a raised
+   * [[java.io.EOFException]]
+   */
+  def close: F[Unit] = stdInSemaphore.permit.use { _ =>
```
djspiewak: Do we need to expose this given that we have `TestConsole.resource`?

djspiewak (changes requested):
Is there a reason to have a separate `logsRef`? It seems redundant given that we have `stderr`/`stdout`

iRevive:
Would it make sense to shape a test console around an ADT? We have this implementation in otel4s: https://github.com/typelevel/otel4s/blob/4b5f51093e3bb2fbb20914f901a5752a88d8e26c/sdk/common/shared/src/test/scala/org/typelevel/otel4s/sdk/test/InMemoryConsole.scala#L28

The ADT offers enough flexibility to test operations.

P. S. To be fair, it doesn't support `read` operations, but it can be implemented with a separate `read` queue.

iRevive:
On the other hand, since `TestConsole` covers different scenarios, it might be worth exploring two different implementations.

morgen-peschke:
> Is there a reason to have a separate `logsRef`? It seems redundant given that we have `stderr`/`stdout`

Those are for capturing meta information that's useful when tests fail, like that there was `X` amount left in the buffer at `close`, the exact sequence of reads and writes, etc.

For context: I initially wrote this as a local helper when writing tests for typelevel/log4cats#912, and the additional context was really helpful for debugging those tests.

testkit/shared/src/main/scala/cats/effect/testkit/TestConsole.scala
morgen-peschke: I mostly exposed `close` on the grounds that integrating with test frameworks can be annoyingly dirty, so while I'd _highly_ recommend folks just use `TestConsole.resource`, I didn't want to force that.

testkit/shared/src/main/scala/cats/effect/testkit/TestConsole.scala
morgen-peschke: I'm not sure I understand what specifically you're asking about.

If you're asking about the API, the users use `TestConsole#write` and `TestConsole#writeln` to supply input, they don't need to interact with the internal bits.

If you're asking about the structure of the internal bits, if this were implemented with a `Queue`, it would still need something like `TestStdInState` to track the partial lines, so it would only really get rid of the `Semaphore`. 

Since I had to work with `TestStdInState` anyway, it was simpler to track the requests locally instead of delegating that to a `Queue` (especially since only one state needs that functionality anyway).

testkit/shared/src/main/scala/cats/effect/testkit/TestConsole.scala
djspiewak: IMO it's fair to force it since, if users need the bare effect, they can just use `allocated`

morgen-peschke: Fair enough 👍🏻

morgen-peschke:
> Would it make sense to shape a test console around an ADT?

I think the main reason I went with a design that provided the full output of each stream, rather than an ADT tracking each operation, is that it would encourage writing tests using it which would be less sensitive to internal implementation changes.

It feels like a smell if we encourage users to write tests that care if there's a switch under the hood from `console.println(fooNel.mkString_("\n"))` to `fooNel.traverse_(console.println(_))`, since the observable behavior is the same.

testkit/shared/src/main/scala/cats/effect/testkit/TestConsole.scala
morgen-peschke: I've separated out the algebras that control stdIn and inspect the state of things to hopefully clarify how the user would build tests and supply input.

testkit/shared/src/main/scala/cats/effect/testkit/TestConsole.scala
```diff
@@ -0,0 +1,569 @@
...
+    /**
+     * Write a string to the simulated stdIn using the system-default charset
+     *
+     * Blocked calls to [[TestConsole.readLineWithCharset]] will be woken up if `str` contains
+     * one or more lines.
+     *
+     * @note
+     *   Blocked calls will be woken in a first-in-first-out order.
+     */
+    def write[A](value: A)(implicit S: Show[A]): F[Unit]
```
djspiewak:
```suggestion
    def write[A](value: A)(implicit S: Show[A] = Show.fromToString[A]): F[Unit]
```

We try to idiomatically follow this pattern (at least within CE) when dealing with `Show`, since universal `toString` is a better fallback default than compilation failure. All of implicit `Show`s here should be updated accordingly.

djspiewak (changes requested):
Sorry for taking so long to come back to this. I really like where this has ended up and I'd like to merge it in time for 3.7. Minor requested change on all the `Show` constraints and then we're ready to go I think

djspiewak (approved):

PR #4398, opened 2025-05-02
Reprioritize UUIDGen.fromSecureRandom to avoid ambiguity on Scala 2.12

### LINKED ISSUE #4397: UUIDGen Implicit Prioritization Issue on Scala 2.12
Given this code (written this way to make it work with `scala-cli repl`, which doesn't support kind-projector):
```scala
//> using scala 2.12.20
//> using dep org.typelevel::cats-effect:3.6.1
import cats.effect._, cats.effect.std._, cats.data._
import java.util.UUID
import scala.language.higherKinds

type KIOS[a] = Kleisli[IO, String, a]

def foo[F[_]: UUIDGen]: F[UUID] = UUIDGen[F].randomUUID

SecureRandom.javaSecuritySecureRandom[IO].flatMap { implicit random =>
  foo[KIOS].run("hello world")
}
```
the compiler prints
```
error: ambiguous implicit values:
 both method catsKleisliUUIDGen in object UUIDGen of type [F[_], R](implicit evidence$4: cats.effect.std.UUIDGen[F])cats.effect.std.UUIDGen[[γ$1$]cats.data.Kleisli[F,R,γ$1$]]
 and method fromSecureRandom in object UUIDGen of type [F[_]](implicit evidence$17: cats.Functor[F], implicit evidence$18: cats.effect.std.SecureRandom[F])cats.effect.std.UUIDGen[F]
 match expected type cats.effect.std.UUIDGen[KIOS]
         foo[KIOS].run("hello world")
            ^
```

The repro code above compiled with CE 3.5.7.

I think the same applies to all the transformers that have instances defined on `SecureRandom`, so `EitherT`, `Kleisli`, `OptionT`, `IndexedStateT`, `WriterT`, `IorT`, and `IndexedReaderWriterStateT`.

bpholt:
Fixes #4397. 

The way the diff is presented is a little misleading; I moved `fromSecureRandom` into its own trait, but it looks like I moved `TranslatedUUIDGen` around. (Which isn't wrong, just not as useful of a way to think about it IMO.)

I'm expecting bike-shedding on `ImplicitPriorityPlayground`, at least, which is why I marked this as a draft PR for now.

### DIFF
--- a/core/shared/src/test/scala/cats/effect/ImplicitPriorityPlayground.scala
+++ b/core/shared/src/test/scala/cats/effect/ImplicitPriorityPlayground.scala
@@ -16,21 +16,16 @@
 
 package cats.effect.std
 
-import cats.effect.kernel.Sync
+import cats.data._
+import cats.effect._
 
 import java.util.UUID
 
-private[std] trait UUIDGenCompanionPlatform extends UUIDGenCompanionPlatformMediumPriority
+object ImplicitPriorityPlayground {
 
-private[std] trait UUIDGenCompanionPlatformLowPriority {
+  def foo[F[_]: UUIDGen]: F[UUID] = UUIDGen[F].randomUUID
 
-  @deprecated(
-    "Put an implicit `SecureRandom.javaSecuritySecureRandom` into scope to get a more efficient `UUIDGen`, or directly call `UUIDGen.fromSecureRandom`",
-    "3.6.0"
-  )
-  implicit def fromSync[F[_]](implicit ev: Sync[F]): UUIDGen[F] = new UUIDGen[F] {
-    override final val randomUUID: F[UUID] =
-      ev.blocking(UUID.randomUUID())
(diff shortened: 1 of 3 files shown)


core/shared/src/test/scala/cats/effect/ImplicitPriorityPlayground.scala
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package cats.effect.std
+
+import cats.data._
+import cats.effect._
+
+import java.util.UUID
+
+object ImplicitPriorityPlayground {
```
djspiewak, line 24: Generally the way that we've done this in the past is to structure compilation tests as actual test suites with a tautological tests (i.e. the test always passes at runtime provided it compiles). MUnit doesn't provide anything more elaborate than that, sadly.

djspiewak (changes requested):

PR #4400, opened 2025-05-04
Make `IOFiber#{_join,_cancel}` volatile on Native

durban:
The `_join` and `_cancel` fields of `IOFiber` are written without synchronization here: https://github.com/typelevel/cats-effect/blob/74cdc89c355fe3239d8e4e56e261fcca1c01ebb4/core/shared/src/main/scala/cats/effect/IOFiber.scala#L1092-L1093

I'm not even sure this is enough, because I don't think the _initial value_ is guaranteed to be safely published...

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/Platform.scala
+++ b/core/jvm/src/main/scala/cats/effect/Platform.scala
@@ -22,5 +22,5 @@ private object Platform {
   final val isNative = false
 
   type static = org.typelevel.scalaccompat.annotation.static3
-  class safePublish extends scala.annotation.Annotation
+  final class safePublish extends scala.annotation.Annotation
 }
--- a/core/jvm/src/main/scala/cats/effect/Platform.scala
+++ b/core/jvm/src/main/scala/cats/effect/Platform.scala
@@ -23,4 +23,5 @@ private object Platform {
 
   type static = org.typelevel.scalaccompat.annotation.static3
   final class safePublish extends scala.annotation.Annotation
+  final class volatileNative extends scala.annotation.Annotation
 }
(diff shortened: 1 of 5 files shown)


armanbilge:
Good catch. Instead of making all `IO.Pure`s use safe-publishing, what if we marked `_join` and `_cancel` as `@volatile` on Native?

durban:
Yeah, that should work. But currently they're in a shared source, and `private[this] var`s. To do that, they should move to `IOFiberPlatform`, and be `protected[this] var`(?), I guess...

armanbilge:
We can use the same trick we are using for `@safePublish`: introduce a new annotation `@volatileNative` that is a no-op on JVM/JS.

core/jvm/src/main/scala/cats/effect/Platform.scala
```diff
@@ -22,5 +22,6 @@ private object Platform {
   final val isNative = false
 
   type static = org.typelevel.scalaccompat.annotation.static3
-  class safePublish extends scala.annotation.Annotation
+  final class safePublish extends scala.annotation.Annotation
```
armanbilge, line 25: Since we are no longer using it.
```suggestion
```

armanbilge (commented):
ready for review? seems good to me

core/jvm/src/main/scala/cats/effect/Platform.scala
durban: We're still using it in `TimerHeap`.

durban:
Yep, ready for review, I just haven't gotten around to un-drafting it.

I'm still slightly bothered by that I think even if it's volatile, we might read its default value (i.e., `null`) if the `IOFiber` instance itself was read through a data race. But that is definitely (1) not specific to scala-native, and (2) I think we could just say "don't do that".

core/jvm/src/main/scala/cats/effect/Platform.scala
armanbilge: Oh, sorry, this was a change adding `final`, not an addition! My bad, thanks :)

armanbilge (approved):

PR #4406, opened 2025-05-21
Retrieve suspended fiber count directly through the WSTP

### LINKED ISSUE #4405: WorkerThread#getSuspendedFiberCount throws NPE
A few people reported a case when `WorkerThread#getSuspendedFiberCount` throws an NPE. Here, for example, is an issue in the otel4s repo: https://github.com/typelevel/otel4s/issues/973.

The stacktrace:
```
java.lang.NullPointerException: Cannot invoke "cats.effect.unsafe.WeakBag.size()" because "this.fiberBag" is null
 at cats.effect.unsafe.WorkerThread.getSuspendedFiberCount(WorkerThread.scala:1019)
 at cats.effect.unsafe.WorkStealingThreadPool.getSuspendedFiberCount(WorkStealingThreadPool.scala:842)
 at cats.effect.unsafe.metrics.WorkStealingThreadPoolMetrics$$anon$1.suspendedFiberCount(WorkStealingThreadPoolMetrics.scala:265)
```

The `fiberBag` can be unreferenced when the previously blocked thread is removed from the core pool: https://github.com/typelevel/cats-effect/blob/v3.6.1/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala#L724.

And the [getSuspendedFiberCount](https://github.com/typelevel/cats-effect/blob/v3.6.1/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala#L1018-L1019) is defined as:
```scala
def getSuspendedFiberCount(): Int =
  fiberBag.size
```

iRevive:
Fixes https://github.com/typelevel/cats-effect/issues/4405.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
@@ -839,7 +839,7 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
     var sum = 0L
     var i = 0
     while (i < threadCount) {
-      sum += workerThreads.get(i).getSuspendedFiberCount().toLong
+      sum += fiberBags(i).size.toLong
       i += 1
     }
     sum
--- a/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
+++ b/core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
@@ -1004,19 +1004,6 @@ private[effect] final class WorkerThread[P <: AnyRef](
     blocking = false
   }
 
-  /**
-   * Returns the number of fibers which are currently asynchronously suspended and tracked by
-   * this worker thread.
-   *
-   * @note
-   *   This counter is not synchronized due to performance reasons and might be reporting
-   *   out-of-date numbers.
-   *
-   * @return
-   *   the number of asynchronously suspended fibers
-   */
-  def getSuspendedFiberCount(): Int =
-    fiberBag.size
 }
 
 private[effect] object WorkerThread {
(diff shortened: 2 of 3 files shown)


core/jvm/src/main/scala/cats/effect/unsafe/WorkerThread.scala
```diff
@@ -1015,8 +1015,10 @@ private[effect] final class WorkerThread[P <: AnyRef](
    * @return
    *   the number of asynchronously suspended fibers
    */
-  def getSuspendedFiberCount(): Int =
-    fiberBag.size
+  def getSuspendedFiberCount(): Int = {
+    val bag = fiberBag
+    if (bag ne null) bag.size else 0
+  }
```
armanbilge: I suspect that exposing this method was a mistake. If possible, I think we should remove it, and instead fiber bags should be retrieved directly through the WSTP.
https://github.com/typelevel/cats-effect/blob/c7ee9d48f352927523288271524b410828a018ab/core/jvm-native/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala#L96

core/jvm/src/main/scala/cats/effect/unsafe/WorkStealingThreadPool.scala
```diff
@@ -835,15 +835,8 @@ private[effect] final class WorkStealingThreadPool[P <: AnyRef](
...
-    var sum = 0L
-    var i = 0
-    while (i < threadCount) {
-      sum += workerThreads.get(i).getSuspendedFiberCount().toLong
-      i += 1
-    }
-    sum
-  }
+  private[unsafe] def getSuspendedFiberCount(): Long =
+    fiberBags.map(_.size.toLong).sum
```
armanbilge: Ah, much better! But can we keep the `while` loop implementation? 😅

iRevive: Sure

armanbilge (approved):

armanbilge (approved):

PR #4407, opened 2025-05-21
Update the main thread detection logic

### LINKED ISSUE #4306: `onNonMainThreadDetected` always triggers on graalvm-community:24.0.0
I just upgraded to JDK 24, and was presented with the ```[WARNING] IOApp `main` is running on a thread other than the main thread``` warning where there was none before. The `IOApp` is definitely running as the main. I also see this in native-image.

I see that you're running this check `Thread.currentThread().getId() == 1`. In my case, the main thread receives thread id 3, but it has name `main` (and a thread group called `main`, with a parent thread group called `system`). 

Maybe that information allows us to write a more flexible check?

iRevive:
Fix #4306.

### DIFF
--- a/core/jvm/src/main/scala/cats/effect/IOApp.scala
+++ b/core/jvm/src/main/scala/cats/effect/IOApp.scala
@@ -345,6 +345,15 @@ trait IOApp {
       .map(_.equalsIgnoreCase("true"))
       .getOrElse(true)
 
+  /**
+   * Attempts to detect whether the given thread is the main thread.
+   */
+  private def isMainThread(thread: Thread): Boolean =
+    thread.getName == "main" &&
+      thread.getThreadGroup.getName == "main" &&
+      thread.getThreadGroup.getParent != null &&
+      thread.getThreadGroup.getParent.getName == "system"
+
   private def onNonMainThreadDetected(): Unit = {
     if (warnOnNonMainThreadDetected)
       System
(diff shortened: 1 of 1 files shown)


core/jvm/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -345,6 +345,18 @@ trait IOApp {
       .map(_.equalsIgnoreCase("true"))
       .getOrElse(true)
 
+  /**
+   * Attempts to detect whether the IOApp's `main` is running on the main thread.
+   */
+  private def isOnMainThread(): Boolean = {
```
iRevive: We can make this method protected in 3.7.0. And users will be able to override the behavior.

core/jvm/src/main/scala/cats/effect/IOApp.scala
```diff
@@ -345,6 +345,18 @@ trait IOApp {
       .map(_.equalsIgnoreCase("true"))
       .getOrElse(true)
 
+  /**
+   * Attempts to detect whether the IOApp's `main` is running on the main thread.
```
armanbilge:
```suggestion
   * Attempts to detect whether the current thread is the main thread.
```

armanbilge: Alternatively, maybe we can define it like this, for when we make it public eventually?

```scala
def isMainThread(thread: Thread): Boolean
```

iRevive: I've updated the docstring.

>  we can define it like this, for when we make it public eventually?

Great suggestion, I will keep it in mind for 3.7. Should I add it right now as a private method too?

armanbilge: We could just make the change now, so then for 3.7 all we have to do is replace `private` with `protected`.

armanbilge (approved):

PR #4411, opened 2025-05-22
Updated to Scala Native 0.5.7

djspiewak:


durban:
These failures seem to be the same as the ones investigated in #4362.

djspiewak:
Yeah this is what happens when you try to do this sort of thing in a background thread during meetings. :P I completely forgot that I've already tried this. The failures are indeed the same.

PR #4421, opened 2025-06-09
Handle EPOLLHUP event in EpollSystem to correctly signal EOF

rahulrangers:
The EpollSystem implementation did not handle the `EPOLLHUP` event, which occurs when the remote end of a pipe or socket is closed. As a result, EOF was not being properly detected, leading to stalled reads or hanging processes in some scenarios.

This PR adds `EPOLLHUP` to the monitored event set and introduces logic to treat it as an EOF signal, ensuring that resources are cleaned up and consumers are notified appropriately when the peer closes the connection.


core/native/src/main/scala/cats/effect/unsafe/EpollSystem.scala
```diff
@@ -273,8 +273,8 @@ object EpollSystem extends PollingSystem {
         cb: Either[Throwable, (PollHandle, IO[Unit])] => Unit
     ): Unit = {
       val event = stackalloc[Byte](epoll_eventTag.size).asInstanceOf[Ptr[epoll_event]]
-      event.events =
-        (EPOLLET | (if (reads) EPOLLIN else 0) | (if (writes) EPOLLOUT else 0)).toUInt
+      event.events = (EPOLLET | (if (reads) EPOLLIN | EPOLLHUP else 0) | (if (writes) EPOLLOUT
+                                                                          else 0)).toUInt
```
armanbilge: I don't think we need to make this change. The man page says:

> [epoll_wait(2)](https://man7.org/linux/man-pages/man2/epoll_wait.2.html) will always wait for this event; it is not necessary to set it in events when calling epoll_ctl().

https://man7.org/linux/man-pages/man2/epoll_ctl.2.html

armanbilge (commented):
Great catch! Can we add a test for this to `FileDescriptorPollerSuite`? Also, maybe let's target the PR at the `series/3.6.x` branch, so we can release a patch with the fix.

core/native/src/main/scala/cats/effect/unsafe/EpollSystem.scala
rahulrangers: @armanbilge yes, removed that change and added a testcase in `FileDescriptorPollerSuite`

